PR #1894, opened 2022-08-05
Add possibility to see the last logged in person in group under projects

sjaanus:
Simple update, that also sends the `last_seen` column to frontend.


nunogois (approved):
Wuu nice, this might make our last seen accessor work correctly for groups as well 👍

PR #1896, opened 2022-08-08
Deleting project does not delete entry group_role table

sjaanus:
1. Added a foreign key, so project is hard-linked to projects table.


sighphyre (approved):
LGTM

PR #1897, opened 2022-08-08
fix: return 400 on invalid JSON instead of 404

olav:


thomasheartman (approved):
Looks great to  me! Maybe we could've done the same in the proxy? I'm not familiar enough with express to know 🙈

gastonfournier (approved):

PR #1898, opened 2022-08-09
Remove full outer joins

sjaanus:
After looking into full outer joins more, we still had some empty rows coming in because of this.

I think there is no benefit of doing full outer joins, replaced everything with left join.


gastonfournier (approved):
I think it makes sense, I wonder if we have some tests that can validate that this works as expected

PR #1904, opened 2022-08-10
Docs: Update compatibility matrix for php

RikudouSage:
## About the changes

Updated the compatibility table for php, `namePrefix` and `tags` filters and IP CIDR syntax were added some time ago, impression data just today (not released yet but will be during today, more specifically after https://github.com/Unleash/unleash-client-php/pull/104 gets merged).

## Discussion points

- Will the global segments be added to the compatibility table? If so, should it be added as part of this PR or a separate one?

### DIFF
--- a/website/docs/advanced/impression-data.md
+++ b/website/docs/advanced/impression-data.md
@@ -3,7 +3,7 @@ title: Impression data
 ---
 
 :::info Availability
-The impression data feature was introduced in **Unleash 4.7**. Listening for events requires [an SDK that supports impression data events](../sdks/index.md#server-side-sdk-compatibility-table). Currently, it's only supported in the [Unleash Proxy client](../sdks/proxy-javascript.md) and [React Proxy client](../sdks/proxy-react.md).
+The impression data feature was introduced in **Unleash 4.7**. Listening for events requires [an SDK that supports impression data events](../sdks/index.md#server-side-sdk-compatibility-table).
 :::
 
 Unleash can provide you with **impression data** about the toggles in your application. Impression data contains information about a specific feature toggle activation check: The client SDK will emit an **impression event** whenever it calls `isEnabled` or `getVariant`.
@@ -79,7 +79,7 @@ You can enable impression data via the impression data toggle in the admin UI's 
 ## Example setup
 
 :::caution
-This functionality is currently only supported in the [Unleash Proxy client](../sdks/proxy-javascript.md) and [React Proxy client](../sdks/proxy-react.md).
+This functionality is currently only supported in some of our SDKs, please see the [compatibility table for SDKs](../sdks/index.md#server-side-sdk-compatibility-table).
 :::
 
 The exact setup will vary depending on your [client SDK](../sdks/index.md). The below example configures the [Unleash Proxy client](/sdks/proxy-javascript) to listen for impression events and log them to the console. If "my-feature-toggle" is configured to emit impression data, then it will trigger an impression event as soon as Unleash is ready.
--- a/website/docs/advanced/impression-data.md
+++ b/website/docs/advanced/impression-data.md
@@ -3,7 +3,7 @@ title: Impression data
 ---
 
 :::info Availability
-The impression data feature was introduced in **Unleash 4.7**. Listening for events requires [an SDK that supports impression data events](../sdks/index.md#server-side-sdk-compatibility-table).
+The impression data feature was introduced in **Unleash 4.7**. Listening for events requires [an SDK that supports impression data events](../sdks/index.md#server-side-sdk-compatibility-table). In addition to supported server side SDKs impression data are available in JavaScript-based proxy clients.
 :::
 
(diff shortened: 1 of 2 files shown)


sighphyre:
@RikudouSage I don't think we have a plan to add segments here at the moment, the way they were designed is that segments are mapped to constraints on the backend unless the backend receives the client spec header and that header is greater than v4.2.0. So they should just work out the box on any SDK that supports constraints. The work on global segments is an optimisation for users that may have very, very large segments, and that's mostly to reduce overhead due to packet size and parsing the response

website/docs/sdks/index.md
```diff
@@ -98,7 +98,7 @@ If you see an item marked with a ❌ that you would find useful, feel free to re
 | Can disable metrics | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | Client registration | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | Basic usage metrics (yes/no) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
-| [Impression data](../advanced/impression-data.md) | ⭕ | ⭕ | ⭕ | ⭕ | ⭕ | ⭕ | ⭕ | ⭕ | N/A |
+| [Impression data](../advanced/impression-data.md) | ⭕ | ⭕ | ⭕ | ⭕ | ⭕ | ⭕ | ✅ | ⭕ | N/A |
```
sighphyre, line 101: I believe this was added to the node SDK as well, by this PR: https://github.com/Unleash/unleash-client-node/pull/293

```suggestion
| [Impression data](../advanced/impression-data.md) | ⭕ | ✅  | ⭕ | ⭕ | ⭕ | ⭕ | ✅ | ⭕ | N/A |
```

sighphyre (commented):
Looks good to me, but I'd like @thomasheartman to do the final approval here, since he's our resident Word Smith

website/docs/sdks/index.md
RikudouSage: I think it would be out of scope, so maybe creating an issue to update it or another PR?

thomasheartman: @sighphyre From what I can tell, the PR only adds it as a field that they expect to get from the server. It doesn't look like they do anything with it, so I don't _think_ the node client supports impression data. I might be wrong, though.

Regardless, I agree with @RikudouSage that that should be a different PR in that case.

website/docs/advanced/impression-data.md
```diff
@@ -3,7 +3,7 @@ title: Impression data
 ---
 
 :::info Availability
-The impression data feature was introduced in **Unleash 4.7**. Listening for events requires [an SDK that supports impression data events](../sdks/index.md#server-side-sdk-compatibility-table). Currently, it's only supported in the [Unleash Proxy client](../sdks/proxy-javascript.md) and [React Proxy client](../sdks/proxy-react.md).
+The impression data feature was introduced in **Unleash 4.7**. Listening for events requires [an SDK that supports impression data events](../sdks/index.md#server-side-sdk-compatibility-table).
```
thomasheartman, line 6: Good catch! Because we still don't have a compatibility table for frontend clients, I think we should still mention which frontend SDKs support it here. Something like this, perhaps?

```suggestion
The impression data feature was introduced in **Unleash 4.7**. Listening for events requires [an SDK that supports impression data events](../sdks/index.md#server-side-sdk-compatibility-table). In addition to server-side SDKs that support it, it's also supported in the JavaScript-based proxy clients.
```

website/docs/advanced/impression-data.md
```diff
@@ -79,7 +79,7 @@ You can enable impression data via the impression data toggle in the admin UI's
 ## Example setup
 
 :::caution
-This functionality is currently only supported in the [Unleash Proxy client](../sdks/proxy-javascript.md) and [React Proxy client](../sdks/proxy-react.md).
+This functionality is currently only supported in some of our SDKs, please see the [compatibility table for SDKs](../sdks/index.md#server-side-sdk-compatibility-table).
```
thomasheartman, line 82: Same thing as above: I think we should keep mention of the proxy clients in the note.

```suggestion
This functionality is currently only supported the JavaScript-based proxy clients and some of some of the server-side clients. Please refer to the [compatibility table for SDKs](../sdks/index.md#server-side-sdk-compatibility-table).
```

thomasheartman (changes requested):
Super! Thanks for updating the docs 🙌🏼  I've got a couple minor suggestions because we don't actually have a proxy client compatibility table just yet. Otherwise, it's all good 😄

website/docs/sdks/index.md
sighphyre: Fair enough! Consider it ignored

website/docs/advanced/impression-data.md
RikudouSage: I reworded it a little bit, let me know if it's ok.

website/docs/advanced/impression-data.md
RikudouSage: Added.

thomasheartman (approved):
Nice one! Thanks for the changes 🙌🏼

PR #1906, opened 2022-08-10
Docs: remove extra warning from impression data doc

thomasheartman:
This PR provides a little bit of polish/maintenance/upkeep on the impression data reference doc:

I don't think we need the extra caution under the "example setup" section, so I've removed that.

I've also applied some formatting and added some minor rewording changes.


sighphyre (approved):
Nice! Looks good

PR #1907, opened 2022-08-10
docs: add OpenAPI spec to Unleash docs

thomasheartman:
> Heads up! The 15.000 lines of code is mostly the website/yarn.lock file that hasn't been committed before! 😅

This PR prepares the Unleash for the following changes:

Add a new OpenAPI subheader to the APIs sidebar item. The OpenAPI category contains the auto-generated API docs based on the OpenAPI schemas that we've built into Unleash.

The PR uses [Palo Alto Networks' docusaurus-openapi-docs package](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs) for the doc generation.

As part of the API generation, I've also changed how we use tags with OpenAPI. Instead of having _one massive_ `admin` group, we now have many more, but much smaller tag groups, such as `tags`, `features`, `users`, etc.

NOTE: the PR does not actually activate this section of the doc just yet. When discussed with @olav and @sighphyre, we landed on not activating this until we could get the automation right. Because of upcoming changes to our repo structure, it was decided to postpone that specific part of the change. However, we can still merge most of these changes to prepare for that.

## Background

As of today, we have _some_ API docs on docs.getunleash.io, but they're all handwritten and hard to keep up to date. Now that we've done all the work to write OpenAPI schemas for our controllers, let's take advantage of that work and include the generated work in our documentation.

## Changes

Visit the files tab for a more detailed overview, but a rough explanation is:

### Version control changes

This is the largest change by line number: it removes `website/yarn.lock` from the .gitignore file and adds it as a staged file.

Files:

-   `.gitignore`
-   `website/yarn.lock`

### Tag changes

The most significant change from previous versions: we add a whole host of new tags, and apply them to existing endpoints.

Files:

-   `src/lib/openapi/util/api-operation.ts`
-   `src/lib/openapi/util/api-operation.ts`
-   everything under `src/lib/routes/`

This does also change the OpenAPI schema we produce and may change how clients are generated. However, it does _not_ break any existing integrations as the endpoints and operations are stable.

#### Capitalization

The tags are currently capitalized, but I think we could leave them all lowercase and instead use their `x-displayName` property to capitalize the name for display purposes. Open to discuss this.

### Add new API docs

In doing so, reference the new docs from the old docs too.

Files:

-   `website/docs/api/index.md`
-   `website/docusaurus.config.js`
-   `website/package.json`
-   `website/sidebars.js`
-   `website/src/css/custom.css` (add the operation buttons)

### Cleanup

I came across a CSS property that wasn't as well-supported as I'd thought, so I changed it to its older, better supported variant.

-   `website/src/components/UserFeedback/styles.module.css`

## Result

When integrated and rendered, it should look a little something like this:

image


website/docs/api/index.md
```diff
@@ -22,10 +22,11 @@ The internal API used by the Admin UI (unleash-frontend). Since v4.0.0 all opera
 
 With versions earlier than v4.0.0 and `insecure` authentication Basic Auth (with curl `-u myemail@test.com:`) is enough.
 
-- [Feature Toggles API](/api/admin/features)
-- [Strategies API](/api/admin/strategies)
 - [Events API](/api/admin/events)
+- [Feature Toggles API](/api/admin/features)
 - [Metrics API](/api/admin/metrics)
+- [Project API](/api/admin/project.md)
+- [Strategies API](/api/admin/strategies)
```
thomasheartman, line 29: These have been sorted and I've added the Projects API link.

website/docusaurus.config.js
```diff
@@ -179,5 +181,28 @@ module.exports = {
...
+        //                     : './openapi-spec.generated.json',
+        //                 outputDir: 'docs/reference/apis/unleash',
+        //                 sidebarOptions: {
+        //                     groupPathsBy: 'tag',
+        //                     categoryLinkSource: 'tag',
+        //                 },
+        //             },
+        //         },
+        //     },
+        // ],
```
thomasheartman, line 205: This generates the OpenAPI docs from a source.

website/sidebars.js
```diff
@@ -172,7 +172,21 @@ module.exports = {
...
+                        //     type: 'category',
+                        //     link: {
+                        //         title: 'Unleash Server APIs',
+                        //         type: 'generated-index',
+                        //         description:
+                        //             'Generated API docs based on the Unleash OpenAPI schema. For the time being, some additional info can also be found in the older API docs.',
+                        //         slug: '/reference/apis/unleash',
+                        //     },
+                        //     items: require('./docs/reference/apis/unleash/sidebar.js'),
+                        // },
```
thomasheartman, line 189: This adds OpenAPI docs to the sidebar.

src/lib/routes/admin-api/config.ts
```diff
@@ -55,7 +55,7 @@ class ConfigController extends Controller {
             permission: NONE,
             middleware: [
                 openApiService.validPath({
-                    tags: ['admin'],
+                    tags: ['Admin UI'],
```
olav, line 58: I wonder how the space and capitalization affects the client API generator, which uses tag names in class names. Hopefully it just works. (Like `AdminApi` from the current `admin` tag.)

website/docs/api/index.md
olav: "Project API" is the only one with `.md`, though not sure if it matters.

website/sidebars.js
olav: Nice, is this everything we have to comment out? The previous docs will still be there?

olav (changes requested):

src/lib/openapi/util/api-operation.ts
```diff
@@ -1,6 +1,89 @@
...
+    {
+        name: 'Tags',
+        description:
+            'Create, update, and delete [tags and tag types](https://docs.getunleash.io/advanced/tags).',
+    },
+    { name: 'Users', description: 'Manage users and passwords.' },
+] as const;
+
+// make the export mutable, so it can be used in a schema
+export const OpenApiTags = [...OPENAPI_TAGS].sort((a, b) =>
```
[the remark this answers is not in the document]
thomasheartman: Oh, right, you're right! Just an oversight on my end.

src/lib/openapi/util/api-operation.ts
```diff
@@ -1,6 +1,89 @@
...
+    a.name.localeCompare(b.name),
+);
+
+type OpenAPIOperationType = typeof OPENAPI_TAGS[number]['name'];
```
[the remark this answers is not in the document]
thomasheartman: Yes! Leftover from previous version

src/lib/routes/admin-api/config.ts
thomasheartman: Yeah, generators do support it (at least the ones I've tried), but I'd be happy to make it admin-ui or something instead 💁🏼

thomasheartman:
@olav The `.md` extension doesn't matter. It just means it references the file instead of the URL. Functionally the same. But now that you mention it, adding the .md extension is probably better for the rest.

And yeah, that's all we need to comment out. I haven't removed any docs yet, so the old API docs are still there. The generated ones aren't ready to take over from them just yet.

src/lib/openapi/util/api-operation.ts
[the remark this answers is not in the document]
thomasheartman: Addressed in 54d24dec6911688a53568de84929d9675e150e48

src/lib/openapi/util/api-operation.ts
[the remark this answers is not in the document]
thomasheartman: Addressed in 54d24dec6911688a53568de84929d9675e150e48

olav (approved):

PR #1917, opened 2022-08-12
Wrong number on the project page regarding members in that project

sjaanus:
Co-Authored by @nunogois .

This PR fixes the number on project page. Also it adds members in group to the count.


nunogois (approved):
LGTM, but it might be a good idea to have someone more experienced with this have a look as well before merging.

sighphyre (commented):
Code looks good but I have a domain question here - this looks like it changes the results for the default project by excluding the root role. Historically, the default project includes all users though a hidden editor role. I'm okay with changing that but I think we need some inputs from the UX folks before we do. WDYT?

sjaanus:
Added new logic and now members are counted from these rules:

1. All the users and groups that are seen on access tab under project
2. All the global editors

Global admins/viewers are not included.

sighphyre (approved):
LGTM!

PR #1926, opened 2022-08-16
feat: embed proxy endpoints

olav:
https://linear.app/unleash/issue/1-154/embedded-proxy-add-apifrontend-endpoints


src/lib/proxy/create-context.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+test('will not blow up if properties is an array', () => {
+    const context = createContext({
+        userId: '123',
+        tenantId: 'some-tenant',
+        region: 'eu',
+        properties: ['some'],
+    });
+
+    // console.log(context);
```
FredrikOseberg, line 58: removable

src/lib/proxy/create-context.ts
```diff
@@ -0,0 +1,34 @@
...
+        environment,
+        userId,
+        sessionId,
+        remoteAddress,
+        properties: Object.assign({}, rest, properties),
+    };
+
+    // Clean undefined properties on the context
+    const cleanContext = Object.keys(context)
+        .filter((k) => context[k])
```
FredrikOseberg, line 30: Could we use full parameter names?

FredrikOseberg (changes requested):

src/lib/proxy/create-context.ts
olav: See above.

src/lib/proxy/create-context.test.ts
FredrikOseberg: We should have that as a discussion point.

PR #1927, opened 2022-08-16
docs: add more standard errors

gastonfournier:
## About the changes
Add a standard error used when the user has not enough privileges to execute an operation


sighphyre (approved):
LGTM

Issue #1875, opened 2022-08-02
🟢  feat: frontend-api (embedded proxy)


ivarconr:
### Describe the feature request

When I am learning Unleash, and for small teams, it is really cumbersome to have to host a separate Unleash Proxy in order to connect frontend clients. By providing an embedded proxy, we take away some friction for the simple Unleash use-cases. 

### Background

* Should be optional to enable the embedded proxy. 
* Must be possible to enable "per environment". 
* Need to have a way to define "proxy public secrets" per env. 
* Must be possible to select projects to expose via the proxy.

ivarconr:
This feature reached GA in v4.18.

PR #1928, opened 2022-08-16
refactor: batch feature metric inserts

olav:
https://linear.app/unleash/issue/1-163/embedded-proxy-add-batching-for-metric-inserts

### DIFF
--- a/src/lib/services/client-metrics/metrics-service-v2.ts
+++ b/src/lib/services/client-metrics/metrics-service-v2.ts
@@ -8,16 +8,19 @@ import {
     IClientMetricsStoreV2,
 } from '../../types/stores/client-metrics-store-v2';
 import { clientMetricsSchema } from './schema';
-import { hoursToMilliseconds, minutesToMilliseconds } from 'date-fns';
+import { hoursToMilliseconds, secondsToMilliseconds } from 'date-fns';
 import { IFeatureToggleStore } from '../../types/stores/feature-toggle-store';
 import EventEmitter from 'events';
 import { CLIENT_METRICS } from '../../types/events';
 import ApiUser from '../../types/api-user';
 import { ALL } from '../../types/models/api-token';
 import User from '../../types/user';
+import { collapseClientMetrics } from '../../util/collapseClientMetrics';
 
 export default class ClientMetricsServiceV2 {
-    private timer: NodeJS.Timeout;
+    private timers: NodeJS.Timeout[] = [];
+
+    private unsavedMetrics: IClientMetricsEnv[] = [];
 
     private clientMetricsStoreV2: IClientMetricsStoreV2;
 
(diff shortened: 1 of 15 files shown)


FredrikOseberg (approved):
LGTM

src/lib/util/collapseHourlyMetrics.ts
```diff
@@ -0,0 +1,41 @@
...
+        metric.timestamp.getTime(),
+    ].join();
+};
+
+export const collapseHourlyMetrics = (
+    metrics: IClientMetricsEnv[],
+): IClientMetricsEnv[] => {
+    const hourlyMetrics = metrics.map((metric) => ({
+        ...metric,
+        timestamp: startOfHour(metric.timestamp),
```
gastonfournier, line 33: Here I believe you might be overriding the `timestamp` of the original metric. If it's intended that's fine, otherwise, maybe you can move this to the group by function sou you'd be using `startOfHour` in the group by but maintaining the original timestamp of the metric

src/lib/util/collapseHourlyMetrics.ts
```diff
@@ -0,0 +1,41 @@
...
+    }));
+
+    return groupBy(hourlyMetrics, createMetricKey).flatMap((group) => ({
+        ...group[0],
+        yes: sum(group.map((metric) => metric.yes)),
+        no: sum(group.map((metric) => metric.no)),
```
gastonfournier, line 39: Here we're iterating twice the list, let me see if I can propose something better

src/lib/util/collapseHourlyMetrics.ts
olav: Yep intended, though a bit unclear. The function is supposed to group metrics by start-of-hour (like in the original code).

src/lib/util/collapseHourlyMetrics.ts
olav: True, could do something like this:

```typescript
const sumYesNo = (
    metrics: IClientMetricsEnv[],
): Pick<IClientMetricsEnv, 'yes' | 'no'> => {
    return metrics.reduce(
        (acc, metric) => ({
            yes: acc.yes + metric.yes,
            no: acc.no + metric.no,
        }),
        {
            yes: 0,
            no: 0,
        },
    );
};

export const collapseHourlyMetrics = (
    metrics: IClientMetricsEnv[],
): IClientMetricsEnv[] => {
    const hourlyMetrics = metrics.map((metric) => ({
        ...metric,
        timestamp: startOfHour(metric.timestamp),
    }));

    return groupBy(hourlyMetrics, createMetricKey).flatMap((group) => ({
        ...group[0],
        ...sumYesNo(group),
    }));
};
```

gastonfournier: looks nice! but now I'm thinking if we can do this in only one loop :sweat_smile: don't worry about it. But your proposal looks like a good improvement :+1:

FredrikOseberg (approved):
We could consider flagging this.

src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -8,47 +8,64 @@ import {
...
+                    this.bulkAdd().catch(console.error);
+                }, bulkInterval).unref(),
+            );
+        }
+
+        this.timers.push(
+            setInterval(() => {
+                this.clientMetricsStoreV2.clearMetrics(48).catch(console.error);
+            }, hoursToMilliseconds(12)).unref(),
+        );
```
gastonfournier, line 68: Should this be part of the else statement?

olav: Nope, this is a separate timer for clearing old metrics.

PR #1931, opened 2022-08-16
Feat/add alias to api tokens

FredrikOseberg:
## About the changes
This is a draft PR that allows us to add metadata to the api key tokens. This will allow us to keep backwards compatability with the old proxy client keys when introducing embedded proxy. 

Relates to: #1875

### Important files
- api-token-service
- api-token-store

### DIFF
--- a/src/lib/db/api-token-store.ts
+++ b/src/lib/db/api-token-store.ts
@@ -66,7 +66,7 @@ const toRow = (newToken: IApiTokenCreate) => ({
     environment:
         newToken.environment === ALL ? undefined : newToken.environment,
     expires_at: newToken.expiresAt,
-    alias: newToken.alias || '',
+    alias: newToken.alias || null,
 });
 
 const toTokens = (rows: any[]): IApiToken[] => {
@@ -153,7 +153,7 @@ export class ApiTokenStore implements IApiTokenStore {
             await Promise.all(updateProjectTasks);
             return {
                 ...newToken,
-                alias: newToken.alias || '',
+                alias: newToken.alias || null,
                 project: newToken.projects?.join(',') || '*',
                 createdAt: row.created_at,
             };
--- a/src/lib/db/api-token-store.ts
+++ b/src/lib/db/api-token-store.ts
@@ -45,7 +45,7 @@ const tokenRowReducer = (acc, tokenRow) => {
             environment: token.environment ? token.environment : ALL,
             expiresAt: token.expires_at,
             createdAt: token.created_at,
-            metadata: token.metadata,
+            alias: token.alias,
         };
     }
     const currentToken = acc[tokenRow.secret];
@@ -66,7 +66,7 @@ const toRow = (newToken: IApiTokenCreate) => ({
     environment:
         newToken.environment === ALL ? undefined : newToken.environment,
     expires_at: newToken.expiresAt,
-    metadata: newToken.metadata || {},
+    alias: newToken.alias || '',
 });
 
 const toTokens = (rows: any[]): IApiToken[] => {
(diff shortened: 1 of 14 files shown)


src/lib/services/api-token-service.ts
```diff
@@ -102,12 +102,16 @@ export class ApiTokenService {
     }
 
     public getUserForToken(secret: string): ApiUser | undefined {
-        let token = this.activeTokens.find((t) => t.secret === secret);
+        let token = this.activeTokens.find(
+            (activeToken) => activeToken.secret === secret,
+        );
 
         // If the token is not found, try to find it in the legacy format with the metadata alias
         // This is to ensure that previous proxies we set up for our customers continue working
```
olav, line 110: Could remove `metadata` from this comment, as well as "customers" since this could be a generic feature.

src/lib/db/api-token-store.ts
```diff
@@ -66,7 +66,7 @@ const toRow = (newToken: IApiTokenCreate) => ({
     environment:
         newToken.environment === ALL ? undefined : newToken.environment,
     expires_at: newToken.expiresAt,
-    metadata: newToken.metadata || {},
+    alias: newToken.alias || '',
```
olav, line 69: Should this default to `null` or `undefined` instead?

src/lib/db/api-token-store.ts
```diff
@@ -153,7 +153,7 @@ export class ApiTokenStore implements IApiTokenStore {
             await Promise.all(updateProjectTasks);
             return {
                 ...newToken,
-                metadata: newToken.metadata || {},
+                alias: newToken.alias || '',
```
olav, line 156: Same here.

src/test/e2e/api/proxy/proxy.e2e.test.ts
```diff
@@ -104,6 +104,37 @@ test('should not allow requests with a client token', async () => {
...
+        alias: tokenAlias,
+    });
+
+    await createFeatureToggle({
+        name: 'enabledFeature1',
+        enabled: true,
+        strategies: [{ name: 'default', constraints: [], parameters: {} }],
+    });
+
+    await app.request
```
olav, line 119: Could test a request with the new key as well, just to make sure that both work.

src/test/fixtures/fake-api-token-store.ts
```diff
@@ -51,8 +51,8 @@ export default class FakeApiTokenStore
         const apiToken = {
             createdAt: new Date(),
             project: newToken.projects?.join(',') || '*',
+            alias: '',
```
olav, line 54: Maybe `null`?

olav (changes requested):

src/lib/services/api-token-service.ts
FredrikOseberg: Thanks

src/test/e2e/api/proxy/proxy.e2e.test.ts
FredrikOseberg: Doesn't this test also verify that, because the old key is just used to resolve the key with the new format?

FredrikOseberg: There is also a test verifying it here: https://github.com/Unleash/unleash/pull/1931/files/69b2b2729c73cabf41399e0f033e37c490e0359d#diff-0527b2389c8d5179e9b86200bd1533d12b2798b54da2d21490989e1cc1dc1cabR178

FredrikOseberg: What would be good to test is that the token retrieves from the correct project and environments.

Tymek (approved):

olav (approved):

PR #1933, opened 2022-08-16
fix: Prototype Pollution in async

gastonfournier:
## About the changes
Closes https://github.com/Unleash/unleash/security/dependabot/18

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "@unleash/express-openapi": "^0.2.0",
     "ajv": "^8.11.0",
     "ajv-formats": "^2.1.1",
-    "async": "^3.2.3",
+    "async": "^3.2.4",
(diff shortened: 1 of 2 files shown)


gastonfournier:
Actually the dependency comes from here: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/blob/2bdf6a86b66d81ea1bdd3c5ed377416f39b1ff92/packages/docusaurus-plugin-openapi-docs/package.json#L47 which is using a particular hotfix and it's not updated with the latest version of https://www.npmjs.com/package/openapi-to-postmanv2 which is 4.2.0, but we'd have to dig deeper into the hotfix version, or check with the maintainers

ivarconr (approved):

PR #1936, opened 2022-08-17
feat: add CORS support to the proxy endpoints

olav:
https://linear.app/unleash/issue/1-154/embedded-proxy-add-apifrontend-endpoints

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -29,7 +29,11 @@ import {
     mapLegacyToken,
     validateApiToken,
 } from './types/models/api-token';
-import { parseEnvVarBoolean, parseEnvVarNumber } from './util/env';
+import {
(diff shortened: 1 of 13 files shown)


src/lib/create-config.ts
```diff
@@ -402,6 +406,10 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
         DEFAULT_STRATEGY_SEGMENTS_LIMIT,
     );
 
+    const frontendApiOrigins =
```
FredrikOseberg, line 409: Should we default to `*` if no origins are passed?

olav: Not sure, seems safer to disallow everything by default. Could always use a different default for `cloud`.

FredrikOseberg (approved):

PR #1941, opened 2022-08-18
chore: type improvements

gastonfournier:
## About the changes
Role validation only requires id and environment from each permission to validate, but `IRoleCreation` refers to `IPermission` that has other mandatory fields. This should help create more accurate OpenAPI specs where we use this method from an API

### DIFF
--- a/src/lib/services/access-service.test.ts
+++ b/src/lib/services/access-service.test.ts
@@ -1,11 +1,7 @@
 import NameExistsError from '../error/name-exists-error';
 import getLogger from '../../test/fixtures/no-logger';
 import createStores from '../../test/fixtures/store';
-import {
-    AccessService,
-    IRoleCreation,
-    IRoleValidation,
-} from './access-service';
+import { AccessService, IRoleValidation } from './access-service';
 
 function getSetup(withNameInUse: boolean) {
     const stores = createStores();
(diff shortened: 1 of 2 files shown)


src/lib/services/access-service.ts
```diff
@@ -42,12 +42,18 @@ const PROJECT_ADMIN = [
     permissions.DELETE_FEATURE,
 ];
 
-interface IRoleCreation {
+export interface IRoleCreation {
```
gastonfournier, line 45: I don't know if it's possible to export this interface only for testing purposes

src/lib/services/access-service.test.ts
```diff
@@ -0,0 +1,162 @@
...
+                id: 1,
+                environment: 'development',
+            },
+        ],
+    });
+});
+
+test('should be able to validate with IRoleCreation', async () => {
+    const { accessService } = getSetup(false);
+    const test: IRoleCreation = {
```
gastonfournier, line 113: This validates backward compatibility with https://github.com/Unleash/unleash/pull/1941/files#diff-a724d524b9f64dc8e2a9432972602c9217f54f70ef1f22d7472b3da25403b2c5L528

src/lib/services/access-service.ts
olav: Not without a dependency, I think. Seems fine to export types for testing.

src/lib/services/access-service.test.ts
gastonfournier: Removed as per conversation it should be fine as we're relaxing restrictions in the original type

olav (approved):

PR #1945, opened 2022-08-19
refactor: improve performance of groupHourlyMetrics

gastonfournier:
Optimize grouping for https://github.com/Unleash/unleash/pull/1928

### DIFF
--- a/src/lib/util/collapseHourlyMetrics.ts
+++ b/src/lib/util/collapseHourlyMetrics.ts
@@ -1,17 +1,6 @@
 import { IClientMetricsEnv } from '../types/stores/client-metrics-store-v2';
 import { startOfHour } from 'date-fns';
 
-const groupBy = <T>(list: T[], createKey: (item: T) => string): T[][] => {
-    const groups = list.reduce((acc, item) => {
-        const key = createKey(item);
-        acc[key] = acc[key] ?? [];
-        acc[key].push(item);
-        return acc;
-    }, {} as Record<string, T[]>);
-
-    return Object.values(groups);
-};
-
 const createMetricKey = (metric: IClientMetricsEnv): string => {
     return [
         metric.featureName,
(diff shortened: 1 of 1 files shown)


src/lib/util/collapseHourlyMetrics.ts
```diff
@@ -21,31 +10,24 @@ const createMetricKey = (metric: IClientMetricsEnv): string => {
...
+    metrics.forEach((metric) => {
+        const hourlyMetric = {
+            ...metric,
+            timestamp: startOfHour(metric.timestamp),
+        };
+        const key = createMetricKey(hourlyMetric);
+        if (!grouped[key]) {
+            grouped[key] = {
+                ...hourlyMetric,
+            };
```
olav, line 26: Nice work! I don't think this has to be a copy, since we just coped `metric` into `hourlyMetric`.

```ts
grouped[key] = hourlyMetric;
```

src/lib/util/collapseHourlyMetrics.ts
```diff
@@ -21,31 +10,24 @@ const createMetricKey = (metric: IClientMetricsEnv): string => {
...
+        } else {
+            grouped[key].yes = metric.yes + (grouped[key].yes || 0);
+            grouped[key].no = metric.no + (grouped[key].no || 0);
```
olav, line 29: Do we need the `|| 0`s here? I think the metrics objects all have yes/no counts by default.

olav (changes requested):

src/lib/util/collapseHourlyMetrics.ts
gastonfournier: Honestly, I don't know. The type does not accept undefined: https://github.com/Unleash/unleash/blob/b5fead41f66b1f7da0484197b0adb710d7f45975/src/lib/types/stores/client-metrics-store-v2.ts#L10-L13 but since we're not strict with this, it could potentially be. We could consider that a bug, but I prefer to play safe. If you think it's not required and safe to be removed that'd be ok for me

src/lib/util/collapseHourlyMetrics.ts
gastonfournier: https://github.com/Unleash/unleash/pull/1945/commits/12207c63871c41bc11c7be6e316a276a6a2940ad

olav (approved):

PR #1950, opened 2022-08-22
fix: clean up proxy clients on API key deletion

olav:
When a proxy key gets deleted, we should also remove its proxy client.


Tymek (approved):
LGTM

PR #1953, opened 2022-08-22
feat: fetch project and environment from body

gastonfournier:
## About the changes
To improve OpenAPI documentation, we'd like to document the required permissions. In order to do that, we need to be able to get the project and environment parameters that come in the request body. Following a similar workaround as we used previously this should be considered a temporary solution until we standardise how we send parameters and the names we use.

### DIFF
--- a/src/lib/middleware/rbac-middleware.ts
+++ b/src/lib/middleware/rbac-middleware.ts
@@ -54,6 +54,9 @@ const rbacMiddleware = (
                 projectId = await featureToggleStore.getProjectId(featureName);
             } else if (permission === CREATE_FEATURE) {
                 projectId = projectId || req.body.project || 'default';
+            } else if (permission.endsWith('FEATURE_STRATEGY')) {
+                projectId = projectId || req.body.projectId;
+                environment = environment || req.body.environmentId;
             }
 
             return accessService.hasPermission(


src/lib/middleware/rbac-middleware.ts
```diff
@@ -54,6 +54,9 @@ const rbacMiddleware = (
                 projectId = await featureToggleStore.getProjectId(featureName);
             } else if (permission === CREATE_FEATURE) {
                 projectId = projectId || req.body.project || 'default';
+            } else if (permission.endsWith('FEATURE_STRATEGY')) {
+                projectId = projectId || req.body.projectId;
+                environment = environment || req.body.environmentId;
```
gastonfournier, line 59: Two ideas here: 
- The permission accepts a callback that will extract the project and environment
- We standardize how we propagate these parameters

We can decide to close this PR in case we want to move to the second option. The first one would require some ideation. Meanwhile, the workaround could help define the proper permission [here](https://github.com/ivarconr/unleash-enterprise/blob/master/src/routes/segment-controller.ts#L63) so it's documented in OpenAPI

ivarconr: I like this idea!

ivarconr (approved):

ivarconr:
@gastonfournier why is this hold back? Is it not a good idea after all?

src/lib/middleware/rbac-middleware.ts
gastonfournier: I'm closing this now and moving to the backlog until I can spend sometime implementing that

PR #1957, opened 2022-08-23
feat: add CORS instance settings

olav:
https://linear.app/unleash/issue/1-171/embedded-proxy-instance-cors-settings-api

### DIFF
--- a/src/lib/util/validateOrigin.test.ts
+++ b/src/lib/util/validateOrigin.test.ts
@@ -1,0 +1,21 @@
+import { validateOrigin } from './validateOrigin';
+
+test('validateOrigin', () => {
+    expect(validateOrigin('')).toEqual(false);
+    expect(validateOrigin(' ')).toEqual(false);
+    expect(validateOrigin('a')).toEqual(false);
+    expect(validateOrigin('**')).toEqual(false);
+    expect(validateOrigin('localhost')).toEqual(false);
+    expect(validateOrigin('localhost:8080')).toEqual(false);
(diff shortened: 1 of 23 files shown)


src/lib/util/validateOrigin.test.ts
```diff
@@ -0,0 +1,21 @@
...
+    expect(validateOrigin('**')).toEqual(false);
+    expect(validateOrigin('localhost')).toEqual(false);
+    expect(validateOrigin('localhost:8080')).toEqual(false);
+    expect(validateOrigin('//localhost:8080')).toEqual(false);
+    expect(validateOrigin('http://localhost/')).toEqual(false);
+    expect(validateOrigin('http://localhost/a')).toEqual(false);
+    expect(validateOrigin('https://example.com/a')).toEqual(false);
+    expect(validateOrigin('https://example.com ')).toEqual(false);
+
+    expect(validateOrigin('*')).toEqual(true);
```
Tymek, line 16: is `*.localhost` valid? or `https://*.example.com`?

olav: Nope, for CORS we have to use `*` or a full origin. Would be nice though!

Tymek: Could be added to as expected to fail

Tymek (approved):

src/lib/util/validateOrigin.test.ts
olav: Yeah, good idea!

PR #1959, opened 2022-08-24
fix: enforce non-nullability for createdBy/modifiedBy arguments

chriswk:
We've been having a couple of problems with our event store not being able to save events, due to createdBy being `null`. We've even had comments telling us to remove the optional nature of these arguments. 

This PR makes these arguments (createdBy, modifiedBy) non-nullable/compulsory, we'll just have to fix enterprise once we've merged and released this.

That should solve these instances of our EventStore not being able to save the events.

### DIFF
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -294,12 +294,11 @@ export default class ProjectService {
         };
     }
 
-    // TODO: Remove the optional nature of createdBy - this in place to make sure enterprise is compatible
     async addUser(
         projectId: string,
         roleId: number,
         userId: number,
-        createdBy?: string,
+        createdBy: string,
     ): Promise<void> {
         const [roles, users] = await this.accessService.getProjectRoleAccess(
             projectId,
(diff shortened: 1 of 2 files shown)


gastonfournier (commented):
it looks fine, but it may break the interfaces of anyone using it. It doesn't seem like a big issue though... but I was wondering:

1. Should this remain optional and just use a default value?
2. If we strictly follow semver, this should be a major change, although I'd have a hard time justifying that...

chriswk:
> it looks fine, but it may break the interfaces of anyone using it. It doesn't seem like a big issue though... but I was wondering:
> 
>     1. Should this remain optional and just use a default value?

I'd prefer to enforce presence

> 
>     2. If we strictly follow semver, this should be a major change, although I'd have a hard time justifying that...

Yes, particularly since I don't think we've said that you can use us a library, I'd argue that our contract is at the HTTP API level, but I can definitely see the argument that this is really a breaking change to the node library API surface. @ivarconr ?

gastonfournier (approved):
Based on our comments I think it's fine, although it would be nice to have confirmation, I think it's fair to assume our contract is based on our API contract

PR #1961, opened 2022-08-24
Feat/exp flag loader

ivarconr:
## About the changes

Introduces the concept of an external flag resolver. This would allow us to use Unleash in Unleash to enable/disable experimental features. 

An example on how to use the Unleash SDK as an flag resolver:

```js
import { start } from './lib/server-impl';
import { createConfig } from './lib/create-config';
import { Unleash } from 'unleash-client';

const unleash = new Unleash({
    appName: 'unleash',
    url: 'https://app.unleash-hosted.com/hosted/api/',
    refreshInterval: 1000,
    customHeaders: {
        Authorization:
            'unleash-cloud:development.xxxxx',
    },
});

process.nextTick(async () => {
    try {
        await start(
            createConfig({
                db: {
                    user: 'unleash_user',
                    password: 'passord',
                    host: 'localhost',
                    port: 5432,
                    database: process.env.UNLEASH_DATABASE_NAME || 'unleash',
                    schema: process.env.UNLEASH_DATABASE_SCHEMA,
                    ssl: false,
                    applicationName: 'unleash',
                },
                experimental: {
                    externalResolver: unleash,
                },
            }),
        );
    } catch (error) {
        console.error(error);
        process.exit();
    }
}, 0);
```

## Discussion points

- **dynamicFlags** provides the names of the flags that can be dynamically enabled. The intention is to protect "non-dynamic flags" to be controlled by an external resolver. It also allows us to "know" which ui flags to check-for in addition to the statically configured UI flags. 
- Enabled static ui flags and experimental flags will not be possible to override with a dynamic flag. The intention is that these are expected to be enabled regardless of what the external resolver (unleash) thinks. 
- Removed userGroups flag. Can @sighphyre validate that (https://github.com/Unleash/unleash/pull/1961/commits/e01b86333a5cd2b30dfe9702a77a1428a67ed881)? 
- Removed the **metricsV2** flag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
     "ts-toolbelt": "^9.6.0",
     "type-is": "^1.6.18",
     "unleash-client": "3.15.0",
-    "unleash-frontend": "4.15.0-beta.1",
+    "unleash-frontend": "4.14.8",
     "uuid": "^8.3.2"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
     "ts-toolbelt": "^9.6.0",
     "type-is": "^1.6.18",
     "unleash-client": "3.15.0",
-    "unleash-frontend": "4.14.8",
+    "unleash-frontend": "4.15.0-beta.1",
     "uuid": "^8.3.2"
   },
   "devDependencies": {
--- a/src/lib/routes/admin-api/user-admin.ts
+++ b/src/lib/routes/admin-api/user-admin.ts
@@ -37,9 +37,10 @@ import {
     usersGroupsBaseSchema,
 } from '../../openapi/spec/users-groups-base-schema';
 import { IGroup } from '../../types/group';
+import { IFlagResolver } from '../../types/experimental';
 
 export default class UserAdminController extends Controller {
-    private anonymise: boolean = false;
+    private flagResolver: IFlagResolver;
 
     private userService: UserService;
 
@@ -90,7 +91,7 @@ export default class UserAdminController extends Controller {
         this.groupService = groupService;
         this.logger = config.getLogger('routes/user-controller.ts');
         this.unleashUrl = config.server.unleashUrl;
-        this.anonymise = config.experimental?.anonymiseEventLog;
+        this.flagResolver = config.flagResolver;
 
         this.route({
             method: 'post',
@@ -294,7 +295,7 @@ export default class UserAdminController extends Controller {
             typeof q === 'string' && q.length > 1
                 ? await this.userService.search(q)
                 : [];
-        if (this.anonymise) {
+        if (this.flagResolver.isEnabled('anonymiseEventLog')) {
             users = this.anonymiseUsers(users);
         }
         this.openApiService.respondWithValidation(
--- a/src/lib/types/experimental.ts
+++ b/src/lib/types/experimental.ts
@@ -1,0 +1,31 @@
+export interface IFlags {
+    [key: string]: boolean;
+}
+export interface IExperimentalOptions {
+    flags?: {
+        [key: string]: boolean;
+    };
+    anonymiseEventLog?: boolean;
+    userGroups?: boolean;
+    embedProxy?: boolean;
+    batchMetrics?: boolean;
+    dynamicFlags?: string[];
+    externalResolver?: IExternalFlagsResolver;
+}
+
+export interface IUIFlags extends IFlags {
+    ENABLE_DARK_MODE_SUPPORT?: boolean;
+}
+
+export interface IFlagContext {
+    [key: string]: string;
+}
+
+export interface IFlagsResolver {
+    getUIFlags: (context?: IFlagContext) => IUIFlags;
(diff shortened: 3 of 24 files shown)


src/server-dev.ts
```diff
@@ -2,6 +2,17 @@
...
 import { ApiTokenType } from './lib/types/models/api-token';
+import { Unleash } from 'unleash-client';
+
+const unleash = new Unleash({
+    appName: 'unleash',
+    url: 'https://app.unleash-hosted.com/hosted/api/',
+    refreshInterval: 1000,
+    customHeaders: {
+        Authorization:
+            'unleash-cloud:development.f719679e65c7367b7c36b4f08f04237116b862897f0c2808c976a48c',
```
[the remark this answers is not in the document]
ivarconr: thanks, I already deactivated this token.

package.json
```diff
@@ -125,7 +125,7 @@
     "ts-toolbelt": "^9.6.0",
     "type-is": "^1.6.18",
     "unleash-client": "3.15.0",
-    "unleash-frontend": "4.15.0-beta.1",
+    "unleash-frontend": "4.14.8",
```
olav, line 128: Intended?

src/lib/types/experimental.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export interface IUIFlags extends IFlags {
+    ENABLE_DARK_MODE_SUPPORT?: boolean;
+}
+
+export interface IFlagContext {
+    [key: string]: string;
+}
+
+export interface IFlagResolver {
```
olav, line 24: Could consider merging UI flags and experimental flags into one flat map of flags. 

Not sure if the distinction between an experimental and an UI flag is worth keeping.

src/server-dev.ts
```diff
@@ -3,6 +3,19 @@ import { createConfig } from './lib/create-config';
 import { LogLevel } from './lib/logger';
 import { ApiTokenType } from './lib/types/models/api-token';
 
+/*
```
olav, line 6: Could remove.

src/lib/util/flag-resolver.ts
```diff
@@ -0,0 +1,53 @@
...
+    private experiments: IFlags;
+
+    private dynamicFlags: string[] = [];
+
+    private externalResolver: IExternalFlagResolver;
+
+    constructor(uiFlags: IUIFlags, expOpt: IExperimentalOptions) {
+        this.uiFlags = uiFlags || {};
+        this.experiments = expOpt.flags || {};
+        this.dynamicFlags = [
```
olav, line 26: Do we need the list of dynamic flags if we always check the static flags first? Seems nice to check static flags first, and then fallback to the resolver (which defaults to false).

src/lib/types/experimental.ts
```diff
@@ -0,0 +1,31 @@
...
+    getUIFlags: (context?: IFlagContext) => IUIFlags;
+    isExperimentEnabled: (flagName: string, context?: IFlagContext) => boolean;
```
olav, line 26: Not sure if we'll need an `isReady` method to wait for at startup, or events that trigger on flag update. 

Maybe something for later.

olav (changes requested):

src/lib/util/flag-resolver.ts
olav: Could also check the presence of the key instead of its true/false value if we'd like to enforce that some flags never use the resolver (as mentioned), though that seems brittle. Maybe the list is better.

ivarconr: > Could also check the presence of the key instead of its true/false value 

I did that first, but it feels a bit strange to always try to resolve something that is a static ui flag that is used for business purposes and not experimental features.

src/lib/types/experimental.ts
ivarconr: you could achieve that by deferring the initialization of unleash until the third party flag resolver is ready.

src/lib/types/experimental.ts
ivarconr: we could. I was actually thinking it would just enrich the ui-flags with experimental flags, makes sense to have them available in the UI as well. What do you think?

olav: Yeah, it would be really nice to automatically have new experimental flags available on the frontend. :+1: 

I was thinking we could also remove the distinction on the backend (and just have "flags"), but not sure.

package.json
ivarconr: yes for testing

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -294,7 +295,7 @@ export default class UserAdminController extends Controller {
             typeof q === 'string' && q.length > 1
                 ? await this.userService.search(q)
                 : [];
-        if (this.anonymise) {
+        if (this.flagResolver.isEnabled('anonymiseEventLog')) {
```
ivarconr, line 298: this is how you can use the flagResolver for the backend stuff.

olav (approved):

PR #1964, opened 2022-08-25
Add a counter for total number of environments

chriswk:
As an attempt from our Customer Success representatives, we'd like to keep track of how much use of our Pro/Enterprise features our customers are making. This PR adds number of environments as a metric (we already had number of projects). If we have users that have more than 2 environments, this will tell us they're using an upgraded feature.


sighphyre (approved):
Nice!

PR #1965, opened 2022-08-25
fix: proxy api check

FredrikOseberg:
Since we are mounting the current proxies under the following paths:
* /clientId/api/environment/proxy

And the current clientKeys are being migrated to frontend keys, we need to amend the api-token middleware to get access to the new endpoints.

#1875

### DIFF
--- a/src/lib/middleware/api-token-middleware.ts
+++ b/src/lib/middleware/api-token-middleware.ts
@@ -7,7 +7,16 @@ const isClientApi = ({ path }) => {
 };
 
 const isProxyApi = ({ path }) => {
-    return path && path.startsWith('/api/frontend');
+    if (!path) {
+        return;
+    }
+
+    return (
+        (!path.startsWith('/api/client') &&
+            !path.startsWith('/api/admin') &&
+            path.includes('proxy')) ||
+        (path && path.startsWith('/api/frontend'))
+    );
 };
 
 export const TOKEN_TYPE_ERROR_MESSAGE =
--- a/src/lib/middleware/api-token-middleware.ts
+++ b/src/lib/middleware/api-token-middleware.ts
@@ -11,11 +11,13 @@ const isProxyApi = ({ path }) => {
         return;
     }
 
+    // Handle all our current proxy paths which will redirect to the new
+    // embedded proxy endpoint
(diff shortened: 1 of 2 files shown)


src/lib/middleware/api-token-middleware.ts
```diff
@@ -7,7 +7,16 @@ const isClientApi = ({ path }) => {
...
 const isProxyApi = ({ path }) => {
-    return path && path.startsWith('/api/frontend');
+    if (!path) {
+        return;
+    }
+
+    return (
+        (!path.startsWith('/api/client') &&
+            !path.startsWith('/api/admin') &&
+            path.includes('proxy')) ||
```
olav, line 17: Seems like `path.includes('proxy')` may match more than we want. Could we be stricter?

src/lib/middleware/api-token-middleware.ts
```diff
@@ -7,7 +7,16 @@ const isClientApi = ({ path }) => {
...
+        (path && path.startsWith('/api/frontend'))
```
olav, line 18: Don't need this `path` check anymore.

src/lib/middleware/api-token-middleware.ts
```diff
@@ -11,11 +11,13 @@ const isProxyApi = ({ path }) => {
...
+    // Handle all our current proxy paths which will redirect to the new
+    // embedded proxy endpoint
     return (
-        (!path.startsWith('/api/client') &&
-            !path.startsWith('/api/admin') &&
-            path.includes('proxy')) ||
-        (path && path.startsWith('/api/frontend'))
+        path.startsWith('/api/default/proxy') ||
+        path.startsWith('/api/development/proxy') ||
+        path.startsWith('/api/production/proxy') ||
```
olav, line 19: Nice! Can there be custom environment names in this case, or will it always be one of these three?

olav (approved):

src/lib/middleware/api-token-middleware.ts
FredrikOseberg: I'm double checking this, but according to our infra as code, those are the only ones we use.

PR #1987, opened 2022-08-29
refactor: simplify login redirect logic

olav:
We currently handle login redirects in both `SWRProvider` and `ProtectedRoute`. The latter should be enough.

Also removes some API retry logic that I don't think we'll need.


chriswk (approved):
LGTM. I like the simplification :)

PR #2003, opened 2022-08-30
Docker cross-compilation with buildx

sjaanus:
Docker build time reduced from 30 minutes to 13 minutes. Frontend is now built with buildx, so it does not have to build on slow arm64.


gardleopard (approved):
great work!

ivarconr (approved):

PR #2013, opened 2022-08-31
refactor: avoid building backend on frontend change

olav:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths


sjaanus (approved):
LGTM!

PR #2018, opened 2022-08-31
fix: avoid stale feature strategy enabled status message

olav:
Ensures that `FeatureStrategyEnabled` uses fresh feature data, and not what's cached in the form.

https://linear.app/unleash/issue/1-199/the-general-message-inside-strategies-if-feature-toggle-is-enabled-in


andreas-unleash (approved):
LGTM!

PR #2022, opened 2022-08-31
feat add count to Environments

chriswk:
By request from Nicolae, this PR adds a counter to the EnvironmentsList.


andreas-unleash (approved):
LGTM

PR #2030, opened 2022-08-31
fix: toggle strategies ui adjustments

Tymek:
https://linear.app/unleash/issue/1-188/environment-cards-ui-adjustments

- Removed the padding from the environment cards
- Removed the background color for the Exposure boxes
- Make these two buttons for new (first) strategy equal


chriswk (approved):
LGTM

PR #2031, opened 2022-08-31
fix: show addon event select validation

Tymek:
## About the changes

https://linear.app/unleash/issue/1-198/addons-missing-fields-validation


chriswk (approved):
LGTM :+1:

PR #2040, opened 2022-09-01
Fix tests not passing when timezone not UTC

sjaanus:
Fix tests not passing when timezone not UTC


chriswk (approved):
LGTM

sighphyre:
Awesome! I'm buying you a beer when we get to Oslo for this one!

PR #2042, opened 2022-09-02
task: use swc instead of ts-jest

chriswk:
So, one of the things that slows down tests is the compile time, this PR uses swc instead of ts-jest for compiling/preparing the tests. I'm seeing 10-20 seconds improvement on my machine, but contrary to auth-app and unleash-cloud, what's really using time in our tests are the actual testing, not compiling our tests. So the win isn't as big here as it is could be. Still feels like it could be worth it.


sighphyre (approved):
Even if it's not as a big an improvement as we'd like, I can't see any obvious downsides to this and 10-20 seconds is 10-20 seconds I didn't have yesterday so... LGTM

PR #2043, opened 2022-09-05
fix: add env and project labels to feature updated metrics.

ivarconr:
## About the changes
Adds environment and project labels to feature updated metrics. It makes it easier to monitor usage of Unleash in larger deployments.

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -64,7 +64,7 @@ export default class MetricsMonitor {
         const featureToggleUpdateTotal = new client.Counter({
             name: 'feature_toggle_update_total',
             help: 'Number of times a toggle has  been updated',
-            labelNames: ['toggle'],
+            labelNames: ['toggle', 'project', 'environment'],
         });
         const featureToggleUsageTotal = new client.Counter({
             name: 'feature_toggle_usage_total',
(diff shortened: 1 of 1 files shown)


chriswk (approved):
LGTM

gardleopard (changes requested):
This doesnt work in my local env. I have tried flipping a kill switch multiple times with no luck
http://localhost:4242/internal-backstage/prometheus
```

# HELP feature_toggle_update_total Number of times a toggle has  been updated
# TYPE feature_toggle_update_total counter

# HELP feature_toggle_usage_total Number of times a feature toggle has been used
# TYPE feature_toggle_usage_total counter

# HELP feature_toggles_total Number of feature toggles
# TYPE feature_toggles_total gauge
feature_toggles_total{version="4.15.0-beta.9"} 1

# HELP users_total Number of users
# TYPE users_total gauge
users_total 2

# HELP projects_total Number of projects
# TYPE projects_total gauge
projects_total 1
```

gardleopard:
is feature toggle update total intended to increase for every time we change the status as well? If that is supposed to happen, there is some flaw there as well

ivarconr:
> Works for normal toggles, but the "n/a" environment is strange as well
> 
> ```
> # HELP feature_toggle_update_total Number of times a toggle has  been updated
> # TYPE feature_toggle_update_total counter
> feature_toggle_update_total{toggle="testSwitch",project="default",environment="n/a"} 1
> feature_toggle_update_total{toggle="testSwitch",project="default",environment="development"} 1
> feature_toggle_update_total{toggle="testSwitch",project="default",environment="production"} 1
> ```

The problem is that some "updates" is not done for an environment (e.g creating a toggle, archiving it, etc). In those scenarios we have no environment to be set as label. The prometheus client fails hard if we do not specify the label value when it is configured.

ivarconr:
> Works for normal toggles, but the "n/a" environment is strange as well

would you prefer "undefined"?

gardleopard:
> 

I think your explanation is good here, so n/a is a valid case here:)

gardleopard (approved):
Now it looks good!

```
feature_toggle_update_total{toggle="dummy",project="default",environment="development"} 4
feature_toggle_update_total{toggle="dummy",project="default",environment="production"} 2
feature_toggle_update_total{toggle="dummy2",project="default",environment="n/a"} 1
feature_toggle_update_total{toggle="dummy2",project="default",environment="development"} 3
feature_toggle_update_total{toggle="dummy2",project="default",environment="production"} 2
feature_toggle_update_total{toggle="testToggle",project="default",environment="production"} 8
feature_toggle_update_total{toggle="testToggle",project="default",environment="development"} 2
```

PR #2044, opened 2022-09-05
fix: strategy parameters - trim string value

Tymek:
There was an inconsistency on custom strategy parameter input. Some values where trimmed, some not, and value sent to server was different from the contents of an input field.

https://linear.app/unleash/issue/1-144/custom-strategies-required-params-can-introduce-empty-space-at-the-end


chriswk (approved):
Nice. LGTM :)

PR #2051, opened 2022-09-07
Revert workflow

sjaanus:
When merging `frontend` and `backend` we kept all the commits. This meant all `frontend` commits were added to this repository.

This brought up an issue, where release notes started failing, because the release notes included the whole history of `frontend`.

To bypass this, we disabled `backend` release notes until a minor release was done, which was 4.15.
Now we can reenable the full release notes. **This PR reverts the workflow to initial state.**


chriswk (approved):
Because the changelog does a diff from previous tag no, so now that we have a tag where we didn't merge 3000 commits, this works out fine again :)

LGTM :+1:

Issue #1877, opened 2022-08-02
🔥feat: Invite team members to Unleash


ivarconr:
### Describe the feature request

Today you need to manually add users with name and email to Unleash (and you must be an Admin to do that). 

### Background

Customizable "root role" for the invite token

Allow project owners to have an invite link only for that project
- Must also solve the case that the user has access to Unleash but is not member of that project.

ivarconr:
This is in beta from v4.16

ivarconr:
GA in v4.18

PR #2053, opened 2022-09-08
PublicSignupTokens

andreas-unleash:
This PR adds the ability to create public signup tokens, to be used in public invite links.  The link will allow Unleash users to invite other people to Unleash and easily signup, granting them the Viewer Role

Relates to  https://github.com/Unleash/unleash/issues/1877

## About the changes

Closes # .


src/lib/db/public-signup-token-store.ts
```diff
@@ -0,0 +1,189 @@
...
+                { user_id: userId, secret },
+                ['created_at'],
+            );
+        });
+    }
+
+    async insert(
+        newToken: IPublicSignupTokenCreate,
+    ): Promise<PublicSignupTokenSchema> {
+        const response = await this.db.transaction(async (tx) => {
```
FredrikOseberg, line 137: Is there a reason this is wrapped in a transaction, it looks to me like this is only doing one interaction with the database?

src/lib/db/public-signup-token-store.ts
```diff
@@ -0,0 +1,189 @@
...
+        );
+        const { present } = result.rows[0];
+        return present;
+    }
+
+    async get(key: string): Promise<PublicSignupTokenSchema> {
+        const row = await this.makeTokenUsersQuery()
+            .where('secret', key)
+            .first();
+        return toTokens([row])[0];
```
FredrikOseberg, line 165: first is an optional, I don't think we're always guaranteed a row here.

src/lib/routes/admin-api/index.ts
```diff
@@ -101,6 +102,13 @@ class AdminApi extends Controller {
             '/constraints',
             new ConstraintsController(config, services).router,
         );
+        this.app.use(
```
FredrikOseberg, line 105: is publicSignup the best wording for this feature? What about inviteLink?

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -0,0 +1,252 @@
...
+                    responses: {
+                        200: emptyResponse,
+                    },
+                }),
+            ],
+        });
+
+        this.route({
+            method: 'get',
+            path: '/token/:token',
```
FredrikOseberg, line 119: `/tokens/:token` maybe

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -0,0 +1,252 @@
...
+                    responses: {
+                        200: createResponseSchema('publicSignupTokenSchema'),
+                    },
+                }),
+            ],
+        });
+
+        this.route({
+            method: 'put',
+            path: '/token/:token',
```
FredrikOseberg, line 135: `/tokens/:token` maybe

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -0,0 +1,252 @@
...
+                    responses: {
+                        200: emptyResponse,
+                    },
+                }),
+            ],
+        });
+
+        this.route({
+            method: 'delete',
+            path: '/token/:token',
```
FredrikOseberg, line 154: `/tokens/:token` maybe

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -0,0 +1,252 @@
...
+                    },
+                }),
+            ],
+        });
+
+        this.route({
+            method: 'get',
+            path: '/token/:token',
+            handler: this.getPublicSignupToken,
+            permission: NONE,
```
FredrikOseberg, line 121: Is this permission correct?

src/lib/services/public-signup-token-service.ts
```diff
@@ -0,0 +1,109 @@
...
+        );
+        this.fetchActiveTokens();
+        this.timer = setInterval(
+            () => this.fetchActiveTokens(),
+            minutesToMilliseconds(1),
+        ).unref();
+    }
+
+    async fetchActiveTokens(): Promise<void> {
+        try {
```
FredrikOseberg, line 43: Is this try catch needed? We try catch in the base controller which wraps the services. Seems like this could be omitted.

FredrikOseberg (changes requested):

FredrikOseberg:
Could we also add a bit more detail to the description and link to the roadmap issue?

src/lib/routes/admin-api/public-signup.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    expiresAt: expireAt(),
+});
+
+test('should create a token', async () => {
+    expect.assertions(2);
+    const appName = '123!23';
+
+    stores.clientApplicationsStore.upsert({ appName });
+
+    const bodyCreate = createBody();
```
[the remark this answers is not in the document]
andreas-unleash: Yes, the invite link holds the role - Viewer only atm

src/lib/routes/admin-api/index.ts
andreas-unleash: Better, thanks!

src/lib/db/public-signup-token-store.ts
andreas-unleash: No specific reason other than assuming all db insert actions are wrapped in transactions

src/lib/db/public-signup-token-store.ts
andreas-unleash: True, fixed

src/lib/services/public-signup-token-service.ts
andreas-unleash: Removed. Thanks!

src/lib/db/public-signup-token-store.ts
FredrikOseberg: The way I understand transactions is that it ensures that if you have operations that require separate calls to the db, the transaction is a way to ensure that both actions succeed. If there is only one call to the db, then the transaction is not needed. WDYT @chriswk @sjaanus ?

src/lib/routes/admin-api/public-signup.ts
andreas-unleash: I am not sure about this. Is Admin better?

Tymek: Seems OK to me. We need a public endpoint to check if secret exists, and I don't think there are details that we should hide in this response.

src/lib/db/public-signup-token-store.ts
```diff
@@ -0,0 +1,193 @@
...
+                { user_id: userId, secret },
+                ['created_at'],
+            );
+        });
+    }
+
+    async insert(
+        newToken: IPublicSignupTokenCreate,
+    ): Promise<PublicSignupTokenSchema> {
+        const response = await this.db.transaction(async (tx) => {
```
FredrikOseberg, line 137: I still think we should remove this transaction, it only adds overhead when there is only one call to the database.

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -0,0 +1,252 @@
...
+                    operationId: 'addPublicSignupTokenUser',
+                    requestBody: createRequestSchema('createUserSchema'),
+                    responses: {
+                        200: emptyResponse,
+                    },
+                }),
+            ],
+        });
+
+        this.route({
```
FredrikOseberg, line 117: Think we should add a separate endpoint for validating the token, and add ADMIN to this.

FredrikOseberg (changes requested):
Nice, this is almost ready.

src/lib/routes/admin-api/public-signup.test.ts
```diff
@@ -0,0 +1,211 @@
...
+    });
+
+    test('should return 404 if token is invalid', async () => {
+        const appName = '123!23';
+
+        stores.clientApplicationsStore.upsert({ appName });
+
+        return request
+            .post('/api/admin/invite-link/tokens/some-invalid-secret/validate')
+            .expect(404);
```
FredrikOseberg, line 209: This should be 401. 404 should only be used if the server can't find a resource matching the URL: https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -0,0 +1,289 @@
...
+            method: 'get',
+            path: '/tokens/:token',
+            handler: this.getPublicSignupToken,
+            permission: NONE,
```
FredrikOseberg, line 122: Am I missing something or does this still not have the ADMIN permission?

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    async validate(
+        req: IAuthRequest<TokenParam, void, CreateUserSchema>,
+        res: Response,
+    ): Promise<void> {
+        const { token } = req.params;
+        const valid = await this.publicSignupTokenService.validate(token);
+        if (valid) return res.status(200).end();
+        else return res.status(404).end();
```
FredrikOseberg, line 223: Should be 401

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -0,0 +1,289 @@
...
+                    responses: {
+                        200: emptyResponse,
+                    },
+                }),
+            ],
+        });
+
+        this.route({
+            method: 'post',
+            path: '/tokens/:token/validate',
```
FredrikOseberg, line 172: Could be `/tokens/validate` and accept a body with the token, but this works too.

FredrikOseberg (changes requested):

src/lib/routes/admin-api/public-signup.test.ts
andreas-unleash: fixed

src/lib/routes/admin-api/public-signup.ts
andreas-unleash: thought I did change, fixed

PR #2055, opened 2022-09-12
Fix validate-edge-tokens OpenAPI schema

thomasheartman:
## What

This change fixes the description of the 'tokens' array of the validate edge tokens schema.

The source of the error was a simple inverson of anyOf and items.

## How

Swap `anyOf`and `items` ordering

## Why

So that the swagger editor and openapi tools don't complain about invalid schemas.

image


sjaanus (approved):
LGTM!

PR #2058, opened 2022-09-13
openapi: improve validation testing

thomasheartman:
## What

This PR adds an extra layer of OpenAPI validation testing to what we already have. It also fixes any issues that make the new tests fail.

## Why

While the current OpenAPI validation takes care of _some_ things, there's also things it misses, as shown by #2055. By adding the OpenAPI Enforcer package, we should hopefully be able to catch more of these errors in the future. The enforcer _does_ flag the issue in #2055 as an error.

## How

By adding the OpenAPI Enforcer package and making whatever changes it picks up on.

By adding location headers to all our 201 endpoints. I also had to change some signatures on `create` store methods so that they actually return something (a lot of them just returned `void`).

### Code changes

Some of the code changes may not be necessary or we may want to change more code to align with what changes have been done. It may be worth standardizing on a pattern for `*store.create` methods, so that they always return an identifier or the stored object, for instance.

### On relative URIs

The 201 location headers use relative URIs to point to the created resources. This seemed to be the easiest way to me, as we don't need to worry about figuring out what the absolute URL of the instance is (though we could probably just concat it to the request URI?). The algorithm for determining relative URIs is described in [RFC 3986 section 5](https://www.rfc-editor.org/rfc/rfc3986#section-5).

There's also some places where I'm not sure we _can_ provide accurate location url. I think they're supposed to point _directly at_ whatever the resource is, but for some resources (such as api tokens), you can only get it as part of a collection. From [RFC 9110 on the location field](https://httpwg.org/specs/rfc9110.html#field.location) (emphasis mine):

> the Location header field in a [201 (Created)](https://httpwg.org/specs/rfc9110.html#status.201) response is supposed to provide a URI that is **specific** to the created resource.

A link to a collection is not specific. I'm not sure what best to do about this.

### Inline comments

I've added a number of inline PR comments that I'd love to get some feedback on too. Have a look and let me know what you think!

### Unfinished business

I've added some juicy comments to some of the files here. They contain non-blocking issues that I'm tracking (via github issues). We should resolve them in the future, but right now they can stay as they are.


src/lib/routes/admin-api/feature.ts
```diff
@@ -221,7 +226,7 @@
             req.body,
             userName,
         );
-        res.status(201).json(tag);
+        res.status(201).header('location', `${featureName}/tags`).json(tag);
```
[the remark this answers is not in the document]
thomasheartman: Ooh, this looks interesting. Don't know if we can do anything about this or whether it's an issue here 💁🏼

src/lib/db/context-field-store.ts
```diff
@@ -16,7 +17,16 @@ const COLUMNS = [
...
+type ContextFieldDB = {
+    name: string;
+    description: string;
+    stickiness: boolean;
+    sort_order: number;
+    legal_values: ILegalValue[];
+    created_at: Date;
+};
+
+const mapRow = (row: ContextFieldDB): IContextField => ({
```
thomasheartman, line 29: Apparently, we hadn't typed this at all before. I've added a type that describes the mapping we do.

src/lib/db/context-field-store.ts
```diff
@@ -88,15 +98,17 @@ class ContextFieldStore implements IContextFieldStore {
...
+        const [row] = await this.db(TABLE)
             .insert(this.fieldToRow(contextField))
             .returning('*');
+
         return mapRow(row);
     }
 
     async update(data: IContextFieldDto): Promise<IContextField> {
-        const row = await this.db(TABLE)
+        const [row] = await this.db(TABLE)
```
thomasheartman, line 111: As far as I understand, neither of these worked correctly before. That is, `mapRow` on a list of rows wouldn't work. However, I don't think we ever used the API to get this data back. 

There doesn't seem to be any tests concerned with this, but I would be more than happy to add them. However, I'm not sure exactly where and what would be a good way to test them? Input is very welcome.

src/lib/openapi/spec/create-api-token-schema.ts
```diff
@@ -1,6 +1,22 @@
...
+//
+// We should be able to annotate this using `oneOf` and `allOf`, but making
+// `oneOf` only valid for _either_ `project` _or_ `projects` is tricky.
+//
+// I've opened an issue to get some help (thought it was a bug initially).
+// There's more info available at:
+//
+// https://github.com/ajv-validator/ajv/issues/2096
+//
+// This also applies to apiTokenSchema and potentially other related schemas.
```
thomasheartman, line 18: I'll leave this as is for now. I'll come back and fix this when I go over the entire spec thoroughly.

src/lib/openapi/validate.ts
```diff
@@ -12,13 +11,15 @@ const ajv = new Ajv({
...
-// let's add it back in!
-ajv.addKeyword('example');
+    // example was superseded by examples in openapi 3.1, but we're still on 3.0, so
+    // let's add it back in!
+    keywords: ['example'],
+    formats: {
+        'date-time': true,
+        uri: true,
+    },
+});
```
thomasheartman, line 22: This is just a simplification of how we write it. Instead of using the `add*` methods, we can specify things in the options.

src/lib/routes/admin-api/context.ts
```diff
@@ -189,13 +194,19 @@ export class ContextController extends Controller {
...
         const userName = extractUsername(req);
 
-        await this.contextService.createContextField(value, userName);
-        res.status(201).end();
+        const result = await this.contextService.createContextField(
+            value,
+            userName,
+        );
+        res.status(201)
+            .header('location', `context/${result.name}`) // todo: how to ensure that the location is (and stays) correct?
```
thomasheartman, line 207: This is a discussion point that we should think about. Do we store api paths as variables anywhere? It would be nice to have this auto-update in case things change later.

src/lib/services/openapi-service.ts
```diff
@@ -69,13 +69,18 @@ export class OpenApiService {
...
         const errors = validateSchema(schema, data);
 
         if (errors) {
             this.logger.debug('Invalid response:', errors);
         }
 
+        Object.entries(headers).forEach(([header, value]) =>
+            res.header(header, value),
+        );
+
```
thomasheartman, line 83: Allow headers in `respondWithValidation` to open for correct 201 responses. The optional parameter means nothing breaks, but it might be worth spinning this out into a separate function or something?

package.json
```diff
@@ -135,6 +135,8 @@
   "devDependencies": {
     "@apidevtools/swagger-parser": "10.1.0",
     "@babel/core": "7.18.13",
+    "@swc/core": "1.2.246",
+    "@swc/jest": "0.2.22",
```
sighphyre, line 139: These little guys intentional in this PR?

src/lib/db/context-field-store.ts
sighphyre: Yeah, I'd definitely sleep better at night if we had a test to poke this with a stick. I want to say that you're right but without a strong, static type system it's an educated guess.

Also don't see any tests but the contents of src/test/e2e/stores/ is probably a good model to build some tests off of. I don't think this needs a massively complex test, just a write, update and then a read back to confirm that what you expect has happened. Happy to help if that's not enough info, grab me on Slack

src/lib/routes/admin-api/context.ts
sighphyre: Yeah, this would be really nice! I do think this is a pretty unlikely situation (always nice to be confident though), changing APIs is rare because of compatibility reasons. Don't think we have a set of paths in variables right now, they're mostly set in strings where they're needed. No reason we can't make that happen, just needs time

src/lib/services/openapi-service.ts
sighphyre: I won't complain if you do break this out but if you hadn't mentioned this in a comment it would't have worried me

src/test/e2e/api/openapi/openapi.e2e.test.ts
```diff
@@ -50,6 +51,24 @@ test('the generated OpenAPI spec is valid', async () => {
         await SwaggerParser.validate(body);
     } catch (err) {
         console.error(err);
-        return false;
+        throw err;
```
sighphyre, line 54: Just want to get this clear in my own head - this suggests that errors will now bubble when previously they wouldn't. This doesn't mean that old behaviour in our APIs will now error if it doesn't like the shape of the data, right?

sighphyre (commented):
I think this looks good, I just want to be certain that this isn't going to cause new errors to bubble in cases where they previously wouldn't?

package.json
thomasheartman: Err, no. No, they're not. I'll remove them. Good spot! 🙌🏼

src/test/e2e/api/openapi/openapi.e2e.test.ts
thomasheartman: Yeah, so this is only for tests that we perform. I'm not sure why I wrote `return false` previously. I might've been under the impression that it would fail the test, but I had to cause an exception later on. In this case, though, I should probably swap it with `expect(true).toBe(false)` or something.

But to be clear: this is _not_ runtime code. This is test code. If it breaks, then that means we've discovered errors we didn't know we had, right?

src/lib/db/context-field-store.ts
thomasheartman: Sweet! I'll have a look and see what I can do. Just to be clear: the manual testing I did (which uncovered that it didn't work) was that I tried to use the data that the method returned. Turns out it was all undefined except for `legalValues: []`. So this fixes the mapping, but yeah, I'll look into the testing bit 👷🏼

src/lib/routes/admin-api/context.ts
thomasheartman: Yeah, fair. In that case, I think I'll leave it for now. With how unlikely it is to change, I think it makes sense to leave it as is.

src/lib/services/openapi-service.ts
thomasheartman: Got it! I'll ... I'll see what I feel like 😄 I think it does make sense that you can pass headers regardless of the status code, so maybe it makes sense to keep it?

src/lib/routes/admin-api/feature.ts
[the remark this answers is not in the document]
thomasheartman: @sighphyre Do you have any thoughts around this comment?

package.json
thomasheartman: No wait, hang on! They've just been sorted! Don't know why, but I'm not against it. I'll leave it in.

src/lib/db/context-field-store.ts
thomasheartman: Addressed in 420546ac759abb48c98587661feba6cb800c6077

src/lib/routes/admin-api/feature.ts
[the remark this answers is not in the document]
sighphyre: Hmm, I don't believe this is an issue, you'd need to successfully add a tag for the location header to get sent and for that to work you'd need to send in a feature name as part of the URL, and that feature needs to exist. I think this is okay

sighphyre (approved):
There's a lot here, but it looks good and seems to have some nice tests so... LGTM!

Issue #1391, opened 2022-02-24
:green_circle: feat: OpenAPI all the things.


ivarconr:
We currently have a manually written Open API specification contributed by the community, still considered a "beta feature". We should iterate on this to make it fully part of our API and all changes to the API should validate that the OAS is still correct. 

It should also be aligned with the new "Create screens" in the UI
![image](https://user-images.githubusercontent.com/158948/155476494-4dfb17bb-531d-4390-8fa7-9e32fbdcc179.png)

olav:
We've started integrating OpenAPI directly into our backend. First step: the Admin API.

This first iteration lets us write OpenAPI specs directly in code. These specs are then used in request validation middleware. We also derive request and response types from the specs, which should help with type safety in our handlers. On the frontend we'll experiment with using an OpenAPI client generated from the specs.

The runtime-generated OpenAPI document will be available at /docs/openapi.json, and its Swagger UI at /docs/openapi.The current spec should remain until most endpoints use this new setup.

ivarconr:
We have documented all Open-Source API resources in Q2-2022. In Q3-2022 we will finish the last few enterprise specific endpoints (e.g. configure SSO)

PR #2062, opened 2022-09-14
Fix(#1391): Expose API version even when not running via npm/yarn

thomasheartman:
## What

Expose the current Unleash version in the generated OpenAPI docs, even when running via docker or other processes.

## Why

An OpenAPI spec without a version isn't valid. This causes some of our generation tools (such as the one for documentation) to fail.

## How

By changing how we fetch the current version:

Previously, we used `process.env.npm_package_version!`. However, when you're not running with yarn or npm, this is `undefined`. That causes the version number to not be included when running tests and when running using the official docker image.

Instead, we now use `version` from `lib/util/version`. This is the same version as the one used by the UI config endpoint, so it should be the same one as what the front end displays. To the best of my knowledge, this _is_ the version of the API.

## Discussion

It _may_ be that I have misunderstood what the version represents, but from what I can tell, it just exports what's listed as the version in package.json.

The source code of the `lib/util/version` file is:

```ts
// export module version
require('pkginfo')(module, 'version');

const { version } = module.exports;
export default version;
module.exports = version;
```

### DIFF
--- a/src/lib/openapi/index.ts
+++ b/src/lib/openapi/index.ts
@@ -113,6 +113,7 @@ import { proxyMetricsSchema } from './spec/proxy-metrics-schema';
 import { setUiConfigSchema } from './spec/set-ui-config-schema';
 import { edgeTokenSchema } from './spec/edge-token-schema';
 import { validateEdgeTokensSchema } from './spec/validate-edge-tokens-schema';
+import version from '../util/version';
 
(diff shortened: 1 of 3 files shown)


sighphyre (approved):
LGTM!

thomasheartman:
Also seems to be working with local docker container, so 🤞🏼

PR #2066, opened 2022-09-15
Add OpenAPI docs (#1391)

thomasheartman:
## What

This PR (finally 🎉) adds generated OpenAPI docs to the official Unleash documentation. In addition to generating docs when things get merged to main, it also pushes new doc updates every day at 12:00 AM (cron `@daily`).

## Why

Now that we have OpenAPI'd all the things, we can finally start using it. This will allow us to remove hand-written api docs from the documentation and should make sure everything is always kept up to date.

### Generating from us-hosted (Unleash enterprise)

Unleash has several different versions (open source, pro, enterprise). The versions do not necessarily have the exact same api surface. In fact, the enterprise version has a few endpoints that open source does not.

Because we want to have _all_ endpoints listed in the documentation we need to generated the docs from an enterprise spec. Which brings us into the next point:

### The need for scheduled jobs

Regarding the daily scheduled tasks to update the documentation: why do we need that?

The docs are generated from the tip of the main branch. For most of the docs, this is good and something that we want. However, because the OpenAPI docs are generated from the enterprise edition, it _will not be in sync_ with the open source main branch.

Also, we probably do not want the docs to list the current bleeding edge api changes. Instead, we should prefer to use the latest enterprise release (roughly). However, because we don't get notified when this version is released and deployed, we'll instead run the API generation on a daily cadence.

This isn't the perfect solution, but it's simple and gets us 80% of the way there. More intricate solutions can be set up later.

## How

-   By adding a scheduled workflow to the generate docs config.
-   By adding .gitignore entries for the generated files

There's also some minor changes in styling etc.

## Dependencies

This is dependent on the changes introduced in #2062 having propagated to the enterprise release, which will probably not be for another week or so.

## Discussion

What should the API reference docs url be? I've set it to be `/reference/api/unleash/*` for now, but I'm on the fence about whether it should be `apis` or `api` in there. I also want to get the proxy and other APIs in there as we grow.


gardleopard (approved):
LGTM! :D

gardleopard:
I take that back, it looks fantastic to me! looked at the deployed version <3

PR #2067, opened 2022-09-15
Feat/use collaborate data

FredrikOseberg:
## About the changes
This PR introduces the concept of handling data that is being collaborated on across browsers. We have received reports that users have made mistakes because two or more people have been trying to make changes to feature toggles at the same time, resulting effectively in an overwrite of the data. 

Here we are introducing a reusable hook that will take a copy of the current state of the data and put it into a cache, we then fetch the new data periodically (5s interval currently, though configurable), and do a deep equality check on the data. If there is a mismatch, we'll display a notification that tells the user that the data they are working is currently out of sync, and provide the option to refresh the data.

Skjermbilde 2022-09-15 kl  16 24 10


frontend/src/component/common/StaleDataNotification/StaleDataNotification.tsx
```diff
@@ -0,0 +1,75 @@
+import { Typography, Button, useTheme, useMediaQuery } from '@mui/material';
+import EventDiff from 'component/events/EventDiff/EventDiff';
+import { useThemeStyles } from 'themes/themeStyles';
+import AnimateOnMount from '../AnimateOnMount/AnimateOnMount';
```
nunogois, line 4:
```suggestion
import AnimateOnMount from 'component/common/AnimateOnMount/AnimateOnMount';
```

frontend/src/component/common/StaleDataNotification/StaleDataNotification.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface IStaleDataNotification {
+    refresh: () => void;
+    afterSubmitAction: Function;
```
nunogois, line 8: Would this be more consistent?
```suggestion
    afterSubmitAction: () => void;
```

frontend/src/component/common/StaleDataNotification/StaleDataNotification.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const StaleDataNotification = ({
+    refresh,
+    show,
+    afterSubmitAction,
+    data,
+    cache,
+}: IStaleDataNotification) => {
+    const { classes: themeStyles } = useThemeStyles();
+    const theme = useTheme();
+    const smallScreen = useMediaQuery(`(max-width:${600}px)`);
```
nunogois, line 23:
```suggestion
    const isExtraSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
```

frontend/src/component/common/StaleDataNotification/StaleDataNotification.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    const getStyles = () => {
+        const base = {
+            padding: `${theme.spacing(3)} ${theme.spacing(4)}`,
+            boxShadow: theme.boxShadows.elevated,
+            borderRadius: theme.shape.borderRadiusLarge,
+            backgroundColor: theme.palette.background.paper,
+            maxWidth: '600px',
```
nunogois, line 31:
```suggestion
            maxWidth: theme.spacing(75),
```

frontend/src/component/common/StaleDataNotification/StaleDataNotification.tsx
```diff
@@ -0,0 +1,75 @@
...
+            zIndex: 1000,
```
nunogois, line 32:
```suggestion
            zIndex: theme.zIndex.mobileStepper,
```

frontend/src/component/common/StaleDataNotification/StaleDataNotification.tsx
```diff
@@ -0,0 +1,75 @@
...
+    };
+
+    return (
+        <AnimateOnMount
+            mounted={show}
+            start={themeStyles.fadeInBottomStartWithoutFixed}
+            enter={themeStyles.fadeInBottomEnter}
+            style={getStyles()}
+        >
+            <Typography variant="h5" sx={{ my: 2, mb: 2 }}>
```
nunogois, line 53: We might want to follow a StyledComponent approach by default instead of using `sx`, even though for things like these it's more convenient.

frontend/src/hooks/useCollaborateData.tsx
```diff
@@ -0,0 +1,84 @@
+import { useState, useLayoutEffect, useEffect } from 'react';
+import { SWRConfiguration, SWRResponse } from 'swr';
```
nunogois, line 2: Unused imports: `SWRResponse`, `useLayoutEffect`.

frontend/src/hooks/useCollaborateData.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+const formatUnleashGetter = <Type,>({
+    unleashGetter,
+    dataKey = '',
+    refetchFunctionKey = '',
+    options = {},
+    params = [''],
+}: IGetterOptions): IFormatUnleashGetterOutput<Type> => {
+    const result = unleashGetter(...params, { ...options, refreshInterval: 5 });
```
nunogois, line 18: Would it make sense to have refreshInterval be part of the options passed down from the hook? That way we could have the flexibility to set it on a case-by-case scenario, while still having the default of 5.

nunogois (changes requested):
GJ - Like I mentioned before, this is really cool 😄 
Noticed a few small things we might want to take a second look at before merging.

frontend/src/hooks/useCollaborateData.tsx
FredrikOseberg: Yeah, it should be in the reverse order here. I'll fix that, thanks!

frontend/src/component/common/StaleDataNotification/StaleDataNotification.tsx
FredrikOseberg: I think it's a bit overkill to use a StyledComponent for these kind of styling changes. It's hard to create an abstraction that makes sense to generalise, because it depends on the situation it's used.

nunogois (approved):
LGTM!

PR #2068, opened 2022-09-15
Fix/oas response

andreas-unleash:
This PR fixes a few bugs for the Invite Team members to unleash project. It also adds e2e tests for all endpoints in the PublicSignupController.

So to give some context: When I created the new api for publicSignup:
I missed to create e2e tests.  I tested the controller using the FakeStore implementations, which pretty much hid all the problems with the actual store

It started with fixing the oas response to the getAll endpoint from PublicSignupTokenSchema to PublicSignupTokensSchema.

Bug fix: extracting username from req - Fixes the (wrong) assumption that a `request.user` was present and instead uses the provided method for extracting the username (for populating the createdBy field in the db)

Added e2e test that revealed some store problems that got fixed:

## About the changes

Closes # .


thomasheartman:
Haven't had a look at all of it yet, but did you delete the SDKs index file (from the docs) on purpose?

andreas-unleash:
> Haven't had a look at all of it yet, but did you delete the SDKs index file (from the docs) on purpose?

No, not really sure what happened there but rolled back

src/lib/db/public-signup-token-store.ts
```diff
@@ -31,21 +31,29 @@ interface ITokenUserRow {
...
             users: [],
         };
     }
     const currentToken = acc[tokenRow.secret];
     if (userId) {
-        currentToken.users.push({ userId, name });
+        currentToken.users.push({
+            id: userId,
+            name: userName,
+            username: userUsername,
```
thomasheartman, line 55: Hmm, I can see these two variables being confusing. Would it be better to use `name` and `username`? Or `user` and `username`? I don't know. You've probably thought about this a fair bit yourself, so I'm assuming you've reached the conclusion that `userName` and `userUsername` are the best way to go 💁🏼

src/lib/db/public-signup-token-store.ts
```diff
@@ -163,14 +174,15 @@ export class PublicSignupTokenStore implements IPublicSignupTokenStore {
...
-        const row = await this.makeTokenUsersQuery()
-            .where('secret', key)
-            .first();
+        const rows = await this.makeTokenUsersQuery().where(
+            'tokens.secret',
+            key,
+        );
 
-        if (!row)
+        if (!rows[0])
```
thomasheartman, line 182: If I'm reading this right, this is actually checking that we got at least one result, right? In that case, maybe it'd be more explicit to check the array length instead of the first element?

```suggestion
        if (!rows.length)
```

src/lib/routes/admin-api/public-signup.test.ts
```diff
@@ -92,22 +92,26 @@ describe('Public Signup API', () => {
...
+        const req = request
             .get('/api/admin/invite-link/tokens')
             .expect(200)
             .expect((res) => {
                 const { tokens } = res.body;
                 expect(tokens[0].name).toBe('some-name');
+                expect(tokens[0].createdBy).toBe('johnDoe');
             });
+        req.user = { username: 'johnDoe', name: 'John Doe' };
+        return req;
```
thomasheartman, line 114: I'm probably missing some context here: what are we returning this for? Isn't this a test? Is this called from somewhere else?

src/lib/routes/admin-api/public-signup.ts
```diff
@@ -245,15 +249,16 @@ export class PublicSignupController extends Controller {
         req: IAuthRequest<void, void, PublicSignupTokenCreateSchema>,
         res: Response<PublicSignupTokenSchema>,
     ): Promise<void> {
+        const userName = extractUsername(req);
         const token =
             await this.publicSignupTokenService.createNewPublicSignupToken(
                 req.body,
-                req.user.name,
+                userName,
```
thomasheartman, line 256: Wondering about the capitalization here: is this a user's name or username? The function name reads like it a username, but the variable reads as if it's a user's name. Maybe change the variable to `username` if that's what it is? The same goes for lines 290 and 292

thomasheartman (commented):
Looks mostly good to me! Got a few comments about `username` vs `userName` vs `name`, but nothing blocking. 

However, I don't understand from the description what the bug this fixes is? Could you describe the bug to help me understand why the changes are being made?

src/lib/routes/admin-api/public-signup.ts
andreas-unleash: yes misspelled it

src/lib/routes/admin-api/public-signup.test.ts
andreas-unleash: leftover. cleaned up thanks

thomasheartman:
Nice work! This looks good to me, but I still don't know what problem it fixes, so I don't want to approve it just yet. 
The description says:
> Bug fix: extracting username from req
> Added e2e test that revealed some store problems that got fixed

What's the bug fix? And what problems do the test fix?

src/lib/db/public-signup-token-store.ts
```diff
@@ -30,22 +30,32 @@ interface ITokenUserRow {
...
+            },
             users: [],
         };
     }
     const currentToken = acc[tokenRow.secret];
     if (userId) {
-        currentToken.users.push({ userId, name });
+        currentToken.users.push({
+            id: userId,
+            name: userName,
```
FredrikOseberg, line 56: What is the difference between userName and userUserName? Could we find a better naming?

thomasheartman (approved):
Looks good to me! Thanks for adding the explanation 🤓 🚂

src/test/e2e/api/admin/public-signup-token.e2e.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        app.use('/api/admin/', async (req, res, next) => {
+            const admin = await accessService.getRootRole(RoleName.ADMIN);
+            await userService.createUser({
+                email: 'admin@example.com',
+                username: 'admin@example.com',
+                rootRole: admin.id,
+            });
+            next();
+        });
+    };
```
FredrikOseberg, line 90: Is this needed in this test?

src/test/e2e/api/admin/public-signup-token.e2e.test.ts
```diff
@@ -0,0 +1,223 @@
...
```
FredrikOseberg, line 120: Do we need to create a user for validate?

src/test/e2e/api/admin/public-signup-token.e2e.test.ts
```diff
@@ -0,0 +1,223 @@
...
```
FredrikOseberg, line 144: Same question here, seems like this should be this should be validated by the secret

FredrikOseberg (approved):
Looks good, few comments on the tests.

src/test/e2e/api/admin/public-signup-token.e2e.test.ts
andreas-unleash: I think we need the admin user for the connection to `createdBy` when inserting the token

src/lib/db/public-signup-token-store.ts
andreas-unleash: userName = the name of the user 
userUsername = the username of the user

src/test/e2e/api/admin/public-signup-token.e2e.test.ts
andreas-unleash: It is, the admin user was needed when creating the token

PR #2069, opened 2022-09-16
Personal access token middleware

sjaanus:
Adds security layer, which allows users to use api with personal access tokens.


src/lib/middleware/api-token-middleware.ts
```diff
@@ -45,21 +45,25 @@ const apiAccessMiddleware = (
         }
 
         try {
-            const apiToken = req.header('authorization');
-            const apiUser = apiTokenService.getUserForToken(apiToken);
-            const { CLIENT, FRONTEND } = ApiTokenType;
+            const apiToken = req.header('authorization') as string;
```
nunogois, line 48: Might not need this.
```suggestion
            const apiToken = req.header('authorization');
```

src/lib/middleware/pat-middleware.ts
```diff
@@ -0,0 +1,38 @@
...
+        return (req, res, next) => next();
+    }
+
+    return async (req, res, next) => {
+        if (req.user) {
+            return next();
+        }
+
+        try {
+            const apiToken = req.header('authorization') as string;
```
nunogois, line 24: Same here. If we want this explicitly as string maybe it would make sense to type this in a different way? See e.g. `unleash-types.ts` - Maybe we could extend the headers in a similar way.
```suggestion
            const apiToken = req.header('authorization');
```

nunogois (approved):
LGTM - Minor things we might want to check though.

src/lib/middleware/pat-middleware.ts
sjaanus: Good one. Changed to correct types.

sjaanus:
Added `flagresolver`.

src/lib/middleware/pat-middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+        authentication,
+        flagResolver,
+    }: Pick<IUnleashConfig, 'getLogger' | 'authentication' | 'flagResolver'>,
+    { userService }: any,
+): any => {
+    const logger = getLogger('/middleware/pat-middleware.ts');
+    logger.debug('Enabling PAT middleware');
+
+    if (
+        !authentication.enablePersonalAccessToken ||
```
sjaanus, line 17: Since we are using `flagresolver` now, do we need the extra flag `authentication.enablePersonalAccessToken`?

src/lib/types/experimental.ts
```diff
@@ -10,6 +10,10 @@ export const defaultExperimentalOptions = {
             process.env.UNLEASH_EXPERIMENTAL_EMBED_PROXY,
             false,
         ),
+        personalAccessTokens: parseEnvVarBoolean(
+            process.env.UNLEASH_EXPERIMENTAL_PERSONAL_ACCESS_TOKENS,
+            true,
```
gardleopard, line 15: isnt this supposed to be false by default?

gardleopard (changes requested):
Nice work! Found two things I want you to look at.

sjaanus:
Add test on `api-token-middleware `to check that it goes next when its `user` token.

src/lib/middleware/pat-middleware.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        user: undefined,
+        path: '/api/client',
+    };
+
+    await func(req, undefined, cb);
+
+    expect(cb).toHaveBeenCalled();
+    expect(req.header).toHaveBeenCalled();
+    expect(req.user).toBe(apiUser);
+});
```
chriswk, line 84: Cool tests. Having not been part of the development I can still track what happens. :+1:

src/lib/types/experimental.ts
sjaanus: Yes, fixed

PR #2075, opened 2022-09-19
fix: s/Never logged/Never/g in frontend

gardleopard:
When a user has never been logged in "last login: never" reads better
than "last login: Never logged".

image


thomasheartman (approved):
Agreed! Looks goooood to me 😄

chriswk (approved):
Agreed.

PR #2076, opened 2022-09-19
Fix: prevent password reset email flooding

Tymek:
## About the changes
Related to security, limit "forget password" requests. [Linear 1-242](https://linear.app/unleash/issue/1-242/security-limit-forget-password-requests)

## Discussion points
Do we want to inform users on the frontend about that? I think we kept the info if mail has been sent very vague before.


src/lib/services/user-service.ts
```diff
@@ -400,11 +402,19 @@ class UserService {
         if (!receiver) {
             throw new NotFoundError(`Could not find ${receiverEmail}`);
         }
+        if (this.passwordResetTimeouts[receiver.id]) {
+            return;
```
Tymek, line 406: Works as a bugfix for now. More suggested improvements in: https://linear.app/unleash/issue/1-245/improve-forgotten-password-api

andreas-unleash (approved):
LGTM

PR #2079, opened 2022-09-20
Wip: #1391: fix url issues

thomasheartman:
## Discussion

As of today, the OpenAPI spec uses only the host for the server address. For our hosted solutions, that means that the URLs listed in the API description start with /ushosted or /customer-x

I don't think that's useful. Instead, I think the first path section should be part of the host url, so that all endpoints start with /api or their equivalent part, and not /ushosted or /customer-x

### Examples

This is what our `ushosted` instance looks like. Notice how the "server" is `https://us.app.unleash-hosted.com` and how all API paths start with `/ushosted`
 
image

The same goes for our demo app

image

As far as I understand, the way we do things today, the actual _unleash instance_ is hosted at `/ushosted`, `/demo` or whatever. As such, I think that that path fragment should be added to the server url (in the dropdown at the top) and removed from the path.

From talks with @chriswk and @sighphyre, that should be configurable based on the variables `unleashUrl` and `baseUriPath`.

## Assumptions

Best I can tell, the `baseUriPath` should be _appended to_ the `unleashUrl` to create the full path to Unleash. This seems to be different than what the existing tests assume. I also assume that if `unleashUrl` ends with a string that is equal to `baseUriPath`, then they should _still_ be concatenated.

## What

This PR changes how we handle URLs in the openapi spec. It uses the entire Unleash url as the "url" in the openapi spec

## Why

See the discussion section above.

## How

### DIFF
--- a/src/lib/openapi/index.test.ts
+++ b/src/lib/openapi/index.test.ts
@@ -45,13 +45,22 @@ describe('createOpenApiSchema', () => {
         ).toEqual('https://example.com');
     });
 
-    test('if baseurl is set strips from serverUrl', () => {
+    test('does not strip baseUri from serverUrl', () => {
         expect(
             createOpenApiSchema({
                 unleashUrl: 'https://example.com/demo2',
                 baseUriPath: '/demo2',
             }).servers[0].url,
-        ).toEqual('https://example.com');
+        ).toEqual('https://example.com/demo2');
+    });
+
+    test('adds baseurl serverUrl if it is not already there', () => {
+        expect(
+            createOpenApiSchema({
+                unleashUrl: 'https://example.com',
+                baseUriPath: '/demo2',
+            }).servers[0].url,
+        ).toEqual('https://example.com/demo2');
     });
 
     test('if baseurl does not end with suffix, cowardly refuses to strip', () => {
@@ -74,6 +83,12 @@ describe('createOpenApiSchema', () => {
             createOpenApiSchema({
                 unleashUrl: 'https://example.com/example/',
                 baseUriPath: '/example',
+            }).servers[0].url,
+        ).toEqual('https://example.com');
+        expect(
+            createOpenApiSchema({
+                unleashUrl: 'https://example.com/example/',
+                baseUriPath: '/example/',
             }).servers[0].url,
         ).toEqual('https://example.com');
     });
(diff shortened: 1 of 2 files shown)


chriswk (approved):
LGTM. Feels like the right thing to do :)

thomasheartman:
# PR update: ideas and solutions

So this has been sitting for a bit without a clear solution. I have found a path that seems to work, but it feels wrong, and I have also identified two solutions that may work, but where I would need some help.

@chriswk do you have any thoughts or insights about this?

## Modifying express-openapi

Looking around, I have found a way to modify the API paths that get written by reaching deep into the express-openapi package that we’re using. We have a fork of it, so we could add that change, but it feels a little hacky.

In [generate-doc-js (before line 57)](https://github.com/Unleash/express-openapi/blob/c1ec5cc171f5f39a57c76ffdd13501f02b8db8ef/lib/generate-doc.js#L57-L60), we could add the following snippet:

```js
        if (doc.servers.length === 1) {
            const serverUrl = doc.servers[0].url || ''
            try {
                const serverPath = new URL(serverUrl).pathname
                path = path.replace(serverPath, '')
            }
            catch (e) {
                // ignored
            }
        }
```

This checks if there is only one server defined, and if so, strips the `pathname` from it. This rewrites the paths correctly as far as I can tell

### Drawbacks

While this seems to work in our one, very narrow use case, it will probably break in other use cases. Because it's our own fork, that doesn't matter much, but it does point to the fact that this is probably not the right way to go.

Because it's so deep down in another package, it also feel brittle and is going to be real hard to debug when something inevitably goes wrong later.

It would also add more changes we need to maintain and keep up to date if the express-openapi package that we forked off of changes.

## Mounting the app at a base path

I did some internet searching and found [this issue about setting base urls ](https://github.com/expressjs/express/issues/1611)on express’s page. It seems to suggest that you can mount the app at a subpath by nesting express apps. That said, the issue was opened 9 years ago (and closed 2 years ago). I’m wondering whether that is something we could/should do and if it would help.

This sounds to me like a more "correct" way to solve the issue.

### Drawbacks

I’m afraid it might cause issues with hosting and running in different environments. I don’t know.

## Inform the express router

Another potential solution (if it’s possible) would be to otherwise inform express or its router that the path to the server (such as `ushosted` or `/demo`) shouldn’t be interpreted as a path in requests, but I’m not sure how I’d do that.

### Drawbacks

Similar to mounting the app at a base path, this may cause breakage somewhere that I'm not aware of.

thomasheartman:
Should be solved by https://github.com/Unleash/unleash/pull/3780.

PR #2081, opened 2022-09-23
docs: Move user groups section to after permissions section

thomasheartman:
## What

This PR moves the permissions section of the RBAC documentation and places it before the user groups section. The permissions section was originally a subsection of CPR, and that seems to be where it should remain.

## Why

The permissions section is a subsection of the CPR section and shouldn't be separated. This seems to have been an oversight.

---

That said, I'm happy to take input that says otherwise 😄


sighphyre (approved):
Yeah, agreed, thanks for fixing this

PR #2083, opened 2022-09-23
Updated name and parameters description

Saketsuraj:
## About the changes

Closes # .

### DIFF
--- a/website/docs/advanced/custom-activation-strategy.md
+++ b/website/docs/advanced/custom-activation-strategy.md
@@ -21,7 +21,7 @@ You define custom activation strategies on your Unleash instance, either via the
 
 - A strategy **name**: You'll use this to refer to the strategy in the UI and in code. The strategy name should make it easy to understand what the strategy does. For instance, if a strategy uses contact numbers to determine whether a feature should be enabled, then _ContactNumbers_ would be a good name.
 - An optional **description**: Use this to describe what the strategy should do.
-- An optional list of **parameters**: This is a very important field when defining your own strategy. Use this to provide the strategy with arguments it should use in its evaluation. For eg. you can use it in the code like parameters.strategyname
+- An optional list of **parameters**: The parameter list lets you pass arguments to your custom activation strategy. These will be made available to your custom strategy implementation. How you interact with them differs between SDKs, but refer to the [Node.js example in the how-to guide](../how-to/how-to-use-custom-strategies.md) for a rough idea.
 
 The strategy **name** is the only required parameter, but adding a good **description** will make it easier to remember what a strategy should do. The list of **parameters** lets you pass data from the Unleash instance to the strategy implementation.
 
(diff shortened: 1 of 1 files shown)


thomasheartman:
Hey, @Saketsuraj! 👋🏼  Thanks for the suggestions! I'll have a look at them and get back to you tomorrow 😄

website/docs/advanced/custom-activation-strategy.md
```diff
@@ -19,9 +19,9 @@ Custom activation strategies work exactly like the built-in activation strategie
 
 You define custom activation strategies on your Unleash instance, either via the admin UI or via the API. A strategy contains:
 
-- A strategy **name**: You'll use this to refer to the strategy in the UI and in code.
+- A strategy **name**: You'll use this to refer to the strategy in the UI and in code. For eg. If you want to create a strategy using which a feature should be enabled for only some particular contact numbers then you can keep the name as ContactNumbers.
 - An optional **description**: Use this to describe what the strategy should do.
-- An optional list of **parameters**: Use this to provide the strategy with arguments it should use in its evaluation.
+- An optional list of **parameters**: This is a very important field when defining your own strategy. Use this to provide the strategy with arguments it should use in its evaluation. For eg. you can use it in the code like parameters.strategyname
```
thomasheartman, line 24: Good suggestion, but not all custom strategies need parameters. For some use cases, just the Unleash context will be enough, if they even need that. Also, how you use it in code depends on the programming language and implementation, so this isn't quite the right place to put the example, I think.

```suggestion
- An optional list of **parameters**: The parameter list lets you pass arguments to your custom activation strategy. These will be made available to your custom strategy implementation. How you interact with them differs between SDKs, but refer to the [Node.js example in the how-to guide](../how-to/how-to-use-custom-strategies.md) for a rough idea.
```

thomasheartman (changes requested):
Thanks for the change suggestions! I assume that this was because you had some trouble understanding the points? Could you elaborate a bit on what you found confusing?

Anyway, I've made some suggestions for some slight rewording that I think retains most of the new added information. What do you think?

website/docs/advanced/custom-activation-strategy.md
```diff
@@ -19,9 +19,9 @@ Custom activation strategies work exactly like the built-in activation strategie
 
 You define custom activation strategies on your Unleash instance, either via the admin UI or via the API. A strategy contains:
 
-- A strategy **name**: You'll use this to refer to the strategy in the UI and in code.
+- A strategy **name**: You'll use this to refer to the strategy in the UI and in code. For eg. If you want to create a strategy using which a feature should be enabled for only some particular contact numbers then you can keep the name as ContactNumbers.
```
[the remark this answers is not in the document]
Saketsuraj: The statement written for the 'name' field was not providing me with the importance of using this field and also what should be the suitable name for the strategy. This led to putting extra effort to understand the correct names to use.

website/docs/advanced/custom-activation-strategy.md
Saketsuraj: The rewording is much better now. With the earlier statement, I could not make any sense of it. It was not clear that these parameters will be available during implementation. Now it is quite clear that there are multiple parameters that will be accessible. Also, there is a link to look for implementation directly from this page. Thanks!!!

thomasheartman (approved):
Great work and thanks for the contribution 👏🏼  Thanks for making us better 😄

PR #2085, opened 2022-09-26
Fix openapi snapshot so tests will not fail

sjaanus:
Seems to be related to https://github.com/Unleash/unleash/pull/2084, OpenAPI spec was changed but new snapshot was not generated.


sjaanus:
Super weird. Everything works perfectly locally for me.

thomasheartman (approved):
Looks good to me! Thanks for catching this! 🚂

PR #2087, opened 2022-09-26
fix: support coverage reports on external PRs

gastonfournier:
## About the changes
External PRs fail when trying to upload coverage reports ([example](https://github.com/Unleash/unleash/actions/runs/3037896692/jobs/4894142964)). As explained in this documentation https://github.com/ArtiomTr/jest-coverage-report-action#outputs 
> this approach doesn't work with pull requests from forks without write permission

which is our case.

With this change, external PRs should be able to post the coverage report as a comment in the PR

### DIFF
--- a/.github/workflows/build_prs_jest_report.yaml
+++ b/.github/workflows/build_prs_jest_report.yaml
@@ -38,7 +38,6 @@ jobs:
         with:
           annotations: none
           package-manager: yarn
-          github-token: ${{ secrets.GITHUB_TOKEN }}
           test-script: yarn run test:coverage:jest
           base-coverage-file: ./coverage/report.json
           output: report-markdown
--- a/.github/workflows/build_prs_jest_report.yaml
+++ b/.github/workflows/build_prs_jest_report.yaml
@@ -47,8 +47,8 @@ jobs:
           DATABASE_URL: postgres://postgres:postgres@localhost:5432/postgres
           NODE_ENV: test
           PORT: 4243
-      - name: Report coverage on ${{ matrix.node-version }}
-        uses: marocchino/sticky-pull-request-comment@v2
-        with:
-          # pass output from the previous step by id.
-          message: ${{ steps.coverage.outputs.report }}
+      # - name: Report coverage on ${{ matrix.node-version }}
+      #   uses: marocchino/sticky-pull-request-comment@v2
+      #   with:
+      #     # pass output from the previous step by id.
+      #     message: ${{ steps.coverage.outputs.report }}
--- a/.github/workflows/build_prs_jest_report.yaml
+++ b/.github/workflows/build_prs_jest_report.yaml
@@ -33,6 +33,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'yarn'
       - name: Tests on ${{ matrix.node-version }}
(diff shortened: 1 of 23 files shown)


chriswk (approved):
Looks great :)

.github/workflows/build_prs_jest_report.yaml
```diff
@@ -33,16 +33,23 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'yarn'
       - name: Tests on ${{ matrix.node-version }}
+        id: coverage
         uses: ArtiomTr/jest-coverage-report-action@v2
         with:
           annotations: none
           package-manager: yarn
           github-token: ${{ secrets.GITHUB_TOKEN }}
```
thomasheartman, line 41: I don't have quite enough context here, but could we use the default github.token here as mentioned in the [documentation](https://github.com/ArtiomTr/jest-coverage-report-action#custom-token):

> By default, this action takes github.token variable to publish reports on your PR.

I couldn't quickly find any info on what this token is or what rights it has.

```suggestion
```

thomasheartman (approved):
Looks good to me, but I'm wondering whether `secrets.GITHUB_TOKEN` will work or not. The docs mention that the default uses `github.token` (whatever that is) and the example it has does _not_ specify the token. Any reason why we need the secret token? If not, maybe we could try and remove it?

.github/workflows/build_prs_jest_report.yaml
chriswk: So, This is the one token you don't have access to when you're an external PR. It is a token that has write rights to REPO, which is why you don't want external PRs to have it. So, I do believe that for the markdown-report publish it should work without the token, but I'm not 100%

gastonfournier:
Here we managed to reproduce the issue: https://github.com/Unleash/unleash/pull/2089#issuecomment-1257935418 because I've opened this from an external repository

.github/workflows/build_prs_jest_report.yaml
thomasheartman: Great! In that case, let's try it without. Doesn't work now regardless, so might as well give it a go? @gastonfournier ?

gardleopard: the secret.GITHUB_TOKEN and github.token are equivalent from what I have read. I think we need to skip the usage of this in prs if we can, at least from external forks.

gastonfournier:
External PRs fail due to lack of permissions. It's still possible to keep the tests and coverage but skip the publishing (that now is split into another step)
image

chriswk (approved):
LGTM

thomasheartman (commented):
Nice one! So what's the current state? Will the coverage report run on external PRs or not? Does it still fail? The [action readme](https://github.com/marocchino/sticky-pull-request-comment#error-resource-not-accessible-by-integration) says:

> Error: Resource not accessible by integration
>
> This message means the requester does not have enough permission. If secrets.GITHUB_TOKEN is explicitly passed, this problem can be solved by just removing it.

But we're not passing it anyway 🤔 [This issue](https://github.com/marocchino/sticky-pull-request-comment/issues/298) also has a number of suggestions, which may or may not work and may or may not be worth it.

PR #2088, opened 2022-09-26
Revert usage of pull_request_target

gastonfournier:
## About the changes
This PR precedes https://github.com/Unleash/unleash/pull/2087 so we can first reproduce the issue and then verify the solution proposed by that PR.


chriswk (approved):
LGTM

PR #2093, opened 2022-09-26
fix: client registration events are on eventStore

gardleopard:
Client registration events are on eventStore and not on eventBus. This change makes us have sdk name and version metrics in unleash.


chriswk (approved):
Nice catch, you're absolutely correct. :)

PR #2099, opened 2022-09-27
Remove/update references to Heroku

thomasheartman:
## What

This PR removes or updates references in the docs to Heroku. Most of the code samples have been replaced with a more generic `unleash.example.com` url, while other references have been removed or updated.

Also removes old OpenAPI files that are out of date and redundant with the new generation.

## Background

Come November and Heroku will no longer offer free deployments of Unleash, so it's about time we remove that claim.

Links to the heroku instance are also outdated because we don't have that instance running anymore.

Finally, the OpenAPI files we do have there are old and static, so they don't match the current reality.


frontend/README.md
```diff
@@ -15,7 +15,9 @@ yarn run start
 
 ## Run with a heroku-hosted instance of unleash-api
 
-Alternatively, instead of running unleash-api on localhost, use a remote instance:
+<!-- todo: update this after updating the package.json -->
```
thomasheartman, line 18: The frontend/package.json file has a "start:heroku" command that we should probably update. Once we do, we'll also have to rework this part of the readme, I reckon.

frontend/package.json
```diff
@@ -13,7 +13,7 @@
     "build": "vite build",
     "lint": "eslint src --max-warnings 0",
     "start": "vite",
-    "start:heroku": "UNLEASH_API=https://unleash.herokuapp.com yarn run start",
+    "start:heroku": "UNLEASH_API=https://unleash.herokuapp.com yarn run start", // <— what do we here?
```
thomasheartman, line 16: Will this continue to work?

website/.gitignore
```diff
@@ -20,6 +20,6 @@ yarn-debug.log*
 yarn-error.log*
 
 # OpenAPI docusaurus generated stuff
-docs/reference/apis/**/sidebar.js
+docs/reference/api/**/sidebar.js
```
thomasheartman, line 23: This is due to a last-minute change in where we generate OpenAPI docs. This change simply makes sure the new generated docs get ignored.

website/docs/user_guide/unleash_overview.md
```diff
@@ -13,6 +13,7 @@ Before you can connect your application to Unleash you need a Unleash server. Yo
 1. **Unleash Open-source**
    - [Docker](../deploy/getting-started.md)
    - [Helm Chart](https://github.com/unleash/helm-charts/)
+   <!-- todo: what do we do about this? -->
```
thomasheartman, line 16: Will the "deploy to heroku" button continue to work or should we remove it?

website/docusaurus.config.js
```diff
@@ -179,6 +179,10 @@ module.exports = {
                         from: '/advanced/audit_log',
                         to: '/reference/event-log',
                     },
+                    {
+                        from: '/api/open_api',
+                        to: '/reference/api/unleash',
+                    },
```
thomasheartman, line 185: I deleted an old (unlisted) OpenAPI doc file, so I added a redirect in case anyone has any old links lying around.

frontend/package.json
chriswk: Technically it should work, but I'm guessing we don't want to keep them running, instead possibly update links to the new sandbox instances @gardleopard has set up.

chriswk (approved):
Seems reasonable :)

frontend/package.json
thomasheartman: I'd be happy to do that, yeah! @gardleopard, do you have any preference on which  targets to use?

gardleopard: sanbox.getunleash.ai/ospro

chriswk (approved):
:+1: LGTM

PR #2100, opened 2022-09-28
Fix/frontendapi synchronization

FredrikOseberg:
Introduces a way to sync the frontend api endpoints. Since we instantiate unleash client with a custom repository, that repository needs a way to refresh it's data. Previously we used the event system in unleash to do this, which doesn't work when we have multiple pods/instances running at the same time. The event system will only refresh the cache of the instance that receives the update request. This ensures that all instances refresh the cache on a staggered interval.

[#1875](https://github.com/Unleash/unleash/issues/1875)

### DIFF
--- a/website/docs/reference/front-end-api.md
+++ b/website/docs/reference/front-end-api.md
@@ -47,3 +47,14 @@ The client needs to point to the correct API endpoint. The front-end API is avai
 ### API token
 
 You can create appropriate token, with type `FRONTEND` on `<YOUR_UNLEASH_URL>/admin/api/create-token` page or with a request to `/api/admin/api-tokens`. See our guide on [how to create API tokens](../user_guide/token.mdx) for more details.
+
+### Refresh interval for tokens
+
+Internally, unleash will set up a new unleash client per token it sees and configure the client with the project and environment specified in the token. By default
+the refresh interval is set to 10 seconds, plus a 10 second stagger to avoid thundering herd. This means that if you have 1000 users, unleash will refresh the client data within a window of 10s + a random number between 0 and 10s. This is to avoid all clients hitting the database simultaneously to refresh their data. If you need to change this interval, you can do so by setting the `FRONTEND_API_REFRESH_INTERVAL_MS` environment variable or by setting it directly in the config on startup. Note that the interval is in milliseconds: 
+
+```
+const unleashConfig = {
+    ... // previous values 
+    frontendApi: { refreshIntervalInMs: 4000 }
+}
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -405,6 +405,13 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
         listen = { host: server.host || undefined, port: server.port };
     }
 
+    const frontendApi = options.frontendApi || {
+        refreshIntervalInMs: parseEnvVarNumber(
+            process.env.FRONTEND_API_REFRESH_INTERVAL_MS,
+            10000,
+        ),
+    };
+
     const secureHeaders =
         options.secureHeaders ||
         parseEnvVarBoolean(process.env.SECURE_HEADERS, false);
@@ -449,6 +456,7 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
         import: importSetting,
         experimental,
         flagResolver,
+        frontendApi,
         email,
         secureHeaders,
         enableOAS,
--- a/src/lib/proxy/proxy-repository.ts
+++ b/src/lib/proxy/proxy-repository.ts
@@ -40,7 +40,9 @@ export class ProxyRepository
 
     private segments: Segment[];
 
-    private timer: NodeJS.Timeout;
+    private interval: number;
+
+    private timer: NodeJS.Timer;
 
     constructor(
         config: Config,
@@ -55,6 +57,7 @@ export class ProxyRepository
         this.services = services;
         this.token = token;
         this.onAnyEvent = this.onAnyEvent.bind(this);
+        this.interval = 5000;
     }
 
     getSegment(id: number): Segment | undefined {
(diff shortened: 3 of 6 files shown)


chriswk (approved):
This looks good to me. I see we ended up with setting the timeout before we do the actual updates (I remember why, after having spent an hour with you trying to figure out why it wasn't being called in tests). 

If the update takes a dreadful long time we are risking that we have timeouts coming while still waiting. Though with 10-20 seconds between each call I do believe that should be dreadfully rare. As well as this not really being an issue since the value will be whatever it was set to last, which is fine.

website/docs/reference/front-end-api.md
```diff
@@ -47,3 +47,14 @@ The client needs to point to the correct API endpoint. The front-end API is avai
...
+### Refresh interval for tokens
+
+Internally, unleash will set up a new unleash client per token it sees and configure the client with the project and environment specified in the token. By default
+the refresh interval is set to 10 seconds, plus a 10 second stagger to avoid thundering herd. This means that if you have 1000 users, unleash will refresh the client data within a window of 10s + a random number between 0 and 10s. This is to avoid all clients hitting the database simultaneously to refresh their data. If you need to change this interval, you can do so by setting the `FRONTEND_API_REFRESH_INTERVAL_MS` environment variable or by setting it directly in the config on startup. Note that the interval is in milliseconds: 
+
+```
+const unleashConfig = {
+    ... // previous values 
+    frontendApi: { refreshIntervalInMs: 4000 }
+}
```
thomasheartman, line 60: > Internally, unleash will set up a new unleash client per token it sees 
So if you have a hundred different tokens with the same project and environment, Unleash will create a hundred different clients?

> the refresh interval is set to 10 seconds, plus a 10 second stagger 

Does the stagger offset change every time or is it set for each client? That is, if client A performs a refresh with a 5 second stagger, will the next refresh also be with a 5 second stagger or could it be anything else, such as 2 seconds or 8 seconds?

> If you need to change this interval, you can do so by setting the `FRONTEND_API_REFRESH_INTERVAL_MS` environment variable

Is this documented in the [configuring Unleash article](https://docs.getunleash.io/deploy/configuring_unleash)?
```suggestion
Internally, Unleash creates a new Unleash client for each token it receives. Each client is configured with the project and environment specified in the token.

Each client updates its feature toggle configuration at a specified refresh interval plus a random offset between 0 and 10 seconds. By default, the refresh interval is set to 10 seconds. The random offset is used to stagger incoming requests to avoid a large number of clients all querying the database simultaneously. 

The refresh interval is specified in milliseconds and can be set by using the `FRONTEND_API_REFRESH_INTERVAL_MS` environment variable or by using the `frontendApi.refreshIntervalInMs` configuration option in code.
```

thomasheartman (changes requested):
Nice work! 🙌🏼  Got some minor change suggestions to the docs and some clarification questions if you don't mind. Also, I suspect that we'll need to update the Configuring Unleash doc with the new config options

FredrikOseberg:
> Nice work! 🙌🏼 Got some minor change suggestions to the docs and some clarification questions if you don't mind. Also, I suspect that we'll need to update the Configuring Unleash doc with the new config options

The stagger is randomized each time. I'll update the configuring unleash section aswell.

website/docs/deploy/configuring-unleash.md
```diff
@@ -129,6 +129,13 @@ unleash.start(unleashOptions);
 - **clientFeatureCaching** - configuring memoization of the /api/client/features endpoint
   - `enabled` - set to true by default - Overridable with (`CLIENT_FEATURE_CACHING_ENABLED`)
   - `maxAge` - the time to cache features, set to 600 milliseconds by default - Overridable with (`CLIENT_FEATURE_CACHING_MAXAGE`) ) (accepts milliseconds)
+- **frontendApi** - configuring the refresh interval for when the frontend clients will refresh it's data from the cache. Overridable with `FRONTEND_API_REFRESH_INTERVAL_MS` (accepts milliseconds)
+  ```
+  Example
+  frontendApi: {
+    refreshIntervalInMs: 5000,
+  }
+  ```
```
thomasheartman, line 138: I suggest we do it like this to stay in line with the previous configuration options

```suggestion
- **frontendApi** - Configuration options for the [Unleash front-end API](../reference/front-end-api.md).
  - `refreshIntervalInMs` - how often (in milliseconds) front-end clients should refresh their data from the cache. Overridable with the `FRONTEND_API_REFRESH_INTERVAL_MS` environment variable.
```

website/docs/reference/front-end-api.md
```diff
@@ -47,3 +47,11 @@ The client needs to point to the correct API endpoint. The front-end API is avai
 ### API token
 
 You can create appropriate token, with type `FRONTEND` on `<YOUR_UNLEASH_URL>/admin/api/create-token` page or with a request to `/api/admin/api-tokens`. See our guide on [how to create API tokens](../user_guide/token.mdx) for more details.
+
+### Refresh interval for tokens
+
+Internally, Unleash creates a new Unleash client for each token it receives. Each client is configured with the project and environment specified in the token.
+
+Each client updates its feature toggle configuration at a specified refresh interval plus a random offset between 0 and 10 seconds. By default, the refresh interval is set to 10 seconds. The random offset is used to stagger incoming requests to avoid a large number of clients all querying the database simultaneously. The staggered interval will be randomized on every pull.
```
thomasheartman, line 55:
```suggestion
Each client updates its feature toggle configuration at a specified refresh interval plus a random offset between 0 and 10 seconds. By default, the refresh interval is set to 10 seconds. The random offset is used to stagger incoming requests to avoid a large number of clients all querying the database simultaneously. A new, random offset is used for every update.
```

thomasheartman (changes requested):
Nice! Thanks for updating the configuration doc too 😄 Just a few more suggestions now and I'm happy 🙌🏼

thomasheartman (approved):
Love it! Ready to go, I think 🚂

ivarconr:
Late to the party, but looks good 👍🏼

PR #2105, opened 2022-09-28
fixed segments not being copied

andreas-unleash:
Currently when you tried to copy a strategy (or all strategies) from one env to another, the segments where not being copied as expected.

This PR fixes the bug by adding an optional `copyOf` parameter in the CreateFeatureStrategy schema, so that the handling of copying the segments over is moved to the backend

The need for handling this this way was created by the desired functionality of copying all strategies from one env to another
## About the changes

Closes # .


src/lib/services/feature-toggle-service.ts
```diff
@@ -718,12 +718,19 @@ class FeatureToggleService {
...
+        const segments = await this.segmentService.getByStrategy(strategyId);
+        let result: Saved<IStrategyConfig> = {
             id: strategy.id,
             name: strategy.strategyName,
             constraints: strategy.constraints || [],
             parameters: strategy.parameters,
         };
+
+        if (segments && segments.length > 0) {
+            result = { ...result, segments: segments.map((s) => s.id) };
```
FredrikOseberg, line 731:
```suggestion
            result = { ...result, segments: segments.map((segment) => segment.id) };
```

FredrikOseberg (changes requested):

src/lib/routes/public-invite.ts
```diff
@@ -70,6 +71,8 @@ export class PublicInviteController extends Controller {
                 openApiService.validPath({
                     tags: ['Public signup tokens'],
                     operationId: 'addPublicSignupTokenUser',
+                    summary:
```
FredrikOseberg, line 74: Unsure if the schema updates for public-invite belongs in this PR. Fine for now, but we should aim to separate PR concerns in the future

src/lib/services/feature-toggle-service.ts
```diff
@@ -718,12 +718,22 @@ class FeatureToggleService {
...
             constraints: strategy.constraints || [],
             parameters: strategy.parameters,
         };
+
+        if (segments && segments.length > 0) {
+            result = {
+                ...result,
+                segments: segments.map((segment) => segment.id),
+            };
+        }
```
FredrikOseberg, line 735: Is there a test that validates that this gives the correct result?

FredrikOseberg (changes requested):
Would like a test to see that we return the expected data on the features call

src/lib/services/feature-toggle-service.ts
FredrikOseberg: Also, if there are no segments defined, should we return an empty array?

andreas-unleash: I added one here https://github.com/ivarconr/unleash-enterprise/pull/138

FredrikOseberg: @andreas-unleash It seems this is not reflected in the specification, at least I can't see a spec update for this data structure.

FredrikOseberg (approved):
LG

PR #2106, opened 2022-09-28
Feat/invite user public controller

andreas-unleash:
We need public access urls to validate an invite-link token and add a user
This PR creates a PublicInviteController at the root level to avoid needed authentication in `api/` urls

## About the changes

Closes # .

## Discussion points

Could this be done in a different way?


FredrikOseberg (approved):
LG

PR #2110, opened 2022-09-28
Add test for api-token-middleware to not make database call when using PAT

sjaanus:
Add test for api-token-middleware to not make database call when using PAT


gardleopard (approved):
LGTM remember squashing this :D

PR #2116, opened 2022-09-30
sort feature strategies by created_at first and then sort_order

andreas-unleash:
Currently the query to list featureStrategies is only sorted by sortOrder.  
When we create feature strategies the sort order is not specified.
This creates a bug where editing any other than the first feature strategy, moves it to the top of the list.  Essentially being sorted by updated_at (since sort order is not defined,  all the feature strategies have the same value for sort order)
## About the changes

Fix: Sort by createdAt before sorting with sortOrder in store

Closes # .


FredrikOseberg (approved):
LGTM

PR #2117, opened 2022-09-30
fix: add appName to http response time metrics

ivarconr:
## About the changes
This small fix adds a appName label to the `http_request_duration_milliseconds` Prometheus metrics. It is taken from the `unleash-appname` header, used by all official Unleash SDKs. This makes it easier to debug chatty applications.

### DIFF
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -73,6 +73,7 @@ exports[`should create default config 1`] = `
       "embedProxyFrontend": false,
       "personalAccessTokens": false,
       "publicSignup": false,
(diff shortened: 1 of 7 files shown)


gardleopard (approved):
LGTM, tested locally and we get the appName in. 

Im a bit afraid of getting too many dimensions by this tho. Maybe we can put it behind a flag so that we can test it a bit before letting it out to everyone?

ivarconr:
> Im a bit afraid of getting too many dimensions by this tho. Maybe we can put it behind a flag so that we can test it a bit before letting it out to everyone?

Good point. Added a experimental flag now, so we can enable it only for selected customers in production. When disabled we will always set `appName="undefined"` so it should not create extra dimensions when disabled.

PR #2118, opened 2022-09-30
If baseUriPath is set format it, but if it's not, don't override default with empty string

chriswk:
From Thomas and mine testing. BaseUriPath can't be set as an environment variable because we override it when trying to format the URI from the server config. This PR makes sure we only format if the custom server options actually have baseUriPath set.

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -234,7 +234,9 @@ const formatServerOptions = (
     /* eslint-disable-next-line */
     return {
         ...serverOptions,
-        baseUriPath: serverOptions.baseUriPath ? formatBaseUri(serverOptions.baseUriPath) : undefined,
+        baseUriPath: serverOptions.baseUriPath
+            ? formatBaseUri(serverOptions.baseUriPath)
+            : undefined,
     };
 };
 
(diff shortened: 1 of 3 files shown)


src/lib/create-config.ts
```diff
@@ -232,10 +232,14 @@ const formatServerOptions = (
...
-        baseUriPath: formatBaseUri(serverOptions.baseUriPath),
-    };
+    if(serverOptions.baseUriPath) {
+        return {
+            ...serverOptions,
+            baseUriPath: formatBaseUri(serverOptions.baseUriPath),
+        };
+    } else {
+        return serverOptions;
+    }
```
thomasheartman, line 242: Could this potentially be rewritten like this or wouldn't that work? Might need to edit it if we want `''` as a valid string.

```suggestion
    return {
        ...serverOptions,
        baseUriPath: serverOptions.baseUriPath ? formatBaseUri(serverOptions.baseUriPath) : undefined,
    };
```

thomasheartman (commented):
I think this looks good to me as long as this is the only field we format! 😄 If we need to format more fields, then we should probably rewrite this into a field-level conditional instead (see inline suggestion). But until that's the case, then I'm happy to leave it as is 🙌🏼

One question though: should we add tests for this?

thomasheartman (approved):
I'm happy to get these changes in if you are 😄 But my question from before still stands: I think we should add tests for this, but I'll leave it up to you ultimately.

Oh, and you'll probably need to fix the formatting. GitHub's suggestions aren't auto-formatted by prettier 🙈

chriswk:
I totally agreed with your request for tests, so have added some. It verified that we were letting config take precedence over environment variables, making it harder rather than easier to override BASE_URI_PATH, so have updated the behaviour accordingly.

chriswk:
I'm not too fond of the double ternary here, I'm not sure I'm going to be able to read it when I come back to this code, so will update that to separate calls.

thomasheartman:
@chriswk : any reason to hold off on merging this one or can we get it in? 🚂

PR #2119, opened 2022-09-30
Add some extra fields for profile

sjaanus:
Fix project structure when sending to profile.
Add extra fields for role.


nunogois (approved):
LGTM

PR #2121, opened 2022-09-30
fix: url encode application name in links

gardleopard:
This ensures that we url encode application names for our links.


ivarconr:
Makes sense, but seems to have some failing tests.

gastonfournier (approved):
LGTM!

PR #2124, opened 2022-10-03
#1391: ensure all tags are valid

thomasheartman:
## What

This PR does two things:

1. Add tests to ensure that all OpenAPI tags are listed in the root-level `tags` property of the spec.
2. Changes the tags for endpoints that use any of the deprecated tags (specifically `admin`).

## Why

When we moved to using the docusaurus OpenAPI doc generator, we made some changes to existing tags. Moving away from having just 'admin', 'client', and one or two more, we now use a more granular system, which makes it easier to find an endpoint by category.

However, there's currently nothing enforcing this and this bit of knowledge hasn't been clearly communicated to anyone. Because of this, we have some endpoints that are currently just grouped under a general 'admin' tag.

This has two drawbacks:

-   In Swagger UI, all these endpoints are grouped together. This means that they break with the expectation that each endpoint has its own category. Further, being lumped together means that the 'admin' group is hard to read.
-   The 'admin' tag is (on purpose) not included in the root-level `tags` property of the generated OpenAPI spec. This means that the OpenAPI docusaurus generator doesn't pick up those endpoints, which means that they're not included in the docs on docs.getunleash.io.

## How

By implementing tests that:

1. Check that the tags included in the spec are a subset of the "approved" tags

2. By checking each path in the OpenAPI spec and making sure that all its tags are listed in the root-level tag list. If this is not the case, the test fails and it logs an error explaining what is wrong to the console. An example of such a message is:

```
The OpenAPI spec contains path-level tags that are not listed in the root-level tags object. The relevant paths, operation ids, and tags are as follows:

POST /api/admin/feedback (operation id: createFeedback) has the following invalid tags: "admin"

PUT /api/admin/feedback/{id} (operation id: updateFeedback) has the following invalid tags: "admin"

For reference, the root-level tags are: "Addons", "Admin UI", "API tokens", "Archive", "Auth", "Client", "Context", "Edge", "Environments", "Events", "Features", "Import/Export", "Metrics", "Operational", "Playground", "Projects", "Public signup tokens", "Strategies", "Tags", "Unstable", "Users"
```


src/test/e2e/api/openapi/openapi.e2e.test.ts
```diff
@@ -100,3 +101,91 @@ test('the generated OpenAPI spec is valid', async () => {
...
+test('all root-level tags are "approved tags"', async () => {
+    const { body: spec } = await app.request
+        .get('/docs/openapi.json')
+        .expect('Content-Type', /json/)
+        .expect(200);
+
+    const specTags = spec.tags;
+    const approvedTags = openApiTags;
+
+    expect(specTags).toStrictEqual(approvedTags);
```
gastonfournier, line 114: I wonder if instead of strictly checking this we should check that all specTags are included in approvedTags. Despite, it's good to have this test for housekeeping reasons

src/test/e2e/api/openapi/openapi.e2e.test.ts
```diff
@@ -100,3 +101,91 @@ test('the generated OpenAPI spec is valid', async () => {
...
+            ),
+        );
+
+        // format message
+        const errorMessage = `The OpenAPI spec contains path-level tags that are not listed in the root-level tags object. The relevant paths, operation ids, and tags are as follows:\n${msgs.join(
+            '\n',
+        )} `;
+
+        console.error(errorMessage);
+    }
```
gastonfournier, line 189: A future developer will thank you for this! I'm doing it ahead of time: thanks!

gastonfournier (approved):
Nice one!

src/test/e2e/api/openapi/openapi.e2e.test.ts
thomasheartman: Actually, that's a very good point! Checking that specTags is a subset of approvedTags is probably better 🧠  I'll fix!

src/test/e2e/api/openapi/openapi.e2e.test.ts
thomasheartman: No worries! 😄 And yeah, that's what I figured. Providing an error message that explains why the test fails felt like a sensible thing to do here. Otherwise it might be hard to understand just what's going on.

PR #2127, opened 2022-10-03
fix: use correct column name for correct member count

### LINKED ISSUE #2123: Total members is invalid on Project page in admin web
### Describe the bug

Total members on Projects page are invalid
After the first installation of Unleash server its shows 0 members. It should be 1 member (default admin user)

1. Go to Projects page (/projects)
2. See the total members of a project

### Expected behavior

Should show total members of project

### Logs, error output, etc.

```bash
0
```

### Screenshots

Screen Shot 2565-10-02 at 21 30 46

### Additional context

I thought it was the wrong key to map with the result of getting projects
according to code from src/lib/db/project-store.ts line 274 selects the project name and then adds field memberCount (line 104) into the result by using project id

image
image

### Unleash version

4.15.4

### Subscription type

Open source

### Hosting type

Self-hosted

_No response_

aneeshrelan:
## About the changes

The project name was being used in the query and was being mapped with `id` to get the `memberCount` which resulted in wrong counts

Closes #2123

### DIFF
--- a/src/lib/db/project-store.ts
+++ b/src/lib/db/project-store.ts
@@ -271,7 +271,7 @@ class ProjectStore implements IProjectStore {
                     })
                     .union((queryBuilder) => {
                         queryBuilder
-                            .select('user_id', 'projects.name as project')
+                            .select('user_id', 'projects.id as project')
                             .from('role_user')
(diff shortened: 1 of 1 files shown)


sjaanus (approved):
LGTM!

PR #2131, opened 2022-10-04
Remove member count for default project

sjaanus:
As discussed, for now we will remove member count from default project, because default project is special compared to other projects. Access tab is not available to them and members can not be added or removed


chriswk (approved):
LGTM

PR #2135, opened 2022-10-05
docs: add link to migration guides

thomasheartman:
## What

This adds a short section on Unleash migration and a link to the migration guide to the readme.

## Why

There was some concern that the readme was too focused on new users and that existing users (especially ones who are still on v3) might not find what they need. To try and make their life easier, this PR adds a short section on upgrading and migrating and links to the official guide.


ivarconr (approved):
lgtm

PR #2136, opened 2022-10-05
feat: add event log to project

### LINKED ISSUE #2060: Overall history logs for a project at one place

I would like unleash to have  complete project level logs at one place. its very pain full to keep track of different feature datewise logs. 

### Background

_No response_

### Solution suggestions

_No response_

aneeshrelan:
## About the changes
Adding Event log tab to Project view to show all events within a project at one place

Closes #2060

## Discussion points
Initially thought that a new `useProjectEvents` would be needed, but getting onto the code realised, it might not be. The current `useEventSearch` hook relies on the common event API and if a feature isn't passed to that it provides all project-level events. Just adding a new component and tab on the Project view seemed to work. Tested it with a few cases of search and multiple toggles and projects, seemed to work fine. Not sure if this is all that was needed

### DIFF
--- a/frontend/src/component/project/Project/ProjectLog/ProjectLog.styles.ts
+++ b/frontend/src/component/project/Project/ProjectLog/ProjectLog.styles.ts
@@ -1,0 +1,9 @@
+import { makeStyles } from 'tss-react/mui';
+
+export const useStyles = makeStyles()(theme => ({
+    container: {
+        borderRadius: '12.5px',
+        backgroundColor: theme.palette.background.paper,
+        padding: '2rem',
+    },
+}));
--- a/frontend/src/component/project/Project/ProjectLog/ProjectLog.styles.ts
+++ b/frontend/src/component/project/Project/ProjectLog/ProjectLog.styles.ts
@@ -1,9 +1,0 @@
-import { makeStyles } from 'tss-react/mui';
-
-export const useStyles = makeStyles()(theme => ({
-    container: {
-        borderRadius: '12.5px',
-        backgroundColor: theme.palette.background.paper,
-        padding: '2rem',
-    },
-}));
--- a/frontend/src/component/project/Project/ProjectLog/ProjectLog.tsx
+++ b/frontend/src/component/project/Project/ProjectLog/ProjectLog.tsx
@@ -1,21 +1,23 @@
 import { EventLog } from 'component/events/EventLog/EventLog';
 import { useRequiredPathParam } from 'hooks/useRequiredPathParam';
-import { useStyles } from './ProjectLog.styles';
+import { styled } from '@mui/material';
 
-const ProjectLog = () => {
+const StyledDiv = styled('div')(({ theme }) => ({
+    borderRadius: '12.5px',
+    backgroundColor: theme.palette.background.paper,
+    padding: '2rem',
+}));
+
+export const ProjectLog = () => {
     const projectId = useRequiredPathParam('projectId');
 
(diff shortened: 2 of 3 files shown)


ivarconr:
Hi @aneeshrelan,

thanks for looking in to this. This will work if the user is on the "Admin" role. There are some limitations with the current "Event" api where the general API do require you to be admin. 

I would argue that all members of a project should be able to retrieve the events for the project(s) they are member of. In order to achieve this, I think the appropriate thing would be to establish an event API under the project path. I imagine a new "events" controller inside the [project route folder](https://github.com/Unleash/unleash/tree/main/src/lib/routes/admin-api/project) which would be running at `/api/admin/projects/:projectId/events`. 

We might also want to establish a new project specific permission "PROJECT_EVENTS_READ", and assign that to all members and owners of a project. This way we can prevent "non members" from seeing the project specific events.

aneeshrelan:
Thanks for the clarifications @ivarconr. Let me try to work on these and update

aneeshrelan:
Hi @ivarconr 

> This will work if the user is on the "Admin" role.

I logged in with an Editor and a viewer and still was able to see the events at the project level? What am I missing? It shouldn't have worked, right?

ivarconr:
> I logged in with an Editor and a viewer and still was able to see the events at the project level? What am I missing? It shouldn't have worked, right?

I might have been mistaken here, this goes against the search endpoint for events which should filter out sensitive events. I will make sure we review this PR next week. If we need new/improved APIs we can always change the data-fetching later. 

Thanks for your contribution.

ivarconr (approved):

ivarconr:
@FredrikOseberg do you mind having a look when you have some capacity?

frontend/src/component/project/Project/ProjectLog/ProjectLog.styles.ts
```diff
@@ -0,0 +1,9 @@
+import { makeStyles } from 'tss-react/mui';
+
+export const useStyles = makeStyles()(theme => ({
+    container: {
+        borderRadius: '12.5px',
+        backgroundColor: theme.palette.background.paper,
+        padding: '2rem',
+    },
+}));
```
FredrikOseberg, line 9: @aneeshrelan We're migrating away from this kind of styling, since the migration to MUI 5 this has been deprecated and is kept alive through a module interop. We should either use styled components, or the built in sx propperty in material ui to style. Using useStyles adds technical debt that we need to clean up later.

frontend/src/component/project/Project/ProjectLog/ProjectLog.tsx
```diff
@@ -0,0 +1,21 @@
...
+            <EventLog
+                title="Event Log"
+                project={projectId}
+                displayInline
+            ></EventLog>
+        </div>
+    );
+};
+
+export default ProjectLog;
```
FredrikOseberg, line 21: We do not use default export since it has a high potential for bugs and confusion, all exports should be named exports.

FredrikOseberg (changes requested):
I noted some small things we'll need to adjust in order to be in compliance with our internal ADRs and styling decisions.

frontend/src/component/project/Project/ProjectLog/ProjectLog.tsx
aneeshrelan: Removed the usage of `useStyles` in favour of a styled component

FredrikOseberg (approved):
LGTM

PR #2137, opened 2022-10-05
fix: auto add stratgy when enabling empty env.

ivarconr:
This makes it easier for the user to enable a feature in an env, simplifying UI and reduces number of steps. 

**Before:**
previous

**After:**
new-flow


src/lib/services/feature-toggle-service.ts
```diff
@@ -849,8 +850,10 @@ class FeatureToggleService {
                     environment,
                 );
                 if (strategies.length === 0) {
-                    throw new InvalidOperationError(
-                        'You can not enable the environment before it has strategies',
+                    await this.createStrategy(
```
ivarconr, line 853: this actually bypasses the built-in permission requirement for being allowed to add a strategy in this environment.
I will add a line validating the correct permission.

FredrikOseberg (approved):
Cool. This should also automatically apply the same logic when we perform this action from the project view from what I can tell

PR #2153, opened 2022-10-07
Docs: Impression events only fire on enabled front-end toggles

thomasheartman:
## What

This change updates the description of when impression events are fired to clarify that impression events are **not** fired in front-end SDKs if the toggle is enabled.

It also adds a 'caution' admonition repeating this information with some more info on why.

## Why

We've had a couple questions about this as it hasn't been clearly stated before. Additionally, I, as a member of the team and the original author of this document, thought that impression events **did** fire for disabled toggles in the front-end.

As such, I consider this to be unexpected and different from what was originally written.

### DIFF
--- a/website/docs/advanced/impression-data.md
+++ b/website/docs/advanced/impression-data.md
@@ -8,7 +8,15 @@ Availability The impression data feature was introduced in **Unleash 4.7**. It i
 
 :::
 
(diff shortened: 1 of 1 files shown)


ivarconr:
I would actually have us fix the frontend SDKs instead. We need to allow you to fire impression events for disabled toggles.

thomasheartman:
@ivarconr I agree that we should allow that, absolutely. But until we _do_ fix that, I would like to get this in. And when we _do_ make the SDKs fire on disabled toggles too, we should update it to mention that on older versions of the SDKs (or on Unleash itself?), it'll still only fire on enabled events. What do you think?

ivarconr (approved):

PR #2162, opened 2022-10-10
fix: PAT secret no longer returned (except new), use id instead

nunogois:
https://linear.app/unleash/issue/2-315/adapt-front-end-to-use-id-instead-of-secret-for-the-pats


sjaanus (approved):
LGTM!

PR #2165, opened 2022-10-10
fix: CORS options path

FredrikOseberg:
## About the changes

* Accounts for basepath when setting up options
* Related to: [#1875](https://github.com/Unleash/unleash/issues/1875)

Closes # .


andreas-unleash (approved):
LG

PR #2166, opened 2022-10-10
fix: invite link page docs

Tymek:
Proper link to documentation on "invite link" page.


thomasheartman (approved):
Looks great; good catch!

PR #2167, opened 2022-10-10
fix: PAT not taking into account expires_at

nunogois:
https://linear.app/unleash/issue/2-318/pats-dont-expire


sighphyre (approved):
Nice! LGTM, tested locally, works like a dream

Issue #2104, opened 2022-09-28
inactive toggle showing up in the overview with 'recently seen' times 


rbtcollins:
### Describe the bug

This toggle is not in use by our code, but shows up as having been reported recently: 
![image](https://user-images.githubusercontent.com/499678/192736652-6c6ed78a-90a8-4ba5-a335-bb2b5605da63.png)

as a side note, this is using the rust SDK, which doesn't filter out metrics for toggles that have 0 active and 0 inactive counts (I would argue that that is something the API server should do, if that matters). 

The toggle knows that no requests have occurred though:
![image](https://user-images.githubusercontent.com/499678/192737026-2e24b68c-fb56-4a5c-bca8-c2dcbd518d30.png)

I would expect even with fresh metric uploads, that this would be based on request count - e.g. 'in the last 2 seconds there was at least one request reported' - a definition of seen that depends on 'the API server provided the toggle to the client, and the client provided it back to the API server' will *never* be inactive otherwise.  And I'm strongly suspecting thats happening...

ivarconr:
Ok so this is kind a funny. 

For the metrics itself we actually do filter metrics where both the yes & no counter is 0:
https://github.com/Unleash/unleash/blob/main/src/lib/services/client-metrics/metrics-service-v2.ts#L81

But we do register the toggle as last seen BEFORE we do that filtering, which is just wrong:
https://github.com/Unleash/unleash/blob/main/src/lib/services/client-metrics/metrics-service-v2.ts#L65-L68

I will write a test case for this and improve the implementation to do the filtering first.

PR #2172, opened 2022-10-11
fix: filter empty metrics before we collect last seen toggles.

ivarconr:
This PR fixes two things related to lastSeen details for feature toggles.

1. Only update lastSeen when `yes` or `no` count is larger than 0. 
2. Allow batching of lastSeen updates, hidden behind feature flag named `batchMetrics`. With the flag enabled lastSeen will only be updated every 30th second. 

fixes: #2104

### DIFF
--- a/src/lib/services/client-metrics/last-seen-service.ts
+++ b/src/lib/services/client-metrics/last-seen-service.ts
@@ -44,8 +44,8 @@ export class LastSeenService {
 
     updateLastSeen(clientMetrics: IClientMetricsEnv[]): void {
         clientMetrics
-            .filter((c) => c.yes > 0 || c.no > 0)
-            .forEach((c) => this.lastSeenToggles.add(c.featureName));
+            .filter((clientMetric) => clientMetric.yes > 0 || clientMetric.no > 0)
+            .forEach((clientMetric) => this.lastSeenToggles.add(clientMetric.featureName));
     }
 
     destroy(): void {
--- a/src/lib/services/client-metrics/last-seen-service.ts
+++ b/src/lib/services/client-metrics/last-seen-service.ts
@@ -1,0 +1,54 @@
+import { secondsToMilliseconds } from 'date-fns';
+import { Logger } from '../../logger';
+import { IUnleashConfig } from '../../server-impl';
+import { IUnleashStores } from '../../types';
+import { IClientMetricsEnv } from '../../types/stores/client-metrics-store-v2';
+import { IFeatureToggleStore } from '../../types/stores/feature-toggle-store';
+
+export class LastSeenService {
+    private timers: NodeJS.Timeout[] = [];
+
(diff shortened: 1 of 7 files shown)


src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -62,23 +62,27 @@ export default class ClientMetricsServiceV2 {
...
-        const toggleNames = Object.keys(value.bucket.toggles);
+        const toggleNames = Object.keys(value.bucket.toggles).filter(
+            (name) =>
+                !(
+                    value.bucket.toggles[name].yes === 0 &&
+                    value.bucket.toggles[name].no === 0
+                ),
+        );
         if (toggleNames.length > 0) {
             await this.featureToggleStore.setLastSeen(toggleNames);
```
ivarconr, line 73: this one has proven to be a bit costly to run on every metrics report as it end up doing an entire lock on the features table in the database. I will look in to moving this in to a separate service and optimize it a bit more.

src/lib/services/client-metrics/last-seen-service.ts
```diff
@@ -0,0 +1,54 @@
...
+            );
+            await this.featureToggleStore.setLastSeen(lastSeenToggles);
+        }
+        return count;
+    }
+
+    updateLastSeen(clientMetrics: IClientMetricsEnv[]): void {
+        clientMetrics
+            .filter((c) => c.yes > 0 || c.no > 0)
+            .forEach((c) => this.lastSeenToggles.add(c.featureName));
```
FredrikOseberg, line 48:
```suggestion
            .filter((clientMetric) => clientMetric.yes > 0 || clientMetric.no > 0)
            .forEach((clientMetric) => this.lastSeenToggles.add(clientMetric.featureName));
```

Personally, I prefer full parameters names because it makes it easier for me to scan the code.

src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -62,30 +67,35 @@ export default class ClientMetricsServiceV2 {
...
+            yes: value.bucket.toggles[name].yes,
+            no: value.bucket.toggles[name].no,
+        }));
 
         if (this.config.flagResolver.isEnabled('batchMetrics')) {
             this.unsavedMetrics = collapseHourlyMetrics([
                 ...this.unsavedMetrics,
                 ...clientMetrics,
             ]);
+            this.lastSeenService.updateLastSeen(clientMetrics);
```
FredrikOseberg, line 94: Should this just be the default?

FredrikOseberg (approved):
LG. Added a small suggestion.

src/lib/services/client-metrics/metrics-service-v2.ts
ivarconr: I prefer to have it behind a flag, and the batch flag seems natural.

PR #2173, opened 2022-10-11
docs: add note about custom activation strats in hosted proxies

thomasheartman:
## What

This change adds a note to the custom activation strategies section of the proxy docs, clarifying that the hosted proxy can not use custom activation strategies.

## Why

We had a customer asking whether this was possible and realized it's not stated explicitly anywhere.


nunogois (approved):
LGTM

PR #2183, opened 2022-10-12
feat: Add group sync settings to front end

sighphyre:
Adds front end properties to the OIDC and SAML pages to allow group syncing to be turned on or off by the user. Requires related PR in enterprise to actually save the settings

### DIFF
--- a/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
+++ b/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
@@ -40,7 +40,6 @@ export const OidcAuth = () => {
     const { updateSettings, errors, loading } = useAuthSettingsApi('oidc');
     const ssoSyncHidden = !uiConfig.flags.syncSSOGroups;
 
-
     useEffect(() => {
         if (config.discoverUrl) {
             setData(config);
--- a/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
+++ b/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
@@ -246,7 +246,8 @@ export const OidcAuth = () => {
                     <Grid item md={5}>
                         <strong>Enable Group Syncing</strong>
                         <p>
-                            Enables automatically syncing of users from the OIDC provider when a user logs in.
+                            Enables automatically syncing of users from the OIDC
+                            provider when a user logs in.
                         </p>
                     </Grid>
                     <Grid item md={6} style={{ padding: '20px' }}>
@@ -270,7 +271,8 @@ export const OidcAuth = () => {
                     <Grid item md={5}>
                         <strong>Group Field JSON Path</strong>
                         <p>
-                            Specifies the path in the OIDC token response from which to read the groups the user belongs to.
+                            Specifies the path in the OIDC token response from
+                            which to read the groups the user belongs to.
                         </p>
                     </Grid>
                     <Grid item md={6}>
--- a/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
+++ b/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
@@ -272,7 +272,7 @@ export const OidcAuth = () => {
                     <Grid item md={5}>
                         <strong>Group Field JSON Path</strong>
                         <p>
-                            Specifies the path in the OIDC token response from which to read the groups the user belongs to
+                            Specifies the path in the OIDC token response from which to read the groups the user belongs to.
                         </p>
                     </Grid>
                     <Grid item md={6}>
--- a/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
+++ b/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
@@ -244,7 +244,7 @@ export const OidcAuth = () => {
                     condition={ssoSyncShown}
                     show={
                         <SsoGroupSettings
-                            ssoType={'OIDC'}
+                            ssoType="OIDC"
                             data={data}
                             setValue={setValue}
                         ></SsoGroupSettings>
--- a/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
+++ b/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
@@ -261,9 +261,7 @@ export const OidcAuth = () => {
                                 />
                             }
                             label={
-                                data.enableGroupSyncing
-                                    ? 'Enabled'
-                                    : 'Disabled'
+                                data.enableGroupSyncing ? 'Enabled' : 'Disabled'
                             }
                         />
                     </Grid>
--- a/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
+++ b/frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
@@ -247,7 +247,7 @@ export const OidcAuth = () => {
                             ssoType="OIDC"
                             data={data}
                             setValue={setValue}
-                        ></SsoGroupSettings>
+                        />
                     }
                 />
                 <AutoCreateForm data={data} setValue={setValue} />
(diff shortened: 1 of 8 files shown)


frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -36,6 +38,8 @@ export const OidcAuth = () => {
     const { hasAccess } = useContext(AccessContext);
     const { config } = useAuthSettings('oidc');
     const { updateSettings, errors, loading } = useAuthSettingsApi('oidc');
+    const ssoSyncHidden = !uiConfig.flags.syncSSOGroups;
+
```
nunogois, line 42:
```suggestion

```

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -235,6 +243,53 @@ export const OidcAuth = () => {
                         />
                     </Grid>
                 </Grid>
+                <Grid hidden={ssoSyncHidden} container spacing={3} mb={2}>
+                    <Grid item md={5}>
+                        <strong>Enable Group Syncing</strong>
+                        <p>
+                            Enables automatically syncing of users from the OIDC provider when a user logs in
```
nunogois, line 250:
```suggestion
                            Enables automatically syncing of users from the OIDC provider when a user logs in.
```

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -235,6 +243,53 @@ export const OidcAuth = () => {
...
+                                    : 'Disabled'
+                            }
+                        />
+                    </Grid>
+                </Grid>
+                <Grid hidden={ssoSyncHidden} container spacing={3} mb={2}>
+                    <Grid item md={5}>
+                        <strong>Group Field JSON Path</strong>
+                        <p>
+                            Specifies the path in the OIDC token response from which to read the groups the user belongs to
```
nunogois, line 276:
```suggestion
                            Specifies the path in the OIDC token response from which to read the groups the user belongs to.
```

frontend/src/component/admin/auth/SamlAuth/SamlAuth.tsx
```diff
@@ -246,6 +253,53 @@ export const SamlAuth = () => {
...
+                            name="groupJsonPath"
+                            value={data.groupJsonPath}
+                            disabled={!data.enableGroupSyncing}
+                            style={{ width: '400px' }}
+                            variant="outlined"
+                            size="small"
+                            required
+                        />
+                    </Grid>
+                </Grid>
```
nunogois, line 302: Comments on `OidcAuth.tsx` also apply here. This could also be a good candidate for a separate component so we don't repeat ourselves.

nunogois (changes requested):
LGTM! Just had a few comments we might want to have a look at before merging.
It also seems like some tests failed, running prettier on `OidcAuth.tsx` and `SamlAuth.tsx` should fix those.

frontend/src/component/admin/auth/SsoGroupSettings.tsx
```diff
@@ -0,0 +1,81 @@
...
+                        }
+                        label={data.enableGroupSyncing ? 'Enabled' : 'Disabled'}
+                    />
+                </Grid>
+            </Grid>
+            <Grid container spacing={3} mb={2}>
+                <Grid item md={5}>
+                    <strong>Group Field JSON Path</strong>
+                    <p>
+                        Specifies the path in the {ssoType} token response from
```
sighphyre, line 61: This kind of templated user-facing string literal is going to be a huge pain in the butt when we eventually do i18n but maybe it's okay for now? Or do we have a better solution in the code that I missed?

frontend/src/component/admin/auth/SamlAuth/SamlAuth.tsx
sighphyre: So I've extracted only the new properties and put them behind a conditionally render flag. There are a few common properties between the SSO providers that already exist but right now I'd rather have the new stuff isolated rather than extract working things for the sake of refactoring. I think we should probably beat the SSO provider frontend with a big refactor stick when we do a v5 of Unleash and leave working things alone for now. Happy to be told this is crazy talk though

frontend/src/component/admin/auth/SsoGroupSettings.tsx
nunogois: From my experience templated i18n strings are a common use-case, so I'm sure this will be fine when we get to do i18n: https://kazupon.github.io/vue-i18n/guide/formatting.html

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -36,6 +40,7 @@ export const OidcAuth = () => {
     const { hasAccess } = useContext(AccessContext);
     const { config } = useAuthSettings('oidc');
     const { updateSettings, errors, loading } = useAuthSettingsApi('oidc');
+    const ssoSyncShown = uiConfig.flags.syncSSOGroups || false;
```
nunogois, line 43:
```suggestion
    const ssoSyncShown = Boolean(uiConfig.flags.syncSSOGroups);
```

frontend/src/component/admin/auth/SsoGroupSettings.tsx
sighphyre: Sets my mind at ease! If you're confident I'm confident!

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -235,7 +240,16 @@ export const OidcAuth = () => {
                         />
                     </Grid>
                 </Grid>
-
+                <ConditionallyRender
+                    condition={ssoSyncShown}
+                    show={
+                        <SsoGroupSettings
+                            ssoType={'OIDC'}
```
nunogois, line 247: Could also be a top-level const, or imported from an enum in `SsoGroupSettings`:
```suggestion
                            ssoType="OIDC"
```

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -235,7 +240,16 @@ export const OidcAuth = () => {
...
+                            data={data}
+                            setValue={setValue}
+                        ></SsoGroupSettings>
```
nunogois, line 250:
```suggestion
                        />
```

frontend/src/component/admin/auth/SamlAuth/SamlAuth.tsx
```diff
@@ -246,6 +251,17 @@ export const SamlAuth = () => {
...
+                            ssoType={'SAML'}
+                            data={data}
+                            setValue={setValue}
+                        ></SsoGroupSettings>
+                    }
+                />
+
                 <AutoCreateForm data={data} setValue={setValue} />
                 <Grid container spacing={3}>
                     <Grid item md={5}>
```
nunogois, line 267: Comments from `OidcAuth.tsx` also apply here.

frontend/src/component/admin/auth/SsoGroupSettings.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { Fragment } from 'react';
+import { FormControlLabel, Grid, Switch, TextField } from '@mui/material';
+
+interface SsoGroupSettingsProps {
+    ssoType: string;
```
nunogois, line 5: We could be a bit stricter here if we wanted to, could also be an enum:
```suggestion
    ssoType: 'OIDC' | 'SAML';
```
But at the end of the day, it's only being used as a string for now anyways, so 🤷‍♂️

nunogois (approved):
LGTM! Mostly some nitpicks that you can optionally address, but I vote we merge this soon so we can put everything together and start testing it 👍

frontend/src/component/admin/auth/SamlAuth/SamlAuth.tsx
sighphyre: Done, thanks!

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
sighphyre: I suspect these kind of things can be done in autoformatter/linter config, we should make a plan to include these

Issue #2185, opened 2022-10-12
Length of feature toggle


jandeschuttere:
### Describe the bug

The length of the feature toggle is limited to 100, a constraint that I could not immediately spot in the documentation. Not sure if this can be updated, or if you could explain?

The length in the database is 255 according to https://github.com/Unleash/unleash/blob/e30ff2b822b13ef7bda11f35f8a2b12a20839759/src/migrations/20141020151056-initial-schema.js and the validation is done based on https://github.com/Unleash/unleash/blob/13ef025fab3c330f3e0197e732bb380a838c9125/src/lib/routes/util.ts#L22 which seems to be used for a couple of elements.

I do understand that having large toggle names might be easy for cluttering, url limitations and adds to bandwidth, however in our team we have a use case that could benefit from having them be larger and I'm not sure if the current constraints really are a concern. Is there a chance you could have this updated to what is supported in the db?

### Expected behavior

Have support on length to what is supported by the database

### Unleash version

4.12.6

### Subscription type

Open source

ivarconr:
Hi @jandeschuttere 

thanks for raising this. 

There are a few concerns about the name length:
- we need to make sure it fit in the database (so we cannot have it bigger than the db support, also need to account for potential extra encoding of special chars, etc)
- we need to make sure the UI is able to display it. The current UI has assumptions about the name length. 
- we currently use the name in the url for API calls. There could be limits we need to account for. 

All the above could be addressed of course. 

Could you explain a bit on your use case on why you would need more than 100 chars? Wold be nice with some real world examples on how and why this would be useful.

jandeschuttere:
Our use case is that we have multiple services moving into using unleash as a feature toggle provider.  One of those services acts as a workflow engine. That workflow engine is desired to have some granular control on the flows it acts upon. The way we structure/identify the flags is based on the service and flow it controls together with one or more parts that define logical elements within the flow itself. 

We would prefer not to move into abbreviations as those could make it harder to identify and manage those flags.

thomasheartman:
@jandeschuttere You raise a good point, though! I don't think this _is_ documented anywhere (I'm not sure I was even entirely aware of it myself), but it definitely belongs in the feature toggle docs. I'll update that, at least. 🙋🏼

ivarconr:
Thanks for clarifying @jandeschuttere.

We have no strong preference that the feature max length has to be exactly 100 chars. But changing this will require some validations on our end:

[...]

PR #2188, opened 2022-10-13
docs: Document the min and max lengths of a feature name.

thomasheartman:
## What

This change adds a the min and max lengths allowed for a feature toggle name to the feature toggle docs.

## Why

As was mentioned in #2185, this isn't documented anywhere as of today, making it a potentially surprising issue to come across.


sighphyre (approved):
LGTM

PR #2191, opened 2022-10-13
fix: PATs should have an unique description (per user)

nunogois:
https://linear.app/unleash/issue/2-327/pats-dont-have-a-unique-description

### DIFF
--- a/frontend/src/component/user/Profile/PersonalAPITokensTab/CreatePersonalAPIToken/CreatePersonalAPIToken.tsx
+++ b/frontend/src/component/user/Profile/PersonalAPITokensTab/CreatePersonalAPIToken/CreatePersonalAPIToken.tsx
@@ -164,7 +164,7 @@ export const CreatePersonalAPIToken: FC<ICreatePersonalAPITokenProps> = ({
 
     const isDescriptionEmpty = (description: string) => description.length;
     const isDescriptionUnique = (description: string) =>
-        !tokens?.filter(token => token.description === description).length;
+        !tokens?.some(token => token.description === description);
     const isValid =
         isDescriptionEmpty(description) && isDescriptionUnique(description);
 
--- a/frontend/src/component/user/Profile/PersonalAPITokensTab/CreatePersonalAPIToken/CreatePersonalAPIToken.tsx
+++ b/frontend/src/component/user/Profile/PersonalAPITokensTab/CreatePersonalAPIToken/CreatePersonalAPIToken.tsx
@@ -93,7 +93,7 @@ export const CreatePersonalAPIToken: FC<ICreatePersonalAPITokenProps> = ({
     setOpen,
     newToken,
 }) => {
-    const { refetchTokens } = usePersonalAPITokens();
+    const { tokens, refetchTokens } = usePersonalAPITokens();
     const { createPersonalAPIToken, loading } = usePersonalAPITokensApi();
     const { setToastApiError } = useToast();
     const { uiConfig } = useUiConfig();
@@ -103,6 +103,11 @@ export const CreatePersonalAPIToken: FC<ICreatePersonalAPITokenProps> = ({
     const [expiration, setExpiration] = useState<ExpirationOption>(
         ExpirationOption['30DAYS']
     );
+    const [errors, setErrors] = useState<{ [key: string]: string }>({});
+
+    const clearErrors = () => {
+        setErrors({});
+    };
 
     const calculateDate = () => {
         const expiresAt = new Date();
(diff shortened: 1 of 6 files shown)


chriswk:
Sorry, why does description need to be unique across all users?

thomasheartman:
Hmm, Chris has a good point there. I think making descriptions unique per user is useful, but across users would probably quickly cause issues when two users want to have the same "local development" description 🤔

src/lib/services/pat-service.ts
```diff
@@ -55,6 +55,20 @@ export default class PatService {
         return this.patStore.deleteForUser(id, userId);
     }
 
+    async validatePat({ description, expiresAt }: IPat): Promise<void> {
+        if (!description) {
+            throw new BadDataError('PAT description cannot be empty');
+        }
+
+        if (new Date(expiresAt) < new Date()) {
+            throw new Error('The expiry date should be in future.');
```
sighphyre, line 64: I know you've just moved this to here and it's not really part of this PR but this error type causes the API to freak out, would it cause trouble if we change this to a BadDataError instead? 

```suggestion
            throw new BadDataError('The expiry date should be in future.');
```

sighphyre (changes requested):
Looks good!

Feel free to tell me I'm crazy with the suggestion, just want eyes on it before we accidentally merge

nunogois:
@chriswk @thomasheartman - Yep, you guys are absolutely right, I'll filter them by user.

frontend/src/component/user/Profile/PersonalAPITokensTab/CreatePersonalAPIToken/CreatePersonalAPIToken.tsx
```diff
@@ -226,6 +250,7 @@ export const CreatePersonalAPIToken: FC<ICreatePersonalAPITokenProps> = ({
                             type="submit"
                             variant="contained"
                             color="primary"
+                            disabled={!isValid}
```
thomasheartman, line 253: I'm always a little on on the fence about disabled buttons due to accessibility concerns (see [this CSS-tricks article](https://css-tricks.com/making-disabled-buttons-more-inclusive/) for why), so I'm wondering whether we can make this more accessible somehow?

src/test/e2e/api/admin/user/pat.e2e.test.ts
```diff
@@ -149,10 +152,72 @@ test('should fail creation of PAT with passed expiry', async () => {
...
+    await app.request
+        .post('/api/admin/user/tokens')
+        .send({
+            expiresAt: tomorrow,
+        } as IPat)
+        .set('Content-Type', 'application/json')
+        .expect(400);
+});
+
+test('should fail creation of PAT with a description that already exists', async () => {
```
thomasheartman, line 172: I think this could be a little more explicit about the context

```suggestion
test('should fail creation of PAT with a description that already exists for the current user', async () => {
```

thomasheartman (commented):
Nice work! Just got some minor comments that I'd like some input on before approving. Depending on how you feel, there may be some changes to implement.

frontend/src/component/user/Profile/PersonalAPITokensTab/CreatePersonalAPIToken/CreatePersonalAPIToken.tsx
sighphyre: This is a pattern that we're using all throughout the frontend. We can definitely make this better but I don't think we should do that in this PR. Very open to suggestions @thomasheartman but it's more work than just this piece of work

thomasheartman: Yeah, that's fair. I'm aware that we do this other places (got an [old trello card for it](https://trello.com/c/ppggUlV6/297-dont-use-the-disabled-attribute-on-buttons), so I'm on board with that. But I figured that when we introduce new changes, we might want to shift and gradually shift over 💁🏼  Very open to being overruled, though 😄

sighphyre: Ye, I don't mind doing a gradual shift over, it's a pretty good idea actually! But before we do, I kinda want a plan to know what we want to achieve and how; and we should do that on features that can sit for a while - PATs are on our critical path for some internal stuff so I'd like to try to keep the work as focused as possible

nunogois: I think this is an interesting discussion and thanks for sharing the article!
If `MUI` default behaviour does not cover this accessibility requirement (apparently not) then I agree that this belongs in a separate PR that pinpoints the exact behaviour that we want, creates a component that implements it, and probably changes every button to use our new component - Or we follow a gradual shift approach after that PR, which I think is more reasonable.

chriswk (approved):
Looks good now to me

sighphyre (approved):
LGTM!

PR #2194, opened 2022-10-14
Syncing external groups with unleash group

sjaanus:
During user login, user groups will be synced with any external groups.


chriswk (approved):
Looks really good

PR #2195, opened 2022-10-14
fix: sandbox codemirror/state 6.1.2

nunogois:
Seems like some `codemirror` dependencies were requesting @codemirror/state v6.1.1 while others requested v6.1.2 - This PR adds a resolution so we use `6.1.2`, fixing the issue.


FredrikOseberg (approved):
LG

PR #2197, opened 2022-10-17
Fix all groups being removed, even when no external groups were defined

sellinjaanus:
Fixed all groups being removed, even when no external groups were defined
Fixed a bug, where SSO payload was wrong and caused crash.
Added more tests


sighphyre (approved):
LG

src/test/e2e/services/group-service.e2e.test.ts
```diff
@@ -69,3 +72,28 @@ test('should add person to completely new group with new name', async () => {
     expect(groups.length).toBe(1);
     expect(groups[0].name).toEqual('dev_group');
 });
+
+test('should not update groups when not string array ', async () => {
+    await groupService.syncExternalGroups(user.id, 'Everyone' as any);
+    const groups = await groupService.getGroupsForUser(user.id);
+    expect(groups.length).toBe(1);
```
chriswk, line 79: `expect(groups).toHaveLength(1)` if you want it to flow better. It does exactly the same behind the scenes so I'm fine with either. Just thought it was nice to know of the API

chriswk (approved):
Nice catch

PR #2198, opened 2022-10-17
PAT: add "never", "custom" options to expiry date

nunogois:
https://linear.app/unleash/issue/2-330/feedback-from-visma-pats-with-no-expirycustom-expiry
Front-end only, assumes "never" as a **very long** time from now.


chriswk (approved):
Hah, seems a bit risky, but I guess it does cover never, cause noone will be around in 1000 years.

PR #2201, opened 2022-10-18
Add suggest-change-store.ts and db migration

andreas-unleash:
## About the changes

Creates the suggest-change-set table and starting store functionality `create` and `get`

Closes # .


src/lib/db/suggest-change-store.ts
```diff
@@ -0,0 +1,123 @@
...
+    private db: Knex;
+
+    constructor(db: Knex, eventBus: EventEmitter, getLogger: LogProvider) {
+        this.db = db;
+        this.eventBus = eventBus;
+        this.logger = getLogger('lib/db/suggest-change-store.ts');
+    }
+
+    async create(
+        suggestChangeSet: PartialSome<ISuggestChangeSet, 'id'>,
```
thomasheartman, line 49: I had to go look this type up, so I might not have it completely correct. Does this make the `id` property optional on the incoming changeset? In that case, what happens to the `id` further down? If it's not there, does the DB set one automatically?

src/lib/types/model.ts
```diff
@@ -365,3 +365,24 @@ export interface IFeatureStrategySegment {
...
 }
+
+export interface ISuggestChangeSet {
+    id: number;
+    environment: string;
+    state?: string;
+    project?: string;
+    createdBy?: string;
+    createdAt?: Date;
+    changes?: ISuggestChange[];
```
thomasheartman, line 376: I thought (at least) some of these options were required: definitely `project`, but probably also `changes` and the `created*` options. Why are they made optional here?

src/lib/types/stores.ts
```diff
@@ -60,4 +61,5 @@ export interface IUnleashStores {
     segmentStore: ISegmentStore;
     patStore: IPatStore;
     publicSignupTokenStore: IPublicSignupTokenStore;
+    suggetChangeStore: ISuggestChangeStore;
```
thomasheartman, line 64: This looks like a little typo 🦅 

```suggestion
    suggestChangeStore: ISuggestChangeStore;
```

thomasheartman (commented):
Nice work! Looks like a good start to me, but I've got a couple of questions regarding typing. Would you mind having a look? Also, there doesn't seem to be any tests included at the moment: is there nothing to test?

src/lib/db/suggest-change-store.ts
andreas-unleash: yes the db should add it

src/lib/types/model.ts
andreas-unleash: resolved, leftover from developing. thanks!

src/lib/db/suggest-change-store.ts
thomasheartman: Thanks for the clarification!

src/lib/db/suggest-change-store.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    prefixColumns(): string[] {
+        return COLUMNS.map((c) => `${Tables.suggestChangeSet}.${c}`);
+    }
+
+    mapRow(row?: ISuggestChangeSetRow): ISuggestChangeSet {
+        if (!row) {
+            throw new NotFoundError('No row');
+        }
+
```
thomasheartman, line 93: Out of curiosity: How do we usually handle cases like this? I'm assuming you're following the same conventions that we use elsewhere in the code base, so it's probably perfectly on brand.

On a personal level, though, I prefer validating values before they're being sent into functions, so that function don't have to check for invalid arguments. So the function would be like this instead:

```suggestion
    mapRow(row: ISuggestChangeSetRow): ISuggestChangeSet {
```

That said, this is more to facilitate a discussion than to point to something we need to change. I also think there's value in keeping with patterns that we use elsewhere in the code, so there's that 💁🏼

src/lib/db/suggest-change-store.ts
```diff
@@ -0,0 +1,123 @@
...
+            .select(this.prefixColumns())
+            .from(Tables.suggestChangeSet)
+            .orderBy('id', 'asc');
+
+        return rows.map(this.mapRow);
+    }
+
+    async get(id: number): Promise<ISuggestChangeSet> {
+        const rows: ISuggestChangeSetRow[] = await this.db
+            .select(this.prefixColumns())
```
thomasheartman, line 78: Does knex have a `first` method or something? DB connectors I've used in the past tend to have a function to get the first or unique item for a given query (or ID) instead of a bunch of rows.

src/lib/services/suggest-change-service.ts
```diff
@@ -0,0 +1,62 @@
...
+    async get(id: number): Promise<ISuggestChangeSet> {
+        return this.suggestChangeStore.get(id);
+    }
+
+    async delete(id: number): Promise<void> {
+        return this.suggestChangeStore.delete(id);
+    }
+}
+
+module.exports = SuggestChangeService;
```
thomasheartman, line 62: Do we use module.exports in the code base? I can't remember seeing that elsewhere. Is there a reason you both `export default class` and `module.exports`?

src/lib/types/events.ts
```diff
@@ -82,6 +82,8 @@ export const PUBLIC_SIGNUP_TOKEN_CREATED = 'public-signup-token-created';
 export const PUBLIC_SIGNUP_TOKEN_USER_ADDED = 'public-signup-token-user-added';
 export const PUBLIC_SIGNUP_TOKEN_TOKEN_UPDATED = 'public-signup-token-updated';
 
+export const SUGGEST_CHANGE_CREATED = 'suggest-change-created';
```
thomasheartman, line 85: Nice! Maybe I can ask you to add this to the docs too? A short description of when it fires and what it is in `website/docs/api/admin/events-api.md`? Along with a note that it is an experimental feature? (I'll make it so that we can feature gate it in the future!)

thomasheartman:
Nice work! I've got some more questions if you don't mind 😇

src/migrations/20221810114644-add-suggest-changes-table.js
```diff
@@ -0,0 +1,28 @@
...
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+CREATE TABLE IF NOT EXISTS suggest_change_set (
+  id serial primary key,
+  environment varchar(255) NOT NULL,
+  state varchar(255) NOT NULL,
+  project varchar(255) NOT NULL,
+  changes text NOT NULL,
```
FredrikOseberg, line 11: Should this be a JSON B column?

src/lib/services/suggest-change-service.ts
```diff
@@ -0,0 +1,62 @@
...
+        }: Pick<IUnleashStores, 'suggestChangeStore' | 'eventStore'>,
+        config: IUnleashConfig,
+    ) {
+        this.config = config;
+        this.logger = config.getLogger('services/suggest-change-service.ts');
+        this.suggestChangeStore = suggestChangeStore;
+        this.eventStore = eventStore;
+    }
+
+    async createChangeRequest(
```
FredrikOseberg, line 31: Do we need a method to record events after it has been created? Does creating a changeRequest always require a changeset?

FredrikOseberg (changes requested):
Initial draft makes sense. I've added some comments / questions for clarification purposes.

src/migrations/20221810114644-add-suggest-changes-table.js
FredrikOseberg: Or should change events be modeled as it's own table?

src/lib/services/suggest-change-service.ts
```diff
@@ -0,0 +1,62 @@
...
+        return this.suggestChangeStore.getAll();
+    }
+
+    async get(id: number): Promise<ISuggestChangeSet> {
+        return this.suggestChangeStore.get(id);
+    }
+
+    async delete(id: number): Promise<void> {
+        return this.suggestChangeStore.delete(id);
+    }
```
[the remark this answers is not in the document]
thomasheartman: I don't think so: Nested promises collapse, so that should be functionally the same, right?

src/lib/db/index.ts
```diff
@@ -32,6 +32,7 @@ import SegmentStore from './segment-store';
 import GroupStore from './group-store';
 import PatStore from './pat-store';
 import { PublicSignupTokenStore } from './public-signup-token-store';
+import SuggestChangeStore from './suggest-change-store';
```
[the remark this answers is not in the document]
andreas-unleash: sure, ok

src/migrations/20221810114644-add-suggest-changes-table.js
andreas-unleash: Could definitely be it's own table, I think that makes more sense

src/lib/services/suggest-change-service.ts
andreas-unleash: A change request requires a changeSet, yes

src/lib/services/suggest-change-service.ts
```diff
@@ -0,0 +1,79 @@
+import { IUnleashConfig, IUnleashStores } from '../types';
```
FredrikOseberg, line 1: The service should move to enterprise

andreas-unleash: The reason is here is to keep the convention (see SegmentService)

src/migrations/20221810114644-add-suggest-changes-table.js
```diff
@@ -0,0 +1,47 @@
...
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+CREATE TABLE IF NOT EXISTS suggest_change_set (
+  id serial primary key,
+  environment varchar(255) NOT NULL,
+  state varchar(255) NOT NULL,
+  project varchar(255) NOT NULL,
+  created_by varchar(255) NOT NULL,
```
sjaanus, line 11: Should we use a foreign key for `user` instead if we want to strongly link the change sets to actual user. In future, we are not displaying the creator just as string, but we will need to run queries on it. Having a `created_by: user.username || user.email` will make it error prone and hard to query.

sjaanus (changes requested):

src/lib/db/suggest-change-store.ts
```diff
@@ -0,0 +1,266 @@
+import { ISuggestChangeStore } from '../types/stores/suggest-change-store';
+import { Logger, LogProvider } from '../logger';
+import EventEmitter from 'events';
+import { Knex } from 'knex';
+import { PartialSome } from '../types/partial';
+import {
+    ISuggestChange,
+    ISuggestChangeEvent,
+    ISuggestChangeEventData,
+    ISuggestChangeSet,
```
Tymek, line 10: `Changeset` should be cammelCased like `ISuggestChangeset`

src/lib/types/model.ts
```diff
@@ -365,3 +365,66 @@ export interface IFeatureStrategySegment {
...
+    id: string;
+    flow: IApprovalFlow;
+}
+
+export interface ISuggestChange {
+    id: number;
+    action: SuggestChangeAction;
+    feature: string;
+    payload?: unknown;
+    createdBy?: string;
```
Tymek, line 392: Is it linked with `User`? How do we store info about who created it, because UI has user name and avatar. Maybe some metadata on this can solve it.

src/lib/db/suggest-change-store.ts
```diff
@@ -0,0 +1,266 @@
...
+    ): Promise<ISuggestChangeSet[]> => {
+        const rows = await this.makeSuggestChangeSetChangesEventsQuery().where({
+            environment,
+        });
+
+        return this.mapRows(rows);
+    };
+
+    get = async (id: number): Promise<ISuggestChangeSet> => {
+        const rows = await this.makeSuggestChangeSetChangesEventsQuery().where({
```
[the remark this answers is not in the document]
andreas-unleash: Wouldn't first only get 1 row ?- There might be many rows based on the connections, that will need to be aggregated

src/migrations/20221810114644-add-suggest-changes-table.js
andreas-unleash: makes sense

src/lib/services/suggest-change-service.ts
FredrikOseberg: I suspect this is an oversight, and represents technical debt that we should move to enterprise also. We do have a setup for services in enterprise. 

@ivarconr

src/migrations/20221810114644-add-suggest-changes-table.js
andreas-unleash: I went through the migrations, almost everywhere `created_by` is set to a text field

FredrikOseberg: I guess it depends on the intent of the table, there may not be a lot of other tables where we need to query for drafts per user, where createdat essentially is metadata and the id is the foreign key.

src/lib/db/suggest-change-store.ts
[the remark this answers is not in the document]
sjaanus: You are correct, checked the database query and seems reasonable.

src/lib/db/suggest-change-store.ts
```diff
@@ -0,0 +1,293 @@
...
+                created_at: suggestChangeSet.createdAt,
+                created_by: user.username || user.email,
+            })
+            .returning('id');
+
+        suggestChangeSet.changes.forEach((change) => {
+            this.addChangeToSet(change, id, user);
+        });
+
+        return this.get(id);
```
FredrikOseberg, line 227: I think we should return the entire changeSet object from the create, this will allow the frontend to save a call and use the returning value instead.

src/lib/types/model.ts
```diff
@@ -365,3 +365,66 @@ export interface IFeatureStrategySegment {
...
+export interface ISuggestChangeset {
+    id: number;
+    state: string;
+    project: string;
+    environment: string;
+    createdBy?: Pick<User, 'username' | 'imageUrl'>;
+    createdAt?: Date;
+    changes?: ISuggestChange[];
+    events?: ISuggestChangeEvent[];
+    // approvalFlow: IChangeSetApprovalFlow;
```
FredrikOseberg, line 378: intentional?

FredrikOseberg (changes requested):
Nice work, this is very close. Minor comment.

website/docs/api/admin/events-api.md
```diff
@@ -1557,3 +1557,24 @@ This event fires when you delete a segment.
...
+  "createdBy": "user@company.com",
+  "createdAt": "2022-06-03T10:30:08.128Z",
+  "data": {},
+  "preData": null,
+  "tags": [],
+  "featureName": null,
+  "project": null,
+  "environment": null
+}
+```
```
thomasheartman, line 1580: Nice! Thank you! Is this what the actual event will look like? Does it contain any data or anything? I'd expect there to be a link to the suggestion that was created, at least? This example should show more or less exactly what the users would see when creating a change suggestion.

src/lib/types/model.ts
```diff
@@ -365,3 +365,66 @@ export interface IFeatureStrategySegment {
...
 }
+
+export interface ISuggestChangeset {
+    id: number;
+    state: string;
+    project: string;
+    environment: string;
+    createdBy?: Pick<User, 'username' | 'imageUrl'>;
+    createdAt?: Date;
+    changes?: ISuggestChange[];
```
thomasheartman, line 376: Can you have a change set without changes? Or a change set with an empty list of changes for that matter? If not, maybe it makes sense to make it a required, non-empty list? That is, something like `changes: [ISuggestChange, ...ISuggestChange[]]`?

src/lib/types/model.ts
```diff
@@ -365,3 +365,66 @@ export interface IFeatureStrategySegment {
...
     segmentId: number;
 }
+
+export interface ISuggestChangeset {
+    id: number;
+    state: string;
+    project: string;
+    environment: string;
+    createdBy?: Pick<User, 'username' | 'imageUrl'>;
+    createdAt?: Date;
```
thomasheartman, line 375: Why are these two properties optional? Are there any situations where we'll have a change set without knowing who made it or when?

src/lib/types/model.ts
```diff
@@ -365,3 +365,66 @@ export interface IFeatureStrategySegment {
...
+    flow: IApprovalFlow;
+}
+
+export interface ISuggestChange {
+    id: number;
+    action: SuggestChangeAction;
+    feature: string;
+    payload?: unknown;
+    createdBy?: Pick<User, 'username' | 'imageUrl'>;
+    createdAt?: Date;
```
thomasheartman, line 386: Does it make sense to have a change without a payload? Maybe there's cases where it's not necessary (like enabling/disabling a toggle)? And as with the change set: why are createdBy and createdAt optional?

src/lib/types/model.ts
```diff
@@ -365,3 +365,66 @@ export interface IFeatureStrategySegment {
...
+    UPDATE_STRATEGY = 'updateStrategyEvent',
+    DELETE_STRATEGY = 'deleteStrategyEvent',
+}
+
+export interface ISuggestChangeEvent {
+    id: number;
+    event: SuggestChangesetEvent;
+    data: ISuggestChangeEventData;
+    createdBy?: Pick<User, 'username' | 'imageUrl'>;
+    createdAt?: Date;
```
thomasheartman, line 417: This is another instance where these fields are listed as optional. Nothing new to add; just wanted to create a comment as a marker 🤓

thomasheartman (commented):
Looks good! Thanks a lot for the docs 🙌🏼  Got a couple questions regarding optional properties, that I've marked inline and some further clarification regarding the event doc, if you don't mind 😄

src/lib/db/suggest-change-store.ts
```diff
@@ -249,6 +250,9 @@ export class SuggestChangeStore implements ISuggestChangeStore {
             { feature: change.feature, data: change.payload },
             user,
         );
+        if (shouldReturn) {
+            return this.get(changeSetID);
+        }
```
thomasheartman, line 255: Why do we need to make this conditional? Can we not just return the change set regardless and ignore it, if not?

thomasheartman: If it's for performance reasons, is there any way to encode that it will always be returned when you pass `shouldReturn`? Seems like a hassle to have to check whether a function returned `void` or an actual object 💁🏼

website/docs/api/admin/events-api.md
andreas-unleash: Would you add the id of the change request or the whole request?

src/migrations/20221810114644-add-suggest-changes-table.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+CREATE TABLE IF NOT EXISTS suggest_change_set (
+  id serial primary key,
+  environment varchar(255) NOT NULL,
+  state varchar(255) NOT NULL,
+  project varchar(255) NOT NULL,
```
sjaanus, line 10: Should project and environment be also foreign keys? What will happen to suggest changes when project is deleted?

src/test/fixtures/fake-suggest-change-store.ts
```diff
@@ -0,0 +1,75 @@
...
+    ): Promise<void> {
+        const changeSet = this.suggestChanges.find((s) => (s.id = changeSetID));
+        changeSet.changes.push(change);
+        return Promise.resolve();
+    }
+
+    getForEnvironment(environment: string): Promise<ISuggestChangeset[]> {
+        return Promise.resolve(
+            this.suggestChanges.filter(
+                (changeSet) => (changeSet.environment = environment),
```
sjaanus, line 37: Am I missing something or this should be `===` and also for other methods  ?

src/lib/types/model.ts
andreas-unleash: They were made optional because they are added when commiting to the db - I can create more objects (SuggestChangeCreate and SuggestChangesetCreate and SuggestChangeEventCreate) that do not have them, if you think that this is a problem.

website/docs/api/admin/events-api.md
thomasheartman: Whatever we want to put in the event. How have you designed the event? Having both would provide the maximum flexibility, but I don't know how large these events would be. At the very least, you should be able to go and see the resource that was created, right?

src/lib/types/model.ts
andreas-unleash: see below

src/lib/types/model.ts
andreas-unleash: see below

src/lib/types/model.ts
thomasheartman: Got it. But when it comes out of the DB, it'll always have this data, right? In that case, my vote goes to using more types that are more specific (using TS's fancy type system). That way, you'll always know you have  createdBy/createdAt data when it's coming out of the DB 💁🏼 But I'm willing to be challenged on this.

src/migrations/20221810114644-add-suggest-changes-table.js
andreas-unleash: Adding it makes sense however, we are already at 5 joins for the query, I am reluctant to add more

src/lib/db/suggest-change-store.ts
thomasheartman: I'm not sure I understand: you need the change set returned in both cases you mentioned over? Why would you then return `void`? Or maybe I misunderstood what you meant.

src/test/fixtures/fake-suggest-change-store.ts
andreas-unleash: Good catch, thanks

src/lib/db/suggest-change-store.ts
andreas-unleash: In the `create` method this code adds the individual changes. I don't want to call get after each add - I want to call it when the whole `create` method is done 
```
suggestChangeSet.changes.forEach((change) => {
            this.addChangeToSet(change, id, user, false);
        });
```

thomasheartman: Ah, I think I see now; thanks for explaining 😄  Yeah, this doesn't sound like anything this function should care about to me. You already need the change set ID to add the change regardless, so if we think it would cause issues to always return the updated change set, then I'd personally vote to leave it up to the caller after updating.

FredrikOseberg: Agree with Thomas on this one

src/migrations/20221810114644-add-suggest-changes-table.js
FredrikOseberg: We can always performance test this and make changes later? Having to manually clean up if an environment / project gets deleted sounds like a nightmare to me.

src/lib/db/suggest-change-store.ts
andreas-unleash: Removed - moved the return to the service layer

src/migrations/20221810114644-add-suggest-changes-table.js
sjaanus: Yea, I don't think the performance hit will be something remarkable. Those joins on foreign keys are quite fast.

src/test/fixtures/fake-suggest-change-store.ts
```diff
@@ -0,0 +1,75 @@
...
+        return Promise.resolve(undefined);
+    }
+
+    addChangeToSet(
+        change: ISuggestChange,
+        changeSetID: number,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        user: Partial<Pick<User, 'username' | 'email'>>,
+    ): Promise<void> {
+        const changeSet = this.suggestChanges.find((s) => (s.id = changeSetID));
```
sjaanus, line 29: This needs `===`

src/migrations/20221810114644-add-suggest-changes-table.js
chriswk: Provided you add index on elements that you do FK joins on

src/migrations/20221810114644-add-suggest-changes-table.js
```diff
@@ -0,0 +1,47 @@
...
+  created_by integer not null references users (id) ON DELETE CASCADE,
+  created_at timestamp default now(),
+  updated_by integer not null references users (id) ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS suggest_change (
+    id serial primary key,
+    action varchar(255) NOT NULL,
+    payload json NOT NULL,
```
chriswk, line 19: We really should be using jsonb.

src/migrations/20221810114644-add-suggest-changes-table.js
```diff
@@ -0,0 +1,47 @@
...
+    created_by integer not null references users (id) ON DELETE CASCADE,
+    created_at timestamp default now(),
+    suggest_change_set_id integer NOT NULL REFERENCES suggest_change_set(id) ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS suggest_change_event (
+    id serial primary key,
+    event varchar(255) NOT NULL,
+    data json NOT NULL,
```
chriswk, line 28: Same thing, please use jsonb unless you have really good reason for not doing so.

src/migrations/20221810114644-add-suggest-changes-table.js
sjaanus: Primary keys are indexed by default.

src/migrations/20221810114644-add-suggest-changes-table.js
sjaanus: Good idea

src/migrations/20221810114644-add-suggest-changes-table.js
sjaanus: Good idea

src/lib/db/suggest-change-store.ts
```diff
@@ -0,0 +1,318 @@
...
+            )
+            .leftJoin(`users as changUser`, 'change.createdBy', 'changUser.id')
+            .leftJoin(
+                `${T.SUGGEST_CHANGE_EVENT} as event`,
+                'changeSet.id',
+                'event.suggest_change_set_id',
+            )
+            .leftJoin(`users as eventUser`, 'eventUser.id', 'event.createdBy')
+            .select(
+                'changeSet.secret',
```
sjaanus, line 155: What is this secret? I dont see it in migrations.

sjaanus (approved):
LGTM!

FredrikOseberg (changes requested):
Looks good, though we should get rid of the console.logs before merging.

Issue #2203, opened 2022-10-18
Bug: Cannot create addon as an Editor


ivarconr:
### Describe the bug

As editor I cannot create a new addon. This seems like a UI bug as I can do it via the API. 

![image](https://user-images.githubusercontent.com/158948/196427831-0d6ab509-fbe1-4ebd-892b-b54af8252109.png)

### Steps to reproduce the bug

1. Log in as "editor"
2. Try to create a new "slack addon",

PR #2204, opened 2022-10-18
fix: should only require CREATE_ADDON when creating addon

ivarconr:
fixes: #2203

### DIFF
--- a/frontend/src/component/addons/AddonForm/AddonForm.tsx
+++ b/frontend/src/component/addons/AddonForm/AddonForm.tsx
@@ -334,7 +334,7 @@ export const AddonForm: VFC<IAddonFormProps> = ({
                             type="submit"
                             color="primary"
                             variant="contained"
-                            permission={ADMIN}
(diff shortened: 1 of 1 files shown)


frontend/src/component/addons/AddonForm/AddonForm.tsx
```diff
@@ -334,7 +334,7 @@ export const AddonForm: VFC<IAddonFormProps> = ({
                             type="submit"
                             color="primary"
                             variant="contained"
-                            permission={ADMIN}
+                            permission={editMode ? 'UPDATE_ADDON' : 'CREATE_ADDON'}
```
nunogois, line 337: You can switch the import to grab the `CREATE_ADDON` and `UPDATE_ADDON` constants instead, so we don't have to use string literals: 
image

ivarconr: will do!

nunogois (approved):
LGTM

sjaanus (approved):
LGTM!

PR #2206, opened 2022-10-19
fix: respect environment if set on context

FredrikOseberg:
This PR fixes an issue with the new frontend API where we would override environment set on the context to what the environment of the API key was. This will break feature toggles for older users of unleash because some might still be using the deprecated environment field on the SDKs.

## About the changes

Closes # .

### DIFF
--- a/src/lib/proxy/create-context.test.ts
+++ b/src/lib/proxy/create-context.test.ts
@@ -1,6 +1,6 @@
 // Copy of https://github.com/Unleash/unleash-proxy/blob/main/src/test/create-context.test.ts.
 
-import { createContext } from './create-context';
+import { createContext, enrichContextWithIp } from './create-context';
 
 test('should remove undefined properties', () => {
     const context = createContext({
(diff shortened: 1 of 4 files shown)


src/lib/proxy/create-context.test.ts
```diff
@@ -55,13 +55,23 @@ test('will not blow up if properties is an array', () => {
...
+test('will respect environment set in context', () => {
+    const context = createContext({
+        userId: '123',
+        tenantId: 'some-tenant',
+        environment: 'development',
+        region: 'eu',
+        properties: ['some'],
+    });
+
+    expect(context.environment).toBe('development');
```
ivarconr, line 72: I would like to see another test proving environment will not be `development` when not set on context.

src/lib/proxy/create-context.test.ts
```diff
@@ -55,13 +55,34 @@ test('will not blow up if properties is an array', () => {
...
+
+test('will not set environment to be development if not set in context', () => {
+    const context = createContext({
+        userId: '123',
+        tenantId: 'some-tenant',
+        region: 'eu',
+        properties: ['some'],
+    });
+
+    expect(context.environment).toBe(undefined);
```
ivarconr, line 83: why will it be `undefined` in this test when we sets it to `req.user.envrionment`?

ivarconr: sorry, you are testing the "create-context" here. I thought this was an API test for some reason..

ivarconr: ...meaning: this test is not relevant for the changes code lines.

ivarconr (approved):

Issue #2186, opened 2022-10-12
Import on startup disables environments


knightsg:
### Describe the bug

When an existing configuration is imported into Unleash server (free version) on startup using the "drop" behaviour, both the development and production environments are disabled even if they are set to 'enabled' in the source configuration file. I've tried removing the environments key from the json source config but it doesn't fix the issue. Only the 'default' environment is enabled, and you have to specifically go to the enviroments tab for the default project and enabled both enviroments manually.

To contrast with the default startup behaviour, if you start the server without importing any configuration then the default environment is disabled and both the development and production environments are enabled.

### Steps to reproduce the bug

1. Prepare an existing configuration file using the export API, and make sure the 'enabled' attribute for both the development and production enviroments is true.
2. Start a copy of unleash server using the IMPORT_FILE to specify the exported configuration and the IMPORT_DROP_BEFORE_IMPORT flag set to 'true'.
3. Go to 'Projects > Default > Environments' and the 'default' environment will be enabled while the 'development' and 'production' environments will be disabled.

### Expected behavior

After a clean server starts up with IMPORT_FILE and IMPORT_DROP_BEFORE_IMPORT set, the environment toggle should be set to the value specified in the source import file.

### Unleash version

4.15.3

### Subscription type

Open source

### Hosting type

Self-hosted

andreas-unleash:
Hi @knightsg , thanks for bringing this to our attention. I will bring it up with the team.

thomasheartman:
Hey, @knightsg! I think this should have been fixed by #2209, so I'm gonna go ahead and close this issue now. If you think there's more to be done, please feel free to reopen this. Thanks again for reporting it ☺️

PR #2209, opened 2022-10-19
Import of feature still showing env on feature, when environment is disabled on project

sellinjaanus:
Should fix the issue https://github.com/Unleash/unleash/issues/2186.

Changes:

1. When importing **featureEnvironments**, it created `feature_environments` always, now only if `project_environments` exists(is enabled).
2. When importing **projects**, it created a `project_environment` only for default project, now creates a `project_environment` for each new project and environment that is new.


chriswk (approved):
This seems like a good fix. I must admit changing the count for environments from 1 to 4 in one of the tests baffled me until I saw the rest of the change.

PR #2211, opened 2022-10-19
Feat/stats service

ivarconr:
## About the changes
This feature introduces instance stats to Unleash. 

image

This makes it easier for admins to get an overview of the Unleash installation and also offer a download CSV file function, to make it shareable (e.g. with support) 

All metrics are also exposed as Prometheus metrics for convenience.


src/lib/services/instance-stats-service.ts
```diff
@@ -0,0 +1,185 @@
...
+            strategies,
+            SAMLenabled,
+            OIDCenabled,
+        };
+    }
+
+    async getSignedStats(): Promise<InstanceStatsSigned> {
+        const instanceStats = await this.getStats();
+
+        const sum = sha256(
```
sighphyre, line 180: So I think this is okay for now but how secure do we want to be with this?

When I read `signed`, I think cryptographically secure signature. Do we want to go to those lengths here? This would technically let people read the source and generate these themselves. This interface would let us do that later though

src/lib/db/role-store.ts
```diff
@@ -141,6 +148,15 @@ export default class RoleStore implements IRoleStore {
             .orWhere('type', 'project');
     }
 
+    async getProjectRolesCount(): Promise<number> {
+        return this.db
+            .from<IRole>(T.ROLES)
+            .count('*')
+            .where('type', 'custom')
+            .orWhere('type', 'project')
```
sighphyre, line 156: Do we want to include the project roles here? If I'm not going crazy, you can't add or remove them so it might just be noise here

sighphyre (commented):
I have a question about the signatures but I think this this looks great!

src/lib/services/instance-stats-service.ts
ivarconr: I am aware that this is not secure at all. 

The problem is that in case of self-hosted there is little way for us to secure it entirely without doing some kind of network call back to us for a secure signature, but then we could just send the data directly from Unleash. If we base it on anything available locally, one can always trick Unleash to show incorrect numbers, you may even remove all users and re-add them again. 

The thinking I had is that users stupid enough to change the CSV file would not be smart enough to re-create correct hash. Smart users will probably not change this anyway, and if they do, they will be able to trick the numbers regardless.

src/lib/db/role-store.ts
ivarconr: Actually I already do this, so this code is not used at all. I will remove it for simplicity.

frontend/src/component/admin/instance-admin/InstanceStats/InstanceStats.tsx
```diff
@@ -0,0 +1,82 @@
...
+                <Table aria-label="Instance statistics">
+                    <TableHead>
+                        <TableRow>
+                            <TableCell>Field</TableCell>
+                            <TableCell align="right">Value</TableCell>
+                        </TableRow>
+                    </TableHead>
+                    <TableBody>
+                        {rows.map(row => (
+                            <TableRow >
```
FredrikOseberg, line 59: This should have a key, technically since you can't mutate the state and the order is guaranteed it doesn't matter. But it would be good to promote best practices.

frontend/src/hooks/api/getters/useInstanceStats/useInstanceStats.ts
```diff
@@ -0,0 +1,58 @@
...
+export interface IInstanceStatsResponse {
+    stats?: InstanceStats;
+    refetchGroup: () => void;
+    loading: boolean;
+    error?: Error;
+}
+
+export const mapGroupUsers = (users: any[]) =>
+    users.map(user => ({
+        ...user.user,
```
FredrikOseberg, line 33: user.user reads a bit weird to me. Is there another word to explain what the parameter is?

frontend/src/component/admin/instance-admin/InstanceAdmin.tsx
```diff
@@ -0,0 +1,11 @@
+import AdminMenu from '../menu/AdminMenu';
+import { InstanceStats } from './InstanceStats/InstanceStats';
+
+export const InstanceAdmin = () => {
+    return (
+        <div>
```
FredrikOseberg, line 6: Should InstanceStats be available for everyone in the system? Currently we don't gate the stats for Admins only.

frontend/src/component/admin/instance-admin/InstanceStats/InstanceStats.tsx
```diff
@@ -0,0 +1,82 @@
...
+        {title: versionTitle, value: version},
+        {title: 'Users', value: stats?.users},
+        {title: 'Feature toggles', value: stats?.featureToggles},
+        {title: 'Projects', value: stats?.projects},
+        {title: 'Environments', value: stats?.environments},
+        {title: 'Roles', value: stats?.roles},
+        {title: 'Groups', value: stats?.groups},
+        {title: 'Context fields', value: stats?.contextFields},
+        {title: 'Strategies', value: stats?.strategies},
+    ];
```
FredrikOseberg, line 38: Hmm. Prettier seems to not be working correctly here.

src/lib/db/role-store.ts
```diff
@@ -141,6 +148,15 @@ export default class RoleStore implements IRoleStore {
...
+            .then((res) => Number(res[0].count));
```
FredrikOseberg, line 157: I'm curious about this. I guess since it's count it always returns a row? But not every knex method does this, as far as I know. Sometimes this causes problems because people try to copy what they have seen elsewhere, and expect it to work the same. Not sure if we should do something about it, but I'm raising the point.

src/lib/routes/admin-api/instance-admin.ts
```diff
@@ -0,0 +1,84 @@
...
+            path: '/statistics/csv',
+            handler: this.getStatisticsCSV,
+            permission: NONE,
+        });
+
+        this.route({
+            method: 'get',
+            path: '/statistics',
+            handler: this.getStatistics,
+            permission: NONE,
```
FredrikOseberg, line 44: I assume this is intentional.

FredrikOseberg (changes requested):
Nice. Some minor details and clarifications, otherwise looks super clean!

frontend/src/component/admin/instance-admin/InstanceStats/InstanceStats.tsx
ivarconr: Thanks 👍🏼 

Maybe prettier could warn about this?

frontend/src/hooks/api/getters/useInstanceStats/useInstanceStats.ts
ivarconr: copy/paste 🤕

frontend/src/component/admin/instance-admin/InstanceAdmin.tsx
ivarconr: That was the intention, all users with access should be able to access it for now, as there is not secrets here. All stats will be implicitly available for read only users anyway (number of projects, feature toggles etc).

frontend/src/component/admin/instance-admin/InstanceStats/InstanceStats.tsx
ivarconr: I have not gotten my IDE to work correctly with prettier after the merge. This is how it looks after auto-format of this file.

src/lib/db/role-store.ts
ivarconr: In knex a `count()` will return a result of type: `Record<string, number | string>`. So we know it will be a single row with a "key" named "count" and the value to be a `number` or a stringed number (`42` vs `"42"`).

src/lib/routes/admin-api/instance-admin.ts
ivarconr: yes

ivarconr:
I have addressed all feedback @FredrikOseberg

src/lib/services/instance-stats-service.ts
sighphyre: Yeah, that makes sense and I agree with the reasoning here, even with a direct API call there's nothing in theory from stopping someone from just cURLing back whatever data they want.

I think this keeps honest users honest and trying to fight smart, malicious people with root access to Unleash isn't worth it. I'm happy!

FredrikOseberg (approved):
LGTM!

PR #2217, opened 2022-10-20
Implement service wide transactions

sighphyre:
# The What

This implements a way to create service level transactions in Unleash. We currently have a few transactions in the stores, but they're a bit painful and fragile to use across methods and bridging them across stores/services means writing a lot of ugly code. We're definitely at the point where a few upcoming features could really use a proper way of doing transactions in a cleaner way (and a few existing features could use a redesign to use transactions). This also uses the new pattern to replace transactions in the group store and move them up to the service level as a sample of what this looks like so that other developers can evaluate the API here.

# Details

To add transactions to a given store requires a few things, this section will walk through what's required and what the relevant chunks of code do.

### Extend Transactor on your store (or implement the interface yourself)

To make transactions work on a given store, you'll need to extend the store's class signature to opt into the new Transactor abstract class like this:

```js
export default class GroupStore
    extends Transactor<GroupStore> // the new stuff, note the abstract class needs to know the type that it's extending
    implements IGroupStore
{
// implementation details    
}
```
You can check out the Transactor class for more details but effectively what this does is implement a new method, `transactable(transaction: Knex.Transaction): T`, to clone the store in a way that keeps the TS compiler happy, while still providing full access to all the instance methods. The new store instance will have its Knex instance swapped out for a transaction instance, which behaves in an identical way to the Knex instance, except that queries run against this instance will be buffered into a transaction. That method can then be used like this:

```js
let newGroups = await this.groupStore.transactional(trx).getNewGroupsForExternalUser(userId, externalGroups);
```
Instead of previously like this:
```js
let newGroups = await this.groupStore.getNewGroupsForExternalUser(userId, externalGroups);
```
Note that is opt in so you don't _have_ to use this in a transactional way. The transactional method is type safe so you can't accidentally pass something else into it.

You'll also need to mark the interface method as `Transactional<T>`, so for example the Group store now looks like this:

```js
export interface IGroupStore
    extends Store<IGroup, number>,
        Transactional<IGroupStore> {
//interface details
}
```
Since the store class now extends an abstract class and you can only extend a single class in TS, this provides an escape hatch for instances of the stores where they already extend a class, the `Transactor` abstract class doesn't do any magic, it's purely for not forcing people to write a ton of these methods each time. For example the `EventStore` already extends `AnyEventEmitter` and so this store gets its own implementation of the interface method (you can check the source for details, but it's not complicated to implement). 

### Receive a Knex instance in your service and use it

Your service will now need to accept a Knex instance, that's just including it in your constructor and keeping it as a field. Most of the work here is already done and the bulk of this PR is just updating test services to take a Knex instance. You'll need to actually start a transaction and consume it, but the API here is fairly nice. You can manage the transaction yourself if you need but there's no obvious need to do that. A simplified example, derived from the actual code:

```js
 async updateGroup(group: IGroupModel, userName: string): Promise<IGroup> {
        const preData = await this.groupStore.get(group.id);
        await this.validateGroup(group, preData); //runs outside of the transaction
        
        return this.db.transaction(async (tx) => { //return the results of the block, in this case a Promise<IGroup>
            const newGroup = await this.groupStore
                .transactional(tx) //runs in the transaction
                .update(group);

            const newUsers = await this.groupStore
                .transactional(tx) //runs in the transaction
                .getAllUsersByGroups([group.id]);
            const existingUserIds = existingUsers.map((g) => g.userId);

            await this.groupStore.transactional(tx).updateGroupUsers( //runs in the transaction
               newGroup.id, newUsers
            );
            await this.eventStore.transactional(tx).store({ //also runs in the same transaction in a different store
                type: GROUP_UPDATED,
                createdBy: userName,
                data: newGroup,
                preData,
            });
            return newGroup; //return the outputs to the top level await, in this case an IGroup
        });
    }
```
If the transaction fails, it'll be automatically rolled back and a Knex error will be raised to the caller. You can optionally wrap the async block in a try/catch, get access to the error and manage things yourself (you can check out the Knex docs on transactions for that) but again, likely no need.

When the top level await completes, the transactional clones will fall off the stack and be garbage collected.

### Optional but recommended - swap your store db field type out for a Union type

While the compiler is happy without it, it's recommended to swap your db field in your store out for a Union type of 

`private db: Knex | Knex.Transaction;`

This won't practically do anything for existing stores, but in future if you try to access methods not present in the `Transaction` type, you'll get a compiler error. 

### Optionally call expectTransaction in your store method

Because of the cognitive load of trying to keep track of which methods really should be called in a transaction, we've also included an opt in way of enforcing that by calling `expectTransaction(this.db);` as the first call in your store methods. You don't have to do this and the transactions will run fine without it, this is a way of signalling to future developers that your store method is not safe to call outside of a transaction. For example the  updateGroupUsers method now looks like this:

```js
    async updateGroupUsers(
   // parameters removed for clarity
    ): Promise<void> {
        expectTransaction(this.db); // call this in methods you don't want to run outside of a transaction
        await this.addUsersToGroup(groupId, newUsers, userName); //We want both of these methods to run or neither  
        await this.deleteUsersFromGroup(deletableUsers); 
    }
``` 

What this method will do is crash hard in a very specific set of circumstances:
- Your store method is called outside of a transaction
- You're running this method in a test harness
This makes this safe for production code - you're expected to pick these issues up during development and not in live. If this code is run outside of a test harness you'll get a warning message but the method _will_ _not_ raise an exception.

# The Ugly

There are a few details we should be aware of here. Firstly, while you can mix transactional and non transactional calls in the service level transaction, this takes us into potentially deadlocking Knex if a transaction fails. Practically this shouldn't be a problem because these should be picked up during development but it's something to be aware of.

Another thing to be aware of is that because this is effectively creating temporary clones of the stores, this means we'll need to keep the store constructors light and sane. This is probably not a major issue since store constructors shouldn't be doing heavy logic anyway but again, something to be aware of.

Finally, this also means re exposing the Knex instance so that the services can get hold of them.


dsusa72:
Great description, congrats! Suggestion (optional): the double use of "a _few_ transactions that..." falls a bit short on specificity. I propose to keep the current paragraph as is (to avoid adding complexity), but to add an extra paragraph like "_Examples: current areas we should apply this new patter: A & B, future areas that will benefit: C & D_" . This bit of information will give current readers some expectation setting on upcoming changes, and future readers the background on where this pattern was originally applied (I'm thinking on the readers who will need to improve the transactional implementation in the future).

src/lib/db/transactional.ts
```diff
@@ -0,0 +1,36 @@
...
+        );
+        for (const attribute in this) {
+            clone[attribute] = this[attribute];
+        }
+        clone.db = transaction;
+        return clone as T;
+    }
+}
+
+export const expectTransaction = (db: Knex | Knex.Transaction): void => {
```
gastonfournier, line 24: It seems that everywhere we use this function is a class that extends Transactor, does it make sense to have it as a method of Transactor?

src/lib/db/group-store.ts
```diff
@@ -60,10 +60,14 @@ const groupToRow = (group: IStoreGroup) => ({
...
-export default class GroupStore implements IGroupStore {
-    private db: Knex;
+export default class GroupStore
+    extends Transactor<GroupStore>
+    implements IGroupStore
+{
+    private db: Knex | Knex.Transaction;
 
     constructor(db: Knex) {
+        super();
```
gastonfournier, line 70: Should this be: 
```suggestion
        super(db);
```

I'm guessing because https://github.com/Unleash/unleash/pull/2217/files#diff-d1c23d22abb0f437d1416e15b6351a785373db023f45dd51090027c060582c93R51 and in most cases we're matching the constructor

src/lib/db/transactional.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export abstract class Transactor<T> implements Transactional<T> {
+    args: any[];
+
+    // eslint-disable-next-line
+    constructor(...args: any) {
+        this.args = args;
+    }
+
+    transactional(transaction: Knex.Transaction): T {
```
gastonfournier, line 12: Would it make sense to rename this as `join` or `joinTransaction`?

src/lib/db/event-store.ts
```diff
@@ -217,6 +224,18 @@ class EventStore extends AnyEventEmitter implements IEventStore {
...
+        let clone = new (this.constructor as {
+            new (db: Knex, getLogger: LogProvider): any;
+        })(this.db, this.logProvider);
+
+        for (const attribute in this) {
+            clone[attribute] = this[attribute];
+        }
+        clone.db = transaction;
+        return clone as IEventStore;
+    }
```
gastonfournier, line 238: Just a random thought: we're cloning the object (this reminds me of our last week struggle with clone-deep library), and I was just thinking, what if some of the inner objects also have a reference to `db` that needs to be replaced by `transaction`? Probably, this is not a problem now, but I'm thinking if this should be recursive somehow :thinking:

src/lib/db/group-store.ts
sighphyre: Yes, good eyes, thank you! It's not _strictly_ necessary for a store that only contains the db object, the clone will still work and the TS compiler will allow it but it's good to be clear

src/lib/db/event-store.ts
sighphyre: I feel we should be blocking a pull request that extends complexity in the stores like that. Part of why this is cheap is because this can be a shallow clone and I think stores that take on that kind of nested data complexity are a strong warning we should rethink our lives a little

src/lib/db/transactional.ts
sighphyre: We can if folks want that. I think that's a very small detail between an instance method and a pure utility function. Transactor exists mostly to make it convenient to implement the clone method with little fuss, which is surprisingly hard to do without a base class. Pulling this into that class is the difference between writing
`this.expectTransaction()`
and 
`expectTransction(this.db)`

I'm personally like it how it is (which might be my Rust showing) but if folks vote to have it moved into the Transactor class I can do that.

src/lib/db/transactional.ts
sighphyre: `join` seems dangerous - I would at least start wondering how on earth a SQL join works in this context. I could go for `joinTransaction` but I kinda want something that expresses that you're getting a new object here too, `transactional` might not be it

src/lib/db/transactional.ts
gastonfournier: Not a big deal tbh, I was just thinking if this is a method that can be used in services, then it'd make more sense to have it as an exported function. Having it as part of the class might help to be aware of it when extending `Transactor` but that'd be the only reason.

sighphyre: Actually, that makes a ton of sense, sir. You're right - this suggests you can and should use this outside of the store implementations and I think that'd be weird. I will shift this into the transactor abstract class.

src/lib/db/transactional.ts
gastonfournier: Then transactional might work well, although `cloneWithTransaction` could work for this at the risk of being too verbose and explicit

src/lib/db/transactional.ts
sighphyre: Right, so I had to make some compromises here - first, I've added `expectTransaction(db: Knex | Knex.Transaction)` to the abstract base class. This is because the abstract class doesn't actually know anything about the underlying db object, in spite of the clone method, and I think it probably shouldn't to prevent us from having to pass the db object explicitly as a named parameter to the super constructor but I do think
`this.expectTransaction(this.db);`
gives a nice sense of how it should be used.

The other compromise is that I've left the utility method exposed as an export, this is so things like the event store that don't actually extend the base class can still get their teeth into the method rather than having to roll their own. WDYT?

gastonfournier: I think it looks nice and indeed it's a good helper method for those cases where you don't extend `Transactor` and instead implement the `Transactional` interface :+1:

gastonfournier (approved):
LG!

nunogois (approved):
LGTM

src/lib/db/event-store.ts
sighphyre: Yeah, we did look but it got unwieldy fast and looked relatively dangerous so we noped out of that. Ultimately we figured we'd want single block transactions at the service level and if you're slinging around existing transactions at the database level you should... probably... not do that, mostly because you're going to shoot your own foot off. But yes, very open to extending or refactoring this if you want to add more magic

src/lib/db/transactional.ts
sighphyre: Yeah `cloneWithTransaction` isn't bad, I think the tricky thing with this name is that I'm both trying to mask the fact that I'm doing a clone in the service layer so that you can ignore that implementation detail while still trying to be expressive about what details you need to be aware of in the calling code (the stuff behind the scenes will fall off the stack when the closing block completes).

I'm open to going with `cloneWithTransaction` but let's let a few more folks weigh in on this one

src/lib/db/event-store.ts
gastonfournier: I think for now this is a good step forward, I was mostly "inspired"/biased by the way some Java frameworks handle transactions and transaction propagation: they can be defined at any level and any internal call talking to the DB (or persisting framework in this case), would be involved in the same transaction (unless explicitly you require a new transaction). But we're not solving the problem of "having a transactional framework to propagate transactions" so :+1:

src/lib/db/transactional.ts
gastonfournier: I agree with you, I'd like to hide the fact that you're cloning, as I think it's an implementation detail

src/lib/middleware/oss-authentication.test.ts
```diff
@@ -21,7 +21,7 @@ async function getSetup(preRouterHook) {
         },
     });
     const stores = createStores();
-    const services = createServices(stores, config);
+    const services = createServices(stores, config, undefined);
```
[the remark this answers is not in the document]
ivarconr: ...or maybe pass in db in tests also?

src/lib/services/group-service.ts
```diff
@@ -103,43 +108,47 @@ export class GroupService {
 
         await this.validateGroup(group, preData);
 
-        const newGroup = await this.groupStore.update(group);
+        return this.db.transaction(async (tx) => {
```
ivarconr, line 111: Do we really want to expose entire knex to all services? Or could transactional stores provide a transaction function (possible via a mixing) instead?

src/lib/db/group-store.ts
```diff
@@ -210,10 +210,9 @@ export default class GroupStore implements IGroupStore {
         deletableUsers: IGroupUser[],
         userName: string,
     ): Promise<void> {
-        await this.db.transaction(async (tx) => {
-            await this.addUsersToGroup(groupId, newUsers, userName, tx);
-            await this.deleteUsersFromGroup(deletableUsers, tx);
-        });
+        this.expectTransaction(this.db);
```
ivarconr, line 213: This got my attention. Does it make sense for a store to require the caller to wrap the call in a transaction? What would happen if we left the original code as is and the service wrapped the transaction in another transaction?

ivarconr: Maybe this method does not belong in the store anymore, if we know move the transaction handling to the service?

src/lib/services/group-service.ts
ivarconr: Maybe we could offer a special u`withTransaction` function in the `stores` container you can use instead of exposing db/knex directly to all services.

src/lib/db/group-store.ts
```diff
@@ -61,10 +61,14 @@ const groupToRow = (group: IStoreGroup) => ({
     mappings_sso: JSON.stringify(group.mappingsSSO),
 });
 
-export default class GroupStore implements IGroupStore {
-    private db: Knex;
+export default class GroupStore
+    extends Transactor<GroupStore>
```
ivarconr, line 65: is it possible to use [Mixins](https://www.typescriptlang.org/docs/handbook/mixins.html) instead of inheritance to provide the transactional capabilities to the store?

src/lib/middleware/oss-authentication.test.ts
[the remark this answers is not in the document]
sighphyre: That's actually better, I'll make it optional

src/lib/services/group-service.ts
sighphyre: There's a reason, maybe a bad one but a reason - it makes a lot more sense in terms of how you end up using this. If we're moving transaction responsibilities into the service layers it makes more sense for the service to both own and start the transaction. You can reach into one of the stores to make that happen but you end up with things like this:

```js
const someStore = getStore();
const someOtherStore = getStore();

someStore.transaction(async (tx) => { //why is this store starting the transaction? Could it be the other one? That's not clear from the syntax 
})

``` 
That's probably not worth the tradeoff of exposing the entire Knex object to the services but we could pass a closure that yields a transaction object. Something like

```
const transactionMaker(knex: Knex){
   startTransaction(){
      return knex.transction()
   }
   return startTransction
}

```
That would allow us to only provide access to creating a transaction to the services while still maintaining a nice API. 

WDYT?

src/lib/db/group-store.ts
sighphyre: This is actually what I tried first, because I would rather avoid inheritance if possible. Unfortunately, you can't hook into the constructor type from a mixin so the cloning approach here doesn't work. If anyone has a suggestion for making that work I would be a ton happier with a mixin though

src/lib/db/group-store.ts
sighphyre: This is a method we added in this PR, in my mind this was less about stores telling services things and more about programmers in the past telling programmers in the future things. What I wanted to try and avoid here is folks making messy store methods that require multiple calls without a way to signal to future developers that it's not safe to be run in a non transactional context. This shouldn't do anything outside of a test harness - this is really just a way to make people's tests crash hard if they do silly things

I... actually haven't tried to wrap a transaction in a child transaction. There's an open issue on Knex that suggests that deadlocks but let me try it and see what happens.

src/lib/services/group-service.ts
ivarconr: yes I kind of corrected myself in the second comment. Transactions should be able to span stores, that is a important need. I am not to worried what we call it, but allowing the services to start a transaction without interacting directly with the database is important to me. 

I am thinking that we could envision a possible future where the entire store/database layer would be replaced with an in-memory implementation at one point (to facilitate a throw-away unleash you spin up for automated integration testing), then it would be nice to not tie the transaction handling directly to knex, but instead provide an abstraction we can replicate with other store adapters as well.

sighphyre: Right, so I've done some stuff here. The services can now depend on a `TransactionCreator<UnleashTransaction>` type instead of directly on anything related to Knex. `TransactionCreator<T>` is basically an alias for a closure type where the closed function is generic over an inner type that's determined by how it's called - that's to allow you to have a start a transaction that will resolve to an `IGroup` for example (S would be `IGroup` in that case). The `UnleashTransaction` is an alias for `KnexTransaction | null`, that doesn't do anything special at the moment, it's just an escape hatch so that  you can mock out the transaction handling in the services and not have to know about or use Knex. So the service call now looks like this:

```js
export class GroupService {
    private startTransaction: TransactionCreator<UnleashTransaction>;

    constructor(
        stores: Pick<IUnleashStores, 'groupStore' | 'eventStore' | 'userStore'>,
        { getLogger }: Pick<IUnleashConfig, 'getLogger'>,
        startTransaction: TransactionCreator<UnleashTransaction>,
    ) {
        //some fields we're ignoring
        this.startTransaction = startTransaction;
    }
   
    //totally made up method
    async getGroups(): Promise<IGroup> {
    
        return this.startTransaction(async (tx) => {  // <- the new magic
            return await this.groupStore
                .transactional(tx)
                .getAllGroups();
        });
    }
}
```

I've also made a little utulity method to generate an `TransactionCreator<KnexTransaction>` to actually use this. That's as simple as doing:

```js
    const db: Knex = someCallThatReturnsAKnex(); 
    const startTransaction = createKnexTransactionStarter(db);
```

I've also cleaned up the FakeGroupStore and used that in a test to show that this can actually be mocked out relatively easily.

If we want to push this in the future, I think we could do something like

```js
class GroupService<T> {
  
  constructor(store: IGroupStore<T>) {
    //stuff
  }
}
```
So that the service transaction type can be type bound to the type of the stores being passed in but I haven't explored that in detail, this is already a fairly big change and that would mean mucking with the types on the stores.

@ivarconr WDYT?

kwasniew:
Some early thoughts. I think the root cause of the problem here is that we have a graph of **application scope** dependencies. In other words every controller, service and repository is a singleton. That's why we need to hack some of them (repositories) to be newed up per service method invocation. I'd consider turning some of the dependencies into **request scope** dependencies that are newed up per HTTP request. Since controllers model request scope in a typical web application we could do it there.
Screenshot 2022-11-07 at 09 28 34

kwasniew:
Also, I'm wondering how we could cover a scenario when one service is making transactional changes in its own store and another service.

sighphyre:
Closing this in favour of a related PR on enterprise

PR #2218, opened 2022-10-20
Achievements

nunogois:
# Achievements

Co-authored by: @sjaanus @NicolaeUnleash 

This adds achievements to Unleash as part of a UX initiative that leans a bit into gamification. Achievements help guide the user and serve as goals the user can look forward to and try to achieve. 

Achievements are defined on the back-end and then triggered by the front-end on specific timings.

[achievements.webm](https://user-images.githubusercontent.com/14320932/197029492-65313419-b0c8-4ece-b8d5-3f39c128e2c1.webm)

## Adding and triggering an achievement

We currently have 2 achievements to play around with, but it's easy to add new ones.

Head to [src/lib/constants/achievements.ts](https://github.com/Unleash/unleash/blob/feat-achievements/src/lib/constants/achievements.ts) where you can find all the achievements and add your own:

```JS
export const Achievements: IAchievement[] = [
    {
        id: Achievement.FIRST_LOGIN,
        title: 'Hello Unleash',
        description: 'Sign in for the first time',
        imageUrl: 'https://www.getunleash.io/logos/unleash_glyph_pos.svg',
    },
    {
        id: Achievement.FIRST_TOGGLE,
        title: 'Toggle Creator',
        description: 'Create your first feature toggle',
        imageUrl: 'https://www.getunleash.io/logos/unleash_glyph_pos.svg',
    },
];
```

On the front-end, add your new achievement identifier to [frontend/src/constants/achievements.ts](https://github.com/Unleash/unleash/blob/feat-achievements/frontend/src/constants/achievements.ts).

Find the exact timing you would like to trigger your achievement unlock:

```JS
import { AchievementContext } from 'component/achievements/AchievementContext/AchievementContext';
import { Achievement } from 'constants/achievements';

// ...

const { unlockAchievement } = useContext(AchievementContext);

// ...

unlockAchievement(Achievement.EPIC_ACHIEVEMENT);
```

It's ok to call `unlockAchievement` every time, even if the user has already unlocked it. This is handled automatically for us on the [store](https://github.com/Unleash/unleash/blob/feat-achievements/src/lib/db/achievements-store.ts#L66) layer and filtered on the [provider](https://github.com/Unleash/unleash/blob/feat-achievements/frontend/src/component/achievements/AchievementProvider/AchievementProvider.tsx#L27).

## Why let the front-end in charge here?

Initially the idea was to let the back-end handle this, but we faced some issues.

 - Triggering new achievements on the back-end meant awkward changes that polluted our code, and this can be seen here: https://github.com/Unleash/unleash/commit/1cc32e2935f8b603f2a868823831cc2e419754f9;
 - It also meant that front-end would need to poll the back-end for new achievements at a very short interval, so that the popup could trigger close to real-time;
 - Besides, we considered achievements extremely non-critical;

### Cheating

Of course, this means you can **cheat** currently and unlock whatever achievements you want. There's no way around it without some server-side logic to guard this. An idea would be to have checks mapped to each achievement, that would run on the context of the current user. Or at least, for some super special rare ones.

## What's left?

Assuming we see value in this and want to merge it, besides a review, we identified a couple of things we might want to look at:

 - [ ] Review the achievements - Do they make sense? Edit them, or add new ones, etc;
 - [ ] Prevent cheating;
 - [ ] Convert [src/lib/db/achievements-store.ts](https://github.com/Unleash/unleash/blob/feat-achievements/src/lib/db/achievements-store.ts) raw queries to Knex;

### DIFF
--- a/frontend/src/component/App.tsx
+++ b/frontend/src/component/App.tsx
@@ -17,6 +17,7 @@ import { useAuthUser } from 'hooks/api/getters/useAuth/useAuthUser';
 import { SplashPageRedirect } from 'component/splash/SplashPageRedirect/SplashPageRedirect';
 import { useStyles } from './App.styles';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
+import { AchievementPopup } from './achievements/AchievementPopup/AchievementPopup';
 
 export const App = () => {
     const { classes: styles } = useStyles();
@@ -75,6 +76,7 @@ export const App = () => {
                                     </Routes>
                                     <FeedbackNPS openUrl="http://feedback.unleash.run" />
                                     <SplashPageRedirect />
+                                    <AchievementPopup />
                                 </div>
                             }
                         />
--- a/frontend/src/component/App.tsx
+++ b/frontend/src/component/App.tsx
@@ -17,7 +17,6 @@ import { useAuthUser } from 'hooks/api/getters/useAuth/useAuthUser';
 import { SplashPageRedirect } from 'component/splash/SplashPageRedirect/SplashPageRedirect';
 import { useStyles } from './App.styles';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
-import { AchievementPopup } from './achievements/AchievementPopup/AchievementPopup';
 
 export const App = () => {
     const { classes: styles } = useStyles();
@@ -76,7 +75,6 @@ export const App = () => {
                                     </Routes>
                                     <FeedbackNPS openUrl="http://feedback.unleash.run" />
                                     <SplashPageRedirect />
-                                    <AchievementPopup />
                                 </div>
                             }
                         />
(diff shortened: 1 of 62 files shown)


ivarconr:
Can you put it behind a feature flag so we can test it for a few customers?

nunogois:
> Can you put it behind a feature flag so we can test it for a few customers?

Thanks @ivarconr, addressed this in https://github.com/Unleash/unleash/pull/2218/commits/534d3f7496bc564bd99ca701f11b231827060079

The way it is implemented means that achievements will still unlock, but the user will see no changes on the UI: 
 - No achievement popup when one gets unlocked;
 - Profile: No achievements badges;
 - Profile: No achievements tab;

Once we enable the flag, the users will be able to see their previously unlocked achievements on their profile, whatever date that happened to be in.

nunogois:
@ivarconr What do you think we should do with this?

PR #2240, opened 2022-10-24
Suggest changes enabled dialog

sellinjaanus:
Suggest changes enabled dialog

Started this, but needs more work.


Tymek:
Moved to #2247

PR #2244, opened 2022-10-24
Update rootRole on login (#2243)

### LINKED ISSUE #2243: [Unleash-server]: customAuthHandler fails to update user
### Describe the bug

When creating an instance of the unleash-server with a customAuthHandler, the method `userService.loginUserSSO()` isn't updating the user if it already exists. By doing `console.log()` on the user object in the OIDCStrategy we see that the `roles` scope is updated, but the user in Unleash isn't. If we delete the user from the DB and try to log in again after the user has been deleted, then the access level / role is updated. Note that this is only tested on root roles.

For this use case the Azure AD handler has been used. 

Code from the custom auth handler: 
```javascript
passport.use(
    'azure',
    // Check passport azure ad documentation for option details: https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/maintenance/passport-azure-ad#4112-options
    new OIDCStrategy(
      {
        identityMetadata: `https://login.microsoftonline.com/${tenantID}/v2.0/.well-known/openid-configuration`,
        clientID,
        clientSecret,
        redirectUrl: `${host}/api/auth/callback`,
        responseType: 'code',
        responseMode: 'query',
        scope: ['openid', 'email', 'profile'],
        allowHttpForRedirectUrl: true,
      },
      async (iss, sub, profile, accessToken, refreshToken, cb) => {
        console.log('debug user', profile)
        let name = profile.displayName;
        let email = profile._json.email;
        let role = profile._json.roles ? profile._json.roles[0] :  unleash.RoleName.VIEWER
        const user = await userService.loginUserSSO({ 
          email, name, rootRole: role, autoCreate: true 
        })
        cb(null, user);
      }
    )
  );
```

To clarify we are configuring custom AzureAD roles for the application representing the three base roles in Unleash `Admin`, `Editor` and `Viewer`. So in the `let role` statement, it's either pulling one of them - and if there is no role assigned it will fall back to the Viewer role as a default.

1. Boot an unleash server
2. Create a custom AzureAD Auth handler
3. Construct the user with the `userService.loginUserSSO()` and `autoCreate: true`
4. Log in with one set of permissions assigned. 
5. Update the permissions either by assigning a new role in AzureAD or hardcode in something new and restart the local unleash-server
6. Log in again and see that your root role hasn't been updated.

### Expected behavior

Unleash don’t look at `rootRole` when updating. Only when creating. We want Unleash to look at the `rootRole` when updating as well to catch updates and support PIM solutions.

### Logs, error output, etc.

```bash
Not any logs, but link to Slack thread: https://unleash-community.slack.com/archives/CGP2MCHPF/p1666621398834659
```

### Screenshots

image
How the mapping is done in AzureAD

### Additional context

_No response_

### Unleash version

4.12.6

### Subscription type

Open source

### Hosting type

Self-hosted

_No response_

dev-slatto:
## About the changes
Don't think this needs much explanation. It will only update the user object on SSO authentication when a rootRole is present on the User object. This should solve the issues as mentioned in #2243 

Closes #2243 .

### DIFF
--- a/src/lib/services/user-service.ts
+++ b/src/lib/services/user-service.ts
@@ -330,6 +330,9 @@ class UserService {
             if (name && user.name !== name) {
                 user = await this.store.update(user.id, { name, email });
             }
+            if (rootRole) {
+                user = await this.store.update(user.id, { rootRole });
+            }
         } catch (e) {
             // User does not exists. Create if "autoCreate" is enabled
             if (autoCreate) {
--- a/src/lib/types/stores/user-store.ts
+++ b/src/lib/types/stores/user-store.ts
@@ -1,3 +1,4 @@
+import { RoleName } from '../model';
 import { IUser } from '../user';
 import { Store } from './store';
 
@@ -17,6 +18,7 @@ export interface IUserLookup {
 export interface IUserUpdateFields {
     name?: string;
     email?: string;
+    rootRole?: number | RoleName;
 }
 
 export interface IUserStore extends Store<IUser, number> {


ivarconr (changes requested):
We can not accept this change as it will be a breaking change to existing users of Unleash. We have two options:

1. Introduce a new method for the new behavior 
2. document how to deal with this. After all this require custom code anyway. 

I lean toward option 2.

dev-slatto:
@ivarconr Will it do that? We're introducing this as a new optional variable on the update user object so that if it's not explicitly set, it should not do anything? That's also the reason I created it as a dedicated if block just targeting the rootRole if updated. I don't know the codebase that good so not quite sure on the intended flow, but shouldn't there be a way when using an customAuthHandler to have the data from that being the truth of access? 

If there is a way to do this that I don't know about I would be happy to learn 😄

ivarconr:
> I don't know the codebase that good so not quite sure on the intended flow, but shouldn't there be a way when using an customAuthHandler to have the data from that being the truth of access?

your custom auth handler can just call [updateUser()](https://github.com/Unleash/unleash/blob/main/src/lib/services/user-service.ts#L231) with correct rootRole to achieve the desired behavior.

dev-slatto:
Thanks! Then I'll do it in the auth handler 😄

PR #2246, opened 2022-10-25
[docs]: Update Prometheus docs with security info

dev-slatto:
## About the changes
Prometheus metrics should not be exposed to the public. Added a note about this to inform people that internal endpoints should be dropped on external access. 

## Discussion points
https://unleash-community.slack.com/archives/CGP2MCHPF/p1666688295670459

### DIFF
--- a/website/docs/api/internal/internal-backstage-api.md
+++ b/website/docs/api/internal/internal-backstage-api.md
@@ -7,7 +7,9 @@ title: /internal-backstage/prometheus
 
 `GET http://unleash.host.com/internal-backstage/prometheus`
 
-Unleash uses prometheus internally to collect metrics. These are available on the given url if the `serverMetrics` option is enabled (default=true).
+Unleash uses Prometheus internally to collect metrics. These are available on the given url if the `serverMetrics` option is enabled (default=true).
+
+Note that it's not recomended to expose Prometheus metrics to the public. You should therefore drop all external access to `/internal-backstage/*` on the network layer to keep your instance secure. 
 
 [Read more about Prometheus](https://prometheus.io/)
 
--- a/website/docs/api/internal/internal-backstage-api.md
+++ b/website/docs/api/internal/internal-backstage-api.md
@@ -9,7 +9,7 @@ title: /internal-backstage/prometheus
 
 Unleash uses Prometheus internally to collect metrics. By default, the metrics are available at `/internal-backstage/prometheus`. You can disable this endpoint by setting the `serverMetrics` option to `false`.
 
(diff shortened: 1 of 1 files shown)


website/docs/api/internal/internal-backstage-api.md
```diff
@@ -7,7 +7,9 @@ title: /internal-backstage/prometheus
 
 `GET http://unleash.host.com/internal-backstage/prometheus`
 
-Unleash uses prometheus internally to collect metrics. These are available on the given url if the `serverMetrics` option is enabled (default=true).
+Unleash uses Prometheus internally to collect metrics. These are available on the given url if the `serverMetrics` option is enabled (default=true).
```
thomasheartman, line 10: I know you haven't really touched this part, but it read a bit awkwardly to me, so I'll add a suggestion for you here. What do you think?


```suggestion
Unleash uses Prometheus internally to collect metrics. By default, the metrics are available at `/internal-backstage/prometheus`. You can disable this endpoint by setting the `serverMetrics` option to `false`.
```

website/docs/api/internal/internal-backstage-api.md
```diff
@@ -7,7 +7,9 @@ title: /internal-backstage/prometheus
...
+
+Note that it's not recomended to expose Prometheus metrics to the public. You should therefore drop all external access to `/internal-backstage/*` on the network layer to keep your instance secure.
```
thomasheartman, line 12: I think adding this bit is a very good suggestion! 💯  However, I'd like to ask you to add a little bit more nuance to it to make it easier to read.

A couple things:
1. do you have a reference for why it's not recommended to expose prometheus metrics? Or can you explain why it's not recommended? It might be obvious to you, but it's not to everyone else.
2. "You should therefore drop all external access to `/internal-backstage/*` on the network layer to keep your instance secure." Could this be something like: 
3. 
```suggestion
Note that it's not recommended to expose Prometheus metrics to the public. Thus, if you want to keep metrics enabled, you should block all external access to `/internal-backstage/*` on the network layer to keep your instance secure. 
```

What do you think?

thomasheartman (changes requested):
Hey, @dev-slatto 👋🏼  Sorry for taking so long to get around to this PR 🙇🏼  Super nice work on this; I definitely agree that it's worth mentioning this in the docs. Just got a few little comments and suggestions I'd like for you to address, if you don't mind? 😄

dev-slatto:
@thomasheartman Hey! 
I've added your suggestions now, and added a reference to the Prometheus external pentest-report that suggest protecting clients metrics endpoint with authentication 😄

thomasheartman (approved):
Great work; thanks for accepting my suggestions and providing some sources for recommendation 🙌🏼  It looks ready to go now, so I'll merge it 🚂 

Let me know if you've got any follow-ups, and thanks so much for contributing and helping us improve! ❤️

PR #2250, opened 2022-10-25
Suggest changes refactoring and schema fixes

sjaanus:
1. Now drafts endpoint will return list of drafts
2. New schema for suggest changes
3. Smaller tweaks

### DIFF
--- a/src/lib/db/suggest-change-store.ts
+++ b/src/lib/db/suggest-change-store.ts
@@ -66,8 +66,8 @@ const suggestChangeRowReducer = (acc, suggestChangeRow) => {
     const currentSuggestChangeSet = acc[suggestChangeSet.id];
 
     if (changeId) {
-        const feature = currentSuggestChangeSet.features.find(
-            (f) => f.name === changeFeature,
+        const featureObject = currentSuggestChangeSet.features.find(
+            (feature) => feature.name === changeFeature,
         );
         const change = {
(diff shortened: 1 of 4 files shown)


src/lib/db/suggest-change-store.ts
```diff
@@ -61,15 +60,17 @@ const suggestChangeRowReducer = (acc, suggestChangeRow) => {
...
-            changes: [],
+            features: [],
         };
     }
     const currentSuggestChangeSet = acc[suggestChangeSet.id];
 
     if (changeId) {
-        currentSuggestChangeSet.changes.push({
+        const feature = currentSuggestChangeSet.features.find(
+            (f) => f.name === changeFeature,
```
FredrikOseberg, line 70:
```suggestion
            (feature) => feature.name === changeFeature,
```

src/test/fixtures/fake-suggest-change-store.ts
```diff
@@ -63,14 +60,11 @@ export default class FakeSuggestChangeStore implements ISuggestChangeStore {
     }
 
     create(
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
FredrikOseberg, line 63: Would prefer to avoid these.

FredrikOseberg (approved):
LGTM - Made some minor suggestions

PR #2260, opened 2022-10-26
chore: Use last version of @unleash/express-openapi

gastonfournier:
## About the changes
Just to use a non-beta version https://www.npmjs.com/package/@unleash/express-openapi?activeTab=versions no changes from v0.2.1-beta.0 meaning that the [change we were testing](https://github.com/Unleash/express-openapi/pull/5) is stable now

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):
Perfect! Thanks for keeping track of this 🙌🏼

PR #2261, opened 2022-10-26
fix: refactor conditional middleware

FredrikOseberg:
Fixes an issue where conditional middleware would return 404 for paths that were registered after the conditional middleware was being used.


kwasniew (approved):
Pair programmed

PR #2266, opened 2022-10-26
Suggest change - update status and get per project

sjaanus:
This PR adds two endpoints to suggest change

1. Update status of change set
2. Get all change sets per project that are not drafts


andreas-unleash (approved):
LG!

PR #2272, opened 2022-10-27
fix: escape single quotes ' for curl

gastonfournier:
## About the changes
When the commit message has a single quote `'`, it breaks our curl command that encloses the request body between single quotes. A similar issue was reported in https://github.com/actions/runner/issues/1656 and the suggestion is to use an environment variable that when used in bash will use bash's built-in escaping of env variables that should help us here as well.

For testing purposes the title of the PR includes also a single quote


thomasheartman (approved):
Looks like a sensible way to do it to me! Just make sure we also escape double quotes (") and other potential problems (don't know if there are any others?)

gastonfournier:
> Looks like a sensible way to do it to me! Just make sure we also escape double quotes (") and other potential problems (don't know if there are any others?)

Everything else should be handled by toJSON

PR #2284, opened 2022-10-28
make prettify large number tests independent of locale

sighphyre:
## What

We recently accepted a PR that includes some tests to prettify large numbers in the front end project. However, these tests are not locale independent and so fail on machines where the number separator is not a comma. This patches those tests to rather check that the length of the output string matches the expected length i.e. that the input strings have had a separator injected into them.

This doesn't touch the tests that use millify for the number formatting. Since we don't set any options on the millify function itself, this should be locale independent already (possibly something to talk about in future). 

I'm not 100% we actually want to keep these tests but if we do this is an option to preserve the logic while keeping them locale independent.

### DIFF
--- a/frontend/src/component/common/PrettifyLargeNumber/PrettifyLargeNumber.test.tsx
+++ b/frontend/src/component/common/PrettifyLargeNumber/PrettifyLargeNumber.test.tsx
@@ -11,7 +11,7 @@ describe('PrettifyLargeNumber', () => {
             LARGE_NUMBER_PRETTIFIED
         );
 
-        expect(prettifiedText.textContent).toBe('999,999');
+        expect(prettifiedText.textContent).toHaveLength('999,999'.length);
     });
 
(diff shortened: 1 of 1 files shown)


thomasheartman:
That seems a sensible solution, but couldn't we also set the locale explicitly for the tests?

FredrikOseberg (approved):

sighphyre:
> That seems a sensible solution, but couldn't we also set the locale explicitly for the tests?

Yeah we almost certainly could do that. I'm not sure that's the right thing to do though (also not sure it's not), the tests in their current state are fairly fragile since they're so specific, I'd rather know that a delimiter has been added than the delimiter has been set to a given thing in a given locale. This will let us know that something went wonky if someone wanders into a locale that breaks this contract. Alternatively we could iterate a set of locales and test each one but given this isn't a critical core component I'd rather we relax the testing here

PR #2291, opened 2022-10-28
fix: only show link to create segment if you have permission

nunogois:
https://linear.app/unleash/issue/2-360/segments-bug


chriswk (approved):
Nice catch :)

PR #2300, opened 2022-10-31
fix: re export knex

FredrikOseberg:
* Takes the knex import and reexports it so we can keep the same types in the enterprise project

## About the changes

Closes # .


sjaanus (approved):
LGTM!

PR #2303, opened 2022-10-31
fix: adapt to new CI builder format

gastonfournier:
## About the changes
This PR adapts to the new format we're expecting on the integration side, which should also help with the problem we faced with commits including single quotes in the commit message


gardleopard (approved):
LGTM

gardleopard (approved):
LGTM

PR #2307, opened 2022-11-01
Hyperlink Injection in People Invitation Emails

sjaanus:
Currently it is possible to send links through emails to users during invitation.
As a workaround, we will start stripping special characters from name, but still leaving language specific letters.


FredrikOseberg (approved):
LGTM

PR #2308, opened 2022-11-01
Hard cap expiry date of public signup token

sjaanus:
Currently there was missing backend validation of expiry for public signup tokens. This PR adds a hard cap to the expiry.


chriswk (approved):
Nice

PR #2310, opened 2022-11-01
Feat/review page timeline

FredrikOseberg:
Adds timeline to the change request overview page, displaying the different states that a change request can be in. Adds a layout component to the review status section (bottom right part of the page).

 
Skjermbilde 2022-11-02 kl  11 12 09

## About the changes

Closes #


sjaanus (approved):
LGTM

frontend/src/component/suggestChanges/SuggestedChangeOverview/SuggestedChangeReviewStatus/SuggestedChangeReviewStatus.tsx
```diff
@@ -30,13 +30,11 @@ export const SuggestedChangeReviewStatus: FC<
...
-                    <ConditionallyRender
-                        condition={approved}
-                        show={<Approved approved={approved} />}
-                        elseShow={<ReviewRequired approved={approved} />}
-                    />
-                </StyledFlexAlignCenterBox>
+                <ConditionallyRender
+                    condition={approved}
+                    show={<Approved approved={approved} />}
+                    elseShow={<ReviewRequired approved={approved} />}
```
thomasheartman, line 36: Just with regards to these two components: is there **ever** a case where `Approved.approved`  isn't `true`? And similarly, is there ever a case where `ReviewRequired.approved` is `false`? Just seems a bit redundant to me, but I don't have all the context here 💁🏼

frontend/src/component/suggestChanges/SuggestedChangeOverview/SuggestedChangeTimeline/SuggestedChangeTimeline.tsx
```diff
@@ -1,52 +1,136 @@
...
+                <StyledTimeline>{renderTimeline()}</StyledTimeline>
+            </StyledBox>
+        </StyledPaper>
+    );
+};
+
+const createTimeline = (
+    data: ITimelineData[],
+    hasActiveStep: boolean,
+    activeIndex: number
```
thomasheartman, line 81: Do these two parameters potentially conflict? What happens if `hasActiveStep` is `true`, but the `activeIndex` is `-1`? Or what if `hasActiveStep` is false, but `activeIndex` is 0? Could that be covered by a union type? Something like:
`type Index = "inactive" | number` (or even `type Index = number?`) or the like?

frontend/src/component/suggestChanges/SuggestedChangeOverview/SuggestedChangeTimeline/SuggestedChangeTimeline.tsx
```diff
@@ -1,52 +1,136 @@
...
+            });
+        }
+
+        return createTimelineItem('grey', title, connector);
+    });
+};
+
+const createCancelledTimeline = (data: ITimelineData[]) => {
+    return data.map(({ title }, index) => {
+        const notLastItem = !(index === data.length - 1);
```
thomasheartman, line 109: I wonder if this could be slightly renamed for clarity? something like the below suggestion or `hasSubsequentItems` or something? I don't feel strongly about this, but I just had to take a second to think about why we used `notLastItem`. 

```suggestion
        const shouldConnectToNextItem = index < data.length - 1;
```

thomasheartman (commented):
Overall this looks fine to me, but I've got a couple minor comments and questions added inline. Also, I have no context and don't know what this PR is supposed to do (except for what I've been able to extract from the code). Would you mind adding a description of what this is? ☺️

frontend/src/component/suggestChanges/SuggestedChangeOverview/SuggestedChangeTimeline/SuggestedChangeTimeline.tsx
FredrikOseberg: Possibly, will check it out.

frontend/src/component/suggestChanges/SuggestedChangeOverview/SuggestedChangeTimeline/SuggestedChangeTimeline.tsx
FredrikOseberg: Makes sense

frontend/src/component/suggestChanges/SuggestedChangeOverview/SuggestedChangeReviewStatus/SuggestedChangeReviewStatus.tsx
thomasheartman: Right, thanks for the clarification. But just so I understand fully: `approved` seems to be a boolean. In what cases is a review required if a change request has already been approved? And similarly, when would you render an `Approved` component if `approved` is false?

FredrikOseberg: I guess it's poorly named. These two components, ReviewRequired and Approved share some styled components that change the color based on the property they receive. It's not a prerequisite for any logic inside these components, but a way to control how the components inside will display when rendered. 

I will follow up with a PR to make this more clear.

PR #2315, opened 2022-11-02
Add stable row id.

kwasniew:
Co-authored-by: Tymoteusz Czech <2625371+Tymek@users.noreply.github.com>

## About the changes

Before:
When switching toggles in the project overview unrelated toggled were flipped. It was due to server returning data in different order and frontend not having stable identity for table rows. 

After:
Now the row identity is inherited from the feature name. react-table wants getRowId to be memoized.

Unfortunately the table data is not fully types so sticking to any param for now.


FredrikOseberg (approved):
LGTM

Tymek (approved):

PR #2316, opened 2022-11-02
fix: server import from frontend

nunogois:
https://unleash-internal.slack.com/archives/C046LV6HH6W/p1667386042150079

This PR makes the import path be a relative path instead of using `tsconfig` paths. If we later decide we want to use `tsconfig` paths (they look a lot nicer IMO) then it might be worth checking out the troubleshooting section in: https://www.npmjs.com/package/vite-tsconfig-paths

I was never able to reproduce this issue on my end.


nunogois:
Will close for now, doesn't seem like we'll need this after https://github.com/Unleash/unleash/pull/2326

PR #2318, opened 2022-11-02
fix: remove duplicate formatApiPath in userUiConfigApi hook

ivarconr:
Fix a small UI bug that incorrectly formats the API url, making the CORS update fail when "baseUriPath" is configured.


sighphyre (approved):
LGTM

Issue #2103, opened 2022-09-28
recently inactive toggles appear active in metrics


rbtcollins:
### Describe the bug

Note that this is 24+  hours after the toggle was removed from our app - so no usage for some time. While the chart is annotated with times, it isn't anchored to a point in time, so I can't tell if thats 4pm today, or yesterday - which wouldn't matter if the chart showed 'the last 48 hours' - but it appears to show 'the last 48 slices from the database' rather than 'the last 48 hours', which are very different things. 
![image](https://user-images.githubusercontent.com/499678/192736304-25d898b8-e6c1-4c72-9ab4-74b30396b2e9.png)

nunogois:
Hi @rbtcollins - Thanks for opening the issue! Let me reach out to the team and we'll get back to you :)

ivarconr:
The problem is that the API should produce the hour buckets for all hours, even if they do not exists in the db. The UI could also be better are showing the actual dates, as it becomes miss-leading the way it is currently displayed.

ivarconr:
A fix is available behind an experimental flag in v4.17, allowing us to validate it in Unleash Cloud. 

It will be GA from unleash v4.18

PR #2319, opened 2022-11-02
fix: generate all hour buckets if missing

ivarconr:
## About the changes
This PR enriches the RAW metrics and adds "containers" for every hour in the request range for the "environment" and "appName" combination present. 

image

TDOO:

- [ ] Write unit tests

Fixes: #2103


src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -150,12 +155,53 @@ export default class ClientMetricsServiceV2 {
...
+
+        hours.forEach((item: HourBucket) => {
+            if (
+                !metrics.some(
+                    (e) => compareAsc(e.timestamp, item.timestamp) === 0,
+                )
+            ) {
+                environments.forEach((environment) => {
+                    applications.forEach((appName) => {
+                        metrics.push({
```
ivarconr, line 191: It feels a bit hackish to have 3 for-loops here. Happy to take input on how we can expand the time-series to make sure we have complete "hour buckets" for the requested range (24 hours, 48 hours, etc) for the two supported dimensions (environment and appName).

chriswk: Since the cardinality here is hours * environment * appName, I'm not seeing a way to avoid three for-loops.
Since I'm used to working with streams instead of array I would probably have put the filter first, and then done a concat of the two lists. Feel free to discard the following code suggestion, I just felt like testing if it was possible to do in a pure way. (Well, pure until the concat :) )
```javascript
const fillers = hours.filter(item => !metrics.some(e => compareAsc(e.timestamp, item.timestamp) === 0))
         .flatMap(hourBucket => 
                environments.flatMap(environment => 
                    applications.map(appName => ({ timestamp: item.timestamp, no: 0, yes: 0, appName, environment, featureName: toggleName })
                )
         );
metrics.concat(fillers).sort((a, b) => compareAsc(a.timestamp, b.timestamp))
```

chriswk (approved):
I do believe this looks reasonable.
Maybe considered using a TreeSet with a custom comparator to have both uniqueness and sort handled.

src/lib/services/client-metrics/metrics-service-v2.ts
ivarconr: Thanks. I think this needs to be changed somehow, because it will not be sufficient to only compare the timestamps, we would need to compare the timestamps, environment and appName, to cover all three dimensions.

sighphyre: Hours is 24, environment can't be more than 15, how many appNames are we likely to have? If that's small too then I think this is okay, it looks hackish to have nested loops but if it doesn't actually result in a crazy long loop I think it's okay

src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -150,12 +155,53 @@ export default class ClientMetricsServiceV2 {
...
+
+        const environments = metrics.reduce(
+            (unique, item) =>
+                unique.includes(item.environment)
+                    ? unique
+                    : [...unique, item.environment],
+            [],
         );
+
+        const applications = metrics.reduce(
```
sighphyre, line 175: I think this is totally fine in it's current state but if you want something that avoids the array assignment/lookups in a loop or something more concise, this should work:
```js
const applications = [...new Set(metrics.map(x => x.appName))];
```

sighphyre (approved):
Looks reasonable to me

src/lib/services/client-metrics/metrics-service-v2.ts
ivarconr: Thanks for your input, it helped me improve the read-ability somewhat. 

In order to guard against one unknown, number of applications, I have capped it to only consider the first 100 appName's, so the absolute worst case would be 75k items in the array for a feature toggles (`15 envs x 100 appName x 48 hours = 72000`. I think that is fine.

src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -149,13 +158,56 @@ export default class ClientMetricsServiceV2 {
...
+
+            const environments = [
+                ...new Set(metrics.map((x) => x.environment)),
+            ];
+
+            const applications = [
+                ...new Set(metrics.map((x) => x.appName)),
+            ].slice(0, 100);
+
+            const result = environments.flatMap((environment) =>
```
ivarconr, line 181: @chriswk this way we avoid the concat also. Thanks for your input!

chriswk (approved):
Still looks reasonable, I'd say go for it :)

PR #2324, opened 2022-11-03
Remove manually added migrations and fix the scripts

sjaanus:
Remove manually added migrations and fix the scripts


FredrikOseberg (approved):
LGTM

PR #2326, opened 2022-11-03
Fix docker build

sighphyre:
This adds two things to resolve the issue with the docker build:

Firstly, it patches the stale lockfile in the docker folder.

Secondly, it patches the dockerfile to respect that the frontend is now dependent on the project level constants


gastonfournier (approved):
This looks fine, also tested the build in my machine and did some manual testing on the application

PR #2327, opened 2022-11-03
feat: display change details

kwasniew:
## About the changes

Screenshot 2022-11-03 at 15 42 31


frontend/src/component/changeRequest/ChangeRequest/ChangeRequest.tsx
```diff
@@ -50,37 +60,45 @@ export const ChangeRequest: VFC<IChangeRequestProps> = ({
...
+                                        onDiscard={onDiscard(change.id)}
+                                    >
+                                        <GetFeatureStrategyIcon
+                                            strategyName={change?.payload?.name}
+                                        />
+                                        {formatStrategyName(
+                                            change?.payload?.name
+                                        )}
+                                    </StrategyEditedChange>
+                                )}
```
FredrikOseberg, line 101: Could we use ConditionallyRender? I prefer the codebase to use the same patterns to not get in a situation where people look at one part of the code and implement it one way, and later look at another part of the code and implement it another way. I'm to discuss how we can make it better, but for now consistency is the key to speed and it's easier to refactor a pattern later if the pattern is the same.

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestFeatureToggleChange/StrategyChange.tsx
```diff
@@ -1,27 +1,59 @@
...
+    );
 };
 
-export const StrategyDeletedChange: FC = () => {
+export const StrategyDeletedChange: FC<IStrategyChangeProps> = ({
+    onDiscard,
+}) => {
     return (
-        <Box sx={{ p: 1 }}>
+        <Box sx={{ p: 1, display: 'flex', justifyContent: 'space-between' }}>
```
FredrikOseberg, line 50: Could be a styled component. Also, should we discard `p` in favor of `padding` ?

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestFeatureToggleChange/StrategyChange.tsx
```diff
@@ -1,27 +1,59 @@
...
+    children,
+    onDiscard,
+}) => {
     return (
-        <Box sx={{ p: 1, display: 'flex', gap: 1 }}>
-            <Typography sx={theme => ({ color: theme.palette.success.main })}>
-                + Strategy Added:
-            </Typography>
-            {children}
+        <Box sx={{ p: 1, display: 'flex', justifyContent: 'space-between' }}>
```
FredrikOseberg, line 13: Same Q as below.

FredrikOseberg (changes requested):
Nice. I've added some comments about consistency.

frontend/src/component/changeRequest/ChangeRequest/ChangeRequest.tsx
kwasniew: I originally started with ConditionallyRender but TypeScript can't infer types in the show and elseShow paths.
E.g. the type guards hasNameField or hasEnabledField are understood by TS in the && condition but not in custom ConditionallyRender component. I'm in two minds here because on the one hand I'd like to stick to the convention on the other hand I'd like to maintain type safety.

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestFeatureToggleChange/StrategyChange.tsx
kwasniew: Sure this was WIP and already switched to Styled components

frontend/src/component/changeRequest/ChangeRequest/ChangeRequest.tsx
kwasniew: Screenshot 2022-11-03 at 15 49 17
Screenshot 2022-11-03 at 15 49 24

FredrikOseberg (approved):
LG

PR #2329, opened 2022-11-04
Fix: Conditionally hide Change Requests tab

andreas-unleash:
This PR adds a conditional to the Change Request Tab in Project so that it can be hidden when not enterprise

## About the changes

Closes #


FredrikOseberg (approved):
LG

PR #2339, opened 2022-11-07
Update environments

nunogois:
https://linear.app/unleash/issue/2-357/update-environments-pages


src/lib/db/environment-store.ts
```diff
@@ -112,6 +156,54 @@ export default class EnvironmentStore implements IEnvironmentStore {
...
+    }
+
+    async getProjectEnvironments(
+        projectId: string,
+    ): Promise<IProjectEnvironment[]> {
+        let qB = this.db<IEnvironmentsWithProjectCountsTable>(TABLE)
+            .select(
+                '*',
+                this.db.raw(
+                    '(SELECT COUNT(*) FROM api_tokens LEFT JOIN api_token_project ON api_tokens.secret = api_token_project.secret WHERE api_tokens.environment = environments.name AND (project = :projectId OR project IS null)) as project_api_token_count',
```
sighphyre, line 191: I _think_ this is okay, but the left join in a sub query worries me a bit

@chriswk thoughts here? It's a rarely run query and it only affects the UI

sighphyre: Okay we spoke, we think this is okay

sighphyre (approved):
This is a biggie but I think this looks good

FredrikOseberg (commented):
@nunogois This is nice, but I was tempted to reject based on the size of the PR. Please try to break up the PR size into chunks in the future. It's much easier to review, and we lower the risk of introducing bugs because the reviewer can more easily reason about the code and take in the entire context of the PR. It was difficult for me to do that here.

PR #2344, opened 2022-11-08
Fix: validate that the project is correct when getting feature by project

### LINKED ISSUE #2337: getFeature by name should filter results by projectId
### Describe the bug

/api/admin/projects/:projectId/features/:featureName
https://docs.getunleash.io/reference/api/unleash/get-feature 

When using the above endpoint, I had assumed that results would be filtered by the projectId provided in the path, but I’m not seeing my results change regardless of which (valid) projectId I put in the endpoint. I know that feature names are a unique identifier across all projects, but I would expect that it should validate the project id as well as the feature name and yield a 404 in the case that the feature name does not exist within that project id. 

_No response_

### Expected behavior

_No response_

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

_No response_

### Subscription type

Enterprise

### Hosting type

Hosted by Unleash

_No response_

thomasheartman:
## What

This PR fixes a bug where fetching a feature toggle via the `/api/admin/projects/:projectId/features/:featureName` endpoint doesn't validate that the feature belongs to the provided project. The same thing applies to the archive functionality. This has also been fixed.

In doing so, it also adds corresponding tests to check for edge cases, updates the 403 error response we use to provide clearer steps for the user, and adds more error responses to the OpenAPI documentation.

## Why

As mentioned in #2337, it's unexpected that the provided project shouldn't matter at all, and after discussions internally, it was also discovered that this was never intended to be the case.

## Discussion points

It might be worth rethinking this for Unleash v5. Why does the features API need the projects part at all when features are unique across the entire instance? Would it be worth reverting to a simpler feature API later or would that introduce issues with regards to how different projects can have different active environments and so on?

### Further improvements

I have _not_ provided schemas for the error responses for the endpoints at this time. I considered it, but because it would introduce new schema code, more tests, etc, I decided to leave it for later. There's a thorough OpenAPI walkthrough coming up, so I think it makes sense to do it as part of that work instead. I am happy to be challenged on this, however, and will implement it if you think it's better.

### Why 403 when the project is wrong?

We could also have used the 404 status code for when the feature exists but doesn't belong to this project, but this would require more (and more complex) code. We also already use 403 for cases like this for post, patch, and put. Finally, the [HTTP spec's section on the 403 status code](https://httpwg.org/specs/rfc9110.html#status.403) says the following (emphasis mine):

> The 403 (Forbidden) status code indicates that the server **_understood the request but refuses to fulfill it_**. A server that wishes to make public why the request has been forbidden can describe that reason in the response content (if any).
>
> If authentication credentials were provided in the request, the server considers them insufficient to grant access. The client SHOULD NOT automatically repeat the request with the same credentials. The client MAY repeat the request with new or different credentials. However, **_a request might be forbidden for reasons unrelated to the credentials_**.

As such, I think using 403 makes sense in this case.

---

Closes #2337.

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -159,7 +159,7 @@ class FeatureToggleService {
         const id = await this.featureToggleStore.getProjectId(featureName);
         if (id !== projectId) {
             throw new InvalidOperationError(
-                'Project id does not match the project that the feature belongs to',
+                `The operation could not be completed. The feature exists, but the provided project id (${projectId}) does not match the project that the feature belongs to (${id}). Try using ${id} in the request URL instead of ${projectId}.`,
             );
         }
     }
@@ -554,13 +554,14 @@ class FeatureToggleService {
         archived: boolean = false,
         projectId?: string,
     ): Promise<FeatureToggleWithEnvironment> {
-        if (projectId) {
-            await this.validateFeatureContext({ featureName, projectId });
-        }
-        return this.featureStrategiesStore.getFeatureToggleWithEnvs(
+        const feature = this.featureStrategiesStore.getFeatureToggleWithEnvs(
             featureName,
             archived,
         );
+        if (projectId) {
+            await this.validateFeatureContext({ featureName, projectId });
+        }
+        return feature;
     }
 
     /**
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -552,7 +552,11 @@ class FeatureToggleService {
     async getFeature(
         featureName: string,
         archived: boolean = false,
+        projectId?: string,
     ): Promise<FeatureToggleWithEnvironment> {
+        if (projectId) {
+            await this.validateFeatureContext({ featureName, projectId });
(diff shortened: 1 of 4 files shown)


chriswk (approved):
Nice :+1: :100:

src/lib/services/feature-toggle-service.ts
```diff
@@ -548,15 +548,22 @@ class FeatureToggleService {
...
         featureName: string,
         archived: boolean = false,
+        projectId?: string,
     ): Promise<FeatureToggleWithEnvironment> {
-        return this.featureStrategiesStore.getFeatureToggleWithEnvs(
-            featureName,
-            archived,
-        );
+        const feature =
+            await this.featureStrategiesStore.getFeatureToggleWithEnvs(
```
ivarconr, line 559: I would rather just pass projectId down to the store as well and have that deal with not finding a toggle with the provided feature name and projectId.

thomasheartman: Fair enough! I didn't think of that. Could certainly have a look if you think it's worth diving into?

PR #2345, opened 2022-11-08
Fix/embedded proxy memory leak

FredrikOseberg:
## About the changes

* Fixes an issue where an event would trigger the function setting up the data polling. Causing the polling to increase by the number of events we received.

Closes #

### DIFF
--- a/src/lib/proxy/proxy-repository.ts
+++ b/src/lib/proxy/proxy-repository.ts
@@ -73,7 +73,7 @@ export class ProxyRepository
     }
 
     async start(): Promise<void> {
-        await this.loadDataForToken();
+        await this.dataPolling();
 
         // Reload cached token data whenever something relevant has changed.
         // For now, simply reload all the data on any EventStore event.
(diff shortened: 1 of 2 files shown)


chriswk (approved):
Looks reasonable. Good catch.

gastonfournier (approved):
LG, I've tested without the codechange and the tests are solid

src/lib/proxy/proxy-repository.ts
```diff
@@ -88,11 +88,15 @@ export class ProxyRepository
...
     }
 
-    private async loadDataForToken() {
+    private async dataPolling() {
         this.timer = setTimeout(async () => {
-            await this.loadDataForToken();
+            await this.dataPolling();
         }, this.randomizeDelay(this.interval, this.interval * 2)).unref();
 
+        await this.loadDataForToken();
```
ivarconr, line 96: why does this solve the problem?

ivarconr: recursive??

nunogois: We made some separation of concerns, since `onAnyEvent` only calls `loadDataForToken`, not `dataPolling`, we no longer create new timers when events happen: https://github.com/Unleash/unleash/blob/96e1758b872ac7e4d23cf6c0b29668dad3efc73e/src/lib/proxy/proxy-repository.ts#L112-L114

PR #2348, opened 2022-11-08
docs: update export/import examples

thomasheartman:
## What

This change updates the availability notice and the code samples in the import/export docs.

Specifically, the code samples have been moved to using the ApiRequest component (and thus get examples in HTTP, cURL, and HTTPie), and the import code sample now says which version it's using (version 3).

## Why

The import code sample didn't say which version of the data it used, so Unleash would assume it was version 1 if you tried to run it. However, the actual data was incompatible with the v1 data format, so you'd get an error if you tried to actually import it.

While fixing this, I figured it'd also be alright to add some other minor updates to bring the docs a bit more up to recent standards.

### DIFF
--- a/website/docs/deploy/import-export.md
+++ b/website/docs/deploy/import-export.md
@@ -59,7 +59,7 @@ You can customize the export with query parameters:
 
 For example if you want to download just feature-toggles as yaml:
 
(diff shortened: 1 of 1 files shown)


chriswk (approved):
Looks good, nice to see the ApiRequest component replacing raw POST examples :)

PR #2357, opened 2022-11-09
Add changeRequestEnabled to project and project_environments

andreas-unleash:
- Add `change_request_enabled` column to `projects` and `project_environments`
- Modified the store to include the new column
- Added new column to Project open api schema 

## About the changes

Closes #


andreas-unleash:
Not sure if this is the right way to go about it to minimise impact to oss (or better to separate the concerns by extending the ProjectStore and ProjectService in unleash-enterprise?) Any thoughts appreciated.

kwasniew (approved):
Looks good. One question: do we need this at both the project level and project_environment level?

andreas-unleash:
I think so, so an admin can enable Change requests for the project and still be able to disable them for eg `development`` env

PR #2360, opened 2022-11-09
fix: remove lexical illusion in playground guidance

thomasheartman:
## What

I have removed a lexical illusion lexical illusion in this guidance step, where "context field" was repeated.

## Why

Because it's unintended and doesn't make any sense in the the sentence.


kwasniew (approved):

andreas-unleash (approved):
LG!

PR #2378, opened 2022-11-10
add get and update functionality to project environments change reque…

andreas-unleash:
Add get and update functionality to project environments change request config

## About the changes

Closes #


sjaanus (commented):
Maybe the table can be like this as we discussed.
```

change_request_settings
	environment string
	project     string
```

PR #2380, opened 2022-11-10
Revert "Add changeRequestEnabled to project and project_environments"

andreas-unleash:
Reverts Unleash/unleash#2357


kwasniew (approved):
I'm not sure if we can delete existing migrations or we should write another one. I remember we had a discussion to make migrations an immutable thing. But since that's the latest one maybe we can do it.

src/migrations/20221107132528-change-request-project-options.js
```diff
@@ -1,21 +0,0 @@
-'use strict';
```
ivarconr, line 1: ⚠️ **DANGER:**  We need to be careful removing migrations.

ivarconr: "okish" as long as we have done 0 releases on it.

andreas-unleash:
Reverting was done by creating a new migration to fix - This is unnecessary

PR #2396, opened 2022-11-10
Feature toggles: Filtering by tags

nunogois:
https://linear.app/unleash/issue/UNL-140/experiment-with-filtering-feature-toggles-by-tags-on-the-ui

Going with a naïve approach for now, tags can be searchable the same way we search for text. The tags column only shows up if at least one toggle has tags set. There's a simple highlightable component that lets us know a match was found and then shows all the tags on a tooltip:
image


src/lib/db/feature-toggle-client-store.ts
```diff
@@ -93,6 +96,11 @@ export default class FeatureToggleClientStore
                 'fs.feature_name',
                 'features.name',
             )
+            .leftJoin(
+                this.db('feature_tag').select('*').as('ft'),
```
sjaanus, line 100: This join seems weird. I think you could just do this:
```

 .leftJoin(
        'feature_tag as ft',
        'ft.feature_name',
        'features.name',
         )
```

src/lib/db/feature-toggle-client-store.ts
```diff
@@ -203,6 +222,20 @@ export default class FeatureToggleClientStore
...
     }
 
+    private isUnseenTag(
+        feature: PartialDeep<IFeatureToggleClient>,
+        row: Record<string, any>,
+    ): boolean {
+        return (
+            row.tag_type &&
+            row.tag_value &&
+            !feature.tags.find(
```
sjaanus, line 232: Probably better to use `!some` instead of returning object and converting to boolean.

src/lib/db/feature-strategy-store.ts
```diff
@@ -317,6 +318,27 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
     }
 
+    private isUnseenTag(
+        feature: Record<string, any>,
+        row: Record<string, any>,
+    ): boolean {
+        return (
+            row.tag_type &&
+            row.tag_value &&
+            !feature.tags.find(
```
sjaanus, line 328: Here also `!some`

src/lib/db/feature-toggle-client-store.ts
nunogois: Thanks, mostly artifacts left from copy/pasting from other places. Addressed in https://github.com/Unleash/unleash/pull/2396/commits/22eff75b2cdf902b20a1da8df2d94a5259330046

src/lib/db/feature-strategy-store.ts
nunogois: Thanks, mostly artifacts left from copy/pasting from other places. Addressed in https://github.com/Unleash/unleash/pull/2396/commits/22eff75b2cdf902b20a1da8df2d94a5259330046

src/lib/db/feature-toggle-client-store.ts
```diff
@@ -93,6 +96,7 @@ export default class FeatureToggleClientStore
                 'fs.feature_name',
                 'features.name',
             )
+            .leftJoin('feature_tag as ft', 'ft.feature_name', 'features.name')
```
ivarconr, line 99: I am unsure if we should do this join for client SDK calls as well, I am considering the unknown implication on performance. Wdyt @chriswk?

ivarconr:
This took a slightly different approach than I imagined. It is a simpler and even "cleaner" solution. The only downside i can think about is a potential risk of performance, particular for the client SDK. 

Maybe we should put this behind a feature flag? I think then we need to use the flagResolver to determine if we will do the join in the db and use the same flag in the ui to turn on/off the tags column.

src/lib/types/model.ts
```diff
@@ -68,6 +68,7 @@ export interface IFeatureToggleClient {
     impressionData?: boolean;
     lastSeenAt?: Date;
     createdAt?: Date;
+    tags: ITag[];
```
ivarconr, line 71: I think we should make this optional.

nunogois:
> This took a slightly different approach than I imagined. It is a simpler and even "cleaner" solution. The only downside i can think about is a potential risk of performance, particular for the client SDK.
> 
> Maybe we should put this behind a feature flag? I think then we need to use the flagResolver to determine if we will do the join in the db and use the same flag in the ui to turn on/off the tags column.

Thanks @ivarconr! Addressed this in https://github.com/Unleash/unleash/pull/2396/commits/9f2c448828732eafde830596245a448e9441e299.

Besides adding the flag, I added `isAdmin` to the condition. Hopefully that covers concerns related to client SDK calls. Since the UI already adapts automatically when we have no tags returned, I didn't have to worry about the flag on the frontend.

frontend/src/interfaces/featureToggle.ts
```diff
@@ -7,6 +8,7 @@ export interface IFeatureToggleListItem {
     lastSeenAt?: string;
     createdAt: string;
     environments: IEnvironments[];
+    tags: ITag[];
```
ivarconr, line 11: probably optional?

src/lib/db/feature-strategy-store.ts
```diff
@@ -345,12 +386,28 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
+
+        const rows = await query;
+
         if (rows.length > 0) {
             const overview = rows.reduce((acc, r) => {
                 if (acc[r.feature_name] !== undefined) {
                     acc[r.feature_name].environments.push(
                         FeatureStrategiesStore.getEnvironment(r),
                     );
+                    if (this.isUnseenTag(acc[r.feature_name], r)) {
```
ivarconr, line 406: why do it need to be unseenTag?

src/lib/db/feature-strategy-store.ts
```diff
@@ -361,10 +418,25 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
                         environments: [
                             FeatureStrategiesStore.getEnvironment(r),
                         ],
+                        tags: [],
                     };
+                    if (this.isUnseenTag(acc[r.feature_name], r)) {
```
ivarconr, line 423: why do it need to be unseenTag?

src/lib/db/feature-strategy-store.ts
```diff
@@ -361,10 +418,25 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
+                        acc[r.feature_name].tags.push(
+                            FeatureStrategiesStore.getTag(r),
+                        );
+                    }
                 }
                 return acc;
             }, {});
+
+            if (!this.flagResolver.isEnabled('toggleTagFiltering')) {
```
ivarconr, line 432: I do not understand why we need this? It would not be added if we do not do the left join? Cant we just add the tags property the first time we try to push to the list?

src/lib/db/feature-toggle-client-store.ts
```diff
@@ -134,12 +156,16 @@ export default class FeatureToggleClientStore
         const featureToggles = rows.reduce((acc, r) => {
             let feature: PartialDeep<IFeatureToggleClient> = acc[r.name] ?? {
                 strategies: [],
+                tags: [],
```
ivarconr, line 159: should we drop this and add it for the first tag instead?

src/lib/db/feature-toggle-client-store.ts
```diff
@@ -134,12 +156,16 @@ export default class FeatureToggleClientStore
...
             };
             if (this.isUnseenStrategyRow(feature, r)) {
                 feature.strategies.push(
                     FeatureToggleClientStore.rowToStrategy(r),
                 );
             }
+            if (this.isUnseenTag(feature, r)) {
+                feature.tags.push(FeatureToggleClientStore.rowToTag(r));
```
ivarconr, line 167: Maybe:

```
const tags = feature.tags || [];
const newTag = FeatureToggleClientStore.rowToTag(r);
feature.tags = [...tags, newTag];
```

src/lib/db/feature-toggle-client-store.ts
```diff
@@ -173,6 +199,10 @@ export default class FeatureToggleClientStore
             FeatureToggleClientStore.removeIdsFromStrategies(features);
         }
 
+        if (!isAdmin || !this.flagResolver.isEnabled('toggleTagFiltering')) {
```
ivarconr, line 202: we should not do this.

nunogois:
@ivarconr Since we're following this `left join` approach to get everything in a single query, it's important to dedupe the tags:
image

I agree with the rest of the comments and tried to address them in https://github.com/Unleash/unleash/pull/2396/commits/f5971786fa2507256c6a151c3b4de86e3c72cde6 - Hopefully this looks a bit clearer 🙂

src/lib/db/feature-strategy-store.ts
nunogois: See: https://github.com/Unleash/unleash/pull/2396#issuecomment-1313539524

src/lib/db/feature-strategy-store.ts
nunogois: See: https://github.com/Unleash/unleash/pull/2396#issuecomment-1313539524

ivarconr:
> @ivarconr Since we're following this `left join` approach to get everything in a single query, it's important to dedupe the tags: image
> 
> I agree with the rest of the comments and tried to address them in [f597178](https://github.com/Unleash/unleash/commit/f5971786fa2507256c6a151c3b4de86e3c72cde6) - Hopefully this looks a bit clearer slightly_smiling_face

thanks for the clarifications. You could consider using a set as the container.

ivarconr (approved):

PR #2404, opened 2022-11-11
fix: project access table

FredrikOseberg:
## About the changes

* Fixes an issue where project access would overflow after moving it to project settings.

Closes #

### DIFF
--- a/frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
+++ b/frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
@@ -75,13 +75,8 @@ const StyledGroupAvatar = styled(UserAvatar)(({ theme }) => ({
     outline: `${theme.spacing(0.25)} solid ${theme.palette.background.paper}`,
 }));
 
-const hiddenColumnsSmall = [
-    'imageUrl',
-    'username',
-    'role',
-    'added',
-    'lastLogin',
-];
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
```diff
@@ -120,29 +116,7 @@ export const ProjectAccessTable: VFC = () => {
...
-                                subtitle={`${row.entity.users?.length} users`}
-                            />
-                        }
-                        elseShow={<HighlightCell value={value} />}
-                    />
-                ),
-                minWidth: 100,
-                searchable: true,
-            },
+
```
sjaanus, line 119: Would this not break displaying of group?

sjaanus: Currently looks like this
image

sjaanus (approved):
Looks good for now, but Last Login feature was actually quite nice. Maybe we can handle the width differently somehow in follow up PR. For example some columns definitely dont need 150px. So instead reduce them.

FredrikOseberg:
> Looks good for now, but Last Login feature was actually quite nice. Maybe we can handle the width differently somehow in follow up PR. For example some columns definitely dont need 150px. So instead reduce them.

Agree.  I will try to tweak it.

PR #2407, opened 2022-11-11
fix: improve drag behaviour when using handle

nunogois:
Fixes a small bug that caused the item to be draggable when clicking outside of the handle, when using a handle. Basically resets the draggable state when (re)entering the container.

### DIFF
--- a/frontend/src/hooks/useDragItem.ts
+++ b/frontend/src/hooks/useDragItem.ts
@@ -74,6 +74,9 @@ const addEventListeners = (
 
     handleEl.addEventListener('mouseenter', onMouseEnter);
     handleEl.addEventListener('mouseleave', onMouseLeave);
+    if (handle) {
+        el.addEventListener('mouseenter', onMouseLeave);
+    }
     el.addEventListener('dragstart', onDragStart);
     el.addEventListener('dragenter', onDragEnter);
     el.addEventListener('dragover', onDragOver);
@@ -82,6 +85,9 @@ const addEventListeners = (
     return () => {
         handleEl.removeEventListener('mouseenter', onMouseEnter);
         handleEl.removeEventListener('mouseleave', onMouseLeave);
+        if (handle) {
+            el.removeEventListener('mouseenter', onMouseLeave);
+        }
         el.removeEventListener('dragstart', onDragStart);
         el.removeEventListener('dragenter', onDragEnter);
         el.removeEventListener('dragover', onDragOver);


frontend/src/hooks/useDragItem.ts
```diff
@@ -74,6 +74,9 @@ const addEventListeners = (
 
     handleEl.addEventListener('mouseenter', onMouseEnter);
     handleEl.addEventListener('mouseleave', onMouseLeave);
+    if (handle) {
+        el.addEventListener('mouseenter', onMouseLeave);
```
gastonfournier, line 78: Question: should line [75](https://github.com/Unleash/unleash/pull/2407/files#diff-f24994d074d568c8e4c916a113e406830a5e76c4eb249fe5dcfa46118e4a1b4fR75) be on the else statement?

nunogois: Hmm I don't think it should, but I'm curious about your reasoning. We want `onMouseEnter` on `handleEl` every time - Whenever the mouse enters the element that will act as our handle, whether that's a specific "handle" or our whole element, we want it to enable the draggable state:
https://github.com/Unleash/unleash/blob/d8ffbe2b6adc5d097aebb798927d8dd69dd73914/frontend/src/hooks/useDragItem.ts#L47-L51

What this PR adds, is that whenever we're using the `handle` parameter (which means drag only on a specific "handle", not the whole element), we want to reset the draggable state when we enter (or re-enter) the container `el`.
image

So:
 - Entering our `handleEl` should enable the draggable state (line 75) - This applies whether we're using a specific "handle" or not;
 - Entering (or re-entering) the container `el`, our whole element, should reset the state (line 78), but only when using a specific "handle", otherwise it breaks the default behaviour of the whole thing being draggable;
 
 Maybe it's easier to understand if you checkout the branch and try it yourself. To toggle between "drag only on handle" and "everything draggable" you can comment this line: https://github.com/Unleash/unleash/blob/d8ffbe2b6adc5d097aebb798927d8dd69dd73914/frontend/src/component/environments/EnvironmentTable/EnvironmentRow/EnvironmentRow.tsx#L33

gastonfournier: Oh! Thanks for the clarification! I didn't realize they were targeting different elements. The names didn't help, but I think it's not easy to give proper names to elements in the UI, it becomes too specific.

gastonfournier (approved):

PR #2416, opened 2022-11-14
refactor: move docs into new structure / fix links for SEO

thomasheartman:
## What

This (admittedly massive) PR updates the "physical" documentation structure and fixes url inconsistencies and SEO problems reported by marketing. The main points are:

- remove or move directories : advanced, user_guide, deploy, api
- move the files contained within to the appropriate one of topics, how-to, tutorials, or reference
- update internal doc links and product links to the content
- create client-side redirects for all the urls that have changed.

A number of the files have been renamed in small ways to better match their url and to make them easier to find. Additionally, the top-level api directory has been moved to /reference/api/legacy/unleash (see the discussion points section for more on this).

## Why

When moving our doc structure to diataxis a while back, we left the "physical' files lying where they were, because it didn't matter much to the new structure. However, that did introduce some inconsistencies with where you place docs and how we organize them. 

There's also the discrepancies in whether urls us underscores or hyphens (which isn't necessarily the same as their file name), which has been annoying me for a while, but now has also been raised by marketing as an issue in terms of SEO.

## Discussion points

The old, hand-written API docs have been moved from /api to /reference/api/legacy/unleash. There _is_ a /reference/api/unleash directory, but this is being populated by the OpenAPI plugin, and mixing those could only cause trouble. However, I'm unsure about putting /legacy/ in the title, because the API isn't legacy, the docs are. Maybe we could use another path? Like /old-docs/ or something? I'd appreciate some input on this.

----

Finally: I know this is a massive PR, but it didn't make a lot of sense to only move some files at a time, because they're all so interconnected. That'd mean that we'd have to update the same links again and again and again in different PRs. However, the commits are (roughly) done on a directory by directory basis, so it might be possible to split up if you want it.


website/sidebars.js
```diff
@@ -308,29 +314,29 @@ module.exports = {
...
-                        'user_guide/technical_debt',
-                        'user_guide/unleash_context',
-                        'sdks/unleash-proxy',
+                        'reference/sso',
+                        'reference/stickiness',
+                        'reference/strategy-constraints',
+                        'reference/tags',
+                        'reference/technical-debt',
+                        'reference/unleash-context',
+                        'reference/unleash-proxy',
```
thomasheartman, line 339: This is the quickest and easiest way to see how everything was renamed ☝🏼

sighphyre (approved):
Gosh this is a monster of a PR! Probably too big to really give more meaningful input outside of the rationale and approach. Those do look good to me though. What's the plan for testing this beast once it goes in?

PR #2418, opened 2022-11-14
Permission guards in existing endpoints interacting with feature toggle configuration

sjaanus:
This PR adds permission guards for operations.

1. Toggling feature flag
2. Adding a strategy
3. Updating a strategy
4. Deleting a strategy

### DIFF
--- a/src/lib/db/access-store.ts
+++ b/src/lib/db/access-store.ts
@@ -27,6 +27,7 @@ const T = {
     ROLE_PERMISSION: 'role_permission',
     PERMISSIONS: 'permissions',
     PERMISSION_TYPES: 'permission_types',
+    CHANGE_REQUEST_SETTINGS: 'change_request_settings',
 };
 
 interface IPermissionRow {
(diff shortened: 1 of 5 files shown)


src/lib/db/access-store.ts
```diff
@@ -448,4 +449,17 @@ export class AccessStore implements IAccessStore {
             [destinationEnvironment, sourceEnvironment],
         );
     }
+
+    async isChangeRequestsEnabled(
+        project: string,
+        environment: string,
+    ): Promise<boolean> {
+        const result = await this.db.raw(
+            `SELECT EXISTS(SELECT 1 FROM ${T.CHANGE_REQUEST_SETTINGS}
```
kwasniew, line 458: why do we need double select? can't we just do one and check it a result or null was returned and map it to boolean?

src/lib/services/feature-toggle-service.ts
```diff
@@ -314,51 +314,58 @@ class FeatureToggleService {
         const { featureName, projectId, environment } = context;
         await this.validateFeatureContext(context);
 
-        if (strategyConfig.constraints?.length > 0) {
-            strategyConfig.constraints = await this.validateConstraints(
-                strategyConfig.constraints,
-            );
-        }
+        if (!(await this.changeRequestsEnabled(projectId, environment))) {
```
kwasniew, line 317: double nested if is usually a pattern of unnecessary complexity. Can we do early exit with the opposite condition?
if(changeRequestEnabled) throw Error

kwasniew (changes requested):

kwasniew (approved):

PR #2419, opened 2022-11-14
task: include variants per environment when cloning a feature

gastonfournier:
## About the changes
Allow cloning features with variants per environment. Pending on other changes in src/lib/db/feature-toggle-store.ts that need to override mine


sighphyre (approved):
G

PR #2420, opened 2022-11-14
feat: update strategy segments with edit / create strategy

FredrikOseberg:
## About the changes

* This PR refactors segments to be directly added when creating or editing a strategy. Instead of making a separate call to the segments API, we directly accept a list of segment ids and apply them when we receive the edit / create call on the strategy. 

Discussion points: 
* To not break existing API, the segmentIds are optional. In the controller we default the segmentIds if they don't exist.
* There is some uglyness in this code, fixing it will be more invasive than I'd like in this PR. But something to look into for the future.

Closes #


andreas-unleash (approved):
LG

sjaanus (approved):
LGTM!

PR #2422, opened 2022-11-14
POC: integration tests

kwasniew:
## About the changes

POC for frontent UI unit tests that test at similar level to Cypress but can be written in unit test framework and run in jsdom.
This example verifies if we can show ChangeRequest banner and if we get Change Request dialog on toggle switch.

Basic idea:
* setup msw returning required data (better option would be to Dependency Inject but it requires bigger refactoring)
* prepare context for rendering (providers for routing, theme, access etc.)
* invoke UI actions as you'd in cypress - we can build page object abstraction etc. but that's not the point of this PR

## Discussion points

This PR also fixed SWR for enterprise and bring useEnterpriseSWR hook that is drop in replacement for SWR observing changes on isOSS


nunogois (approved):
LGTM 👍

PR #2430, opened 2022-11-15
fix: copy strategy change request

FredrikOseberg:
## About the changes

* Fixes a bug where change request configured was instead governed by the feature flag

Closes #


sjaanus (approved):
LGTM!

PR #2433, opened 2022-11-15
fix: update text in environments

nunogois:
Updates the text in some places related with the new environments updates. Hopefully this makes things even clearer 🙂


NicolaeUnleash (approved):
LGTM.

PR #2435, opened 2022-11-15
Define exports for enterprise

andreas-unleash:
This PR sets up exports so that we can import in enterprise with just "unleash-server".
This will free us to refactor unleash internals without breaking enterprise

## About the changes

Closes #


kwasniew (approved):
I understand that in a transitional period we ca use both new and old approach

PR #2438, opened 2022-11-15
Add variants per env gaston

gastonfournier:
Diff I worked on

### DIFF
--- a/src/lib/db/project-store.ts
+++ b/src/lib/db/project-store.ts
@@ -178,7 +178,7 @@ class ProjectStore implements IProjectStore {
             .returning(COLUMNS)
             .onConflict('id')
             .ignore();
-        if (rows.length > 0) {
+        if (environments && rows.length > 0) {
             environments.forEach((env) => {
                 projects.forEach(async (project) => {
                     await this.addEnvironmentToProject(project.id, env.name);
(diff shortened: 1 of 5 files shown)


src/lib/services/feature-toggle-service.ts
```diff
@@ -165,7 +165,7 @@ class FeatureToggleService {
         projectId,
     }: IFeatureContext): Promise<void> {
         const id = await this.featureToggleStore.getProjectId(featureName);
-        if (id !== projectId) {
+        if (id && id !== projectId) {
```
gastonfournier, line 168: Don't fail if we don't find a project id associated with the feature, instead, let the application return not found.

src/lib/db/project-store.ts
```diff
@@ -178,7 +178,7 @@ class ProjectStore implements IProjectStore {
             .returning(COLUMNS)
             .onConflict('id')
             .ignore();
-        if (rows.length > 0) {
+        if (environments && rows.length > 0) {
```
gastonfournier, line 181: This doesn't affect any test if removed, but at some point, I got an error here and it seems to be fine, if environments is not defined but the SQL statement inserts some rows, the next line will fail, so I think it still makes sense but is one of those unneeded changes that may add noise to our not so small set of changes

PR #2447, opened 2022-11-16
Check assumption·

gastonfournier:
## About the changes
After importing we should have an enabled feature in the environment, but it seems we're not:
```
[
   {
      "enabled":false,
      "name":"state-visible-environment",
      "sortOrder":9999,
      "strategies":[
         {
            "constraints":[
               
            ],
            "id":"2ea91298-4565-4db2-8a23-50757001a076",
            "name":"gradualRolloutRandom",
            "parameters":{
               "percentage":"100"
            },
            "sortOrder":9999
         }
      ],
      "type":"production"
   }
]
```

Adding an extra check to validate that shows that the feature is not enabled in the environment.

### DIFF
--- a/src/test/e2e/api/admin/state.e2e.test.ts
+++ b/src/test/e2e/api/admin/state.e2e.test.ts
@@ -463,4 +463,5 @@ test(`should not show environment on feature toggle, when environment is disable
 
     expect(body.environments).toHaveLength(1);
     expect(body.environments[0].name).toBe('state-visible-environment');
+    expect(body.environments[0].isEnabled).toBeTruthy();
 });


src/test/e2e/api/admin/state.e2e.test.ts
```diff
@@ -463,4 +463,5 @@ test(`should not show environment on feature toggle, when environment is disable
 
     expect(body.environments).toHaveLength(1);
     expect(body.environments[0].name).toBe('state-visible-environment');
+    expect(body.environments[0].enabled).toBeTruthy();
```
gastonfournier, line 466: From line 464:466, this is what we should test:
```javascript
    expect(body.environments).toHaveLength(2);
    expect(body.environments[0].name).toBe('state-visible-environment');
    expect(body.environments[0].enabled).toBeTruthy();
    expect(body.environments[1].name).toBe('state-hidden-environment');
    expect(body.environments[1].enabled).toBeFalsy();
```

src/lib/db/feature-environment-store.ts
```diff
@@ -217,11 +217,17 @@ export class FeatureEnvironmentStore implements IFeatureEnvironmentStore {
     async connectProject(
         environment: string,
         projectId: string,
+        idempotent?: boolean, // default false to respect old behavior
```
gastonfournier, line 220: Other tests depend on this method fail when the project is already connected to the environment. Personally, I think that this method could be independent and the other methods depending on its failure can be rewritten (there's only one [here](https://github.com/Unleash/unleash/blob/38f351871918a98ea73dedade6a966d92b608369/src/lib/services/environment-service.ts#L80-L100))

src/lib/db/feature-environment-store.ts
```diff
@@ -258,17 +264,22 @@ export class FeatureEnvironmentStore implements IFeatureEnvironmentStore {
     async connectFeatureToEnvironmentsForProject(
         featureName: string,
         projectId: string,
+        config?: [{ environment: string; enabled: boolean }],
```
gastonfournier, line 267: This adds support to pass the desired state of the feature in that environment, which is required for this assertion: https://github.com/Unleash/unleash/pull/2447/files#diff-696f6242d5d467fdce764b004ca9dda90678e55f5051153dd5196d21a9c49e03R463

src/lib/services/state-service.ts
```diff
@@ -200,6 +200,30 @@ export default class StateService {
                 dropBeforeImport,
                 keepExisting,
             });
+
+            if (
+                featureEnvironments
+                    .map((fe) => fe.environment)
+                    .filter((eName) => eName !== 'default').length > 0
+            ) {
```
gastonfournier, line 208: This is hairy: we want to create this link only in very specific cases. We can't rely on the version number, but this situation is likely to happen if you have multiple environments. I'd like some input on this one @chriswk and @sighphyre

src/lib/services/state-service.ts
```diff
@@ -266,6 +290,7 @@ export default class StateService {
                         this.featureEnvironmentStore.connectFeatureToEnvironmentsForProject(
                             env.featureName,
                             id,
+                            featureEnvironments,
```
gastonfournier, line 293: Send configurations to set the desired "enabled state" of the feature in each environment (note that `connectFeatureToEnvironmentsForProject` will iterate and create connections for all environments)

src/test/e2e/api/admin/state.e2e.test.ts
```diff
@@ -452,15 +452,15 @@ test(`should not show environment on feature toggle, when environment is disable
         .attach('file', 'src/test/examples/import-state.json')
         .expect(202);
 
-    await app.request
-        .post('/api/admin/projects/default/environments')
-        .send({ environment: 'state-visible-environment' })
-        .expect(200);
```
gastonfournier, line 458: This method was connecting the environments in the test, but when customers run the import and not this method they may run into this issue

src/test/examples/import-state.json
```diff
@@ -28,26 +28,26 @@
...
+      "name": "development",
+      "type": "development",
+      "displayName": "Dev"
     },
     {
-      "name": "state-hidden-environment",
+      "name": "production",
       "type": "production",
-      "displayName": "Hidden"
+      "displayName": "Prod"
```
gastonfournier, line 38: All environments are enabled by default, renaming them to avoid misleading terms such as `visible` or `hidden` cause both environments are enabled and visible

src/lib/services/state-service.ts
gastonfournier: So, in the end, there was an issue with my logic to find whether the feature should be enabled in the environment or not, which I found thanks to @chriswk suggestion to refactor the code. After I fixed that, I can now remove the restriction I had here

src/lib/db/feature-environment-store.ts
```diff
@@ -258,6 +264,7 @@ export class FeatureEnvironmentStore implements IFeatureEnvironmentStore {
     async connectFeatureToEnvironmentsForProject(
         featureName: string,
         projectId: string,
+        enabledIn: { [environment: string]: boolean } = {},
```
gastonfournier, line 267: This adds support to pass the desired state of the feature in that environment, which is required for this assertion: https://github.com/Unleash/unleash/pull/2447/files#diff-696f6242d5d467fdce764b004ca9dda90678e55f5051153dd5196d21a9c49e03R463

We were always setting the relationship to `enabled: false` https://github.com/Unleash/unleash/pull/2447/files#diff-299d69b93e5da7b84238a0d5564d36d4129bee60cfce0e0e84511a00375e04a7L271

andreas-unleash (approved):
LG!

sjaanus (approved):
LGTM!

PR #2449, opened 2022-11-16
stop making enterprise calls in oss

kwasniew:
## About the changes

We were fetching change request config and change request status in OSS version. This PR stops this behavior.

Design decisions:
* keep the change surface area small - don't touch most of the existing code
* make fetcher return a fallback value for OSS


FredrikOseberg (approved):
LGTM

Maybe we could make an abstraction in the future, this will probably happen again. But for now, this is simple enough.

PR #2451, opened 2022-11-16
fix: make the tags column have a fixed width

nunogois:
By setting a fixed width in the tags column we can save some extra space.


chriswk (approved):
Seems reasonable

PR #2459, opened 2022-11-17
Updated personal access tokens for accuracy

JarrodSC:
Updated **personal access tokens** section to reflect that users technically can set the expiration date to **never**.

## About the changes

Closes #

### DIFF
--- a/website/docs/reference/api-tokens-and-client-keys.mdx
+++ b/website/docs/reference/api-tokens-and-client-keys.mdx
@@ -61,17 +61,19 @@ Support for scoped admin tokens with more fine-grained permissions is currently 
 
 When using a personal access token to modify resources, the event log will list the token creator's name for that operation.
 
-All personal access tokens have a lifetime and **will stop working after the expiration** date.
+Personal access tokens with a lifetime **will stop working after the expiration date**.
 
 Use personal access tokens to:
 
 - Provide more fine-grained permissions for automation than an admin token provides
 - Give temporary access to an automation tool
 
+Note: While it is possible to set the expiration date to **never**, we don't recommend it. 
+
 Do **not** use personal access tokens for:
 
 -   [Client SDKs](../sdks/index.md): You will _not_ be able to read toggle data from multiple environments. Use [client tokens](#client-tokens) instead.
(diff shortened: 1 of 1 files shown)


website/docs/reference/api-tokens-and-client-keys.mdx
```diff
@@ -61,17 +61,19 @@ Support for scoped admin tokens with more fine-grained permissions is currently
...
 
-All personal access tokens have a lifetime and **will stop working after the expiration** date.
+Personal access tokens with a lifetime **will stop working after the expiration date**.
 
 Use personal access tokens to:
 
 - Provide more fine-grained permissions for automation than an admin token provides
 - Give temporary access to an automation tool
 
+Note: While it is possible to set the expiration date to **never**, we don't recommend it.
```
thomasheartman, line 71: As discussed, I think we should add some basic info about why we recommend that tokens always have an expiry date here. How's this sound? 

Also, I've wrapped this in a [docusaurus admonition](https://docusaurus.io/docs/markdown-features/admonitions) for more visibility. This helps call attention to it.

```suggestion
:::info On token expiry

It is possible to set a token's expiration date to **never**. However, a token that doesn't expire has the potential to cause a lot of harm if a bad actor gets hold of it, so we recommend that you use tokens with expiry dates whenever possible.

:::
```

thomasheartman (changes requested):
Super nice! Thanks for picking up on this and fixing it 💯  Before we merge it, I think we should explain _why_ we don't recommend using non-expiring tokens to the user. I've added a suggestion for that inline. Take a look at it, make the changes you deem appropriate, and then send me another review request when you're ready ☺️

website/docs/reference/api-tokens-and-client-keys.mdx
JarrodSC:
```suggestion
:::info On token expiration

It is possible to set a token's expiration date to **never**. However, a token that doesn't expire has the potential to cause a lot of harm if a bad actor gets hold of it, so we recommend that you use tokens with expiration dates whenever possible.

:::
```

thomasheartman (approved):
🥳 🚂

PR #2460, opened 2022-11-17
implement proxy all endpoint

andreas-unleash:

This PR implements the `all` endpoint of unleash-proxy, by adding an experimental flag that can control the behaviour

## About the changes

Closes #


thomasheartman (changes requested):
Nice! This looks pretty great 😄 I just have few questions and cases we should test inline. Let me know what you think 🙌🏼

FredrikOseberg (changes requested):
We dont control the query parameter in the SDK. Doing this with a query parameter would require us to make changes in all client SDKs. i would prefer if this was a setting in Unleash instead.

FredrikOseberg (approved):
LGTM. Just need to fix prettier to pass the checks.

thomasheartman (approved):
Nice work 💯

PR #2463, opened 2022-11-17
fix: flaky test on number of projects statistic

gastonfournier:
This test fails from time to time due to a race condition between the http statement and the creation of the new project.
Adding an await statement should fix this.


chriswk (approved):
LGTM

PR #2465, opened 2022-11-17
conditional message for change request strategy form

kwasniew:
## About the changes

Green box message changes for change requests
Screenshot 2022-11-17 at 14 59 20

Design decisions:
* composition over configuration - I'm composing by passing children props instead of branching inside the child component


Tymek (approved):
nice

PR #2466, opened 2022-11-17
separate protected and unprotected routes

kwasniew:
## About the changes

With change requests functionality we need to separate feature toggle service methods exposed over API (protected) and service method calls that change requests service can invoke directly (unprotecte)


FredrikOseberg (approved):
LGTM

PR #2470, opened 2022-11-18
older safari does not support gap

kwasniew:
## About the changes

replacing gap with margins for older safari.
Screenshot 2022-11-18 at 09 55 41


andreas-unleash (approved):
LG

Issue #2254, opened 2022-10-25
🔥 feat: Variants per environment


ivarconr:
Currently variants is a global feature config shared across all environments. This is causing some friction, especially for larger customers with stricter access controls.

ivarconr:
This feature will be in beta from v.4.18

ivarconr:
This is still in progress. We are working on:

- integrating it with the "Change Request" feature 
- make it easy to manage variants across environments, to avoid unnecessary work guaranteeing they are not diverging.

ivarconr:
is GA from v4.21

PR #2471, opened 2022-11-18
feat: Add variants per env

gastonfournier:
## About the changes
Variants are now stored in each environment rather than in the feature toggle. This enables RBAC, suggest changes, etc to also apply to variants.

### Important files
- **src/lib/db/feature-strategy-store.ts** a complex query was moved to a view named `features_view`
- **src/lib/services/state-service.ts** export version number increased due to the new format


src/lib/db/feature-environment-store.ts
```diff
@@ -7,20 +7,22 @@ import {
...
 import { DB_TIME } from '../metric-events';
-import { IFeatureEnvironment } from '../types/model';
+import { IFeatureEnvironment, IVariant } from '../types/model';
 import NotFoundError from '../error/notfound-error';
 import { v4 as uuidv4 } from 'uuid';
 
 const T = {
     featureEnvs: 'feature_environments',
     featureStrategies: 'feature_strategies',
+    projectEnvs: 'project_environments',
```
gastonfournier, line 17: @sighphyre is this needed? I feel it's a leftover that can be removed
```suggestion
```

gastonfournier: Tested locally and it's not needed

src/migrations/20221107121635-move-variants-to-per-environment.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+exports.up = function (db, callback) {
```
sighphyre, line 3: I think we need to move this to be after the change request migrations so that it actually runs

sighphyre: Lemme check that quick...

sighphyre: Nevermind, Knex is smart enough to make sure this goes in

sighphyre (approved):
G

PR #2472, opened 2022-11-18
fix: flaky tests on segments order

gastonfournier:
## About the changes
This test fails randomly, example: https://github.com/Unleash/unleash/actions/runs/3496006349/jobs/5853415518#step:5:376 

It's not so frequent but I tested locally one out of 4 or 8 runs fails because the rows returned are in the wrong order.

## Discussion points
If the order of segments is relevant, we should look into how we're fetching them or applying a `sort()` before returning from the query. I've validated on my machine that even if the segments are sent in order, when pulled from the db sometimes they are processed in a different order

### DIFF
--- a/src/test/e2e/api/admin/project/features.e2e.test.ts
+++ b/src/test/e2e/api/admin/project/features.e2e.test.ts
@@ -2703,7 +2703,7 @@ test.only('should add multiple segments to a strategy', async () => {
                 (strat) => strat.id === strategyOne.id,
             );
 
-            expect(strategy.segments).toEqual([
+            expect(strategy.segments.sort()).toEqual([
                 segment.id,
                 segmentTwo.id,
                 segmentThree.id,


src/test/e2e/api/admin/project/features.e2e.test.ts
```diff
@@ -2703,7 +2703,7 @@ test.only('should add multiple segments to a strategy', async () => {
                 (strat) => strat.id === strategyOne.id,
             );
 
-            expect(strategy.segments).toEqual([
+            expect(strategy.segments.sort()).toEqual([
```
gastonfournier, line 2706: We're sorting in the test, so we're basically saying: the order is not important. If it is, we should consider a different change

gastonfournier:
Here's a sample on main: https://github.com/Unleash/unleash/actions/runs/3511158144/jobs/5881663038#step:7:370

ivarconr (approved):

PR #2474, opened 2022-11-18
Permission groups

Tymek:
## About the changes
Group permissions in role settings.
image

## Discussion points
Should we move this 'group' to the database in the future?

### DIFF
--- a/frontend/src/component/admin/projectRoles/ProjectRoleForm/PermissionAccordion/PermissionAccordion.tsx
+++ b/frontend/src/component/admin/projectRoles/ProjectRoleForm/PermissionAccordion/PermissionAccordion.tsx
@@ -48,7 +48,7 @@ export const PermissionAccordion: VFC<IEnvironmentPermissionAccordionProps> = ({
     permissions,
     checkedPermissions,
     Icon,
-    isInitiallyExpanded,
+    isInitiallyExpanded = false,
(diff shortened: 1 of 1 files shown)


frontend/src/component/admin/projectRoles/ProjectRoleForm/PermissionAccordion/PermissionAccordion.tsx
```diff
@@ -43,12 +43,36 @@ const StyledTitle = styled(StringTruncator)(({ theme }) => ({
...
+    }
+    if (name.endsWith('_STRATEGY')) {
+        return ['Strategies', 3];
+    }
+    if (name.endsWith('_CHANGE_REQUEST')) {
+        return ['Change requests', 4];
+    }
+
+    return ['Other', 5];
+};
```
FredrikOseberg, line 68: While this works now, it seems to introduce a complexity that can be problematic in the future. For example, what if we added a new permission tomorrow that ended with _STRATEGY, for example: `CHANGE_REQUEST_CONFLICT_RESOLTION_STRATEGY` (just making stuff up here). Now as the creator of a new permission you need to be aware that this naming convention grouping exists in the frontend. Over time it might be hard to manage. Perhaps we should consider a more explicit structure?

PR #2478, opened 2022-11-18
Docs/update unleash proxy img

NicolaeUnleash:
updating the image with the right details


thomasheartman (approved):
Love it! Thanks for fixing it 😁

PR #2493, opened 2022-11-22
add events and expose them for webhooks

kwasniew:
## About the changes

Added change request events and exposed them for webhooks

Also fixing toggle change calling protected method that was not allowing to apply changes in change requests.


FredrikOseberg (approved):
LGTM

PR #2495, opened 2022-11-22
fix: Pagination of event log

chriswk:
Currently, our event log is keeping all the previous data and then fetching and appending to our state list. This causes duplicates, since we're keeping the existing list and might be getting the same data, but with added events at the end.

This fix changes the logic to slice existing data at offset. Which should make sure we only append the new data.

Co-authored-by: Fredrik Strand Oseberg <fredrik.no@gmail.com>


FredrikOseberg (approved):
LGTM

PR #2497, opened 2022-11-22
Add possibility to soft delete users

sjaanus:
Previously we hard deleted the users, but due to change requests and possibly other features in future, we really want to hard-link user table and have meaningful relationships.

But this means, when user is deleted, all linked data is also deleted.
**Workaround is to soft delete users and just clear users data and keep the relationships alive for audit logs.**

This PR implements this feature.

### DIFF
--- a/src/lib/db/user-store.ts
+++ b/src/lib/db/user-store.ts
@@ -77,7 +77,7 @@ class UserStore implements IUserStore {
     }
 
     async update(id: number, fields: IUserUpdateFields): Promise<User> {
-        await this.buildBaseQuery()
+        await this.activeUsers()
             .where('id', id)
             .update(mapUserToColumns(fields));
         return this.get(id);
@@ -85,7 +85,7 @@ class UserStore implements IUserStore {
 
     async insert(user: ICreateUser): Promise<User> {
         const rows = await this.db(TABLE)
-            .insert({ ...mapUserToColumns(user), is_deleted: false })
+            .insert({ ...mapUserToColumns(user), deleted_at: null })
             .returning(USER_COLUMNS);
         return rowToUser(rows[0]);
     }
@@ -100,7 +100,7 @@ class UserStore implements IUserStore {
     }
 
     buildSelectUser(q: IUserLookup): any {
-        const query = this.buildBaseQuery();
+        const query = this.activeUsers();
         if (q.id) {
             return query.where('id', q.id);
         }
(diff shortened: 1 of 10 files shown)


src/migrations/20221121133546-soft-delete-user.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+            ALTER TABLE users ADD COLUMN IF NOT EXISTS is_deleted boolean not null default false;
```
kwasniew, line 6: it's better to have deleted_at with date because it opens more options in future

src/test/e2e/stores/user-store.e2e.test.ts
```diff
@@ -160,3 +153,14 @@ test('should always lowercase emails on updates', async () => {
     storedUser = await store.get(storedUser.id);
     expect(storedUser.email).toBe(updatedUser.email.toLowerCase());
 });
+
+test('should delete user', async () => {
+    const user = await stores.userStore.upsert({
+        email: 'deleteuser@mail.com',
+    });
+    await stores.userStore.delete(user.id);
```
kwasniew, line 161: I'd add space between given/when/then sections

src/lib/db/user-store.ts
```diff
@@ -111,31 +113,33 @@ class UserStore implements IUserStore {
         throw new Error('Can only find users with id, username or email.');
     }
 
+    buildBaseQuery(): any {
```
kwasniew, line 116: I'd rename it to what it does. BaseQuery is not very meaningful. Sth like activeUsers()

kwasniew (changes requested):

kwasniew (approved):

src/lib/services/access-service.ts
```diff
@@ -351,8 +351,12 @@ export class AccessService {
...
     }
 
-    async unlinkUserRoles(userId: number): Promise<void> {
-        return this.store.unlinkUserRoles(userId);
+    async wipeUserPermissions(userId: number): Promise<void> {
+        await this.store.unlinkUserRoles(userId);
+        await this.store.unlinkUserGroups(userId);
+        await this.store.clearUserPersonalAccessTokens(userId);
+        await this.store.clearPublicSignupUserTokens(userId);
+        return Promise.resolve();
```
FredrikOseberg, line 359: Why do we need this? An async function returns a promise by default.

Also: We are making 4 calls here. Could we do them in parallel? Something like:

```
const actions = [this.store.unlinkUserRoles, this.store.unlinkUserGroups, this.store.clearPersonaAccessTokens, this.store.clearPublichSignupUser]

await Promise.all(actions.map(action => action(userId)))
```

FredrikOseberg (changes requested):
Marking an opportunity for improvement, otherwise it looks good!

src/lib/services/access-service.ts
sjaanus: Made it parallel, but I am not sure about this mapping. It definitely does not help with readability and does not make it much shorter.

FredrikOseberg (approved):

PR #2498, opened 2022-11-22
fix: force clone feature to correctly copy variant environments

sighphyre:
This patches an issue where cloning a feature would clone variants for it's last found environment in the db. This now will clone the feature environments correctly

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -768,18 +768,30 @@ class FeatureToggleService {
         await this.validateName(newFeatureName);
 
         const cToggle =
-            await this.featureStrategiesStore.getFeatureToggleWithEnvs(
+            await this.featureStrategiesStore.getFeatureToggleWithVariantEnvs(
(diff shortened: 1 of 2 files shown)


src/lib/services/feature-toggle-service.ts
```diff
@@ -768,18 +768,30 @@ class FeatureToggleService {
...
-            await this.featureStrategiesStore.getFeatureToggleWithEnvs(
+            await this.featureStrategiesStore.getFeatureToggleWithVariantEnvs(
                 featureName,
             );
 
-        const newToggle = { ...cToggle, name: newFeatureName };
+        const newToggle = {
+            ...cToggle,
+            name: newFeatureName,
+            variants: undefined,
```
sighphyre, line 778: We do always return the variants for legacy purposes here, so we intentionally yeet them so that the creation doesn't create variants we don't want

gastonfournier (approved):
LG! And I loved the names in the test :clap:

PR #2501, opened 2022-11-22
fix cr banner

kwasniew:
## About the changes

Banner was not disappearing after apply and not changing status after approve.


FredrikOseberg (approved):
LGTM!

PR #2505, opened 2022-11-22
fix: improve UI in env variants, simplify code

nunogois:
https://linear.app/unleash/issue/2-446/ui-simplify-design-with-some-improvements


sighphyre (approved):
G

PR #2510, opened 2022-11-23
cancel changes

kwasniew:
## About the changes

ChangeRequest Author and Admin can Cancel ChangeRequest as long as it's not Applied or Cancelled.
Other users won't see a button.

Important:
* Cancel happens when the ChangeRequest is outside of a Draft state
* In Draft state we Discard ChangeRequest

Screenshot 2022-11-23 at 13 41 02


FredrikOseberg (approved):
LGTM

PR #2514, opened 2022-11-23
Feat update token seen at

nunogois:
https://linear.app/unleash/issue/2-448/update-last-seen-column-for-api-tokens

### DIFF
--- a/src/lib/db/api-token-store.ts
+++ b/src/lib/db/api-token-store.ts
@@ -203,7 +203,7 @@ export class ApiTokenStore implements IApiTokenStore {
         const now = new Date();
         try {
             await this.db(TABLE)
-                .whereIn('secrets', secrets)
+                .whereIn('secret', secrets)
                 .update({ seen_at: now });
         } catch (err) {
             this.logger.error('Could not update lastSeen, error: ', err);
--- a/src/lib/services/api-token-service.ts
+++ b/src/lib/services/api-token-service.ts
@@ -26,6 +26,7 @@ import {
     ApiTokenUpdatedEvent,
 } from '../types';
 import { omitKeys } from '../util';
+import { IFlagResolver } from 'lib/types/experimental';
 
 const resolveTokenPermissions = (tokenType: string) => {
     if (tokenType === ApiTokenType.ADMIN) {
@@ -60,6 +61,8 @@ export class ApiTokenService {
 
     private lastSeenSecrets: string[] = [];
 
+    private flagResolver: IFlagResolver;
+
     constructor(
         {
             apiTokenStore,
@@ -69,7 +72,10 @@ export class ApiTokenService {
             IUnleashStores,
             'apiTokenStore' | 'environmentStore' | 'eventStore'
         >,
-        config: Pick<IUnleashConfig, 'getLogger' | 'authentication'>,
+        config: Pick<
+            IUnleashConfig,
+            'getLogger' | 'authentication' | 'flagResolver'
+        >,
     ) {
         this.store = apiTokenStore;
         this.eventStore = eventStore;
@@ -89,6 +95,7 @@ export class ApiTokenService {
                 this.initApiTokens(config.authentication.initApiTokens),
             );
         }
+        this.flagResolver = config.flagResolver;
     }
 
     async fetchActiveTokens(): Promise<void> {
@@ -101,7 +108,10 @@ export class ApiTokenService {
     }
 
     async updateLastSeen(): Promise<void> {
-        if (this.lastSeenSecrets.length > 0) {
+        if (
+            this.lastSeenSecrets.length > 0 &&
+            this.flagResolver.isEnabled('tokensLastSeen')
+        ) {
             await this.store.markSeenAt(this.lastSeenSecrets);
             this.lastSeenSecrets = [];
         }
@@ -152,7 +162,9 @@ export class ApiTokenService {
         }
 
         if (token) {
-            this.lastSeenSecrets.push(token.secret);
+            if (this.flagResolver.isEnabled('tokensLastSeen')) {
+                this.lastSeenSecrets.push(token.secret);
+            }
 
             return new ApiUser({
                 username: token.username,
--- a/src/lib/services/api-token-service.ts
+++ b/src/lib/services/api-token-service.ts
@@ -108,7 +108,7 @@ export class ApiTokenService {
 
     async updateLastSeen(): Promise<void> {
         if (this.lastSeenSecrets.size > 0) {
-            let toStore = [...this.lastSeenSecrets];
+            const toStore = [...this.lastSeenSecrets];
             this.lastSeenSecrets = new Set<string>();
             await this.store.markSeenAt(toStore);
         }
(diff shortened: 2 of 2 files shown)


src/lib/db/api-token-store.ts
```diff
@@ -200,7 +200,7 @@ export class ApiTokenStore implements IApiTokenStore {
         const now = new Date();
         try {
             await this.db(TABLE)
-                .whereIn('secrets', secrets)
+                .whereIn('secret', secrets)
```
gastonfournier, line 203: Is this a bug unrelated to this PR? Just curious

nunogois: Yep, this `markSeenAt` method was already implemented from a while ago, git says March 2021, except it was not being used AFAIK and had this typo. So I guess I'm following the approach that was originally thought of, with the `secrets string[]` parameter and all.

src/lib/services/api-token-service.ts
```diff
@@ -46,15 +47,25 @@ export class ApiTokenService {
 
     private timer: NodeJS.Timeout;
 
+    private seenTimer: NodeJS.Timeout;
+
     private activeTokens: IApiToken[] = [];
 
+    private lastSeenSecrets: string[] = [];
```
gastonfournier, line 54: One problem with this approach could be memory consumption, cause we might end up with a large array in memory. I think it'd be better to have a Map structure and then use the keys to obtain the last seen secrets.

nunogois: The array gets cleared in https://github.com/Unleash/unleash/pull/2514/files/e31c6d6296e7b5317e86e502abbbe4ff158c0b76#diff-c7747d86059ec3422885e79411b8ef7323c0804df081c110268dd3250ace3696R102 - But we can also try a Map approach.

nunogois: I remembered that you may be thinking about having an unique array (don't have the same token more than once in the array). I agree with that and I think we could maybe use a Set here.

src/lib/services/api-token-service.ts
```diff
@@ -79,6 +96,13 @@ export class ApiTokenService {
         }
     }
 
+    async updateLastSeen(): Promise<void> {
+        if (this.lastSeenSecrets.length > 0) {
+            await this.store.markSeenAt(this.lastSeenSecrets);
+            this.lastSeenSecrets = [];
```
gastonfournier, line 102: We could lose some last seen in between due to the interleaving of different operations. I don't know if there's an atomic way of doing this... you may know better, but I googled this:
```typescript
let toStore = [];
[this.lastSeenSecrets, toStore] = [toStore, this.lastSeenSecrets];
await this.store.markSeenAt(toStore);
```

nunogois: Interesting! That could work, thanks 👍

src/lib/services/api-token-service.ts
gastonfournier: Yes, a set would work as well, the problem is with high-throughput applications, say 1000 req/s we'd be queuing 60K strings in memory every minute (which is 4.2Mb considering a token size of 70 characters). It's not much but if we're speaking of a single token, it'd be much better to have only 70 bytes in memory :D

nunogois: Thanks, tried addressing this in https://github.com/Unleash/unleash/pull/2514/commits/ed33001b04d69c056ed25827d2530a777e88ecf1 - Let me know what you think 👍 
I'll do the same change in the PAT PR.

src/lib/services/api-token-service.ts
```diff
@@ -63,6 +74,12 @@ export class ApiTokenService {
             () => this.fetchActiveTokens(),
             minutesToMilliseconds(1),
         ).unref();
+        if (this.flagResolver.isEnabled('tokensLastSeen')) {
+            this.seenTimer = setInterval(
+                () => this.updateLastSeen(),
```
gastonfournier, line 79: Should we await here? Something like:
```suggestion
                () => await this.updateLastSeen(),
```
The concern is: if `updateLastSeen` executes asynchronously, some updateLastSeen might overlap over time (i.e. run 2 in parallel). It'd take quite a lot of time to happen with 1 minute between operations, but I've seen this before. 

I usually wait until the previous operation finishes before triggering the timer again. I'm not sure how this is done in node (or any best practice). Just wanted to touch base on this but it's not a blocker

gastonfournier (commented):
This looks great IMO, just a small question but I might be overthinking it.

src/lib/services/api-token-service.ts
nunogois: I don't think it works like this, as it still doesn't ensure sequentiality, but I may be wrong.
What do you think about an approach like https://github.com/Unleash/unleash/pull/2514/commits/47a1f0a8b8da574c23253076b1941e6231b516b6 inspired by: https://github.com/Unleash/unleash/blob/a992aca228c3bc0c17abcac22d027796db8059c0/src/lib/proxy/proxy-repository.ts#L91-L97

gastonfournier: Yes, I read a bit about the topic and your change is in line with my intuition: setTimeout is less risky (operationally speaking) cause there's no risk of executing multiple instances of the code concurrently. It does lag if the function takes longer, but that's exactly what I'd like to see in this particular use case: to execute less frequently in situations of high CPU load or other situations that may cause this function to execute slower.

Resources:
- https://blog.skay.dev/when-to-use-settimeout-vs-setinterval#conclusion
- https://stackoverflow.com/questions/729921/settimeout-or-setinterval

src/lib/db/api-token-store.ts
ivarconr: nice catch! 🗡️

src/lib/services/api-token-service.ts
```diff
@@ -92,6 +106,19 @@ export class ApiTokenService {
         }
     }
 
+    async updateLastSeen(): Promise<void> {
+        if (this.lastSeenSecrets.size > 0) {
+            let toStore = new Set<string>();
+            [this.lastSeenSecrets, toStore] = [toStore, this.lastSeenSecrets];
+            await this.store.markSeenAt([...toStore]);
```
ivarconr, line 113: This syntax makes me a bit nervous, as it requires the next line to be a spread operator. Remember that node processes are single threaded, so we can have two lines in a row and assume order (as long as they are not async). 

**How about:**

```suggestion
            let toStore = [...this.lastSeenSecrets];
            this.lastSeenSecrets = new Set<string>();
            await this.store.markSeenAt(toStore);
```

src/lib/services/api-token-service.ts
```diff
@@ -52,10 +53,16 @@ export class ApiTokenService {
 
     private timer: NodeJS.Timeout;
 
+    private seenTimer: NodeJS.Timeout;
```
ivarconr, line 56: We could have made the precious timer in to an array, we only need to keep the reference to be able to clear them all.

nunogois: Hmm I'm not sure I understood this one, could you please clarify? 
I guess they could be a `NodeJS.Timeout[]` but if we want a single reference to each we would store them in `[0]` and `[1]` I guess? 🤔 Besides one of them is an interval and the other a timeout. 
Seems like decreased readability without much to gain, so I'm probably missing something.

PR #2515, opened 2022-11-23
changes count

kwasniew:
## About the changes

Change count should reflect number of changes not number of feature toggles

Screenshot 2022-11-23 at 16 40 57


sjaanus (approved):
LGTM!

PR #2516, opened 2022-11-23
feat: update seen_at pat column

nunogois:
https://linear.app/unleash/issue/2-451/update-last-seen-column-for-pats

### DIFF
--- a/src/lib/services/user-service.ts
+++ b/src/lib/services/user-service.ts
@@ -117,10 +117,7 @@ class UserService {
             process.nextTick(() => this.initAdminUser());
         }
         if (this.flagResolver.isEnabled('tokensLastSeen')) {
-            this.seenTimer = setInterval(
-                () => this.updateLastSeen(),
-                minutesToMilliseconds(1),
-            ).unref();
+            this.updateLastSeen();
         }
     }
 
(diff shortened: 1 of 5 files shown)


gastonfournier (commented):
LGTM! Let's focus on the other PR https://github.com/Unleash/unleash/pull/2514 and if everything is fine with that one, this one will be as well

src/lib/services/user-service.ts
```diff
@@ -426,6 +442,30 @@ class UserService {
...
+    async updateLastSeen(): Promise<void> {
+        if (this.lastSeenSecrets.size > 0) {
+            const toStore = [...this.lastSeenSecrets];
+            this.lastSeenSecrets = new Set<string>();
+            await this.store.markSeenAt(toStore);
+        }
+
+        this.seenTimer = setTimeout(
+            async () => this.updateLastSeen(),
+            minutesToMilliseconds(3),
```
sighphyre, line 455: I don't have a problem with it, just curious, why 3? It feels like such a carefully chosen number

sighphyre (approved):
LGTM

src/lib/services/user-service.ts
nunogois: It's just so it's aligned with the other PR: https://github.com/Unleash/unleash/pull/2514#discussion_r1033938370

sighphyre: Ah!

PR #2518, opened 2022-11-24
Change fetcher in OSS chang

kwasniew:
## About the changes

Fetchers that depend on OSS/enterprise version should not be stuck on initial selection


sjaanus (approved):
LGTM!

PR #2519, opened 2022-11-24
feat: add last seen column to PAT table

nunogois:
https://linear.app/unleash/issue/2-452/add-table-column-to-pats-table-ui


chriswk (approved):
Had to read that toggle check twice before seeing that you push to hidden if it's not enabled. Nifty way of doing it.

PR #2520, opened 2022-11-24
feat: add last seen column to token table

nunogois:
https://linear.app/unleash/issue/2-449/add-table-column-to-api-tokens-table-ui


sighphyre (approved):
LGTM

PR #2521, opened 2022-11-24
fix: feature overview env count should only count enabled

nunogois:
https://linear.app/unleash/issue/2-457/fix-feature-overview-sidepanel-environment-counter-to-only-include


chriswk (approved):
LGTM

PR #2522, opened 2022-11-24
fix: adds cors caching

FredrikOseberg:
* This PR adds a configurable maxAge header to the CORS middleware. This allows the preflight request to be cached so that we can reduce the request load on our end for the frontend clients starting to utilise the frontend api.


gardleopard (approved):
LGTM

PR #2529, opened 2022-11-24
task: continue to return export v3 when variants per env disabled

gastonfournier:
## About the changes
With the latest changes of variants per environment, we switched to export schema v4 without having the feature toggle enabled. This moves the variants to `featureEnvironments` when they were previously in `features`. The main problem is that it can create confusion as the exported file has the same variants for each one of the environments but after importing the file the UI will only show one set of variants attached to the feature.

With this change, we're maintaining the previous schema until the feature toggle is enabled.


src/lib/services/state-service.ts
```diff
@@ -697,7 +704,50 @@ export default class StateService {
...
+        featureTags: IFeatureTag[];
+        featureStrategies: IFeatureStrategy[];
+        environments: IEnvironment[];
+        featureEnvironments: IFeatureEnvironment[];
+    }> {
+        if (this.flagResolver.isEnabled('variantsPerEnvironment')) {
+            return this.exportV4(opts);
+        }
+        // adapt v4 to v3
+        const v4 = await this.exportV4({ ...opts, includeEnvironments: true });
```
chriswk, line 723: Should we add a comment why we need to always have environments set to true?

src/lib/services/state-service.ts
```diff
@@ -697,7 +704,50 @@ export default class StateService {
...
+        if (opts.includeFeatureToggles || true) {
+            const keepEnv = v4.environments
+                .filter((env) => env.type === 'production' && env.enabled)
+                .sort((e1, e2) => e2.sortOrder - e1.sortOrder)[0]; // TODO test order
```
chriswk, line 727: We're assuming we have environments, should we be more defensive here. I realise our discussion before lunch only covered the happy path.

src/test/e2e/api/admin/state.e2e.test.ts
```diff
@@ -464,3 +466,33 @@ test(`should not show environment on feature toggle, when environment is disable
...
+        })
+        .sort();
+    let importedFeatures = variantsv3.features
+        .map((f) => {
+            delete f.createdAt;
+            return f;
+        })
+        .sort();
+    expect(exportedFeatures).toStrictEqual(importedFeatures);
+});
```
chriswk, line 498: we probably should test all variants of the export with the old format, to make sure we handle things being turned on/off?

src/lib/services/state-service.ts
gastonfournier: That's a good point and also reminds me that I should validate the order is correct

src/test/e2e/api/admin/state.e2e.test.ts
chriswk: Then again, we have really only changed the logic for feature and featureEnvironment, so maybe I'm being overly defensive

src/lib/services/state-service.ts
gastonfournier: shoot, my answer was going to be: no, this statement is only executed when includeFeatureToggles is `undefined` or true, but after looking at the code I see an error in my logic :see_no_evil: I'll fix the condition

src/lib/services/state-service.ts
```diff
@@ -697,7 +704,50 @@ export default class StateService {
...
+        featureEnvironments: IFeatureEnvironment[];
+    }> {
+        if (this.flagResolver.isEnabled('variantsPerEnvironment')) {
+            return this.exportV4(opts);
+        }
+        // adapt v4 to v3
+        const v4 = await this.exportV4({ ...opts, includeEnvironments: true });
+        if (opts.includeFeatureToggles || true) {
+            const keepEnv = v4.environments
+                .filter((env) => env.type === 'production' && env.enabled)
```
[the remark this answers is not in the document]
gastonfournier: Screenshot from 2022-11-24 20-10-41

Only, development is enabled (environment type != `production`), but as the error message shows we need at least one environment enabled

thomasheartman:
Just a point of order here, I feel like the main problem is that this **breaks the export API**, and not that it causes confusion with the number of variants shown. Is that wrong? Previously, the variants would be listed directly on the toggle, but it isn't anymore when this is enabled, making this a breaking change. Would it be sensible to default to returning v3 unless a v4 flag is passed (as a query param, for instance)?

gastonfournier:
> Would it be sensible to default to returning v3 unless a v4 flag is passed (as a query param, for instance)?

The problem is that this is an intended breaking change, once you turn this flag on, there will be more data that we won't be able to represent with the previous format. Even if we add that capability of returning v3 through a query parameter we'd need to arbitrarily decide to drop some data.

I don't know if this happened before with other changes, but I'd like to know how we handled this situation in previous circumstances

thomasheartman:
> The problem is that this is an intended breaking change,

Intended or not, it's still a breaking change. And even if it's only enabled if the user explicitly turns the feature on (so it can't be GA until v5), I still think there's a fair chance that someone isn't going to see all the ramifications and inadvertently break their own integration somewhere. Unless we're very explicit about it.

> I don't know if this happened before with other changes, but I'd like to know how we handled this situation in previous circumstances.

Good question! I also don't know. 

> BTW this change goes back to v3, unless variants per environment is enabled because it was not an intentional change to break backward compatibility (similar to your idea of adding a query parameter)

I'm not sure I follow. Could you elaborate a bit? Did we break compatibility between v2 and v3? I'm sure fields were added, sure, but would it break existing integrations?

gastonfournier:
> I still think there's a fair chance that someone isn't going to see all the ramifications and inadvertently break their own integration somewhere. 

The only way I see it'd break is if you're exporting with variants per environment enabled (say unleash v5) and trying to import in a server with variants per environment disabled (unleash v4). Let me know if you think of other way of breaking this but what we are considering is:
- exported from unleash v4 and imported in unleash v5 should work
- exported from unleash v5 and imported in unleash v5 should work

> Unless we're very explicit about it.

We plan to add the necessary documentation https://linear.app/unleash/issue/2-390/documentation and considered making it an unleash v5 feature, but open to other ideas.

> > I don't know if this happened before with other changes, but I'd like to know how we handled this situation in previous circumstances.
> 
> Good question! I also don't know.

Maybe @chriswk would be a good person to ask

> > BTW this change goes back to v3, unless variants per environment is enabled because it was not an intentional change to break backward compatibility (similar to your idea of adding a query parameter)
> 
> I'm not sure I follow. Could you elaborate a bit? Did we break compatibility between v2 and v3? I'm sure fields were added, sure, but would it break existing integrations?

I think this is just a confusion about different versions: 
- I think you're speaking about unleash versions (v4 and future v5)
- I was speaking about the version of our export file format/schema, which is v3 in unleash v4 and we upgraded to v4 (accidentally in unleash v4), but should have kept v3 due to the compatibility issue.

BTW in my comment, I'm being more explicit about what version I'm speaking about, so I hope this helps clarify, but do let me know if that's not the case

thomasheartman:
Great; thanks for the input! I'll respond in turn. But just for the record, if variants per environment doesn't go GA before Unleash 5, then it's all good.

> The only way I see it'd break is if you're exporting with variants per environment enabled (say unleash v5) and trying to import in a server with variants per environment disabled (unleash v4). Let me know if you think of other way of breaking this but what we are considering is:
>
>-    exported from unleash v4 and imported in unleash v5 should work
>-    exported from unleash v5 and imported in unleash v5 should work

got it! Let me try and illustrate my point a bit better:

For clarity: this customer is always on Unleash v4.
- Customer A runs Unleash  v4, and uses the export functionality for some internal integration. They may or may not try to import the exported data back into Unleash. 
- Their current integration is built against v4 of the export API.
- They decide to try enabling variants per environment (either as an experimental option or because it's GA). Because it's included in a minor release, they figure everything should keep working.
- When they activate variants per environment, their homegrown export tooling breaks because the variants aren't where the tool expects them to be anymore.

Does that make it clearer? Again, if this feature doesn't become available until Unleash 5, then this is fine.

> I think this is just a confusion about different versions:
>
>    I think you're speaking about unleash versions (v4 and future v5)
>    I was speaking about the version of our export file format/schema, which is v3 in unleash v4 and we upgraded to v4 (accidentally in unleash v4), but should have kept v3 due to the compatibility issue.

I think we were talking about the same versions, actually. I was definitely referring to the schema version. Did we break compatibility between v3 and v4?

gastonfournier:
> Great; thanks for the input! I'll respond in turn. But just for the record, if variants per environment doesn't go GA before Unleash 5, then it's all good.
> [...]
> For clarity: this customer is always on Unleash v4.
> [...]
> Does that make it clearer? Again, if this feature doesn't become available until Unleash 5, then this is fine.

Yes, clear! I think we are all on the same page here. This is a breaking change and shouldn't be released as a minor change. We still haven't agreed on whether or not this should wait until Unleash 5 or it should be part of Unleash 4 with a big warning message (I think we all lean towards an internal beta on V4 and GA on V5).

> I think we were talking about the same versions, actually. I was definitely referring to the schema version. Did we break compatibility between v3 and v4?

We didn't break syntactic compatibility, but we did break the semantics of the exported file: if you enable variants per environment, the variants that were previously present at the feature will no longer be there. Syntactically the exported file will be valid, but semantically you'd see that the feature has no variants when it does.

chriswk (approved):
This looks good to me

PR #2531, opened 2022-11-25
task: Make operations on the API Token store auditable.

chriswk:
## About the changes
We need a way to have an audit log for operations made on Api Tokens. These changes add three new event types, API_TOKEN_CREATED, API_TOKEN_UPDATED, API_TOKEN_DELETED and extends api-token-service to store these to our event store to reflect the action being taken.

## Discussion points
I'd love to have an discussion on the event format. 
I'm currently exporting the token (minus the secret) as data, and then using it as preData for deletion events and both preData and data in the update events.

Also, not too fond of the Omit use here, since if you pass in an object with the secret that will still be kept. Now I have tests in place for our current three places where we add these three events.

### DIFF
--- a/src/lib/db/api-token-store.ts
+++ b/src/lib/db/api-token-store.ts
@@ -173,7 +173,10 @@ export class ApiTokenStore implements IApiTokenStore {
     }
 
     async get(key: string): Promise<IApiToken> {
-        const row = await this.makeTokenProjectQuery().where('secret', key);
+        const row = await this.makeTokenProjectQuery().where(
+            'tokens.secret',
+            key,
+        );
         return toTokens(row)[0];
     }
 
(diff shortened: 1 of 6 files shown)


src/lib/db/api-token-store.ts
```diff
@@ -173,7 +173,10 @@ export class ApiTokenStore implements IApiTokenStore {
     }
 
     async get(key: string): Promise<IApiToken> {
-        const row = await this.makeTokenProjectQuery().where('secret', key);
+        const row = await this.makeTokenProjectQuery().where(
+            'tokens.secret',
+            key,
+        );
```
chriswk, line 179: tests started failing here due to ambiguous column names, so I took the liberty of adding the table name we use in makeTokenProjectQuery as a prefix.

nunogois (approved):
LGTM - Noticed there are some literal strings we could possibly move to constants, if you think it makes sense.

src/lib/types/events.ts
```diff
@@ -604,3 +609,61 @@ export class PublicSignupTokenUserAddedEvent extends BaseEvent {
...
+export class ApiTokenCreatedEvent extends BaseEvent {
+    readonly data: any;
+
+    readonly environment: string;
+
+    readonly project: string;
+
+    constructor(eventData: {
+        createdBy: string;
+        apiToken: Omit<IApiToken, 'secret'>;
```
ivarconr, line 622: great that you drop the secret 🍡

ivarconr (approved):

PR #2541, opened 2022-11-28
fix: add ADR for domain language

FredrikOseberg:
This PR puts our contributing guidelines in the sidebar of the unleash documentation. Currently there was no way of navigating to them easily, which made our contribution guides and ADRs less useful. This PR adds them to the sidebar as their own category, and adds an ADR for domain centric language.


website/docs/contributing/general/ADR/domain-language.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Decision
+
+We have decided to use the same domain language for the features we develop. Each feature will have it's own domain language to keep it consistent across the codebase.
+
+## Change requests domain language
+
+* *Change request*
+
+An entity referring to the overarching data structure of a change request. A change request contains changes, and can be approved or rejected.
```
dsusa72, line 17: Alternative Suggestion "A  change request contains a collection of changes" (or a group, a set, a list if  you find a problem with _collection_ wording). Basically I suggest emphasizing there's a **group** of changes, in a stronger way than just using the plural.

website/docs/contributing/general/ADR/domain-language.md
```diff
@@ -0,0 +1,34 @@
...
+
+* *Changes*
+
+An entity referring to a single change within a change request
+
+* *Discard*
+
+A naming convention used for deleting a single change of a change request, or discarding an entire change request
```
dsusa72, line 25: I suggest being more precise than "a naming convention". This is fine, but as reader I doubt: _a naming convention for whom ? where ?_  
Maybe using "_a naming convention in the UI_" or "_a naming convention to simplify UX_".

website/docs/contributing/general/ADR/domain-language.md
```diff
@@ -0,0 +1,34 @@
...
+
+* *Pending*
+
+A naming convention used to describe a change requests state. A pending change request is in one of these three states: `Draft | In review | Approved`
```
dsusa72, line 29: Apparently there are 3 well defined underlying states, that we're "masking" under simpler terms for the end user. If that's not the case, please ignore the rest of this comment.
If each sub-state falls under the bigger umbrella of "Pending", I'd appreciate a clarification for the end user point of view. Examples:
- a Draft change is _Pending to be assigned a Reviewer_ 
- an In Review change is _Pending to be Approved_
- an Approved change is _Pending to be applied_

website/docs/contributing/ADRs.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Architectural Decision Records
+---
+
+## Introduction {#introduction}
+
+Architectural decision records are a record of design decisions we have made in the past because we belived they would help our code quality over time. Any ADR can be challenged, but two conditions must be met to change an ADR: (1) The proposed solution must provide a tangible benefit in terms of code quality, and (2) the proposed solutions benefit must be worth retroactively refitted to the entire codebase. One such example is the decision to re-write unleash to typescript.
```
thomasheartman, line 7: I propose splitting this into a list for legibility

```suggestion
Architectural decision records are a record of design decisions we have made in the past because we belived they would help our code quality over time. Any ADR can be challenged, but two conditions must be met to change an ADR: 
1. The proposed solution must provide a tangible benefit in terms of code quality.
2. The benefits of the proposed solution must outweigh the effort of retroactively changing the entire codebase. 
One such example is the decision to re-write Unleash to TypeScript.
```

kwasniew (approved):

website/docs/contributing/ADRs.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Crosscutting ADRs
```
thomasheartman, line 9: I wanna say this should be with a hyphen instead (i.e. "cross-cutting"), but maybe we can make it clearer by changing the word and/or by adding a description:

```suggestion
## Overarching ADRs

These ADRs describe decisions that concern the entire codebase. They apply to back-end code, front-end code, and code that doesn't neatly fit into either of those categories.
```

website/docs/contributing/ADRs.md
```diff
@@ -0,0 +1,34 @@
...
+
+* [Domain language](./general/ADR/domain-language.md)
+
+## Backend ADRs
```
thomasheartman, line 13: Now that this goes into the "official" docs, I think we should align on how we spell "back end" and "front end". I've been going off Chris Coyier's short [post on whether it's "front-end" or front end"](https://css-tricks.com/poll-results-front-end-front-end/) for a bit, which says that:
- as a noun, it's two words. For instance: "I work on the front end"
- as a compound adjective, it's with a hyphen. E.g: "I'm a front-end engineer"

But I would actually be very interested to hear @JarrodSC's thoughts on this.

```suggestion
## Back-end ADRs
```

website/docs/contributing/ADRs.md
```diff
@@ -0,0 +1,34 @@
...
+
+We are currently in the process of defining ADRs for the backend, currently we have created the following ADRS:
```
thomasheartman, line 15:
```suggestion
We are in the process of defining ADRs for the back end. At the time of writing we have created the following ADRS:
```

website/docs/contributing/ADRs.md
```diff
@@ -0,0 +1,34 @@
...
+
+* [Naming](./backend/ADR/naming.md)
+* [Preferred export](./backend/ADR/preferred-export.md)
+
+## Frontend ADRs
+
+We have created a set of ADRs to help guide the development of the frontend:
+
```
thomasheartman, line 23:
```suggestion
## Front-end ADRs

We have created a set of ADRs to help guide the development of the front end:

```

website/docs/contributing/ADRs.md
```diff
@@ -0,0 +1,34 @@
...
+* [Interface naming](./frontend/ADR/interface-naming.md)
+* [Preferred component props usage](./frontend/ADR/preferred-component-props-usage.md)
+* [Preferred export](./frontend/ADR/preferred-export.md)
+* [Preferred function type](./frontend/ADR/preferred-function-type.md)
+* [Preferred style import placement](./frontend/ADR/preferred-styles-import-placement.md)
+* [Preferred styling method](./frontend/ADR/preferred-styling-method.md)
+* [Preferred data mutation method](./frontend/ADR/preferred-data-mutation-method.md)
+* [Preferred data fetching method](./frontend/ADR/preferred-data-fetching-method.md)
+* [Preferred folder structure](./frontend/ADR/preferred-folder-structure.md)
+* [Preferred form architecture](./frontend/ADR/preferred-form-architecture.md)
```
thomasheartman, line 34: These don't seem to be included in the sidebar: is there any reason they're left out?

website/docs/contributing/general/ADR/domain-language.md
thomasheartman, line 17: We can use lists for listing terms, but please make sure they're treated as a single list in that case, and not as multiple lists (as is the case right now). Optionally, use a [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) (sadly, markdown doesn't have anything that maps directly to it, so you'll need to mark it up with raw html in that case).

If you just want to turn it into a single list, consider placing the term and the definition on the same line, separated by a colon.

```suggestion
* *Change request*: An entity referring to the overarching data structure of a change request. A change request contains changes, and can be approved or rejected.
```

website/docs/contributing/general/ADR/domain-language.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Change requests domain language
+
+* *Change request*
+
+An entity referring to the overarching data structure of a change request. A change request contains changes, and can be approved or rejected.
+
+* *Changes*
+
+An entity referring to a single change within a change request
```
thomasheartman, line 21: Is this correct? Isn't "changes" a "list of changes" and not a single change? This is unclear to me. Also, what is "an entity" in this context?

website/docs/contributing/general/ADR/domain-language.md
thomasheartman, line 29: I sort of agree with @dsusa72's comments on considering UX and legibility as well as the term "naming convention" Would something like this work?

```suggestion
A *pending* change request is one that has not yet been applied or discarded. In other words, it is in one of these three states:
1. `Draft`
2. `In review`
3. `Approved`
```

website/docs/contributing/general/ADR/domain-language.md
```diff
@@ -0,0 +1,34 @@
...
+
+A naming convention used for deleting a single change of a change request, or discarding an entire change request
+
+* *Pending*
+
+A naming convention used to describe a change requests state. A pending change request is in one of these three states: `Draft | In review | Approved`
+
+* *Closed*
+
+A naming convention used to describe a change requests state. A closed change request is in one of two states: `Applied | Cancelled`
```
thomasheartman, line 33: Maybe this makes it a little clearer? What do you think?

```suggestion
A *closed* change request has either been applied or cancelled and can no longer be changed. Change requests that are either `Applied` or `Cancelled` are considered closed.
```

website/sidebars.js
```diff
@@ -341,5 +341,22 @@ module.exports = {
                 },
             ],
         },
+        {
+            label: 'Contributing',
+            type: 'category',
+            collapsed: false,
```
thomasheartman, line 347: I think this should be collapsed by default. No need to take up additional space in the side bar for something that's only applicable to a small number of users.

```suggestion
            collapsed: true,
```

website/sidebars.js
```diff
@@ -341,5 +341,22 @@ module.exports = {
                 },
             ],
         },
+        {
+            label: 'Contributing',
```
thomasheartman, line 345: Can we make this clearer?  Something like this?

```suggestion
            label: 'Developer contribution guides',
```

website/sidebars.js
```diff
@@ -341,5 +341,22 @@ module.exports = {
...
+                type: 'generated-index',
+                title: 'Developer Guide',
+                description: 'Learn how to contribute to unleash.',
+                slug: '/developer-guide',
+            },
+            items: [
+                'contributing/developer-guide',
+                'contributing/frontend/overview',
+                'contributing/backend/overview',
+                'contributing/ADRs',
```
thomasheartman, line 358: We should list all the contribution docs from the sidebar (as remarked on in a previous comment). However, instead of listing each file individually (as we have done in other parts of the sidebar), I think we should try and autogenerate this section if possible. Refer to the [autogenerated sidebar docs](https://docusaurus.io/docs/sidebar/autogenerated) for more info. Give it a go and see if it produces satisfying results?

thomasheartman (commented):
Nice work! And yeah, I agree we should list this (or optionally host it on its own domain). I've added some comments and questions inline that I'd love to get some input on before approving this 🙌🏼

website/docs/contributing/ADRs.md
FredrikOseberg: Did not want to clutter the sidebar. Seemed like an acceptable tradeoff.

website/docs/contributing/general/ADR/domain-language.md
FredrikOseberg: We don't use this term in the user facing code, but rather as our own term for naming the grouping of these states. For the user, they will only relate to the state themselves as modeled: `Draft | In review | Approved | Applied | Cancelled`

website/docs/contributing/ADRs/ADRs.md
```diff
@@ -13,27 +13,27 @@ One such example is the decision to re-write Unleash to TypeScript.
...
-* [Domain language](./general/ADR/domain-language.md)
+* [Domain language](./Overarching/domain-language.mddomain-language.md)
 
 ## Back-end ADRs
 
 We are in the process of defining ADRs for the back end. At the time of writing we have created the following ADRS:
 
-* [Naming](./back-end/ADR/naming.md)
-* [Preferred export](./back-end/ADR/preferred-export.md)
+* [Naming](./Back-end/naming.md)
```
thomasheartman, line 22: Out of curiosity, why do the directory names contain uppercase letters? Does it have to do with how they're displayed in the sidebar? In that case, I think I'd prefer to fix the casing with CSS instead of in the folder tree. Either that, or move the autogeneration one step lower, so that you explicitly declare the "Back end" and "Front end" categories, but autogenerate the content 🤔 

We generally use all-lowercase names everywhere else (I think), so I'd like to keep doing that as much as possible 💁🏼

FredrikOseberg: That's the only way I could find to fit it with the auto generation for now. Do you know the quick fix for this? I'm at the limit of my timebox for this.

thomasheartman: I've never used the auto-generation method, so I don't have a quick fix for it, no. What's the issue you're trying to solve? If it's sidebar casing, then we can just ignore that for now, I'd say? 🤷🏼

PR #2542, opened 2022-11-28
perf: remove indexes covered by pks

gastonfournier:
## About the changes
While experimenting with https://github.com/ankane/pghero it reported some indexes that can be removed:

Screenshot from 2022-11-25 17-22-49

This can improve write performance and disk usage but in both cases, the impact might be rather low (except maybe for metrics).

## Discussion points
This is not a must do, but because the potential impact on metrics performance I opened this PR as a conversation opener

### DIFF
--- a/src/migrations/20221125164718-remove-duplicated-indexes.js
+++ b/src/migrations/20221125164718-remove-duplicated-indexes.js
@@ -1,0 +1,21 @@
+'use strict';
+
+export function up(db, callback) {
+    db.runSql(
+        `DROP INDEX idx_client_metrics_f_name;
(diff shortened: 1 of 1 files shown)


chriswk:
Hmm, that's interesting, I believe those indices were added due to recommendations from a different db analyzer (https://github.com/channable/dbcritic)  due to being used as part of a foreign key (FK). When doing selects where only one of the columns that are part of a composite index is used, I believe it would fall back to FTS instead of index use.

If we can prove that's not the case, I'm more than happy to remove extra indexes to save some time on writes.

PR #2547, opened 2022-11-28
fix: addon parameter validation on empty strings

nunogois:
https://linear.app/unleash/issue/2-408/addon-parameter-validation


gastonfournier (approved):
Provided `event.target` can't be undefined, this looks good :+1:

PR #2544, opened 2022-11-28
Feat/add pro feature icon to oss projects

andreas-unleash:
Add pro feature icon to oss projects.

## About the changes

Closes #

PR #2549, opened 2022-11-28
Fix plan upgrade tooltip

nunogois:
Related: https://github.com/Unleash/unleash/pull/2544
From: https://linear.app/unleash/issue/1-414/add-a-badge-for-oss-users-on-the-projects-page-to-see-that-extra

Includes some fixes and suggestions:
 - Should only show when OSS (icon was showing every time);
 - Link now works correctly, opens on new tab;
 - Improves styling to be more aligned with original design: https://www.figma.com/file/qdwpPfuitJUNinm6mvmCmG/Unleash-application?node-id=6961%3A38443&t=zsXNAnOurapoLm9j-0;
 - Fixes text to be like in the design;
 - Improves `HtmlTooltip` look and behaviour: white arrow, custom width, etc;
 - Some refactoring and cleanup;
 
Let me know if I should change anything 👍

### DIFF
--- a/frontend/src/component/common/TooltipResolver/TooltipResolver.tsx
+++ b/frontend/src/component/common/TooltipResolver/TooltipResolver.tsx
@@ -5,7 +5,7 @@ import { HtmlTooltip } from '../HtmlTooltip/HtmlTooltip';
 export interface ITooltipResolverProps extends Omit<TooltipProps, 'title'> {
     title?: string;
     titleComponent?: ReactNode;
-    variant?: 'default' | 'white';
+    variant?: 'default' | 'custom';
 }
 
 export const TooltipResolver = ({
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/TooltipResolver/TooltipResolver.tsx
```diff
@@ -1,11 +1,11 @@
...
 import { Tooltip, TooltipProps } from '@mui/material';
 import { HtmlTooltip } from '../HtmlTooltip/HtmlTooltip';
 
 export interface ITooltipResolverProps extends Omit<TooltipProps, 'title'> {
-    title?: string | undefined;
-    titleComponent?: ReactNode | undefined;
-    variant?: 'default' | 'white';
+    title?: string;
+    titleComponent?: ReactNode;
+    variant?: 'default' | 'custom';
```
nunogois, line 8: I think this should be `custom` (or `html` or something else). `white` doesn't fully represent all the `HtmlTooltip` component differences and e.g. dark mode would not exactly make this a "white" tooltip.

andreas-unleash: Makes sense

andreas-unleash (approved):
LGTM! Thanks for improving :)

PR #2550, opened 2022-11-28
Favorite features

sjaanus:
This PR adds following functionality.

1. Feature flag for favorites
2. Ability to add features as favorites
3. Ability to delete features from favorites
4. `/api/admin/projects/default/features/${featureName}` will return  favorites
6. `/api/admin/features` will return favorites
7. `/api/admin/projects/default/features` will return favorites


src/test/e2e/custom-auth.test.ts
```diff
@@ -1,9 +1,21 @@
 import dbInit from './helpers/database-init';
 import { setupAppWithCustomAuth } from './helpers/test-helper';
+import { RoleName } from '../../lib/types';
 
 let db;
 let stores;
 
+const preHook = (app, config, { userService, accessService }) => {
```
kwasniew, line 8: I'd use a name for a preHook that explains the role it plays in those tests

src/test/e2e/api/admin/favorites.e2e.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+beforeEach(async () => {
+    await loginRegularUser();
+});
+
+test('should have favorites true in project endpoint', async () => {
+    const featureName = 'test-feature';
+    await createFeature(featureName);
+
+    await app.request
```
kwasniew, line 74: I'd extract those into helper methods explaining what those steps are. Otherwise the reader of the code is distracted by the HTTP details

src/test/e2e/api/admin/favorites.e2e.test.ts
```diff
@@ -0,0 +1,171 @@
...
+const createFeature = async (featureName: string) => {
+    await app.request
+        .post('/api/admin/projects/default/features')
+        .send({
+            name: featureName,
+        })
+        .set('Content-Type', 'application/json')
+        .expect(201);
+
+    // await projectService.addEnvironmentToProject('default', environment);
```
kwasniew, line 24: why do we have this comment here?

src/test/e2e/api/admin/favorites.e2e.test.ts
```diff
@@ -0,0 +1,171 @@
...
+afterEach(async () => {
+    await db.stores.favoriteFeaturesStore.deleteAll();
+    await db.stores.featureToggleStore.deleteAll();
+});
+
+beforeEach(async () => {
+    await loginRegularUser();
+});
+
+test('should have favorites true in project endpoint', async () => {
```
kwasniew, line 70: what about: should be able to favorite project? This is the business intent. favorites true and project endpoint are implementation details.

src/test/e2e/api/admin/favorites.e2e.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        .expect(200);
+
+    expect(body.features).toHaveLength(1);
+    expect(body.features[0]).toMatchObject({
+        name: featureName,
+        favorite: true,
+    });
+});
+
+test('should have favorites false by default', async () => {
```
kwasniew, line 91: similar to the previous one. Feature should not be favorited by default.

src/test/e2e/api/admin/favorites.e2e.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        .set('Content-Type', 'application/json')
+        .expect(200);
+
+    expect(body).toMatchObject({
+        name: featureName,
+        favorite: true,
+    });
+});
+
+test('should have favorites false after deleting favorite', async () => {
```
kwasniew, line 148: Should be able to unfavorite the feature

src/migrations/20221124123914-add-favorites.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+            CREATE TABLE IF NOT EXISTS favorite_features
+            (
+                feature varchar(255) not null references features (name) on DELETE CASCADE,
```
kwasniew, line 8: I'd use upper case letters for SQL stuff and lower case for our stuff. It's easier for the reader to parse it like this.
e.g. feature VARCHAR(255) NOT NULL REFERENCES features (name) ON DELETE CASCADE

src/lib/services/favorites-service.ts
```diff
@@ -0,0 +1,37 @@
+import { IUnleashConfig } from '../types/option';
+import { IUnleashStores } from '../types/stores';
+import { Logger } from '../logger';
+import {
+    IFavoriteFeatureKey,
+    IFavoriteFeaturesStore,
+} from '../types/stores/favorite-features';
+import { IFavoriteFeature } from '../types/favorites';
+
+export class FavoritesService {
```
kwasniew, line 10: Do we need this service? It's a CRUD operation so controller can call the store directly. Is there anything this service adds on top of the store?

src/lib/db/feature-strategy-store.ts
```diff
@@ -214,27 +222,53 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
         const stopTimer = this.timer('getFeatureAdmin');
-        const rows = await this.db('features_view')
+        let query = this.db('features_view')
             .where('name', featureName)
             .modify(FeatureToggleStore.filterByArchived, archived);
+
+        let selectColumns = ['features_view.*'];
+        if (userId && this.flagResolver.isEnabled('favorites')) {
+            query = query.leftJoin(`favorite_features as ff`, function () {
+                this.on('ff.feature', 'features_view.name').andOnVal(
```
kwasniew, line 263: I'd not mix abbreviations with full names. If we have features_view I'd use favorite_features instead of ff

kwasniew (approved):
Non blocking approve but please check the comments

kwasniew:
@sjaanus @Tymek please followup with PR cleanup in another PR. Merging this one for now to unblock work.

src/migrations/20221124123914-add-favorites.js
sjaanus: Now that this is merged, it is not best to edit the migration. Will keep it in mind for future

src/lib/services/favorites-service.ts
sjaanus: I guess just best practice, aligns the rest of the codebase and disencourages to call stores directly from controller.

PR #2552, opened 2022-11-28
fix: state-service should always keep api keys

ivarconr:
We have experienced side-effects where the import was unexpected and resulted in environments thought to be removed. This had the unexpected side-effect of also deleting API keys for some environments not part of the import file. 

This commit removes the ability of the state-service to mutate api keys directly. There is no compelling reasons why we should remove API keys as part of an import query.


sighphyre (approved):
LGTM!

gastonfournier:
Just resolved merge conflicts

website/docs/reference/deploy/import-export.md
```diff
@@ -64,7 +64,9 @@ For example if you want to download just feature-toggles as yaml:
 
 ### API Import {#api-import}
 
-:::caution Importing environments
+:::caution Importing environments in Unleash 4.20 and below
+
+This is only relevant if you use **Unleash 4.20 or earlier**:
```
thomasheartman, line 69: @ivarconr We should make sure that the versions here are correct. If we just released 4.20, I can imagine this would be part of 4.21.

thomasheartman (commented):
Makes sense to me. I think we did that so that no one would get unexpected access to the environments you imported, right? 

Anyway, I've updated the little bit of the doc that mentions this, but we need to make sure the version mentioned is correct too.

PR #2556, opened 2022-11-29
min approvals from cr not config

kwasniew:
## About the changes

Min approvals should be fetched from change request, not from config


FredrikOseberg (approved):
LGTM

PR #2563, opened 2022-11-30
Some tweaks to favorites

sjaanus:
This is follow up PR to https://github.com/Unleash/unleash/pull/2550, which addresses the issues that came up from reviews.


kwasniew (approved):
LGTM!

PR #2569, opened 2022-11-30
Favorite project

sjaanus:
Adds ability to favorite projects.

1. Can favorite project
2. Can unfavorite project
3. Favorite field is returned on `/api/admin/projects/default`
4. Favorite field is returned on` /api/admin/projects`

### DIFF
--- a/src/lib/db/project-store.ts
+++ b/src/lib/db/project-store.ts
@@ -88,7 +88,6 @@ class ProjectStore implements IProjectStore {
         const projectTimer = this.timer('getProjectsWithCount');
         let projects = this.db(TABLE)
             .leftJoin('features', 'features.project', 'projects.id')
-            .groupBy('projects.id')
             .orderBy('projects.name', 'asc');
         if (query) {
             projects = projects.where(query);
@@ -97,7 +96,9 @@ class ProjectStore implements IProjectStore {
             this.db.raw(
                 'projects.id, projects.name, projects.description, projects.health, projects.updated_at, count(features.name) AS number_of_features',
             ),
-        ] as (String | Raw<any>)[];
+        ] as (string | Raw<any>)[];
+
+        let groupByColumns = ['projects.id'];
 
         if (userId && this.flagResolver.isEnabled('favorites')) {
             projects = projects.leftJoin(`favorite_projects`, function () {
@@ -109,11 +110,16 @@ class ProjectStore implements IProjectStore {
             });
             selectColumns = [
                 ...selectColumns,
-                this.db.raw('min(favorite_projects.project) as project'),
+                this.db.raw(
+                    'favorite_projects.project is not null as favorite',
+                ),
             ];
+            groupByColumns = [...groupByColumns, 'favorite_projects.project'];
         }
 
-        const projectAndFeatureCount = await projects.select(selectColumns);
+        const projectAndFeatureCount = await projects
+            .select(selectColumns)
+            .groupBy(groupByColumns);
 
         const projectsWithFeatureCount = projectAndFeatureCount.map(
             this.mapProjectWithCountRow,
@@ -138,7 +144,7 @@ class ProjectStore implements IProjectStore {
             id: row.id,
             description: row.description,
             health: row.health,
-            favorite: row.project !== null,
+            favorite: row.favorite,
             featureCount: Number(row.number_of_features) || 0,
             memberCount: Number(row.number_of_users) || 0,
             updatedAt: row.updated_at,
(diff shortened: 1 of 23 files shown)


src/test/e2e/api/admin/favorites.e2e.test.ts
```diff
@@ -162,3 +175,47 @@ test('should be able to unfavorite feature', async () => {
         favorite: false,
     });
 });
+
+test('should be favorited in projects list', async () => {
+    await favoriteProject();
+
+    const { body } = await app.request
```
kwasniew, line 182: getProject() to keep same level of abstraction as favoriteProject.

src/test/e2e/api/admin/favorites.e2e.test.ts
```diff
@@ -162,3 +175,47 @@ test('should be able to unfavorite feature', async () => {
...
+    expect(body.projects[0]).toMatchObject({
+        name: 'Default',
+        favorite: true,
+    });
+});
+
+test('should be favorited in single project endpoint', async () => {
+    await favoriteProject();
+
+    const { body } = await app.request
```
kwasniew, line 197: same as above, I'd keep one abstraction level in a test. When we go for helper domain method let's stick to it

src/lib/types/model.ts
```diff
@@ -176,6 +175,7 @@ export interface IProjectOverview {
     members: number;
     version: number;
     health: number;
+    favorite?: boolean;
```
kwasniew, line 178: does it have to be optional? can't the store guarantee the false value for old projects?

src/lib/db/project-store.ts
```diff
@@ -73,21 +83,37 @@ class ProjectStore implements IProjectStore {
...
             .orderBy('projects.name', 'asc');
         if (query) {
             projects = projects.where(query);
         }
-        const projectAndFeatureCount = await projects;
+        let selectColumns = [
+            this.db.raw(
+                'projects.id, projects.name, projects.description, projects.health, projects.updated_at, count(features.name) AS number_of_features',
+            ),
+        ] as (String | Raw<any>)[];
```
kwasniew, line 100: what is this upper case String type here?

src/lib/db/project-store.ts
```diff
@@ -73,21 +83,37 @@ class ProjectStore implements IProjectStore {
...
+            projects = projects.leftJoin(`favorite_projects`, function () {
+                this.on('favorite_projects.project', 'projects.id').andOnVal(
+                    'favorite_projects.user_id',
+                    '=',
+                    userId,
+                );
+            });
+            selectColumns = [
+                ...selectColumns,
+                this.db.raw('min(favorite_projects.project) as project'),
```
kwasniew, line 112: can you remind me why we have this min function here?

src/lib/db/project-store.ts
```diff
@@ -112,6 +138,7 @@ class ProjectStore implements IProjectStore {
             id: row.id,
             description: row.description,
             health: row.health,
+            favorite: row.project !== null,
```
kwasniew, line 141: this condition looks weird, why not row.favorite?

src/lib/services/project-health-service.ts
```diff
@@ -52,11 +56,13 @@ export default class ProjectHealthService {
             hoursToMilliseconds(1),
         ).unref();
         this.featureToggleService = featureToggleService;
+        this.favoritesService = favoritesService;
     }
 
     // TODO: duplicate from project-service.
     async getProjectOverview(
         projectId: string,
+        userId?: number,
```
kwasniew, line 65: I'd add optional field at the end of a method signature so that the default param for archived can work in more situations

src/lib/services/project-health-service.ts
```diff
@@ -85,7 +97,11 @@ export default class ProjectHealthService {
     async getProjectHealthReport(
         projectId: string,
     ): Promise<IProjectHealthReport> {
-        const overview = await this.getProjectOverview(projectId, false);
+        const overview = await this.getProjectOverview(
+            projectId,
+            undefined,
```
kwasniew, line 102: passing undefined in the middle of a call feels weird to me

kwasniew (approved):
LG. Minor comments inside

src/lib/types/model.ts
sjaanus: This interface is used in multiple endpoints, not all of them need favorites.

src/lib/db/project-store.ts
sjaanus: Yes, lets do that!

PR #2571, opened 2022-11-30
refactor: flag resolver should use stricter types

nunogois:
Adding stricter types to `FlagResolver` can possibly help improve our DX - Help us prevent errors like typos, guide us to correctly add a flag when needed, and warn us of stray checks whenever we do a clean up at a later stage.

### DIFF
--- a/src/lib/types/experimental.ts
+++ b/src/lib/types/experimental.ts
@@ -1,73 +1,61 @@
 import { parseEnvVarBoolean } from '../util';
 
-export type IFlags = IExperimentalOptions['flags'];
+export type IFlags = Partial<typeof flags>;
 export type IFlagKey = keyof IFlags;
 
+const flags = {
+    E: false,
+    ENABLE_DARK_MODE_SUPPORT: false,
+    anonymiseEventLog: false,
+    embedProxy: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_EMBED_PROXY,
+        true,
+    ),
+    changeRequests: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_CHANGE_REQUESTS,
+        false,
+    ),
+    embedProxyFrontend: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_EMBED_PROXY_FRONTEND,
+        true,
+    ),
+    batchMetrics: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_BATCH_METRICS,
+        false,
+    ),
+    responseTimeWithAppName: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_RESPONSE_TIME_WITH_APP_NAME,
+        false,
+    ),
+    proxyReturnAllToggles: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_PROXY_RETURN_ALL_TOGGLES,
+        false,
+    ),
+    variantsPerEnvironment: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_VARIANTS_PER_ENVIRONMENT,
+        false,
+    ),
+    favorites: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_FAVORITES,
+        false,
+    ),
+    networkView: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_NETWORK_VIEW,
+        false,
+    ),
+    maintenance: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_MAINTENANCE,
+        false,
+    ),
+};
+
 export const defaultExperimentalOptions: IExperimentalOptions = {
-    flags: {
-        ENABLE_DARK_MODE_SUPPORT: false,
-        anonymiseEventLog: false,
-        embedProxy: parseEnvVarBoolean(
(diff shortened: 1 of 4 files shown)


src/lib/routes/admin-api/index.ts
```diff
@@ -112,6 +112,7 @@ class AdminApi extends Controller {
         this.app.use(
             '/invite-link',
             conditionalMiddleware(
+                // @ts-expect-error
                 () => config.flagResolver.isEnabled('publicSignup'),
```
nunogois, line 116: One of the strays caught by the PR - Maybe it's a feature flag we should clean up? Or, on the other hand, correctly add to the remaining feature flags?

FredrikOseberg: This should be removed.

src/lib/types/experimental.ts
```diff
@@ -64,15 +65,17 @@ export const defaultExperimentalOptions = {
 
 export interface IExperimentalOptions {
     flags: {
-        [key: string]: boolean;
+        E?: boolean;
```
kwasniew, line 68: It it supposed to be some generic or a typo?

src/lib/util/flag-resolver.test.ts
```diff
@@ -14,9 +14,11 @@ test('should produce UI flags with extra dynamic flags', () => {
         ...defaultExperimentalOptions,
         flags: { extraFlag: false },
     };
+    // @ts-expect-error
```
kwasniew, line 17: can we get rid of those?

kwasniew (commented):
questions inline

src/lib/util/flag-resolver.test.ts
kwasniew: e.g.  const resolver = new FlagResolver(config as IExperimentalOptions);

kwasniew: casting seems to be more idiomatic in TS tests than expecting errors

src/lib/types/experimental.ts
nunogois: Discussed in: https://unleash-internal.slack.com/archives/C046LV6HH6W/p1669880769980979 - Needed to include this option since it's being passed in:
https://github.com/Unleash/unleash/blob/bc3744d56566b6a173f5473344557f0ad7f72e0b/src/lib/create-config.ts#L101-L112

src/lib/util/flag-resolver.test.ts
nunogois: Good point, addressed in https://github.com/Unleash/unleash/pull/2571/commits/e226e30584d98ff504660ed1c20d9ab206f855cf.

FredrikOseberg:
Before we merge this, will this break enterprise and cloud builds since they don't explicitly type? We should at least be aware that this can happen and update these in quick succession.

FredrikOseberg:
Thinking a bit more about this, the IFlagKey type might be too limiting. If we only allow the flags defined in unleash, then the external flag provider has to implement an interface that only allows these flags. This can be problematic when we have flags that we want to define outside of unleash. For example, what if we wanted to have a feature flag specifically for unleash-cloud or unleash-enterprise? One specific example that comes to mind is change requests, which is a flag that is not in experimental options in unleash, because it's only used in enterprise. I think we need to consider a less restrictive type, but one that still allows the autocompletion. There should be a possibility to a looser type using the index access type of an object.

nunogois:
Hey @FredrikOseberg,

> Before we merge this, will this break enterprise and cloud builds since they don't explicitly type? We should at least be aware that this can happen and update these in quick succession.

A quick test locally seemed to work as intended, but it would be great if you could have a look as well just to double check.

> Thinking a bit more about this, the IFlagKey type might be too limiting. If we only allow the flags defined in unleash, then the external flag provider has to implement an interface that only allows these flags. This can be problematic when we have flags that we want to define outside of unleash. For example, what if we wanted to have a feature flag specifically for unleash-cloud or unleash-enterprise? One specific example that comes to mind is change requests, which is a flag that is not in experimental options in unleash, because it's only used in enterprise. I think we need to consider a less restrictive type, but one that still allows the autocompletion. There should be a possibility to a looser type using the index access type of an object.

Seems like `changeRequests` is already part of experimental options in Unleash, even if it's an enterprise only feature: https://github.com/Unleash/unleash/blob/4a3d26065fe475fbcfff4f09e8096bb9eb9023c0/src/lib/types/experimental.ts#L74

This seems like the right approach to me.

A looser type would also mean we would lose some things, like easily catching a stray like this: https://github.com/Unleash/unleash/pull/2571#discussion_r1035841427

You would still have these options:
 - Adding the feature flag to Unleash so it's known, which is the ideal route;
 - Casting your flag string as `IFlagKey` for specific scenarios like tests;
 - `// @ts-expect-error`ing the line, which should be avoided;
 
This flow seems better to me and more explicit than what we had before, WDYT?

FredrikOseberg (approved):
LGTM for now. If we discover any limitations we could always go back.

chriswk (approved):
I like that this should allow us to see which features are enabled, and what our legal keys to use are, rather than like what we have now, where you could theoretically just ask for any key. 

The disadvantage is of course that we have to remember to code it in here, rather than having to just add a feature toggle in the Unleash Hosted UI. I find that I like the trade-off. 

But would really like someone else to chime in as well.

Also, this looks quite like Cognite's way of doing the Rust client, where you need your feature key in an enum to be able to query for its enabled status. When writing tools like Edge and the proxy, it's a hindrance, but when using in our own controlled App, having a type that catches typos and non-existing toggles is :+1:  for me

PR #2573, opened 2022-11-30
feat: default the frontend api flags to on

FredrikOseberg:
* This PR turns on the flag for the frontendApi to be on by default. We should still be able to override them from the unleash-cloud distribution.


sjaanus (approved):
LGTM!

PR #2586, opened 2022-12-01
task: Expose prometheus metrics

gastonfournier:
## About the changes
This connects our backend with Prometheus (or compatible) metrics service, acting as a proxy. 
The changes are behind a feature toggle and should not be considered feature complete.

## Discussion points
1. Should this be part of src/lib/services/client-metrics/instance-service.ts under `/api/admin/metrics/rps`?
2. Is rps the best resource name for this?


chriswk (approved):
Cool. Time to expose how much our customers are using us :)

PR #2590, opened 2022-12-02
docs: new variants per environment

gastonfournier:
## About the changes
Documentation about feature toggle variants per environment

## Discussion points
The version when this will be available is still to be defined.


website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -378,16 +378,7 @@ Vary: Accept-Encoding
...
-    "type": "release",
-    "variants": [
-        {
-            "name": "blue",
-            "overrides": [],
-            "stickiness": "default",
-            "weight": 1000,
-            "weightType": "variable"
-        }
-    ]
```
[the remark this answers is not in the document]
thomasheartman: Were they previously included or was the doc wrong here?

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -609,9 +600,17 @@ Transfer-Encoding: chunked
 
 ## Feature Variants
 
+**Important notes:** 
+1. Variants were previously linked to feature toggles making it impossible to have different variants per environment. We are migrating to variants per environment which will allow you to manage variants independently in each environment. 
+2. As part of the migration, existing variants will be copied to all the environments of its feature which will respect the behaviour prior to having variants per environment.
+3. Clients and SDKs are not affected by this changes as they always access features within one environment.
+4. The export format will change to have variants linked to environments. Old exported formats will be gracefully handled if used to import data.
+5. Because this change adds more data, it won't be possible to down migrate without data loss (i.e. if variants are different per environment, there's no way of representing that in the model where variants were linked to features and applying to all environments)
```
thomasheartman, line 608: This is great information, but feels like more of a blog post than reference documentation, I think. We could rewrite it to a section about changes from 4.x to 4.x+1 or 5 or whatever. What do you think about that?

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -609,9 +600,17 @@ Transfer-Encoding: chunked
...
+
 ### Put variants for Feature Toggle {#update-variants}
+**Warning** this affects all environments at once. It's recommended to use [#update-variants-per-environment](#update-variants-per-environment)
```
thomasheartman, line 611: We should use [docusaurus' admonitions](https://docusaurus.io/docs/markdown-features/admonitions) when making call-outs in the text.

```suggestion
:::caution 

This endpoint affects all environments at once. If you only want to update a single environment, use [#update-variants-per-environment](#update-variants-per-environment) instead.

:::
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -696,6 +695,7 @@ Content-Type: application/json; charset=utf-8
 ```
 
 ### PATCH variants for a feature toggle
+**Warning** this affects all environments at once. It's recommended to use [#update-variants-per-environment](#update-variants-per-environment)
```
thomasheartman, line 698:
```suggestion
:::caution

 This endpoint affects all environments at once. If you only want to update a single environment, use [#update-variants-per-environment](#update-variants-per-environment) instead.
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -740,6 +740,137 @@ Authorization:$KEY
...
+		"weightType": "variable",
+		"weight": 123
+	}
+]}/>
+
+This overwrites the current variants for the feature toggle specified in the :featureName parameter for the :environment parameter.
+The backend will validate the input for the following invariants
+
+* If there are variants, there needs to be at least one variant with `weightType: variable`
+* The sum of the weights of variants with `weightType: fix` must be below 1000 (< 1000)
```
thomasheartman, line 771: Is this correct? It's < and not <= 1000, right?

```suggestion
* The sum of the weights of variants with `weightType: fix` must be strictly less than 1000 (< 1000)
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -740,6 +740,137 @@ Authorization:$KEY
...
+
+The backend will also distribute remaining weight up to 1000 after adding the variants with `weightType: fix` together amongst the variants of `weightType: variable`
```
thomasheartman, line 773:
```suggestion
The back end will also distribute remaining weight up to 1000 after adding the variants with `weightType: fix` together amongst the variants of `weightType: variable`
```

website/docs/reference/feature-toggle-variants.md
```diff
@@ -3,6 +3,7 @@ title: Feature Toggle Variants
 ---
 <div class="alert alert--info" role="alert">
   <em>Feature toggle variants</em> were introduced in <em>Unleash v3.2.0</em>.
+  <em>Feature toggle variants</em> are attached to environments since <em>Unleash v4.19.0</em>.
 </div>
 <br/>
```
thomasheartman, line 8: This should be rewritten as an admonition. Something like:

```suggestion
:::info Availability

**Feature toggle variants** were introduced in Unleash 3.2.0.

**Per-environment variants** (link to a description?) were introduced in Unleash 4.x.

:::
```

website/docs/reference/feature-toggle-variants.md
```diff
@@ -12,7 +13,7 @@ Variants facilitate A/B testing and experimentation by letting you create contro
 
 ## What are variants?
 
-Whenever you create a feature toggle, you can assign it any number of _variants_. This is commonly done in cases where you want to serve your users different versions of a feature to see which performs better.
+Whenever you create a feature toggle, you can assign it any number of _variants_ which can be different depending on the enviroment. This is commonly done in cases where you want to serve your users different versions of a feature to see which performs better.
```
thomasheartman, line 16:
```suggestion
Whenever you create a feature toggle, you can assign it any number of _variants_. This is commonly done in cases where you want to serve your users different versions of a feature to see which performs better. You can assign different variants in each environment.
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
gastonfournier: Actually, we require to have at least one variable variant, and the sum of the fixed ones can be 1000 (leaving the variable variant at zero). So we should change the doc, but at the same time I'm thinking that putting this behaviour here might lead to inconsistencies if we change it later and makes me wonder about the level of detail we want to have because these nuances could be in the OpenAPI doc

thomasheartman: Got it, so it should be something like this instead, then:

```suggestion
* The sum of the weights of variants with `weightType: fix` must be less than or equal to 1000 (≤ 1000)
```

> So we should change the doc, but at the same time I'm thinking that putting this behaviour here might lead to inconsistencies if we change it later and makes me wonder about the level of detail we want to have because these nuances could be in the OpenAPI doc

I'm not entirely sure what you're saying here; could you explain a bit? Anyway, I think this should absolutely go in the OpenAPI doc. This is behavior that people using the endpoints will want to know about. And if it's something that we do end up changing later, then we'll need to update the API docs too.

website/docs/reference/api/legacy/unleash/admin/features-v2.md
gastonfournier: Yes, I agree, it could be part of the release notes or something :thinking:

website/docs/reference/api/legacy/unleash/admin/features-v2.md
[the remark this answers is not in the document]
gastonfournier: I went back in time and this was present up until Unleash 4.17. With 4.18 it got removed, probably because variants were no longer at the feature level. Unfortunately, a breaking change.

website/docs/reference/api/legacy/unleash/admin/features-v2.md
gastonfournier: So, the thing is to add a fixed variant, you first need to add a variable variant. This variable variant needs to have at least weight of 1. So the fixed variant can weight at most 999, but never 1000.

gastonfournier: We could say here something like:
> The sum of the weights of variants with `weightType: fix` must be less than or equal to 1000 (which should include at least one variant with `weightType: variable`)

gastonfournier:
> I'm also unsure about updating the `website/docs/reference/api/legacy/unleash/admin/features-v2.md` and other legacy docs. **All of this information should instead be put into the OpenAPI docs** and we should link to that instead. Because we're striving to be semver-compliant, the old docs should continue to work, just without describing all the new options.

I think if we keep the legacy documentation without updating it, we should at least mark it as stale. I'd like to add a link to https://docs.getunleash.io/reference/api/unleash/features which is where all the endpoints are documented and maybe we can improve the descriptions in OpenAPI. WDYT?

website/docs/reference/api/legacy/unleash/admin/features-v2.md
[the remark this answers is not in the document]
gastonfournier: I left it untouched as it's legacy documentation

website/docs/reference/feature-toggle-variants.md
```diff
@@ -1,23 +1,23 @@
...
 title: Feature Toggle Variants
 ---
-<div class="alert alert--info" role="alert">
-  <em>Feature toggle variants</em> were introduced in <em>Unleash v3.2.0</em>.
-</div>
-<br/>
+:::info Availability
+  **Feature toggle variants** were introduced in Unleash 3.2.
+  **Per-environment variants** were introduced in Unleash 4.21.
+:::
```
thomasheartman, line 7:
```suggestion
:::info Availability

**Feature toggle variants** were first introduced in Unleash 3.2.

The option to have **different variant configurations in each environment** was introduced in Unleash 4.21.

:::
```

website/docs/reference/feature-toggle-variants.md
```diff
@@ -1,23 +1,23 @@
...
 
 Variants facilitate A/B testing and experimentation by letting you create controlled and measurable experiments. Check [our blog post on using Unleash for A/B/n experiments](https://www.getunleash.io/blog/a-b-n-experiments-in-3-simple-steps) for some more insights into how you can set it up.
 
 ## What are variants?
 
-Whenever you create a feature toggle, you can assign it any number of _variants_. This is commonly done in cases where you want to serve your users different versions of a feature to see which performs better.
+Whenever you create a feature toggle, you can assign it any number of _variants_. This is commonly done in cases where you want to serve your users different versions of a feature to see which performs better. You can assign different variants in each environment.
```
thomasheartman, line 15:
```suggestion
Whenever you create a feature toggle, you can assign it any number of _variants_. This is commonly done in cases where you want to serve your users different versions of a feature to see which performs better. A feature can have different variants in each of its environments.
```

website/docs/reference/feature-toggle-variants.md
```diff
@@ -1,23 +1,23 @@
...
 
 A variant has four components that define it:
 - a **name**:
 
-    This must be unique among the toggle's variants. When working with a toggle with variants in a client, you will typically use the variant's name to find out which variant it is.
+    This must be unique among the toggle's variants in an environment. When working with a toggle with variants in a client, you will typically use the variant's name to find out which variant it is.
```
thomasheartman, line 20:
```suggestion
    This must be unique among the toggle's variants in the specified environment. When working with a feature with variants in a client, you will typically use the variant's name to find out which variant it is.
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -696,8 +703,15 @@ Content-Type: application/json; charset=utf-8
 ```
 
 ### PATCH variants for a feature toggle
+:::caution
```
thomasheartman, line 706: This is just for formatting reasons.

```suggestion

:::caution
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -610,8 +610,15 @@ Transfer-Encoding: chunked
 ## Feature Variants
 
 ### Put variants for Feature Toggle {#update-variants}
+:::caution
```
thomasheartman, line 613: Formatting again.

```suggestion

:::caution 
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -696,8 +703,15 @@ Content-Type: application/json; charset=utf-8
 ```
 
 ### PATCH variants for a feature toggle
+:::caution
+
+From 4.21 variants are tied to an environment. Check our Open API docs for the up-to-date docs: https://docs.getunleash.io/reference/api/unleash/features
```
thomasheartman, line 708: I'm not sure what the link does here? What's the purpose of it? Why are you linking them to that page exactly?

```suggestion
Since Unleash 4.21, variants are tied to environments.
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -696,8 +703,15 @@ Content-Type: application/json; charset=utf-8
...
+
+ This endpoint affects all environments at once. If you only want to update a single environment, use [#update-variants-per-environment](#update-variants-per-environment) instead.
+
+:::
 
-<ApiRequest verb="patch" url="api/admin/projects/:projectId/features/:featureName/variants" title="Patch variants for a feature toggle (example data)" payload={[{"op": "add", "path": "/1", "value": {
+<ApiRequest verb="patch" url="api/admin/projects/:projectId/features/:featureName/variants" title="Patch variants for a feature toggle trying to apply the changes in all environments (example data)" payload={[{"op": "add", "path": "/1", "value": {
```
thomasheartman, line 714: Could we simplify it to this?

```suggestion
<ApiRequest verb="patch" url="api/admin/projects/:projectId/features/:featureName/variants" title="Patch variants for a feature toggle in all environments (example data)" payload={[{"op": "add", "path": "/1", "value": {
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -610,8 +610,15 @@ Transfer-Encoding: chunked
 ## Feature Variants
 
 ### Put variants for Feature Toggle {#update-variants}
+:::caution 
 
-<ApiRequest verb="put" url="api/admin/projects/:projectId/features/:featureName/variants" title="Create (overwrite) variants for a feature toggle (example data)" payload={[
+From 4.21 variants are tied to an environment. Check our Open API docs for the up-to-date docs: https://docs.getunleash.io/reference/api/unleash/features
```
thomasheartman, line 615: Same question here regarding the link? Why are we linking them there?

```suggestion
Since Unleash, 4.21 variants are tied to environments.
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -610,8 +610,15 @@ Transfer-Encoding: chunked
...
+
+This endpoint affects all environments at once. If you only want to update a single environment, use [#update-variants-per-environment](#update-variants-per-environment) instead.
```
thomasheartman, line 617: Where is that link supposed to go? I can't seem to find that ID anywhere in this document? And why is the link text just the anchor?

```suggestion
This endpoint affects all environments at once. If you only want to update a single environment, use [#update-variants-per-environment](#update-variants-per-environment) instead.
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -65,6 +65,9 @@ export default class VariantsController extends Controller {
             handler: this.getVariants,
             middleware: [
                 openApiService.validPath({
+                    summary:
```
thomasheartman, line 68: This is too long to be a summary and should be a description instead. Let's try and add a summary at the same time.
```suggestion
                    summary: 'Retrieve variants for a feature (deprecated) ',
                    description:
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -65,6 +65,9 @@ export default class VariantsController extends Controller {
...
+                        '(deprecated from 4.21) Since variants per environment this endpoint will attempt to choose one of the production environments as the source of truth. If more than one production environment is found, the first one will be used.',
```
thomasheartman, line 69:
```suggestion
                        '(deprecated from 4.21) Retrieve the variants for the specified feature. From Unleash 4.21 onwards, this endpoint will attempt to choose a [production-type environment](https://docs.getunleash.io/reference/environments) as the source of truth. If more than one production environment is found, the first one will be used.',
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -80,6 +83,10 @@ export default class VariantsController extends Controller {
             handler: this.patchVariants,
             middleware: [
                 openApiService.validPath({
+                    summary:
+                        'This will apply the same patch to variants in all environments.',
```
thomasheartman, line 87:
```suggestion
                        'Apply a patch to a feature's variants (in all environments).',
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -80,6 +83,10 @@ export default class VariantsController extends Controller {
...
+                    description:
+                        'This method is not atomic which is the main reason why it is not recommended to use.',
```
thomasheartman, line 89: What do we recommend that you do instead of using this endpoint?

```suggestion
                        `Apply the provided patch to the specified feature's variants. The patch conform to [some format] @gastonfournier 
                        
                        ⚠️ **Warning**: This method is not atomic. If something fails in the middle of applying the patch, you can be left with a half-applied patch. We recommend that you instead [do what] ... @gastonfournier?`,
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -96,6 +103,16 @@ export default class VariantsController extends Controller {
...
+                    summary:
+                        'Create (overwrite) variants for a feature toggle in all environments',
+                    description: `This overwrites the current variants for the feature toggle specified in the :featureName parameter in all environments.
+
+                    The backend will validate the input for the following invariants
+
+                    * If there are variants, there needs to be at least one variant with ${'`'}weightType: variable${'`'}
+                    * The sum of the weights of variants with ${'`'}weightType: fix${'`'} must be below 1000 (< 1000)
+
+                    The backend will also distribute remaining weight up to 1000 after adding the variants with ${'`'}weightType: fix${'`'} together amongst the variants of ${'`'}weightType: variable${'`'}`,
```
thomasheartman, line 115: If you just want to escape the backticks, you should be able to use backslashes:

```suggestion
                    * If there are variants, there needs to be at least one variant with \`weightType: variable\`
                    * The sum of the weights of variants with \`weightType: fix\` must be strictly less than 1000 (< 1000)

                    The backend will also distribute remaining weight up to 1000 after adding the variants with \`weightType: fix\` together amongst the variants of \`weightType: variable\``,
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -112,6 +129,9 @@ export default class VariantsController extends Controller {
             handler: this.getVariantsOnEnv,
             middleware: [
                 openApiService.validPath({
+                    summary:
+                        'Get variants for a feature in a specific environment',
+                    description: `Returns the variants for a feature in a specific environment.`,
```
thomasheartman, line 134: What happens if you have no variants?

```suggestion
                    description: `Returns the variants for a feature in a specific environment. If the feature has no variants in this environment ...`,
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -127,6 +147,9 @@ export default class VariantsController extends Controller {
             handler: this.patchVariantsOnEnv,
             middleware: [
                 openApiService.validPath({
+                    summary:
+                        'Patch variants for a feature toggle under the provided environment',
+                    description: `Returns the variants for a feature in a specific environment.`,
```
thomasheartman, line 152: This seems like a copy-paste mistake? Also, what format is the patch in? Is this operation atomic?

```suggestion
                    description: `Apply a patch to the features environments in the specified environment. The patch must follow [some format]. `,
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -127,6 +147,9 @@ export default class VariantsController extends Controller {
             handler: this.patchVariantsOnEnv,
             middleware: [
                 openApiService.validPath({
+                    summary:
+                        'Patch variants for a feature toggle under the provided environment',
```
thomasheartman, line 151:
```suggestion
                        'Patch a feature's variants in an environment',
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -112,6 +129,9 @@ export default class VariantsController extends Controller {
             handler: this.getVariantsOnEnv,
             middleware: [
                 openApiService.validPath({
+                    summary:
+                        'Get variants for a feature in a specific environment',
```
thomasheartman, line 133:
```suggestion
                        'Get variants for a feature in an environment',
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -143,6 +166,17 @@ export default class VariantsController extends Controller {
             handler: this.overwriteVariantsOnEnv,
             middleware: [
                 openApiService.validPath({
+                    summary:
+                        'Create (overwrite) variants for a feature toggle under the provided environment',
```
thomasheartman, line 170:
```suggestion
                        'Create (overwrite) variants for a feature in an environment',
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -96,6 +103,16 @@ export default class VariantsController extends Controller {
             handler: this.overwriteVariants,
             middleware: [
                 openApiService.validPath({
+                    summary:
+                        'Create (overwrite) variants for a feature toggle in all environments',
+                    description: `This overwrites the current variants for the feature toggle specified in the :featureName parameter in all environments.
```
thomasheartman, line 108:
```suggestion
                    description: `This overwrites the current variants for the feature specified in the :featureName parameter in all environments.
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -143,6 +166,17 @@ export default class VariantsController extends Controller {
             handler: this.overwriteVariantsOnEnv,
             middleware: [
                 openApiService.validPath({
+                    summary:
+                        'Create (overwrite) variants for a feature toggle under the provided environment',
+                    description: `This overwrites the current variants for the feature toggle specified in the :featureName parameter for the :environment parameter.
```
thomasheartman, line 171:
```suggestion
                    description: `This overwrites the current variants for the feature toggle in the :featureName parameter for the :environment parameter.
```

src/lib/routes/admin-api/project/variants.ts
```diff
@@ -143,6 +166,17 @@ export default class VariantsController extends Controller {
...
+                        
+                        The backend will validate the input for the following invariants:
+                        
+                        * If there are variants, there needs to be at least one variant with ${'`'}weightType: variable${'`'}
+                        * The sum of the weights of variants with ${'`'}weightType: fix${'`'} must be below 1000 (< 1000)
+                        
+                        The backend will also distribute remaining weight up to 1000 after adding the variants with ${'`'}weightType: fix${'`'} together amongst the variants of ${'`'}weightType: variable${'`'}.
```
thomasheartman, line 178: Same thing for the backticks here.

```suggestion
                    * If there are variants, there needs to be at least one variant with \`weightType: variable\`
                    * The sum of the weights of variants with \`weightType: fix\` must be strictly less than 1000 (< 1000)

                    The backend will also distribute remaining weight up to 1000 after adding the variants with \`weightType: fix\` together amongst the variants of \`weightType: variable\``,
```

thomasheartman (changes requested):
Very nice going here! 💯 I've got some questions and suggestions we should look at before getting this merged, though. Refer to the inline comments for details.

website/docs/reference/feature-toggle-variants.md
gastonfournier: I like it better, but I'm not sure I'd call it an option, because it's not optional. You have to deal with different variants configurations in each environment

src/lib/routes/admin-api/project/variants.ts
gastonfournier: > Apply the provided patch to the specified feature's variants. The patch conforms to [some format] @gastonfournier 

Not sure what you want to say here. Shouldn't our open API schema be enough? 

Internally, we use an implementation of JSON-Patch [(RFC6902)](http://tools.ietf.org/html/rfc6902), but we wrap it with some additional data. 

> ⚠️ **Warning**: This method is not atomic. If something fails in the middle of applying the patch, you can be left with a half-applied patch. We recommend that you instead [do what] ... @gastonfournier?

In this case, we recommend you use the same method by environment. The operation is: https://docs.getunleash.io/reference/api/unleash/patch-environments-feature-variants

src/lib/routes/admin-api/project/variants.ts
gastonfournier:
```suggestion
                    description: `Returns the variants for a feature in a specific environment. If the feature has no variants it will return an empty array of variants`,
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
gastonfournier: It was pointing to the section I added in the previous version. Now it should be pointing to the OpenAPI spec maybe

gastonfournier (commented):
I've included most of the comments but left some others open to continue the conversation

website/docs/reference/feature-toggle-variants.md
thomasheartman: Ah, that's a very good point. I mean, technically, you can have the same variants for all environments still: you just need to keep them in sync. But I agree that it's confusing and potentially misleading 😅 

How about this?

```suggestion
:::info Availability

**Feature toggle variants** were first introduced in Unleash 3.2.
In Unleash 4.21, variants were updated to be **environment-dependent**, meaning the same feature could have different variant configurations in different environments.

:::
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
thomasheartman: Sorry, I'm not sure I understand. Why are you linking them there? Are you saying that the OpenAPI docs you're pointing to are the more updated versions of the doc for this specific endpoint?

gastonfournier: Isn't this documentation "legacy"? We've found some payloads which are not up-to-date (due to a non-intentional change), but OpenApi should have the correct shapes for these API endpoints. This would avoid maintaining the same documentation in two places... 

But we can talk about it, I'm sensing there's something I'm missing here

website/docs/reference/api/legacy/unleash/admin/features-v2.md
thomasheartman: Got it. So as above: it should be linking the the openapi doc for this specific endpoint, right? `/docs/reference/api/unleash/overwrite-feature-variants.api.mdx`? I'll make a suggestion for that.

website/docs/reference/api/legacy/unleash/admin/features-v2.md
thomasheartman: Yes, it is. I don't think you're missing anything. This is legacy documentation and shouldn't really be updated (though I like the idea of adding links to the new docs). So yes, this is a good change. I just wanted to understand where you want to link them, but I think I got it! 😄

website/docs/reference/api/legacy/unleash/admin/features-v2.md
```diff
@@ -697,7 +705,15 @@ Content-Type: application/json; charset=utf-8
 
 ### PATCH variants for a feature toggle
 
-<ApiRequest verb="patch" url="api/admin/projects/:projectId/features/:featureName/variants" title="Patch variants for a feature toggle (example data)" payload={[{"op": "add", "path": "/1", "value": {
+:::caution
+
+From 4.21 variants are tied to an environment. Check our Open API docs for the up-to-date docs: https://docs.getunleash.io/reference/api/unleash/features
+
+ This endpoint affects all environments at once. If you only want to update a single environment, use [#update-variants-per-environment](#update-variants-per-environment) instead.
```
thomasheartman, line 712: So something like this? 

```suggestion
This API documentation is no longer maintained. You should use the [OpenAPI docs for this endpoint](/docs/reference/api/unleash/overwrite-feature-variants.api.mdx) instead.

This endpoint affects all environments at once. If you only want to update a single environment, use the [operation for updating variants per enviroment](/docs/reference/api/unleash/overwrite-feature-variants-on-environments.api.mdx) instead.
```

src/lib/routes/admin-api/project/variants.ts
thomasheartman: > Not sure what you want to say here. Shouldn't our open API schema be enough?

In theory, yes! I had a look now, and I think we've improved the payload for patch schemas. I think it used to be just any object (`{}`). However, the patch schema format is still pretty inscrutable if you haven't used it before (I always need to look it up), so it would be nice to at least have some examples and have it documented.

... buuut that's not the place of this PR, so I think we can leave that as is.

> Internally, we use an implementation of JSON-Patch (RFC6902), but we wrap it with some additional data.

What kind of additional wrapping do we do? Would doing something like this be appropriate?

As for the suggestion, then:

```suggestion
                        `Apply the provided patch to the specified feature's variants. The patch object should conform to the [JSON-patch format (RFC 6902)](https://www.rfc-editor.org/rfc/rfc6902).
                        
                        ⚠️ **Warning**: This method is not atomic. If something fails in the middle of applying the patch, you can be left with a half-applied patch. We recommend that you instead [patch variants on a per-environment basis](/docs/reference/api/unleash/patch-environments-feature-variants.api.mdx), which **is** an atomic operation.`,
```

src/lib/routes/admin-api/project/variants.ts
thomasheartman:
```suggestion
                    description: `Apply a patch to the features environments in the specified environment. The patch object should conform to the [JSON-patch format (RFC 6902)](https://www.rfc-editor.org/rfc/rfc6902).`,
```

website/docs/reference/api/legacy/unleash/admin/features-v2.md
gastonfournier: yes, I think that'd be the best to start making it our source of truth

src/lib/routes/admin-api/project/variants.ts
gastonfournier: It is inscrutable for me as well, I think it's only helpful if you generate it from a diff (which I believe that's what we do in the UI). 

> What kind of additional wrapping do we do? Would doing something like this be appropriate?

To be precise the objects in the body array of https://docs.getunleash.io/reference/api/unleash/patch-environments-feature-variants is what should conform to the RFC spec

website/docs/reference/api/legacy/unleash/admin/features-v2.md
gastonfournier: Yes, but I can't find the linked doc, maybe it's autogenerated? You may know better than I do, so feel free to commit this suggestion because despite the link being correct is what I think we should say

src/lib/routes/admin-api/project/variants.ts
gastonfournier: Yes, also to be precise, the patch object should conform to an array of JSON-patch format...

website/docs/reference/api/legacy/unleash/admin/features-v2.md
thomasheartman: Yeah, the doc is generated, but you can generate it yourself. Anyway, this should link to the right place.

src/lib/routes/admin-api/project/variants.ts
thomasheartman: A great!
```suggestion
                    description: `Apply a list of patches to the features environments in the specified environment. The patch objects should conform to the [JSON-patch format (RFC 6902)](https://www.rfc-editor.org/rfc/rfc6902).`,
```

src/lib/routes/admin-api/project/variants.ts
thomasheartman: Sooo ... this?

```suggestion
                        `Apply a list of patches patch to the specified feature's variants. The patch objects should conform to the [JSON-patch format (RFC 6902)](https://www.rfc-editor.org/rfc/rfc6902).
                        
                        ⚠️ **Warning**: This method is not atomic. If something fails in the middle of applying the patch, you can be left with a half-applied patch. We recommend that you instead [patch variants on a per-environment basis](/docs/reference/api/unleash/patch-environments-feature-variants.api.mdx), which **is** an atomic operation.`,
```

thomasheartman (approved):
Perfect! Let's finally get this in 😄 Thanks a lot for listening to all my input 🙌🏼

PR #2591, opened 2022-12-02
Fix PATCH variants (old endpoint) when variants per environment are enabled

gastonfournier:
## About the changes
This PR addresses some issues when working with variants after migrating to variants per environment.

**Problem:** since PATCH `/api/admin/projects/default/features/${featureName}/variants` does not take into account `featureEnvironments`, when variantsPerEnvironment gets enabled, this method will override the variants in other environments (i.e. not doing a patch). This method has to be maintained because of backward compatibility but it has to be adapted to deal with variants per environment

https://linear.app/unleash/issue/2-476/when-using-patch-for-variants-without-environments-it-wipes-out

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -645,6 +645,7 @@ class FeatureToggleService {
 
     /**
      * GET /api/admin/projects/:project/features/:featureName/variants
+     * @deprecated - Variants should be fetched from FeatureEnvironmentStore (since variants are now; since 4.18, connected to environments)
      * @param featureName
      * @return The list of variants
      */
@@ -1247,9 +1248,26 @@ class FeatureToggleService {
         newVariants: Operation[],
         createdBy: string,
     ): Promise<FeatureToggle> {
-        const oldVariants = await this.getVariants(featureName);
-        const { newDocument } = applyPatch(oldVariants, newVariants);
-        return this.saveVariants(featureName, project, newDocument, createdBy);
+        // const oldVariants = await this.getVariants(featureName);
+        // const { newDocument } = applyPatch(oldVariants, newVariants);
+        // NOTE: saveVariants iterates over all feature_environments overriding the variants in all environments
+        // this.saveVariants(featureName, project, newDocument, createdBy);
+        const ft =
+            await this.featureStrategiesStore.getFeatureToggleWithVariantEnvs(
+                featureName,
+            );
+        const promises = ft.environments.map((env) =>
+            this.updateVariantsOnEnv(
+                featureName,
+                project,
+                env.name,
+                newVariants,
+                createdBy,
+            ).then((resultingVariants) => (env.variants = resultingVariants)),
+        );
+        await Promise.all(promises);
+        ft.variants = ft.environments[0].variants;
+        return ft;
     }
 
     async updateVariantsOnEnv(
@@ -1270,6 +1288,7 @@ class FeatureToggleService {
             environment,
             newDocument,
             createdBy,
+            oldVariants,
         );
     }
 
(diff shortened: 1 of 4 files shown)


src/lib/services/feature-toggle-service.ts
```diff
@@ -1247,9 +1248,26 @@ class FeatureToggleService {
         newVariants: Operation[],
         createdBy: string,
     ): Promise<FeatureToggle> {
-        const oldVariants = await this.getVariants(featureName);
-        const { newDocument } = applyPatch(oldVariants, newVariants);
-        return this.saveVariants(featureName, project, newDocument, createdBy);
```
gastonfournier, line 1252: `saveVariants` iterates over all feature_environments overriding the variants in all environments which was causing this issue

src/lib/services/feature-toggle-service.ts
```diff
@@ -1309,23 +1328,26 @@ class FeatureToggleService {
         environment: string,
         newVariants: IVariant[],
         createdBy: string,
+        oldVariants?: IVariant[],
```
gastonfournier, line 1331: Just a performance optimization to avoid an extra query

src/test/e2e/api/admin/project/variants.e2e.test.ts
```diff
@@ -120,6 +120,104 @@ test('Can patch variants for a feature and get a response of new variant', async
         });
 });
 
+test('Can patch variants for a feature patches all environments independently', async () => {
```
gastonfournier, line 123: This test confirms the issue and validates the fix

src/test/e2e/api/admin/project/variants.e2e.test.ts
```diff
@@ -120,6 +120,104 @@ test('Can patch variants for a feature and get a response of new variant', async
...
+            expect(res.body.variants).toHaveLength(2);
+            // it picks variants from the first environment (sorted by name)
+            expect(res.body.variants[0].name).toBe('dev-variant');
+            expect(res.body.variants[1].name).toBe(addedVariantName);
+        });
+
+    await app.request
+        .get(
+            `/api/admin/projects/default/features/${featureName}?variantEnvironments=true`,
+        )
```
gastonfournier, line 194: We can't use the output of variants to validate how it affects multiple environments, so instead we use this method to gather the variants per environment

src/test/e2e/api/admin/project/variants.e2e.test.ts
```diff
@@ -315,6 +413,11 @@ test('Invalid variant in PATCH also throws 400 exception', async () => {
     await db.stores.featureToggleStore.create('default', {
         name: featureName,
     });
+    await db.stores.featureEnvironmentStore.addEnvironmentToFeature(
+        featureName,
+        'default',
+        true,
+    );
```
gastonfournier, line 420: Because now features should be assigned to environments, we need to modify the test to account for this. I'm not fond of modifying tests to adapt to code changes made in production so I want to bring your attention as a reviewer to this part.

This is where we're fetching the feature environments https://github.com/Unleash/unleash/pull/2591/files#diff-24ad1ee0838798068127b66ed470a16e72f16cd368932059dbb6c6bd9f055b23R1256 where we were previously fetching just features: https://github.com/Unleash/unleash/pull/2591/files#diff-24ad1ee0838798068127b66ed470a16e72f16cd368932059dbb6c6bd9f055b23L1250

nunogois:
Correctly deduping the variants in https://github.com/Unleash/unleash/pull/2591/files#diff-56f835ce2f1786005832407e3f73888d4da6d27a48b7cc806ef847be4ba52161R308-R314 seems to fix the remaining tests.
 
However I'm still unsure if this is the path we should go. Is applying the same patch to every environment individually an intuitive enough behaviour that it can be considered to be what's expected of this endpoint? 🤔

sighphyre:
> Is applying the same patch to every environment individually an intuitive enough behaviour that it can be considered to be what's expected of this endpoint?

It's not great, yeah. Do you have a cunning plan here?

What I'm thinking is that prior to turning on our toggle here, this behaviour should still be rational. After turning on the toggle but making no changes, the behaviour here is... janky but okay. After making changes, this should probably  return a... I dunno, 409 maybe?

sighphyre:
> > Is applying the same patch to every environment individually an intuitive enough behaviour that it can be considered to be what's expected of this endpoint?
> 
> It's not great, yeah. Do you have a cunning plan here?
> 
> What I'm thinking is that prior to turning on our toggle here, this behaviour should still be rational. After turning on the toggle but making no changes, the behaviour here is... janky but okay. After making changes, this should probably return a... I dunno, 409 maybe?

Okay we spoke offline. For a summary for anyone coming here later, each option has edge cases. As it stands, this does what it says on the tin, it's not perfect but only a v5 and a removal of this endpoint will solve that

sighphyre (approved):
LTM

PR #2597, opened 2022-12-05
CR applier should be able to create strategy

kwasniew:
## About the changes

Let the unprotected route to create strategy so that Change Request Applier doesn't need a permission for strategy creation


FredrikOseberg (approved):
LGTM

PR #2605, opened 2022-12-06
fix: favorites column visibility

Tymek:
## About the changes
Favorites column did not work because of `react-table` reloading `initialState` if `columns` change.


sjaanus (approved):
LG

PR #2607, opened 2022-12-06
Fix error in migration when updating feature_environments

sjaanus:
Fixes a issue, where `variants` was null in features table, but column is not nullable.

This is error we got:
```
Failing row contains (default, Retesting, t, null).":"Failed to migrate db null value in column \"variants\" violates not-null constraint
","table":"feature_environments"}
```


sighphyre (approved):
#sad
Nice patch

PR #2611, opened 2022-12-06
feat: add capability to write heap snapshot.

ivarconr:
Is protected behind a config option called `enableHeapSnapshotEnpoint`. When API endpoint is triggered it will write a heapSnapshot to disk. 

Steps to create heap snapshot:

1. Add env variable `ENABLE_HEAP_SNAPSHOT_ENPOINT=true`. 
2. Call this endpoint: `/internal-backstage/heapSnapshot`
3. Download the created heapdump from the server. 

The snapshot can be loaded in the chrome dev-tool to analysis.

### DIFF
--- a/src/lib/routes/backstage.ts
+++ b/src/lib/routes/backstage.ts
@@ -19,7 +19,7 @@ class BackstageController extends Controller {
         }
 
         if (config.server.enableHeapSnapshotEnpoint) {
-            this.get('/heapSnapshot', async (req, res) => {
+            this.get('/heap-snapshot', async (req, res) => {
                 writeHeapSnapshot();
                 res.status(200);
                 res.end('Snapshot written');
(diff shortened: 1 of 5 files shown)


nunogois (approved):
LGTM

src/lib/routes/backstage.ts
```diff
@@ -16,6 +17,14 @@ class BackstageController extends Controller {
                 res.end(await prometheusRegister.metrics());
             });
         }
+
+        if (config.server.enableHeapSnapshotEnpoint) {
+            this.get('/heapSnapshot', async (req, res) => {
```
ivarconr, line 22: Maybe better to do kebab-case?

```suggestion
            this.get('/heap-snapshot', async (req, res) => {
```

PR #2617, opened 2022-12-07
ui flags replaced in index.html

andreas-unleash:

This PR puts the ui flags in the index.html in a meta tag. This makes them accessible without logging in and allows us to track user signups with invite links
## About the changes

Closes #

### DIFF
--- a/frontend/src/component/providers/PlausibleProvider/PlausibleProvider.tsx
+++ b/frontend/src/component/providers/PlausibleProvider/PlausibleProvider.tsx
@@ -11,10 +11,22 @@ export const PlausibleProvider: FC = ({ children }) => {
     const [context, setContext] = useState<ReturnType<typeof Plausible> | null>(
         null
     );
-    const uiFlagsStr =
-        (document.querySelector('meta[name="uiFlags"]') as HTMLMetaElement)
-            ?.content || '';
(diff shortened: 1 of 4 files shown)


frontend/src/component/providers/PlausibleProvider/PlausibleProvider.tsx
```diff
@@ -11,8 +11,13 @@ export const PlausibleProvider: FC = ({ children }) => {
     const [context, setContext] = useState<ReturnType<typeof Plausible> | null>(
         null
     );
+    const uiFlagsStr =
+        (document.querySelector('meta[name="uiFlags"]') as HTMLMetaElement)
+            ?.content || '';
```
FredrikOseberg, line 16: If this alternative clause ever triggers, JSON.parse will throw an error on the next line. Empty string is not valid JSON. Could we put this in a function and wrap it in a try catch?

FredrikOseberg (changes requested):

frontend/src/component/providers/PlausibleProvider/PlausibleProvider.tsx
andreas-unleash: sure

FredrikOseberg (approved):

PR #2625, opened 2022-12-07
fix: update package json and remove empty exports

FredrikOseberg:
* Removes exports statements in e2e tests
* Updates package.json to optionally use heroku url for e2e tests when running locally


kwasniew (approved):
Nice change

PR #2635, opened 2022-12-08
adding missing enterprise check

kwasniew:
## About the changes

One hook hasn't been migrated to the new enterprise check


nunogois (approved):
LGTM

PR #2637, opened 2022-12-08
Conditional swr

kwasniew:
## About the changes

More idiomatic conditional SWR (https://swr.vercel.app/docs/conditional-fetching#conditional)


nunogois (approved):
Nice! 
I assume it's correctly reactive when the condition changes, since it even updates the key. 
We could make sure of this by writing some tests but I think this looks great 👍

PR #2650, opened 2022-12-09
fix: too many reqs to feature toggle endpoint

nunogois:
Fixes an issue where the feature toggle endpoint gets request too many times repeatedly.


kwasniew (approved):
Nice

PR #2666, opened 2022-12-11
fix: move docker-compose to this repo

ivarconr:
Use latest official image for Unleash for simplicity. 
Also takes away the proxy as it is not needed anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,10 @@ If you use the docker compose file from the previous step, here's the configurat
 
 - For front-end SDKs, use:
   - URL: `http://localhost:4242/api/frontend/`
-  - `clientKey`: `<frontend key you create>`
+  - `clientKey`: `default:development.unleash-insecure-frontend-api-token`
 - For server-side SDKs, use:
   - Unleash API URL: `http://localhost:4242/api/`
-  - API token: `<client key you create>`
+  - API token: `default:development.unleash-insecure-api-token`
 
 If you use a different setup, your configuration details will most likely also be different.
 
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ With [`git`](https://git-scm.com/) and [`docker`](https://www.docker.com/) insta
 Run this script:
 
 ```bash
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -53,11 +53,11 @@ Find your preferred SDK in [our list of official SDKs](#unleash-sdks) and import
 If you use the docker compose file from the previous step, here's the configuration details you'll need to get going:
 
 - For front-end SDKs, use:
-  - URL: `http://localhost:3000/proxy`
-  - `clientKey`: `proxy-client-key`
+  - URL: `http://localhost:4242/api/frontend/`
+  - `clientKey`: `<frontend key you create>`
 - For server-side SDKs, use:
   - Unleash API URL: `http://localhost:4242/api/`
-  - API token: `default:development.unleash-insecure-api-token`
```
ivarconr, line 60: should we keep the init key for simplicity?

README.md
```diff
@@ -53,11 +53,11 @@ Find your preferred SDK in [our list of official SDKs](#unleash-sdks) and import
 If you use the docker compose file from the previous step, here's the configuration details you'll need to get going:
 
 - For front-end SDKs, use:
-  - URL: `http://localhost:3000/proxy`
-  - `clientKey`: `proxy-client-key`
+  - URL: `http://localhost:4242/api/frontend/`
+  - `clientKey`: `<frontend key you create>`
```
ivarconr, line 57: should we have an init key for frontend as well?

thomasheartman: I think yes. The more people can just copy and paste, the better. I think having them create their own is going to be more confusing.

README.md
thomasheartman: Again, yeah, I think that's the way to go. That way, people can just go and connect directly instead of having to fiddle in the UI first.

thomasheartman (commented):
Looks great! But yeah, I think putting using init keys makes the most sense. that minimizes the amount of things that new users need to know before getting started.

PR #2667, opened 2022-12-12
docs: update group docs to reflect that users is not optional

sighphyre:
The reference documentation for groups is incorrect, groups always require a list of users

### DIFF
--- a/website/docs/reference/rbac.md
+++ b/website/docs/reference/rbac.md
@@ -118,7 +118,7 @@ A user group consists of the following:
 
 - a **name** (required)
 - a **description** (optional)
-- a list of users (optional)
+- a list of users (required)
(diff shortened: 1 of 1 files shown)


website/docs/reference/rbac.md
```diff
@@ -118,7 +118,7 @@ A user group consists of the following:
 
 - a **name** (required)
 - a **description** (optional)
-- a list of users (optional)
+- a list of users (required)
```
thomasheartman, line 121:
```suggestion
- a **list of users** (required)
```

thomasheartman (approved):
Ooh, did we change this? I feel like we talked about this exact thing when we added the docs in the first place? Anyway, it looks good, but it looks like we're bolding other required fields, so I've added a suggestion to do that for this one as well 😄

PR #2670, opened 2022-12-12
feat: first draft of chart for instance traffic in frontend

chriswk:
## What
We've already added the backend for this. This is the initial work for drawing a chart for instance traffic in the frontend.


frontend/src/component/admin/traffic/Traffic.tsx
```diff
@@ -0,0 +1,11 @@
+import InstanceMetricsChart from '../instance-admin/InstanceMetrics/InstanceMetricsChart'; // TODO move this to traffic folder (pending conversation on how to organize this)
```
gastonfournier, line 1: @chriswk maybe it's worth moving `InstanceMetricsChart` to frontend/src/component/admin/traffic and perhaps renaming it. I'm leaving this as a stylish comment.

gastonfournier (approved):
Looks good! Just a small comment but not a blocker

PR #2673, opened 2022-12-12
remove feature flag: tokens last seen

nunogois:
https://linear.app/unleash/issue/2-470/clean-up-flag


chriswk (commented):
Does this still expose the lastSeen? seems like you removed the useHiddenColumns entry for lastSeen; but I'm not seeing any column being added?

nunogois:
> Does this still expose the lastSeen? seems like you removed the useHiddenColumns entry for lastSeen; but I'm not seeing any column being added?

The columns were always there from this feature's PR, just hidden behind the feature flag that we are now removing.

**API Tokens:**
https://github.com/Unleash/unleash/blob/c27773e8887dd7e56004e9838538bbb69ae5f96e/frontend/src/component/admin/apiToken/ApiTokenTable/ApiTokenTable.tsx#L210-L216
**PATs:**
https://github.com/Unleash/unleash/blob/c27773e8887dd7e56004e9838538bbb69ae5f96e/frontend/src/component/user/Profile/PersonalAPITokensTab/PersonalAPITokensTab.tsx#L134-L140

chriswk (approved):
Perfect :)

PR #2679, opened 2022-12-13
Remove full width access tab

sjaanus:
1. Now all redirect to access tab are going to narrow access tab
2. Remove wide access tab completely, so now it is not possible to access it via url
3. Project settings tab is visible by default now for **ALL USERS**


FredrikOseberg (commented):
Should we put a redirect in, in case people have bookmarked direct URLs to /access ?

sjaanus:
> Should we put a redirect in, in case people have bookmarked direct URLs to /access ?

Good idea, added the redirect.

FredrikOseberg (approved):
LGTM

PR #2692, opened 2022-12-14
feat: expose number of registered applications metric

gastonfournier:
## About the changes
This metric will expose an aggregated view of how many client applications are registered in Unleash. Since applications are ephemeral we are exposing this metric in different time windows based on when the application was last seen. 

The caveat is that we issue a database query for each new range we want to add. Hopefully, this should not be a problem because:
a) the amount of ranges we'd expose is small and unlikely to grow
b) this is currently updated at startup time and even if we update it on a scheduled basis the refresh rate will be rather sparse

## Sample data
This is how metrics will look like
```
# HELP client_apps_total Number of registered client apps aggregated by range by last seen
# TYPE client_apps_total gauge
client_apps_total{range="allTime"} 3
client_apps_total{range="30d"} 3
client_apps_total{range="7d"} 2
```

### DIFF
--- a/src/lib/db/client-instance-store.ts
+++ b/src/lib/db/client-instance-store.ts
@@ -182,14 +182,16 @@ export default class ClientInstanceStore implements IClientInstanceStore {
         return rows.map((r) => r.app_name);
     }
 
-    async getDistinctApplicationsCount(): Promise<number> {
-        return this.db
-            .from(TABLE)
+    async getDistinctApplicationsCount(daysBefore?: number): Promise<number> {
+        let query = this.db.from(TABLE);
+        if (daysBefore) {
+            const after = new Date();
+            after.setDate(after.getDate() - daysBefore);
+            query = query.where('last_seen', '>', after);
+        }
+        return query
             .countDistinct('app_name')
-            .then((res) => {
-                console.log(res);
-                return Number(res[0].count);
-            });
+            .then((res) => Number(res[0].count));
     }
 
     async deleteForApplication(appName: string): Promise<void> {
--- a/src/lib/db/client-instance-store.ts
+++ b/src/lib/db/client-instance-store.ts
@@ -6,7 +6,7 @@ import {
     IClientInstanceStore,
     INewClientInstance,
 } from '../types/stores/client-instance-store';
-import { hoursToMilliseconds } from 'date-fns';
+import { hoursToMilliseconds, subDays } from 'date-fns';
 import Timeout = NodeJS.Timeout;
 
 const metricsHelper = require('../util/metrics-helper');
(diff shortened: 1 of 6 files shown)


src/lib/services/instance-stats-service.ts
```diff
@@ -171,9 +180,33 @@ export class InstanceStatsService {
...
+                    30,
+                ),
+            },
+            {
+                range: '7d',
+                count: await this.clientInstanceStore.getDistinctApplicationsCount(
+                    7,
+                ),
+            },
+        ];
```
gastonfournier, line 207: This will trigger 3 queries against the DB (but right now it only runs at startup time). An alternative would be to do something like:
```sql
select t.range as [score range], count(*) as [number of occurences]
from (
  select case 
    when score between  0 and  9 then ' 0-9 '
    when score between 10 and 19 then '10-19'
    when score between 20 and 29 then '20-29'
    ...
    else '90-99' end as range
  from scores) t
group by t.range
```
which involves nested queries with temporal tables and I'm not sure which one will perform better, but I'm leaning towards keeping this simple because it's not executed frequently. 

The current query does a full scan but adding an index by last_seen might not be good because that column will be updated frequently resulting in multiple index updates.

src/lib/db/client-instance-store.ts
```diff
@@ -182,6 +182,18 @@ export default class ClientInstanceStore implements IClientInstanceStore {
         return rows.map((r) => r.app_name);
     }
 
+    async getDistinctApplicationsCount(daysBefore?: number): Promise<number> {
+        let query = this.db.from(TABLE);
+        if (daysBefore) {
+            const after = new Date();
+            after.setDate(after.getDate() - daysBefore);
```
chriswk, line 189: `subDays(after, daysBefore)` from date-fns? just to be absolutely clear? I see it works surprisingly well with setDate as well, I didn't think `setDate(-1)` would be a good call, but it does what  you hope for, second to last day in previous month.

src/lib/services/instance-stats-service.ts
chriswk: I agree, keep it simple. :)

chriswk (approved):
LGTM

src/lib/db/client-instance-store.ts
gastonfournier: I like the subDays, what I don't like from my solution is that you need to mutate a date to get the result

gastonfournier:
```typescript
test('subdays', () => {
    const theDate = new Date();
    const minus3 = (d: Date) => {
        d.setDate(d.getDate() - 3);
        return d;
    };
    expect(subDays(theDate, 3)).toEqual(minus3(theDate));
});
```
Just double checked this locally (but I don't think I need to add this test)

PR #2693, opened 2022-12-14
feature: add query support to features endpoint

### LINKED ISSUE #2306: Add tag filter to new project features endpoint

I would like to be able filter features by their tags. Something similar to what we already have for /api/admin/features (which is deprecated).

### Background

Previously with our now deprecated /api/admin/features endpoint, we could filter by tags using ?tag[]=TAGTYPE:TAGVALUE (curious detail, this was the first feature I added when I joined in January 2021).

### Solution suggestions

Stealing the logic from our previous /api/admin/features for extracting the tags parameter from query params seems like a good way to go. Our feature service/stores already supports filtering by tags as evidenced by the fact that we can still filter by tags through the deprecated endpoint.

chriswk:
## About the changes
The deprecated /api/admin/features endpoint supported querying with tag and namePrefix parameters. 

This PR adds this functionality to /api/admin/projects/<project>/features as well, allowing to replicate queries that used to work.

Closes #2306

### Important files
src/lib/db/feature-strategy-store.ts
src/test/e2e/stores/feature-strategies-store.e2e.test.ts

## Discussion points
I'm extending our query parameters support for /api/admin/projects/<projectId>/features endpoint. This will be reflected in our open-api spec, so I also made an adminFeaturesQuerySchema for this.

Also, very open for something similar to what we did for the modifyQuery for the archived parameter, but couldn't come up with a good way to support subselects using the query builder, it just ended up blowing the stack. If anyone has a suggestion, I'm all ears.

### DIFF
--- a/src/lib/openapi/spec/admin-features-query-schema.test.ts
+++ b/src/lib/openapi/spec/admin-features-query-schema.test.ts
@@ -1,0 +1,21 @@
+import { validateSchema } from '../validate';
+import { AdminFeaturesQuerySchema } from './admin-features-query-schema';
+
+test('adminFeaturesQuerySchema empty', () => {
+    const data: AdminFeaturesQuerySchema = {};
+
+    expect(
+        validateSchema('#/components/schemas/adminFeaturesQuerySchema', data),
+    ).toBeUndefined();
+});
+
+test('adminFeatureQuerySchema all fields', () => {
+    const data: AdminFeaturesQuerySchema = {
+        tag: [['some-tag', 'some-other-tag']],
+        namePrefix: 'some-prefix',
+    };
+
+    expect(
+        validateSchema('#/components/schemas/adminFeaturesQuerySchema', data),
+    ).toBeUndefined();
+});
--- a/src/lib/openapi/spec/admin-features-query-schema.test.ts
+++ b/src/lib/openapi/spec/admin-features-query-schema.test.ts
@@ -11,11 +11,21 @@ test('adminFeaturesQuerySchema empty', () => {
 
 test('adminFeatureQuerySchema all fields', () => {
     const data: AdminFeaturesQuerySchema = {
-        tag: ['some-tag', 'some-other-tag'],
+        tag: ['simple:some-tag', 'simple:some-other-tag'],
         namePrefix: 'some-prefix',
     };
 
     expect(
         validateSchema('#/components/schemas/adminFeaturesQuerySchema', data),
     ).toBeUndefined();
+});
+
+test('pattern validation should deny invalid tags', () => {
+    const data: AdminFeaturesQuerySchema = {
+        tag: ['something', 'somethingelse'],
+    };
+
+    expect(
+        validateSchema('#/components/schemas/adminFeaturesQuerySchema', data),
+    ).toBeDefined();
 });
--- a/src/lib/openapi/spec/admin-features-query-schema.test.ts
+++ b/src/lib/openapi/spec/admin-features-query-schema.test.ts
@@ -11,7 +11,7 @@ test('adminFeaturesQuerySchema empty', () => {
 
 test('adminFeatureQuerySchema all fields', () => {
     const data: AdminFeaturesQuerySchema = {
-        tag: [['some-tag', 'some-other-tag']],
+        tag: ['some-tag', 'some-other-tag'],
         namePrefix: 'some-prefix',
     };
 
--- a/src/lib/openapi/spec/admin-features-query-schema.ts
+++ b/src/lib/openapi/spec/admin-features-query-schema.ts
@@ -4,18 +4,20 @@ export const adminFeaturesQuerySchema = {
     $id: '#/components/schemas/adminFeaturesQuerySchema',
     type: 'object',
     additionalProperties: true,
+    description: 'Used to filter feature toggles from the admin-api',
     properties: {
         tag: {
             type: 'array',
             items: {
-                type: 'array',
-                items: {
-                    type: 'string',
-                },
+                type: 'string',
             },
+            description:
+                'Used to filter by tags. For each entry, a TAGTYPE:TAGVALUE is expected',
         },
         namePrefix: {
             type: 'string',
+            description:
+                'A case-insensitive prefix filter for the names of feature toggles',
         },
     },
     components: {},
--- a/src/lib/openapi/spec/admin-features-query-schema.ts
+++ b/src/lib/openapi/spec/admin-features-query-schema.ts
@@ -1,0 +1,26 @@
+import { FromSchema } from 'json-schema-to-ts';
+
+export const adminFeaturesQuerySchema = {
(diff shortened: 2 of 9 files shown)


src/lib/openapi/spec/admin-features-query-schema.ts
```diff
@@ -0,0 +1,26 @@
+import { FromSchema } from 'json-schema-to-ts';
+
+export const adminFeaturesQuerySchema = {
+    $id: '#/components/schemas/adminFeaturesQuerySchema',
+    type: 'object',
+    additionalProperties: true,
```
thomasheartman, line 6: What is this schema used for? Is there anything the user should be aware of about the schema in general?

```suggestion
    additionalProperties: true,
    description: 'todo',
```

src/lib/openapi/spec/admin-features-query-schema.ts
```diff
@@ -0,0 +1,26 @@
...
+    properties: {
+        tag: {
+            type: 'array',
```
thomasheartman, line 9: What does this parameter do? And what is it?

```suggestion
            type: 'array',
            description: 'todo',
```

src/lib/openapi/spec/admin-features-query-schema.ts
```diff
@@ -0,0 +1,26 @@
...
+            items: {
+                type: 'array',
+                items: {
+                    type: 'string',
+                },
+            },
+        },
+        namePrefix: {
+            type: 'string',
```
thomasheartman, line 18: What does this query param do? 

```suggestion
            type: 'string',
            description: '',
```

src/lib/openapi/spec/admin-features-query-schema.ts
```diff
@@ -0,0 +1,26 @@
...
+    type: 'object',
+    additionalProperties: true,
+    properties: {
+        tag: {
+            type: 'array',
+            items: {
+                type: 'array',
+                items: {
+                    type: 'string',
+                },
```
thomasheartman, line 14: Just to be clear, is it a list of list of tags? A `Tag[][]`? or is there an extra layer of nesting here?

If it is a list of lists, it definitely needs some extra explanation.

src/lib/openapi/spec/admin-features-query-schema.ts
```diff
@@ -0,0 +1,26 @@
...
+    $id: '#/components/schemas/adminFeaturesQuerySchema',
+    type: 'object',
+    additionalProperties: true,
+    properties: {
+        tag: {
+            type: 'array',
+            items: {
+                type: 'array',
+                items: {
+                    type: 'string',
```
thomasheartman, line 13: What are these strings? Are these tags? What do they correspond to?

```suggestion
                    type: 'string',
                    description: 'todo',
```

thomasheartman (changes requested):
Nice work 🙌🏼  I'd like to see some more information in the schema, both on the overall level and for each property. Also, it'd be nice to add some examples in cases where the auto-generated ones don't make any sense. 

A user should be able to read the schema and understand what it describes, what it's used for, and what all the parts are.

src/lib/openapi/spec/admin-features-query-schema.ts
chriswk: Thank you Thomas. Good point. I'll update the docs.

src/lib/openapi/spec/admin-features-query-schema.ts
chriswk: It used to be a list of lists in the old version. Maybe we can dumb it down to simple an array of strings. Let me see how my tests run.

src/lib/openapi/spec/admin-features-query-schema.ts
```diff
@@ -4,18 +4,20 @@ export const adminFeaturesQuerySchema = {
...
             type: 'array',
             items: {
-                type: 'array',
-                items: {
-                    type: 'string',
-                },
+                type: 'string',
             },
+            description:
+                'Used to filter by tags. For each entry, a TAGTYPE:TAGVALUE is expected',
```
thomasheartman, line 15: Is this correct? The tests above use `['some-tag', 'some-other-tag']`. Should they actually be `['tag-type-a:some-tag', 'tag-type-b:some-other-tag']`? If this is correct, then we should encode that into the validator using the `pattern` property. Something like `pattern: '[A-Za-z0-9_-]+:[A-Za-z0-9_-]+'`. And the test above (as it is written today) should fail because the format is incorrect.

We should also add an example to make it easier to understand.

```suggestion
            description:
                'Used to filter by tags. For each entry, a TAGTYPE:TAGVALUE is expected',
             example: ['simple:ui', 'simple:red'],
```

src/lib/openapi/spec/admin-features-query-schema.ts
```diff
@@ -4,18 +4,20 @@ export const adminFeaturesQuerySchema = {
...
         },
         namePrefix: {
             type: 'string',
+            description:
+                'A case-insensitive prefix filter for the names of feature toggles',
```
thomasheartman, line 20: Can we add an example here? Something like this?

```suggestion
            description:
                'A case-insensitive prefix filter for the names of feature toggles',
            example: 'top_secret_',
```

src/lib/openapi/spec/admin-features-query-schema.test.ts
```diff
@@ -11,7 +11,7 @@ test('adminFeaturesQuerySchema empty', () => {
 
 test('adminFeatureQuerySchema all fields', () => {
     const data: AdminFeaturesQuerySchema = {
-        tag: [['some-tag', 'some-other-tag']],
+        tag: ['some-tag', 'some-other-tag'],
```
thomasheartman, line 14: If tags need to be `type:value`, then this test should fail because these tags are not.

src/lib/openapi/spec/admin-features-query-schema.ts
```diff
@@ -4,18 +4,20 @@ export const adminFeaturesQuerySchema = {
     $id: '#/components/schemas/adminFeaturesQuerySchema',
     type: 'object',
     additionalProperties: true,
+    description: 'Used to filter feature toggles from the admin-api',
```
thomasheartman, line 7: Sorry, I was a bit quick here. I thought this was a data object and not query parameters. The query params object doesn't accept a top-level description (from what I can tell based on [the query params description from swagger.io](https://swagger.io/docs/specification/describing-parameters/#query-parameters). 

```suggestion
```

thomasheartman (changes requested):
Super! It seems I was a bit overzealous with descriptions. They don't go on the top-level for parameters. But I've added some suggestions for examples to make it clearer for the user how to format the params. I also think we can make the spec better by constraining the valid values for tags. It's described closer inline, but I think the test you added should fail because the tag values are in the wrong format.

andreas-unleash (approved):
LG

PR #2694, opened 2022-12-14
Fix/remove settings cache

ivarconr:
In this PR we remove the general SettingService cache, as it will not work across multiple horizontal unleash instances, events are not published across. 

We also fix the CORS origin to: 
- Access-Control-Allow-Origin set to "*" if no Origin is configured
- Access-Control-Allow-Origin set to "*" if any Origin is configured to "*"
- - Access-Control-Allow-Origin set to array and have the "cors" middleware to return an exact match on the user provided Origin.


FredrikOseberg (approved):
LGTM

PR #2697, opened 2022-12-14
fix: speedup build by turning off postgres fsync

ivarconr:
## About the changes

Closes #


ivarconr:
it was not faster after all.

PR #2701, opened 2022-12-15
fix: disable networkView for dev, fail more gracefully

nunogois:
Disables networkView for dev for now. Attempts to fail more gracefully both on the service and front-end.


gastonfournier (approved):
Nicely done :raised_hands: thanks Nuno! :rocket:

chriswk (approved):
LGTM

PR #2705, opened 2022-12-15
fix: for hosted we want to limit the query to that instance

gastonfournier:
## About the changes
This follows up on the traffic UI, right now it shows no data because none of the metrics starts with /api (which is true for all hosted instances). We also want to limit the query to the instance querying the data.

### DIFF
--- a/src/lib/routes/admin-api/metrics.ts
+++ b/src/lib/routes/admin-api/metrics.ts
@@ -21,6 +21,8 @@ class MetricsController extends Controller {
 
     private clientInstanceService: ClientInstanceService;
 
+    readonly baseUriPath: string;
+
     constructor(
         config: IUnleashConfig,
         {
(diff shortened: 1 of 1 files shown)


nunogois (approved):
LG

src/lib/routes/admin-api/metrics.ts
```diff
@@ -189,11 +192,11 @@ class MetricsController extends Controller {
             const hoursToQuery = 6;
             const [clientMetrics, adminMetrics] = await Promise.all([
                 this.clientInstanceService.getRPSForPath(
-                    '/api/client/.*',
+                    `${this.baseUriPath}/api/client/.*`,
```
chriswk, line 195: This will fail if baseUriPath has a trailing slash.

chriswk (approved):
As you pointed out. baseUriPath is expected to never have a trailing slash. :)

PR #2706, opened 2022-12-16
perf: Simplify queries to prometheus

gastonfournier:
## About the changes
Instead of making multiple queries do only one query to prometheus

Screenshot from 2022-12-16 19-05-31
Screenshot from 2022-12-16 19-05-18


src/lib/services/client-metrics/instance-service.ts
```diff
@@ -224,16 +224,17 @@ export default class ClientInstanceService {
...
         }
         const timeoutSeconds = 5;
         const basePath = this.serverOption.baseUriPath;
-        const compositePath = `${basePath}/${path}`.replaceAll('//', '/');
-        const step = '5m'; // validate: I'm using the step both for step in query_range and for irate
-        const query = `sum by(appName) (irate (http_request_duration_milliseconds_count{path=~"${compositePath}"} [${step}]))`;
+        const pathQuery = `${basePath}/api/.*`.replaceAll('//', '/');
+        const step = '5m';
+        const rpsQuery = `irate (http_request_duration_milliseconds_count{path=~"${pathQuery}"} [${step}])`;
+        const query = `sum by(appName, endpoint) (label_replace(${rpsQuery}, "endpoint", "$1", "path", ".*(/api/(?:client/)?[^/]+).*"))`;
```
gastonfournier, line 237: TIL: https://www.regular-expressions.info/brackets.html

`(?:client/)?` is a non-capturing group. This allows us to capture:
- `/api/admin`: because `client/` is optional
- `/api/client/features`
- `/api/client/register`
- `/api/client/metrics`
- `/api/client/frontend`

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
```diff
@@ -22,24 +22,20 @@ export const NetworkOverview = () => {
...
-                        ?.map(result => ({
-                            label: result.metric?.appName,
-                            reqs: parseFloat(
-                                result.values?.[
-                                    result.values?.length - 1
-                                ][1].toString() || '0'
-                            ).toFixed(2),
-                            type: metric.split('Metrics')[0],
-                        }))
-                        .filter(app => app.label !== 'undefined') || []
```
gastonfournier, line 39: Just a personal opinion, but I believe we should also show `undefined` ones

frontend/src/component/admin/network/NetworkTraffic/NetworkTraffic.tsx
```diff
@@ -123,65 +117,66 @@ const createInstanceChartOptions = (
...
-                    metrics[key],
-                    colors.next(),
-                    metricName => `${metricName}: ${key}`
-                )
-            );
-        }
-        return { datasets };
+        return rps.data.result.map(dataset => {
+            const endpoint = dataset.metric?.endpoint || 'unknown';
+            const appName = dataset.metric?.appName || 'unknown';
```
gastonfournier, line 145: This can be 'undefined'  as well, we need to check if we want to have only 'unknown'

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
gastonfournier: Screenshot from 2022-12-19 11-31-15
Screenshot from 2022-12-19 11-31-00

gastonfournier: For now I'm not showing 'undefined' (should we renamed to 'unknown' if we want to show it), just to avoid changing  the current behavior

frontend/src/component/admin/network/NetworkTraffic/NetworkTraffic.tsx
gastonfournier: Used: `(dataset.metric?.appName!).replace(/^undefined$/, 'unknown');`

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
```diff
@@ -22,24 +22,33 @@ export const NetworkOverview = () => {
     const { metrics } = useInstanceMetrics();
 
     const apps: INetworkApp[] = [];
-    if (Boolean(metrics)) {
```
chriswk, line 25: Why the coercion to Boolean? isn't it enough to say `if (metrics) {}`. I'm genuinely curious, not trying to enforce one or the other.

src/lib/services/client-metrics/instance-service.ts
```diff
@@ -224,16 +224,17 @@ export default class ClientInstanceService {
...
         const timeoutSeconds = 5;
-        const basePath = this.serverOption.baseUriPath;
-        const compositePath = `${basePath}/${path}`.replaceAll('//', '/');
-        const step = '5m'; // validate: I'm using the step both for step in query_range and for irate
-        const query = `sum by(appName) (irate (http_request_duration_milliseconds_count{path=~"${compositePath}"} [${step}]))`;
+        const basePath = this.serverOption.baseUriPath.replace(/\/$/, '');
+        const pathQuery = `${basePath}/api/.*`;
+        const step = '5m';
+        const rpsQuery = `irate (http_request_duration_milliseconds_count{path=~"${pathQuery}"} [${step}])`;
+        const query = `sum by(appName, endpoint) (label_replace(${rpsQuery}, "endpoint", "$1", "path", "${basePath}(/api/(?:client/)?[^/]*).*"))`;
```
chriswk, line 237: I kinda liked the possibility to add new queries to this method, but at the same time. This seems safer, so :+1:

chriswk (approved):
Looks sensible :)

nunogois:
Have a PR with some suggestions: https://github.com/Unleash/unleash/pull/2711

It also seems like both `frontend` and `frontend*` are being reported:

image

Even though they are the exact same data, as seen below:

### frontend*
image

### frontend
image

nunogois (changes requested):
I think we should address https://github.com/Unleash/unleash/pull/2706#issuecomment-1357786531 before merging.

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
nunogois: I don't remember the rationale behind it, but I'm pretty sure you're correct that it's not doing anything and should be cleaned up.

gastonfournier:
> I think we should address [#2706 (comment)](https://github.com/Unleash/unleash/pull/2706#issuecomment-1357786531) before merging.

Opened a [thread](https://unleash-internal.slack.com/archives/C048ELND3QD/p1671463310840969) but it seems ok for now to sum the two together as our data in Prometheus shows they might be different things

nunogois (approved):

Issue #2699, opened 2022-12-15
Unleash maintenance mode


sjaanus:
### Describe the feature request

Allow unleash to enter maintenance mode. 

The needs:

- Lock down all changes to unleash during maintenance mode
- Unleash is still accessible in read-only mode
- Needs to have a clear message in the UI
- Client API still works as before. 
    - can still fetch toggles
    - can still post metrics 

### Solution suggestions

- First iteration could just be an env-flag
- locking down POST,PUT, DELETE for /api/admin
- "override permsissions" when fething users. 
- Some kind of a middleware handling this

Iteration 2:

Evaluate if certain users roles should be able to bypass this
Time schedule for when it enters and leaves maintenance mode.

ivarconr:
This feature will go in to beta in v4.20

ivarconr:
Will go in to GA in v4.22

PR #2707, opened 2022-12-16
Middleware mode middlware

sjaanus:
Related to #2699 

1. Added middleware to block all requests to /api/admin during maintenance mode
2. Normal users permissions will be wiped, which means they will see most things locked in UI
3. Admin users will keep their ADMIN permission, so they still see admin panel and can turn off maintenance mode in future

### DIFF
--- a/src/lib/middleware/maintenance-middleware.ts
+++ b/src/lib/middleware/maintenance-middleware.ts
@@ -1,0 +1,24 @@
+import { IUnleashConfig } from '../types';
+import { Request } from 'express';
+
+const maintenanceMiddleware = ({
+    getLogger,
+    flagResolver,
+}: Pick<IUnleashConfig, 'getLogger' | 'flagResolver'>): any => {
+    const logger = getLogger('/middleware/maintenance-middleware.ts');
+    logger.debug('Enabling Maintenance middleware');
+
+    return async (req: Request, res, next) => {
+        try {
+            const writeMethod = ['POST', 'PUT', 'DELETE'].includes(req.method);
+            if (writeMethod && flagResolver.isEnabled('maintenance')) {
+                res.status(503).send({});
+            }
+        } catch (error) {
+            logger.error(error);
+        }
+        next();
+    };
+};
+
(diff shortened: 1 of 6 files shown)


src/lib/middleware/maintenance-middleware.ts
```diff
@@ -0,0 +1,24 @@
...
+    return async (req: Request, res, next) => {
+        try {
+            const writeMethod = ['POST', 'PUT', 'DELETE'].includes(req.method);
+            if (writeMethod && flagResolver.isEnabled('maintenance')) {
+                res.status(503).send({});
+            }
+        } catch (error) {
+            logger.error(error);
+        }
+        next();
```
kwasniew, line 20: we may get an error here with double response sending. One from res.status(503) and one from whatever is next. it's safer to either res.status in middleware or next but not both.

kwasniew (changes requested):

src/lib/middleware/maintenance-middleware.ts
kwasniew: In other words we don't return after line 15 with res.status and the function continues.

sjaanus: I believe send() ends the function. If I would just set res.status() then it would continue.

kwasniew: without return it continues

kwasniew: Very often it leads to this error message in the console: "Can't set headers after they are sent to the client"

kwasniew: try this example reproducing the problem
```js
const express = require("express");
const app = express();

app.use((req, res, next) => {
    res.send("middleware");
    next();
});
app.get('/', (req, res, next) => {
    res.send("handler");
});

app.listen(3000);
```

sjaanus: Yep, seems to be so. Should be fine now.

kwasniew (approved):

PR #2709, opened 2022-12-16
Feat lazy loading network

nunogois:
Adds lazy loading to the network routes, so we end up with smaller chunks in the build.

Also adds a `start:prod` script that can prove useful to test the chunk loading behaviour locally.


gastonfournier (approved):
LGTM!

ivarconr (commented):
Too easy;)

PR #2711, opened 2022-12-19
Simplify prometheus queries suggestions

nunogois:
Some suggestions for https://github.com/Unleash/unleash/pull/2706
Unrelated, but also fixes some random issues with the Mermaid component.


gastonfournier (approved):
Thanks!

PR #2713, opened 2022-12-19
fix: archived feature layout

Tymek:
## About the changes
Update layout and link if you try to open feature toggle that has been archived.

Closes [1-508](https://linear.app/unleash/issue/1-508/refreshing-feature-view-of-the-recently-archived-feature-shows)

## Discussion points
We use soft-delete for projects. I think we should have a banner or other indication, that you cannot change deleted project.


kwasniew (approved):
Approving but I'm wondering if FeatureView should know about the MainLayout. It feels wrong to me. Some outer component should probably own the layout.

FredrikOseberg (changes requested):
I think the correct approach here would be to refactor the DraftBanner so we don't need to use MainLayout in lower components. It feels a bit hacky that we are now leaking MainLayout out into these components.

FredrikOseberg (approved):
LGTM

PR #2714, opened 2022-12-20
fix: set a more generous maxWidth in access timeago cells

nunogois:
https://unleash-internal.slack.com/archives/C046LV6HH6W/p1671533789519689


chriswk (approved):
LGTM

PR #2717, opened 2022-12-20
do etag calculation when memoizing feature toggles

chriswk:
To prevent express from having to do etag calculation of all feature toggle responses even if hitting a memoized feature toggle this PR adds etag calculation to the memoized function. This will cost us an extra etag calculation for client SDK requests without ETag headers, but for SDK requests with ETag headers, it should reduce the number of etag calculations we do down to 1 per 0.6 seconds per query

Co-authored-by: Nuno Gois <nuno@getunleash.ai>

## Still missing
A good way of testing this. We'd like to verify that on average we're not giving Unleash more to do. We believe it will cause some more work for low traffic instances, but save high traffic instances lots of work. So very similar to how our current cached features really work.

## Discussion points we've already covered
* What happens when a features response expires from cache
    * The memoized function will calculate the ETag for the response before returning it to the calling function
    * The new `cachedGetAll` function will then see that the ETags matches and return a 304 instead of passing the full body to express to calculate the ETag for.


src/lib/routes/client-api/feature.ts
```diff
@@ -64,12 +66,13 @@ export default class FeatureController extends Controller {
...
     ) {
         super(config);
-        const { clientFeatureCaching } = config;
+        const { clientFeatureCaching, flagResolver } = config;
         this.featureToggleServiceV2 = featureToggleServiceV2;
         this.segmentService = segmentService;
         this.clientSpecService = clientSpecService;
         this.openApiService = openApiService;
         this.logger = config.getLogger('client-api/feature.js');
+        this.seenEtags = new Map<string, string>();
```
gastonfournier, line 75: This makes me afraid of filling up the memory with garbage. Maybe we can have a fixed-size data structure discarding the oldest values.

I'm not sure if this is a valid concern, but I do expect the key to change frequently over time.

src/lib/routes/client-api/feature.ts
```diff
@@ -138,7 +183,7 @@ export default class FeatureController extends Controller {
             if (!isAllProjects(user.projects)) {
                 override.project = user.projects;
             }
-            if (user.environment !== ALL) {
+            if (user.environment != ALL) {
```
gastonfournier, line 186: Out of curiosity, why are we changing this? Despite being similar, `!==` might perform better as it doesn't do type conversion.

src/lib/routes/client-api/feature.ts
chriswk: Ivar is experimenting with etag calculation based on query params (req.query and key joined together) and the most recent eventId instead of the full body of the response. That experiment will supersede this PR. So we might be closing this.

chriswk:
Superseded by #3062

PR #2718, opened 2022-12-20
feat: allows creation of PATs for other users

nunogois:
https://linear.app/unleash/issue/2-530/api-allow-creation-of-pats-for-other-users

image

Adds and takes into account the following permissions:
 - **READ_USER_PAT**;
 - **CREATE_USER_PAT**;
 - **DELETE_USER_PAT**;
 
 API only, will make some exploration on UI soon.


src/migrations/20221220160345-user-pat-permissions.js
```diff
@@ -0,0 +1,33 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        INSERT INTO permissions (permission, display_name, type) VALUES ('READ_USER_PAT', 'Read PATs for a specific user', 'root');
+        INSERT INTO permissions (permission, display_name, type) VALUES ('CREATE_USER_PAT', 'Create a PAT for a specific user', 'root');
+        INSERT INTO permissions (permission, display_name, type) VALUES ('DELETE_USER_PAT', 'Delete a PAT for a specific user', 'root');
+
+        INSERT INTO role_permission (role_id, permission_id)
```
sjaanus, line 8: Does admin need entry in role_permission? I thought they had access to everything automatically. Just wondering.

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -123,6 +137,50 @@ export default class UserAdminController extends Controller {
...
+                    requestBody: createRequestSchema('patSchema'),
+                    responses: { 200: createResponseSchema('patSchema') },
+                }),
+            ],
+        });
+
+        this.route({
+            method: 'delete',
+            path: '/:id/pat/:patId',
+            acceptAnyContentType: true,
```
sjaanus, line 172: Any specific reason for this?

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -123,6 +137,50 @@ export default class UserAdminController extends Controller {
...
         });
 
+        this.route({
+            method: 'get',
+            path: '/:id/pat',
+            handler: this.getUserPats,
+            permission: READ_USER_PAT,
+            middleware: [
+                openApiService.validPath({
+                    tags: ['Users'],
```
ivarconr, line 147: I do not think we should use this tag. Should start with the experimental tag and be "Admin" eventually.

src/lib/routes/admin-api/user-admin.ts
ivarconr: I think you get invalid content type if not defining it, which really does not make sense for a delete request with empty body. Would maybe be better to not require a content type for delete requests in general (that would be outside of this PR)?

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -123,6 +137,50 @@ export default class UserAdminController extends Controller {
             ],
         });
 
+        this.route({
```
ivarconr, line 140: I would like to have these routes in a separate handler, to avoid to much in this file. Also I think it make sense to move it to the enterprise repo eventually, as service-accounts will be an enterprise feature.

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -475,4 +533,48 @@ export default class UserAdminController extends Controller {
...
+    async createUserPat(
+        req: IAuthRequest<{ id: number }>,
+        res: Response,
+    ): Promise<void> {
+        if (!this.flagResolver.isEnabled('serviceAccounts')) {
+            throw new NotFoundError('Not enabled');
+        }
+        const { id } = req.params;
+        const pat = req.body;
+        const createdPat = await this.patService.createPat(pat, id, req.user);
```
ivarconr, line 560: What happens here if the user is not define? Would love to see some e2e tests validating all happy cases and the edge cases.

src/migrations/20221220160345-user-pat-permissions.js
ivarconr: You are right @sjaanus, admin is special and we do not need to map these permission for that role.

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -475,4 +533,48 @@ export default class UserAdminController extends Controller {
...
+
+    async deleteUserPat(
+        req: IAuthRequest<{ id: number; patId: number }>,
+        res: Response,
+    ): Promise<void> {
+        if (!this.flagResolver.isEnabled('serviceAccounts')) {
+            throw new NotFoundError('Not enabled');
+        }
+        const { id, patId } = req.params;
+        await this.patService.deletePat(patId, id);
```
ivarconr, line 577: What happens here if the user is not define? Would love to see some e2e tests validating all happy cases and the edge cases.

ivarconr (commented):
Looks like a solid first iteration. Would like to move the route handlers into a separate file, maybe named `user-admin-pa.ts`.

src/migrations/20221220160345-user-pat-permissions.js
nunogois: Thanks, addressed in https://github.com/Unleash/unleash/pull/2718/commits/b379484102dabb8d3826535e5cc9fe1e9b3cb8c7

frontend/src/component/providers/AccessProvider/permissions.ts
```diff
@@ -36,3 +36,6 @@ export const DELETE_SEGMENT = 'DELETE_SEGMENT';
 export const APPLY_CHANGE_REQUEST = 'APPLY_CHANGE_REQUEST';
 export const APPROVE_CHANGE_REQUEST = 'APPROVE_CHANGE_REQUEST';
 export const SKIP_CHANGE_REQUEST = 'SKIP_CHANGE_REQUEST';
+export const READ_USER_PAT = 'READ_USER_PAT';
```
gastonfournier, line 39: Do we need READ_USER_PAT? If the user owns the PAT, then they should be the only ones having access to it. Now if we think about service accounts, probably it's not a Personal Access Token but a Service Access Token (implementation-wise they'll be pretty similar, but regarding permissions, it may imply that different "users" or services can access the token). 

Just some random thoughts, let me know what you think.

src/lib/services/pat-service.ts
```diff
@@ -33,24 +33,24 @@ export default class PatService {
         this.eventStore = eventStore;
     }
 
-    async createPat(pat: IPat, user: User): Promise<IPat> {
-        await this.validatePat(pat, user.id);
+    async createPat(pat: IPat, userId: number, creator: User): Promise<IPat> {
```
gastonfournier, line 36: So, the userId is the target of the PAT (or will be the owner), while the creator can be a different person, right? I think this is when an administrator creates a token on behalf of a user (who might not have permission to create it themselves).

Nitpicking, maybe a small rename in the variable can give more clarity

```suggestion
    async createPat(pat: IPat, forUserId: number, creator: User): Promise<IPat> {
```

gastonfournier (approved):
Looks good, just a small stylish comment.

frontend/src/component/providers/AccessProvider/permissions.ts
nunogois: I feel like for all intents and purposes they are still PATs, whether they belong to a user or a service account. The idea is that there's not really much of a difference between regular users and SAs. If we are introducing other-users-PAT-management then I think it makes sense that we should have a permission that controls whether we are able to read them.

Reminder that reading/fetching PATs, whether our own or belonging to other users, never return the secret (PAT itself), only the auxiliary info.

gastonfournier: True, reading a service account PAT would be a good use case for a user reading another ~user~ service account PAT

ivarconr (approved):

PR #2719, opened 2022-12-20
[feat] export features with variants event when feature is disabled

matlockx:
## About the changes

When exporting features one is normally also interested in disabled features, so they are also included in the export file, while the variants are not. I do not see a good reason for that, so this PR removes the check and exports the variants then as well.

Closes #

### DIFF
--- a/src/lib/services/state-service.ts
+++ b/src/lib/services/state-service.ts
@@ -743,7 +743,7 @@ export default class StateService {
             );
             v4.features = v4.features.map((f) => {
                 const variants = featureEnvs.find(
-                    (fe) => fe.enabled !== false && fe.featureName === f.name,
+                    (fe) => fe.featureName === f.name,
                 )?.variants;
                 return { ...f, variants };
             });


gastonfournier:
Hi @matlockx! Thanks for opening this PR! I'm not completely sure why, it might just be something we overlooked. But let me talk about it with the team next week to confirm. From the top of my mind, this makes sense.

gastonfournier:
Hi @matlockx sorry about the delay, I've opened another PR https://github.com/Unleash/unleash/pull/2824 to add tests which confirm the misbehaviour: https://github.com/Unleash/unleash/actions/runs/3838645588/jobs/6535361951#step:5:392 and merged your change on top which confirms fixing the issue: https://github.com/Unleash/unleash/actions/runs/3838676975. 

I'll continue with that one (cause I can't update your PR) and add you as co-author.

matlockx:
👍

sighphyre (approved):
LGTM

PR #2721, opened 2022-12-21
Create nuxt.md

conejerock:
## About the changes
Added  Nuxt2 documentation as `website/docs/reference/sdks/nuxt.md`

### DIFF
--- a/website/docs/reference/sdks/nuxt.md
+++ b/website/docs/reference/sdks/nuxt.md
@@ -1,0 +1,187 @@
+# nuxt-unleash
+
+[![npm version][npm-version-src]][npm-version-href]
+[![npm downloads][npm-downloads-src]][npm-downloads-href]
+[![Github Actions CI][github-actions-ci-src]][github-actions-ci-href]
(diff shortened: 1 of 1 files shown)


thomasheartman:
Hey, @conejerock! Thanks for the work on the module and for the contribution 🙌🏼 However, we don't put community-contributed SDKs docs in the official documentation at the current time, so I'm afraid we can't accept this as it stands now.

Regardless, I had a little look over the package and I have a few questions:

1. Does the package allow the use of authentication / client tokens? I couldn't find an option for it.

2. It seems that the package fetches the toggle configuration directly from the API, but I can't find any place where you evaluate feature toggles. [UnleashFlags.js](https://github.com/conejerock/nuxt-unleash/blob/main/templates/UnleashFlags.js) seems to extract some data about the flags, but does not seem to take into account constraints, segments, etc. I also think that the `isEnabled` call might not be entirely correct. Have you tested the SDK against the [Unleash Client specification](https://github.com/Unleash/client-specification)?

---

I'll close the PR for now, but you can let me know if you think we should keep it open.

Thank you 😄

Issue #2715, opened 2022-12-20
feat: OpenAPI generator for frontend


Tymek:
### Inconsistencies that we get by writing API integration by hand:
- A lot of duplicated code  
- We had bugs where for example HTTP method didn't match  
- It's not clear how to handle errors  
- Cache invalidation on different levels of abstraction

### Things to take into consideration
- Possible errors in parsing data are a big concern  
	- If generated code is trying to be smart and parse JSON and Date, this can cause serious bugs.  
- To test: Can we generate types out of OpenAPI `anyOf` & `oneOf`?  
- We would like to avoid lock-in with the library that we end up using.  
	- We need to be able to compose functionality on top of it, for example with custom fetcher.  
	- Something like "middleware" is the pattern we are looking for. Example use is **conditionally fetching** from an endpoint.  
- Right now to generate types we need to run Enterprise to get OpenAPI, and we generate frontend  
	- It's hard to synchronize changes  
	- We can crate a separate package for entire schema  

### Possible solutions
<details>
  <summary>Dependencies that we can use (click to expand)</summary>
  
- `@openapitools/openapi-generator`  
	- https://github.com/OpenAPITools/openapi-generator  
	- It's **what we currently have**
	- Generates `fetch` API  
	- Doesn't have React hooks generator  
- `orval`  
	- https://github.com/anymaniax/orval  
	- Can generate `SWR` or `React-Query` hooks  
	- Unfortunately it is using `axios`  
		- Shouldn't be hard to modify for `fetch` or `got`  
        - https://github.com/anymaniax/orval/blob/master/packages/swr/src/index.ts  
- `rapini`  
	- https://github.com/rametta/rapini  
	- More opinionated  
	- Also with `axios`  
- `@openapi-codegen`  
	- `React-Query`+`fetch`  
	- https://xata.io/blog/openapi-typesafe-react-query-hooks  
</details>

cobraz:
Why not try GraphQL?

Tymek:
Because it's a huge rewrite with no clear benefits. You're welcome to state your arguments and/or provide a POC.

cobraz:
In my personal experience I find that OpenAPI doesn't have that great tools for generating clients, and I think this mainly comes down to people misunderstanding the OpenAPI specification (or not caring enough, maybe).

AFAIK, GraphQL has better tooling at this point, not to mention better documentation. Another option might be to look at tRPC? The benefits for both options are end-to-end type-safe APIs.

[...]

PR #2724, opened 2022-12-21
Orval generator POC

Tymek:
## About the changes
POC for #2715

### Important files
- `frontend/orval.config.js`
- `frontend/src/openapi/fetcher.ts`


kwasniew:
Can we put this (the actual usage) behind a feature flag so that we can turn it on/off at any time. This is a very dangerous change so I'd like to be on the safe side.

frontend/src/hooks/api/actions/useFeatureApi/useFeatureApi.ts
```diff
@@ -38,15 +37,7 @@ const useFeatureApi = () => {
...
-    const createFeatureToggle = async (
-        projectId: string,
-        createFeatureSchema: CreateFeatureSchema
-    ) => {
-        return openApiAdmin.createFeature({
-            projectId,
-            createFeatureSchema,
-        });
-    };
+    const createFeatureToggle = createFeature;
```
FredrikOseberg, line 40: I realize we had this before, but a few things to check here:
* Does it expose a loading mechanism like useAPI does?
* Does it do basic error handling like useAPI handles for us?
* Does it format the path correctly in terms of using unleash with a base path?

FredrikOseberg (commented):
I think this makes sense. For types I'm happy to start right away. For the API integration I think we should wait until we can validate properly and over time. We can for example try the fetchers on new features, put them behind feature flags and try it out as we are developing instead of modifying existing endpoints.

This way, we'll find edge cases while we are developing, and we can figure out if the generator provides a pleasant experience, or if we are just moving around the technical debt.

frontend/src/hooks/api/actions/useFeatureApi/useFeatureApi.ts
Tymek: Yes to all.
Loading can be done, for **actions** (not SWR getters) since it's not a hook but a `Promise`. We can wrap it if it's more convenient to use a hook for both loading and error handling.

As for path, I thing I already did that in [`frontend/src/openapi/fetcher.ts`](https://github.com/Unleash/unleash/pull/2724/files#diff-31f4e58c2ecb71b1e94d813edc845f92b804a548b43347964891b79dc6ac7a70R23). This approach it very flexible.

Tymek:
@FredrikOseberg @kwasniew About putting it behind a flag: I only substituted 2 OpenAPI getters that where used before (_get features_ and _get features archive_). All model/Schema names where compatible. We still have `frontend/src/interfaces` written by hand (that IMHO we should mostly deprecate).

Files that where using previous auto-generated fetcher that I refactored:
- `frontend/src/component/archive/ProjectFeaturesArchiveTable.tsx` 
- `frontend/src/hooks/api/actions/useFeatureApi/useFeatureApi.ts`
- `frontend/src/hooks/api/getters/useFeaturesArchive/useFeaturesArchive.ts`

frontend/src/hooks/api/getters/useSegments/useSegments.ts
```diff
@@ -16,9 +15,13 @@ export interface IUseSegmentsOutput {
...
 
+    const url = strategyId
+        ? formatApiPath(`api/admin/segments/strategies/${strategyId}`)
+        : formatApiPath('api/admin/segments');
+
     const { data, error, mutate } = useSWR(
-        [strategyId, uiConfig.flags],
-        fetchSegments,
+        url,
+        () => (uiConfig.flags?.SE ? fetchSegments(url) : []),
```
kwasniew, line 24: We tried this approach in change requests and it didn't work. Please check SWR recommended approach: https://swr.vercel.app/docs/conditional-fetching. We have our own hook useConditionalSWR encapsulating the logic.

FredrikOseberg (approved):
Can be merged after adressing comments.

sjaanus:
Couple of questions to cover everything.

1. Do we have example how we regenerate the schema, when we add new endpoint to the backend?
2. If I understand correctly, for **useFeatureApi** we are now using Orval. When looking at that file, does this mean, we still need to know the path and http method by heart? Orval does not generate that?

Tymek:
> Couple of questions to cover everything.
> 
> 1. Do we have example how we regenerate the schema, when we add new endpoint to the backend?

There is a comment in `orval.config.js`. Scripts in `package.json` are fairly easy to understand:
```
"gen:api": "orval --config orval.config.js",
"gen:api:demo": "UNLEASH_OPENAPI_URL=https://app.unleash-hosted.com/demo/docs/openapi.json yarn run gen:api",
```

> 2. If I understand correctly, for **useFeatureApi** we are now using Orval. When looking at that file, does this mean, we still need to know the path and http method by heart? Orval does not generate that?

We don't use methods/actions yet. For now we only use generated types.

PR #2725, opened 2022-12-21
feat: report app names metric dimension only below a threshold

gastonfournier:
## About the changes
Only adds appName dimension in response time metrics when the total number of app names is below a threshold. This reduces the risk of high cardinality

## Discussion points
1. This will apply to every request, so caching properly is really important to avoid impacting latency.
2. Alternatively to memoizee, we could use a fixed interval updating a value in memory that will be used in the comparison.
3. The previous point can also be extended to some properties inside InstantceStatsService that are only calculated at startup time, which would give us a more up-to-date view of the systems

### DIFF
--- a/src/lib/middleware/response-time-metrics.ts
+++ b/src/lib/middleware/response-time-metrics.ts
@@ -2,21 +2,36 @@ import * as responseTime from 'response-time';
 import EventEmitter from 'events';
 import { REQUEST_TIME } from '../metric-events';
 import { IFlagResolver } from '../types/experimental';
+import { InstanceStatsService } from 'lib/services';
+import memoizee from 'memoizee';
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
 const _responseTime = responseTime.default;
 
+const countAppNames = async (instanceStatsService: InstanceStatsService) => {
+    const response = await instanceStatsService.getLabeledAppCounts();
+    return response.find((c) => c.range === '7d').count;
+};
 export function responseTimeMetrics(
     eventBus: EventEmitter,
     flagResolver: IFlagResolver,
+    instanceStatsService: InstanceStatsService,
 ): any {
-    return _responseTime((req, res, time) => {
+    const appNameReportingThreshold = 100;
+    let appNameCount = memoizee(() => countAppNames(instanceStatsService), {
+        promise: true,
+        maxAge: 30000, // milliseconds
+    });
+    return _responseTime(async (req, res, time) => {
         const { statusCode } = res;
 
         const pathname = req.route ? req.baseUrl + req.route.path : '(hidden)';
 
         let appName;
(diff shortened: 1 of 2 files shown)


src/lib/middleware/response-time-metrics.ts
```diff
@@ -2,21 +2,36 @@ import * as responseTime from 'response-time';
...
+    const response = await instanceStatsService.getLabeledAppCounts();
+    return response.find((c) => c.range === '7d').count;
+};
 export function responseTimeMetrics(
     eventBus: EventEmitter,
     flagResolver: IFlagResolver,
+    instanceStatsService: InstanceStatsService,
 ): any {
-    return _responseTime((req, res, time) => {
+    const appNameReportingThreshold = 100;
```
nunogois, line 20: Maybe add this as a const in `src/lib/util/constants.ts` and import it here?

gastonfournier: It could be one option, but I was also thinking that we could use variants (we're already using a feature toggle to control the behavior), so we can modify it if needed (close to being a configuration more than a toggle though)

nunogois: Ooh, that's smart!

src/lib/middleware/response-time-metrics.ts
```diff
@@ -2,21 +2,36 @@ import * as responseTime from 'response-time';
...
+    return _responseTime(async (req, res, time) => {
         const { statusCode } = res;
 
         const pathname = req.route ? req.baseUrl + req.route.path : '(hidden)';
 
         let appName;
-        if (flagResolver.isEnabled('responseTimeWithAppName')) {
+        if (
+            flagResolver.isEnabled('responseTimeWithAppName') &&
+            (await appNameCount()) < appNameReportingThreshold
```
ivarconr, line 33: I am unsure if this does exactly what we want? I know we are deleting instance stats, but will we keep application information forever? If yes, then we should be fine.

gastonfournier: Based on our conversation, appNameCount means `countDistinctAppNamesFromInstances` (i.e. we're maintaining a sliding window of active instances from which we fetch the app names). 

We're caching this count to avoid recalculating on every request, but it's currently done as a write-ahead cache, so it'll likely impact p99 latency. The other option would be to implement a write-behind cache so we don't spend time going to the DB in the user thread

> I know we are deleting instance stats, but will we keep application information forever? If yes, then we should be fine.

We do keep application information forever (we're not changing that), but we're using the instance stats for this count which are deleted after some time. This allows customers who had a surge of instances registered with random application names to recover and re-gain this metric once the older instances are deleted (because this count will go down)

gastonfournier:
Prefer: https://github.com/Unleash/unleash/pull/2737

PR #2727, opened 2022-12-22
Allow hiding environments from the feature overview screen

sjaanus:
1. Allow hiding environments from the feature overview screen
2. Stored in localstorage
3. Configured per user


Tymek (approved):
👍🏻

PR #2729, opened 2022-12-22
fix feature-schema.ts

andreas-unleash:

This PR fixes the admin feature schema by moving the `strategies` (featureStrategySchema) from the root of `feature` object to `feature.environment`
## About the changes

Closes #


thomasheartman (commented):
This looks great! Thanks for making these changes; they are long overdue 🙈 

But before I approve it: What endpoints return this schema? Have they been typed correctly? They _should_ in theory pick up any errors like this, so I suspect that something hasn't been wired right.

PR #2732, opened 2022-12-23
feat: service accounts (hooks)

nunogois:
https://linear.app/unleash/issue/2-539/hooks-for-service-accounts

Creates and adapts hooks for the Service Accounts feature, to be used by the FE.


frontend/src/hooks/api/actions/usePersonalAPITokensApi/usePersonalAPITokensApi.ts
```diff
@@ -37,9 +37,26 @@ export const usePersonalAPITokensApi = () => {
...
+        const req = createRequest(`api/admin/user-admin/${userId}/pat`, {
+            method: 'POST',
+            body: JSON.stringify(payload),
+        });
+        try {
+            const response = await makeRequest(req.caller, req.id);
+            return await response.json();
+        } catch (e) {
+            throw e;
+        }
```
gastonfournier, line 53: Out of curiosity, wouldn't this be the same?
```suggestion
        const response = await makeRequest(req.caller, req.id);
        return await response.json();
```

frontend/src/hooks/api/actions/useServiceAccountsApi/useServiceAccountsApi.ts
```diff
@@ -0,0 +1,66 @@
+import useAPI from '../useApi/useApi';
+
+export interface IServiceAccountPayload {
+    name: string;
+    username: string;
+    rootRole: number;
```
gastonfournier, line 6: Not sure what the `rootRole` is, just asking to learn about it.

gastonfournier (approved):
Just a small comment that won't make a difference, and a question for me to learn ;) Looks good!

frontend/src/hooks/api/actions/usePersonalAPITokensApi/usePersonalAPITokensApi.ts
nunogois: You're right that it should be the same, but this is consistent with the other hooks we have. No idea what the original reasoning was behind it.

frontend/src/hooks/api/actions/useServiceAccountsApi/useServiceAccountsApi.ts
nunogois: `rootRole` should map to one of our root roles in Unleash: 
 - Admin;
 - Editor;
 - Viewer;
 
 Just like regular users, service accounts also have a root role associated with them: 
image

PR #2733, opened 2022-12-23
feat: service accounts - extract PAT form

nunogois:
https://linear.app/unleash/issue/2-540/extract-pat-form

By refactoring the UI logic and extracting the PAT form we can use the same component on our service account creation form.


gastonfournier (approved):
Looks like a good refactor, I can't find anything fishy :D

PR #2734, opened 2022-12-23
feat: service accounts (UI)

nunogois:
https://linear.app/unleash/issue/2-541/ui-for-service-accounts

Adds basic CRUD UI to Service Accounts - New tab/page, table, create/edit form in a modal, token dialog, delete dialog, actions cell.


frontend/src/component/admin/serviceAccounts/ServiceAccounts.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+    return (
+        <div>
+            <AdminMenu />
+            <ConditionallyRender
+                condition={hasAccess(ADMIN)}
+                show={<ServiceAccountsTable />}
+                elseShow={<AdminAlert />}
+            />
+        </div>
```
gastonfournier, line 20: I wonder if we'll need something like https://github.com/Unleash/unleash/blob/fe1c9c70bf57362a38d71b8a6171ece49d52123a/frontend/src/component/admin/auth/AuthSettings.tsx#L39-L45

But probably, not now. At least not until this goes beta :D

gastonfournier (approved):
Looks good! Should there be any UI thingy we can definitely sort it out later

PR #2735, opened 2022-12-23
fix: simplify OAS for state service

ivarconr:
It feels unnecessary to define support for boolean, string and number to define a bool value. The API will accepts all, but the specification should push the user in to use the most appropriate type.

This also ease the down-conversion to Swagger 2.0.

### DIFF
--- a/src/lib/openapi/spec/export-query-parameters.ts
+++ b/src/lib/openapi/spec/export-query-parameters.ts
@@ -15,18 +15,7 @@ export const exportQueryParameters = [
         name: 'download',
         schema: {
             default: false,
-            anyOf: [
-                {
-                    type: 'boolean',
-                },
-                {
-                    type: 'string',
-                    minLength: 1,
-                },
-                {
-                    type: 'number',
-                },
-            ],
+            type: 'boolean',
         },
         description: 'Whether exported data should be downloaded as a file.',
         in: 'query',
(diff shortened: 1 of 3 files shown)


thomasheartman:
Great suggestion, but there's a few things I'd like to discuss before it gets merged.

## Technical: Can we use only "boolean"? 

If we could, then I agree that it would be simpler, but I don't think it will work, and that that is exactly why I ended up using `string | bool | number`.

If I remember correctly, then using "boolean" alone breaks the API because it rejects numbers and strings that it can't coerce into `true` or `false`. If this isn't the case, and if we can use "boolean" and still have it accept everything it accepts today, then yes, we can technically use "boolean" instead.

@ivarconr I notice that you've changed some of the tests to use `true` instead of `1`. Is that because they fail? I feel like I've added tests to check for that before.

## Philosophical: what's the purpose of the specification?

> The API will accepts all, but the specification should push the user in to use the most appropriate type.

I'm not sure I agree here, but I'm also not sure I disagree. The OpenAPI specification is a specification, and the API does accept all of them. So based on that, I think the spec should say something about it. However, if it's technically feasible, then I might be willing to compromise and use "boolean" as the type, but clarify in the parameter description that you can also use numbers and arbitrary strings.

In short, I think reference documentation should try and be as close to reality as possible, even if that reality isn't ideal. 

---

Anyway, if using "boolean" here breaks the API (as I think it might), then the rest of the discussion is redundant (though potentially quite interesting).

Looking forward to hearing your thoughts around this.

ivarconr:
> @ivarconr I notice that you've changed some of the tests to use true instead of 1. Is that because they fail? I feel like I've added tests to check for that before.

I do not think I tested what happens then. I guess it all comes down to how the implementation handle boolean, is it trueish or strictly true/false.

thomasheartman:
If there isn't, then there **should** be a test for that, I think. Again, I'm pretty sure that openapi is quite strict about this, and that only "true" and "false" exactly are accepted as values. Anything else will give you a 400 error. There may be a setting you can tweak to make it more lenient, though 💁🏼

PR #2737, opened 2022-12-23
feat: report app names only if below a threshold

gastonfournier:
## About the changes
Introduce a snapshot version of instanceStats inside instance-stats-service to provide a cached state of the statistics without compromising the DB.

### Important notes
Some rule-of-thumb applied in the PR that can be changed:
1. The snapshot refresh time
2. The threshold to report appName with the metrics

## Discussion points
1. The snapshot could be limited to just the information needed (things like `hasOIDC` don't change until there's a restart)
3. metrics.ts (used to expose Prometheus metrics) has a [refresh interval of 2hs](https://github.com/Unleash/unleash/blob/2d16730cc24ca9d62bd7eaff451853e58cb83482/src/lib/metrics.ts#L189-L195), but with this implementation, we could remove that background task and rely on the snapshot
4. We could additionally update the snapshot every time someone queries the DB to fetch stats (`getStats()` method), but it may increase complexity without a significant benefit


src/lib/middleware/response-time-metrics.ts
```diff
@@ -2,21 +2,44 @@ import * as responseTime from 'response-time';
...
+                const currentCount = stats?.clientApps?.find(
+                    (t) => t.range === '7d',
+                )?.count;
+                // update app count
+                appCount =
+                    currentCount === undefined
+                        ? appNameReportingThreshold
+                        : currentCount;
+            })
+            .catch(() => {});
```
gastonfournier, line 36: I'm not sure if this can be done better. It basically inspects the promise, and if it's resolved it uses the stats information to set the `appCount` variable. If not settled, it just moves on. 

This avoids increasing latency on responseTime calculation

src/lib/services/instance-stats-service.ts
```diff
@@ -109,6 +115,29 @@ export class InstanceStatsService {
...
+
+    async refreshStatsSnapshot(): Promise<void> {
+        // check if stores are present (required for some tests)
+        if (
+            this.featureToggleStore === undefined ||
+            this.userStore === undefined ||
+            this.clientInstanceStore === undefined
+        ) {
+            return;
+        }
```
gastonfournier, line 135: Without this some tests would fail trying to read from userStore when userStore is undefined. I'm not sure why it happens, but it's also not so bad to validate this

src/lib/middleware/response-time-metrics.ts
```diff
@@ -2,21 +2,44 @@ import * as responseTime from 'response-time';
...
+
         let appName;
-        if (flagResolver.isEnabled('responseTimeWithAppName')) {
+        if (
+            flagResolver.isEnabled('responseTimeWithAppName') &&
+            appCount < appNameReportingThreshold
```
gastonfournier, line 41: We can avoid calculating appCount when the flag is disabled

src/lib/middleware/response-time-metrics.ts
```diff
@@ -2,21 +2,28 @@ import * as responseTime from 'response-time';
...
+    return _responseTime(async (req, res, time) => {
         const { statusCode } = res;
-
         const pathname = req.route ? req.baseUrl + req.route.path : '(hidden)';
 
         let appName;
-        if (flagResolver.isEnabled('responseTimeWithAppName')) {
+        if (
+            flagResolver.isEnabled('responseTimeWithAppName') &&
+            (await instanceStatsService.getAppCountSnapshot('7d')) <
```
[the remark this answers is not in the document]
sjaanus: We could utilize memoizee here and cache the response, resulting it querying the database once every hour/day per pod.

src/lib/services/instance-stats-service.ts
ivarconr: I think this is the incorrect approach. I think you instead need to wrap the next line in a try/catch, because the way to start the fetching as an interval a lot of tests will fail due to the database connection being closed just before the refreshStatsSnapshot is called.

src/lib/services/instance-stats-service.ts
```diff
@@ -207,6 +243,19 @@ export class InstanceStatsService {
         ];
     }
 
+    async getAppCountSnapshot(range: string): Promise<number> {
+        return this.snapshot
+            ?.then((stats) => {
+                return (
+                    stats?.clientApps?.find((t) => t.range === range)?.count ||
```
FredrikOseberg, line 250: `t` throws me off here. If it's a clientApp, why not use `clientApp` as the parameter name?

FredrikOseberg (commented):
Couldn't the refresher handle the asynchronicity and just set the data directly on the instance instead of returning a promise? I don't think the async actions should leak out into the middlewares. Then getAppCountSnapshot could just check if the data is there and return the data, or if not, return a default value.

While the async action probably does not add a lot of overhead in this case, it's using a pattern that may lead to someone mistakingly using this pattern in another scenario that actually touches the database. I would prefer to not use async in middlewares to avoid such problems in the future.

src/lib/services/instance-stats-service.ts
gastonfournier: Yeah, naming naming naming :P it's `t` for tuple, but I think `clientApp` representing (range, count) pair is more meaningful

src/lib/services/instance-stats-service.ts
gastonfournier: Oh, that's a good idea! I'll look into it. Thanks!!!

src/lib/middleware/response-time-metrics.ts
```diff
@@ -2,21 +2,28 @@ import * as responseTime from 'response-time';
...
         const { statusCode } = res;
-
         const pathname = req.route ? req.baseUrl + req.route.path : '(hidden)';
 
         let appName;
-        if (flagResolver.isEnabled('responseTimeWithAppName')) {
+        if (
+            flagResolver.isEnabled('responseTimeWithAppName') &&
+            (instanceStatsService.getAppCountSnapshot('7d') ||
+                appNameReportingThreshold) < appNameReportingThreshold
```
gastonfournier, line 25: If app count was undefined, fallback to `appNameReportingThreshold`. This prevents us from reporting app names until we have the data which might protect us from some edge cases

src/lib/services/instance-stats-service.ts
gastonfournier: The try catch: https://github.com/Unleash/unleash/blob/cb4b765eeea0c47ddaa3ee36f982e2c38d691ef0/src/lib/services/instance-stats-service.ts#L129-L142 doesn't seem to catch `Cannot read properties of undefined` exception types... :thinking: 

I'm not sure how to fix this https://github.com/Unleash/unleash/actions/runs/3849713375/jobs/6559023513#step:5:576

gastonfournier: https://github.com/Unleash/unleash/pull/2737#discussion_r1062817463

src/lib/middleware/response-time-metrics.ts
gardleopard: if flagresolver is false, will it evaluate the stuff after &&?

gardleopard: what will appname be if we have this disabled or too many apps?

gastonfournier: if flagresolver is false it won't resolve the other part of the `&&`, it's a short-circuit and ;)

> what will appname be if we have this disabled or too many apps?

It won't have appname, which will translate to `"undefined"` (as the current behavior).

Example from prometheus metrics:
```
# HELP http_request_duration_milliseconds App response time
# TYPE http_request_duration_milliseconds summary
http_request_duration_milliseconds{quantile="0.1",path="/*",method="GET",status="304",appName="undefined"} 8.32329
http_request_duration_milliseconds{quantile="0.5",path="/*",method="GET",status="304",appName="undefined"} 8.32329
http_request_duration_milliseconds{quantile="0.9",path="/*",method="GET",status="304",appName="undefined"} 8.32329
http_request_duration_milliseconds{quantile="0.95",path="/*",method="GET",status="304",appName="undefined"} 8.32329
http_request_duration_milliseconds{quantile="0.99",path="/*",method="GET",status="304",appName="undefined"} 8.32329
http_request_duration_milliseconds_sum{path="/*",method="GET",status="304",appName="undefined"} 8.32329
http_request_duration_milliseconds_count{path="/*",method="GET",status="304",appName="undefined"} 1
http_request_duration_milliseconds{quantile="0.1",path="(hidden)",method="GET",status="200",appName="undefined"} 8.339421
http_request_duration_milliseconds{quantile="0.5",path="(hidden)",method="GET",status="200",appName="undefined"} 9.418512
http_request_duration_milliseconds{quantile="0.9",path="(hidden)",method="GET",status="200",appName="undefined"} 10.466320999999999
http_request_duration_milliseconds{quantile="0.95",path="(hidden)",method="GET",status="200",appName="undefined"} 10.466320999999999
http_request_duration_milliseconds{quantile="0.99",path="(hidden)",method="GET",status="200",appName="undefined"} 10.466320999999999
http_request_duration_milliseconds_sum{path="(hidden)",method="GET",status="200",appName="undefined"} 28.224254000000002
http_request_duration_milliseconds_count{path="(hidden)",method="GET",status="200",appName="undefined"} 3
http_request_duration_milliseconds{quantile="0.1",path="(hidden)",method="GET",status="401",appName="undefined"} 1.847349
http_request_duration_milliseconds{quantile="0.5",path="(hidden)",method="GET",status="401",appName="undefined"} 4.5549409999999995
http_request_duration_milliseconds{quantile="0.9",path="(hidden)",method="GET",status="401",appName="undefined"} 11.237288
http_request_duration_milliseconds{quantile="0.95",path="(hidden)",method="GET",status="401",appName="undefined"} 11.237288
http_request_duration_milliseconds{quantile="0.99",path="(hidden)",method="GET",status="401",appName="undefined"} 11.237288
http_request_duration_milliseconds_sum{path="(hidden)",method="GET",status="401",appName="undefined"} 27.669517999999997
http_request_duration_milliseconds_count{path="(hidden)",method="GET",status="401",appName="undefined"} 5
```

src/lib/services/instance-stats-service.ts
```diff
@@ -111,7 +119,37 @@ export class InstanceStatsService {
         this.logger = getLogger('services/stats-service.js');
     }
 
-    async getToggleCount(): Promise<number> {
+    async start(): Promise<void> {
+        if (!this.snapshotRefresher) {
+            this.refreshStatsSnapshot();
```
sighphyre, line 124: This doesn't get awaited, is there a reason for that?

src/lib/services/instance-stats-service.ts
```diff
@@ -111,7 +119,37 @@ export class InstanceStatsService {
...
+                () => this.refreshStatsSnapshot(),
+                minutesToMilliseconds(5),
+            ).unref();
+        }
+    }
+
+    async refreshStatsSnapshot(): Promise<void> {
+        try {
+            this.snapshot = await this.getStats();
+            let appCountReplacement = {};
```
sighphyre, line 135: Minor detail, this doesn't look like it ever gets reassigned, it can be const
```suggestion
            const appCountReplacement = {};
```

src/lib/services/instance-stats-service.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    expect(instanceStatsService.getStats).toBeCalledTimes(2);
+
+    jest.advanceTimersByTime(minutesToMilliseconds(5) + 1);
+
+    // after 5 minutes snapshot should have been refreshed
+    expect(instanceStatsService.refreshStatsSnapshot).toBeCalledTimes(3);
+    expect(instanceStatsService.getStats).toBeCalledTimes(3);
+});
+
+test('before the snapshot is refreshed we can still get the appCount', async () => {
```
sighphyre, line 66: What does this test do? It says we can still get the appCount but then it asserts that the result is undefined?

sighphyre (approved):
LGTM, I've left some small comments

src/lib/services/instance-stats-service.ts
gastonfournier: My idea is that by not waiting on it, it'll run in the background. Anyhow this `start` method will go away as I plan to migrate to the newly scheduler service introduced a few days ago

src/lib/services/instance-stats-service.test.ts
gastonfournier: This is to test that if you call `getAppCountSnapshot` before the snapshot is refreshed it does not fail/explode. Maybe it'd be more meaningful to say: `expect(...).not.toThrow()` or something like that...

PR #2738, opened 2022-12-26
fix: variant weight display when decimals

### LINKED ISSUE #2222: bug: Variants "Fix Setting" has Incorrect Behavior
### Describe the bug

Using the "fix" percentage on variants causes issues. Setting to 33.33% can cause an error (event logs show the weight being set as 333.29999999999995 while the other (variable set) percentage is set to 667.
This can also be seen by using any more than 1 decimal point. Using two or more causes the weights to add up over 1000. This causes the UI to show a percentage, but the actual weight is different. Ex. setting a variant to 55.555% causes one weight to be 505.55 and the other to be 495. This means the total is 1000.55, meaning 495/1000.55 is actually 49.47%. Not a huge issue but it is incorrect. 
Screenshot 2022-10-20 at 15 29 41
Screenshot 2022-10-20 at 15 26 22

1. Create a feature toggle
2. add at least two variants
3. set one variant percentage manually with more than 1 decimal point
4. check event logs for actual weights set in the system 

### Expected behavior

I would expect weights to accurately add up to 1000 and to be accurate in the UI so users can correctly set up experiments. 

### Logs, error output, etc.

_No response_

### Screenshots

Screenshot 2022-10-20 at 15 29 41
Screenshot 2022-10-20 at 15 26 22

### Additional context

_No response_

### Unleash version

4.16.1

### Subscription type

Enterprise

### Hosting type

Hosted by Unleash

_No response_

gastonfournier:
## About the changes
When displaying decimal values for variants fixed percentages, we divide the value by `10.0` which causes some fix point imprecision:
Screenshot from 2022-12-26 12-14-24

To avoid rounding issues, we're manually shifting the comma.

After this change the numbers should show correctly:
Screenshot from 2022-12-26 12-14-10

Closes #2222 

## Discussion points
If there's a better way to avoid precision issues with float divisions I'm happy to change the implementation. The only alternative I found is using `toFixed`:
`(weight / 10.0).toFixed(precision)`

But the challenge with this approach is finding what `precision` is (because we don't impose a limit on the precision). The solution is to find the precision of `weight` and subtract 1 from it, but the method I've found to get the precision of a float implies converting it to a string, so instead I decided to use the string to shift the comma.

### DIFF
--- a/frontend/src/component/common/util.ts
+++ b/frontend/src/component/common/util.ts
@@ -45,6 +45,15 @@ export const parseValidDate = (value: string): Date | undefined => {
 };
 
 export const calculateVariantWeight = (weight: number) => {
+    if (!Number.isInteger(weight)) {
+        const parts = String(weight).split('.');
+        return parseFloat(
+            `${parts[0].slice(0, -1)}.${parts[0][parts[0].length - 1]}${
+                parts[1]
+            }`
+        );
+    }
+
     return weight / 10.0;
 };
 
--- a/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantModal/EnvironmentVariantModal.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantModal/EnvironmentVariantModal.tsx
@@ -27,6 +27,7 @@ import cloneDeep from 'lodash.clonedeep';
 import { CloudCircle } from '@mui/icons-material';
 import PermissionSwitch from 'component/common/PermissionSwitch/PermissionSwitch';
 import { UPDATE_FEATURE_VARIANTS } from 'component/providers/AccessProvider/permissions';
+import { calculateVariantWeight } from 'component/common/util';
 
 const StyledFormSubtitle = styled('div')(({ theme }) => ({
     display: 'flex',
@@ -193,7 +194,7 @@ export const EnvironmentVariantModal = ({
         if (variant) {
             setName(variant.name);
             setCustomPercentage(variant.weightType === WeightType.FIX);
-            setPercentage(String(variant.weight / 10));
+            setPercentage(String(calculateVariantWeight(variant.weight)));
             setPayload(variant.payload || EMPTY_PAYLOAD);
             overridesDispatch(
                 variant.overrides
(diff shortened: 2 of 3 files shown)


kwasniew:
Shouldn't the percentages add up to 100%? In the attached image they don't.
I'd check what libraries like this one are doing: https://github.com/super-ienien/percent-round
We can even use it directly as it's very small and has no transitive deps

gastonfournier:
> Shouldn't the percentages add up to 100%?

Yes, but I believe the problem is elsewhere in the math we're doing to calculate the remainder variable percentage. But there's a question we need to address:
1. How many decimals do we allow customers to have? This may require changes in the UI. If we want to maintain this unconstrained (some customers may want/need 3 or 4 decimals), then we need to fix the math (I believe I know where the problem is and this library can help)

I'll look at the math, probably we're assuming only 1 decimal place and storing that as Integer. Well spotted!

gastonfournier:
> I'll look at the math, probably we're assuming only 1 decimal place and storing that as Integer

This is where things mess up: https://github.com/Unleash/unleash/blob/2d16730cc24ca9d62bd7eaff451853e58cb83482/frontend/src/component/common/util.ts#L80-L82

The parseInt means the result will have no decimals and since we're dividing this result by `10.0` it means variable percentages will have only 1 decimal. I'll look into fixing that, but the question remains. If we agree to support only 1 decimal, we can easily use the library you suggested (i.e.  `percentRound([10.34, 20.043, 30.04, 39.567], 1);`), but I feel we should lean towards flexibility

gastonfournier:
The value in the DB (another relevant piece of data), is stored as follows:
```json
[
 {"name": "variant1", "weight": 445, "overrides": [], "stickiness": "default", "weightType": "variable"}, 
 {"name": "variant2", "weight": 555.55, "overrides": [], "stickiness": "default", "weightType": "fix"}
]
```
with the weight of variable weight type having no decimals, but the other weight (which we store as given by the user input), do have decimals

gastonfournier:
Found the docs: https://docs.getunleash.io/reference/feature-toggle-variants#variant-weight
> It is a numeric value between 0 and 100 (inclusive) with one decimal's worth of precision.

gastonfournier:
Closing in favor of https://github.com/Unleash/unleash/pull/2749

PR #2739, opened 2022-12-26
feat: add error logs for addons

### LINKED ISSUE #2057: Improve logging for failing addon event handling

To help with debugging failing webhooks, I would like Unleash to start logging the actual error being caught in the fetch call

### Background

Trying to help a customer with failing webhooks that were simply logging
```
[INFO] addon/webhook - Handled event "feature-environment-disabled". Status code: undefined
```
today, made me realise that we're not giving ourselves or self-hosted users very much to go on when addons fail.

### Solution suggestions

We should probably in addon.ts fetchRetry's catch log the error with a separate logger so that it can be muted for our hosted solution and only turned on if we need to help debug it.

Something like

```
 async fetchRetry(
        url: string,
        options = {},
        retries: number = 1,
    ): Promise<Response> {
        let res;
        try {
            res = await fetch(url, {
                retry: {
                    retries,
                },
                ...options,
            });
            return res;
        } catch (e) {
    //THIS LINE: this.addonFetchLogger.error(e);
            res = { statusCode: e.code, ok: false };
        }
        return res;
    }
```

gastonfournier:
## About the changes
This uses log-level warning so we can turn off the log without turning off all error logs. This could potentially be tweaked further by using log categories in the log configuration.

Closes #2057

### DIFF
--- a/src/lib/addons/addon.ts
+++ b/src/lib/addons/addon.ts
@@ -39,7 +39,7 @@ export default abstract class Addon {
 
     async fetchRetry(
         url: string,
-        options = {},
+        options: any = {},
         retries: number = 1,
     ): Promise<Response> {
         let res;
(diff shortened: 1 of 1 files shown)


sighphyre (approved):
LGTM

PR #2749, opened 2022-12-27
fix: enforce weight precision to 1 decimal

### LINKED ISSUE #2222: bug: Variants "Fix Setting" has Incorrect Behavior

gastonfournier:
## About the changes
According to our docs, we only support up to 1 decimal place for weights. This is to use integers to represent the percentages (we divide them by 10) and supporting more decimals results in bad maths due to floating point arithmetics. 

This PRs adds Frontend and Backend validations to enforce this restriction

Closes #2222

## Discussion points
Should we reconsider supporting more decimal places, that door remains open, but for now we'll just adhere to our documentation because that change would require some development.

### DIFF
--- a/src/lib/schema/feature-schema.test.ts
+++ b/src/lib/schema/feature-schema.test.ts
@@ -70,6 +70,56 @@ test('should allow weightType=fix', () => {
     expect(value).toEqual(toggle);
 });
 
+test('should not allow weightType=fix with floats', () => {
+    const toggle = {
+        name: 'app.name',
+        type: 'release',
+        project: 'default',
+        enabled: false,
+        impressionData: false,
+        stale: false,
+        archived: false,
+        strategies: [{ name: 'default' }],
+        variants: [
+            {
+                name: 'variant-a',
+                weight: 1.5,
+                weightType: 'fix',
+                stickiness: 'default',
+            },
+        ],
+    };
+
+    const { error } = featureSchema.validate(toggle);
+    expect(error.details[0].message).toEqual('Weight only supports 1 decimal');
+});
+
+test('should not allow weightType=fix with more than 1000', () => {
+    const toggle = {
+        name: 'app.name',
+        type: 'release',
+        project: 'default',
+        enabled: false,
+        impressionData: false,
+        stale: false,
+        archived: false,
+        strategies: [{ name: 'default' }],
+        variants: [
+            {
+                name: 'variant-a',
+                weight: 1001,
+                weightType: 'fix',
+                stickiness: 'default',
+            },
+        ],
+    };
+
+    const { error } = featureSchema.validate(toggle);
+    expect(error.details[0].message).toEqual(
+        '"variants[0].weight" must be less than or equal to 1000',
+    );
+});
+
 test('should disallow weightType=unknown', () => {
     const toggle = {
         name: 'app.name',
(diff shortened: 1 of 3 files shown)


src/lib/schema/feature-schema.ts
```diff
@@ -50,7 +50,13 @@ export const variantValueSchema = joi
 
 export const variantsSchema = joi.object().keys({
     name: nameType,
-    weight: joi.number().min(0).max(1000).required(),
+    weight: joi
+        .number()
+        .integer()
+        .message('Weight only supports 1 decimal')
```
gastonfournier, line 56: This is just to improve the error message. Otherwise it'll show that the value has to be integer which is misleading for oru users

src/lib/schema/feature-schema.test.ts
```diff
@@ -70,6 +70,56 @@ test('should allow weightType=fix', () => {
...
+                name: 'variant-a',
+                weight: 1.5,
+                weightType: 'fix',
+                stickiness: 'default',
+            },
+        ],
+    };
+
+    const { error } = featureSchema.validate(toggle);
+    expect(error.details[0].message).toEqual('Weight only supports 1 decimal');
```
chriswk, line 94: Why should this fail? There's only 1 decimal in the actual variant?

gastonfournier: The problem is that our weight range goes from 0 to 1000, where 1000 represents 100% and 1.5 represents 0.15%. This way of representing values as integers is common in representing money, but it is a bit confusing IMO, especially when you see things like this: 
1. I send 1 to the API (or 1.5 to the API)
2. I get 0.1% as result (I get only 1 decimal is supported when I sent only 1 decimal)

It has the overhead of having to know that the API payload is being interpreted differently

nunogois (approved):
LGTM 👍

src/lib/schema/feature-schema.test.ts
gastonfournier: Exactly, the error message is confusing but I wanted to test the two error messages (the old one in the text below and the new one in this test). The test description seems right IMO: `should not allow weightType=fix with floats` which is exactly what we want, but the message we're validating is a bit misleading from a backend perspective. If we could just validate against some error type (e.g. `INVALID_TYPE`) that'd be easier.

PR #2765, opened 2022-12-28
Fix crashing search bars

sjaanus:
image
If you enter characters like ()[] in the search, following error occurs:
image

**This PR escapes regex characters and makes the app not crash.**


gastonfournier (approved):
Nice! Just double checked and found https://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex with the same implementation :+1:

PR #2776, opened 2022-12-29
feat: limit the amount of results coming from Prometheus

gastonfournier:
## About the changes
To avoid showing too much data in the traffic screen, limit the number of results to `topk`. 

## Discussion points
Top 10 is a rule of thumb, but maybe we could do top 25. Until we gather more data, I believe this should be good enough


gardleopard (approved):
Fantastic!

PR #2777, opened 2022-12-29
formatApiPath / formatAssetPath - Implement smart functionality to detect double subpath

sjaanus:
Today we have two functions that wrap our paths with the subpath. Since all of our customers are hosted on a subpath, such as /eubb1001, we need to account for this path when building API paths and asset paths. 

These functions could be smarter, we could, for example detect if we have already added the base path and ignore it if it already exists.

**This PR implements this and adds 2 capabilities:**

1. If there is list of paths that need to be joined and one is subset of another, it is removed.
2. All duplicate paths in the list are removed

### DIFF
--- a/frontend/src/utils/formatPath.ts
+++ b/frontend/src/utils/formatPath.ts
@@ -24,7 +24,9 @@ export const parseBasePath = (value = basePathMetaTagContent()): string => {
 // Join paths with a leading separator and without a trailing separator.
 const joinPaths = (...paths: string[]): string => {
     const filteredPaths = paths.filter(path => {
-        return !paths.some(i => i !== path && i.includes(path));
+        return !paths.some(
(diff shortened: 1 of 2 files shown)


frontend/src/utils/formatPath.ts
```diff
@@ -23,7 +23,12 @@ export const parseBasePath = (value = basePathMetaTagContent()): string => {
 
 // Join paths with a leading separator and without a trailing separator.
 const joinPaths = (...paths: string[]): string => {
-    return ['', ...paths]
+    const filteredPaths = paths.filter(path => {
+        return !paths.some(i => i !== path && i.includes(path));
```
FredrikOseberg, line 27: Could we be more explicit here, I find `i` to be a bit confusing in this context. `currentPath` ?

FredrikOseberg (approved):

PR #2778, opened 2022-12-29
UI Flags cleanup

sjaanus:
We had multiple flags that were unused.
This PR removes unused flags.

Also added a small writeup about how UI flags are working.

https://docs.google.com/document/d/1-jenEiazoG6MkD9thBQuEw_jm8CkUUjeEMcsxfOBUzM/edit#heading=h.6wl0ylk15vcj


frontend/src/interfaces/uiConfig.ts
```diff
@@ -29,20 +29,16 @@ export interface IProclamationToast {
 }
 
 export interface IFlags {
-    C: boolean;
     P: boolean;
     E: boolean;
     RE: boolean;
     EEA?: boolean;
-    OIDC?: boolean;
```
FredrikOseberg, line 37: I'm surprised this isn't in use, but this is perhaps a leftover from implementing OIDC as an alternative to Google login.

FredrikOseberg (approved):
Nice cleanup.

PR #2781, opened 2022-12-30
Make it possible to validate enterprise schema with respondWithValidation

sjaanus:
Now respondWithValidation<T, S = SchemaId> can be called in oss and enterprise to validate against needed schema.


FredrikOseberg (approved):
LGTM

PR #2797, opened 2023-01-03
Docs(fix): broken links

thomasheartman:
This change fixes some broken links in the "about the docs" document. Most of these links go to concrete documents and have been replaced with their appropriate relative file links. The few links that go to auto-generated pages have been corrected.

Unfortunately, there does not seem to be an easy way to link safely to generated pages.


gastonfournier (approved):
:+1:

Issue #2782, opened 2022-12-30
Features are always disabled


anogr:
I'm new to Unleash. I have installed the self-hosted version and got it up and running on a K8s cluster. Everything appears fine.

So I created a feature toggle, and enabled it in development and production. Then, I created a Node.js app as described in the Node.js SDK docs. I configured the API key, set the environment to development and the call to startUnleash() appears to be fine.

Then, if I check if the feature toggle is enabled, it returns false. I can see that the feature toggle was found because the fallback function was not called. I tried with a feature toggle that isn't defined and the fallback function was called so Unleash actually knows that the feature toggle exists. Strange.

Even more strange, if I make an explicit HTTP call to my.server.host/api/client/features I get the feature toggle in the response and can see that enabled is false. The environments, development and production, are not listed so it's unclear what the value actually means. Why isn't the environments included in the response?

If I make a call to my.server.host/api/admin/features I get a similar response with the same problem. No environments in the response and the wrong "enabled" value for the feature toggle.

If I make a call to my.server.host/api/admin/projects/default I can see a correct response with both environments and correct values for enabled.

In summary, the Node.js SDK and two out of three APIs appear to return the wrong value for the feature toggle. Why is that? It appears to be some bug in the two APIs and the Node.js SDK appears to be using the failing APIs instead of the one that actually works.

anogr:
I found the error. I must use an API key that's targeted for clients with a specific environment, and I need to provide the environment as a query parameter for the API calls to respond with correct values.

Well, if you read this, a little more helpful responses or error messages for stupid people like me wouldn't be a bad thing.

thomasheartman:
Hey, @anogr! Thanks for the issue and sorry you found it confusing. I think you've partially found the solution, but not quite:

When you use an [client API token](https://docs.getunleash.io/reference/api-tokens-and-client-keys#client-tokens), the environment is encoded into the token itself, and you don't need to specify it as a query parameter. The `environment` config option is **not used to set the environment**, but only left in for compatibility with older versions  of Unleash. 

So if you [create an API token (server-side)](https://docs.getunleash.io/how-to/how-to-create-api-tokens) for the desired environment, that should be enough.

That said, looking through the node docs, I see that it's a bit out of date. I'll see if I can't make some small changes there to make it more obvious. Thanks for letting us know 😄

kjvalencik:
[...]

PR #2799, opened 2023-01-03
docs: clarify node sdk instructions wrt envs and tokens

thomasheartman:
## What

This change does two things:

It **removes the `environment` option** from the initialization examples. This option is deprecated and only causes confusion. There's no good reason to have it lying around in the examples.

It also **updates the explanation of the code samples**, telling you that you need to generate an API key and linking you to the reference doc and how to guide.

Relates to: #2782.

### DIFF
--- a/website/docs/reference/sdks/node.md
+++ b/website/docs/reference/sdks/node.md
@@ -37,7 +37,6 @@ const unleash = require('unleash-client');
 unleash.initialize({
   url: 'https://YOUR-API-URL',
   appName: 'my-node-name',
-  environment: process.env.APP_ENV,
   customHeaders: { Authorization: 'SOME-SECRET' },
 });
 ```
@@ -51,7 +50,6 @@ const { startUnleash } = require('unleash-client');
 const unleash = await startUnleash({
   url: 'https://YOUR-API-URL',
   appName: 'my-node-name',
-  environment: process.env.APP_ENV,
   customHeaders: { Authorization: 'SOME-SECRET' },
 });
 ```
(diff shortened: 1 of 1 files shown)


kwasniew (approved):
Good change

website/docs/reference/sdks/node.md
```diff
@@ -51,15 +50,14 @@ const { startUnleash } = require('unleash-client');
...
-  environment: process.env.APP_ENV,
   customHeaders: { Authorization: 'SOME-SECRET' },
 });
 ```
 
   </TabItem>
 </Tabs>
 
-The example code above will initialize the client SDK, and connect to the Unleash-hosted demo instance. It also uses the API token for the demo instance. You should change the URL and the Authorization header (API token) with the correct values for your instance, which you may locate under “Instance admin” in the menu.
+The example code above will initialize the client SDK, and try to the Unleash instance you point it to. You will need to use your own Unleash instance URL and a [server-side API token](../api-tokens-and-client-keys.mdx#client-tokens) to connect. For steps on how to create an API token, refer to the [_how to create API tokens_](../../how-to/how-to-create-api-tokens.mdx) guide.
```
nunogois, line 60: Seems like a typo, should it be `try to connect to` instead?

thomasheartman: Yup! I caught that right after merging, so quick-added it in a commit d2245583eca4c76dc6f144a5f498955fd8c924ec 😅  Good catch 🙌🏼

PR #2801, opened 2023-01-03
docs: Try to find broken links

gastonfournier:
## About the changes
This is a test for detecting broken links


gastonfournier:
In Docusaurus there are relative paths and absolute ones, some containing .md are then transformed and we also have made up urls that don't correlate with a file in the system because they're defined in sidebars.js 

Overall, the task of testing these files outside Docusaurus is not worthy

PR #2805, opened 2023-01-03
Makestyles refactor #7/1

sjaanus:
This is half of refactoring makestyles part 7


frontend/src/component/segments/SegmentFormStepTwo.tsx
```diff
@@ -38,6 +37,67 @@ interface ISegmentFormPartTwoProps {
...
+const StyledNoConstraintText = styled('div')(({ theme }) => ({
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center',
+    marginTop: '6rem',
+}));
+
+const StyledSubtitle = styled('p')(({ theme }) => ({
+    fontSize: theme.fontSizes.bodySize,
+    color: theme.palette.grey[600],
```
FredrikOseberg, line 73: Should be using a semantic color from theme instead of leaking color specifics into the component.

frontend/src/component/segments/SegmentFormStepTwo.tsx
```diff
@@ -38,6 +37,67 @@ interface ISegmentFormPartTwoProps {
...
+    maxWidth: 515,
+    marginBottom: 20,
+    wordBreak: 'break-word',
+    whiteSpace: 'normal',
+    textAlign: 'center',
+}));
+
+const StyledConstraintContainer = styled('div')(({ theme }) => ({
+    marginBlock: '2rem',
```
FredrikOseberg, line 82: Could be theme.spacing ?

frontend/src/component/segments/SegmentFormStepTwo.tsx
```diff
@@ -38,6 +37,67 @@ interface ISegmentFormPartTwoProps {
...
+}));
+
+const StyledButtonContainer = styled('div')(({ theme }) => ({
+    marginTop: 'auto',
+    display: 'flex',
+    justifyContent: 'flex-end',
+    borderTop: `1px solid ${theme.palette.grey[300]}`,
```
FredrikOseberg, line 89: Should use semantic theme color

FredrikOseberg (commented):
Just made a few general comments, we have the opportunity to make it a lot more ergonomic by switching out all the old tech debt such as leaking color details out to the components, and using theme.spacing instead of rem units. It's a bit more work, but it's something we'd have to tackle when doing the dark mode anyway. 

I think it makes sense to do it now while we are refactoring? WDYT?

frontend/src/component/segments/SegmentFormStepTwo.tsx
sjaanus: Yes, fixed

sjaanus:
> Just made a few general comments, we have the opportunity to make it a lot more ergonomic by switching out all the old tech debt such as leaking color details out to the components, and using theme.spacing instead of rem units. It's a bit more work, but it's something we'd have to tackle when doing the dark mode anyway.
> 
> I think it makes sense to do it now while we are refactoring? WDYT?

Good point, made all the needed changes.

FredrikOseberg (approved):
LGTM

PR #2731, opened 2022-12-23
feat: service accounts (BE)

nunogois:
https://linear.app/unleash/issue/2-537/exploration-back-end-for-service-accounts

Adds the concept of Service Accounts from a back-end perspective.

PR #2806, opened 2023-01-03
feat: enterprise service accounts (BE)

nunogois:
Supersedes https://github.com/Unleash/unleash/pull/2731

### DIFF
--- a/src/lib/db/user-store.ts
+++ b/src/lib/db/user-store.ts
@@ -100,7 +100,7 @@ class UserStore implements IUserStore {
     }
 
     buildSelectUser(q: IUserLookup): any {
-        const query = this.activeUsers();
+        const query = this.activeAll();
         if (q.id) {
             return query.where('id', q.id);
         }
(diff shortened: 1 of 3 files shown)


src/lib/db/user-store.ts
```diff
@@ -113,8 +113,17 @@ class UserStore implements IUserStore {
...
+            deleted_at: null,
+        });
+    }
+
     activeUsers(): any {
-        return this.db(TABLE).where('deleted_at', null);
+        return this.db(TABLE).where({
+            deleted_at: null,
+            is_service: false,
+        });
```
nunogois, line 126: Reasoning for these changes is: If we're adapting `activeUsers` to only handle real users (not service accounts), then `buildSelectUser` should be adapted to still query on all active accounts (both users and SAs). Since the fields queried by `buildSelectUser` are unique this should be fine, and this also means we should be able to safely check if a username/email is unique across both users and SAs.

gastonfournier: Maybe we don't need to make this distinction in OSS since there should be zero SA, but I do see a problem while using the same `users` table from enterprise. You've added a new `service-account-store` in the other PR, so I'm wondering: is it worth having users and service accounts in the same table?

I'm thinking this could require additional work on validating permissions or other functionality that relies on the user's table...

PR #2807, opened 2023-01-03
Fix(typo): Fix "login"/"log in", "logout"/"log out" issues

thomasheartman:
## What

This change updates some places in the docs where we use the terms "login" and "logout" incorrectly.

A "login" is a noun, typically referring to the set of credentials you need to _log in_ to a service. The verb form, the act of signing in, is written in two words: to "log in".

A similar logic applies to "logout" and "log out", although I don't find the term "logout" in my dictionary. However, I think it makes sense to talk about "logout requests" (and I see references to logout in other services and documentation), so I'm happy to use that as a noun. Regardless, the act of logging out is to "log out".

### DIFF
--- a/frontend/src/component/user/Login/Login.tsx
+++ b/frontend/src/component/user/Login/Login.tsx
@@ -48,7 +48,7 @@ const Login = () => {
                     condition={authDetails?.type !== DEMO_TYPE}
                     show={
                         <h2 className={styles.title}>
-                            Login to continue the great work
+                            Log in to continue the great work
                         </h2>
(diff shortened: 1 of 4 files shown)


kwasniew (approved):
We have the same mistake in 2 github workflows "Login to docker hub"

thomasheartman:
@kwasniew Good catch. Feel free to open a PR for those and add me as a reviewer 😄  I'm less concerned about those, though, because they're not client-facing in the same way 💁🏼

PR #2808, opened 2023-01-03
refactor: styled component in header

kwasniew:
Proposal: Example of replacing makeStyles with themable functions and shared CSS props literals. Check themeStyles.ts.


frontend/src/component/menu/Header/Header.tsx
```diff
@@ -20,21 +22,90 @@ import { ReactComponent as UnleashLogoWhite } from 'assets/img/logoWithWhiteText
...
+    color: 'inherit',
+    cursor: 'pointer',
+}));
+
+const styledIconProps = (theme: Theme) => ({
+    color: theme.palette.neutral.main,
+});
+
+const StyledLink = styled(Link)(({ theme }) => ({
+    ...focusable(theme),
```
kwasniew, line 101: themable function usage

frontend/src/component/menu/Header/Header.tsx
```diff
@@ -196,7 +244,7 @@ const Header: VFC = () => {
                                 rel="noopener noreferrer"
                                 size="large"
                                 disableRipple
-                                className={themeStyles.focusable}
+                                sx={focusable}
```
kwasniew, line 247: one off themable function usage in sx

frontend/src/component/menu/Header/Header.tsx
```diff
@@ -20,21 +22,90 @@ import { ReactComponent as UnleashLogoWhite } from 'assets/img/logoWithWhiteText
...
+        alignItems: 'center',
+    },
+}));
+
+const StyledAdvancedNavButton = styled('button')(({ theme }) => ({
+    border: 'none',
+    background: 'transparent',
+    height: '100%',
+    display: 'flex',
+    fontSize: '1rem',
```
sjaanus, line 89: Probably better to use theme here.

frontend/src/themes/themeStyles.ts
```diff
@@ -1,4 +1,20 @@
...
+        outlineColor: theme.palette.primary.main,
+    },
+});
+
+export const flexRow = {
+    display: 'flex',
+    alignItems: 'center',
+};
 
 export const useThemeStyles = makeStyles()(theme => ({
```
kwasniew, line 19: I leave the old code for now so that it can be migrated in other batches

FredrikOseberg (approved):
Lovely, how do we document this and spread the knowledge in the team?

frontend/src/themes/themeStyles.ts
sjaanus: Can we mark it as deprecated or similar?

sjaanus (approved):

frontend/src/component/menu/Header/Header.tsx
kwasniew: I don't think we have anything in our theme that only specifies a font size without some extra CSS properties. That's why I skipped fontSize in previous PRs too. I also want to avoid theme.spacing here because it's for spacing not font sizes.

sjaanus: Could it be theme.fontsizes.bodySize?

kwasniew:
@FredrikOseberg what about adding a note about shared CSS literals and functions to preferred-styling-method? WDYT?

FredrikOseberg:
> 

Good idea. Let's also write it up in slack and share the pattern?

frontend/src/component/menu/Header/Header.tsx
kwasniew: Oh I didn't see this one, thanks!

frontend/src/themes/themeStyles.ts
kwasniew: sure, marking it as deprecated with a comment what to do instead

frontend/src/component/menu/Header/Header.tsx
nunogois: Can be simplified as `const StyledLink = styled(Link)(({ theme }) => focusable(theme));`

PR #2809, opened 2023-01-03
docs: enable "remote content" plugin, add Rust SDK readme

thomasheartman:
## What

This PR does a few connected things:

1. It adds the ["docusaurus-plugin-remote-content" package](https://github.com/rdilweb/docusaurus-plugin-remote-content).

2. It adds configuration to make it work with Readmes found on GitHub.

3. It adds the Rust SDK's readme (replacing the link we used to have) as a proof of concept on how to do it.

## Why

With documentation split between GitHub readmes and the official docs, it's hard to keep everything up to date and in sync. It's also quite confusing that some information is only available in some places, but not in others. 

We've talked about auto-including readmes from GitHub for a while, so here's a proof of concept (finally) 🥳 

The intention is to get this merged and then to migrate the other SDK docs one by one, ensuring that everything in the documentation is also in the readme (so that no info is lost). 

### Generation directory

The current generation method generates the files into `/reference/sdks/<sdk name>`. I think this works for now, but it means it adds auto-generated files into a directory that you can't ignore (at least not yet). 

We could instead generate them into `/generated/sdks` and update the slugs so that they still match the expected pattern. 

However, this would make the sidebar a little harder to work with (for now). That said, there may be ways around it. It's worth exploring.

### Generation method

By default, this plugin will generate files whenever you build. That (probably) means that you need an internet connection _and_ that you'll end up with a bunch of untracked files.

An option is to only generate the files "manually" and commit them to the repo. That would allow you to build the project without an internet connection and would also remove the need for ignoring the files. We could automate the generation if we wanted to. 

## Preview / Screenies

Visit [/reference/sdks/rust](https://unleash-docs-git-docs-include-sdk-readmes-unleash-team.vercel.app/reference/sdks/rust) in the preview to see what it looks like live. 

image

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -1,3 +1,27 @@
+// all SDK repos and what they map to for the sidebar.
+//
+// Each key should return an object that contains *at least* the
+// `sidebarSdkName` property. This is the name that is placed before "SDK" in
+// the sidebar. An object can also optionally define a `slugName`. This will be
+// used to construct the slug. If no "slugName" is defined, the `sidebarSdkName`
+// should be used to create the slug.
+const readmeRepos = {
+    'unleash-client-rust': {
+        sidebarSdkName: 'Rust',
+    },
+    // as an example of when you'd use both
+    // 'unleash-android-proxy-sdk': {
+    //     sidebarSdkName: 'Android',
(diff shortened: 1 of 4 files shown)


thomasheartman:
Closing in favor of the merged #2816 which builds off of this PR.

PR #2816, opened 2023-01-04
docs: Use Go readme

thomasheartman:
This PR changes the docs generation to use the Go SDK's GitHub readme for the SDK docs instead of a separate document.

## What

The changes in this PR are:

-   Delete the existing Go SDK documentation. All the content in this guide already exists in the Go readme.
-   Add the Go SDK to the list of auto-generated readme docs
-   Move the readme-related code into a separate module, `readme-fns.js` (I'm not bullish about the file name: we can change it if you have suggestions)
-   Add a note to the top of all generated readmes saying you'll need an API url and an API token. The note also links you to the relevant reference and how-to docs.

## Why

Having two different bits of documentation for the same SDK is troublesome. By only having the data in one place, we can avoid it going out of sync and getting stale.


FredrikOseberg (approved):
LGTM

PR #2824, opened 2023-01-04
fix: export features with variants event when feature is disabled

gastonfournier:
## About the changes
When exporting features one is normally also interested in disabled features, so they are also included in the export file, while the variants are not. I do not see a good reason for that, so this PR removes the check and exports the variants then as well.
I could also add an option as well, but as long as there is no good reason for ignoring the variants I would just export them with the features.

Tests for #2719

Closes #2719


matlockx:
👍

sighphyre (approved):
LG

PR #2829, opened 2023-01-05
Scheduler abstraction

kwasniew:
Introducing a scheduler service abstraction to separate scheduling capability from individual services.

Design constraints: there should be one setInterval in the entire codebase. Applications services should not know anything about them being scheduled.

Chosen scheduling style: orchestration - a scheduler service calling other service methods directly. In future we may consider choreography/event emitting style but it will require extra abstraction.


sjaanus (approved):
LGTM

src/lib/services/scheduler-service.ts
```diff
@@ -0,0 +1,11 @@
+export default class SchedulerService {
+    private intervalIds: NodeJS.Timer[] = [];
+
+    schedule(scheduledFunction: () => void, timeMs: number): void {
+        this.intervalIds.push(setInterval(scheduledFunction, timeMs).unref());
```
ivarconr, line 5: In the future I can imagine we will improve this implementation to not create a new interval per background task, but rather run at a regular interval and pick tasks in that interval and run them in sequence. I think that would use database connections more efficiently.

src/lib/services/api-token-service.ts
```diff
@@ -88,27 +79,23 @@ export class ApiTokenService {
         }
     }
 
-    async fetchActiveTokens(): Promise<void> {
+    fetchActiveTokens = async (): Promise<void> => {
         try {
```
ivarconr, line 83: I suggest that the scheduler will take the responsibility to run background task in a try/catch. I cannot think of any scenarios where we would allow a background job to crash Unleash with an unhanded exceptions.

src/lib/services/index.ts
```diff
@@ -137,6 +139,17 @@ export const createServices = (
...
+    schedulerService.schedule(
+        apiTokenService.fetchActiveTokens,
+        minutesToMilliseconds(1),
+    );
+
+    schedulerService.schedule(
+        apiTokenService.updateLastSeen,
+        minutesToMilliseconds(3),
+    );
+
```
ivarconr, line 152: when we move to package by feature I would think we could have a pattern where we have a separate file `scheduled-jobs.ts` to deal with the registering these things.

src/lib/services/api-token-service.ts
```diff
@@ -287,10 +274,5 @@ export class ApiTokenService {
         }
     }
 
-    destroy(): void {
-        clearInterval(this.timer);
-        clearTimeout(this.seenTimer);
-        this.timer = null;
-        this.seenTimer = null;
-    }
+    destroy(): void {}
```
ivarconr, line 277: destroy method could be removed. Does not make sense anymore.

src/lib/services/scheduler-service.ts
kwasniew: Good idea. Once we have more code schduled externally some task collapsing would work really nice.

src/lib/services/api-token-service.ts
kwasniew: Good idea. I will handle both sync and async handlers. Also, I'd log those to the logger.

src/lib/services/api-token-service.ts
```diff
@@ -88,27 +79,23 @@ export class ApiTokenService {
...
             this.activeTokens = await this.getAllActiveTokens();
         } finally {
             // eslint-disable-next-line no-unsafe-finally
             return;
         }
-    }
+    };
 
-    async updateLastSeen(): Promise<void> {
+    // arrow function to prevent this reference problems in the schedulerService
```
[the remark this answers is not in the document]
kwasniew: Yeah, with "this" in there's no elegant solution just tradeoffs :) I'll play around with an API that gives more freedom to function implementers.

src/lib/services/index.ts
kwasniew: definitely, that's the plan. I can't wait to split everything by feature :)

ivarconr (approved):

PR #2831, opened 2023-01-05
Allow hiding environments from the feature overview screen fix

sjaanus:
This feature was already implemented, but this PR adds couple of fixes to it.

1. Currently it was only visible, when **variantsPerEnvironment** flag was enabled. Now it is visible always.
2. Now the hide button is more centered on the line.


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewSidePanel/FeatureOverviewSidePanelEnvironmentSwitches/FeatureOverviewSidePanelEnvironmentSwitch/FeatureOverviewSidePanelEnvironmentHider.tsx
```diff
@@ -6,6 +6,7 @@ import { ConditionallyRender } from 'component/common/ConditionallyRender/Condit
 const Visible = styled(Visibility)(({ theme }) => ({
     cursor: 'pointer',
     marginLeft: 'auto',
+    marginTop: theme.spacing(0.5),
     color: theme.palette.grey[700],
```
FredrikOseberg, line 10: Can we change this also, even though it wasn't changed in this PR?

FredrikOseberg (approved):
LGTM. Could be nice to fix the color in the Visible component also.

PR #2832, opened 2023-01-05
Remove E flag

sjaanus:
Remove environments flag, because it is not needed anymore. Everyone can use environments.


kwasniew (approved):
We should merge this one because I can't add new envs in sandbox w/o it

PR #2833, opened 2023-01-05
fix: when health is 0 we should not show 100

kwasniew:
When health report is 0% we should report 0%. Previously we were reporting 100%.

Screenshot 2023-01-05 at 14 48 04

### DIFF
--- a/src/lib/db/project-store.ts
+++ b/src/lib/db/project-store.ts
@@ -419,7 +419,7 @@ class ProjectStore implements IProjectStore {
             name: row.name,
             description: row.description,
             createdAt: row.created_at,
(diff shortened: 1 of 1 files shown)


src/lib/db/project-store.ts
```diff
@@ -419,7 +419,7 @@ class ProjectStore implements IProjectStore {
             name: row.name,
             description: row.description,
             createdAt: row.created_at,
-            health: row.health || 100,
+            health: row.health ?? 100,
```
ivarconr, line 422: The danger of trueish

ivarconr (approved):

PR #2842, opened 2023-01-06
Refactor/lazy load

FredrikOseberg:
## About the changes

Currently our bundle size is clocking in at: 1,798.28 kB │ gzip: 558.42 kB
After the changes: 1,299.32 kB │ gzip: 403.26 kB

This PR aims to reduce the bundle size by leveraging [code splitting](https://reactjs.org/docs/code-splitting.html), which will break specific parts of the application up into separate bundles that are loaded on demand when the user navigates to the page. This is the first part of this effort.

In order to effectively code the larger areas of our codebase, I did some refactoring to the admin section. It groups the routes under a single component that can be lazy loaded, which also allows us to remove the admin menu code duplication and prepares the ground for a more uniform admin menu section in the future.


frontend/src/interfaces/route.ts
```diff
@@ -16,6 +16,14 @@ export interface IRoute {
     isStandalone?: boolean;
 }
 
+export interface ICondensedRoute {
```
kwasniew, line 19: I don't understand the meaning of condensed here

frontend/src/component/admin/LazyAdminExport.tsx
```diff
@@ -0,0 +1,7 @@
+// Lazy loading only supports default export. We have an ADR
```
kwasniew, line 1: We can do it without default export: https://dev.to/iamandrewluca/react-lazy-without-default-export-4b65

frontend/src/interfaces/route.ts
kwasniew, line 19: What is the meaning of "condensed" in this context?

kwasniew (approved):
Non blocking approval with minor comments

frontend/src/interfaces/route.ts
FredrikOseberg: Essentially it's just a navigation route. It doesn't have all the properties of IRoute, and it does not have a component because it's not being used to render a route, just populate a navigation menu. It could be clearer: `INavigationMenuItem` ?

frontend/src/component/admin/LazyAdminExport.tsx
FredrikOseberg: I tried this approach first, but couldn't get it to work. Let me have another look before merging.

frontend/src/component/admin/groups/RemoveGroup/RemoveGroup.tsx
```diff
@@ -19,15 +21,14 @@ export const RemoveGroup: FC<IRemoveGroupProps> = ({
...
-    const { refetchGroups } = useGroups();
     const { removeGroup } = useGroupApi();
     const { setToastData, setToastApiError } = useToast();
     const navigate = useNavigate();
 
     const onRemoveClick = async () => {
         try {
             await removeGroup(group.id);
-            refetchGroups();
+            mutate(formatApiPath('api/admin/groups'));
```
FredrikOseberg, line 31: This is super ugly, but it solves an issue we found related to deleting groups. This is currently failing our e2e tests and logs an uncaught error in production: 

Skjermbilde 2023-01-11 kl  15 59 01

Digging locally I found that this happens because of useSWR trying run this function: https://github.com/vercel/swr/blob/main/core/use-swr.ts#L287

My theory here is that using useGroups() and exposing the data property from useSWR sets up a subscription. UseSWR needs to keep an internal cache of components to re-render once the data is refetched. My theory is that it does this by tracking the invocations of useSWR, and re-renders these components when the data changes. In this instance, the hook is used in a modal that is removed once the refetch is complete. Using the global mutate avoids setting up a subscription to re-render this function.

kwasniew: isn't there some equivalent of useEffect unsubscribe function to clean this up?

PR #2846, opened 2023-01-09
feat: RBAC read params from body

gastonfournier:
## About the changes
This is a follow-up on #1953

This implementation generalizes how we fetch some standard parameters from the query parameters or request body. 

## Discussion points
Unfortunately, we have not used standard names for our APIs and one example is our `projectId` (in some cases we just used `project`). Ideally, we're only using one way of sending these parameters either `projectId` or `project` (same applies to `environment` vs `environmentId`). 

If both parameters are present, due to historical reasons, we'll give precedence to:
- `projectId` over `project`
- `environment` over `environmentId` 

In the presence of both query parameters and body, we'll give precedence to query parameters also for historical reasons.


sighphyre (approved):
This looks a bit hacky to me (no fault of the code, just that the problem is ugly to solve without hacks). I'm open to this but perhaps, getting a standard in place for v5 of Unleash could help us remove these kind of small things at a later stage

PR #2848, opened 2023-01-09
Feat sa table info

nunogois:
https://linear.app/unleash/issue/2-543/show-relevant-information-on-the-service-accounts-table

Shows relevant information on the table, like total PATs and the last time a service account was active based on latest seen PAT for that account. Adapts to the latest related PR on enterprise.

image


gastonfournier (approved):
LG!

PR #2851, opened 2023-01-09
fix: disable autoResetHiddenColumns when useConditionallyHiddenColumns

nunogois:
https://linear.app/unleash/issue/2-563/fix-issue-with-useconditionallyhiddencolumns-and-react-table

It seems like we should add `autoResetHiddenColumns: false` to `useTable` whenever we use `useConditionallyHiddenColumns`.

Basically the thought is that, if we're controlling column visibility in our own way, we should not want other things to change that state unpredictably, otherwise this may make React go _brrrrrr_. And it can be very hard to pinpoint what exactly may be causing React to go _brrrrrr_.

image

First detected this issue apparently randomly while developing the new SA table. Around 10-20 page refreshes would eventually trigger it. Was not easy to find, but hopefully this fixes it permanently. At least I haven't been able to reproduce it since. Maybe someone has a better idea of where the issue could be or if this is a pretty good guess. Doesn't seem like this change hurts us anyways.

I love React, `useEffect` and these very to-the-point error messages. Very fun and productive.

Reference: https://react-table-v7.tanstack.com/docs/api/useTable

### DIFF
--- a/frontend/src/component/admin/apiToken/ApiTokenTable/ApiTokenTable.tsx
+++ b/frontend/src/component/admin/apiToken/ApiTokenTable/ApiTokenTable.tsx
@@ -52,6 +52,7 @@ export const ApiTokenTable = () => {
             data: tokens as any,
             initialState,
             sortTypes,
+            autoResetHiddenColumns: false,
             disableSortRemove: true,
         },
         useGlobalFilter,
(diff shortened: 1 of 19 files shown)


gastonfournier:
Do we get client error metrics? Maybe Plausible can help to track how many times the UI crashes like this?

gastonfournier (approved):
I wonder if we should have some global/central configuration for default `TableOptions` so we can do something like: 
```javascript
const { headerGroups, rows, prepareRow, setHiddenColumns } = useTable(
        {
            ...defaultTableOptions,
            columns: columns as any[],
            data: users as any[],
            sortTypes,
        }
);
```

I don't know if it makes sense, but it would make this change easier and we'll be able to modify tables more consistently or provide some sensible defaults.

Regardless this is still a good PR IMO

sighphyre (approved):
> I love React, useEffect and these very to-the-point error messages. Very fun and productive.

There's so much salt in this PR, I got heart disease reading this line. Died laughing

Nice change though, looks like it needs to happen. 

> Do we get client error metrics? Maybe Plausible can help to track how many times the UI crashes like this?

@gastonfournier Might be nice to know about all crashes, be cool if we could find these things faster

PR #2811, opened 2023-01-03
Refactor/make styles batch 6 part 2

FredrikOseberg:
Adds another batch of refactored components

PR #2852, opened 2023-01-09
fix: PageContent bodyContainer overflowX

nunogois:
https://linear.app/unleash/issue/1-546/project-overview-table-is-missing-overflow-x-auto

Fixes an issue where the PageContent body doesn't scroll horizontally when needed.

This regression seems to have been included with https://github.com/Unleash/unleash/pull/2811, more specifically:
https://github.com/Unleash/unleash/pull/2811/files#diff-a1c10fdcd539d988f74d4c3742cc6b8844152999a6c7e509ed8b1765c604524fL491

### DIFF
--- a/frontend/src/component/common/PageContent/PageContent.styles.ts
+++ b/frontend/src/component/common/PageContent/PageContent.styles.ts
@@ -9,9 +9,7 @@ export const useStyles = makeStyles()(theme => ({
         [theme.breakpoints.down('md')]: {
             padding: theme.spacing(2),
         },
-        [theme.breakpoints.down('sm')]: {
-            overflowX: 'auto',
-        },
+        overflowX: 'auto',
     },
     paddingDisabled: {
         padding: '0',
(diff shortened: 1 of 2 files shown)


frontend/src/component/common/PageContent/PageContent.styles.ts
```diff
@@ -9,9 +9,7 @@ export const useStyles = makeStyles()(theme => ({
         [theme.breakpoints.down('md')]: {
             padding: theme.spacing(2),
         },
-        [theme.breakpoints.down('sm')]: {
-            overflowX: 'auto',
-        },
+        overflowX: 'auto',
```
nunogois, line 12: Seems like `PageContent` hasn't been migrated to StyledComponents yet. Having `overflowX: 'auto'` applied every time instead of only on the breakpoint doesn't seem to hurt and fixes the issue.

FredrikOseberg (approved):
Perhaps the behavior needs a unit test?

PR #2853, opened 2023-01-09
feat: move createdAt col on users

nunogois:
Small UI/UX change. 

After discussing with @NicolaeUnleash we decided that moving this column in the users table over to the right next to "last login" would improve readability and consistency.

image


gastonfournier (approved):
G!

PR #2855, opened 2023-01-09
docs: use php readme for doc generation

thomasheartman:


thomasheartman:
Closing in favor of #2858, which includes these changes and the rest of the server-side sdks.

Issue #1386, opened 2022-02-23
The documentation for the node client-sdk doesn't mention events


SimonSimCity:
The documentation of the client-sdk for node (https://docs.getunleash.io/sdks/node_sdk) does not mention any of the available events (https://github.com/Unleash/unleash-client-node#events). Maybe its best to just link to the readme of the repository instead of having two docs you need to keep in sync.

ivarconr:
Thanks for pointing this out. None of the SDK tutorials in the docs currently explains all the capabilities of the SDKs. We plan to move more of the individual SDK docs in the the combined documentation. Stay tuned!

thomasheartman:
@SimonSimCity Yes, you're right! Thanks for raising this 😄 What Ivar says is correct: we've got plans to start integrating the documentation from the SDK repos into docs.getunleash.io, so we'll only need to keep one set of docs up to date. It's been researched and we've found a way forward, so it's on the way ✌🏼

ivarconr:
any updates on this @thomasheartman ?

thomasheartman:
Not yet, I'm afraid. With the OpenAPI work and other things going on, we haven't gotten around to actually implementing this, though.

PR #2858, opened 2023-01-09
docs: auto-generate remaining server-side SDK docs

thomasheartman:
This PR builds on the preceding doc auto-generation PRs and generates documentation for the remaining server-side SDKs.

## Why

Refer to https://github.com/Unleash/unleash/pull/2809 for more context about generating SDK docs.

## What

-   Adds generation for the remaining server-side SDKs
-   Moves generated docs from the `/reference/sdks` directory to `/generated` directory.
-   Makes sure that the URLs do not change because of the move by using the `slug` frontmatter property.
-   replaces relative github links in the markdown documents so that they become absolute github links. (refer to the next section)
-   Updates some image styling so that it doesn't apply to readme badges (we don't need them using `display: block`)

### On link replacing:

This PR adds handling of links in the generated documentation. Specifically, it changes links in one case:

    Relative links to github. Links to code and other files in the repository. These are prefixed with the repository's URL.

While this should work in most cases, it will fail in cases where the links to the files are not on the repository's primary branch. (typically main, but could also be "v3", for instance). In these cases, the links will get a double branch in the URL and will fail. However, I see no easy way around this (though suggestions are definitely accepted!), and think it's a fair tradeoff. It takes the links from "definitely failing" to "will work in the vast majority of cases".

Note: I originally also wanted to handle the case where the link is an absolute link to docs.getunleash.io. We could turn these into relative urls to avoid full page reloads and enjoy a smoother experience. However, the client-side redirects don't work correctly if the relative URL goes to a redirect page, so you end up with a 404 page. As such, I think it's better to leave the links as absolute for now.


website/docusaurus.config.js
```diff
@@ -566,7 +566,7 @@ module.exports = {
                 // more info at https://github.com/rdilweb/docusaurus-plugin-remote-content#options
                 name: 'content-sdks',
                 sourceBaseUrl: 'https://raw.githubusercontent.com/Unleash/', // gets prepended to all of the documents when fetching
-                outDir: 'docs/reference/sdks', // the base directory to output to.
+                outDir: 'docs/generated/sdks', // the base directory to output to.
```
thomasheartman, line 569: We _might_ consider changing this line to:

```suggestion
                outDir: 'docs/generated/reference/sdks', // the base directory to output to.
```

But I also think that's something we can do later if we need to. No need create an arbitrarily nested folder structure if we're not gonna need it.

ivarconr:
fixes: #1386

PR #2861, opened 2023-01-10
fix: SA duplicate autoResetHiddenColumns

nunogois:
Seems like merging both https://github.com/Unleash/unleash/pull/2851 and https://github.com/Unleash/unleash/pull/2848 made us end up with a duplicate `autoResetHiddenColumns` option in the `ServiceAccountsTable` table component.


sighphyre (approved):
G

gardleopard (approved):
LG

PR #2872, opened 2023-01-11
Update change-requests.md

JarrodSC:
With the release of 4.19, the info box at the top was outdated. I've removed it.

## About the changes

Closes #

### DIFF
--- a/website/docs/reference/change-requests.md
+++ b/website/docs/reference/change-requests.md
@@ -2,14 +2,6 @@
 title: Change requests
 ---
 
-:::info
-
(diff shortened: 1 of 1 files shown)


website/docs/reference/change-requests.md
```diff
@@ -2,14 +2,6 @@
 title: Change requests
 ---
```
thomasheartman, line 4: How about this?

```suggestion

:::info Availability

The change requests feature is an enterprise-only feature that was introduced in **Unleash 4.19.0**.

:::
```

thomasheartman (approved):

PR #2877, opened 2023-01-11
Export with strategies

sjaanus:
1. Added OpenAPI schemas
2. Added features strategies


src/test/e2e/api/admin/export-import.e2e.test.ts
```diff
@@ -71,11 +98,34 @@ test('exports features', async () => {
...
         features: [
             {
                 name: 'first_feature',
             },
         ],
+        featureStrategies: [resultStrategy],
+    });
+});
+
+test('returns all features, when no feature was defined', async () => {
```
kwasniew, line 112: suggested rename => returns all features by default

I'd also consider: features not specified at all or set to undefined

kwasniew (approved):
Nice

PR #2879, opened 2023-01-11
fix: prevent margin collapsing

kwasniew:
After styled component migration we broke the header. Grid element margin was going outside the main parent element and entering the area of the header (margin collapsing). Added display: flex to main element to prevent child elements escaping with their margins.

Fixed:
Screenshot 2023-01-11 at 10 40 40

Before:
Screenshot 2023-01-11 at 10 40 35

Solution: Added flex to parent of the element that was collapsing its margin.

### DIFF
--- a/frontend/src/component/layout/MainLayout/MainLayout.tsx
+++ b/frontend/src/component/layout/MainLayout/MainLayout.tsx
@@ -25,7 +25,7 @@ const MainLayoutContainer = styled(Grid)(() => ({
 
 const MainLayoutContentWrapper = styled('main')(({ theme }) => ({
     margin: theme.spacing(0, 'auto'),
-    display: 'flex',
+    display: 'flex', // prevent margin collapsing
     flex: 1,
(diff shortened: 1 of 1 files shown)


thomasheartman (approved):
Nice one! Seems like a reasonable change 😄  Are there any other display modes that would take care of the issue too? I don't think it's gonna be a performance issue, but seeing code where display is 'flex' or 'grid' without any flexing or gridding happening is always a little confusing at first 💁🏼

kwasniew:
There's a few ways to prevent margins collapsing: https://stackoverflow.com/questions/13573653/css-margin-terror-margin-adds-space-outside-parent-element
I've tried all of the suggested solutions and display: flex was the least hacky one and didn't break other styles as e.g. inline-block.

kwasniew:
@thomasheartman added a comment explaining why we have this display flex. That's the best I know for weird CSS corner cases :)

PR #2886, opened 2023-01-12
test: fix test runs to 1

kwasniew:
## About the changes

property based tests will run only once because otherwise e2e is flaky


thomasheartman (approved):
Nice change! This is a simplification that should relieve the flakiness for sure. Obviously, it makes the test less thorough, but if they stop breaking, then that's a net win 🥳

PR #2887, opened 2023-01-12
fix: improve hidden envs UI/UX

nunogois:
Some small UI/UX fixes related to user-scoped hidden environments in the feature overview after aligning with @NicolaeUnleash 

image


gastonfournier (approved):
LG!

PR #2889, opened 2023-01-12
fix: improve breadcrumb spacing

nunogois:
https://linear.app/unleash/issue/2-575/improve-the-breadcrumb-spacing

image

image

### DIFF
--- a/frontend/src/component/common/BreadcrumbNav/BreadcrumbNav.tsx
+++ b/frontend/src/component/common/BreadcrumbNav/BreadcrumbNav.tsx
@@ -6,9 +6,9 @@ import { useContext } from 'react';
 import StringTruncator from '../StringTruncator/StringTruncator';
 import { styled } from '@mui/material';
 
-const StyledBreadCrumbs = styled(Breadcrumbs)(({ theme }) => ({
-    position: 'absolute',
-    top: theme.spacing(0.25),
+const StyledBreadcrumbContainer = styled('div')(({ theme }) => ({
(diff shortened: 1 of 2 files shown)


frontend/src/component/common/BreadcrumbNav/BreadcrumbNav.tsx
```diff
@@ -50,55 +50,57 @@ const BreadcrumbNav = () => {
...
+
+                                    let link = '/';
+
+                                    paths.forEach((path, i) => {
+                                        if (i !== index && i < index) {
+                                            link += path + '/';
+                                        } else if (i === index) {
+                                            link += path;
+                                        }
+                                    });
```
gastonfournier, line 86: Just for fun, this is a for-loop (`paths.map`) inside a for-loop (`paths.forEach`), so it's `O(length(path)²)`. Perhaps, this can be done with a an accumulator (for the link) and make it linear.

But not a blocker for the PR as this is the same the previous code was doing

gastonfournier: Also, I assume lenght is usually no longer than 5

gastonfournier (approved):
LG!

frontend/src/component/common/BreadcrumbNav/BreadcrumbNav.tsx
nunogois: Outside of the scope of this PR, but I agree that we could improve this 👍

PR #2891, opened 2023-01-12
Feat better tooltip links

nunogois:
https://linear.app/unleash/issue/2-576/improve-how-text-that-has-tooltip-should-look-so-they-are-not

image

image

- Adapts the existing `HtmlTooltip` component to support setting `maxHeight` and `maxWidth` properties;
- Introduces a new common component: `TooltipLink`;
- Adapts SA (tokens), features (tags), variants (overrides, payloads) and project access (role and role description);
- Role description in project access now uses this component instead of the old jankier popover component;


chriswk (approved):
Looks sensible. With the explanation I even understood what's going on here :P

PR #2892, opened 2023-01-12
fix: stop forwarding visibilityOff prop in toggle

nunogois:
Tiny fix to stop forwarding the new `visibilityOff` prop in the `StyledVisibilityToggle`. 
Bug introduced with https://github.com/Unleash/unleash/pull/2887


nunogois:
https://github.com/Unleash/unleash/pull/2895#discussion_r1069045457 - @kwasniew was faster so I'm closing this.

PR #2893, opened 2023-01-12
fix: small breadcrumb adjustments

nunogois:
https://linear.app/unleash/issue/2-577/small-breadcrumb-adjustments

Includes small adjustments to the breadcrumb navigation component.

### Long chain path (ellipts if width > 200px)
image

### Long final path (no longer ellipts as long as it has enough remaining space)
image

### Long final path with resized window (still ellipts when needed)
image


chriswk (approved):
Cool. Looks good.

PR #2899, opened 2023-01-13
Fix/redirect woes

FredrikOseberg:
This PR fixes two problems: 

(1) The initial redirect put us into an infinite loop when redirecting, because trying to go back to the root would always trigger the initial redirect component. Throwing you back to project screen.

(2) Using UI config in the useLastViewedProject to get the basePath introduced a race condition where you needed data from the uiConfig in order to fetch the correct key from local storage. The fix here was to use the basePath coded into the HTML file, so we can synchronously retrieve the correct key at startup.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -14,6 +14,7 @@
     "start": "vite",
     "start:prod": "vite build && vite preview",
     "start:sandbox": "UNLEASH_API=https://sandbox.getunleash.io/ospro yarn run start",
+    "start:demo2": "UNLEASH_API=https://sandbox.getunleash.io/ UNLEASH_BASE_PATH=/demo2/ yarn run start",
     "start:enterprise": "UNLEASH_API=https://unleash.herokuapp.com yarn run start",
     "start:demo": "UNLEASH_BASE_PATH=/demo/ yarn start",
     "test": "tsc && vitest run",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -15,7 +15,7 @@
     "start:prod": "vite build && vite preview",
     "start:sandbox": "UNLEASH_API=https://sandbox.getunleash.io/ospro yarn run start",
     "start:demo2": "UNLEASH_API=https://sandbox.getunleash.io/ UNLEASH_BASE_PATH=/demo2/ yarn run start",
-    "start:enterprise": "UNLEASH_API=https://unleash.herokuapp.com yarn run start",
+    "start:enterprise": "UNLEASH_API=https://unleash.herokuapp.com VITE_TEST_REDIRECT=true yarn run start",
     "start:demo": "UNLEASH_BASE_PATH=/demo/ yarn start",
     "test": "tsc && vitest run",
     "test:snapshot": "yarn test -u",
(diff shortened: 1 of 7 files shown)


frontend/package.json
```diff
@@ -14,6 +14,7 @@
     "start": "vite",
     "start:prod": "vite build && vite preview",
     "start:sandbox": "UNLEASH_API=https://sandbox.getunleash.io/ospro yarn run start",
+    "start:demo2": "UNLEASH_API=https://sandbox.getunleash.io/ UNLEASH_BASE_PATH=/demo2/ yarn run start",
```
thomasheartman, line 17: Oh, nice 😄

frontend/src/component/menu/routes.ts
```diff
@@ -45,6 +45,13 @@ import { LazyProject } from 'component/project/Project/LazyProject';
 import { AdminRedirect } from 'component/admin/AdminRedirect';
 
 export const routes: IRoute[] = [
+    {
+        path: '/',
+        title: 'Unleash',
+        component: ProjectListNew,
+        type: 'protected',
+        menu: {},
+    },
```
thomasheartman, line 54: Interesting: was this not here before? Maybe because the redirect would take care of taking you to `/projects` or something?

thomasheartman (approved):
Yeeeees! This looks good to me 🚂  

My only (non-blocking) suggestion would be if we should move the redirect logic into the app component? It never renders anything, so should it be a component? On the other hand, it's an easy way to separate out the logic (especially because of hooks), so I think it's a viable use case. (And in fact, bailing from the app component if need to redirect might present its own problems, I suppose 💁🏼 ). Just for discussion; I'll leave it up to you, but I would be interested to hear your take on it 🤓

frontend/src/component/menu/routes.ts
FredrikOseberg: The redirect was set to render on the root path, which caused the loop to trigger. Because you can still navigate to the root by clicking the unleash logo, I set this as the default root path.

FredrikOseberg:
> Yeeeees! This looks good to me 🚂
> 
> My only (non-blocking) suggestion would be if we should move the redirect logic into the app component? It never renders anything, so should it be a component? On the other hand, it's an easy way to separate out the logic (especially because of hooks), so I think it's a viable use case. (And in fact, bailing from the app component if need to redirect might present its own problems, I suppose 💁🏼 ). Just for discussion; I'll leave it up to you, but I would be interested to hear your take on it 🤓

To me it makes sense to encapsulate the logic, if only for readability. It could perhaps be a hook, but I think this is good enough for now.

frontend/src/component/menu/routes.ts
thomasheartman: Thanks! Does that mean that the URL will be the root URL, but that the rendered component is the project list?

FredrikOseberg: Yes, I suppose we could setup a redirect to: `/projects`

thomasheartman: I don't really mind, to be honest. It used to be that that page only rendered "all features", right? So if that's not useful anymore, then rendering projects instead might make sense? I don't have enough context to say what's the right thing to do here, and I suspect that exploring the solution space for this would take up a lot more time than is warranted for this PR, so I'm happy for you to make a choice here. If you think it's worth a wider discussion (it might be), then we should probably loop in the UX team et al., and make it its own task.

PR #2900, opened 2023-01-13
fix: found an edge case exporting variants

gastonfournier:
## About the changes
When exporting v3, for variants backward compatibility, we need to find one featureEnvironment and fetch variants from there. 
In cases where the default environment is disabled (therefore does not get variants per environment when added), it can be still be selected for the export process. Therefore variants don't appear in the feature when they should be there.

An e2e test that fails with the previous implementation was added to validate the behavior

This comes from our support ticket 404

### DIFF
--- a/src/lib/services/state-service.ts
+++ b/src/lib/services/state-service.ts
@@ -712,26 +712,35 @@ export default class StateService {
         const v4 = await this.exportV4({ ...opts, includeEnvironments: true });
         // undefined defaults to true
         if (opts.includeFeatureToggles !== false) {
-            const keepEnv = v4.environments
-                .filter((env) => env.enabled !== false)
-                .sort((e1, e2) => {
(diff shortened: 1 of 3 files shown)


src/lib/services/state-service.ts
```diff
@@ -712,26 +712,35 @@ export default class StateService {
...
+                    (env) => fE.environment === env.name,
+                );
+                return { ...fE, ...envDetails, active: fE.enabled };
+            });
             v4.features = v4.features.map((f) => {
-                const variants = featureEnvs.find(
-                    (fe) => fe.featureName === f.name,
-                )?.variants;
+                const variants = featureAndEnvs
+                    .sort((e1, e2) => {
```
gastonfournier, line 727: The sorting of the environments is not absolute, it depends on whether the feature in a particular project has that environment active or not plus the other conditions.

sighphyre (approved):
G

PR #2901, opened 2023-01-13
fix: SA unique token bug

nunogois:
Fixes a small bug on service account token creation where the tokens the new one was being validated against were not the most up-to-date ones.


chriswk (approved):
LGTM

PR #2904, opened 2023-01-16
docs: update footer links

thomasheartman:
## What

This PR:

-   adds a footer link to GitHub discussions
-   updates the link to the public roadmap

## Why

Because we want more people to use discussions as a resource and because the old roadmap link leads to a deprecated roadmap.


chriswk (approved):
LGTM

PR #2909, opened 2023-01-17
docs: add change requests video

thomasheartman:
This change adds the change requests explainer video to the change requests reference documentation.

As a bonus, it also cleans up some redundant whitespace.


EliseBrevald:
It's looking great @thomasheartman

PR #2911, opened 2023-01-17
add selectable timezone

andreas-unleash:

1st Iteration = Show the user the timezone and UTC offset

## About the changes

Closes #
[issue 1-568](https://linear.app/unleash/issue/1-568/show-the-timezone-for-the-currenttime-constraint)


frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,41 +1,116 @@
...
 export const DateSingleValue = ({
     setValue,
     value,
     error,
     setError,
 }: IDateSingleValueProps) => {
+    const { uiConfig } = useUiConfig();
+    const showSelectableTimezone = Boolean(uiConfig.flags.selectableTimezone);
+    const timezones = useTimezones();
+    const { timeZone: local } = Intl.DateTimeFormat().resolvedOptions();
```
thomasheartman, line 40: I was a bit unsure where the `local` variable was defined. It also wasn't clear what it was, so maybe renaming it to `localTimeZone` would help?

```suggestion
    const { timeZone: localTimeZone } = Intl.DateTimeFormat().resolvedOptions();
```

But on further inspection, I see that `localTimezone` is another variable that's defined. But that's only ever used in cases where `showSelectableTimezone` is false 🤔

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,41 +1,116 @@
...
+                                    value={timezone}
+                                    sx={{ minWidth: 180 }}
+                                />
+                            }
+                        />
+                    }
+                    elseShow={
+                        <p>
+                            {localTimezone?.key}{' '}
+                            {`(UTC ${localTimezone?.utcOffset})`}
```
thomasheartman, line 101: What happens if `localTimezone` is undefined here? Do we just show an empty `p` tag? Would it make sense to show something else if it _is_ undefined?

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,41 +1,116 @@
...
+}));
+
 export const DateSingleValue = ({
     setValue,
     value,
     error,
     setError,
 }: IDateSingleValueProps) => {
+    const { uiConfig } = useUiConfig();
+    const showSelectableTimezone = Boolean(uiConfig.flags.selectableTimezone);
```
thomasheartman, line 38: If I understand correctly: for the first iteration, this will always be false. Is that correct?

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,41 +1,116 @@
 import { ConstraintFormHeader } from '../ConstraintFormHeader/ConstraintFormHeader';
 import Input from 'component/common/Input/Input';
 import { parseDateValue, parseValidDate } from 'component/common/util';
+import {
+    ITimezoneSelect,
+    useTimezones,
+} from '../../../../../../hooks/useTimezones';
```
Tymek, line 7: Please clean up imports. I know it's autocompleted by IDE, but it's hard to read. I'll not adding ESlint rule for this in the future.

thomasheartman (commented):
Nice work! Looks pretty good to me, but I've got a few questions and suggestions (mostly related to naming and clarity). 

If I read the code correctly, though, it also attempts to deliver what's necessary for version 2 of the feature. Is that correct? If this is only version one (unselectable time zone), shouldn't we leave the extra code out until the next version is ready?

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,41 +1,116 @@
...
     setValue: (value: string) => void;
     value?: string;
     error: string;
     setError: React.Dispatch<React.SetStateAction<string>>;
 }
 
+const StyledWrapper = styled('div')(({ theme }) => ({
+    display: 'flex',
+    flexDirection: 'row',
+    marginBottom: 8,
```
Tymek, line 26: `theme.spacing`

frontend/src/hooks/useTimezones.ts
```diff
@@ -0,0 +1,26 @@
+import TimezoneCountries from 'countries-and-timezones';
+import { useEffect, useState } from 'react';
+import { ISelectOption } from '../component/common/GeneralSelect/GeneralSelect';
+
+export interface ITimezoneSelect extends ISelectOption {
+    utcOffset: string;
+}
+export const useTimezones = (): ITimezoneSelect[] => {
```
Tymek, line 8: Why is this a hook without props, that is setting things on `useEffect`? Can this be just a getter function? Why does this require internal state?

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ConstraintFormHeader/ConstraintFormHeader.tsx
```diff
@@ -1,22 +1,15 @@
...
-const useStyles = makeStyles()(theme => ({
-    header: {
-        fontSize: theme.fontSizes.bodySize,
-        fontWeight: 'normal',
-        marginTop: '1rem',
-        marginBottom: '0.25rem',
-    },
+const StyledHeader = styled('h3')(({ theme }) => ({
+    fontSize: theme.fontSizes.bodySize,
+    fontWeight: 'normal',
```
Tymek, line 6: `theme.typography.fontWeightMedium`

Tymek (changes requested):

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
andreas-unleash: Yes, the implementation for selectable timezone is there behind the flag until the sdks catch up

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
andreas-unleash: sure, I always forget

frontend/src/hooks/useTimezones.ts
andreas-unleash: Getter makes sense. Thanks

frontend/src/hooks/useGetTimezones.ts
```diff
@@ -0,0 +1,22 @@
+import TimezoneCountries from 'countries-and-timezones';
+import { useCallback } from 'react';
+import { ISelectOption } from '../component/common/GeneralSelect/GeneralSelect';
+
+export interface ITimezoneSelect extends ISelectOption {
+    utcOffset: string;
+}
+export const useGetTimezones = () => {
```
Tymek, line 8: I don't think this needs to be a hook. `const getTimezones = () => ...` or even `const timezones = ` will work the same.

frontend/package.json
```diff
@@ -152,5 +152,9 @@
     "ignorePatterns": [
       "cypress"
     ]
+  },
```
Tymek, line 155: For frontend we where using only `devDependencies`. I'm all for separating build tools from bundled dependencies, maybe in a separate PR or with tech debt task.

Tymek (changes requested):

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
thomasheartman: Got it. But that is for iteration 2, right? Do we want to merge the code to deal with that now, or wait until it's time? (No right or wrong answers; just opening for discussion)

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,41 +1,122 @@
...
+                        />
+                    }
+                    elseShow={
+                        <ConditionallyRender
+                            condition={Boolean(localTimezone)}
+                            show={
+                                <p>
+                                    {localTimezone?.key}{' '}
+                                    {`(UTC ${localTimezone?.utcOffset})`}
+                                </p>
```
thomasheartman, line 104: What do we render if localTimezone is undefined? Is this just an empty paragraph? Can we render something sensible here in that case?

frontend/package.json
andreas-unleash: Moved

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
andreas-unleash: How about 'No timezone information available'

andreas-unleash:
@thomasheartman I think taking the code out is not necessary since it is behind a flag

FredrikOseberg:
I think that unless we are absolutely certain that we want to have this, we should take it out. I don't think we have enough data points to warrant going down the rabbit hole of saving timezone information. I would prefer if we reverted it and evaluated the need. If we ever end up doing iteration 2, let's link to this PR in the project to not lose the context and the work.

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
thomasheartman: Sure, that's an option, but it doesn't tell the user what time zone the time we're showing them is? I don't know exactly when this will be undefined, but what about something like:

> The time shown is in your local time zone according to your browser.

It seems that the only times we won't have it, is when we can't find the key in the list of time zones, yeah?

andreas-unleash: yes

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,42 +1,88 @@
...
+                    error={Boolean(error)}
+                    errorText={error}
+                    required
+                />
+                <ConditionallyRender
+                    condition={Boolean(localTimezone)}
+                    show={
+                        <p>
+                            {localTimezone?.key}{' '}
+                            {`(UTC ${localTimezone?.utcOffset})`}
```
thomasheartman, line 80: Can we remove these now? Or is there a thing where it doesn't recognize that the condition guarantees this?

```suggestion
                            {localTimezone.key}{' '}
                            {`(UTC ${localTimezone.utcOffset})`}
```

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,42 +1,88 @@
...
+                        </p>
+                    }
+                    elseShow={<p>No timezone information available</p>}
```
thomasheartman, line 83: So something like this, yeah?

```suggestion
                    elseShow={<p>The time shown is in your local time zone according to your browser.</p>}
```

thomasheartman (changes requested):
Nice one! I'm happy once we change the "No timezone information available" text. I've added a suggestion for it. I've also added a suggestion for conditional accessors, but I'm not sure whether that'll work or not.

thomasheartman (approved):
Thanks! Looks good to me 😄

thomasheartman:
@andreas-unleash Well, except that the build / test is failing. As expected, the compiler doesn't understand the implication between the check for `localTimezone`'s presence in the `ConditionallyRender` component's condition and it's render method.

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/DateSingleValue/DateSingleValue.tsx
```diff
@@ -1,42 +1,93 @@
...
+                    }}
+                    error={Boolean(error)}
+                    errorText={error}
+                    required
+                />
+                <ConditionallyRender
+                    condition={Boolean(localTimezone)}
+                    show={
+                        <p>
+                            {localTimezone!.key}{' '}
```
andreas-unleash, line 79: I don't usually like the `!` operator, but in this case it is safe since we already check for truthiness

thomasheartman: Yes, I agree. However, it does introduce a potential for error later if the condition were to change. Could this perhaps be changed to regular ternary? E.g.

```
Boolean(localTimezone) ? <p>{localTimezone.key[...]</p> : <p>The time shown is your browser's local time</p>
```

Maybe even extract the string to a variable higher up?

```
const timeZoneDescription = localTimezone ? `${localTimezone.key} [...]` : "The time shown is your browser's local time"
```

and then skip the last conditional altogether?

thomasheartman: If the condition suddenly changes to `condition={true}` (or any other condition that doesn't rely on `localTimezone` being present), then the `!` will try and access an object that is undefined.

thomasheartman: Or, well, it could 💁🏼 So I think if we can avoid introducing a potential error like that, then we should.

thomasheartman (approved):
Safe! Thanks for implementing the changes 🙌🏼  This looks good to me 🚂

Tymek (approved):

PR #2914, opened 2023-01-17
feat: Add ability to push variants to multiple environments

gastonfournier:
## About the changes
This PR adds the ability to push variants to multiple environments overriding the existing variants.

## Discussion points
1. Should this fail if there are variants in other environments? Initially no, because otherwise the operation wouldn't be idempotent and it should have that property (setting variants to 1 or more environments once or twice should not make a difference)


src/lib/routes/admin-api/project/variants.ts
```diff
@@ -193,6 +223,62 @@ export default class VariantsController extends Controller {
...
+        await this.featureService.setVariantsOnEnvs(
+            projectId,
+            featureName,
+            environments,
+            variants,
+            userName,
+        );
+        res.status(200).json({
+            version: 1,
+            variants: variants,
```
gastonfournier, line 255: Worth noticing that here we're just echoing the variants received as parameter we could potentially collect the responses from each environment and return an object with variants per each environment or we could also just return a 202 without body...

chriswk (approved):
Nice! :)

PR #2917, opened 2023-01-17
feat: adapt user logic to better adapt to SAs

nunogois:
https://linear.app/unleash/issue/2-579/improve-user-like-behaviour-for-service-accounts-accounts-concept

image

Makes SAs behave more like users. 

Even though they share the same `users` database table, the `is_service` column distinguishes them. This PR makes the distinction a bit less obvious by not filtering out SAs for some methods in the user store, returning both account types and their respective account type information so we can handle them properly on the UI.

We felt like this was a good enough approach for now, and a decent compromise to move SAs forward. In the future, we may want to make a full refactor with the `accounts` concept in mind, which we've experimented with in the [accounts-refactoring](https://github.com/Unleash/unleash/tree/accounts-refactoring) branches (both OSS and Enterprise).
 
https://github.com/Unleash/unleash/pull/2918 - Moves this a bit further, by introducing the account service and store.

### DIFF
--- a/frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
+++ b/frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
@@ -7,6 +7,8 @@ import { useUsers } from 'hooks/api/getters/useUsers/useUsers';
 import { IGroupUser } from 'interfaces/group';
 import { UG_USERS_ID } from 'utils/testIds';
 import { caseInsensitiveSearch } from 'utils/search';
+import { useServiceAccounts } from 'hooks/api/getters/useServiceAccounts/useServiceAccounts';
+import { IServiceAccount } from 'interfaces/service-account';
 
 const StyledOption = styled('div')(({ theme }) => ({
     display: 'flex',
@@ -44,7 +46,11 @@ const renderOption = (
         />
         <StyledOption>
             <span>{option.name || option.username}</span>
-            <span>{option.email}</span>
+            <span>
+                {option.name && option.username
+                    ? option.username
+                    : option.email}
+            </span>
         </StyledOption>
     </li>
 );
@@ -57,6 +63,10 @@ const renderTags = (value: IGroupUser[]) => (
     </StyledTags>
 );
 
+type UserOption = IUser & {
+    type: string;
+};
+
 interface IGroupFormUsersSelectProps {
     users: IGroupUser[];
     setUsers: React.Dispatch<React.SetStateAction<IGroupUser[]>>;
@@ -67,6 +77,27 @@ export const GroupFormUsersSelect: VFC<IGroupFormUsersSelectProps> = ({
     setUsers,
 }) => {
     const { users: usersAll } = useUsers();
+    const { serviceAccounts } = useServiceAccounts();
+
+    const options = [
+        ...usersAll
+            .map((user: IUser) => ({ ...user, type: 'USERS' }))
+            .sort((a: IUser, b: IUser) => {
+                const aName = a.name || a.username || '';
+                const bName = b.name || b.username || '';
+                return aName.localeCompare(bName);
+            }),
+        ...serviceAccounts
+            .map((serviceAccount: IServiceAccount) => ({
+                ...serviceAccount,
+                type: 'SERVICE ACCOUNTS',
+            }))
+            .sort((a, b) => {
+                const aName = a.name || a.username || '';
+                const bName = b.name || b.username || '';
+                return aName.localeCompare(bName);
+            }),
+    ];
 
     return (
         <StyledGroupFormUsersSelect>
@@ -77,7 +108,7 @@ export const GroupFormUsersSelect: VFC<IGroupFormUsersSelectProps> = ({
                 limitTags={1}
                 openOnFocus
                 disableCloseOnSelect
-                value={users}
+                value={users as UserOption[]}
                 onChange={(event, newValue, reason) => {
                     if (
                         event.type === 'keydown' &&
@@ -88,13 +119,10 @@ export const GroupFormUsersSelect: VFC<IGroupFormUsersSelectProps> = ({
                     }
                     setUsers(newValue);
                 }}
-                options={[...usersAll].sort((a, b) => {
-                    const aName = a.name || a.username || '';
-                    const bName = b.name || b.username || '';
-                    return aName.localeCompare(bName);
-                })}
+                groupBy={option => option.type}
+                options={options}
                 renderOption={(props, option, { selected }) =>
-                    renderOption(props, option as IUser, selected)
+                    renderOption(props, option as UserOption, selected)
                 }
                 filterOptions={(options, { inputValue }) =>
                     options.filter(
@@ -105,7 +133,7 @@ export const GroupFormUsersSelect: VFC<IGroupFormUsersSelectProps> = ({
                     )
                 }
                 isOptionEqualToValue={(option, value) => option.id === value.id}
-                getOptionLabel={(option: IUser) =>
+                getOptionLabel={(option: UserOption) =>
                     option.email || option.name || option.username || ''
                 }
                 renderInput={params => (
--- a/frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
+++ b/frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
@@ -34,6 +34,7 @@ import {
     PA_USERS_GROUPS_TITLE_ID,
 } from 'utils/testIds';
 import { caseInsensitiveSearch } from 'utils/search';
+import { IServiceAccount } from 'interfaces/service-account';
 
 const StyledForm = styled('form')(() => ({
     display: 'flex',
@@ -99,6 +100,7 @@ interface IProjectAccessAssignProps {
     selected?: IProjectAccess;
     accesses: IProjectAccess[];
     users: IUser[];
+    serviceAccounts: IServiceAccount[];
     groups: IGroup[];
     roles: IProjectRole[];
 }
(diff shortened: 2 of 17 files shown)


frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
```diff
@@ -44,7 +46,11 @@ const renderOption = (
         />
         <StyledOption>
             <span>{option.name || option.username}</span>
-            <span>{option.email}</span>
+            <span>
+                {option.name && option.username
+                    ? option.username
+                    : option.email}
```
gastonfournier, line 52: I understand the logic here is: 
- If we've shown the name in the previous span and we also have a username, just display the username. 
- Otherwise, just show the email

This leaves us with either:
1. name + username (if both are present)
2. name + email (if username is not present)
3. username + email (if name is not present)

I think it'd be simpler if we just have conditional spans for each one of the values. Would that be an option?

frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
```diff
@@ -67,6 +77,27 @@ export const GroupFormUsersSelect: VFC<IGroupFormUsersSelectProps> = ({
...
+            .map((serviceAccount: IServiceAccount) => ({
+                ...serviceAccount,
+                type: 'SERVICE ACCOUNTS',
+            }))
+            .sort((a, b) => {
+                const aName = a.name || a.username || '';
+                const bName = b.name || b.username || '';
+                return aName.localeCompare(bName);
+            }),
+    ];
```
gastonfournier, line 100: Would this have the same result? Feel free to dismiss it

```suggestion
    const options = [
        ...usersAll
            .map((user: IUser) => ({ ...user, type: 'USERS' } as UserOption)),
        ...serviceAccounts
            .map((serviceAccount: IServiceAccount) => ({
                ...serviceAccount,
                type: 'SERVICE ACCOUNTS',
            } as UserOption)),
    ].sort((a: UserOption, b: UserOption) => {
                const aName = a.name || a.username || '';
                const bName = b.name || b.username || '';
                return aName.localeCompare(bName);
    });
```

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
```diff
@@ -250,7 +278,11 @@ export const ProjectAccessAssign = ({
                             <span>
                                 {optionUser?.name || optionUser?.username}
                             </span>
-                            <span>{optionUser?.email}</span>
+                            <span>
+                                {optionUser?.name && optionUser?.username
+                                    ? optionUser?.username
+                                    : optionUser?.email}
```
gastonfournier, line 284: Looks like the same logic as above... I'm just wondering if this logic can be encapsulated in a method maybe inside `IUser` as `identifier()` or `accountIdentifier()` which seems to be the purpose of showing either email (for users) or username (for service accounts)

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
```diff
@@ -336,7 +372,11 @@ export const ProjectAccessAssign = ({
                                 }}
                                 filterOptions={(options, { inputValue }) =>
                                     options.filter((option: IAccessOption) => {
-                                        if (option.type === ENTITY_TYPE.USER) {
+                                        if (
+                                            option.type === ENTITY_TYPE.USER ||
+                                            option.type ===
+                                                ENTITY_TYPE.SERVICE_ACCOUNT
+                                        ) {
```
gastonfournier, line 379: Same suggestion, maybe you can encapsulate this logic in a method `isAccount()` inside `IAccessOption`

frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
```diff
@@ -228,7 +228,9 @@ export const ProjectAccessTable: VFC = () => {
                             permission={UPDATE_PROJECT}
                             projectId={projectId}
                             to={`edit/${
-                                row.type === ENTITY_TYPE.USER ? 'user' : 'group'
+                                row.type === ENTITY_TYPE.GROUP
+                                    ? 'group'
+                                    : 'user'
```
gastonfournier, line 233: Should this `'user'` be a `'service account'` or is it ok to aggregate both under `user` in this case?

chriswk (approved):
:+1:

src/lib/db/user-store.ts
```diff
@@ -133,6 +126,11 @@ class UserStore implements IUserStore {
     }
 
     async getAll(): Promise<User[]> {
+        const users = await this.activeAll().select(USER_COLUMNS);
```
gastonfournier, line 129: Should we guard this change behind a flag?

gastonfournier (commented):
I see it fine, I just have a couple of clarification questions but other than that I'm ready to hit the approve button

frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
nunogois: Can you show an example of what you have in mind?

It's important to keep the total spans as 2, a main "title" one and a "subtitle" one, like the screenshot I shared above in the PR description. We need to cover the following scenarios:

1. Service accounts have a name and username;
2. Normal user accounts usually have a name and email;
3. Accounts like `admin` don't have a name or email, only username;

frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
nunogois: I believe it wouldn't, as according to the current implementation each option should be sorted inside its own category. 

Here's what happens with the suggested code:

image

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
nunogois: Could maybe be represented as a method inside the `User` class or another auxiliary method. Something that would resolve the `IUser` model to a `{ title: string, subtitle: string }` object?

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
nunogois: Unsure if that would bring much more clarity, and if I understood correctly it would also mean having to introduce a new class or class-like implementing this method, but feel free to share a suggestion of what you have in mind.

I guess we could have an auxiliary method here `const isAccount = (option: IAccessOption) => option.type === ENTITY_TYPE.USER || option.type === ENTITY_TYPE.SERVICE_ACCOUNT` if you think it's clearer.

FYI: https://github.com/Unleash/unleash/pull/2918 changes this to check for `!== ENTITY_TYPE.GROUP` instead, which may help solve this by simplifying the check.

frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
nunogois: Nice catch! In this case it's important that both types are handled as users, as both types should be handled by `frontend/src/component/project/ProjectAccess/ProjectAccessEditUser/ProjectAccessEditUser.tsx`

frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
gastonfournier: I thought that sorting in "sections" would create some sort of sub-ordering. Of course we can first order by `type` and then by name, but I think your code would be clearer and easy to maintain

src/lib/db/user-store.ts
nunogois: Well, my reasoning is that if there are no service accounts it shouldn't matter, and there should only be service accounts if the flag is enabled.

With this PR, the only place this method is called from is `getBaseUsersAndGroups` in `user-admin` which fetches access options and should encompass both account types:
https://github.com/Unleash/unleash/blob/5ff9fa45e582a3bc1a1132c9ff751935d5ae5363/src/lib/routes/admin-api/user-admin.ts#L313

Besides, https://github.com/Unleash/unleash/pull/2918 reverts this back by making the `user` entity only handle "users" instead of all account types. It changes `getBaseUsersAndGroups` to use the new `accountService.getAll()` method instead.

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
gastonfournier: I thought about the IUser because that's the type of optionUser, but also because it could be a method of `IAccount` in the future. I feel that `username` or `email` is a way of identifying the user in a more user-friendly way than just `id`, that's why I thought about `accountIdentifier` but I don't know if transforming an IUser to {title, subtiltle} would be a good improvement compared to the current solution

frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
gastonfournier: Got it, so what I had in mind is potentially going to 3 spans, but if that's a restriction then forget about it. The only I'd consider is https://github.com/Unleash/unleash/pull/2917#discussion_r1073296806

src/lib/db/user-store.ts
gastonfournier: Yeah, it makes sense, it's controlled by the ability to add new service accounts. Fair enough :+1:

gastonfournier (approved):

PR #2918, opened 2023-01-17
feat: add the account abstraction logic

nunogois:
https://linear.app/unleash/issue/2-579/improve-user-like-behaviour-for-service-accounts-accounts-concept

Builds on top of https://github.com/Unleash/unleash/pull/2917 by moving the responsibility of handling both account types from `users` to `accounts`.

Ideally:
 - `users` - Should only handle users;
 - `service-accounts` - Should only handle service accounts;
 - `accounts` - Should handle any type of account;

This should hopefully also provide a good building block in case we later decide to refactor this further down the `accounts` path.

### DIFF
--- a/src/lib/db/account-store.ts
+++ b/src/lib/db/account-store.ts
@@ -1,15 +1,9 @@
-/* eslint camelcase: "off" */
-
 import { Knex } from 'knex';
 import { Logger, LogProvider } from '../logger';
 import User from '../types/user';
 
 import NotFoundError from '../error/notfound-error';
-import {
-    ICreateUser,
-    IUserLookup,
-    IUserUpdateFields,
-} from '../types/stores/user-store';
+import { IUserLookup } from '../types/stores/user-store';
 import { IAccountStore } from '../types';
 
 const TABLE = 'users';
@@ -34,13 +28,6 @@ const emptify = (value) => {
 };
 
 const safeToLower = (s?: string) => (s ? s.toLowerCase() : s);
-
-const mapUserToColumns = (user: ICreateUser) => ({
-    name: user.name,
-    username: user.username,
-    email: safeToLower(user.email),
-    image_url: user.imageUrl,
-});
 
 const rowToUser = (row) => {
     if (!row) {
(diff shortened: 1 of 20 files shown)


src/lib/types/stores/account-store.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface IUserUpdateFields {
+    name?: string;
+    email?: string;
+}
+
+export interface IAccountStore extends Store<IUser, number> {
+    update(id: number, fields: IUserUpdateFields): Promise<IUser>;
+    insert(user: ICreateUser): Promise<IUser>;
+    upsert(user: ICreateUser): Promise<IUser>;
```
gastonfournier, line 25: If we are adding these methods here, I wonder if we should remove the ones from user-store (or proxy them to this method), but because we're not calling them (at least not in this PR apparently), maybe these two can be removed from here.

src/lib/db/account-store.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    async upsert(user: ICreateUser): Promise<User> {
+        const id = await this.hasAccount(user);
+
+        if (id) {
+            return this.update(id, user);
+        }
+        return this.insert(user);
+    }
```
gastonfournier, line 93: If we remove the modification methods we can get rid of all this code

gastonfournier (approved):
Looks good but I think there's some additional code carried over from user-store that can potentially be removed.

src/lib/types/stores/account-store.ts
nunogois: There are very good points, and if we ever want to create a new `user` or `service-account` we should use their respective stores, not `account-store`. Addressed in: https://github.com/Unleash/unleash/pull/2918/commits/bd671eb83267b91274d027cbe615d261de59800d

src/lib/db/account-store.ts
nunogois: See: https://github.com/Unleash/unleash/pull/2918#discussion_r1073607177

PR #2922, opened 2023-01-18
Segment export

sjaanus:
Segment export

### DIFF
--- a/src/lib/services/export-import-service.ts
+++ b/src/lib/services/export-import-service.ts
@@ -4,6 +4,7 @@ import {
     IFeatureEnvironment,
     IFeatureStrategy,
     IFeatureStrategySegment,
+    ISegment,
     ITag,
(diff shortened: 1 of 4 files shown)


FredrikOseberg (approved):
LGTM

src/lib/services/export-import-service.ts
```diff
@@ -121,6 +124,7 @@ export default class ExportImportService {
             this.segmentStore.getAllFeatureStrategySegments(),
             this.contextFieldStore.getAll(),
             this.featureTagStore.getAll(),
+            this.segmentStore.getAll(),
```
sjaanus, line 127: Segments depend on the `featureStrategies`. This means we need to do some` .then() `here, or possibly even better, for future, is to rewrite it entirely in SQL to join features, strategies, segments etc.

PR #2924, opened 2023-01-18
Add support for case for IN and NOT_IN operators

andreas-unleash:

Adds support for case for IN and NOT_IN operators

## About the changes

Closes #

### Important files

Screenshot 2023-01-18 at 11 35 10


kwasniew (approved):

src/lib/util/feature-evaluator/constraint.ts
```diff
@@ -39,10 +39,15 @@ const cleanValues = (values: string[]) =>
...
 const InOperator = (constraint: Constraint, context: Context) => {
     const field = constraint.contextName;
+    const caseInsensitive = Boolean(constraint.caseInsensitive);
     const values = cleanValues(constraint.values);
     const contextValue = resolveContextValue(context, field);
 
-    const isIn = values.some((val) => val === contextValue);
+    const isIn = values.some((val) =>
+        caseInsensitive
+            ? val.toLowerCase() === contextValue?.toLowerCase()
```
sighphyre, line 48: IIRC we don't have any we have any restrictions on unicode so localCompare or toLocaleLowerCase might be better here

sighphyre (commented):
Do we have a plan for rolling this out? The SDKs will need support to handle this (not a big deal, we just need to plan it), the bigger worry is that folks using a version that has this patch with an SDK that doesn't will get confused because the behaviour won't be what they expect. Last time we got ourselves into this situation we added a warning to the constraint page but adding more warnings is probably going to look horrid

PR #2925, opened 2023-01-18
refactor/clean-up-get-project

FredrikOseberg:
This PR moves the getProjectOverview method out from the project health controller. It doesn't make sense that this method lives here anymore, as over time it has grown into method that relays all information about a single project. It makes more sense that this now lives on the root of the project api. Also removes unwanted duplication of getProjectOverview from the project-service and the project-health-service.


kwasniew (approved):

chriswk (commented):
Definitely looks good :+1:

PR #2931, opened 2023-01-19
feat: status API

FredrikOseberg:
Draft PR for status API


src/lib/db/event-store.ts
```diff
@@ -120,6 +120,31 @@ class EventStore extends AnyEventEmitter implements IEventStore {
         return present;
     }
 
+    async getForFeatures(
+        features: FeatureToggle[],
```
sjaanus, line 124: Probably should be list of string[], so it will be better to reuse in future.

src/lib/db/event-store.ts
```diff
@@ -120,6 +120,31 @@ class EventStore extends AnyEventEmitter implements IEventStore {
...
+        environments: IProjectEnvironment[],
```
sjaanus, line 125: Same here

src/lib/db/environment-store.ts
```diff
@@ -201,6 +202,11 @@ export default class EnvironmentStore implements IEnvironmentStore {
                 { column: 'created_at', order: 'asc' },
             ]);
         const rows = await qB;
+
+        if (query) {
```
sjaanus, line 206: This seems wrong. Await is already executing the query. Where statement should be before await.

PR #2936, opened 2023-01-19
Add feature name to strategy schema

sjaanus:


kwasniew (approved):

src/lib/openapi/spec/feature-strategy-schema.ts
```diff
@@ -14,6 +14,9 @@ export const featureStrategySchema = {
         name: {
             type: 'string',
         },
+        featureName: {
+            type: 'string',
+        },
```
gastonfournier, line 19: Do we need the featureName here? We may be creating unnecessary coupling here. Usually, I'd expect a directed graph, starting from feature entity down to the strategy. That means that when you get to the strategy you already know the feature name. 

There might be use cases where we want to send a plain object with all properties, but I wouldn't overload types for these purposes... I just want to understand the need of this

kwasniew: We're just describing the shape of the data we return anyway. It's needed for the export/import functionality. This PR just describe something we already had. We can model it differently though.

gastonfournier: Well, that kind of describes my concern. We're changing the shape of the data but the typesystem is not helping to detect this. In some cases like https://github.com/Unleash/unleash/pull/2940 we even have a debug log, but no compilation error (which should be the goal IMO). I believe this could be because of a relaxed configuration in tsconfig.json but it happens that sometimes we modify shapes that are re-used elsewhere and it's causing mismatches with the API schemas...

Another idea is to use the same types for open API schemas and our shapes in the controller layer, because today we're relying on automatic type casting (I don't know if it's called that way):

Screenshot from 2023-01-20 08-53-09

which by some reason is not very strict. I think you'll know better than me, that's why I wanted to share

PR #2948, opened 2023-01-19
feat: add warning hints on potential misconfiguration

gastonfournier:
## About the changes
Add warnings when we detect something might  be wrong with the customer configuration, in particular with regard to variants configuration

## Rationale
Moving from variants per feature to variants per environment will allow users to have fine-grained permissions and more control over variants on different environments: #2254

But because this requires an additional step of copying variants to other environments, we identified the potential risk of users forgetting to follow this step. To keep them informed about this, we're introducing a warning sign after a toggle is enabled when we detect that:
1. The environment is enabled without variants
2. Other enabled environments have variants

This situation would be a problem if you rely on `getVariant` method from the SDK, because without variants you'll receive the default variant. Probably, not what you'd expect after enabling the toggle, but there are situations where this might be correct. Because of the latter, we thought that adding a warning and letting the user handle the situation was the best solution.

## UI sketches
image (6)
Screenshot from 2023-01-19 08-55-10

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewSidePanel/FeatureOverviewSidePanelEnvironmentSwitches/FeatureOverviewSidePanelEnvironmentSwitches.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewSidePanel/FeatureOverviewSidePanelEnvironmentSwitches/FeatureOverviewSidePanelEnvironmentSwitches.tsx
@@ -5,8 +5,7 @@ import { FeatureOverviewSidePanelEnvironmentSwitch } from 'component/feature/Fea
 import { Link, styled, Tooltip } from '@mui/material';
 import { Link as RouterLink } from 'react-router-dom';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
-import { HtmlTooltip } from 'component/common/HtmlTooltip/HtmlTooltip';
-import { WarningAmber } from '@mui/icons-material';
+import VariantsWarningTooltip from 'component/feature/FeatureView/FeatureVariants/VariantsTooltipWarning';
 
 const StyledContainer = styled('div')(({ theme }) => ({
     padding: theme.spacing(3),
@@ -15,12 +14,6 @@ const StyledContainer = styled('div')(({ theme }) => ({
 const StyledSwitchLabel = styled('div')(() => ({
     display: 'flex',
     flexDirection: 'column',
-}));
-
-const StyledWarningAmber = styled(WarningAmber)(({ theme }) => ({
-    color: theme.palette.warning.main,
-    fontSize: theme.fontSizes.bodySize,
-    marginLeft: theme.spacing(1),
 }));
 
 const StyledLabel = styled('p')(({ theme }) => ({
@@ -93,6 +86,10 @@ export const FeatureOverviewSidePanelEnvironmentSwitches = ({
                     </>
                 );
 
+                const hasWarning =
+                    environment.enabled &&
+                    variants.length == 0 &&
+                    someEnabledEnvironmentHasVariants;
                 return (
                     <FeatureOverviewSidePanelEnvironmentSwitch
                         key={environment.name}
@@ -110,34 +107,12 @@ export const FeatureOverviewSidePanelEnvironmentSwitches = ({
                                 {strategiesLabel}
                                 {variantsLink}
                                 <ConditionallyRender
-                                    condition={
-                                        variants.length == 0 &&
-                                        environment.enabled &&
-                                        someEnabledEnvironmentHasVariants
-                                    }
+                                    condition={hasWarning}
                                     show={
-                                        <HtmlTooltip
-                                            arrow
-                                            title={
-                                                <>
-                                                    This environment has no
-                                                    variants enabled. If you
-                                                    check this feature's
-                                                    variants in this
-                                                    environment, you will get
-                                                    the{' '}
-                                                    <a
-                                                        href="https://docs.getunleash.io/reference/feature-toggle-variants#the-disabled-variant"
-                                                        target="_blank"
-                                                    >
-                                                        disabled variant
-                                                    </a>
-                                                    .
-                                                </>
-                                            }
-                                        >
-                                            <StyledWarningAmber />
-                                        </HtmlTooltip>
+                                        <>
+                                            <StyledSeparator />
+                                            <VariantsWarningTooltip />
+                                        </>
                                     }
                                 />
                             </StyledSubLabel>
--- a/frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
+++ b/frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { styled, useMediaQuery, useTheme } from '@mui/material';
-import { Add, WarningAmber } from '@mui/icons-material';
+import { Add } from '@mui/icons-material';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 import { SortingRule, useFlexLayout, useSortBy, useTable } from 'react-table';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
@@ -45,16 +45,11 @@ import { useFavoriteFeaturesApi } from 'hooks/api/actions/useFavoriteFeaturesApi
 import { FeatureTagCell } from 'component/common/Table/cells/FeatureTagCell/FeatureTagCell';
 import { useGlobalLocalStorage } from 'hooks/useGlobalLocalStorage';
 import { useConditionallyHiddenColumns } from 'hooks/useConditionallyHiddenColumns';
-import { HtmlTooltip } from 'component/common/HtmlTooltip/HtmlTooltip';
 import { flexRow } from 'themes/themeStyles';
+import VariantsWarningTooltip from 'component/feature/FeatureView/FeatureVariants/VariantsTooltipWarning';
 
 const StyledResponsiveButton = styled(ResponsiveButton)(() => ({
     whiteSpace: 'nowrap',
-}));
-
-const StyledWarningAmber = styled(WarningAmber)(({ theme }) => ({
-    color: theme.palette.warning.main,
-    fontSize: theme.fontSizes.bodySize,
 }));
 
 const StyledSwitchContainer = styled('div', {
@@ -314,28 +309,7 @@ export const ProjectFeatureToggles = ({
                             />
                             <ConditionallyRender
                                 condition={hasWarning}
-                                show={
-                                    <HtmlTooltip
-                                        arrow
-                                        title={
-                                            <>
-                                                This environment has no variants
-                                                enabled. If you check this
-                                                feature's variants in this
-                                                environment, you will get the{' '}
-                                                <a
-                                                    href="https://docs.getunleash.io/reference/feature-toggle-variants#the-disabled-variant"
-                                                    target="_blank"
-                                                >
-                                                    disabled variant
-                                                </a>
-                                                .
-                                            </>
-                                        }
-                                    >
-                                        <StyledWarningAmber />
-                                    </HtmlTooltip>
-                                }
+                                show={<VariantsWarningTooltip />}
                             />
                         </StyledSwitchContainer>
                     );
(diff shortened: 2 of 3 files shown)


frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -127,6 +135,16 @@ export const ProjectFeatureToggles = ({
...
+    const hasOtherEnvsWithVariants = (featureName: string) => {
+        const featureEnvs =
+            features.find(feature => feature.name === featureName)
+                ?.environments || [];
+        const found = featureEnvs.find(
+            env => env.enabled && env.variantCount > 0
+        );
+        return found !== undefined;
+    };
+
```
gastonfournier, line 147: @nunogois I'd like to get your advice on this piece of code. I don't like it but I'm limited in my React knowledge.

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -336,8 +372,13 @@ export const ProjectFeatureToggles = ({
...
                                 enabled:
                                     featureEnvironments?.find(
-                                        feature => feature?.name === env
+                                        fe => fe?.name === env
                                     )?.enabled || false,
+                                variantCount:
+                                    featureEnvironments?.find(
+                                        fe => fe.name === env
+                                    )?.variantCount || 0,
+                                // othersUseVariants: featureEnvironments?.find(fe => fe.variantCount > 0 && fe.enabled)?.enabled || false,
```
gastonfournier, line 381: Also @nunogois I don't understand this part... without adding `variantCount` here I don't see that field in another part of the code, but I tried using `othersUseVariants` but that's not visible... so I have the feeling that I'm doing something wrong

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
nunogois: Hmm on a first look I see it could be simplified as:

```suggestion
    const hasOtherEnvsWithVariants = (featureName: string) => {
        const featureEnvs =
            features.find(feature => feature.name === featureName)
                ?.environments || [];
        return featureEnvs.some(env => env.enabled && env.variantCount > 0);
    };

```

Or take it one step further as:

```suggestion
    const hasOtherEnvsWithVariants = (featureName: string) =>
      features.some(
          feature =>
              feature.name === featureName &&
              feature.environments.some(
                  env => env.enabled && env.variantCount > 0
              )
      );

```

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -274,13 +292,31 @@ export const ProjectFeatureToggles = ({
...
+                            featureName={feature?.name}
+                            environmentName={name}
+                            onToggle={onToggle}
+                        />
+                        <ConditionallyRender
+                            condition={
+                                hasOtherEnvsWithVariants(feature?.name) &&
+                                feature.environments[name]?.variantCount ===
+                                    0 &&
+                                feature.environments[name].enabled
```
nunogois, line 308: Could maybe be simplified and extracted into a function that takes the feature and environment (name), but I'm not sure if it would add much value.

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -336,8 +372,13 @@ export const ProjectFeatureToggles = ({
                                 name: env,
                                 enabled:
                                     featureEnvironments?.find(
-                                        feature => feature?.name === env
+                                        fe => fe?.name === env
```
nunogois, line 375: It's nice that we changed `feature` to `fe` to better represent what we're dealing with here, since it's an environment from the feature, not the feature itself. I would still prefer something more explicit, like `environment` or `featureEnvironment` instead of `fe`.

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
gastonfournier: Yes, but I wanted to keep this logic closer to [this line](https://github.com/Unleash/unleash/pull/2948/files#diff-a1c10fdcd539d988f74d4c3742cc6b8844152999a6c7e509ed8b1765c604524fR305) because here I'm using global `features` variable to find a feature. But at line 305 I already have the feature. 

The problem I faced is that the shape of that feature is different than the shape of the feature in this `features` array, in particular, it's missing access to other environments (the type of the `environments` is different IIRC). If I could do that, then I wouldn't have to iterate on features to find the feature, because I'd already have it

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
gastonfournier: Well, yes, but it gets too long. I chose `fe` because we're using that acronym elsewhere and also because it's close to the collection `featureEnvironments` so you can quickly look it up, but having `feature` was misleading because you'd trust it to be a feature without looking at the definition. Still, I agree `featureEnvironment` could be a good variable name

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
gastonfournier: Yes, I wanted to extract it to a constant and keep it close, but it was not possible in this context. I'm happy to jump into another pairing session to make this better

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
nunogois: Hmm if I understood correctly you probably want something like:

```TS
othersUseVariants:
  featureEnvironments?.some(
      environment =>
          environment.variantCount > 0 &&
          environment.enabled
  ) || false,
```

You can then add `othersUseVariants: boolean;` to `type ListItemType` on line 66.

Which means you could do something like this on the cell conditional rendering:

```TS
feature.environments[name]?.othersUseVariants &&
feature.environments[name]?.variantCount ===
    0 &&
feature.environments[name].enabled
```

You could probably also just compute whether to show the warning inside the memo, since it seems you have everything you need in there.

src/test/fixtures/fake-feature-environment-store.ts
```diff
@@ -22,12 +22,24 @@ export default class FakeFeatureEnvironmentStore
...
+        featureName: string,
+        environments: string[],
+        variants: IVariant[],
     ): Promise<void> {
         this.featureEnvironments
             .filter(
                 (fe) =>
                     fe.featureName === featureName &&
-                    fe.environment === environment,
+                    environments.indexOf(fe.environment) !== -1,
```
nunogois, line 42:
```suggestion
                    environments.includes(fe.environment),
```

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
gastonfournier: Thanks! an awesome suggestion!

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
gastonfournier: [bdf3e4a](https://github.com/Unleash/unleash/pull/2948/commits/bdf3e4a126d8f889369ae62060c10deb11c41694) as gratitude for all your help :hugs:

nunogois: Ah, thanks! 😁 I think I got the habit of being more extensive and explicit with my naming because of PR comments like these, especially from @FredrikOseberg - Otherwise I always opted for short variable names in these cases. 

But I think it's a good habit to have and it doesn't represent that much more effort for potentially more readability and clarity.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewSidePanel/FeatureOverviewSidePanelEnvironmentSwitches/FeatureOverviewSidePanelEnvironmentSwitches.tsx
```diff
@@ -89,6 +102,34 @@ export const FeatureOverviewSidePanelEnvironmentSwitches = ({
                             <StyledSubLabel>
                                 {strategiesLabel}
                                 {variantsLink}
+                                <ConditionallyRender
+                                    condition={
```
kwasniew, line 106: I'd extract the condition to the intention-revealing name

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewSidePanel/FeatureOverviewSidePanelEnvironmentSwitches/FeatureOverviewSidePanelEnvironmentSwitches.tsx
```diff
@@ -89,6 +102,34 @@ export const FeatureOverviewSidePanelEnvironmentSwitches = ({
...
+                                        variants.length == 0 &&
+                                        environment.enabled &&
+                                        someEnabledEnvironmentHasVariants
+                                    }
+                                    show={
+                                        <HtmlTooltip
```
kwasniew, line 112: We can extract this to const VariantsTooltip = <HtmlTooltip... so that show is a one liner

kwasniew (approved):
LGTM

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewSidePanel/FeatureOverviewSidePanelEnvironmentSwitches/FeatureOverviewSidePanelEnvironmentSwitches.tsx
gastonfournier: Not sure if I can make it better than:
```
environment.enabled &&
variants.length == 0 &&
someEnabledEnvironmentHasVariants
```
If you're suggesting to extract this in a single variable/method it would be something like: `isEnabledWithoutVariantsAndOtherEnabledEnvironmentsHaveVariants(environment)`, in which case I prefer my version, but if you're suggesting something different please let me know (I can also change it after merging)

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewSidePanel/FeatureOverviewSidePanelEnvironmentSwitches/FeatureOverviewSidePanelEnvironmentSwitches.tsx
gastonfournier: I've extracted it to a component: https://github.com/Unleash/unleash/pull/2948/files#diff-b9629cb7366c55ca30629ef8fc0c813a999819557d3e886511c3e2001cd00db6 I was also concerned about having the content of the text duplicated.

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -274,13 +288,41 @@ export const ProjectFeatureToggles = ({
...
+                    <StyledSwitchContainer>
+                        <FeatureToggleSwitch
+                            value={value}
+                            projectId={projectId}
+                            featureName={feature?.name}
+                            environmentName={name}
+                            onToggle={onToggle}
+                        />
+                        <ConditionallyRender
+                            condition={
```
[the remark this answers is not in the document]
gastonfournier: Followed the same approach Nuno used for the other file

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewSidePanel/FeatureOverviewSidePanelEnvironmentSwitches/FeatureOverviewSidePanelEnvironmentSwitches.tsx
kwasniew: maybe showVariantWarning: boolean?

gastonfournier: I decided to use `hasWarning` because there was also the idea of using the same warning for other purposes in the future.

PR #2957, opened 2023-01-20
fix: filtering environments in the project features table

nunogois:
image

Fixes an issue where environments were not filtered correctly on the search, most likely due to some regression accidentally introduced with changes in the meantime.


sighphyre (approved):
LGTM

PR #2965, opened 2023-01-23
feat: adds the PAT_DELETE event

nunogois:
Adds the `PAT_DELETE` event so we can log the relevant information when a PAT is deleted.

Should cover the following scenarios:

- User deletes their own PAT;
- Admin deletes another user's PAT;
- Admin deletes a Service Account token;


sighphyre (approved):
G

PR #2968, opened 2023-01-23
chore: increase max number of environments from 15 to 50

sighphyre:
Currently we have a soft limit of 15 environments, there's no technical or business reason for this, only that it's handled inelegantly in the UI. Users are asking for more

### DIFF
--- a/frontend/src/constants/values.ts
+++ b/frontend/src/constants/values.ts
@@ -1,1 +1,1 @@
-export const ENV_LIMIT = 15;
+export const ENV_LIMIT = 50;


nunogois (approved):
About this "handled inelegantly by the UI" thing, if I understood correctly, should we add some tasks do address it?

sighphyre:
> About this "handled inelegantly by the UI" thing, if I understood correctly, should we add some tasks do address it?

That's maybe not a bad idea actually. We probably need to start by discussing whether we want to or not. If the side bar gets a scroll bar, is that so bad? No idea

PR #2969, opened 2023-01-23
feat: add push to all button to UI

gastonfournier:
## About the changes
This adds the push to all button to the UI

### UI
Screenshot from 2023-01-16 12-25-49

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/PushVariantsButton/PushVariantsButton.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/PushVariantsButton/PushVariantsButton.tsx
@@ -68,6 +68,15 @@ export const PushVariantsButton = ({
         setPushToAnchorEl(null);
     };
 
+    const hasAccess = environments.reduce((ret, env) => {
+        ret[env.name] = useHasProjectEnvironmentAccess(
+            permission,
+            projectId,
+            env.name
+        );
+        return ret;
+    }, {} as { [env: string]: boolean });
+
     const variants =
         environments.find(environment => environment.name === current)
             ?.variants ?? [];
@@ -104,11 +113,7 @@ export const PushVariantsButton = ({
                                 <MenuItem key={otherEnvironment.name}>
                                     <FormControlLabel
                                         disabled={
-                                            !useHasProjectEnvironmentAccess(
-                                                permission,
-                                                projectId,
-                                                otherEnvironment.name
-                                            )
+                                            !hasAccess[otherEnvironment.name]
                                         }
                                         control={
                                             <Checkbox
--- a/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/PushVariantsButton/PushVariantsButton.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/PushVariantsButton/PushVariantsButton.tsx
@@ -10,7 +10,7 @@ import {
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 import { IFeatureEnvironment } from 'interfaces/featureToggle';
 import { useState } from 'react';
-import { useHasProjectEnvironmentAccess } from 'hooks/useHasAccess';
+import { useCheckProjectAccess } from 'hooks/useHasAccess';
 
 const StyledMenu = styled(Menu)(({ theme }) => ({
     '&>div>ul': {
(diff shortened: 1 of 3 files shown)


frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/PushVariantsButton/PushVariantsButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+    '&>div>ul': {
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        padding: theme.spacing(2),
+        '&>li': {
+            padding: theme.spacing(0),
+        },
+    },
+}));
```
nunogois, line 25: I was thinking and if we want to make it perfect maybe we shouldn't remove the padding from the `li`s. It looks very good overall but the hover effect on the items looks a bit weird. Maybe we should remove some of these specifics, clean some of this up, and try to only style a `div` for the separator and button below, but I'm not sure.

Something we can run by @NicolaeUnleash and see what he thinks.

frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/PushVariantsButton/PushVariantsButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+const StyledButton = styled(Button)(({ theme }) => ({
+    marginTop: theme.spacing(1),
+}));
```
nunogois, line 29:
```suggestion
}));

```

frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/PushVariantsButton/PushVariantsButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+    current,
+    environments,
+    permission,
+    projectId,
+    onSubmit,
+}: IPushVariantsButtonProps) => {
+    const [pushToAnchorEl, setPushToAnchorEl] = useState<null | HTMLElement>(
+        null
+    );
+    const pushToOpen = Boolean(pushToAnchorEl);
```
nunogois, line 48: In terms of code organisation since this is a computation based on a state, maybe it belongs some lines below instead, but it's a tiny nitpick based on preference.

frontend/src/hooks/api/actions/useFeatureApi/useFeatureApi.ts
```diff
@@ -223,6 +224,26 @@ const useFeatureApi = () => {
...
+            body: JSON.stringify({ variants, environments }),
+        });
+
+        try {
+            const res = await makeRequest(req.caller, req.id);
+            return res;
+        } catch (e) {
+            throw e;
+        }
+    };
```
nunogois, line 245: I know that this is copy pasted from the other methods, and we always do it like this, me included. Not sure about the original reasoning but the fact is that this could optionally be simplified.

nunogois (approved):
Looks great! But since I'm a bit biased from us pairing you may want another pair of eyes just in case 😄

frontend/src/hooks/api/actions/useFeatureApi/useFeatureApi.ts
gastonfournier: I believe the try catch throwing an exception without doing anything is pointless... I could just replace:
```typescript
        try {
            const res = await makeRequest(req.caller, req.id);
            return res;
        } catch (e) {
            throw e;
        }
```
with 
```typescript
        return makeRequest(req.caller, req.id);
```
and now there's no need to `await` because it's the return statement. 

But it's specified that way in [this ADR](https://github.com/Unleash/unleash/blob/main/website/docs/contributing/ADRs/front-end/preferred-data-mutation-method.md).  Maybe the intent of the ADR was to say that you should do something with the exception? Or perhaps it's something we carried away from another place...  :call_me_hand: @FredrikOseberg who may have more context

gastonfournier (commented):
I'm inclined to merge and move fast, then come back if needed, just to be able to release this week. I'm going to share in standup and make a decision there

sighphyre (approved):
LGTM

I see there's small detail chatter going on here. Agree some of it makes sense but it also might be worth doing in a separate PR due to time pressure around this one?

frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/PushVariantsButton/PushVariantsButton.tsx
nunogois: Yeah, good point. 
We are actually moving away from `.styles.ts` files and preferring styled components like you're doing here instead, so this looks good as is.

chriswk (approved):
I'm concurring with Simon and Nuno here. This looks good. The small nits around catching and rethrowing can be solved in a separate PR; it won't block us moving forward.

gastonfournier:
> Agree some of it makes sense but it also might be worth doing in a separate PR due to time pressure around this one?

This is what I plan to do, and I'll take in all comments in another PR. For others:

PR #2970, opened 2023-01-23
validation stage

kwasniew:
## About the changes

Validation stage displays:
* errors
* warnings
* bad data error toast message

Validation state:
* goes back to import screen
* prevents import on errors

Screenshot 2023-01-23 at 16 50 11

Further steps:
* UI tweaks and remove inline styles
* connect to the next ("import") stage
* generate type safe API client


sjaanus (approved):
LGTM!

PR #2981, opened 2023-01-24
fix: redirect only happening on root path with replace

FredrikOseberg:
Adds a fix for overeager redirects, now we will redirect on the root path and replace on our redirects from login.


gastonfournier (approved):
Tested in Vercel preview and it works nicely :+1:

PR #2986, opened 2023-01-25
Feat/project status monthly

FredrikOseberg:
This PR takes the project status API a step further by adding the capability of providing a date to control the selection. We are currently making calculations based on a gliding 30 day window, updated once a day. The initial database structure and method for updating the UI is outlined in this PR.

Discussion points: 
* Should the data filtering on gliding 30 window be offloaded to the database layer?
* Does the new items make sense on the project table?


src/lib/services/project-service.ts
```diff
@@ -57,6 +59,20 @@ export interface AccessWithRoles {
     groups: IGroupModelWithProjectRole[];
 }
 
+export interface IStatusUpdate {
+    avgTimeToProdCurrentWindow: number;
```
kwasniew, line 63: is the number days/hours/minutes/ms? maybe type alias (tiny types) would help?
e.g. type Hours = number;

src/lib/services/project-service.ts
```diff
@@ -57,6 +59,20 @@ export interface AccessWithRoles {
     groups: IGroupModelWithProjectRole[];
 }
 
+export interface IStatusUpdate {
```
kwasniew, line 62: I'm in two minds about this name. Aren't we doing project stats? Status sounds like on track/late etc.

src/lib/services/project-service.ts
```diff
@@ -596,20 +617,61 @@ export default class ProjectService {
...
         const features = await this.featureToggleStore.getAll({
             type: 'release',
             project: projectId,
         });
 
+        const [createdCurrentWindow, createdPastWindow] = await Promise.all([
+            await this.featureToggleStore.getByDate({
+                project: projectId,
+                dateAccessor: 'created_at',
+                date: subDays(new Date(), 30).toISOString(),
```
kwasniew, line 645: it's difficult to test it with the "new Date()" hardcoded. What about clock pattern where we inject time provider as a dependency? https://twitter.com/jarek000000/status/897470232159236096

src/lib/services/project-service.ts
```diff
@@ -596,20 +617,61 @@ export default class ProjectService {
...
+            }),
+            await this.featureToggleStore.getByDate({
+                project: projectId,
+                dateAccessor: 'created_at',
+                range: [
+                    subDays(new Date(), 60).toISOString(),
```
kwasniew, line 651: what about intention revealing local variable name for the -60 days and -30 days?

kwasniew (approved):

src/lib/read-models/project-status/project-status.test.ts
```diff
@@ -202,7 +208,29 @@ describe('calculate average time to production', () => {
             ...events,
         ]);
 
-        const timeToProduction = projectStatus.calculateAverageTimeToProd();
-        expect(timeToProduction).toBe(9.75);
+        const timeToProduction = projectStatus.calculateAverageTimeToProd(
```
kwasniew, line 211: I'm not sure if mean instead of average wouldn't be better

src/lib/services/project-service.ts
FredrikOseberg: Sure, let's change the name.

src/lib/services/project-service.ts
FredrikOseberg: I would like to explore this option, but I don't think I can get it done before my vacation and I don't want it to block the feature.

src/lib/read-models/project-status/project-status.test.ts
FredrikOseberg: Maybe, but since we scoped it to only toggles of type release average might give us a better dataset. We can always review after some time to see if we get a lot of outliers.

src/lib/services/project-service.ts
FredrikOseberg: Good idea.

src/lib/services/project-service.ts
sighphyre: @kwasniew I've been using that approach for years, didn't even realise it had a name, I learned something, thank you!

PR #2990, opened 2023-01-25
fix: revert table virtualization in variants per env

nunogois:
Fixes a bug where some variant tables did not render properly if scrolled all the way down.

Since we are not expecting to need virtualisation in these tables anyways (each table should hopefully have a relatively low amount of variants), reverting the `VirtualizedTable` usage to a normal one seemed like the most straightforward solution at this stage, fixing the issue without any noticeable side effects.

image


sighphyre (approved):
LGTM

PR #3004, opened 2023-01-26
feat: new variants per env form

nunogois:
https://linear.app/unleash/issue/2-647/adapt-current-variants-ui-to-better-align-with-cr

## About the changes
Big refactor to the variants per environment UI/UX logic, making the variants management happen on a side modal. This makes it so variants per environment play a lot nicer with change requests.

image

Relates to [roadmap](https://github.com/orgs/Unleash/projects/10) item: [#2254](https://github.com/Unleash/unleash/issues/2254)

### Important files
A big chunk of the changes is mostly moving things around or straight up removing them.

 - EnvironmentVariantModal - The modal itself that controls all of the variants that you're editing;
 - VariantForm - The extracted form for editing each of the variants;


chriswk (approved):
LGTM
---
Sidenote:
I'm missing tests of the weight balancing logic. We really should do something about the ease of testing setup in frontend; that isn't the responsibility of this PR.

frontend/src/component/common/util.ts
```diff
@@ -90,6 +91,40 @@ export function updateWeight(variants: IFeatureVariant[], totalWeight: number) {
...
+    const percentage = parseInt(
+        String(remainingPercentage / variableVariantCount)
+    );
+
+    return variants.map(variant => {
+        if (variant.weightType !== weightTypes.FIX) {
+            variant.weight = percentage;
+        }
+        return variant;
+    });
```
gastonfournier, line 125: This is probably (not sure), the same logic we have in the backend. Is there a way that can be shared? Just to avoid having it duplicated, because that means we need to keep them in sync

gastonfournier (approved):
I think we can merge it, give it a test in sandbox (eg instance) and if we see fit, we can add it into 4.20. Even though this is guarded when CRs are disabled, we need to test also with CR enabled

PR #3007, opened 2023-01-27
exclude tags for other features

kwasniew:
## About the changes

Import was failing on import of tags for features that don't exist in the new unleash. We filter out tags from other features that we don't export.

Also feature strategy query seems to be incorrect so I fixed it

### DIFF
--- a/src/lib/db/feature-strategy-store.ts
+++ b/src/lib/db/feature-strategy-store.ts
@@ -208,9 +208,9 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
         const query = this.db
             .select(COLUMNS)
             .from<IFeatureStrategiesTable>(T.featureStrategies)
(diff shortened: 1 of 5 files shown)


src/lib/db/feature-strategy-store.ts
```diff
@@ -208,9 +208,9 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
         const query = this.db
             .select(COLUMNS)
             .from<IFeatureStrategiesTable>(T.featureStrategies)
-            .where('environment', environment);
-        if (features) {
-            query.whereIn('feature_name', features);
+            .whereIn('feature_name', features);
+        if (environment) {
+            query.where('environment', environment);
```
kwasniew, line 213: environment is optional not a list of features

sjaanus (approved):

PR #3015, opened 2023-01-27
fix: down migration

FredrikOseberg:
* Fixes broken down migration


sighphyre (approved):
Fair enough, seems safe given it's in the down and a bug. LGTM

PR #3019, opened 2023-01-30
fix: limit creation of other users PATs (adapting)

nunogois:
https://linear.app/unleash/issue/2-656/limit-the-ability-of-creating-a-token-on-behalf-of-another-user

Adapts to the refactor that reverts the initial experimental idea of Service Accounts before they existed in the current implementation: Managing other user's PATs.


gastonfournier (approved):
LGTM!

Issue #2923, opened 2023-01-18
Archived toggles are included in the project view


ChristopheWanteeed:
### Describe the bug

Hello,
Archived toggles are included in the project view.
I think it is inconsistent.
Would it be possible to exclude them from this view please ?

For example on the screenshot, 45 toggles displayed but with 12 archived.

### Screenshots

![2023-01-18 Unleash](https://user-images.githubusercontent.com/112936569/213120328-023b94e4-0b0d-4ac5-ac55-744e4cb40bb9.png)

### Unleash version

Unleash enterprise 4.18.7

### Subscription type

Enterprise

### Hosting type

Hosted by Unleash

PR #3020, opened 2023-01-30
fix: list projects with all archived toggles

gastonfournier:
## About the changes
While trying to count only features that are not archived to display the amount of features of a project, accidentally we filtered out projects with all features archived (they should show up in the list but with count of features zero)

Improves: #2923

### DIFF
--- a/src/lib/db/project-store.ts
+++ b/src/lib/db/project-store.ts
@@ -95,7 +95,7 @@ class ProjectStore implements IProjectStore {
         }
         let selectColumns = [
             this.db.raw(
-                'projects.id, projects.name, projects.description, projects.health, projects.updated_at, count(case when features.archived_at is null then features.name end) AS number_of_features',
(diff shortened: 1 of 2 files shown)


ivarconr (approved):
Nice fix, it feels so obvious why mine did not work when the solution is in front of me.

src/lib/db/project-store.ts
```diff
@@ -89,14 +89,13 @@ class ProjectStore implements IProjectStore {
...
             .leftJoin('features', 'features.project', 'projects.id')
-            .where('features.archived_at', null)
             .orderBy('projects.name', 'asc');
         if (query) {
             projects = projects.where(query);
         }
         let selectColumns = [
             this.db.raw(
-                'projects.id, projects.name, projects.description, projects.health, projects.updated_at, count(features.name) AS number_of_features',
+                'projects.id, projects.name, projects.description, projects.health, projects.updated_at, count(features.name) FILTER (WHERE features.archived_at is null) AS number_of_features',
```
gastonfournier, line 98: Docs about filter on aggregate functions: https://www.postgresql.org/docs/current/sql-expressions.html

PR #3027, opened 2023-02-01
feat/enable_batch_metrics_for_all

ivarconr:
Batch Metrics as a capability developed to support the frontend API to handle more metrics from SDKs without overloading the DB to much. It has been running in Unleash Cloud for months and has proven to work quite nice. 

This PR simply removes the flag to make the capability GA, also for self-hosted users.


nunogois (approved):
LGTM

PR #3029, opened 2023-02-01
fix: when app count is zero because it just started

gastonfournier:
## About the changes
This fixes response time metrics with app names when the app just starts and has zero which is falsy. We want to compare against undefined (which means the snapshot is not yet ready)


nunogois (approved):
LG

PR #3013, opened 2023-01-27
ProjectMemberWidget live data

andreas-unleash:
ProjectMemberWidget live data
Styling fixes.
Refactored  toggle type widget

Signed-off-by: andreas-unleash <andreas@getunleash.ai>
Screenshot 2023-01-27 at 15 36 49

## About the changes

Closes #

PR #3030, opened 2023-02-01
fix: @server/openapi import breaking tsc

nunogois:
Seems like a `@server/openapi` import introduced with https://github.com/Unleash/unleash/pull/3013 was breaking `tsc`. Using a more explicit/extensive path to the schema seems to fix the issue for now.

See: https://unleash-internal.slack.com/archives/C048ELND3QD/p1675269006875629


nunogois:
Seems like this doesn't work after all 😦

Issue #3005, opened 2023-01-27
no auth is not working for frontend endpoint


DimaGolomozy:
### Describe the bug

when using locally the `unleash-server:v4.20.0` with the environment `AUTH_TYPE:NONE`
the `/api/client/...` endpoints work fine with no auth needed, this allows us to run locally unleash and test stuff without the need to provide a valid API token
but the `/api/frontend` for the client SDK still needs to pass auth header with a valid frontend API key.

this makes using unleash for local mode client testing very inconvenient, as each developer will need to spin up a local `unleash` server, login, and create a frontend token. we wanted to reduce those steps by providing `AUTH_TYPE:NONE`

### Steps to reproduce the bug

1. sping unleash up with docker compose:
```
    unleash-db:
        container_name: unleash-db
        image: postgres:13.5-alpine3.14
        environment:
            POSTGRES_PASSWORD: "unleash"
            POSTGRES_USER: "unleash"
            POSTGRES_DB: "unleash"
            PGUSER: "unleash"
        healthcheck:
            test: [ "CMD", "pg_isready", "--username=unleash", "--host=127.0.0.1", "--port=5432" ]
            interval: 2s
            timeout: 1m
            retries: 5
            start_period: 5s

    unleash:
        container_name: unleash
        image: unleashorg/unleash-server:4.20.0
        ports:
            - "4242:4242"
        environment:
            DATABASE_HOST: "unleash-db"
            DATABASE_NAME: "unleash"
            DATABASE_USERNAME: "unleash"
            DATABASE_PASSWORD: "unleash"
            DATABASE_SSL: "false"
            LOG_LEVEL: "debug"
            ENABLED_ENVIRONMENTS: "default"
            AUTH_TYPE: "NONE"
        healthcheck:
            test: [ "CMD", "nc",  "-z", "unleash-db", "5432" ]
            interval: 1s
            timeout: 1m
            retries: 5
            start_period: 10s
        depends_on:
            - unleash-db
```

2. simple curl that works fine:
```
curl http://localhost:4242/api/client/features
```
3. curl the frontend endpoint to see the error:
```
curl 'http://localhost:4242/api/frontend?appName=dima&environment=default'
```

this endpoint is not working and also there are errors on the unleash server:
```
[2023-01-27T06:47:47.709] [WARN] controller/ProxyController - Cannot destructure property 'type' of 'undefined' as it is undefined.
[...]

PR #3031, opened 2023-02-01
Support environment in API token even with auth disabled

kjvalencik:
## About the changes

The preferred method of providing the environment in API calls is implicitly in the API token. It is the [_only_ way to set it in the go client](https://github.com/Unleash/unleash-client-go/issues/95) since it does provide a way to add query parameters.

However, when using `AuthType.NONE`, the API Token middleware isn't included. Without the middleware, the environment isn't parsed from the API token in the authorization header and it always falls back to `default`.

After this change, when using `AuthType.NONE`, both of these provide a `200` with the expected results:

```
curl -H "Authorization: $API_TOKEN" http://localhost:4242/api/client/features
curl http://localhost:4242/api/client/features?environment=development
```

Closes #
https://github.com/Unleash/unleash/issues/2782
https://github.com/Unleash/unleash/issues/3005

### Important files

N/A

## Discussion points

It wasn't clear if there was a specific reason the API Token middleware was omitted. The middleware calls [`next()` without error when no token was provided](https://github.com/Unleash/unleash/blob/627958d30f01a8124c6f25c76efb16befd815d2c/src/lib/middleware/api-token-middleware.ts#L50). Since it functioned nearly identical to before (continued providing API results without `401`) when no `Authorization` header was present, I chose to unconditionally include the middleware.

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -103,25 +103,22 @@ export default async function getApp(
     );
 
     app.use(baseUriPath, patMiddleware(config, services));
+    app.use(baseUriPath, apiTokenMiddleware(config, services));
 
     switch (config.authentication.type) {
         case IAuthType.OPEN_SOURCE: {
-            app.use(baseUriPath, apiTokenMiddleware(config, services));
             ossAuthentication(app, config.server.baseUriPath);
             break;
         }
         case IAuthType.ENTERPRISE: {
-            app.use(baseUriPath, apiTokenMiddleware(config, services));
             config.authentication.customAuthHandler(app, config, services);
             break;
         }
         case IAuthType.HOSTED: {
-            app.use(baseUriPath, apiTokenMiddleware(config, services));
             config.authentication.customAuthHandler(app, config, services);
             break;
         }
         case IAuthType.DEMO: {
-            app.use(baseUriPath, apiTokenMiddleware(config, services));
             demoAuthentication(
                 app,
                 config.server.baseUriPath,
@@ -131,7 +128,6 @@ export default async function getApp(
             break;
         }
         case IAuthType.CUSTOM: {
-            app.use(baseUriPath, apiTokenMiddleware(config, services));
             config.authentication.customAuthHandler(app, config, services);
             break;
         }
@@ -140,7 +136,6 @@ export default async function getApp(
             break;
         }
         default: {
-            app.use(baseUriPath, apiTokenMiddleware(config, services));
             demoAuthentication(
                 app,
                 config.server.baseUriPath,


thomasheartman:
Hey, @kjvalencik 🙋🏼 Thanks for the PR!

Regarding the changes (and the feature request):
We've discussed this a bit internally and we've reached the conclusion that this could open up for a number of weird edge cases in SDKs and that it's not something we want to do.

For a little bit of background: `AuthType.NONE` was originally added as a way to enable folks to transition from v3 to v4 of Unleash and is not something we recommend you use. However, when you introduce something like that, it _does_ stick around.

However, to address your issue, there might be several ways to work around it:

If you need a **predictable API token** that you can just put into the SDK, you could create one and start Unleash with `INIT_API_TOKENS` to make that work. 

You could also write a custom bit of middleware that adds a fake token with your preferred environment to the request.

That last bit might be the closest to what you want to do here. Would either of those two serve your needs at all? 

I'm closing the PR for now, but let's continue the discussion. I'd be very interested in trying to understand your use case ☺️

thomasheartman:
@kjvalencik Sorry for the delay! 

>  I am looking to use AuthType.NONE [...] to make the management UI easier to use

Just for my understanding: are you talking about the Unleash management UI here? How does using `AuthType.NONE` make it easier to use?

Is it that if you use `AuthType.NONE`, then you don't need to log in to use the UI, but that also causes issues with the clients, because then they can't use tokens? Is that how it works? Because if so, then that sounds to me like a potential conflation of two separate things (on our end; not yours).

> Am I understanding correctly that AuthType.NONE is intentionally not working with current clients (e.g., the [go client](https://github.com/Unleash/unleash-client-go/issues/95))? Why not completely remove it?

Unfortunately, we can't remove it due to backwards compatibility. So for people who have had older installations of Unleash and still use no auth and the default environment: removing `AuthType.NONE` now would be a breaking change. It's something we may do in v5, but it can't be done until then.

As for whether it's "intentionally not working": that's not quite accurate. It should - to the best of my understanding - work with `AuthType.NONE` if you still use the default environment. However, that environment is only available to users who upgraded from versions before 4.3.

> My goal with this PR was to help others from spending a few hours debugging like me. I don't personally need the feature and can live with the OSS authentication and a default user. I can even make it a little easier by putting the default local dev username/password in our password manager.

For sure! And I think this is an important point. If the issue is indeed that you can't disable the password prompt but still keep using tokens, then that might be something we want to look into.

> I think this may need some doc improvements since there have been several cases where users have reported this issue and using an API token was recommended, despite it not working:

This is also quite possible. Regarding the issues you mention:
- #2782 This was closed by the user before we managed to respond. I added a response but never heard back. I might have misread exactly what they were saying, but it's hard to know without them reporting back. They do not mention the use of `AuthType.NONE` anywhere in their posts.
- #3005 This one is still open and being investigated.

---

However, if I understand your situation correctly, and that you can't disable auth for the admin UI without also disabling auth (or at least token parsing) for SDKs, then I think that warrants reviving the discussion. I'll go and investigate a bit and get back to you. 

Thanks for you patience.

kjvalencik:
> However, if I understand your situation correctly, and that you can't disable auth for the admin UI without also disabling auth (or at least token parsing) for SDKs, then I think that warrants reviving the discussion.

This is exactly right.

thomasheartman:
Thanks for the clarification! In that case, let me see what I can find out and get back to you.

thomasheartman:
Alright, so I've done some digging, and I think you're absolutely right. 

So I'd like to thank you for sticking to it and helping me understand what was going on, and I'd like to apologize for closing the PR early 🙇🏼  This is a part of Unleash that doesn't get a lot of attention and that has been around for a while, and this is the first time I've interacted closely with it, so I wasn't aware of the weird edge cases here.

I've consulted with @sighphyre on this, and we agree that `AuthType.<whatever>` should determine the login flow to the admin UI and not affect the client API. Because the client API is dependent on tokens to provide important information such as environment and projects, the tokens will be required for that regardless of `AuthType`. And because of this, they should also obviously be parsed regardless of `AuthType`. In other words, how it works right now is a bug, not a feature.

So with that in mind, I'll reopen the PR. The changes look good in and of themselves, but we'd like to have a test for this before it goes in. Would you be willing to contribute that too?

Thanks again!

kjvalencik:
@thomasheartman Thanks for following up. I can add a test. 👍

thomasheartman:
Hey, @kjvalencik 👋🏼  Did you get anywhere with that tests? It would be good to get this closed if you have the capacity for it. Of course, there's no obligation, so no pressure ☺️

PR #3035, opened 2023-02-02
Refactor: Project Overview fetching

Tymek:
## About the changes
https://linear.app/unleash/issue/1-616/refactor-project-endpoint


frontend/src/hooks/useLoading.ts
```diff
@@ -2,11 +2,13 @@ import { createRef, useLayoutEffect } from 'react';
 
 type refElement = HTMLDivElement;
 
-const useLoading = (loading: boolean) => {
+const useLoading = (loading: boolean, label = 'true') => {
```
kwasniew, line 5: IMO this API is too specific. As a user of useLoading I'd expect (loading: boolean, selector = '[data-loading]') that I can override. label that goes into data-loading selector is not very intuitive from the user's perspective.

frontend/src/component/project/Project/ProjectInfo/ProjectInfo.tsx
```diff
@@ -50,8 +51,11 @@ const ProjectInfo = ({
     const { uiConfig, isEnterprise } = useUiConfig();
     const theme = useTheme();
     const isSmallScreen = useMediaQuery(theme.breakpoints.down('md'));
+    const hasChangeRequestsEnabled = useChangeRequestConfig(id).data?.some(
```
kwasniew, line 54: There's a hook useChangeRequestsEnabled.ts that returns isChangeRequestConfiguredInAnyEnv

frontend/src/component/project/Project/ProjectInfo/ChangeRequestsWidget.tsx
```diff
@@ -11,6 +11,8 @@ import {
 import { useProjectChangeRequests } from 'hooks/api/getters/useProjectChangeRequests/useProjectChangeRequests';
 import { WidgetFooterLink } from './WidgetFooterLink';
 
+const LOADING_LABEL = 'change-requests-widget';
```
kwasniew, line 14: why do we need custom selector here?

kwasniew (changes requested):

frontend/src/component/project/Project/ProjectInfo/ChangeRequestsWidget.tsx
Tymek: Data for change requests is being loaded from another endpoint (that isn't available in unleash-core)

kwasniew (approved):

PR #3037, opened 2023-02-02
Remove console log

gastonfournier:
It seems to have slipped but let me know if it's needed. The thing is it logs in every screen.


sjaanus (approved):
Super!

PR #3038, opened 2023-02-02
fix: use utc to compare time

gastonfournier:
## About the changes
We've found that 60000 was just 1 minute and we wanted it to be 10 minutes so we were missing 1 zero and sometimes in 1 minute we didn't have data. This is still an assumption we'll verify in sandbox

Co-authored-by: Nuno Góis <github@nunogois.com>


nunogois (approved):
🚀

PR #3046, opened 2023-02-03
docs: update Anatomy doc with variants per environment notice

thomasheartman:
## What

This change updates the "variants" section of the Anatomy of Unleash document with more information about how variants work with multiple environments.

At the moment, it states that variants per env are GA from 4.21, but that might need to be changed.

## Why

With the upcoming release of variants per environment, the doc should reflect the current state of things. However, it should also still contain info for those not yet on v4.21


sebastian-bury (approved):
looks good!

PR #3047, opened 2023-02-03
feat: aggregate by label and type

gastonfournier:
## About the changes
We noticed in our instance that when we have same label it shows duplicated in the overview:
image

We're changing to this way of displaying info:
Screenshot from 2023-02-03 11-35-26
when the data in traffic looks like this:
Screenshot from 2023-02-03 11-35-35

### DIFF
--- a/frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
+++ b/frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
@@ -34,7 +34,7 @@ interface INetworkApp {
 
 const asNetworkAppData = (result: RequestsPerSecondSchemaDataResultItem) => {
     const values = (result.values || []) as ResultValue[];
-    const data = values.filter(value => isRecent(value)) || [];
+    const data = values.filter(value => isRecent(value));
     const reqs = data.length ? parseFloat(data[data.length - 1][1]) : 0;
     return {
         label: unknownify(result.metric?.appName),
--- a/frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
+++ b/frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
@@ -5,7 +5,10 @@ import { ConditionallyRender } from 'component/common/ConditionallyRender/Condit
 import { Alert, styled } from '@mui/material';
 import { unknownify } from 'utils/unknownify';
 import { useMemo } from 'react';
-import { RequestsPerSecondSchema } from 'openapi';
+import {
+    RequestsPerSecondSchema,
+    RequestsPerSecondSchemaDataResultItem,
+} from 'openapi';
 import logoIcon from 'assets/icons/logoBg.svg';
 import { formatAssetPath } from 'utils/formatPath';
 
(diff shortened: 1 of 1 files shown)


frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
```diff
@@ -24,30 +27,51 @@ const isRecent = (value: ResultValue) => {
...
     type: string;
 }
 
+const asNetworkAppData = (result: RequestsPerSecondSchemaDataResultItem) => {
+    const values = (result.values || []) as ResultValue[];
+    const data = values.filter(value => isRecent(value)) || [];
+    let reqs = 0;
+    if (data.length) {
+        reqs = parseFloat(data[data.length - 1][1]);
+    }
```
nunogois, line 41: Could be a ternary if it's simpler:

```ts
const reqs = data.length ? parseFloat(data[data.length - 1][1]) : 0
```

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
```diff
@@ -24,30 +27,51 @@ const isRecent = (value: ResultValue) => {
...
-    label?: string;
-    reqs: string;
+    label: string;
+    reqs: number;
     type: string;
 }
 
+const asNetworkAppData = (result: RequestsPerSecondSchemaDataResultItem) => {
+    const values = (result.values || []) as ResultValue[];
+    const data = values.filter(value => isRecent(value)) || [];
```
nunogois, line 37: Don't think the `|| []` is needed 🤔

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
```diff
@@ -24,30 +27,51 @@ const isRecent = (value: ResultValue) => {
...
+    let reqs = 0;
+    if (data.length) {
+        reqs = parseFloat(data[data.length - 1][1]);
+    }
+    return {
+        label: unknownify(result.metric?.appName),
+        reqs: reqs,
```
nunogois, line 44:
```suggestion
        reqs,
```

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
```diff
@@ -24,30 +27,51 @@ const isRecent = (value: ResultValue) => {
...
+        reqs: current.reqs + (acc[groupBy]?.reqs ?? 0),
+    };
+    return acc;
+};
+
 const toGraphData = (metrics?: RequestsPerSecondSchema) => {
-    const results = metrics?.data?.result;
+    const results =
+        metrics?.data?.result?.filter(
+            result => unknownify(result.metric?.appName) !== 'unknown'
```
nunogois, line 66: We're already unknownifying in the map after, should we do it on the filter as well or just check for truthy since it's simpler?

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
```diff
@@ -24,30 +27,51 @@ const isRecent = (value: ResultValue) => {
...
-                return {
-                    label: unknownify(result.metric?.appName),
-                    reqs: reqs.toFixed(2),
-                    type: unknownify(result.metric?.endpoint?.split('/')[2]),
-                };
-            })
-            .filter(app => app.label !== 'unknown')
-            .filter(app => app.reqs !== '0.00') ?? []
+        Object.values(aggregated).filter(
+            app => app.reqs.toFixed(2) !== '0.00'
```
nunogois, line 73: Also feels like it could be done once instead of in the filter and then the graph below.

nunogois (approved):
LG, added some nitpicks

PR #3049, opened 2023-02-03
chore: make sure we run tsc on build

gastonfournier:
## About the changes
This helps give more visibility to errors in the frontend by running `tsc` before building the frontend.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -10,7 +10,7 @@
     "node": ">=14"
   },
   "scripts": {
-    "build": "vite build",
+    "build": "tsc && vite build",
     "start": "vite",
     "start:prod": "vite build && vite preview",
     "start:sandbox": "UNLEASH_API=https://sandbox.getunleash.io/ospro yarn run start",


frontend/package.json
```diff
@@ -10,7 +10,7 @@
     "node": ">=14"
   },
   "scripts": {
-    "build": "vite build",
+    "build": "tsc && vite build",
```
gastonfournier, line 13: The downside of this and why we might want to discard this is because it introduces an overhead and `tsc` is already executed in the `test` script.

gastonfournier: If we are adding it to the build, I think it's a reasonable thing to do.

gastonfournier: :heavy_check_mark:

nunogois: Should be fine, however do (or should) we really build the frontend that often when developing locally? I believe `tsc` is already checked in `ts:check` on the workflows.

gastonfournier: Not in main. It's on the PRs but if we merge a PR that breaks main, main will still report as green, so that might be actually more important

gastonfournier: Should `yarn build` also build the frontend? Or should this be an additional workflow step of GH actions?

gastonfournier: https://github.com/Unleash/unleash/pull/3049/commits/99c567dc5f7b1d60a3ca5c202b20e4b1b7b9c452 maybe?

gastonfournier: That didn't work (you can see the failed builds)

PR #3050, opened 2023-02-06
fix: use the correct hook for permissions and CR

gastonfournier:
## About the changes
When checking for permissions we have 2 methods, one that is not change request aware and one that is. We were using the one that is not aware of CR while the feature we developed is aware of CR.

This PR switches to the correct method and documents the hooks


nunogois (approved):
LGTM

kwasniew (approved):

PR #3052, opened 2023-02-06
Add maintenance toast

sjaanus:
Currently, there was no toast, when maintenance mode was changed. Now user gets nice toast to have better feedback.


kwasniew (approved):
minor comment

PR #3060, opened 2023-02-07
docs: remove the proxy from the sdks overview table

thomasheartman:
## What

This change removes the "unleash proxy" column from the SDK compat table.

## Why

Because the proxy isn't technically an SDK. Having the column there can be misleading to users.

In the latest case, a user pointed out that the table says the proxy supports `isEnabled` with fallback values, but that there is no way to use this via the API (or via client-side sdks) at the moment.

I think we should add a section to the proxy doc about its capabilities instead.

In fact, I think we should do the same thing with the proxy that we did with the SDKs: merge the docs.getunleash content with the github readme and instead source it from there.

### DIFF
--- a/website/docs/reference/sdks/index.md
+++ b/website/docs/reference/sdks/index.md
@@ -51,57 +51,57 @@ If you see an item marked with a ❌ that you would find useful, feel free to re
 
 :::
 
-| Capability | [Java](/docs/generated/sdks/server-side/java.md) | [Node.js](/docs/generated/sdks/server-side/node.md) | [Go](/docs/generated/sdks/server-side/go.md) | [Python](/docs/generated/sdks/server-side/python.md) | [Ruby](/docs/generated/sdks/server-side/ruby.md) | [.NET](/docs/generated/sdks/server-side/dotnet.md) | [PHP](/docs/generated/sdks/server-side/php.md) | [Rust](/docs/generated/sdks/server-side/rust.md) | [Unleash Proxy](../../generated/unleash-proxy.md) |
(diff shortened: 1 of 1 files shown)


sighphyre:
Yeah I think this is the right thing to do. Do we have a plan to put this info somewhere else or does it not really make sense?

ivarconr:
I see your point, but I think we can cause some confusion on what the frontend SDKs will support.

thomasheartman:
@ivarconr 
> I see your point, but I think we can cause some confusion on what the frontend SDKs will support.
Could you elaborate on that? I'm not sure I understand.

thomasheartman:
As mentioned in the description, the content from this repo's proxy docs have been added to the proxy readme in https://github.com/Unleash/unleash-proxy/pull/116. [This commit adds a section about capabilities](https://github.com/Unleash/unleash-proxy/commit/a86f9c56a8c7282ce1a8209c79a28b4775873f24) so that they don't get lost.

ivarconr:
> Could you elaborate on that? I'm not sure I understand.

I'm happy with merging this if we also creates an issue to create a simoil if we create an issue to establish a similar table for frontend SDKs. Some capabilities will be automatically inherited from the proxy (e.g. constraints, ip range etc). 

This would also make it easier to highlight differences in frontend SDK capabilites (bootstrap, impression events etc). 

@thomasheartman

ivarconr (approved):

thomasheartman:
Yeah, I think that's fair (and honestly, it's pretty overdue 😅).  Issue created in #3179

PR #3061, opened 2023-02-08
fix: client query should not include favorute

ivarconr:
Calls to /api/client/features should not return favorites

### DIFF
--- a/src/lib/db/feature-toggle-client-store.ts
+++ b/src/lib/db/feature-toggle-client-store.ts
@@ -211,7 +211,6 @@ export default class FeatureToggleClientStore
             feature.impressionData = r.impression_data;
             feature.enabled = !!r.enabled;
             feature.name = r.name;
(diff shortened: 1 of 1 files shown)


nunogois (approved):
LGTM

chriswk (approved):
LGTM

FredrikOseberg (approved):
I find it counter-intuitive that both the admin API and the client API use the same getAll method. It kind of defeats the purpose of having the client API and admin API separate because there exists the possibility of introducing mistakes in the data by manipulating one shared method. 

It seems like the original architecture would be preserved better if we separated the two explicitly.

thomasheartman:
Just gonna jump in and ask: is there any reason this isn't merged yet?

kwasniew:
This was fixed here: https://github.com/Unleash/unleash/pull/3276

PR #3062, opened 2023-02-08
Poc: calculate etag based on query and latest revison id

ivarconr:
This is very much POC and WIP


src/lib/routes/client-api/feature.ts
```diff
@@ -222,6 +255,57 @@ export default class FeatureController extends Controller {
...
+    async optimal304(
+        req: IAuthRequest,
+        res: Response<ClientFeaturesSchema>,
+    ): Promise<void> {
+        const query = await this.resolveQuery(req);
+
+        const revisionId = await this.eventService.getMaxRevisionId();
+
+        // TODO: We will need to standardize this to be able to implement this a cross languages (Edge in Rust?).
+        const queryHash = hasSum(query);
```
chriswk, line 267: Using a standardized hash would probably be the best solution. We use xxh3-128 in Edge. There's a known implementation for node as well, but it currently requires native libraries.

chriswk: Something like: https://www.npmjs.com/package/xxhash however seems to be native node.

chriswk: But it doesn't implement xxh3  128bit though

ivarconr: In a perfect world it makes sense to hash in a predictable way. But it also have limits, as it needs to be correct, and puts extra constraints to the  hashing algorithm and how the content is parsed and hashed. E.g. a different order of the query args could yield a different hash. Its not enough to agree on the hash-algorithm itself, but also the payload going in to the hashing algorithm. The payload will be a combination of URL query args, and the environment.

Imho each component can (and should ?) define their own way to query-hash. The edge would be responsible for their clients, and the Unleash API should be responsible for its clients. The edge still gets the config version number, and can use that to determine that the client still has the latest config. 

This is a reversible decision. By implementing this without assumption that different components use the same hashing algorithm, we can still iterate on the solution later, if we see clear benefits of doing that. The other way is more tricky, as we might drive the implementation on a conclusion that it needs to be the same.   

Wdyt @chriswk ?

chriswk: Sounds reasonable. Edge is set up to believe/store the hash that comes from upstream (whether that is another Edge, or an Unleash server) for caching. This means that if you change the URL from an Unleash Server to an Edge Server you'll invalidate the cache, but since we (as of yet) cannot change the Unleash URL at runtime for our SDKs that should be safe.

src/lib/db/event-store.ts
```diff
@@ -149,6 +149,16 @@ class EventStore implements IEventStore {
         }
     }
 
+    async getMaxRevisionId(largerThan: number = 0): Promise<number> {
+        const row = await this.db(TABLE)
+            .max('id')
+            .whereNotNull('feature_name')
+            .orWhere('type', 'segment-update')
```
gastonfournier, line 156: I just wonder what types of events to target (or which ones do we exclude?), if we ever expand and add new stuff we might forget to add another `orWhere` which might result in non-errors that would cause weird behaviors...

src/lib/db/event-store.ts
```diff
@@ -149,6 +149,16 @@ class EventStore implements IEventStore {
...
+            .andWhere('id', '>=', largerThan)
+            .first();
```
gastonfournier, line 158: Not sure if the ids are sorted desc or asc, I wonder if `max` is better.

src/lib/routes/client-api/feature.ts
```diff
@@ -199,6 +227,10 @@ export default class FeatureController extends Controller {
         req: IAuthRequest,
         res: Response<ClientFeaturesSchema>,
     ): Promise<void> {
+        if (this.flagResolver.isEnabled('optimal304')) {
+            return this.optimal304(req, res);
+        }
```
gastonfournier, line 232: Maybe we can shadow traffic to this branch and compare to detect if there's a different behavior using one method or the other. We can do it in sandbox at expense of some additional latency

src/lib/routes/client-api/feature.ts
ivarconr: Yes. But I think that's ok for now, the design allows us to normalize the hashing later, if we see extreme value in that

src/lib/services/event-service.ts
```diff
@@ -35,6 +37,21 @@ export default class EventService {
             totalEvents,
         };
     }
+
+    async getMaxRevisionId(): Promise<number> {
+        if (this.revisionId) {
+            return this.revisionId;
+        } else {
+            this.revisionId = await this.eventStore.getMaxRevisionId();
```
gastonfournier, line 45: Nit: maybe
```suggestion
            await this.updateMaxRevisionId();
```

src/lib/routes/client-api/feature.ts
gastonfournier: We can even compare time/size

src/lib/db/event-store.ts
gastonfournier: My understanding is the same, `max` would be similar in cost but might be more meaningful

src/lib/routes/client-api/feature.ts
ivarconr: Oh, in that case the new behavior should be available on a separate path also.

src/lib/db/event-store.ts
gastonfournier: I was tempted to say categorizing our event log with event types :D

src/lib/routes/client-api/feature.ts
```diff
@@ -199,12 +236,32 @@ export default class FeatureController extends Controller {
...
+            try {
+                const { etag } = await this.calculateMeta(query);
+                const [featuresNew] = await this.cachedFeatures2(query, etag);
+                const theDiffedObject = diff(features, featuresNew);
+
+                const message = isEmpty(theDiffedObject)
+                    ? 'The diff is: <Empty>'
+                    : `The diff is: ${JSON.stringify(theDiffedObject)}`;
+
+                this.logger.warn(message);
```
ivarconr, line 259: we should also have a Prometheus metric for this.

src/lib/routes/client-api/feature.ts
```diff
@@ -199,12 +236,32 @@ export default class FeatureController extends Controller {
...
+            return this.optimal304(req, res);
+        }
+
         const query = await this.resolveQuery(req);
 
         const [features, segments] = this.cache
             ? await this.cachedFeatures(query)
             : await this.resolveFeaturesAndSegments(query);
 
+        if (this.flagResolver.isEnabled('optimal304Differ')) {
```
ivarconr, line 249: explain a bit what we expect to see here. 
Old cache is 600ms, while new cache should be faster.

src/lib/routes/client-api/feature.ts
ivarconr: I added a separate toggle called "optimal304Differ". It will allow us to perform actual diffing of the way of caching feature toggles (which end up with a db query every 600ms) and the new way of cache feature toggles, by using revision id as part of cache key and cache for 1 hour. Thus would allow us to validate the new cache implementation behind the scene for real customers.

src/lib/routes/client-api/feature.ts
```diff
@@ -1,7 +1,11 @@
 import memoizee from 'memoizee';
 import { Response } from 'express';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import hasSum from 'hash-sum';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
gastonfournier, line 5: I don't fully understand why we need no-extraneous-dependencies. Is this a leftover from the initial tests?

At the risk of breaking compilation:
```suggestion
import hasSum from 'hash-sum';
```

gastonfournier (approved):
I like the flexibility feature toggles gives us in this scenario, we can both test the diff or enable the new way of calculating 304s without having to deploy more changes and with quick feedback :rocket:

src/lib/db/event-store.ts
ivarconr: lets add that to the todo.

PR #3065, opened 2023-02-08
Feat/project api token permissions

andreas-unleash:
## About the changes

Define and implements Project api token permissions
Assign permissions to existing roles
Adjust UI to support them 
Adjust BE to implement


sjaanus (dismissed):

frontend/src/component/admin/apiToken/RemoveApiTokenButton/RemoveApiTokenButton.tsx
```diff
@@ -11,24 +13,33 @@ import { useContext, useState } from 'react';
 import { Dialogue } from 'component/common/Dialogue/Dialogue';
 import useToast from 'hooks/useToast';
 import useApiTokensApi from 'hooks/api/actions/useApiTokensApi/useApiTokensApi';
+import PermissionIconButton from '../../../common/PermissionIconButton/PermissionIconButton';
```
sjaanus, line 16: Can shorten this

frontend/src/hooks/api/actions/useApiTokensApi/useApiTokensApi.ts
```diff
@@ -12,8 +13,12 @@ const useApiTokensApi = () => {
         propagateErrors: true,
     });
 
-    const deleteToken = async (secret: string) => {
-        const path = `api/admin/api-tokens/${secret}`;
+    const deleteToken = async (secret: string, project?: string) => {
+        const path = formatApiPath(
+            `api/admin/api-tokens/${secret}${
```
sjaanus, line 18: This should be **api/admin/project/:projectId/api-tokens/:secret** rather.

frontend/src/hooks/api/actions/useApiTokensApi/useApiTokensApi.ts
```diff
@@ -1,4 +1,5 @@
 import useAPI from '../useApi/useApi';
+import { formatApiPath } from '../../../../utils/formatPath';
```
sjaanus, line 2: Shorten

frontend/src/hooks/api/actions/useApiTokensApi/useApiTokensApi.ts
sjaanus: Also, since you dont have /project word inside the path, are you sure it works currently?

frontend/src/hooks/api/actions/useApiTokensApi/useApiTokensApi.ts
```diff
@@ -12,8 +13,12 @@ const useApiTokensApi = () => {
         propagateErrors: true,
     });
 
-    const deleteToken = async (secret: string) => {
-        const path = `api/admin/api-tokens/${secret}`;
+    const deleteToken = async (secret: string, project?: string) => {
+        const path = formatApiPath(
+            `api/admin/api-tokens/${secret}${
+                Boolean(project) && project !== '' ? `/${project}` : ''
```
sjaanus, line 19: Boolean(project) should be enough, because Boolean(empty string) should return false and Boolean(non empty string) returns always true

frontend/src/hooks/api/actions/useApiTokensApi/useApiTokensApi.ts
```diff
@@ -25,8 +30,12 @@ const useApiTokensApi = () => {
         }
     };
 
-    const createToken = async (newToken: IApiTokenCreate) => {
-        const path = `api/admin/api-tokens`;
+    const createToken = async (newToken: IApiTokenCreate, project?: string) => {
+        const path = formatApiPath(
```
sjaanus, line 34: Same thing here. It should contain /project/:projectId and correct order.

sjaanus: You should create a separate hook for these probably

src/lib/routes/admin-api/api-token.ts
```diff
@@ -89,11 +92,44 @@ export class ApiTokenController extends Controller {
...
+                }),
+            ],
+        });
+
         this.route({
             method: 'post',
             path: '',
             handler: this.createApiToken,
-            permission: CREATE_API_TOKEN,
+            permission: CREATE_API_TOKEN || CREATE_PROJECT_API_TOKEN,
```
sjaanus, line 115: Not sure this works.

andreas-unleash: bug

frontend/src/hooks/api/actions/useApiTokensApi/useApiTokensApi.ts
andreas-unleash: Yes, that would require a different Controller in the BE - which is an overkill for what we are doing imo.  The endpoint sitting under api-token controlleris `admin/api-tokens/${secret}/{project}` mostly serves to solve the double permission problem

andreas-unleash: Yes it works, the rbac middleware check for `product` or `productId` in the request params
Screenshot 2023-02-09 at 16 50 24

frontend/src/component/admin/apiToken/CopyApiTokenButton/CopyApiTokenButton.tsx
```diff
@@ -1,16 +1,46 @@
...
+        if (token && token.projects && project && permission) {
+            const { projects } = token;
+            for (const tokenProject of projects) {
+                if (!hasAccess(permission, tokenProject)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+    }, [hasAccess, token, project, permission]);
```
FredrikOseberg, line 42: I don't think this is worth memoizing, there are not many iterations here. Could we remove it?

frontend/src/component/admin/apiToken/RemoveApiTokenButton/RemoveApiTokenButton.tsx
```diff
@@ -1,34 +1,66 @@
...
+        if (token && token.projects && project && permission) {
+            const { projects } = token;
+            for (const tokenProject of projects) {
+                if (!hasAccess(permission, tokenProject)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+    }, [hasAccess, token, project, permission]);
```
FredrikOseberg, line 56: Don't think this optimization is worth the cost.

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -0,0 +1,214 @@
...
+            apiTokensSchema.$id,
+            { tokens: serializeDates(projectTokens) },
+        );
+    }
+
+    async createProjectApiToken(
+        req: IAuthRequest,
+        res: Response<ApiTokenSchema>,
+    ): Promise<any> {
+        const createToken = await createApiToken.validateAsync(req.body);
```
FredrikOseberg, line 146: This validation is using joi, which is the old way of validating data. Since we swapped to openAPI validation, we should use that instead so we don't accrue more technical debt.

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -0,0 +1,214 @@
...
+        const { projectId } = req.params;
+        if (!createToken.project) {
+            createToken.project = projectId;
+        }
+
+        if (
+            createToken.projects.length === 1 &&
+            createToken.projects[0] === projectId
+        ) {
```
FredrikOseberg, line 155: What happens if these conditions aren't satisfied here? We return 201 Created without actually performing the action in that case?

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    async deleteProjectApiToken(
+        req: IAuthRequest<ProjectTokenParam>,
+        res: Response,
+    ): Promise<void> {
+        const { user } = req;
+        const { projectId, token } = req.params;
+        const storedToken = (await this.accessibleTokens(user, projectId)).find(
+            (t) => this.tokenEquals(t.secret, token),
```
FredrikOseberg, line 177:
```suggestion
            (currentToken) => this.tokenEquals(currentToken.secret, token),
```

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -0,0 +1,214 @@
...
+        );
+        if (
+            storedToken &&
+            (storedToken.project === projectId ||
+                (storedToken.projects.length === 1 &&
+                    storedToken.project[0] === projectId))
```
FredrikOseberg, line 183: Consider extracting condition to readable name.

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -0,0 +1,214 @@
...
+        if (user.isAPI && user.permissions.includes(ADMIN)) {
+            return allTokens;
+        }
+
+        return allTokens.filter(
+            (token) =>
+                token.type !== ApiTokenType.ADMIN &&
+                (token.project === project || token.projects.includes(project)),
+        );
+    }
```
FredrikOseberg, line 213: Isn't this a service level responsibility?

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    async deleteProjectApiToken(
+        req: IAuthRequest<ProjectTokenParam>,
+        res: Response,
+    ): Promise<void> {
+        const { user } = req;
+        const { projectId, token } = req.params;
+        const storedToken = (await this.accessibleTokens(user, projectId)).find(
```
FredrikOseberg, line 176: Service level responsibility? Seems like this controller is performing business logic around the deletion.

src/migrations/20230208091810-assign-project-api-token-permissions_admin.js
```diff
@@ -0,0 +1,23 @@
...
+                   p.id  as permission_id
+            FROM permissions p
+            WHERE p.permission IN
+                  ('READ_PROJECT_API_TOKEN',
+                   'CREATE_PROJECT_API_TOKEN',
+                   'DELETE_PROJECT_API_TOKEN');
+        `,
+        cb,
+    );
+};
```
FredrikOseberg, line 15: I don't think the admin needs to have these permissions, because the ADMIN permission will short circuit the permissions checks.

src/migrations/20230208093627-assign-project-api-token-permissions-editor.js
```diff
@@ -0,0 +1,20 @@
...
+    );
+};
+
+exports.down = function (db, cb) {
+    db.runSql(
+        `
+  `,
+        cb,
+    );
+};
```
FredrikOseberg, line 20: Should have a down migration

src/migrations/20230208093750-assign-project-api-token-permissions-owner.js
```diff
@@ -0,0 +1,23 @@
...
+    );
+};
+
+exports.down = function (db, cb) {
+    db.runSql(
+        `
+  `,
+        cb,
+    );
+};
```
FredrikOseberg, line 23: Should add a down migration here

src/migrations/20230208093942-assign-project-api-token-permissions-member.js
```diff
@@ -0,0 +1,20 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+            INSERT INTO role_permission (role_id, permission_id)
+            SELECT (SELECT id as role_id from roles WHERE name = 'Member' LIMIT 1),
+                   p.id   as permission_id
+            FROM permissions p
+            WHERE p.permission='READ_PROJECT_API_TOKEN';
```
FredrikOseberg, line 8: Should members not have the ability to CREATE/DELETE? Something to discuss at least.

src/migrations/20230208093942-assign-project-api-token-permissions-member.js
```diff
@@ -0,0 +1,20 @@
...
+    );
+};
+
+exports.down = function (db, cb) {
+    db.runSql(
+        `
+  `,
+        cb,
+    );
+};
```
FredrikOseberg, line 20: Down migration?

frontend/src/component/admin/apiToken/ApiTokenForm/ApiTokenForm.tsx
```diff
@@ -95,23 +94,29 @@ const ApiTokenForm: React.FC<IApiTokenFormProps> = ({
...
     errors,
     clearErrors,
+    scope = 'global',
 }) => {
     const { uiConfig } = useUiConfig();
     const { environments } = useEnvironments();
     const { projects: availableProjects } = useProjects();
 
+    const disableProjectSelection = scope === 'project';
+
```
FredrikOseberg, line 104: Wondering if we should change the architecture so these can be composed together instead of injecting internal state into one component.

Something like:

```
<ApiTokenForm>
   <ProjectSelector />
</ApiTokenForm>
```

FredrikOseberg (changes requested):

src/lib/routes/admin-api/project/api-token.ts
andreas-unleash: makes sense

frontend/src/component/providers/AccessProvider/AccessProvider.tsx
```diff
@@ -79,7 +78,7 @@ const checkPermission = (
     if (
         p.permission === permission &&
         (p.project === project || p.project === '*') &&
-        p.environment === null
+        p.environment == null
```
FredrikOseberg, line 81: Why was this change needed? It's generally considered bad practice to allow type coercion in equal checks in javascript.

src/lib/routes/admin-api/project/api-token.ts
andreas-unleash: These are there to verify that project level tokens have access to 1 project and matches the path param.  I can respond with 409 when these are not met. wdty?

frontend/src/component/providers/AccessProvider/AccessProvider.tsx
andreas-unleash: I wanted the check to be for undefined as well

FredrikOseberg: I would change it to explicit conversion instead: `Boolean(p.environment) === false`

src/lib/routes/admin-api/project/api-token.ts
FredrikOseberg: Maybe 400 Bad Request is more appropriate? If these conditions aren't met, then the request isn't valid for this endpoint.

frontend/src/hooks/api/actions/useProjectApiTokensApi/useProjectApiTokensApi.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const useProjectApiTokensApi = () => {
+    const { makeRequest, createRequest, errors, loading } = useAPI({
+        propagateErrors: true,
+    });
+
+    const deleteToken = async (secret: string, project: string) => {
+        const path = formatApiPath(
+            `api/admin/projects/${project}/api-tokens/${secret}`
+        );
```
FredrikOseberg, line 19: createRequest will call formatApiPath, this will result in a double path.

frontend/src/hooks/api/actions/useProjectApiTokensApi/useProjectApiTokensApi.ts
```diff
@@ -0,0 +1,50 @@
...
+            const res = await makeRequest(req.caller, req.id);
+
+            return res;
+        } catch (e) {
+            throw e;
+        }
+    };
+
+    const createToken = async (newToken: IApiTokenCreate, project: string) => {
+        const path = formatApiPath(`api/admin/project/${project}/api-tokens`);
```
FredrikOseberg, line 32: This will result in a double path.

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -0,0 +1,218 @@
...
+                201,
+                res,
+                apiTokenSchema.$id,
+                serializeDates(token),
+                { location: `api-tokens` },
+            );
+        } else {
+            res.statusMessage =
+                'Project level tokens can only be created for one project';
+            res.status(409);
```
FredrikOseberg: I think 400 is more appropriate.

src/lib/routes/admin-api/project/api-token.ts
andreas-unleash: ok 400 it is

src/migrations/20230208091810-assign-project-api-token-permissions_admin.js
```diff
@@ -0,0 +1,30 @@
...
```
FredrikOseberg, line 15: This whole migration is not neccesary, the admin role only needs the admin permission.

FredrikOseberg (changes requested):
Some of the original feedback is still not adressed.

FredrikOseberg (approved):

PR #3066, opened 2023-02-08
feat: continuous beta version npm release

kwasniew:
## About the changes

Introducing **continuous npm version release**. Every conventional commit to main triggers a new version automatically.

Consequences: 
* for day to day beta releases we don't need to take any manual actions
* for official releases you can still use https://github.com/Unleash/unleash/actions/workflows/publish-new-version.yaml

How it works?
* you squash your commit with a conventional commit **feat:** or **fix:** or **refactor:**
* this action bumps up the beta version
* this action pushes the bumped up version back to the **github repo** and to **npm registry**

Corner cases:
* how will it work after official release of e.g. 4.22.0? new version: 4.23.0-beta.0
* how will it work after previous beta release e.g. 4.23.0-beta.0? new version 4.23.0-beta.1

Why is it using beta.0, beta.1? 
* because `npm version prerelease --preid=beta` gives us the version upgrade logic for free
* in future we may consider adding a timestamp instead of simple version upgrade (cc: @gardleopard )

### DIFF
--- a/.github/workflows/continuous-publish-new-version.yaml
+++ b/.github/workflows/continuous-publish-new-version.yaml
@@ -13,7 +13,7 @@ on:
       - main
 
 # To trigger auto version publish name your squashed commits using conventional notation
-# We support: feat:, fix: and :refactor
+# We support: feat:, fix:m :refactor and :chore
 jobs:
   build:
     if: ${{ startsWith(github.event.head_commit.message, 'feat:') || startsWith(github.event.head_commit.message, 'fix:') || startsWith(github.event.head_commit.message, 'refactor:') || startsWith(github.event.head_commit.message, 'chore:') }}
--- a/.github/workflows/continuous-publish-new-version.yaml
+++ b/.github/workflows/continuous-publish-new-version.yaml
@@ -16,7 +16,7 @@ on:
 # We support: feat:, fix: and :refactor
 jobs:
   build:
-    if: ${{ startsWith(github.event.head_commit.message, 'feat:') || startsWith(github.event.head_commit.message, 'fix:') || startsWith(github.event.head_commit.message, 'refactor:') }}
+    if: ${{ startsWith(github.event.head_commit.message, 'feat:') || startsWith(github.event.head_commit.message, 'fix:') || startsWith(github.event.head_commit.message, 'refactor:') || startsWith(github.event.head_commit.message, 'chore:') }}
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-publish-new-version.yaml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  push:
+    branches:
+      - main
+
+# To trigger auto version publish name your squashed commits using conventional notation
+# We support: feat:, fix: and :refactor
+jobs:
+  build:
+    if: ${{ startsWith(github.event.head_commit.message, 'feat:') || startsWith(github.event.head_commit.message, 'fix:') || startsWith(github.event.head_commit.message, 'refactor:') }}
```
thomasheartman, line 19: Just a thought: is this check case sensitive? I know my editor sometimes auto-capitalizes the start of a sentence, and I can see it being very confusing why some commit doesn't trigger the flow just because it starts with "Fix" instead of "fix".

thomasheartman (commented):
I really like this idea! The execution seems good too; just wondering about the case issue (as mentioned in the comment). And just for clarification, if it's a fix or a refactor after 4.22.0, then it'll be 4.22.1-beta.0, right?

.github/workflows/continuous-publish-new-version.yaml
kwasniew: feat, fix or refactor after 4.22.0 will be 4.23.0-beta.1.  This is the logic from `npm version prerelease --preid=beta`. I think it makes sense in the continuous world we strive for. If you want to fix something on the official 4.22.0 you should work on a branch off the 4.22.0 tag without blocking people who are already contributing to the next 4.23 official release candidate.

gardleopard: Actually two things in one here. Ill just address 4.22.0 -> fix/refactor. After a release is cut on the main branch its normal that the next commits go to the next version. If we have fixes for a patch release I think it makes sense to use a release_branch and cherry pick commits from main that you want in the patch version.

gardleopard (approved):
I love it!

.github/workflows/continuous-publish-new-version.yaml
kwasniew: the lowercase function from what I know is currently missing in GH actions and if often requested. I prefer to keep the code free of the complex workarounds for now. But we can iterate on this idea next.

kwasniew: One more comment about versioning. We could follow traditional conventional commits without alpha/beta and always upgrade major/minor/patch based on conventional commits. this would work if we didn't distribute the package to many clients and only self hosted. Due to the nature of internal releases and open source client hosted releases we make a tradeoff. We continuously bump up the beta version for our internal continuous delivery purposes. But for the client facing stuff we make a manual release on a regular cadence. This GH action only codifies things that we already do manually.

PR #3068, opened 2023-02-08
fix: x axis not displaying the right date

gastonfournier:
## About the changes
The x-axis is showing 10:31 which is not the right date. If you look at the tooltips the date is correct but not the x-axis

How it looks:
Screenshot from 2023-02-08 15-46-45

How it will look like:
image


nunogois (approved):
LGTM

PR #3069, opened 2023-02-08
fix: unknown in network view

gastonfournier:
## About the changes
This replicates the logic of unknowify library to filter out undefined


nunogois (approved):
LGTM

PR #3070, opened 2023-02-09
feat: remove CI build trigger

gardleopard:
We do not need this trigger anymore as we auto-release beta versions from this repository and auto bump the dependency in unleash-enterprise.


sighphyre (approved):
g

PR #3071, opened 2023-02-09
refactor: move repo sync signal after npm publish

kwasniew:
## About the changes
Moving repo sync signal after npm publish


gardleopard (approved):
LGTM

PR #3072, opened 2023-02-09
fix: make sure we have a user in event store

gastonfournier:
## About the changes
Spotted some issues in logs:
```json
{
  "level":"warn",
  "message":"Failed to store \"feature-environment-variants-updated\" event: error: insert into \"events\" (\"created_by\", \"data\", \"environment\", \"feature_name\", \"pre_data\", \"project\", \"tags\", \"type\") values (DEFAULT, $1, $2, $3, $4, $5, $6, $7) returning \"id\", \"type\", \"created_by\", \"created_at\", \"data\", \"pre_data\", \"tags\", \"feature_name\", \"project\", \"environment\" - null value in column \"created_by\" violates not-null constraint",
  "name":"lib/db/event-store.ts"
}
```

In all other events we're doing the following: https://github.com/Unleash/unleash/blob/b7fdcd36c08fcd9461fef930767606962e73d59c/src/lib/services/segment-service.ts#L80

So this is just mimicking that to quickly release a patch, but I'll look into a safer (type-checked) solution so this problem does not happen again

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -1331,7 +1331,7 @@ class FeatureToggleService {
         featureName: string,
         environment: string,
         newVariants: IVariant[],
-        createdBy: string,
+        user: User,
         oldVariants?: IVariant[],
     ): Promise<IVariant[]> {
         await variantsArraySchema.validateAsync(newVariants);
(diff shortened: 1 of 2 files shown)


nunogois (approved):
LGTM

PR #3076, opened 2023-02-09
chore: make event store accept IUser

gastonfournier:
## About the changes
Currently, we need to remember of using the email or else the username of a user when storing into EventStore, because we don't have [strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks), it's error-prone. Fix for a production issue: #3072

I tried enabling strictNullChecks in our [tsconfig file](https://github.com/Unleash/unleash/blob/354cfa41352aa8a8739b0d9f833b146483952317/tsconfig.json#L30), but it reports 509 issues in 143 files.

This means we also need to check for undefined here: https://github.com/Unleash/unleash/blob/479d2f14bf0803818b30b1fae3d931307ed956ef/src/lib/types/events.ts#L144-L147


src/lib/types/events.ts
```diff
@@ -135,9 +136,15 @@ class BaseEvent implements IBaseEvent {
...
-    constructor(type: string, createdBy: string, tags: ITag[] = []) {
+    /**
+     * @param createdBy accepts a string for backward compatibility. Prefer using IUser for standardization
+     */
+    constructor(type: string, createdBy: string | IUser, tags: ITag[] = []) {
         this.type = type;
-        this.createdBy = createdBy;
+        this.createdBy =
+            typeof createdBy === 'string' || typeof createdBy === 'undefined'
+                ? createdBy
```
gastonfournier, line 146: This maintains the same behaviour as before, if `createdBy` is passed as undefined we propagate that (which should break DB constraints and display an error)

chriswk (approved):
:)

src/lib/types/events.ts
```diff
@@ -135,9 +137,15 @@ class BaseEvent implements IBaseEvent {
...
+    /**
+     * @param createdBy accepts a string for backward compatibility. Prefer using IUser for standardization
+     */
+    constructor(type: string, createdBy: string | IUser, tags: ITag[] = []) {
         this.type = type;
-        this.createdBy = createdBy;
+        this.createdBy =
+            typeof createdBy === 'string'
+                ? createdBy
+                : extractUsernameFromUser(createdBy);
```
gastonfournier, line 148: Reusing an existing function: https://github.com/Unleash/unleash/blob/03c34995466ab8f13f188c588b28c8d74c819c49/src/lib/util/extract-user.ts

PR #3077, opened 2023-02-09
chore: attempt to make tsconfig stricter

gastonfournier:
## About the changes
Not being able to enable strictNullChecks: https://github.com/Unleash/unleash/pull/3076 but at least we can start enabling other flags


chriswk (approved):
If this works, I love it :)

gastonfournier:
> > If this works, I love it :)
> 
> I'll rebase just to check with the latest main

Comment was for https://github.com/Unleash/unleash/pull/3076

thomasheartman (approved):

PR #3079, opened 2023-02-10
chore: add bulk endpoint for metrics and app registration for edge

gastonfournier:
## About the changes
Implementation of bulk metrics and registration endpoint. This will be used by edge nodes to send all collected information.

Types around metrics were improved and `IClientApp.bucket` with type `any` is no longer needed


src/lib/types/model.ts
```diff
@@ -342,7 +341,7 @@ export interface IMetricCounts {
 export interface IMetricsBucket {
     start: Date;
     stop: Date;
-    toggles: IMetricCounts;
+    toggles: { [key: string]: IMetricCounts };
```
gastonfournier, line 344: This type was not used and it was not correct. If we look at the schema, toggles is an object with each key representing the toggle name and the value being the metrics count

sighphyre (approved):
LGTM

PR #3081, opened 2023-02-10
fix: enabling more tests with strict schema validation 

kwasniew:
## About the changes

Run more tests with strict response schema validation and fixing health schema accordingly.


src/test/e2e/api/openapi/__snapshots__/openapi.e2e.test.ts.snap
```diff
@@ -1323,6 +1327,13 @@ exports[`should serve the OpenAPI spec 1`] = `
             "example": false,
             "type": "boolean",
           },
+          "strategies": {
```
kwasniew, line 1330: added this as some clients may depend on this legacy structure

sjaanus (approved):

src/test/e2e/api/admin/constraints.e2e.test.ts
```diff
@@ -9,7 +9,13 @@ const PATH = '/api/admin/constraints/validate';
 
 beforeAll(async () => {
     db = await dbInit('constraints', getLogger);
-    app = await setupApp(db.stores);
+    app = await setupAppWithCustomConfig(db.stores, {
+        experimental: {
+            flags: {
+                strictSchemaValidation: true,
```
gastonfournier, line 15: :raised_hands:

src/lib/openapi/spec/health-overview-schema.ts
```diff
@@ -23,6 +23,7 @@ export const healthOverviewSchema = {
         },
         description: {
             type: 'string',
+            nullable: true,
```
gastonfournier, line 26: I guess this is different than not being required. Right?

gastonfournier (approved):
Nice!

src/lib/openapi/spec/health-overview-schema.ts
kwasniew: this is when description is there and set to null. We were getting those values in the value space, so adding them to the type space

PR #3094, opened 2023-02-13
chore: gradually reduce null-check errors

gastonfournier:
## About the changes

In order to move us towards enabling `strictNullChecks` we'd want to have a way of gradually enabling this without having to fix all errors at once, this will force us to start reducing the number of null check issues.

This new workflow:
1. [Checks out the current branch and main into 2 different folders](https://github.com/Unleash/unleash/pull/3094/files#diff-068f2ace1d1d2e773fb5e4240c83ccab251556fd5524fe13847122878e40da3bR15-R23)
2. Uses the **same** script `gradual-strict-null-checks.sh` (from the current branch) [against each folder in parallel](https://github.com/Unleash/unleash/pull/3094/files#diff-068f2ace1d1d2e773fb5e4240c83ccab251556fd5524fe13847122878e40da3bR34-R38) to count the number of errors if `strictNullChecks` was enabled
3. If the number of potential errors in the current branch is higher than the number of potential errors in main [it fails](https://github.com/Unleash/unleash/pull/3094/files#diff-068f2ace1d1d2e773fb5e4240c83ccab251556fd5524fe13847122878e40da3bR41-R46)

As an example, a [new issue was introduced in this PR](https://github.com/Unleash/unleash/pull/3094/commits/753f57223c2107278dd7ee387444847e5cc4496a) (and then [reverted](https://github.com/Unleash/unleash/pull/3094/commits/e4deb62965bdc12b22c2a78a85588b237943483a)), so we can test the build failure:
https://github.com/Unleash/unleash/actions/runs/4163632636/jobs/7204268519#step:5:10

## Discussion points
This could be a non-mandatory check, just advising, or even adding a comment in the PR. It might be good to start with a non-strict check, but at the same time we can decide to make it non-strict if a problem appears

In some situations, an additional null check error might require us to fix a bunch of them, increasing the time to deliver. In these cases we can suppress an individual line with `// @ts-ignore: Object is possibly 'null'.` although might defeat the purpose of this workflow

### DIFF
--- a/.github/workflows/gradual-strict-null-checks.yml
+++ b/.github/workflows/gradual-strict-null-checks.yml
@@ -26,4 +26,10 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'yarn'
-      - run: ./current/scripts/gradual-strict-null-checks.sh
+          cache-dependency-path: |
+            current/yarn.lock
+            main/yarn.lock
+      - run: ./current/scripts/gradual-strict-null-checks.sh ./current > ./current-count
+      - run: ./current/scripts/gradual-strict-null-checks.sh ./main > ./main-count
+      - run: cat ./current-count
+      - run: cat ./main-count
--- a/.github/workflows/gradual-strict-null-checks.yml
+++ b/.github/workflows/gradual-strict-null-checks.yml
@@ -12,10 +12,18 @@ jobs:
         node-version: [16.x]
 
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout current branch
+        uses: actions/checkout@v3
+        with:
+          path: current
+      - name: Checkout main branch
+        uses: actions/checkout@v3
+        with:
+          ref: main
+          path: main
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'yarn'
-      - run: ./scripts/gradual-strict-null-checks.sh
+      - run: ./current/scripts/gradual-strict-null-checks.sh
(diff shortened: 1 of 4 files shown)


scripts/gradual-strict-null-checks.sh
```diff
@@ -0,0 +1,31 @@
...
+git checkout "${CURRENT_BRANCH}"
+git stash pop
+
+# count errors
+BRANCH_ERRORS=$(yarn 2> /dev/null | grep "Found [0-9]* errors" | sed 's/Found \(.*\) errors in .* files./\1/')
+
+# reset tsconfig changes
+git checkout -- "${SCRIPT_DIR}/../tsconfig.json"
+
+if [ ${BRANCH_ERRORS} -gt ${MAIN_ERRORS} ]; then
```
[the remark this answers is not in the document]
gastonfournier: @sonatype-lift ignore

chriswk (commented):
I like the idea. I'm a +1 on this

.github/workflows/gradual-strict-null-checks.yml
```diff
@@ -0,0 +1,46 @@
...
+      - run: |
+          ./current/scripts/gradual-strict-null-checks.sh ./current > ./current-count &
+          pid1=$!
+          ./current/scripts/gradual-strict-null-checks.sh ./main > ./main-count &
+          pid2=$!
+          wait $pid1 && wait $pid2
+          MAIN=$(cat ./main-count)
+          CURRENT=$(cat ./current-count)
+          if [ $CURRENT -gt $MAIN ]; then
+            echo "The PR is increasing the number of null check errors from ${MAIN} to ${CURRENT}. Consider fixing them before merging"
```
gastonfournier, line 42:
```suggestion
            echo "The PR is increasing the number of null check errors from ${MAIN} to ${CURRENT}. Check if your branch is up-to-date and consider fixing them before merging"
```

gastonfournier: Also having a diff might be useful

PR #3101, opened 2023-02-14
refactor: docker wait for db with condition

Tymek:
## About the changes
Change script to https://docs.docker.com/compose/compose-file/#depends_on `service_healthy`.

Closes [1-598/docker-composeyml-cant-be-run-on-its-own](https://linear.app/unleash/issue/1-598/docker-composeyml-cant-be-run-on-its-own)


sjaanus (approved):
LGTM

PR #2453, opened 2022-11-16
Variants per environment (frontend)

nunogois:
image
## About the changes
https://linear.app/unleash/issue/2-425/variants-crud-new-environment-cards-with-tables-inside-add-edit-and

Basically created parallel components for the **variants per environments** feature, so both flows should work correctly depending on the feature flag state. Some of the duplication means that cleanup should be straightforward - Once we're happy with this feature it should be enough to delete `frontend/src/component/feature/FeatureView/FeatureVariants/FeatureVariantsList` and do some little extra cleanup.

I noticed we had some legacy-looking code in variants, so this involved *some* rewriting of the current variants logic. Hopefully this new code looks nicer, more maintainable, and more importantly **doesn't break anything**.

Relates to [roadmap](https://github.com/orgs/Unleash/projects/10) item: #2254 

### Important files
Everything inside the `frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants` folder.

PR #3110, opened 2023-02-14
refactor: remove old components after variants per env

nunogois:
https://linear.app/unleash/issue/2-427/clean-up-previous-components-from-the-old-ui

Major clean up after we fully migrated to variants per environment, removing old components. 
You can read more about it in the original PR: https://github.com/Unleash/unleash/pull/2453


gastonfournier (approved):
-1,813!!! :partying_face:

PR #3119, opened 2023-02-15
chore: moved GH action script to the yaml file

gastonfournier:
## About the changes
- Actions are not counting successfully: https://github.com/Unleash/unleash/actions/runs/4181713524/jobs/7243938807#step:5:17
- This should fix an issue with running yarn concurrently (which might be leading to the zero counts)
- Move the code to a GH action so it's easier to debug 
- Added a diff of the reported errors to help the PR author identify the potential cause


.github/workflows/gradual-strict-null-checks.yml
```diff
@@ -34,16 +34,30 @@ jobs:
             current/yarn.lock
             main/yarn.lock
       # intentionally use the same script from current branch against both repositories
-      - run: |
-          ./current/scripts/gradual-strict-null-checks.sh ./current > ./current-count &
+      - name: Compare errors if enabling strictNullChecks
+        run: |
+          set -x
```
nunogois, line 39: Should we keep this?

```suggestion

```

nunogois (approved):
LGTM

PR #3120, opened 2023-02-15
feat: export project level feature toggles

kwasniew:
## About the changes

In addition to global toggles export we now support project level export
Screenshot 2023-02-15 at 11 16 49


sjaanus (approved):
LGTM

PR #3122, opened 2023-02-15
Source proxy and Edge docs from GitHub

thomasheartman:
## What

The main purpose of this PR is to

1. Delete the proxy docs in this repo and replace them with the proxy's GitHub readme.
2. Add the docs for Unleash Edge.

### Detailed change description

This PR contains a lot of small changes in a large number of files. To make it easier to get an overview, here's a detailed description of what happens where:

#### In the `website/docs`directory

Except for the deletion of the proxy doc, all changes in this directory are rewriting internal links, so that they point to the newly generated document instead.

#### `package.json` and `yarn.lock`

When including the documentation for Edge, we also want to render the mermaid diagrams it uses. Docusaurus supports this via a plugin. All changes in these files are related to installing that plugin.

#### `docusaurus.config.js`

There's two types of changes in this file:

1. Mermaid-related changes: we ask docusaurus to render mermaid in markdown files and add the plugin

2. Document generation. There's some rewrites to the sdk doc generation plus an entirely new section that generates docs for Edge and the proxy

#### `sidebars.js`

Two things:

1. Add the edge docs
2. Move both the Edge and the proxy docs up a level, so that they're directly under "reference docs" instead of nested inside "unleash concepts".

#### In the `website/remote-content` directory

These are the remote content files. Previously, all of this lived only in a `readme-fns.js` file, but with the introduction of Edge and proxy docs, this has been moved into its own directory and refactored into three files (`shared`, `sdks`, `edge-proxy`).

#### `custom.css`

Style updates to center mermaid diagrams and provide more space around them.

#### In `static/img`

The image files that were included in the proxy doc and that have been deleted.

## Why

For two reasons:

1. Reduce duplication for the proxy. Have one source of truth.
2. Add docs for edge.

## Discussion points and review wishes

This is a big PR, and I don't expect anyone to do a line-by-line review of it, nor do I think that is particularly useful. Instead, I'd like to ask reviewers to:

1. Visit the [documentation preview](https://unleash-docs-git-docs-source-proxy-gh-unleash-team.vercel.app/reference/unleash-proxy) and have a look at both the proxy docs and the Edge docs. Potentially have a look at the SDK docs too to verify that everything still works.

2. Consider whether they think moving the proxy and edge docs up a level (in the sidebar) makes sense.

3. Let me know what slug they'd prefer for the Edge docs. I've gone with `unleash-edge` for now (so that it's `docs.getunleash.io/reference/unleash-edge`), but we could potentially also just use `edge`. WDYT?

4. Read through the detailed changes section.

5. Let me know if they have any other concerns or questions.

## Screenies

The new proxy doc:

image

The new edge doc:

image


sighphyre (approved):
> Visit the [documentation preview](https://unleash-docs-git-docs-source-proxy-gh-unleash-team.vercel.app/reference/unleash-proxy) and have a look at both the proxy docs and the Edge docs. Potentially have a look at the SDK docs too to verify that everything still works.

Yes, this looks good to me, better than good actually, these render really really nicely and I see the Rust docs table got fixed :)

> Consider whether they think moving the proxy and edge docs up a level (in the sidebar) makes sense.

I agree with where you've put them in this. They're deserve their own place and it should be more visible 

> Let me know what slug they'd prefer for the Edge docs. I've gone with unleash-edge for now (so that it's docs.getunleash.io/reference/unleash-edge), but we could potentially also just use edge. WDYT?

I like unleash-edge to distinguish it from a cloud edge. We call it edge a lot internally but as soon as we start talking about running it on the edge, it becomes unleash-edge so this looks good to me 

> Read through the detailed changes section.

LGTM

> Let me know if they have any other concerns or questions.

I think this looks fine. We've started to slip in some code here so long term if this grows in complexity, some tests might be nice but I think we're far away from being at the point we need that

PR #3054, opened 2023-02-07
chore: Move rps endpoint to pro

gastonfournier:
## About the changes
Network view is part of our cloud offering.

PR #3124, opened 2023-02-15
Network view not experimental

gastonfournier:
## About the changes
Promoted experimental networkView flag into a configuration that relies on prometheusApi being configured.

Also, a follow-up on https://github.com/Unleash/unleash/pull/3054 moving this code to enterprise because it doesn't make sense to maintain this code in OSS where it's not being used.


nunogois (approved):
LGTM

PR #3131, opened 2023-02-16
Project stats feedback

Tymek:
## About the changes
image

https://linear.app/unleash/issue/1-694/widgets-explanation-plausible-buttons

### DIFF
--- a/frontend/src/component/project/Project/ProjectStats/Feedback.tsx
+++ b/frontend/src/component/project/Project/ProjectStats/Feedback.tsx
@@ -1,0 +1,61 @@
+import { useState, VFC } from 'react';
+import { Box, Paper, Button } from '@mui/material';
+import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
+
+interface IFeedbackProps {
+    id: string;
+}
+
+export const Feedback: VFC<IFeedbackProps> = ({ id }) => {
+    const [selected, setSelected] = useState<'yes' | 'no'>();
+    const { trackEvent } = usePlausibleTracker();
+
+    const onTrackFeedback = (value: 'yes' | 'no') => {
+        setSelected(value);
+        trackEvent('project_overview', {
+            props: {
+                eventType: id,
+                wasHelpful: value === 'yes',
+            },
+        });
+    };
+
+    return (
+        <Paper
+            elevation={0}
+            sx={{
+                background: theme => theme.palette.neutral.light,
+                padding: theme => theme.spacing(1.5, 2),
+                marginTop: theme => theme.spacing(1.5),
+            }}
+        >
+            Was this information useful to you?
+            <Box
+                sx={{
+                    display: 'flex',
+                    gap: theme => theme.spacing(1),
+                    marginTop: theme => theme.spacing(0.5),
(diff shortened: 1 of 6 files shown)


frontend/src/component/project/Project/ProjectStats/HelpPopper.tsx
```diff
@@ -0,0 +1,81 @@
...
+                            <Close
+                                sx={{
+                                    fontSize: theme =>
+                                        theme.typography.body1.fontSize,
+                                }}
+                            />
+                        </IconButton>
+                        {children}
+                        <ConditionallyRender
+                            condition={isPro()}
```
FredrikOseberg, line 73: Or enterprise?

FredrikOseberg: We could use the uiConfig flag "T" to determine whether or not we are in a cloud setting. The tracking flag is only true if we are hosted in our domain.

frontend/src/component/project/Project/ProjectStats/HelpPopper.tsx
```diff
@@ -0,0 +1,81 @@
...
+            >
+                <ClickAwayListener onClickAway={onClose}>
+                    <Paper
+                        sx={theme => ({
+                            padding: theme.spacing(3, 3),
+                            maxWidth: '350px',
+                            borderRadius: `${theme.shape.borderRadiusMedium}px`,
+                            border: `1px solid ${theme.palette.neutral.border}`,
+                            fontSize: theme.typography.body2.fontSize,
+                        })}
```
FredrikOseberg, line 57: I suspect @nunogois would like these to be Styled components instead of inline.

frontend/src/component/project/Project/ProjectStats/Feedback.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+    const onTrackFeedback = (value: 'yes' | 'no') => {
+        setSelected(value);
+        trackEvent('project_overview', {
+            props: {
+                eventType: id,
+                wasHelpful: value === 'yes',
+            },
+        });
+    };
```
FredrikOseberg, line 21: Should we record this in the DB/localstorage? Or should we accept that a user can provide the feedback multiple times?

The drawback of localstorage is that we clear it on logout.

frontend/src/component/project/Project/ProjectStats/Feedback.tsx
```diff
@@ -0,0 +1,61 @@
...
+            }}
+        >
+            Was this information useful to you?
+            <Box
+                sx={{
+                    display: 'flex',
+                    gap: theme => theme.spacing(1),
+                    marginTop: theme => theme.spacing(0.5),
+                }}
+            >
```
FredrikOseberg, line 39: Extract to Styled?

FredrikOseberg (changes requested):
Nice! Added some clarifications and suggestions. The only blocker is the condition for showing the feedback, I think we need to be more explicit here.

FredrikOseberg (approved):
LGTM

PR #3133, opened 2023-02-16
chore: comment outcome of null check action

gastonfournier:
## About the change
Post a comment in the Pull Request rather than echo in the console in case of an error. Also, [reduced the null check errors by 26](https://github.com/Unleash/unleash/actions/runs/4196924954/jobs/7278564622#step:5:40)

### DIFF
--- a/.github/workflows/gradual-strict-null-checks.yml
+++ b/.github/workflows/gradual-strict-null-checks.yml
@@ -40,7 +40,7 @@ jobs:
           URL: ${{ github.event.pull_request.comments_url }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          comment() {
+          comment () {
             curl -X POST $URL \
               -H "Content-Type: application/json" \
               -H "Authorization: token $GITHUB_TOKEN" \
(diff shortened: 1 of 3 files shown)


gastonfournier:
> After enabling strictNullChecks this PR would be increasing the number of null check errors from 537 to 538. Make sure your branch is up-to-date with main and check the diff in the console output to gather more details

This comment is to prove that the comment works after [bfefbc9](https://github.com/Unleash/unleash/pull/3133/commits/bfefbc9f82f54e3cc26f8762322541fcdd61054e)

thomasheartman (approved):
Love it 🙌🏼  My only question is whether we'd want the "The PR has $CURRENT null check errors against $MAIN in main. You're good to go!" message to be commented as well? But you're probably right here, and it's something we can easily change later 😄

gastonfournier:
> Love it 🙌🏼 My only question is whether we'd want the "The PR has $CURRENT null check errors against $MAIN in main. You're good to go!" message to be commented as well? But you're probably right here, and it's something we can easily change later smile

At first, I had added it, but then I realized it would be annoying unless we manage to "update the current comment", because currently I'm posting a new comment every time this runs, and if you're adding commits to the PR you'd get a comment on every one of them (I mean, it's good having someone tapping at your back say "Good job", but it might be too much doing it on every new push to your branch :sweat_smile: )

PR #3139, opened 2023-02-17
docs: network-view documentation

### LINKED ISSUE #2603: 🔥 feat: Show API traffic inside Unleash
### Description
Make it easy for instance administrator to debug the amount of traffic being sent to the Unleash instance, and where it is coming from.

image

### Availability
Pro and Enterprise

gastonfournier:
## About the changes
This documentation describes the basics of our network view pages

Closes #2603


website/docs/reference/network.md
```diff
@@ -0,0 +1,31 @@
+---
+title: Network
```
thomasheartman, line 2: Maybe this should be "Network view"? Wdyt? Just "Network" doesn't really tell you all that much. Network overview? 

```suggestion
title: Network view
```

website/docs/reference/network.md
```diff
@@ -0,0 +1,31 @@
...
+---
+
+:::info Availability
+
+The Unleash network view was released in Unleash 4.21 for our paid plans.
```
thomasheartman, line 7: Maybe reword it into something like this? (absolutely non-blocking, though)

```suggestion
The Network view was released in Unleash 4.21. It is available to Pro and Enterprise users.
```

website/docs/reference/network.md
```diff
@@ -0,0 +1,31 @@
...
+## Network traffic
+
+Network traffic can help you identify misconfigured apps, surges in traffic or other anomalies.
+
+![Network traffic showing 3 sources and unregistered apps as unknown](/img/network-traffic.png)
+
+# Prerequisites
+This features uses a Prometheus-like API to query data produced by Unleash server. Our managed instances receive this a part of our service.
+
+It's most useful when having the toggle `responseTimeWithAppName` enabled (for hosted customers we enabled it for all customers), because it will allow you to identify the applications generating the traffic.
```
thomasheartman, line 31: I'm having a little trouble understanding this section. 

> Our managed instances receive this a part of our service.

What about non-managed instances?

> It's most useful when having the toggle `responseTimeWithAppName` enabled (for hosted customers we enabled it for all customers), because it will allow you to identify the applications generating the traffic.

What is this toggle? Where is it available?

thomasheartman (changes requested):
This looks very nice; really appreciate the images 💯  But I have a few comments I'd like addressed. 

About the network: could we say anything about how the stats are calculated or something? How does the Unleash server get this data? Is it based on ... incoming SDKs? Proxy? Metrics calls?  Maybe that's redundant?

website/docs/reference/network.md
gastonfournier: Network view is fine, it's an overarching term over "Network overview" (which is one tab) and "Network traffic"

website/docs/reference/network.md
gastonfournier: > What about non-managed instances?

Spot on! This is one of the things I wanted to get feedback on! 

Non-managed are Open Source self-hosted (you don't get Network view), and Enterprise self-hosted. For the latter we should say something: if you need help setting this up reach out or specify that they need an additional environment variable `PROMETHEUS_API=http://localhost:9090` for Prometheus or `PROMETHEUS_API=http://localhost:8481/select/0:0/prometheus`  for VictoriaMetrics


> > It's most useful when having the toggle responseTimeWithAppName enabled (for hosted customers we enabled it for all customers), because it will allow you to identify the applications generating the traffic.
> 
> What is this toggle? Where is it available?

I don't know if we have other configurations like this. We have an environment variable `UNLEASH_EXPERIMENTAL_RESPONSE_TIME_WITH_APP_NAME` (we should probably rename it as non-experimental), which defaults to false. We can easily switch this to true, but still we have a feature toggle (which is resolved by our feature toggle resolver) that is helpful to disable this from misbehaving customers. 

Maybe let's talk about expectations:
1. All customers should have it enabled by default.
2. If some customer if using randomly generated app-names, the combination of having it enabled and the random app names might cause their storage in Prometheus to spike. 

I think there are 2 options, if we consider the risk of 2 is high and we don't want that, we can have the feature disabled and only enabled consciously by customers (therefore we need an environment variable), or we have it enabled and add a warning here (we expect everyone to have alerts on disk usage, but...)

gastonfournier: I'll be removing this last part because #3161 should make the toggle disappear (or be enabled by default)

thomasheartman: Thanks. I'm trying to form my thoughts on this, so apologies if this isn't quite coherent.

I'm not sure "prerequisites" is the right title for the section. For several reasons:
1. It's not a how-to guide
2. It doesn't actually explicitly state what it is you need anywhere

What does "prometheus-like" mean in this case? The links / env vars you list make it seem like it **is** a prometheus API? Is it documented in our API docs anywhere? Or is it just an API that conforms to the Prometheus interface?

What does the `UNLEASH_EXPERIMENTAL_RESPONSE_TIME_WITH_APP_NAME` feature do? What happens when you turn it on and off?

How about moving it up to after the introductory section and rewording it to something like this:

```suggestion
## Data source

The network view sources its data from Unleash's Prometheus API, which is controlled by the environment variable `PROMETHEUS_API`, as explained in the configuring Unleash document (@gastonfournier do we have documentation on this somewhere?). 
 
## Uses

The network view is most useful when you enable response times with app name in metrics ... [I need more information to understand what we're trying to say here]
```

But I obviously still do not quite understand what you're trying to convey in this section, so any input is much appreciated.

thomasheartman: > The complexity is why we offer this as a cloud offering, but if you're self-hosted and still want it, we can help you out, but it's not straightforward.

Right. In that case, it sounds like this might be worth writing a how-to guide for? (As a separate document)

And just so I'm clear: when you say to set the prometheus API env var: is that for the Unleash process? Does Unleash try to expose an API on that URL, or will Unleash **use** that URL to fetch data for this?

> Regarding this other topic, it allows you to show the traffic by app name, otherwise, you'll just receive a line with `unknown` which is not so valuable. 

This would definitely be worth documenting. Both what the default is and what it means when the metrics say `unknown`.

gastonfournier: I've tried to explain some of the details of `PROMETHEUS_API` configuration, but I still have my concerns about it, especially because it's not part of the standard Unleash setup. But, at least it's under a `Data source` section as an appendix of this documentation. I've noted down that `How to set up Prometheus to collect metrics from that API is outside of the scope of this document.`

wrt. `unknown`,  I've added a note under `Network traffic` which is where you'd see it.

gastonfournier:
Closing in favor of #3183

PR #3142, opened 2023-02-17
docs: add notes to constraints and segment docs about large values

thomasheartman:
## What

This PR adds a note to the constraint and the segment docs about using a large number of constraint values.

## Why

This is a fairly common question in slack, and there is no clear documentation around this at the moment. As such, it's likely that people do this a fair bit without telling us. We don't think this is the right approach, so adding this bit of documentation should hopefully make it a little clearer.

### DIFF
--- a/website/docs/reference/strategy-constraints.md
+++ b/website/docs/reference/strategy-constraints.md
@@ -180,6 +180,16 @@ If you set a context field to a value that the SDKs cannot parse correctly for a
 In other words: if you have a strategy constraint operator that expects a number, such as `NUM_GT`, but you set the corresponding context field to a string value, then the expression will be false: `"some string"` is **not** greater than `5`.
 This value can still be negated as explained in [the section on negating values](#constraint-negation).
 
+## Constraint limitations (or "how many user IDs can I add to a constraint") {#limitations}
+
+When using a constraint operator that accepts a list of values, it might be tempting to add a large number of values to that list. However, we strongly advise you **not** to do that: Unleash is **not** a database, and is not intended to store large amounts of data. Instead you should try and find a different way to achieve what you want.
+
+Because Unleash's server-side SDKs fetch the full feature toggle configuration from Unleash, every value that you add to that constraint value list will increase the payload size. For small numbers, this isn't an issue, but as the list grows, so will the payload, and so will the time and processing power used by the SDK to evaluate the feature.
+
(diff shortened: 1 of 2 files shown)


website/docs/reference/strategy-constraints.md
```diff
@@ -180,6 +180,16 @@ If you set a context field to a value that the SDKs cannot parse correctly for a
 In other words: if you have a strategy constraint operator that expects a number, such as `NUM_GT`, but you set the corresponding context field to a string value, then the expression will be false: `"some string"` is **not** greater than `5`.
 This value can still be negated as explained in [the section on negating values](#constraint-negation).
 
+## Constraint limitations (or "how many user IDs can I add to a constraint") {#limitations}
+
+When using a constraint operator that accepts a list of values, it might be tempting to add a large number of values to that list. However, we strongly advise you **not** to do that: Unleash is **not** a database, and is not intended to store large amounts of data. Instead you should try and find a different way to achieve what you want.
```
gastonfournier, line 185:
```suggestion
When using a constraint operator that accepts a list of values, it might be tempting to add a large number of values to that list. However, we strongly advise you **not** to do that: Unleash is **not** a database, and is not intended to store large amounts of data. Instead you should try and find a different way to achieve what you want (e.g. using segmentation to classify your customers and the segment name as the constraint).
```
There's a better explanation below, but I feel like having this small pill here will help customers stay engaged (the paragraph states: don't use Unleash as a database and find your way around this limitation, which is a bit harsh). Having a suggestion of what could be that way might smoothen the message

gastonfournier: Another alternative (because the explanation below is great), would be to move the second paragraph to the bottom: 
> Because Unleash's server-side SDKs fetch the full feature toggle configuration from Unleash, every value that you add to that constraint value list will increase the payload size. For small numbers, this isn't an issue, but as the list grows, so will the payload, and so will the time and processing power used by the SDK to evaluate the feature.

You can start it with **Technical reasons** or similar. That way the customers will get the potential solution right away, and the technical details later (if they're interested).

gastonfournier (commented):
This looks great, just a small re-arrangement considering that as a customer I just want to solve the problem and maybe not be interested in the reasons why this is not a good idea.

website/docs/reference/strategy-constraints.md
thomasheartman: Love it! Addressed in 024f1f86f1317858918e47733e4b1ded47dae2ac; what do you think?

gastonfournier: Looks awesome!

gastonfournier (approved):
:clap:

PR #3143, opened 2023-02-17
Update demo start script

Tymek:
Starting with [Demo](https://app.unleash-hosted.com/demo/) backend should work with `cd frontend && yarn && yarn start:demo`


FredrikOseberg (approved):
Nice, maybe we could change `start:sandbox` also? It seems like that one also has the base path in the API url.

PR #3161, opened 2023-02-20
chore: make response with app time true by default

gastonfournier:
## About the changes
This defaults `responseWithAppNames` to true, but still allows us to control it through a feature toggle rather than an environment variable.

Self-hosted and open source can toggle it off by setting `UNLEASH_RESPONSE_TIME_WITH_APP_NAME` to false


chriswk (approved):
LGTM

gastonfournier:
Cancelled in favor of https://github.com/Unleash/unleash/pull/3178

PR #3162, opened 2023-02-20
fix: decouple forms

FredrikOseberg:
This PR decouples the forms for creating API tokens and project level API tokens. The point of having a hook that provides the functionality for the form is that we can create specific forms that take care of implementing the logic needed for that form instead of having one form serving multiple use cases.


frontend/src/component/admin/apiToken/ApiTokenTable/ProjectApiTokenCreate.tsx
```diff
@@ -1,29 +0,0 @@
...
-export const ProjectApiTokenCreate = () => {
-    const projectId = useRequiredPathParam('projectId');
-    const navigate = useNavigate();
-
-    const { access } = useProjectAccess(projectId);
-    const { users, serviceAccounts, groups } = useAccess();
-
-    if (!access || !users || !serviceAccounts || !groups) {
-        return null;
-    }
```
FredrikOseberg, line 18: Removed this from the new component because it didn't make sense to me. Why do we need to check access, users, service accounts or groups to create an API token here?

andreas-unleash: Thanks, I think that was left there by mistake

andreas-unleash (commented):
LGTM! Sorry for the trouble

andreas-unleash (approved):

Tymek (approved):

PR #3170, opened 2023-02-21
feat: extend refs in openapi

kwasniew:
## About the changes

```
payload: {
                    required: ['id'],
                    $ref: '#/components/schemas/updateFeatureStrategySchema',
                    type: 'object',
                    properties: {
                        id: { type: 'string' },
                    },
                },
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     ]
   },
   "dependencies": {
-    "@unleash/express-openapi": "^0.2.1",
+    "@unleash/express-openapi": "^0.2.2",
     "ajv": "^8.11.0",
     "ajv-formats": "^2.1.1",
     "async": "^3.2.4",
--- a/src/lib/openapi/validate.ts
+++ b/src/lib/openapi/validate.ts
@@ -11,7 +11,6 @@ const ajv = new Ajv({
     schemas: Object.values(schemas).map((schema) =>
         omitKeys(schema, 'components'),
     ),
-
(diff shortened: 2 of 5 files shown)


thomasheartman:
@kwasniew Could you explain what this does? I don't understand what this option does? Maybe include an example of what we couldn't do before but that we can do now? Does it have any drawbacks?

kwasniew:
@thomasheartman The example I put in the commit description was printing warning to the stdout. After this change it will allow to use $ref and other properties e.g. id in the example above.

thomasheartman:
@thomasheartman Right! Yeah. So what happens when when you ref something that already has those properties? Are they merged? Do the new properties take precedence? How does it work?

FredrikOseberg (approved):
LGTM

kwasniew:
@thomasheartman we don't have this use case but I'll try to check the next time I touch something in this area.

thomasheartman:
@kwasniew So are you saying that when we use refs, we only reference schemas that don't have these properties defined? That doesn't sound right to me 🤔 

I couldn't find any docs at AJV about this, and the closest thing I found in the swagger docs was this bit that says that [all sibling properties are ignored and overwritten when using $ref](https://swagger.io/docs/specification/using-ref#sibling). I also didn't see anything in the express-openapi readme about this. 

>  In our version of the openapi it has to be opted-in (with extendsRefs). In newer versions of openapi this is a default.

Could you point me to some documentation where this is explained?

PR #3178, opened 2023-02-22
chore: GA responseTimeWithAppNames

gastonfournier:
## About the changes
This makes response time with app names enabled for everyone but also kept a way of turning it off (kill switch) in case it cause some issues because of misconfigured app names


nunogois (approved):
LGTM

PR #3182, opened 2023-02-22
docs: add tables for client-side sdks

thomasheartman:


website/docs/reference/sdks/index.md
```diff
@@ -103,6 +90,102 @@ If you see an item marked with a ❌ that you would find useful, feel free to re
...
+Because client-side SDKs don't evaluate their features locally, some of their capabilities depend on what they are connected to.
+
+One clear difference between server-side SDKs and client-side SDKs, is that client-side SDKs don't support using fallback values in `isEnabled` calls. This is because the evaluation happens upstream from them.
+
+| Capability | [Android](/docs/generated/sdks/client-side/android-proxy.md) | [Flutter](/docs/generated/sdks/client-side/flutter.md) | [iOS](/docs/generated/sdks/client-side/ios-proxy.md) | [JavaScript](/docs/generated/sdks/client-side/javascript-browser.md) | [React](/docs/generated/sdks/client-side/react.md) | [Svelte](/docs/generated/sdks/client-side/svelte.md) | [Vue](/docs/generated/sdks/client-side/vue.md) |
+| --- | :-: | :-: | :-: | :-: | :-: | :-: | :-: |
+| **Category: Initialization** |  |  |  |  |  |  |  |
+| Async initialization | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ |
+| Can block until synchronized | ✅ | ✅ | ✅ | ⭕ | ⭕ | ✅ | ✅ |
+| Default refresh interval | 10s | 15s | 15s | 15s | 15s | 30s | 30s |
```
kwasniew, line 122: E.g for Flutter it's 30 seconds for the refresh interval and the metrics interval. But maybe instead of syncing the table we should sync the SDKs?

website/docs/reference/sdks/index.md
```diff
@@ -103,6 +90,102 @@ If you see an item marked with a ❌ that you would find useful, feel free to re
...
+| Default metrics interval | 60s | 60s | 60s | 60s | 60s | 60s | 30s |
+| **Category: Custom Headers** |  |  |  |  |  |  |  |
+| static | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| function | ✅ | ✅ | ⭕ | ✅ | ✅ (4.3) | ✅ | ✅ |
```
kwasniew, line 126: Which SDK supports function based custom headers? What is the use case? Dynamic header names/values that change over time?

website/docs/reference/sdks/index.md
```diff
@@ -103,6 +90,102 @@ If you see an item marked with a ❌ that you would find useful, feel free to re
...
+| **Category: [Unleash Context](../reference/unleash-context)** |  |  |  |  |  |  |  |
+| Static fields (`environment`, `appName`) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
```
kwasniew, line 128: E.g. in flutter sdk we don't have environment field. You can add it in your custom fields but it's not built-in.

website/docs/reference/sdks/index.md
```diff
@@ -103,6 +90,102 @@ If you see an item marked with a ❌ that you would find useful, feel free to re
...
+| Defined fields | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Custom properties | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| **Category: [Variants](../feature-toggle-variants.md)** |  |  |  |  |  |  |  |
+| Basic support | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| [Custom stickiness (beta)](../stickiness.md#custom-stickiness-beta) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
```
kwasniew, line 133: Does it make sense for the client side SDKs to be aware of this?

website/docs/reference/sdks/index.md
```diff
@@ -103,6 +90,102 @@ If you see an item marked with a ❌ that you would find useful, feel free to re
...
+| **Category: Local backup** |  |  |  |  |  |  |  |
+| File based backup | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| **Category: Usage metrics** |  |  |  |  |  |  |  |
+| Can disable metrics | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Client registration | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Basic usage metrics (yes/no) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| [Impression data](../impression-data.md) | ⭕ | ✅ | ⭕ | ⭕ | ⭕ | ⭕ | ✅ |
+| **Category: Bootstrap (beta)** |  |  |  |  |  |  |  |
+| Bootstrap from file | ✅ | ✅ | ✅ | ⭕ | ✅ | ✅ | ✅ |
```
kwasniew, line 142: I thought that bootstrap is just an object. You can read this object from a file but it doesn't seem to be Unleash client responsibility to reach out to a file. I'd rather pipe the output of a file to Unleash client (Unix philosophy)

website/docs/reference/sdks/index.md
thomasheartman: That we could, but it might also make sense for different SDKs to have different intervals. I'd be all for synchronizing the defaults, but I think we can do that as a separate thing.

website/docs/reference/sdks/index.md
thomasheartman: Oh, that's interesting! (And I think it's the right way to go). I think that this can be reworded a bit so that we can get away with it.

PR #3186, opened 2023-02-23
fix: notifications tables drop order

nunogois:
Since `user_notifications` references `notifications`, it makes sense to drop `user_notifications` before dropping `notifications` and not the other way around. This should fix the failing tests on Enterprise with latest main.

Thanks @gastonfournier for the help finding the bug 👍


gastonfournier (approved):
LG!

PR #3188, opened 2023-02-23
fix: Import change tense of help message

chriswk:
### What
Spotted during a demo today, this PR fixes a small grammar issue for the Import stage hint for import to a CR enabled project.


kwasniew (approved):
Good catch!

Issue #2951, opened 2023-01-19
feat: Sign-on history log


ivarconr:
### Describe the feature request

Capture all sign-on attempts and if they are successful or not and give information on why if they are not successful. 

### Availability

Enterprise

ivarconr:
will be part of v4.23.

PR #3193, opened 2023-02-24
feat: display sign on log

nunogois:
## About the changes
https://linear.app/unleash/issue/2-719/display-basic-sign-on-log-in-frontend
Adds a basic, naïve way of displaying sign-on history. 
We can be OK with merging this for now if we want, given that it's behind a feature flag, and then iterate over it and implement the features and improvements we need in the meantime.

image

Relates to [roadmap](https://github.com/orgs/Unleash/projects/10) item: #2951

### DIFF
--- a/frontend/src/component/common/Table/cells/TimeAgoCell/TimeAgoCell.tsx
+++ b/frontend/src/component/common/Table/cells/TimeAgoCell/TimeAgoCell.tsx
@@ -1,7 +1,7 @@
 import { Tooltip, Typography } from '@mui/material';
 import { useLocationSettings } from 'hooks/useLocationSettings';
 import { VFC } from 'react';
-import { formatDateYMD } from 'utils/formatDate';
+import { formatDateYMD, formatDateYMDHMS } from 'utils/formatDate';
 import { TextCell } from '../TextCell/TextCell';
 import TimeAgo from 'react-timeago';
 
@@ -10,6 +10,7 @@ interface ITimeAgoCellProps {
     live?: boolean;
     emptyText?: string;
     title?: (date: string) => string;
+    timestamp?: boolean;
 }
 
 export const TimeAgoCell: VFC<ITimeAgoCellProps> = ({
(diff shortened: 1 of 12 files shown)


frontend/src/component/common/Table/cells/TimeAgoCell/TimeAgoCell.tsx
```diff
@@ -10,19 +10,23 @@ interface ITimeAgoCellProps {
     live?: boolean;
     emptyText?: string;
     title?: (date: string) => string;
+    timestamp?: boolean;
```
gastonfournier, line 13: What about receiving the format function as with the title?
```suggestion
    dateFormat?: (value?: string | number | Date, locale?: string) => string;
```
Because timestamp as a boolean limits you to have 2 formats, if you want to add a third one things gets complicated. If the function is not provided you just use the default formatting function:
```typescript
const fn = dateFormat || formatDateYMD;
const date = fn(value, locationSettings.locale);
```

frontend/src/component/signOnLog/SignOnLogTable/SignOnLogTable.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const isExtraSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
+    const isSmallScreen = useMediaQuery(theme.breakpoints.down('md'));
+
+    const columns = useMemo(
+        () => [
+            {
+                Header: 'Created',
+                accessor: 'created_at',
+                Cell: ({ value }: { value: Date }) => (
+                    <TimeAgoCell value={value} timestamp />
```
gastonfournier, line 59: With the change suggested this would become:
```typescript
<TimeAgoCell value={value} dateFormat={formatDateYMDHMS} />
```

gastonfournier (approved):
Looks good, just a small suggestion, but you can ignore it if you feel it's a premature optimization

nunogois:
Thanks @gastonfournier, I think this is a great optimization for `TimeAgoCell` we can implement right away. Addressed in https://github.com/Unleash/unleash/pull/3202

PR #3194, opened 2023-02-27
docs: Update change requests permissions docs

thomasheartman:
This change updates the "change requests permissions" section of the change request doc to explicitly call out a few things we've seen confusion around recently.

These are:

-   project owners can't approve/apply change requests. You must create custom roles
-   viewers can create change requests. This is intentional
-   skip change requests only works via the API

It also:

-   elaborates a little on the circumventing change requests section to clarify how it works.
-   adds the change requests project roles to the project roles list. This section should be kept up to date with all the relevant roles that exist for projects.


kwasniew (approved):
Really good explanation

PR #3195, opened 2023-02-27
docs: update documentation around permissions for variants

gastonfournier:
## About the changes
This updates the documentation for permissions around variants


website/docs/reference/rbac.md
```diff
@@ -104,6 +104,10 @@ You can assign the following permissions on a per-environment level within the p
 
   Lets the user enable and disable toggles within the environment.
 
+- **update variants on environment**
```
thomasheartman, line 107: I know this is what the UI says, but I think we should shorten it. We should also update the UI: there's some inconsistencies right now with how it's used.

```suggestion
- **update variants**
```

thomasheartman (changes requested):
Nice one 💯  A few rewording suggestions and one question that I'm really curious about.

website/docs/how-to/how-to-create-feature-toggles.md
```diff
@@ -20,7 +20,7 @@ To be able to create a feature toggle in an Unleash system you will need:
 - A project to hold the toggle
 - A user with an **editor** or **admin** role OR a user with the following permissions inside the target project:
   - create feature toggles
-  - create/edit variants
+  - update variants (for the right environment)
```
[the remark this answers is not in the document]
gastonfournier: It's just the same as below, if it's unclear we should update the lines below

website/docs/reference/rbac.md
gastonfournier: I think I can also update the message on the UI to say update variants. In other permissions we also prepend "in this environment", do you think that's valuable or should we keep it short? I think the context of the environment it's right there so it's probably not needed

image

thomasheartman: Yeah, that's my point, exactly 😄  I think for this list, it shouldn't mention "in this environment", so it should also be "enable/disable toggles"

website/docs/how-to/how-to-create-feature-toggles.md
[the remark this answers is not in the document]
thomasheartman: Oh, interesting. Yeah, I see now that this is a how-to guide and not a reference document 😅  Yeah, in that case, I think this is fine..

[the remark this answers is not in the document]
gastonfournier: I also found it a bit misleading, because it says:
> To be able to create a feature toggle in an Unleash system you will need:
> - update variants 

and some other permissions which are not needed to create a feature toggle...

website/docs/reference/rbac.md
gastonfournier: I'm making it better for variants: https://github.com/Unleash/unleash/pull/3195/commits/ac017c785f88f65ddfc8dfecb0b8c8d3f778529e but let me know if you have better proposals for the others, this could be a good time to improve them, because it's unfortunate but we do have to run an SQL migration to change the display name

thomasheartman: Yeah, so my proposal is to leave most unchanged. I would change these, though:
- "Enable/disable toggles in this environment" -> "Enable/disable toggles" (or should it be features? maybe later 🤔 )
- "Approve a change request" -> "Approve change requests"
- "Apply a change request" -> "Apply change requests"

And maybe
- "skip change request process" -> "skip change request process (API-only)" for clarity?
wdyt?

website/docs/reference/rbac.md
```diff
@@ -82,7 +82,7 @@ You can assign the following project permissions. The permissions will be valid
 
 - **create/edit variants**
 
-  Lets the user create and edit variants within the project.
+  Lets the user create and edit variants within the project. (Deprecated with v4.21 in favor of environment-specific permissions for working with variants)
```
thomasheartman, line 85: Can you add a footnote here?

```suggestion
  Lets the user create and edit variants within the project. (Deprecated with v4.21 in favor of environment-specific permissions for working with variants[^1].)
```

And then something like this at the end of the document?
```
[^1]: The project-level permission is still required for the [**create/overwrite variants** (PUT)](api/unleash/overwriteFeatureVariants.api.mdx) and [**update variants** (PATCH)](api/unleash/patchFeatureVariants.api.mdx) API endpoints, but it is not used for anything within the admin UI. The API endpoints have been superseded by the [**create/overwrite environment variants** (PUT)](api/unleash/overwrite-environment-feature-variants.api.mdx) and [**update environment variants** (PATCH)](api/unleash/patch-environment-feature-variants.api.mdx) endpoints, respectively.
```

gastonfournier: Yes, on it!

website/docs/reference/rbac.md
gastonfournier: I like the proposals, also not adding a lot of unrelated changes is fine, but I agree with your recommendations. Applied!

thomasheartman (approved):
Love it! Let's get it in, you hero 🏆

PR #3198, opened 2023-02-27
docs: better comment in strictNullCheck action

gastonfournier:
## About the changes
This is a small improvement adding more context to the comment message when this action fails. This is what it'll comment:
https://github.com/Unleash/unleash/pull/3198#issuecomment-1446172703

### DIFF
--- a/.github/workflows/gradual-strict-null-checks.yml
+++ b/.github/workflows/gradual-strict-null-checks.yml
@@ -66,9 +66,10 @@ jobs:
           MAIN=$(grep "Found [0-9]* errors" .out-main | sed 's/Found \(.*\) errors in .* files./\1/')
 
           if [ $CURRENT -gt $MAIN ]; then
-            comment "After enabling strictNullChecks this PR would be increasing the number of null check errors from ${MAIN} to ${CURRENT}. Make sure your branch is up-to-date with ${MAIN_BRANCH} and check the diff in the console output to gather more details"
+            comment "After enabling \`strictNullChecks\` this PR would be **increasing** the number of null check errors from ${MAIN} to ${CURRENT}. <br /> Make sure your branch is up-to-date with ${MAIN_BRANCH} and **check the diff in the console output** to gather more details.<br /><br />**So what are strict null checks?** When strictNullChecks is \`false\`, null and undefined are effectively ignored by the language. This can lead to unexpected errors at runtime.<br/>When strictNullChecks is \`true\`, null and undefined have their own distinct types and you'll get a type error if you try to use them where a concrete value is expected.<br>[Documentation](https://www.typescriptlang.org/tsconfig#strictNullChecks)"
             diff .out-current .out-main
             exit 1
           else
             echo "The PR has $CURRENT null check errors against $MAIN in main. You're good to go!"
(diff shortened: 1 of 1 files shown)


.github/workflows/gradual-strict-null-checks.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           MAIN=$(grep "Found [0-9]* errors" .out-main | sed 's/Found \(.*\) errors in .* files./\1/')
 
           if [ $CURRENT -gt $MAIN ]; then
-            comment "After enabling strictNullChecks this PR would be increasing the number of null check errors from ${MAIN} to ${CURRENT}. Make sure your branch is up-to-date with ${MAIN_BRANCH} and check the diff in the console output to gather more details"
+            comment "After enabling \`strictNullChecks\` this PR would be **increasing** the number of null check errors from ${MAIN} to ${CURRENT}. <br /> Make sure your branch is up-to-date with ${MAIN_BRANCH} and **check the diff in the console output** to gather more details.<br /><br />**So what are strict null checks?** When strictNullChecks is \`false\`, null and undefined are effectively ignored by the language. This can lead to unexpected errors at runtime.<br/>When strictNullChecks is \`true\`, null and undefined have their own distinct types and you'll get a type error if you try to use them where a concrete value is expected.<br>[Documentation](https://www.typescriptlang.org/tsconfig#strictNullChecks)"
```
thomasheartman, line 69:
```suggestion
            comment "After enabling \`strictNullChecks\` this PR would be **increasing** the number of null check errors from ${MAIN} to ${CURRENT}. <br /> Make sure your branch is up-to-date with ${MAIN_BRANCH} and **check the diff in the console output** to gather more details.<br /><br />## What are strict null checks?<br/><br/>From the [`strictNullChecks` reference documentation](https://www.typescriptlang.org/tsconfig#strictNullChecks):<br/><br/>When \`strictNullChecks\` is \`false\`, \`null\` and \`undefined\` are effectively ignored by the language. This can lead to unexpected errors at runtime.<br/>When \`strictNullChecks\` is \`true\`, \`null\` and \`undefined\` have their own distinct types and you'll get a type error if you try to use them where a concrete value is expected."
```

thomasheartman: How about something like this? Moving the link to the docs to the front and clarifying that that's where the content is from?

gastonfournier: Sure! We just need to escape the `strictNullChecks` inside the reference documentation link

thomasheartman: Or we could do a simpler version to keep the message short (less text to read):

```suggestion
            comment "After enabling [\`strictNullChecks\`](https://www.typescriptlang.org/tsconfig#strictNullChecks) this PR would be **increasing** the number of null check errors from ${MAIN} to ${CURRENT}. <br /> Make sure your branch is up-to-date with ${MAIN_BRANCH} and **check the diff in the console output** to gather more details."
```

thomasheartman (approved):
Yeah, I think that's perfect. Keeps it nice and short but lets you go and find what you need

PR #3201, opened 2023-02-27
docs: clarify note around permissions required

thomasheartman:
This change does two things:

It clarifies that the permissions listed in the guide are permissions needed to perform all the steps in the guide, and not the permissions needed to create a feature.

Separates the project-level and environment-level permissions needed into separate sublists.


gastonfournier (approved):
Nice!

PR #3202, opened 2023-02-27
feat: dateFormat in TimeAgoCell

nunogois:
Slight refactor to `TimeAgoCell` that allows passing in a date format function.
Suggested by @gastonfournier in https://github.com/Unleash/unleash/pull/3193#pullrequestreview-1315064957


gastonfournier (approved):
You've made my suggestion even better! Looks great!

PR #3203, opened 2023-02-27
chore: Renames login_events to sign_on_log

chriswk:
This makes the distinction from the event services clearer.

In enterprise we'll also rename LoginEventService etc to reflect this rename.


nunogois (approved):
LGTM

PR #3206, opened 2023-02-27
feat: add authentication as a table filter

nunogois:
Adds `auth` as a table filter.
Also fixes the hook to correctly fetch the events based on the new return type.

image


gastonfournier (approved):
LG!

PR #3212, opened 2023-02-28
feat: implement plausible tracking

FredrikOseberg:
Adds plausible event tracking for notifications. Refactors Feedback component to be reusable and to ask whether or not this functionality is useful.

### DIFF
--- a/frontend/src/component/common/Notifications/Notifications.tsx
+++ b/frontend/src/component/common/Notifications/Notifications.tsx
@@ -9,7 +9,7 @@ import {
     Button,
 } from '@mui/material';
 import { useNotifications } from 'hooks/api/getters/useNotifications/useNotifications';
-import { ConditionallyRender } from '../ConditionallyRender/ConditionallyRender';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/Notifications/Notifications.tsx
```diff
@@ -63,11 +66,20 @@ export const Notifications = () => {
...
     const { markAsRead } = useNotificationsApi();
+    const { uiConfig } = useUiConfig();
+    const { trackEvent } = usePlausibleTracker();
 
     const onNotificationClick = (notification: NotificationsSchemaItem) => {
         if (notification.link) {
             navigate(notification.link);
         }
+
+        if (uiConfig?.flags?.T) {
```
kwasniew, line 77: do we need this flag if notifications are not even shown without this flag enabled?

kwasniew (approved):

frontend/src/component/common/Notifications/Notifications.tsx
FredrikOseberg: Yes, because we only want to track in the context of our cloud offering where tracking is available for us. For OSS and Enterprise-self hosted, this is not the case. I'd rather add the flag now, then to forget about it when we go to production.

PR #3214, opened 2023-02-28
refactor: make permission check type less demanding

kwasniew:
## About the changes
I have a use case where I need to check permissions for all project users. I only know their IDs. This new type will allow me to use this method. We only list parts of the IUser that are used inside the check.


chriswk (approved):
Seems reasonable to me

FredrikOseberg (approved):
LGTM

PR #3215, opened 2023-02-28
feat: add url/query sort/search to sign on log table

nunogois:
https://linear.app/unleash/issue/2-729/adapt-signontable-to-use-querystored-sortsearch-params

Adds the feature where the current sort + search are reflected in the URL.

image
(notice the URL at the top)


chriswk (approved):
LGTM

PR #3220, opened 2023-02-28
feat: smarter auth type labels on sign-on log table

nunogois:
Small change to make the `Authentication` column in the Sign-on Log table display nice-to-read labels for the multiple authentication types.

image


chriswk (approved):
Nice :)

PR #3231, opened 2023-03-01
feat: add download and delete all buttons to sign-on log UI

nunogois:
Adds `Download sign-on log` and `Clear sign-on log` buttons to the UI, making it easy for admins to manage their sign-on log.

image
image


gastonfournier (approved):
Looks fine!

PR #3233, opened 2023-03-01
fix: pass shared event store

kwasniew:
## About the changes
Pass event store around for event emitter semantics. This will fix import not triggering notifications


FredrikOseberg (approved):
LGTM

PR #3234, opened 2023-03-01
fix: load export dialog after envs loaded

kwasniew:
## About the changes
Fix hard refresh on project page affecting exported envs


FredrikOseberg (approved):
LGTM

PR #3241, opened 2023-03-02
feat: shared event emitter

kwasniew:
## About the changes

EventStore is used as part of DB transactions (events should be stored when business operations succeed) and EventStores are used as EventEmitters to notify other modules (including notifications module).
We want to have correctness of DB transactions (impossible to store event after failed business operation) and correctness of notifications (impossible not to get a notification). This PR combines those 2 architectural drivers. 

Problem in a diagram:
Screenshot 2023-03-02 at 09 19 05

Solution in code:

Composition over inheritance. Each EventEmitter should have HAS-A relationship to a shared EventEmitter. Previously it was IS-A relationship so we were getting a new emitter per EventStore instance. By making this dependency a singleton we make it impossible not to be notified about events no matter if we have one EventStore of many EventStores.

Screenshot 2023-03-02 at 09 26 20

### DIFF
--- a/src/lib/db/event-store.ts
+++ b/src/lib/db/event-store.ts
@@ -337,8 +337,8 @@ class EventStore implements IEventStore {
         };
     }
 
-    setMaxListeners(n: number): EventEmitter {
-        return this.eventEmitter.setMaxListeners(n);
+    setMaxListeners(number: number): EventEmitter {
+        return this.eventEmitter.setMaxListeners(number);
     }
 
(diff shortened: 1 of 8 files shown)


chriswk (approved):
Elegantly solved. I like how small it ended up being :)

sighphyre (approved):
LGTM

gastonfournier (approved):
Also looked at it, looks fine

PR #3243, opened 2023-03-02
feat: better segment info

nunogois:
## About the changes

- Updates the segment information on top to be clearer - No longer an experimental feature, but we do have some limits in place;
- Also updates the documentation to better reflect this;

Co-authored-by: @thomasheartman

image

### DIFF
--- a/website/docs/reference/deploy/configuring-unleash.md
+++ b/website/docs/reference/deploy/configuring-unleash.md
@@ -184,7 +184,7 @@ start();
 
 :::caution
 
-Changing segment limits could have adverse impacts on the performance of Unleash SDKs and cause network congestion. Think twice before changing these values.
+Changing segment limits could have a negative impact on the performance of Unleash SDKs and cause network congestion. Think twice before changing these values.
 
 :::
 
--- a/website/docs/reference/deploy/configuring-unleash.md
+++ b/website/docs/reference/deploy/configuring-unleash.md
@@ -180,6 +180,20 @@ const start = async () => {
 start();
 ```
 
+### Segment limits {#segments}
+
+:::caution
+
+Changing segment limits could have adverse impacts on the performance of Unleash SDKs and cause network congestion. Think twice before changing these values.
+
+:::
+
+Some facets of the [segments feature](../segments.mdx) can be customized via environment variables. This lets you change the [segment limits](../segments.mdx#segment-limits) that Unleash uses.
+
+`UNLEASH_STRATEGY_SEGMENTS_LIMIT` controls the maximum number of segments that can be applied to a single strategy. The default is 5.
+
+`UNLEASH_SEGMENT_VALUES_LIMIT` controls the maximum number of values that you can assign across a segment's constraints. The default is 100.
+
 ## Securing Unleash {#securing-unleash}
 
 You can integrate Unleash with your authentication provider (OAuth 2.0). Read more about [securing unleash](./securing-unleash.md).
--- a/website/docs/reference/segments.mdx
+++ b/website/docs/reference/segments.mdx
@@ -40,7 +40,7 @@ If an activation strategy has a segment _and_ additional constraints applied, th
 
 ## Segment limits
 
-In theory, you could you create segments with a thousand constraints, each with a million values. But this wouldn't scale well, so there are limitations in place to stop you from doing this. Unleash enforces the following limits on use of segments:
+In theory, you could create segments with a thousand constraints, each with a million values. But this wouldn't scale well, so there are limitations in place to stop you from doing this. Unleash enforces the following limits on use of segments:
 
 1. By default, a segment can have **at most 100 values** specified across all of its constraints. That means that if you add a constraint that uses 10 values, you will have 90 more values to use for any other constraints you add to the same segment.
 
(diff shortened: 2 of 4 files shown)


nunogois:
@thomasheartman Looking at [SegmentDocs.tsx](https://github.com/Unleash/unleash/blob/feat-better-segment-info/frontend/src/component/segments/SegmentDocs.tsx) we may want to make further changes:

1. Should we reword this as well?
https://github.com/Unleash/unleash/blob/90bb557854343455bc133bb11266093f32786598/frontend/src/component/segments/SegmentDocs.tsx#L30-L34

2. Should we reword this **and** update the docs accordingly as well?
https://github.com/Unleash/unleash/blob/90bb557854343455bc133bb11266093f32786598/frontend/src/component/segments/SegmentDocs.tsx#L46-L49

3. Should we keep this *as is*?
https://github.com/Unleash/unleash/blob/90bb557854343455bc133bb11266093f32786598/frontend/src/component/segments/SegmentDocs.tsx#L55-L65

thomasheartman:
@nunogois Great points! 

```tsx
 <Alert severity="error"> 
     A segments can have at most {segmentValuesLimit} across all of its constraints. This 
     segment has {props.values}{' '} 
     {props.values === 1 ? 'value' : 'values'}. 
 </Alert> 
```

What does this mean? Is it that strategies can have at most `strategySegmentsLimit` segments applied? If so, what about something like this:

```tsx
 <Alert severity="warning"> 
     You can't apply more than {strategySegmentsLimit} segments to a strategy.{' '} 
     <SegmentLimitsLink /> 
 </Alert> 
```

As for the last one: I think linking to slack is fine, but we could also link to github discussions? https://github.com/unleash/unleash/discussions/

nunogois:
> @nunogois Great points!
> 
> How about something like this?
> 
> ```tsx
>  <Alert severity="error"> 
>      A segments can have at most {segmentValuesLimit} across all of its constraints. This 
>      segment has {props.values}{' '} 
>      {props.values === 1 ? 'value' : 'values'}. 
>  </Alert> 
> ```
> 
> What does this mean? Is it that strategies can have at most `strategySegmentsLimit` segments applied? If so, what about something like this:
> 
> ```tsx
>  <Alert severity="warning"> 
>      You can't apply more than {strategySegmentsLimit} segments to a strategy.{' '} 
>      <SegmentLimitsLink /> 
>  </Alert> 
> ```
> 
> As for the last one: I think linking to slack is fine, but we could also link to github discussions? https://github.com/unleash/unleash/discussions/

Thanks, addressed in https://github.com/Unleash/unleash/pull/3243/commits/468b560b81daff2003e1bd2d30a7451928335ddb.

I believe you are correct with your interpretation of point 2.

I've kept the Slack link for now, but I think in the future it would be nice to somehow give the option to the users of how they would prefer to get in touch, maybe by showing these options in a dialog before redirecting directly 🤷‍♂️ 

Let me know if we decide GitHub discussions is how we prefer to be reached about this instead and I'll quickly make the change.

thomasheartman (approved):
Code changes look good to me 🙋🏼  @nunogois , do you wanna have a look at the docs?

website/docs/reference/segments.mdx
```diff
@@ -38,16 +38,24 @@ Segments are collections of strategy constraints. To satisfy a segment, _all_ th
 
 If an activation strategy has a segment _and_ additional constraints applied, the segment _and_ the strategies must all be satisfied. Similarly, if an activation strategy has multiple segments, then they must _must all be satisfied_.
 
-## Creating, updating, and deleting segments
+## Segment limits
 
-Segments can be created, edited, and deleted from the segments page in the admin UI or via the API (see the [segments API documentation](/reference/api/legacy/unleash/admin/segments.mdx)).
+In theory, you could you create segments with a thousand constraints, each with a million values. But this wouldn't scale well, so there are limitations in place to stop you from doing this. Unleash enforces the following limits on use of segments:
```
nunogois, line 43:
```suggestion
In theory, you could create segments with a thousand constraints, each with a million values. But this wouldn't scale well, so there are limitations in place to stop you from doing this. Unleash enforces the following limits on use of segments:
```

nunogois:
> Code changes look good to me 🙋🏼 @nunogois , do you wanna have a look at the docs?

Docs look good to me as well! Let's have someone else take a look just in case before we merge this 🙂

thomasheartman:
Sure! 

I'll re-request some of them, because I can't remember doing this:

image

website/docs/reference/deploy/configuring-unleash.md
```diff
@@ -180,6 +180,20 @@ const start = async () => {
 start();
 ```
 
+### Segment limits {#segments}
+
+:::caution
+
+Changing segment limits could have adverse impacts on the performance of Unleash SDKs and cause network congestion. Think twice before changing these values.
```
gastonfournier, line 187:
```suggestion
Changing segment limits could have an adverse impact on the performance of Unleash SDKs and cause network congestion. Think twice before changing these values.
```
or maybe 
```suggestion
Changing segment limits could have a negative impact on the performance of Unleash SDKs and cause network congestion. Think twice before changing these values.
```
Which I heard more frequently

gastonfournier (approved):
I like it!

website/docs/reference/deploy/configuring-unleash.md
thomasheartman: Sure! Negative is probably clearer than adverse, so that's good 😄

PR #3245, opened 2023-03-02
refactor: rename sign-on log to login history

nunogois:
Renames `sign-on log` to `login history`.


thomasheartman (commented):
This looks pretty good to me. Is the login history an experimental thing (I see it's listed under `types/experimental`)? If not, we'll need to maintain the previous endpoint too.

nunogois:
> This looks pretty good to me. Is the login history an experimental thing (I see it's listed under `types/experimental`)? If not, we'll need to maintain the previous endpoint too.

Yep, it is behind a flag currently and it always has been since we're still in early stages, so no need to worry about breaking changes here.

PR #3246, opened 2023-03-02
fix: Return all toggles when env var is set and flag is on

andreas-unleash:
Fixes bug where strategies where not evaluated when returning all toggles.
Added env var to control the feature per client in addition to the flag

## About the changes

Closes #


src/lib/services/proxy-service.ts
```diff
@@ -86,7 +86,7 @@ export class ProxyService {
 
         return definitions.map((feature) => ({
             name: feature.name,
-            enabled: Boolean(feature.enabled),
+            enabled: Boolean(client.isEnabled(feature.name)),
```
kwasniew, line 89: no context?

src/test/e2e/api/proxy/proxy.e2e.test.ts
```diff
@@ -934,7 +934,10 @@ test('Should not recursively set off timers on events', async () => {
 });
 
 test('should return all features when specified', async () => {
```
thomasheartman, line 936: Just for discussion: So this test does double duty as something that also checks that "strategies are correctly evaluated" when using proxy all (because the `disabledToggle` is included with `enabled: false`. Would it be worth adding that in a separate test or put that in the description somehow? Just so it's clear what it's doing? Or would it be redundant? 

In a way, it's already covered by this test, but if you're only reading test names, it's not clear that it's being tested for.

andreas-unleash: True, I will separate it out to its own test

src/test/e2e/api/proxy/proxy.e2e.test.ts
```diff
@@ -982,6 +1006,133 @@ test('should return all features when specified', async () => {
...
+                        variant: { enabled: false, name: 'disabled' },
+                    },
+                    {
+                        name: 'enabledFeature2',
+                        enabled: true,
+                        impressionData: false,
+                        variant: { enabled: false, name: 'disabled' },
+                    },
+                ],
+            });
```
thomasheartman, line 1132: Just a thought I had (and not something that needs to be addressed in this PR or maybe ever):

Do these tests test too much? The test name only indicates that it returns the corrects toggles, and not necessarily that they contain the right kind of data (other than status). That makes them kind of brittle in case we change the payload later (by adding more data, like how we did for `impressionData`). If this test is only to test the name, would it be better to do something like this?

```suggestion
            expect(res.body.toggles.map(toggle => toggle.name).toEqual(["enabledFeature1", "enabledFeature2"]);
```

(The same thing could probably be done to separate the duplicate testing that we discussed earlier 💁🏼 )

src/lib/types/experimental.ts
```diff
@@ -26,8 +26,12 @@ const flags = {
         process.env.UNLEASH_RESPONSE_TIME_WITH_APP_NAME_KILL_SWITCH,
         false,
     ),
+    proxyReturnAllTogglesKillSwitch: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_PROXY_RETURN_ALL_TOGGLES_KILL_SWITCH,
+        false,
+    ),
     proxyReturnAllToggles: parseEnvVarBoolean(
-        process.env.UNLEASH_EXPERIMENTAL_PROXY_RETURN_ALL_TOGGLES,
+        process.env.UNLEASH_PROXY_RETURN_ALL_TOGGLES,
```
thomasheartman, line 34: Why is one of these marked as experimental and the other one not?

src/lib/routes/proxy-api/index.ts
```diff
@@ -145,7 +145,13 @@ export default class ProxyController extends Controller {
         res: Response<ProxyFeaturesSchema>,
     ) {
         let toggles;
-        if (this.flagResolver.isEnabled('proxyReturnAllToggles')) {
+        const releaseFlag = this.flagResolver.isEnabled(
+            'proxyReturnAllTogglesKillSwitch',
```
thomasheartman, line 149: On the semantics of kill switches: Should they kill something when they're enabled or when they're disabled? 

I thought that kill switches generally meant that a feature is turned off when the kill switch is engaged? At least that's what the [kill switches best practice blog post](https://www.getunleash.io/blog/kill-switches-best-practice) from our blog says. 

I don't know if we've used kill switches anywhere else in our code base. We should probably be consistent internally, but ideally also with the material we put out on how to use it.

So if that's correct, then the `releaseFlag` variable is the opposite of `isEnabled('proxyReturnAllTogglesKillSwitch')`, I think 🤔

thomasheartman (commented):
So this looks pretty good to me, but I have a question regarding kill switches that we should at least resolve before we merge this. There's also a few more questions in there that I'd like your take on.

Also, as @FredrikOseberg mentioned in the planning, we might want to not include this functionality in the front-end API after all. So we should probably clear that up before we merge too.

FredrikOseberg:
Based on the discussion in the post mortem and afterwards, I believe that we should actually remove this functionality. But let's leave in the test that proves the functionality. cc: @andreas-unleash @thomasheartman

PR #3252, opened 2023-03-03
refactor: light theme colors

NicolaeUnleash:
## About the changes
Refactoring the colors for the light theme to be much easier to continue with dark mode

This is the first step to finish dark mode
https://linear.app/unleash/project/[low][s][alpha]-dark-mode-in-unleash-admin-ui-31b407d13c4b/1

This PR uses `main-theme` as a placeholder for `dark-theme` for now due to the new changes. Still need to set the correct values here.


nunogois (approved):
LGTM, had a quick check with Nicolae through the UI and everything looks great, didn't spot anything broken. I vote we merge this earlier rather than later so we can start using the new variables right away and get started with dark mode 🙂

PR #3263, opened 2023-03-07
chore: add a new project column to segments table

gastonfournier:
## About the changes
Adds a migration and persistence layer with a new column `segment_project_id` to bind a segment to a project. 

## Discussion points
The reasoning behind the column name instead of `project` is due to collisions with other columns. The alternative is to modify existing unrelated queries that will fail due to a name collision because of `left join "segments"` that will also bring a column named `project`. Example:

> select "features"."name" as "name", "features"."description" as "description", "features"."type" as "type", 
**"features"."project" as "project"**, "features"."stale" as "stale", "features"."impression_data" as "impression_data", "fe"."variants" as "variants", "features"."created_at" as "created_at", "features"."last_seen_at" as "last_seen_at", "fe"."enabled" as "enabled", "fe"."environment" as "environment", "fs"."id" as "strategy_id", "fs"."strategy_name" as "strategy_name", "fs"."parameters" as "parameters", "fs"."constraints" as "constraints", "segments"."id" as "segment_id", "segments"."constraints" as "segment_constraints" from "features" left join (select * from "feature_strategies" where "environment" = $1) as "fs" on "fs"."feature_name" = "features"."name" left join (select "feature_name", "enabled", "environment", "variants" from "feature_environments" where "environment" = $2) as "fe" on "fe"."feature_name" = "features"."name" left join "feature_strategy_segment" as "fss" on "fss"."feature_strategy_id" = "fs"."id" **left join "segments" on "segments"."id" = "fss"."segment_id"** where "archived_at" is null and "project" in ($3) 
> - column reference "project" is ambiguous

Still open to suggestions as refactoring these queries may bring more clarity to the code


src/lib/db/feature-toggle-client-store.ts
```diff
@@ -51,28 +50,16 @@ export default class FeatureToggleClientStore
...
     private timer: Function;
 
-    private flagResolver: IFlagResolver;
-
-    constructor(
-        db: Db,
-        eventBus: EventEmitter,
-        getLogger: LogProvider,
-        inlineSegmentConstraints: boolean,
-        flagResolver: IFlagResolver,
```
gastonfournier, line 65: Unrelated cleanup of unused parameters to simplify test setup src/test/e2e/stores/feature-toggle-client-store.e2e.test.ts

src/migrations/20230306103400-add-project-column-to-segments.js
```diff
@@ -0,0 +1,11 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `ALTER TABLE segments ADD COLUMN IF NOT EXISTS segment_project_id VARCHAR(255);`,
```
chriswk, line 3: USE text as type here. Also, don't use if not exists, if the column exist I believe we want the migration to fail. Also, I believe we should put an index on this field

chriswk (changes requested):
I like rationalisation around the field name, but I do believe at least the index and the unnecessary if not exists should be removed

src/migrations/20230306103400-add-project-column-to-segments.js
```diff
@@ -0,0 +1,11 @@
...
+        cb,
+    );
+    cb();
+};
+
+exports.down = function (db, cb) {
+    cb();
```
nunogois, line 10: Shouldn't we have a migration down here as well?

chriswk: Good point, we really should

src/migrations/20230306103400-add-project-column-to-segments.js
gastonfournier: I've chosen `VARCHAR(255)` to use the same type as projects table: https://github.com/Unleash/unleash/blob/e4b84bc2a3620336b36067afbfd11cf5d2edd47f/src/migrations/20200928194947-add-projects.js#L10-L16

And also following other references to project.id https://github.com/Unleash/unleash/blob/e4b84bc2a3620336b36067afbfd11cf5d2edd47f/src/migrations/20210618091331-project-environments-table.js#L5 but in trying to fix the test issues I've accidentally removed the cascading configuration. 

Adding an index is worthwhile

src/migrations/20230306103400-add-project-column-to-segments.js
gastonfournier: We have [many migrations without a down migration](https://sourcegraph.com/search?q=context%3Aglobal+repo%3A%5Egithub%5C.com%2FUnleash%2Funleash%24+cb%28%29&patternType=standard&sm=1&groupBy=path). In this case, adding a column shouldn't be harmful (although tests have shown that the wrong column name might cause issues), but dropping the column might unlink the segment from the project (which might not be that bad).

But, all in all, it doesn't hurt to add it if you consider it valuable. Particularly, I don't see much risk in not doing the down migration, unless we remove IF NOT EXISTS as suggested [here](https://github.com/Unleash/unleash/pull/3263#discussion_r1127731408)

src/migrations/20230306103400-add-project-column-to-segments.js
gastonfournier: I looked at the different access patterns we have from the issues in our backlog, and I believe the main one is "get all segments except the ones bound to other projects", which can be solved by a single query or two queries: one by project id and another one for `project_id is NULL`. Only in the last case, the index will be helpful. So I'd like to delay the index creation until we tackle https://linear.app/unleash/issue/2-737/when-listing-segments-project-specific-segments-should-also-be-present

I'm still open to add the index but perhaps the FK constraint is enough

src/migrations/20230306103400-add-project-column-to-segments.js
gastonfournier: Added and removed `IF NOT EXISTS`

nunogois (approved):
LGTM

chriswk (approved):
LGTM

PR #3266, opened 2023-03-07
Remove link that's not necessary

gastonfournier:
## About the changes
Tested manually as admin in enterprise:

image

Admin should have a special permission that gives super-powers. There's no need to have specific permissions linked to it.

Based of: https://github.com/Unleash/unleash/blob/a077967760f39df2f0fd7adf1db64d14df039326/src/migrations/20220307130902-add-segments.js#L32-L46


sighphyre (approved):
LGTM

chriswk (approved):
LGTM

PR #3268, opened 2023-03-08
fix: polling in proxy repository now stops correctly

sighphyre:
### What

This patches two very subtle bugs in the proxy repository that cause it to never actually stop polling the db in the background

## Details - Issue 1

We've recently started to get the following output when running `yarn test`:

` Attempted to log "Error: Unable to acquire a connection
        at Object.queryBuilder (/home/simon/dev/unleash/node_modules/knex/lib/knex-builder/make-knex.js:111:26)`
        
This seems to occur for every test suite after running the proxy tests and the full stack trace doesn't point to anything related to the running tests that produce this output. Running a `git bisect` points to this commit: https://github.com/Unleash/unleash/commit/6e44a65c58d8e28668f0d3459b62c0ce0b84849a being the culprit but I believe that this may have surfaced the bug rather than causing it.

Layering in a few console logs and running Unleash, seems to point to the proxy repository setting up data polling but never actually terminating it when `stop` was called, which is inline with the output here - effectively the tests were continuing to run the polling in the background after the suite had exited and jest freaks out that an async task is running when it shouldn't be. This is easy to reproduce once the console logs are in place in the `dataPolling` function, by running Unleash - creating and deleting a front end token never terminates the poll cycle. 

I believe the cause here is some subtlety around using async functions with timers - stop was being called, which results in the timer being cleared but a scheduled async call was already on the stack, causing the recursive call to resolve after stop, resurrecting the timer and reinitializing the poll cycle.

I've moved the terminating code into the async callback. Which seems to solve the problem here.

## Details - Issue 2

Related to the first issue, when the proxy service stops the underlying Unleash Client, it never actually calls destroy on the client, it only removes it from its internal map. That in turn means that the Client never calls stop on the injected repository, it only removes it from memory. However, the scheduled task is `async` and `unref`, meaning it continues to spin in the background until every other process also exits. This is patched by simply calling destroy on the client when cleaning up

## The Ugly

This is really hard to test effectively, mostly because this is an issue caused by internals within NodeJS and async. I've added a test that reads the output from the debug log (and also placed a debug log in the termination code). This also requires the test code to wait until the async task completes. This is horribly fragile so if someone has a better idea on how to prove this I would be a very happy human.

The second ugly part is that this is a subtle issue in complex code that really, really needs to work correctly. I'm nervous about making changes here without lots of eyes on this


src/lib/proxy/proxy-repository.ts
```diff
@@ -85,11 +88,19 @@ export class ProxyRepository
 
     stop(): void {
         this.stores.eventStore.off(ANY_EVENT, this.onAnyEvent);
-        clearTimeout(this.timer);
+        this.running = false;
```
ivarconr, line 91: Nice catch; this would be correct if we actually used a setTimeout. In this example we are setting up new timeouts. 

Q: Should this have been a scheduled job?

sighphyre: I think the setTimeout is in place here so that we can introduce jitter on the db calls. Otherwise we risk hammering the db at at once when there's multiple active frontend tokens

I'm not quite sure a scheduled job fits here. That seems to be designed to tackle setting up schedules for service level tasks that are consistent and live for the application lifetime. This is a set of schedules on the repository layer with jitter that are added and removed on an adhoc basis depending on what's happening with the frontend tokens that you have active.

src/test/e2e/api/proxy/proxy.e2e.test.ts
```diff
@@ -991,3 +992,47 @@ test('should return maxAge header on options call', async () => {
...
+        },
+        db.stores,
+        app.services,
+        user,
+    );
+
+    await proxyRepository.start();
+    proxyRepository.stop();
+    // Polling here is an async recursive call, so we gotta give it a bit of time
+    await new Promise((r) => setTimeout(r, 10));
```
sighphyre, line 1034: Had a very good suggestion to tackle this test with fake timers. However, that never seems to surface the bug on main. I believe that's because this is fundamentally a race condition. I think this is okay since it only adds a 10ms wait here but I'm open to suggestions

src/lib/proxy/proxy-repository.ts
```diff
@@ -85,11 +88,19 @@ export class ProxyRepository
...
     }
 
     private async dataPolling() {
         this.timer = setTimeout(async () => {
+            if (!this.running && this.timer) {
+                clearTimeout(this.timer);
+                this.timer = null;
+                this.logger.debug(
+                    'Shutting down data polling for proxy repository',
+                );
```
gastonfournier, line 101: **Alternative idea**: you can have an optional callback function as a parameter. 
This callback function will never be injected, except for the test code.
In the test setup you can inject `jest.fn()` and expect it to be called once

sighphyre: I really like the approach there but it doesn't actually catch the issue on main and it seems to break other tests :( I've reset that change out of the branch but I have a copy on my machine if we want to build on it

gastonfournier: Yeah, let's go with the improvement

gastonfournier (approved):

PR #3272, opened 2023-03-08
chore: add a toggle for project segments

gastonfournier:
## About the changes
This adds a new flag that can be used in the UI to display project specific segments


frontend/src/interfaces/uiConfig.ts
```diff
@@ -48,6 +48,7 @@ export interface IFlags {
     proPlanAutoCharge?: boolean;
     notifications?: boolean;
     loginHistory?: boolean;
+    projectScopedSegments?: boolean;
```
gastonfournier, line 51: Is there any reason why this is optional? I think it can be made non optional without any compilation issue

chriswk (approved):
LGTM

PR #3273, opened 2023-03-08
chore: enable strictNullCheck but ignore problems at compile time

gastonfournier:
## About the changes
This enables strictNullChecks which will give us nice hints in our IDEs to avoid introducing more and hopefully will encourage us to fix some of the existing problems.

Also:
1. The compiler explicitly ignores these errors
2. The "null checks action" still verifies we're not introducing new ones.

The combination of these two things should help us to reduce the number of nulls

For testing that the action still works (cause it was modified), [a commit](https://github.com/Unleash/unleash/pull/3273/commits/5c4b818d1acc3f46053f22c32ded167b01a045b8) was added introducing a bunch of null check errors: https://github.com/Unleash/unleash/actions/runs/4364648191/jobs/7632224720


.github/workflows/gradual-strict-null-checks.yml
```diff
@@ -47,13 +47,12 @@ jobs:
               --data "{ \"body\": \"${1}\" }"
           }
 
-          sed -i 's/\/\/\s*"strictNullChecks":\s*true,/"strictNullChecks": true,/' "current/tsconfig.json"
-          sed -i 's/\/\/\s*"strictNullChecks":\s*true,/"strictNullChecks": true,/' "main/tsconfig.json"
+          YARN_1="yarn --mutex network --cwd ./current"
+          YARN_2="yarn --mutex network --cwd ./main"
 
-          yarn --mutex network --cwd ./current 2> .stderr-current > .out-current &
+          $YARN_1 install --ignore-scripts &> /dev/null && $YARN_1 build --strictNullChecks true 2> .stderr-current > .out-current &
```
gastonfournier, line 53: As an explanation for this change:
`install` runs the `prepare` script that transitively calls `build`. To improve performance we just want to download dependencies and then run `build`

sighphyre (approved):
I'm excited for this one!

PR #3274, opened 2023-03-08
feat: bulk update tags

sjaanus:
Now the bulk update tag can add/remove tags in the same time.


src/lib/types/experimental.ts
```diff
@@ -64,6 +64,7 @@ const flags = {
     ),
     notifications: parseEnvVarBoolean(process.env.NOTIFICATIONS, false),
     loginHistory: parseEnvVarBoolean(process.env.UNLEASH_LOGIN_HISTORY, false),
+    bulkOperations: parseEnvVarBoolean(process.env.BULK_OPERATIONS, false),
```
Tymek, line 67:
```suggestion
    bulkOperations: parseEnvVarBoolean(process.env.UNLEASH_BULK_OPERATIONS, false),
```
Previous convention was to keep `UNLEASH_` prefix. I think it's good practice to have it

PR #3276, opened 2023-03-08
fix: exclude favorites from the client api

kwasniew:
## About the changes

Client API should not return favorites.
Enabling check in tests so that we don't make the same mistake again.
For now moved favorites inclusion in the admin if path. Separate PR may split the store into client and admin if this turns out to be a viable strategy.


sjaanus (approved):
LGTM!

PR #3277, opened 2023-03-08
refactor: separate client store and admin toggle store

kwasniew:
## About the changes

Separating client toggle store from the admin feature store (used for UI listing toggles across all projects).

In the new world:
* client store doesn't have any admin stuff leaked out by accident (main goal of this change)
* admin store still has some client stuff leaking out because of the segments tests

The change was made as a copy-paste and following the if trail (isAdmin).


src/lib/db/feature-toggle-admin-store.ts
```diff
@@ -0,0 +1,287 @@
...
+                store: 'feature-toggle',
+                action,
+            });
+    }
+
+    private async getAll({
+        featureQuery,
+        archived,
+        userId,
+    }: IGetAllFeatures): Promise<(IFeatureToggleClient & FeatureToggle)[]> {
```
kwasniew, line 58: we could remove 50% of the code and not return IFeatureToggleClient specific stuff if the segments.e2e.test.ts wasn't using admin endpoint to fetch strategies that are only part of the client API. Are there any reasons why segments tests do that? @chriswk @sighphyre

src/lib/services/feature-toggle-service.ts
```diff
@@ -704,7 +710,7 @@ class FeatureToggleService {
         userId?: number,
         archived: boolean = false,
     ): Promise<FeatureToggle[]> {
-        return this.featureToggleClientStore.getAdmin({
+        return this.featureToggleAdminStore.getAdmin({
```
kwasniew, line 713: admin is returning not just FeatureToggle but also the client stuff that some tests rely on

PR #3278, opened 2023-03-08
chore: avoid building frontend when possible

gastonfournier:
## About the changes
This is based on @nunogois suggestion to split the build in two, so we don't build the frontend every time we run `yarn`

e2e frontend tests were forced to run by modifying frontend/README.md

### Important files
Some github actions had to be updated to also build the frontend. The Dockerfile building our docker image was also looked into but it should work as is

## Discussion points
This is a potentially risky operation as we might overlook something that requires building the frontend which might lead to invalid builds. We need to make sure when we do this we don't have any release planned.


package.json
```diff
@@ -38,13 +38,15 @@
     "build:watch": "tsc -w",
     "prebuild": "yarn run clean",
     "build": "yarn run copy-templates && tsc --pretty",
-    "prepare": "node scripts/husky-install && cd frontend && yarn && cd .. && yarn run build",
+    "build:frontend": "yarn --cwd ./frontend",
+    "prepare": "node scripts/husky-install && yarn run build",
```
nunogois, line 42: Would be nice to somehow double-check if we really need this. Also, do we need to explicitly `build` here? Would be great if we could get rid of the `prepare` script altogether.

package.json
```diff
@@ -38,13 +38,15 @@
     "build:watch": "tsc -w",
     "prebuild": "yarn run clean",
     "build": "yarn run copy-templates && tsc --pretty",
-    "prepare": "node scripts/husky-install && cd frontend && yarn && cd .. && yarn run build",
+    "build:frontend": "yarn --cwd ./frontend",
```
nunogois, line 41: Maybe this one could explicitly build `frontend` instead of relying on frontend's `prepare` script? That way maybe we could also get rid of `prepare` script in frontend and be more explicit instead.

nunogois (commented):
Also a bit scared of us overlooking things and breaking builds, so I would feel more comfortable if others could have a look before we merge this.

Still, I feel like there is a lot of space for improvement regarding this.
To me, ideal DX in terms of scripts would be:
 - `yarn` - Install dependencies (maybe on both BE and FE?);
 - `yarn dev` - Start both BE and FE in watch/dev mode (this way we would prevent having to run both at the same time ourselves) - I also added `yarn` to `dev` scripts in the past to double check if dependencies are up-to-date before starting, but maybe there's a better way;
 - `yarn build` - Build both BE and FE for prod (used for e.g. deploys/releases);

package.json
gastonfournier: Yes, but I'd do that in another step to reduce the amount of changes

.github/workflows/build_prs.yaml
```diff
@@ -20,3 +20,4 @@ jobs:
       - run: yarn install --frozen-lockfile --ignore-scripts
       - run: yarn lint
       - run: yarn build
+      #- run: yarn build:frontend # not needed
```
gastonfournier, line 23: We don't need this because we have https://github.com/Unleash/unleash/blob/improve-dev-build-time/.github/workflows/build_frontend_prs.yml

sighphyre (approved):
Yeah, this is a huge improvement, it would be really nice to have

Makes me nervous to fiddle with our scripts like this so maybe we merge it and test it thoroughly before next release?

PR #3282, opened 2023-03-09
fix: hide change password when simple auth is disabled

nunogois:
https://linear.app/unleash/issue/2-764/change-password-should-be-disabled-when-password-login-is-disabled

Hides "Change password" option on the profile page when "Password based login" is disabled.

image


thomasheartman (approved):
Simple and clean! This looks very good to me! 🔥

PR #3284, opened 2023-03-09
feat: filter out all data fields other than explicit openapi fields

kwasniew:
## About the changes

Service returning client feature data filters in only relevant fields. Anything that is not part of the openapi schema will be removed.


gastonfournier (approved):
This looks safe, I wonder if we can use the schema type to filter out additional properties :thinking: we can leave that question open for later.

If we're unsure about this change we can always put a flag behind it, but since we're dogfooding I think we can do more thorough testing in our pre-prod environment

kwasniew:
@gastonfournier https://ajv.js.org/options.html#removeadditional - this option removed things outside the schema. However removeadditional I'd only add behind a flag. This one seems to be innocent

PR #3290, opened 2023-03-09
feat: allow adding project to segment

nunogois:
https://linear.app/unleash/issue/2-741/add-project-dropdown-into-creation-of-segments

Allows users to bind a segment to a project when creating or editing a segment.

image

### DIFF
--- a/frontend/src/component/segments/SegmentForm.tsx
+++ b/frontend/src/component/segments/SegmentForm.tsx
@@ -12,11 +12,11 @@ export type SegmentFormMode = 'create' | 'edit';
 interface ISegmentProps {
     name: string;
     description: string;
-    project: string;
+    project: string | null;
     constraints: IConstraint[];
     setName: React.Dispatch<React.SetStateAction<string>>;
     setDescription: React.Dispatch<React.SetStateAction<string>>;
-    setProject: React.Dispatch<React.SetStateAction<string>>;
+    setProject: React.Dispatch<React.SetStateAction<string | null>>;
     setConstraints: React.Dispatch<React.SetStateAction<IConstraint[]>>;
     handleSubmit: (e: any) => void;
     errors: { [key: string]: string };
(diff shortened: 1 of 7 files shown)


src/lib/db/segment-store.ts
```diff
@@ -203,7 +203,7 @@ export default class SegmentStore implements ISegmentStore {
             id: row.id,
             name: row.name,
             description: row.description,
-            project: row.segment_project_id,
+            project: row.segment_project_id || '',
```
nunogois, line 206: Needed some of this back-and-forth here for consistency and due to the way we're binding the project column to projects: Otherwise if we try to set project as `""` we're not able to, as that project does not exist.
This way we have a simple mapping from `""` to `null` and back.

gastonfournier: I'm not sure I'm 100% convinced on this one, I think the project should be made optional in all the UI code around segments, but maybe I'm not fully understanding what you mean in this comment

frontend/src/component/segments/SegmentForm.tsx
```diff
@@ -12,9 +12,11 @@ export type SegmentFormMode = 'create' | 'edit';
 interface ISegmentProps {
     name: string;
     description: string;
+    project: string;
```
gastonfournier, line 15: Should this be 
```suggestion
    project: string | undefined;
```
?

frontend/src/component/segments/SegmentForm.tsx
```diff
@@ -12,9 +12,11 @@ export type SegmentFormMode = 'create' | 'edit';
...
     constraints: IConstraint[];
     setName: React.Dispatch<React.SetStateAction<string>>;
     setDescription: React.Dispatch<React.SetStateAction<string>>;
+    setProject: React.Dispatch<React.SetStateAction<string>>;
```
gastonfournier, line 19: And here
```suggestion
    setProject: React.Dispatch<React.SetStateAction<string | undefined>>;
```

nunogois:
Fair comments @gastonfournier - Tried addressing them in https://github.com/Unleash/unleash/pull/3290/commits/18bf2826cf2bc29ddd52b7adeb6b85ff6ff1bbb9 by explicitly using `null` instead of `""`.

gastonfournier (approved):
LG!

PR #3293, opened 2023-03-10
makes the 'Should update last seen for known toggles test' no longer …

sighphyre:
The patched test is currently depending on runtime to take more than a millisecond to update the tested property. That's not always true and more so on a fast machine, which makes this test flakey. This forces the old timestamp to be 100 ms in the past so that the checked property must be at least 100 ms different if the update occured


gastonfournier (approved):
G!

PR #3295, opened 2023-03-10
chore: add project-specific-segment permission

gastonfournier:
## About the changes
This adds new permission to control access to project specific segments 

This should be used later for https://linear.app/unleash/issue/2-743/have-a-project-specific-configuration-section:
image (11)


nunogois (approved):
LGTM

PR #3296, opened 2023-03-10
fix: remove remaining warnings in tests

kwasniew:
## About the changes

Remove warning from test that was touching scheduler, so that we're one step closer to zero warnings in CI build.
Warnings came from scheduler calling nonexistent methods from the test mock.

Also: missing mock methods for PAT last seen cc: @nunogois


nunogois (approved):
LGTM

PR #3300, opened 2023-03-10
fix: segment project null handling

nunogois:
~~Should we handle this on the store layer instead~~? 🤔
Fixing this on the store layer. Effectively, frontend is able to send `project: null` and even if that gets magically converted to `""` it's OK since we're covering that use case on the store layer. Backend response will be `project: null` as well, so it should be consistent.


sighphyre (approved):
G

daveleek (approved):
G!

PR #3301, opened 2023-03-10
feat: disable verbosity from db-migrate

kwasniew:
## About the changes

* db migrate was reading --verbose setting that WebStorm/Intellij set by default so it was impossible to silence DB migrations in dev
* also fixing last e2e test logging unnecessary stuff to stderr


sjaanus (approved):
LGTM!

PR #3304, opened 2023-03-13
fix: assume undefined instead of null on segment project

nunogois:
Assume `undefined` instead of `null` for project in terms of interfacing with segments: If the `project` field is not present, that means that it is not set, which means we're dealing with a "global" segment. If we want to unset `project` and make a segment global, we simply don't send the `project` property on our PUT method.


gastonfournier (approved):
LGTM! We have to take a look at coercion and what are the consequences

PR #3306, opened 2023-03-14
fix: type checker happy

kwasniew:
## About the changes
Fixing all type error in the project service. In the next PR I'm planning to add new code here but wanted to start with a clean slate.


sjaanus (approved):

src/lib/services/project-service.ts
```diff
@@ -269,10 +269,12 @@ export default class ProjectService {
         if (!isCompatibleWithTargetProject) {
             throw new IncompatibleProjectError(newProjectId);
         }
+        const createdBy = getCreatedBy(user);
+        if (!createdBy) return;
```
gastonfournier, line 273:
```suggestion
```
After adding `|| 'unknown'` is it possible to get undefined here?

src/lib/services/project-service.ts
```diff
@@ -412,6 +414,7 @@ export default class ProjectService {
         const role = await this.accessService.getRole(roleId);
         const group = await this.groupService.getGroup(groupId);
         const project = await this.getProject(projectId);
+        if (!group.id) return;
```
gastonfournier, line 417: If group is not found should we exit or assign no group?

gastonfournier (commented):
Let's be careful with things like:
```
if (!group.id) return;
```
Because it may silently succeed but not make any change. I'm not sure if this is the expected behavior or not, it depends on the use case

src/lib/services/project-service.ts
kwasniew: Currently we'd fail with an error. I don't think this situation should ever happen but because some other type in unrelated part of the codebase we have to deal with it here. Another option would be to throw IllegalStateException. I'd expect one of our existing tests to tell me if I broke something.

src/lib/services/project-service.ts
kwasniew: Good catch. I also noticed that we have createdBy || 'system-user' in other places so I'm thinking about defaulting to 'system-user' instead of 'unknown'. Again this should probably not happen in practice but we should be defensive.

gastonfournier: I think we're using `'unknown'` in other cases, and I wonder if there's such need... throwing an exception might be better, although it might also uncover some hidden errors we already have. Not a problem but I'd leave it in sandbox and maybe demo instance for a while and monitor logs.

src/lib/services/project-service.ts
gastonfournier: If we'd fail, then I'd fail here (fail fast), but if we return we're changing the behavior. Might not be that bad because it's an edge case, but the takeaway is that we need to be careful

kwasniew: After this comment I'll default to replicating current behavior and failing but with a more explicit error message

src/lib/services/project-service.ts
kwasniew: Leaving 'unknown' for now. And agree not to change behavior. So throwing TypeError for now.

PR #3307, opened 2023-03-14
feat: improve variants modal UI/UX

nunogois:
https://linear.app/unleash/issue/2-758/add-variant-improve-the-flow

image
image

### Variants form

- Fix variants edit form to follow natural tab order;
- Update variants form UI to new design with multiple improvements and fixes, including a sticky header;
- New variants are now added at the bottom of the edit form instead of at the top, with a smooth scroll and focus;

### Change requests

- On the variants diff, use variant names instead of index;
- Use an object-based diff logic (instead of array-based) for cleaner diffs on variants (thanks @thomasheartman !);
- Display a table with the new variants data and display the diff on a `TooltipLink`;
- Adapt strategy CR changes to the new `TooltipLink` logic for consistency;

### Other

- `TooltipLink` and `Badge` components are now tab-selectable;
- Small enhancements, refactors and improvements;


frontend/src/component/common/FormTemplate/FormTemplate.tsx
```diff
@@ -36,7 +36,6 @@ const StyledContainer = styled('section', {
     width: '100%',
     display: 'flex',
     margin: '0 auto',
-    overflow: 'hidden',
```
nunogois, line 39: Having this set broke `position: sticky` in child components. Removing this had no noticeable negative side-effects that I could find 🤷 

Seems like an old setting: 
https://github.com/Unleash/unleash-frontend/pull/739/files#diff-548c3ebed76c71107c101d96f139fb5fb32649ddd404a87fa1d87b12bf2254b5R10

chriswk (approved):
LGTM

NicolaeUnleash (commented):
This looks great Nuno

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/VariantPatch/VariantPatch.tsx
```diff
@@ -39,12 +52,39 @@ export const VariantPatch = ({
...
+                            preData={preData}
+                            data={change.payload.variants}
+                        />
+                    }
+                    tooltipProps={{
+                        maxWidth: 500,
+                        maxHeight: 600,
+                    }}
+                >
+                    Updating variants:
```
gastonfournier, line 69:
```suggestion
                    Updating variants to:
```
I think this makes it clear that we're showing the end result after the CR is applied.

gastonfournier (approved):
Just a small detail, but looks great!

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/VariantPatch/VariantPatch.tsx
nunogois: WDYT @NicolaeUnleash @thomasheartman ?

thomasheartman: Yes, I very much agree with this suggestion 🙋🏼

frontend/src/component/events/EventDiff/EventDiff.tsx
```diff
@@ -12,9 +12,13 @@ const DIFF_PREFIXES: Record<string, string> = {
 
 interface IEventDiffProps {
     entry: Partial<IEvent>;
+    sort?: (a: any, b: any) => number;
```
thomasheartman, line 15: Could we avoid the use of `any` here? I've got the impression that we're generally trying to avoid it in the codebase. How about one of these?

I don't know how the interface is used other places, though, so it might be incovenient 🤷🏼 

Either type param (with a an optional default to avoid causing issues at other places where it's used):

```suggestion
interface IEventDiffProps<T = { key: string }> {
    entry: Partial<IEvent>;
    sort?: (a: T, b: T) => number;
```

Or lock it down to only use the specific type

```suggestion
interface IEventDiffProps {
    entry: Partial<IEvent>;
    sort?: (a: {key: string}, b: {key: string}) => number;
```

wdyt?

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/VariantPatch/Diff.tsx
```diff
@@ -17,13 +17,17 @@ interface IDiffProps {
     data: any;
 }
 
+const variantsArrayToObject = (arr: any) =>
+    arr.reduce((a: any, v: any) => ({ ...a, [v.name]: v }), {});
```
thomasheartman, line 21: Same question regarding any: is it something we're avoiding or is it fine?

If we don't want anys, how about this (also a couple of parameter naming suggestions)? 

```suggestion
const variantsArrayToObject = (variants: { name: string }[]) =>
    variants.reduce((acc, { name, ...rest }) => ({ ...acc, [name]: rest }), {});
```

Would the linter be happy with eliding the types in the lambda? Because otherwise it gets kind of verbose 🙈 

```suggestion
const variantsArrayToObject = (variants: { name: string }[]) =>
    variants.reduce(
        (acc: { [name: string]: {} }, { name, ...rest }: { name: string }) => ({
            ...acc,
            [name]: rest,
        }),
        {},
    );
```

thomasheartman (approved):
Yeah, this looks great to me 😄  I've got a few code-level questions / suggestions, but you're free to do with them what you will. I don't want them to be blocking.

On the UI-side, I've only got a little suggestion: when you delete a variant, it currently shows up as:
> \- variantName

Would it make sense to add a "(deleted)" or something to make it clearer what happened:
> \- variantName (deleted)

Maybe? Maybe not. And again, definitely not a blocking issue.

Well done 🙌🏼

frontend/src/component/events/EventDiff/EventDiff.tsx
nunogois: That's a valid concern, thanks. Addressed in https://github.com/Unleash/unleash/pull/3307/commits/b9e702fe146512908f390b687bae4e99188f47bb.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/VariantPatch/Diff.tsx
nunogois: Yep, all great suggestions. Addressed in https://github.com/Unleash/unleash/pull/3307/commits/4e802032923a3e557411516547617897f7998784.

nunogois:
> 

https://github.com/Unleash/unleash/pull/3307/commits/5fdf808bee36ee56244f8e218361620d4fe396b4 seems to achieve what you're describing:

image

thomasheartman:
Looks great! 💯

PR #3310, opened 2023-03-14
feat: expose project members

kwasniew:
## About the changes

Exposing project members and checking if a user is a member. We need this in the notifications service to send notifications to all project members. We also need this for checking who should be able to submit change requests in a protected model. Access service seems to be the holder of the core information so he owns the membership check.

Also fixed TS errors in all the files I touched.

### DIFF
--- a/src/lib/services/access-service.ts
+++ b/src/lib/services/access-service.ts
@@ -31,6 +31,7 @@ import InvalidOperationError from '../error/invalid-operation-error';
 import BadDataError from '../error/bad-data-error';
 import { IGroupModelWithProjectRole } from '../types/group';
 import { GroupService } from './group-service';
+import { uniqueByKey } from '../util/unique';
 
 const { ADMIN } = permissions;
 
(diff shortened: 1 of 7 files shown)


src/test/e2e/services/access-service.e2e.test.ts
```diff
@@ -18,7 +18,7 @@ import { FavoritesService } from '../../../lib/services';
 
 let db: ITestDb;
 let stores: IUnleashStores;
-let accessService;
+let accessService: AccessService;
```
kwasniew, line 21: this triggered many TS error that I fixed as part of this PR

src/lib/services/access-service.ts
```diff
@@ -381,10 +382,10 @@ export class AccessService {
             const userIdList = userRoleList.map((u) => u.userId);
             const users = await this.accountStore.getAllWithId(userIdList);
             return users.map((user) => {
-                const role = userRoleList.find((r) => r.userId == user.id);
+                const role = userRoleList.find((r) => r.userId == user.id)!;
```
FredrikOseberg, line 385:
```suggestion
                const role = userRoleList.find((role) => role.userId == user.id)!;
```

FredrikOseberg (approved):
LGTM

Issue #3221, opened 2023-02-28
Connection string does not support multiple hosts


lukos:
### Describe the bug

On startup, the unleash Docker server fails with an error, "Failed to migrate db error: database ",10.240.0.4,10.0.0.5/unleash" does not exist" in the pod logs.

This appears to be caused by a connection string in the format: `postgresql://unleash_user:mypassword@10.240.0.5,10.240.0.4,10.0.0.5/unleash?sslmode=require`

### Expected behavior

The documentation [here](https://docs.getunleash.io/reference/deploy/configuring-unleash#database-configuration) says that the property `databaseUrl` is "A string that matches the [libpq connection string](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING), such as...."

And if you click on the link, the connection string should support multiple hosts (and we use the same format on other microservices) since it permits us to failover postgres as needed without breaking a critical service like unleash in the process..

### Unleash version

4.21.0

### Subscription type

Open source

### Hosting type

Self-hosted

thomasheartman:
@lukos Sorry for the delayed response and thanks for reporting this! 🙌🏼 

I've had a look at it now, and I can confirm that I **can** replicate this locally by running Unleash in development mode (that is: without docker). 

We'll look into this for sure!

thomasheartman:
So diving a little deeper on this, this seems to be an [issue with the library we use](https://github.com/brianc/node-postgres/issues/2308). That means that fixing this is going to be a little trickier. It might be that the docs are a little misleading and that they need updating. Anyway, I'll bring it up with the team and get their input to see what they say.

thomasheartman:
I'd like to follow up on this. It doesn't look like swapping out the library is something we're going to prioritize any time soon. I've updated the docs to specify that we only support single-host strings for now. 

That said, I think this functionality would be great, so if you're willing to make a PR that fixes it, we'd be happy to help guide you? ☺️

PR #3312, opened 2023-03-14
docs: clarify that we don't support multi-host db urls

thomasheartman:
This PR updates the docs to clarify that we don't support multi-host strings in the databaseUrl configuration variable.

The docs used to just mention that we accept libpq strings, but as reported in #3221, we **don't** support multi-host strings.

I did some digging and found that this is because [node-postgres -- the library we use to connect to postgres -- itself does not support multi-host strings](https://github.com/brianc/node-postgres/issues/2308).

To fix this, we would need to replace the library we use or otherwise wrap it. While this would be a better long-term solution, we don't have the capacity to it right now. But we can update the docs to better reflect reality.


chriswk (approved):
LGTM

PR #3314, opened 2023-03-15
feat: bulk delete features

sjaanus:


src/lib/routes/admin-api/project/project-archive.ts
Tymek, line 1: How things related to the archive are structured is not intuitive. It would be nice to separate it to a domain in the future.

Tymek (approved):

PR #3315, opened 2023-03-15
chore: refactor segments to stop depending on the implementation

gastonfournier:
## About the changes
- Introducing ISegmentService interface to decouple from the actual implementation
- Moving UpsertSegmentSchema to OSS to be able to use types
- Added comments where our code is coupled with segments just to highlight and have a conversation about some use cases if needed, but they can be removed before merging
- Removed segment service from some project features as it was not used


src/lib/services/feature-toggle-service.ts
```diff
@@ -352,7 +352,10 @@ class FeatureToggleService {
         const { featureName, projectId, environment } = context;
         await this.validateFeatureContext(context);
 
-        if (strategyConfig.constraints?.length > 0) {
+        if (
+            strategyConfig.constraints &&
+            strategyConfig.constraints.length > 0
+        ) {
```
gastonfournier, line 358: This makes the compiler happier

gastonfournier:
> I think this looks good, I like the separation into a new upsert schema. :)

Thanks! this was already present in enterprise

PR #3316, opened 2023-03-15
feat: Default project mode open

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/db/project-store.ts
+++ b/src/lib/db/project-store.ts
@@ -134,7 +134,7 @@ class ProjectStore implements IProjectStore {
             memberCount.map((c) => [c.project, Number(c.count)]),
         );
         return projectsWithFeatureCount.map((r) => {
-            return { ...r, memberCount: memberMap.get(r.id) };
+            return { ...r, memberCount: memberMap.get(r.id) || 0 };
         });
     }
 
@@ -149,6 +149,7 @@ class ProjectStore implements IProjectStore {
             featureCount: Number(row.number_of_features) || 0,
             memberCount: Number(row.number_of_users) || 0,
             updatedAt: row.updated_at,
+            mode: 'open',
         };
     }
 
@@ -456,6 +457,7 @@ class ProjectStore implements IProjectStore {
             createdAt: row.created_at,
             health: row.health ?? 100,
             updatedAt: row.updated_at || new Date(),
+            mode: 'open',
         };
     }
 }
--- a/src/test/fixtures/fake-project-store.ts
+++ b/src/test/fixtures/fake-project-store.ts
@@ -52,6 +52,7 @@ export default class FakeProjectStore implements IProjectStore {
             ...project,
             health: 100,
             createdAt: new Date(),
+            mode: 'open',
         };
         this.projects.push(newProj);
         return newProj;
@@ -129,7 +130,7 @@ export default class FakeProjectStore implements IProjectStore {
     }
 
     async updateHealth(healthUpdate: IProjectHealthUpdate): Promise<void> {
-        this.projects.find((p) => p.id === healthUpdate.id).health =
+        this.projects.find((p) => p.id === healthUpdate.id)!.health =
             healthUpdate.health;
     }
 
--- a/src/lib/types/model.ts
+++ b/src/lib/types/model.ts
@@ -366,7 +366,10 @@ export interface IProject {
     createdAt?: Date;
     updatedAt?: Date;
     changeRequestsEnabled?: boolean;
+    mode: ProjectMode;
 }
+export type ProjectMode = 'open' | 'protected';
+
 export interface ICustomRole {
     id: number;
     name: string;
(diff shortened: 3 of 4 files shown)


src/lib/db/project-store.ts
```diff
@@ -134,7 +134,7 @@ class ProjectStore implements IProjectStore {
             memberCount.map((c) => [c.project, Number(c.count)]),
         );
         return projectsWithFeatureCount.map((r) => {
-            return { ...r, memberCount: memberMap.get(r.id) };
+            return { ...r, memberCount: memberMap.get(r.id) || 0 };
```
FredrikOseberg, line 137: What the hell is `r` ? Let's clean up these legacy one character identifiers while we are here.

src/test/fixtures/fake-project-store.ts
```diff
@@ -129,7 +130,7 @@ export default class FakeProjectStore implements IProjectStore {
     }
 
     async updateHealth(healthUpdate: IProjectHealthUpdate): Promise<void> {
-        this.projects.find((p) => p.id === healthUpdate.id).health =
+        this.projects.find((p) => p.id === healthUpdate.id)!.health =
```
FredrikOseberg, line 133:
```suggestion
        this.projects.find((project) => project.id === healthUpdate.id)!.health =
```

FredrikOseberg (approved):
LGTM

src/lib/openapi/spec/project-schema.ts
```diff
@@ -55,6 +55,14 @@ export const projectSchema = {
...
                 '`true` if the project was favorited, otherwise `false`.',
         },
+        mode: {
+            type: 'string',
+            enum: ['open', 'protected'],
+            example: 'open',
+            nullable: true,
+            description:
+                'A mode of the project affecting what actions are possible in this project.',
+        },
```
thomasheartman, line 65: Out of curiosity, is there a difference between `undefined` and `null` in this case? 

The property isn't listed in the schema's required properties, so we're saying it might not be there (`undefined`), but we're also saying that it could be `null`. Is there a need for that? Isn't one or the other enough?

src/lib/db/project-store.ts
kwasniew: Usually that's a name clash with some other scope that triggers linter

src/lib/openapi/spec/project-schema.ts
kwasniew: Good question. Theoretically it should be either open | protected or not defined. We can probably skip null. But since I don't know full implications of the change I wanted to be on the safe side and keep it as loose as possible. Over time I'd like to make it more precise. It's easier to go this way (from loose to strict) than the other way round (from strict to loose).

thomasheartman: Wait, so is this a response schema or a request schema? Because whether it's easier to go from loose to strict or the other way depends on that. If it's a response schema, then yes, it's easier to narrow it down and make responses more accurate. On the other hand, "be conservative in what you send, be liberal in what you accept" says that we should know which one we're sending and use the types to enforce that.

src/lib/db/project-store.ts
kwasniew: This one should be easy to clean up though. Didn't want to blow up the PR size but since you mentioned this I'm renaming those.

src/lib/openapi/spec/project-schema.ts
```diff
@@ -55,6 +55,14 @@ export const projectSchema = {
             description:
                 '`true` if the project was favorited, otherwise `false`.',
         },
+        mode: {
+            type: 'string',
+            enum: ['open', 'protected'],
+            example: 'open',
+            nullable: true,
+            description:
+                'A mode of the project affecting what actions are possible in this project.',
```
thomasheartman, line 64: Can we add to the description what it means when the property is `null` or missing? So that consumers don't have to wonder?

```suggestion
                'A mode of the project affecting what actions are possible in this project.',
```

src/lib/openapi/spec/project-schema.ts
kwasniew: Read schema as there's memberCount, featureCount etc.

src/lib/openapi/spec/project-schema.ts
kwasniew: What documentation strategy do we have for experimental/WIP capabilities?

kwasniew: Also we don't have this type of information for all other fields in this schema and they are mostly optional.

kwasniew: Added extra info about a rollout of a new feature so that people on the bleeding edge are not confused about this field

thomasheartman: Great! So what you're saying (if I understand the changes in d87a8e8205c6245486aeee0494255237df3a2219 correctly), is that this will only be `null` during rolling out this feature? Also the description doesn't say anything about when it's undefined. What does that mean?

And you're right. The other fields don't have this info today, but that doesn't mean that they can't be better. We should strive to explain all the fields completely to the end user. That can either be on the property-level or on the schema-level.

PR #3317, opened 2023-03-15
fix: project permissions

FredrikOseberg:
## About the changes
This PR fixes an issue where permissions would not allow you to create an API token in the new project API management view. In addition, it fixes an issue where if you missed the permission to read API tokens, you would be blocked from creating one as well. 

Closes #


thomasheartman (approved):
Yeah, this looks great to me! But GitHub seems to complain that you don't need to import the `CREATE_API_TOKEN`  on line 7 anymore, so let's remove that before merging?

andreas-unleash (approved):
LG

PR #3318, opened 2023-03-15
fix: add container for EditGroups

FredrikOseberg:
This PR fixes an issue that would manifest when you navigated directly to the groups and tried to edit the group from the group card. When you were redirected to the edit groups view, the data fields would be empty. The root cause of this issue is that the data has not arrived over the wire before we load the UI, resulting in the UI showing an empty dataset. 

Normally, if this data came directly from useSWR, the app would re-render and display the data correctly. However, in this instance we are passing the data to an intermediary hook (useGroupForm) which does not have the same re-render logic. Therefore we are effectively working from stale data. We solve this issue by adding a container that resolves the group before the edit form is rendered, and that invokes the intermediary hook only when the data is available.

### DIFF
--- a/frontend/src/component/admin/groups/EditGroup/EditGroup.tsx
+++ b/frontend/src/component/admin/groups/EditGroup/EditGroup.tsx
@@ -13,10 +13,24 @@ import { useGroup } from 'hooks/api/getters/useGroup/useGroup';
 import { UG_SAVE_BTN_ID } from 'utils/testIds';
 import { GO_BACK } from 'constants/navigate';
 import { useGroups } from 'hooks/api/getters/useGroups/useGroups';
+import { IGroup } from 'interfaces/group';
 
-export const EditGroup = () => {
+export const EditGroupContainer = () => {
     const groupId = Number(useRequiredPathParam('groupId'));
-    const { group, refetchGroup } = useGroup(groupId);
+    const { group } = useGroup(groupId);
+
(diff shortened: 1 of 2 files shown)


kwasniew (approved):

frontend/src/component/admin/groups/EditGroup/EditGroup.tsx
```diff
@@ -13,10 +13,24 @@ import { useGroup } from 'hooks/api/getters/useGroup/useGroup';
...
+    return <EditGroup group={group} groupId={groupId} />;
+};
+
+interface IEditGroupProps {
+    group: IGroup;
+    groupId: number;
+}
+
+export const EditGroup = ({ group, groupId }: IEditGroupProps) => {
+    const { refetchGroup } = useGroup(groupId);
```
nunogois, line 33: I'm guessing in the end it doesn't matter much, but does it make sense to call the hook again for `refetchGroup` instead of passing it as a prop?

nunogois (approved):
LGTM - Thanks for the great PR description ❤️

frontend/src/component/admin/groups/EditGroup/EditGroup.tsx
FredrikOseberg: True. I wanted to pass as little as possible, but I think your suggestion makes more sense after thinking about it. I'll make the change!

PR #3322, opened 2023-03-15
refactor: move project membership check from access to project

kwasniew:
## About the changes
Project user checking move from access service to project service. It was misplaced before. Calling it a project user to distinguish it from a member that has established meaning in our domain.


andreas-unleash (approved):
LGTM!

PR #3323, opened 2023-03-15
refactor: fix compilation issues in features service

gastonfournier:
## About the changes
This PR fixes a bunch of null check errors: https://github.com/Unleash/unleash/actions/runs/4509279284/jobs/7938853559#step:5:39


src/lib/services/feature-toggle-service.ts
```diff
@@ -260,24 +260,20 @@ class FeatureToggleService {
...
-            }
-        } else {
-            if (contextDefinition?.legalValues?.length > 0) {
-                validateLegalValues(
-                    contextDefinition.legalValues,
-                    constraint.values,
-                );
-            }
+                ? constraint.value
+                : constraint.values;
```
gastonfournier, line 272: Maybe here we could do:
`const valuesToValidate = constraint.value || constraint.values;` but I wanted to retain the same logic. I don't know if we have cases where we are receiving both value and values in constraints

kwasniew (approved):
Thanks for reducing unsafe typing footprint

PR #3325, opened 2023-03-15
feat: filter segments by project on the strategy

nunogois:
https://linear.app/unleash/issue/2-737/when-listing-segments-project-specific-segments-should-also-be-present

Simple filter on the UI for segments that are either global or for the specific project.


gastonfournier (approved):
LG! Let's try it out!

PR #3331, opened 2023-03-16
fix: remove proxy return all toggles functionality

andreas-unleash:
This PR removes the return all toggles functionality. Removes the flag as well
## About the changes

Closes # [1-778](https://linear.app/unleash/issue/1-778/remove-proxyalltoggles-functionality)


thomasheartman (approved):
I like it! I don't know if there's a way to verify that all code related to this has been removed, but this seems fairly comprehensive 😄 

Question for a discussion: is there any way we can test that it's not there anymore? I mean, it's hard to test the absence of something. Anyway, not worth stalling it for that, I think.

In short: nice! 🥳

FredrikOseberg (approved):
LGTM

PR #3332, opened 2023-03-16
fix: bug fix - `end` the response when projectScopedStickiness is disabled

andreas-unleash:
This PR fixes a bug where the api call to getProjectSettings was left hanging because no `end` was specified on the server

## About the changes

Closes #

### DIFF
--- a/src/lib/routes/admin-api/project/index.ts
+++ b/src/lib/routes/admin-api/project/index.ts
@@ -28,7 +28,7 @@ import { OpenApiService, SettingService } from '../../../services';
 import { IAuthRequest } from '../../unleash-types';
 import { ProjectApiTokenController } from './api-token';
 import ProjectArchiveController from './project-archive';
-import NotFoundError from 'lib/error/notfound-error';
+import NotFoundError from '../../../error/notfound-error';
 
 const STICKINESS_KEY = 'stickiness';
 const DEFAULT_STICKINESS = 'default';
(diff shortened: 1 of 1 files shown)


nunogois (approved):
Thanks for the quick fix @andreas-unleash 🙂 

In the future in such cases you could also simply throw `NotFoundError` which will achieve the same end result in a simpler way:

```TS
if (!this.config.flagResolver.isEnabled('projectScopedStickiness')) {
  throw new NotFoundError('Project scoped stickiness is not enabled');
}
```

See: https://github.com/Unleash/unleash/blob/0db5c14f3999628e5ea1265a89fecb217dd811e7/src/lib/routes/admin-api/tag.ts#L217-L219

andreas-unleash:
> Thanks for the quick fix @andreas-unleash 🙂
> 
> In the future in such cases you could also simply throw `NotFoundError` which will achieve the same end result in a simpler way:
> 
> ```ts
> if (!this.config.flagResolver.isEnabled('projectScopedStickiness')) {
>   throw new NotFoundError('Project scoped stickiness is not enabled');
> }
> ```
> 
> See:
> 
> https://github.com/Unleash/unleash/blob/0db5c14f3999628e5ea1265a89fecb217dd811e7/src/lib/routes/admin-api/tag.ts#L217-L219

Thanks, I didn't know that. That is much better

PR #3334, opened 2023-03-16
feat: project mode

kwasniew:
## About the changes

Screenshot 2023-03-16 at 15 23 47

* Adding UI and backend service for project mode (open or protected)
* fixing type errors
* project mode is made mandatory for the backend internals since create and update project have joi schema that can provide a default value at the edges
* similarly our read query always default to open for projects that never had this option set explicitly

Followup work:
* extend cypress tests for this feature
* hide buttons for change requests client side when project mode is protected and user is not a member

### DIFF
--- a/frontend/src/component/project/Project/CreateProject/CreateProject.tsx
+++ b/frontend/src/component/project/Project/CreateProject/CreateProject.tsx
@@ -19,6 +19,7 @@ const CreateProject = () => {
     const {
         projectId,
         projectName,
+        projectMode,
         projectDesc,
         setProjectId,
         setProjectName,
@@ -28,6 +29,7 @@ const CreateProject = () => {
         validateProjectId,
         validateName,
         setProjectStickiness,
+        setProjectMode,
         projectStickiness,
         errors,
     } = useProjectForm();
@@ -92,8 +94,10 @@ const CreateProject = () => {
                 projectId={projectId}
                 setProjectId={setProjectId}
                 projectName={projectName}
+                projectMode={projectMode}
                 projectStickiness={projectStickiness}
                 setProjectStickiness={setProjectStickiness}
+                setProjectMode={setProjectMode}
                 setProjectName={setProjectName}
                 projectDesc={projectDesc}
                 setProjectDesc={setProjectDesc}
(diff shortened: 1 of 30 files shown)


frontend/src/component/project/Project/CreateProject/CreateProject.tsx
```diff
@@ -92,8 +94,10 @@ const CreateProject = () => {
                 projectId={projectId}
                 setProjectId={setProjectId}
                 projectName={projectName}
+                projectMode={projectMode}
```
kwasniew, line 97: this is props drilling antipattern but I don't want to refactor as part of this PR

frontend/src/component/project/Project/ProjectForm/ProjectForm.tsx
```diff
@@ -113,6 +120,30 @@ const ProjectForm: React.FC<IProjectForm> = ({
                         </>
                     }
                 />
+                <ConditionallyRender
+                    condition={Boolean(projectModeFlag)}
```
kwasniew, line 124: this flag should be enough w/o additional enterprise check since the parent component hides create project screen

src/lib/db/project-store.ts
```diff
@@ -460,7 +486,7 @@ class ProjectStore implements IProjectStore {
             createdAt: row.created_at,
             health: row.health ?? 100,
             updatedAt: row.updated_at || new Date(),
-            mode: 'open',
+            mode: row.project_mode || 'open',
```
kwasniew, line 489: always have a default for projects that don't have explicit setting

src/lib/features/export-import-toggles/export-import-permissions.e2e.test.ts
```diff
@@ -87,6 +87,7 @@ const createProject = async () => {
         name: DEFAULT_PROJECT,
         description: '',
         id: DEFAULT_PROJECT,
+        mode: 'open' as const,
```
kwasniew, line 90: now it's required so I had to touch lots of places with the same change

andreas-unleash (approved):
LG

PR #3335, opened 2023-03-16
feat: implement project-scoped segments in project settings

nunogois:
https://linear.app/unleash/issue/2-743/have-a-project-specific-configuration-section

image

Adds the "segments" option to project settings, providing the usual CRUD operations but scoped to the specific project.

### DIFF
--- a/frontend/src/component/common/FormTemplate/FormTemplate.tsx
+++ b/frontend/src/component/common/FormTemplate/FormTemplate.tsx
@@ -34,6 +34,7 @@ const StyledContainer = styled('section', {
     minHeight: modal ? '100vh' : '80vh',
     borderRadius: modal ? 0 : theme.spacing(2),
     width: '100%',
+    height: '100%',
     display: 'flex',
     margin: '0 auto',
     [theme.breakpoints.down(1100)]: {
--- a/frontend/src/component/segments/SegmentTable.tsx
+++ b/frontend/src/component/segments/SegmentTable.tsx
@@ -28,8 +28,10 @@ import { Search } from 'component/common/Search/Search';
 import { useConditionallyHiddenColumns } from 'hooks/useConditionallyHiddenColumns';
 import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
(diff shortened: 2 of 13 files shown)


frontend/src/component/common/FormTemplate/FormTemplate.tsx
```diff
@@ -34,6 +34,7 @@ const StyledContainer = styled('section', {
     minHeight: modal ? '100vh' : '80vh',
     borderRadius: modal ? 0 : theme.spacing(2),
     width: '100%',
+    height: '100%',
```
nunogois, line 37: Not sure why this was not set before, but it seems to provide a more straightforward behavior without having to set the modal prop.

frontend/src/component/segments/SegmentTable.tsx
```diff
@@ -39,17 +41,22 @@ export const SegmentTable = () => {
...
                 projectId: 'Project',
-            })
-        );
-    }, [segments]);
+            });
+        }
+
+        if (projectId) {
+            return segments.filter(({ project }) => project === projectId);
+        }
```
gastonfournier, line 56: I believe you should move the segments filter before line 44 where you check `if (!segments)`, because you could have segments at the beginning but maybe not after filtering by project

gastonfournier (approved):
LG! There's only one doubt I have but can be addressed later

chriswk (approved):
Gaston seems to have addressed my worries already, so it's a :+1: from me. :)

frontend/src/component/segments/SegmentTable.tsx
nunogois: I think I understand what you mean but I believe the logic is a bit different here than what you assumed in your comment. 
If you check line 44 again, we are checking for truthy `segments`, not their length. The purpose of the first check on line 44 is to fill segments with placeholder information for loading purposes (in case `segments` is still `undefined`, let's fill them with mock info to show as loading skeletons or what not). 
From that point forward, we are totally fine with having `[]` for segments here, and the table should handle that by displaying an appropriate message. In fact, we could not even filter them on line 55 if they were still `undefined`.

gastonfournier: Gotcha! I thought it was something different. I also see it was there before :+1:

PR #3340, opened 2023-03-17
Feat/add cypress tests for project scoped stickiness

andreas-unleash:
## About the changes

Closes #


kwasniew (approved):
Same question as in the enterprise repo. Are we gonna piggy back on the project or create separate settings path?

PR #3341, opened 2023-03-17
docs(a11y): update syntax highlighting theme to fix contrast issues

thomasheartman:
This PR changes the syntax highlighting theme we use in the docs. The reason for the change is that the previous theme we used didn't have sufficient color contrast between its comment strings and its background.

## Options

I decided to use the Oceanic Next color theme that is included with docusaurus. Looking through, the only other dark theme that also passed AA contrast checks was VS Dark. The reason I picked Oceanic over VS Dark is simply that I like it better.

For demonstration, here's what the Oceanic theme looks like:
image

And here's the VS Dark theme:
image

## Bonus change

The PR also includes a fix to render a pricing admonition correctly. It's not related to the PR, but I noticed it rendered incorrectly while I was working on it.

### DIFF
--- a/website/docs/tutorials/quickstart.md
+++ b/website/docs/tutorials/quickstart.md
@@ -110,7 +110,7 @@ Now you can fetch the feature toggle you created and try turning it on / off in 
 
 If you want to set up your own instance for testing purposes you can easily do so by using one of our premade setup kits for Heroku or DigitalOcean.
 
-::: info Pricing
+:::info Pricing
 
 Both DigitalOcean and Heroku will charge you for running Unleash on their services. However, if you use the DigitalOcean link below and are a new user, you will receive $100 in free credits.
 
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -88,6 +88,7 @@ module.exports = {
             ],
         },
         prism: {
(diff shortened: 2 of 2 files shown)


NicolaeUnleash (approved):
Nice.
As long as both pass contrast check, any of this will be a good choice.

Which one is better is very subjective:
- personally i can follow the syntax easier with VS Dark theme, but it's because that is the theme i'm using on my VSC (the comments are green and its exactly the opposite on Oceanic)
- what i noticed is the the screenshot looked much saturated/brighter than the demo on vercel (the image was better that what I'm trying to say here)

thomasheartman:
>Which one is better is very subjective:
>
>   - personally i can follow the syntax easier with VS Dark theme, but it's because that is the theme i'm using on my VSC (the comments are green and its exactly the opposite on Oceanic)
>    - what i noticed is the the screenshot looked much saturated/brighter than the demo on vercel (the image was better that's what I'm trying to say here)

Yeah, safe! It's the opposite for me on the first one 😅 But yeah, that's a matter of preference and what you're used to. 
And as for the screenshot: interesting. Don't know why that'd be.

But anyway, yeah: both of them passed any contrast checks I threw at them, so I just made a choice. We can always change it later 😄

PR #3328, opened 2023-03-15
fix: Eta vulnerable to Code Injection

gastonfournier:
## About the changes
Closes security vulnerabilities [50](https://github.com/Unleash/unleash/security/dependabot/50) [51](https://github.com/Unleash/unleash/security/dependabot/51)

PR #3343, opened 2023-03-17
chore: update docusaurus and deps to v2.3.1

thomasheartman:
This also updates the openapi-docs plugin we use to the most recent beta. Support for 2.3.1 (or 2.3) is not available in any of the stable packages.

## Discussion point

If we can, I think we should hold off on merging this until v2.0.0 of the openapi plugin we use is released. That way, we don't need to take the beta package into production. They've been pretty good with moving fast so it shouldn't be long.

On the other hand, if fixing the ETA issue (#3328) is very urgent, then we can do it right away.


gastonfournier (approved):
I think this looks good, I don't think using the beta versions is a huge problem provided this will help solve 2 security vulnerabilities (which maybe can't be exploited but :shrug:)

gastonfournier (approved):

PR #3346, opened 2023-03-17
Feat project segments permissions

nunogois:
- Adds `UPDATE_PROJECT_SEGMENT` permission checks;
- Allows `PermissionIconButton` to evaluate multiple permissions, just like `PermissionButton`;
- Also includes a possible fix for `hasAccess` in `AccessProvider`.


frontend/src/component/common/PermissionIconButton/PermissionIconButton.tsx
```diff
@@ -13,7 +13,7 @@ import {
 } from 'hooks/useHasAccess';
 
 interface IPermissionIconButtonProps {
-    permission: string;
+    permission: string | string[];
```
nunogois, line 16: Allows `PermissionIconButton` to evaluate multiple permissions, just like `PermissionButton`.

frontend/src/component/providers/AccessProvider/AccessProvider.tsx
```diff
@@ -70,22 +70,12 @@ const checkPermission = (
     if (
         p.permission === permission &&
         (p.project === project || p.project === '*') &&
-        (p.environment === environment || p.environment === '*')
+        (!p.environment ||
+            p.environment === environment ||
+            p.environment === '*')
```
nunogois, line 75: Unrelated slight refactor from the 2 conditions we had before:

```TS
if (
    p.permission === permission &&
    (p.project === project || p.project === '*') &&
    (p.environment === environment || p.environment === '*')
) {
    return true;
}

if (
    p.permission === permission &&
    (p.project === project || p.project === '*') &&
    !Boolean(p.environment)
) {
    return true;
}
```

Should achieve the same end result.

frontend/src/component/providers/AccessProvider/AccessProvider.tsx
```diff
@@ -70,22 +70,12 @@ const checkPermission = (
...
-    ) {
-        return true;
-    }
-
-    return (
-        p.permission === permission &&
-        p.project === undefined &&
-        p.environment === null
-    );
+    return p.permission === permission && !p.project && !p.environment;
```
nunogois, line 80: My use-case was to check both "root" and "project" permissions at the same time (e.g. `permission={[UPDATE_SEGMENT, UPDATE_PROJECT_SEGMENT]}`), however my "root" permission was not correctly evaluated when specifying a `project`. E.g. see: https://github.com/Unleash/unleash/pull/3346/files#diff-09482643904b9601b01cf445d86cbc5fc1c589ed115ce166891ae67358a940a6R23-R24

This change seems to make that possible without breaking anything (hopefully). From my understanding, `p.environment` may be returned as `undefined` now instead of `null`, so we simply check if those fields are falsy instead, which should mean that the underlying permission doesn't care about which project or environment we sent here to check, which should mean we're evaluating a root permission.

gastonfournier: Note this also follows the same logic as the backend: https://github.com/Unleash/unleash/blob/6b165f36635fd633e9c92c163e14593249cd8437/src/lib/services/access-service.ts#L120-L123

gastonfournier (approved):
LGTM!

PR #3348, opened 2023-03-20
refactor: test composition and other error codes

gastonfournier:
## About the changes
Small refactor to showcase how to use [composition to validate different aspects of the response](https://github.com/Unleash/unleash/pull/3348/files#diff-ee4c1bd501b1195162b7a85ed6be348a665288f871abc8e74f64d94361213f9eR361-R367) and checking [different status codes](https://github.com/Unleash/unleash/pull/3348/files#diff-4044a5da3280ef76960bbffd5f36eccb395ac319fe58c4d59ef68a878cbb1a5dR95)


src/test/e2e/helpers/test-helper.ts
```diff
@@ -45,15 +55,29 @@ function httpApis(
...
+            feature: string | CreateFeatureSchema,
             project: string = DEFAULT_PROJECT,
             expectedResponseCode: number = 201,
         ) => {
+            let body = feature;
+            if (typeof feature === 'string') {
+                body = {
+                    name: feature,
+                };
+            }
```
gastonfournier, line 67: I'm not a big fan of having this logic here, but to make it simpler in testing I don't think it hurts as much as if I'm doing this in production code.

src/test/e2e/helpers/test-helper.ts
```diff
@@ -45,15 +55,29 @@ function httpApis(
...
+                .set('Content-Type', 'application/json')
+                .expect(expectedResponseCode);
+        },
+
+        getFeatures(
+            name?: string,
+            expectedResponseCode: number = 200,
+        ): supertest.Test {
+            return request
+                .get(`/api/admin/features${name ? '/' + name : ''}`)
```
gastonfournier, line 80: This follows REST conventions of having the feature resource and being able to get a single one of them. Alternatively, we could have two methods, but again, because this is test code, I thought it was better to simplify

src/test/e2e/api/admin/feature-archive.e2e.test.ts
```diff
@@ -25,78 +19,42 @@ beforeAll(async () => {
...
-            name: 'feature.with.variants',
-            description: 'A feature toggle with variants',
-            variants: [
-                { name: 'control', weight: 50 },
-                { name: 'new', weight: 50 },
-            ],
-        },
-        'test',
-    );
+    await app.createFeature({
```
gastonfournier, line 22: Here we're changing from direct service calls to HTTP API calls. Arguably, we could use services in the setup of the test (or we can even load the desired state of the DB), but because all of our services are exposed through HTTP APIs it doesn't feel bad to do it this way.

src/test/e2e/helpers/test-helper.ts
```diff
@@ -45,15 +59,33 @@ function httpApis(
...
+        },
+
+        getFeatures(
+            name?: string,
+            project?: string,
+            expectedResponseCode: number = 200,
+        ): supertest.Test {
+            const projectSubPath = project ? `projects/${project}/` : '';
+            const featureSubPath = name ? `/${name}` : '';
+            const featuresUrl = `/api/admin/${projectSubPath}features${featureSubPath}`;
```
gastonfournier, line 86: This follows REST conventions of having the feature resource and being able to get a single one of them. Alternatively, we could have two methods, but again, because this is test code, I thought it was better to simplify.

Features and project features responses are different (the first one does not return environments) and I only discovered this difference because of using these functions

gastonfournier: I decided to split the two APIs to make it clear that they're not the same: [8c9edf4](https://github.com/Unleash/unleash/pull/3348/commits/8c9edf444498e14f2931262ee654a040a6e526eb)

src/test/e2e/helpers/test-helper.ts
nunogois: Also a bit unsure about this one, my mind is telling me maybe we should have both `createFeature` and e.g. `createFeatureByName` methods with strict typing, where the second one does this for us before sending it in to `createFeature`. 

But it's like you said, maybe it doesn't hurt as much as we're thinking.

src/test/e2e/helpers/test-helper.ts
```diff
@@ -45,15 +73,44 @@ function httpApis(
...
+            expectedResponseCode: number = 200,
+        ): supertest.Test {
+            const featuresUrl = `/api/admin/features${name ? `/${name}` : ''}`;
+            return request
+                .get(featuresUrl)
+                .set('Content-Type', 'application/json')
+                .expect(expectedResponseCode);
+        },
+
+        getProjectFeatures(
```
nunogois, line 104: These ones also make me think the same as https://github.com/Unleash/unleash/pull/3348/files#r1145853611 - Having plural methods `getFeatures` and `getProjectFeatures` that don't take in a name, and singular `getFeature` and `getProjectFeature` ones that do. Whatever we decide there should maybe apply here as well, either separate into different methods or leave them be and accept that this is good enough.

nunogois (approved):
LGTM - Non-blocking approval but a couple comments below 🙂

src/test/e2e/api/admin/feature-archive.e2e.test.ts
nunogois: I would argue HTTP API calls feel more `e2e` than direct calls to service methods that skip the controller and any other logic before reaching the service.

gastonfournier: Indeed, but in the setup, you just want to get you application state to a known state as soon as possible, it doesn't matter much how you put it in that stage, but if we can do it with HTTP calls exercising all our code while doing so, I think it increases the coverage

src/test/e2e/helpers/test-helper.ts
gastonfournier: I'm more inclined to remove the create feature by name and use the correct schema. But, it's more like a shortcut. In Java I'd use method overloading, but I think this is the best way of accomplishing this in Typescript (or alternatively having 2 functions with different names)

src/test/e2e/helpers/test-helper.ts
gastonfournier: Yes, I feel the same way, but because this is test code and a thin layer on top of a REST API, I think we can let it be

src/test/e2e/helpers/test-helper.ts
nunogois: Yeah exactly. I like the idea of having this helper for easier use that allows creating features simply by passing in the name. Whether that's through a different method that refactors things to be stricter typed or this single method that handles both cases, I'll leave it up to you.

gastonfournier: I think this is fine for now, we can continue to refactor later on

PR #3358, opened 2023-03-21
feat: support updating existing projects without existing settings

kwasniew:
## About the changes
Legacy projects don't have settings and we still need to support updating them.


andreas-unleash (approved):
LG

PR #3359, opened 2023-03-21
Feat segments project safeguards

nunogois:
https://linear.app/unleash/issue/2-811/ability-to-move-project-specific-segments

image

Provides some safeguards to project-specific segments when managing them on a global level (segments page):
- You can always promote a segment to global;
- You can't specify a project if the segment is currently in use in multiple projects;
- You can't specify a different project if the segment is currently already in use in a project;

### DIFF
--- a/frontend/src/component/segments/SegmentProjectAlert.tsx
+++ b/frontend/src/component/segments/SegmentProjectAlert.tsx
@@ -72,7 +72,7 @@ export const SegmentProjectAlert = ({
             <StyledAlert severity="info">
                 You can't specify a project other than{' '}
                 <strong>{availableProjects[0].name}</strong> for this segment
-                because it is used in that project:
+                because it is used here:
                 {projectList}
             </StyledAlert>
         );
--- a/frontend/src/component/segments/SegmentProjectAlert.tsx
+++ b/frontend/src/component/segments/SegmentProjectAlert.tsx
@@ -5,6 +5,11 @@ import { IFeatureStrategy } from 'interfaces/strategy';
 import { Link } from 'react-router-dom';
 import { formatStrategyName } from 'utils/strategyNames';
 
+const StyledUl = styled('ul')(({ theme }) => ({
+    listStyle: 'none',
+    paddingLeft: 0,
+}));
+
 const StyledAlert = styled(Alert)(({ theme }) => ({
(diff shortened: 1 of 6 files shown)


frontend/src/component/segments/SegmentProjectAlert.tsx
```diff
@@ -0,0 +1,90 @@
...
+            </StyledAlert>
+        );
+    }
+
+    if (availableProjects.length === 1) {
+        return (
+            <StyledAlert severity="info">
+                You can't specify a project other than{' '}
+                <strong>{availableProjects[0].name}</strong> for this segment
+                because it is used here:
```
nunogois, line 75: @thomasheartman @NicolaeUnleash we were a bit unsure of what text to use here (both this scenario and the one above), so if you have better suggestions they're welcome as always! ❤️

thomasheartman: does it pertain to this bit?

> You can't specify a different project if the segment is currently already in use in a project;

NicolaeUnleash: nice, i guess this is ok for now, i will let Thomas to have the last word on the text :) 

- Please remove the project bullets and padding, it's enough if we have bullets for feature toggles
- what will happen if the list is very long? does the info has a max height and scroll inside?

nunogois: @thomasheartman `projectList` is the list of projects and features you see in the screenshot on the PR description where the segment is currently being used.

This specific message pertains to:
> You can't specify a different project if the segment is currently already in use in a project

While the one above in the code pertains to:
> You can't specify a project if the segment is currently in use in multiple projects

Thanks for the feedback, I think we'll go with the current text then 🙂 

@NicolaeUnleash Thanks, addressed your first point in https://github.com/Unleash/unleash/pull/3359/commits/da0e351070d4818696b3228099762ecd0c00067d - Hope it looks OK now:
image

Regarding your second point, I didn't specify a max-height. My guess is that the behavior will be consistent with other places in Unleash: If the info alert grows, the modal will scroll if needed, which seems OK to me. If you'd like to specify a height, let me know the value and I can set it, otherwise I vote we leave it as is.

PR #3360, opened 2023-03-21
Project collaboration mode docs

kwasniew:
## About the changes

Project collaboration reference documentation.
This PR is a prerequisite for https://github.com/Unleash/unleash/pull/3357


kwasniew:
Update: in the UI added info the everyone except from viewers outside of the project can submit change requests.

Update: in the docs listed all protected CR submitters explicitly with a link to rbac

website/docs/reference/project-collaboration-mode.md
```diff
@@ -0,0 +1,22 @@
...
+
+:::info Availability
+
+The change requests feature is an enterprise-only feature that was introduced in **Unleash 4.22.0**.
+
+:::
+
+Unleash project collaboration mode is a feature that allows to specify who can submit change requests. Currently, there are two modes:
+- **open** where everyone can submit change requests. This is a default setting for the existing and new Unleash projects.
+- **protected** where admins, editors and project members can submit change requests. Viewers who are not added to projects will not be able to submit change requests.
```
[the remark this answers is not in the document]
kwasniew: Yes, fixing it

website/docs/reference/project-collaboration-mode.md
```diff
@@ -0,0 +1,28 @@
...
+
+:::info Availability
+
+The change requests feature is an enterprise-only feature that was introduced in **Unleash 4.22.0**.
+
+:::
+
+Unleash project collaboration mode is a feature that allows to specify who can submit [change requests](change-requests.md). Currently, there are two modes:
+- **open** where everyone can submit change requests. This is a default setting for the existing and new Unleash projects.
+- **protected** where [admins, project members and owners](rbac.md#standard-roles) can submit change requests. Users who are not added to projects will not be able to submit change requests.
```
FredrikOseberg, line 13: Actually, maybe we could change the wording here. Project members and owners suggest that only these roles can submit change requests. That is not true in the implementation, during my testing I tried giving my viewer a custom role with the least privileges possible and I was allowed to submit CRs. There's a difference between being a project member (holding any role in the project) and having the project role of member / owner in the project.

kwasniew: Do you suggest just saying project members?

FredrikOseberg: Yeah, I think that's an ok terminology. Maybe we also define it in this context, since we have the project member role. Something like `project members (any user holding a role in the project)`

website/docs/reference/project-collaboration-mode.md
```diff
@@ -0,0 +1,28 @@
...
+Project collaboration mode is controlled from the "Create Project" and "Edit Project" screens.
+
+![Create project with collaboration mode](/img/collaboration-mode.png)
+
+When project collaboration mode is protected all users who are not project members or admins will have all the
+change request operations disabled.
+
+## Enabling protected mode for an existing project
+
+To change the project collaboration mode for an existing project you have to edit the project.
```
thomasheartman, line 26: This feels like it's veering into how-to territory, but maybe we can rephrase it a bit to keep it as a reference guide.

Also, what happens if you change a project's collaboration mode after a viewer submits a CR? Does it stay? And what about editing it?

```suggestion
## Project collaboration mode setting

You can set a project's collaboration mode when you create a project and at any point after creation.

![Project creation form with a collaboration mode field and corresponding explanation.](/img/collaboration-mode.png)

### Change requests 

When you change a project's collaboration mode from open to protected, users who do not have the right permissions will lose the ability to create new change requests. 

However, existing change requests created by these users will not be deleted. Any users with open change requests will still be able to edit the change request, regardless of their current permissions in the project.
```

website/docs/reference/project-collaboration-mode.md
kwasniew: I thought that the following sentence "Users who are not added to projects will not be able to submit change requests." was clarifying this. WDYT?

thomasheartman (changes requested):
Nice work! I've got some questions and suggestions that I'd love to discuss with you. See the inline comments for specifics 😄

website/docs/reference/project-collaboration-mode.md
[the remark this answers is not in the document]
thomasheartman: Oh, for context: maybe we can explode this out into two separate sections. That makes it easier to explain each mode in more depth (using multiple paragraphs)

website/docs/reference/project-collaboration-mode.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Project collaboration mode
+---
+
+:::info Availability
+
+The change requests feature is an enterprise-only feature that was introduced in **Unleash 4.22.0**.
```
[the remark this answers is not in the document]
kwasniew: Good catch. Definitely a typo

website/docs/reference/project-collaboration-mode.md
kwasniew: So when you make the project protected then the permission checker kicks in and you can only cancel your change request from this moment.

website/docs/reference/project-collaboration-mode.md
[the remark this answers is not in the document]
kwasniew: Makes sense. Picking and applying your changes. With one minor change. Instead of "Only [admins, project members and owners]" we'll say "only admins and project members"

website/docs/reference/project-collaboration-mode.md
thomasheartman: So you can cancel or have it approved & applied. It cannot be updated?

kwasniew: Correct. I updated the text in docs to reflect it.

kwasniew:
@thomasheartman Please check again since I applied all your suggestions with minor modifications. I had to do it on my machine as I had some other conflicts.

website/docs/reference/project-collaboration-mode.md
```diff
@@ -0,0 +1,43 @@
...
+
+## **Protected** collaboration mode
+
+Only admins and project members can submit change requests in a **protected** project. Other users can not submit change requests.
+
+### Change requests
+
+When you change a project's collaboration mode from open to protected, users who do not have the right permissions will lose the ability to create new change requests.
+
+However, existing change requests created by these users will not be deleted. Any users with open change requests will still be able to cancel the change request.
```
thomasheartman, line 29: I'd like to be explicit about these users not being able to update the change request.

```suggestion
However, existing change requests created by these users will not be deleted. Any users with open change requests will still be able to cancel the change request, but they will **not** be able to update the change request.
```

website/docs/reference/project-collaboration-mode.md
```diff
@@ -0,0 +1,28 @@
...
+![Create project with collaboration mode](/img/collaboration-mode.png)
+
+When project collaboration mode is protected all users who are not project members or admins will have all the
+change request operations disabled.
+
+## Enabling protected mode for an existing project
+
+To change the project collaboration mode for an existing project you have to edit the project.
+
+![Edit project](/img/edit-project.png)
```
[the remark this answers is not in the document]
thomasheartman: Right, because this is one of the few settings that are available from "edit project" instead of "project settings"? What's the reason for that? And why do we have those two separate sections anyway? 

I've added a suggestion for this link higher up, so I'll add one to remove this line too.

```suggestion
```

website/docs/reference/project-collaboration-mode.md
```diff
@@ -0,0 +1,43 @@
...
+When you change a project's collaboration mode from open to protected, users who do not have the right permissions will lose the ability to create new change requests.
+
+However, existing change requests created by these users will not be deleted. Any users with open change requests will still be able to cancel the change request.
+
+## Project collaboration mode setting
+
+You can set a project's collaboration mode when you create a project and at any point after creation.
+
+![Project creation form with a collaboration mode field and corresponding explanation.](/img/collaboration-mode.png)
+
```
thomasheartman, line 36: Mayyyybe something like this?

```suggestion
Because the project's collaboration mode is a core part of the project, the controls to change it are available on the "edit project" page.

![The project-level header section with the "edit project" button highlighted.](/img/edit-project.png)

```

thomasheartman (commented):
Thanks for implementing my suggestions! I've got two more small suggestions regarding the edit description and the what people without permissions can do to existing change requests, but otherwise it's good to me 😄

website/docs/reference/project-collaboration-mode.md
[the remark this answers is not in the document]
kwasniew: We wanted to delivery this capability very soon and it was the quickest and deliberate route. Eventually we'll create project settings in the next quarter.

PR #3361, opened 2023-03-21
docs: notifications

kwasniew:
## About the changes
Documenting notifications


website/docs/reference/notifications.md
```diff
@@ -0,0 +1,31 @@
...
+
+:::info Availability
+
+Notifications were introduced in Unleash 4.22.0 for pro and enteprise customers.
+
+:::
+
+Notifications give your project members hints about interesting events happening in their projects.
+To see your notifications, click on the ring bell icon in the top-right corner of the screen.
+You will only see the notifications from the other project members without the events triggered by yourself.
```
thomasheartman, line 13: I wonder if rewriting it to something like this works better?

Also, is it possible to disable notifications entirely?

```suggestion
Unleash's notifications give you updates when certain events occur in projects that you are part of. The notifications are accessible from the notifications button in the navigation section of the admin UI.

Your notifications will only contain updates from other members; you will not get notifications about actions you perform yourself.
```

For future reference: describing the visual appearance of a button (or where on the page it is) has no real value to users with visual impairments. I also checked the markup of it, and the button doesn't seem to have any text associated with it. So it's a side task, but we should try and make sure that the button has some text that screen readers can latch onto before this goes out (or as soon as possible).

website/docs/reference/notifications.md
```diff
@@ -0,0 +1,31 @@
...
+## Notification events
+
+The following events are supported in notifications:
+* Feature created
+* Feature enabled
+* Feature archived
+* Feature enabled in environment
+* Change request sent to review (sent only to project members who can approve the change request)
+* Change request approved
+* Change request applied
```
thomasheartman, line 31: I think this would be clearer in a more active voice. Plus, I have some questions about some of these.

```suggestion
The following actions will trigger notifications:
* Creating a feature
* Feature enabled # @kwasniew what is the difference between this and enabling a feature in an env?
* Archiving a feature
* Enabling a feature in an environment
* Submitting a change request (sent only to project members who can approve the change request)
* Any change request is approved (or is it only **your** change requests?)
* Any change request is applied (or is it only **your** change requests being applied?)
```

thomasheartman (changes requested):
Won't have the time to play around with the feature today, but I've read through the docs and have some suggestions here.

website/docs/reference/notifications.md
kwasniew: It was a mistake to list feature enabled twice. I took your suggestions but the last two decided to start with a verb to have a consistent list. I also repeated the information that only your project events will trigger notifications.

website/docs/reference/notifications.md
kwasniew: Good catch with being overly specific with the UI details

website/docs/reference/notifications.md
thomasheartman: Ah, yes, good catch! That consistency works better 🙌🏼  But I'm not sure I understand the last two items: does everyone in that project get notifications when you approve/apply a change request or just the person who created the change reqest?

website/docs/reference/notifications.md
```diff
@@ -0,0 +1,30 @@
...
+
+Notifications were introduced in Unleash 4.22.0 for pro and enteprise customers.
+
+:::
+
+Unleash's notifications give you updates when certain events occur in projects that you are part of. The notifications are accessible from the notifications button in the navigation section of the admin UI.
+
+Your notifications will only contain updates from other members; you will not get notifications about actions you perform yourself.
+
+![Notifications](/img/notifications.png)
```
thomasheartman, line 15: Oh, I missed this 😅  Please make sure the alt text is useful to people who can't see the image for whatever reason. "Notifications" doesn't really explain much at all. I've added a suggestion below:

```suggestion
![The notifications overview. It's telling the user about a feature that was enabled in production and that there is a change request ready for review.](/img/notifications.png)
```

website/docs/reference/notifications.md
thomasheartman: @kwasniew Did you see this comments regarding the last two points?

kwasniew: Every project member gets notification about approved/applied Change Request.

kwasniew: The only exception was review notification that's why it got extra description

website/sidebars.js
```diff
@@ -333,6 +333,7 @@ module.exports = {
                         'reference/unleash-context',
                         'reference/change-requests',
                         'reference/maintenance-mode',
+                        'reference/notifications',
```
thomasheartman, line 336: Just realized maintenance mode is also in the wrong place. Would you mind placing these last two around network view, so that it goes:

```markdown
- ...
- login-history
- maintenance-mode
- network-view
- notifications
- ...
```

website/docs/reference/notifications.md
```diff
@@ -0,0 +1,30 @@
...
+
+Your notifications will only contain updates from other members; you will not get notifications about actions you perform yourself.
+
+![The notifications overview. It's telling the user about a feature that was enabled in production and that there is a change request ready for review.](/img/notifications.png)
+
+From within the notifications list, you can:
+* navigate to the features or change request that the notification is for
+* mark all notifications as read
+* filter out unread notifications
+
```
thomasheartman, line 21: Can you disable notifications (either for a project or as a whole) today? Do we plan to add that later?

```suggestion
Notifications can not be disabled.

```

website/docs/reference/notifications.md
```diff
@@ -0,0 +1,30 @@
...
+title: Notifications
+---
+
+:::info Availability
+
+Notifications were introduced in Unleash 4.22.0 for pro and enteprise customers.
+
+:::
+
+Unleash's notifications give you updates when certain events occur in projects that you are part of. The notifications are accessible from the notifications button in the navigation section of the admin UI.
```
thomasheartman, line 11: Are notifications available via the API or are they a UI-only thing today? Are the OpenAPI specs up to date with this already?

thomasheartman (commented):
I think this is pretty much ready, just a few more (tiny) things to sort out 😄

website/docs/reference/notifications.md
kwasniew: OpenAPI specs are up to date even though this API is mostly for the UI. If someone wants to use the API though it's there and fully documented.

website/docs/reference/notifications.md
kwasniew: Currently there's no way to disable notifications so your comment is spot on. We're considering adding configuration options based on the adoption of the capability.

website/docs/reference/notifications.md
```diff
@@ -0,0 +1,31 @@
...
+
+Your notifications will only contain updates from other members; you will not get notifications about actions you perform yourself.
+
+![The notifications overview. It's telling the user about a feature that was enabled in production and that there is a change request ready for review.](/img/notifications.png)
+
+From within the notifications list, you can:
+* navigate to the features or change request that the notification is for
+* mark all notifications as read
+* filter out unread notifications
+Notifications can not be disabled.
```
thomasheartman, line 21: Last thing: you'll need a space here for it not to be part of the last list item.

```suggestion

Notifications can not be disabled.
```

thomasheartman (approved):
Great! Looks good 😄  I added a newline to format the text correctly (hope you don't mind).

Issue #1265, opened 2022-01-13
🟢 Unleash v5 - Breaking changes


ivarconr:
This issue is used to highlight important changes for v5. We have not decided on when we will release v5 of Unleash, but this issue will allow us to collect all breaking changes we want to do. 

Please note that this is not a commitment on our side, and more an indication of items we are currently discussing and possible impacts for that. 

- **DROP "/api/admin/features" API** [(link)](https://docs.getunleash.io/api/admin/features) - Does not support environments, custom permission, etc. Replaced with the new new features api, which supports environments [(link)](https://docs.getunleash.io/api/admin/feature-toggles-v2)
- **DROP [eventHook](https://docs.getunleash.io/deploy/configuring_unleash)** - its out-dated, does not support all event types etc. You should use web-hook or the event-store instead. 
- (✅) **Support for node.js v18 and above (only LTS)**
- **DROP direct support for Auth With Google (SSO)** - Not needed as Sign-In with google supports OIDC out of the box. 
... more to come
- Consider: Migrate to ES modules. This will break CommonJS (`require...`) support 
- (✅) Change default password for Unleash DB connections from `passord` to `password` to avoid confusion 
 - ~**DROP express and migrate to [fastify](https://www.fastify.io/)**. - faster, more secure, and will allow us to automatically generate Open API specifications.~

Helm chart:
- The default unleash chart should not bundle postgres by default from v5 (we can have a "experimental" chart that does this, clearly marked as something you should not run in production. 

Things that will be deprecated:
* state-api (and services)

Under consideration for being deprecated (still not decided)
* Custom strategies 
* Unnecessary built in strategies (e.g. userId strategy is better solved via constraints today).

thomasheartman:
I'd like to propose a few things:

1. Align APIs on how you PUT vs POST vs PATCH and what the different methods do. That means that PUT _should_ do a complete replace, and that a PATCH is what you do for partial updates.
2. Reconsider the PATCH format we used. We discussed this during the PUT discussions previously. PATCHing is hard because the JSON patch syntax is honestly a bit of a hassle. Instead, we should use an alternative format. A suggestion was to simply send pure JSON data and do a regular merge, leaving untouched fields unchanged, and overwriting any fields you pass in.
3. Do more API-side validation of values passed in instead of blindly accepting them: this includes things such as numbers that are invalid in a given context, strings that don't match some casing, etc. The most recent example I can think of was an issue about passing invalid weights to the variants API.
[...]

PR #3362, opened 2023-03-21
fix: containers for node 16, 18 and 20

ivarconr:
Will produce docker images for node 16, 18 and 20.

Related to #1265.

## About the changes

Closes #

### DIFF
--- a/.github/workflows/docker_publish_tags.yaml
+++ b/.github/workflows/docker_publish_tags.yaml
@@ -8,6 +8,9 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        version: [16-alpine, 18-alpine, 20-alpine]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


chriswk (changes requested):
This won't work. It will have a race condition on which part of the matrix finishes last for which container actually gets pushed. If you're going to do multiple platforms, you'll need to extend our tag format. 

This will build the same container, pushing to the same tag for all three containers.

Unfortunately, build-args is not a differentiator.

chriswk:
Adding a <nodeversion> suffix or prefix to our tags would work. Why do we need to do this though? I agree with building for multiple node versions and confirming that they work, but for docker I'm not a 100% sure why we want to publish multiple version?

ivarconr:
> I agree with building for multiple node versions and confirming that they work, but for docker I'm not a 100% sure why we want to publish multiple version?

Its a fair question. The main motivation is that changing the default node version would (in theory?) be a breaking change, while we should start to offer versions for people who wants to start using node-18 today.

chriswk (commented):
Is 20-alpine out?

PR #3363, opened 2023-03-21
Feat: plausible tracking stickiness

andreas-unleash:
Adds plausible tracking when default project stickiness is changed

## About the changes

Closes #


frontend/src/component/project/Project/CreateProject/CreateProject.tsx
```diff
@@ -10,6 +12,7 @@ import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 import useToast from 'hooks/useToast';
 import { formatUnknownError } from 'utils/formatUnknownError';
 import { GO_BACK } from 'constants/navigate';
+import { usePlausibleTracker } from '../../../../hooks/usePlausibleTracker';
```
kwasniew, line 15: dots

frontend/src/component/project/Project/EditProject/EditProject.tsx
```diff
@@ -15,6 +17,7 @@ import AccessContext from 'contexts/AccessContext';
 import { Alert } from '@mui/material';
 import { GO_BACK } from 'constants/navigate';
 import { useDefaultProjectSettings } from 'hooks/useDefaultProjectSettings';
+import { usePlausibleTracker } from '../../../../hooks/usePlausibleTracker';
```
kwasniew, line 20: dots

kwasniew (approved):
Originally when I was thinking how I'd implement it I thought about useEffect tracking an attempt to change the select box to something other than default. But now that I see how you added this inside the submit handler also makes sense.

andreas-unleash:
Initially I was planning to use it within a useEffect, but that would give false results if someone is changing the select over and over. Having it in the submit handler, ensures we only track what we need

PR #3366, opened 2023-03-21
Sticky batch actions bar

Tymek:
Animation


frontend/src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -21,12 +21,15 @@ interface IMainLayoutProps {
...
     justifyContent: 'space-between',
+    display: 'flex',
+    flexDirection: 'column',
+    flexGrow: 1,
+    position: 'relative',
 }));
 
 const MainLayoutContentWrapper = styled('main')(({ theme }) => ({
     margin: theme.spacing(0, 'auto'),
-    overflow: 'auto', // prevent margin collapsing
```
Tymek, line 28: I'm worried that this is important, but without removing it `sticky` isn't working

sjaanus (approved):
Good job!

PR #3368, opened 2023-03-22
fix: properly escaping app names

gastonfournier:
## About the changes
Network overview was not escaping square brackets from app names and that caused mermaid graph to break. We've also identified problems with quotes and some characters not showing properly when using utf-8 (but still no solution for this). We're going to be testing the utf-8 issue in our sandbox instances.


nunogois (approved):
LGTM 👍

PR #3372, opened 2023-03-23
fix: use projectId from param

sjaanus:
Previously every time there was permission **DELETE_FEATURE** or **UPDATE_FEATURE** needed, it expected to have `featureName` in path parameters.

Now that we introduced batch operations, for example batch deletion of features, the `featureName` will not be in parameters, but instead the `projectId` is provided as path parameter.

This PR makes following change:  now `projectId` from path parameters will be used instead of making a database call to find the `projectId` from feature.


andreas-unleash (approved):
Much better! LGTM

PR #3379, opened 2023-03-23
fix: Make api token environment selector accept environments

andreas-unleash:
Makes the api token environment selector accept options to allow different environment to be passed in according to the use case ie use all envs when creating api tokens and use project envs when creating project scoped tokens
## About the changes

Closes [# 1-811](https://linear.app/unleash/issue/1-811/change-environment-dropdown-to-show-all-environments-enabled-for-the)

### DIFF
--- a/frontend/src/component/project/Project/ProjectSettings/ProjectApiAccess/CreateProjectApiTokenForm.tsx
+++ b/frontend/src/component/project/Project/ProjectSettings/ProjectApiAccess/CreateProjectApiTokenForm.tsx
@@ -21,6 +21,7 @@ import { TokenTypeSelector } from 'component/admin/apiToken/ApiTokenForm/TokenTy
 import { ConfirmToken } from 'component/admin/apiToken/ConfirmToken/ConfirmToken';
 import { useProjectApiTokens } from 'hooks/api/getters/useProjectApiTokens/useProjectApiTokens';
 import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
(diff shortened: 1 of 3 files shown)


frontend/src/component/project/Project/ProjectSettings/ProjectApiAccess/CreateProjectApiTokenForm.tsx
```diff
@@ -49,6 +50,7 @@ export const CreateProjectApiTokenForm = () => {
         useProjectApiTokensApi();
     const { refetch: refetchProjectTokens } = useProjectApiTokens(project);
     const { trackEvent } = usePlausibleTracker();
+    const { environments } = useProjectEnvironments(project);
```
ivarconr, line 53: Will this only be enabled environments for the project?

andreas-unleash: Yes, though the intention as I understand would be to use `projectVisible` instead of `enabled` so that what we have in Environments project settings
Screenshot 2023-03-23 at 21 12 42

is available in the dropdown

Screenshot 2023-03-23 at 21 21 30

sebastian-bury: yes exactly! thank you @andreas-unleash

FredrikOseberg (approved):

PR #3380, opened 2023-03-23
refactor: read model for change request access checking

kwasniew:
## About the changes

Introducing a new **read model**  pattern to have more decoupling between modules and answer cross module questions.

What we have before this PR?
Feature toggle service behavior depends on change requests.
Change request status and ability to bypass change requests was incorrectly added to the access service and feature toggle service respectively.

What happens after this PR?
Read model answers a question if change requests can be bypassed.
Feature toggle service depends on the read model.

Why is this pattern important?
In the past we shared information between modules through services. 
But services introduce too much coupling. Why do I bring the whole service as a dependency with all its commands and transitive deps just to answer one query. 
Read model is a solution for those one-off queries across modules. 
It has much more tiny API surface area and doesn't require so many transitive deps.

Can you visualize this?

Screenshot 2023-03-23 at 17 08 54

How is read model related to a store?

Store is for storing and reading write models. Read model is for answering questions from other modules or returning view models to the UI (view model = everything I need in the UI screen)


src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public async canBypassChangeRequest(
+        project: string,
+        environment: string,
+        user?: User,
+    ): Promise<boolean> {
+        const [canSkipChangeRequest, changeRequestEnabled] = await Promise.all([
+            user
+                ? this.accessService.hasPermission(
```
kwasniew, line 26: this should be another read model because to check permissions I don't need all commands on the access service. But for now let's leave this as another refactor

src/lib/features/change-request-access-service/createChangeRequestAccessReadModel.ts
```diff
@@ -0,0 +1,19 @@
...
+): IChangeRequestAccessReadModel => {
+    const accessService = createAccessService(db, config);
+
+    return new ChangeRequestAccessReadModel(db, accessService);
+};
+
+export const createFakeChangeRequestAccessService =
+    (): IChangeRequestAccessReadModel => {
+        return new FakeChangeRequestAccessReadModel(true);
+    };
```
gastonfournier, line 19: I noticed this pattern of real object + fake object available in the same file. I assume fake is usually for testing, would it make sense to have it in a `.test.ts` file? As usual, this could be just my java-centric mind, but also having this in "production accessible code" makes me fear that by accident someone forgets to remove the fake implementation before going to production... I have some ideas, but first I'd like to understand the context of this design decision

src/lib/features/change-request-access-service/fake-change-request-access-read-model.ts
```diff
@@ -0,0 +1,15 @@
+import { IChangeRequestAccessReadModel } from './change-request-access-read-model';
+
+export class FakeChangeRequestAccessReadModel
```
gastonfournier, line 3: Do we need the auxiliary function `createFakeChangeRequestAccessService` if we can just call `new FakeChangeRequestAccessReadModel(true)`? The main risk I see is having multiple ways of creating the same object in different places. Normally, I like to keep the creation of objects close to the object definition itself, this could mean:
1. Moving that helper method to this file and making the constructor private
2. Not having the constructor accessible outside change-request-access-service (either private or protected IIRC)

src/lib/services/index.ts
```diff
@@ -149,11 +153,15 @@ export const createServices = (
     const healthService = new HealthService(stores, config);
     const userFeedbackService = new UserFeedbackService(stores, config);
     const segmentService = new SegmentService(stores, config); // TODO coupled with enterprise feature
+    const changeRequestAccessReadModel = db
+        ? createChangeRequestAccessReadModel(db, config)
+        : createFakeChangeRequestAccessService();
```
gastonfournier, line 158: This explains why the 2 constructor methods, but still I'm not convinced we should have this logic in this file. Despite this, I see we're already doing that with `exportImportService`. At first sight, this seems to be a blurry boundary between test code and production code and even if it's now limited to this file, this may leak in the future and cause issues. In any case, I think this is a different topic and we can talk about it separately from this PR :+1:

gastonfournier (commented):
I like when a refactor comes in so cleanly. I have a couple of questions but no blocker. Happy to :+1:, but because this is a refactor I think we can go over the questions first :)

src/lib/features/change-request-access-service/fake-change-request-access-read-model.ts
kwasniew: I added it for consistency reason but since this graph of objects is trivial for now we can skip it. If we wanted to inject fake access service (for now) then having a function for building a dependency graph is useful. Think of each of those methods as mini DI container for real/fake set of objects with type checking and without magic.

src/lib/services/index.ts
kwasniew: This logic doesn't belong here. There's a comment I made a few lines below. It's a transitional pattern before we migrate to fully feature driven architecture. After the migration we won't need this shared services file anymore. It was only useful in the technology driven architecture. And at the feature dependency graph level there will always be separate graphs for SQL based deps and in-memory based deps.

kwasniew:
@gastonfournier Keep the questions coming. Everything you're asking is things I should've put in the PR description and now you're showing me the gaps :)

src/lib/features/change-request-access-service/createChangeRequestAccessReadModel.ts
gastonfournier: I'm curious about the use cases where you'd use this. If this is a new feature behind a flag, you can easily write the fake code in the real implementation which later can evolve to be the real implementation. That way you only deal with "the implementation" and not a fake vs a real one. 

Other scenarios I think can easily be dealt with Feature Toggles which would give you control at runtime rather than at boot time and you can still have that in-memory behavior if you want.

The risk I see in having these fakes mixed with production code is that you may end up running something that differs from what customers run in production, and this is one thing I don't like to sacrifice.

src/lib/features/change-request-access-service/fake-change-request-access-read-model.ts
gastonfournier: I usually like to think about constructors as functions creating an object. Factory functions might be useful when the build is complex or if you get rid of classes and just use functions, but I think we didn't make a decision on that area

src/lib/services/index.ts
gastonfournier: It's fine if it's a transitional pattern, but what puzzles me is the fact that we can use this code with or without a DB. It's like implying: when there's no db we're in test, which means we're mixing production code with test code

kwasniew: With the technology driven architecture we were injecting either real stores or fake stores to the big services factory. So you could also pass stores with DB an without DB. But now with the migration to feature driven architecture this is the best I found so far before we get rid of a big services factory.

src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
```diff
@@ -0,0 +1,52 @@
...
+                      user,
+                      SKIP_CHANGE_REQUEST,
+                      project,
+                      environment,
+                  )
+                : Promise.resolve(false),
+            this.accessService.isChangeRequestsEnabled(project, environment),
```
nunogois, line 33: From your comment above maybe you meant this instead? Otherwise I don't see the point of the `async isChangeRequestsEnabled` method in the class.

```suggestion
            this.isChangeRequestsEnabled(project, environment),
```

src/lib/features/change-request-access-service/createChangeRequestAccessReadModel.ts
kwasniew: What about a common closure principle then? Things that change together should be kept together. When I change a contract for real and fake impl I want to do it inside one directory. I don't want to do [shotgun surgery](https://refactoring.guru/smells/shotgun-surgery). I'd like all my feature changes to be done in a feature directory. If I delete a feature I want to delete one directory. Etc.

kwasniew: Actually if you look at the tests they are also in the same directory. This is where Java and JS/TS nodes diverged. I see JS/TS to follow common closure principle to a greater extent. So when we say feature-oriented architecture then everything about this feature is in a feature directory.

src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
kwasniew: Brilliant finding @nunogois Thanks!

src/lib/features/change-request-access-service/createChangeRequestAccessReadModel.ts
gastonfournier: Yes, but one thing is having tests in the same directory, but not being able to refer to a test class/object from production code because somehow `.test.ts` files are not accessible (I don't know much about the implementation details of the typescript compiler here). But having test code being accessible from production code seems a bit risky because the boundary becomes blurry.

src/lib/services/index.ts
gastonfournier: I'm ok with this as a temporal solution, but I wouldn't like this pattern to spread. The main difference with https://github.com/Unleash/unleash/blob/main/src/test/fixtures/store.ts is that it's in a `src/test` folder which should not be accessible from production code (but from a quick test I did that doesn't seem to be the case... :disappointed:)

kwasniew: So if it's visible anyway it's better to have it in one place at least :)

gastonfournier: Maybe the difference is that now in test code you need to import the appropriate code: https://github.com/Unleash/unleash/blob/9cdd870365035077359fc98afbd5659777784aa7/src/lib/metrics.test.ts#L12 which then is used to create the objects in test code: https://github.com/Unleash/unleash/blob/9cdd870365035077359fc98afbd5659777784aa7/src/lib/metrics.test.ts#L28-L30

In the new solution, the dependency is magically changed/injected depending on the db. I'd rather have the db as non-optional and keep doing what we were doing, but I do not consider that a blocker for this PR

gastonfournier (approved):
Please, keep some of the comments open, but overall I think this is good as is and a step in the right direction. The parts that are more opinionated I think are reversible decisions that can easily be refactored later, so I'm ok with this PR as it's adding value in terms of separation of concerns :+1:

PR #3381, opened 2023-03-24
fix: header-responsiveness

NicolaeUnleash:
- when introducing dark mode icon in the header, on tablet the menu items don't have enough space anymore, and this fix it

- fixing also the color for dark mode of the keyboard nav button "skip to content"

### DIFF
--- a/frontend/src/component/menu/Header/Header.tsx
+++ b/frontend/src/component/menu/Header/Header.tsx
@@ -54,9 +54,7 @@ const StyledContainer = styled(Container)(({ theme }) => ({
     display: 'flex',
     alignItems: 'center',
     maxWidth: 1280,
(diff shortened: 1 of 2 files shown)


frontend/src/component/menu/Header/Header.tsx
```diff
@@ -54,9 +54,7 @@ const StyledContainer = styled(Container)(({ theme }) => ({
     display: 'flex',
     alignItems: 'center',
     maxWidth: 1280,
-    [theme.breakpoints.down('md')]: {
-        padding: '0',
-    },
+    '&&&': { padding: 0 },
```
NicolaeUnleash, line 57: i used this method because there is some styling that is applied to these component and overwrite a normal padding

nunogois (approved):
LGTM

PR #3382, opened 2023-03-24
fix: count events instead of loading them in memory

FredrikOseberg:
Refactor project events to use count instead of loading the events in memory

### DIFF
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -35,6 +35,7 @@ import {
     ProjectUserRemovedEvent,
     ProjectUserUpdateRoleEvent,
     RoleName,
+    IFlagResolver,
 } from '../types';
 import {
     IProjectQuery,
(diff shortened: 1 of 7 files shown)


kwasniew (approved):
pair programmed. eventually we'll move parts of the store to a dedicated read model

src/lib/services/project-service.ts
```diff
@@ -664,20 +668,24 @@ export default class ProjectService {
...
-        await Promise.all(
-            statusUpdates.map((statusUpdate) => {
-                return this.projectStatsStore.updateProjectStats(
-                    statusUpdate.projectId,
-                    statusUpdate.updates,
-                );
-            }),
-        );
+            await Promise.all(
+                statusUpdates.map((statusUpdate) => {
```
ivarconr, line 679: How many queries can we potentially generate here?

PR #3385, opened 2023-03-24
task: enclose ETag in double quotes.

chriswk:
According to the
[RFC](https://www.rfc-editor.org/rfc/rfc7232#section-2.3) Etags should be enclosed in double quotes. This PR makes sure we do actually wrap it in double quotes


ivarconr:
Thanks 🙏
This was unexpected, should have read the RFC more carefully

PR #3389, opened 2023-03-27
fix error boundary

kwasniew:
## About the changes

Move plausible available for the error boundary.

Previously:
* plausible provider was much deeper than the error boundary so the error handler didn't get access to the tracker

After:
* error boundary moved below the plausible provider 
* another error boundary added for the plausible provider itself

This is just a few LOC but the formatting kicked in

### DIFF
--- a/frontend/src/component/App.tsx
+++ b/frontend/src/component/App.tsx
@@ -48,61 +48,67 @@ export const App = () => {
 
     return (
         <ErrorBoundary FallbackComponent={Error}>
-            <SWRProvider>
-                <Suspense fallback={<Loader />}>
-                    <PlausibleProvider>
-                        <ConditionallyRender
-                            condition={!hasFetchedAuth}
-                            show={<Loader />}
-                            elseShow={
-                                <>
-                                    <ConditionallyRender
-                                        condition={Boolean(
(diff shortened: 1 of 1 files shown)


frontend/src/component/App.tsx
```diff
@@ -48,61 +48,67 @@ export const App = () => {
...
-                                            uiConfig?.maintenanceMode
-                                        )}
-                                        show={<MaintenanceBanner />}
-                                    />
-                                    <StyledContainer>
-                                        <ToastRenderer />
-                                        <Routes>
-                                            {availableRoutes.map(route => (
+            <PlausibleProvider>
+                <ErrorBoundary FallbackComponent={Error}>
```
kwasniew, line 52: this will catch all the unhandled errors except from the plausible errors

frontend/src/component/App.tsx
```diff
@@ -48,61 +48,67 @@ export const App = () => {
 
     return (
         <ErrorBoundary FallbackComponent={Error}>
```
kwasniew, line 50: this is to handle plausible errors from the plausible provider

Tymek (approved):

nunogois (approved):
LGTM

PR #3390, opened 2023-03-27
fix: remove notification feature flag from oss

FredrikOseberg:
Removes the notifications feature flag


kwasniew (changes requested):
I think there's still a usage of the flag in Header.tsx

FredrikOseberg:
> I think there's still a usage of the flag in Header.tsx

True. We should improve how we keep track of these, it wasn't easy to find with pure searching.

PR #3391, opened 2023-03-27
feat: stop change project on change request enabled

kwasniew:
## About the changes
Backend check to verify we cannot switch toggle project to a project that has change requests enabled in any environment. We can add extra UI enhancement on top of it in a separate PR.

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -1350,6 +1350,15 @@ class FeatureToggleService {
         newProject: string,
         createdBy: string,
     ): Promise<void> {
+        const changeRequestEnabled =
+            await this.changeRequestAccessReadModel.isChangeRequestsEnabledForProject(
+                newProject,
+            );
+        if (changeRequestEnabled) {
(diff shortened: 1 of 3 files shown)


src/lib/services/feature-toggle-service.ts
```diff
@@ -1350,6 +1350,15 @@ class FeatureToggleService {
         newProject: string,
         createdBy: string,
     ): Promise<void> {
+        const changeRequestEnabled =
+            await this.changeRequestAccessReadModel.isChangeRequestsEnabledForProject(
+                newProject,
+            );
+        if (changeRequestEnabled) {
```
ivarconr, line 1357: what happens if the it is an API user with "skip change request" permissions?

kwasniew: Currently it's still blocked. SKIP_CHANGE_REQUEST was mainly created for emergency situations/monitoring scripts to quickly turn off the kill switch without going through the change request process. The question is should we do the same for cloning a toggle/moving a toggle between projects.

PR #3402, opened 2023-03-28
fix: input used to read file should be sanitized

gastonfournier:
## About the changes
Following the recommendation to fix https://github.com/Unleash/unleash/security/code-scanning/2 and https://github.com/Unleash/unleash/security/code-scanning/3 

The endpoint seems to be used for developing purposes only (to preview an email under development) but it's available in every Unleash installation and can potentially be exploited.


chriswk (approved):
LGTM. Nice catch.

PR #3404, opened 2023-03-28
feat: disable apply button on loading

kwasniew:
## About the changes
Apply change request button should be disabled when the changes are being applied. Otherwise we allow for double apply submission.


FredrikOseberg (approved):
LGTM

PR #3405, opened 2023-03-28
docs: Project api token

andreas-unleash:
## About the changes

Closes #


website/docs/reference/api/unleash/unleash-api.info.mdx
thomasheartman, line 1: What's this file? This doesn't look like it should be added here? Did you do it manually or is it an automatic thing?

thomasheartman (commented):
This is a super good starting point! 👏🏼 👏🏼 👏🏼  But I think we should add a small subsection to the API tokens and client keys doc explaining what they are. And that's probably also where we should put the thing about visibility in the admonition. I'll leave it up to you, but I think it might be a subsection under the ["parts of an API token"](https://docs.getunleash.io/reference/api-tokens-and-client-keys#token-data) section. Maybe 🤔

website/docs/reference/api/unleash/unleash-api.info.mdx
andreas-unleash: Automatic. I will remove it

website/docs/how-to/how-to-create-project-api-tokens.mdx
```diff
@@ -0,0 +1,35 @@
+---
+title: How to create Project API Tokens
+---
+
+:::info Permissions
+
+Creating Project API tokens requires you to have the `CREATE_PROJECT_API_TOKEN` permission.
+
+All project users can see tokens with `CLIENT` and `FRONTEND` level access.
```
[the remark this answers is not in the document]
andreas-unleash: Yes, thank you for clarifying it

sjaanus (approved):

website/docs/reference/api-tokens-and-client-keys.mdx
```diff
@@ -55,6 +56,18 @@ Do **not** use admin tokens for:
...
 Support for scoped admin tokens with more fine-grained permissions is currently in the planning stage.
 
+
+### Project level access tokens
+
+All project members can see any `CLIENT` (server-side client token) and `FRONTEND` (client-side client token) respectively.
+
+Use personal access tokens to:
+
+- Provide more fine-grained permissions for automation than an admin token provides
```
thomasheartman, line 66: Looks like you forgot to remove this line when copying the PAT section 💁🏼 

```suggestion
```

website/docs/reference/api-tokens-and-client-keys.mdx
```diff
@@ -21,6 +21,7 @@ This section describes what API tokens are. For information on how to create the
 Use API tokens to connect to the Unleash server API. API tokens come in four distinct types:
 
 -   [Admin tokens](#admin-tokens)
+-   [Project level access tokens](#project-access-tokens)
```
thomasheartman, line 24: It's not a different type of API token, so I don't think it should be listed here.

```suggestion
```

website/docs/reference/api-tokens-and-client-keys.mdx
```diff
@@ -55,6 +56,18 @@ Do **not** use admin tokens for:
 
 Support for scoped admin tokens with more fine-grained permissions is currently in the planning stage.
 
+
+### Project level access tokens
+
+All project members can see any `CLIENT` (server-side client token) and `FRONTEND` (client-side client token) respectively.
```
thomasheartman, line 62: Thanks! Looking at this, I think we could reword it to instead be about visibility. But I need you to help me out with the details here: how visible are tokens usually? I've got a suggestion below, but it needs to be modified based on what is the reality. I'd be happy to help rewrite this again, but I don't quite know how all the permissions around API tokens work these days.

Oh, and I also think this should be moved to before the admin tokens section.

```suggestion
### API token visibility

:::info project-level visibility

Project-level visibility and access to API tokens was introduced in Unleash 4.22.

:::


By default, only admin users can create API tokens, and only admins can see their values.

However, any [client](#client-tokens client tokens) and [front-end tokens](#front-end-tokens) that are applicable to a project, will also be visible to any members of that project that have the `READ_PROJECT_API_TOKEN` permission (all project members by default). 

Similarly, any project members with the `CREATE_PROJECT_API_TOKEN` permission can also create client and front-end tokens for that specific project ([how to create project API tokens](../how-to/how-to-create-project-api-tokens.mdx)).
```

website/docs/reference/api-tokens-and-client-keys.mdx
```diff
@@ -55,6 +56,18 @@ Do **not** use admin tokens for:
...
+
+Use personal access tokens to:
+
+- Provide more fine-grained permissions for automation than an admin token provides
+- Give access to all project members (or anyone with `READ_PROJECT_API_TOKEN` permission) to read the token
+- Give access to the project owner (or anyone with `CREATE_PROJECT_API_TOKEN` permission) to create a token,
```
thomasheartman, line 68: If you agree with my suggestion regarding visibility, then I think we can also delete this and the rest of the other content from this section. 

But before you do: isn't the `read_project_api_token` permission a project-level permission? So don't you have to be BOTH a project member AND have that permission? And the same for creating.

thomasheartman (commented):
Nice! But there's still a few things I'd like to address and ask you about / discuss. I know the PR was just merged, so we can follow up on these in a subsequent PR.

website/docs/reference/api-tokens-and-client-keys.mdx
andreas-unleash: Yes, we assign the `read_project_api_token` permission to Project Members by default and the same for project owner and `create_project_api_token`

website/docs/reference/api-tokens-and-client-keys.mdx
andreas-unleash: agreed and well done

PR #3408, opened 2023-03-28
Fix: make stickiness accept any string

andreas-unleash:
Changes the schema and api to accept any string for defaultStickiness
## About the changes

Closes #


kwasniew (approved):
One minor comment about omit

PR #3421, opened 2023-03-29
OpenAPI: addon operations

thomasheartman:
This PR updates the OpenAPI schemas for all the operations tagged with "addons". In doing so, I also uncovered a few bugs and inconsistencies. These have also been fixed.

## Changes

I've added inline comments to the changed files to call out anything that I think is worth clarifying specifically. As an overall description, this PR does the following:

Splits `addon-schema` into `addon-schema` and `addon-create-update-schema`. The former is used when describing addons that exist within Unleash and contain IDs and `created_at` timestamps. The latter is used when creating or updating addons.

Adds examples and descriptions to all relevant schemas (and their dependencies).

Updates addons operations descriptions and response codes (including the recently introduced 413 and 415).

Fixes a bug where the server would crash if it didn't recognize the addon provider (test added).

Fixes a bug where updating an addon wouldn't return anything, even if the API said that it would. (test added)

Resolves some inconsistencies in handling of addon description. (tests added)

### Addon descriptions

when creating addons, descriptions are optional. The original `addonSchema` said they could be `null | string | undefined`. This caused some inconsistencies in return values. Sometimes they were returned, other times not. I've made it so that `descriptions` are now always returned from the API. If it's not defined or if it's set to `null`, the API will return `description: null`. 

### `IAddonDto`

`IAddonDto`, the type we used internally to model the incoming addons (for create and update) says that `description` is required. This hasn't been true at least since we introduced OpenAPI schemas. As such, the update and insert methods that the service uses were incompatible with the **actual** data that we require. 

I've changed the type to reflect reality for now. Assuming the tests pass, this **should** all be good, but I'd like the reviewer(s) to give this a think too.


sjaanus:
Also should enable `strictSchemaValidation` on e2e test.

src/lib/db/addon-store.ts
```diff
@@ -16,6 +16,7 @@ const COLUMNS = [
     'events',
     'projects',
     'environments',
+    'created_at',
```
thomasheartman, line 19: This was missing from our return value. We claimed that we returned it, but we never did. Seems like an oversight.

src/lib/db/addon-store.ts
```diff
@@ -71,18 +72,19 @@ export default class AddonStore implements IAddonStore {
         stopTimer();
         // eslint-disable-next-line @typescript-eslint/naming-convention
         const { id, created_at } = rows[0];
-        return { id, createdAt: created_at, ...addon };
+        return this.rowToAddon({ id, createdAt: created_at, ...addon });
```
thomasheartman, line 75: Without mapping this, we could return a "created" payload with a missing description property. Using the `rowToAddon` method, the `description` property will be set to `null` if it is missing. This seems like a bug. Getting the same addon via a GET (or PUT) request would return the `description` property (as `null` if it was missing)

src/lib/openapi/spec/addon-create-update-schema.ts
```diff
@@ -0,0 +1,73 @@
...
+        provider: {
+            type: 'string',
+
+            description: `The addon provider, such as "webhook" or "slack". The list of all providers supported can be found in the [addons reference documentation](https://docs.getunleash.io/reference/addons). This string is **case sensitive**. Supported options are:
+- \`datadog\` for [Datadog](https://docs.getunleash.io/reference/addons/datadog)
+- \`slack\` for [Slack](https://docs.getunleash.io/reference/addons/slack)
+- \`teams\` for [Microsoft Teams](https://docs.getunleash.io/reference/addons/teams)
+- \`webhook\` for [webhooks](https://docs.getunleash.io/reference/addons/webhook)
+
+The provider you choose for your addon dictates what properties the \`parameters\` object needs. Refer to the documentation for each provider for more information.
```
thomasheartman, line 17: We have very scant documentation on this, so until those are improved, I think putting this in here would be useful.

src/lib/openapi/spec/addon-create-update-schema.ts
thomasheartman, line 1: Split out from the normal addon schema. This schema is used when creating or updating addons. It has some different requirements and some of the descriptions are tailored to the use case.

src/lib/openapi/spec/addon-parameter-schema.ts
```diff
@@ -4,27 +4,49 @@ export const addonParameterSchema = {
     $id: '#/components/schemas/addonParameterSchema',
     type: 'object',
     required: ['name', 'displayName', 'type', 'required', 'sensitive'],
+    description: 'Parameter definition for an addon. .... TODO',
```
thomasheartman, line 7: This needs to be fleshed out a bit. Should be fairly trivial.

src/lib/openapi/spec/addon-parameter-schema.ts
```diff
@@ -4,27 +4,49 @@ export const addonParameterSchema = {
...
     properties: {
         name: {
             type: 'string',
+            example: 'emojiIcon',
+            description:
+                'The name of the parameter as it will be used in code.',
```
thomasheartman, line 13: Where is this actually used? What does this mean?

src/lib/openapi/spec/addon-parameter-schema.ts
```diff
@@ -4,27 +4,49 @@ export const addonParameterSchema = {
...
         displayName: {
             type: 'string',
+            example: 'Emoji Icon',
+            description:
+                'The name of the parameter as it will be shown to the end user.',
         },
         type: {
             type: 'string',
+            description:
+                "The type of the parameter. TODO: what are these? Are they defined anywhere? What do they do? I found `text`, `url`, and `textfield`, but they don't seem to do much?",
```
thomasheartman, line 24: What does the type mean? What does it dictate?

src/lib/openapi/spec/addon-schema.test.ts
```diff
@@ -5,6 +5,8 @@ test('addonSchema', () => {
     const data: AddonSchema = {
         provider: 'some-provider',
         enabled: true,
+        description: null,
+        id: 5,
```
thomasheartman, line 9: These fields are now mandatory when when returning the schema. In effect they were there before, but the schema didn't mark them as mandatory because it was also used for creating addons.

src/lib/openapi/spec/addon-type-schema.ts
```diff
@@ -5,37 +5,121 @@ import { tagTypeSchema } from './tag-type-schema';
 export const addonTypeSchema = {
     $id: '#/components/schemas/addonTypeSchema',
     type: 'object',
+    additionalProperties: false,
     required: ['name', 'displayName', 'documentationUrl', 'description'],
+    description:
+        'An addon provider definiton. Addon providers define ... TODO. Think of them as type definitions.',
```
thomasheartman, line 11: I'm not entirely sure how best to describe this yet. Needs more thought.

src/lib/openapi/spec/addon-type-schema.ts
```diff
@@ -5,37 +5,121 @@ import { tagTypeSchema } from './tag-type-schema';
...
         },
         description: {
             type: 'string',
+            description: 'A description of the addon provider.',
+            example: 'Allows Unleash to post updates to Slack.',
         },
         tagTypes: {
             type: 'array',
+            description:
+                "A list of tags that the addon provider uses. TODO: I don't know what this is.",
```
thomasheartman, line 39: What actually is this?

src/lib/openapi/spec/addons-schema.ts
```diff
@@ -8,15 +8,256 @@ export const addonsSchema = {
...
+                    tagTypes: [
+                        {
+                            name: 'datadog',
+                            description:
+                                'All Datadog tags added to a specific feature are sent to datadog event stream.',
+                            icon: 'D',
+                        },
+                    ],
+                },
+            ],
```
thomasheartman, line 260: This is what you actually get when you call the endpoint. All this stuff is not documented anywhere else.

src/lib/openapi/util/standard-responses.ts
```diff
@@ -25,12 +25,17 @@ const conflictResponse = {
...
+    description: `The operation does not support request payloads of the provided type. Please ensure that you're using one of the listed payload types and that you have specified the right content type in the "content-type" header.`,
+} as const;
+
 const standardResponses = {
     400: badRequestResponse,
     401: unauthorizedResponse,
     403: forbiddenResponse,
     404: notFoundResponse,
     409: conflictResponse,
+    415: unsupportedMediaTypeResponse,
```
thomasheartman, line 38: This is probably also covered in a different PR.

src/lib/routes/admin-api/addon.ts
```diff
@@ -63,10 +71,16 @@ class AddonController extends Controller {
...
+                    description:
+                        'Create an addon instance. The addon must use one of the providers available on this Unleash instance.',
                     tags: ['Addons'],
                     operationId: 'createAddon',
-                    requestBody: createRequestSchema('addonSchema'),
-                    responses: { 200: createResponseSchema('addonSchema') },
+                    requestBody: createRequestSchema('addonCreateUpdateSchema'),
+                    responses: {
+                        200: createResponseSchema('addonSchema'),
+                        ...getStandardResponses(400, 401, 403, 415),
```
thomasheartman, line 82: Should also have 413, I suppose

src/lib/routes/admin-api/addon.ts
```diff
@@ -92,10 +112,17 @@ class AddonController extends Controller {
...
+
+Note: passing \`null\` as a value for the description property will set it to an empty string.`,
                     tags: ['Addons'],
                     operationId: 'updateAddon',
-                    requestBody: createRequestSchema('addonSchema'),
-                    responses: { 200: createResponseSchema('addonSchema') },
+                    requestBody: createRequestSchema('addonCreateUpdateSchema'),
+                    responses: {
+                        200: createResponseSchema('addonSchema'),
+                        ...getStandardResponses(400, 401, 403, 415),
```
thomasheartman, line 124: 413

src/lib/routes/admin-api/addon.ts
```diff
@@ -141,11 +174,12 @@ class AddonController extends Controller {
     }
 
     async updateAddon(
-        req: IAuthRequest<{ id: number }, any, any, any>,
+        req: IAuthRequest<{ id: number }, any, AddonCreateUpdateSchema, any>,
         res: Response<AddonSchema>,
     ): Promise<void> {
         const { id } = req.params;
         const createdBy = extractUsername(req);
+
```
thomasheartman, line 182:
```suggestion
```

src/lib/services/addon-service.ts
```diff
@@ -12,13 +12,18 @@ import { IAddon, IAddonDto, IAddonStore } from '../types/stores/addon-store';
...
 
 const SUPPORTED_EVENTS = Object.keys(events).map((k) => events[k]);
 
 const MASKED_VALUE = '*****';
 
 const WILDCARD_OPTION = '*';
 
+// The *actual* shape of the data we receive. The description field can't be
+// enforced now without it being breaking change.
+type ActualAddonDto = Omit<IAddonDto, 'description'> & { description?: string };
```
thomasheartman, line 25: The type `IAddonDto` expects `description` to be required, but it is *not*. There's probably better ways to handle this, but changing the type itself had some far-reaching issues, so this is a patch for now. Suggestions are welcome.

src/lib/services/addon-service.ts
```diff
@@ -186,10 +191,14 @@ export default class AddonService {
...
+    async createAddon(data: ActualAddonDto, userName: string): Promise<IAddon> {
         const addonConfig = await addonSchema.validateAsync(data);
-        await this.validateKnownProvider(addonConfig);
-        await this.validateRequiredParameters(addonConfig);
+        try {
+            await this.validateKnownProvider(addonConfig);
+            await this.validateRequiredParameters(addonConfig);
+        } catch (e) {
+            throw new BadDataError(e);
+        }
```
thomasheartman, line 201: This gets around some API errors where the response is either 500 or not clear (I forget which one)

src/test/e2e/api/admin/addon.e2e.test.ts
```diff
@@ -9,7 +9,13 @@ let db;
 
 beforeAll(async () => {
     db = await dbInit('addon_api_serial', getLogger);
-    app = await setupApp(db.stores);
+    app = await setupAppWithCustomConfig(db.stores, {
+        experimental: {
+            flags: {
+                strictSchemaValidation: true,
```
thomasheartman, line 15: Enable strict schema validation for e2e tests with addon schemas

src/test/e2e/api/admin/addon.e2e.test.ts
```diff
@@ -196,3 +202,98 @@ test('should not delete unknown addon configuration', async () => {
...
+test("should return 400 if it doesn't recognize the provider", async () => {
+    const payload = {
+        provider: 'htni',
+        enabled: true,
+        parameters: {},
+        events: [],
+    };
+
+    return app.request.post('/api/admin/addons').send(payload).expect(400);
+});
```
thomasheartman, line 215: This used to crash the API (500)

src/test/e2e/api/admin/addon.e2e.test.ts
```diff
@@ -196,3 +202,98 @@ test('should not delete unknown addon configuration', async () => {
...
+    };
+
+    return app.request
+        .put(`/api/admin/addons/${body.id}`)
+        .send(updatedConfig)
+        .expect((res) => {
+            expect(res.body).toMatchObject(updatedConfig);
+        });
+});
+
```
thomasheartman, line 241: This used to return nothing (even if the API promised it returned the addon).

src/test/e2e/api/admin/addon.e2e.test.ts
```diff
@@ -196,3 +202,98 @@ test('should not delete unknown addon configuration', async () => {
...
+                .send({
+                    ...addonWithoutDescription,
+                    description,
+                })
+                .expect((res) => {
+                    expect(res.body.description).toStrictEqual('');
+                });
+        },
+    );
+});
```
thomasheartman, line 299: encoded how we deal with addon descriptions today. The last test (that it gets set to an empty string) is because of how the OpenAPI coerces a null into an empty string. Unsure what do about that here.

src/lib/openapi/spec/addon-parameter-schema.ts
thomasheartman: html input types

src/lib/openapi/spec/addon-type-schema.ts
thomasheartman: It **is** supposed to be an Unleash tag type

thomasheartman: will extend your list of tags by adding custom tag types

src/lib/types/stores/addon-store.ts
thomasheartman, line 1: The changes in this file better reflect how we actually use addons (to the best of my understanding). Refer to the review description for more information.

src/lib/services/addon-service.test.ts
thomasheartman, line 1: The changes here are to allow the API to return the correct error code (400) in cases of bad data. If this is incorrectly configured (that the other errors here should be caught and reported automatically), then we can look at that. However, this was causing 500 errors (I believe) prior to this.

src/lib/openapi/spec/addon-parameter-schema.ts
```diff
@@ -3,28 +3,51 @@ import { FromSchema } from 'json-schema-to-ts';
...
+            example:
+                'The emoji_icon to use when posting messages to slack. Defaults to ":unleash:".',
+            description:
+                'A description of the parameter. This should explain to the end user what the parameter is used for.',
         },
         placeholder: {
             type: 'string',
+            example: ':unleash:',
+            description:
+                'The default value for this parameter. This value be used if no other value is provided.',
```
chriswk, line 39:
```suggestion
                'The default value for this parameter. This value is used if no other value is provided.',
```

src/lib/openapi/spec/addon-parameter-schema.ts
```diff
@@ -3,28 +3,51 @@ import { FromSchema } from 'json-schema-to-ts';
...
         required: {
             type: 'boolean',
+            example: false,
+            description:
+                'Whether this parameter is required or not. If a parameter is required, you must give it a value when you create the addon. If it is not required it can be left out. It may receive a default value in those cases.',
         },
         sensitive: {
             type: 'boolean',
+            example: false,
+            description: `Indicates whether this parameter is **sensitive** or not. Unleash will not return sensitive parameters to API request. It will instead use a number of asterisks to indicate that a value is set, e.g. "******". The number of asterisks does not correlate to the parameter\'s value.`,
```
chriswk, line 50:
```suggestion
            description: `Indicates whether this parameter is **sensitive** or not. Unleash will not return sensitive parameters to an API request. It will instead use a number of asterisks to indicate that a value is set, e.g. "******". The number of asterisks does not correlate to the parameter\'s value.`,
```

src/lib/openapi/spec/addon-schema.ts
```diff
@@ -3,43 +3,70 @@ import { FromSchema } from 'json-schema-to-ts';
...
+            description: 'The event types that trigger this specific addon.',
             items: {
                 type: 'string',
             },
+            example: ['feature-created', 'feature-updated'],
         },
         projects: {
             type: 'array',
+            description:
+                'The projects that this addon listens to events from. An empty list means listens to events from **all** projects.',
```
chriswk, line 59:
```suggestion
                'The projects that this addon listens to events from. An empty list means it listens to events from **all** projects.',
```

src/lib/openapi/spec/addon-create-update-schema.ts
chriswk: I like it. Splitting response and request types seems logical.

src/lib/openapi/spec/addon-create-update-schema.ts
chriswk: Registered providers might refer to the endpoint in the /addons endpoint, where we can see all known providers for the instance. I guess this has exposed the need for fleshing out the addon documentation.

src/lib/openapi/spec/addon-schema.test.ts
chriswk: Same goes here, I like that we're doing the work with separating request/response to avoid having optional fields, or having to allow extra properties in responses because we're reusing the schema as a request

src/lib/services/addon-service.test.ts
```diff
@@ -627,7 +628,7 @@ test('should reject addon config with missing required parameter when creating',
 
     await expect(async () =>
         addonService.createAddon(config, 'me@mail.com'),
-    ).rejects.toThrow(ValidationError);
+    ).rejects.toThrow(BadDataError);
```
chriswk, line 631: We are handling ValidationError by returning 400, it feels like that is more verbose and explicit.

src/lib/services/addon-service.test.ts
```diff
@@ -73,7 +74,7 @@ test('should not allow addon-config for unknown provider', async () => {
             },
             'test',
         );
-    }).rejects.toThrow(new TypeError('Unknown addon provider unknown'));
+    }).rejects.toThrow(BadDataError);
```
chriswk, line 77: Here, I'm wondering if we shouldn't have fixed the src/lib/routes/util.ts handleError match statement instead to return 400 on TypeErrors.

chriswk (approved):
Some concerns about using a more generic error, but love the fixes and added tests confirming that we're delivering what we promise.

src/lib/services/addon-service.test.ts
thomasheartman: Resolved!

PR #3417, opened 2023-03-29
feat: add API pagination feature flag

nunogois:
https://linear.app/unleash/issue/2-826/add-pagination-feature-flag

Adds `apiPagination` feature flag in order to start https://linear.app/unleash/project/[med][m][none]-api-filter-and-pagination-support-29ef4909dd2e/2

PR #3433, opened 2023-03-31
refactor: revert adding apiPagination feature flag

nunogois:
Reverts https://github.com/Unleash/unleash/pull/3417 - If we're not going forward with this project at this stage, we should clean up after ourselves and remove the unused flag that we added.


chriswk (approved):
This seems like the logical step since we decided to put this on hold for a while longer.

PR #3434, opened 2023-03-31
docs: point link to unleash/unleash instead of unleash-docker

thomasheartman:
## why

This link used to post to the now-outdated unleash-docker repo. They should instead go directly to the unleash repo now.


sebastian-bury (approved):
Looks good to me! :)

PR #3439, opened 2023-04-03
docs: adr for db changes

kwasniew:
## About the changes

ADR for making safe breaking DB changes

### DIFF
--- a/website/docs/contributing/ADRs/back-end/breaking-db-changes.md
+++ b/website/docs/contributing/ADRs/back-end/breaking-db-changes.md
@@ -1,0 +1,28 @@
+---
+title: "ADR: Braking DB changes"
+---
+
+## Background
+
+During the evolution of a feature different clients may use different version of code e.g. behind a feature flag.
+If the code relies on breaking DB changes it may lead to errors.
+
+The very same problem occurs when you apply a breaking migration just before the new version of the application starts. The code is still running 
+against the old schema as the migration takes a few seconds to apply.
+
+## Decision
+
+First please make sure to avoid breaking DB changes in the first place if possible.
+
+If breaking change is inevitable please use the "expand/contract" pattern. 
+
+In the "expand phase":
+* maintain old and new DB schema in parallel
+* maintain code that works with old and new DB schema
+* keep it for 2 minor releases to give all clients a chance to upgrade the code
+
+In the "contract phase":
+* remove the old schema when you know that no client is using the old version
+
+Action for a code reviewer:
+* when you spot a migration with `ALTER table DROP COLUMN` or `ALTER table RENAME TO` please raise a flag if the "expand phase" was missed


gardleopard (approved):
Looks good

sjaanus (commented):
Probably should mention that when deleting, changing entries of table, need to be sure that these entries are not needed in database in old version of Unleash.

website/docs/contributing/ADRs/back-end/breaking-db-changes.md
```diff
@@ -0,0 +1,29 @@
...
+
+First please make sure to avoid breaking DB changes in the first place if possible.
+
+If breaking change is inevitable please use the "expand/contract" pattern. 
+
+In the "expand phase":
+* maintain old and new DB schema in parallel
+* maintain code that works with old and new DB schema
+* keep it for 2 minor releases to give all clients a chance to upgrade the code
+* with a fallback of 2 version we can also downgrade in this range without running down migrations
```
gastonfournier, line 23: I think we need to discuss if 2 minor releases are required. I believe 2 deployments should be ok for cloud instances, no need of a minor release.

For self-hosted, it might be a different story and I'd use one minor version for the "expand phase" and another for the "contract phase", but it's also tricky because some customers might be lagging for a few minor versions and upgrading two at the same time might not be advisable. 

If we just focus on our cloud customers and consider that no matter what we do, self-hosted customers might run into this issue if we just increment minor versions, we could consider just using patch versions instead of minor versions. 

But if we really want to play nice with our self-hosted, the right thing to do, would be to push breaking changes until we have a major release... that of course, will require us to be more careful with DB changes and perhaps less experimental, which I wouldn't like that much.

My worry here is not to add additional waiting steps (such as 2 months for 2 minor versions) as default when in some cases we might just get along with patching

gastonfournier (commented):
I agree with the process, but I think we need to think a bit about how we introduce these changes. Even with minor version changes this could be a major change to some customers (although depending on their deployment strategy they might just experience a minor hiccup). The question is: should these changes be considered major at expense of higher maintenance costs and slowing development? Are we just tackling a small percentage of potential bugs but impacting the majority of our DB changes?

website/docs/contributing/ADRs/back-end/breaking-db-changes.md
kwasniew: This ADR is mostly for us hosting the infra. We can't cover every DB migration scenario out there, so to start small I focused on our real world needs. In practice we do one minor version up and down. E.g. 4.21 -> 4.22 -> 4.21. This is sth we did in the past. The two versions recommendation is to give use 1 version buffer.

gastonfournier: I understand, but with the current cadence of minor releases (once a month) that requires 2 months before applying the migration, it's a lot of time to keep that into our heads/backlogs. If we're just focusing on us hosting (which I think is the right thing to do, cause as you mentioned we can't cover all scenarios), then I'd like us to focus on a continuous deployment scenario, where minor versions become irrelevant, but it's more about waiting for a few deployments. WDYT?

kwasniew: Just to be clear, I'm not saying you can't make one migration after the other for non breaking changes. It's only about the exceptional migrations removing our design mistakes (renaming column, dropping column or table). Those should be very rare. Some companies never perform those. 
How would you handle the scenario we had when downgrading 4.22 to 4.21 where 4.22 removed the columns that 4.21 needed.

gastonfournier: Yes, you're right, but I feel we were intentionally adding tables and columns without thinking too much about them because we can easily change them later, and that proved to be an issue. Maybe we should also suggest not taking DB schema lightly and consider any change to it as an important change. 

But I agree with you so I'm approving the PR :+1:

gastonfournier (approved):

PR #3440, opened 2023-04-03
docs: Flesh out openapi for tag: Environments

chriswk:
### What
This fleshes out the documentation for actions tagged with Environments. 

A rather small PR since previous efforts had covered most of the operations already.

### DIFF
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -15,7 +15,7 @@ export const environmentProjectSchema = {
             type: 'string',
             example: 'production',
             description:
-                'The type of environment. Typically either development (non-production) or production',
+                'The [type of environment](https://docs.getunleash.io/reference/environments#environment-types).',
         },
         enabled: {
             type: 'boolean',
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -42,7 +42,7 @@ export const environmentProjectSchema = {
             minimum: 0,
             example: 5,
             description:
-                'How many client and frontend API tokens that have access to this project',
+                'The number of client and front-end API tokens that have access to this project',
         },
         projectEnabledToggleCount: {
             type: 'number',
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -6,7 +6,7 @@ export const environmentProjectSchema = {
     additionalProperties: false,
     required: ['name', 'type', 'enabled'],
     description:
-        "Describes an environment as it's configured for the project it's connected to",
+        "Describes a project's configuration in a given environment.",
     properties: {
         name: {
             type: 'string',
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -27,7 +27,7 @@ export const environmentProjectSchema = {
             type: 'boolean',
             example: false,
             description:
-                '`true` if the environment is not removable, otherwise `false`',
+                '`true` if the environment is protected, otherwise `false`. A *protected* environment can not be deleted.',
         },
         sortOrder: {
             type: 'number',
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -35,7 +35,7 @@ export const environmentProjectSchema = {
             type: 'integer',
             example: 1,
             description:
-                'Priority of the environment in a list of environments, the lower value, the higher up in the list the environment will appear',
+                'Priority of the environment in a list of environments, the lower the value, the higher up in the list the environment will appear',
         },
         projectApiTokenCount: {
             type: 'integer',
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -8,22 +8,37 @@ export const environmentProjectSchema = {
     properties: {
         name: {
             type: 'string',
+            example: 'development',
+            description: 'The name of the environment',
         },
         type: {
             type: 'string',
+            example: 'production',
+            description:
+                'The type of environment. Typically either development (non-production) or production',
         },
         enabled: {
             type: 'boolean',
+            example: true,
+            description:
+                '`true` if the environment is enabled for the project, otherwise `false`',
         },
         protected: {
             type: 'boolean',
+            example: false,
+            description:
+                '`true` if the environment is not removable, otherwise `false`',
         },
         sortOrder: {
             type: 'number',
+            example: 1,
+            description:
+                'Priority of the environment in a list of environments, the lower value, the higher up in the list the environment will appear',
         },
         projectApiTokenCount: {
             type: 'number',
             nullable: true,
+            description: 'How many API tokens are register',
         },
         projectEnabledToggleCount: {
             type: 'number',
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -5,6 +5,8 @@ export const environmentProjectSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['name', 'type', 'enabled'],
+    description:
+        "Describes an environment as it's configured for the project it's connected to",
     properties: {
         name: {
             type: 'string',
@@ -30,19 +32,24 @@ export const environmentProjectSchema = {
                 '`true` if the environment is protected, otherwise `false`. A *protected* environment can not be deleted.',
         },
         sortOrder: {
-            type: 'number',
+            type: 'integer',
             example: 1,
             description:
                 'Priority of the environment in a list of environments, the lower value, the higher up in the list the environment will appear',
         },
         projectApiTokenCount: {
-            type: 'number',
-            nullable: true,
-            description: 'How many API tokens are register',
+            type: 'integer',
+            minimum: 0,
+            example: 5,
+            description:
+                'How many client and frontend API tokens that have access to this project',
         },
         projectEnabledToggleCount: {
             type: 'number',
-            nullable: true,
+            minimum: 0,
+            example: 7,
+            description:
+                'Number of features enabled in this environment for this project',
         },
     },
     components: {},
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -4,9 +4,16 @@ export const environmentProjectSchema = {
     $id: '#/components/schemas/environmentProjectSchema',
     type: 'object',
     additionalProperties: false,
-    required: ['name', 'type', 'enabled'],
-    description:
-        "Describes a project's configuration in a given environment.",
+    required: [
+        'name',
+        'type',
+        'enabled',
+        'protected',
+        'sortOrder',
+        'projectApiTokenCount',
+        'projectEnabledToggleCount',
+    ],
+    description: "Describes a project's configuration in a given environment.",
     properties: {
         name: {
             type: 'string',
--- a/src/lib/openapi/spec/environment-project-schema.ts
+++ b/src/lib/openapi/spec/environment-project-schema.ts
@@ -4,15 +4,7 @@ export const environmentProjectSchema = {
     $id: '#/components/schemas/environmentProjectSchema',
     type: 'object',
     additionalProperties: false,
-    required: [
-        'name',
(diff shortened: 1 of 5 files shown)


src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -8,22 +8,37 @@ export const environmentProjectSchema = {
...
         name: {
             type: 'string',
+            example: 'development',
+            description: 'The name of the environment',
         },
         type: {
             type: 'string',
+            example: 'production',
+            description:
+                'The type of environment. Typically either development (non-production) or production',
```
thomasheartman, line 18: A couple things:
1. We should link to environment types here (added in the suggestion).
2. Is this a "read" schema / is it returned by a GET call? If so, I don't think we need the sentence about what it probably is. (removed in suggestion)

```suggestion
                'The [type of environment](https://docs.getunleash.io/reference/environments#environment-types).',
```

src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -8,22 +8,37 @@ export const environmentProjectSchema = {
...
             type: 'boolean',
+            example: true,
+            description:
+                '`true` if the environment is enabled for the project, otherwise `false`',
         },
         protected: {
             type: 'boolean',
+            example: false,
+            description:
+                '`true` if the environment is not removable, otherwise `false`',
```
thomasheartman, line 30: What does removable mean here? Does that mean that it can't be deleted? If so, how about something like this?

```suggestion
                '`true` if the environment is protected, otherwise `false`. A *protected* environment can not be deleted.',
```

src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -8,22 +8,37 @@ export const environmentProjectSchema = {
...
         },
         sortOrder: {
             type: 'number',
+            example: 1,
+            description:
+                'Priority of the environment in a list of environments, the lower value, the higher up in the list the environment will appear',
```
thomasheartman, line 36: I think I've asked this before, but there's no minimum or maximum for this right? Negative numbers are also okay? Do they have to be integers?

```suggestion
            example: 1,
            description:
                'Priority of the environment in a list of environments, the lower the value, the higher up in the list the environment will appear',
```

src/lib/openapi/spec/environment-project-schema.ts
thomasheartman, line 1: I'd like the schema itself to have a description as well, and not just its fields. What does the schema describe?

src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -8,22 +8,37 @@ export const environmentProjectSchema = {
...
         },
         projectApiTokenCount: {
             type: 'number',
             nullable: true,
+            description: 'How many API tokens are register',
```
thomasheartman, line 41: So this field can be a number OR null OR undefined (equivalent to `number | null | undefined` in TS)? When is it anything but a number? That should be mentioned in the description.

Also, I'm not sure I understand the description here. What does it mean? And what kind of API tokens does it count?

This suggestion needs to be reworked before it's ready.

```suggestion
            nullable: true,
            description: 'The number of client API tokens that have access to this project. `null` if .... `undefined` if ...'.
```

src/lib/openapi/spec/environment-schema.ts
```diff
@@ -25,12 +25,14 @@ export const environmentSchema = {
         },
         protected: {
             type: 'boolean',
+            example: true,
+            description: '`true` if the environment is a protected environment',
```
thomasheartman, line 29: As in the previous comment. What is a protected environment? If this shows up in a lot of places, we should add it to the docs. I couldn't find anything about protected environments in the docs, so we should probably add it anyway.

src/lib/routes/admin-api/environments.ts
```diff
@@ -73,8 +80,10 @@ export class EnvironmentsController extends Controller {
                 openApiService.validPath({
                     tags: ['Environments'],
                     operationId: 'getEnvironment',
+                    description: 'Gets the environment with `name`',
```
thomasheartman, line 83: This is the environment ID, right? Or is it its name? We should also have a summary.

```suggestion
                    summary: 'Get environment by ID',
                    description: 'Gets the environment whose ID is `name`',
```

src/lib/routes/admin-api/environments.ts
```diff
@@ -58,8 +61,12 @@ export class EnvironmentsController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Environments'],
+                    description: 'Gets all environments',
```
thomasheartman, line 64: This should probably be a summary and not a description.

We haven't discussed this previously and it's not written down (I'll do it 🙋🏼), but summaries should probably be expressed as a "bare infinitive", as in "call this endpoint to [summary]".



```suggestion
                    summary: 'Get all environments',
                    description: 'Retrieves all environments that exist in this Unleash instance.'
```

src/lib/routes/admin-api/environments.ts
```diff
@@ -89,8 +98,11 @@ export class EnvironmentsController extends Controller {
                 openApiService.validPath({
                     tags: ['Environments'],
                     operationId: 'getProjectEnvironments',
+                    description:
+                        'Gets the environments that are available for this project',
```
thomasheartman, line 102: Something like this, maybe? What makes an environment available?

```suggestion
                    summary: 'Get environments available to a project',
                    description:
                        'Gets all environments that are available for the specified project. An environment is considered available if ... what?',
```

src/lib/routes/admin-api/environments.ts
```diff
@@ -104,9 +116,14 @@ export class EnvironmentsController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Environments'],
+                    description:
+                        'Updates sort orders for the named environments',
```
thomasheartman, line 120:
```suggestion
                    summary: 'Update environment sort orders'
                    description:
                        'Updates sort orders for the named environments. Environments not specified are unaffected.',
```

src/lib/routes/admin-api/environments.ts
```diff
@@ -120,8 +137,12 @@ export class EnvironmentsController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Environments'],
+                    description: 'Toggles the environment with `name` on',
```
thomasheartman, line 140: Is this enable/disable environments for the global unleash instance or for projects? If the former, is it whether an env should be deprecated or not? 

I think the description could be a bit clearer on what this does. The operation is also missing a summary.

src/lib/routes/admin-api/environments.ts
```diff
@@ -135,8 +156,12 @@ export class EnvironmentsController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Environments'],
+                    description: 'Toggles the environment with `name` off',
```
thomasheartman, line 159: Same comment here as for the `on` endpoint above: there's no summary and I'm don't understand what turning an environment on or off means.

thomasheartman (changes requested):
Ayy, that was fast 🥳 Nice work, but I have a few suggestions, a few questions, and a few recommendations 😄 If you're wondering about anything, please let me know. I'll go ahead and update the parent task in the meantime

src/lib/openapi/spec/environment-project-schema.ts
chriswk: Agreed.

src/lib/openapi/spec/environment-project-schema.ts
chriswk: No min or max, should be integers yes, I don't think we accept decimals.

src/lib/openapi/spec/environment-project-schema.ts
chriswk: Agreed. Will fix

src/lib/openapi/spec/environment-project-schema.ts
thomasheartman: Nice, thanks! In that case, I think there's only a missing "the" in the sentence (as mentioned in the suggestion) 😄

src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -5,29 +5,51 @@ export const environmentProjectSchema = {
...
+                'Priority of the environment in a list of environments, the lower value, the higher up in the list the environment will appear',
         },
         projectApiTokenCount: {
-            type: 'number',
-            nullable: true,
+            type: 'integer',
+            minimum: 0,
+            example: 5,
+            description:
+                'How many client and frontend API tokens that have access to this project',
```
thomasheartman, line 45: Great! I suggest rewording it to the suggestion below, but I'll leave it up to you.


```suggestion
                'The number of client and front-end API tokens that have access to this project',
```

src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -5,29 +5,51 @@ export const environmentProjectSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['name', 'type', 'enabled'],
```
thomasheartman, line 7: So here's a thing to discuss. The way OpenAPI works, **all** properties that aren't specified in the `required` list are considered optional. From [the OpenAPI docs](https://swagger.io/docs/specification/data-models/data-types/#required):

>  By default, all object properties are optional. You can specify the required properties in the required list: 

That means, they can be undefined (as in: not included in the object). My guess is that most of the other fields on this schema aren't optional: they're always there. Can we specify all the properties that aren't optional in this list. And if a property really is optional, then its description should explicitly state when it will be missing. What do you think?

src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -5,29 +5,51 @@ export const environmentProjectSchema = {
...
+    description:
+        "Describes an environment as it's configured for the project it's connected to",
```
thomasheartman, line 9: Non-blocking and up to you, but I feel like this is actually describing a project and not an environment. Maybe we can flip it on its head? It's a tricky thing to explain, this 😅 

```suggestion
        "Describes a project's configuration in a given environment.",
```

src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -5,29 +5,51 @@ export const environmentProjectSchema = {
...
+            description:
+                'How many client and frontend API tokens that have access to this project',
         },
         projectEnabledToggleCount: {
             type: 'number',
-            nullable: true,
+            minimum: 0,
+            example: 7,
+            description:
+                'Number of features enabled in this environment for this project',
```
thomasheartman, line 52: Missing a definite article here

```suggestion
                'The number of features enabled in this environment for this project',
```

src/lib/openapi/spec/environments-project-schema.ts
```diff
@@ -6,6 +6,7 @@ export const environmentsProjectSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['version', 'environments'],
+    description: 'Environments defined for the current project',
```
thomasheartman, line 9: I'd switch this to something like this:

```suggestion
    description: 'Environments defined for a given project',
```

There's not necessarily a "current" project when you're working with APIs or reading specs, so I think this is clearer.

thomasheartman (commented):
Good going! 🙌🏼  Looks like you're still looking through some of the comments I left, but I had another look now. I think an important thing that we should get out of the way is how `required` properties work in OpenAPI (refer to the comment about that specifically): everything not mentioned explicitly can be undefined according to the spec. 

Let's figure out how we want to treat that.

src/lib/openapi/spec/environment-project-schema.ts
chriswk: I think you're right, since this is a response schema, not a request schema, it should be safe to mark them all as required. I'll update the PR

chriswk: Hmm, interesting discussion. The types we're returning here has these fields set as optional, so should we guarantee that they are here using required and change the Typescript type as well?

chriswk: I see we have a call for Environments which does not return counts.

chriswk: I'm guessing this is an instance of reusing types multiple places, where what we'd like is at least two different types, one with counts and one without, instead of trying to reuse the same type by making the fields optional

thomasheartman (approved):
Perfect! Yeah, this looks great to me for now 😄 I agree with the separate PR to split out the environment schemas from the other schemas, so let's get this done for now 💯

PR #3442, opened 2023-04-03
fix: concurrency issue when running multiple requests

gastonfournier:
## About the changes

Running multiple calls to the frontend concurrently creates many instances of unleash SDK client.

The test added in commit 34004b5be5f41d5cad646ce9e18cd1e307119ceb is to verify we can reproduce the issue with a test case, and we can see the error [here](https://github.com/Unleash/unleash/actions/runs/4596689332/jobs/8118467530?pr=3442#step:5:571)

The next commit introduces a fix for the issue

### How to reproduce manually
1. Start Unleash (it has to be a fresh application: to retry the test, first restart Unleash)
2. Run: 
```shell
#!/usr/bin/env bash

i=0
while [ $i -ne 15 ]
do
  curl http://localhost:4242/api/frontend -H 'Authorization: *:development.5487836a72f472d3494d92134fab6b4e07fcffe84c6c81a1d78448d0' &

  echo $i
  (( i=i+1 ));
done
```
You should see some errors, such as: 
```shell
[2023-04-03T10:48:49.818] [ERROR] services/proxy-service.ts - Error: The unleash SDK has been initialized more than 10 times
    at /home/ivar/code/unleash/node_modules/unleash-client/lib/unleash.js:96:29
```

### DIFF
--- a/src/lib/services/proxy-service.ts
+++ b/src/lib/services/proxy-service.ts
@@ -42,7 +42,8 @@ export class ProxyService {
 
     private readonly services: Services;
 
-    private readonly clients: Map<ApiUser['secret'], Unleash> = new Map();
+    private readonly clients: Map<ApiUser['secret'], Promise<Unleash>> =
+        new Map();
 
     private cachedFrontendSettings?: FrontendSettings;
 
@@ -99,14 +100,13 @@ export class ProxyService {
     private async clientForProxyToken(token: ApiUser): Promise<Unleash> {
         ProxyService.assertExpectedTokenType(token);
 
-        if (!this.clients.has(token.secret)) {
-            this.clients.set(
-                token.secret,
-                await this.createClientForProxyToken(token),
-            );
+        let client = this.clients.get(token.secret);
+        if (!client) {
+            client = this.createClientForProxyToken(token);
+            this.clients.set(token.secret, client);
         }
 
-        return this.clients.get(token.secret);
+        return client;
     }
 
     private async createClientForProxyToken(token: ApiUser): Promise<Unleash> {
@@ -134,13 +134,17 @@ export class ProxyService {
         return client;
     }
 
-    deleteClientForProxyToken(secret: string): void {
-        this.clients.get(secret)?.destroy();
-        this.clients.delete(secret);
+    async deleteClientForProxyToken(secret: string): Promise<void> {
+        let clientPromise = this.clients.get(secret);
+        if (clientPromise) {
+            const client = await clientPromise;
+            client.destroy();
+            this.clients.delete(secret);
+        }
     }
 
     stopAll(): void {
-        this.clients.forEach((client) => client.destroy());
+        this.clients.forEach((promise) => promise.then((c) => c.destroy()));
     }
 
     private static assertExpectedTokenType({ type }: ApiUser) {
--- a/src/lib/services/proxy-service.ts
+++ b/src/lib/services/proxy-service.ts
@@ -42,6 +42,11 @@ export class ProxyService {
 
     private readonly services: Services;
 
+    /**
+     * This is intentionally a Promise becasue we want to be able to await
+     * until the client (which might be being created by a different request) is ready
+     * Check this test that fails if we don't use a Promise: src/test/e2e/api/proxy/proxy.concurrency.e2e.test.ts
+     */
     private readonly clients: Map<ApiUser['secret'], Promise<Unleash>> =
         new Map();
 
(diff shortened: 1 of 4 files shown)


chriswk (approved):
Great job Gaston. :+1:

src/lib/services/proxy-service.ts
```diff
@@ -42,7 +42,8 @@ export class ProxyService {
 
     private readonly services: Services;
 
-    private readonly clients: Map<ApiUser['secret'], Unleash> = new Map();
+    private readonly clients: Map<ApiUser['secret'], Promise<Unleash>> =
```
kwasniew, line 45: It's weird that we store clients wrapped in promises instead of clients themselves. Why is that?

src/lib/services/proxy-service.ts
```diff
@@ -134,13 +134,17 @@ export class ProxyService {
         return client;
     }
 
-    deleteClientForProxyToken(secret: string): void {
-        this.clients.get(secret)?.destroy();
-        this.clients.delete(secret);
+    async deleteClientForProxyToken(secret: string): Promise<void> {
+        let clientPromise = this.clients.get(secret);
+        if (clientPromise) {
+            const client = await clientPromise;
```
kwasniew, line 140: this is promise storage leaking out to this method. It would be easier to work with the client here

src/lib/services/proxy-service.ts
```diff
@@ -134,13 +134,17 @@ export class ProxyService {
...
+            client.destroy();
+            this.clients.delete(secret);
+        }
     }
 
     stopAll(): void {
-        this.clients.forEach((client) => client.destroy());
+        this.clients.forEach((promise) => promise.then((c) => c.destroy()));
```
kwasniew, line 147: same here, we pay the price of storing promises and not clients

src/lib/services/proxy-service.ts
gastonfournier: Good question! And open to alternative solutions, but this is to store "we're processing/creating this unleash client, it might be ready soon", but we don't want to do that asynchronously as before because another request (or 15) might come while we're resolving the promise, and another client will be created. Ultimately, we want to know if someone is already processing a client for that secret, and wait until they finished before we proceed to create a new one

src/lib/services/proxy-service.ts
```diff
@@ -99,14 +100,13 @@ export class ProxyService {
...
         ProxyService.assertExpectedTokenType(token);
 
-        if (!this.clients.has(token.secret)) {
-            this.clients.set(
-                token.secret,
-                await this.createClientForProxyToken(token),
-            );
+        let client = this.clients.get(token.secret);
+        if (!client) {
+            client = this.createClientForProxyToken(token);
```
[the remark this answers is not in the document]
gastonfournier: This causes the same problem as before (used the test to validate my assumption). The await will be executed in the event loop and meanwhile, another thread will execute `clients.get`, receive an empty response and yield a new `createClientForProxyToken`

sighphyre:
@gastonfournier I think @kwasniew 's comments are on point here. I know why we went this way but my main worry is that some helpful person will read this code, decide this could be optimised and bring back the bug. If there's no cleaner way to do this, would a comment explaining why this code looks funny help prevent that in future?

src/lib/services/proxy-service.ts
```diff
@@ -99,14 +100,13 @@ export class ProxyService {
     private async clientForProxyToken(token: ApiUser): Promise<Unleash> {
```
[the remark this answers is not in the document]
gastonfournier: This is currently unchanged and I was lucky it was that way because it fits well with the change I did. 
The reason for this is probably that we need to execute a `client.start()` [here](https://github.com/Unleash/unleash/blob/217e4a491798dbfa343d623c6c547f39a7fbb999/src/lib/services/proxy-service.ts#L112-L132) and that's async, which bubbles up until this line

src/lib/services/proxy-service.ts
ivarconr: The challenge is that the current frontend API creates new node SDK instances on the fly for new tokens discovered. 
Creating an SDK instance is an async operation, as it will require to fetch data to be ready.

This is probably not the best design, but fixing the bug at hand we need to make sure we await the client. I do not think there is a way to force that in to not being async, with the current implementation, where SDK instances are created on the fly. 

In an optimal implementation of this, we should not use the SDK for other than the evaluation part. This would allow us to keep the caching and async part separate from the evaluation part.

gastonfournier:
As we discussed, we're going to merge the PR and continue the conversation around improving this offline.

PR #3443, opened 2023-04-03
feat: postgres lock

kwasniew:
## About the changes

Lock for DB migrations (or any other async operation) to run only from one instance at a time. The lock works for lock id + database name so will work with infra level multi-tenancy. 

`pg_advisory_lock` is like zookeeper inside postgres :) 

## Discussion points

* This lock has a timeout of 5 seconds. After 5 seconds of waiting for a lock no migration is run but the application startup proceeds. Another option would be to run a migration after 5 seconds anyway.
* It is behind a feature flag for now to keep it safe


sjaanus (approved):
Nice!

src/lib/util/db-lock.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const withDbLock =
+    (dbConfig: IDBOption, timeout = 5000, lockKey = defaultLockKey) =>
+    <A extends any[], R>(fn: (...args: A) => Promise<R>) =>
+    async (...args: A): Promise<R> => {
+        const client = new Client({ ...dbConfig, query_timeout: timeout });
+        try {
+            await client.connect();
+            // wait to obtain a lock
+            await client.query('SELECT pg_advisory_lock($1)', [lockKey]);
```
kwasniew, line 14: only one instance with a given key and DB name can run the SELECT. Other clients will await on the lock to free up. The timeout of 5000 prevents from infinite waiting. Realistically our migrations should run in under 5 seconds.

src/lib/util/db-lock.ts
```diff
@@ -0,0 +1,24 @@
...
+            const result = await fn(...args);
+            return result;
+        } catch (e) {
+            console.error('Locking error:', e);
+            throw e;
+        } finally {
+            await client.query('SELECT pg_advisory_unlock($1)', [lockKey]);
```
kwasniew, line 21: free up the lock

src/lib/util/db-lock.ts
```diff
@@ -0,0 +1,24 @@
...
+export const withDbLock =
+    (dbConfig: IDBOption, timeout = 5000, lockKey = defaultLockKey) =>
+    <A extends any[], R>(fn: (...args: A) => Promise<R>) =>
+    async (...args: A): Promise<R> => {
+        const client = new Client({ ...dbConfig, query_timeout: timeout });
+        try {
+            await client.connect();
+            // wait to obtain a lock
+            await client.query('SELECT pg_advisory_lock($1)', [lockKey]);
+            const result = await fn(...args);
```
kwasniew, line 15: the original async function that we're protecting with a lock

PR #3450, opened 2023-04-04
feat: add PAT kill switch

gastonfournier:
## Add the ability to disable Personal Access Tokens (PAT). 

This PR disables PAT middleware, making it impossible to use PAT if disabled, the UI is hidden and but the administrative APIs are still enabled, so it's possible to use it to create new PATs but they won't be effective.

This also stops existing users owning a PAT from using it if the kill switch is turned on.

By default the kill switch is disabled (i.e. PAT is enabled by default)


src/lib/app.ts
```diff
@@ -102,7 +102,9 @@ export default async function getApp(
         ),
     );
 
-    app.use(baseUriPath, patMiddleware(config, services));
+    if (!config.flagResolver.isEnabled('personalAccessTokensKillSwitch')) {
```
ivarconr, line 105: this will not work as intended. Feature flags are dynamic and can change during runtime of the application. This will either load/not load the middle-ware at startup depending on the flags status on Unleash startup time. 

We need to move the check in to the middleware itself to ensure it can change over time.

ivarconr:
How will it impact service-accounts? 

(I suspect you will be able to create them, but not use the token as long as the kill-switch is enabled).

src/lib/app.ts
gastonfournier: Yes, I considered this but I thought that the way they want this to work would be ok, but there's no harm in having it as a runtime configuration. It was actually like that before but in this case, I believed the customer wanted it off for good. In any case, let's err for the most general use case which should consider enabling this without having to restart :+1:

gastonfournier:
> How will it impact service-accounts?
> 
> (I suspect you will be able to create them, but not use the token as long as the kill-switch is enabled).

:thinking: maybe this is what's failing in e2e tests... I'll double check this part

src/lib/app.ts
gastonfournier: [9d3e900](https://github.com/Unleash/unleash/pull/3450/commits/9d3e9002c0b46622b19f265a063183f074721afb) should address this

gastonfournier:
> How will it impact service-accounts?
> 
> (I suspect you will be able to create them, but not use the token as long as the kill-switch is enabled).

Yes, I just checked this solution does not allow you to use service accounts... we could disable it if they don't want to use them... the other alternative would be to just hide the UI for PAT and only have the admin APIs enabled for users who are service accounts (that way we don't allow users to create PAT if they're normal users)

ivarconr:
> @ivarconr this alternative: #3454 allows us to still have service accounts

For now I prefer the alternative as it feels simpler and less risky. If we end up doing this we would have a lot more things to consider.

PR #3460, opened 2023-04-05
task: Follow up environment openapi docs

chriswk:
Cleans up Environment usages, using separate interfaces instead of having nullable fields that are set or not set depending on which endpoint we're working with. This also allows us to set these extra count fields as required in the schema for the response.


gastonfournier (approved):
This looks neat!

thomasheartman (commented):
This looks good to me! But have you introduced any new schemas in this iteration? It seems that even with the new required properties, you didn't need to change the schemas you used anywhere. Was this taken care of in the previous iteration (sorry, it's been a week 😅 )

thomasheartman:
@chriswk How's it looking for this PR? Can we move it one way or the other? Any blockers? I'd still love for you to tell me about whether any new schemas are introduced etc. It's been ... a few weeks now, so I don't have all the context anymore.

src/lib/services/environment-service.ts
```diff
@@ -52,10 +57,20 @@ export default class EnvironmentService {
...
-    async getAll(): Promise<IEnvironment[]> {
+    async getAll(): Promise<IEnvironmentWithCounts[]> {
         return this.environmentStore.getAllWithCounts();
     }
 
+    async getWithCounts(name: string): Promise<IEnvironmentWithCounts> {
+        let matches = await this.environmentStore.getAllWithCounts({
+            name: name,
+        });
+        if (matches.length > 0) {
```
prabodhmeshram, line 68: Minor suggestion 
```suggestion
        if (matches.length) {
```

src/lib/services/environment-service.ts
```diff
@@ -52,10 +57,20 @@ export default class EnvironmentService {
...
     }
 
-    async getAll(): Promise<IEnvironment[]> {
+    async getAll(): Promise<IEnvironmentWithCounts[]> {
         return this.environmentStore.getAllWithCounts();
     }
 
+    async getWithCounts(name: string): Promise<IEnvironmentWithCounts> {
+        let matches = await this.environmentStore.getAllWithCounts({
+            name: name,
```
prabodhmeshram, line 66: More concise way to do it
```suggestion
            name,
```

PR #3473, opened 2023-04-06
WIP: working but hacky

gastonfournier:
## About the changes
This PR is an attempt to do virtualized Autocomplete by using react-window as mentioned in MUI documentation: https://mui.com/material-ui/react-autocomplete/#virtualization

To implement this we need to write our own ListboxComponent which is what's used to render the listbox. 

By doing this the list is much more responsive but we lose the group name stickiness.

The implementation should be considered a proof of concept. I kept the older implementation that can be used to compare the behavior.

To create 1k users you can use the following script (just specify your token)
```bash
#!/usr/bin/env bash
ITERATIONS=${1:-1000}  # use first parameter as number of iterations
START=${START:-1}      # env variable to tweak the initial number
ROOT_ROLE=${ROLE:-2}   # default to editor role (1: admin, 2: editor, 3: viewer)
PREFIX=${PREFIX:-test} # prefix for name and email
TOKEN=${TOKEN:-*:*.secret-123}

set -x
for i in $( seq ${START} ${ITERATIONS} ); do
    curl --location --request POST 'http://localhost:4242/api/admin/user-admin' \
            --header "Authorization: ${TOKEN}" \
            --header 'Content-Type: application/json' \
            --data-raw "{ \"name\": \"${PREFIX} ${i}\", \"email\": \"${PREFIX}${i}@getunleash.io\", \"sendEmail\": false, \"rootRole\": ${ROOT_ROLE} }"
done
```

### DIFF
--- a/frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
+++ b/frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
@@ -1,4 +1,12 @@
-import { Autocomplete, Checkbox, styled, TextField } from '@mui/material';
+import {
+    Autocomplete,
+    autocompleteClasses,
+    Checkbox,
+    Popper,
+    styled,
+    TextField,
+} from '@mui/material';
+import { VariableSizeList, ListChildComponentProps } from 'react-window';
 import CheckBoxOutlineBlankIcon from '@mui/icons-material/CheckBoxOutlineBlank';
 import CheckBoxIcon from '@mui/icons-material/CheckBox';
 import { IUser } from 'interfaces/user';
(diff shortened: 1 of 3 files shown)


frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
```diff
@@ -99,48 +215,112 @@ export const GroupFormUsersSelect: VFC<IGroupFormUsersSelectProps> = ({
...
-                value={users as UserOption[]}
-                onChange={(event, newValue, reason) => {
-                    if (
-                        event.type === 'keydown' &&
-                        (event as React.KeyboardEvent).key === 'Backspace' &&
-                        reason === 'removeOption'
-                    ) {
-                        return;
+    const virtualized = true;
+    if (virtualized)
```
gastonfournier, line 219: We can have a threshold to decide when to show the virtualized dropdown list (e.g. if length > 200 show virtualized), but this requires us to maintain two implementations. Also, we can have a feature flag to always display this.

I think we need to abstract this in a reusable component, but as a first step this could be good enough to prove the solution.

PR #3475, opened 2023-04-07
Improved lead time calculation

kwasniew:
## About the changes

* adding new flag to compare avg time to prod against the original impementation
* in a transitional period we gonna run both until we sunset the old implementation
* what is needed in DB query results in old impl: all features, all feature enabled events for those features
* what is needed in DB query results in the new impl: we only read feature created/first enabled tuples
* what is the new DB query: select first enabled event for each feature but only after feature creation (handles feature deletions in the past), exclude only kill-switch long lived toggles, select only prod environments
* why do we calculate avg and time to prod for each feature in memory: we need date-fns for advanced full day/daylight time savings aware calculation which I couldn't reliably reproduce in the DB itself

Closes #


src/lib/read-models/time-to-production/time-to-production.ts
```diff
@@ -27,6 +29,7 @@ export class TimeToProduction {
         this.events = events;
     }
 
+    // todo: remove
```
Tymek, line 32:
```suggestion
    /**
     * @deprecated TODO: remove
     */
```

src/lib/db/project-stats-store.ts
```diff
@@ -107,6 +110,41 @@ class ProjectStatsStore implements IProjectStatsStore {
...
+            .where('environments.type', '=', 'production')
+            // kill-switch is long lived
+            .where('features.type', '<>', 'kill-switch')
+            // exclude events for features that were previously deleted
+            .where(this.db.raw('events.created_at > features.created_at'))
+            .where('features.project', '=', projectId)
+            .orderBy('events.feature_name')
+            // first enabled event
+            .orderBy('events.created_at', 'asc');
+        return result;
```
Tymek, line 146: that query is less complex then I expected

src/lib/services/project-service.ts
```diff
@@ -778,12 +784,32 @@ export default class ProjectService {
...
 
+        const avgTimeToProdCurrentWindowFast =
+            TimeToProduction.calculateAverageTimeToProd(
+                await this.projectStatsStore.getTimeToProdDates(projectId),
+            );
+        const avgTimeToProdCurrentWindowSlow =
+            timeToProduction.calculateAverageTimeToProd();
+
+        const avgTimeToProdCurrentWindow = this.flagResolver.isEnabled(
+            'projectStatusApiImprovements',
```
Tymek, line 802: I see you're not using variants, OK, new flag is easier to understand

Tymek (approved):
Let's test it in sandbox 👍🏻

src/lib/services/project-service.ts
kwasniew: I found that flagResolver interface doesn't expose variants

src/lib/db/project-stats-store.ts
kwasniew: I moved time difference calcs to memory since we need date-fns or similar advanced time calculator

src/lib/read-models/time-to-production/time-to-production.ts
kwasniew: I'm hoping to remove those TODOs no later than next week so I don't invest time in them :)

PR #3477, opened 2023-04-07
refactor: switching to new stats calculations

kwasniew:
## About the changes

* removes old code with stats calculations in memory (confirmed in sandbox that results are the same for old and new code)
* adds tests for all corner cases of DB querying for stats
* deletes unused code since DB is doing most of the heavylifting


andreas-unleash (commented):
LG! Nice job

andreas-unleash (approved):

sjaanus (approved):

PR #3486, opened 2023-04-10
chore:  docs for public signup token tag

andreas-unleash:
## About the changes

Closes # [1-863](https://linear.app/unleash/issue/1-836/update-endpoints-for-tag-public-signup-tokens)


src/lib/openapi/spec/public-signup-token-schema.ts
thomasheartman, line 1: Can you add a top-level description for this schema? What is it used for?

We should probably include a link to the docs for public invite links too: https://docs.getunleash.io/reference/public-signup#public-sign-up-tokens

src/lib/openapi/spec/public-signup-token-schema.ts
```diff
@@ -18,28 +18,40 @@ export const publicSignupTokenSchema = {
...
             description:
                 'The public signup link for the token. Users who follow this link will be taken to a signup page where they can create an Unleash user.',
             type: 'string',
+            example:
+                'https://sandbox.getunleash.io/enterprise/new-user?invite=a3c84b25409ea8ca1782ef17f94a42fc',
         },
         name: {
+            description: "The token's name",
             type: 'string',
+            example: 'default',
```
thomasheartman, line 35: Is "default" a good name for a token? Is it something we would expect to see? It feels a little ... generic. We could use the example to promote what we think are good names, I think 💁🏼

src/lib/openapi/spec/public-signup-token-schema.ts
```diff
@@ -18,28 +18,40 @@ export const publicSignupTokenSchema = {
...
         },
         enabled: {
+            description: 'Whether the token is active',
```
thomasheartman, line 38: Can a token be active **and** past its expiration date? Or is that an invalid state? If so, maybe we could say something like this?

```suggestion
            description: 'Whether the token is active. This property will always be `false` for a token that has expired.',
```

src/lib/openapi/spec/public-signup-token-schema.ts
```diff
@@ -18,28 +18,40 @@ export const publicSignupTokenSchema = {
...
             format: 'date-time',
+            example: '2023-04-12T11:13:31.960Z',
         },
         createdAt: {
             type: 'string',
             format: 'date-time',
+            example: '2023-04-12T11:13:31.960Z',
         },
         createdBy: {
+            description: 'The user that created the token',
```
thomasheartman, line 53: It's a bit unclear to me what part of the user this is. Is it their username? Their email? Their displayname? Because it's just a string, I assume it's one of them. Do we use a fallback method if the first isn't available? For instance:

If it's the user's email, but the user doesn't have an email, does it fall back to using the username?

src/lib/openapi/spec/public-signup-token-update-schema.ts
thomasheartman, line 1: This schema also needs a top-level description. Also, **can** you update a token's expiration? I thought you couldn't?

src/lib/openapi/spec/public-signup-tokens-schema.ts
```diff
@@ -11,6 +11,7 @@ export const publicSignupTokensSchema = {
     properties: {
         tokens: {
             type: 'array',
+            description: 'An array of all the public signup tokens',
```
thomasheartman, line 14: This property is missing an example. What does it look like?

thomasheartman (changes requested):
Nice going! 🏆  There's a few places where I think you've forgotten examples/descriptions, but I've added comments for all of that. I also have a few questions about some of the descriptions and behaviors, but it's all there 🌸

src/lib/openapi/spec/public-signup-token-schema.ts
andreas-unleash: sounds good

src/lib/openapi/spec/public-signup-tokens-schema.ts
andreas-unleash: added

thomasheartman (commented):
Looking good! Any reason why you haven't accepted the suggestions for descriptions on [`expiresAt`](https://github.com/Unleash/unleash/pull/3486/files#r1162842318) and [`createdAt`](https://github.com/Unleash/unleash/pull/3486/files#r1162843754)?

Once we have resolved those, I'm happy to get it in 😄

thomasheartman (approved):
Nice one! ✅

PR #3487, opened 2023-04-10
chore: fix docs

andreas-unleash:
Adds `mode` and `defaultStickiness` to the project form.
Adds a new img

## About the changes

Closes #


kwasniew (approved):

website/docs/reference/projects.md
```diff
@@ -43,13 +43,15 @@ The UI shows the currently available projects. To create a new project, use the
...
 The configuration of a new Project is now available. the following input is available to create the new Project.
 
-![A project creation form. The form fields are labeled "project ID", "name", and "description". The "Create" button is highlighted.](/img/projects_save_new_project.png)
-
-| Item         | Description                        |
-| ------------ | ---------------------------------- |
-| Project Id   | Id for this Project                |
-| Project name | The name of the Project.           |
-| Description  | A short description of the project |
+![A project creation form. The form fields are labeled "project ID", "name", "description", "mode" and "defaultStickiness". The "Create" button is highlighted.](/img/projects_save_new_project_v2.png)
```
thomasheartman, line 46: Non-blocking: This is my fault, but I think that description is more wordy than what it has to be. Something like this, maybe?

```suggestion
![A project creation form. The "Create" button is highlighted.](/img/projects_save_new_project_v2.png)
```

website/docs/reference/projects.md
```diff
@@ -43,13 +43,15 @@ The UI shows the currently available projects. To create a new project, use the
...
+
+| Item               | Description                                                                                 |
+|--------------------|---------------------------------------------------------------------------------------------|
+| Project Id         | Id for this Project                                                                         |
+| Project name       | The name of the Project.                                                                    |
+| Description        | A short description of the project                                                          |
+| Mode               | The project [collaboration mode](/reference/project-collaboration-mode.md)                  |
+| Default Stickiness | The default stickiness for the project affecting the default stickiness value for variants and Gradual Rollout strategy                                                                 |
```
thomasheartman, line 54: Splitting into two sentences makes it a little easier to read I think. What do you think?

```suggestion
| Default Stickiness | The default stickiness for the project. This setting controls the default stickiness value for variants and for the gradual rollout strategy.                                                                 |
```

thomasheartman (changes requested):
So this looks pretty good to me! 😄 I've got two minor change suggestions that you can choose to accept or not. 

There's only one thing that's keeping me from approving this:
It doesn't look like you've deleted the old image. Could you do that? I don't think it's being used anywhere else, is it? If it is, then ignore this 💁🏼

PR #3496, opened 2023-04-11
chore: Improve openapi documentation for tags

gastonfournier:
## About the changes
1. Create tag should not throw a 500 when bad data is provided
2. Added summary, description and examples to open API endpoints


src/lib/services/feature-tag-service.ts
```diff
@@ -129,12 +130,20 @@ class FeatureTagService {
...
+                        createdBy: userName,
+                        data: tag,
+                    });
+                } catch (err) {
+                    if (err.code === FOREIGN_KEY_VIOLATION) {
+                        throw new BadDataError(
+                            `Tag type '${tag.type}' does not exist`,
+                        );
+                    }
+                }
```
gastonfournier, line 146: I'm not a big fan of nested try catches, the problem IMO is the getTag throwing an exception at the store level rather than returning `undefined`. Alternatively, we can move handling the `FOREIGN_KEY_VIOLATION` to the store level.

gastonfournier: From a consistency point of view, we're catching FK violations at the service layer: 

1. https://github.com/Unleash/unleash/blob/0b0191a2f746a146e7b731714aff7fc9415ec43a/src/lib/services/api-token-service.ts#L241
2. https://github.com/Unleash/unleash/blob/0b0191a2f746a146e7b731714aff7fc9415ec43a/src/lib/services/feature-toggle-service.ts#L456

chriswk: I feel like cleaning up the usage and handling it in the service layer is probably better. Does that mean you'd like to swap to undefined as the return value at the store level? That feels fair, what do we do in the feature store?

gastonfournier: At the feature-store we're also throwing the exception: https://github.com/Unleash/unleash/blob/openapi-tags/src/lib/db/feature-strategy-store.ts#L370-L372

So, now I'm unsure... Even if returning `undefined` and handling that at the service layer seems like a good thing to do, I also like to keep the code consistent and maybe change this as a whole later (if we decide to). 

I rather not have a try-catch inside a catch block, so I'm considering moving the FK handling to the store layer (although that also breaks consistency)

src/lib/services/tag-schema.ts
```diff
@@ -1,11 +1,16 @@
 import Joi from 'joi';
 
 import { customJoi } from '../routes/util';
+import { TAG_MAX_LENGTH, TAG_MIN_LENGTH } from '../../lib/openapi';
```
gastonfournier, line 4: This is to gradually make openapi schema the source of truth

src/lib/routes/admin-api/feature.ts
```diff
@@ -110,9 +113,15 @@ class FeatureController extends Controller {
             permission: NONE,
             middleware: [
                 openApiService.validPath({
+                    summary: 'Get all tags for a feature.',
+                    description:
+                        'Retrieves all the tags for a feature name. If the feature does not exist it returns an empty list.',
```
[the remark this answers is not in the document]
gastonfournier: I think so, but that could be a breaking change... maybe the UI is not handling that situation... so I preferred to stick to just documenting the behavior, but because it's strange, I wanted to be explicit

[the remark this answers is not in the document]
gastonfournier: We can add a task though and consider this for v5, wdyt?

src/lib/routes/admin-api/feature.ts
```diff
@@ -159,9 +176,15 @@ class FeatureController extends Controller {
             handler: this.removeTag,
             middleware: [
                 openApiService.validPath({
+                    summary: 'Removes a tag from a feature.',
+                    description:
+                        'Removes a tag from a feature if it exists. If the feature exist and the tag does not, it returns a successful response.',
```
nunogois, line 181:
```suggestion
                        'Removes a tag from a feature. If the feature exists but the tag does not, it returns a successful response.',
```

src/lib/routes/admin-api/feature.ts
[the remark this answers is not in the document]
nunogois: Ah right, that makes sense. Yep, I would consider it for v5.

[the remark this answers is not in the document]
gastonfournier: Added a linear issue: https://linear.app/unleash/issue/2-905/featuretags-should-return-404-when-feature-does-not-exist

src/lib/services/feature-tag-service.ts
nunogois: +1 to try to keep consistency for now until we can apply a better generic solution, as long as we're aware that this is something we should try to improve soon-ish. Returning `undefined` in that scenario and handling in the service layer seems OK to me.

nunogois (approved):
LGTM

PR #3497, opened 2023-04-11
fix: security vulnerability

gastonfournier:
This was an oversight of https://github.com/Unleash/unleash/pull/3402


thomasheartman (approved):
Makes sense 😄

PR #3501, opened 2023-04-11
docs: Added docs for edge endpoints

chriswk:
Adding documentation for the edge endpoints. Also separating request and response schema for our validate endpoint to make clear that we expect a list of strings as input, but yield tokens as output.

### DIFF
--- a/src/lib/openapi/spec/bulk-metrics-schema.ts
+++ b/src/lib/openapi/spec/bulk-metrics-schema.ts
@@ -7,6 +7,8 @@ export const bulkMetricsSchema = {
     $id: '#/components/schemas/bulkMetricsSchema',
     type: 'object',
     required: ['applications', 'metrics'],
+    description:
+        'A batch of metrics accumulated by Edge (or other compatible applications). Includes both application registrations as well usage metrics from clients',
     properties: {
         applications: {
             type: 'array',
--- a/src/lib/openapi/spec/bulk-metrics-schema.ts
+++ b/src/lib/openapi/spec/bulk-metrics-schema.ts
@@ -6,6 +6,7 @@ import { clientMetricsEnvSchema } from './client-metrics-env-schema';
 export const bulkMetricsSchema = {
     $id: '#/components/schemas/bulkMetricsSchema',
     type: 'object',
+    required: ['applications', 'metrics'],
     properties: {
         applications: {
             type: 'array',
--- a/src/lib/openapi/spec/bulk-registration-schema.ts
+++ b/src/lib/openapi/spec/bulk-registration-schema.ts
@@ -9,7 +9,7 @@ export const bulkRegistrationSchema = {
         connectVia: {
             type: 'array',
             description:
-                'a list of applications this app registration has been registered through',
+                'A list of applications this app registration has been registered through or empty',
             items: {
                 type: 'object',
                 required: ['appName', 'instanceId'],
--- a/src/lib/openapi/spec/bulk-registration-schema.ts
+++ b/src/lib/openapi/spec/bulk-registration-schema.ts
@@ -8,6 +8,8 @@ export const bulkRegistrationSchema = {
     properties: {
         connectVia: {
             type: 'array',
+            description:
+                'a list of applications this app registration has been registered through',
             items: {
                 type: 'object',
                 required: ['appName', 'instanceId'],
@@ -20,29 +22,48 @@ export const bulkRegistrationSchema = {
                     },
                 },
             },
+            example: ['unleash-edge'],
         },
         appName: {
+            description:
+                'The name of the application that is evaluating toggles',
             type: 'string',
+            example: 'Ingress load balancer',
         },
         environment: {
+            description: 'Which environment is the application running in',
             type: 'string',
+            example: 'development',
         },
         instanceId: {
+            description: 'A unique identifier for the application',
             type: 'string',
+            example: 'Random UUID',
         },
         interval: {
+            description:
+                'How often (in seconds) does the application refresh its features',
             type: 'number',
+            example: 10,
         },
         started: {
+            description: 'When did the application start',
+            example: '1952-03-11T12:00:00.000Z', //GNU Douglas Adams
             $ref: '#/components/schemas/dateSchema',
         },
         strategies: {
+            description:
+                'Which [strategies](https://docs.getunleash.io/reference/activation-strategies) are enabled in the application',
             type: 'array',
+            example: ['standard', 'gradualRollout'],
             items: {
                 type: 'string',
             },
         },
         sdkVersion: {
+            summary: 'Version identifier for the SDK used',
+            description: 'Typically <client>:<version>',
+            example: 'unleash-client-java:8.0.0',
             type: 'string',
         },
     },
--- a/src/lib/openapi/spec/bulk-registration-schema.ts
+++ b/src/lib/openapi/spec/bulk-registration-schema.ts
@@ -10,7 +10,7 @@ export const bulkRegistrationSchema = {
         connectVia: {
             type: 'array',
             description:
-                'A list of applications this app registration has been registered through or empty',
+                'A list of applications this app registration has been registered through. If connected directly to Unleash, this is an empty list. \n This can be used in later visualizations to tell how many levels of proxy or Edge instances our SDKs have connected through',
             items: {
                 type: 'object',
                 required: ['appName', 'instanceId'],
@@ -23,7 +23,9 @@ export const bulkRegistrationSchema = {
                     },
                 },
             },
-            example: ['unleash-edge'],
+            example: [
+                { appName: 'unleash-edge', instanceId: 'edge-pod-bghzv5' },
+            ],
         },
         appName: {
             description:
@@ -63,8 +65,8 @@ export const bulkRegistrationSchema = {
             },
         },
         sdkVersion: {
-            summary: 'Version identifier for the SDK used',
-            description: 'Typically <client>:<version>',
+            description:
+                'The version the sdk is running. Typically <client>:<version>',
             example: 'unleash-client-java:8.0.0',
             type: 'string',
         },
--- a/src/lib/openapi/spec/bulk-registration-schema.ts
+++ b/src/lib/openapi/spec/bulk-registration-schema.ts
@@ -4,7 +4,8 @@ import { dateSchema } from './date-schema';
 export const bulkRegistrationSchema = {
     $id: '#/components/schemas/bulkRegistrationSchema',
     type: 'object',
-    required: ['appName', 'instanceId'],
+    required: ['appName', 'instanceId', 'environment'],
+    description: `An application registration. Defines the format POSTed by our server-side SDKs when they're starting up`,
     properties: {
         connectVia: {
             type: 'array',
@@ -31,29 +32,30 @@ export const bulkRegistrationSchema = {
             example: 'Ingress load balancer',
         },
         environment: {
-            description: 'Which environment is the application running in',
+            description: 'Which environment the application is running in',
             type: 'string',
             example: 'development',
         },
         instanceId: {
-            description: 'A unique identifier for the application',
+            description:
+                'A [(somewhat) unique identifier](https://docs.getunleash.io/reference/sdks/node#advanced-usage) for the application',
             type: 'string',
-            example: 'Random UUID',
+            example: 'application-name-dacb1234',
         },
         interval: {
             description:
-                'How often (in seconds) does the application refresh its features',
+                'How often (in seconds) the application refreshes its features',
             type: 'number',
             example: 10,
         },
         started: {
-            description: 'When did the application start',
+            description: 'The application started at',
             example: '1952-03-11T12:00:00.000Z', //GNU Douglas Adams
             $ref: '#/components/schemas/dateSchema',
         },
         strategies: {
             description:
-                'Which [strategies](https://docs.getunleash.io/reference/activation-strategies) are enabled in the application',
+                'Enabled [strategies](https://docs.getunleash.io/reference/activation-strategies) in the application',
             type: 'array',
             example: ['standard', 'gradualRollout'],
             items: {
--- a/src/lib/openapi/spec/client-metrics-env-schema.ts
+++ b/src/lib/openapi/spec/client-metrics-env-schema.ts
@@ -9,27 +9,45 @@ export const clientMetricsEnvSchema = {
     properties: {
         featureName: {
             type: 'string',
+            description: 'Name of the feature checked by the SDK',
+            example: 'my.special.feature',
         },
         appName: {
+            description: 'The name of the application the SDK is being used in',
             type: 'string',
+            example: 'accounting',
         },
         environment: {
+            description: 'Which environment the SDK is being used in',
             type: 'string',
+            example: 'development',
         },
         timestamp: {
+            description: 'The time window these metrics are valid for',
+            example: '1926-05-08T12:00:00.000Z',
             $ref: '#/components/schemas/dateSchema',
         },
         yes: {
+            description: 'How many times the toggle evaluated to true',
             type: 'number',
+            example: 974,
         },
         no: {
+            description: 'How many times the toggle evaluated to false',
             type: 'number',
+            example: 50,
         },
         variants: {
+            description: 'How many times each variant was returned',
             type: 'object',
             additionalProperties: {
                 type: 'integer',
                 minimum: 0,
+            },
+            example: {
+                variantA: 15,
+                variantB: 25,
+                variantC: 5,
             },
         },
     },
--- a/src/lib/openapi/spec/client-metrics-env-schema.ts
+++ b/src/lib/openapi/spec/client-metrics-env-schema.ts
@@ -4,8 +4,9 @@ import { FromSchema } from 'json-schema-to-ts';
 export const clientMetricsEnvSchema = {
     $id: '#/components/schemas/clientMetricsEnvSchema',
     type: 'object',
-    required: ['featureName', 'appName'],
+    required: ['featureName', 'appName', 'environment'],
     additionalProperties: true,
+    description: 'Used for reporting feature evaluation results from SDKs',
     properties: {
         featureName: {
             type: 'string',
@@ -23,7 +24,8 @@ export const clientMetricsEnvSchema = {
             example: 'development',
         },
         timestamp: {
-            description: 'The time window these metrics are valid for',
+            description:
+                'The start of the time window these metrics are valid for. The window is 1 hour wide',
             example: '1926-05-08T12:00:00.000Z',
             $ref: '#/components/schemas/dateSchema',
         },
--- a/src/lib/openapi/spec/edge-token-schema.ts
+++ b/src/lib/openapi/spec/edge-token-schema.ts
@@ -26,7 +26,7 @@ export const edgeTokenSchema = {
         },
         token: {
             description:
-                '[Unleash API tokens](https://docs.getunleash.io/reference/api-tokens-and-client-keys) are comprised of three parts. <project(s)>:<environment>.randomcharacters',
+                'The actual token value. [Unleash API tokens](https://docs.getunleash.io/reference/api-tokens-and-client-keys) are comprised of three parts. <project(s)>:<environment>.randomcharacters',
             type: 'string',
             example:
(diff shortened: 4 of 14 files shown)


src/lib/openapi/spec/bulk-registration-schema.ts
```diff
@@ -8,6 +8,8 @@ export const bulkRegistrationSchema = {
     properties: {
         connectVia: {
             type: 'array',
+            description:
+                'a list of applications this app registration has been registered through',
```
gastonfournier, line 12:
```suggestion
                'A list of applications this app registration has been registered through or empty',
```

src/lib/openapi/spec/client-metrics-env-schema.ts
```diff
@@ -9,28 +9,46 @@ export const clientMetricsEnvSchema = {
...
             type: 'string',
+            example: 'accounting',
         },
         environment: {
+            description: 'Which environment the SDK is being used in',
             type: 'string',
+            example: 'development',
         },
         timestamp: {
+            description: 'The time window these metrics are valid for',
```
gastonfournier, line 26: Time window makes me think about a range of dates, but the timestamp is just a date. Is this the start date of the window and is the end date: `start date + push metrics interval`? I think it's worth clarifying this

src/lib/routes/edge-api/index.ts
```diff
@@ -59,12 +63,15 @@ export default class EdgeController extends Controller {
             middleware: [
                 this.openApiService.validPath({
                     tags: ['Edge'],
+                    summary:
+                        'Accepts a list of tokens, returns tokens that are valid and which projects they can access',
+                    description:
+                        'Send a list of tokens, only returns the valid ones',
```
gastonfournier, line 69: I think the description should be longer than the summary, also a small suggestion:
```suggestion
                    summary: 'Filter a list of tokens returning only the valid ones',
                    description:
                        'Accepts a list of tokens, and returns those that are valid with the projects they can access',
```

src/lib/routes/edge-api/index.ts
```diff
@@ -78,23 +85,24 @@ export default class EdgeController extends Controller {
             middleware: [
                 this.openApiService.validPath({
                     tags: ['Edge'],
+                    summary:
+                        'Accepts batched metrics from Edge, for insertion into our metrics storage',
                     operationId: 'bulkMetrics',
                     requestBody: createRequestSchema('bulkMetricsSchema'),
                     responses: {
                         202: emptyResponse,
+                        ...getStandardResponses(400),
```
gastonfournier, line 94: Don't we have any permission validation that could yield a 401? I noticed [this comment](https://github.com/Unleash/unleash/blob/19d318d684a63c5690c54995f5103625499733dc/src/lib/routes/edge-api/index.ts#L77) so maybe is something we missed? I think the idea was to do a manual validation, but it seems we didn't do it

gastonfournier (approved):
Looks fine, just some minor comments

src/lib/openapi/spec/bulk-metrics-schema.ts
thomasheartman, line 1: Could you add a top-level description to this schema? What does the schema describe?

It's also missing descriptions and examples for its properties.

src/lib/openapi/spec/bulk-registration-schema.ts
thomasheartman, line 1: This one is also missing a top-level description

src/lib/openapi/spec/bulk-registration-schema.ts
```diff
@@ -20,29 +22,48 @@ export const bulkRegistrationSchema = {
...
+            example: 'development',
         },
         instanceId: {
+            description: 'A unique identifier for the application',
             type: 'string',
+            example: 'Random UUID',
         },
         interval: {
+            description:
+                'How often (in seconds) does the application refresh its features',
```
thomasheartman, line 45: I think questions should be rephrased as statements, as suggested here. The same goes for other similar sentences

```suggestion
                'How often (in seconds) the application refreshes its features',
```

src/lib/openapi/spec/bulk-registration-schema.ts
```diff
@@ -20,29 +22,48 @@ export const bulkRegistrationSchema = {
...
+            example: ['unleash-edge'],
         },
         appName: {
+            description:
+                'The name of the application that is evaluating toggles',
             type: 'string',
+            example: 'Ingress load balancer',
         },
         environment: {
+            description: 'Which environment is the application running in',
```
thomasheartman, line 34:
```suggestion
            description: 'Which environment the application is running in',
```

src/lib/openapi/spec/bulk-registration-schema.ts
```diff
@@ -20,29 +22,48 @@ export const bulkRegistrationSchema = {
...
             type: 'string',
+            example: 'development',
         },
         instanceId: {
+            description: 'A unique identifier for the application',
             type: 'string',
+            example: 'Random UUID',
```
thomasheartman, line 41: I suspect that the example here is more of a description than an example of what we would usually expect. If so, would it make sense to do something like this?

```suggestion
            description: 'A unique identifier for the application. Commonly a UUID.',
            type: 'string',
            example: 'AD4F7B4C-0C0C-4AF1-B3C4-95C174E28227',
```

src/lib/openapi/spec/bulk-registration-schema.ts
```diff
@@ -20,29 +22,48 @@ export const bulkRegistrationSchema = {
...
         },
         interval: {
+            description:
+                'How often (in seconds) does the application refresh its features',
             type: 'number',
+            example: 10,
         },
         started: {
+            description: 'When did the application start',
```
thomasheartman, line 50:
```suggestion
            description: 'When the application started',
```

src/lib/openapi/spec/bulk-registration-schema.ts
```diff
@@ -20,29 +22,48 @@ export const bulkRegistrationSchema = {
...
+                'Which [strategies](https://docs.getunleash.io/reference/activation-strategies) are enabled in the application',
             type: 'array',
+            example: ['standard', 'gradualRollout'],
             items: {
                 type: 'string',
             },
         },
         sdkVersion: {
+            summary: 'Version identifier for the SDK used',
+            description: 'Typically <client>:<version>',
```
thomasheartman, line 65: Properties don't have a "summary" field. It's all description. How about this?

```suggestion
            summary: 'Version identifier for the SDK used. Typically in the format of "<client>:<version>"',
```

src/lib/openapi/spec/client-metrics-env-schema.ts
thomasheartman, line 1: Can we add a top-level description here?

src/lib/openapi/spec/client-metrics-env-schema.ts
```diff
@@ -9,28 +9,46 @@ export const clientMetricsEnvSchema = {
...
         },
         no: {
+            description: 'How many times the toggle evaluated to false',
             type: 'number',
+            example: 50,
         },
         variants: {
+            description: 'How many times each variant was returned',
             type: 'object',
             additionalProperties: {
```
thomasheartman, line 43: Should this be additional properties? This looks like a required property to me. The syntax for child objects is the same as for top-level objects.

src/lib/openapi/spec/token-string-list-schema.ts
thomasheartman, line 1: This schema should also have a top-level description

src/lib/openapi/spec/token-string-list-schema.ts
```diff
@@ -0,0 +1,23 @@
...
+    $id: '#/components/schemas/tokenStringListSchema',
+    type: 'object',
+    additionalProperties: false,
+    required: ['tokens'],
+    properties: {
+        tokens: {
+            description: 'Tokens used to authenticate against Unleash',
+            type: 'array',
+            items: { type: 'string' },
+        },
```
thomasheartman, line 14: Do you have an example of what this might look like?

src/lib/openapi/spec/token-string-list-schema.ts
```diff
@@ -0,0 +1,23 @@
...
+import { FromSchema } from 'json-schema-to-ts';
+
+export const tokenStringListSchema = {
+    $id: '#/components/schemas/tokenStringListSchema',
+    type: 'object',
+    additionalProperties: false,
+    required: ['tokens'],
+    properties: {
+        tokens: {
+            description: 'Tokens used to authenticate against Unleash',
```
thomasheartman, line 11: How are the tokens presented? Just as raw strings? This might be clearer with the example, but are admin, front-end, client tokens, etc all in the same list? (Unleash itself knows what kind of tokens they are, right?)

src/lib/openapi/spec/validate-edge-tokens-schema.ts
thomasheartman, line 1: Top-level schema description here too 💁🏼

src/lib/openapi/spec/validate-edge-tokens-schema.ts
```diff
@@ -9,12 +9,7 @@ export const validateEdgeTokensSchema = {
     properties: {
         tokens: {
             type: 'array',
-            items: {
-                anyOf: [
-                    { $ref: '#/components/schemas/edgeTokenSchema' },
-                    { type: 'string' },
-                ],
-            },
+            items: { $ref: '#/components/schemas/edgeTokenSchema' },
```
thomasheartman, line 12: Can we get an example and a description going for this property?

src/lib/routes/edge-api/index.ts
```diff
@@ -59,12 +63,15 @@ export default class EdgeController extends Controller {
...
+                    description:
+                        'Send a list of tokens, only returns the valid ones',
                     operationId: 'getValidTokens',
-                    requestBody: createRequestSchema(
-                        'validateEdgeTokensSchema',
-                    ),
+                    requestBody: createRequestSchema('tokenStringListSchema'),
                     responses: {
                         200: createResponseSchema('validateEdgeTokensSchema'),
+                        ...getStandardResponses(400),
```
thomasheartman, line 74: Can't this return 401/403? And what about 415?

src/lib/routes/edge-api/index.ts
```diff
@@ -78,23 +85,24 @@ export default class EdgeController extends Controller {
             middleware: [
                 this.openApiService.validPath({
                     tags: ['Edge'],
+                    summary:
+                        'Accepts batched metrics from Edge, for insertion into our metrics storage',
```
thomasheartman, line 89: This looks like something between a description and a summary (and there is no description).

Something like this?
```suggestion
                    summary:
                        'Send metrics from Edge',
                     description:
                        'This operation accepts batched metrics from Edge. Metrics will be inserted into Unleash's metrics storage',
                     
```

thomasheartman (changes requested):
Very nice work on this! 🏅  I've got a minor suggestions, and I'm missing a few things in general:

The schemas have no top-level description. This makes it harder to understand what they are for. (I realize this wasn't expressed clearly in the linear task. I've tried to make it clearer now)

Examples and descriptions of nested schemas. I don't know if OpenAPI is smart enough to generate the examples itself, but the descriptions are almost definitely needed. If you can check whether it generates decent examples, then we might be able to get away without it, but I suspect we won't.

src/lib/openapi/spec/client-metrics-env-schema.ts
chriswk: Good point. Yes. Will clarify

src/lib/openapi/spec/bulk-registration-schema.ts
chriswk: Agreed, will fix

src/lib/routes/edge-api/index.ts
chriswk: Can't return 401/403, doesn't require a authentication header, there's no auth happening here. 415 is possible

src/lib/routes/edge-api/index.ts
chriswk: Yeah, will update so the description is actually a valid string :)

src/lib/routes/edge-api/index.ts
chriswk: We don't do auth here, so no 401,403. But Thomas is right 415 is a possibility, we don't accept any other content types than application/json. Technically I guess we could yield 413 as well if you try to post more than 100kB of metrics at the time

chriswk:
Added 413 and 415 to the standard responses here as well. I think the 413 message was fine. I was a bit uncertain about the 415.

src/lib/openapi/spec/validate-edge-tokens-schema.ts
chriswk: Added in 6ac9cc0

src/lib/routes/edge-api/index.ts
chriswk: Updated with 6ac9cc0

src/lib/openapi/util/standard-responses.ts
```diff
@@ -25,12 +25,24 @@ const conflictResponse = {
...
 
+const tooBigResponse = {
+    description:
+        'The body POSTed was too large. By default we only accept bodies of 100kB or less',
+} as const;
+
+const wrongContentTypeResponse = {
+    description:
+        'The provided resource does not accept requests with this content-type',
+} as const;
```
thomasheartman, line 36: I'll look at the PR in more detail later, but for now: how about this? I snagged it from the work I'm doing on Addons: it also explains what you're doing wrong and how to fix it. 💁🏼

```suggestion
const unsupportedMediaTypeResponse = {
    description: `The operation does not support request payloads of the provided type. Please ensure that you're using one of the listed payload types and that you have specified the right content type in the "content-type" header.`,
} as const;
```

chriswk: I love it. Thanks :)

src/lib/routes/edge-api/index.ts
```diff
@@ -59,12 +63,15 @@ export default class EdgeController extends Controller {
             middleware: [
                 this.openApiService.validPath({
                     tags: ['Edge'],
+                    summary:
+                        'Filter a list of tokens returning only the valid ones',
+                    description:
+                        'Accepts a list of tokens, and returns those that are valid with the projects they can access',
```
thomasheartman, line 69: What does "valid with the projects they can access" mean? Is it "valid with the projects" or does it return the tokens with their projects?

But ooh, this one is tricky. What about something like this? Trying to be as clear and explicit as possible. But it might not be necessary to describe in the description **everything** that is included in the edge-token-schema 🤔 

```suggestion
                    summary:
                        'Check which tokens are valid',
                    description:
                        'This operation accepts a list of tokens to validate. Unleash will validate each token you provide. For each valid token you provide, Unleash will return the token along with its type and which projects it has access to.',
```

src/lib/routes/edge-api/index.ts
```diff
@@ -78,27 +85,28 @@ export default class EdgeController extends Controller {
             middleware: [
                 this.openApiService.validPath({
                     tags: ['Edge'],
+                    summary: 'Receive metrics from Edge',
```
thomasheartman, line 88: When you're calling this endpoint you want to **send** metrics, right? 

```suggestion
                    summary: 'Send metrics from Edge',
```

src/lib/openapi/spec/edge-token-schema.ts
```diff
@@ -6,19 +6,30 @@ export const edgeTokenSchema = {
...
         },
         type: {
+            description: `Unleash supports three different types of API tokens ([ADMIN](https://docs.getunleash.io/reference/api-tokens-and-client-keys#admin-tokens), [CLIENT](https://docs.getunleash.io/reference/api-tokens-and-client-keys#client-tokens), [FRONTEND](https://docs.getunleash.io/reference/api-tokens-and-client-keys#front-end-tokens)). They all have varying access, so when validating a token it's important to know what kind you're dealing with`,
             type: 'string',
             enum: Object.values(ApiTokenType),
+            example: 'client',
         },
         token: {
+            description:
+                '[Unleash API tokens](https://docs.getunleash.io/reference/api-tokens-and-client-keys) are comprised of three parts. <project(s)>:<environment>.randomcharacters',
```
thomasheartman, line 29: My question here is "what is a token"? And this is a token schema, right? So if you think of the schema as a variable, then you might have `token.type`, `token.projects`, and ... `token.token`. `token.token` sounds kind of awkward and is confusing. 

We can't change the name of the property, but would you have a different way to describe what this is? We've used `token.secret`, but I don't think that's very illuminating. Because in essence, **this** is the actual token, right? The type and the project access is metadata about the token in a way. 

Something like this?

```suggestion
            description:
                'The actual token value. [Unleash API tokens](https://docs.getunleash.io/reference/api-tokens-and-client-keys) are comprised of three parts. <project(s)>:<environment>.randomcharacters',
```

thomasheartman (commented):
Great updates! 😄 I've got a few more suggestions and questions, but this is looking really good now 💯

src/lib/openapi/spec/edge-token-schema.ts
```diff
@@ -6,19 +6,30 @@ export const edgeTokenSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['token', 'projects', 'type'],
+    description:
+        'A representation of a client token, limiting access to [CLIENT](https://docs.getunleash.io/reference/api-tokens-and-client-keys#client-tokens) (used by serverside SDKs) or [FRONTEND](https://docs.getunleash.io/reference/api-tokens-and-client-keys#front-end-tokens) (used by proxy SDKs)',
     properties: {
         projects: {
+            description:
+                'Since a token can have access to multiple projects, but that is represented with a `[]` in the token, this can be used to expand the []',
```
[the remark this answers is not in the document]
chriswk: A token has to have access a project. It's the tokenString that can have `[]` as its project. This means we have to fetch projects from the database. Technically `['*', 'a']` could be possible, but it would still mean the same as `['*']`. Since 'a' would be subsumed by the wildcard. I don't think I've seen that happen yet

src/lib/openapi/spec/validated-edge-tokens-schema.ts
```diff
@@ -0,0 +1,27 @@
...
+export const validatedEdgeTokensSchema = {
+    $id: '#/components/schemas/validatedEdgeTokensSchema',
+    type: 'object',
+    additionalProperties: false,
+    required: ['tokens'],
+    description: `A list of Unleash client tokens with what projects they can access included`,
+    properties: {
+        tokens: {
+            description:
+                'Includes which projects the client tokens can access as well as the token itself',
```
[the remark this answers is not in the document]
chriswk: It's not really directly connected to Edge, it's a generic way of validating an API token. It just so happens to be available under an edge endpoint.

src/lib/openapi/spec/edge-token-schema.ts
```diff
@@ -6,19 +6,30 @@ export const edgeTokenSchema = {
...
             type: 'array',
             items: {
                 type: 'string',
             },
+            example: ['developerexperience', 'enterprisegrowth'],
         },
         type: {
+            description: `Unleash supports three different types of API tokens ([ADMIN](https://docs.getunleash.io/reference/api-tokens-and-client-keys#admin-tokens), [CLIENT](https://docs.getunleash.io/reference/api-tokens-and-client-keys#client-tokens), [FRONTEND](https://docs.getunleash.io/reference/api-tokens-and-client-keys#front-end-tokens)). They all have varying access, so when validating a token it's important to know what kind you're dealing with`,
```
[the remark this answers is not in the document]
chriswk: Updated the description to include linking to the api-tokens chapter in the initial sentence.

src/lib/openapi/spec/validated-edge-tokens-schema.ts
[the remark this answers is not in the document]
chriswk: And the intended users are edge instances. However, it's in response to the queried tokenStringList.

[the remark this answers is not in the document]
thomasheartman:
```suggestion
    description: `A object containing a list of valid Unleash tokens.`,
    properties: {
        tokens: {
            description:
                'The list of Unleash token objects. Each object contains the token itself and some additional metadata.',
```

thomasheartman (approved):
🏆

PR #3503, opened 2023-04-12
docs: open api examples for segment schemas

gastonfournier:
## About the changes
Add additional documentation for segments

### DIFF
--- a/src/lib/openapi/spec/upsert-segment-schema.ts
+++ b/src/lib/openapi/spec/upsert-segment-schema.ts
@@ -16,13 +16,29 @@ export const upsertSegmentSchema = {
         project: {
             type: 'string',
             nullable: true,
+            description:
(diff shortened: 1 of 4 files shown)


src/lib/openapi/spec/upsert-segment-schema.ts
```diff
@@ -16,14 +16,30 @@ export const upsertSegmentSchema = {
...
+                'List of constraints that determine which users will be part of the segment',
             items: {
                 $ref: '#/components/schemas/constraintSchema',
             },
         },
     },
+    example: {
+        name: 'segment name',
+        description: 'segment description',
+        project: 'optional project name',
```
nunogois, line 34: I believe in most cases it's the same, but it could still be useful to be explicit.
```suggestion
        project: 'optional project id',
```

nunogois (approved):
LGTM

PR #3510, opened 2023-04-12
feat: add title to strategy

andreas-unleash:
Adds title column to strategies, feature_strategies and features_view in the db
Updates model/schemas
## About the changes

Closes # [1-855](https://linear.app/unleash/issue/1-855/allow-for-title-on-strategy-backend)


src/lib/openapi/spec/strategy-schema.ts
```diff
@@ -13,6 +13,9 @@ export const strategySchema = {
         'parameters',
     ],
     properties: {
+        title: {
```
kwasniew, line 16: I'd add descriptions and examples in this or next PR

src/lib/services/strategy-schema.ts
```diff
@@ -5,6 +5,7 @@ const strategySchema = joi
     .object()
     .keys({
         name: nameType,
+        title: joi.string().default('').optional(),
```
kwasniew, line 8: so do we want to persist empty string for new strategies? 

Also is we have optional and default isn't default enough?

kwasniew (approved):
A few minor comments inside

src/lib/services/strategy-schema.ts
andreas-unleash: Actually, I will make it the same as description here, thanks for catching

src/lib/openapi/spec/strategy-schema.ts
andreas-unleash: Good call, I will fix the schema since I am here

andreas-unleash: Good call.

sjaanus:
Can we update e2e tests also to check that title gets saved on creation and on update?

PR #3511, opened 2023-04-12
chore: test requirements for openapi

gastonfournier:
## About the changes
This enables us to validate the shape of our OpenAPI schemas by defining specific json-schema rules that will be evaluated against all our open API schemas.

Because we know there are things we need to improve, we've added a list of `knownExceptions`. When fixing some of the known exceptions the tests will force us to remove the exception from the list, that way contributing to reducing the number of violations to our own rules.

Co-authored-by: Mateusz Kwasniewski <kwasniewski.mateusz@gmail.com>
Co-authored-by: Thomas Heartman <thomas@getunleash.io>

### DIFF
--- a/src/lib/openapi/index.ts
+++ b/src/lib/openapi/index.ts
@@ -171,9 +171,9 @@ interface UnleashSchemas {
     [name: string]: SchemaWithMandatoryFields;
 }
 
-type OpenAPIV3DocumentWithServers = Omit<OpenAPIV3.Document, 'servers'> & {
+interface OpenAPIV3DocumentWithServers extends OpenAPIV3.Document {
     servers: OpenAPIV3.ServerObject[];
-};
+}
 
 // All schemas in `openapi/spec` should be listed here.
 export const schemas: UnleashSchemas = {
--- a/src/lib/openapi/index.ts
+++ b/src/lib/openapi/index.ts
@@ -147,8 +147,36 @@ import { clientMetricsEnvSchema } from './spec/client-metrics-env-schema';
 import { updateTagsSchema } from './spec/update-tags-schema';
 import { batchStaleSchema } from './spec/batch-stale-schema';
 
+// Schemas must have an $id property on the form "#/components/schemas/mySchema".
+export type SchemaId = typeof schemas[keyof typeof schemas]['$id'];
+
+// Schemas must list all their $refs in `components`, including nested schemas.
+export type SchemaRef = typeof schemas[keyof typeof schemas]['components'];
+
+// JSON schema properties that should not be included in the OpenAPI spec.
+export interface JsonSchemaProps {
+    $id: string;
+    components: object;
+}
+
+type SchemaWithMandatoryFields = Partial<
+    Omit<
+        OpenAPIV3.ReferenceObject | OpenAPIV3.SchemaObject,
+        '$id' | 'components'
+    >
+> &
+    JsonSchemaProps;
+
+interface UnleashSchemas {
+    [name: string]: SchemaWithMandatoryFields;
+}
+
+type OpenAPIV3DocumentWithServers = Omit<OpenAPIV3.Document, 'servers'> & {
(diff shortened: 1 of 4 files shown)


src/lib/openapi/spec/meta-schema.test.ts
```diff
@@ -0,0 +1,33 @@
...
+            countErrors += validateMetaSchema.errors?.length ?? 0;
+            // expect(validateMetaSchema.errors).toBeNull(); // commented until we fix all schemas
+            if (validateMetaSchema.errors?.length) {
+                // this function outputs instead of failing the test
+                console.error(
+                    `${name} has the following errors: ${JSON.stringify(
+                        validateMetaSchema.errors,
+                    )}`,
+                );
+            }
```
gastonfournier, line 29: This is a very naive workaround to avoid failing the tests on line 21

thomasheartman:
Oh, this is pretty cool! What does it check? Operation descriptions (endpoints) or request/response payloads?

src/lib/openapi/spec/meta-schema.test.ts
```diff
@@ -0,0 +1,33 @@
...
+        description: { type: 'string' },
+    },
+    required: ['description'],
+};
+
+describe('OpenAPI schemas meta validation', () => {
+    let countErrors = 0;
+    Object.entries(schemas).forEach(([name, schema]) => {
+        it(`${name} should be valid`, () => {
+            const validateMetaSchema = ajv.compile(openApiMetaSchema);
```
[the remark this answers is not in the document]
gastonfournier: I thought about it, but apparently, the way `ajv` works (and correct me if I'm wrong because this is just speculation), is by storing some state inside the `validateMetaSchema` function/object/hybrid? because after you use it as a function on a schema (`validateMetaSchema(schema)`) you use it as an object to check if the previous usage as a function raised an error (`validateMetaSchema.errors?.length`). 

Therefore, I assume it's storing the evaluation result in the function. Because of that, I should either recreate the function/object/hybrid by recompiling it or maybe clean up the errors list (but I don't know if there's more state in that object that needs to be cleaned up).

gastonfournier:
> Oh, this is pretty cool! What does it check? Operation descriptions (endpoints) or request/response payloads?

Ir checks all schemas, and this question is exactly why I wanted to check this with you :+1: I've updated the description with the whole list of schemas that don't have a description.

If we need to target a subset of all schemas, this PR won't be useful unless we find a way of identifying which "schema types" need a description.

src/lib/openapi/spec/meta-schema.test.ts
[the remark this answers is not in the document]
gastonfournier: Ok! It's possible, maybe when the function is called it clears the state... :shrug: but it works so :heavy_check_mark:

src/lib/openapi/index.ts
```diff
@@ -328,7 +344,7 @@ export const createOpenApiSchema = ({
     unleashUrl,
     baseUriPath,
 }: Pick<IServerOption, 'unleashUrl' | 'baseUriPath'>): Omit<
-    OpenAPIV3.Document,
+    OpenAPIV3DocumentWithServers,
```
gastonfournier, line 347: Servers are always returned from this function, so adding this fixes some null pointer compilation errors

src/lib/openapi/index.ts
```diff
@@ -147,8 +147,36 @@ import { clientMetricsEnvSchema } from './spec/client-metrics-env-schema';
...
+    [name: string]: SchemaWithMandatoryFields;
+}
+
+interface OpenAPIV3DocumentWithServers extends OpenAPIV3.Document {
+    servers: OpenAPIV3.ServerObject[];
+}
+
 // All schemas in `openapi/spec` should be listed here.
-export const schemas = {
+export const schemas: UnleashSchemas = {
```
gastonfournier, line 179: This just forces some validations at compilation time, giving us a faster feedback loop

PR #3512, opened 2023-04-12
feat: combo component (virtualization)

nunogois:
Alternative proposal to @gastonfournier https://github.com/Unleash/unleash/pull/3473

Won't be able to prioritize this currently, and it also doesn't seem to be part of our current priorities, but maybe someone else can pick this up in the meantime.

Adds a drop-in replacement component with virtualization capabilities called `Combo`.

Still a bit buggy sometimes. Most of the complexity comes from handling edge cases from the original `Autocomplete`.


frontend/package.json
```diff
@@ -96,6 +98,7 @@
     "react-table": "7.8.0",
     "react-test-renderer": "17.0.2",
     "react-timeago": "7.1.0",
+    "react-virtuoso": "^4.2.1",
```
nunogois, line 101: Only dependency we're adding is [react-virtuoso](https://github.com/petyosi/react-virtuoso)

frontend/src/component/admin/groups/GroupForm/GroupFormUsersSelect/GroupFormUsersSelect.tsx
```diff
@@ -101,7 +102,7 @@ export const GroupFormUsersSelect: VFC<IGroupFormUsersSelectProps> = ({
 
     return (
         <StyledGroupFormUsersSelect>
-            <Autocomplete
+            <Combo
```
nunogois, line 105: Nice drop-in replacement should make us happy coders and provide a DRYer approach 🙂

frontend/src/component/common/Combo/Combo.tsx
```diff
@@ -0,0 +1,150 @@
...
+    FreeSolo extends boolean | undefined = false
+>({
+    virtualized = 'auto',
+    ...props
+}: IComboProps<T, Multiple, DisableClearable, FreeSolo>) => {
+    if (
+        virtualized === 'never' ||
+        (virtualized === 'auto' && props.options.length < VIRTUALIZED_THRESHOLD)
+    )
+        return <Autocomplete {...props} />;
```
nunogois, line 43: Use plain old `Autocomplete` if we don't match the necessary conditions for virtualization.

frontend/src/component/common/Combo/Combo.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    return (
+        <Autocomplete
+            {...props}
+            disableListWrap
+            ListboxComponent={ListboxComponent}
+        />
```
nunogois, line 50: Otherwise we use a version with a custom `ListboxComponent` which will allow us to virtualize the list.

We can probably leverage something here to simplify the `ListboxComponent` event handlers (instead of adding to window) and interactions below. I'm thinking maybe adding a ref to the `Autocomplete` and passing it down to the `ListboxComponent`, but needs exploring.

frontend/src/component/common/Combo/Combo.tsx
```diff
@@ -0,0 +1,150 @@
...
+                window.addEventListener('click', selectItem);
+                window.addEventListener('keypress', scrollUp);
+            } else {
+                window.removeEventListener('keydown', keyDownCallback);
+                window.removeEventListener('click', selectItem);
+                window.removeEventListener('keypress', scrollUp);
+            }
+        },
+        [keyDownCallback]
+    );
```
nunogois, line 128: Most of the lines above are just handling specific edge cases to try to play nicely with the `Autocomplete` component.

Other than that, the ones we really need are:

```TS
const elements = (props.children || []) as JSX.Element[];

const groups = elements.map(({ props }) => {
    const [header, list] = props.children;
    const items = list.props.children;
    return {
        total: items.length,
        header,
        items,
    };
});

const groupCounts = groups.map(({ total }) => total);
```

frontend/src/component/common/Combo/Combo.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    return (
+        <StyledDiv ref={ref} {...props}>
+            <GroupedVirtuoso
+                ref={virtuosoRef}
+                scrollerRef={scrollerRef}
+                increaseViewportBy={{ top: 100, bottom: 100 }}
```
nunogois, line 135: This improves keyboard navigation a bit by having some more items pre-rendered out of view.

frontend/src/component/common/Combo/Combo.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { GroupedVirtuoso, GroupedVirtuosoHandle } from 'react-virtuoso';
+
+const VIRTUALIZED_THRESHOLD = 100;
+
+const StyledDiv = styled('div')(({ theme }) => ({
+    'div[data-item-group-index]': {
+        '.MuiAutocomplete-option': {
+            paddingLeft: theme.spacing(3),
+        },
+    },
```
nunogois, line 18: Tried adding `overflow: 'hidden'` or `padding: 0` here to improve UI/UX and avoid some annoying extra scrolling, but somehow it made scrolling the dropdown more sluggish? Maybe it's a compromise we want, but needs more exploring.

frontend/package.json
gastonfournier: I just found https://github.com/bvaughn/react-virtualized-select which is something else we can look into

PR #3515, opened 2023-04-13
fix: remove e2e tests

andreas-unleash:
After discussions around e2e tests:
- Became clear that the intention is to test core functionality only

Removing 2 test files:
- notifications.spec.ts : because all PRs are run against the same heroku backend - no guarantee that more notifications will not happen as other tests run - so the test is flaky
- settings.spec.ts: depends on instance being enterprise and a flag on  
## About the changes

Closes #


andreas-unleash:
I was the one that created both those files - without knowing the original intention or implementation.  There is a thread in Engineering with @gastonfournier when we investigated

PR #3519, opened 2023-04-13
chore: amend open api tags descriptions and examples

gastonfournier:
## About the changes
Adds descriptions and examples to tag schemas


chriswk (approved):
LGTM

src/lib/openapi/spec/feature-tag-schema.ts
```diff
@@ -3,23 +3,36 @@ import { FromSchema } from 'json-schema-to-ts';
...
         tagValue: {
             type: 'string',
+            example: 'my-tag',
+            description: 'The value of the tag',
         },
         type: {
+            deprecated: true,
             type: 'string',
+            description:
+                'This field is deprecated and currently unused, use tagType instead',
```
thomasheartman, line 29: Wait: this is a response schema, right? 

If so, maybe something like this?

```suggestion
            description:
                'The [type](https://docs.getunleash.io/reference/tags#tag-types tag types) of the tag. This property is deprecated and will be removed in a future version of Unleash. Superseded by the `tagType` property.',,
            example: 'simple',
```

src/lib/openapi/spec/feature-tag-schema.ts
```diff
@@ -3,23 +3,36 @@ import { FromSchema } from 'json-schema-to-ts';
...
         },
         value: {
+            deprecated: true,
             type: 'string',
+            description:
+                'This field is deprecated and currently unused, use tagValue instead',
```
thomasheartman, line 35:
```suggestion
Wait: this is a response schema, right? 

```suggestion
            description:
                'The value of the tag. This property is deprecated and will be removed in a future version of Unleash. Superseded by the `tagValue` property.',
            example: 'my-tag-value',
```
```

File: src/lib/openapi/spec/tag-schema.ts [hunk 3]
```diff
@@ -5,25 +5,26 @@ export const TAG_MAX_LENGTH = 50;
 export const tagSchema = {
     $id: '#/components/schemas/tagSchema',
     type: 'object',
+    description: 'Representation of a tag',
```
thomasheartman, line 8:
```suggestion
    description: 'Representation of a [tag](https://docs.getunleash.io/reference/tags)',
```

src/lib/openapi/spec/tag-schema.ts
```diff
@@ -5,25 +5,26 @@ export const TAG_MAX_LENGTH = 50;
...
             maxLength: TAG_MAX_LENGTH,
+            description: 'The value of the tag',
+            example: 'a-tag-value',
         },
         type: {
             type: 'string',
             minLength: TAG_MIN_LENGTH,
             maxLength: TAG_MAX_LENGTH,
             default: 'simple',
+            description: 'The type of the tag',
```
thomasheartman, line 24:
```suggestion
            description: 'The [type](https://docs.getunleash.io/reference/tags#tag-types) of the tag',
```

src/lib/openapi/spec/update-tags-schema.ts
```diff
@@ -4,17 +4,20 @@ import { tagSchema } from './tag-schema';
 export const updateTagsSchema = {
     $id: '#/components/schemas/updateTagsSchema',
     type: 'object',
+    description: 'Represents a set of changes to tags of a feature.',
```
thomasheartman, line 7:
```suggestion
    description: 'Represents a set of changes to a feature's tags, such as adding or removing tags.',
```

thomasheartman (commented):
I think this is great 🙌🏼  Just a few tiny suggestions I've added. I know it's been merged already, but if they can be easily applied to a new set of changes, then we can look at it. if not, that's fine too. They're pretty trivial here.

Also: this made me realize how ... bare-bones the tags documentation is 🙈

src/lib/openapi/spec/feature-tag-schema.ts
gastonfournier: Sure, I believe this was an oversight on the past, I couldn't find where this is being used and it's currently returned as `undefined` (i.e. field is not present in the response). I tested the APIs and none of the APIs using this schema returned these fields

PR #3521, opened 2023-04-14
feat: visualize adding a default strategy

kwasniew:
## About the changes

When CR is not applied we show a default strategy that will be added:
Screenshot 2023-04-14 at 08 59 53
The default one doesn't count into all the changes count since it's not something a user added.

After applying the default one goes away:

Screenshot 2023-04-14 at 09 02 05

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/ChangeRequest.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/ChangeRequest.tsx
@@ -49,7 +49,10 @@ export const ChangeRequest: VFC<IChangeRequestProps> = ({
                                     variant="body2"
                                     color="text.secondary"
                                 >
-                                    Default strategy will be added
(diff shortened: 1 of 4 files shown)


frontend/src/component/changeRequest/ChangeRequest/ChangeRequest.tsx
```diff
@@ -42,6 +42,25 @@ export const ChangeRequest: VFC<IChangeRequestProps> = ({
...
                         />
                     ))}
+                    {feature.defaultChange ? (
+                        <Change
+                            discard={
+                                <Typography
+                                    variant="body2"
+                                    color="text.secondary"
+                                >
+                                    {feature.defaultChange.action ===
```
sjaanus, line 52: Will the action be same for both?

sjaanus (approved):

PR #3523, opened 2023-04-14
chore: rule for description on properties

gastonfournier:
## About the changes
This rule defines that all properties should have a description

We also considered enforcing an example but we noticed that in some cases that's not needed.


thomasheartman (approved):
Love it! ❤️

PR #3528, opened 2023-04-14
Feat demo guide

nunogois:
https://linear.app/unleash/issue/2-915/implement-guide-steps-logic

https://user-images.githubusercontent.com/14320932/232099388-a8138b29-8256-4ed2-b8f4-f7607cf3ab9c.mp4


frontend/src/component/demo/Demo.tsx
```diff
@@ -0,0 +1,193 @@
...
+                            Join our community in Slack. This is just an example
+                            and not part of the final guide.
+                        </Typography>
+                    </>
+                ),
+                disableBeacon: true,
+            },
+        ],
+    },
+];
```
nunogois, line 141: These are just examples for now, I'm including the real steps as a separate PR.

frontend/src/component/demo/DemoTopics/DemoTopics.tsx
```diff
@@ -0,0 +1,210 @@
...
+    expanded: boolean;
+    setExpanded: React.Dispatch<React.SetStateAction<boolean>>;
+    steps: number[];
+    currentTopic: number;
+    setCurrentTopic: (topic: number) => void;
+    topics: ITutorialTopic[];
+}
+
+export const DemoTopics = ({
+    expanded,
```
kwasniew, line 144: I'm not sure how much this component will be reused in different context but usually my recommendation is to avoid passing props and use slots (ReadNode children) instead. Or composition over configuration as it's called. https://artursabirov.blog/composition-over-configuration-in-react/. E.g. in HTML you're not passing <select options=[props]> but <select><option><option></select>. This make for much more extensible APIs.

nunogois: Hi @kwasniew - From reading your comment I got the impression that maybe it was left unfinished. 
I don't foresee this component to be reused, at least not at this stage, and there's little drill down happening. The component currently exists only to abstract way some of the complexity and isolated logic of displaying the topics on the bottom left corner. Going for a composition over configuration refactor here may leak things to the parent component that it doesn't need / should care about.

Given this specific scenario, could you please share an example of what you have in mind? Maybe it would help me visualize it and better understand the pros vs cons here.

kwasniew: GH actions didn't display my full comment :)

I'd expect this type of API, but again if it's not reused then I'm ok with your solution:
```
<DemoTopics>
   <Topic selected/>
  <Topic/>
  <Topic/>
</DemoTopics>
```
BTW, The point of composition is not only to avoid prop drilling but to have extensible UI. E.g. If you wan to have enhanced topic you can do it without touching DemoTopics every single time.

```
<DemoTopics>
   <WithPicture><Topic selected/></WithPicture>
  <Topic/>
  <Topic/>
</DemoTopics>
```

gastonfournier (approved):
G!

PR #3529, opened 2023-04-17
V5 node 18 frontend

gastonfournier:
## About the changes
Follow up on https://github.com/Unleash/unleash/pull/3527 refactoring cyclic dependencies in frontend


gastonfournier:
https://github.com/Unleash/unleash/pull/3527 solves this one

PR #3530, opened 2023-04-17
Fix main after OpenAPI fixes

gastonfournier:
## About the changes
After fixing some OpenAPI descriptions: https://github.com/Unleash/unleash/commit/6cc51b7ba03da1fe4b37023ac36576e97999dc79 this test needs to be updated


chriswk (approved):
I notice you're removing some of our schemas from the rules test. Do we not want to test them?

gastonfournier:
> I notice you're removing some of our schemas from the rules test. Do we not want to test them?

No, just removing them as exceptions to the rule, because those are the ones improved in https://github.com/Unleash/unleash/commit/6cc51b7ba03da1fe4b37023ac36576e97999dc79

PR #3549, opened 2023-04-18
Fix openapi tests

gastonfournier:
## About the changes
Update tests due to fixes introduced in https://github.com/Unleash/unleash/commit/2da279b7fbda59f8d8199e4f4fdc2e86d0c1cea8

This PR removes the fixes from the exceptions' list


thomasheartman (approved):
Good catch! 💯

PR #3550, opened 2023-04-18
fix: gh actions that run on main

gastonfournier:
## About the changes
These also need to be migrated to node 18


chriswk (approved):
Nice catch :)

PR #3552, opened 2023-04-18
fix: missing node engine upgrade to 18

gastonfournier:
## About the changes
Updates frontend engine minimum version to node 18 and also improves some messages in the release GHA task


gardleopard (approved):
LGTM

PR #3553, opened 2023-04-18
chore: remove uses of type errors from user-facing code

thomasheartman:
This change removes uses of the `TypeError` type from user-facing code. Type errors are used by typescript when you provide it the wrong type. This is a valid concern. However, in the API, they're usually a signal that **we've** done something wrong rather than the user having done something wrong. As such, it makes more sense to return them as validation errors or bad request errors.

## Breaking changes

Note that because of the way we handle errors, some of these changes will be made visible to the end user, but only in the response body.

```ts
{ "name": "TypeError", "message": "Something is wrong", "isJoi": true }
```

will become

```ts
{ "name": "ValidationError", "message": "Something is wrong", "isJoi": true }
```

Technically, this could be considered a breaking change. However, as we're gearing up for v5, this might be a good time to merge that?

## A return to 500

This PR also makes TypeErrors a 500-type error again because they should never be caused by invalid data provided by the user

### DIFF
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -425,7 +425,11 @@ export default class ProjectService {
         const group = await this.groupService.getGroup(groupId);
         const project = await this.getProject(projectId);
         if (group.id == null)
-            throw new ValidationError('Unexpected empty group id');
+            throw new ValidationError(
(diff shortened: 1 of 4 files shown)


chriswk (approved):
Looks pretty good to me.

PR #3554, opened 2023-04-18
Test npm publish

gastonfournier:
## About the changes

https://github.blog/changelog/2022-10-24-npm-v9-0-0-released/ introduced a breaking change in the way they handle files inside package.json which caused some issues with the way we pack and distribute Unleash:

> npm pack now follows a strict order of operations when applying ignore rules. If a files array is present in the package.json, then rules in .gitignore and .npmignore files from the root will be ignored.

What we discovered is that when having a nested .gitignore (the one we have inside frontend), `npm publish` was taking that nested .gitignore into account (despite the fact that we also have a package.json with files inside the same folder). We tricked this by removing the `build` folder from `frontend/.gitignore` and instead adding it into the root `.gitignore` which is being ignored by `npm publish` following what's stated in the release note above. 

-----------------


gardleopard (approved):
LGTM:)

PR #3556, opened 2023-04-18
Strategy title frontend

Tymek:
## About the changes
Ability to add a title to a strategy
image

Closes https://linear.app/unleash/issue/1-858/allow-for-title-on-strategy-frontend

### Discussion questions

How do we track adoption? On every strategy creation/edit?


sjaanus:
Should the input width be 100% as in figma? And use same input component?

sjaanus (approved):
Other than UX tweaks, looks ok

Issue #3539, opened 2023-04-17
🟢  feat: Allow SSO Sync to Set Root Role


ivarconr:
### Describe the feature request

Currently, SSO syncing only allows accepting a group id to sync in Unleash. Users would also like to be able to set a root role for the user so that they can accept admin/editor/viewer roles from the SSO provider as well.

Availability: **Enterprise**

PR #3559, opened 2023-04-19
feat: root roles from groups

sighphyre:
# What

This allows the option setting a root role on a group (only Editor or Admin). Any user that enters that group will receive the permissions from that root role but will not have their base root role changed. This property is allowed to be undefined so that we have a little layer of backwards compatibility and the db migration shouldn't impact the code flow whether or not the feature is enabled.

# Relevant Decisions

- Users entering a group with a root role take on that role without needing to be added to a project directly
- On the backbone of the above, groups with root roles cannot be added directly to a project, this is to prevent confusion
- The base role of users is unaffected by entering a group with a root role - they will still reflect as Viewer/Editor/Admin in the user information screen, but they will have the permissions from the group root role

# Screens

Modifications to group add/edit screen

image

Displaying relevant properties on the group. Each group is tagged with a root role description if present. Groups with root roles will not display "not used".

image

Changes to add user/group to project - if a group has a root role it can't be added directly to a project

image


frontend/src/component/admin/groups/GroupForm/GroupForm.tsx
```diff
@@ -83,23 +105,47 @@ export const GroupForm: FC<IGroupForm> = ({
...
+    const { roles } = useUsers();
 
     const isGroupSyncingEnabled =
         (oidcSettings?.enabled && oidcSettings.enableGroupSyncing) ||
         (samlSettings?.enabled && samlSettings.enableGroupSyncing);
 
+    const groupRootRolesEnabled = Boolean(uiConfig.flags.groupRootRoles);
+
+    const roleIdToRole = (rootRoleId: number | null): IProjectRole | null => {
+        return (
```
sighphyre, line 132: This little dance is because the `StyledAutocompleteWrapper` does not tolerate `undefined` correctly. It very much wants a `null`

frontend/src/component/admin/groups/GroupForm/GroupForm.tsx
```diff
@@ -146,16 +192,50 @@ export const GroupForm: FC<IGroupForm> = ({
                     elseShow={() => (
                         <StyledDescriptionBlock>
                             <Box sx={{ display: 'flex' }}>
-                                You can enable SSO groups syncronization if
+                                You can enable SSO groups synchronization if
```
sighphyre, line 195: Spelling :)

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
```diff
@@ -143,7 +177,10 @@ export const GroupCard = ({
                                         arrow
                                         describeChild
                                     >
-                                        <Badge>Not used</Badge>
+                                        <ConditionallyRender
```
sighphyre, line 180: I'm not sure of the sanity of nesting `ConditionallyRender` but it does work

src/lib/db/group-store.ts
```diff
@@ -53,13 +55,15 @@ const rowToGroupUser = (row) => {
...
         createdBy: row.created_by,
+        rootRoleId: row.root_role_id,
     };
 };
 
 const groupToRow = (group: IStoreGroup) => ({
     name: group.name,
     description: group.description,
     mappings_sso: JSON.stringify(group.mappingsSSO),
+    root_role_id: group.rootRole || null,
```
sighphyre, line 66: This is a problem with our validation layer. We can't send `null` numbers without them being coerced to `0`. So we don't send the property from the front end and if we don't find it, it gets coerced to `null` so that when we update the group the root role can be removed. In practise this is only done through a `PUT` endpoint so it feels okay. Happy to be told I'm mad

src/lib/openapi/spec/group-schema.ts
```diff
@@ -24,6 +24,10 @@ export const groupSchema = {
                 type: 'string',
             },
         },
+        rootRole: {
+            type: 'number',
+            nullable: true,
+        },
```
chriswk, line 30: Boyscout: Adding description to the new field?

frontend/src/component/admin/groups/GroupForm/GroupForm.tsx
```diff
@@ -146,16 +192,50 @@ export const GroupForm: FC<IGroupForm> = ({
...
+                            <StyledInputDescription>
+                                <Box sx={{ display: 'flex' }}>
+                                    Do you want to associate a root role with
+                                    this group?
+                                    <HelpIcon tooltip="When you associate an Admin or Editor role with this group, users in this group will automatically inherit the role globally. Note that groups with a root role association cannot be assigned to projects." />
+                                </Box>
+                            </StyledInputDescription>
+                            <StyledAutocompleteWrapper>
+                                <Autocomplete
+                                    data-testid={'test'}
```
nunogois, line 218: Perhaps use something a bit more specific and explicit?
```suggestion
                                    data-testid="GROUP_ROOT_ROLE"
```

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
nunogois: Yep, should be fine 🙂

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
```diff
@@ -101,6 +113,28 @@ export const GroupCard = ({
                             />
                         </StyledHeaderActions>
                     </StyledTitleRow>
+                    <ConditionallyRender
+                        condition={!!group.rootRole}
```
nunogois, line 117:
```suggestion
                        condition={Boolean(group.rootRole)}
```

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
```diff
@@ -101,6 +113,28 @@ export const GroupCard = ({
...
+                        show={
+                            <>
```
nunogois, line 119: No need for fragment `<>` since it seems like you only have 1 child in this case: `<InfoBadgeDescription>`

frontend/src/component/admin/groups/GroupsList/GroupsList.tsx
```diff
@@ -134,6 +137,9 @@ export const GroupsList: VFC = () => {
                         <Grid key={group.id} item xs={12} md={6}>
                             <GroupCard
                                 group={group}
+                                rootRoles={roles.filter(
+                                    (role: IProjectRole) => role.type === 'root'
+                                )}
```
nunogois, line 142: You may want to extract stuff like this, even though this one is pretty simple.

```
// ...
const rootRoles = roles.filter(
    (role: IProjectRole) => role.type === 'root'
)

// ...

rootRoles={rootRoles}
```

frontend/src/component/admin/groups/GroupForm/GroupForm.tsx
nunogois: Yep, but like I mentioned in https://github.com/Unleash/unleash/pull/3559/files#r1171626519 I think if it's `undefined` you're already covering your bases, since it's going to be `null` here. Up to you if you'd like to try it out in case you feel like it's more straightforward.

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
```diff
@@ -255,6 +256,12 @@ export const ProjectAccessAssign = ({
         --data-raw '${JSON.stringify(payload, undefined, 2)}'`;
     };
 
+    const createRootGroupWarning = (group?: IGroup): string | undefined => {
+        if (group && !!group.rootRole) {
```
nunogois, line 260:
```suggestion
        if (group && Boolean(group.rootRole)) {
```

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
```diff
@@ -268,35 +275,45 @@ export const ProjectAccessAssign = ({
...
-                            <span>
-                                {optionUser?.name && optionUser?.username
-                                    ? optionUser?.username
-                                    : optionUser?.email}
-                            </span>
-                        </StyledUserOption>
-                    }
-                />
-            </li>
+            <Tooltip title={createRootGroupWarning(optionGroup)}>
```
nunogois, line 278: That's a pretty smart approach 👍

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
```diff
@@ -346,6 +363,14 @@ export const ProjectAccessAssign = ({
                                 disableCloseOnSelect
                                 disabled={edit}
                                 value={selectedOptions}
+                                getOptionDisabled={option => {
+                                    if (option.type === ENTITY_TYPE.GROUP) {
+                                        const optionGroup =
+                                            option.entity as IGroup;
+                                        return !!optionGroup.rootRole;
```
nunogois, line 370: Again, just a matter of preference, but maybe it's more readable and consistent with the rest of the codebase.
```suggestion
                                        return Boolean(optionGroup.rootRole);
```

src/lib/db/group-store.ts
nunogois: Yep, reminds me of discussions we had like https://unleash-internal.slack.com/archives/C048ELND3QD/p1678447540849379 - May be worth having a look at how we're handling `project` in segments to see if we can keep it consistent.

nunogois (approved):
Amazing job with this, left a few optional comments.

I'm a bit confused with the feature if I'm completely honest. If you're adding to the documentation I wouldn't mind having a look as well just to see if I get a better understanding of it.

It also makes me feel like we urgently need some kind of permission matrix or similar system to easily see all the permissions each user has access to. I imagine it can be very confusing for a user to have higher access but to see it on the users page as `Viewer` and not understanding exactly where that access is coming from.

frontend/src/component/admin/groups/GroupForm/GroupForm.tsx
sighphyre: I like!

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
sighphyre: Happy to take this one, is this a style thing or is there a sharp edge to using `!!` that I don't know about?

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
sighphyre: Ah! I asked in a previous comment. Yes makes sense, I'll accept both of these

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
sighphyre: Update: Answered elsewhere. So accepted!

src/lib/db/group-store.ts
sighphyre: Yeah, I remember you guys talking about this one. It's a very similar approach - I had a chat with Gaston before I took a stab at this. The subtle differences are mostly due string vs number things in the validation layer

frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
sighphyre: You're just saying that 'cos you helped me come up with it :P

frontend/src/component/admin/groups/GroupsList/GroupsList.tsx
sighphyre: Yes! Thank you, nice suggestion!

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
sighphyre: Oh my word, that finally makes sense to me. Thank you!

frontend/src/component/admin/groups/GroupForm/GroupForm.tsx
sighphyre: Yeah *sigh* I did try. It doesn't work as well as I'd like. The find returns `undefined` if it uh... doesn't find anything. It does work but it splashes a huge warning to the console. Not a fan of warnings there since they tend to hide things we really should care about

frontend/src/component/admin/groups/hooks/useGroupForm.ts
```diff
@@ -24,6 +26,7 @@ export const useGroupForm = (
             users: users.map(({ id }) => ({
                 user: { id },
             })),
+            rootRole: rootRole || undefined,
```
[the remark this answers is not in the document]
sighphyre: This is coercing the outputs of the MUI Autocomplete., which likes null. That should work fine except that our validation layer doesn't handle undefined nicely. That should also be fine except that our DB layer won't update properties that are undefined.

So this is less about the Types expect and more about what's _actually_ happening in practice

I would sacrifice an entire kingdom of goats at this point to have proper Optionals

src/lib/openapi/spec/group-schema.ts
sighphyre: Good call! Done!

sighphyre:
> Amazing job with this, left a few optional comments.
> 
> I'm a bit confused with the feature if I'm completely honest. If you're adding to the documentation I wouldn't mind having a look as well just to see if I get a better understanding of it.
> 
> It also makes me feel like we urgently need some kind of permission matrix or similar system to easily see all the permissions each user has access to. I imagine it can be very confusing for a user to have higher access but to see it on the users page as `Viewer` and not understanding exactly where that access is coming from.

Appreciate the feedback sir! Docs are en route, but haven't been done yet. Happy to chat before then though because I do agree, we have a ton of ways that a user can have a permissions modified and it's getting hella confusing at this point. I think we need these changes to serve some of our users with large sets of users that are painful to manage but this is definitely adding fuel to an already burning fire. It'd be nice if we head that off at the pass before we start getting bombarded with help requests because people don't understand why permissions are behaving the way they are

sighphyre:
Related to #3539

PR #3560, opened 2023-04-19
fix: return 404 when gettings tags for a non existing feature

chriswk:
From the discussion here https://github.com/Unleash/unleash/pull/3496#discussion_r1163745860

This PR checks the existence of the feature before trying to get tags for the feature. Doing so by stealing the exists method from the feature store, since that's what we need to know exists, and avoiding having the feature store as a dependency to the featureTagStore seemed reasonable.

### DIFF
--- a/src/lib/db/feature-tag-store.ts
+++ b/src/lib/db/feature-tag-store.ts
@@ -9,6 +9,7 @@ import {
     IFeatureTagStore,
 } from '../types/stores/feature-tag-store';
 import { Db } from './db';
+import NotFoundError from '../error/notfound-error';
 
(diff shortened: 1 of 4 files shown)


chriswk:
A feature might exist, but not have tags, if so, the correct answer is an empty list, not a NotFoundError.
An exists call on the PK column is fast, and you'll also reuse the connection. I'll have a look to see if I could get it down to single query.

gastonfournier (approved):
As we discussed, there's no other way of doing this because we're querying the relationship table and it may return empty even if the feature does not exist. The only way of checking if it exists is by adding an additional query.

PR #3562, opened 2023-04-19
chore: update node-version file

gastonfournier:
## About the changes
There are a few version managers for node.js respecting .node-version file so it makes sense to also update it


.node-version
```diff
@@ -1 +1 @@
-14.21.3
+18.7.0
```
chriswk, line 1: Should probably be 18.15
```suggestion
18.15.0
```

chriswk (changes requested):
Change to most recent 18 version :)

PR #3563, opened 2023-04-19
fix: override built in stickiness options

andreas-unleash:
Created migration for updating context fields
Context fields with stickiness are now the source of truth for stickiness options.  

## About the changes
`userId` context field is updated to have stickiness on by default
`sessionId` context field is added with stickiness on by default
If `default` or `random` are not part of the context field - we add them
This will allow for the option to disable them, by creating a eg. `random` context field with stickiness off

https://user-images.githubusercontent.com/104830839/233581147-9a514af4-1fdd-4c2e-b954-3cfa13b1a203.mov

Closes to # [1-794](https://linear.app/unleash/issue/1-794/reach-out-to-yousician-for-early-beta-test)


src/migrations/20230420211308-update-context-fields-add-sessionId.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+            INSERT INTO context_fields(name, description, sort_order, stickiness) VALUES('sessionId', 'Allows you to constrain on sessionId', 4, true);
+
+            UPDATE context_fields
+            SET stickiness = true
+                WHERE name LIKE 'userId' AND stickiness is null;
```
andreas-unleash, line 10: Added the `stickiness` AND clause so that we do not override existing setting

Tymek (approved):

PR #3565, opened 2023-04-19
chore: drop event hook

gastonfournier:
## About the changes
Ref: https://docs.getunleash.io/reference/deploy/configuring-unleash#further-customization
> **eventHook** (`function(event, data)`) - (_deprecated in Unleash 4.3_ in favor of the [Webhook addon](../addons/webhook.md)) If provided, this function will be invoked whenever a feature is mutated. The possible values for `event` are `'feature-created'`, `'feature-archived'` and `'feature-revived'`. The `data` argument contains information about the mutation. Its fields are `type` (string) - the event type (same as `event`); `createdBy` (string) - the user who performed the mutation; `data` - the contents of the change. The contents in `data` differs based on the event type; For `'feature-archived'` and `'feature-revived'`, the only field will be `name` - the name of the feature. For `'feature-created'` the data follows a schema defined in the code [here](https://github.com/Unleash/unleash/blob/7b7f0b84e8cddd5880dcf29c231672113224b9a7/src/lib/schema/feature-schema.ts#L77). See an [api here](/reference/api/legacy/unleash/admin/events).

Related to: https://github.com/Unleash/unleash/issues/1265


src/lib/util/format-base-uri.ts
```diff
@@ -1,14 +1,14 @@
-export const formatBaseUri = (input: string): string => {
+export const formatBaseUri = (input?: string): string => {
     if (!input) return '';
     const firstChar = input[0];
     const lastChar = input[input.length - 1];
 
     if (firstChar === '/' && lastChar === '/') {
-        return input.substr(0, input.length - 1);
+        return input.substring(0, input.length - 1);
```
gastonfournier, line 7: `substr` is deprecated, following advice from: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

src/lib/util/parseEnvVar.ts
```diff
@@ -1,4 +1,10 @@
-export function parseEnvVarNumber(envVar: string, defaultVal: number): number {
+export function parseEnvVarNumber(
+    envVar: string | undefined,
```
gastonfournier, line 2: Same solution as below: https://github.com/Unleash/unleash/pull/3565/files#diff-f6361816dec3aa605263e21913ebc719f23d6f62da6c270d42afdf7b41df0b3dR18

chriswk (approved):
LG

PR #3567, opened 2023-04-19
fix: seed setup and remove admin/features call

gastonfournier:
## About the changes
Seed setup fails because of compilation errors, we need to skip errors until we fix them. Also, this removes one usage of `/api/admin/features` that was used only for testing


chriswk (approved):
LGTM

PR #3568, opened 2023-04-19
fix: compilation issues from event-store

gastonfournier:
## About the changes
Reduce compilation errors on event-store


chriswk (approved):
LGTM

Issue #3537, opened 2023-04-17
🟢  feat:  Revamp Unleash Demo


ivarconr:
### Describe the feature request

Revamp Unleash demo to have an interactive guided tour to improve end-user experience in getting to know the product. Both Unleash and the demo website should be improved to achieve this.

### Background

Make it easier to understand how Unleash works and demonstrate use cases for new users.

willianantunes:
Hi @ivarconr! I develop [this project](https://github.com/willianantunes/tutorials/tree/a31d8657b04f8b0d69a2b919e7f9b65a9b2b83f3/2023/05/feature_toggle_unleash). You simply execute  `docker-compose up` and have three up-to-date projects with Python/Django, JavaScript/Next.js, and C#/Razor.

This may come in handy for you. If you open the link, check out the `README.md`.

ivarconr:
Super nice @willianantunes 🚀 
Want to write a blog about it? Check out https://github.com/Unleash/community-content

PR #3569, opened 2023-04-19
feat: implement demo guide steps

nunogois:
https://linear.app/unleash/issue/2-924/implement-the-guide-steps

While https://github.com/Unleash/unleash/pull/3528 implemented the basic guide logic with some example steps, this PR adds the guide steps as per the current demo, along with any necessary changes and features that were needed to accomplish this.

https://user-images.githubusercontent.com/14320932/233133252-a3790f11-ec5d-4ee3-952d-88348212bd3c.mp4


frontend/src/component/demo/demo-topics.tsx
```diff
@@ -0,0 +1,489 @@
...
+            {
+                target: 'div[data-testid="variant_list"] > div:last-of-type > div:last-of-type > button',
+                content: (
+                    <Typography variant="body2" color="text.secondary">
+                        Let's also add an override for our user.
+                    </Typography>
+                ),
+            },
+            {
+                target: 'div[data-testid="variant_list"] > div:last-of-type div[data-testid="context_field"]',
```
gastonfournier, line 433: These are sensitive to refactors and we don't have a good testing strategy to validate these continue to work. Can we use a more direct way of identifying the targets? What worries me is the `> div: last-of-type`. 

Something like: `'div[data-testid="variant_list"] div[data-testid="context_field"]:last'` indicating the last context_field inside variant_list would be better IMO. But it also means that these `data-testid`'s are not unique (they are repeated, right?).

Alternatively, if we could **identify** the last context_field in variants' list, something like `[data-testid="variants_list_last_context_field"]`, this would make it less error prone, but it requires some additional code to only add this to the last element we want to target. But at the same time this can open up the possibility of adding tests to the page to guarantee that the element exists.

gastonfournier (approved):
Approving because this looks fine, but I'd like to challenge the way we target elements. I don't consider it a blocker, so feel free to move on, create a task (if you consider the suggestion) or have a conversation to prove me wrong ;)

frontend/src/component/demo/demo-topics.tsx
nunogois: Yep, had the same thought, I think these are valid concerns. I think I'll have another look through the steps and try to use more explicit selectors whenever possible.

However, if you think about it, the example of `> div:last-of-type` may not be so bad. We're looking at the `variants-list` and selecting the last element of the list (the last variant we just added). I don't see another trivial way of doing this that would be much more explicit...

I also prefer to adapt the demo guide to the product whenever possible, instead of the other way around.

Worst case scenario, if we eventually end up with scenarios where the user gets stuck in one of the steps, they can always escape from it.

nunogois: Tried addressing this with more explicit selectors in https://github.com/Unleash/unleash/pull/3569/commits/6f1258b337b43a45cd717ef766da492e29ee8730 but it's something to keep in mind for the future.

PR #3571, opened 2023-04-20
Docs: openapi schema specifications for Projects tag

daveleek:
## About the changes

Improves the openapi schema specifications for the schemas belonging to the "Projects" tag.
Expected error codes/http statues, descriptions, and example data


src/lib/openapi/spec/project-environment-schema.ts
```diff
@@ -3,14 +3,21 @@ import { FromSchema } from 'json-schema-to-ts';
 export const projectEnvironmentSchema = {
     $id: '#/components/schemas/projectEnvironmentSchema',
     type: 'object',
-    additionalProperties: false,
```
daveleek, line 6: I could only find this type being used as a post body, but I'll have an extra look

src/lib/routes/admin-api/project/environments.ts
```diff
@@ -42,10 +45,16 @@ export default class EnvironmentsController extends Controller {
...
+                    summary: 'Add an environment to a project.',
+                    description:
+                        'This endpoint adds the provided environment to the specified project, with optional support for enabling and disabling change requests for the environment and project.',
                     requestBody: createRequestSchema(
                         'projectEnvironmentSchema',
                     ),
-                    responses: { 200: emptyResponse },
+                    responses: {
+                        200: emptyResponse,
+                        ...getStandardResponses(401, 403, 409),
```
daveleek, line 56: This endpoint interestingly enough returns a 500 (with the underlying error being a foreign key constraint violation on a missing key) when sending it a junk-project

src/lib/routes/admin-api/project/environments.ts
```diff
@@ -60,7 +69,13 @@ export default class EnvironmentsController extends Controller {
                 openApiService.validPath({
                     tags: ['Projects'],
                     operationId: 'removeEnvironmentFromProject',
-                    responses: { 200: emptyResponse },
+                    summary: 'Remove a environment to a project.',
+                    description:
+                        'This endpoint removes the provided environment to the specified project, with optional support for enabling and disabling change requests for the environment and project.',
+                    responses: {
+                        200: emptyResponse,
+                        ...getStandardResponses(400, 401, 403),
```
daveleek, line 77: Returns a 400 when trying to delete an environment from a junk project

src/lib/openapi/spec/health-overview-schema.ts
```diff
@@ -14,16 +14,23 @@ export const healthOverviewSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['version', 'name'],
+    description: `An overview of a projects and it's health as described in the documentation on [technical debt](https://docs.getunleash.io/reference/technical-debt)`,
```
chriswk, line 17:
```suggestion
    description: `An overview of a project and its health as described in the documentation on [technical debt](https://docs.getunleash.io/reference/technical-debt)`,
```

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -81,8 +82,12 @@ export class ProjectApiTokenController extends Controller {
                 openApiService.validPath({
                     tags: ['Projects'],
                     operationId: 'getProjectApiTokens',
+                    summary: 'Get api tokens for project.',
+                    description:
+                        'Returns the [project API tokens](https://docs.getunleash.io/how-to/how-to-create-project-api-tokens) that have been created for this project.',
                     responses: {
                         200: createResponseSchema('apiTokensSchema'),
+                        ...getStandardResponses(401, 403),
```
chriswk, line 90: Wouldn't 404 also be a possibility if the project doesn't exist?

src/lib/openapi/spec/health-overview-schema.ts
daveleek: Whoops. Cheers!

src/lib/routes/admin-api/project/api-token.ts
daveleek: Yeah I'm gonna have a look and see, it should be, but I'm not entirely sure. Some of the others here only return a 400

daveleek: This appears to just return an empty 200 which I find a bit odd

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -98,9 +103,13 @@ export class ProjectApiTokenController extends Controller {
                     tags: ['Projects'],
                     operationId: 'createProjectApiToken',
                     requestBody: createRequestSchema('createApiTokenSchema'),
+                    summary: 'Create a project API token.',
+                    description:
+                        'Endpoint that allows creation of [project API tokens](https://docs.getunleash.io/how-to/how-to-create-project-api-tokens) for the specified project.',
                     responses: {
                         201: resourceCreatedResponseSchema('apiTokenSchema'),
                         400: emptyResponse,
+                        ...getStandardResponses(401, 403),
```
daveleek, line 112: This also returns a 400, so I'll add that

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -116,8 +125,12 @@ export class ProjectApiTokenController extends Controller {
                 openApiService.validPath({
                     tags: ['Projects'],
                     operationId: 'deleteProjectApiToken',
+                    summary: 'Delete a project API token.',
+                    description:
+                        'Endpoint that allows deletion of a project API token.',
                     responses: {
                         200: emptyResponse,
+                        ...getStandardResponses(401, 403),
```
daveleek, line 133: This one actually seems to hang in tests when given a wrong project/token

chriswk (approved):
I am going to approve this, but I think we'd like some feedback from @thomasheartman as well before merging

src/lib/routes/admin-api/project/api-token.ts
daveleek: Never mind, it already had 400

src/lib/openapi/spec/api-token-schema.ts
```diff
@@ -6,47 +6,72 @@ export const apiTokenSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['username', 'type'],
+    description: 'An overview of an API token.',
```
thomasheartman, line 9: Maybe we could link to the docs here?

```suggestion
    description: 'An overview of an [Unleash API token](https://docs.getunleash.io/reference/api-tokens-and-client-keys).',
```

src/lib/openapi/spec/api-token-schema.ts
```diff
@@ -6,47 +6,72 @@ export const apiTokenSchema = {
...
         type: {
             type: 'string',
             enum: Object.values(ApiTokenType),
+            description: 'The type of API token',
+            example: 'client',
         },
         environment: {
             type: 'string',
+            description: 'The environment the token has access to.',
+            nullable: true,
```
thomasheartman, line 30: Can `environment` be both `null` and `undefined` (not present)?  When is it `null`? And when is it undefined?

src/lib/openapi/spec/api-token-schema.ts
```diff
@@ -6,47 +6,72 @@ export const apiTokenSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['username', 'type'],
```
thomasheartman, line 8: Are these the only two properties that are guaranteed to be on the object?

src/lib/openapi/spec/api-token-schema.ts
```diff
@@ -6,47 +6,72 @@ export const apiTokenSchema = {
...
         },
         project: {
             type: 'string',
+            description: 'The project this token belongs to.',
+            example: 'developerexperience',
         },
         projects: {
             type: 'array',
+            description:
+                'The list of projects this token has access to. If the token has access to specific projects they will be listed here. If the token has access to all projects it will be represented as [`*`]',
```
thomasheartman, line 41: Just for my understanding: can this be an empty list (if the token has access to no projects)?

src/lib/openapi/spec/api-token-schema.ts
thomasheartman: This question goes for every property that's marked as nullable here.

src/lib/openapi/spec/api-token-schema.ts
```diff
@@ -6,47 +6,72 @@ export const apiTokenSchema = {
...
             nullable: true,
+            example: '2023-04-19T08:15:14.000Z',
+            description:
+                'When the token was last seen/used to authenticate with.',
         },
         alias: {
             type: 'string',
             nullable: true,
+            description:
+                'Not actively used. Present as a legacy format for identifying older tokens',
```
thomasheartman, line 73: What does it represent? I don't understand what it is. It's fine that it's not used anymore, but I don't understand what it is.

src/lib/openapi/spec/health-overview-schema.ts
```diff
@@ -14,16 +14,23 @@ export const healthOverviewSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['version', 'name'],
```
thomasheartman, line 16: Like with the other schema: are these the only required properties? Can everything else be undefined?

src/lib/openapi/spec/health-overview-schema.ts
```diff
@@ -14,16 +14,23 @@ export const healthOverviewSchema = {
...
+    description: `An overview of a project and its health as described in the documentation on [technical debt](https://docs.getunleash.io/reference/technical-debt)`,
     properties: {
         version: {
             type: 'number',
+            description: 'The project overview version.',
```
thomasheartman, line 21: For clarity: is this the version of the format used for this object? If so, maybe we could make it a little clearer. But also: wouldn't changing the format require a new schema? If so, maybe this should be:
 
```suggestion
            description: 'The project overview version.',
            enum: [1],
```

src/lib/openapi/spec/health-overview-schema.ts
```diff
@@ -14,16 +14,23 @@ export const healthOverviewSchema = {
...
+            example: 1,
         },
         name: {
             type: 'string',
+            description: 'Project name',
```
thomasheartman, line 26:
```suggestion
            description: 'The project's name',
```

src/lib/openapi/spec/health-overview-schema.ts
```diff
@@ -40,29 +47,44 @@ export const healthOverviewSchema = {
         },
         members: {
             type: 'number',
```
thomasheartman, line 49: My guess is this is a non-zero natural number, right? A project **must** have an owner?
```suggestion
            type: 'integer',
            minimum: 1,
```

thomasheartman (commented):
Consider reviewing all the `type: 'number'` instances in these schemas and see if they should be `type: 'integer'` instead. Also, if it makes sense, set a `minimum` or `maximum` value.

src/lib/routes/admin-api/project/api-token.ts
thomasheartman: Can we add 400 as part of the standard response? Maybe?

src/lib/routes/admin-api/project/environments.ts
thomasheartman: Yeah. I've seen this a lot with other endpoints too. My suggestion is to add a test for it. We shouldn't really return 500s on bad data, I think.

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -116,8 +125,12 @@ export class ProjectApiTokenController extends Controller {
                 openApiService.validPath({
                     tags: ['Projects'],
                     operationId: 'deleteProjectApiToken',
+                    summary: 'Delete a project API token.',
+                    description:
+                        'Endpoint that allows deletion of a project API token.',
```
thomasheartman, line 130: Do you specify the token in the URL? And if it doesn't exist, is it a 200? Or a 404?

```suggestion
                        'This operation deletes the API token specified in the request URL. If the token doesn't exist, returns an OK response (status code 200).',
```

src/lib/routes/admin-api/project/environments.ts
thomasheartman: Also, shouldn't there be a 404 here if the project doesn't exist?

src/test/e2e/api/admin/project/project.api.tokens.e2e.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    return app.request
+        .post('/api/admin/projects/wrong/api-tokens')
+        .send({
+            username: 'default-client',
+            type: 'client',
+            projects: ['wrong'],
+            environment: 'default',
+        })
+        .set('Content-Type', 'application/json')
+        .expect(400);
```
thomasheartman, line 81: Ooh, should this be a 400 or 404? I don't have the answer, but it feels like if the URL is wrong, then it's a 404, but if the request payload is bad, then that's a 400 to me. What do **you** think?

thomasheartman (commented):
This looks pretty good! I've got some suggestions on using `integer` and using certain other constraints for data types, but nothing major.

Regarding documenting operations: It seems like many of your descriptions are just longer rewrites of the summaries. As a general comment: would it be possible to elaborate more in the description? I like to think that the end user should be able to see the docs for an operation and get all the context they need, including links to external docs, what kind of errors can happen, and anything else that's useful, such as what forms the data must take etc. This isn't blocking here, but worth a thought, I think.

src/lib/routes/admin-api/project/environments.ts
daveleek: That's a good question! I'll add a task for this, maybe we should discuss that there? Seems like there should be some kind of decision/guideline on this kind of use case, doubt this would be the only instance of it?

src/lib/routes/admin-api/project/api-token.ts
daveleek: Yeah fair enough. I can add a 404 standard response now, and set up a task for implementing it!

src/lib/routes/admin-api/project/api-token.ts
daveleek: I don't see why not

src/lib/routes/admin-api/project/api-token.ts
daveleek: I'm unsure if that's just in my unit tests or if this is something that affects it during runtime as well. I'll add a task

src/lib/openapi/spec/project-environment-schema.ts
daveleek: Well, it was. it's not anymore. I'll re-add this

src/lib/openapi/spec/api-token-schema.ts
daveleek: Yeah absolutely!

src/lib/openapi/spec/api-token-schema.ts
daveleek: No - for project api-tokens it has to be exactly 1 - the specific project it is tied to.
For other api-tokens it gets the ALL parameter if project/list of projects isn't provided.

This should perhaps be reflected in the documentation.

src/lib/openapi/spec/api-token-schema.ts
daveleek: As far as I can gather it was a type of name/public id for the token, but I have not found a definitive answer on that

src/lib/openapi/spec/api-token-schema.ts
thomasheartman: Right. Could you put that in the description? That is the bit of docs that is the closest to the actual API and is at least where I would look.

src/lib/openapi/spec/api-token-schema.ts
thomasheartman: Got it; thanks!

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -98,9 +103,13 @@ export class ProjectApiTokenController extends Controller {
                     tags: ['Projects'],
                     operationId: 'createProjectApiToken',
                     requestBody: createRequestSchema('createApiTokenSchema'),
+                    summary: 'Create a project API token.',
+                    description:
+                        'Endpoint that allows creation of [project API tokens](https://docs.getunleash.io/how-to/how-to-create-project-api-tokens) for the specified project.',
```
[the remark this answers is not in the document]
daveleek: Indeed! Changing!

src/test/e2e/api/admin/project/project.api.tokens.e2e.test.ts
daveleek: It feels like a 404 to me - I'll add a task to fix the endpoint

src/lib/openapi/spec/api-token-schema.ts
daveleek: In this particular case I can conclude after having checked the code that no - it's not actually nullable. It's occasionally stored as NULL in the DB - but will get translated to ALL (or '*') when converted to an object. I'll fix

src/lib/openapi/spec/health-overview-schema.ts
daveleek: I suspect you are correct, but I'll have a look

src/lib/openapi/spec/health-overview-schema.ts
thomasheartman: This is a good question. So if Unleash always returns only a single schema, then we wouldn't need to delete it, we could just evolve it between versions and update the version from `[1]` to `[2]`, for instance.

If we want to return multiple schemas, we'd have to keep them around. 

am I making sense? did that answer your question?

src/lib/openapi/spec/health-overview-schema.ts
daveleek: It defaults to being 0 - so it's not nullable, I am unsure if it's possible to completely abandon a project?

src/lib/openapi/spec/project-environment-schema.ts
```diff
@@ -3,14 +3,21 @@ import { FromSchema } from 'json-schema-to-ts';
...
     required: ['environment'],
     properties: {
         environment: {
             type: 'string',
+            description: 'The environment to add to the project',
+            example: 'development',
         },
         changeRequestsEnabled: {
             type: 'boolean',
+            nullable: true,
```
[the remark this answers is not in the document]
daveleek: Appears it can only be undefined. Fixing

src/lib/openapi/spec/health-overview-schema.ts
daveleek: The interesting part is that changing this to enum: [1] introduces build errors as it's being passed around as a 'number' in the domain

src/lib/openapi/spec/health-overview-schema.ts
thomasheartman: Got it. In that case, I'd go for `type: integer, minimum: 0`, I think.

src/lib/openapi/spec/api-token-schema.ts
```diff
@@ -24,57 +36,57 @@ export const apiTokenSchema = {
...
         },
         project: {
             type: 'string',
+            description: 'The project this token belongs to.',
+            example: 'developerexperience',
         },
         projects: {
             type: 'array',
+            description:
+                'The list of projects this token has access to. If the token has access to specific projects they will be listed here. If the token has access to all projects it will be represented as [`*`]',
```
thomasheartman, line 56: Shouldn't those quotes and brackets be swapped?

```suggestion
                'The list of projects this token has access to. If the token has access to specific projects they will be listed here. If the token has access to all projects it will be represented as `[*]`',
```

src/lib/openapi/spec/health-overview-schema.ts
```diff
@@ -41,30 +57,45 @@ export const healthOverviewSchema = {
...
             type: 'string',
             format: 'date-time',
             nullable: true,
+            description: 'When the project was last updated.',
+            example: '2023-04-19T08:15:14.000Z',
         },
         favorite: {
             type: 'boolean',
+            description:
+                'Indicates if the project has been favourited by the current user requesting the project health overview.',
```
thomasheartman, line 97: I think we've decided to go for American English, so I'll remove that and suggest a slight restructuring:

```suggestion
                'Indicates if the project has been marked as a favorite by the current user requesting the project health overview.',
```

thomasheartman (approved):
Super! A few minor suggestions, but it's all non-blocking.

Very well done! 👏🏼  Thanks for making the effort to work all the way through this marathon of a task 🏆

PR #3574, opened 2023-04-20
feat: add demo guide welcome dialog

nunogois:
https://linear.app/unleash/issue/2-923/add-initial-getting-started-dialog

Adds the welcome dialog with the static QR code pointing to our demo website. Also changes how the run logic works a bit, so that we have better control over it (only start demo steps once we click the start button, etc).

image


frontend/src/component/demo/Demo.tsx
```diff
@@ -18,34 +20,47 @@ const { value: storedProgress, setValue: setStoredProgress } =
 
 export const Demo = () => {
     const { uiConfig } = useUiConfig();
-    const [loaded, setLoaded] = useState(false);
+    const [welcomeOpen, setWelcomeOpen] = useState(
+        storedProgress.welcomeOpen ?? true
```
gastonfournier, line 24: Will this reset every time we deploy a new version? I don't remember how storedProgress is persisted, but I just thought asking the question

gastonfournier (approved):
LGTM!

frontend/src/component/demo/Demo.tsx
nunogois: It's persisted on local storage, I don't think deploying new versions of Unleash would affect it. 

This local storage set is also versioned `Tutorial:v1`, so if we later release a new version of the tutorial and want to reset it for end-users we just need to increase this version.

Users have the choice of re-entering the welcome screen at any point through the topic list by clicking "View demo link again" at the bottom of the list:
image
Clicking any of the topics at any point would start that topic as well.

PR #3575, opened 2023-04-20
chore: remove deprecated strategies and allow deprecate default

gastonfournier:
## About the changes
Improvements around strategies to steer users toward using [strategy constraints](https://docs.getunleash.io/reference/strategy-constraints) instead of [custom strategies](https://docs.getunleash.io/reference/custom-activation-strategies)

Related to #1265 

Important changes:
1.[ Ability to deprecate default strategy](https://github.com/Unleash/unleash/pull/3575/files#diff-3b49e566565b01580ff7a15f662d2cb824547fdfb3935d0a0baa7fae326d28d6L232-L235)
2. [Remove strategies that were deprecated in v4 only if they're not used](https://github.com/Unleash/unleash/pull/3575/files#diff-a04de7adf327c48d54b030bed70d9b1d62ff14df3efc668eb6897a210124ebe0R7-R10)
3. [Fresh Unleash installations will not include deprecated strategies from v4](https://github.com/Unleash/unleash/pull/3575/files#diff-46265b0d51725a5acbf251589b98a8970e45d3a1c6f3d7ea77bdcc3714e672d8L31-L78)
4. [Deprecate userWithId strategy](https://github.com/Unleash/unleash/pull/3575/files#diff-a04de7adf327c48d54b030bed70d9b1d62ff14df3efc668eb6897a210124ebe0R13) (note: if in use it will continue to work, it'll be not listed when selecting strategies for a feature toggle).
5. [Switch the order of strategies to make gradual rollout first and change the comment to suggest using gradual rollout instead of other existing strategies](https://github.com/Unleash/unleash/pull/3575/files#diff-a04de7adf327c48d54b030bed70d9b1d62ff14df3efc668eb6897a210124ebe0R16-R18)

### DIFF
--- a/src/migrations/20230420125500-v5-strategy-changes.js
+++ b/src/migrations/20230420125500-v5-strategy-changes.js
@@ -1,0 +1,72 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+        -- delete deprecated strategies still present in v4
+        delete from strategies 
+            where name in ('gradualRolloutUserId', 'gradualRolloutRandom', 'gradualRolloutSessionId') 
+            and deprecated 
+            and not exists (select * from feature_strategies where strategy_name = name limit 1);
+        
+        -- deprecate strategies on v5
+        update strategies set deprecated = true where name in ('userWithId');
+        
+        -- update strategy descriptions and sort order
+        update strategies set sort_order = 1, description = 'This strategy turns on / off for your entire userbase.' WHERE name = 'default';
+        update strategies set sort_order = 0 WHERE name = 'flexibleRollout';
+        update strategies set description = 'Prefer using "Gradual rollout" strategy with user id constraints. Enable the feature for a specific set of userIds.' WHERE name = 'userWithId';
+        `,
+        callback,
+    );
+};
+
+exports.down = function (db, callback) {
+    db.runSql(
+        `
+        -- restore deleted strategies
+        insert into strategies (name, description, parameters, deprecated, sort_order) values
+            ('gradualRolloutRandom', 'Randomly activate the feature toggle. No stickiness.', [
+                {
+                  "name": "percentage",
+                  "type": "percentage",
+                  "description": "",
+                  "required": false
+                }
+              ], true, 3),
+            ('gradualRolloutSessionId', 'Gradually activate feature toggle. Stickiness based on session id.', [
+                {
+                  "name": "percentage",
+                  "type": "percentage",
+                  "description": "",
+                  "required": false
+                },
+                {
+                  "name": "groupId",
+                  "type": "string",
+                  "description": "Used to define a activation groups, which allows you to correlate across feature toggles.",
+                  "required": true
+                }
+              ], true, 4),
+            ('gradualRolloutUserId', 'Gradually activate feature toggle for logged in users. Stickiness based on user id.', [
+                {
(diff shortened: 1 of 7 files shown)


src/migrations/20230420125500-v5-strategy-changes.js
```diff
@@ -0,0 +1,72 @@
...
+                  "required": false
+                },
+                {
+                  "name": "groupId",
+                  "type": "string",
+                  "description": "Used to define a activation groups, which allows you to correlate across feature toggles.",
+                  "required": true
+                }
+              ], true, 5);
+
```
gastonfournier, line 65: We don't want to reverse the [deprecation of userWithId](https://github.com/Unleash/unleash/pull/3575/files#diff-a04de7adf327c48d54b030bed70d9b1d62ff14df3efc668eb6897a210124ebe0R13) because the user can change that manually and there might be cases where the customer could have manually deprecated this strategy and we don't want to enable it back

src/migrations/20230420125500-v5-strategy-changes.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+        -- delete deprecated strategies still present in v4
+        delete from strategies 
+            where name in ('gradualRolloutUserId', 'gradualRolloutRandom', 'gradualRolloutSessionId') 
+            and deprecated 
+            and not exists (select * from feature_strategies where strategy_name = name limit 1);
```
gardleopard, line 10: This migration only deletes the strategies if they match the happy path. As we discussed there are two blockers here for deleting them:
1. the strategy is in use
2. the strategy is in use and it has been modified.

For 1. we can probably migrate the use into using gradualRollout with the same parameters as before.
For 2. we can change the name and make it clear that its handled by the user and not unleash anymore just like other custom strategies.

src/migrations/20230420125500-v5-strategy-changes.js
```diff
@@ -0,0 +1,72 @@
...
+        
+        -- deprecate strategies on v5
+        update strategies set deprecated = true where name in ('userWithId');
+        
+        -- update strategy descriptions and sort order
+        update strategies set sort_order = 1, description = 'This strategy turns on / off for your entire userbase. Prefer using "Gradual rollout" strategy (100%=on, 0%=off).' WHERE name = 'default';
```
chriswk, line 16:
```suggestion
        update strategies set sort_order = 1, description = 'This strategy turns the feature on / off for your entire userbase. Prefer using "Gradual rollout" strategy (100%=on, 0%=off).' WHERE name = 'default';
```

src/migrations/20230420125500-v5-strategy-changes.js
gastonfournier: This is a very good comment. Ideally, we'd like to get rid of them, but there might be unexpected side effects that make me worry.

> For 1. we can probably migrate the use into using gradualRollout with the same parameters as before.

Yes, but we need to merge 3 strategies into one, which might be doable in some situations, but still won't cover all cases. I'm also not that sure how stickiness will behave if we change the strategy, because even if the strategy implementation is the same, because of the name change we might reassign users to another treatment. So I rather keep it like this

chriswk (approved):
I like that we now unprotect the default strategy, it feels like a small enough footgun to trust our users with.

PR #3578, opened 2023-04-20
feat: add demo guide finish dialog

nunogois:
https://linear.app/unleash/issue/2-925/add-final-tutorial-complete-dialog

Adds the finish dialog, similar to https://github.com/Unleash/unleash/pull/3574

Also includes some fixes and improvements identified in the meantime.

image


frontend/src/component/demo/Demo.tsx
```diff
@@ -1,52 +1,57 @@
...
-    const { uiConfig } = useUiConfig();
     const [welcomeOpen, setWelcomeOpen] = useState(
         storedProgress.welcomeOpen ?? true
     );
-    const [active, setActive] = useState(false);
+    const [finishOpen, setFinishOpen] = useState(false);
+
     const [expanded, setExpanded] = useState(storedProgress.expanded ?? true);
-    const [topic, setTopic] = useState(storedProgress.topic ?? 0);
+    const [topic, setTopic] = useState(storedProgress.topic ?? -1);
```
gastonfournier, line 26: Would this be better?
```suggestion
    const [topic, setTopic] = useState(storedProgress.topic ?? defaultProgress.topic);
```

gastonfournier: If yes, the line right below might also benefit from a similar change

frontend/src/component/demo/Demo.tsx
```diff
@@ -1,52 +1,57 @@
...
             topic,
             steps,
         });
-    }, [welcomeOpen, expanded, active, topic, steps]);
+    }, [welcomeOpen, expanded, topic, steps]);
 
-    if (!uiConfig.flags.demo) return null;
+    const onStart = () => {
+        setTopic(0);
+        setSteps([0]);
```
gastonfournier, line 40: On the topic field, I'm curious why it's not `-1` as in the default. Maybe it's just that zero is the first topic? I'm tempted to suggest `setTopic(defaultProgress.topic + 1);` but that's an assumption that the default topic is set to -1 and feels like unnecessary coupling, so I don't like it, but: 
```suggestion
        setTopic(0); 
        setSteps(defaultProgress.steps);
```

frontend/src/component/demo/DemoSteps/DemoSteps.tsx
```diff
@@ -59,54 +60,58 @@ export const DemoSteps = ({
...
     const back = () => {
         setFlow('back');
         if (steps[topic] === 0) {
             const newTopic = topic - 1;
-            setTopic(newTopic);
-            setStep(newTopic, topics[newTopic].steps.length - 1);
+            setTopicStep(newTopic, topics[newTopic].steps.length - 1);
         } else {
-            setStep(topic, steps[topic] - 1);
+            setTopicStep(topic, steps[topic] - 1);
```
gastonfournier, line 97: Does it make sense to protect from potentially a `-2` here? Something like:
```suggestion
            setTopicStep(topic, min(steps[topic] - 1, -1));
```

frontend/src/component/demo/DemoSteps/DemoSteps.tsx
```diff
@@ -59,54 +60,58 @@ export const DemoSteps = ({
...
-            setTopic(newTopic);
-            setStep(newTopic, 0);
+            setTopicStep(newTopic, 0);
         }
     };
 
     const next = (index = steps[topic]) => {
         setFlow('next');
-        setStep(topic, index + 1);
+        setTopicStep(topic, index + 1);
```
gastonfournier, line 114: Likewise here, we could protect this:
```suggestion
        setTopicStep(topic, max(index + 1, topics[topic].steps.length - 1));
```
An alternative way of protecting against this would be in the next line changing the `===` with a `>=`

gastonfournier (approved):
I really like it! Just some minor comments, non of them a blocker, just sharing some ideas.

frontend/src/component/demo/DemoSteps/DemoSteps.tsx
nunogois: Maybe something to refactor and change the logic a bit in the future. Will not prioritize this for now but will keep it in mind.

frontend/src/component/demo/DemoSteps/DemoSteps.tsx
nunogois: Maybe something to refactor and change the logic a bit in the future. Will not prioritize this for now but will keep it in mind.

PR #3582, opened 2023-04-21
Disable and enable strategies - frontend

Tymek:
image


kwasniew (approved):
Very nice. 
A few non blocking comments:
* I'd add the 2 new flags to our hosted instance (even before you start work on a new feature)
* Disabling last strategy doesn't disable a toggle - is it by design?
* For all new code I write I try to increase test coverage. Maybe after merging this PR try to add at least one cypress tests for the Change Request user journey?

sjaanus:
> * For all new code I write I try to increase test coverage. Maybe after merging this PR try to add at least one cypress tests for the Change Request user journey?

@kwasniew we agreed to avoid adding Cypress tests for the use cases that are not essential to the core functionality of the application. These tests tend to have a high maintenance overhead and can be flaky, which diminishes their value.

PR #3583, opened 2023-04-21
chore: GA project specific segments

gastonfournier:
## About the changes
General availability of project-specific segments 

Now new permissions can be assigned to non-admin users to allow them to create segments specifically bounded to a project (note that unique naming restriction applies across projects, this limitation can be lifted later).
image

This enables a section in the project configuration to add project-specific segments
image


gardleopard (approved):
LGTM

PR #3586, opened 2023-04-21
docs: removal of event hook in v5

gastonfournier:
## About the changes
Make it explicit that eventHook is no longer an option in v5, the migration should be to https://docs.getunleash.io/reference/addons/webhook which is already linked in the docs since v4


chriswk (approved):
Nice, good to see removal of this. :+1:

PR #3604, opened 2023-04-24
chore: make login history GA (remove flag)

nunogois:
https://linear.app/unleash/issue/2-956/remove-loginhistory-feature-flag

Promotes [login history](https://docs.getunleash.io/reference/login-history) to GA by removing the `loginHistory` feature flag.


chriswk (approved):
Good. :)

PR #3605, opened 2023-04-24
docs: openapi schema for PATs

daveleek:
## About the changes

OpenAPI schema updates for Personal Access Tokens, http statuses, property types and examples, return types


src/lib/openapi/spec/pat-schema.ts
```diff
@@ -3,27 +3,42 @@ import { FromSchema } from 'json-schema-to-ts';
...
+            type: 'integer',
+            description:
+                'The unique identification number for this Personal Access Token.',
+            example: 1,
+            minimum: 1,
         },
         secret: {
             type: 'string',
+            description: 'The token used for authentication.',
+            example: 'user.xyzrandomstring',
```
nunogois, line 19:
```suggestion
            example: 'user:xyzrandomstring',
```

src/lib/routes/admin-api/user/pat.ts
```diff
@@ -48,7 +52,17 @@ export default class PatController extends Controller {
                 openApiService.validPath({
                     tags: ['Personal access tokens'],
                     operationId: 'getPats',
-                    responses: { 200: createResponseSchema('patsSchema') },
+                    summary: 'Get all current users Personal Access Tokens.',
+                    description:
+                        'Returns all of the [Personal Access Tokens](https://docs.getunleash.io/how-to/how-to-create-personal-access-tokens) belonging to the current user.',
+                    responses: {
+                        200: createResponseSchema('patsSchema'),
```
daveleek, line 59: Should we split `patsSchema` into separate request and response types?

src/lib/routes/admin-api/user/pat.ts
```diff
@@ -48,7 +52,17 @@ export default class PatController extends Controller {
                 openApiService.validPath({
                     tags: ['Personal access tokens'],
                     operationId: 'getPats',
-                    responses: { 200: createResponseSchema('patsSchema') },
+                    summary: 'Get all current users Personal Access Tokens.',
```
nunogois, line 55: May sound a bit unclear 🤔 Would this be clearer?
```suggestion
                    summary: 'Get all Personal Access Tokens for the current user.',
```

src/lib/routes/admin-api/user/pat.ts
thomasheartman: Yeah, probably. Because you can't provide the same data when you create one, right?

src/lib/routes/admin-api/user/pat.ts
```diff
@@ -61,8 +75,18 @@ export default class PatController extends Controller {
...
+                        'Creates a new [Personal Access Token](https://docs.getunleash.io/how-to/how-to-create-personal-access-tokens) for the current user.',
                     requestBody: createRequestSchema('patSchema'),
-                    responses: { 200: createResponseSchema('patSchema') },
+                    responses: {
+                        201: resourceCreatedResponseSchema('patSchema'),
+                        ...getStandardResponses(401, 403),
+                        404: {
+                            description:
+                                'Is returned when Personal Access Tokens are disabled.',
+                        },
```
thomasheartman, line 88: Oh, this is very interesting. Can you just use 404 for these for now? We have some changes coming to the standard responses with the new error type, so we probably want to find a way to handle this. Can you instead put this in the operation description?

thomasheartman (commented):
Generally looks pretty good to me, but there's a few things that you've identified yourself that we should look at.

src/lib/routes/admin-api/user/pat.ts
daveleek: Yeah it's just a 404 for now. Should I just put the description in the general description of the method, and make the 404 a standardResponse?

src/lib/routes/admin-api/user/pat.ts
daveleek: Yeah indeed, it would be ignored if you tried no? I'll create a task for this under the technical debt tag

src/lib/openapi/spec/pat-schema.ts
```diff
@@ -3,27 +3,42 @@ import { FromSchema } from 'json-schema-to-ts';
...
             type: 'string',
             format: 'date-time',
             nullable: true,
+            example: '2023-04-19T08:15:14.000Z',
+            description: 'When the token was created.',
         },
         seenAt: {
             type: 'string',
             format: 'date-time',
             nullable: true,
```
thomasheartman, line 38: These three properties (`expiresAt`, `createdAt`, `seenAt`) are marked as nullable. in what cases are they null? Especially interesting is `createdAt`. surely that shouldn't ever be null, right?

If this is because this schema still describes both read and write models, then that should be clearly marked. Any properties that are optional (or ignored) no write models should call it out in their description.

daveleek: Of the 3, only `seenAt` is actually nullable, and all three are ignored on write. I've updated the description to reflect this

src/lib/openapi/spec/pat-schema.ts
```diff
@@ -3,27 +3,42 @@ import { FromSchema } from 'json-schema-to-ts';
...
     $id: '#/components/schemas/patSchema',
     type: 'object',
+    description:
+        'An overview of a [Personal Access Token](https://docs.getunleash.io/how-to/how-to-create-personal-access-tokens).',
     properties: {
         id: {
-            type: 'number',
+            type: 'integer',
+            description:
+                'The unique identification number for this Personal Access Token. This property is set backend when token is created.',
```
thomasheartman, line 12: It's a little verbose, but is this clearer? You could probably copy this to any of the other descriptions where it's relevant.

```suggestion
                'The unique identification number for this Personal Access Token. This property is set by Unleash when the token is created and cannot be set manually: if you provide a value when creating a PAT, Unleash will ignore it.',
```

thomasheartman, line 12: It's a little verbose, but is this clearer? You could probably copy this to any of the other descriptions where it's relevant.

```suggestion
                'The unique identification number for this Personal Access Token. (This property is set by Unleash when the token is created and cannot be set manually: if you provide a value when creating a PAT, Unleash will ignore it.)',
```

src/lib/openapi/spec/pat-schema.ts
```diff
@@ -3,27 +3,42 @@ import { FromSchema } from 'json-schema-to-ts';
...
+            description:
+                'When the token was created. This property is set backend when token is created.',
         },
         seenAt: {
             type: 'string',
             format: 'date-time',
             nullable: true,
+            example: '2023-04-19T08:15:14.000Z',
+            description:
+                'When the token was last seen/used to authenticate with. Null if it has not been used yet, this property is ignored on write.',
```
thomasheartman, line 41: Minor formatting suggestion:

```suggestion
                'When the token was last seen/used to authenticate with. `null` if it has not been used yet. (This property is set by Unleash when the token is created and cannot be set manually: if you provide a value when creating a PAT, Unleash will ignore it.)',
```

src/lib/openapi/spec/pat-schema.ts
daveleek: I picked the other one you suggested! Thanks

PR #3606, opened 2023-04-24
chore: move google auth ui behind flag

sighphyre:
This moves the access tab for Google Auth behind a flag. The elements are still accessible but hidden by default so this is a soft change. This is a deprecated feature and is on its way out.

Related to https://github.com/ivarconr/unleash-enterprise/pull/444, which locks out the backend from using Google Auth if this flag is not set

UI when the flag is off:
image


chriswk (approved):
LGTM

PR #3607, opened 2023-04-24
feat: unify error responses

thomasheartman:
This PR implements the first version of a suggested unification (and documentation) of the errors that we return from the API today.

The goal is for this to be the first step towards the error type defined in this internal [linear task](https://linear.app/unleash/issue/1-629/define-the-error-type 'Define the new API error type').

## The state of things today

As things stand, we currently have no (or **very** little) documentation of the errors that are returned from the API. We mention error codes, but never what the errors may contain.

Second, there is no specified format for errors, so what they return is arbitrary, and based on ... Who knows? As a result, we have multiple different errors returned by the API depending on what operation you're trying to do. What's more, with OpenAPI validation in the mix, it's absolutely possible for you to get two completely different error objects for operations to the same endpoint.

Third, the errors we do return are usually pretty vague and don't really provide any real help to the user. "You don't have the right permissions". Great. Well what permissions do I need? And how would I know? "BadDataError". Sick. Why is it bad?

... You get it.

## What we want to achieve

The ultimate goal is for error messages to serve both humans and machines. When the user provides bad data, we should tell them what parts of the data are bad and what they can do to fix it. When they don't have the right permissions, we should tell them what permissions they need.

Additionally, it would be nice if we could provide an ID for each error instance, so that you (or an admin) can look through the logs and locate the incident.

## What's included in **this** PR?

This PR does not aim to implement everything above. It's not intended to magically fix everything. Its goal is to implement the necessary **breaking** changes, so that they can be included in v5. Changing error messages is a slightly grayer area than changing APIs directly, but changing the format is definitely something I'd consider breaking.

So this PR:

-   defines a minimal version of the error type defined in the [API error definition linear task](https://linear.app/unleash/issue/1-629/define-the-error-type).
-   aims to catch all errors we return today and wrap them in the error type
-   updates tests to match the new expectations.

An important point: because we are cutting v5 very soon and because work for this wasn't started until last week, the code here isn't necessarily very polished. But it doesn't need to be. The internals can be as messy as we want, as long as the API surface is stable.

That said, I'm very open to feedback about design and code completeness, etc, but this has intentionally been done quickly.

Please also see my inline comments on the changes for more specific details.

### Proposed follow-ups

As mentioned, this is the first step to implementing the error type. The public API error type only exposes `id`, `name`, and `message`. This is barely any more than most of the previous messages, but they are now all using the same format. Any additional properties, such as `suggestion`, `help`, `documentationLink` etc can be added as features without breaking the current format. This is an intentional limitation of this PR.

Regarding additional properties: there are some error responses that must contain extra properties. Some of these are documented in the types of the new error constructor, but not all. This includes `path` and `type` properties on 401 errors, `errors` on validation errors, and more.

Also, because it was put together quickly, I don't yet know exactly how we (as developers) would **prefer** to use these new error messages within the code, so the internal API (the new type, name, etc), is just a suggestion. This can evolve naturally over time if (based on feedback and experience) without changing the public API.

## Error types:

We have `ValidationError`, `BadDataError`, and `BadRequestError` today. Do we really need all of these? I think we all of these are essentially the same thing?

## Returning multiple errors

Most of the time when we return errors today, we only return a single error (even if many things are wrong). AJV, the OpenAPI integration we use does have a setting that allows it to return all errors in a request instead of a single one. I suggest we turn that on, but that we do it in a separate PR (because it updates a number of other snapshots).

I've suggested we use the following format as a base to work off of:
```ts
type: ErrorDesecription = { description: string }

{
  errors: [ErrorDescription, ...ErrorDescription[]] // a non-empty list of descriptors
}
``` 

That gives json that looks a bit like this:

```json
{
  "name": "BadDataError",
  "message": "Something went wrong. Check the `errors` property for more information."
  "errors": [{
    "description": "The .params property must be an object. You provided an array.",
  }]
}
```

However, we already have used `details` and `message` in some places, which would look like this.

```json
{
  "name": "BadDataError",
  "message": "Something went wrong. Check the `errors` property for more information."
  "details": [{
    "message": "The .params property must be an object. You provided an array.",
  }]
}
```

I propose changing that to `errors` and `descriptions`, but am willing to discuss it. I think the term `description` is a better descriptor of what's contained in each separate item, but I don't have very strong feelings regarding `errors` vs `details`. Let me know if you have any thoughts.

## Failing tests

As of right now, there are a number of failing tests. I'm working on fixing them, but would like to get this reviewed as soon as possible to agree on the direction.

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -28,7 +28,7 @@ import { Knex } from 'knex';
 import maintenanceMiddleware from './middleware/maintenance-middleware';
 import { unless } from './middleware/unless-middleware';
 import { catchAllErrorHandler } from './middleware/catch-all-error-handler';
-import NotFoundError from './error/notfound-error';
+import { UnleashError } from './error/api-error';
 
 export default async function getApp(
     config: IUnleashConfig,
@@ -174,13 +174,6 @@ export default async function getApp(
         services.openApiService.useErrorHandler(app);
     }
 
-    // handle all API 404s
-    app.use(`${baseUriPath}/api`, (req) => {
-        throw new NotFoundError(
-            `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
-        );
-    });
-
     if (process.env.NODE_ENV !== 'production') {
         app.use(errorHandler());
     } else {
@@ -189,6 +182,16 @@ export default async function getApp(
 
     app.get(`${baseUriPath}`, (req, res) => {
         res.send(indexHTML);
+    });
+
+    // handle all API 404s
+    app.use(`${baseUriPath}/api`, (req, res) => {
+        const error = new UnleashError({
+            name: 'NotFoundError',
+            message: `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
+        });
+        res.status(error.statusCode).send(error);
+        return;
     });
 
     app.get(`${baseUriPath}/*`, (req, res) => {
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -184,27 +184,17 @@ export default async function getApp(
         res.send(indexHTML);
     });
 
-    // // handle all API 404s
-    // app.use(`${baseUriPath}/api`, (req, res) => {
-    //     const error = new UnleashError({
-    //         name: 'NotFoundError',
-    //         message: `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
-    //         suggestion: 'Try using a path that does exist.',
-    //     });
-    //     res.status(error.statusCode).send(error);
-    //     return;
-    // });
+    // handle all API 404s
+    app.use(`${baseUriPath}/api`, (req, res) => {
+        const error = new UnleashError({
+            name: 'NotFoundError',
+            message: `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
+        });
+        res.status(error.statusCode).send(error);
+        return;
+    });
 
     app.get(`${baseUriPath}/*`, (req, res) => {
-        if (req.path.startsWith(`${baseUriPath}/api`)) {
-            const error = new UnleashError({
-                name: 'NotFoundError',
-                message: `The path you were looking for (${req.path}) is not available.`,
-            });
-            res.status(error.statusCode).send(error);
-            return;
-        }
-
         res.send(indexHTML);
     });
 
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -28,7 +28,7 @@ import { Knex } from 'knex';
 import maintenanceMiddleware from './middleware/maintenance-middleware';
 import { unless } from './middleware/unless-middleware';
 import { catchAllErrorHandler } from './middleware/catch-all-error-handler';
-import { UnleashError } from './error/api-error';
+import NotFoundError from './error/notfound-error';
 
 export default async function getApp(
     config: IUnleashConfig,
@@ -174,6 +174,13 @@ export default async function getApp(
         services.openApiService.useErrorHandler(app);
     }
 
+    // handle all API 404s
+    app.use(`${baseUriPath}/api`, (req) => {
+        throw new NotFoundError(
+            `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
+        );
+    });
+
     if (process.env.NODE_ENV !== 'production') {
         app.use(errorHandler());
     } else {
@@ -182,16 +189,6 @@ export default async function getApp(
 
     app.get(`${baseUriPath}`, (req, res) => {
         res.send(indexHTML);
-    });
-
-    // handle all API 404s
-    app.use(`${baseUriPath}/api`, (req, res) => {
-        const error = new UnleashError({
-            name: 'NotFoundError',
-            message: `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
-        });
-        res.status(error.statusCode).send(error);
-        return;
     });
 
     app.get(`${baseUriPath}/*`, (req, res) => {
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -28,6 +28,7 @@ import { Knex } from 'knex';
 import maintenanceMiddleware from './middleware/maintenance-middleware';
 import { unless } from './middleware/unless-middleware';
 import { catchAllErrorHandler } from './middleware/catch-all-error-handler';
+import { UnleashError } from './error/api-error';
 
 export default async function getApp(
     config: IUnleashConfig,
@@ -183,9 +184,24 @@ export default async function getApp(
         res.send(indexHTML);
     });
 
+    // // handle all API 404s
+    // app.use(`${baseUriPath}/api`, (req, res) => {
+    //     const error = new UnleashError({
+    //         name: 'NotFoundError',
+    //         message: `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
+    //         suggestion: 'Try using a path that does exist.',
+    //     });
+    //     res.status(error.statusCode).send(error);
+    //     return;
+    // });
+
     app.get(`${baseUriPath}/*`, (req, res) => {
         if (req.path.startsWith(`${baseUriPath}/api`)) {
-            res.status(404).send({ message: 'Not found' });
+            const error = new UnleashError({
+                name: 'NotFoundError',
+                message: `The path you were looking for (${req.path}) is not available.`,
+            });
+            res.status(error.statusCode).send(error);
             return;
         }
 
--- a/src/lib/error/api-error.ts
+++ b/src/lib/error/api-error.ts
@@ -22,6 +22,7 @@ const UnleashApiErrorTypes = [
     'UnknownError',
     'PasswordMismatch',
     'DisabledError',
+    'ContentTypeError',
 
     // server errors; not the end user's fault
     'InternalError',
@@ -31,6 +32,8 @@ type UnleashApiErrorKind = typeof UnleashApiErrorTypes[number];
 
 export const statusCode = (errorKind: UnleashApiErrorKind): number => {
     switch (errorKind) {
+        case 'ContentTypeError':
+            return 415;
         case 'ValidationError':
             return 400;
         case 'BadDataError':
--- a/src/lib/error/api-error.ts
+++ b/src/lib/error/api-error.ts
@@ -1,5 +1,6 @@
 import { v4 as uuidV4 } from 'uuid';
 import { FromSchema } from 'json-schema-to-ts';
+import { ErrorObject } from 'ajv';
 
 const UnleashApiErrorTypes = [
     'OwaspValidationError',
@@ -258,6 +259,22 @@ export const fromLegacyError = (e: Error): UnleashError => {
         message: e.message,
         errors: [{ description: 'whoops' }],
     });
+};
+
+export const fromOpenApiValidationError =
+    (requestBody: object) =>
+    (validationError: ErrorObject): ValidationErrorDescription => {
+        console.log(requestBody, validationError);
+
+        throw new Error();
+    };
+
+export const fromOpenApiValidationErrors = (
+    requestBody: object,
+    validationErrors: ErrorObject[],
+): UnleashError => {
+    console.log(requestBody, validationErrors);
+    throw new Error();
 };
 
 export type ApiErrorSchema = FromSchema<typeof apiErrorSchema>;
--- a/src/lib/error/api-error.ts
+++ b/src/lib/error/api-error.ts
@@ -4,7 +4,6 @@ import { FromSchema } from 'json-schema-to-ts';
 const UnleashApiErrorTypes = [
     'OwaspValidationError',
     'PasswordUndefinedError',
-    'MinimumOneEnvironmentError',
     'InvalidTokenError',
     'NoAccessError',
     'UsedTokenError',
@@ -29,6 +28,7 @@ const UnleashApiErrorTypes = [
 
 // types that have extra data associated with them
 const UnleashApiErrorTypes2 = [
+    'MinimumOneEnvironmentError',
     'BadDataError',
     'BadRequestError',
     'ValidationError',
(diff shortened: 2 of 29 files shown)


src/lib/app.ts
```diff
@@ -183,9 +184,24 @@ export default async function getApp(
...
+    // // handle all API 404s
+    // app.use(`${baseUriPath}/api`, (req, res) => {
+    //     const error = new UnleashError({
+    //         name: 'NotFoundError',
+    //         message: `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
+    //         suggestion: 'Try using a path that does exist.',
+    //     });
+    //     res.status(error.statusCode).send(error);
+    //     return;
+    // });
```
thomasheartman, line 196: note to self: figure out why this doesn't work

src/lib/error/api-error.ts
thomasheartman, line 1: This is, in many ways, the meat of this PR. This file contains the new error type (both its implementation and its public schema). It covers a lot of the ground previously covered in `util.ts` (the massive error code switch, for instance.)

src/lib/app.ts
```diff
@@ -183,12 +184,17 @@ export default async function getApp(
...
-        }
+    // handle all API 404s
+    app.use(`${baseUriPath}/api`, (req, res) => {
+        const error = new UnleashError({
+            name: 'NotFoundError',
+            message: `The path you were looking for (${baseUriPath}/api${req.path}) is not available.`,
+        });
+        res.status(error.statusCode).send(error);
+        return;
+    });
```
thomasheartman, line 195: The idea here is that this should cover all API paths regardless of the method used (as opposed to the previous one that only covered GETs). I had some problems with this not working correctly, but I can't reproduce this now. If you don't think this is a good idea for some reason or if there is a better way to solve this, let me know.

src/lib/error/no-access-error.ts
thomasheartman, line 1: This class is changed to extend the new API error class instead.

src/lib/middleware/authorization-middleware.ts
thomasheartman, line 1: This changes auth middleware uses of 401 errors.

src/lib/openapi/util/standard-responses.ts
thomasheartman, line 1: There is probably a better way to do this, but this is a brute force way of making sure that all endpoints that use the standard responses for errors now have included error examples.

src/lib/routes/admin-api/addon.ts
thomasheartman, line 1: This isn't related to the PR directly, but I realized these 404s were missing. Happy to take it out of the PR if you think it's relevant.

src/test/e2e/api/admin/feature.e2e.test.ts
```diff
@@ -324,9 +324,9 @@ test('refuses to create a new feature toggle with variant when type is json but
         .set('Content-Type', 'application/json')
         .expect(400)
         .expect((res) => {
-            expect(res.body.isJoi).toBe(true);
-            expect(res.body.details[0].type).toBe('invalidJsonString');
-            expect(res.body.details[0].message).toBe(
+            console.log(res.body);
+            expect(res.body.errors[0].type).toBe('invalidJsonString');
```
thomasheartman, line 328: Not actually sure we should have a `type` property on errors like this, but I'm also not against it. This isn't defined on the error type just yet, so maybe we can postpone that conversation?

src/lib/app.ts
kwasniew: It's common in the middleware stack architecture for the 404 handler to be put just before the last error handler and pass the error with next(new NotFoundError()) and have all the error formatting in one error handler

src/lib/services/user-service.ts
```diff
@@ -24,10 +24,11 @@ import SettingService from './setting-service';
 import { SimpleAuthSettings } from '../server-impl';
 import { simpleAuthSettingsKey } from '../types/settings/simple-auth-settings';
 import DisabledError from '../error/disabled-error';
-import PasswordMismatch from '../error/password-mismatch';
+// import PasswordMismatch from '../error/password-mismatch';
```
kwasniew, line 27: Is this intentional?

src/lib/error/api-error.ts
```diff
@@ -0,0 +1,270 @@
...
+    'DisabledError',
+    'ContentTypeError',
+    'NotImplementedError',
+
+    // server errors; not the end user's fault
+    'InternalError',
+] as const;
+
+// types that have extra data associated with them
+const UnleashApiErrorTypes2 = [
```
kwasniew, line 30: I'd put the comment information into the const name itself. Adding number 2 seems to be obscure.

src/lib/error/api-error.ts
```diff
@@ -0,0 +1,270 @@
...
+    'AuthenticationRequired',
+    'NoAccessError',
+] as const;
+
+const AllUnleashApiErrorTypes = [
+    ...UnleashApiErrorTypes,
+    ...UnleashApiErrorTypes2,
+] as const;
+
+type UnleashApiErrorKind =
```
kwasniew, line 44: can it be `typeof AllUnleashApiErrorTypes[number]` since you already did the merging in lines 39-42?

src/lib/error/api-error.ts
```diff
@@ -0,0 +1,270 @@
...
+export class UnleashError extends Error {
+    id: string;
+
+    name: UnleashApiErrorKind;
+
+    documentationLink: string | null;
+
+    statusCode: number;
+
+    additionalParameters: object;
```
kwasniew, line 152: do we have a use case for this one? I'd avoid adding speculative fields.

src/lib/services/user-service.ts
thomasheartman: Nope. I'll handle it.

src/lib/services/user-service.ts
```diff
@@ -298,7 +299,12 @@ class UserService {
             await this.store.successfullyLogin(user);
             return user;
         }
-        throw new PasswordMismatch();
+
+        throw new UnleashError({
+            name: 'PasswordMismatchError',
+            message:
+                'The password you provided does not match the username. If you have forgotten your password ...',
```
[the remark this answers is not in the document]
thomasheartman: It is not, but I don't know what you can do if you have forgotten your username. Do you?

src/lib/error/api-error.ts
thomasheartman: Absolutely! I just forgot to rename it after hashing it out.

src/lib/app.ts
thomasheartman: > Why not new NotFoundError? Why do you prefer to put the absence semantics in the name string vs the type itself?

Good question! This is actually sort of at the heart of this issue: I don't know whether we should ask people to always use the new error type or if they should keep using the old ones. We can use the new one to enforce that you pass in certain required data (such as errors, path and type, etc), but we might not want people to switch regardless.

> It's common in the middleware stack architecture for the 404 handler to be put just before the last error handler and pass the error with next(new NotFoundError()) and have all the error formatting in one error handler

Right! I'm not very familiar with express, so that's very good to know. I'm not sure where it'd go, then, but I'd be happy to take a look.

src/lib/error/api-error.ts
thomasheartman: Yeah, probably! Good shout.

src/lib/error/api-error.ts
thomasheartman: We have multiple! I'd like to ultimately get rid of this myself, but we can't for now:
Some of our error messages must pass additional parameters for the Unleash app to work correctly. For instance, if the 401 you get before you log in to the UI doesn't contain `path` and `type`, you won't be redirected to login screen. Similarly, there's other error versions we have that contain extra data. 

There may well be a better way to ensure these are also passed, though, so if you have suggestions, I'm open to listening!

src/lib/error/api-error.ts
```diff
@@ -0,0 +1,270 @@
...
+                    ValidationErrorDescription,
+                    ...ValidationErrorDescription[],
+                ];
+            }
+      );
+
+export class UnleashError extends Error {
+    id: string;
+
+    name: UnleashApiErrorKind;
```
kwasniew, line 146: In the bounded context (linguistic boundaries) of Unleash error I'd stick to either name or kind but not both. Since we're not in the bounded context of type theory I vote for name over kind.

thomasheartman: Sure! So would you call the type `UnleashApiErrorName`?

src/lib/services/user-service.ts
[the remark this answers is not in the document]
kwasniew: From a security perspective you shouldn't indicate whether the username exists or not. There should be one generic message that username and password don't work + a link to forgotten password page.

src/test/e2e/api/admin/project/features.e2e.test.ts
```diff
@@ -777,8 +776,11 @@ test('Trying to patch variants on a feature toggle should trigger an OperationDe
         ])
         .expect(403)
         .expect((res) => {
-            expect(res.body.details[0].message).toEqual(
-                'Changing variants is done via PATCH operation to /api/admin/projects/:project/features/:feature/variants',
+            expect(res.body.message.includes('PATCH'));
+            expect(
+                res.body.message.includes(
+                    '/api/admin/projects/:project/features/:feature/variants',
```
kwasniew, line 782: why are we relaxing the check here?

src/lib/services/openapi-service.ts
```diff
@@ -58,10 +59,36 @@ export class OpenApiService {
     useErrorHandler(app: Express): void {
         app.use((err, req, res, next) => {
             if (err && err.status && err.validationErrors) {
-                res.status(err.status).json({
-                    error: err.message,
-                    validation: err.validationErrors,
+                const errors = err.validationErrors.map((validationError) => {
```
kwasniew, line 62: Lines 62 to  90 are getting quite involved. I'd extract them into a function and unit test in isolation.

thomasheartman: Aye, I agree with that.

src/lib/error/api-error.ts
```diff
@@ -0,0 +1,270 @@
...
+    additionalParameters: object;
+
+    constructor({
+        name,
+        message,
+        documentationLink,
+        ...rest
+    }: UnleashErrorData) {
+        super();
+        this.id = uuidV4();
```
kwasniew, line 161: it's convenient and useful but not testable since we're baking randomness without a way to control it. But maybe it's a tradeoff worth taking.

kwasniew (approved):
Wow, you put hell lot of work into this. Kudos. I'm assuming this should not break any UI code that was reading details[index].message. TBH I preferred our old naming with error having details+messages. Error with errors inside feels weird. But hey, this is naming so it's very personal.

src/lib/error/api-error.ts
kwasniew: Yeap

src/lib/app.ts
thomasheartman: Attempted to address this in d481fa02d4c7e0b5182e17257becd1c4bc63e37d. I would very much appreciate it if you had a look to make sure that it's done correctly 🙏🏼

src/lib/error/api-error.ts
thomasheartman: Addressed in 6a16be1437cd929212acd8c95b303c7e2683bfcd

src/lib/error/api-error.ts
thomasheartman: Interesting point! I wouldn't mind taking this out and doing it in some other way, but maybe save that for a later PR?

src/lib/error/api-error.ts
```diff
@@ -0,0 +1,270 @@
...
+        this.statusCode = statusCode(name);
+
+        this.additionalParameters = rest;
+    }
+
+    help(): string {
+        return `Get help for id ${this.id}`;
+    }
+
+    serialize(): ApiErrorSchema {
```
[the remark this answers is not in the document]
thomasheartman: Addressed in 9ae66c505874bbd38d5116b755e88920fbf2341b.

src/lib/services/user-service.ts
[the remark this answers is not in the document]
thomasheartman: That's great! Yeah, I don't think the code knows whether the username exists or not, but maybe my phrasing implies it. I've addressed this comment in f6dce47d7e06dc0b8122e89a0fef7cd69704c157. Does that work better, you think?

src/lib/error/api-error.ts
thomasheartman: Addressed in 9ae66c505874bbd38d5116b755e88920fbf2341b

thomasheartman:
@kwasniew Thanks a lot for a very thorough and thoroughly helpful review! 🙏🏼  I think I've addressed all your comments now except for the one about unit testing the error messages for openapi validation errors. I'll have a look at that now.

In the meantime, there's a few comments I haven't resolved because I would like some more input from you before I do. They are: 
- [Password mismatch error](https://github.com/Unleash/unleash/pull/3607#discussion_r1176155073)
- [additional error parameters](https://github.com/Unleash/unleash/pull/3607#discussion_r1176169354)
- [relaxing checks](https://github.com/Unleash/unleash/pull/3607#discussion_r1176195092)
- [random id creation in constructor](https://github.com/Unleash/unleash/pull/3607#discussion_r1176200740)

Please resolve them if you think they're OK.

As for your general comment on error object properties:

> TBH I preferred our old naming with error having details+messages. Error with errors inside feels weird. But hey, this is naming so it's very personal.

Yeah, I see what you mean. `error.errors` isn't great. I'm happy to leave it at `details` (I **do** like that), but I still prefer `description` over `message`.  Would that compromise work for you?

src/lib/services/openapi-service.ts
thomasheartman: Addressed in 2d4aaf9cb22c4e4f6cc8dec958696a9998437c9f

src/lib/services/user-service.ts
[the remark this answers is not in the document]
kwasniew: Works for me :)

src/test/e2e/api/admin/project/features.e2e.test.ts
kwasniew: Fair point

src/lib/error/api-error.ts
kwasniew: Sure, that was a non-blocking comment about design principles (testability vs convenience)

src/lib/error/api-error.ts
kwasniew: Fair point

PR #3608, opened 2023-04-24
feat: update the user-specific topic

nunogois:
https://linear.app/unleash/issue/2-961/update-demo-guide-steps-to-not-include-a-userids-strategy

Updates the user-specific toggle topic to use a standard strategy with a constraint instead (deprecate UserID strategy).

image


gardleopard (approved):
Cool! Is it by design that its data-testid and not data-testId?

nunogois:
> Cool! Is it by design that its data-testid and not data-testId?

It's an [HTML data attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes), so all lowercase makes sense 🙂 this is also consistent with the rest of the codebase.

PR #3616, opened 2023-04-25
chore: deprecate username on api-tokens

daveleek:
## About the changes

This deprecates the `username` properties on api-token schemas, and adds a `tokenName` property.
DB field `username` has been renamed to `token_name`, migration added for the rename.
Both `username` and `tokenName` can be used when consuming the service, but only one of them.

## Discussion points

There's a couple of things I'd like to get opinions on and discuss:
- Frontend still uses the deprecated `username` property
- ApiTokenSchema is used both for input and output of `Create` controller endpoints and should be split out into separate schemas. I'll set up a task for this


src/lib/openapi/spec/api-token-schema.ts
thomasheartman, line 1: I definitely suggest splitting this, but we can do it in a different PR. 

My reasons are:
`secret`, `createdAt`, and `providedAt` are definitely **not** something we accept when you create the token.

src/lib/openapi/spec/api-token-schema.ts
```diff
@@ -12,6 +12,13 @@ export const apiTokenSchema = {
         },
         username: {
             type: 'string',
+            deprecated: true,
+        },
+        tokenName: {
+            type: 'string',
```
thomasheartman, line 18: Can it be both `null` and `undefined`? Why? 

I'd suggest having it always present on response schemas. On request schemas it could be optional and used instead of `username` if provided. Ideally we'd specify that you must **either** provide `username` or `tokenName` but that the other one could be missing. That requires some clever `oneOf` trickery, though.

src/lib/routes/admin-api/api-token.ts
```diff
@@ -168,7 +168,7 @@ export class ApiTokenController extends Controller {
             201,
             res,
             apiTokenSchema.$id,
-            serializeDates(token),
+            serializeDates({ ...token, tokenName: token.username }),
```
thomasheartman, line 171: Should the `tokenName` be added by the service/store when you `createApiToken`? That way, you don't need to remember to map it manually? If we don't want to store it in the database, you could still use the `fromRow` method (that I assume exists in the store) to populate it.

src/lib/routes/admin-api/api-token.ts
```diff
@@ -204,8 +204,15 @@ export class ApiTokenController extends Controller {
...
     }
 
-    private async accessibleTokens(user: User): Promise<IApiToken[]> {
-        const allTokens = await this.apiTokenService.getAllTokens();
+    private async accessibleTokens(user: User): Promise<IApiTokenExt[]> {
+        const allTokens = (await this.apiTokenService.getAllTokens()).map(
+            (token) => {
+                return {
+                    ...token,
+                    tokenName: token.username,
```
thomasheartman, line 212: Same concern here as above.

src/test/e2e/api/admin/api-token.e2e.test.ts
thomasheartman, line 1: To avoid these getting out of sync, could they all say:

```ts
            expect(res.body.tokenName).toBe(res.token.username);
```

?

src/lib/openapi/spec/api-token-schema.ts
```diff
@@ -12,6 +12,13 @@ export const apiTokenSchema = {
         },
         username: {
             type: 'string',
+            deprecated: true,
```
thomasheartman, line 15:
```suggestion
            deprecated: true,
            description: 'This property was deprecated in Unleash v5. Prefer the `tokenName` property instead.'
```

thomasheartman (commented):
Nice work! 🙌🏼  I have a few comments inline, and then there's this:
I **really** think we should split create/get schemas for the api token schema. There's a lot of properties you can't set, so I don't see the value in not splitting it out for now. What's the reasoning here?

I'm also assuming you're not updating the api token schema for OpenAPI just yet, so I'll ignore the missing descriptions etc.

Anyway: good stuff and I'm very happy to see this deprecation happen 😄

src/lib/openapi/spec/api-token-schema.ts
daveleek: Indeed, and also the new property `tokenName` isn't used in creation either. I brought the split up in the standup, and I think the consensus was to make a small task out of this, stick to deprecating, and make follow up tasks for the technical debt that isn't time sensitive

src/lib/openapi/spec/api-token-schema.ts
daveleek: I guess of the two - `null` is the least likely one. On response it shouldn't be either, but for request it's either not set at all, or if a consumer has seen it and passed in a `null` which might not be very likely.

Yeah what to do during the migration phase of this property is an interesting discussion, but I'll set up tasks for that and we can discuss that there?

src/lib/routes/admin-api/api-token.ts
daveleek: Thank you, this was one of the things I specifically wanted feedback on! Is putting this in the model closer to DB when it's not backed by the DB schema something to be cautious about?

src/test/e2e/api/admin/api-token.e2e.test.ts
daveleek: I don't see why not! I'll fix!

src/lib/openapi/spec/api-token-schema.ts
daveleek: Brilliant, thanks!

src/lib/openapi/spec/api-token-schema.ts
thomasheartman: Yeah, good point. And you're right. When you think about it that way, it does make sense. Yeah, I'm happy to do that to get the deprecation into v5.

src/lib/openapi/spec/api-token-schema.ts
thomasheartman: Hmm, I'd set it to be non-nullable, I think. Because if the user provides a string for `username` but `null` for tokenname, which is suppose to supersede it: what do we do?

src/lib/routes/admin-api/api-token.ts
thomasheartman: Ah, that's fair. To be honest, I don't know what the pattern we prefer to use for this is, but my thinking is that if tha mapping is done in the store, then nowhere else needs to worry about it. And even if the backing store doesn't change the column names, the store interface still acts as if it has changed. but you'd need to ask someone else about whether that's a pattern we follow in this code base or not.

thomasheartman:
@daveleek Why is the `tokenName` property ignored on creation now? If that's the preferred new property, shouldn't it take precedence over `username` if both are provided?

daveleek:
@thomasheartman Initially wanted to keep the scope down, and since I wasn't changing the frontend to accommodate the changes I figured I'd just ignore it until we started implementing the change. But your point is good, it does make sense doing implementing that property both ways as we're deprecating the other property both ways. I'll get that sorted!

src/lib/routes/admin-api/api-token.ts
sighphyre: > if tha mapping is done in the store, then nowhere else needs to worry about it

Yes please! I don't see a problem with this approach and it makes this much less risky. I hate touching token things

ivarconr: I concur.

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -202,8 +202,15 @@ export class ProjectApiTokenController extends Controller {
     private async accessibleTokens(
         user: User,
         project: string,
-    ): Promise<IApiToken[]> {
-        const allTokens = await this.apiTokenService.getAllTokens();
+    ): Promise<IApiTokenExt[]> {
+        const allTokens = (await this.apiTokenService.getAllTokens()).map(
+            (token) => {
```
ivarconr, line 207: mapping concern.

src/lib/types/models/api-token.ts
```diff
@@ -37,6 +37,10 @@ export interface IApiToken extends Omit<IApiTokenCreate, 'alias'> {
     alias: string | null;
 }
 
+export interface IApiTokenExt extends IApiToken {
```
ivarconr, line 40: I struggle a bit to understand the different use cases for `IApiToken` vs `IApiTokenExt`. Can you explain how they are different and when would we use `IApiTokenExt` without a `tokenName` when we will always map username to the tokenName?

I am left wondering if we in our internal model should just rename `username` to `tokenName` and just accept either one on creation, but keep it as `tokenName` internally.

daveleek: I wanted to get some discussion going around this, but at the same time have something reasonably functional that had a minimal effect on the type system close to the db. (in case there were some reflection-schema magic that could trip us up). So I kept the changes as external and unobtrusive as possible for the initial conversation.

> I am left wondering if we in our internal model should just rename username to tokenName and just accept either one on creation, but keep it as tokenName internally.
Should we also rename the database field if so?

daveleek:
I've renamed the property in the database and added the dual-property support to the store. I've also added a bunch of dual-property support-schema changes. I've seen the [strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks) and I'll have a look at that! But I'd love some feedback on the direction of this pull request @thomasheartman @ivarconr @sighphyre @chriswk

chriswk (approved):
This looks mature enough now. Great job listening to all the feedback David :+1:

src/lib/db/api-token-store.ts
```diff
@@ -27,6 +27,7 @@ interface ITokenInsert {
     created_at: Date;
     seen_at?: Date;
     environment: string;
+    tokenName?: string;
```
gastonfournier, line 30:
```suggestion
    tokenName?: string; // we're migrating off using username to using tokenName, after which tokenName should be mandatory
```

gastonfournier: And maybe mark username as optional

src/lib/db/api-token-store.ts
```diff
@@ -154,6 +156,7 @@ export class ApiTokenStore implements IApiTokenStore {
             await Promise.all(updateProjectTasks);
             return {
                 ...newToken,
+                username: newToken.tokenName,
```
gastonfournier, line 159:
```suggestion
                username: newToken.tokenName, // for backward compatibility
```

gastonfournier (approved):
This looks good after the changes we discussed to document this better :+1:

PR #3622, opened 2023-04-26
openapi: enable `allErrors` and update snapshots

thomasheartman:
## What

This PR changes our AJV configuration to return all errors it finds with a schema instead of stopping at the first one.

Because of this, a number of the snapshot tests that we have must be updated.

## Why

DX! As someone using an API: if I send a faulty request, I'd rather have the API tell me about the five things that are wrong than for me to learn about one thing, send a new request, learn about another thing, send a new request, .... etc.


kwasniew (approved):
Agree, at the boundaries of the system we should return all errors in one go.

PR #3626, opened 2023-04-26
docs: update docs to reflect that google SSO is removed

sighphyre:
Updates docs to point out that Google SSO is removed and how to temporarily bring it back

### DIFF
--- a/website/docs/how-to/how-to-add-sso-google.md
+++ b/website/docs/how-to/how-to-add-sso-google.md
@@ -4,7 +4,7 @@ title: '[Deprecated] How to add SSO with Google'
 
 :::caution Deprecation notice
 
-Single Sign-on via the Google Authenticator provider is deprecated. We recommend using [OpenID Connect](./how-to-add-sso-open-id-connect.md) instead.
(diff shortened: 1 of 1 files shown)


website/docs/how-to/how-to-add-sso-google.md
```diff
@@ -4,7 +4,7 @@ title: '[Deprecated] How to add SSO with Google'
 
 :::caution Deprecation notice
 
-Single Sign-on via the Google Authenticator provider is deprecated. We recommend using [OpenID Connect](./how-to-add-sso-open-id-connect.md) instead.
+Single Sign-on via the Google Authenticator provider has been removed in Unleash v5. We recommend using [OpenID Connect](./how-to-add-sso-open-id-connect.md) instead. If you're running a self hosted version of Unleash and you need to temporarily re-enable Google SSO, you can do so by setting the `GOOGLE_AUTH_ENABLED` environment variable to `true`. If you're running a hosted version of Unleash, you'll need to reach out to us and ask us to re-enable the flag. Note that this code will be removed in a future release and this is not safe to depend on.
```
chriswk, line 7:
```suggestion
Single Sign-on via the Google Authenticator provider has been removed in Unleash v5 (deprecated in v4). We recommend using [OpenID Connect](./how-to-add-sso-open-id-connect.md) instead. If you're running a self hosted version of Unleash and you need to temporarily re-enable Google SSO, you can do so by setting the `GOOGLE_AUTH_ENABLED` environment variable to `true`. If you're running a hosted version of Unleash, you'll need to reach out to us and ask us to re-enable the flag. Note that this code will be removed in a future release and this is not safe to depend on.
```

chriswk (approved):
Added a small comment on keeping the note that it was deprecated in v4.

Still. LGTM

thomasheartman (approved):
Looks good to me!

PR #3629, opened 2023-04-26
chore: remove @ts-ignore-error

thomasheartman:
This PR removes two instances of @ts-ignore-error by proving to the compiler that the lists are indeed not empty.

## Background

These two directives were introduced in #3607. The reason they were introduced is that they provide a non-empty list to the error. We have just asserted that the list have at least 1 element in them, but the compiler doesn't trust this. So it's safe to ignore that.

However, in doing so, we're also ignoring the error that said this we provided an unexpected property. That's bad.


src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -362,24 +362,23 @@ export default class ExportImportService {
...
+                unsupportedContextFields.map((field) => {
                     const description = `${field.name} is not supported.`;
                     return {
                         description,
                         message: description,
                     };
-                }),
-            });
+                });
+            if (firstError) {
```
thomasheartman, line 374: Ensure that `firstError` actually exists. The compiler is pretty smart, but not infallible. For instance, if you do:

```ts
// create an empty list
const xs: { name: string }[] = []

// map and split it
const [head, ...tail] = xs

// try and assign it to a non-empty list
details: [head, ...tail]
```
It will gladly accept it. 

So instead, we need to verify that `firstError` actually **does** exist.

kwasniew (approved):

PR #3633, opened 2023-04-26
Clean up old errors

thomasheartman:
This PR attempts to improve the error handling introduced in #3607.

## **tl;dr:**
- Make `UnleashError` constructor protected
- Make all custom errors inherit from `UnleashError`.
- Add tests to ensure that all special error cases include their relevant data
- Remove `PasswordMismatchError` and `BadRequestError`. These don't exist.
- Add a few new error types: `ContentTypeError`, `NotImplementedError`, `UnauthorizedError`
- Remove the `...rest` parameter from error constructor
- Add an unexported `GenericUnleashError` class
- Move OpenAPI conversion function to `BadDataError` clas
- Remove explicit `Error.captureStackTrace`. This is done automatically.
- Extract `getPropFromString` function and add tests

### **In a more verbose fashion**

The main thing is that all our internal errors now inherit from`UnleashError`. This allows us to simplify the `UnleashError` constructor and error handling in general while still giving us the extra benefits we added to that class. However, it _does_ also mean that I've had to update **all** existing error classes.

The constructor for `UnleashError` is now protected and all places that called that constructor directly have been updated. Because the base error isn't available anymore, I've added three new errors to cover use cases that we didn't already have covered: `NotImplementedError`, `UnauthorizedError`, `ContentTypeError`. This is to stay consistent in how we report errors to the user.

There is also an internal class, `GenericUnleashError` that inherits from the base error. This class is only used in conversions for cases where we don't know what the error is. It is not exported.

In making all the errors inherit, I've also removed the `...rest` parameter from the `UnleashError` constructor. We don't need this anymore.

Following on from the fixes with missing properties in #3638, I have added tests for all errors that contain extra data.

Some of the error names that were originally used when creating the list don't exist in the backend. `BadRequestError` and `PasswordMismatchError` have been removed.

The `BadDataError` class now contains the conversion code for OpenAPI validation errors. In doing so, I extracted and tested the `getPropFromString` function.

### Main files

Due to the nature of the changes, there's a lot of files to look at. So to make it easier to know where to turn your attention:

The changes in `api-error.ts` contain the main changes: protected constructor, removal of OpenAPI conversion (moved into `BadDataError`.

`api-error.test.ts` contains tests to make sure that errors work as expected.

Aside from `get-prop-from-string.ts` and the tests, everything else is just the required updates to go through with the changes. 

## Discussion points

I've gone for inheritance of the Error type over composition. This is in large part because throwing actual Error instances instead of just objects is preferable (because they collect stack traces, for instance). However, it's quite possible that we could solve the same thing in a more elegant fashion using composition.

## For later / suggestions for further improvements

The `api-error` files still contain a lot of code. I think it might be beneficial to break each Error into a separate folder that includes the error, its tests, and its schema (if required). It would help decouple it a bit.

We don't currently expose the schema anywhere, so it's not available in the openapi spec. We should look at exposing it too.

Finally, it would be good to go through each individual error message and update each one to be as helpful as possible.


src/lib/util/get-prop-from-string.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * This function does not do any processing or validation of the property path.
+ * Empty paths and non-existent properties yield `undefined`. Whitespace is
+ * valid as a property name (so `x. .y`) is a valid path.
+ *
+ * @param {string} propertyPath The path to the property you want to extract. First part of the path is assumed to be on the root level of the `object` you provide.
+ * @param {object} object The object you want to extract the property from.
+ * @return {T = unknown | undefined} If the nested property exists, return that as `unknown` or try to coerce it into to the type you've provided. Otherwise, return `undefined`.
+ */
+export const getPropFromString = <T = unknown>(
```
kwasniew, line 12: I'd suggest we use https://lodash.com/docs/4.17.15#get that is exposed as micro lib here: https://www.npmjs.com/package/lodash.get

thomasheartman: Interesting. I'd be on board with doing that, but adding an external dependency for that feels a little ... unnecessary? But happy to do it if you think it's worth the tradeoff. 

Out of curiosity: what's your reason for preferring an external lib to writing the function ourself?

kwasniew: * This dependency is deliberately a single function not the entire lib so it doesn't bring anything extra we don't need (I'm against bringing code we don't need)
* non-differentiator, just a generic boring fn I'd rather not review, maintain, look at the code no matter how well written it is
* We don't have to maintain tests for it, lodash does

src/lib/error/api-error.ts
```diff
@@ -102,52 +91,11 @@ const statusCode = (errorName: UnleashApiErrorName): number => {
...
-                details: [
-                    {
-                        validationErrors: string[];
-                        message: string;
-                    },
-                ];
-            }
-      );
-
 export class UnleashError extends Error {
```
kwasniew, line 99: Why is UnleashError in api-error.ts while all the other errors have their own files?

src/lib/error/disabled-error.ts
```diff
@@ -1,9 +1,8 @@
-import BaseError from './base-error';
+import { UnleashError } from './api-error';
 
-class DisabledError extends BaseError {
+class DisabledError extends UnleashError {
     constructor(message: string) {
-        super(message, 422, 'DisabledError');
-        Error.captureStackTrace(this, this.constructor);
+        super({ message, name: 'DisabledError' });
```
kwasniew, line 5: I've realized you're always passing `name: 'ClassName'` in every child error implementation. This is unnecessary and may lead to subtle typos. My suggestion is to stop passing name into super constructor and get it in UnleashError using this code as an inspiration:

```js
class ParentError extends Error {
    constructor(message) {
        super(message);
        console.log(
            `ParentError constructor called from ${this.constructor.name}`,
        );
    }
}

class ChildError extends ParentError {
    constructor(message) {
        super(message);
        console.log(
            `ChildError constructor called from ${this.constructor.name}`,
        );
    }
}

const error = new ChildError('This is an error message');
```

kwasniew (changes requested):
I started reviewing the first two points from the TL;DR and found something that should be addressed before I continue the review so I'm submitting a request changes proposal.

src/lib/error/disabled-error.ts
thomasheartman: I love this suggestion. Commit 35f3d4910a36e03d5ee6c6a45476918a5c88ff84:
- removes explicit class name passing, instead relying on `this.constructor.name`
- simplifies the base class constructor, allowing us to remove most of the constructors that we currently do have (based on an eslint feedback)

However, because the base constructor was protected, the compiler had some issues with building it. As such, I've made the constructor public for the base class. But I realized that JS does have abstract classes, which is helpful in this case. So I made the base class abstract (this seems to be the closest we can get to an interface while having it remain an Error type?).

Using constructor names instead of static strings does also mean that there's no such thing as exhaustive string matching for error types anymore, so I've had to relax switch statements and some types to allow general strings too. 

I've added an optional `name` parameter to the base error class to allow you to override the name that's set. This is primarily an escape hatch to use for the generic error class. It's just a suggestion and we can for sure change it.

src/lib/error/bad-data-error.ts
```diff
@@ -1,23 +1,90 @@
...
-            ],
+            ...super.toJSON(),
+            details: this.details,
         };
     }
 }
 
 export default BadDataError;
+
+export const fromOpenApiValidationError =
```
kwasniew, line 37: do we have any tests describing what this function does? The logic is quite involved and I'd appreciate some examples showing how a given input translates to a given output.

src/lib/error/bad-data-error.ts
```diff
@@ -1,23 +1,90 @@
...
+    (requestBody: object) =>
+    (validationError: ErrorObject): ValidationErrorDescription => {
+        // @ts-expect-error Unsure why, but the `dataPath` isn't listed on the type definition for error objects. However, it's always there. Suspect this is a bug in the library.
+        const propertyName = validationError.dataPath.substring(
+            '.body.'.length,
+        );
+        if (validationError.keyword === 'required') {
```
kwasniew, line 44: I'd use "extract method" refactoring for each branch behind if/else if to explain the intent of each branch

src/lib/error/bad-data-error.ts
```diff
@@ -1,23 +1,90 @@
...
+            ...super.toJSON(),
+            details: this.details,
         };
     }
 }
 
 export default BadDataError;
+
+export const fromOpenApiValidationError =
+    (requestBody: object) =>
```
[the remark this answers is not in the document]
thomasheartman: Depends on what you mean by "at this stage"? This is a function, so we're saying that we want an object, right?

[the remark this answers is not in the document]
kwasniew: is it possible to put our system in a state where requestBody is not an object?

src/lib/error/api-error.test.ts
```diff
@@ -333,3 +338,79 @@ describe('Error serialization special cases', () => {
...
+        expect(error.details![0].message).toBe(description);
+        expect(error.details![0].description).toBe(description);
+    });
+
+    it('OwaspValidationErrors: adds `validationErrors` to `details`', () => {
+        const results = owasp.test('123');
+        const error = new OwaspValidationError(results);
+        const json = error.toJSON();
+
+        expect(json.message).toBe(results.errors[0]);
```
kwasniew, line 379: IMO it's much more readable when tests use the actual object shape in one expect assertions instead of 3 different expect calls. The key is to use https://jestjs.io/docs/expect#tomatchobjectobject

src/lib/error/bad-data-error.ts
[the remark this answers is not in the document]
kwasniew: When the req.body arrives over the wire its type is `unknown` until we validate/parse it into an object. My question is if we always call this function after validate/parse was done somewhere else in the code.

kwasniew (approved):
Very nice improvements. Non blocking approve with some comments.

src/lib/error/bad-data-error.ts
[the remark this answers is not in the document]
thomasheartman: Oh, right. TypeScript should prevent us from doing that, right (unknown isn't an object)? But yeah, this is only called if you get OpenAPI validation errors because something is wrong with your request. Shouldn't happen, but I'll verify that it doesn't break with an empty body. God catche

src/lib/error/api-error.test.ts
thomasheartman: I agree in most cases. I'll look aver the code and make any necessary adjustments.

src/lib/util/get-prop-from-string.ts
thomasheartman: Yeah, I buy that. Done in e002678050454a967aa4a3164a5e10a83ebc7268.

src/lib/error/api-error.test.ts
thomasheartman: Yeah, I agree! Didn't realize you could put matchers inside `toMatchObject`, but TIL 😄  c21a299af6124d0a0dfb9c4ffd7ef13e28f66cba

src/lib/error/bad-data-error.ts
thomasheartman: Done in b1b7a2938a22795dbf35aff385b07200b605f032

src/lib/error/bad-data-error.ts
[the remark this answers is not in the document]
thomasheartman: Can confirm: if you try and post something without a body (and you mark it as json), it gets parsed into an empty object (presumably by express).

src/lib/error/api-error.ts
thomasheartman: Tests and source file have been renamed in 71cf90741920d08d545d4133880b979d0c61daa5.

PR #3635, opened 2023-04-27
feat: export by tags

kwasniew:
## About the changes
Export environment toggle config by tags. This will be needed for promote feature toggles script.

* currently tags take precedence over feature list
* removed some typechecker error in the export-import-service


sjaanus (approved):
LGTM!

PR #3638, opened 2023-04-27
fix: ensure that the owaspvalidationerrors always contain all errors

thomasheartman:
## What

This fixes a bug where the owasp validation response that came back from the server no longer contained all its validation errors. This caused the UI to slightly break and for the password validation box not to update correctly.

This is a quick fix (but with tests!). I'm working on a better design for this on the side, but it seemed pertinent to get this out ASAP.

### DIFF
--- a/src/lib/error/api-error.test.ts
+++ b/src/lib/error/api-error.test.ts
@@ -1,3 +1,4 @@
+import owasp from 'owasp-password-strength-test';
 import { ErrorObject } from 'ajv';
 import {
     ApiErrorSchema,
@@ -9,6 +10,7 @@ import {
     UnleashError,
 } from './api-error';
 import BadDataError from './bad-data-error';
+import OwaspValidationError from './owasp-validation-error';
 
 describe('v5 deprecation: backwards compatibility', () => {
     it.each(UnleashApiErrorTypes)(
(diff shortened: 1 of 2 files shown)


src/lib/error/api-error.test.ts
```diff
@@ -320,3 +322,14 @@ describe('OpenAPI error conversion', () => {
...
     });
 });
+
+describe('Error serialization special cases', () => {
+    it('OwaspValidationErrors: adds `validationErrors` to `details`', () => {
+        const results = owasp.test('123');
+        const error = new OwaspValidationError(results);
+        const json = fromLegacyError(error).toJSON();
+
+        expect(json.details!![0].message).toBe(results.errors[0]);
```
kwasniew, line 332: what's the double exclamation mark here?

gardleopard (approved):
LGTM

kwasniew (approved):

src/lib/error/api-error.test.ts
thomasheartman: But maybe a single one'll do? I saw we did the double somewhere else, so I copied that.

thomasheartman: Update: yeah, a single one is enough.

kwasniew: yeah I was surprised by a double exclamation mark :)

PR #3639, opened 2023-04-27
feat: promote toggles script

kwasniew:
## About the changes

Added first version of the promote feature toggles script + fixed one query that was failing on empty toggles list


sjaanus (approved):
Good first version. Does not allow to run it from CLI, because need to modify file and cannot use parameters, but can improve later.

PR #3647, opened 2023-04-28
fix: correct error for missing context field

chriswk:
When adding a strategy using a context field that did not exist, we threw an unknown error. 

This changes to throw NotFoundError so that our users can better know what they did wrong.

### DIFF
--- a/src/lib/db/context-field-store.ts
+++ b/src/lib/db/context-field-store.ts
@@ -6,6 +6,7 @@ import {
     IContextFieldStore,
     ILegalValue,
 } from '../types/stores/context-field-store';
+import NotFoundError from '../error/notfound-error';
 
 const COLUMNS = [
     'name',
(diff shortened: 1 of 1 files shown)


sjaanus (approved):

kwasniew (approved):

thomasheartman (commented):
Interesting. I'd usually expect NotFoundErrors to be 404s. But if what can't be found is a nested part of a request object, shouldn't it be a 400 because the data is bad? If so, how do we differentiate between "route doesn't exist" and "your data is bad because it points to something that doesn't exist"?

chriswk:
We discussed the same thing, and came to the conclusion that the 404 is still a better error message than 500 UnknownError.

The next step is having a strategy for the validation of constraints that changes the 404 to 400.

andreas-unleash (commented):
LGTM!

PR #3650, opened 2023-04-28
docs: add migration guide entry for v4 -> v5

thomasheartman:
This lists the breaking changes between the versions to make it easier (and safer) for users to upgrade


JarrodSC (approved):
Looks good to me

PR #3652, opened 2023-04-28
docs: unleash instance synchronization

kwasniew:
## About the changes
Documenting how to synchronize Unleash instances

### DIFF
--- a/website/docs/how-to/how-to-synchronize-unleash-instances.md
+++ b/website/docs/how-to/how-to-synchronize-unleash-instances.md
@@ -1,0 +1,64 @@
+---
+title: 'How to synchronize Unleash instances'
+---
+
+:::note 
+
+This is experimental feature
+
+:::
+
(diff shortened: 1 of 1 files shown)


website/docs/how-to/how-to-synchronize-unleash-instances.md
```diff
@@ -0,0 +1,64 @@
...
+
+:::
+
+# Unleash Instance Synchronization Script 
+
+This script allows you to synchronize feature toggles between two Unleash instances using the export and import APIs provided by Unleash. The script exports feature toggles from the source instance and imports them into the target instance.
+
+You can find this script in the following location within the project:
+
+`src/scripts/promote.sh`
```
sjaanus, line 17: Should be just /scripts/ without src

sjaanus (approved):
Small remark but other than that looks ok. Still wondering if we should have an example to run it with single command.

PR #3663, opened 2023-05-02
feat: demo sorts features by name

nunogois:
https://linear.app/unleash/issue/2-983/can-we-we-reverse-order-of-the-toggles

This PR makes it so that, on topic navigation to the project overview page, we assume a feature toggle name sorting by default:

image


NicolaeUnleash (approved):
Nice!

PR #3665, opened 2023-05-02
chore: remove optimal 304 flag

gastonfournier:
## About the changes
This PR removes the optimal304 flag after being tested in production. We're keeping the existing configuration that allows users to disable cache mainly because it's useful for testing.


src/lib/routes/client-api/feature.ts
```diff
@@ -130,39 +119,30 @@ export default class FeatureController extends Controller {
             ],
         });
 
-        if (clientFeatureCaching?.enabled) {
-            this.cache = true;
-            this.cachedFeatures = memoizee(
-                (query) => this.resolveFeaturesAndSegments(query),
+        if (clientFeatureCaching.enabled) {
+            this.featuresAndSegments = memoizee(
```
gastonfournier, line 123: This one uses the code of `cachedFeatures2` from the old code, but instead of defining a field holding the cache, we're dynamically defining the behavior of the method at construction time, to either use the cache or call directly to the method

chriswk (approved):
I love this. You know someone is going to ask how to I prevent caching , but I say we go for it :)

src/lib/routes/client-api/feature.ts
```diff
@@ -130,39 +119,30 @@ export default class FeatureController extends Controller {
...
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                (query: IFeatureToggleQuery, etag: string) =>
+                    this.resolveFeaturesAndSegments(query),
                 {
                     promise: true,
-                    maxAge: clientFeatureCaching.maxAge,
+                    maxAge: clientFeatureCaching.maxAge
```
ivarconr, line 129: do we then default "clientFeatureCaching.maxAge" to undefined? I think this logic should be moved to where we setup all the options instead of in-lining it here.

ivarconr: The current default maxAge will probably always win?

gastonfournier: You mean for clarity, because the linter puts the Elvis operator in 3 lines, right? I think that makes sense

gastonfournier: Oh, I know what you mean!

gastonfournier: Check out: https://github.com/Unleash/unleash/pull/3665/commits/11e71ade467ddddda4a811c747556f0729f936bc

gastonfournier:
> I love this. You know someone is going to ask how to prevent caching , but I say we go for it :)

It can still be disabled: https://github.com/Unleash/unleash/pull/3665/files#diff-a38cc7ae9f963264333b6535498970a33e046dba50a6af554f438343a6edd2ddR77-R79 I'm not sure it should be disabled, but I think it's ok to let users decide

PR #3666, opened 2023-05-02
feat: set timezone to utc

sjaanus:
If Unleash is running in a timezone other than UTC, the metrics bucket values will be incorrect by the corresponding number of hours. We need to decide whether we should allow Unleash to run in other timezones or if we should modify the code to enforce UTC.

We decided to set timezone to UTC. There are two ways to do it, during runtime and before starting node. We decided to go with environment variable, because setting at runtime produces non-deterministic behaviour. Source https://github.com/nodejs/node-v0.x-archive/issues/3286. 

An example, metrics was posted at 13:00+3, and was stored in database 13:00+0. After setting application timezone, it is now stored 10:00+0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,8 @@ FROM node:$NODE_VERSION
 
 ENV NODE_ENV production
 
+ENV TZ UTC
+
 WORKDIR /unleash
 
 COPY --from=builder /unleash/docker /unleash
--- a/package.json
+++ b/package.json
@@ -30,9 +30,9 @@
   "license": "Apache-2.0",
   "main": "./dist/lib/server-impl.js",
   "scripts": {
-    "start": "node ./dist/server.js",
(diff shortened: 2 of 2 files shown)


kwasniew (approved):
Let's add info in PR desc why we're adding it before the Node.js process so that someone can have context for this decision.

sighphyre (commented):
This seems like a small change but it also seems like a big change. Does this really not break a single test? I kinda expected it to break a few. If this doesn't break any tests are we missing a sane test case or does this do less than we think it does

Also, definitely agree with @kwasniew, this kinda PR is one of those small changes that's easy to miss and easy to get confused about later, a record of why we're doing this and what it impacts would be amazing here before we merge

sjaanus:
@sighphyre the tests have been running with UTC timezone since 2019.

PR #3672, opened 2023-05-03
fix: anonymize email in event payload

nunogois:
https://linear.app/unleash/issue/2-987/unleash-is-not-correctly-anonymizing-emails-in-the-project-events

Context: https://unleash-internal.slack.com/archives/C048ELND3QD/p1683097636083299

We should also anonymize email fields in `data` and `preData` of event objects when `anonymiseEventLog` is enabled.

image


sighphyre (approved):
Nice one! LGTM

ivarconr (approved):

PR #3683, opened 2023-05-04
fix: log project access events

nunogois:
https://linear.app/unleash/issue/2-992/register-event-when-granting-access-to-a-project

 - Correctly registers a new `project-access-added` event;
 - Adds an anonymization method that supports anonymizing objects, recursively searching for keys to anonymize;
 - Fixes the event label used by `ProjectUserUpdateRoleEvent` to be `PROJECT_USER_ROLE_CHANGED`;

image

### DIFF
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -521,14 +521,12 @@ export default class ProjectService {
                     roleId,
                     roleName: role.name,
                     groups: addedGroups.map(
-                        ({ id, name, users: groupUsers }) => ({
-                            id,
+                        ({ id: groupId, name, users: groupUsers }) => ({
+                            id: groupId,
                             name,
                             users: groupUsers.map(
-                                ({
-                                    user: { id: userId, email, username },
-                                }) => ({
-                                    id: userId,
+                                ({ user: { id, email, username } }) => ({
+                                    id,
                                     email,
                                     username,
                                 }),
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -520,19 +520,10 @@ export default class ProjectService {
                 data: {
                     roleId,
                     roleName: role.name,
-                    groups: addedGroups.map(
-                        ({ id: groupId, name, users: groupUsers }) => ({
-                            id: groupId,
-                            name,
-                            users: groupUsers.map(
-                                ({ user: { id, email, username } }) => ({
-                                    id,
-                                    email,
-                                    username,
-                                }),
-                            ),
-                        }),
-                    ),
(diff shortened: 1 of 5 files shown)


src/lib/services/project-service.ts
```diff
@@ -492,13 +494,44 @@ export default class ProjectService {
...
+        const role = await this.accessService.getRole(roleId);
+        const { users, groups } = await this.getAccessToProject(projectId);
+
+        const addedUsers = users.filter(({ id }) =>
+            usersAndGroups.users.some(({ id: userId }) => userId === id),
+        );
+
+        const addedGroups = groups.filter(({ id }) =>
+            usersAndGroups.groups.some(({ id: groupId }) => groupId === id),
+        );
```
gastonfournier, line 514: Do we need this? I assume this is to filter out users who already had access? I'm not sure what's the intention here because I believe `users` should be the same as `usersAndGroups` (and therefore `addedUsers` should be the same). But I might be wrong

nunogois: I think so. `usersAndGroups` only include the `id`s of users and groups, so this is a way of fetching some extra information like group names, emails, usernames to add to the payload.

gastonfournier: Yes, but I don't like that apparently we are loading all users and groups to add that information (unless I'm understanding this wrong).
1. I'm not sure we need those details in the payload (ofc it's nice to have, but if it goes against performance I start to doubt it)
2. Maybe we can just query the user ids we have rather than fetching all and filtering out the ones we don't want

nunogois: @ivarconr WDYT? 
1. Should we only include user and group `id`'s in the payload?
2. Should we add user and group store methods to fetch multiple users and groups based on arrays of `id`'s?

ivarconr: we should not include users in a group imho

nunogois: @ivarconr We're no longer including the users that currently belong to the group in the payload:
image
What I mean is, on the `users` array, do we need to fetch the users info to add it to the payload, or are `id`s enough?

nunogois: After aligning on standup we're keeping it simple and going for `id`'s only for now: https://github.com/Unleash/unleash/pull/3683/commits/ba836681ecbdaa6aa1e504f47ce693b0f6bdbb34

image

gastonfournier (approved):
There are integration tests for this (mostly for happy case), but looking at the code looks fine. Still, I'd encourage you to have unit tests on that library `anonymise.ts`

PR #3689, opened 2023-05-04
docs: update to node 18

gastonfournier:
## About the changes
Update documentation to the minimum version of node. This was already updated in other parts of the documentation

### DIFF
--- a/website/docs/reference/deploy/getting-started.md
+++ b/website/docs/reference/deploy/getting-started.md
@@ -8,7 +8,7 @@ title: Getting Started
 
 You will need:
 
-- [Node.js](https://nodejs.org/en/download/) (version 14 or later)
+- [Node.js](https://nodejs.org/en/download/) (version 18 or later)
 - [PostgreSQL](https://www.postgresql.org/download/) (version 10 or later)
 - [Create an unleash user and database](./database-setup).
 
--- a/website/package.json
+++ b/website/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "engines": {
-    "node": ">=16.14"
+    "node": ">=18"
   },
   "scripts": {
     "docusaurus": "docusaurus",


thomasheartman (approved):
Yeah, I'm good with this change for sure 😄

thomasheartman:
@gastonfournier , I took the liberty of also updating the `engines` property of the website/package.json file. Do you agree with that change?

gastonfournier:
> @gastonfournier , I took the liberty of also updating the `engines` property of the website/package.json file. Do you agree with that change?

Yes, I remember I didn't change this, because the website runs differently than Unleash, so I wasn't sure this change was needed, but if it does work, then better have only one runtime :) thanks!

thomasheartman:
Can't see any reason why it wouldn't work? And if the docs say it requires node 18, then surely we should say that everywhere?

gastonfournier:
> Can't see any reason why it wouldn't work? And if the docs say it requires node 18, then surely we should say that everywhere?

I mean, the website is not run on the same infrastructure as Unleash, actually... I don't know where it's running... but just for the sake of thinking, imagine this is running in say Vercel, and Vercel does not support node 18 yet (or it can be a docker container that needs to switch its image from node-16 to node-18). Let me know if I'm speaking nonsense (sometimes I do, specially in the early morning :laughing:)

thomasheartman:
No, I think it makes sense, and I think I'm the one who's speaking nonsense 😄 I didn't read the PR closely enough and thought you were adjusting the docs for the docs 😅 I realize now that it's the docs for the application and that "docs: update to node 18" doesn't mean "update the docs to 18" but "update the docs to say that the app needs 18+". 

Sooo ... yeah. If you want to, we can remove that commit, but it seems that the build has worked just fine, so I see no reason not to do it?

gastonfournier:
> it seems that the build has worked just fine, so I see no reason not to do it?

Exactly! Let's do it, we can revert if we find some issue down the road :+1:

PR #3695, opened 2023-05-05
task: Add Unit test result check task

chriswk:
After a Team Retro, one of our squads felt like we needed more data on our test suites. This is the first effort to make our test results easier to grab. It uses the test-reporter action to add a github check to our main build and PR builds with our test results.

This at least should make it easier to parse which tests are failing. However, it does not give us trends. So it does not yet make it easier to decide which tests are flaky just from a quick view.


sighphyre (approved):
Awesome! Where does this report end up, out of curiousity?

.github/workflows/build_prs_test_report.yaml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'yarn'
+      - run: yarn install --frozen-lockfile --ignore-scripts
+      - run: yarn build
+      - run: yarn build:frontend
```
gastonfournier, line 35: This is out of date, build is already building the frontend, but if we want to build frontend, I think we also need to remove --ignore-scripts (I'm not sure why we have them)

.github/workflows/build_prs.yaml
```diff
@@ -19,5 +19,4 @@ jobs:
           cache: 'yarn'
       - run: yarn install --frozen-lockfile --ignore-scripts
       - run: yarn lint
-      - run: yarn build:backend
-      #- run: yarn build:frontend # not needed
+      - run: yarn build
```
gastonfournier, line 22: Not sure if this is intended, but build will also build the frontend, in which case we need to remove --ignore-scripts from yarn install to also install frontend dependencies

.github/workflows/build_prs.yaml
```diff
@@ -20,4 +20,10 @@ jobs:
       - run: yarn install --frozen-lockfile --ignore-scripts
       - run: yarn lint
       - run: yarn build:backend
-      #- run: yarn build:frontend # not needed
+      - run: yarn test:report
+      - name: Upload test report to build # Done this way since external PRs would not be able to write the check. See https://github.com/marketplace/actions/test-reporter#recommended-setup-for-public-repositories
+        uses: actions/upload-artifact@v2
+        if: success() || failure()
```
gastonfournier, line 26: Here you could use
```suggestion
        if: always()
```
But this is also fine IMO

gastonfournier (approved):
LGTM!

PR #3696, opened 2023-05-05
docs: extend group documentation to include information on setting root roles

sighphyre:
This adds documentation to the RBAC section on how to use root roles on groups and updates a few screenshots for the group pages.


website/docs/reference/rbac.md
```diff
@@ -184,4 +189,4 @@ To enable group sync, you'll need to set two fields in your SSO provider configu
...
\ No newline at end of file
+[^1]: The project-level permission is still required for the [**create/overwrite variants** (PUT)](/docs/reference/api/unleash/overwrite-feature-variants.api.mdx) and [**update variants** (PATCH)](/docs/reference/api/unleash/patch-feature-variants.api.mdx) API endpoints, but it is not used for anything within the admin UI. The API endpoints have been superseded by the [**create/overwrite environment variants** (PUT)](/docs/reference/api/unleash/overwrite-feature-variants-on-environments.api.mdx) and [**update environment variants** (PATCH)](/docs/reference/api/unleash/patch-environments-feature-variants.api.mdx) endpoints, respectively.
```
sighphyre, line 192: Gotta be honest, I have no idea what Github is telling me here, my difftool says these are identical.

sighphyre:
Huh. Spec tests have found an issue. I am 99.99% sure it's unrelated given the test but really nice to see they found something!

website/docs/reference/rbac.md
```diff
@@ -124,7 +124,7 @@ You can assign the following permissions on a per-environment level within the p
 
 :::info availability
 
-User groups are available to Unleash Enterprise users since **Unleash 4.14**.
+User groups are available to Unleash Enterprise users since **Unleash 4.14** and root role groups are planned to be released in **Unleash 5.1**.
```
thomasheartman, line 127: Just think this is clearer if it's in two separate sentences.

```suggestion
User groups are available to Unleash Enterprise users since **Unleash 4.14**. Root role groups are planned to be released in **Unleash 5.1**.
```

website/docs/reference/rbac.md
```diff
@@ -136,10 +136,15 @@ A user group consists of the following:
 - a **description** (optional)
 - a **list of users** (required)
 - a list of SSO groups to sync from (optional)
+- a root role associated with the group (optional) (only available in **Unleash 5.1** and later)
```
thomasheartman, line 139: I'm unsure about whether this should have the "(only available in ...)" or not. I'm leaning towards yes because it's unreleased for now, but I don't actually think it should stay there. So ... maybe no? What do you think?

website/docs/reference/rbac.md
thomasheartman: I'll trust you on that!

thomasheartman (approved):
Looks great! Solid work 💪🏼  I've got one suggestion and one discussion point, but I'll leave it up to you what you wanna do.

website/docs/reference/rbac.md
sighphyre: Yeah, I'm also in two minds. We've used this pattern in the past but I'm not 100% sure it's working for us since we apparently never go clean these things up.

That being said, I don't want users to find the docs and then ask why things don't work the way the docs say they should so while I don't like it, I think it should stay

We do seem to be pretty consistent with these version strings in the docs, though, could we script something to check this?

website/docs/reference/rbac.md
sighphyre: Love it!

website/docs/reference/rbac.md
thomasheartman: Sure, I'm happy to do either, so I'll leave it up to you. 

> We do seem to be pretty consistent with these version strings in the docs, though, could we script something to check this?

Depends on what you're thinking about? We can definitely script things, but what do you want to check exactly?

PR #3708, opened 2023-05-07
feat(devenv): reproducible dev environment via https://devenv.sh

ghuntley:
## About the changes

The contributing/onboarding devloop can be simplified via https://devenv./sh (which is chrome around the Nix package manager). 

```bash
$ cd unleash
$ devenv up
```

CleanShot 2023-05-07 at 13 17 43

This will automatically:
- Install the correct version of nodejs, yarn and postgres in a sandbox (ie. no sudo, no changes to a developers computer)
- Create the Postgres database and the required users.
- Start the Postgres database
- Run the two yarn scripts and merge the output logs into a single console log where the process names are tagged.

### Important files

- Installation instructions: https://devenv.sh/getting-started/
- Configuration options: https://devenv.sh/reference/options/

## Discussion points

- Install https://devenv.sh/getting-started/ on your machine to test then run `$ devenv up`
- https://devenv./sh works on macOS, Linux and Windows (via WSL)
- https://devenv./sh is completely sandboxed meaning it does not mutate or make changes to a users computer (state is kept in `unleash/.devenv`)

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -1,0 +1,41 @@
+{ pkgs, ... }:
+
+{
+  # https://devenv.sh/packages/
+  packages = [
+    pkgs.nodejs-18_x
+    pkgs.postgresql_14
+    pkgs.yarn
+  ];
+
+  # https://devenv.sh/processes/
+  processes = {
+    yarn-build-watch.exec = "yarn build:watch";
+    yarn-start-dev.exec = "yarn start:dev";
+  };
+
+  # https://devenv.sh/reference/options/#servicespostgresenable
+  services.postgres.enable = true;
+  services.postgres.listen_addresses = "127.0.0.1";
+
+  services.postgres.createDatabase = true;
+  services.postgres.initdbArgs = [
+    "--locale=C"
+    "--encoding=UTF8"
+  ];
+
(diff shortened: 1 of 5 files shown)


devenv.nix
```diff
@@ -0,0 +1,41 @@
...
+  # https://devenv.sh/packages/
+  packages = [
+    pkgs.nodejs-18_x
+    pkgs.postgresql_14
+    pkgs.yarn
+  ];
+
+  # https://devenv.sh/processes/
+  processes = {
+    yarn-build-watch.exec = "yarn build:watch";
```
thomasheartman, line 13: Why are we running both these processes? It _should_ be enough to run just `yarn start:dev`, unless you've identified something I haven't?

thomasheartman (commented):
@ghuntley This is very cool; thanks for the PR! 💯  I'm a Nix user, but haven't used devenv before: looking at the docs, I think it seems like a solid layer on top of barebones Nix. If we add this, then I might also be able to remove my custom nix shells and envrcs 😄 

I checked out your branch, though, and I have a few questions. One, regarding the processes, I've left inline. The other is:

It seems to be running on port 5200 or 5100 (does it change?). Unleash normally runs on port 4242. Is there a reason why this doesn't run Unleash on 4242?

PR #3710, opened 2023-05-08
chore: expand default segment limits to 1000

sighphyre:
This expands the segment limits to 1000, this should have no impact on OSS since this feature isn't exposed. This is overridden to 250 in hosted `pro` instances and 1000 in hosted `enterprise` customers. This only affects self hosted enterprise instances


gardleopard (approved):
LGTM

PR #3712, opened 2023-05-08
TESTING: CI optimization

sjaanus:


.github/workflows/docker_publish_main.yaml
```diff
@@ -1,9 +1,7 @@
 name: Publish main branch to dockerhub
 
 on:
-  push:
-    branches:
-      - main
+  pull_request:
```
chriswk, line 4: This feels strange. This workflow says Publish Main branch to dockerhub, and now it's only going to run on pull_requests ? :)

sjaanus: Yeah, it was very early testing, just tested to see how long it builds without arm.

PR #3717, opened 2023-05-09
fix: project api token type to lowercase

andreas-unleash:
## About the changes

Closes #


sighphyre (approved):
LGTM!

Optional but can I suggest adding a test for this? This is a really critical area to have working correctly and a test here would help prevent bugs that try sneak in

andreas-unleash:
> LGTM!
> 
> Optional but can I suggest adding a test for this? This is a really critical area to have working correctly and a test here would help prevent bugs that try sneak in

Sure thing

kwasniew (approved):

PR #3719, opened 2023-05-09
chore: add another migration that remigrates the proper way

daveleek:
## About the changes

Adds a migration that renames `token_name` back to `username`, then adds a new optional column named `token_name`

## Discussion points

I've added fallbacks for resolving username/tokenname on insert and on making rows from results.
But this adds another column renaming, which is worth discussing properly

### DIFF
--- a/src/migrations/20230509065107-recreate-username-field.js
+++ b/src/migrations/20230509065107-recreate-username-field.js
@@ -1,0 +1,17 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        'ALTER TABLE api_tokens RENAME COLUMN token_name TO username;',
+        callback,
+    );
+    db.runSql('ALTER TABLE api_tokens ADD COLUMN "token_name" text;', callback);
+};
+
+exports.down = function (db, callback) {
+    db.runSql('ALTER TABLE api_tokens DROP COLUMN "token_name";', callback);
+    db.runSql(
+        'ALTER TABLE api_tokens RENAME COLUMN username TO token_name;',
+        callback,
+    );
+};
--- a/src/lib/db/api-token-store.ts
+++ b/src/lib/db/api-token-store.ts
@@ -39,7 +39,7 @@ const tokenRowReducer = (acc, tokenRow) => {
     if (!acc[tokenRow.secret]) {
         acc[tokenRow.secret] = {
             secret: token.secret,
-            tokenName: token.token_name,
+            tokenName: token.token_name ? token.token_name : token.username,
             type: token.type,
             project: ALL,
             projects: [ALL],
@@ -48,7 +48,7 @@ const tokenRowReducer = (acc, tokenRow) => {
             createdAt: token.created_at,
             alias: token.alias,
             seenAt: token.seen_at,
-            username: token.token_name,
+            username: token.token_name ? token.token_name : token.username,
         };
     }
     const currentToken = acc[tokenRow.secret];
@@ -63,6 +63,7 @@ const tokenRowReducer = (acc, tokenRow) => {
 };
 
 const toRow = (newToken: IApiTokenCreate) => ({
+    username: newToken.tokenName ?? newToken.username,
     token_name: newToken.tokenName ?? newToken.username,
     secret: newToken.secret,
     type: newToken.type,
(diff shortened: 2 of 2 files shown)


src/migrations/20230509065107-recreate-username-field.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        'ALTER TABLE api_tokens RENAME COLUMN token_name TO username;',
```
gardleopard, line 5: is db.runSql or the entire function a transaction? I think these two should run as a transaction together? Also we should probably copy the data from username -> token_name.

daveleek: I'll look into the transaction part, that's a good point. Yup, I can add copying data as well

daveleek: I can't find anything documenting any transaction taking place using runSql.
I found two places in our migrations where we explicitly start transactions in the SQL passed to runSql. In the migrations where we run several SQL statements they are typically run in the same call to runSql.
I can change this to be inside the same runSql call, and I'll see if I can find more info on transactions

gastonfournier: Let's have a quick catch up about this

daveleek: Ah interesting. Yeah lets talk!

src/migrations/20230510113903-fix-api-token-username-migration.js
```diff
@@ -0,0 +1,36 @@
...
+                    `
+                ALTER TABLE api_tokens ADD COLUMN "token_name" text;
+                UPDATE api_tokens SET token_name = username;
+              `,
+                );
+            }
+            callback();
+        },
+    );
+};
```
gastonfournier, line 29: I mean... Neat!! :star_struck:

src/migrations/20230510113903-fix-api-token-username-migration.js
```diff
@@ -0,0 +1,36 @@
...
+        `
+        SELECT * FROM migrations WHERE name = '/20230426110026-rename-api-token-username-field';
+        `,
+        (err, results) => {
+            if (results.rows.length > 0) {
+                db.runSql(
+                    `
+                ALTER TABLE api_tokens RENAME COLUMN token_name TO username;
+                ALTER TABLE api_tokens ADD COLUMN "token_name" text;
+                UPDATE api_tokens SET token_name = username;
```
gastonfournier, line 14: Wouldn't this be the same but without the RENAME?
```suggestion
                ALTER TABLE api_tokens ADD COLUMN "username" text;
                UPDATE api_tokens SET username = token_name;
```

daveleek: The reason for the rename is that the original `username` column is NOT NULL - and the added column becomes nullable, so we keep the contract to keep old clients running. They would be getting non-null errors on `token_name`

gastonfournier: Ok now I get it! If you run `ALTER TABLE api_tokens ADD COLUMN "username" text NOT NULL;` it will fail because there are rows in the table

gastonfournier (approved):
LGTM!

src/migrations/20230510113903-fix-api-token-username-migration.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+        SELECT * FROM migrations WHERE name = '/20230426110026-rename-api-token-username-field';
```
gastonfournier, line 6: So, this didn't work in `demo` and the reason is how we did the manual down-migration with @gardleopard: we run the down migration scripts but we didn't remove the row from the `migrations` table. Therefore, this script will believe that the migration was applied but actually, it was not.

This would have been more robust if we had checked if the column existed rather than checking if the migration was applied. Unfortunately it's now late to change this again, so we're aiming to manually re-apply this migration in `demo` :facepalm:

gastonfournier: Ok, we fix by running `DELETE FROM migrations WHERE name = '/20230426110026-rename-api-token-username-field'` manually

PR #3722, opened 2023-05-09
test: add anonymise unit tests

nunogois:
Adds unit tests to `anonymise` functions. 

See: https://github.com/Unleash/unleash/pull/3683#pullrequestreview-1418223563


src/lib/util/anonymise.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+test('anonymiseKeys', () => {
+    expect(anonymiseKeys(null, [])).toBe(null);
+    expect(anonymiseKeys(undefined, [])).toBe(undefined);
+    expect(anonymiseKeys(true, [])).toBe(true);
+    expect(anonymiseKeys(false, [])).toBe(false);
+    expect(anonymiseKeys('test', [])).toBe('test');
+    expect(anonymiseKeys(123, [])).toBe(123);
+    expect(anonymiseKeys([1, 2, 3], [])).toStrictEqual([1, 2, 3]);
+    expect(anonymiseKeys({ test: 'test' }, [])).toStrictEqual({ test: 'test' });
```
gastonfournier, line 17: I think here you could do something like:
```typescript
test.each([null, undefined, true, false, 'test', 123, [1, 2, 3], { test: 'test' })
    ('Without keys should return the same',
        (obj) -> { expect(anonymiseKeys(obj, []).toStrictEqual(obj); }
    );
```

gastonfournier (approved):
I think this covers many cases, you can group them, but this looks good

src/lib/util/anonymise.test.ts
nunogois: Tried addressing this in: https://github.com/Unleash/unleash/pull/3722/commits/e24dcaef1f0ffb087bc018d5ffa4837baec53cfd

PR #3735, opened 2023-05-10
fix: log missing user at warn level

chriswk:
When using PATs if the user that the PAT is for has been removed, we currently log the missing user at ERROR level. Since this is not something our SREs can fix, this PR downgrades the NotFoundError to WARN, instead of ERROR.


sighphyre (approved):
Nice!

PR #3736, opened 2023-05-10
chore: simplify package scripts

nunogois:
# Simplify package scripts

This PR's purpose is to raise a discussion surrounding our current package scripts.

It includes some suggestions that aim to simplify the scripts and hopefully bring a much more straightforward approach to developing and contributing to Unleash.

Building (prod) should only happen **explicitly** and when needed.

## Before PR (current behavior)

- Clone the project;
- Open 2 terminals: One for `unleash` and another for `unleash/frontend`;
- On `unleash`: 
  - Run `yarn` (which will also build, for some reason?);
  - Run `yarn start:dev` to start backend in dev mode (`tsc-watch`);
- On `unleash/frontend`:
  - Run `yarn` (which will also build, for some reason?);
  - Run `yarn start` to start frontend in dev mode (`vite`);

So it seems to me like we build unnecessarily every time we install dependencies. Neither dev scripts need to build the project, as backend uses `tsc-watch` and frontend uses `vite`. I'm unsure why this is the case, as building can take a very long time.

image

There's also some complexity in the way we need to split the terminal to `cd` into `frontend` and treat it as a different project. The fact that we have different script names is also confusing (`yarn start`, `yarn start:dev`, etc).

## After PR

- Clone the project;
- Run `yarn` to install all dependencies;
- Run `yarn dev` to get started developing Unleash;

Running `yarn` should take care of everything needed to start developing. This includes installing dependencies for frontend as well. It should not build projects if we are not being explicit about it, especially since we don't need to build them at this stage.

image

Running `yarn dev` should start the project in dev mode. This means running both projects in `dev` mode, which for `backend` means running `tsc-watch` and for `frontend` means running `vite`.

Here this PR attempts to provide a better DX by using [concurrently](https://www.npmjs.com/package/concurrently) and [wait-on](https://www.npmjs.com/package/wait-on) - This means both tasks are ran simultaneously, stdout is labeled accordingly, and are stopped together. It also means that `frontend` waits for `backend` to be serving at `4242` before starting, since `frontend` starts pretty much immediately with `vite` and `backend` takes a bit longer. Of course, when the `backend` is hot-reloading you may still find some `ECONNREFUSED`s on `frontend` stdout while it recompiles.

image

No more splitting your terminal and treating `frontend` as a separate project.

## Discussion points

Maybe there's a better alternative to `tsc-watch`? I briefly explored some alternatives and while they had a much faster starting speed, hot-reload was sometimes slower. IMO we should aspire to run `src/server-dev.ts` directly and only compile when needed.

Running `dev:backend` still serves a version of the frontend (at 4242). **Why? Can we remove that behavior?** 
I can't imagine a scenario in dev where we wouldn't want to run the latest version of the frontend with `vite`.

~~**Note:** This PR removes all other out-of-scope scripts to focus on this revamp. If we decide to merge it, we should evaluate what other existing scripts we still want to include. May be a good opportunity to clean up unused ones and only include the ones we really use. This includes scripts that our GH actions rely on.~~

**Update:** In an effort to minimize impact surface of this PR and make it a bit more ready for merging:
 - It updates some docs in https://github.com/Unleash/unleash/pull/3736/commits/2a4ff805e87b65d9c1256effaa189ddcccba15fb and https://github.com/Unleash/unleash/pull/3736/commits/1bbc4882519b5a82e3116f0be255ad24a6f3ce53 to reflect our new simplified flow;
 - It includes the old package scripts for now in https://github.com/Unleash/unleash/pull/3736/commits/039bc04699ac880e491fd3ce01f9bcd6f97a94b9;
 - It updates some of our GH actions to reflect the new scripts in https://github.com/Unleash/unleash/pull/3736/commits/7782cb9b12e37ee844507e41ef2b7137eaf55666;

Given its current status I'll promote the PR to "ready for review". 

I still think we should have a second look at our existing scripts and GH actions to see what we really need and/or should adapt, but it should be a team effort so we have a broader context. Maybe on a follow-up PR.

Does this require any changes to related projects (e.g. Enterprise)?


thomasheartman (commented):
This looks great to me and I am **so** on board with these changes 😄 

I think my only question is: why are we setting on 3000 for the port instead of 4242? I feel like 4242 will have fewer collisions (such as with the proxy etc).

nunogois:
> This looks great to me and I am **so** on board with these changes 😄
> 
> I think my only question is: why are we setting on 3000 for the port instead of 4242? I feel like 4242 will have fewer collisions (such as with the proxy etc).

Thank you for the feedback 😄 

`3000` is where `vite` runs by default (current behavior). If we're worried about collisions we can easily specify a different port. I believe `vite` also [attempts to use a different port](https://vitejs.dev/config/server-options.html#server-port) in case of collision.

thomasheartman:
Thanks! Now that I think about it, it's not actually the collisions that worry me, it's just the muscle memory of going to 4242. Now I'll have to change that to 3000. But I suppose the same can be said for people who work mostly on the frontend 💁🏼 

But there **is** the thing where having it always launch on the same port is easier to remember. And that is more likely to be true if we default to 4242. Either that, or we pick a different port that we don't already use for something else. 

Anyway, I may be wrong and I don't think this should be blocking in any sense, so I'm good! I'll give this an approval in the meantime 🏆  What is the path forward here? Did you want more input from other members too?

thomasheartman (approved):

nunogois:
> Thanks! Now that I think about it, it's not actually the collisions that worry me, it's just the muscle memory of going to 4242. Now I'll have to change that to 3000. But I suppose the same can be said for people who work mostly on the frontend 💁🏼
> 
> But there **is** the thing where having it always launch on the same port is easier to remember. And that is more likely to be true if we default to 4242. Either that, or we pick a different port that we don't already use for something else.
> 
> Anyway, I may be wrong and I don't think this should be blocking in any sense, so I'm good! I'll give this an approval in the meantime 🏆 What is the path forward here? Did you want more input from other members too?

When running `yarn dev` vite will open a new browser tab on `3000` for you automatically, if it helps.

We can't specify `4242` as the default port of the frontend because the backend still runs on that port (we're running both). I'm open to suggestions if you'd like to try a different port other than `3000` - Not sure we are depending on it being `3000` anywhere else outside this scope.

Thanks! Yes, I would love more input from other members as well just to be sure we have a consensus, and also to address these points:

- Maybe there's a better alternative to tsc-watch? I briefly explored some alternatives and while they had a much faster starting speed, hot-reload was sometimes slower. IMO we should aspire to run src/server-dev.ts directly and only compile when needed.

- Running dev:backend still serves a version of the frontend (at 4242). Why? Can we remove that behavior?
I can't imagine a scenario in dev where we wouldn't want to run the latest version of the frontend with vite.

- I still think we should have a second look at our existing scripts and GH actions to see what we really need and/or should adapt, but it should be a team effort so we have a broader context. Maybe on a follow-up PR.

- Does this require any changes to related projects (e.g. Enterprise)?

We don't need to solve them now, but I think it's important to be aware of them and maybe get some input.

frontend/README.md
```diff
@@ -4,13 +4,12 @@ This directory contains the Unleash Admin UI frontend app.
 
 ## Run with a local instance of the unleash-api
 
-First, start the unleash-api backend on port 4242.
-Then, start the frontend dev server:
+Refer to the [Contributing to Unleash](../CONTRIBUTING.md#how-to-run-the-project) guide for instructions.
+The frontend dev server runs simultaneously with the backend dev server:
```
gastonfournier, line 8:
```suggestion
The frontend dev server runs (in port 3000) simultaneously with the backend dev server (in port 4242):
```

package.json
```diff
@@ -31,6 +31,13 @@
   "main": "./dist/lib/server-impl.js",
   "scripts": {
     "start": "TZ=UTC node ./dist/server.js",
+    "build:backend": "copyfiles -u 1 src/mailtemplates/**/*.mustache dist/ && tsc --pretty --strictNullChecks false",
+    "build:frontend": "cd ./frontend && yarn run build",
+    "build": "yarn run build:frontend && yarn run build:backend",
```
gastonfournier, line 36: Yes! With @sjaanus we noticed we'll need `build` to build both backend and frontend :+1:

package.json
```diff
@@ -31,6 +31,13 @@
   "main": "./dist/lib/server-impl.js",
   "scripts": {
     "start": "TZ=UTC node ./dist/server.js",
+    "build:backend": "copyfiles -u 1 src/mailtemplates/**/*.mustache dist/ && tsc --pretty --strictNullChecks false",
+    "build:frontend": "cd ./frontend && yarn run build",
```
gastonfournier, line 35: I like this because it doesn't require a change directory. Although it might seem just a preference it may matter when you combine `build:frontend` with something else (but not 100% sure)
```suggestion
    "build:frontend": "yarn --cwd ./frontend run build",
```

package.json
```diff
@@ -31,6 +31,13 @@
...
+    "build": "yarn run build:frontend && yarn run build:backend",
+    "dev:backend": "TZ=UTC NODE_ENV=development tsc-watch --strictNullChecks false --onSuccess \"node dist/server-dev.js\"",
+    "dev:frontend": "wait-on tcp:4242 && cd ./frontend && yarn run dev",
```
gastonfournier, line 38: Nice solution with `wait-on` and concurrently!! 
```suggestion
    "dev:frontend": "wait-on tcp:4242 && yarn --cwd ./frontend run dev",
```

package.json
```diff
@@ -31,6 +31,13 @@
...
+    "dev": "concurrently \"yarn:dev:backend\" \"yarn:dev:frontend\"",
+    "prepare": "node scripts/husky-install && cd ./frontend && yarn install",
```
gastonfournier, line 40:
```suggestion
    "prepare": "node scripts/husky-install && yarn --cwd ./frontend install",
```

gastonfournier (approved):
I love it! just a few comments but not blockers. I particularly like the --cwd in package.json because it doesn't change directory, I found it useful for scripting, but definitely not a blocker

sjaanus (approved):

PR #3737, opened 2023-05-10
feat: use new demo website

nunogois:
https://linear.app/unleash/issue/2-1023/use-new-demo-website-for-testing-purposes

Use https://unleash-demo-app.vercel.app/ for now, which will help us testing.
Should be reverted/updated before we release to the current `demo` instance.


NicolaeUnleash (approved):
Nice!

PR #3739, opened 2023-05-10
fix: Default email sender to getunleash.io domain

chriswk:
As part of the move to a unified domain this PR updates the default EMAIL_SENDER to noreply@getunleash.io . Should not be merged/deployed until we've verified DMARC, DKIM for the new domain.


ivarconr (approved):
Looks good. You just need to adjust the snapshots for tests!

PR #3749, opened 2023-05-11
feat: improve frontend config freshness to < 1s

ivarconr:
This PR reuses the revision Id information from the "optimal 304 for server SDKs" to improve the freshness of the frontend API config data. 

In addition it allows us to reduce the polling (and eventually remove it when we are confident).

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -431,7 +431,7 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
     const frontendApi = options.frontendApi || {
         refreshIntervalInMs: parseEnvVarNumber(
             process.env.FRONTEND_API_REFRESH_INTERVAL_MS,
-            10000,
+            20000,
         ),
     };
 
--- a/src/lib/features/feature-toggle/configuration-revision-service.ts
+++ b/src/lib/features/feature-toggle/configuration-revision-service.ts
@@ -1,7 +1,10 @@
-import { Logger } from 'lib/logger';
-import { IEventStore } from 'lib/types';
+import { EventEmitter } from 'stream';
+import { Logger } from '../../logger';
+import { IEventStore, IUnleashConfig, IUnleashStores } from '../../types';
 
-export default class ToggleConfigChangeEmitter {
+export const UPDATE_REVISION = 'UPDATE_REVISION';
+
+export default class ConfigurationRevisionService extends EventEmitter {
     private logger: Logger;
 
     private eventStore: IEventStore;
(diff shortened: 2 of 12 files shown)


src/lib/create-config.ts
```diff
@@ -431,7 +431,7 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
     const frontendApi = options.frontendApi || {
         refreshIntervalInMs: parseEnvVarNumber(
             process.env.FRONTEND_API_REFRESH_INTERVAL_MS,
-            10000,
+            20000,
```
ivarconr, line 434: updated on purpose. Polling frequency on 20s is still very conservative, and we should be able to go away from polling entirely when we trust the revision id event.

ivarconr:
I have manually tested with multiple unleash instances connected to the same DB and can observe they all updates they fronted caches in less than a second for any change to any toggle.

nunogois (approved):
: 🚀 Fantastic. Only thing I'm wondering is if we should have a feature toggle here just in case.

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    private revisionId: number;
+
+    constructor(
+        { eventStore }: Pick<IUnleashStores, 'eventStore'>,
+        { getLogger }: Pick<IUnleashConfig, 'getLogger'>,
+    ) {
+        super();
+        this.logger = getLogger(
+            'features/config-client/toggle-config-change-emitter',
```
gastonfournier, line 20: Nitpicking: I think we're usually doing this:
```suggestion
            'configuration-revision-service.ts',
```
But maybe there's benefit in having the full path (although I think there's a copy&paste leftover here):
```suggestion
            'features/feature-toggle/configuration-revision-service',
```

gastonfournier, line 20: Nitpicking: I think we're usually doing this:
```suggestion
            'configuration-revision-service.ts',
```
But maybe there's benefit in having the full path (although I think there's a copy&paste leftover here):
```suggestion
            'features/feature-toggle/configuration-revision-service',
```

src/lib/proxy/proxy-repository.ts
```diff
@@ -67,6 +73,7 @@ export class ProxyRepository
     }
 
     getToggle(name: string): FeatureInterface {
+        //@ts-ignore (we must update the node SDK to allow undefined)
```
gastonfournier, line 76: Yes! I faced the same issue, but it might be a breaking change. I'll propose a PR to discuss because other methods take into account that finding something might yield undefined

gastonfournier (approved):
LG! Just a minor comment on the logger configuration but not a blocker

gastonfournier (approved):
LG! Just a minor comment on the logger configuration but not a blocker

src/lib/features/feature-toggle/configuration-revision-service.ts
gastonfournier: GH double-posted this? :-1:

src/lib/proxy/proxy-repository.ts
ivarconr: we should do it. Its not really a breaking to allow more (I think). If you look at the actual implementation I would almost consider that a bug-fix, but let not take the full discussion here 🗡️

PR #3756, opened 2023-05-12
feat: release inputs added to dispatcher

sjaanus:
This disables the local bump as the way of triggering a new version and adds the parameters to trigger the new process


gastonfournier (approved):
LG!

PR #3759, opened 2023-05-12
fix: build is not present in npm package

gastonfournier:
## About the changes
Before publishing we should build. Not sure how this was working before


sjaanus (approved):
LGTM!

PR #3762, opened 2023-05-12
fix: demo guide dialogs behavior

nunogois:
Improves the dialogs behavior to address https://linear.app/unleash/issue/2-1053/bug-when-big-dialog-is-open-if-you-click-on-a-step-the-flow-is-broken

Also prevents close on backdrop click for the welcome dialog.


NicolaeUnleash (approved):
LGTM!

PR #3770, opened 2023-05-15
fix: add missing options parameter back in

thomasheartman:
This PR adds the missing serialization of the AuthenticationRequired response back in. It was mistakenly removed in #3633.

This PR also adds another test to verify that it the options property is present.


chriswk (approved):
Nice. LGTM.

PR #3771, opened 2023-05-15
chore: add additional methods we want exposed

gastonfournier:
## About the changes
These methods are being used externally as part of our enterprise package


chriswk (approved):
LGTM

PR #3773, opened 2023-05-15
fix: anonymise PII fields in user access if flag is set

chriswk:
### What
In the demo when listing possible users to grant access to your project, we inadvertently expose emails when listing users you can grant access to. This PR anonymises the access list on the way out.


sighphyre (approved):
LGTM!

PR #3775, opened 2023-05-15
fix: block adding a root role to a group with a project role

sighphyre:
## What

Blocks adding a root role to a group that already has a project role.

## Why

This is a bug and a confusing one at that. The group role mapping is an either/or - you either get project permissions _or_ you get a root role, but not both. Blocking this flow makes this a little saner to use.

### DIFF
--- a/src/test/e2e/services/group-service.e2e.test.ts
+++ b/src/test/e2e/services/group-service.e2e.test.ts
@@ -106,8 +106,8 @@ test('adding a root role to a group with a project role should fail', async () =
 
     stores.accessStore.addGroupToRole(group.id, 1, 'test', 'default');
 
-    try {
-        await groupService.updateGroup(
+    await expect(() => {
+        return groupService.updateGroup(
             {
                 id: group.id,
                 name: group.name,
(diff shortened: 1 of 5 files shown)


src/test/e2e/services/group-service.e2e.test.ts
```diff
@@ -97,3 +97,31 @@ test('should not remove user from no SSO definition group', async () => {
...
+                createdBy: 'test',
+            },
+            'test',
+        );
+    } catch (any) {
+        expect(any.message).toContain(
+            'This group already has a project role and cannot also be given a root role',
+        );
+    }
+    expect.assertions(1);
```
gastonfournier, line 126: I believe there's a better way of doing this: https://github.com/Unleash/unleash/blob/bd53193dc95d16ec13e147070c7fb1071f8911f3/src/test/e2e/api/client/segment.e2e.test.ts#L469-L476 but I'm not picky with tests

gastonfournier (approved):
LG

src/test/e2e/services/group-service.e2e.test.ts
sighphyre: Outside of being fewer lines is it actually better (genuinely asking, I wanna know if this does something cool I'm not aware of)? Honestly I'm trying to avoid these goopy mixed await/promise chains because we keep on getting them wrong in subtle ways and so tired of our tests doing weird things

sighphyre: ...that being said, I see we do this in a lot of places. I've taken the change, gonna merge it

PR #3777, opened 2023-05-15
fix: prevent variant name from containing extra whitespace

sighphyre:
## What

This trims off the white space surrounding a variant name when adding/editing a variant.

## Why

The behaviour here differs from feature toggle add/update.  This patches the behaviour on variants to bring this inline with adding/updating a feature toggle name.

Variant names need to be URL safe, while we do get a nice error message back, we know on the frontend that a variant name with a white space buffer will fail. This should block the user from the most obvious ways of inputting text that will fail


nunogois (approved):
LGTM, however I would like to see a smarter behavior like it's described in the task: https://linear.app/unleash/issue/2-1052/bug-variants-name-with-space-in-the-name-return-error-on-save

This approach doesn't cover some edge cases, and in general it would be better to warn the user of an invalid variant name in real-time instead of allowing the user to save and only then providing an error, therefore losing the edited data.

sighphyre:
> LGTM, however I would like to see a smarter behavior like it's described in the task: https://linear.app/unleash/issue/2-1052/bug-variants-name-with-space-in-the-name-return-error-on-save
> 
> This approach doesn't cover some edge cases, and in general it would be better to warn the user of an invalid variant name right away instead of allowing the user to save and only then providing an error, therefore losing the edited data.

Fair enough! Unpacked a little elsewhere, sounds like we should do a second pass at this to improve this app wide but this should still be merged

PR #3779, opened 2023-05-16
fix: allow null checker to not fail if it gets no output

sighphyre:
## What

Patches the build scripts so that the null checker doesn't fail


.github/workflows/gradual-strict-null-checks.yml
```diff
@@ -64,6 +64,11 @@ jobs:
           CURRENT=$(grep "Found [0-9]* errors" .out-current | sed 's/Found \(.*\) errors in .* files./\1/')
           MAIN=$(grep "Found [0-9]* errors" .out-main | sed 's/Found \(.*\) errors in .* files./\1/')
 
+          if [[ -z $CURRENT ]] || [[ -z $MAIN ]]; then
+            comment "No null checks were emitted from this patch. This is likely because no backend code was changed."
```
gastonfournier, line 68: I had this error recently, but is it because of no changes? Yarn should run on the source and generate the output, I think something else might be going on... :thinking:

gastonfournier: This was due to me adding some debug echo statements that messed up with the script

sighphyre: You are 100% correct, looked like this was the cause until literally everyone else had it as well

PR #3780, opened 2023-05-16
base path support for openapi

kwasniew:
## About the changes

Using full URL with base path in the dropdown which is different from proxy (where we only show the base path). TBH I don't know which one is better but decided to stick to full URL here because I don't know the implications of using just base path. Swagger should support both.
Screenshot 2023-05-16 at 09 10 15


kwasniew:
@thomasheartman I modified some of the tests. New logic is like this:
* if no base path is provided then do not set the servers section (which is used for base path settings)
* if base path provided then do add a servers section with a UI dropdown

How to do the interplay of base path and unleash url with the same base path included can be changed later based on the actual use case we find. I tried to make minimal changes to our current code.

thomasheartman (approved):
Nice one! Yeah, it's a good point: we can definitely adjust this later. If it does break the swagger ui requests, then that's easy enough to fix and definitely not critical. Looks great to me! 😄

PR #3785, opened 2023-05-16
chore: use concurrently to run parallel builds

gastonfournier:
## About the changes
Small improvement to speed up the build, tested locally validating the output with and without concurrently.
**Note** this also removes copy-templates from the backend build


sjaanus (approved):
LG

PR #3793, opened 2023-05-17
fix: demo QR code

nunogois:
https://linear.app/unleash/issue/2-1062/update-the-qr-code

Updates the QR code to a correct one, since the last one expired.


nicolaesocaciu (approved):
LGTM!

PR #3794, opened 2023-05-17
Basic bulk update implementation

kwasniew:
## About the changes

First draft of the bulk toggle on/off with Promise.all and no optimizations. This is to get it functional and available for integration with the UI.

In the followup PR I will implement transactions and error handling, but only once we have the UI.


src/lib/services/feature-toggle-service.ts
```diff
@@ -1257,6 +1257,30 @@ class FeatureToggleService {
...
+    async bulkUpdateEnabled(
+        project: string,
+        featureNames: string[],
+        environment: string,
+        enabled: boolean,
+        createdBy: string,
+        user?: User,
+        shouldActivateDisabledStrategies = false,
+    ): Promise<void> {
+        await Promise.all(
```
kwasniew, line 1269: I will test later if sequential creation is needed or if we can run it in parallel

kwasniew: Also I will do error handling once I have the UI and transactions

andreas-unleash (approved):
LG

PR #3798, opened 2023-05-17
Fix title bug

Tymek:


andreas-unleash (approved):
Already handled in https://github.com/Unleash/unleash/pull/3787

PR #3808, opened 2023-05-18
fix: properly handle flag resolver variants

nunogois:
Variants were not being properly handled in the `flag-resolver`: The fact that the default value of the variant is not falsy made it so we never asked the external flag resolver for the value.

This also moves the logic from `Variant | undefined` to `Variant` where we use the `getDefaultVariant()` helper method to return us a [default variant](https://github.com/Unleash/unleash-client-node/blob/55274e4953164cdfdf14469a7e318147aa4c37cc/src/variant.ts#L37-L42).

### DIFF
--- a/src/lib/util/flag-resolver.test.ts
+++ b/src/lib/util/flag-resolver.test.ts
@@ -2,6 +2,7 @@ import { PayloadType } from 'unleash-client';
 import { defaultExperimentalOptions, IFlagKey } from '../types/experimental';
 import FlagResolver, { getVariantValue } from './flag-resolver';
 import { IExperimentalOptions } from '../types/experimental';
+import { getDefaultVariant } from 'unleash-client/lib/variant';
 
 test('should produce empty exposed flags', () => {
     const resolver = new FlagResolver(defaultExperimentalOptions);
@@ -30,7 +31,7 @@ test('should use external resolver for dynamic flags', () => {
                 return true;
             }
         },
-        getVariant: () => undefined,
+        getVariant: () => getDefaultVariant(),
     };
 
     const config = {
@@ -50,7 +51,7 @@ test('should not use external resolver for enabled experiments', () => {
         isEnabled: () => {
             return false;
         },
-        getVariant: () => undefined,
+        getVariant: () => getDefaultVariant(),
     };
 
     const config = {
@@ -70,7 +71,7 @@ test('should load experimental flags', () => {
         isEnabled: () => {
             return false;
         },
-        getVariant: () => undefined,
+        getVariant: () => getDefaultVariant(),
     };
 
     const config = {
@@ -91,7 +92,7 @@ test('should load experimental flags from external provider', () => {
                 return true;
             }
         },
-        getVariant: () => undefined,
+        getVariant: () => getDefaultVariant(),
     };
 
     const config = {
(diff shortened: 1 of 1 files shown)


src/lib/util/flag-resolver.ts
```diff
@@ -22,13 +23,16 @@ export default class FlagResolver implements IFlagResolver {
         const flags: IFlags = { ...this.experiments };
 
         Object.keys(flags).forEach((flagName: IFlagKey) => {
-            if (!this.experiments[flagName]) {
-                if (typeof flags[flagName] === 'boolean') {
+            const flag = flags[flagName];
+            if (typeof flag === 'boolean') {
+                if (!flag) {
```
gastonfournier, line 28: Wouldn't this prevent the flag to be evaluated again? If it's turned true once, it seems you can't disable it because you'll skip the `isEnabled` evaluation

src/lib/util/flag-resolver.ts
```diff
@@ -22,13 +23,16 @@ export default class FlagResolver implements IFlagResolver {
...
                     flags[flagName] = this.externalResolver.isEnabled(
                         flagName,
                         context,
                     );
-                } else {
+                }
+            } else {
+                if (!flag?.enabled) {
```
gastonfournier, line 35: Should we still resolve `getVariant` even if it's not enabled? It would give us the default variant

frontend/src/utils/variants.ts
```diff
@@ -18,7 +18,8 @@ export interface Variant {
 export const getVariantValue = <T = string>(
     variant: Variant | undefined
 ): T | undefined => {
-    if (variant?.payload !== undefined) {
+    if (variant?.enabled) {
+        if (!variant.payload) return variant.name as T;
```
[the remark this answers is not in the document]
nunogois: That was the intention, yes. In case you don't specify a payload, the only thing you can use to identify the variant and runtime control, effectively its value in the perspective of this helper, turns out to be its name.

src/lib/util/flag-resolver.ts
nunogois: If I understood the original reasoning, our `flag` here is the default value we get from either `experimental` or `server-dev`, not the current state of the flag. 

What we're saying here is, unless one those two places above override the flag by setting it as something `truthy` (e.g. env variable), we want to fetch it from the external resolver.

src/lib/util/flag-resolver.ts
nunogois: Same as [above](https://github.com/Unleash/unleash/pull/3808#discussion_r1197983537), if the initial value is `falsy` (not overridden be either `experimental` or `server-dev` to be `truthy`), we want to see what our external resolver has to say about it.

It may not be the default variant in this case. We may have a `falsy` initial value, but our external resolver (e.g. Unleash hosted) saying that this flag should be enabled, etc.

frontend/src/utils/variants.ts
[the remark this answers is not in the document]
nunogois: In case it helps visualizing what I mean in our UI:
image

This means that we could easily do something like:

```TS
const variant = getVariantValue(uiConfig.flags.ourVariant)

if (variant === 'variant-A') {
  // Do cool thing
}
```

src/lib/util/flag-resolver.ts
gastonfournier: Ok, I saw it now, and flags is a local variable of this method

gastonfournier (approved):
LG!

gastonfournier (approved):
LG!

PR #3810, opened 2023-05-19
refactor: change plausible events to be more specific at the top level

nunogois:
Splits the `demo` event into multiple more specific events so it's easier to track on Plausible (fix `(none)` in Plausible).


nicolaesocaciu (approved):
LGTM!

PR #3827, opened 2023-05-22
feat: disable bulk toggles flag

kwasniew:
## About the changes

Adding a flag to disable bulk toggle enable/disable.
In frontend we hide the buttons, in backend we return 409 conflict when the feature is disabled.

We want the feature to be enabled by default (GA feature) and have an option to turn it off if we need to (operational/killswitch toggle).


sjaanus (approved):
LG

PR #3829, opened 2023-05-22
fix: add confirmation to disable password login

nunogois:
https://linear.app/unleash/issue/2-1071/prevent-users-from-disabling-password-authentication-when-there-are-no

Improves the behavior of disabling password based login by adding some relevant information and a confirmation dialog with a warning. This felt better than trying to disable the toggle, by still allowing the end users to make the decision, except now it should be a properly informed decision with confirmation.

image

- **Password based administrators**: Admin accounts that have a password set;
- **Other administrators**: Other admin users that do not have a password. May be SSO, but may also be users that did not set a password yet;
- **Admin service accounts**: Service accounts that have the admin root role. Depending on how you're using the SA this may not necessarily mean locking yourself out of an admin account, especially if you secured its token beforehand;
- **Admin API tokens**: Similar to the above. If you secured an admin API token beforehand, you still have access to all features through the API;

Each one of them link to the respective page inside Unleash (e.g. users page, service accounts page, tokens page...);

If you try to disable and press "save", and only in that scenario, you are presented with the following confirmation dialog:

image

### DIFF
--- a/frontend/src/component/admin/auth/PasswordAuth/PasswordAuthDialog.tsx
+++ b/frontend/src/component/admin/auth/PasswordAuth/PasswordAuthDialog.tsx
@@ -1,0 +1,53 @@
+import { Alert, Typography } from '@mui/material';
+import { Dialogue } from 'component/common/Dialogue/Dialogue';
+import { IAdminCount } from 'hooks/api/getters/useAdminCount/useAdminCount';
+import { IApiToken } from 'hooks/api/getters/useApiTokens/useApiTokens';
+
+interface IPasswordAuthDialogProps {
+    open: boolean;
+    setOpen: (open: boolean) => void;
+    onClick: () => void;
+    adminCount: IAdminCount;
+    tokens: IApiToken[];
+}
+
+export const PasswordAuthDialog = ({
+    open,
+    setOpen,
+    onClick,
+    adminCount,
+    tokens,
+}: IPasswordAuthDialogProps) => (
+    <Dialogue
+        open={open}
+        onClose={() => {
+            setOpen(false);
+        }}
+        onClick={onClick}
+        title="Disable password based login?"
+        primaryButtonText="Disable password based login"
+        secondaryButtonText="Cancel"
+    >
+        <Alert severity="warning">
+            <strong>Warning!</strong> Disabling password based login may lock
+            you out of the system permanently if you do not have any alternative
+            admin credentials (such as an admin SSO account or admin API token)
+            secured beforehand.
+            <br />
+            <br />
+            <strong>Password based administrators: </strong>{' '}
+            {adminCount?.password}
+            <br />
+            <strong>Other administrators: </strong> {adminCount?.no_password}
+            <br />
+            <strong>Admin service accounts: </strong> {adminCount?.service}
+            <br />
+            <strong>Admin API tokens: </strong>{' '}
+            {tokens.filter(({ type }) => type === 'admin').length}
+        </Alert>
+        <Typography sx={{ mt: 3 }}>
+            You are about to disable password based login. Are you sure you want
+            to proceed?
+        </Typography>
+    </Dialogue>
+);
--- a/frontend/src/component/admin/auth/PasswordAuth/PasswordAuthDialog.tsx
+++ b/frontend/src/component/admin/auth/PasswordAuth/PasswordAuthDialog.tsx
@@ -38,7 +38,7 @@ export const PasswordAuthDialog = ({
             <strong>Password based administrators: </strong>{' '}
             {adminCount?.password}
             <br />
-            <strong>Other administrators: </strong> {adminCount?.no_password}
+            <strong>Other administrators: </strong> {adminCount?.noPassword}
             <br />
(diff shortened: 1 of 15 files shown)


src/lib/openapi/meta-schema-rules.test.ts
```diff
@@ -79,6 +79,7 @@ const metaRules: Rule[] = [
             },
         },
         knownExceptions: [
+            'adminCountSchema',
```
gastonfournier, line 82: I think we should avoid adding more exceptions to the meta-schema-rules and instead add the descriptions

frontend/src/hooks/api/getters/useAdminCount/useAdminCount.ts
```diff
@@ -0,0 +1,29 @@
+import useSWR from 'swr';
+import { formatApiPath } from 'utils/formatPath';
+import handleErrorResponses from '../httpErrorResponseHandler';
+
+export interface IAdminCount {
+    password: number;
+    no_password: number;
```
gastonfournier, line 7: Maybe nitpicking but should we use camelCase here? I know this is coming from the database table which means our type is coupled with the DB structure and that's fine at the store level but it shouldn't leak into the API

src/lib/services/account-service.ts
```diff
@@ -52,6 +53,10 @@ export class AccountService {
         return this.store.getAccountByPersonalAccessToken(secret);
     }
 
+    async getAdminCount(): Promise<IAdminCount> {
+        return this.store.getAdminCount();
```
gastonfournier, line 57: Here it might be a good place to translate from DB types (snake_case) into API types (camelCase)

gastonfournier (commented):
I think this is good, I just have 2 small things I'd like to improve

src/lib/db/account-store.ts
```diff
@@ -169,4 +170,27 @@ export class AccountStore implements IAccountStore {
             this.logger.error('Could not update lastSeen, error: ', err);
         }
     }
+
+    async getAdminCount(): Promise<IAdminCount> {
```
sighphyre, line 174: This doesn't respect group root roles. But that might be a intentional?

frontend/src/component/admin/auth/PasswordAuth/PasswordAuthDialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+        open={open}
+        onClose={() => {
+            setOpen(false);
+        }}
+        onClick={onClick}
+        title="Disable password based login?"
+        primaryButtonText="Disable password based login"
+        secondaryButtonText="Cancel"
+    >
+        <Alert severity="warning">
```
sighphyre, line 31: Am I reading this right if I say, we're not actually stopping people from locking themselves out and trusting that they won't click silly buttons?

src/lib/services/account-service.ts
nunogois: Addressed in https://github.com/Unleash/unleash/pull/3829/commits/ecc6f81040f6fd619842bae1552515ba26143568 - I preferred to translate it in the store layer so we use the same interface through all the layers.

src/lib/types/stores/account-store.ts
```diff
@@ -7,6 +7,12 @@ export interface IUserLookup {
     email?: string;
 }
 
+export interface IAdminCount {
+    password: number;
+    no_password: number;
```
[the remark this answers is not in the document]
nunogois: Addressed in https://github.com/Unleash/unleash/pull/3829/commits/ecc6f81040f6fd619842bae1552515ba26143568 - I preferred to translate it in the store layer so we use the same interface through all the layers.

frontend/src/component/admin/auth/PasswordAuth/PasswordAuthDialog.tsx
nunogois: Correct, we still trust our users to do the correct thing here, we're just helping them making a more informed decision.

As per the PR description:
>This felt better than trying to disable the toggle, by still allowing the end users to make the decision, except now it should be a properly informed decision with confirmation.

It also doesn't feel trivial to make a condition that completely stops people from proceeding because it depends entirely on their use case. Maybe they want their instance to be managed by an admin SA and keep an admin API token secured just in case 🤷

src/lib/openapi/meta-schema-rules.test.ts
nunogois: Fair point. Addressed in https://github.com/Unleash/unleash/pull/3829/commits/4c52804bb3ec68d675edcf884ff041093e4e3b52

gastonfournier (approved):
This LGTM!

frontend/src/component/admin/auth/PasswordAuth/PasswordAuthDialog.tsx
sighphyre: Yeah fair enough, seems like an reasonable tradeoff. Someone will absolutely lock themselves out but at least at that point we can cry on their shoulder and ask them why they would do that

sighphyre (approved):
LGTM!

PR #3830, opened 2023-05-22
chore: disable null checks until we have time to fix this

gastonfournier:
## About the changes
This action started to fail and despite being safe to ignore it's still annoying. I'm disabling it for now until I can look into what's going on with it

Example:
https://github.com/Unleash/unleash/pull/3829#issuecomment-1557372501


nunogois (approved):
LGTM

PR #3835, opened 2023-05-23
feat: add instance stats to version check

daveleek:
## About the changes

Adds feature usage info and custom strategy counters to the version check object.

Closes #


src/lib/services/version-service.ts
```diff
@@ -23,11 +26,20 @@ export interface IVersionResponse {
     latest: boolean;
 }
 
+export interface IFeatureUsageInfo extends InstanceStats {
```
sighphyre, line 29: My gut feels is that IFeatureUsageInfo is a sub interface of InstanceStats by coincidence rather than a thing that should always be true. We may very well choose to change both these data blobs in ways that make the other one sad. I would just create a new interface here with the same fields

daveleek: Yeah good point! thanks!

sighphyre (approved):
LGTM!

PR #3836, opened 2023-05-23
chore: improve joi errors

thomasheartman:
This PR improves our handling of internal Joi errors, to make them more sensible to the end user. It does that by providing a better description of the errors and by telling the user what they value they provided was.

Previous conversion:
```json
{
  "id": "705a8dc0-1198-4894-9015-f1e5b9992b48",
  "name": "BadDataError",
  "message": "\"value\" must contain at least 1 items",
  "details": [
    {
      "message": "\"value\" must contain at least 1 items",
      "description": "\"value\" must contain at least 1 items"
    }
  ]
}
```

New conversion:
```json
{
  "id": "87fb4715-cbdd-48bb-b4d7-d354e7d97380",
  "name": "BadDataError",
  "message": "Request validation failed: your request body contains invalid data. Refer to the `details` list for more information.",
  "details": [
    {
      "description": "\"value\" must contain at least 1 items. You provided [].",
      "message": "\"value\" must contain at least 1 items. You provided []."
    }
  ]
}
```

## Restructuring

This PR moves some code out of `unleash-error.ts` and into a new file. The purpose of this is twofold:
1. avoid circular dependencies (we need imports from both UnleashError and BadDataError)
2. carve out a clearer separation of concerns, keeping `unleash-error` a little more focused.

### DIFF
--- a/src/lib/error/bad-data-error.ts
+++ b/src/lib/error/bad-data-error.ts
@@ -18,7 +18,7 @@ class BadDataError extends UnleashError {
         const topLevelMessage =
             'Request validation failed: your request body contains invalid data' +
             (errors
-                ? '. Refer to the `details` list to see what happened.'
+                ? '. Refer to the `details` list for more information.'
                 : `: ${message}`);
         super(topLevelMessage);
 
--- a/src/lib/error/bad-data-error.ts
+++ b/src/lib/error/bad-data-error.ts
@@ -1,4 +1,5 @@
 import { ErrorObject } from 'ajv';
+import { ValidationError } from 'joi';
 import getProp from 'lodash.get';
 import { ApiErrorSchema, UnleashError } from './unleash-error';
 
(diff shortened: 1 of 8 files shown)


src/lib/error/bad-data-error.ts
```diff
@@ -121,3 +122,29 @@ export const fromOpenApiValidationErrors = (
...
+            ? `. You provided ${JSON.stringify(detail.context.value)}.`
+            : '.';
+        const description = detail.message + messageEnd;
+        return {
+            description,
+            message: description,
+        };
+    });
+
+    const [first, ...rest] = details;
```
nunogois, line 138: Why do we need this and check for `first` afterwards? 🤔 
Couldn't we check `details.length` on line 140 and pass in `details` in the second parameter in 143?

nunogois (approved):
It seems like we may need to update some of the existing tests to fix them, but this looks good to me!

src/lib/error/bad-data-error.ts
thomasheartman: Good question! I've tried that multiple times, but typescript doesn't accept that, I'm afraid 🤷🏼

thomasheartman:
Yeah, you're right. Those errors also provide more details now, so I've updated the tests in 30f007e5da703716145b62ca2969a48fd13e9cbc. Thanks for the review 🙏🏼

PR #3837, opened 2023-05-23
chore: lower log-level for unexpected errors

thomasheartman:
This change lowers the log level from warning to debug for when we see unexpected error types.

Right now this triggers for Joi errors, which we still rely on pretty heavily. Lowering this should clear up logs for most users.

### DIFF
--- a/src/lib/routes/util.ts
+++ b/src/lib/routes/util.ts
@@ -31,7 +31,7 @@ export const handleErrors: (
 
     if (!(error instanceof UnleashError)) {
         logger.debug(
-            `I encountered an error that wasn't an instance of the \`UnleashError\` type. This probably means that we had an unexpected crash. The original error and what it was mapped to are:`,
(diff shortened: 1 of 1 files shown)


thomasheartman:
Good question. So this is global, but there's a few things to consider:

1. This only tells us if the error wasn't spawned by one of our custom error types. With the exception of Joi validation errors and unexpected server errors (type errors, null errors, etc), they all should be. This was primarily put in place to discover cases like Joi.
2. On line 40 of the same file, we log the error's message with a warn-level regardless. So the error won't disappear from the logs. You'll just not get the big "it was something we didn't expect" message.

Does that make sense?

thomasheartman:
@sjaanus Just wanted to remind you of this one now that we're back from Poland. Any more thoughts on this?

sjaanus (approved):
LG

PR #3842, opened 2023-05-23
fix: Use the correct actor

gastonfournier:
Use the right author name as tested here https://github.com/Unleash/unleash/actions/runs/5056589150/jobs/9074236865#step:6:833


nunogois (approved):
LGTM

PR #3843, opened 2023-05-23
fix: Only show names as changed when titles have changed.

thomasheartman:
Related to [linear task 1-954](https://linear.app/unleash/issue/1-954/disabling-last-strategy-in-change-request-shows-strikethrough).

This PR changes the display logic for showing titles as changed: it previously fell back to always being `true` if there was a custom title set for a strategy. This PR makes it so that it only shows as changed if the title has actually changed, either from one custom title to another, or to and from the display name.

To accommodate the last bit, it also shows display names with a strikethrough if the strategy had no title previously, but now it does.

Here's a number of different examples: 

image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.tsx
@@ -15,6 +15,7 @@ import { Typography, styled } from '@mui/material';
 import { IFeatureStrategy } from 'interfaces/strategy';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 import { textTruncated } from 'themes/themeStyles';
+import { useStrategies } from 'hooks/api/getters/useStrategies/useStrategies';
 
 const StyledCodeSection = styled('div')(({ theme }) => ({
     overflowX: 'auto',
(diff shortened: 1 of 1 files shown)


frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.tsx
```diff
@@ -73,37 +74,47 @@ export const StrategyTooltipLink: FC<IStrategyTooltipLinkProps> = ({
...
-                    </Truncated>
-                }
-            />
+}) => {
+    const { strategies } = useStrategies();
+    const titleHasChanged = Boolean(
+        (previousTitle && previousTitle !== change.payload.title) ||
+            (!previousTitle && change.payload.title)
+    );
+    const previousTitleOrDefault = () =>
```
Tymek, line 83: I don't think this has to be a function

Tymek (approved):

frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.tsx
thomasheartman: Doesn't have to, but that way we don't evaluate it when we don't have to 💁🏼 I'm happy to make it a variable though

thomasheartman:
Discussion point: if you give a strategy a title that matches its display name: should it be shown as a change?

I think "yes", but I'm happy to hear otherwise. It currently _does_ do that:

image

PR #3845, opened 2023-05-23
fix: new workflow name

gastonfournier:
Adopt the new workflow name: https://github.com/ivarconr/unleash-enterprise/pull/529


gastonfournier:
I skipped tests because there are no code changes

PR #3846, opened 2023-05-23
fix: move application logic to service

kwasniew:
## About the changes

Fixing problem that manifested in change requests application but the root cause was in the toggle controller:
* moving application logic to a service from a controller since it's easier to reuse from a service
* adding transaction boundaries around update of strategies so that strategy change and disabling of the toggle either both succeed of both fail but never leave the system in the inconsistent state

### DIFF
--- a/src/lib/routes/admin-api/project/project-features.ts
+++ b/src/lib/routes/admin-api/project/project-features.ts
@@ -887,13 +887,16 @@ export default class ProjectFeaturesController extends Controller {
             req.body.segmentIds = [];
         }
 
-        const updatedStrategy = await this.featureService.updateStrategy(
-            strategyId,
-            req.body,
-            { environment, projectId, featureName },
-            userName,
-            req.user,
+        const updatedStrategy = await this.startTransaction(async (tx) =>
+            this.transactionalFeatureToggleService(tx).updateStrategy(
+                strategyId,
+                req.body,
+                { environment, projectId, featureName },
+                userName,
+                req.user,
+            ),
         );
+
         res.status(200).json(updatedStrategy);
     }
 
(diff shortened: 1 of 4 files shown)


src/lib/routes/admin-api/project/project-features.ts
```diff
@@ -907,28 +910,16 @@ export default class ProjectFeaturesController extends Controller {
...
+                newDocument,
+                { environment, projectId, featureName },
+                userName,
+                req.user,
+            ),
         );
-        const feature = await this.featureService.getFeature({ featureName });
 
-        const env = feature.environments.find((e) => e.name === environment);
-        const hasOnlyDisabledStrategies = env!.strategies.every(
```
kwasniew, line 920: the ! to ignore type system in production code is not a good idea. I replaced it with env?. because I managed to get null pointer here

src/lib/routes/admin-api/project/project-features.ts
```diff
@@ -907,28 +910,16 @@ export default class ProjectFeaturesController extends Controller {
         const strategy = await this.featureService.getStrategy(strategyId);
 
         const { newDocument } = applyPatch(strategy, patch);
-        const updatedStrategy = await this.featureService.updateStrategy(
-            strategyId,
-            newDocument,
-            { environment, projectId, featureName },
-            userName,
-            req.user,
+        const updatedStrategy = await this.startTransaction(async (tx) =>
```
kwasniew, line 913: updating strategy and toggle should be part of one DB transaction

src/lib/services/feature-toggle-service.ts
```diff
@@ -510,6 +510,29 @@ class FeatureToggleService {
         return this.unprotectedUpdateStrategy(id, updates, context, userName);
     }
 
+    async optionallyDisableFeature(
```
kwasniew, line 513: this was previously in a controller but it means that change requests calling service methods don't have access to full functionality. Overall controller shouldn't orchestrate application logic, application service should do it.

src/lib/services/feature-toggle-service.ts
```diff
@@ -510,6 +510,29 @@ class FeatureToggleService {
...
+                projectId,
+                featureName,
+                environment,
+                false,
+                userName,
+            );
+        }
+    }
+
     async unprotectedUpdateStrategy(
```
kwasniew, line 536: this method is called by change requests apply method

andreas-unleash (approved):
LGTM

PR #3851, opened 2023-05-24
fix: hide password login when it's disabled

nunogois:
https://linear.app/unleash/issue/2-1085/bug-password-based-login-still-shows-on-the-login-page-even-if

Fixes a regression introduced with the changes related with #3633 where we still show the password login even though it's disabled.


thomasheartman (approved):
Yeah, this looks great! Good catch!

PR #3853, opened 2023-05-24
feat: add usage of segment in list

sjaanus:
image


src/lib/types/model.ts
```diff
@@ -402,6 +402,8 @@ export interface ISegment {
     description?: string;
     project?: string;
     constraints: IConstraint[];
+    usedInProjects?: number;
```
kwasniew, line 405: does it have to be optional? Is there any scenario where we don't know the number?

src/lib/db/segment-store.ts
```diff
@@ -97,8 +99,29 @@ export default class SegmentStore implements ISegmentStore {
 
     async getAll(): Promise<ISegment[]> {
         const rows: ISegmentRow[] = await this.db
-            .select(this.prefixColumns())
+            .select(
+                this.prefixColumns(),
+                'used_in_projects',
+                'used_in_features',
+            )
+            .countDistinct(
```
kwasniew, line 107: have you tested it with a big data set in terms of performance?

frontend/src/component/segments/SegmentTable.tsx
```diff
@@ -182,8 +182,34 @@ const COLUMNS = [
...
+    },
+    {
+        Header: 'Used in',
+        width: '60%',
         Cell: ({ value, row: { original } }: any) => (
-            <HighlightCell value={value} subtitle={original.description} />
+            <TextCell
+                sx={{
+                    color:
+                        original.usedInProjects == 0 &&
```
kwasniew, line 204: === as GH actions bots says

src/lib/db/segment-store.ts
```diff
@@ -97,8 +99,29 @@ export default class SegmentStore implements ISegmentStore {
 
     async getAll(): Promise<ISegment[]> {
         const rows: ISegmentRow[] = await this.db
-            .select(this.prefixColumns())
+            .select(
```
kwasniew, line 102: do we have any backend test that will tell us when this query breaks?

frontend/src/component/segments/SegmentTable.tsx
```diff
@@ -182,8 +182,34 @@ const COLUMNS = [
         Header: 'Name',
         accessor: 'name',
         width: '60%',
+        Cell: ({
```
kwasniew, line 185: I'd enhance existing test for a segmenttable or create a new one to test enabled and disabled color display logic

kwasniew (changes requested):

frontend/src/component/segments/SegmentTable.tsx
kwasniew: can be in a separate PR

src/lib/db/segment-store.ts
sjaanus: Analyzed this query on enterprise, it is just regular join, quite cheap.

src/lib/types/model.ts
sjaanus: Yes, only case when we know the number is the when we load the list. Throughout app, other places, we dont know it about segment

kwasniew (approved):

frontend/src/component/segments/SegmentTable.tsx
```diff
@@ -56,6 +65,10 @@ export const SegmentTable = () => {
         return segments;
     }, [segments, projectId]);
 
+    const columns = useMemo(
```
sjaanus, line 68: Is this part fine?

PR #3854, opened 2023-05-24
feat: maintenance mode disables scheduler

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/server-impl.ts
+++ b/src/lib/server-impl.ts
@@ -43,7 +43,7 @@ async function createApp(
     const db = createDb(config);
     const stores = createStores(config, db);
     const services = createServices(stores, config, db);
-    scheduleServices(services);
+    await scheduleServices(services);
 
     const metricsMonitor = createMetricsMonitor();
     const unleashSession = sessionDb(config, db);
--- a/src/lib/services/index.ts
+++ b/src/lib/services/index.ts
@@ -60,7 +60,9 @@ import ConfigurationRevisionService from '../features/feature-toggle/configurati
 import { createFeatureToggleService } from '../features';
 
 // TODO: will be moved to scheduler feature directory
-export const scheduleServices = (services: IUnleashServices): void => {
+export const scheduleServices = async (
+    services: IUnleashServices,
+): Promise<void> => {
     const {
         schedulerService,
         apiTokenService,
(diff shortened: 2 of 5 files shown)


src/lib/services/maintenance-service.ts
```diff
@@ -17,19 +18,23 @@ export default class MaintenanceService {
...
+    private schedulerService: SchedulerService;
+
     constructor(
         {
             patStore,
             eventStore,
         }: Pick<IUnleashStores, 'patStore' | 'eventStore'>,
         config: IUnleashConfig,
         settingService: SettingService,
+        schedulerService: SchedulerService,
```
kwasniew, line 30: new explicit dependency so that maintenance mode can affect the scheduler

src/lib/services/maintenance-service.ts
```diff
@@ -51,6 +56,11 @@ export default class MaintenanceService {
         setting: MaintenanceSchema,
         user: string,
     ): Promise<void> {
+        if (setting.enabled) {
```
kwasniew, line 59: this is the core of the logic for the runtime behavior

src/lib/services/index.ts
```diff
@@ -69,8 +71,13 @@ export const scheduleServices = (services: IUnleashServices): void => {
         projectService,
         projectHealthService,
         configurationRevisionService,
+        maintenanceService,
     } = services;
 
+    if (await maintenanceService.isMaintenanceMode()) {
```
kwasniew, line 77: this is the core of the logic for the startup behavior

src/lib/server-impl.ts
```diff
@@ -43,7 +43,7 @@ async function createApp(
     const db = createDb(config);
     const stores = createStores(config, db);
     const services = createServices(stores, config, db);
-    scheduleServices(services);
+    await scheduleServices(services);
```
kwasniew, line 46: waiting on maintenance mode telling us if it's on/off

src/lib/services/scheduler-service.ts
```diff
@@ -1,12 +1,17 @@
 import { Logger, LogProvider } from '../logger';
 
+export type SchedulerMode = 'active' | 'paused';
```
kwasniew, line 3: scheduler can be in paused mode now

sjaanus (approved):
LG

PR #3855, opened 2023-05-24
fix: profile should wait for loaded state before rendering

nunogois:
Small change that makes it so that the profile page waits for the loaded state before rendering.

E.g. Before this, there was a flash of the "change password" tab being visible for a very short time before the auth settings loaded.


gardleopard (approved):
LGTM

gardleopard (approved):
LGTM

PR #3873, opened 2023-05-26
Fix: laggy toggles

andreas-unleash:
This PR removes the `disabled` state for the feature toggle env switches.  

This is what causes the lag from when you toggle the switch to when it becomes available for actions again.

## About the changes

Closes #


gastonfournier (approved):
LG! Let's test it out!

PR #3880, opened 2023-05-27
security: Reject multiple successive slashes in path

chriswk:


src/test/e2e/api/auth/leading-slashes-are-stripped.e2e.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        environment: 'default',
+        projects: ['default'],
+        tokenName: 'test',
+        type: ApiTokenType.CLIENT,
+    });
+    await app.request
+        .get('/api/client/features')
+        .set('Authorization', token.secret)
+        .expect(200);
+});
```
chriswk, line 59: I know we're testing this already, so this test could probably be removed.

ivarconr (approved):

PR #3894, opened 2023-05-29
feat: change own password confirmation

kwasniew:
## About the changes

When changing your own password you should specify the old password. I decided to place old password field above the new one.

Show new password error:
Screenshot 2023-05-29 at 16 10 51

Show old password error:
Screenshot 2023-05-29 at 16 10 41

### DIFF
--- a/frontend/src/component/user/Profile/PasswordTab/PasswordTab.tsx
+++ b/frontend/src/component/user/Profile/PasswordTab/PasswordTab.tsx
@@ -32,20 +32,26 @@ export const PasswordTab = () => {
     const [password, setPassword] = useState('');
     const [confirmPassword, setConfirmPassword] = useState('');
     const [oldPassword, setOldPassword] = useState('');
-    const { changePassword, errors } = usePasswordApi();
+    const { changePassword } = usePasswordApi();
 
     const submit = async (e: SyntheticEvent) => {
         e.preventDefault();
 
         if (password !== confirmPassword) {
             return;
-        } else if (!validPassword) {
+        } else if (false && !validPassword) {
+            console.log('*** invalid password');
             setError(PASSWORD_FORMAT_MESSAGE);
         } else {
             setLoading(true);
             setError(undefined);
             setAuthenticationError(undefined);
             try {
+                console.log('**** changing', {
+                    password,
+                    confirmPassword,
+                    oldPassword,
+                });
                 await changePassword({
                     password,
                     confirmPassword,
@@ -57,6 +63,7 @@ export const PasswordTab = () => {
                     type: 'success',
                 });
             } catch (error: unknown) {
+                console.log('**** error', error);
                 const formattedError = formatUnknownError(error);
                 if (error instanceof AuthenticationError) {
                     setAuthenticationError(formattedError);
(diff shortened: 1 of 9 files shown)


src/lib/routes/admin-api/user/user.ts
```diff
@@ -167,10 +168,14 @@ class UserController extends Controller {
         res: Response,
     ): Promise<void> {
         const { user } = req;
-        const { password, confirmPassword } = req.body;
-        if (password === confirmPassword) {
+        const { password, confirmPassword, oldPassword } = req.body;
+        if (password === confirmPassword && oldPassword != null) {
```
kwasniew, line 172: this should be part of the service but didn't want to change too many things in this PR

src/lib/services/user-service.ts
```diff
@@ -362,6 +362,22 @@ class UserService {
         await this.resetTokenService.expireExistingTokensForUser(userId);
     }
 
+    async changePasswordWithVerification(
```
kwasniew, line 365: delegates to changePassword but performs extra check of old password hash

src/lib/services/user-service.ts
```diff
@@ -362,6 +362,22 @@ class UserService {
...
+        userId: number,
+        newPassword: string,
+        oldPassword: string,
+    ): Promise<void> {
+        const currentPasswordHash = await this.store.getPasswordHash(userId);
+        const match = await bcrypt.compare(oldPassword, currentPasswordHash);
+        if (!match) {
+            throw new PasswordMismatch(
```
kwasniew, line 373: this results in 401 in the error translation layer

src/lib/routes/admin-api/user/user.ts
```diff
@@ -167,10 +168,14 @@ class UserController extends Controller {
...
             this.userService.validatePassword(password);
-            await this.userService.changePassword(user.id, password);
+            await this.userService.changePasswordWithVerification(
+                user.id,
+                password,
+                oldPassword,
+            );
             res.status(200).end();
         } else {
             res.status(400).end();
```
kwasniew, line 181: not sure if we shouldn't make it 401 to be more precise

frontend/src/component/user/Profile/PasswordTab/PasswordTab.tsx
```diff
@@ -79,6 +89,18 @@ export const PasswordTab = () => {
...
                             data-loading
                         />
+                        <PasswordField
+                            data-loading
+                            label="Old password"
+                            name="oldPassword"
+                            value={oldPassword}
+                            error={Boolean(authenticationError)}
+                            helperText={authenticationError}
+                            autoComplete="new-password"
```
nunogois, line 99:
```suggestion

```

nunogois, line 99:
```suggestion
                            autoComplete="current-password"
```

src/lib/routes/admin-api/user/user.ts
nunogois, line 172: Maybe it doesn't hurt to check for truthy here instead of `!= null` so we can cover `undefined` as well. But I'm guessing `undefined` is converted to `null` anyways.

```suggestion
        if (password === confirmPassword && oldPassword) {
```

nunogois (approved):
LGTM

src/lib/routes/admin-api/user/user.ts
kwasniew: This part was test-driven and missing oldPassword is turned into null as you mentioned so I'd rather stick to the original version since I can't even trigger a test scenario for undefined to test it.

PR #3899, opened 2023-06-05
fix: author and email should be for PRs

gastonfournier:
## About the changes
Previous PR https://github.com/Unleash/unleash/pull/3871 we were supposed to change this for PRs but the change was made on `release.yaml` file. This fixes the issue


sighphyre (approved):
g

PR #3901, opened 2023-06-05
feat: autocomplete off on login password

kwasniew:
## About the changes

Please note that modern web browsers may ignore this directive (to encourage the usage of password managers). In spite of this there is a chance that
not disabling autocomplete may cause problems obtaining PCI compliance.


sjaanus (approved):
LG

PR #3902, opened 2023-06-05
chore: simplify workflows

gastonfournier:
## About the changes
This PR removes one workflow file and merges it into a similar workflow file, these are the differences:
image

This should help improving maintainability of the workflows


chriswk (approved):
I like the change, but I still have to ask if this really lowers cognitive load?
 Now we have to parse a more complicated workflow file, and pay attention to the step ids to know what's going on.

Still a :thumbsup:, because I do most heartily agree that we have too many workflows going on.

gastonfournier:
> I like the change, but I still have to ask if this really lowers cognitive load? Now we have to parse a more complicated workflow file, and pay attention to the step ids to know what's going on.
> 
> Still a +1, because I do most heartily agree that we have too many workflows going on.

@chriswk I was also concerned about the cognitive load when adding two conditional steps. I dig a bit further and came up with some changes using something similar to https://github.com/docker/metadata-action#major-version-zero

I believe with this, the cognitive load is more or less the same. LMKWYT

gardleopard (approved):
LGTM

PR #3903, opened 2023-06-05
Add Memo to AdminAlert component

talhatahir:
## About the changes
- First open source contribution to this repository
- Adding `Memo` to `AdminAlert` component as its being called in multiple places.
- Import only the necessary components: Since its only using the Alert component from @mui/material, we can import it directly instead of importing the entire library. This can help improve the bundle size of your application.


nunogois:
Hi @talhatahir, thank you for your contribution.

Unfortunately, it appears that there was a miscommunication and we were not able to connect with you in a timely manner. I regret to inform you that we no longer have the component mentioned in your PR, rendering it irrelevant. As a result, I will be closing the PR for now. Nonetheless, please know that we sincerely appreciate your effort and the intention behind it.

PR #3906, opened 2023-06-06
feat: usage on context fields in list

sjaanus:


src/lib/db/context-field-store.ts
```diff
@@ -68,18 +83,48 @@ class ContextFieldStore implements IContextFieldStore {
...
 
     async getAll(): Promise<IContextField[]> {
-        const rows = await this.db
-            .select(COLUMNS)
-            .from(TABLE)
-            .orderBy('name', 'asc');
-
-        return rows.map(mapRow);
+        if (this.flagResolver.isEnabled('segmentContextFieldUsage')) {
+            const rows = await this.db
```
sjaanus, line 87: The query looks scary, but tested and query plan is reasonable. Furthermore since the base query is getting **context fields** and there are not so many of them, it is not a issue.

src/test/e2e/api/admin/context.e2e.test.ts
```diff
@@ -1,9 +1,9 @@
 import dbInit from '../../helpers/database-init';
-import { setupApp } from '../../helpers/test-helper';
+import { IUnleashTest, setupApp } from '../../helpers/test-helper';
 import getLogger from '../../../fixtures/no-logger';
 
 let db;
-let app;
+let app: IUnleashTest;
```
sjaanus, line 6: Will add in separate PR

frontend/src/component/context/ContextList/ContextList.tsx
```diff
@@ -24,12 +24,17 @@ import { ContextActionsCell } from './ContextActionsCell/ContextActionsCell';
 import { Adjust } from '@mui/icons-material';
 import { IconCell } from 'component/common/Table/cells/IconCell/IconCell';
 import { Search } from 'component/common/Search/Search';
+import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
+import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
+import theme from 'themes/theme';
+import { Box } from '@mui/material';
 
 const ContextList: VFC = () => {
```
kwasniew, line 32: In a separate PR I'd increase the test coverage for the context list if we don't have any tests in this area. I'd only test the new behavior we're adding which is the usage numbers.

frontend/src/component/context/ContextList/ContextList.tsx
```diff
@@ -73,6 +88,30 @@ const ContextList: VFC = () => {
...
             },
+            ...(uiConfig.flags.segmentContextFieldUsage
+                ? [
+                      {
+                          Header: 'Used in',
+                          width: '60%',
+                          Cell: ({ row: { original } }: any) => (
+                              <TextCell
+                                  sx={{
+                                      color:
```
kwasniew, line 99: This logic should be touched by some unit test

frontend/src/interfaces/context.ts
```diff
@@ -4,6 +4,8 @@ export interface IUnleashContextDefinition {
     createdAt: string;
     sortOrder: number;
     stickiness: boolean;
+    usedInProjects?: number;
```
kwasniew, line 7: can we remove the optionals in the long term? I understand that in a transitional period we may do expand/contract so optionality may be needed.

src/lib/db/context-field-store.ts
```diff
@@ -16,13 +17,18 @@ const COLUMNS = [
...
+    contextFields: 'context_fields',
+    featureStrategies: 'feature_strategies',
+};
 
 type ContextFieldDB = {
     name: string;
     description: string;
     stickiness: boolean;
     sort_order: number;
+    used_in_projects?: number;
```
kwasniew, line 30: can we default to 0 and not make it optional?

src/lib/db/context-field-store.ts
kwasniew: what was the biggest data set that you run it on? E.g. have you checked context fields used in 5k features?

sjaanus: Yes, 100-200ms. Lighting fast.

frontend/src/component/context/ContextList/ContextList.tsx
sjaanus: Added

PR #3918, opened 2023-06-07
chore: avoid building frontend twice

gastonfournier:
## About the changes
Running `yarn install` without the `dist` folder will trigger a `yarn build` automatically as part of the `prepare script`

As we can see here: https://github.com/Unleash/unleash/actions/runs/5200272004/jobs/9378770279
```
[build:frontend] ✓ built in 41.10s
// more output
[build:frontend] ✓ built in 39.81s
```


thomasheartman (approved):
Nice!

PR #3921, opened 2023-06-08
feat: context field usage backend

sjaanus:
Adds possibilty to find all strategies where context field is used.
Will be used in single context field view to display usage.

### DIFF
--- a/src/lib/routes/admin-api/context.ts
+++ b/src/lib/routes/admin-api/context.ts
@@ -32,6 +32,7 @@ import { serializeDates } from '../../types/serialize-dates';
 import NotFoundError from '../../error/notfound-error';
 import { NameSchema } from '../../openapi/spec/name-schema';
 import { emptyResponse } from '../../openapi/util/standard-responses';
+import { contextFieldStrategiesSchema } from '../../openapi/spec/context-field-strategies-schema';
 
 interface ContextParam {
     contextField: string;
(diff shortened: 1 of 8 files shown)


src/test/e2e/api/admin/context.e2e.test.ts
```diff
@@ -1,13 +1,22 @@
...
 let db;
 let app: IUnleashTest;
 
 beforeAll(async () => {
     db = await dbInit('context_api_serial', getLogger);
-    app = await setupApp(db.stores);
+    app = await setupAppWithCustomConfig(db.stores, {
+        experimental: {
+            flags: {
+                segmentContextFieldUsage: true,
```
kwasniew, line 16: I don't see us using this flag anywhere in the PR

src/lib/routes/admin-api/context.ts
```diff
@@ -243,4 +262,31 @@ export class ContextController extends Controller {
...
+    async getStrategiesByContextField(
+        req: IAuthRequest<{ contextField: string }>,
+        res: Response,
+    ): Promise<void> {
+        const { contextField } = req.params;
+        const contextFields =
+            await this.contextService.getStrategiesByContextField(contextField);
+
+        // Remove unnecessary IFeatureStrategy fields from the response.
+        const contextFieldStrategies = contextFields.map((strategy) => ({
```
kwasniew, line 275: why does service return data that we don't need and forces this mapping? if the service used anywhere else?
Could a service return a fit for purpose read model instead?

kwasniew (approved):

src/test/e2e/api/admin/context.e2e.test.ts
kwasniew: also strict schema validation flag

PR #3922, opened 2023-06-08
fix: propagate http-errors as they are

thomasheartman:
This PR aims to handle the increased log alarm volume seen by the SREs.

It appears that we get a large number of alarms because a client disconnects early from the front-end API. These errors are then converted into 500s because of missing handling. These errors appear to be caused by the `http-errors` library in a dependency.

We also introduced a log line whenever we see errors now a while back, and I don't think we need this logging (I was also the one who introduced it).

The changes in this PR are specifically:
- When converting from arbitrary errors, give `BadRequestError` a 400 status code, not a 500.
- Add a dependency on `http-errors` (which is already a transitive dependency because of the body parser) and use that to check whether an error is an http-error.
- If an error is an http error, then propagate it to the user with the status code and message.
- Remove warning logs when an error occurs. This was introduced to make it easier to correlate an API error and the logs, but the system hasn't been set up for that (yet?), so it's just noise now.
- When logging errors as errors, only do that if their status code would be 500.


andreas-unleash (approved):
LGTM!

PR #3928, opened 2023-06-08
Docs: update stickiness docs

andreas-unleash:
## About the changes

Closes #


website/docs/reference/stickiness.md
```diff
@@ -25,11 +25,21 @@ Custom stickiness is available starting from Unleash Enterprise v4.
 When using [the gradual rollout strategy](../reference/activation-strategies.md#gradual-rollout) or [feature toggle variants](./feature-toggle-variants.md), you can use parameters other than the user id to calculate stickiness. More specifically, you can use any field, custom or otherwise, of the [Unleash Context](../reference/unleash-context.md) as long as you have enabled custom stickiness for these fields.
 
 :::note
-This feature is currently in beta and is not yet supported by all our SDKs. Check out the [SDK compatibility table](../reference/sdks/index.md#server-side-sdk-compatibility-table) to see what SDKs support it at the moment.
+This feature is currently generally available but is not yet supported by all our SDKs. Check out the [SDK compatibility table](../reference/sdks/index.md#server-side-sdk-compatibility-table) to see what SDKs support it at the moment.
```
thomasheartman, line 28: I suggest replacing this line and surrounding `:::` lines with this:

```suggestion
:::note SDK compatibility

Custom stickiness is supported by all of our SDKs except for the Rust SDK. You can always refer to the [SDK compatibility table](../reference/sdks/index.md#server-side-sdk-compatibility-table) for the full overview.

:::
```

The extra spacing is just for formatting reasons. The extra text after `note` gives the admonition a little header. It's only the Rust SDK that doesn't support it, so I think we could call that out explicitly to save people finding it in the table.

website/docs/reference/stickiness.md
```diff
@@ -25,11 +25,21 @@ Custom stickiness is available starting from Unleash Enterprise v4.
...
+
+## Project default stickiness
+
+To set a default stickiness for an existing project, navigate to the Edit Project screen in the UI and find the stickiness dropdown. The values that you see in the dropdown are the context fields with stickiness enabled.  Setting a default project stickiness means that every time a Gradual Rollout activation strategy or a variant is added or edited, the default (intial) setting for stickiness will be the value set at the project level.  This option is also available when creating a new project.
+
+![The Edit Project screen.  There is a dropdown for setting the default stickiness](/img/proect_default_stickiness.png)
+
+:::info
+Project default stickiness is available starting from Unleash v5.
+:::
```
thomasheartman, line 45: The content is good, but I had a few questions, and I think we could rewrite it slightly. Because this is a reference doc and not a how-to guide, the format changes a little.

```suggestion
:::info Availability

Project default stickiness was introduced in **Unleash v5**.

:::

Each project in Unleash can have its own default stickiness context field. Whenever you add a gradual rollout strategy or variants to a feature in that project, Unleash will use the configured context field as the initial value. 

Only context fields that have the [custom stickiness](unleash-context.md#custom-stickiness) option turned on can be used as default stickiness fields.

If you don't specify a default custom stickiness, the project will use the "default" stickiness option, which works as described in the [calculation section](#calculation).

You can configure project default stickiness when you create a project or by editing the project later. 

![The Edit Project screen.  There is a dropdown for setting the default stickiness](/img/project_default_stickiness.png)
```

website/static/img/proect_default_stickiness.png
thomasheartman, line 1: This file looks like it's missing a 'j' in its title. Can we rename it to:  
`website/static/img/project_default_stickiness.png`

thomasheartman (changes requested):
This looks pretty good, but I have some suggestions I'd like to run by you before we merge this. There's also one more update that we should do: in `unleash-context.md`, there is a custom stickiness section that has a similar note to the one in stickiness. I think we can do the same thing to that note: mention that only the Rust SDK doesn't support it, and link to the table.

What do you think? 😄

website/docs/reference/stickiness.md
andreas-unleash: Sounds good

andreas-unleash:
> This looks pretty good, but I have some suggestions I'd like to run by you before we merge this. There's also one more update that we should do: in `unleash-context.md`, there is a custom stickiness section that has a similar note to the one in stickiness. I think we can do the same thing to that note: mention that only the Rust SDK doesn't support it, and link to the table.
> 
> What do you think? 😄

Sounds good, found another place (activation_strategies.md) and replaced the note there as well

thomasheartman (approved):
Very good! Thank you 👏🏼

PR #3929, opened 2023-06-08
fix: add createdAt in projects API response

ivarconr:
This PR adds the "createdAt" field to the /api/projects response, so that we are compliant with the schema.

### DIFF
--- a/src/test/e2e/api/admin/projects.e2e.test.ts
+++ b/src/test/e2e/api/admin/projects.e2e.test.ts
@@ -1,0 +1,31 @@
+import dbInit, { ITestDb } from '../../helpers/database-init';
+import { IUnleashTest, setupApp } from '../../helpers/test-helper';
+import getLogger from '../../../fixtures/no-logger';
+let db: ITestDb;
+let app: IUnleashTest;
+
+beforeAll(async () => {
+    db = await dbInit('projects_api_serial', getLogger);
+    app = await setupApp(db.stores);
+});
+
+afterAll(async () => {
+    await app.destroy();
+    await db.destroy();
+});
+
+test('response should include created_at', async () => {
+    const { body } = await app.request
+        .get('/api/admin/projects')
+        .expect('Content-Type', /json/)
+        .expect(200);
+    expect(body.projects[0].createdAt).toBeDefined();
+});
+
+test('response for default project should include created_at', async () => {
+    const { body } = await app.request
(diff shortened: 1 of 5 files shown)


thomasheartman (approved):
Looks good to me! 

But what about the `mapRow` function on line 555 of `src/lib/db/project-store.ts`? It also doesn't include `createdAt`. Is this something we should update too? I'm not sure exactly where it's called from (transitively).

src/test/e2e/api/admin/projects.e2e.test.ts
```diff
@@ -0,0 +1,31 @@
+import dbInit, { ITestDb } from '../../helpers/database-init';
+import { IUnleashTest, setupApp } from '../../helpers/test-helper';
+import getLogger from '../../../fixtures/no-logger';
+let db: ITestDb;
+let app: IUnleashTest;
+
+beforeAll(async () => {
+    db = await dbInit('projects_api_serial', getLogger);
+    app = await setupApp(db.stores);
```
sjaanus, line 9: It would be nice to test the schema within the test.

```
 app = await setupAppWithCustomConfig(db.stores, {
      experimental: {
          flags: {
              strictSchemaValidation: true,
          },
      },
  });
```

src/test/e2e/api/admin/projects.e2e.test.ts
```diff
@@ -0,0 +1,31 @@
+import dbInit, { ITestDb } from '../../helpers/database-init';
+import { IUnleashTest, setupApp } from '../../helpers/test-helper';
+import getLogger from '../../../fixtures/no-logger';
+let db: ITestDb;
+let app: IUnleashTest;
+
+beforeAll(async () => {
```
sjaanus, line 7: Any reason why not using the existing `projects.e2e.test.ts`?

ivarconr: No, I did not see it, sense it was in a separate folder. I will fix.

src/test/e2e/api/admin/projects.e2e.test.ts
ivarconr: Ah, nice. I will take a look.

thomasheartman:
> I am confused 🤔 does it not map the createdAt on line 564?

Oh, sorry, **I'm** confused. I must've missed that (even though I looked for it specifically). My bad! As you were 😅

PR #3938, opened 2023-06-09
feat: context field usage frontend

sjaanus:
image
image

### DIFF
--- a/frontend/src/component/context/ContextFieldUsage/ContextFieldUsage.test.tsx
+++ b/frontend/src/component/context/ContextFieldUsage/ContextFieldUsage.test.tsx
@@ -1,7 +1,6 @@
 import { render } from '../../../utils/testRenderer';
 import { screen } from '@testing-library/react';
 import React from 'react';
-import { SegmentTable } from './SegmentTable';
 import { testServerRoute, testServerSetup } from '../../../utils/testServer';
 import { UIProviderContainer } from '../../providers/UIProvider/UIProviderContainer';
 import { ContextFieldUsage } from './ContextFieldUsage';
--- a/frontend/src/component/context/ContextFieldUsage/ContextFieldUsage.test.tsx
+++ b/frontend/src/component/context/ContextFieldUsage/ContextFieldUsage.test.tsx
@@ -1,0 +1,56 @@
+import { render } from '../../../utils/testRenderer';
+import { screen } from '@testing-library/react';
+import React from 'react';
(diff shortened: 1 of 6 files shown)


src/lib/db/segment-store.ts
```diff
@@ -240,23 +240,21 @@ export default class SegmentStore implements ISegmentStore {
...
             constraints: row.constraints,
             createdBy: row.created_by,
             createdAt: row.created_at,
-            usedInProjects: row.used_in_projects
-                ? Number(row.used_in_projects)
-                : 0,
-            usedInFeatures: row.used_in_projects
-                ? Number(row.used_in_features)
-                : 0,
+            ...(row.used_in_projects && {
```
kwasniew, line 251: why not 0 by default?

frontend/src/component/context/ContextFieldUsage/ContextFieldUsage.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    <Link
+                        to={`/projects/${projectId}`}
+                        target="_blank"
+                        rel="noreferrer"
+                    >
+                        {projects.find(({ id }) => id === projectId)?.name ??
+                            projectId}
+                    </Link>
+                    <ul>
+                        {strategies
```
kwasniew, line 45: I'd prepare projectStrategies variable outside of JSX to minimize scripting inside HTML tags. I'd only use map operation in JSX, not filter

frontend/src/component/context/ContextFieldUsage/ContextFieldUsage.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    const contextFieldName = 'appName';
+    render(
+        <UIProviderContainer>
+            <ContextFieldUsage contextName={contextFieldName} />
+        </UIProviderContainer>
+    );
+
+    await screen.findByText('Usage of this context field:');
+    await screen.findByText('tests (Gradual rollout)');
```
kwasniew, line 54: where do you check project and feature name where the feature is used?

kwasniew (approved):
Minor comments

src/lib/db/segment-store.ts
sjaanus: The same mapper method is used when loading **single context field**, but we are using  this value only **in list view**, so it will confuse customers to have the count, but 0 values.

PR #3942, opened 2023-06-09
refactor: playground in feature oriented architecture

kwasniew:
## About the changes

Migrated playground backend to feature-oriented architecture. Now all backend files for this feature are in one directory so that we don't have to jump between 3 different locations (routes, services and utils).

Screenshot 2023-06-09 at 12 04 34


thomasheartman (approved):
Slick 💯

PR #3947, opened 2023-06-09
fix: fix variant metrics

sjaanus:
Currently, Unleash is storing metrics every 5 seconds. However, a problem arises when multiple SDKs query the same feature flag, and one of them sends variants data while the other does not. If the SDK that does not send variants data is registered first, it leads to a error in the system.

This PR fixes the issue, by always storing variants object.

### DIFF
--- a/src/lib/util/collapseHourlyMetrics.test.ts
+++ b/src/lib/util/collapseHourlyMetrics.test.ts
@@ -108,3 +108,48 @@ test('collapseHourlyMetrics', () => {
         },
     ]);
 });
+
+test('collapseHourlyMetrics variants', () => {
+    const timestamp = startOfHour(new Date());
+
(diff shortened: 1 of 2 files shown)


src/lib/util/collapseHourlyMetrics.test.ts
```diff
@@ -108,3 +108,48 @@ test('collapseHourlyMetrics', () => {
         },
     ]);
 });
+
+test('collapseHourlyMetrics variants', () => {
+    const timestamp = startOfHour(new Date());
+
+    const metricAX1: IClientMetricsEnv = {
```
kwasniew, line 115: I'd call this one metricsWithoutVariants and the other two metricsWithVariants

kwasniew (approved):

PR #3949, opened 2023-06-09
feat: walking skeleton of the advanced playground

kwasniew:
## About the changes

* calling new playground "advanced playground" since it supports advanced search by multiple envs and combinations of context field values
* new path /playground/advanced (added advanced suffix to the existing URL)
* added a release toggle called advancedPlayground
* returning hardcoded response for now to get feedback from UI
* modelling disabled toggles as 409 -  "indicates a request conflict with the current state of the target resource" to express that the current state of the resource is unavailable for the end user and should not be requests at this point in time
* fixing a bug with missing project in the old playground resource

Next PRs:
* describe input and output schemas
* start calling real service to calculate the results


src/lib/features/playground/playground.ts
```diff
@@ -58,7 +71,7 @@ export default class PlaygroundController extends Controller {
         const response = {
             input: req.body,
             features: await this.playgroundService.evaluateQuery(
-                req.body.projects,
+                req.body.projects || '*',
```
kwasniew, line 74: fixing a bug we had before when no projects are specified

thomasheartman (approved):
This looks good to me! Super work 😄  My only thought before merging this is whether we should discuss the URL first. I'm not entirely sold on /playground/advanced, but I'm not sure whether this needs to be handled yet?

kwasniew:
@thomasheartman Since we're not advertising it as stable API yet and it's behind a flag the name is up for a debate. Want to unblock the UI work first and we can rename it later.

thomasheartman:
That's fair. I noticed there's no OpenAPI docs for it yet, so that's ... kinda good? that said, the `Unstable` tag was made for things like this, so that should allow us to put it in the docs and work with it.

kwasniew:
@thomasheartman We will be adding OpenAPI docs after we see the data shape is useful.

Issue #3580, opened 2023-04-20
"Edit application" page fails validation


dylannz:
### Describe the bug

When trying to edit any of the 3 available attributes on the "Edit application" page (icon, application URL, description) within the web panel, I get the following error:

<img width="455" alt="Screen Shot 2023-04-21 at 11 30 02 AM" src="https://user-images.githubusercontent.com/6373352/233507490-c573d08f-2d12-4d89-95d0-b184675dbd20.png">

Digging a bit deeper, I can see that the server is returning an HTTP 400 status with the following error in the response body:

```
{"error":"Request validation failed","validation":[{"keyword":"required","dataPath":".body","schemaPath":"#/components/schemas/applicationSchema/required","params":{"missingProperty":"appName"},"message":"should have required property 'appName'"}]}
```

But as you can see on the form, appName is not a field that is visible on the page:

<img width="946" alt="Screen Shot 2023-04-21 at 11 31 08 AM" src="https://user-images.githubusercontent.com/6373352/233507602-3ab5397f-d35d-4327-81b1-2cfc0fb2c180.png">

So there is no way for the user to fix this. Maybe some of these fields are normally configured via the SDK configuration within the code and nobody usually touches this part of the app? Either way, seems like a bug!

### Steps to reproduce the bug

Configure ruby on rails application to use unleash SDK like so (where unleash is installed on https://example.com):

**config/initializers/unleash.rb**:
```
Unleash.configure do |config|
  config.app_name            = 'MyRailsApplication'
  config.url                 = 'https://example.com/api'
  config.custom_http_headers = {'Authorization': ENV['UNLEASH_SERVER_API_KEY']}
end

UNLEASH = Unleash::Client.new
```

Start up the application, then navigate to:
https://example.com/applications/MyRailsApplication

Click the "Edit application" tab

Try to change any of the 3 fields displayed (icon, application URL, description)

Then you will see the error

### Expected behavior

No error

### Unleash version

4.22.5

### Subscription type

Open source

### Hosting type

Self-hosted

### SDK information (language and version)

unleash-client-ruby:4.4.2

andreas-unleash:
Hi, @dylannz ! Thanks for bringing this to our attention. I will forward it to the right team and get back to you.

gastonfournier:
I've been able to reproduce this and I've raised a bug internally to look into it in our upcoming cycle

PR #3953, opened 2023-06-11
fix: allow empty appName as it may come in the url

gastonfournier:
## About the changes
Edit application under https://app.unleash-hosted.com/demo/applications/test-app is currently not working as the appName is expected to come in the request body, but it's actually part of the url. We have two options here:
1. We change the UI to adapt to the expectations of the request by adding appName to the request body (and eventually removing appName from the URL, which would be a breaking change)
2. We remove the restriction of only sending the appName in the body and take the one that comes in the URL. We have a validation that verifies that at least one of the two sets the appName ([here](https://github.com/Unleash/unleash/blob/e376088668428b3ec9016de9fd481fbbb03ad5cb/src/lib/services/client-metrics/instance-service.ts#L208) we validate using [this schema](https://github.com/Unleash/unleash/blob/e376088668428b3ec9016de9fd481fbbb03ad5cb/src/lib/services/client-metrics/schema.ts#L55-L70))

In terms of REST API, we can assume that the appName will be present in the resource `/api/admin/metrics/applications` (an endpoint we don't have), but when we're updating an application we should refer to that application by its URL: `/api/admin/metrics/applications/<appName>` and the presence of an appName in the body might indicate that we're trying to change the name of the application (something we currently not support)

Based on the above, I believe going with the second option is best, as it adheres to REST principles and does not require a breaking change. Despite that, we only support updating applications as the creation is done from metrics ingestion

Fixes: #3580

### DIFF
--- a/src/lib/openapi/spec/create-application-schema.ts
+++ b/src/lib/openapi/spec/create-application-schema.ts
@@ -5,7 +5,6 @@ export const createApplicationSchema = {
     type: 'object',
     additionalProperties: true,
     description: 'Reported application information from Unleash SDKs',
-    required: [],
     properties: {
         appName: {
             description: 'Name of the application',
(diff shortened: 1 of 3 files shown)


chriswk (approved):
Nice catch Gaston. :+1:

PR #3954, opened 2023-06-12
docs: remove API doc clean script

thomasheartman:
Now that we have fixed the issue with server URLs prefixing the API
paths (#3780), we can remove the cleaning script.


thomasheartman:
Turns out I was wrong. We still need this. 

Otherwise, the openapi plugin concatenates the server URL to each path, like this:

image

PR #3958, opened 2023-06-12
fix: feature not found in project yields 404

### LINKED ISSUE #3726: API returns 403 - Forbidden when toggle does not exist
Dear Unleash team, 

I noticed a peculiar behavior using Unleash server `v4.23.2` and `v5`. When doing a GET request to `<unleash-url>/api/admin/projects/:projectId/features/:featureName` if the toggle does not exist, the API returns 403 - Forbidden (I am using an admin API token), see:

Toggle exists (200 OK):
image

Toggle does not exist (403 - Forbidden)
image

The error message is a bit misleading too.

- Perform a GET request to `<unleash-url>/api/admin/projects/:projectId/features/:featureName` by replacing :projectId with an existing project ID and :featureName with a feature toggle that does not exist in the db.
- See 403 - Forbidden error in return 

### Expected behavior

The API should return 404 - Not Found

### Logs, error output, etc.

```bash
{
    "id": "f73da250-2dec-4516-a31b-a3bec40df943",
    "name": "InvalidOperationError",
    "message": "The operation could not be completed. The feature exists, but the provided project id (\"default\") does not match the project that the feature belongs to (\"undefined\"). Try using \"undefined\" in the request URL instead of \"default\".",
    "details": [
        {
            "message": "The operation could not be completed. The feature exists, but the provided project id (\"default\") does not match the project that the feature belongs to (\"undefined\"). Try using \"undefined\" in the request URL instead of \"default\".",
            "description": "The operation could not be completed. The feature exists, but the provided project id (\"default\") does not match the project that the feature belongs to (\"undefined\"). Try using \"undefined\" in the request URL instead of \"default\"."
        }
    ]
}
```
```

### Screenshots
See above

### Additional context
N/A

### Unleash version
v5

### Subscription type
Open source

### Hosting type
Self-hosted

N/A

### PR DESCRIPTION
**Pull request by gastonfournier**
## About the changes
When a feature is not found in a project we should fail with a NotFound error. If the feature belongs to a different project, it should not be a permission issue, because the user might not be aware (lack of permissions/visibility) of that other project, so even in this case the error should be NotFound (this also works if we ever allow the same feature name in different projects)

Fixes #3726

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -207,10 +207,10 @@ class FeatureToggleService {
 
         if (id !== projectId) {
             throw new NotFoundError(
-                `There's no feature named ${featureName} in project ${projectId} ${
+                `There's no feature named "${featureName}" in project "${projectId}"${
                     id === undefined
                         ? '.'
-                        : `, but there's a feature by that name in project "${id}"`
+                        : `, but there's a feature with that name in project "${id}"`
                 }`,
             );
         }
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -206,8 +206,12 @@ class FeatureToggleService {
         const id = await this.featureToggleStore.getProjectId(featureName);
 
         if (id !== projectId) {
-            throw new InvalidOperationError(
-                `The operation could not be completed. The feature exists, but the provided project id ("${projectId}") does not match the project that the feature belongs to ("${id}"). Try using "${id}" in the request URL instead of "${projectId}".`,
+            throw new NotFoundError(
+                `There's no feature named ${featureName} in project ${projectId} ${
+                    id === undefined
+                        ? '.'
+                        : `, but there's a feature by that name in project "${id}"`
+                }`,
             );
         }
     }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/lib/types/stores/feature-toggle-store.ts [hunk 1]
```diff
@@ -11,7 +11,7 @@ export interface IFeatureToggleQuery {
 export interface IFeatureToggleStore extends Store<FeatureToggle, string> {
     count(query?: Partial<IFeatureToggleQuery>): Promise<number>;
     setLastSeen(toggleNames: string[]): Promise<void>;
-    getProjectId(name: string): Promise<string>;
+    getProjectId(name: string): Promise<string | undefined>;
```
gastonfournier, line 14: There's a chance that we don't find the project id if the feature does not exist at all

src/lib/services/feature-toggle-service.ts
```diff
@@ -206,8 +206,12 @@ class FeatureToggleService {
         const id = await this.featureToggleStore.getProjectId(featureName);
 
         if (id !== projectId) {
-            throw new InvalidOperationError(
-                `The operation could not be completed. The feature exists, but the provided project id ("${projectId}") does not match the project that the feature belongs to ("${id}"). Try using "${id}" in the request URL instead of "${projectId}".`,
+            throw new NotFoundError(
+                `There's no feature named ${featureName} in project ${projectId} ${
+                    id === undefined
+                        ? '.'
+                        : `, but there's a feature by that name in project "${id}"`
```
gastonfournier, line 213: Not sure if we should provide this information in the API, but I'm just keeping the message closer to the one in the left

thomasheartman, line 213:
```suggestion
                `There's no feature named "${featureName}" in project "${projectId}"${
                    id === undefined
                        ? '.'
                        : `, but there's a feature with that name in project "${id}"`
```

thomasheartman (approved):
LGTM 🙌🏼

PR #3959, opened 2023-06-12
docs: Health check endpoint

chriswk:
### What
Adds documentation for the health check endpoint.


src/lib/openapi/spec/health-check-schema.ts
```diff
@@ -3,10 +3,14 @@ import { FromSchema } from 'json-schema-to-ts';
 export const healthCheckSchema = {
     $id: '#/components/schemas/healthCheckSchema',
     type: 'object',
+    description:
+        'Used by service orchestrators to decide if instance should be marked as healthy or unhealthy',
```
thomasheartman, line 7:
```suggestion
        'Used by service orchestrators to decide whether this Unleash instance should be marked as healthy or unhealthy',
```

src/lib/openapi/spec/health-check-schema.ts
```diff
@@ -3,10 +3,14 @@ import { FromSchema } from 'json-schema-to-ts';
...
     additionalProperties: false,
     required: ['health'],
     properties: {
         health: {
+            description:
+                'The state this instance is in. GOOD if everything is ok, BAD if instance should be restarted',
```
thomasheartman, line 13:
```suggestion
                'The state this Unleash instance is in. GOOD if everything is ok, BAD if the instance should be restarted',
```

src/lib/routes/health-check.ts
```diff
@@ -31,6 +31,9 @@ export class HealthCheckController extends Controller {
                 openApiService.validPath({
                     tags: ['Operational'],
                     operationId: 'getHealth',
+                    summary: 'The status of the instance',
```
thomasheartman, line 34:
```suggestion
                    summary: 'Get instance operational status',
```

...mmaaybe?

thomasheartman (approved):
This looks good to me!  😄 I have a few small suggestions, but nothing really very important. Some clarification would be nice some times, but I'll leave it up to you 🙌🏼

src/lib/openapi/spec/health-check-schema.ts
```diff
@@ -3,10 +3,14 @@ import { FromSchema } from 'json-schema-to-ts';
...
             type: 'string',
             enum: ['GOOD', 'BAD'],
```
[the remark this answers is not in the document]
chriswk: Ok, that's creepy, I was sure I'd added that. Agreed

chriswk:
We're running into an overloaded concept with operator. This healthcheck endpoint is used in k8s, mesos, docker swarm, pm2. So any application orchestrator. Maybe that's a better word, including somehing like `meant for usage by your application orchestrator (e.g. Kubernetes, Docker Swarm, Mesos, et.al.)`

thomasheartman:
Yeah, that sounds like a good call!

PR #3961, opened 2023-06-12
docs: Instance admin 

chriswk:
### What
Adds the documentation for the instance admin endpoints.


src/lib/openapi/spec/instance-admin-stats-schema.ts
```diff
@@ -4,57 +4,123 @@ export const instanceAdminStatsSchema = {
...
+                    range: {
+                        type: 'string',
+                        description: 'A description of a time range',
+                        enum: ['allTime', '30d', '7d'],
+                        example: '30d',
+                    },
+                    count: {
+                        type: 'number',
+                        description:
+                            'The number of client applications that has been observed in this period',
```
thomasheartman, line 112:
```suggestion
                            'The number of client applications that have been observed in this period',
```

src/lib/openapi/util/create-response-schema.ts
```diff
@@ -36,6 +36,14 @@ export const createResponseSchema = (
     });
 };
 
+export const createCsvResponseSchema = (
+    schemaName: string,
+): OpenAPIV3.ResponseObject => {
+    return createResponseSchemas(schemaName, {
+        'text/csv': schemaNamed(schemaName),
```
thomasheartman, line 43: Cool! Does this convert it to CSV (for examples etc)? Does the example response show as CSV in swagger UI?

src/lib/routes/admin-api/instance-admin.ts
```diff
@@ -35,6 +38,17 @@ class InstanceAdminController extends Controller {
...
+            middleware: [
+                openApiService.validPath({
+                    tags: ['Instance Admin'],
+                    operationId: 'getInstanceAdminStatsCsv',
+                    responses: {
+                        200: createCsvResponseSchema(
+                            'instanceAdminStatsSchema',
+                        ),
+                    },
+                }),
```
thomasheartman, line 50: This should have a summary and a description.

src/lib/routes/admin-api/instance-admin.ts
```diff
@@ -46,6 +60,10 @@ class InstanceAdminController extends Controller {
                 openApiService.validPath({
                     tags: ['Instance Admin'],
                     operationId: 'getInstanceAdminStats',
+                    summary:
+                        'An overview of usage of various features of Unleash',
```
thomasheartman, line 64: Maybe we can shorten it down a bit?

```suggestion
                        'Get feature usage overview',
```

src/lib/routes/admin-api/instance-admin.ts
```diff
@@ -46,6 +60,10 @@ class InstanceAdminController extends Controller {
...
+                    description:
+                        'Lists number of users, groups, features, strategies to allow for reporting of usage for self-hosted customers',
```
thomasheartman, line 66:
```suggestion
                        'Provides statistics about various features of Unleash to allow for reporting of usage for self-hosted customers. The response contains data such as the number of users, groups, features, strategies, versions, etc.',
```

src/lib/openapi/spec/instance-admin-stats-schema.ts
```diff
@@ -4,57 +4,123 @@ export const instanceAdminStatsSchema = {
...
         },
         segments: {
             type: 'number',
+            description: 'The number of segments defined in this instance',
+            example: 19,
         },
         strategies: {
             type: 'number',
+            description: 'The number of strategies defined in this instance',
+            example: 8,
```
thomasheartman, line 80: We **could** mark all these numbers in the admin stats schema with a `minimum: 0`. That would be more accurate. However, because this is a response schema, we're not in any danger of the users submitting wrong data based on this, so it's not that important. 

What do you think? I think it'd be nice, but I'll leave it up to you.

thomasheartman (changes requested):
Overall this looks very good! 💯 

I have left a few small comments and questions that I think we should address before merging, but it should all be pretty trivial.

src/lib/openapi/util/create-response-schema.ts
chriswk: Unfortunately swagger does not show examples for CSV

src/lib/routes/admin-api/instance-admin.ts
chriswk: Indeed. Oversight from my side.

src/lib/routes/admin-api/instance-admin.ts
chriswk: I feel like we've conflated the word feature here. Maybe "Get usage statistics" instead?

thomasheartman: Yeah, that's much better 😄

src/lib/openapi/spec/instance-admin-stats-schema.ts
chriswk: Sure, I mean, when parsing this that would allow me to use unsigned types safely. Compared to how little work it is to add, happy to add it.

src/lib/openapi/util/create-response-schema.ts
thomasheartman: Hmm ... Yeah, it doesn't look like it's well supported. Can we still show examples, though? according to this stack overflow thread, you should [mark CSV payloads as strings](https://stackoverflow.com/questions/59722686/how-to-properly-document-text-csv-responses-on-swagger-yaml-docs) if I read correctly. And they can probably have examples. Right?

chriswk: This is annoying, the schema still shows the json example, but we definitely have an object here, so I'm inclined to keep it.

src/lib/routes/admin-api/instance-admin.ts
chriswk: Ended up using `Instance usage statistics`

src/lib/openapi/util/create-response-schema.ts
chriswk: yeah, which means we should just drop the `schemaNamed` here then. and just use string

chriswk: I fixed it by making the type of the response string, and then allowing the makeCsvResponseSchema take an example string. Our example is now a csv serialized object of the type we're expecting.

PR #3965, opened 2023-06-13
refactor: read project ids in memory

kwasniew:
## About the changes

* instead of making hundreds of DB calls in parallel to fetch project id for every feature we can look it up in memory

## Discussion points

* we could make it even faster if we stored input bootstrap data with project and returned it as part of the getFeatureToggleDefinitions output interface

### DIFF
--- a/src/lib/features/playground/playground-service.ts
+++ b/src/lib/features/playground/playground-service.ts
@@ -68,37 +68,31 @@ export class PlaygroundService {
                     ? new Date(context.currentTime)
                     : undefined,
             };
-            const output: PlaygroundFeatureSchema[] = await Promise.all(
-                client
-                    .getFeatureToggleDefinitions()
-                    .map(async (feature: FeatureInterface) => {
-                        const strategyEvaluationResult: FeatureStrategiesEvaluationResult =
-                            client.isEnabled(feature.name, clientContext);
+            const output: PlaygroundFeatureSchema[] = client
+                .getFeatureToggleDefinitions()
+                .map((feature: FeatureInterface) => {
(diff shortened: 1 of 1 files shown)


src/lib/features/playground/playground-service.ts
```diff
@@ -68,37 +68,31 @@ export class PlaygroundService {
...
-            );
+                    return {
+                        isEnabled,
+                        isEnabledInCurrentEnvironment: feature.enabled,
+                        strategies: {
+                            result: strategyEvaluationResult.result,
+                            data: strategyEvaluationResult.strategies,
+                        },
+                        projectId:
+                            features.find((f) => f.name === feature.name)
```
kwasniew, line 89: features list is fetched upfront and has all project ids

thomasheartman: Yeah, this seems sensible. According to the type in `src/lib/types/stores/feature-strategies-store.ts`, "project" is always a string, though. Do you think it'll ever be empty? But then again, I guess this is JS, so you never really know 💁🏼  

Just one suggestion: if this is to help with performance, would it make sense to also convert the features list into an object/map and access it with `features[feature.name]` instead of linearly searching through the list again and again? Especially if the list of features is long, we'll end up with a O(n) instead of O(nm) (where n = number of features and m = number of client features), right?

thomasheartman (approved):
Looks pretty good to me! 🙌🏼  I have inlined two questions that might be worth considering, but I'll leave it up to you.

kwasniew:
@thomasheartman I was considering O(n) optimization and since you raised it I decided to add it :)

PR #3967, opened 2023-06-13
feat: draft branch playground evaluation

kwasniew:
## About the changes

Drafting batch playground/advanced evaluations instead of a hardcoded response.

Next PR:
* define OpenAPI schema if the data makes sense in the UI
* stabilize the backend by adding tests and maybe some refactoring


src/lib/features/playground/playground-service.ts
```diff
@@ -11,12 +11,20 @@ import { FeatureStrategiesEvaluationResult } from 'lib/features/playground/featu
...
 
 type EvaluationInput = {
     features: FeatureConfigurationClient[];
     segments: ISegment[];
     featureProject: Record<string, string>;
     context: SdkContextSchema;
+    environment: string;
+};
+
+type AdvancedPlaygroundFeatureSchema = PlaygroundFeatureSchema & {
```
kwasniew, line 25: This will be openapi schema once we define it, but for now I neded to add extra context and environment info to prepare UI specific read model

src/lib/features/playground/playground-service.ts
```diff
@@ -42,34 +50,49 @@ export class PlaygroundService {
...
         );
-        return results.flat();
+        const items = results.flat();
+        const itemsByName = groupBy(items, (item) => item.name);
+        return Object.entries(itemsByName).map(([name, entries]) => {
+            const groupedEnvironments = groupBy(
+                entries,
+                (entry) => entry.environment,
+            );
+            return {
```
kwasniew, line 81: it creates the same structure as we had in the hardcoded response

src/lib/features/playground/playground-service.ts
```diff
@@ -42,34 +50,49 @@ export class PlaygroundService {
...
+                name,
+                projectId: entries[0]?.projectId,
```
kwasniew, line 83: we can take any entry in the group so I take the first one since it's always there

src/lib/features/playground/playground-service.ts
```diff
@@ -112,18 +135,22 @@ export class PlaygroundService {
                         projectId: featureProject[feature.name],
                         variant: client.getVariant(feature.name, clientContext),
                         name: feature.name,
+                        environment,
```
kwasniew, line 138: I'm passing environment and context from input to output because the new UI requires those values

andreas-unleash (approved):
LGTM

PR #3968, opened 2023-06-13
test: advanced playground

kwasniew:
## About the changes

* added test for the advanced search (e2e/component test for now so that we can do aggressive refactoring inside the code)
* removed hardcoded response because we have a test documenting example API response

Next PR:
* start adding openapi types


src/lib/features/playground/advanced-playground.test.ts
```diff
@@ -0,0 +1,254 @@
...
+let db: ITestDb;
+
+beforeAll(async () => {
+    db = await dbInit('advanced_playground', getLogger);
+    app = await setupAppWithCustomConfig(
+        db.stores,
+        {
+            experimental: {
+                flags: {
+                    advancedPlayground: true,
```
sjaanus, line 18: While we are at it, perhaps strictSchemaValidation?

sjaanus: Unless it touches too difficult schemas, like strategies etc, which need more work.

kwasniew: we'll be adding openapi in the next PR so flag will be added only then

PR #3974, opened 2023-06-13
chore: set unleash version to be either the enterprise version or OSS

gastonfournier:
## About the changes
Unleash version should be the one best representing its runtime. In this regard, enterpriseVersion trumps unleash-server version.

This version is the one that's sent as part of our metrics.


chriswk (approved):
I think you're right here. Enterprise version is more interesting than the oss version.

gastonfournier:
I'm going to merge and test. If it goes wrong I'll revert it

PR #3975, opened 2023-06-13
feat: custom root roles

nunogois:
## About the changes
Implements custom root roles, encompassing a lot of different areas of the project, and slightly refactoring the current roles logic. It includes quite a clean up.

This feature itself is behind a flag: `customRootRoles`

This feature covers root roles in:
 - Users;
 - Service Accounts;
 - Groups;

Apologies in advance. I may have gotten a bit carried away 🙈 

### Roles

We now have a new admin tab called "Roles" where we can see all root roles and manage custom ones. We are not allowed to edit or remove *predefined* roles.
image
This meant slightly pushing away the existing roles to `project-roles` instead. One idea we want to explore in the future is to unify both types of roles in the UI instead of having 2 separate tabs. This includes modernizing project roles to fit more into our current design and decisions.

Hovering the permissions cell expands detailed information about the role:
image

### Create and edit role

Here's how the role form looks like (create / edit):
image
Here I categorized permissions so it's easier to visualize and manage from a UX perspective.

I'm using the same endpoint as before. I tried to unify the logic and get rid of the `projectRole` specific hooks. What distinguishes custom root roles from custom project roles is the extra `root-custom` type we see on the payload. By default we assume `custom` (custom project role) instead, which should help in terms of backwards compatibility.

### Delete role

When we delete a custom role we try to help the end user make an informed decision by listing all the entities which currently use this custom root role:
image
~~As mentioned in the screenshot, when deleting a custom role, we demote all entities associated with it to the predefined `Viewer` role.~~ 
**EDIT**: Apparently we currently block this from the API (access-service deleteRole) with a message:
image
What should the correct behavior be?

### Role selector

I added a new easy-to-use role selector component that is present in:
 - Users 
image
 - Service Accounts
image
- Groups
image

### Role description

I also added a new role description component that you can see below the dropdown in the selector component, but it's also used to better describe each role in the respective tables:
image

I'm not listing all the permissions of predefined roles. Those simply show the description in the tooltip:
image

### Role badge

Groups is a bit different, since it uses a list of cards, so I added yet another component - Role badge:
image

I'm using this same component on the profile tab:
image

## Discussion points
 - Are we being defensive enough with the use of the flag? Should we cover more?
 - Are we breaking backwards compatibility in any way?
 - What should we do when removing a role? Block or demote?
 - Maybe some existing permission-related issues will surface with this change: Are we being specific enough with our permissions? A lot of places are simply checking for `ADMIN`;
 - We may want to get rid of the API roles coupling we have with the users and SAs and instead use the new hooks (e.g. `useRoles`) explicitly;
 - We should update the docs;
 - Maybe we could allow the user to add a custom role directly from the role selector component;


src/lib/services/access-service.ts
```diff
@@ -228,7 +247,7 @@ export class AccessService {
                 await this.store.removeRolesOfTypeForUser(
                     userId,
                     RoleType.ROOT,
-                );
+                ); // TODO
```
gastonfournier, line 250: TODO?

frontend/src/interfaces/role.ts
```diff
@@ -1,9 +1,12 @@
+import { IPermission } from './permissions';
+
 interface IRole {
     id: number;
     name: string;
     project: string | null;
     description: string;
     type: string;
+    permissions?: IPermission[];
```
gastonfournier, line 9: Here I'd add a new interface extending IRole, something like:
```typescript
interface IRoleWithPermissions extends IRole {
    permissions?: IPermission[];
}
```

gastonfournier (commented):
WIP some comments

src/lib/services/access-service.ts
```diff
@@ -467,38 +486,80 @@ export class AccessService {
     }
 
     async createRole(role: IRoleCreation): Promise<ICustomRole> {
+        const roleType =
+            role.type === CUSTOM_ROOT_ROLE_TYPE
+                ? CUSTOM_ROOT_ROLE_TYPE
+                : CUSTOM_PROJECT_ROLE_TYPE;
```
gastonfournier, line 492:
```suggestion
        // CUSTOM_PROJECT_ROLE_TYPE is assumed by default for backward compatibility
        const roleType =
            role.type === CUSTOM_ROOT_ROLE_TYPE
                ? CUSTOM_ROOT_ROLE_TYPE
                : CUSTOM_PROJECT_ROLE_TYPE;
```

gastonfournier (approved):
I think this is good enough to merge and test. There's a couple of minor things but I don't think they're blockers

src/lib/services/access-service.ts
nunogois: Addressed in https://github.com/Unleash/unleash/pull/3975/commits/4fb9eb691ff6cfbd1e449eed7eb0ed68867247f2

frontend/src/component/admin/Admin.tsx
```diff
@@ -27,8 +28,11 @@ export const Admin = () => (
         <AdminMenu />
         <Routes>
             <Route path="users" element={<UsersAdmin />} />
-            <Route path="create-project-role" element={<CreateProjectRole />} />
-            <Route path="roles/:id/edit" element={<EditProjectRole />} />
+            <Route path="project-roles/new" element={<CreateProjectRole />} />
+            <Route
+                path="project-roles/:id/edit"
+                element={<EditProjectRole />}
+            />
```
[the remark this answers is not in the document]
nunogois: Since the goal is to eventually merge both root roles and project roles on the same tab, I don't think we should prioritize this for now.

PR #3978, opened 2023-06-14
feat: Advanced playground table

andreas-unleash:
Implements the Advanced Playground Table

## About the changes

Closes # [1-1007](https://linear.app/unleash/issue/1-1007/env-aware-results-table)

### Important files

Screenshot 2023-06-14 at 15 04 08


frontend/src/component/playground/Playground/AdvancedPlaygroundResultsTable/AdvancedPlaygroundEnvironmentCell/AdvancedPlaygroundEnvironmentCell.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const [anchor, setAnchorEl] = useState<null | Element>(null);
+
+    const onOpen = (event: React.FormEvent<HTMLButtonElement>) =>
+        setAnchorEl(event.currentTarget);
+
+    const onClose = () => setAnchorEl(null);
+
+    const open = Boolean(anchor);
+
+    const enabled = value?.filter(evaluation => evaluation.isEnabled) || [];
```
kwasniew, line 43: (value || []).filter() would be simpler

kwasniew (changes requested):
I added a bunch of comments. I know many of them apply to the original playground that you copied over here but I though we can clean it up a bit. 
As commented in slack I'd definitely add tests for changing UI inputs and displaying a table (can be separate tests) since there's so much conditional logic here. 

Feel free to decide which changes you want to make now, which later and which are not needed. Also the test PR can be a separate one since this one is getting big and it's behind a flag.

kwasniew (approved):
non blocking approval

frontend/src/hooks/api/actions/usePlayground/usePlayground.ts
```diff
@@ -27,8 +27,25 @@ export const usePlaygroundApi = () => {
         }
     };
 
+    const evaluateAdvancedPlayground = async (payload: any) => {
```
kwasniew, line 30: can we have explicit type from orval here?

frontend/src/component/playground/Playground/playground.utils.ts
```diff
@@ -50,3 +59,7 @@ export const resolveResultsWidth = (
 
     return '50%';
 };
+
+export const capitalizeFirst = (string: string) => {
```
kwasniew, line 63: I can imaging this fn to be general util not just playground

PR #3979, opened 2023-06-14
docs: OpenAPI Client tag

chriswk:
### What
Adds documentation for open api endpoints tagged with `Client`.


src/lib/openapi/spec/client-feature-schema.ts
```diff
@@ -9,55 +9,67 @@ export const clientFeatureSchema = {
...
-        createdAt: {
-            type: 'string',
-            format: 'date-time',
-            nullable: true,
-        },
-        lastSeenAt: {
-            type: 'string',
-            format: 'date-time',
-            nullable: true,
+            example: 'No variants here',
```
thomasheartman, line 31: While this is funny, I'm afraid it could also be misleading. I'd suggest something (sadly) more mundane. How about this? It links together with the feature name example

```suggestion
            example: 'Enable the new payment flow',
```

src/lib/openapi/spec/client-feature-schema.ts
```diff
@@ -9,55 +9,67 @@ export const clientFeatureSchema = {
...
         },
         enabled: {
             type: 'boolean',
+            description:
+                'Whether the feature flag is enabled for the current API key or not. This is ANDed with the evaluation results of the strategies list, so if this is false, the evaluation result will always be false',
```
thomasheartman, line 36: Should this be 'environment' or 'API key'? 

```suggestion
                'Whether the feature flag is enabled for the current environment or not. This is ANDed with the evaluation results of the strategies list, so if this is false, the evaluation result will always be false',
```

src/lib/openapi/spec/client-feature-schema.ts
```diff
@@ -9,55 +9,67 @@ export const clientFeatureSchema = {
...
+            example: true,
         },
         stale: {
+            description:
+                'If this is true Unleash believes this feature toggle has been active longer than Unleash expects a toggle of this type to be active',
```
thomasheartman, line 41: Doesn't this have to be explicitly set by the users? I thought Unleash only said features *may* be stale, but that it required a user to mark them as such?

```suggestion
                'Whether this feature toggle has become [stale](https://docs.getunleash.io/reference/technical-debt#stale-and-potentially-stale-toggles "Stale toggles") or not.',
```

src/lib/openapi/spec/client-features-query-schema.ts
```diff
@@ -4,30 +4,46 @@ export const clientFeaturesQuerySchema = {
...
         tag: {
             type: 'array',
+            description: 'Features tagged with one of these tags are included',
             items: {
                 type: 'array',
                 items: {
                     type: 'string',
                 },
             },
+            example: [['simple:payment', 'simple:stripejourney']],
```
thomasheartman, line 18: Interesting. Why is it a list of lists here? It's probably how the API is designed, but it seems weird.

src/lib/openapi/spec/client-features-query-schema.ts
```diff
@@ -4,30 +4,46 @@ export const clientFeaturesQuerySchema = {
     $id: '#/components/schemas/clientFeaturesQuerySchema',
     type: 'object',
     additionalProperties: false,
+    description: 'Query parameters active for a client features request',
     properties: {
         tag: {
             type: 'array',
+            description: 'Features tagged with one of these tags are included',
```
thomasheartman, line 11: Hmm maybe something like this to make it clearer that it excludes features without the tags?

```suggestion
            description: 'Only return features tagged with at least one of the provided tags',
```

src/lib/openapi/spec/client-features-schema.ts
```diff
@@ -13,23 +13,34 @@ export const clientFeaturesSchema = {
     $id: '#/components/schemas/clientFeaturesSchema',
     type: 'object',
     required: ['version', 'features'],
+    description:
+        'Configuration data for server-side SDKs for evaluating feature flags.',
     properties: {
         version: {
             type: 'number',
+            description:
+                'A version number for the format used in the response. Most Unleash instances now return version 2, which includes segments as a separate array',
```
thomasheartman, line 22: Discussion: should this property be an enum instead of a number? Maybe, maybe not. Do you have any thoughts? If all we have is 1 and 2, then it's not very helpful to put 564, for instance. Not very important in a response schema, but interesting none the less.

src/lib/openapi/spec/override-schema.ts
```diff
@@ -5,15 +5,22 @@ export const overrideSchema = {
...
+        'An override for deciding which variant should be assigned to a user based on the context name',
     properties: {
         contextName: {
+            description:
+                'The name of the context field used to determine overrides',
             type: 'string',
+            example: 'userId',
         },
         values: {
+            description: 'Which values that should be overriden',
```
thomasheartman, line 18:
```suggestion
            description: 'The list of values that should trigger this override. If the context field (in `contextName`) has one of these values, activate the override.',
```

src/lib/openapi/spec/variant-schema.ts
```diff
@@ -5,23 +5,39 @@ export const variantSchema = {
     $id: '#/components/schemas/variantSchema',
     type: 'object',
     additionalProperties: false,
+    description:
+        'A variant allows for further separation of users into segments. See [our excellent documentation](https://docs.getunleash.io/reference/feature-toggle-variants#what-are-variants) for a more detailed description',
```
thomasheartman, line 9: I feel like saying "our **excellent** documentation" is a little too ... grand (being a Norwegian and all).
But also, the link should explain what bit of the docs its pointing to. And we should also try to avoid using the term "see" to be more inclusive towards vision-impaired people 💁🏼 How about this?

```suggestion
        'A variant allows for further separation of users into segments. Check out the [variants documentation](https://docs.getunleash.io/reference/feature-toggle-variants#what-are-variants) for a more detailed description',
```

src/lib/openapi/spec/variant-schema.ts
```diff
@@ -5,23 +5,39 @@ export const variantSchema = {
...
     required: ['name', 'weight'],
     properties: {
         name: {
             type: 'string',
+            description: 'The variants name. Is unique for this feature toggle',
```
thomasheartman, line 14:
```suggestion
            description: 'The variant's name. The name will be unique for this feature toggle',
```

src/lib/routes/client-api/register.ts
```diff
@@ -41,6 +41,9 @@ export default class RegisterController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Client'],
+                    summary: 'Register a client SDK',
+                    description:
+                        'Called by SDKs on startup to tell Unleash about their existence. Used to track custom strategies in use as well as SDK versions.',
```
thomasheartman, line 46:
```suggestion
                        'Register a client SDK with Unleash. SDKs call this endpoint on startup to tell Unleash about their existence. Used to track custom strategies in use as well as SDK versions.',
```

thomasheartman (approved):
Very nice work here! A lot of changes, so it's super good to see it done 💯  I've added a few suggestions that should probably be looked at before merging, but I trust you to do that without needing a second approval. Well done 😄

src/lib/openapi/spec/client-application-schema.ts
```diff
@@ -4,33 +4,53 @@ export const clientApplicationSchema = {
...
         environment: {
+            description: 'Which environment the SDK is active for',
             type: 'string',
+            example: 'development',
         },
         interval: {
             type: 'number',
+            description:
+                'How often (in seconds) does the client refresh its toggles',
+            example: 10,
```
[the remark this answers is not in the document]
chriswk: Agreed. Thanks for the suggestion, applied

PR #3980, opened 2023-06-14
test: meta schema rules should not check description on ref

kwasniew:
## About the changes

Our meta rules checker will only verify description is present when the field is not a ref based field which has its own description.

This change automatically fixes a few schemas.

### DIFF
--- a/src/lib/openapi/meta-schema-rules.test.ts
+++ b/src/lib/openapi/meta-schema-rules.test.ts
@@ -68,6 +68,7 @@ const metaRules: Rule[] = [
                 properties: {
                     type: 'object', // properties of the schema should be an object
                     additionalProperties: {
+                        // with the following shape
                         anyOf: [
                             {
                                 type: 'object',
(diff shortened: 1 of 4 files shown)


gastonfournier (approved):
LGTM! :rocket:

thomasheartman:
I'm not sure I'm on board with changing these schemas. Have you confirmed that `$ref` descriptions get propagated correctly (in swagger UI etc)? Second, even if they do, you will sometimes want a separate description where they are referenced, so I'm not sure I agree with just removing the descriptions where you have. Have you checked what the descriptions say up against what the $ref.descriptions say?

kwasniew:
The ones I removed are the ones I added a few hours ago myself so I know they are duplicates. I did some experimentation and $ref from the included schema are the ones being shown while the description on the including side was ignored. So we're removing stuff that was already ignored.

thomasheartman:
Got it. In that case it sounds like a definite win! Good work! 🙌🏼

PR #3987, opened 2023-06-15
feat: add support for turning telemetry off with environment variable

daveleek:
## About the changes

Adds an environment variable for switching off feature telemetry in version check


sighphyre (approved):
Yeah looks sane to me

PR #3994, opened 2023-06-15
refactor: address custom root roles PR comments

nunogois:
https://linear.app/unleash/issue/2-1135/address-3975-pr-comments-by-refactoring-some-of-the-new-custom-root

This pull request addresses the majority of the comments raised in issue #3975 and lays the groundwork for unifying roles. The idea is for project roles to also be managed in the "Roles" tab, and several components, such as `RoleForm` and the `useRoleForm` can potentially be reused.

I'll leave the further investigation and implementation of unifying roles to be addressed in a separate task.

As a mostly unrelated UI fix, this also adds an arrow to the tooltip in the `RoleBadge` component.


gastonfournier (approved):
LGTM! And clearer :)

PR #3996, opened 2023-06-16
feat: expose kapi as part of docs

ivarconr:
Small poc to add ask kapi.ai to our docs. UI not great yet. 

image

And this is the widget in action:
image


website/docs/reference/api/unleash/unleash-api.info.mdx
thomasheartman, line 1: This file doesn't seem to be related to Kapa.ai at all and it's not very useful. Was it added by mistake?
image

thomasheartman (commented):
This is cool! Search doesn't seem to work in the preview though (just inifinite loading). Is that just because it's a preview or is there an issue somewhere?

image

Second, I don't think the unleash-api.info.mdx is part of this. Did you add that by mistake?

website/docs/reference/api/unleash/unleash-api.info.mdx
ivarconr: I did not add this one on purpose. Have no clue why it ended up here.

ivarconr:
> This is cool! Search doesn't seem to work in the preview though (just inifinite loading). Is that just because it's a preview or is there an issue somewhere?

jeah, that domain is not whitelisted on the kapi side.

thomasheartman (approved):
Yeah, I'm happy to get this in now 😄 I noticed that the response kapa gave you about Java still uses the `no.finn.` classes. I don't think that appears anywhere in our current docs (did a search now and didn't find it). Is that something we can help it sort out?

nunogois (approved):
UI does look weird, should we wait until it gets fixed before enabling this?
But otherwise 👍

ivarconr:
> UI does look weird, should we wait until it gets fixed before enabling this?

I think yes, but unsure of to fix it tbh.

ivarconr:
> I noticed that the response kapa gave you about Java still uses the no.finn. classes. I don't think that appears anywhere in our current docs (did a search now and didn't find it). Is that something we can help it sort out?

Nice catch. I will ask them.

ivarconr:
> UI does look weird, should we wait until it gets fixed before enabling this?

Yes it does. The Widget CSS is crashing with out CSS.

Emil-Sorensen:
Hey Unleash team! Emil from kapa.ai here. :)

Trying replicating above. Looks like it's an odd case where the unleash custom figure component clashes with global styles which affects the kapa widget. We will try to get this fixed on our end by ensuring all styles are reset and unable to affect the component.

In case you are eager to deploy it now, I was able to get it working by commenting out one line in `src/components/Figure/styles.module.css`:
```
figure {
    box-shadow: var(--ifm-global-shadow-lw);
    border-radius: var(--ifm-global-radius);
    border: var(--ifm-global-border-width) solid var(--unleash-color-gray);
    /* padding: var(--ifm-pre-padding); */
    margin-inline: 0;
}
```

thomasheartman:
Ah, nice find! Thanks for having a look. We can adjust that on our end to only target figures within the main content, so that should work fine. Otherwise, a `padding: unset` on the kappa widget fixes it too.

Thanks!

thomasheartman:
Better? 
image

Emil-Sorensen:
Awesome Thomas - thanks for the tip with `padding: unset` - will deploy a bug fix shortly on the kapa-widget. :) 

Otherwise LGTM. 👍

PR #3997, opened 2023-06-16
fix: add trial expired warning for enterprise

ivarconr:
Adds the trial expired warning for enterprise as well.

### DIFF
--- a/frontend/src/component/admin/billing/BillingDashboard/BillingPlan/BillingPlan.tsx
+++ b/frontend/src/component/admin/billing/BillingDashboard/BillingPlan/BillingPlan.tsx
@@ -82,6 +82,7 @@ export const BillingPlan: FC<IBillingPlanProps> = ({ instanceStatus }) => {
         [InstancePlan.PRO]: 80,
         [InstancePlan.COMPANY]: 0,
         [InstancePlan.TEAM]: 0,
+        [InstancePlan.ENTERPRISE]: 0,
         [InstancePlan.UNKNOWN]: 0,
         user: 15,
     };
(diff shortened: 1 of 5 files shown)


frontend/src/component/common/InstanceStatus/InstanceStatus.tsx
```diff
@@ -104,7 +104,7 @@ export const InstanceStatus: FC = ({ children }) => {
     return (
         <>
             <ConditionallyRender
-                condition={isBilling && Boolean(instanceStatus)}
+                condition={Boolean(instanceStatus)}
```
nunogois, line 107: If I understood this correctly, we want `isBilling` to only represent plans that we show the billing page for (invoices, etc). That's why we removed it from this condition and have not added ENTERPRISE to the billing plans.

I think this is a safe change, both the bar and the dialog should only get displayed in case we have specific information from the instance status (e.g. has expired).

frontend/src/component/admin/billing/BillingDashboard/BillingPlan/BillingPlan.tsx
```diff
@@ -82,6 +82,7 @@ export const BillingPlan: FC<IBillingPlanProps> = ({ instanceStatus }) => {
         [InstancePlan.PRO]: 80,
         [InstancePlan.COMPANY]: 0,
         [InstancePlan.TEAM]: 0,
+        [InstancePlan.ENTERPRISE]: 0,
```
nunogois, line 85: According to https://github.com/Unleash/unleash/pull/3997/files#r1232089348 I don't think we need this.
```suggestion
```

nunogois (approved):
LGTM, added a few comments.

frontend/src/component/common/InstanceStatus/InstanceStatus.tsx
ivarconr: > I would recommend removing isBilling from the destructor above since it's now unused.

Yes!

PR #3998, opened 2023-06-16
fix: can review CR with skip change request

kwasniew:
## About the changes

Roles with skip change request should still be able to review (approve/apply/comment) on change requests.


frontend/src/hooks/useChangeRequestsEnabled.ts
```diff
@@ -25,12 +25,25 @@ export const useChangeRequestsEnabled = (projectId: string) => {
         [JSON.stringify(data)]
     );
 
+    const isChangeRequestConfiguredForReview = React.useCallback(
```
kwasniew, line 28: same as isChangeRequestConfigured but doesn't check skip change request

Tymek (approved):

PR #3999, opened 2023-06-16
feat: roles unification

nunogois:
https://linear.app/unleash/issue/2-1137/roles-unification-on-the-ui

Root and project roles should be managed in a similar manner, which means using the same roles route and tab for both. 

Additionally, this includes a big revamp to the project roles to align them more closely with the modern and standardized custom root roles that were recently developed. They mostly use the same components.

There are still more things we want to improve and unify, but we've left some of that out of this PR due to PR size concerns.


chriswk (approved):
Great job. Also noticed the fact that this PR manages to reuse more components (visible from the fact that the delta sum is -700 lines of code)

PR #4002, opened 2023-06-19
Playground environment diff table

kwasniew:
## About the changes

* Environment diff table with context fields and enabled/disabled status in each environment
* All columns are sortable
* AppName is also displayed even though originally I was considering removing it, but users can override it and it would be confusing to hide it

Screenshot 2023-06-19 at 11 02 00

Next PRs:
* test for the table
* add info popover for each enabled/disabled label to see why it was enabled/disabled


frontend/src/component/playground/Playground/AdvancedPlayground.tsx
```diff
@@ -24,7 +24,7 @@ import { AdvancedPlaygroundResponseSchema } from 'openapi';
 export const AdvancedPlayground: VFC<{}> = () => {
     const { environments: availableEnvironments } = useEnvironments();
     const theme = useTheme();
-    const matches = useMediaQuery(theme.breakpoints.down('lg'));
+    const matches = true;
```
kwasniew, line 27: UI breaks on limited space we have in the right area so forcing mobile view for now

frontend/src/component/playground/Playground/PlaygroundEnvironmentTable/PlaygroundEnvironmentTable.tsx
```diff
@@ -108,6 +108,11 @@ export const PlaygroundEnvironmentTable = ({
             columns: COLUMNS as any,
             data: features,
             sortTypes,
+            autoResetGlobalFilter: false,
```
kwasniew, line 111: fixing sorting to be consistent with other views

frontend/src/component/playground/Playground/AdvancedPlayground.tsx
andreas-unleash: With `breaks`, do you mean the overflow ?

PR #4005, opened 2023-06-19
chore: Add advanced playground table test

andreas-unleash:
Adds a frontend test for AdvancedPlaygroundResultsTable
## About the changes

Closes #


frontend/src/component/playground/Playground/AdvancedPlaygroundResultsTable/AdvancedPlaygroundResultsTable.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <UIProviderContainer>
+            <AdvancedPlaygroundResultsTable
+                loading={false}
+                features={[
+                    {
+                        name: 'Infinite',
+                        projectId: 'ChangeReqs',
+                        environments: {
+                            development: [
+                                {
```
kwasniew, line 17: I'd extract a method `feature(name: string)` that is parametrized with data that matters to the test and hide the irrelevant details inside method. You can e.g have identical strategies in each feature because you don't test them

As a result of that when I read the test my eyes can focus on the data that matters and filter out irrelevant plumbing

kwasniew (approved):
Minor comment

PR #4006, opened 2023-06-19
fix: filter out usage for export

sjaanus:
Will be filtering out usage counts for export, because:

1. This is the same pattern we are doing for other objects. Filtering out not needed fields.
2. Reducing the payload. This data is not needed for importing.


kwasniew (approved):
So was it breaking the output schema?

sjaanus:
> So was it breaking the output schema?

It was not breaking output schema.

PR #4007, opened 2023-06-19
test: environment diff

kwasniew:
## About the changes

* Testing environment diff table
* Including diff table link only when more than 1 environment


frontend/src/component/playground/Playground/PlaygroundEnvironmentTable/PlaygroundEnvironmentDiffTable.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+                        {
+                            name: 'featureA',
+                            isEnabled: true,
+                            environment: 'development',
+                            context: {
+                                channel: 'web',
+                                client: 'clientA',
+                                appName: 'myapp',
+                            },
+                            ...irrelevantDetails,
```
kwasniew, line 39: I use this to separate data that matters to this test from data that only satisfies a type signature

andreas-unleash (approved):
LG

PR #4009, opened 2023-06-19
feat: implement better roles sub-tabs

nunogois:
https://linear.app/unleash/issue/2-1145/improve-roles-sub-tabs

Improves UI/UX of the roles sub-tabs.

Some of the logic is a bit specific due to the feature flag, will be nice to clean this up once we remove it.

Before:
image

After:
image


chriswk (approved):
LGTM

PR #4015, opened 2023-06-20
task: Make keepalive configurable via an environment variable

chriswk:
As requested in [Linear](https://linear.app/unleash/issue/2-1147/unleash-cloud-make-keepalive-configurable) this PR makes the serverKeepAliveTimeout configurable via the SERVER_KEEPALIVE_TIMEOUT environment variable. This was already configurable when starting Unleash programmatically, but it's nice to have as an env variable as well


gardleopard (approved):
Kickass!!!

PR #4016, opened 2023-06-20
fix: reject API admin tokens when importing features

thomasheartman:
This PR fixes an issue where trying to use an admin token to import features via the API resulted in a 500 (due to missing properties).

The solution is to catch when the user is using and admin token in the controller, throw a 400, and tell them to use personal access tokens or service accounts.

The PR includes a new test file for this specific use case. We don't really test these cases many other places, so it seemed the logical choice.

### DIFF
--- a/src/lib/features/export-import-toggles/export-import-api-usage.test.ts
+++ b/src/lib/features/export-import-toggles/export-import-api-usage.test.ts
@@ -1,139 +1,34 @@
 import { setupAppWithCustomAuth } from '../../../test/e2e/helpers/test-helper';
 import dbInit, { ITestDb } from '../../../test/e2e/helpers/database-init';
 import getLogger from '../../../test/fixtures/no-logger';
-import {
-    DEFAULT_PROJECT,
-    IContextFieldStore,
-    IEnvironmentStore,
-    IEventStore,
-    IFeatureToggleStore,
-    IProjectStore,
-    ITagStore,
-} from '../../types';
+import { DEFAULT_PROJECT } from '../../types';
 import { DEFAULT_ENV } from '../../util';
-import { ImportTogglesSchema, VariantsSchema } from '../../openapi';
+import { ImportTogglesSchema } from '../../openapi';
 import { ApiTokenType } from '../../types/models/api-token';
 import { ApiUser } from '../../server-impl';
 
 let db: ITestDb;
-let eventStore: IEventStore;
-let environmentStore: IEnvironmentStore;
-let contextFieldStore: IContextFieldStore;
-let projectStore: IProjectStore;
-let toggleStore: IFeatureToggleStore;
-let tagStore: ITagStore;
 
 beforeAll(async () => {
     db = await dbInit('export_import_api_admin', getLogger);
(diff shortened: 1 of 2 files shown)


src/lib/features/export-import-toggles/export-import-api-usage.test.ts
```diff
@@ -0,0 +1,182 @@
...
+            next();
+        });
+    };
+
+    const { request, destroy } = await setupAppWithCustomAuth(
+        db.stores,
+        preHook,
+    );
+
+    await request
```
kwasniew, line 165: usually we follow a different testing pattern:
// async call on one line
const {body} = await request.post(....).expect(400)
// sync assertions below
expect().to...

src/lib/features/export-import-toggles/export-import-api-usage.test.ts
```diff
@@ -0,0 +1,182 @@
...
+
+let db: ITestDb;
+let eventStore: IEventStore;
+let environmentStore: IEnvironmentStore;
+let contextFieldStore: IContextFieldStore;
+let projectStore: IProjectStore;
+let toggleStore: IFeatureToggleStore;
+let tagStore: ITagStore;
+
+beforeAll(async () => {
```
kwasniew, line 26: do we use all of those stores from the beforeAll setup? If not we can drop them and keep the test file smaller

src/lib/features/export-import-toggles/export-import-api-usage.test.ts
```diff
@@ -0,0 +1,182 @@
...
+    await contextFieldStore.deleteAll();
+});
+
+afterAll(async () => {
+    await db.destroy();
+});
+
+const defaultFeature = 'first_feature';
+
+const variants: VariantsSchema = [
```
kwasniew, line 52: I'd craft a minimal input to test the scenario and pass the input validation. E.g. to test the import is rejected we don't need those variants.

src/lib/features/export-import-toggles/export-import-api-usage.test.ts
```diff
@@ -0,0 +1,182 @@
...
+    ];
+const exportedStrategy: ImportTogglesSchema['data']['featureStrategies'][0] = {
+    featureName: defaultFeature,
+    id: '798cb25a-2abd-47bd-8a95-40ec13472309',
+    name: 'default',
+    parameters: {},
+    constraints,
+};
+
+const tags = [
```
kwasniew, line 98: we don't need tags to test the failing input

kwasniew (approved):
Nice handled! Minor comment: I'd remove code that's not necessary to test the scenario so that the test reader is not distracted by irrelevant details.

thomasheartman:
Yeah, great catch. I've shaved off a bunch of lines for that 😄

src/lib/features/export-import-toggles/export-import-api-usage.test.ts
thomasheartman: Fixed in f2d0702bc6b07ff7b2bac05bf050e25aac1e61dc

PR #4017, opened 2023-06-20
Query complexity validation

kwasniew:
## About the changes

This PR adds capability to reject playground queries that are too complex (over 30k results)
Screenshot 2023-06-20 at 13 10 01

Details:
* complexity calculated by multiplying number of features, envs, context combinations
* complexity is calculated in the backend code and results are returned to UI
* UI alert box is reset after another form resubmission, not when the input data is changed. You may want to see the error for some time and not have it disappear instantly after the first change

Next PR:
* UI tests for failed form submission


andreas-unleash (approved):
Nice!

PR #4018, opened 2023-06-20
feat: use new role components in project access

nunogois:
https://linear.app/unleash/issue/2-1143/adapt-project-roles-to-use-the-new-role-selector-and-role-description

This PR further unifies roles, by having a single `IRole` interface to cover both types, and re-using the same components for project roles.


gastonfournier (approved):
Really nice :broom:

PR #4020, opened 2023-06-20
docs: document how and why we collect data when using Unleash

sighphyre:
This adds documentation for how we collect data in Unleash. Previously this had been a very small amount of data but it was quite hidden in our docs in terms of what it did and why. This makes this extremely explicit, with guidelines on how to disable it and confirmation that previously disabling data collection will continue to work as expected

### DIFF
--- a/website/docs/topics/data-collection.md
+++ b/website/docs/topics/data-collection.md
@@ -24,8 +24,6 @@ This includes the following data points:
 - The number of environments defined in the system
 - The number of segments in active use by feature toggles
 - The number of custom strategies defined and in use
-- Whether the instance uses SAML (Security Assertion Markup Language) for Single Sign-On
-- Whether the instance uses OIDC (OpenID Connect) for Single Sign-On
 - The number of feature exports/imports made
 
 Please note that all collected data is anonymous, and we only collect usage counts. This data helps us understand how features are used in Unleash, enabling us to prioritize important features and make informed decisions about deprecating features that are no longer relevant to our users.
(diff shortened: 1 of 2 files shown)


website/docs/topics/data-collection.md
```diff
@@ -0,0 +1,43 @@
...
+- The total number of users in the system
+- The total number of projects in the system
+- The number of custom context fields defined and in use
+- The number of user groups defined in the system
+- The number of custom roles defined in the system
+- The number of environments defined in the system
+- The number of segments in active use by feature toggles
+- The number of custom strategies defined and in use
+- Whether the instance uses SAML (Security Assertion Markup Language) for Single Sign-On
+- Whether the instance uses OIDC (OpenID Connect) for Single Sign-On
```
daveleek, line 28: I think we concluded that we don't collect this?

website/docs/topics/data-collection.md
```diff
@@ -0,0 +1,43 @@
...
+
+Please note that we do not collect personally identifiable information (PII) through Unleash.
+
+## Managing data collection settings {#managing-data-collection-settings}
+We understand that privacy preferences may vary among our users. While the data collected by Unleash is limited and anonymous, we provide options to manage data collection settings:
+
+**Disabling All Telemetry**: If you have previously disabled the version telemetry by setting the environment variable `CHECK_VERSION` to anything other than "true," both the version telemetry and the feature telemetry will be disabled. This respects your choice to opt out of all telemetry data if you had previously disabled it.
+
+**Turning Off Feature Telemetry**: To disable the collection of the new telemetry data while still allowing the version telemetry, set the environment variable `SEND_TELEMETRY` to anything other than "true" before starting Unleash. This will ensure that the new telemetry data is not sent, but the version information is still sent.
```
daveleek, line 40: >  to anything other than "true"

We should definitely test this to be 100% sure it's the case

daveleek (changes requested):
Looks good, remove the OIDC and SAML parts then it's awesome!

website/sidebars.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
                 'topics/a-b-testing',
                 'topics/managing-constraints',
                 'topics/proxy-hosting',
+                'topics/data-collection',
```
thomasheartman, line 47: Suggest you move this up to before "managing-constraints" so that the list stays sorted alphabetically.

website/docs/topics/data-collection.md
```diff
@@ -0,0 +1,43 @@
...
+- The number of feature exports/imports made
+
+Please note that all collected data is anonymous, and we only collect usage counts. This data helps us understand how features are used in Unleash, enabling us to prioritize important features and make informed decisions about deprecating features that are no longer relevant to our users.
+
+Please note that we do not collect personally identifiable information (PII) through Unleash.
+
+## Managing data collection settings {#managing-data-collection-settings}
+We understand that privacy preferences may vary among our users. While the data collected by Unleash is limited and anonymous, we provide options to manage data collection settings:
+
+**Disabling All Telemetry**: If you have previously disabled the version telemetry by setting the environment variable `CHECK_VERSION` to anything other than "true," both the version telemetry and the feature telemetry will be disabled. This respects your choice to opt out of all telemetry data if you had previously disabled it.
```
thomasheartman, line 38: Just wondering: is the env variable case sensitive? That is, will "True" also collect metrics? We should probably be clear about that.

thomasheartman (approved):
Good work! It's clear, well structured, and makes it easy for end users to see what we collect and how they can disable it. Only got a small request to reorder the sidebar. 💯

website/docs/topics/data-collection.md
sighphyre: Yuh, shot, yeeted

website/sidebars.js
sighphyre: Thank you, good call, done!

website/docs/topics/data-collection.md
sighphyre: Ah, that was a great a question. So I went and looked. In practise this actually does 3 checks: a case sensitive check for "true", a case sensitive check for "t" and a case sensitive check for the string "1". I've updated the docs here to reflect this

website/docs/topics/data-collection.md
sighphyre: I looked at the code, it's partially true. I've updated this with the full and correct details

daveleek (approved):
LGTM!

PR #4021, opened 2023-06-20
feat: enable oas by default

kwasniew:
## About the changes
Enable open api by default for everyone.

Next steps: 
* remove enableOas option from enterprise 
* remove enableOas option in OSS (after enterprise)


ivarconr (approved):

kwasniew:
@ivarconr I will update docs in a separate PR that we'll merge when the setting goes live

PR #4023, opened 2023-06-20
Test advanced playground error

kwasniew:
## About the changes

Playground form evaluation: testing that error banner is displayed on backend returning 400 and error message.


frontend/src/component/playground/Playground/AdvancedPlayground.test.tsx
```diff
@@ -22,6 +24,18 @@ const testComponent = (
     </UIProviderContainer>
 );
 
+const testEvaluateComponent = (
+    <UIProviderContainer>
+        <AdvancedPlayground
+            FormComponent={props => (
+                <form onSubmit={props.onSubmit}>
```
kwasniew, line 31: instead of using real form with hard to test code mirror I use the test form. The important stuff is that it calls onSubmit from the parent component owning the state.

thomasheartman (approved):
Yeah, this looks pretty good to me 😄

PR #4027, opened 2023-06-21
feat: add "edit" link to playground strategies

thomasheartman:
This change adds an "edit" link to all playground strategies when they
are returned from the API, allowing the user to jump directly to the
evaluated strategy edit screen.

This change applies to both old and new strategies, so it should even
work in the old playground.

This does not use this information in the frontend yet.

## Discussion points:

Should "links" be an object or a singular string? I know the
notifications service uses just "link", but using an object will make
it easier to potentially add more actions in the future (such as
"enable"/"disable", maybe?)

Do we need to supply basePath? I noticed that the notifications links
only ever use an empty string for base path, so it might not be
necessary and can potentially be left out.

## Changes

I've implemented the link building in a new view model file. Inspecting the output after the result is fully ready requires some gnarly introspection and mapping, but it's tested.

Further, I've done a little bit of work to stop the playground service using the schema types directly as the schema types now contain extra information.

This PR also updates the `urlFriendlyString` arbitrary to not produce strings that contain only periods. This causes issues when parsing URLs (and is also something we struggle with in the UI).


src/lib/features/playground/playground-service.ts
```diff
@@ -134,7 +134,19 @@ export class PlaygroundService {
                         isEnabledInCurrentEnvironment: feature.enabled,
                         strategies: {
                             result: strategyEvaluationResult.result,
-                            data: strategyEvaluationResult.strategies,
+                            data: strategyEvaluationResult.strategies.map(
```
kwasniew, line 137: personally I prefer to enhance HTTP specific things in the view model/function preparing JSON representation since service is about application logic, not HTTP representations. I'd place the view model function call in the controller since controller is the HTTP delivery mechanism

src/lib/features/playground/playground-service.ts
```diff
@@ -147,6 +159,14 @@ export class PlaygroundService {
         }
     }
 
+    private buildStrategyLink = (
```
kwasniew, line 162: I like to test this type of code in a fast and focused unit test

kwasniew (approved):
Very nice! I'd only move HTTP links building to controller handling HTTP logic.

src/lib/features/playground/playground-service.ts
thomasheartman: Yeah, that makes sense. I'll see if that can be done without extensive code changes 🙋🏼

PR #4031, opened 2023-06-21
fix: infinite playground rendering

kwasniew:
## About the changes

Env array was causing infinite re-render


andreas-unleash (approved):
LG

andreas-unleash (approved):

PR #4032, opened 2023-06-21
fix: consider ADMIN in API tokens fetch permissions

nunogois:
https://github.com/Unleash/unleash/pull/4019 introduced a bug where API token filtering was not taking into account ADMIN permissions, which means the API tokens were not being displayed on the UI.


kwasniew (approved):

chriswk (approved):
Thank you.

PR #4040, opened 2023-06-21
fix: default segments should only be selected when using default stra…

sjaanus:


andreas-unleash (approved):
LG

frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
```diff
@@ -56,7 +56,9 @@ export const FeatureStrategyCreate = () => {
 
     const [strategy, setStrategy] = useState<Partial<IFeatureStrategy>>({});
 
-    const [segments, setSegments] = useState<ISegment[]>(strategySegments);
+    const [segments, setSegments] = useState<ISegment[]>(
+        useDefaultStrategy ? strategySegments : []
```
kwasniew, line 60: please rename useDefaultStrategy because it looks like a hook which it isn't

kwasniew (approved):

PR #4041, opened 2023-06-21
chore: add paths-ignore to more workflows

chriswk:
### What
To reduce the work needed, this PR adds paths-ignore to avoid building unnecessary workflows.


gastonfournier (approved):
LG!

PR #4047, opened 2023-06-21
feat: configurable playground limit

kwasniew:
## About the changes

Make playground evaluation limit configurable through variants so that we can experiment easily with different values.


thomasheartman (approved):
Clever solution! 🙌🏼

PR #4060, opened 2023-06-22
fix: disallow empty list of envs and invalid env names in advanced playground

thomasheartman:
This PR changes the OpenAPI schema for the advanced playground to not accept empty lists of environments and to not accept environment names that don't match the env name pattern we use.

The pattern is the same as the one we use for controlling environment names on creation.

However, there is a (small) chance that these may get out of sync later, so we could do something to only define this pattern once (and import it in the enterprise package), but that may be more work than is necessary, and I'd suggest we do that later.

I've also added a minLength to the string items although it isn't strictly necessary. It's primarily to give the users better feedback if the name is empty.


kwasniew (approved):
Thanks for finding and fixing it!

PR #4061, opened 2023-06-22
fix: multi env select should always have an environment selected

andreas-unleash:
## About the changes

Closes #


kwasniew (approved):

thomasheartman:
Does this mark the input field as required in the form too?

andreas-unleash:
> Does this mark the input field as required in the form too?

No, it adds a default value

thomasheartman:
But it *is* a required field, so I think we should indicate that on the form too. So that it actually won't submit if you somehow get into a state where it has no value.

PR #4063, opened 2023-06-22
feat: link to strategy edit screens from playground strategy results

thomasheartman:
This PR adds links from the strategy evaluation results directly to the strategy edit screen.

If the code doesn't find any links, nothing changes.

The link text will be the name of the strategy and the title (if it exists). 

image

### DIFF
--- a/frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
+++ b/frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
@@ -107,12 +107,10 @@ export const StrategyItemContainer: FC<IStrategyItemContainerProps> = ({
     const Icon = getFeatureStrategyIcon(strategy.name);
     const { uiConfig } = useUiConfig();
 
-    // @ts-expect-error IFeatureStrategy doesn't have links
-    const link = strategy.links?.edit;
-
-    const StrategyHeaderLink: React.FC = link
-        ? ({ children }) => <Link to={link}>{children}</Link>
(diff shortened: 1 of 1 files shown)


frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -106,6 +107,13 @@ export const StrategyItemContainer: FC<IStrategyItemContainerProps> = ({
     const Icon = getFeatureStrategyIcon(strategy.name);
     const { uiConfig } = useUiConfig();
 
+    // @ts-expect-error IFeatureStrategy doesn't have links
```
kwasniew, line 110: I would solve it without ignoring typechecker. When I checked out your branch none of the union elements had a links field

kwasniew (changes requested):
Something is not right with the type

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
kwasniew: AFter we get the links on one of the union element we can play around with type aliases and ternany operators so that we don't have to opt out of type safety

thomasheartman: Fair! I'll rebase off main now that the Orval updates have been merged and see if that fixes it.

kwasniew (approved):
Very elegant solution together with the backend changes you made before.

PR #4068, opened 2023-06-22
fix: fetching user root roles include custom ones

gastonfournier:
## About the changes
`getUserRootRoles` should also consider custom root roles

This introduces test cases that unveiled a dependency between stores (this happens actually at the DB layer having access-service access tables from two different stores but skipping the store layer).

https://linear.app/unleash/issue/2-1161/a-user-with-custom-root-role-and-permission-to-create-client-api

### DIFF
--- a/src/lib/features/access/createAccessService.ts
+++ b/src/lib/features/access/createAccessService.ts
@@ -45,7 +45,7 @@ export const createFakeAccessService = (
     const accountStore = new FakeAccountStore();
     const roleStore = new FakeRoleStore();
     const environmentStore = new FakeEnvironmentStore();
-    const accessStore = new FakeAccessStore();
+    const accessStore = new FakeAccessStore(roleStore);
     const groupService = new GroupService(
         { groupStore, eventStore, accountStore },
         { getLogger },
--- a/src/lib/services/access-service.ts
+++ b/src/lib/services/access-service.ts
@@ -212,7 +212,7 @@ export class AccessService {
     async addUserToRole(
         userId: number,
         roleId: number,
-        projectId?: string,
+        projectId: string,
     ): Promise<void> {
         return this.store.addUserToRole(userId, roleId, projectId);
     }
--- a/src/lib/services/access-service.ts
+++ b/src/lib/services/access-service.ts
@@ -18,7 +18,6 @@ import {
     IRoleData,
     IUserWithRole,
     RoleName,
-    RoleType,
 } from '../types/model';
 import { IRoleStore } from 'lib/types/stores/role-store';
 import NameExistsError from '../error/name-exists-error';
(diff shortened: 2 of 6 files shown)


src/lib/features/access/createAccessService.ts
```diff
@@ -45,7 +45,7 @@ export const createFakeAccessService = (
     const accountStore = new FakeAccountStore();
     const roleStore = new FakeRoleStore();
     const environmentStore = new FakeEnvironmentStore();
-    const accessStore = new FakeAccessStore();
+    const accessStore = new FakeAccessStore(roleStore);
```
gastonfournier, line 48: @kwasniew I'm reusing your `createAccessService` method which simplifies the test setup, but note I'm explicitly adding this dependency because access-store directly queries roles table: https://github.com/Unleash/unleash/blob/0457ae42ee78e9481b295bae3161d049051ededa/src/lib/db/access-store.ts#L23

To simulate this DB coupling I'm adding this dependency.

I couldn't find a better way which is why I'm tagging you in case you have some ideas

src/lib/services/access-service.ts
```diff
@@ -275,7 +275,9 @@ export class AccessService {
 
     async getUserRootRoles(userId: number): Promise<IRoleWithProject[]> {
         const userRoles = await this.store.getRolesForUserId(userId);
-        return userRoles.filter((r) => r.type === RoleType.ROOT);
+        return userRoles.filter(
+            (r) => r.type === RoleType.ROOT || r.type === RoleType.ROOT_CUSTOM,
```
gastonfournier, line 279: This is the fix

nunogois: Maybe consider `({type}) => ROOT_ROLE_TYPES.includes(type)` in case it's cleaner.

src/lib/types/model.ts
```diff
@@ -386,11 +386,11 @@ export interface IProject {
     defaultStickiness: string;
 }
 
-export interface ICustomRole {
-    id: number;
-    name: string;
+/**
+ * Extends IRole making description mandatory
+ */
```
nunogois, line 391: I would still like to drop `ICustomRole` and only have `IRole`. Description is already required in the frontend interface `frontend/src/interfaces/role.ts` and AFAIK there's always a description. 

Could be out of scope for this PR and maybe belong in a separate one though.

src/test/fixtures/fake-access-store.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
+import { P } from 'ts-toolbelt/out/Object/_api';
```
nunogois, line 2: I'm guessing this was added by mistake?
```suggestion

```

nunogois (approved):
LGTM but added a few comments.

src/lib/types/model.ts
gastonfournier: :+1: but I thought chasing down why the description was optional required a deeper investigation and analyze potential backward incompatibility problems as I see that some project role in our sandbox instance were created without a description

src/lib/services/access-service.ts
gastonfournier: Addressed in [9956e9e](https://github.com/Unleash/unleash/pull/4068/commits/9956e9e9caf3940b1eefcad3c903593b7ead9804)

nunogois:
This also fixes https://linear.app/unleash/issue/2-1160/editing-a-user-that-has-a-custom-root-role-assigned-pre-selects-the 🙌

PR #4071, opened 2023-06-22
fix: add strategy bug when strategySplittedButton flag is on

andreas-unleash:
Fixes 2 bugs after the recent strategy improvements v2 changes:
- When creating a strategy the `groupId` param of the Gradual Rollout strategy now populates the groupId (when using default strategy, the groupId will only be overwritten when it is an empty string ) with the feature name (as it was before)

- When editing/setting a default strategy for an environment the `groupId` param should be an empty string, but editable.

## About the changes

Closes #

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
@@ -33,7 +33,7 @@ import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
 import useQueryParams from 'hooks/useQueryParams';
 import useProject from 'hooks/api/getters/useProject/useProject';
 import { useSegments } from 'hooks/api/getters/useSegments/useSegments';
-import { DEFAULT_STRATEGY } from '../../../project/Project/ProjectSettings/ProjectDefaultStrategySettings/ProjectEnvironment/ProjectEnvironmentDefaultStrategy/EditDefaultStrategy';
+import { DEFAULT_STRATEGY } from 'component/project/Project/ProjectSettings/ProjectDefaultStrategySettings/ProjectEnvironment/ProjectEnvironmentDefaultStrategy/EditDefaultStrategy';
 
 export const FeatureStrategyCreate = () => {
     const projectId = useRequiredPathParam('projectId');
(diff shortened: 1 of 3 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
```diff
@@ -101,11 +101,24 @@ export const FeatureStrategyCreate = () => {
 
     useEffect(() => {
         if (shouldUseDefaultStrategy) {
-            setStrategy((defaultStrategy as any) || DEFAULT_STRATEGY);
+            const strategyTemplate = defaultStrategy || DEFAULT_STRATEGY;
+            if (
```
kwasniew, line 105: Can be simplified to 
```
if (strategyTemplate.parameters?.groupId === '' && featureId) {
     strategyTemplate.parameters.groupId = featureId;
}
```

frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
```diff
@@ -101,11 +101,24 @@ export const FeatureStrategyCreate = () => {
...
+                strategyTemplate.parameters &&
+                'groupId' in strategyTemplate.parameters &&
+                strategyTemplate.parameters.groupId === '' &&
+                featureId
+            ) {
+                strategyTemplate.parameters.groupId = featureId;
+            }
+            setStrategy(strategyTemplate as any);
```
kwasniew, line 113: as any can be avoided by changing

```
export interface IFeatureStrategy {
    ...other fields
    title?: string | null;
    disabled?: boolean | null;
}
```
Our FE code didn't match openapi schema so let's adjust them instead of ignorign types

kwasniew: I added nullable constraint ont itle and disabled as indicated by the openapi schema

kwasniew (approved):
Thenks for a quick fix. Added minor improvements suggestions in two places.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
andreas-unleash: Thanks

frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
andreas-unleash: Turns out doing the type change opens a huge can of worms. I think more important to fix the functionality. Decided to skip this for now

PR #4072, opened 2023-06-22
Feat/more granular permissions check in create apitoken

gastonfournier:
## About the changes
This PR enables or disables create API token button based on the permissions.

**Note:** the button is only displayed if you have READ permissions on some API token. This is a minor limitation as having CREATE permissions should also grant READ permissions, but right now this is up to the user to set up the custom role with the correct permissions

**Note 2:** Project-specific API tokens are also ruled by the project-specific permission to create API tokens in a project (just having the root permissions to create a client token or frontend token does not grant access to create a project-specific API token). The permissions to access the creation of a project-specific API token then rely on the root permissions to allow the user to create either a client token or a frontend token.


src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -157,6 +159,19 @@ export class ProjectApiTokenController extends Controller {
...
         const createToken = await createApiToken.validateAsync(req.body);
         const { projectId } = req.params;
+        const permissionRequired = tokenTypeToCreatePermission(
+            createToken.type,
+        );
+        const hasPermission = await this.accessService.hasPermission(
+            req.user,
+            permissionRequired,
+            projectId,
+        );
```
gastonfournier, line 169: This combines the permission to create project-specific API tokens with the root role permission that determines whether you can create CLIENT or FRONTEND tokens

gastonfournier:
> Still need to have a better look through the code but this is great!
> 
> I noticed that when we only have permissions for the FRONTEND tokens we are still selecting CLIENT tokens as the default. Maybe we should cover this by having a default behavior where we select the first type we have access to?
> 
> image

Maybe something like this: https://github.com/Unleash/unleash/pull/4075 ?

chriswk (changes requested):
You have a warning about using a react hook inside a callback here. Are you able to set it up as it recommends.

gastonfournier:
> You have a warning about using a react hook inside a callback here. Are you able to set it up as it recommends.

https://github.com/Unleash/unleash/pull/4075 should fix it

frontend/src/component/admin/apiToken/ApiTokenForm/TokenTypeSelector/TokenTypeSelector.tsx
```diff
@@ -7,45 +7,24 @@ import {
...
     type: string;
-    setType: (value: string) => void;
+    setType: (value: TokenType) => void;
+    apiTokenTypes: {
+        key: string;
+        label: string;
+        title: string;
+        permission: string | string[];
+        hasAccess: boolean;
+    }[];
```
nunogois, line 21: We could maybe extract this type.

frontend/src/component/admin/apiToken/ApiTokenForm/TokenTypeSelector/TokenTypeSelector.tsx
```diff
@@ -57,36 +36,39 @@ export const TokenTypeSelector = ({
...
-                                    sx={{
-                                        ml: 0.75,
-                                        alignSelf: 'flex-start',
-                                    }}
-                                />
-                            }
-                            label={
-                                <Box>
+                    {apiTokenTypes.map(
+                        ({ key, label, title, permission, hasAccess }) => (
```
nunogois, line 42: Unused `permission`

nunogois (approved):
LG!

frontend/src/component/admin/apiToken/ApiTokenForm/TokenTypeSelector/TokenTypeSelector.tsx
gastonfournier: After removing permission this got simplified: [6d67940](https://github.com/Unleash/unleash/pull/4072/commits/6d67940d64d9c7d24d44faf360d5fc9aad061b24)

frontend/src/component/admin/apiToken/ApiTokenForm/TokenTypeSelector/TokenTypeSelector.tsx
gastonfournier: Done in [6d67940](https://github.com/Unleash/unleash/pull/4072/commits/6d67940d64d9c7d24d44faf360d5fc9aad061b24)

chriswk (approved):
Looks good now :+1:

PR #4074, opened 2023-06-22
feat: upgrade AdminAlert to PermissionGuard

nunogois:
Upgrades our `AdminAlert` to a new `PermissionGuard`.

**Question**: We don't **need** to, but **should** we be specific about the `ADMIN` permission every time?
Technically `PermissionGuard` could have `permissions` as optional and assume `[]` by default, which will add `ADMIN` anyways. However, I feel like we gain some readability if we're specific about it. WDYT?

Single permission:
image

Multiple permissions:
image


chriswk (approved):
I concur with Gaston here. I really like this feature, it might even help people request the correct access from their manager / admin. 

Ping @thomasheartman - What do you think?

PR #4076, opened 2023-06-23
fix: remove playground results flip

kwasniew:
## About the changes

Before:
* when try configuration was clicked in playground and the results fetching was over there was a quick moment where playground guidance/instructions showed up before the result table was rendered

After:
* instruction if only shown before form submission
* after form submission loaded appears
* when loading finishes table starts rendering

### DIFF
--- a/frontend/src/component/playground/Playground/AdvancedPlayground.tsx
+++ b/frontend/src/component/playground/Playground/AdvancedPlayground.tsx
@@ -57,6 +57,7 @@ export const AdvancedPlayground: VFC<{
     const [searchParams, setSearchParams] = useSearchParams();
     const searchParamsLength = Array.from(searchParams.entries()).length;
     const { evaluateAdvancedPlayground, loading } = usePlaygroundApi();
+    const [hasFormBeenSubmitted, setHasFormBeenSubmitted] = useState(false);
 
     useEffect(() => {
         if (environments?.length === 0) {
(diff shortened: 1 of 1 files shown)


frontend/src/component/playground/Playground/AdvancedPlayground.tsx
```diff
@@ -57,6 +57,7 @@ export const AdvancedPlayground: VFC<{
     const [searchParams, setSearchParams] = useSearchParams();
     const searchParamsLength = Array.from(searchParams.entries()).length;
     const { evaluateAdvancedPlayground, loading } = usePlaygroundApi();
+    const [hasFormBeenSubmitted, setHasFormBeenSubmitted] = useState(false);
```
kwasniew, line 60: new variable to store if form was submitted

frontend/src/component/playground/Playground/AdvancedPlayground.tsx
```diff
@@ -272,17 +275,25 @@ export const AdvancedPlayground: VFC<{
...
+                                            loading={loading}
+                                            features={results?.features}
+                                            input={results?.input}
+                                        />
+                                    }
+                                />
+                                <ConditionallyRender
+                                    condition={
+                                        !Boolean(results) &&
+                                        !hasFormBeenSubmitted
```
kwasniew, line 292: this removing flipping effect

thomasheartman (approved):
This looks good to me! My only question would be if we could use data state to check whether the form has been submitted or not instead of introducing a new state variable, but I'm assuming you've already explored that possibility? But definitely non-blocking. Nice work! 🏆

kwasniew:
@thomasheartman You're right. I tried to do it w/o a new state variable but couldn't get it to work as expected.

Issue #3545, opened 2023-04-17
🟢  feat: Playground iteration 2


ivarconr:
### Describe the feature request

One particular need coming from multiple users is the ability to compare multiple values for a context fields. 

E.g. compare user1 with user2 in order to spot differences in their configurations. 

### Background

This feature needs a bit more discovery.

PR #4077, opened 2023-06-23
feat: count number of combinations from playground

thomasheartman:
This PR adds plausible tracking of the number of feature combinations that we get from the advanced playground API.

The event type has been added to plausible

 Relates to #3545

### DIFF
--- a/frontend/src/component/playground/Playground/AdvancedPlaygroundResultsTable/combinationCounter.ts
+++ b/frontend/src/component/playground/Playground/AdvancedPlaygroundResultsTable/combinationCounter.ts
@@ -1,0 +1,12 @@
+import { AdvancedPlaygroundFeatureSchema } from 'openapi';
+
+export const countCombinations = (
+    features: AdvancedPlaygroundFeatureSchema[]
+) =>
+    features.reduce(
+        (total, feature) =>
+            total +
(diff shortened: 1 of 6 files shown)


andreas-unleash (approved):
LGTM

frontend/src/component/playground/Playground/AdvancedPlaygroundResultsTable/combinationCounter.ts
```diff
@@ -0,0 +1,12 @@
+import { AdvancedPlaygroundFeatureSchema } from 'openapi';
+
+export const countCombinations = (
```
kwasniew, line 3: Maybe backend should return it?

thomasheartman: (For future visibility; as mentioned in the slack discussion)

That's a good point! I put it in the frontend for now because that’s where we track events. Seemed sensible to do the calculations as close to where we track them as possible instead of sending them over the wire. That said, we could update the advanced playground response schema to include the count instead, but this seems like it would incur more complexity. I think the best reason for not doing it in the frontend is to reduce the amount of computation client side (and the bundle size, I suppose). That might be a worthwhile tradeoff, but how about we see whether it causes any performance problems first?

PR #4083, opened 2023-06-23
chore: filter out deprecated permissions

gastonfournier:
## About the changes
This makes these permissions not available for selection. In particular `UPDATE_ROLE`, `CREATE_API_TOKEN`, `UPDATE_API_TOKEN`, `DELETE_API_TOKEN`, `READ_API_TOKEN` are long-lived and should be taken out with special care which is why we have https://linear.app/unleash/issue/2-1158/add-delete-migration-to-clean-up-no-longer-used-permissions

## Discussion points
If a role has this permission assigned, it will be displayed but will not be able to remove it. Because the application code does not rely on these permissions, this shouldn't be a problem. Later when we remove them from the DB, the permission will be removed as well from the role by the migration

### DIFF
--- a/src/lib/db/access-store.ts
+++ b/src/lib/db/access-store.ts
@@ -43,6 +43,18 @@ interface IPermissionRow {
 }
 
 export class AccessStore implements IAccessStore {
+    private readonly DEPRECATED_PERMISSIONS = [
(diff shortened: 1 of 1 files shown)


nunogois (approved):
Perfect 👍

PR #4084, opened 2023-06-23
bug(#3545): include strategy titles on playground evaluation results

thomasheartman:
This PR adds strategy titles as an optional bit of data added to client features. It's only added when prompted. 

image

### getPlaygroundFeatures

The optional `includeStrategyId` parameter has been replaced by a `getPlaygroundFeatures` in the service (and in the underlying store). The playground was the only place that used this specific include, so instead of adding more and making the interface for that method more complex, I created a new method that deals specifically with the playground. 

The underlying store still uses an `optionalIncludes` parameter, however. I have a plan to make that interface more fluid, but I'd like to propose that in a follow-up PR.

### DIFF
--- a/src/lib/db/feature-toggle-client-store.ts
+++ b/src/lib/db/feature-toggle-client-store.ts
@@ -304,13 +304,22 @@ export default class FeatureToggleClientStore
 
     async getClient(
         featureQuery?: IFeatureToggleQuery,
-        optionalIncludes?: OptionalClientFeatures,
     ): Promise<IFeatureToggleClient[]> {
         return this.getAll({
             featureQuery,
             archived: false,
             isAdmin: false,
-            optionalIncludes,
+        });
+    }
+
+    async getPlayground(
+        featureQuery?: IFeatureToggleQuery,
+    ): Promise<IFeatureToggleClient[]> {
+        return this.getAll({
+            featureQuery,
+            archived: false,
+            isAdmin: false,
+            optionalIncludes: new Set(['strategy titles', 'strategy IDs']),
         });
     }
 
--- a/src/lib/db/feature-toggle-client-store.ts
+++ b/src/lib/db/feature-toggle-client-store.ts
@@ -23,6 +23,7 @@ export interface IGetAllFeatures {
     includeStrategyIds?: boolean;
     includeDisabledStrategies?: boolean;
     userId?: number;
+    includeStrategyTitles?: boolean;
 }
 
 export interface IGetAdminFeatures {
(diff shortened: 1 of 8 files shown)


src/lib/db/feature-toggle-client-store.ts
```diff
@@ -238,6 +245,16 @@ export default class FeatureToggleClientStore
         });
     }
 
+    private static removeTitlesFromStrategies(
+        features: IFeatureToggleClient[],
+    ) {
+        features.forEach((feature) => {
+            feature.strategies.forEach((strategy) => {
+                delete strategy.title;
```
kwasniew, line 253: JS engines like V8 don't like delete as it affects perf. What if we iterated over features with map by omiting what is not needed and combining removeTitlesFromStrategies with removeIdsFromStrategies to avoid re-iterating

src/lib/db/feature-toggle-client-store.ts
```diff
@@ -300,13 +317,15 @@ export default class FeatureToggleClientStore
         featureQuery?: IFeatureToggleQuery,
         includeStrategyIds?: boolean,
         includeDisabledStrategies?: boolean,
+        includeStrategyTitles?: boolean,
```
kwasniew, line 320: what's the story behind those include special cases? Are they for playground only while SDKs never include anythig?

kwasniew (approved):
Non blocking approve. I'd triple check if the client API for SDKs is unaffected.

src/lib/db/feature-toggle-client-store.ts
thomasheartman: I don't know about the disabled strategies (but @andreas-unleash might?). The strategy IDs one **is** for playground, though. Don't know if it's been used anywhere else since, but probably not. Might be beneficial to abstract it away into a `getPlaygroundClient` function instead?

src/lib/db/feature-toggle-client-store.ts
thomasheartman: Ah, I wasn't aware of that. But yeah, that seems like a fine suggestion. I left the simplest case in here for now assuming that the number of features is usually insignificant, but avoiding an iteration is better if we can. I'll have a look!

src/lib/db/feature-toggle-client-store.ts
kwasniew: BTW from the API design perspective includeXYZ: boolean is not very elegant. Maybe sth like:
` type OptionalInclude = 'strategyIds' | 'disabledStratgies' | 'strategyTitles' `
+ array of OptionalIncludes

chriswk (approved):
Nice. Also a nice catch to not return id/title to client endpoint. Exposing internal id's is really not a good idea (https://blog.acolyer.org/2016/09/13/data-on-the-outside-versus-data-on-the-inside/), but for our internal playground I can see the need.

PR #4096, opened 2023-06-27
fix: max revision query order

sjaanus:
The previous query utilized an ID limit for the "type" condition but not for the "feature_name" condition.

The improved query now correctly handles both conditions by ensuring that the "feature_name" is not null or the "type" matches the specified value, and then checking if the ID is greater than or equal to the provided value.
```
OLD: select max("id") from "events" where "feature_name" is not null or "type" = $1 and "id" >= $2 limit $3
NEW: select max("id") from "events" where ("feature_name" is not null or "type" = $1) and "id" >= $2 limit $3
```


ivarconr (approved):
Nice catch!

PR #4101, opened 2023-06-27
fix: roles plan assumption

nunogois:
Fixes a wrong assumption that project roles were an Enterprise feature.
This PR only focuses on getting the fix out, would be nice to clean this up afterwards.


nunogois:
This also had some wrong assumptions, so I'm closing it.

PR #4108, opened 2023-06-28
fix: metrics performance patch

sjaanus:
In our **hosted** instance, we faced issue of metrics loading 5-10 seconds. We had two environments, around 70 applications, and 48 hour buckets, resulting in ~7000 permutations.

The main issue arose when the metrics.filter function was called, which involved running the compare function on all metrics for each permutation. With 3000 metrics in total, this led to **over 20 million calls** to the compareAsc function worse case scenario, causing a delay of 5-10 seconds for metric loading.

To improve performance, I optimized the logic by prefiltering the metrics based on the environment and appName. This eliminated the need to run the same function on all metrics repeatedly. Now, the compareAsc function is expected to be called only around 5000 times, resulting in a significant **tenfold performance increase**.


chriswk (approved):
Definitely a nice improvement. :)

PR #4115, opened 2023-06-29
refactor: use `requestType` instead of `isAdmin`, `optionalIncludes`

thomasheartman:
This removes the burden of knowing what optionalIncludes is and also
prevents weird edge cases like isAdmin with optionalIncludes.

### DIFF
--- a/src/lib/db/feature-toggle-client-store.ts
+++ b/src/lib/db/feature-toggle-client-store.ts
@@ -16,14 +16,11 @@ import FeatureToggleStore from './feature-toggle-store';
 import { Db } from './db';
 import Raw = Knex.Raw;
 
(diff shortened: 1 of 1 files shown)


kwasniew (approved):
It's beautiful :) I'd double check if the client query works fine (from the SDK) since it's the most impactful part. Looking at the code nothing should break but with client API facing stuff I'm always very careful.

thomasheartman:
Had a feeling you'd like this one 😄  And yeah, good point! I'll double check!

thomasheartman:
The Rust SDK (which we know is a little finicky) runs fine with these changes, as does the Go SDK, so I'm happy to merge this.

PR #4116, opened 2023-06-29
openapi: `strategies` tag

thomasheartman:
Update OpenAPI schemas and operation descriptions for the strategies tag.


sjaanus (commented):
Context and staregies controller should respondWithValidation. This will allow our e2e tests to validate schema.

Or we need to skip validation to send extra headers?

thomasheartman:
Should be addressed in 6c6251ff8216303fa2e768d17502a4b41db7ffb7 now. I'm assuming we don't use `respondWithValidation` for responses without bodies, right?

sjaanus:
> Should be addressed in [6c6251f](https://github.com/Unleash/unleash/commit/6c6251ff8216303fa2e768d17502a4b41db7ffb7) now. I'm assuming we don't use `respondWithValidation` for responses without bodies, right?

Yes, we dont use it for responses without bodies.

sjaanus (approved):

PR #4117, opened 2023-06-29
docs: Context api tag

chriswk:
### What
Added documentation for context fields and endpoints tagged with the "Context" tag.


src/lib/openapi/spec/context-field-schema.ts
```diff
@@ -5,41 +5,59 @@ export const contextFieldSchema = {
     $id: '#/components/schemas/contextFieldSchema',
     type: 'object',
     additionalProperties: false,
+    description:
+        'A [context field](https://docs.getunleash.io/reference/unleash-context) used to configure strategies with data used to evaluate isEnabled calls',
```
thomasheartman, line 9: This sentence runs on a bit and I'm not sure what it's saying. Would it be fine to shorten it down a bit like this?

```suggestion
        'A representation of a [context field](https://docs.getunleash.io/reference/unleash-context).',
```

src/lib/openapi/spec/context-field-schema.ts
```diff
@@ -5,41 +5,59 @@ export const contextFieldSchema = {
...
         },
         stickiness: {
+            description:
+                'Does this context field support being used for [stickiness](https://docs.getunleash.io/reference/stickiness) calculations',
             type: 'boolean',
+            example: true,
         },
         sortOrder: {
+            description:
+                'Used to have deterministic sorting of a list of context fields. Can also be used as a draw-breaker if list of context fields is sorted alphabetically.',
```
thomasheartman, line 31: I've never heard draw-breaker before, but I'm assuming it's the same as a tiebreaker? I'd suggest the latter because it seems to be more widely used.

What's the deterministic sort order you mention? What's it based on? 

```suggestion
                'Used when sorting a list of context fields. Is also used as a tiebreaker if a list of context fields is sorted alphabetically.',
```

src/lib/openapi/spec/context-field-schema.ts
```diff
@@ -5,41 +5,59 @@ export const contextFieldSchema = {
...
             type: 'number',
+            example: 900,
```
thomasheartman, line 33: Is this a `number` or an `integer`? Both are valid types in OpenAPI and if we always assume integers, then that may be a good way to make it more precise.

src/lib/openapi/spec/context-field-schema.ts
```diff
@@ -5,41 +5,59 @@ export const contextFieldSchema = {
...
         },
         createdAt: {
+            description: 'When this context field was created',
             type: 'string',
             format: 'date-time',
             nullable: true,
+            example: '2023-06-29T10:19:00.000Z',
         },
         usedInFeatures: {
             type: 'number',
```
thomasheartman, line 43: this and `usedInProjects` should probably also be `integers` instead of numbers: https://swagger.io/docs/specification/data-models/data-types/#numbers

src/lib/openapi/spec/legal-value-schema.ts
```diff
@@ -4,13 +4,19 @@ export const legalValueSchema = {
...
     properties: {
         value: {
+            description: 'The valid value',
             type: 'string',
+            example: 'red',
         },
         description: {
+            description: 'Describes the intended usage of this value',
             type: 'string',
+            example: 'We only carry these colors in our shop',
```
thomasheartman, line 19: This description's example feels like it's for legal values as a whole and not this specific value. How about something like this (if we're sticking with colors)?


```suggestion
            description: 'The valid value',
            type: 'string',
            example: '#c154c1',
        },
        description: {
            description: 'Describes this specific legal value',
            type: 'string',
            example: 'Deep fuchsia'
```

src/lib/routes/admin-api/context.ts
```diff
@@ -65,6 +65,9 @@ export class ContextController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Context'],
+                    summary: 'Gets configured context fields',
+                    description:
+                        'Returns all configured [Context fields](https://docs.getunleash.io/how-to/how-to-define-custom-context-fields) that have been created.',
```
thomasheartman, line 70: This is a little unclear: Can a context field be configured and not created? What about the other way: can it be created and not configured?

src/lib/routes/admin-api/context.ts
```diff
@@ -81,6 +84,9 @@ export class ContextController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Context'],
+                    summary: 'Gets context field',
+                    description:
+                        'Returns specific [context field](https://docs.getunleash.io/reference/unleash-context) identified by the name in the path',
```
thomasheartman, line 89:
```suggestion
                    summary: 'Get a context field',
                    description:
                        'Returns a specific [context field](https://docs.getunleash.io/reference/unleash-context) identified by the name in the path',
```

src/lib/routes/admin-api/context.ts
```diff
@@ -65,6 +65,9 @@ export class ContextController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Context'],
+                    summary: 'Gets configured context fields',
```
thomasheartman, line 68: Would this be correct?

```suggestion
                    summary: 'Get all context fields',
```

src/lib/routes/admin-api/context.ts
```diff
@@ -97,6 +103,8 @@ export class ContextController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Strategies'],
+                    summary: 'Gets strategies using this context field',
+                    description: `Used to see which strategies are using this context field. Can aid in cleanup of no longer used context fields, if this list is empty it's safe to delete the context field`,
```
thomasheartman, line 107:
```suggestion
                    summary: 'Get strategies using this context field',
                    description: `Returns all strategies that use the specified context field. Useful when cleaning up context fields: if this list is empty, it's safe to delete the context field.`,
```

src/lib/routes/admin-api/context.ts
```diff
@@ -136,6 +147,8 @@ export class ContextController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Context'],
+                    summary: 'Update an existing context field',
+                    description: `Endpoint that allows updating a custom context field. Used to toggle stickiness and add/remove legal values for this context field`,
```
thomasheartman, line 151: How is the PUT treated here? Does it replace the object wholesale, or is it more of a patch? If it works more like a patch (if any fields are ignored or unchanged if not provided), then that should be mentioned here.

src/lib/routes/admin-api/context.ts
```diff
@@ -156,6 +169,9 @@ export class ContextController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Context'],
+                    summary: 'Delete an existing context field',
+                    description:
+                        'Endpoint that allows deletion of a custom context field. Does not validate that context field is not in use, but since context field configuration is stored in a json blob for the strategy, existing strategies are safe.',
```
thomasheartman, line 174: The explanation here is probably useful, but seems to ... internal to Unleash (end users don't need to know how we store things). What does this mean to the user? If you delete a context field that is in use, what happens? 

```suggestion
                        'Endpoint that allows deletion of a custom context field. Does not validate that context field is not in use, but since context field configuration is stored in a json blob for the strategy, existing strategies are safe.',
```

src/lib/routes/admin-api/context.ts
```diff
@@ -172,6 +188,9 @@ export class ContextController extends Controller {
             middleware: [
                 openApiService.validPath({
                     tags: ['Context'],
+                    summary: 'Validates a context field',
```
thomasheartman, line 191:
```suggestion
                    summary: 'Validate a context field',
```

src/lib/routes/admin-api/context.ts
```diff
@@ -172,6 +188,9 @@ export class ContextController extends Controller {
...
+                    description:
+                        'Used by the Unleash frontend to validate that the form for creating a context field is valid (e.g. that the name of the context field is URL safe)',
```
thomasheartman, line 193: I don't think we need to tell the user where we use this? But I'm not sure it matters much.

What does it do if it can't? Does it return anything about why the field can't be created? And is it returned as a 200 with a body or as a 400?

```suggestion
                        'Check whether the provided data can be used to create a context field. If the data is not valid, ...?',
```

thomasheartman (commented):
Good job! 😄  I've got a few small suggestions that you can address and a few questions about some of your descriptions, but overall this is solid work. Let me know if you've got any questions ☺️

src/lib/openapi/spec/context-field-schema.ts
chriswk: Yes, thank you. The link should cover what a context field actually is.

src/lib/openapi/spec/context-field-schema.ts
chriswk: Good point, I don't know where drawbreaker came from either... :P

src/lib/openapi/spec/context-field-schema.ts
chriswk: agreed, It's kinda hard to be used in 2.3 projects. Will fix.

src/lib/openapi/spec/legal-value-schema.ts
chriswk: Yeah, that works for me :) Thanks for the improved example and description

PR #4122, opened 2023-06-30
ux: return better error message if a segment doesn't exist

thomasheartman:
Catch cases where the segment doesn't exist and populate that error message with more info: it now says that a segment
with <id> doesn't exist instead of just 'No row'.

### DIFF
--- a/src/lib/db/segment-store.ts
+++ b/src/lib/db/segment-store.ts
@@ -7,6 +7,7 @@ import { PartialSome } from '../types/partial';
 import User from '../types/user';
 import { Db } from './db';
 import { IFlagResolver } from '../types';
+import { BadDataError } from '../error';
 
 const T = {
     segments: 'segments',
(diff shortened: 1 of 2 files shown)


src/lib/db/segment-store.ts
```diff
@@ -193,9 +194,19 @@ export default class SegmentStore implements ISegmentStore {
         const rows: ISegmentRow[] = await this.db
             .select(this.prefixColumns())
             .from(T.segments)
-            .where({ id });
+            .where({ id })
```
sjaanus, line 197: I think ID validation ideally should not be done on store. Can we do it in schema?

thomasheartman: Good point. We don't use an OpenAPI schema for this because it's part of the request URL in express. I'll see if it's something we can handle there, though.

thomasheartman: We can definitely handle this in the controller, though, so happy to do that. That said. I'd like to keep the improved NotFoundError: would you be okay with that?

sjaanus: I think the improved error is fine, but the catch part should not be there and validated on controller level.

sjaanus: On the other hand, this makes me wonder what else we need to validate. Maybe we should just catch the error in service, and give generic response message? When you are putting in numbers like -1491751759715971759, then its not a actual user flow, but someone trying to do something stupid, and we do not need to give nice error messages for every corner case people try to do.

sjaanus: That being said, UI seems to handle it nicely, nothing wrong on the UI side, so it is just in server logs. So maybe we should not change it at all and keep this ugly error, because again, this is if someone is trying to break your application, and we want to know when this is happening.

sjaanus (approved):

PR #4125, opened 2023-06-30
fix: demo flow with split strategy button by making step optional

nunogois:
This fixes the demo flow by making the strategy popup step optional. 

This way the demo works whether the flag is enabled or not. Once we make the split strategy button GA, we can remove this step entirely.

Feel free to re-enable the flag once this is deployed to demo.


nicolaesocaciu (approved):
LGTM!

PR #4129, opened 2023-07-03
feat/admin menu reorganize

daveleek:
## About the changes

Reorganizes the items in the menu to align with the tabs on the admin page.
Also makes admin menu available to all users, they can get there anyways when using API access link, and all admin-only pages are disabled for non-admins.
Also adds API access to the mobile drawer menu, in accordance with how the configure menu is laid out.


ivarconr (commented):
Has this change been coordinated with @nicolaesocaciu ?

nunogois:
> Has this change been coordinated with @nicolaesocaciu ?

Should be aligned with what @nicolaesocaciu had in mind for now. I think @daveleek forgot to link the Linear task: https://linear.app/unleash/issue/2-1163/having-a-role-with-access-to-some-of-the-administrative-tasks-such-as

I can take lead if needed and double check with @nicolaesocaciu to move this forward.

nicolaesocaciu:
> Has this change been coordinated with @nicolaesocaciu ?

yes its based on the task that Nuno share

Context: Even if we were hidden the settings icon in the menu all the users could get to those links from the UI on a different path (of course they don't see the page content if they don't have permissions), and it was not logic to hide it for some users anymore, and now with the new permissions will be even more complex

nunogois (approved):
LGTM 👍

PR #4130, opened 2023-07-03
fix: SERVER_KEEPALIVE_TIMEOUT env variable should be seconds

gardleopard:
This changes SERVER_KEEPALIVE_TIMEOUT to take in seconds instead of milliseconds.


chriswk (approved):
LGTM

chriswk (approved):
LGTM

PR #4131, opened 2023-07-03
fix: improve users search

nunogois:
https://linear.app/unleash/issue/2-1183/improve-users-list-search

This adapts the users list to adapt the new `useSearch` hook so you can have a better search that also includes "username" and "email" as fields.


gastonfournier (approved):
LGTM! Let's test it out

PR #4134, opened 2023-07-03
docs: Documentation around metrics API

gastonfournier:
## About the changes
Mark some legacy endpoints as deprecated (they are long deprecated but documentation was out of date) and fix some of the URLs

### DIFF
--- a/website/docs/reference/api/legacy/unleash/admin/metrics.md
+++ b/website/docs/reference/api/legacy/unleash/admin/metrics.md
@@ -8,8 +8,13 @@ title: /api/admin/metrics
 # This document describes the metrics endpoint for admin ui
 
 ### Seen-toggles {#seen-toggles}
+:::caution Deprecation notice
 
-`GET http://unleash.host.com/api/admin/seen-toggles`
+This endpoint has been deprecated
+
+:::
+
+`GET http://unleash.host.com/api/admin/metrics/seen-toggles`
 
 This enpoints returns a list of applications and what toogles unleash has seen for each application. It will only guarantee toggles reported by client applications within the last hour, but will in most cases remember seen toggles for applications longer.
 
@@ -84,9 +89,9 @@ This endpoint gives _last minute_ and _last hour_ metrics for all active toggles
 
 ### Applications {#applications}
 
-`GET http://unleash.host.com/api/admin/applications`
+`GET http://unleash.host.com/api/admin/metrics/applications`
 
-This endpoint returns a list of known applications (seen the last two days) and a link to follow for more details.
+This endpoint returns a list of known applications (seen the last day) and a link to follow for more details.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


website/docs/reference/api/legacy/unleash/admin/metrics.md
```diff
@@ -8,8 +8,13 @@ title: /api/admin/metrics
 # This document describes the metrics endpoint for admin ui
 
 ### Seen-toggles {#seen-toggles}
+:::caution Deprecation notice
 
-`GET http://unleash.host.com/api/admin/seen-toggles`
+This endpoint has been deprecated
```
thomasheartman, line 13: What should they use instead?

```suggestion
This endpoint has been deprecated. Use the [what?] endpoint instead.
```

website/docs/reference/api/legacy/unleash/admin/metrics.md
```diff
@@ -84,9 +89,9 @@ This endpoint gives _last minute_ and _last hour_ metrics for all active toggles
 
 ### Applications {#applications}
 
-`GET http://unleash.host.com/api/admin/applications`
+`GET http://unleash.host.com/api/admin/metrics/applications`
 
-This endpoint returns a list of known applications (seen the last two days) and a link to follow for more details.
+This endpoint returns a list of known applications (seen the last day) and a link to follow for more details.
```
thomasheartman, line 94:
```suggestion
This endpoint returns a list of known applications (seen in the last day) and a link to follow for more details.
```

website/docs/reference/api/legacy/unleash/admin/metrics.md
```diff
@@ -152,7 +157,13 @@ This endpoint gives insight into details about a client application, such as ins
 
 ### Seen applications {#seen-applications}
 
-`GET http://unleash.host.com/api/admin/seen-apps`
+:::caution Deprecation notice
+
+This endpoint has been deprecated
```
thomasheartman, line 162: Same as above: what do they use instead?t

```suggestion
This endpoint has been deprecated. Use ... instead.
```

thomasheartman (approved):
The changes here look good! My only real ask before you merge is that you point them to what endpoints they should use instead of the deprecated ones, something like:
```
Use the [delete a user](link to new endpoint docs) endpoint instead.
```

good work! 🙌🏼

website/docs/reference/api/legacy/unleash/admin/metrics.md
gastonfournier: There's no replacement AFAIK

website/docs/reference/api/legacy/unleash/admin/metrics.md
thomasheartman: So this functionality is just going away? In that case, we should say that instead.

gastonfournier: At least it's not listed in the new docs: https://docs.getunleash.io/reference/api/unleash/metrics there are some new endpoints but I don't think it'd be easy for customers to use them. But we can say that this will be removed in v6

website/docs/reference/api/legacy/unleash/admin/metrics.md
gastonfournier: And they're returning an empty response with 410 response code:
https://github.com/Unleash/unleash/blob/0b18491237686be627edc52fc953d37e9624371d/src/lib/routes/admin-api/metrics.ts#L120-L126

PR #4135, opened 2023-07-04
docs: switch order of api doc sidebar items; put legacy docs under legacy header

thomasheartman:
Also adds a caution admonition to the legacy doc index page


gastonfournier (approved):
Looks nice! especially the caution note

PR #4137, opened 2023-07-04
openapi: update API tokens tag

thomasheartman:
This PR updates endpoints and schemas for the API tokens tag.


src/lib/openapi/spec/create-api-token-schema.ts
```diff
@@ -1,68 +1,83 @@
...
-            required: ['username'],
         },
         {
+            required: ['type'],
             properties: {
-                tokenName: {
+                type: {
                     type: 'string',
+                    pattern:
+                        '[Cc][Ll][Ii][Ee][Nn][Tt]|[Ff][Rr][Oo][Nn][Tt][Ee][Nn][Dd]',
```
kwasniew, line 55: can we do it with case insensitive regex instead of typing every letter in []?

src/lib/routes/admin-api/api-token.ts
```diff
@@ -154,8 +157,12 @@ export class ApiTokenController extends Controller {
                 openApiService.validPath({
                     tags: ['API tokens'],
                     operationId: 'getAllApiTokens',
+                    summary: 'Get API tokens',
+                    description:
+                        'Retrieves all API tokens that exists in the Unleash instance.',
```
kwasniew, line 162: that exist?

src/lib/openapi/spec/create-api-token-schema.ts
```diff
@@ -1,68 +1,83 @@
...
     },
-    anyOf: [
+    oneOf: [
         {
+            required: ['type'],
             properties: {
-                username: {
+                type: {
                     type: 'string',
+                    pattern: '[Aa][Dd][Mm][Ii][Nn]',
```
kwasniew, line 43: Would /admin/i work?

kwasniew: In a typical regex syntax it should work but I'm not sure about the openapi engine

kwasniew (approved):
Nice one. Minor comments.

src/lib/openapi/spec/create-api-token-schema.ts
thomasheartman: Sadly, the OpenAPI spec doesn't support the regex flags to make it case insensitive, so this is the best way. Ideally I'd use the `enum` property, but that is also case sensitive, so this seems to be the only way. Because we verify this on the backend anyway, I'd prefer to put this in the OpenAPI handling.

src/lib/routes/admin-api/api-token.ts
thomasheartman: Yes!

src/lib/error/unleash-error.test.ts
```diff
@@ -99,6 +99,46 @@ describe('OpenAPI error conversion', () => {
...
+            // it tells the user what happened
+            expect(result.description).toContain(
+                'matches more than one option',
+            );
+            // it tells the user what part of the request body this pertains to
+            expect(result.description).toContain(
+                dataPath === '.body'
+                    ? 'root object'
+                    : `${dataPath.substring('.body.'.length)} property`,
+            );
```
thomasheartman, line 138: Turns out that using multipling `expect.stringContaining` on on object property doesn't work 😬

thomasheartman:
Good catch @sjaanus 🥇

PR #4139, opened 2023-07-04
task: Add customHeaders as possible parameter.

chriswk:
###What
Adds an optional sensitive parameter for customHeaders to all current addons. When formatted as a valid json object, the object will be included as headers in the request the addon makes upstream to it's target URL

### DIFF
--- a/src/lib/addons/__snapshots__/slack.test.ts.snap
+++ b/src/lib/addons/__snapshots__/slack.test.ts.snap
@@ -1,9 +1,18 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Should call slack webhook 1`] = `"{"username":"Unleash","icon_emoji":":unleash:","text":"some@user.com created feature toggle <http://some-url.com/projects/default/features/some-toggle|some-toggle> in project *default*","channel":"#undefined","attachments":[{"actions":[{"name":"featureToggle","text":"Open in Unleash","type":"button","value":"featureToggle","style":"primary","url":"http://some-url.com/projects/default/features/some-toggle"}]}]}"`;
+exports[`Should call slack webhook 1`] = `"{"username":"Unleash","icon_emoji":":unleash:","text":"some@user.com created feature toggle <http://some-url.com/projects/default/features/some-toggle|some-toggle> in project *default*","channel":"#general","attachments":[{"actions":[{"name":"featureToggle","text":"Open in Unleash","type":"button","value":"featureToggle","style":"primary","url":"http://some-url.com/projects/default/features/some-toggle"}]}]}"`;
 
-exports[`Should call slack webhook for archived toggle 1`] = `"{"username":"Unleash","icon_emoji":":unleash:","text":" some@user.com just archived feature toggle *<http://some-url.com/archive|some-toggle>*","channel":"#undefined","attachments":[{"actions":[{"name":"featureToggle","text":"Open in Unleash","type":"button","value":"featureToggle","style":"primary","url":"http://some-url.com/archive"}]}]}"`;
+exports[`Should call slack webhook for archived toggle 1`] = `"{"username":"Unleash","icon_emoji":":unleash:","text":" some@user.com just archived feature toggle *<http://some-url.com/archive|some-toggle>*","channel":"#general","attachments":[{"actions":[{"name":"featureToggle","text":"Open in Unleash","type":"button","value":"featureToggle","style":"primary","url":"http://some-url.com/archive"}]}]}"`;
 
-exports[`Should call slack webhook for archived toggle with project info 1`] = `"{"username":"Unleash","icon_emoji":":unleash:","text":" some@user.com just archived feature toggle *<http://some-url.com/projects/some-project/archive|some-toggle>*","channel":"#undefined","attachments":[{"actions":[{"name":"featureToggle","text":"Open in Unleash","type":"button","value":"featureToggle","style":"primary","url":"http://some-url.com/projects/some-project/archive"}]}]}"`;
+exports[`Should call slack webhook for archived toggle with project info 1`] = `"{"username":"Unleash","icon_emoji":":unleash:","text":" some@user.com just archived feature toggle *<http://some-url.com/projects/some-project/archive|some-toggle>*","channel":"#general","attachments":[{"actions":[{"name":"featureToggle","text":"Open in Unleash","type":"button","value":"featureToggle","style":"primary","url":"http://some-url.com/projects/some-project/archive"}]}]}"`;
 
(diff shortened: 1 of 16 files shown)


src/lib/addons/webhook.test.ts
```diff
@@ -114,3 +114,35 @@
...
+            enabled: false,
+            strategies: [{ name: 'default' }],
+        },
+    };
+
+    const parameters = {
+        url: 'http://test.webhook.com/plain',
+        bodyTemplate: '{{event.type}} on toggle {{event.data.name}}',
+        contentType: 'text/plain',
+        authorization: 'API KEY 123abc',
```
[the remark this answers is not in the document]
gastonfournier: I think it can safely be dismissed

src/lib/addons/__snapshots__/slack.test.ts.snap
```diff
@@ -1,9 +1,18 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Should call slack webhook 1`] = `"{"username":"Unleash","icon_emoji":":unleash:","text":"some@user.com created feature toggle <http://some-url.com/projects/default/features/some-toggle|some-toggle> in project *default*","channel":"#undefined","attachments":[{"actions":[{"name":"featureToggle","text":"Open in Unleash","type":"button","value":"featureToggle","style":"primary","url":"http://some-url.com/projects/default/features/some-toggle"}]}]}"`;
+exports[`Should call slack webhook 1`] = `"{"username":"Unleash","icon_emoji":":unleash:","text":"some@user.com created feature toggle <http://some-url.com/projects/default/features/some-toggle|some-toggle> in project *default*","channel":"#general","attachments":[{"actions":[{"name":"featureToggle","text":"Open in Unleash","type":"button","value":"featureToggle","style":"primary","url":"http://some-url.com/projects/default/features/some-toggle"}]}]}"`;
```
gastonfournier, line 3: Have you fixed slack integration by chance?

src/lib/addons/datadog.ts
```diff
@@ -9,6 +9,12 @@ import {
 } from './feature-event-formatter-md';
 import { IEvent } from '../types/events';
 
+interface IDatadogParameters {
+    url: string;
+    apiKey: string;
+    customHeaders?: string;
```
gastonfournier, line 15: Why not something like:
```
    customHeaders?: { [key: string]: string } | Record<string, string> ;
```
That would save you from doing `JSON.parse` later

Inspired by https://github.com/search?q=%22type+HeadersInit+%3D%22+language%3ATypeScript&type=code&l=TypeScript

src/lib/addons/datadog.ts
```diff
@@ -35,12 +47,22 @@ export default class DatadogAddon extends Addon {
...
             tags,
         };
-
+        let extraHeaders = {};
+        if (typeof customHeaders === 'string' && customHeaders.length > 1) {
+            try {
+                extraHeaders = JSON.parse(customHeaders);
+            } catch (e) {
+                this.logger.warn(
+                    'Could not parse the json in the customHeaders parameters',
```
gastonfournier, line 56:
```suggestion
                    `Could not parse the json in the customHeaders parameters ${customHeaders}`,
```

but I think this exception should bubble up as it'd be a configuration issue

src/lib/addons/slack.ts
```diff
@@ -10,6 +10,13 @@ import {
 } from './feature-event-formatter-md';
 import { IEvent } from '../types/events';
 
+interface ISlackAddonParameters {
+    url: string;
+    username?: string;
+    defaultChannel: string;
+    emojiIcon?: string;
+    customHeaders?: string;
```
gastonfournier, line 18: Maybe this would benefit from the same suggestion:
```
customHeaders?: { [key: string]: string } | Record<string, string> ;
```

src/lib/addons/teams.ts
```diff
@@ -8,6 +8,10 @@ import {
 } from './feature-event-formatter-md';
 import { IEvent } from '../types/events';
 
+interface ITeamsParameters {
+    url: string;
+    customHeaders?: string;
```
gastonfournier, line 13: Same :D

src/lib/addons/webhook.ts
```diff
@@ -9,6 +9,7 @@ interface IParameters {
     bodyTemplate?: string;
     contentType?: string;
     authorization?: string;
+    customHeaders?: string;
```
gastonfournier, line 12: And one more... although this could be a bit trickier because we need to handle a web request.

src/lib/types/model.ts
```diff
@@ -108,6 +108,10 @@ export interface ISortOrder {
     [index: string]: number;
 }
 
+export interface ICustomHeaders {
+    [name: string]: string;
+}
+
```
gastonfournier, line 114: Ok, I think you had the idea of doing something as I suggested, but I believe this is not used right now

gastonfournier (commented):
I think this will work, just maybe a few improvements

src/lib/addons/webhook.test.ts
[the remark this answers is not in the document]
chriswk: Indeed, since it's a test, but it's good to see CodeQL catching these, just in case we stupidly add them to production code. :)

src/lib/addons/__snapshots__/slack.test.ts.snap
chriswk: No, have not changed it, just given it support for custom headers

src/lib/addons/datadog.ts
chriswk: Hmm, if it works that would be good, we just don't have a way to define the input parameter in the addon definition as anything but a string. So somewhere we'd need to do the parsing.

src/lib/addons/datadog.ts
chriswk: But you wouldn't want to throw an exception when creating a feature because one of the addons failed to parse data.

src/lib/types/model.ts
chriswk: You're right. I forgot to remove this one.

src/lib/addons/datadog.ts
chriswk: Though, this is on a different thread.

src/lib/addons/slack.ts
chriswk: Same comment here, since we're doing a post of parameters, it isn't currently possible to have it as anything but strings, otherwise our template for event data would also be something else but string. I tested locally to ensure that this was correct.

src/lib/addons/__snapshots__/slack.test.ts.snap
gastonfournier: It changed from `undefined` to `general` so the support for custom headers might have helped. Not sure but :+1:

gastonfournier (approved):
LG!

src/lib/addons/__snapshots__/slack.test.ts.snap
chriswk: Oh, that's just because I'm now type checking the parameters, so our test no longer allowed us to activate the slack addon with no defined defaultChannel (it's marked as required in our definition). Earlier, the test that wrote this snapshot did not have the defaultChannel parameter set, which caused javascript to happily send `undefined` :)

PR #4141, opened 2023-07-04
fix: change to the proper author

gastonfournier:
## About the changes
Based on the information displayed here https://github.com/Unleash/unleash/actions/runs/5456442710/jobs/9929252672#step:3:11 this is how we should get the email and name of the committer


nunogois (approved):
LG

PR #4142, opened 2023-07-05
fix: hide users list extra searchable columns

nunogois:
https://linear.app/unleash/issue/2-1187/fix-two-extra-columns-in-users-list

This fixes a regression introduced in https://github.com/Unleash/unleash/pull/4131 where two extra columns were added for search only but were not properly hidden.

Thanks @nicolaesocaciu for the heads up!

Before:
image

After:
image


chriswk (approved):
Nice. LGTM

PR #4144, opened 2023-07-05
fix: ensure userId context exists when running demo

nunogois:
https://linear.app/unleash/issue/2-1168/demo-ensure-userid-context-field-in-setup-steps

This ensures that the `userId` context field exists when we reach specific demo topics that require it in order to be successfully completed. This uses the `setup` property on those topics, where we'll check for the existence of this context field and create it if it's not found.


gastonfournier (commented):
Shouldn't we prevent it from being deleted instead?

chriswk (approved):
LGTM

PR #4145, opened 2023-07-05
Inline endpoint descriptions

Tymek:
Update OpenApi docs


src/lib/features/export-import-toggles/export-import-controller.ts
```diff
@@ -74,7 +73,9 @@ class ExportImportController extends Controller {
                         200: createResponseSchema('exportResultSchema'),
                         ...getStandardResponses(404),
                     },
-                    ...endpointDescriptions.admin.export,
+                    description:
+                        "Exports all features listed in the `features` property from the environment specified in the request body. If set to `true`, the `downloadFile` property will let you download a file with the exported data. Otherwise, the export data is returned directly as JSON. Refer to the documentation for more information about [Unleash's export functionality](https://docs.getunleash.io/reference/deploy/environment-import-export#export).",
```
kwasniew, line 77: We can also export by tags, not just by features

src/lib/features/export-import-toggles/export-import-controller.ts
```diff
@@ -74,7 +73,9 @@ class ExportImportController extends Controller {
                         200: createResponseSchema('exportResultSchema'),
                         ...getStandardResponses(404),
```
kwasniew, line 74: If you don't provide a key you may probably get 401 - I'd test it in the openapi UI

kwasniew (approved):
What's our strategy on inline vs external file?
As we're making changes we can improve the descriptions as we added export by tag and also we need to list 401 codes.

thomasheartman:
On inlining vs internal file descriptions: when we first started this, it was suggested that we store them in an external file to keep the code more compact. However, as it grew, we kind of forgot about it and now most of the descriptions live next to the code. Personally I think I prefer this for these reasons:
- it lives closer to the code, so it's easier to understand what it's describing and easier to find. Saves you jumping to a different file and then finding the right key in that file.
- it's easier for developers. Less thinking about finding the right file, placing the descriptions under the right key name etc. Just less overhead.
- almost all our descriptions are inlined by now, so changing it would be a lot of work.

But I'd love to hear other views or suggestions too.

thomasheartman (approved):
I think this is a good change and it clears up some of our inconsistencies; good work! 🏆  And while it's not related to the changes you made, I think @kwasniew is right when it comes to the standard responses. It might be worth doing in a separate PR, though 💁🏼  (And yeah, it just means that we missed this in a previous PR)

PR #4146, opened 2023-07-05
feat: openapi schema for user admin

kwasniew:
## About the changes

Full OpenAPI spec for user-admin controller.

* This should be around 100-200 LOC. 1k lines is from the snapshot test.
* Removed all working schemas from the meta schema exception list
* adding 400 when body or query/path params are validated, 401 always, 403 when ADMIN is required, 404 when resource may not exist
* fixing a few 500 when id was not a number, those should be 400 errors
* making BadDataError reflect incorrect data in body or params
* making user roles being number or string explicit. When creating/updating a user we can specify both and we return what you specified. When reading a user we always return a number.


src/lib/error/bad-data-error.ts
```diff
@@ -16,7 +16,7 @@ class BadDataError extends UnleashError {
         errors?: [ValidationErrorDescription, ...ValidationErrorDescription[]],
     ) {
         const topLevelMessage =
-            'Request validation failed: your request body contains invalid data' +
+            'Request validation failed: your request body or params contain invalid data' +
```
kwasniew, line 19: we throw this error when query or path params are incorrect so the error message should be more generic

src/lib/openapi/spec/create-user-response-schema.ts
```diff
@@ -0,0 +1,34 @@
...
+import { userSchema } from './user-schema';
+
+export const createUserResponseSchema = {
+    $id: '#/components/schemas/createUserResponseSchema',
+    type: 'object',
+    additionalProperties: false,
+    description: 'An Unleash user after creation',
+    required: ['id'],
+    properties: {
+        ...userSchema.properties,
```
kwasniew, line 11: same as user schema but it returns either number or string for the role name depending on what was in the create/update request

src/lib/services/user-service.ts
```diff
@@ -192,7 +192,7 @@ class UserService {
 
         const exists = await this.store.hasUser({ username, email });
         if (exists) {
-            throw new Error('User already exists');
+            throw new BadDataError('User already exists');
```
kwasniew, line 195: Error was 500, BadDataError is 400

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -451,48 +529,66 @@ export default class UserAdminController extends Controller {
...
+                ...serializeDates(updateUser),
+                rootRole: normalizedRootRole,
+            },
+        );
     }
 
     async deleteUser(req: IAuthRequest, res: Response): Promise<void> {
         const { user, params } = req;
         const { id } = params;
+        if (!Number.isInteger(Number(id))) {
```
kwasniew, line 589: previously it was 500, not it is 400

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -451,48 +529,66 @@ export default class UserAdminController extends Controller {
...
+        >,
+        res: Response<CreateUserResponseSchema>,
     ): Promise<void> {
         const { user, params, body } = req;
         const { id } = params;
         const { name, email, rootRole } = body;
+        if (!Number.isInteger(Number(id))) {
+            throw new BadDataError('User id should be an integer');
+        }
+        const normalizedRootRole = Number.isInteger(Number(rootRole))
```
kwasniew, line 561: openapi coerces root role to string on number | string union so we need to coerce it back to number if possible

src/lib/openapi/spec/create-user-schema.ts
```diff
@@ -5,24 +5,52 @@ export const createUserSchema = {
...
         password: {
             type: 'string',
+            example: 'k!5As3HquUrQ',
+            description: 'Password for the user',
         },
         rootRole: {
-            type: 'number',
+            description:
+                "The role to assign to the user. Can be either the role's ID or its unique name.",
+            oneOf: [
```
kwasniew, line 36: oneOf with integer or string + coercion: true in our openapi lib causes rootRole to always be a string. We need to compensate for it in the controller by coercing back

src/lib/openapi/spec/create-user-response-schema.ts
kwasniew: user schema hasn't changed because this schema was added

sjaanus:
Also enable `strictSchemaValidation` in user-admin e2e test to validate all of this.

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -114,8 +122,14 @@ export default class UserAdminController extends Controller {
...
                     tags: ['Users'],
                     operationId: 'validateUserPassword',
+                    summary: 'Validate password for a user',
+                    description:
+                        'Validate the password strength. Minimum 10 characters, uppercase letter, number, special character.',
                     requestBody: createRequestSchema('passwordSchema'),
-                    responses: { 200: emptyResponse },
+                    responses: {
+                        200: emptyResponse,
+                        ...getStandardResponses(400, 401),
```
thomasheartman, line 131: Any user request that sends data can also get a 415 if the data is in a format we don't accept (such as xml). This also goes for the other endpoints in this pr.

```suggestion
                        ...getStandardResponses(400, 401, 415),
```

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -175,7 +204,22 @@ export default class UserAdminController extends Controller {
...
+                        ' It will preform a simple search based on name and email matching the given query. Requires minimum 2 characters',
+                    parameters: [
+                        {
+                            name: 'q',
+                            description:
+                                'The pattern to search in the username or email',
+                            schema: { type: 'string' },
+                            in: 'query',
+                        },
+                    ],
```
thomasheartman, line 218: If you need to type this somehow (you might not), then you could look into using the `FromQueryParams` method and putting it in its own file (similar to how we do it in `src/lib/openapi/spec/export-query-parameters.ts`). I'm not sure whether that's necessary here. As long as it shows up correctly in the output schema, I'm happy 😄

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -256,8 +318,13 @@ export default class UserAdminController extends Controller {
                 openApiService.validPath({
                     tags: ['Users'],
                     operationId: 'updateUser',
+                    summary: 'Update a user',
+                    description: 'Updates use with new fields',
```
thomasheartman, line 322: Are all fields required? Are non-submitted fields ignored? In other words: does this act more like a "true PUT" or a "PATCH"? It'd be good to mention that in the description.

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -451,48 +529,66 @@ export default class UserAdminController extends Controller {
...
             ...serializeDates(createdUser),
             inviteLink: inviteLink || this.unleashUrl,
             emailSent,
-            rootRole,
+            rootRole: normalizedRootRole,
         };
 
         this.openApiService.respondWithValidation(
-            201,
+            200,
```
thomasheartman, line 540: Why did we change this from a 201 to a 200?

thomasheartman (approved):
The changes look good to me overall! 💯  Got a few questions and minor typo fixes that I'd like to have addressed, though, but I trust you to take that into account before merging.

src/lib/routes/admin-api/user-admin.ts
kwasniew: I think one of our validators was complaining that 201 should have a Location header

thomasheartman: Yeah, that sounds about right. You should add one: `respondWithValidation` takes an optional last parameter called `headers`. You can add it in there, pointing to the newly created user.

kwasniew: Ok I found the location header setting pattern in other parts of the code. Our schema was lying before with 200 and the code was returning 201. Let's stick to 201 in both

src/lib/routes/admin-api/user-admin.ts
kwasniew: I copied this pattern from other places where it was done inline. It does show up in the swagger UI and works fine :)

PR #4147, opened 2023-07-05
chore: avoids code injection through git commit

gastonfournier:
## About the changes
Avoids code injection via the committer name or email

Also removes pushing to s3 on `main` folder as it's not used and we're already pushing to `commits` folder


gardleopard (approved):
LGTM :D

PR #4149, opened 2023-07-05
fix: disallow deletion of single login history entries

nunogois:
https://linear.app/unleash/issue/2-1185/disallow-deletion-of-single-login-entries-in-history-ui

Disallows deletion of single login history entries on the UI.

image


gastonfournier (approved):
LGTM!

ivarconr (approved):

PR #4152, opened 2023-07-05
docs: Events tag

chriswk:
### What
This PR adds documentation for our endpoints that are covered by our "Events" tag. It also adds a type for all valid events, and then uses this as valid values for type argument.


src/lib/openapi/spec/event-schema.ts
```diff
@@ -1,51 +1,136 @@
...
+    additionalProperties: true,
+    properties: {
+        name: {
+            type: 'string',
+        },
+        description: {
+            type: 'string',
+        },
+        type: {
+            type: 'string',
```
gastonfournier, line 17: // I wonder if we can inject the enum type here so we don't have to maintain the list in two places
```suggestion
            type: 'string',
            example: 'APPLICATION_CREATED',
```

src/lib/openapi/spec/event-schema.ts
```diff
@@ -1,51 +1,136 @@
...
+        },
+        stale: {
+            type: 'boolean',
+        },
+        variants: {
+            type: 'array',
+            items: {
+                $ref: '#/components/schemas/variantSchema',
+            },
+        },
```
thomasheartman, line 30: I know these properties are only here to allow the example (or at least that's what I seem to remember), but could we still give them a description? It might be hard because they can vary between events, though 🤔

src/lib/openapi/spec/feature-events-schema.ts
```diff
@@ -7,18 +7,30 @@ export const featureEventsSchema = {
...
     additionalProperties: false,
     required: ['events'],
+    description: 'One or more events happening to a specific feature toggle',
     properties: {
-        version: { type: 'number' },
+        version: {
+            type: 'integer',
+            description: 'An API versioning number',
+            minimum: 1,
+            example: 1,
```
thomasheartman, line 16: Same comment as above regarding enums

thomasheartman (approved):
As always: very good work on this! I've got a few suggestions going through, but nothing major. Feel free to handle and merge as you please.

src/lib/openapi/spec/event-schema.ts
```diff
@@ -1,51 +1,136 @@
 import { FromSchema } from 'json-schema-to-ts';
 import { tagSchema } from './tag-schema';
+import { IEventTypes } from '../../types';
+import { variantSchema } from './variant-schema';
 
+const eventDataSchema = {
+    type: 'object',
+    additionalProperties: true,
```
gastonfournier, line 8: Do we have some required properties? Maybe
```suggestion
    additionalProperties: true,
    required: ['name', 'type'],
```

src/lib/types/events.ts
```diff
@@ -3,115 +3,222 @@ import { FeatureToggle, IStrategyConfig, ITag, IVariant } from './model';
 import { IApiToken } from './models/api-token';
 import { IUser } from './user';
 
-export const APPLICATION_CREATED = 'application-created';
+export const APPLICATION_CREATED = 'application-created' as const;
```
gastonfournier, line 6: TIL: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

gastonfournier (approved):
LGTM!

src/lib/openapi/spec/event-schema.ts
chriswk: Well, I can still give it a go certainly, I can steal with pride from our feature schema :+1:

src/lib/openapi/spec/events-schema.ts
```diff
@@ -7,18 +7,24 @@ export const eventsSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['version', 'events'],
+    description: 'A list of events that has happened in the system',
     properties: {
         version: {
             type: 'integer',
             minimum: 1,
+            description:
+                'The api version of this response. A natural increasing number. Only increases if format changes',
```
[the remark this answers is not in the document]
chriswk: Sure.

src/lib/openapi/spec/event-schema.ts
thomasheartman: Sounds good! Just try and make it clear in some way that the descriptions may apply to different things depending on what the event is for

src/lib/openapi/spec/event-schema.ts
thomasheartman: Good shout! That should be possible

src/test/e2e/services/user-service.e2e.test.ts
```diff
@@ -47,8 +47,8 @@ beforeAll(async () => {
     });
     userStore = stores.userStore;
     const rootRoles = await accessService.getRootRoles();
-    adminRole = rootRoles.find((r) => r.name === RoleName.ADMIN);
-    viewerRole = rootRoles.find((r) => r.name === RoleName.VIEWER);
+    adminRole = rootRoles.find((r) => r.name === RoleName.ADMIN)!!;
+    viewerRole = rootRoles.find((r) => r.name === RoleName.VIEWER)!!;
```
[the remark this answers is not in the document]
chriswk: I think I might've inadvertently used Kotlin syntax here. Let me see what the compiler says with a single bang :P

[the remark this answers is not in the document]
chriswk: Indeed, a single bang is enough. Will update.

src/lib/openapi/spec/event-schema.ts
chriswk: Not really, cause it's not the event type, it could be the feature type here.

chriswk: The example for our feature-created events has `type: 'release'` here. So unfortunately, we can't inject the event type enum here.

src/lib/openapi/spec/event-schema.ts
chriswk: I don't believe we do. There's nothing preventing you from creating an event of `{}`, if you just want to use the event to trigger some other behaviour.

src/lib/openapi/spec/event-schema.ts
gastonfournier: maybe we can have a union type for event-type + feature-type, but I think that'd be requesting too much on this PR. This is good for me :+1:

src/lib/types/events.ts
```diff
@@ -3,115 +3,222 @@ import { FeatureToggle, IStrategyConfig, ITag, IVariant } from './model';
...
+export const API_TOKEN_DELETED = 'api-token-deleted' as const;
+
+export const FEATURE_FAVORITED = 'feature-favorited' as const;
+export const FEATURE_UNFAVORITED = 'feature-unfavorited' as const;
+export const PROJECT_FAVORITED = 'project-favorited' as const;
+export const PROJECT_UNFAVORITED = 'project-unfavorited' as const;
+export const FEATURES_EXPORTED = 'features-exported' as const;
+export const FEATURES_IMPORTED = 'features-imported' as const;
+
+export const IEventTypes = [
```
gastonfournier, line 120: @chriswk by making it this strict, we can't build unleash-enterprise, as there are types in enterprise which are not defined here: https://github.com/ivarconr/unleash-enterprise/actions/runs/5473805611/jobs/9967754171 I think we should make it string again

PR #4157, opened 2023-07-06
fix: avoid expression injection

gastonfournier:
## About the changes
Using toJSON should escape any potentially harmful content from the username and email input


sighphyre (approved):
G

PR #4159, opened 2023-07-06
fix: disallow deletion of all login history entries

nunogois:
https://linear.app/unleash/issue/2-1191/disallow-deletion-of-all-login-entries-in-history-ui

Disallows deletion of all login history entries on the UI.

image


thomasheartman (approved):
🚀

PR #4161, opened 2023-07-06
chore: install csrf protection

gastonfournier:
## About the changes
Following the recommendation on https://github.com/Unleash/unleash/security/code-scanning/47


gardleopard:
We have cookie policy same site so this is not something we need to fix.

PR #4163, opened 2023-07-06
fix: enterprise types not recognized

gastonfournier:
## About the changes
Allow extending event type, something that's needed on enterprise where we have additional event types: https://github.com/ivarconr/unleash-enterprise/actions/runs/5473805611/jobs/9967754171#step:10:473

This is a workaround to unblock the CI/CD pipeline, we might have to think this a bit further


chriswk:
A better solution is to add `as const` to the change request events in enterprise. That should fix your problem

chriswk (changes requested):
I vote we keep this, and fix enterprise instead.

chriswk:
Superseded by fixes to type and `as const` ing variables in enterprise

PR #4164, opened 2023-07-06
fix: added service-account events

chriswk:
We're using this in enterprise as well, so we need to add events that could happen there as well.


gastonfournier (approved):
Nice!

gastonfournier:
Still, this requires changes on enterprise to depend on changes in OSS, not sure if this is ideal, but that can be a problem for later

PR #4165, opened 2023-07-06
fix: project tokens can now be created with the correct permissions

sighphyre:
This fixes the creation of project only tokens (done through the project settings). When we redesigned the way tokens are created, we missed project specific tokens, which have their own permission set that wasn't respected. This forces only the project api for token creation to use the CREATE_PROJECT_API_TOKEN instead of the global CREATE_CLIENT_API_TOKEN or CREATE_FRONTEND_API_TOKEN

### DIFF
--- a/frontend/src/component/admin/apiToken/ApiTokenForm/useApiTokenForm.ts
+++ b/frontend/src/component/admin/apiToken/ApiTokenForm/useApiTokenForm.ts
@@ -7,6 +7,7 @@ import {
     ADMIN,
     CREATE_FRONTEND_API_TOKEN,
     CREATE_CLIENT_API_TOKEN,
+    CREATE_PROJECT_API_TOKEN,
 } from '@server/types/permissions';
 import { useHasRootAccess } from 'hooks/useHasAccess';
 import { SelectOption } from './TokenTypeSelector/TokenTypeSelector';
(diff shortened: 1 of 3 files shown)


src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -159,9 +158,7 @@ export class ProjectApiTokenController extends Controller {
     ): Promise<any> {
         const createToken = await createApiToken.validateAsync(req.body);
         const { projectId } = req.params;
-        const permissionRequired = tokenTypeToCreatePermission(
-            createToken.type,
-        );
+        const permissionRequired = CREATE_PROJECT_API_TOKEN;
```
sighphyre, line 161: Creation of both frontend and client tokens use the same permission here. Seems to be a feature, not a bug so I've left it alone and allowed it to use that exact permission

frontend/src/component/admin/apiToken/ApiTokenForm/useApiTokenForm.ts
```diff
@@ -17,17 +18,21 @@ export const useApiTokenForm = (project?: string) => {
...
             label: `Server-side SDK (${TokenType.CLIENT})`,
             title: 'Connect server-side SDK or Unleash Proxy',
-            enabled: useHasRootAccess(CREATE_CLIENT_API_TOKEN),
+            enabled: hasCreateTokenPermission || hasCreateProjectTokenPermission,
         },
     ];
 
     const hasAdminAccess = useHasRootAccess(ADMIN);
     const hasCreateFrontendAccess = useHasRootAccess(CREATE_FRONTEND_API_TOKEN);
+    const hasCreateFrontendTokenAccess = useHasRootAccess(CREATE_PROJECT_API_TOKEN, project);
```
sighphyre, line 35: CREATE_PROJECT_API_TOKEN gives access to create both front end and client tokens so this is intentional

chriswk (approved):
Nice. We just had a customer report this as well, nice to have a fix ready.

gastonfournier (approved):
Looks good!

PR #4174, opened 2023-07-07
1-1049 Emit events after db transaction is complete

thomasheartman:
This PR fixes an issue where events generated during a db transaction would get published before the transaction was complete. This caused errors in some of our services that expected the data to be stored before the transaction had been commited. Refer to [linear issue 1-1049](https://linear.app/unleash/issue/1-1049/event-emitter-should-emit-events-after-db-transaction-is-commited-not) for more info.

Fixes 1-1049.

## Changes

The most important change here is that the `eventStore` no longer emits events when they happen (because that can be in the middle of a transaction). Instead, events are stored with a new `announced` column. The new event announcer service runs on a schedule (every second) and publishes any new events that have not been published. 

Parts of the code have largely been lifted from the `client-application-store`, which uses a similar logic.

I have kept the emitting of the event within the event store because a lot of other services listen to events from this store, so removing that would require a large rewrite. It's something we could look into down the line, but it seems like too much of a change to do right now.

### Terminology:

Published vs announced? We should settle on one or the other. Announced is consistent with the client-application store, but published sounds more fitting for events.

### Publishing and marking events as published

The current implementation fetches all events that haven't been marked as announced, sets them as announced, and then emits them. It's possible that Unleash would crash in the interim or something else might happen, causing the events not to get published. Maybe it would make sense to just fetch the events and only mark them as published after the announcement? On the other hand, that might get us into other problems. Any thoughts on this would be much appreciated.

### DIFF
--- a/src/lib/services/event-announcer-service.ts
+++ b/src/lib/services/event-announcer-service.ts
@@ -23,9 +23,6 @@ export default class EventAnnouncer {
 
     async publishUnannouncedEvents(): Promise<void> {
         const events = await this.eventStore.setUnannouncedToAnnounced();
-        if (events.length) {
-            console.log('about to announce these events', events);
-        }
 
         events.forEach((e) => this.eventEmitter.emit(e.type, e));
     }
--- a/src/lib/services/event-announcer-service.ts
+++ b/src/lib/services/event-announcer-service.ts
@@ -1,16 +1,12 @@
 import { IUnleashConfig } from '../types/option';
 import { IUnleashStores } from '../types/stores';
-import { sharedEventEmitter } from '../util/anyEventEmitter';
 import { Logger } from '../logger';
 import { IEventStore } from '../types/stores/event-store';
-import EventEmitter from 'events';
 
 export default class EventAnnouncer {
     private logger: Logger;
 
     private eventStore: IEventStore;
-
-    private eventEmitter: EventEmitter;
 
     constructor(
         { eventStore }: Pick<IUnleashStores, 'eventStore'>,
@@ -18,7 +14,6 @@ export default class EventAnnouncer {
     ) {
         this.logger = getLogger('services/event-service.ts');
         this.eventStore = eventStore;
-        this.eventEmitter = sharedEventEmitter;
     }
 
     async publishUnannouncedEvents(): Promise<void> {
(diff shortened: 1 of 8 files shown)


src/lib/services/event-announcer-service.ts
```diff
@@ -0,0 +1,22 @@
+import { IUnleashConfig } from '../types/option';
+import { IUnleashStores } from '../types/stores';
+import { Logger } from '../logger';
+import { IEventStore } from '../types/stores/event-store';
+
+export default class EventAnnouncer {
```
kwasniew, line 6: Now that I see this code it seems like the announcer is not doing anything on top of the event store. I'd consider using event store directly and it taking a role of an event announcer without creating a new abstraction. The variable can event be called eventAnnouncer and assigned to the EventStore instance.

kwasniew (approved):
Very very nice. I'm so glad to see it took you so few lines of code to achieve it. Before we merge it I'd remove the setTimeout from the notifications service and test if it solved the original problem that triggered this whole refactoring effort.

src/lib/services/event-announcer-service.ts
thomasheartman: Yeah, I thought about that. The reason I didn't is that we're currently doing all our scheduling in `src/lib/services/index.ts`, which only works with services. 

We could potentially do something similar in `src/lib/db/index.ts`, but that would require getting hold of the services, which we currently don't. 

We could look into either of these solutions, but they both break with an established pattern, so I'm not sure that's preferable over having this admittedly anemic service. 

what do you think?

thomasheartman:
On testing the notifications service: I've linked them locally and can confirm that it works as expected without the timeout. I don't know if there's a good way for us to set that up in a test, though. Got any ideas?

src/lib/services/event-announcer-service.ts
kwasniew: given this context I'd go with your original solution

kwasniew:
It's actually quite tricky to test and even if we find a way it's still part of the notification module test in the enterprise

thomasheartman:
Fair. But that's a good point. I'll go and make sure the necessary changes are ready in enterprise before merging this, so we don't break it without a fix ready.

PR #4177, opened 2023-07-07
fix: bulk tags will work now with project permissions

sjaanus:
Previously creation of tags required UPDATE_FEATURE permission, but creating tags is not related to feature or project, so it was not able to work.
Also bulk update tags was missing projectId from url, so it also did not work properly.

### DIFF
--- a/src/lib/routes/admin-api/project/project-features.ts
+++ b/src/lib/routes/admin-api/project/project-features.ts
@@ -36,10 +36,16 @@ import {
     getStandardResponses,
     ParametersSchema,
     SetStrategySortOrderSchema,
+    TagsBulkAddSchema,
+    TagSchema,
     UpdateFeatureSchema,
     UpdateFeatureStrategySchema,
 } from '../../../openapi';
-import { OpenApiService, FeatureToggleService } from '../../../services';
+import {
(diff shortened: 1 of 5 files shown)


src/lib/routes/admin-api/project/project-features.ts
```diff
@@ -85,6 +91,7 @@ export interface IFeatureProjectUserParams extends ProjectParam {
 
 const PATH = '/:projectId/features';
 const PATH_STALE = '/:projectId/stale';
+const PATH_TAGS = `/:projectId/tags`;
```
sjaanus, line 94: Now updating tags takes in projectId, so it can evaluate permissions.

src/lib/routes/admin-api/tag.ts
```diff
@@ -74,7 +73,7 @@ class TagController extends Controller {
             method: 'post',
             path: '',
             handler: this.createTag,
-            permission: UPDATE_FEATURE,
+            permission: NONE,
```
sjaanus, line 76: Removed project level permission to create tag, because it is not related to project/feature

kwasniew (approved):

PR #4179, opened 2023-07-07
fix: validate min constraint values in openapi

kwasniew:
## About the changes

Screenshot 2023-07-07 at 11 32 45

Previously constraint values were validated inside the feature toggle service not in the openapi. What it means is that change requests that are using the same schema, don't get a validation since it's inside the feature service. So I'm moving the validation of values min length to the openapi schema. Since this field is optional the validation only kicks in when values are provided, not when single value is provided.

Solves: https://linear.app/unleash/issue/1-1095/cr-allows-0-values-for-constraints

### DIFF
--- a/src/lib/openapi/spec/constraint-schema.ts
+++ b/src/lib/openapi/spec/constraint-schema.ts
@@ -39,6 +39,7 @@ export const constraintSchemaBase = {
             items: {
                 type: 'string',
             },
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):
This looks good to me! Because this has always been the intended use case, then I think it's a great change to move forward with 🙌🏼

PR #4181, opened 2023-07-07
fix: remove dangerouslySetInnerHTML

sjaanus:
With using dangerouslySetInnerHTML we are often vulnerable to different injections. Lets try to avoid it until really necessary.

### DIFF
--- a/frontend/src/component/common/Highlighter/Highlighter.tsx
+++ b/frontend/src/component/common/Highlighter/Highlighter.tsx
@@ -1,4 +1,4 @@
-import { VFC } from 'react';
+import { Fragment, VFC } from 'react';
 import { safeRegExp } from '@server/util/escape-regex';
 import { styled } from '@mui/material';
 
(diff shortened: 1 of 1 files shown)


frontend/src/component/common/Highlighter/Highlighter.tsx
```diff
@@ -29,11 +29,18 @@ export const Highlighter: VFC<IHighlighterProps> = ({
...
 
-    return (
-        <StyledSpan
-            dangerouslySetInnerHTML={{
-                __html: children?.replaceAll(regex, '<mark>$&</mark>') || '',
-            }}
-        />
+    const parts = children.split(regex);
+
+    const highlightedText = parts.map((part, index) =>
```
sjaanus, line 34: Instead of using dangerouslySetInnerHTML, we can build the same behaviour from react fragments.

thomasheartman (approved):
Ah, this is cool ⭐

PR #4182, opened 2023-07-07
docs: add description to requests per seconds schemas

chriswk:
### What
Adds a quick and dirty description to requestPerSeconds and segmentedRequestPerSecondsSchema so the enterprise /rps endpoint has better API docs.


src/lib/openapi/spec/requests-per-second-schema.ts
```diff
@@ -3,33 +3,51 @@ import { FromSchema } from 'json-schema-to-ts';
...
+                            'A representation of a single metric to build a line in a graph',
                         properties: {
                             metric: {
                                 description:
                                     'A key value set representing the metric',
                                 type: 'object',
                                 properties: {
                                     appName: {
+                                        description:
+                                            'Name of the application this metric relate to',
```
sighphyre, line 42:
```suggestion
                                            'Name of the application this metric relates to',
```

sighphyre (approved):
LGTM, just one small language thing

gastonfournier (approved):
great!

PR #3914, opened 2023-06-07
fix: remove unneseccary constraint validation request

andreas-unleash:
Fixes over-validating constraint
## About the changes

Closes #  [1-949](https://linear.app/unleash/issue/1-949/fix-overvalidating-constraint)

PR #4183, opened 2023-07-07
fix: constraint validation affecting disabled button

kwasniew:
## About the changes

Screenshot 2023-07-07 at 14 31 02

Fixing regression introduced by https://github.com/Unleash/unleash/pull/3914.
We were not disabling save strategy or add to change request button on invalid constraints because we were filtering out invalid values locally in the frontend and excluding them from checking.

This PR fixes the regression and also introduces unit tests describing the new behavior.

### DIFF
--- a/frontend/src/hooks/api/getters/useConstraintsValidation/useConstraintsValidation.ts
+++ b/frontend/src/hooks/api/getters/useConstraintsValidation/useConstraintsValidation.ts
@@ -2,6 +2,14 @@ import useFeatureApi from 'hooks/api/actions/useFeatureApi/useFeatureApi';
 import { useEffect, useState } from 'react';
 import { IConstraint } from 'interfaces/strategy';
 
+const isValid = (constraint: IConstraint) => {
+    const hasValues =
+        Array.isArray(constraint.values) && constraint.values.length > 0;
+    const hasValue = Boolean(constraint.value);
+
+    return hasValues || hasValue;
+};
+
 export const useConstraintsValidation = (
     constraints?: IConstraint[]
 ): boolean => {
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/api/getters/useConstraintsValidation/useConstraintsValidation.ts
```diff
@@ -16,25 +24,29 @@ export const useConstraintsValidation = (
...
-                    Array.isArray(constraint.values) &&
-                    constraint.values.length > 0;
-                const hasValue = Boolean(constraint.value);
+        const invalidConstraints = constraints.find(item => !isValid(item));
+        if (invalidConstraints) {
+            setValid(false);
+            return;
+        }
 
-                return hasValues || hasValue;
```
kwasniew, line 26: this is wrong. we were excluding locally invalid constraints from the remote check without saying those items are invalid

frontend/src/hooks/api/getters/useConstraintsValidation/useConstraintsValidation.ts
```diff
@@ -16,25 +24,29 @@ export const useConstraintsValidation = (
             return;
         }
 
-        const validationRequests = constraints
-            .filter(constraint => {
-                const hasValues =
-                    Array.isArray(constraint.values) &&
-                    constraint.values.length > 0;
-                const hasValue = Boolean(constraint.value);
+        const invalidConstraints = constraints.find(item => !isValid(item));
```
kwasniew, line 27: check locally invalid constraints that don't require server checks

Tymek: do we highlight what's wrong? it can be confusing when button is disabled and there is no validation error

Tymek (approved):

kwasniew:
@Tymek previously we were not showing anything in this scenario. This is only when you add an empty constraint and try to submit it. When you try to save the constraint we have proper validation.

PR #4184, opened 2023-07-07
fix: delete project dialog cancel redirect

nunogois:
https://linear.app/unleash/issue/2-1204/delete-project-dialog-cancel-button-wrong-redirect

Fixes an issue where the project deletion dialog would redirect if canceled.


chriswk (approved):
LGTM

PR #4186, opened 2023-07-07
fix: add focus style to vertical tabs

nunogois:
https://linear.app/unleash/issue/2-1206/we-need-focus-state-for-vertical-menu

Adds a `focus-visible` style to vertical tabs in order to improve keyboard navigation.

image


chriswk (approved):
LGTM

PR #4190, opened 2023-07-10
refactor: split NoAccessError into ForbiddenError + PermissionError

thomasheartman:
In some of the places we used `NoAccessError` for permissions, other
places we used it for a more generic 403 error with a different
message. This refactoring splits the error type into two distinct
types instead to make the error messages more consistent.


rishiraj88:
.

chriswk (approved):
I like it. LGTM

PR #4191, opened 2023-07-10
docs: Update proxy hosting to point to Frontend API

chriswk:
### What
We had a customer wonder why `/api/proxy/development` wasn't configured for them. After some digging it was found that they'd followed our documentation and assumed that the proxy was configured for them. However, since we no longer host proxies, this was not the case. This PR updates the documentation to point out that "Unleash hosts everything" now means that our clients should use Frontend API keys and the `/api/frontend` endpoint to get what they used to get from embedded proxies.

## Missing
We should make a new illustration in the same vein as the other illustrations on this page to demonstrate what the Frontend API really is. This PR removes the reference to the picture of the topology of having Unleash hosting proxies.

### DIFF
--- a/website/docs/topics/proxy-hosting.mdx
+++ b/website/docs/topics/proxy-hosting.mdx
@@ -36,7 +36,7 @@ This setup is only available to Pro and Enterprise customers.
 
 :::
 
-Unleash no longer hosts instances of the proxy, but makes the [Frontend API](../reference/front-end-api) available to all Pro and Enterprise customers. The API is backed by an Amazon RDS database. Your applications can connect to the frontend API from your own cloud or other hosting solution.
+Unleash no longer hosts instances of the proxy, but makes the [Frontend API](../reference/front-end-api) available to all Pro and Enterprise customers. The API is backed by an Amazon RDS database. Your applications can connect to the frontend API from your own cloud or from other hosting solutions.
 
 In order to access the frontend API you'll need
 - to create a [Frontend API key](../reference/api-tokens-and-client-keys#front-end-tokens) for the environment you'd like to use.
--- a/website/docs/topics/proxy-hosting.mdx
+++ b/website/docs/topics/proxy-hosting.mdx
@@ -50,7 +50,7 @@ While this is easy to get started with, it comes with the limitations described 
 
 :::info Recommendation
 
-We recommend you use [Unleash Edge](../reference/unleash-edge) instead of Unleash Proxy. See our [Edge migration guide](https://github.com/Unleash/unleash-edge/blob/main/migration-guide.md) for a guide on how to migrate
+We recommend you use [Unleash Edge](../reference/unleash-edge) instead of the Unleash proxy. See our [Edge migration guide](https://github.com/Unleash/unleash-edge/blob/main/migration-guide.md) for a guide on how to migrate
 
 :::info Availability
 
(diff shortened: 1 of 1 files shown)


website/docs/topics/proxy-hosting.mdx
```diff
@@ -36,19 +36,22 @@ This setup is only available to Pro and Enterprise customers.
...
+This is the simplest, but also most limited (by far) way to use Unleash client SDKs. In this setup, Unleash hosts both the Unleash API and the Unleash frontend API. With Unleash hosting, you'll only need to worry about the Frontend API keys and the URL to access the endpoint.
 
 While this is easy to get started with, it comes with the limitations described in the [section on tradeoffs between self-hosted and Unleash-hosted setups](#unleash-hosted-or-self-hosted).
 
 ## Unleash hosts the API, you host the proxy
 
+:::info Recommendation
+
+We recommend you use [Unleash Edge](../reference/unleash-edge) instead of Unleash Proxy. See our [Edge migration guide](https://github.com/Unleash/unleash-edge/blob/main/migration-guide.md) for a guide on how to migrate
+
```
chriswk, line 54: I couldn't find a better way to make the recommendation to switch to Edge. Definitely open to suggestions here.

website/docs/topics/proxy-hosting.mdx
```diff
@@ -27,6 +26,7 @@ Hosting the proxy yourself requires a little more setup on your part than the Un
 -   You can arrange for the proxy to be close to your applications, minimizing response times.
 -   You have full control of all your user data. None of the data that the proxy receives is ever sent to the Unleash API. This keeps _your_ data in _your_ hands.
 -   You can easily add more proxies in different regions, as described in the [multi-region](#multi-region) section.
+-   If you have scaling challenges, it could be worth it to have a look at [Unleash Edge](../reference/unleash-edge)
```
thomasheartman, line 29:
```suggestion
-   If you're having scaling challenges, it could be worth it to have a look at [Unleash Edge](../reference/unleash-edge)
```

website/docs/topics/proxy-hosting.mdx
```diff
@@ -36,19 +36,22 @@ This setup is only available to Pro and Enterprise customers.
 
 :::
 
-![An architecture diagram of using the proxy in a setup where Unleash hosts both the API and the proxy.](/img/proxy-hosting-all-unleash.svg)
+Unleash no longer hosts instances of the proxy, but makes the [Frontend API](../reference/front-end-api) available to all Pro and Enterprise customers. The API is backed by an Amazon RDS database. Your applications can connect to the frontend API from your own cloud or other hosting solution.
```
thomasheartman, line 39:
```suggestion
Unleash no longer hosts instances of the proxy, but makes the [Frontend API](../reference/front-end-api) available to all Pro and Enterprise customers. The API is backed by an Amazon RDS database. Your applications can connect to the frontend API from your own cloud or from other hosting solutions.
```

website/docs/topics/proxy-hosting.mdx
thomasheartman: Nah, I think this works well 😄

thomasheartman (approved):
Looks good to me 🧯 Just a few small suggestions on wording, but nothing stopping this from going through!

PR #4193, opened 2023-07-10
fix: return 400 on incorrect client metrics input

andreas-unleash:
Wraps the whole `registerClientMetrics` function with try/catch to return 400 on error
## About the changes

Closes #

### DIFF
--- a/src/lib/routes/client-api/metrics.ts
+++ b/src/lib/routes/client-api/metrics.ts
@@ -65,11 +65,14 @@ export default class ClientMetricsController extends Controller {
     }
 
     async registerMetrics(req: IAuthRequest, res: Response): Promise<void> {
-        const { body: data, ip: clientIp, user } = req;
-        data.environment = this.metricsV2.resolveMetricsEnvironment(user, data);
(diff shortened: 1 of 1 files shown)


thomasheartman (commented):
Seems good. What did it return before? 500? And is the error empty now? What happens when it catches?

ivarconr (approved):

thomasheartman:
Just for reference: my comment is because the error message shown in the screenshot you posted looks like it's coming from OpenAPI and not from the try-catch block. It looks like the code is just returning 400 with an empty body, but I might be misreading that.

andreas-unleash:
I was able to reproduce the original error by passing `../../../../../../../../../../../../../../../../etc/passwd\u0000tap-cancelCoreSubscription` as a feature name when posting metrics.
Turns out the are 2 errors:
1. ClientMetricsStore 
2. FeatureToggleStore (via LastSeenService)
Screenshot 2023-07-11 at 09 39 37

sjaanus (approved):
LG

PR #4194, opened 2023-07-10
docs: add descriptions and examples to tag schemas

gastonfournier:
## About the changes
- Adding descriptions and examples to tag and tag types schemas
- Adding standard errors, summaries, and descriptions to tag and tag types endpoints
- Some improvements on compilation errors


src/lib/services/tag-type-service.ts
```diff
@@ -68,9 +68,11 @@ export default class TagTypeService {
         return Promise.resolve(true);
     }
 
-    async validate(tagType: ITagType): Promise<void> {
+    async validate(tagType: Partial<ITagType> | undefined): Promise<void> {
```
gastonfournier, line 71: Solving a compilation problem

src/lib/services/tag-type-service.ts
```diff
@@ -58,7 +58,7 @@ export default class TagTypeService {
         return data;
     }
 
-    async validateUnique({ name }: Pick<ITagType, 'name'>): Promise<boolean> {
+    async validateUnique(name: string): Promise<boolean> {
```
gastonfournier, line 61: This solves a compilation problem

src/lib/openapi/spec/tag-type-schema.ts
```diff
@@ -4,17 +4,24 @@ export const tagTypeSchema = {
...
         description: {
             type: 'string',
+            description: 'The description of the tag type.',
+            example: 'A tag type for describing the color of a tag.',
         },
         icon: {
             type: 'string',
             nullable: true,
+            description: 'The icon of the tag type.',
+            example: 'not-really-used',
```
thomasheartman, line 24: This example is a little confusing. Are you saying we don't use the icons provided? If they provide icons, what format should it be in?

src/lib/openapi/spec/tag-types-schema.ts
```diff
@@ -6,12 +6,18 @@ export const tagTypesSchema = {
...
     additionalProperties: false,
     required: ['version', 'tagTypes'],
+    description:
+        'A list of tag types with a version number representing the schema used to model the tag types.',
     properties: {
         version: {
             type: 'integer',
+            description:
+                'The version of the schema used to model the tag types.',
+            example: 1,
```
thomasheartman, line 16: How many versions do we have? Could we maybe use an enum to show all the versions here?

src/lib/openapi/spec/tag-with-version-schema.ts
```diff
@@ -6,9 +6,13 @@ export const tagWithVersionSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['version', 'tag'],
+    description:
+        'A tag with a version number representing the schema used to model the tag.',
     properties: {
         version: {
             type: 'integer',
+            description: 'The version of the schema used to model the tag.',
+            example: 1,
```
thomasheartman, line 15: Same as above. Can we and do we want to use enums here?

src/lib/openapi/spec/update-tag-type-schema.ts
```diff
@@ -3,12 +3,17 @@ import { FromSchema } from 'json-schema-to-ts';
...
     properties: {
         description: {
             type: 'string',
+            description: 'The description of the tag type.',
+            example: 'A tag type for describing the color of a tag.',
         },
         icon: {
             type: 'string',
+            description: 'The icon of the tag type.',
+            example: 'not-really-used',
```
thomasheartman, line 16: Same as the previous question regarding tag types.

src/lib/routes/admin-api/tag-type.ts
```diff
@@ -69,8 +77,11 @@ class TagTypeController extends Controller {
                 openApiService.validPath({
                     tags: ['Tags'],
                     operationId: 'createTagType',
+                    summary: 'Create a tag type',
+                    description: 'Create a new tag type.',
```
thomasheartman, line 81: What happens if a tag type with the same name already exists? Is that a 409? We should communicate that to the end user.

src/lib/routes/admin-api/tag-type.ts
```diff
@@ -85,8 +96,12 @@ class TagTypeController extends Controller {
                 openApiService.validPath({
                     tags: ['Tags'],
                     operationId: 'validateTagType',
+                    summary: 'Validate a tag type',
+                    description:
+                        'Validate if the body of the request is a valid tag schema and if the tame given to the tag type already exists, returning a 409 status code if it does.',
```
thomasheartman, line 101: Small clarification suggestion:

```suggestion
                        'Validates whether if the body of the request is a valid tag and whether the a tag type with that name already exists or not. If a tag type with the same name exists, this operation will return a 409 status code.',
```

src/lib/routes/admin-api/tag-type.ts
```diff
@@ -116,8 +134,12 @@ class TagTypeController extends Controller {
                 openApiService.validPath({
                     tags: ['Tags'],
                     operationId: 'updateTagType',
+                    summary: 'Update a tag type',
+                    description:
+                        'Update a tag type data except the name that cannot be changed.',
```
thomasheartman, line 139: We don't list `name` as one of the properties for the update schema, so I don't think we need to mention that explicitly. If you think that's confusing, we can add a note, of course.

```suggestion
                        'Update the configuration for the specified tag type.',
```

src/lib/routes/admin-api/tag-type.ts
```diff
@@ -133,8 +155,12 @@ class TagTypeController extends Controller {
                 openApiService.validPath({
                     tags: ['Tags'],
                     operationId: 'deleteTagType',
+                    summary: 'Delete a tag type',
+                    description:
+                        'Delete a tag type. This will remove the tags associated to that tag type from all features.',
```
thomasheartman, line 160:
```suggestion
                        'Deletes a tag type. If any features have tags of this type, those tags will be deleted.',
```

src/lib/routes/admin-api/tag-type.ts
```diff
@@ -133,8 +155,12 @@ class TagTypeController extends Controller {
...
                     responses: {
                         200: emptyResponse,
+                        ...getStandardResponses(401, 403),
```
thomasheartman, line 163: Do you get a 404 if the tag doesn't exist or just a 200?

src/lib/routes/admin-api/tag.ts
```diff
@@ -78,10 +86,13 @@ class TagController extends Controller {
                 openApiService.validPath({
                     tags: ['Tags'],
                     operationId: 'createTag',
+                    summary: 'Create a new tag.',
+                    description: 'Create a new tag.',
```
thomasheartman, line 90:
```suggestion
                    description: 'Create a new tag with the specified data.',
```

src/lib/routes/admin-api/tag.ts
```diff
@@ -112,8 +127,12 @@ class TagController extends Controller {
                 openApiService.validPath({
                     tags: ['Tags'],
                     operationId: 'getTag',
+                    summary: 'Get a tag by type and value.',
+                    description:
+                        "Get a tag by type and value. It's useful to check if a given tag type and tag value already exists.",
```
thomasheartman, line 132:
```suggestion
                        "Get a tag by type and value. Can be used to check whether a given tag already exists in Unleash or not.",
```

thomasheartman (commented):
Looks good! Most comments are non-blocking suggestions, but I have a few questions I'd like to get an answer to before accepting the change.

src/lib/openapi/spec/tag-type-schema.ts
gastonfournier: We're definitely not using them in the admin UI 

https://sandbox.getunleash.io/enterprise/tag-types/create
image

https://sandbox.getunleash.io/enterprise/tag-types/edit/group
image

But icons do exist in our database and in our API. So I think this could be just an idea that never got implemented. 

Maybe @chriswk who worked on this 2 years ago might remember something.

src/lib/openapi/spec/tag-types-schema.ts
gastonfournier: Only 1 AFAIK, and it will probably never change. This is reserved for major changes, but I don't think modeling a version as an enum is a good idea as adding a new version would be a breaking change for older clients (which will not be gracefully handled)

src/lib/openapi/spec/tag-with-version-schema.ts
gastonfournier: I don't think it's a good idea to model auto-incremental values with enums

src/lib/routes/admin-api/tag-type.ts
gastonfournier: Added here: [bf74c88](https://github.com/Unleash/unleash/pull/4194/commits/bf74c88ce74fd68414cacd8cc89a0c33f7e83ad3)

src/lib/openapi/spec/tag-type-schema.ts
chriswk: What Gaston said, it was meant to be an extra feature so you could paste an URL to a valid icon, so that :slack: / :jira:  had a more obvious illustration than just the first letter of the type capitalised.

chriswk: I don't think it ever got prioritised.

thomasheartman: Got it. Do we have a format for that in any way?

Based on that, I'd definitely put that in the description and maybe even deprecate the properties?

src/lib/openapi/spec/tag-types-schema.ts
thomasheartman: Hmm, I see your reasoning here. However, I feel like giving users a clear description of what values we expect to deliver would be also be good. Maybe an enum and stating explicitly that more versions can be added in the future, but that any breaking changes would go in a major version? 

If not, I'm happy to leave it as is too.

thomasheartman (approved):
I think this is good to go in now 😄 I'd like some clarification on the `icon` property for tag types, but that can always be added later 💁🏼

src/lib/openapi/spec/tag-type-schema.ts
gastonfournier: Right now it can be anything, it will depend on what we want to do when we finally implement this. It's like predicting the future... we can tell customers to write it as a full-url, but maybe in the future we want to use an enum that is not a URL. So :shrug: 

Deprecating it also doesn't feel right because it's like some feature that has always been a work in progress. Maybe we can hide it from the documentation, but it was already there... so this is one of those unfortunate cases where we've dropped the ball and we didn't know about it. 

We can update the description, but also the example helps documenting that it's not being used

thomasheartman: Yeah, that sounds sensible. I'd add a note to the description that this isn't used yet, but might be in the future, then.

src/lib/openapi/spec/tag-types-schema.ts
gastonfournier: > Maybe an enum and stating explicitly that more versions can be added in the future

I think the version being an integer states exactly that

gastonfournier: Talking about it with @chriswk there might be some benefits of having it as an enum. I did some digging and found [this](https://stackoverflow.com/questions/50760778/how-to-do-version-control-with-enums) and [this](https://softwareengineering.stackexchange.com/questions/345773/semantic-versioning-in-openapi-with-string-enum-new-values-ok), none of those were conclusive. I asked ChatGPT and this is what I got:

> Including a version field in the response of an OpenAPI specification can be a good idea in certain situations, especially when dealing with APIs that are subject to frequent updates or have different versions coexisting. However, whether to represent the version field as an enum or a string/integer depends on your specific use case and requirements. Let's discuss the considerations for each option:
> 
> 1. Enum:
> Using an enum to represent the version field in the response has some benefits. It provides a predefined set of allowed values, ensuring that the version is always one of the specified options. This can help with consistency, validation, and avoiding potential typos or accidental variations in the version representation.
> 
> However, using an enum for versioning can also introduce some challenges. Enums are typically used for representing a finite set of known values, assuming that the set won't change frequently. If your API has a dynamic versioning system with a large number of potential versions, it may be impractical to maintain an enum that includes all possible values. Additionally, if the versioning system allows for arbitrary or user-defined values, an enum may not be suitable.
> 
> 2. String or Integer:
> Representing the version field as a string or an integer provides more flexibility compared to an enum. It allows for arbitrary version values, including custom or vendor-specific formats. This flexibility can be advantageous when dealing with complex versioning schemes or when the API consumers need to parse and interpret the version information in a specific way.
> 
> Using a string or integer for the version field also makes it easier to add new versions without modifying the API specification. You can simply introduce a new value without needing to update the enum definition or regenerate client code.
> 
> Ultimately, the decision between using an enum, string, or integer for the version field depends on the specific requirements of your API and how you want to manage versioning. Consider factors such as the expected number of versions, whether the set of versions is known and stable, and how the version information will be used by API consumers.

Overall, this doesn't seem like a decision to take on lightly in this particular API. I think this is worth some general alignment and changing this across the board. 

Most of our versions are numbers, a few are enums and there's no consistency on how we define versions. I think this should be our focus, but for now, I'd like to keep it as an integer which is what majority of our APIs are using

thomasheartman: Yeah, I agree. Thanks for flagging that and having the discussion! Let's figure it out later 😄

PR #4196, opened 2023-07-10
chore: Datadog addon, ability to include source type name

### LINKED ISSUE #4109: Add source_type_name field in Datadog addon payload

I want Unleash addon to have a field `source_type_name` which is required by Datadog to properly render the app source in the UI.
image

### Background

We've been trying to setup a [datadog event integration](https://docs.getunleash.io/reference/addons/datadog) with Unleash enterprise, but we've been having an issue with it being rendered as "My Apps" and not as "Unleash" in the source field, which is confusing.
Looking at [Datadog docs](https://docs.datadoghq.com/api/latest/events/#post-an-event), seems like the request from unleash is missing source_type_name  ([list of other sources](https://docs.datadoghq.com/integrations/faq/list-of-api-source-attribute-value/))
We've tried sending the same events using generic webhook add-on, but that does not seem to go through. Is there anything that we're missing there or how can we get that working?

### Solution suggestions

_No response_

gastonfournier:
## About the changes
Include a new configuration parameter to be able to specify source_type_name. This is an opt-in feature which provides backward compatibility to our existing users.

image

Closes #4109 

## Discussion points
Maybe this should be hardcoded to `Unleash` but this gives additional flexibility


chriswk (approved):
LGTM

PR #4198, opened 2023-07-10
chore: add auto-generated doc index to gitignore

thomasheartman:
I've seen this one pop up a couple times. It's part of our openapi docusaurus integration and should be ignored.


daveleek (approved):
LGTM!

PR #4215, opened 2023-07-11
docs: suggest to use strategy constraints instead of custom strategies

gastonfournier:
## About the changes
Custom activation strategies are still the recommended way to go in our documentation, but now most of the use cases for custom activation strategies can be covered by strategy constraints (released with Unleash 4.16)

With this change, we try to drive people to use strategy constraints.


website/docs/how-to/how-to-use-custom-strategies.md
```diff
@@ -2,6 +2,14 @@
 title: How to use custom activation strategies
 ---
 
+:::tip
+
+Use custom activation strategies only for cases where [strategy constraints](https://docs.getunleash.io/reference/strategy-constraints) don't give you enough control. 
+
+The downside of using a custom activation strategy instead of constraints is that you need to distribute the code of your strategy with Unleash client SDK while on the other hand, [strategy constraints](https://docs.getunleash.io/reference/strategy-constraints) should work out of the box.
```
thomasheartman, line 9: Small rewording suggestion, non-blocking

```suggestion
The downside of using a custom activation strategy instead of constraints is that you need to distribute the code of your strategy with Unleash client SDK while on the other hand, [strategy constraints](https://docs.getunleash.io/reference/strategy-constraints) work without any extra code or maintenance.
```

website/docs/reference/custom-activation-strategies.md
```diff
@@ -6,6 +6,8 @@ title: Custom Activation Strategies
 
 This document is a reference for custom activation strategies. If you're looking for a guide on how to use them, see the [_how to use custom strategies_ guide](../how-to/how-to-use-custom-strategies.md).
 
+Custom activation strategies should be considered an advanced feature because in most cases [strategy constraints](https://docs.getunleash.io/reference/strategy-constraints) provides enough control without the additional complexity.
```
thomasheartman, line 9: Non-blocking

```suggestion
Custom activation strategies should be considered an advanced feature. In most cases [strategy constraints](https://docs.getunleash.io/reference/strategy-constraints) provides enough control without the additional complexity.
```

thomasheartman (approved):
This is great! Two minor suggestions, but both are non-blocking. Super nice work ❤️

PR #4216, opened 2023-07-11
fix: reduce severity of api token middleware errors

chriswk:
This isn't something our on-calls can do something about when it fails, so downgrading to WARN seems like the right level.


sjaanus (approved):
LG

Issue #4205, opened 2023-07-10
feat: Improve staleness handing


ivarconr:
### Describe the feature request

- Make toggle lifetime configurable. 
- Emit events when Unleash marks toggles as potentially stale.

dramamask:
I would love this.

ivarconr:
Your in luck @dramamask, this was included in v5.4 which was released today!

PR #4217, opened 2023-07-11
#4205: mark potentially stale features

thomasheartman:
This PR lays most of the groundwork required for emitting events when features are marked as potentially stale by Unleash. It does **not** emit any events just yet. The summary is:
- periodically look for features that are potentially stale and mark them (set to run every 10 seconds for now; can be changed)
- when features are updated, if the update data contains changes to the feature's type or createdAt date, also update the potentially stale status.

It is currently about 220 lines of tests and about 100 lines of application code (primarily db migration and two new methods on the IFeatureToggleStore interface). 

The reason I wanted to put this into a single PR (instead of just the db migration, then just the potentially stale marking, then the update logic) is:
If users get the db migration first, but not the rest of the update logic until the events are fired, then they could get a bunch of new events for features that should have been marked as potentially stale several days/weeks/months ago. That seemed undesirable to me, so I decided to bunch those changes together. Of course, I'd be happy to break it into smaller parts.

## Rules

A toggle will be marked as potentially stale iff:
- it is not already stale
- its createdAt date is older than its feature type's expected lifetime would dictate

## Migration

The migration adds a new `potentially_stale` column to the features
table and sets this to true for any toggles that have exceeded their
expected lifetime and that have not already been marked as `stale`.

### The `currentTime` parameter of `markPotentiallyStaleFeatures`

The `markPotentiallyStaleFetaures` method takes an optional `currentTime` parameter. This was added to make it easier to test (so you can test "into the future"), but it's not used in the application. We can rewrite the tests to instead update feature toggles manually, but that wouldn't test the actual marking method. Happy to discuss.

### DIFF
--- a/src/lib/db/feature-toggle-store.ts
+++ b/src/lib/db/feature-toggle-store.ts
@@ -377,6 +377,114 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
 
         return toggle;
     }
+
+    async markPotentiallyStaleFeatures(
+        currentTime?: string,
+    ): Promise<string[]> {
+        const query = this.db(TABLE)
+            .update('potentially_stale', true)
+            .whereRaw(
+                `? > (features.created_at + ((
+    SELECT feature_types.lifetime_days
+    FROM feature_types
+    WHERE feature_types.id = features.type
+  ) * INTERVAL '1 day'))`,
+                [currentTime || this.db.fn.now()],
+            );
+
+        const updatedFeatures = await query.returning(FEATURE_COLUMNS);
+        return updatedFeatures.map(({ name }) => name);
+
+        console.log(
+            this.db
+                .raw(
+                    `
+            UPDATE features
+            SET features.potentially_stale = TRUE
+            FROM feature_types
+            WHERE feature_types.id = features.type
+            AND current_time > (features.created_at + (feature_types.lifetime_days * INTERVAL '1 day'))
+            AND features.stale IS NOT TRUE;`,
+                )
+                .toSQL()
+                .toNative(),
+        );
+
+        console.log(
+            this.db(TABLE)
+                .update({ stale: true })
+                .where((builder) =>
+                    builder
+                        .join(
+                            'feature_types',
+                            'feature_types.id',
+                            '=',
+                            'features.type',
+                        )
+                        // .where('feature_types.id', '=', 'features.type')
+                        .where(
+                            this.db.fn.now(),
+                            '>',
+                            "(features.created_at + (feature_types.lifetime_days * INTERVAL '1 day'))",
+                        ),
+                )
+
+                // .join('feature_types', 'feature_types.id', '=', 'features.type')
+                // .from('feature_types')
+                // .whereRaw(
+                //     `CURRENT_TIMESTAMP > (features.created_at + (feature_types.lifetime_days * INTERVAL '1 day'))`,
+                // )
+                // .where(
+                //     'now()',
+                //     '>',
+                //     `(features.created_at + (feature_types.lifetime_days * INTERVAL '1 day'))`,
+                // )
+                .andWhere('features.stale', '!=', true)
+                .returning(FEATURE_COLUMNS)
+                .toSQL()
+                .toNative(),
+        );
+
+        console.log(x);
+
+        return [];
+
+        // const updatedFeatures = await this.db(TABLE)
+        //     .update({ stale: true })
+        //     .where((builder) =>
+        //         builder
+        //             // .from('feature_types')
+        //             .join(
+        //                 'feature_types',
+        //                 'feature_types.id',
+        //                 '=',
+        //                 'features.type',
+        //             )
+        //             // .where('feature_types.id', '=', 'features.type')
+        //             .where(
+        //                 this.db.fn.now(),
+        //                 '>',
+        //                 "(features.created_at + (feature_types.lifetime_days * INTERVAL '1 day'))",
+        //             ),
+        //     )
+
+        //     // .join('feature_types', 'feature_types.id', '=', 'features.type')
+        //     // .from('feature_types')
+        //     // .whereRaw(
+        //     //     `CURRENT_TIMESTAMP > (features.created_at + (feature_types.lifetime_days * INTERVAL '1 day'))`,
+        //     // )
+        //     // .where(
+        //     //     'now()',
+        //     //     '>',
+        //     //     `(features.created_at + (feature_types.lifetime_days * INTERVAL '1 day'))`,
+        //     // )
+        //     .andWhere('features.stale', '!=', true)
+        //     .returning(FEATURE_COLUMNS);
+
+        console.log('updated features:', updatedFeatures);
+
+        return updatedFeatures.map(({ name }) => name);
+    }
 }
 
 module.exports = FeatureToggleStore;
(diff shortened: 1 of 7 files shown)


sjaanus (commented):
While the update is fine, it will be out of date already tomorrow. We need a scheduler to do it anyways, so question is if there is value of putting this out of date data with migration.

src/lib/db/feature-toggle-store.ts
```diff
@@ -262,7 +262,26 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
+
+        const feature = this.rowToFeature(row[0]);
+        // if a feature toggle's type or createdAt has changed, update its potentially stale status
+        if (!feature.stale && (data.type || data.createdAt)) {
+            await this.db(TABLE)
+                .where({ name: data.name })
+                .update(
+                    'potentially_stale',
+                    this.db.raw(
+                        `COALESCE (? > (features.created_at + ((
```
sjaanus, line 274: Why coalesce?

src/lib/db/feature-toggle-store.ts
```diff
@@ -262,7 +262,26 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             .where({ name: data.name })
             .update(this.dtoToRow(project, data))
             .returning(FEATURE_COLUMNS);
-        return this.rowToFeature(row[0]);
+
+        const feature = this.rowToFeature(row[0]);
+        // if a feature toggle's type or createdAt has changed, update its potentially stale status
```
sjaanus, line 267: Can createdAt change? Is that correct?

src/lib/db/feature-toggle-store.ts
```diff
@@ -377,6 +396,40 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
+                    false,
+                );
+            })
+            .andWhereNot('stale', true);
+
+        const updatedFeatures = await query.returning(FEATURE_COLUMNS);
+        return updatedFeatures.map(({ name }) => name);
+    }
+
+    async getPotentiallyStaleStatus(featureName: string): Promise<boolean> {
```
sjaanus, line 425: `isPotentiallyStale` perhaps is better name

src/lib/services/index.ts
```diff
@@ -125,6 +126,13 @@ export const scheduleServices = async (
         ),
         secondsToMilliseconds(1),
     );
+
+    schedulerService.schedule(
+        featureToggleService.markPotentiallyStaleFeatures.bind(
+            featureToggleService,
+        ),
+        secondsToMilliseconds(10),
```
sjaanus, line 134: This might be a little excessive? I understand we want our event to be as accurate as possible, but perhaps once a minute?

src/migrations/20230711094214-add-potentially-stale-flag.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        ALTER TABLE features ADD COLUMN IF NOT EXISTS potentially_stale boolean;
+        UPDATE features
```
sjaanus, line 7: Do we need this now that we have scheduler?

src/lib/db/feature-toggle-store.ts
```diff
@@ -262,7 +262,26 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             .where({ name: data.name })
             .update(this.dtoToRow(project, data))
             .returning(FEATURE_COLUMNS);
-        return this.rowToFeature(row[0]);
+
+        const feature = this.rowToFeature(row[0]);
+        // if a feature toggle's type or createdAt has changed, update its potentially stale status
+        if (!feature.stale && (data.type || data.createdAt)) {
```
sjaanus, line 268: Should we unstale features also?

src/lib/db/feature-toggle-store.ts
thomasheartman: You can set it manually (and in fact some of the tests rely on it). That said, I don't think we allow you to send it through from the API. Maybe.

src/lib/db/feature-toggle-store.ts
thomasheartman: Yeah, I like it! 341302140a29d9db21748847ea0926f80aa561fc

src/lib/services/index.ts
thomasheartman: Absolutely! It was just something I put in for now. Thanks for picking up on it! I agree that once a minute should be more than enough.

src/migrations/20230711094214-add-potentially-stale-flag.js
thomasheartman: Ah, good point! Maybe we don't. 

However, when the events come into play, we probably do want to keep it. This migration won't cause any events to emit, but if you rely on the scheduler to do it, then that could potentially emit thousands of events because it marks all your old toggles (that you haven't cleaned up) as stale.

What do you think?

src/lib/db/feature-toggle-store.ts
sjaanus: Fair enough, I was mismatching stale and potentially stale.

src/lib/db/feature-toggle-store.ts
sjaanus: Just removing coalesce should work.

src/lib/db/feature-toggle-store.ts
```diff
@@ -377,6 +396,40 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
 
         return toggle;
     }
+
+    async markPotentiallyStaleFeatures(
+        currentTime?: string,
+    ): Promise<string[]> {
+        const query = this.db(TABLE)
+            .update('potentially_stale', true)
```
sjaanus, line 404: Should we also unmark them if configuration changes?

src/migrations/20230711094214-add-potentially-stale-flag.js
sjaanus: Sure, good point.

src/lib/db/feature-toggle-store.ts
```diff
@@ -377,6 +396,40 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
+            })
+            .andWhereNot('stale', true);
+
+        const updatedFeatures = await query.returning(FEATURE_COLUMNS);
+        return updatedFeatures.map(({ name }) => name);
+    }
+
+    async isPotentiallyStale(featureName: string): Promise<boolean> {
+        const result = await this.db(TABLE)
+            .first([...FEATURE_COLUMNS, 'potentially_stale'])
```
sjaanus, line 427: I think we dont need to return more columns than `potentially_stale`

thomasheartman: Agreed! 5e5a3d1c9a253740682eef4bad537e4bd0ab2949

src/lib/db/feature-toggle-store.ts
thomasheartman: Huh, thought I tried that, but .. fixed in 5e5a3d1c9a253740682eef4bad537e4bd0ab2949

src/lib/db/feature-toggle-store.ts
thomasheartman: That should be handled in the update method,  I think. Do you disagree?

sjaanus (approved):
LG

PR #4218, opened 2023-07-11
feat: sort feature strategies

kwasniew:
## About the changes

Feature strategies will be sorted in client features. This is required for the new variant strategies where the order of strategies matters for the variant that will be selected.

## Discussion points

Client toggles cache is updated on event log change. We need to add re-ordering event for the re-order to trigger changes.

I'm wondering if this change could be behind a flag.

### DIFF
--- a/src/lib/db/feature-toggle-client-store.ts
+++ b/src/lib/db/feature-toggle-client-store.ts
@@ -77,6 +77,7 @@ export default class FeatureToggleClientStore
             'fs.disabled as strategy_disabled',
             'fs.parameters as parameters',
             'fs.constraints as constraints',
+            'fs.sort_order as sort_order',
             'segments.id as segment_id',
             'segments.constraints as segment_constraints',
         ] as (string | Raw<any>)[];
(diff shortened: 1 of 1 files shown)


andreas-unleash (approved):
LG

src/lib/db/feature-toggle-client-store.ts
```diff
@@ -208,15 +209,25 @@ export default class FeatureToggleClientStore
...
-            })),
+            strategies: strategies
+                ?.sort((strategy1, strategy2) => {
+                    if (
+                        typeof strategy1.sortOrder === 'number' &&
+                        typeof strategy2.sortOrder === 'number'
+                    ) {
+                        return strategy1.sortOrder - strategy2.sortOrder;
+                    }
+                    return 0;
```
sighphyre, line 220: I feel silly saying this but where do items that end up resolving to 0 actually end up in the final list? I'm guessing this is dependent on the JS engine's implementation of the sort? What if we do something like:

```
strategies.sort((strategy1, strategy2) => {
    const sortOrder1 = typeof strategy1.sortOrder === 'number' ? strategy1.sortOrder : -Infinity;
    const sortOrder2 = typeof strategy2.sortOrder === 'number' ? strategy2.sortOrder : -Infinity;

    return sortOrder1 - sortOrder2;
});
```

I want to say that's both very explicit and deterministic independent of whatever JS feels like doing

sighphyre (approved):
Left a (probably) optional comment but it's not worth blocking the merge if you feel it's okay to ignore so I'm gonna hit approve

sighphyre:
> I'm wondering if this change could be behind a flag.

Does this need a flag? This should be effectively the same result for all of our SDKs so long as we don't change the actual API response

src/lib/db/feature-toggle-client-store.ts
kwasniew: They way I read it is 0 means don't change the order. I'm afraid of arithmetics with Infinity in JS since we'll get NaN soon

sighphyre: Yeah I think it's probably okay, from some reading, most modern JS engines should what we expect here. NaN should also behave like a 0 here too.

This is mostly me nitpicking, I think it's probably okay as is

PR #4219, opened 2023-07-11
fix: min items was breaking import

kwasniew:
## About the changes
Prior to introducing minItems constraint some toggles were created that break this constraint. When we do batch import we get an error from schema validation.

minItems values constraint was introduced to prevent invalid constraints creation in a single item mode. When we do batch import mode it difficult to clean up all the items.


thomasheartman (approved):
Looks like a good change if it's necessary to not break things 💁🏼

PR #4221, opened 2023-07-12
feat: feature creation limit crud together with frontend

sjaanus:
1. Add saving and reading limit to backend with migration
2. Remove old project settings methods
3. UI to disable creation on frontend

image

### DIFF
--- a/frontend/src/component/feature/CreateFeature/CreateFeature.tsx
+++ b/frontend/src/component/feature/CreateFeature/CreateFeature.tsx
@@ -1,5 +1,5 @@
 import FormTemplate from 'component/common/FormTemplate/FormTemplate';
-import { Link, useNavigate } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
 import FeatureForm from '../FeatureForm/FeatureForm';
 import useFeatureForm from '../hooks/useFeatureForm';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
--- a/frontend/src/component/feature/CreateFeature/CreateFeature.tsx
+++ b/frontend/src/component/feature/CreateFeature/CreateFeature.tsx
@@ -13,8 +13,8 @@ import { CF_CREATE_BTN_ID } from 'utils/testIds';
 import { formatUnknownError } from 'utils/formatUnknownError';
 import { GO_BACK } from 'constants/navigate';
 import { Alert, styled } from '@mui/material';
-import useProject from '../../../hooks/api/getters/useProject/useProject';
-import { ConditionallyRender } from '../../common/ConditionallyRender/ConditionallyRender';
+import useProject from 'hooks/api/getters/useProject/useProject';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 
 const StyledAlert = styled(Alert)(({ theme }) => ({
     marginBottom: theme.spacing(2),
--- a/frontend/src/component/feature/CreateFeature/CreateFeature.tsx
+++ b/frontend/src/component/feature/CreateFeature/CreateFeature.tsx
@@ -20,6 +20,17 @@ const StyledAlert = styled(Alert)(({ theme }) => ({
     marginBottom: theme.spacing(2),
 }));
 
+export const isFeatureLimitReached = (
+    featureLimit: number | null | undefined,
+    currentFeatureCount: number
+): boolean => {
+    return (
+        featureLimit !== null &&
+        featureLimit !== undefined &&
+        featureLimit <= currentFeatureCount
+    );
+};
+
 const CreateFeature = () => {
     const { setToastData, setToastApiError } = useToast();
     const { setShowFeedback } = useContext(UIContext);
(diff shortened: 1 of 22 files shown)


src/lib/error/from-legacy-error.ts
```diff
@@ -21,14 +21,14 @@ const getStatusCode = (errorName: string): number => {
             return 400;
         case 'MinimumOneEnvironmentError':
             return 400;
+        case 'InvalidOperationError':
+            return 400;
```
sjaanus, line 25: InvalidOperation should not be related to authorization, so not 403, but rather 400

src/lib/db/project-store.ts
```diff
@@ -486,24 +505,6 @@ class ProjectStore implements IProjectStore {
         return Number(members.count);
     }
 
-    async getProjectSettings(projectId: string): Promise<IProjectSettings> {
```
sjaanus, line 489: Cleaning up leftover code

frontend/src/component/feature/CreateFeature/CreateFeature.tsx
```diff
@@ -1,5 +1,5 @@
 import FormTemplate from 'component/common/FormTemplate/FormTemplate';
-import { useNavigate } from 'react-router-dom';
+import { Link, useNavigate } from 'react-router-dom';
```
kwasniew, line 2: linter waring

frontend/src/component/feature/CreateFeature/CreateFeature.tsx
```diff
@@ -74,6 +83,11 @@ const CreateFeature = () => {
         navigate(GO_BACK);
     };
 
+    const featureLimitReached =
```
kwasniew, line 86: are those conditionals covered with a test?

src/lib/error/from-legacy-error.ts
kwasniew: We use it for features behind a feature flag. We had a discussion about it and agreed on 403. Since it's forbidden behind a feature flag. I wouldn't change this one.

src/lib/openapi/spec/health-overview-schema.ts
```diff
@@ -56,6 +56,13 @@ export const healthOverviewSchema = {
             description:
                 "The project's [collaboration mode](https://docs.getunleash.io/reference/project-collaboration-mode). Determines whether non-project members can submit change requests or not.",
         },
+        featureLimit: {
+            type: 'number',
+            nullable: true,
+            example: 100,
+            description:
+                'A limit of the number of features allowed in the project. Null means no limit.',
```
kwasniew, line 64: `null` if no limit.  -- this is the language pattern we normally use for nulls

src/lib/openapi/spec/project-overview-schema.ts
```diff
@@ -51,6 +51,13 @@ export const projectOverviewSchema = {
             description:
                 "The project's [collaboration mode](https://docs.getunleash.io/reference/project-collaboration-mode). Determines whether non-project members can submit change requests or not.",
         },
+        featureLimit: {
+            type: 'number',
+            nullable: true,
+            example: 100,
+            description:
+                'A limit of the number of features allowed in the project. Null means no limit.',
```
kwasniew, line 59: `null` if no description exists.

kwasniew: "A limit on the number of features" is more idiomatic

src/lib/services/feature-toggle-service.ts
```diff
@@ -921,6 +921,15 @@ class FeatureToggleService {
         this.logger.info(`${createdBy} creates feature toggle ${value.name}`);
         await this.validateName(value.name);
         const exists = await this.projectStore.hasProject(projectId);
+
+        if (
+            this.flagResolver.isEnabled('newProjectLayout') &&
+            (await this.projectStore.isFeatureCreationLimitReached(projectId))
+        ) {
+            throw new InvalidOperationError(
```
kwasniew, line 929: ...and this means it's forbidden so 403

src/lib/types/model.ts
```diff
@@ -196,7 +196,7 @@ export interface IProjectOverview {
     createdAt: Date | undefined;
     stats?: IProjectStats;
     mode: ProjectMode;
-
+    featureLimit?: number;
```
kwasniew, line 199: can we make it always defined and return number | null?

src/lib/types/stores/project-store.ts
```diff
@@ -125,4 +115,6 @@ export interface IProjectStore extends Store<IProject, string> {
         environment: string,
         strategy: CreateFeatureStrategySchema,
     ): Promise<CreateFeatureStrategySchema>;
+
+    isFeatureCreationLimitReached(id: string): Promise<boolean>;
```
kwasniew, line 119: isFeatureLimitReached?

kwasniew (changes requested):
Overall a very good PR. I'd not change the HTTP status code

src/lib/error/from-legacy-error.ts
sjaanus: We either need to change it to something else than 403 or move alot of things away from InvalidOperationError.
We are using 403 for these and it does not make any sense.

```
You can not have an empty string for stickiness.
You can not delete the default project!
You cannot change built in roles.
You can not have an empty string for stickiness.
```

and this is just start of the list. these are not 403s.

If we want to have features that are behind a feature flag, maybe the error should be more explicit by name or something.

src/lib/types/model.ts
sjaanus: Using` featureLimit?: number; `allows you to omit the featureLimit property when creating an object or defining a variable of that type. This can be useful in scenarios where the featureLimit may or may not be provided.

frontend/src/component/feature/CreateFeature/CreateFeature.tsx
sjaanus: Yes now they are

kwasniew (approved):

PR #4222, opened 2023-07-12
Fix: variants-batch

Tymek:
## About the changes
Fix un-awaited promise on batch variant update - reduce function allowed TS to skip Promise type.


gastonfournier (approved):
lg!

PR #4226, opened 2023-07-13
openapi: document operations for admin ui feedback

thomasheartman:
This PR updates the endpoint documentation and schemas related to Admin UI feedback.


kwasniew (approved):

src/lib/routes/admin-api/user-feedback.ts
```diff
@@ -60,8 +67,14 @@ class UserFeedbackController extends Controller {
...
                     tags: ['Admin UI'],
                     operationId: 'updateFeedback',
+                    summary: 'Update Unleash feedback',
+                    description:
+                        'Updates the feedback with the provided ID. Only provided fields are updated. Fields left out are left untouched.',
                     requestBody: createRequestSchema('feedbackSchema'),
-                    responses: { 200: createResponseSchema('feedbackSchema') },
+                    responses: {
+                        200: createResponseSchema('feedbackSchema'),
+                        ...getStandardResponses(400, 401, 415),
```
gastonfournier, line 76: Can we get a 404 if the id does not exist?

thomasheartman: Good question! Let me find out!

thomasheartman: You should! Right now you might get a 500 😳

gastonfournier: I found similar situations on other update endpoints!

src/test/e2e/api/admin/feedback.e2e.test.ts
```diff
@@ -33,7 +33,13 @@ beforeAll(async () => {
         );
     };
 
-    app = await setupAppWithCustomAuth(stores, preHook);
+    app = await setupAppWithCustomAuth(stores, preHook, {
+        experimental: {
+            flags: {
+                strictSchemaValidation: true,
```
gastonfournier, line 39: Nice!

gastonfournier (approved):
Looks good! Maybe we can use this opportunity to improve the NotFound returning a 500, but can also be part of a different PR

src/lib/routes/admin-api/user-feedback.ts
kwasniew: Same here. I try to turn those 5xx into client 4xx errors whenever I test some endpoint. As as general rule of thumb if the endpoint has id or some other query/path param we may need to test for 400 vs 500 handling

thomasheartman: No, it does not. If it doesn't exist, we create that feedback id..

thomasheartman:
@gastonfournier The 500 wasn't because the ID was not found, but because I used an admin token. You need an identifiable user to post and put these.

src/lib/routes/admin-api/user-feedback.ts
thomasheartman: Update: this only returns 500 if you use an admin token (and thus have no identifiable user). @kwasniew , I think we might have discussed cases like this before? I don't know if it's worth adding the handling for this in this one specific place now, but I can definitely add it to the description.

kwasniew: I remember us having this conversation but I don't remember how we solved it. @sjaanus do you remember what we did for the admin tokens that had no user id?

sjaanus: I think we never solved it, but the consensus was that are moving away from admin tokens and endpoints that require user as foreign key, can not work for admin tokens.

thomasheartman: In that case, I'm going to merge it for now and we can update that later if we want to.

PR #4228, opened 2023-07-13
fix: existing stickiness value should be available in the dropdown 

andreas-unleash:
existing stickiness value should be available in the dropdown even if the context field is no longer sticky

Closes # [1-1115](https://linear.app/unleash/issue/1-1115/current-gradual-rollout-stickiness-configurations-should-take-priority)


kwasniew (approved):
I'd add a missing test case if it wasn't present before. There's 3 if statements in the stickiness function without coverage.

PR #4229, opened 2023-07-13
docs: highlight unleash edge

gastonfournier:
Highlight edge in the first paragraph.

Not sure if we should suggest that it should be the preferred choice, but at least this helps to realize there's an alternative already in the first paragraph.


sighphyre (approved):
Nice, thank you!

PR #4230, opened 2023-07-13
test: try serving openapi.json from docs

gastonfournier:
## About the changes
It'd be nice to have our openapi.json (maybe openapi.yaml as well) being part of our docs. This can help auto-generate clients.

The problem is having the only available version under https://us.app.unleash-hosted.com/ushosted/docs/openapi.json that can run out of sync with the generated documentation. Having it as a static asset of the current version of our documentation can be helpful.

Also, the documentation https://us.app.unleash-hosted.com/ushosted/docs/openapi/ allows you to download the raw spec (under the title) but the page generated by docusaurus does not

image

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -590,16 +590,26 @@ module.exports = {
             },
         ],
         [
+            'docusaurus-plugin-remote-content',
+            {
+                name: 'openapi-schema',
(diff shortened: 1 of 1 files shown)


gastonfournier:
https://github.com/Unleash/unleash/issues/4231 this might require some additional work

PR #4232, opened 2023-07-13
feat: strategy variant schema openapi

kwasniew:
## About the changes

Introducing a new openapi concept for the strategy variant. It differs from the old variant in two ways:
* overrides are not supported
* ['name', 'weight', 'weightType', 'stickiness'] are required so weightType and stickiness were added. 

I don't know why weightType and stickiness were not required for the old variants but it makes types align with the application services (tested in a separate POC).

Added WIP in description so that people don't have expectations that it's already implemented.

Create and read variant differ by additionalProperties requirement.

### DIFF
--- a/src/lib/openapi/spec/create-strategy-variant-schema.ts
+++ b/src/lib/openapi/spec/create-strategy-variant-schema.ts
@@ -10,7 +10,8 @@ export const createStrategyVariantSchema = {
     properties: {
         name: {
             type: 'string',
-            description: 'The variant name. Must be unique for this feature toggle',
+            description:
+                'The variant name. Must be unique for this feature toggle',
             example: 'blue_group',
         },
         weight: {
@@ -38,9 +39,12 @@ export const createStrategyVariantSchema = {
             description: 'Extra data configured for this variant',
             properties: {
                 type: {
+                    description:
+                        'The type of the value. Commonly used types are string, json and csv.',
                     type: 'string',
                 },
                 value: {
+                    description: 'The actual value of payload',
                     type: 'string',
                 },
             },
--- a/src/lib/openapi/spec/create-strategy-variant-schema.ts
+++ b/src/lib/openapi/spec/create-strategy-variant-schema.ts
@@ -5,7 +5,7 @@ export const createStrategyVariantSchema = {
     type: 'object',
     additionalProperties: true,
     description:
-        '[WIP] A strategy variant allows you to attach any data to strategies instead of only returning `true`/`false`. Strategy variants take precedence over feature variants.',
+        "This is an experimental property. It may change or be removed as we work on it. Please don't depend on it yet. A strategy variant allows you to attach any data to strategies instead of only returning `true`/`false`. Strategy variants take precedence over feature variants.",
     required: ['name', 'weight', 'weightType', 'stickiness'],
     properties: {
         name: {
--- a/src/lib/openapi/spec/create-strategy-variant-schema.ts
+++ b/src/lib/openapi/spec/create-strategy-variant-schema.ts
@@ -1,0 +1,55 @@
+import { FromSchema } from 'json-schema-to-ts';
+
+export const createStrategyVariantSchema = {
+    $id: '#/components/schemas/createStrategyVariantSchema',
+    type: 'object',
+    additionalProperties: true,
+    description:
+        '[WIP] A strategy variant allows to attach any data to strategies instead of only returning `true`/`false`. Strategy variants take precedence over feature variants.',
+    required: ['name', 'weight', 'weightType', 'stickiness'],
+    properties: {
+        name: {
+            type: 'string',
+            description: 'The variants name. Is unique for this feature toggle',
+            example: 'blue_group',
+        },
+        weight: {
+            type: 'number',
+            description:
+                'The weight is the likelihood of any one user getting this variant. It is a number between 0 and 1000. See the section on [variant weights](https://docs.getunleash.io/reference/feature-toggle-variants#variant-weight) for more information',
+            minimum: 0,
+            maximum: 1000,
+        },
+        weightType: {
+            description:
+                'Set to fix if this variant must have exactly the weight allocated to it. If the type is variable, the weight will adjust so that the total weight of all variants adds up to 1000',
+            type: 'string',
+            example: 'fix',
+        },
+        stickiness: {
+            type: 'string',
+            description:
+                '[Stickiness](https://docs.getunleash.io/reference/feature-toggle-variants#variant-stickiness) is how Unleash guarantees that the same user gets the same variant every time',
+            example: 'custom.context.field',
+        },
+        payload: {
+            type: 'object',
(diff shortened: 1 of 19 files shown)


src/lib/openapi/spec/create-strategy-variant-schema.ts
```diff
@@ -0,0 +1,55 @@
+import { FromSchema } from 'json-schema-to-ts';
+
+export const createStrategyVariantSchema = {
+    $id: '#/components/schemas/createStrategyVariantSchema',
+    type: 'object',
+    additionalProperties: true,
```
thomasheartman, line 6: The default value of this property is `true`, so you don't strictly **need** this line, but you can leave it in if you think it makes sense.

```suggestion
```

src/lib/openapi/spec/create-strategy-variant-schema.ts
```diff
@@ -0,0 +1,55 @@
...
+    description:
+        '[WIP] A strategy variant allows to attach any data to strategies instead of only returning `true`/`false`. Strategy variants take precedence over feature variants.',
```
thomasheartman, line 8:
```suggestion
        '[WIP] A strategy variant allows you to attach any data to strategies instead of only returning `true`/`false`. Strategy variants take precedence over feature variants.',
```

src/lib/openapi/spec/create-strategy-variant-schema.ts
```diff
@@ -0,0 +1,55 @@
...
+    required: ['name', 'weight', 'weightType', 'stickiness'],
+    properties: {
+        name: {
+            type: 'string',
+            description: 'The variants name. Is unique for this feature toggle',
```
thomasheartman, line 13: Out of curiosity: do we use the same variants as for strategies or do you now have two different sets of variants (one for strategies and one for the feature)?

```suggestion
            description: 'The variant name. Must be unique for this feature toggle',
```

src/lib/openapi/spec/create-strategy-variant-schema.ts
```diff
@@ -0,0 +1,55 @@
...
+            example: 'blue_group',
+        },
+        weight: {
+            type: 'number',
```
thomasheartman, line 17: This should probably be an integer, right?

```suggestion
            type: 'integer',
```

src/lib/openapi/spec/create-strategy-variant-schema.ts
```diff
@@ -0,0 +1,55 @@
...
+            description:
+                'The weight is the likelihood of any one user getting this variant. It is a number between 0 and 1000. See the section on [variant weights](https://docs.getunleash.io/reference/feature-toggle-variants#variant-weight) for more information',
```
thomasheartman, line 19: Relates to the previous comment. Discard or accept depending on that.

```suggestion
                'The weight is the likelihood of any one user getting this variant. It is an integer between 0 and 1000. See the section on [variant weights](https://docs.getunleash.io/reference/feature-toggle-variants#variant-weight) for more information',
```

src/lib/openapi/spec/create-strategy-variant-schema.ts
```diff
@@ -0,0 +1,55 @@
...
+            minimum: 0,
+            maximum: 1000,
+        },
+        weightType: {
+            description:
+                'Set to fix if this variant must have exactly the weight allocated to it. If the type is variable, the weight will adjust so that the total weight of all variants adds up to 1000',
```
thomasheartman, line 25: Suggest we add a link here.

```suggestion
                'Set to `fix` if this variant must have exactly the weight allocated to it. If the type is `variable`, the weight will adjust so that the total weight of all variants adds up to 1000. Refer to the [variant weight documentation](https://docs.getunleash.io/reference/feature-toggle-variants#variant-weight).',
```

src/lib/openapi/spec/create-strategy-variant-schema.ts
```diff
@@ -0,0 +1,55 @@
...
+            type: 'string',
+            example: 'fix',
+        },
+        stickiness: {
+            type: 'string',
+            description:
+                '[Stickiness](https://docs.getunleash.io/reference/feature-toggle-variants#variant-stickiness) is how Unleash guarantees that the same user gets the same variant every time',
```
thomasheartman, line 32: Is this either the context field or default? 

Maybe something like this? Or we could drop the second sentence entirely. Up to you.

```suggestion
                'The [stickiness](https://docs.getunleash.io/reference/feature-toggle-variants#variant-stickiness) to use for distribution of this variant. Stickiness is how Unleash guarantees that the same user gets the same variant every time',
```

src/lib/openapi/spec/create-strategy-variant-schema.ts
```diff
@@ -0,0 +1,55 @@
...
+            type: 'object',
+            required: ['type', 'value'],
+            description: 'Extra data configured for this variant',
+            properties: {
+                type: {
+                    type: 'string',
+                },
+                value: {
+                    type: 'string',
+                },
```
thomasheartman, line 45: Can we add description for these properties too? Should be the same as in the variant schema, right?

src/lib/openapi/spec/strategy-variant-schema.ts
```diff
@@ -0,0 +1,17 @@
+import { FromSchema } from 'json-schema-to-ts';
+import { createStrategyVariantSchema } from './create-strategy-variant-schema';
+
+export const strategyVariantSchema = {
+    $id: '#/components/schemas/strategyVariantSchema',
+    type: 'object',
+    additionalProperties: false,
+    description:
+        '[WIP] A strategy variant allows to attach any data to strategies instead of only returning `true`/`false`. Strategy variants take precedence over feature variants.',
```
thomasheartman, line 9:
```suggestion
        '[WIP] A strategy variant allows you to attach any data to strategies instead of only returning `true`/`false`. Strategy variants take precedence over feature variants.',
```

thomasheartman (approved):
This is very nice! 💯 I've got a few minor suggestions but no real blockers. What's interesting here, though, is how we add this to the schema. I know you added the `[WIP]` annotation at the start of every description for these schemas. I don't think there is an `unstable` property for schemas, so it might be the best we have for now. But maybe we could make it more explicit in case you don't understand what WIP means. Maybe something like: `This is an experimental property. It may change or be removed as we work on it. Please don't depend on it yet.` or something like that? What do you think?

src/lib/openapi/spec/create-strategy-variant-schema.ts
kwasniew: I prefer the explicit version since I never remember the defaults

src/lib/openapi/spec/create-strategy-variant-schema.ts
kwasniew: I prefer to always provide the stickiness even if it's default. I can relax it later once I start doing UI work

PR #4233, opened 2023-07-13
openapi: add routeWithOpenApi

thomasheartman:
This change adds a routeWithOpenApi method to the base controller
class that we use. That method modifies the open api
configuration passed in and adds standard error codes. It uses that
method in the addon controller.

The goal here is to simplify how we add error codes to the openapi
doc, potentially removing the need for us to do that entirely.

This is just a proof of concept to show that it works. The API for it
could use some work (for instance, having to pass in the service is a
little inconvenient) and possibly some tests, but this shows that
it **is** possible. Might be a nice QOL improvement.

An idea would be to make the controller class `abstract` and require implementers to have an OpenApiService. All of our controllers should already have that, so that might be an easy win.

### DIFF
--- a/src/lib/routes/admin-api/addon.ts
+++ b/src/lib/routes/admin-api/addon.ts
@@ -14,7 +14,6 @@ import {
 import { IAuthRequest } from '../unleash-types';
 import { createRequestSchema } from '../../openapi/util/create-request-schema';
 import { createResponseSchema } from '../../openapi/util/create-response-schema';
-import { OpenApiService } from '../../services/openapi-service';
 import { AddonSchema, addonSchema } from '../../openapi/spec/addon-schema';
 import { serializeDates } from '../../types/serialize-dates';
 import { AddonsSchema, addonsSchema } from '../../openapi/spec/addons-schema';
@@ -30,17 +29,13 @@ class AddonController extends Controller {
 
     private addonService: AddonService;
 
-    private openApiService: OpenApiService;
-
     constructor(
         config: IUnleashConfig,
         { addonService, openApiService }: AddonServices,
     ) {
-        super(config);
+        super(config, { openApiService });
         this.logger = config.getLogger('/admin-api/addon.ts');
         this.addonService = addonService;
-        this.openApiService = openApiService;
-
         this.routeWithOpenApi(this.openApiService)({
             method: 'get',
             path: '',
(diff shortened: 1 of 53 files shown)


src/lib/routes/admin-api/addon.ts
```diff
@@ -44,108 +44,96 @@ class AddonController extends Controller {
...
+                requestBody: createRequestSchema('addonCreateUpdateSchema'),
+                responses: {
+                    200: createResponseSchema('addonSchema'),
+                    ...getStandardResponses(404),
+                },
+            },
         });
 
-        this.route({
+        this.routeWithOpenApi(this.openApiService)({
```
kwasniew, line 120: maybe openApiService can be passed with the super() call in the constructor so that we don't have to repeat it in each route

src/lib/routes/controller.ts
```diff
@@ -105,6 +112,44 @@ export default class Controller {
...
+            ...options
+        }: IRouteOptions & { openApi: ApiOperation }): void => {
+            const errorCodes = new Set<StandardResponseCodes>([401]);
+
+            if (
+                ['put', 'post', 'patch'].includes(
+                    options?.method?.toLowerCase() || '',
+                )
+            ) {
+                errorCodes.add(400);
```
kwasniew, line 127: we can also analyze the path for path params and generate 404 automatically

thomasheartman: Yeah, I thought about that! The problem is that we don't always return 404s if the path param doesn't match anything, though. But I agree that we should, ideally. However, I wouldn't necessarily be that against adding it. What do you think?

src/lib/routes/admin-api/addon.ts
thomasheartman: Yeah, that's definitely also an option! I think I'm personally leaning towards making the class abstract and requiring subclasses to implement it, but it's definitely an option we can explore!

src/lib/routes/controller.ts
thomasheartman: Sure! I think that's primarily for our DELETEs, though, so we could do a check on that too, I suppose? If it has path params and isn't a DELETE, then ...

thomasheartman: Coming back to this: I just read up a bit on it, and I think that deletes actually **should** return 404 if the resource doesn't exist. So I'd be happy to mark that as the intention and we could maybe strive towards making that a reality.

thomasheartman: Realized right now that the 404 won't apply correctly to the token name endpoint you created yesterday, if it returns an empty list when no tokens are found. Would you call that an exception to the rule, or should it actually be a 404? 🤔 I'm kinda torn, to be honest. I can see both being valid options.

PR #4234, opened 2023-07-13
Feat/add strategy update event on strategy ordering

andreas-unleash:
Adds a `feature-strategy-update-event` when the strategy sort-order is changed.

Makes all fields in the eventDataSchema nullable

Closes # [1-11120](https://linear.app/unleash/issue/1-1112/we-should-have-event-for-re-ordering-strategies)


src/lib/services/feature-toggle-service.ts
```diff
@@ -377,17 +377,53 @@ class FeatureToggleService {
             constraints: featureStrategy.constraints || [],
             parameters: featureStrategy.parameters,
             segments: segments.map((segment) => segment.id) ?? [],
+            sortOrder: featureStrategy.sortOrder,
         };
     }
 
     async updateStrategiesSortOrder(
```
kwasniew, line 384: I'd wrap the controller code that calls this method with this.startTransaction. project-features.ts already has everything set up so it's just a matter of adding a transaction boundary.

src/lib/services/feature-toggle-service.ts
```diff
@@ -377,17 +377,53 @@ class FeatureToggleService {
...
         sortOrders: SetStrategySortOrderSchema,
     ): Promise<Saved<any>> {
         await Promise.all(
-            sortOrders.map(async ({ id, sortOrder }) =>
-                this.featureStrategiesStore.updateSortOrder(id, sortOrder),
-            ),
+            sortOrders.map(async ({ id, sortOrder }) => {
+                const strategyToUpdate =
+                    await this.featureStrategiesStore.getStrategyById(id);
+                await this.featureStrategiesStore.updateSortOrder(
```
kwasniew, line 395: can we return the updated strategy from this method so that we don't need to make another DB call on line 399?

src/lib/services/feature-toggle-service.ts
```diff
@@ -377,17 +377,53 @@ class FeatureToggleService {
...
+                    id,
+                    sortOrder,
+                );
+                const updatedStrategy =
+                    await this.featureStrategiesStore.getStrategyById(id);
+
+                const tags = await this.tagStore.getAllTagsForFeature(
+                    featureName,
+                );
+                const segments = await this.segmentService.getByStrategy(
```
kwasniew, line 405: do we need to fetch those segments? i think we only ned data and preData for diffing. Since we know only the order change maybe we can have segments: [] in both variants?

kwasniew: I guess the only problem would be data in the event log showing incorrect value so probably it's not a good idea

src/test/e2e/api/admin/project/features.e2e.test.ts
```diff
@@ -3153,3 +3157,52 @@ test('Enabling a feature environment should add the default strategy when only d
...
+    strategies.forEach((strategy) => {
+        sortOrders.push({ id: strategy.id!, sortOrder: order++ });
+    });
+
+    await app.request
+        .post(
+            `/api/admin/projects/default/features/${featureName}/environments/${envName}/strategies/set-sort-order`,
+        )
+        .send(sortOrders);
+    // .expect()
```
kwasniew, line 3200: missing expect code

src/test/e2e/api/admin/project/features.e2e.test.ts
```diff
@@ -3153,3 +3157,52 @@ test('Enabling a feature environment should add the default strategy when only d
...
+    const response = await app.request.get(`/api/admin/events`);
+    const { body: eventsBody } = response;
+    let { events } = eventsBody;
+
+    for (let i = 0; i < strategies.length; i++) {
```
kwasniew, line 3205: I'd use manual indexes and assertions since for loops break the declarative nature of the tests (e.g. people have to mentally process i++)

src/test/e2e/api/admin/project/features.e2e.test.ts
```diff
@@ -3153,3 +3157,52 @@ test('Enabling a feature environment should add the default strategy when only d
...
+        .post('/api/admin/projects/default/features')
+        .send({ name: featureName })
+        .expect(201);
+
+    await addStrategies(featureName, envName);
+    const { body } = await app.request.get(
+        `/api/admin/projects/default/features/${featureName}/environments/${envName}/strategies`,
+    );
+
+    const strategies: FeatureStrategySchema[] = body;
```
kwasniew, line 3187: can we do something easier to reason about e.g. taking last strategy and moving it to the beginning of the list?

kwasniew (approved):
Non blocking approve but please check the comments. I guess if we migrate this to update strategy then most of the code will disappear.

src/test/e2e/api/admin/project/features.e2e.test.ts
andreas-unleash: 👍

src/test/e2e/api/admin/project/features.e2e.test.ts
andreas-unleash: we would still need all strategies sort-order (as the frontend does this now)

src/lib/services/feature-toggle-service.ts
andreas-unleash: done

src/lib/services/feature-toggle-service.ts
```diff
@@ -369,7 +369,7 @@ class FeatureToggleService {
         featureStrategy: IFeatureStrategy,
         segments: ISegment[] = [],
     ): Saved<IStrategyConfig> {
-        return {
+        let result: Saved<IStrategyConfig> = {
```
kwasniew, line 372: const since you're not re-assigning this variable only modifying one of the properties

src/test/e2e/api/admin/project/features.e2e.test.ts
```diff
@@ -3153,3 +3157,52 @@ test('Enabling a feature environment should add the default strategy when only d
             expect(res.body.strategies[1].disabled).toBeFalsy();
         });
 });
+
+test('Updating feature strategy sort-order should trigger a FeatureStrategyUpdatedEvent', async () => {
+    const envName = 'sort-order-within-environment-one';
+    const featureName = 'feature.sort.order.event.list';
+    // Create environment
```
kwasniew, line 3177: comment not needed

kwasniew (approved):
Nice. 2 minor comments inside

PR #4236, opened 2023-07-13
Persist strategy variants

kwasniew:
## About the changes

Strategy variants can be creates/update/read. When variants are present they are validates with joi and weights are re-calculated as for the old variants.

The test documents how it all works.


andreas-unleash (approved):
non-blocking approve - looks good

PR #4239, opened 2023-07-13
#4205: update to prepare for emitting potentially stale events

thomasheartman:
This PR adds updates the potentially stale status change events whenever the potentially stale update function is run.

No events are emitted yet. While the emission is only a few lines of code, I'd like to do that in a separate PR so that we can give it the attention it deserves in the form of tests, etc.

This PR also moves the potentially stale update functionality from the `update` method to only being done in the `updatePotentiallyStaleFeatures` method. This keeps all functionality related to marking `potentiallyStale` in one place.

The emission implementation was removed in https://github.com/Unleash/unleash/commit/4fb7cbde03b43320442633888f5ba4c0c3c77b31

## The update queries

While it would be possible to do the state updates in a single query instead of three separate ones, wrangling this into knex proved to be troublesome (and would also probably be harder to understand and reason about). The current solution uses three smaller queries (one select, two updates), as Jaanus suggested in a private slack thread.

Input is welcome if you have any.

### DIFF
--- a/src/lib/db/feature-toggle-store.ts
+++ b/src/lib/db/feature-toggle-store.ts
@@ -382,25 +382,7 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
     async updatePotentiallyStaleFeatures(
         currentTime?: string,
     ): Promise<{ name: string; potentiallyStale: boolean }[]> {
-        const previousState = (
-            await this.db(TABLE)
-                .select(['name', 'potentially_stale'])
-                .whereNot('stale', true)
-        ).reduce((previousValue, current) => {
-            previousValue[current.name] = current.potentially_stale;
-            return previousValue;
-        }, {});
-
-        const isPotentiallyStale = this.db.raw(
-            `(? > (features.created_at + ((
-                            SELECT feature_types.lifetime_days
-                            FROM feature_types
-                            WHERE feature_types.id = features.type
-                        ) * INTERVAL '1 day')))`,
-            [currentTime || this.db.fn.now()],
-        );
-
-        const simpleQuery = this.db.raw(
+        const query = this.db.raw(
             `SELECT name, potentially_stale, (? > (features.created_at + ((
                             SELECT feature_types.lifetime_days
                             FROM feature_types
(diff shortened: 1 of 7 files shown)


src/lib/db/feature-toggle-store.ts
```diff
@@ -397,29 +379,53 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
+        const previousState = (
+            await this.db(TABLE)
+                .select(['name', 'potentially_stale'])
+                .whereNot('stale', true)
+        ).reduce((previousValue, current) => {
+            previousValue[current.name] = current.potentially_stale;
+            return previousValue;
+        }, {});
+
+        const updateQuery = this.db(TABLE)
```
sjaanus, line 394: I have a feeling this will update all features every minute. I think we want to write only to database when the flag should change.

thomasheartman: Yeah, agree! I'll see if I can make knex do that. That should also alleviate the need for the manual diff.

sjaanus: Maybe the previousState can already query the `state` the feature has to be after update and previousState.
Then on the next query, you can just send the filtered data.

thomasheartman: That's also a good idea! I thought I'd put it into a single query instead, something like this:

```sql
WITH
  feature_types_data AS (
    SELECT id, lifetime_days
    FROM feature_types),
  current_ts AS (
    SELECT CURRENT_TIMESTAMP AS current_ts),
  comparison AS (
    SELECT
	features.name,
	(current_ts.current_ts > (features.created_at + (feature_types_data.lifetime_days * INTERVAL '1 day'))) AS is_stale
    FROM
	features
    JOIN
	feature_types_data ON features.type = feature_types_data.id
    CROSS JOIN
	current_ts)
UPDATE features
SET potentially_stale = comparison.is_stale
FROM comparison
WHERE features.name = comparison.name
  AND (
    (comparison.is_stale AND NOT potentially_stale = TRUE)
    OR (NOT comparison.is_stale AND potentially_stale = TRUE)
  )
  AND NOT stale = TRUE
RETURNING features.name, potentially_stale;
```

But I can explore both options 💁🏼

thomasheartman: Solution discussed with @sjaanus in Slack was implemented in 66e838d642d363cc06a25ae3adcb9cee4820b57d

PR #4240, opened 2023-07-14
#4205: activate event emission

thomasheartman:
This PR activates the event emission that was prepared for in https://github.com/Unleash/unleash/pull/4239.

It emits events (behind a flag) when something is marked as potentially stale or the opposite.
It takes the features returned from the store and creates events out of them.

The events only contain data, no preData. This is because the preData can easily be inferred and because it gives a nicer event in the event log.

Here is an image of the difference. The top event uses only data, so it shows the name of the feature and the new potentiallyStale status. The bottom event uses both preData and data, so it only shows the new potentiallyStale status and not the feature name (unless you show the raw event):

image

Should not be merged before #4239. Merge that and then rebase this off main or cherry the commit.

### `preData`

Should we also use preData or is it enough to use only data? It seems unnecessary in this event, but I'm open to hearing your thoughts.

### event author: `createdBy`

I've set `unleash-system` as the `createdBy` property on these events because they are generated by the system. I found the same string used some other places. However, it may be that there we want to use a different author.


src/lib/services/feature-toggle-service.ts
```diff
@@ -1967,7 +1968,26 @@ class FeatureToggleService {
     }
 
     async updatePotentiallyStaleFeatures(): Promise<void> {
-        await this.featureToggleStore.updatePotentiallyStaleFeatures();
+        const potentiallyStaleFeatures =
+            await this.featureToggleStore.updatePotentiallyStaleFeatures();
+        if (this.flagResolver.isEnabled('emitPotentiallyStaleEvents')) {
+            if (potentiallyStaleFeatures.length) {
```
thomasheartman, line 1974: We need to check that there are actual events in here, because otherwise the `batchStore` methods throws an error when it fails to insert the empty event data when there are no changes.

src/lib/services/feature-toggle-service.ts
```diff
@@ -1967,7 +1968,26 @@ class FeatureToggleService {
...
+                            data: {
+                                name,
+                                potentiallyStale,
+                            },
+                        }),
+                    ),
+                );
+            }
+        } else {
+            console.log('NOT ENABLED\n\n\n\n');
```
sjaanus, line 1989: Can clear this

src/lib/services/feature-toggle-service.ts
sjaanus, line 1974: Should you compare length to 0?

src/lib/services/feature-toggle-service.ts
thomasheartman: Oh, for sure! Just left that in by mistake.

thomasheartman: 30f1f49e51005a0ddefd7376253c44672a20e1c9

src/lib/services/feature-toggle-service.ts
thomasheartman: Sure, we can do that! Probably more explicit, so I think that's a sensible change.

thomasheartman: cf2cf015ced7d37548c6c7c3e1829ba4ca390550

sjaanus (approved):
LG

PR #4247, opened 2023-07-17
refactor: error param prop

kwasniew:
## About the changes

VariantForm was coupled to feature variant API payload. This level of coupling is not necessary since we only need the information if there was some error outside of the component. This refactoring reduces coupling between the VariantForm and it's parent component. After this change VariantForm can be used for strategy variants without creating artificial API payload.


nunogois (approved):
Yep, smart change 👍

PR #4248, opened 2023-07-17
Feat/feature environment strategy execution reorder

andreas-unleash:
When reordering strategies for a feature environment:
- Adds stop when CR are enabled
- Emits an event 

## About the changes

Closes #

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -431,7 +431,7 @@ class FeatureToggleService {
             })
             .map((strategy) => strategy.id);
 
-        const eventPreData: StrategyIds = existingOrder;
+        const eventPreData: StrategyIds = { strategyIds: existingOrder };
 
         await Promise.all(
             sortOrders.map(async ({ id, sortOrder }) => {
@@ -459,7 +459,7 @@ class FeatureToggleService {
             })
             .map((strategy) => strategy.id);
 
-        const eventData: StrategyIds = newOrder;
+        const eventData: StrategyIds = { strategyIds: newOrder };
         const tags = await this.tagStore.getAllTagsForFeature(featureName);
         const event = new StrategiesOrderChangedEvent({
             featureName,
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -442,18 +442,17 @@ class FeatureToggleService {
         );
 
         const tags = await this.tagStore.getAllTagsForFeature(featureName);
-
-        await this.eventStore.store(
-            new FeatureEnvironmentStrategyExecutionOrderUpdatedEvent({
-                featureName,
-                environment,
-                project,
-                createdBy,
-                preData: eventPreData,
-                data: eventData,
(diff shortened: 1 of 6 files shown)


src/lib/services/feature-toggle-service.ts
```diff
@@ -378,17 +380,92 @@ class FeatureToggleService {
...
+        const { featureName, environment, projectId: project } = context;
+        const existingOrder = (
+            await this.getStrategiesForEnvironment(
+                project,
+                featureName,
+                environment,
+            )
+        )
+            .sort((a, b) =>
+                !!a.sortOrder && !!b.sortOrder
```
sighphyre, line 423: Can we reuse the sort done here: https://github.com/Unleash/unleash/pull/4218, it's a bit easier to read imo and while we should have strategies without sorting order, if we do, I'd rather they were handled in the same way

sighphyre (commented):
This looks significant enough to want tests, or is it covered by existing tests?

src/lib/services/feature-toggle-service.ts
kwasniew: +1

src/lib/types/events.ts
```diff
@@ -326,6 +328,43 @@ export class FeatureEnvironmentEvent extends BaseEvent {
         this.environment = p.environment;
     }
 }
+export type EnvironmentStrategyExecutionOrder = string[];
```
kwasniew, line 331: Can we call it StrategyIds?

src/lib/types/events.ts
```diff
@@ -326,6 +328,43 @@ export class FeatureEnvironmentEvent extends BaseEvent {
...
+export class FeatureEnvironmentStrategyExecutionOrderUpdatedEvent extends BaseEvent {
```
kwasniew, line 332: I'd go for something easier to use in a conversation e.g. StrategiesOrderChangedEvent

kwasniew (changes requested):
I'd write a few tests that show:
* updated order
* event emitted 
* stopped when CR enabled

src/test/e2e/api/admin/project/features.e2e.test.ts
```diff
@@ -3227,3 +3228,130 @@ test('Enabling a feature environment should add the default strategy when only d
...
+    await addStrategies(featureName, envName);
+    const { body } = await app.request.get(
+        `/api/admin/projects/default/features/${featureName}/environments/${envName}/strategies`,
+    );
+
+    const strategies: FeatureStrategySchema[] = body;
+    let order = 1;
+    const sortOrders: SetStrategySortOrderSchema = [];
+
+    strategies.forEach((strategy) => {
```
kwasniew, line 3335: I'd replace forEach with map for readability
```
const strategies: FeatureStrategySchema[] = body;
const sortOrders: SetStrategySortOrderSchema[] = strategies.map((strategy, index) => ({
    id: strategy.id!,
    sortOrder: index + 1
}));
```

kwasniew (approved):
Nice. Minor comments inside.

PR #4254, opened 2023-07-17
feat: remove legacy playground

kwasniew:
## About the changes

* everyone gets a new playground and old one is removed from code
* flag is still left for variant/runtime value control of the results limit
* since the old playground evaluation URL is gone the new playground can take it over (since this API is used for the UI I'm wondering if we can make a breaking change and take over the old URL)

## Discussion points

* are there any other UI components we can remove cc: @andreas-unleash 
* can we have a breaking change in the API cc: @thomasheartman


thomasheartman (changes requested):
This is very nice! But unfortunately, I'm going to say that no, we cannot arbitrarily break the API without doing a new major version bump of Unleash. Sure, it's intended to be used for the UI, but it's not marked as such, and it's just as possible to use it via the API. So we can deprecate the endpoint, but we can't remove it until Unleash 6.

At least, that's my opinion. But I might be missing something and am open to hearing your thoughts.

kwasniew:
@thomasheartman fair point. We can deprecate it in the first step.

kwasniew:
Superseded by https://github.com/Unleash/unleash/pull/4262

PR #4256, opened 2023-07-17
#4205: add openapi definition for the new endpoint

thomasheartman:
This PR adds an operation and accompanying openapi docs for the new "update feature type lifetime" API operation.

It also fixes an oversight where the other endpoint on the same controller didn't use `respondWithValidation`.

Note: the API here is a suggestion. I'd like to hear whether you agree with this implementation or not.


src/lib/routes/admin-api/feature-type.ts
```diff
@@ -50,15 +64,61 @@ export class FeatureTypeController extends Controller {
...
+            UpdateFeatureTypeLifetimeSchema
+        >,
+        res: Response<FeatureTypeSchema>,
+    ): Promise<void> {
+        if (this.flagResolver.isEnabled('configurableFeatureTypeLifetimes')) {
+            throw new NotImplementedError(
+                "This operation isn't implemented yet",
+            );
+        } else {
+            res.status(409).end();
```
Tymek, line 121: `409` is not in responses. It's temporary, right?

Tymek (approved):

src/lib/routes/admin-api/feature-type.ts
thomasheartman: Good point! Yeah, it's only temporary until the endpoint gets stabilized. But good point; I'll add it to the endpoint description for now.

PR #4258, opened 2023-07-17
Group schema updates

kwasniew:
## About the changes

OpenAPI schemas for group, groups and group user


src/lib/openapi/spec/group-schema.ts
```diff
@@ -7,49 +7,67 @@ export const groupSchema = {
     type: 'object',
     additionalProperties: true,
```
thomasheartman, line 8: This is a response schema, right? In that case, this should be `false`
```suggestion
    additionalProperties: false,
```

src/lib/openapi/spec/group-schema.ts
```diff
@@ -7,49 +7,67 @@ export const groupSchema = {
...
     required: ['name'],
+    description: 'A detailed information about a user group',
     properties: {
         id: {
-            type: 'number',
+            description: 'The group id',
+            type: 'integer',
+            example: 1,
```
thomasheartman, line 15: Not blocking and not really important on response schemas, but what's the minimum ID we use? We could indicate that here 💁🏼 

```suggestion
            example: 1,
            minimum: 1, // or 0?
```

src/lib/openapi/spec/group-user-model-schema.ts
```diff
@@ -6,14 +6,19 @@ export const groupUserModelSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['user'],
+    description: 'Details for the single user belonging to a group',
     properties: {
         joinedAt: {
+            description: 'A date when the user join the group',
```
thomasheartman, line 12:
```suggestion
            description: 'The date when the user joined the group',
```

src/lib/openapi/spec/group-user-model-schema.ts
```diff
@@ -6,14 +6,19 @@ export const groupUserModelSchema = {
     type: 'object',
     additionalProperties: false,
     required: ['user'],
+    description: 'Details for the single user belonging to a group',
```
thomasheartman, line 9:
```suggestion
    description: 'Details for a single user belonging to a group',
```

src/lib/openapi/spec/group-user-model-schema.ts
```diff
@@ -6,14 +6,19 @@ export const groupUserModelSchema = {
...
     properties: {
         joinedAt: {
+            description: 'A date when the user join the group',
             type: 'string',
             format: 'date-time',
+            example: '2023-06-30T11:41:00.123Z',
         },
         createdBy: {
+            description: 'A user who added the group user',
```
thomasheartman, line 18: Wow, that's a tricky thing to try to formulate. Maybe something like this? I'm not sure it's better or worse 😅 

```suggestion
            description: 'The username of the user who added this user to this group',
```

thomasheartman (approved):
This looks good! Got a few small suggestions, but nothing preventing this from going in 💯

src/lib/openapi/spec/group-schema.ts
kwasniew: This one is shared for request and response schema. Once I get to the create schemas I may split those two. But for now I only wanted to add examples/descriptions

src/lib/openapi/spec/group-schema.ts
kwasniew: In my DB it's 1 but it's a DB implementation detail so not sure if it adds much value

thomasheartman: Yeah, and because it's not something you can set yourself, it doesn't matter much. I'd set it to `0` so that people can used unsigned integers if they want to, but I'll leave it up to you.

kwasniew: Actually 0 should never occur because postgres starts counting from 1 at least on my machine. But I don't want to make too many assumptions here

PR #4259, opened 2023-07-17
test: add some basic tests to the new slack app

nunogois:
https://linear.app/unleash/issue/2-1244/write-some-tests-for-the-new-slack-app-addon

Pretty self-explanatory. These are based on the tests we have for the other addons.

If you're reading this and have any suggestions of tests we could add at this stage, please let me know.
Thanks!

### DIFF
--- a/src/lib/addons/__snapshots__/slack-app.test.ts.snap
+++ b/src/lib/addons/__snapshots__/slack-app.test.ts.snap
@@ -1,0 +1,127 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Should call Slack API 1`] = `
+{
+  "attachments": [
+    {
+      "actions": [
+        {
+          "name": "featureToggle",
+          "style": "primary",
+          "text": "Open in Unleash",
+          "type": "button",
+          "url": "http://some-url.com/projects/default/features/some-toggle",
+          "value": "featureToggle",
+        },
+      ],
+    },
+  ],
+  "channel": 1,
+  "text": "some@user.com created feature toggle <http://some-url.com/projects/default/features/some-toggle|some-toggle> in project *default*",
+}
+`;
+
+exports[`Should call Slack API for archived toggle 1`] = `
+{
+  "attachments": [
+    {
+      "actions": [
+        {
+          "name": "featureToggle",
+          "style": "primary",
+          "text": "Open in Unleash",
+          "type": "button",
+          "url": "http://some-url.com/projects/default/archive",
+          "value": "featureToggle",
+        },
+      ],
+    },
+  ],
+  "channel": 1,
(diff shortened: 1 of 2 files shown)


src/lib/addons/slack-app.test.ts
```diff
@@ -0,0 +1,173 @@
...
+            name: 'some-toggle',
+        },
+        tags: [{ type: 'slack', value: 'general' }],
+    };
+
+    await addon.handleEvent(event, { accessToken });
+    expect(slackApiCalls.length).toBe(1);
+    expect(slackApiCalls[0].channel).toBe(1);
+    expect(slackApiCalls[0]).toMatchSnapshot();
+});
```
gastonfournier, line 93: This is probably not needed, AFAIK the logic to deliver an event to an addon is elsewhere. SlackAppAddon doesn't have any rejection logic. Doesn't hurt to have more tests though, but just mentioning it :D

src/lib/addons/slack-app.test.ts
```diff
@@ -0,0 +1,173 @@
...
```
gastonfournier, line 117: Ok, similar comment, unless there's something about the mapping that you're covering here

src/lib/addons/__snapshots__/slack-app.test.ts.snap
```diff
@@ -0,0 +1,127 @@
...
+          "text": "Open in Unleash",
+          "type": "button",
+          "url": "http://some-url.com/projects/default/features/some-toggle",
+          "value": "featureToggle",
+        },
+      ],
+    },
+  ],
+  "channel": 2,
+  "text": "some@user.com *disabled* <http://some-url.com/projects/default/features/some-toggle|some-toggle> in *undefined* environment in project *default*",
```
gastonfournier, line 83: Maybe add environment to one of the tests ;)

gastonfournier (approved):
This looks good, just some comments but more tests don't hurt (well, until they do, because your build slows down :sweat_smile:)

src/lib/addons/slack-app.test.ts
nunogois: Yep, but to be fair these tests are there in the old slack addon. I guess they are there simply for the extra snapshots. I can remove these for now.

nunogois: Addressed in https://github.com/Unleash/unleash/pull/4259/commits/34f9b6c4f26d8c885bfc177e4a4ab3d67b5cac42

src/lib/addons/__snapshots__/slack-app.test.ts.snap
nunogois: Addressed in https://github.com/Unleash/unleash/pull/4259/commits/34f9b6c4f26d8c885bfc177e4a4ab3d67b5cac42

gastonfournier (approved):

PR #4262, opened 2023-07-18
feat: advancedPlayground flag used only for runtime control

kwasniew:
## About the changes

* everyone gets a new playground UI and the old UI is removed from code
* flag is still left for variant/runtime value control of the results limit


src/lib/types/experimental.ts
```diff
@@ -95,10 +95,6 @@ const flags: IFlags = {
         process.env.DISABLE_NOTIFICATIONS,
         false,
     ),
-    advancedPlayground: parseEnvVarBoolean(
```
kwasniew, line 98: we don't use this flag for enabled/disabled control

thomasheartman (approved):
I like it! 😄

PR #4266, opened 2023-07-18
docs: update playground docs to mention advanced features

thomasheartman:
This change updates the playground reference docs with some new images
and an explanation of how the results work now.

I have not removed any of the old images because I think they still
illustrate well how the playground works. And because they're not
direct screenshots anyway, that sounds like a fair trade-off to me.


kwasniew (approved):
I think we're good with the old images for now. Nice update!

PR #4268, opened 2023-07-18
Chore: move event types into a separate reference doc.

thomasheartman:
Remove the list from the events API page and instead point to the new doc.


sighphyre (approved):
LGTM

PR #4273, opened 2023-07-18
docs(#4205): update feature toggle types doc to add lifetime info

thomasheartman:
This change updates the existing feature toggle types document to add
a new `expected lifetime` section. This section explains better what
lifetimes are, where they come from, and that admins can (soon) change
the expected lifetimes for feature toggles.


sighphyre (approved):
Nice!

PR #4281, opened 2023-07-19
feat: strategy variants in playground

kwasniew:
## About the changes

* feature strategies variants in playground - based on Node SDK implementation 
* basic algorithm: for each strategy we calculate one variant and also return a list of all possible variants for the strategy. For the entire evaluation result we choose a first matching strategy and copy it's variant/variants
* by bumping up the Unleash Client SDK also frontend API gets support for strategy variants. It's thoroughly tested in the Node SDK. We can get feedback from prod how it works. After this I'd like to make node SDK non-beta. I prefer to go with frontend API first since we control it and can revert. Once we release Node SDK we don't control it anymore

## Discussion points

I'm pretty confident of the change since the property based tests caught many errors as I was doing the refactor

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "stoppable": "^1.1.0",
     "ts-toolbelt": "^9.6.0",
     "type-is": "^1.6.18",
-    "unleash-client": "4.1.0-beta.3",
+    "unleash-client": "4.1.0-beta.4",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "stoppable": "^1.1.0",
     "ts-toolbelt": "^9.6.0",
     "type-is": "^1.6.18",
-    "unleash-client": "3.21.0",
+    "unleash-client": "4.1.0-beta.3",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "stoppable": "^1.1.0",
     "ts-toolbelt": "^9.6.0",
     "type-is": "^1.6.18",
-    "unleash-client": "4.1.0-beta.4",
+    "unleash-client": "4.1.0-beta.5",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
--- a/src/lib/features/playground/feature-evaluator/client.ts
+++ b/src/lib/features/playground/feature-evaluator/client.ts
@@ -210,7 +210,7 @@ export default class UnleashClient {
         context: Context,
         fallbackVariant?: Variant,
     ): Variant {
-        return this.resolveVariant(name, context, true, fallbackVariant);
+        return this.resolveVariant(name, context, false, fallbackVariant);
     }
 
     private resolveVariant(
@@ -242,10 +242,10 @@ export default class UnleashClient {
         }
 
         if (
-            !enabled ||
             !feature.variants ||
             !Array.isArray(feature.variants) ||
-            feature.variants.length === 0
+            feature.variants.length === 0 ||
+            !feature.enabled
         ) {
             return fallback;
         }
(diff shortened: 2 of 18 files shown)


package.json
```diff
@@ -161,7 +161,7 @@
     "stoppable": "^1.1.0",
     "ts-toolbelt": "^9.6.0",
     "type-is": "^1.6.18",
-    "unleash-client": "3.21.0",
+    "unleash-client": "4.1.0-beta.3",
```
kwasniew, line 164: support for strategy variants that will be used in frontend API but not playground which has its own impl

src/lib/features/playground/feature-evaluator/client.ts
```diff
@@ -110,30 +112,45 @@ export default class UnleashClient {
...
-                        strategySelector.disabled,
-                    ),
+                    ...evaluationResult,
                 };
             },
         );
 
         // Feature evaluation
-        const overallStrategyResult = () => {
+        const overallStrategyResult = (): [
```
kwasniew, line 136: choosing a first matching strategy and its variant/variants

src/lib/features/playground/feature-evaluator/client.ts
```diff
@@ -197,27 +221,35 @@ export default class UnleashClient {
...
             return fallback;
         }
 
         let enabled = true;
         if (checkToggle) {
-            enabled =
-                this.isFeatureEnabled(feature, context, () =>
-                    fallbackVariant ? fallbackVariant.enabled : false,
-                ).result === true;
+            const result = this.isFeatureEnabled(feature, context, () =>
```
kwasniew, line 231: based on Node SDK

src/lib/features/playground/feature-evaluator/strategy/strategy.ts
```diff
@@ -127,10 +130,27 @@ export class Strategy {
         const overallResult =
             constraintResults.result && enabledResult && segmentResults.result;
 
+        const variantDefinition = variantDefinitions
```
kwasniew, line 133: based on Node SDK

src/lib/openapi/spec/playground-strategy-schema.ts
```diff
@@ -51,6 +53,54 @@ export const strategyEvaluationResults = {
                     description:
                         'Whether this strategy evaluates to true or not.',
                 },
+                variant: {
```
kwasniew, line 56: Created inline variant schema that reflects what SDK sees. It's different from the variant in the admin API which has more fields.

src/lib/openapi/spec/playground-strategy-schema.ts
```diff
@@ -51,6 +53,54 @@ export const strategyEvaluationResults = {
...
+                            },
+                        },
+                    },
+                    nullable: true,
+                    example: { name: 'green', enabled: true },
+                },
+                variants: {
+                    type: 'array',
+                    description: 'The feature variants.',
+                    items: { $ref: variantSchema.$id },
```
kwasniew, line 102: this is full variants since we need to display weights in the playground UI

yarn.lock
```diff
@@ -7475,15 +7475,15 @@ universalify@^0.1.0:
   resolved "https://registry.yarnpkg.com/universalify/-/universalify-0.1.2.tgz#b646f69be3942dabcecc9d6639c80dc105efaa66"
   integrity sha512-rBJeI5CXAlmy1pV+617WB9J63U6XcazHHF2f2dbJix4XzpUF0RS3Zbj0FGIOCAva5P/d/GBOYaACQ1w+0azUkg==
 
-unleash-client@3.21.0:
-  version "3.21.0"
-  resolved "https://registry.yarnpkg.com/unleash-client/-/unleash-client-3.21.0.tgz#a31ab30acb42abfb3a21180aa83e4415a3124ec1"
-  integrity sha512-I7eYhRyOia3oBZ9Tu1v+IlNO+XJgsjcMEO2+j+e4A7LTTKZvGoV8WPfDGGxiMPKBPHNUACkERB3YhCQ9jzTGoQ==
+unleash-client@4.1.0-beta.4:
```
kwasniew, line 7478: this also adds strategy variant support for frontend API. no coding needed

Tymek (approved):
LGTM

PR #4283, opened 2023-07-19
chore: Update OpenAPI definitions generated for frontend

Tymek:
## About the changes
`frontend> yarn gen:api` after recent updates


kwasniew (approved):
If compiler is happy I'm happy

PR #4284, opened 2023-07-19
feat: improve slack app addon scalability

nunogois:
https://linear.app/unleash/issue/2-1237/explore-slack-app-addon-scalability-and-limitations

Relevant document: https://linear.app/unleash/document/894e12b7-802c-4bc5-8c22-75af0e66fa4b

 - Implements 30s cache layer for Slack channels;
 - Adds error logging;
 - Adds respective tests;
 - Slight refactors and improvements for overall robustness;

### DIFF
--- a/src/lib/addons/slack-app.test.ts
+++ b/src/lib/addons/slack-app.test.ts
@@ -1,16 +1,17 @@
 import { IEvent, FEATURE_ENVIRONMENT_ENABLED } from '../types/events';
-
 import SlackAppAddon from './slack-app';
-
 import noLogger from '../../test/fixtures/no-logger';
 import { ChatPostMessageArguments } from '@slack/web-api';
 
+let addon;
 const accessToken = 'test-access-token';
 const slackApiCalls: ChatPostMessageArguments[] = [];
+let conversationsList = jest.fn();
+
 jest.mock('@slack/web-api', () => ({
     WebClient: jest.fn().mockImplementation(() => ({
         conversations: {
-            list: () => ({
+            list: conversationsList.mockImplementation(() => ({
                 channels: [
                     {
                         id: 1,
@@ -21,7 +22,7 @@ jest.mock('@slack/web-api', () => ({
                         name: 'another-channel-1',
                     },
                 ],
-            }),
+            })),
         },
         chat: {
             postMessage: jest.fn().mockImplementation((options) => {
@@ -33,14 +34,21 @@ jest.mock('@slack/web-api', () => ({
 }));
 
 beforeEach(() => {
+    jest.useFakeTimers();
     slackApiCalls.length = 0;
-});
-
-test('Should post message when feature is toggled', async () => {
-    const addon = new SlackAppAddon({
+    conversationsList.mockClear();
+    addon = new SlackAppAddon({
         getLogger: noLogger,
         unleashUrl: 'http://some-url.com',
     });
+});
+
+afterEach(() => {
+    jest.useRealTimers();
+    addon.destroy();
+});
+
+test('Should post message when feature is toggled', async () => {
     const event: IEvent = {
         id: 1,
         createdAt: new Date(),
@@ -65,10 +73,6 @@ test('Should post message when feature is toggled', async () => {
 });
 
 test('Should post to all channels in tags', async () => {
-    const addon = new SlackAppAddon({
-        getLogger: noLogger,
-        unleashUrl: 'http://some-url.com',
-    });
     const event: IEvent = {
         id: 2,
         createdAt: new Date(),
@@ -95,10 +99,6 @@ test('Should post to all channels in tags', async () => {
 });
 
 test('Should not post to unexisting tagged channels', async () => {
-    const addon = new SlackAppAddon({
-        getLogger: noLogger,
-        unleashUrl: 'http://some-url.com',
-    });
     const event: IEvent = {
         id: 3,
         createdAt: new Date(),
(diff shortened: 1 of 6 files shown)


src/lib/addons/slack-app.ts
```diff
@@ -17,64 +21,103 @@ interface ISlackAppAddonParameters {
...
+                                        style: 'primary',
+                                        url: featureLink,
+                                    },
+                                ],
+                            },
+                        ],
+                    }),
+                );
+
+                await Promise.all(requests);
```
gastonfournier, line 94: I'm wondering whether we should use `Promise.allSettled` here and just map rejected promises to a warning message, rather than handling an exception if any of the promises failed.

src/lib/addons/slack-app.ts
```diff
@@ -17,64 +21,103 @@ interface ISlackAppAddonParameters {
...
+            } else if (error.code === ErrorCode.RequestError) {
+                this.logger.error(
+                    `Error handling event ${event.type}. A request error occurred: ${error.original}`,
+                    error,
+                );
+            } else if (error.code === ErrorCode.RateLimitedError) {
+                this.logger.error(
+                    `Error handling event ${event.type}. A rate limit error occurred: retry after ${error.retryAfter} seconds`,
+                    error,
+                );
```
gastonfournier, line 112: IIRC, WebClient will handle all these for us, right? If that's the case, should the log be an error or just a warn? Maybe it could be an info message saying that "It will be retried" more than "Retry later" which seems to indicate the client/user has to do something

src/lib/addons/slack-app.ts
```diff
@@ -17,64 +21,103 @@ interface ISlackAppAddonParameters {
...
+
+                await Promise.all(requests);
+                this.logger.info(`Handled event ${event.type}.`);
+            }
+        } catch (error) {
+            if (error.code === ErrorCode.PlatformError) {
+                this.logger.error(
+                    `Error handling event ${event.type}. A platform error occurred: ${error.data}`,
+                    error,
+                );
```
gastonfournier, line 102: Should these errors be "errors"? I don't think we can do much (as SRE oncalls) if there's a surge of errors because SlackAPI is down... I think a warning would be better here

src/lib/addons/slack-app.ts
```diff
@@ -17,64 +21,103 @@ interface ISlackAppAddonParameters {
...
+                    await this.slackClient.conversations.list({
+                        types: 'public_channel,private_channel',
+                    });
+                this.slackChannels = slackConversationsList.channels;
+            }
+
+            const taggedChannels = this.findTaggedChannels(event);
+
+            if (this.slackChannels?.length && taggedChannels.length) {
+                const slackChannelsToPostTo = this.slackChannels.filter(
```
gastonfournier, line 65: I'm not that familiar with the threading model of node, but what if `this.slackChannels` are set to `undefined` after we checked the length? In Java I'd create a local variable or constant to hold the current "view" of `this.slackChannels` to make sure it doesn't change in the middle of my execution

src/lib/addons/slack-app.ts
```diff
@@ -17,64 +21,103 @@ interface ISlackAppAddonParameters {
...
 
-        if (!accessToken) return;
+            if (!accessToken) return;
 
-        if (!this.slackClient) {
-            this.slackClient = new WebClient(accessToken);
-        }
+            if (!this.slackClient || this.accessToken !== accessToken) {
+                this.slackClient = new WebClient(accessToken);
+            }
```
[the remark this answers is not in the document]
nunogois: Oops, forgot about assigning this 🙈 Addressed in https://github.com/Unleash/unleash/pull/4284/commits/b4fa168f4005e6a0d504d6dc60fab4f3b4da9e0c

This check is done in order to initialize a new `WebClient` with the new token in case we change it in the meantime.

src/lib/addons/slack-app.ts
```diff
@@ -17,64 +21,103 @@ interface ISlackAppAddonParameters {
...
+
+            if (!this.slackChannels) {
+                const slackConversationsList =
+                    await this.slackClient.conversations.list({
+                        types: 'public_channel,private_channel',
+                    });
+                this.slackChannels = slackConversationsList.channels;
```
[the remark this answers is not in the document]
nunogois: Addressed in https://github.com/Unleash/unleash/pull/4284/commits/1434f1aebde661e26fc7d55b9029019f1883c8a3

src/lib/addons/slack-app.ts
```diff
@@ -17,64 +21,103 @@ interface ISlackAppAddonParameters {
...
+            }
+
+            const taggedChannels = this.findTaggedChannels(event);
+
+            if (this.slackChannels?.length && taggedChannels.length) {
```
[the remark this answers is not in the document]
nunogois: Agreed. Addressed in https://github.com/Unleash/unleash/pull/4284/commits/4a2d24db668228a3130e34ab1910e48e4a7d9707

We are actually planning on having a `defaultChannel` parameter for when we don't find any tags (some events do not have tags at all). Coming at a later PR.

src/lib/addons/slack-app.ts
nunogois: In this case, this is still an error and should only happen in case `WebClient` does not handle rate limiting for us.
https://slack.dev/node-slack-sdk/web-api#rate-limits
>You might not want to the WebClient to handle rate limits in this way. Perhaps the operation was time sensitive, and it won’t be useful by the time Slack is ready for another request. Or, you have a more sophisticated approach. In these cases, you can set the rejectRateLimitedCalls option on the client to true. Once you set this option, method calls can fail with rate limiting related errors. These errors have a code property set to ErrorCode.RateLimitedError. See [error handling](https://slack.dev/node-slack-sdk/web-api#handle-errors) for more details.

I've added a callback for the `RATE_LIMITED` event so we can log it. I'm keeping this error handling for now, but we can remove it if you think it's better.

src/lib/addons/slack-app.ts
nunogois: Yep, addressed in https://github.com/Unleash/unleash/pull/4284/commits/1d2db10de70b3c2b1cdb3bda3df57cb9b6683131

src/lib/addons/slack-app.ts
nunogois: Fair point, this should be safer: https://github.com/Unleash/unleash/pull/4284/commits/6fdb18ef6041a18d4211c583fcd994153750d210

src/lib/addons/slack-app.ts
nunogois: Nice! Did a major refactor that addresses this in https://github.com/Unleash/unleash/pull/4284/commits/4883ea1eade9e938c1d8246a681273ad35b1af30

gastonfournier: Looks nice!

src/lib/addons/slack-app.ts
gastonfournier: Yeah, I wasn't concerned about logging when `WebClient` receives a rate limit and handles it internally, I was rather thinking that the error might not happen.

It's ok what you did in https://github.com/Unleash/unleash/commit/1d2db10de70b3c2b1cdb3bda3df57cb9b6683131 I'm not sure if `client.on(WebClientEvent.RATE_LIMITED` is needed (or if it has to be a warn, could be maybe a debug), but we can change this later, so no worries for now

src/lib/addons/slack-app.test.ts
```diff
@@ -58,66 +75,120 @@ test('Should post message when feature is toggled', async () => {
...
+    it('should refresh Slack channels cache after 30 seconds', async () => {
+        await addon.handleEvent(event, { accessToken });
+
+        jest.advanceTimersByTime(30000);
+
+        await addon.handleEvent(event, { accessToken });
+
+        expect(slackApiCalls.length).toBe(2);
+        expect(conversationsList).toHaveBeenCalledTimes(2);
+    });
```
gastonfournier, line 154: Nice one!

gastonfournier (approved):
Looks good! Thanks for looking into all my comments!

src/lib/addons/slack-app.ts
nunogois: Ah yeah, fair point. I'll keep it for now then, but I changed it to `debug` for now: https://github.com/Unleash/unleash/pull/4284/commits/a8a7410209bcc74480b03c27c0cf0951c9e457f1

PR #4308, opened 2023-07-21
feat: slack app addon default channels

nunogois:
https://linear.app/unleash/issue/2-1249/add-support-for-default-slack-channels

Adds support for default Slack channels (multiple, comma-separated). 

Some of the events we are handling do not have associated tags, or maybe the tags are empty. This adds a "default Slack channels" parameter to the addon configuration in order to post messages to those channels in those cases.

image

### DIFF
--- a/src/lib/addons/slack-app.ts
+++ b/src/lib/addons/slack-app.ts
@@ -65,7 +65,7 @@ export default class SlackAppAddon extends Addon {
                 : this.getDefaultChannels(defaultChannels);
 
             if (!eventChannels.length) {
-                this.logger.warn(
+                this.logger.debug(
                     `No Slack channels found for event ${event.type}.`,
                 );
                 return;
--- a/src/lib/addons/slack-app.ts
+++ b/src/lib/addons/slack-app.ts
@@ -25,6 +25,7 @@ const CACHE_SECONDS = 30;
 
 interface ISlackAppAddonParameters {
     accessToken: string;
+    defaultChannels: string;
 }
 
 export default class SlackAppAddon extends Addon {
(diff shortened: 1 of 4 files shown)


src/lib/addons/slack-app.ts
```diff
@@ -52,17 +53,20 @@ export default class SlackAppAddon extends Addon {
...
-                this.logger.debug(
-                    `No Slack channels tagged for event ${event.type}`,
-                    event,
+            const eventChannels = taggedChannels.length
+                ? taggedChannels
+                : this.getDefaultChannels(defaultChannels);
+
+            if (!eventChannels.length) {
+                this.logger.warn(
+                    `No Slack channels found for event ${event.type}.`,
```
gastonfournier, line 69: I think this should be just a debug message

src/lib/addons/slack-app.ts
```diff
@@ -52,17 +53,20 @@ export default class SlackAppAddon extends Addon {
         parameters: ISlackAppAddonParameters,
     ): Promise<void> {
         try {
-            const { accessToken } = parameters;
+            const { accessToken, defaultChannels } = parameters;
```
gastonfournier, line 56: I'm wondering if we should opt-in on the events we send to the default channels, I don't think all events are relevant for Slack and it can potentially affect our rate limiting. I know default channels are optional, but I'm just wondering if it's a good thing to do.

gastonfournier: Ok, the events received by handlEvent are already filtered here: https://github.com/Unleash/unleash/blob/34c4dd573ac5081419609128ac64e6d6d8830f24/src/lib/services/addon-service.ts#L114

gastonfournier (approved):

PR #1100, opened 2021-11-10
fix: prevent deadlock for batchinserting usage metrics

ivarconr:
In client metrics v2 we utilize postgres to count the usage
across a few dimensions (feature_name, app_name, environment).

It turns out that if the UPDATE values are not executed in a predictable
order we can end up in a deadlock scenario with postgresql.

**We discovered this while testing in production, behind a feature flag.** 

In this fix we thus sort the metrics on the `feature_name`, `app_name` and
`environment`, to make sure they always are executed in a predictable
order, and thus avoiding independent inserts colliding in to a deadlock
waiting for each-other.

PR #4309, opened 2023-07-21
fix: variant table deadlocks

kwasniew:
## About the changes

Same as https://github.com/Unleash/unleash/pull/1100 but for the variant metrics table.

In client variant metrics v2 we utilize postgres to count the usage
across a few dimensions (feature_name, app_name, environment, variant).

It turns out that if the UPDATE values are not executed in a predictable
order we can end up in a deadlock scenario with postgresql.

In this fix we thus sort the variant metrics on the feature_name, app_name,
environment and variant, to make sure they always are executed in a predictable
order, and thus avoiding independent inserts colliding in to a deadlock
waiting for each-other.


gastonfournier (approved):
Well done!! Thanks!

PR #4312, opened 2023-07-21
fix: Addons toggle

Tymek:
## Issue
- When you you enable/disable an addon, in the same time another one is disabled/enabled
- Sometimes, when clicking enable/disable an addon the toggle switches quick back to the initial state

## About the changes
- Stable table sorting
- Switch toggle `key`

Closes https://linear.app/unleash/issue/1-1154/addons-enabledisable-action-misbehaves-changing-other-items


gastonfournier (approved):
LG!

PR #4334, opened 2023-07-24
fix: group cleanup

kwasniew:
## About the changes

* Group API cleanup triggered by the OpenAPI improvements
* Replacing 500 errors with 400 errors when foreign key violations happen (no role, no user)
* Separating write/create model and read model for groups
* Removing transaction boundaries from the store to the controller in group-controller to have the same pattern across the entire codebase and possibly span multiple stores

### DIFF
--- a/src/lib/db/group-store.ts
+++ b/src/lib/db/group-store.ts
@@ -1,16 +1,15 @@
 import { IGroupStore, IStoreGroup } from '../types/stores/group-store';
-import { Knex } from 'knex';
 import NotFoundError from '../error/notfound-error';
 import Group, {
+    ICreateGroupModel,
+    ICreateGroupUserModel,
     IGroup,
-    IGroupModel,
     IGroupProject,
     IGroupRole,
     IGroupUser,
-    IGroupUserModel,
 } from '../types/group';
-import Transaction = Knex.Transaction;
 import { Db } from './db';
+import { BadDataError, FOREIGN_KEY_VIOLATION } from '../error';
 
 const T = {
     GROUPS: 'groups',
@@ -81,13 +80,23 @@ export default class GroupStore implements IGroupStore {
         return groups.map(rowToGroup);
     }
 
-    async update(group: IGroupModel): Promise<IGroup> {
-        const rows = await this.db(T.GROUPS)
-            .where({ id: group.id })
-            .update(groupToRow(group))
-            .returning(GROUP_COLUMNS);
+    async update(group: ICreateGroupModel): Promise<IGroup> {
+        try {
+            const rows = await this.db(T.GROUPS)
+                .where({ id: group.id })
+                .update(groupToRow(group))
+                .returning(GROUP_COLUMNS);
 
-        return rowToGroup(rows[0]);
+            return rowToGroup(rows[0]);
+        } catch (error) {
+            if (
+                error.code === FOREIGN_KEY_VIOLATION &&
+                error.constraint === 'fk_group_role_id'
+            ) {
+                throw new BadDataError(`Incorrect role id ${group.rootRole}`);
+            }
+            throw error;
+        }
     }
 
     async getProjectGroupRoles(projectId: string): Promise<IGroupRole[]> {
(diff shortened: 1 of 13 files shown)


src/lib/db/group-store.ts
```diff
@@ -176,10 +185,20 @@ export default class GroupStore implements IGroupStore {
...
-            .insert(groupToRow(group))
-            .returning('*');
-        return rowToGroup(row[0]);
+        try {
+            const row = await this.db(T.GROUPS)
+                .insert(groupToRow(group))
+                .returning('*');
+            return rowToGroup(row[0]);
+        } catch (error) {
+            if (
```
kwasniew, line 194: same as update

src/lib/db/group-store.ts
```diff
@@ -190,25 +209,31 @@ export default class GroupStore implements IGroupStore {
...
+            const rows = (users || []).map((user) => {
+                return {
+                    group_id: groupId,
+                    user_id: user.user.id,
+                    created_by: userName,
+                };
+            });
+            return await this.db.batchInsert(T.GROUP_USER, rows);
+        } catch (error) {
+            if (
```
kwasniew, line 225: Screenshot 2023-07-24 at 09 22 02
If user list has invalid user id. We could check this upfront when we validate group in the service but it was more invasive since we'd need to inject read models from the user subdomain. I don't mind having this code here since it's DB related code wrapped in a DB specific store so the details are not leaking out.

src/lib/openapi/spec/create-group-schema.ts
```diff
@@ -0,0 +1,43 @@
+import { FromSchema } from 'json-schema-to-ts';
+import { groupSchema } from './group-schema';
+
+export const createGroupSchema = {
```
kwasniew, line 4: separate schema from the read schema we were sharing before

src/lib/types/group.ts
```diff
@@ -33,6 +33,11 @@ export interface IGroupModel extends IGroup {
     projects?: string[];
 }
 
+export interface ICreateGroupModel extends IGroup {
```
kwasniew, line 36: new TS model reflecting openapi schema change. Users are optional and each user is represented by id only

src/lib/types/services.ts
```diff
@@ -96,4 +96,5 @@ export interface IUnleashServices {
     transactionalFeatureToggleService: (
         db: Knex.Transaction,
     ) => FeatureToggleService;
+    transactionalGroupService: (db: Knex.Transaction) => GroupService;
```
kwasniew, line 99: we'll use it in the enterprise controller

src/test/e2e/services/group-service.e2e.test.ts
```diff
@@ -124,3 +124,26 @@ test('adding a root role to a group with a project role should fail', async () =
 
     expect.assertions(1);
 });
+
+test('adding a nonexistent role to a group should fail', async () => {
```
kwasniew, line 128: confirming the try/catch I added in the store works

src/lib/db/group-store.ts
```diff
@@ -190,25 +209,31 @@ export default class GroupStore implements IGroupStore {
...
-        transaction?: Transaction,
     ): Promise<void> {
-        const rows = (users || []).map((user) => {
-            return {
-                group_id: groupId,
-                user_id: user.user.id,
-                created_by: userName,
-            };
-        });
-        return (transaction || this.db).batchInsert(T.GROUP_USER, rows);
```
kwasniew, line 204: removing transactions from the store and moving them to the controller/around the group service in the enterprise repo

src/lib/db/group-store.ts
sighphyre: Yeah agreed, I think it's fine living in this logical layer

src/lib/db/group-store.ts
```diff
@@ -81,13 +80,23 @@ export default class GroupStore implements IGroupStore {
...
-        return rowToGroup(rows[0]);
+    async update(group: ICreateGroupModel): Promise<IGroup> {
+        try {
+            const rows = await this.db(T.GROUPS)
+                .where({ id: group.id })
+                .update(groupToRow(group))
+                .returning(GROUP_COLUMNS);
+
+            return rowToGroup(rows[0]);
+        } catch (error) {
```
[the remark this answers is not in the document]
sighphyre: If the role doesn't exist at all, it should fail, which I think this now does, thanks to this PR. If the role is not a root role then this will succeed but effectively be a no-op. Probably needs some love but I think this PR takes us forward

sighphyre (approved):
LGTM

PR #4336, opened 2023-07-24
fix: global role is called root role

kwasniew:
## About the changes

We have 'root' role, not 'global' role.


gastonfournier (approved):
LGTM!

PR #4339, opened 2023-07-25
fix: Client metrics name validation

gastonfournier:
## About the changes
1. Add a test for the failing use case (we can see it [here](https://github.com/Unleash/unleash/actions/runs/5656229196/job/15322845002?pr=4339#step:5:783)):
```
FAIL src/lib/services/client-metrics/metrics-service-v2.test.ts
  ● process metrics properly even when some names are not url friendly

    ValidationError: "name" must be URL friendly
```
2. Fix and handle this gracefully

Fixes: https://github.com/Unleash/unleash/pull/4193

### DIFF
--- a/src/lib/services/client-metrics/metrics-service-v2.test.ts
+++ b/src/lib/services/client-metrics/metrics-service-v2.test.ts
@@ -7,7 +7,7 @@ import EventEmitter from 'events';
 import { LastSeenService } from './last-seen-service';
 import { IUnleashConfig } from 'lib/types';
 
-function testSClientMetrics() {
+function testSClientMetrics(flagEnabled = true) {
     const stores = createStores();
 
     const eventBus = new EventEmitter();
@@ -19,7 +19,18 @@ function testSClientMetrics() {
     const lastSeenService = new LastSeenService(stores, config);
     lastSeenService.updateLastSeen = jest.fn();
 
-    const service = new ClientMetricsServiceV2(stores, config, lastSeenService);
+    const flagResolver = {
+        isEnabled: () => {
+            return flagEnabled;
+        },
+    };
+
+    const service = new ClientMetricsServiceV2(
+        stores,
+        config,
+        lastSeenService,
+        flagResolver,
+    );
     return { clientMetricsService: service, eventBus, lastSeenService };
 }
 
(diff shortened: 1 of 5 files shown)


sighphyre (approved):
LGTM

src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -60,6 +64,33 @@ export default class ClientMetricsServiceV2 {
         );
     }
 
+    async validToggleNames(toggleNames: string[]): Promise<string[]> {
+        const nameValidations: Promise<
```
kwasniew, line 68: for of is not very idiomatic. I'd suggest map instead:
```ts
const nameValidations = toggleNames.map((toggle) =>
        nameSchema.validateAsync({ name: toggle })
    );
```

src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -81,20 +112,28 @@ export default class ClientMetricsServiceV2 {
                 ),
         );
 
-        this.logger.debug(`got metrics from ${clientIp}`);
+        const validatedToggleNames = await this.validToggleNames(toggleNames);
```
kwasniew, line 115: I'd prefer functions to be verbs. Maybe `filterValidToggleNames`? Also the variable should be `validToggleNames`

src/lib/services/client-metrics/metrics-service-v2.test.ts
```diff
@@ -0,0 +1,117 @@
+import ClientMetricsServiceV2 from './metrics-service-v2';
+
+import getLogger from '../../../test/fixtures/no-logger';
+
+import createStores from '../../../test/fixtures/store';
+import EventEmitter from 'events';
+import { LastSeenService } from './last-seen-service';
+import { IUnleashConfig } from 'lib/types';
+
+function testSClientMetrics(flagEnabled = true) {
```
kwasniew, line 10: testClientMetrics? Extra 'S'

src/lib/services/client-metrics/metrics-service-v2.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    const stores = createStores();
+
+    const eventBus = new EventEmitter();
+    eventBus.emit = jest.fn();
+
+    // @ts-ignore only add config we care about
```
kwasniew, line 16: do we need to use type assertion (as) and ts-ignore at the same time. Type assertion should do the work. There's also 'as unknown as IUnleashConfig' pattern if something doesn't match.

src/lib/services/client-metrics/metrics-service-v2.test.ts
```diff
@@ -0,0 +1,117 @@
...
+            environment: 'test',
+        },
+        '127.0.0.1',
+    );
+
+    expect(eventBus.emit).toHaveBeenCalledTimes(1);
+    expect(lastSeenService.updateLastSeen).toHaveBeenCalledTimes(1);
+});
+
+test('flag on: process metrics properly even when some names are not url friendly, filtering out invalid names', async () => {
```
kwasniew, line 68: I'd move flag on to the end of the sentence

kwasniew (approved):
Nice tests! Minor comments inside.

src/lib/services/client-metrics/metrics-service-v2.test.ts
gastonfournier: `as IUnleashConfig` didn't work, but `as unknown as IUnleashConfig` worked but I'm not sure which one I prefer. The comment seems to be more explicit

PR #4340, opened 2023-07-25
fix: default strategy stickiness

kwasniew:
## About the changes

Default strategy should get default stickiness from the project settings not from a hardcoded const. 

Create a custom hook that returns default stickiness, fallback and a way to refresh it.

Will add a test in the next PR.


andreas-unleash (approved):
LG

PR #4342, opened 2023-07-25
fix: Client metrics name validation (#4339)

gastonfournier:
1. Add a test for the failing use case (we can see it [here](https://github.com/Unleash/unleash/actions/runs/5656229196/job/15322845002?pr=4339#step:5:783)):
```
FAIL src/lib/services/client-metrics/metrics-service-v2.test.ts
  ● process metrics properly even when some names are not url friendly

    ValidationError: "name" must be URL friendly
```
2. Fix and handle this gracefully
3. Added a new toggle to silently ignore bad names: filterInvalidClientMetrics

Fixes: https://github.com/Unleash/unleash/pull/4193


andreas-unleash (approved):
LG

PR #4345, opened 2023-07-26
fix: sync enabled and variant status

kwasniew:
## About the changes

When we have non 100% gradual rollout with random stickiness enabled status and variant value were getting out of sync. This was because variant is checking isEnabled under the hood. In the new version we also verify isEnabled result (than contains enabled strategy variant). The new force version accepts the explicit evaluation result instead of a boolean flag.

### DIFF
--- a/src/lib/features/playground/feature-evaluator/feature-evaluator.ts
+++ b/src/lib/features/playground/feature-evaluator/feature-evaluator.ts
@@ -104,6 +104,10 @@ export class FeatureEvaluator {
 
     forceGetVariant(
         name: string,
+        forcedResults: Pick<
+            FeatureStrategiesEvaluationResult,
+            'result' | 'variant'
+        >,
         context: Context = {},
         fallbackVariant?: Variant,
     ): Variant {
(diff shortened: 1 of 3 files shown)


src/lib/features/playground/playground-service.ts
```diff
@@ -176,7 +176,11 @@ export class PlaygroundService {
                             data: strategyEvaluationResult.strategies,
                         },
                         projectId: featureProject[feature.name],
-                        variant: client.getVariant(feature.name, clientContext),
+                        variant: client.forceGetVariant(
```
kwasniew, line 179: we're forcing variant to the strategy evaluation result

src/lib/features/playground/feature-evaluator/feature-evaluator.ts
```diff
@@ -104,13 +104,18 @@ export class FeatureEvaluator {
 
     forceGetVariant(
         name: string,
+        forcedResults: Pick<
```
kwasniew, line 107: explicit value to force it to

Tymek (approved):

sighphyre (approved):
Super nice catch, LGTM

PR #4347, opened 2023-07-26
fix: frontend variant weights distribution

Tymek:
## About the changes
Unit-tested way of distributing weights between variants.


frontend/src/component/common/util.ts
```diff
@@ -98,11 +98,12 @@ export function updateWeightEdit(
...
         return [];
     }
-    const { remainingPercentage, variableVariantCount } = variants.reduce(
+    let { remainingPercentage, variableVariantCount } = variants.reduce(
         ({ remainingPercentage, variableVariantCount }, variant) => {
             if (variant.weight && variant.weightType === weightTypes.FIX) {
                 remainingPercentage -= Number(variant.weight);
-            } else {
+            }
+            if (variant.weightType === weightTypes.VARIABLE) {
```
Tymek, line 106: there is was a bugfix here - it was not counting correctly when `weight == 0`

frontend/src/component/common/util.ts
```diff
@@ -113,14 +114,18 @@ export function updateWeightEdit(
         { remainingPercentage: totalWeight, variableVariantCount: 0 }
     );
 
-    const percentage = parseInt(
-        String(remainingPercentage / variableVariantCount)
-    );
+    const getPercentage = () =>
+        Math.round(remainingPercentage / variableVariantCount);
```
Tymek, line 118: When `variableVariantCount == 1` it will return `remainingPercentage`, so there is no special case at the end

frontend/src/component/common/util.ts
```diff
@@ -113,14 +114,18 @@ export function updateWeightEdit(
...
 
     return variants.map(variant => {
         if (variant.weightType !== weightTypes.FIX) {
+            const percentage = getPercentage(); // round "as we go" - clean best effort approach
```
Tymek, line 122: Instead of dividing everything at once, calculate "head" and adjust "tail"

frontend/src/component/common/util.ts
kwasniew: do we need let?

frontend/src/interfaces/featureToggle.ts
```diff
@@ -52,7 +52,7 @@ export interface IFeatureVariant {
     name: string;
     stickiness: string;
     weight: number;
-    weightType: string;
+    weightType: 'fix' | 'variable';
```
kwasniew, line 55: isn't it gonna break openapi schemas?

kwasniew (approved):

frontend/src/interfaces/featureToggle.ts
Tymek: That's in frontend, before we where generating it from OpenAPI. Something to clean up and refactor

Tymek:
we need `let` to substract percentages and variants count as we go

PR #4349, opened 2023-07-26
test: matching variants

kwasniew:
## About the changes

Test selected variant and list of variants match. E.g. disabled variant should not have variants list available.


sighphyre (approved):
Nice

PR #4352, opened 2023-07-27
fix: stable bulk updates

Tymek:
## About the changes
Prevent re-fetched state change from de-selecting rows.


thomasheartman (approved):
This looks like a sensible change to me! Are there any cases where this'll cause your selected rows to change (might happen if it was based on list ordering, for instance)? I'm guessing it's based on some sort of identifier to prevent that?

PR #4355, opened 2023-07-28
openapi: enable strict schema validation by default and fix

thomasheartman:
Enable strict schema validation by default. It can still be overridden by explicitly setting it to false.

I've also fixed the validation errors that appeared when turning it on. I've opted for the simplest route and changed the schemas to comply with the tests.

### DIFF
--- a/src/lib/openapi/spec/public-signup-token-schema.ts
+++ b/src/lib/openapi/spec/public-signup-token-schema.ts
@@ -29,6 +29,7 @@ export const publicSignupTokenSchema = {
             description:
                 'The public signup link for the token. Users who follow this link will be taken to a signup page where they can create an Unleash user.',
             type: 'string',
+            nullable: true,
             example:
                 'https://sandbox.getunleash.io/enterprise/new-user?invite=a3c84b25409ea8ca1782ef17f94a42fc',
         },
--- a/src/lib/openapi/spec/token-user-schema.ts
+++ b/src/lib/openapi/spec/token-user-schema.ts
@@ -6,7 +6,7 @@ export const tokenUserSchema = {
     type: 'object',
     additionalProperties: false,
     description: 'A user identified by a token',
-    required: ['id', 'name', 'email', 'token', 'createdBy', 'role'],
+    required: ['id', 'email', 'token', 'createdBy', 'role'],
     properties: {
         id: {
             type: 'integer',
--- a/src/lib/openapi/spec/user-schema.ts
+++ b/src/lib/openapi/spec/user-schema.ts
@@ -25,6 +25,7 @@ export const userSchema = {
             description: 'Name of the user',
             type: 'string',
             example: 'User',
+            nullable: true,
         },
         email: {
             description: 'Email of the user',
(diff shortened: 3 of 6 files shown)


src/lib/openapi/spec/public-signup-token-schema.ts
```diff
@@ -29,6 +29,7 @@ export const publicSignupTokenSchema = {
             description:
                 'The public signup link for the token. Users who follow this link will be taken to a signup page where they can create an Unleash user.',
             type: 'string',
+            nullable: true,
```
thomasheartman, line 32: I don't know what it means for this to be `null`, but it definitely can be in our tests. Does that mean we need to change our tests or is it actually nullable?

src/lib/openapi/spec/token-user-schema.ts
```diff
@@ -6,7 +6,7 @@ export const tokenUserSchema = {
     type: 'object',
     additionalProperties: false,
     description: 'A user identified by a token',
-    required: ['id', 'name', 'email', 'token', 'createdBy', 'role'],
+    required: ['id', 'email', 'token', 'createdBy', 'role'],
```
thomasheartman, line 9: The `name` property is missing in responses in pretty much all of our tests, so I'm guessing it's not there. Should it be?

In fact,  I don't know if it is ever used 🤔

src/lib/openapi/spec/user-schema.ts
```diff
@@ -25,6 +25,7 @@ export const userSchema = {
             description: 'Name of the user',
             type: 'string',
             example: 'User',
+            nullable: true,
```
thomasheartman, line 28: `name` is null if the user has no name.

src/lib/routes/admin-api/tag-type.ts
```diff
@@ -180,10 +180,16 @@ class TagTypeController extends Controller {
...
+
+        this.openApiService.respondWithValidation(
+            200,
+            res,
+            validateTagTypeSchema.$id,
+            {
+                valid: true,
+                tagType: req.body,
+            },
+        );
```
thomasheartman, line 192: The only thing that's changed here (aside from formatting) is that we use the correct `validateTagTypeSchema.$id` instead of `tagTypeSchema.$id` in the response. This was probably a typo or copy/paste error.

src/test/e2e/helpers/test-helper.ts
```diff
@@ -179,7 +179,16 @@ async function createApp(
...
+        ...{
+            ...customOptions,
+            experimental: {
+                ...(customOptions?.experimental ?? {}),
+                flags: {
+                    strictSchemaValidation: true,
+                    ...(customOptions?.experimental?.flags ?? {}),
+                },
+            },
+        },
```
thomasheartman, line 191: Enable strict schema validation unless it's been explicitly disabled. This applies to all app setup methods.

src/lib/openapi/spec/token-user-schema.ts
thomasheartman: No, we definitely **do** use it sometimes: it's explicitly assigned to in UserService. I'm guessing that we do return it from the reset-password-controller as well if the user has a name.

Tymek (approved):

PR #4368, opened 2023-07-31
chore: clean client api flag removed

kwasniew:
## About the changes

We were explicitly picking what fields to show in the client API behind a flag. Since this flag has been tested in prod for a long time we can remove it now.

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -940,37 +940,33 @@ class FeatureToggleService {
         const result = await this.featureToggleClientStore.getClient(
             query || {},
         );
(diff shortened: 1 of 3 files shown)


src/lib/services/feature-toggle-service.ts
```diff
@@ -940,37 +940,29 @@ class FeatureToggleService {
...
-                }) => ({
-                    name,
-                    type,
-                    enabled,
-                    project,
-                    stale,
-                    strategies,
-                    variants,
-                    description,
-                    createdAt,
```
kwasniew, line 966: createdAt and lastSeenAt should not be exposed anymore according to openapi aschema

thomasheartman (approved):
Looks good! 😄

PR #4371, opened 2023-07-31
feat: strategy variants alert

kwasniew:
## About the changes

Static strategy variants SDK upgrade alert. 
Screenshot 2023-07-31 at 12 08 34

## Discussion points

I'm thinking whether we should do SDK version detection in the backend and return this warning conditionally.


Tymek (approved):

thomasheartman (approved):
Seems like a good change! Regarding SDK detection: we could do that, but what if they haven't used any SDKs yet? In that case it's likely that they'll use the most recent versions of the SDKs, but maybe not 🤷🏼  I think it's safer to do it this way, but I really like the idea 😄

kwasniew:
@thomasheartman I'd say building infrastructure for SDK detection is a bigger initiative that can span beyond this feature. Let's see if the static box will suffice for now :)

PR #4376, opened 2023-08-01
feat: segments service DI

kwasniew:
## About the changes

Segments service assembly so that change requests can instantiate it easily.


andreas-unleash (approved):
LG

PR #4384, opened 2023-08-01
fix: EventStore#getMaxRevisionId can return null

acr92:
In a new fresh Unleash instance with cache enabled this can cause feature toggles to never get updated.

We saw in our client that the ETag was ETag: "60e35fba:null" Which looked incorrect for us.

I also did manual testing and if the andWhere had a value of largerThan higher than whatever the id was then we would get back { max: null }.

This should fix that issue.

### DIFF
--- a/src/lib/db/event-store.ts
+++ b/src/lib/db/event-store.ts
@@ -162,7 +162,7 @@ class EventStore implements IEventStore {
             )
             .andWhere('id', '>=', largerThan)
             .first();
-        return row ? row.max : -1;
+        return row?.max ?? 0;
     }
 
     async delete(key: number): Promise<void> {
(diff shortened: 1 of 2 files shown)


ivarconr (approved):

ivarconr:
A bit surprised the tests are failing.

SimenB:
> A bit surprised the tests are failing.

https://github.com/Unleash/unleash/blob/6c5df9f2c7397ed473144b3866f809f766c68fc2/src/lib/features/feature-toggle/configuration-revision-service.ts#L23-L29 hits `else` now, so maybe whatever sets it to `16` is never called?

The changes in`src/lib/features/feature-toggle/configuration-revision-service.ts` aren't needed (even though they're more correct), but maybe it should be set to `0`? Or just left alone even though the `number` type should then be `number | undefined`

PR #4390, opened 2023-08-02
chore: update ts and ts-jest

thomasheartman:


thomasheartman:
Closing for now due to complications in upgrading. Not that we won't do it, just that I have other things to work on right now.

PR #4391, opened 2023-08-02
Feat: last seen in feature environment

andreas-unleash:
- Adds last_seen_at column in feature_environments and lastSeenAt property to the FeatureEnvironment models

Closes [1-1181](https://linear.app/unleash/issue/1-1181/implement-storing-last-seen-per-environment-be)


sjaanus (approved):
LG, but is it safe enough so feature flag is not needed? Up to you.

PR #4392, opened 2023-08-02
fix: Password hash is null should yield PasswordMissmatch

ivarconr:
We are giving a strange error message if the user does not have a password_hash defined in the database.


src/lib/routes/index.ts
```diff
@@ -34,6 +34,7 @@ class IndexRouter extends Controller {
                 max: 10,
                 standardHeaders: true,
                 legacyHeaders: false,
+                validate: false,
```
ivarconr, line 37: oh, this was a mistake.

ivarconr, line 37:
```suggestion
```

PR #4394, opened 2023-08-03
feat: optional change request feature

kwasniew:
## About the changes

Screenshot 2023-08-03 at 11 01 33

In change request changes we had required feature name. Now we introduce segment changes that don't have any single feature attached. So I'm making the feature column optional. Segment id will be part of the payload same as strategy id is. In the long term we may even consider adding feature name to the payload.

### DIFF
--- a/src/migrations/20230803061359-change-request-optional-feature.js
+++ b/src/migrations/20230803061359-change-request-optional-feature.js
@@ -1,0 +1,20 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+            ALTER TABLE change_request_events ALTER COLUMN feature DROP NOT NULL;
+        `,
+        callback,
+    );
+};
+
+exports.down = function (db, callback) {
+    db.runSql(
+        `
+            UPDATE change_request_events SET feature = 'unknown' WHERE feature IS NULL;
+            ALTER TABLE change_request_events ALTER COLUMN feature SET NOT NULL;
+        `,
+        callback,
+    );
+};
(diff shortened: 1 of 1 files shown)


thomasheartman (approved):
This looks good to me 😄 Did you already handle the necessary API changes for it?

src/migrations/20230803061359-change-request-optional-feature.js
```diff
@@ -0,0 +1,20 @@
...
+            ALTER TABLE change_request_events ALTER COLUMN feature DROP NOT NULL;
+        `,
+        callback,
+    );
+};
+
+exports.down = function (db, callback) {
+    db.runSql(
+        `
+            UPDATE change_request_events SET feature = 'unknown' WHERE feature IS NULL;
```
sjaanus, line 15: This will not work, because it is invalid foreign key.

kwasniew: good point, what options do we have? delete those?

kwasniew:
@thomasheartman I have a PR coming soon that uses this migration and allows to perform full segment update workflow through change request. It will have a corresponding test showing how to use it.

sjaanus (approved):

src/migrations/20230803061359-change-request-optional-feature.js
sjaanus: Yeah, not ideal, but acceptable.

sjaanus: FYI, there is no ideal solution. We kind of want to retain all user data, but it will be super hard anyways, if they do down migrations.

PR #4395, opened 2023-08-03
1-1192: track the feature type and the new lifetime

thomasheartman:
This PR adds tracking of the feature type and the new lifetime to the
existing plausible event. This allows us extra insights into what users change it to.


frontend/src/component/featureTypes/FeatureTypeEdit/FeatureTypeForm/FeatureTypeForm.tsx
```diff
@@ -79,14 +79,19 @@ export const FeatureTypeForm: VFC<FeatureTypeFormProps> = ({
                 throw new Error('No feature toggle type loaded');
 
             const value = doesntExpire ? 0 : lifetime;
-            await updateFeatureTypeLifetime(featureType?.id, value);
+            await updateFeatureTypeLifetime(featureType.id, value);
```
thomasheartman, line 82: We've already verified that it's non-empty here, so it should be safe to remove the `?`

sjaanus (approved):

Tymek (approved):
LGTM

PR #4396, opened 2023-08-03
feat: allow trust proxy

sjaanus:
We have configured the trust proxy setting to true. With this configuration, the application places trust in the "X-Forwarded-For" header and considers the left-most entry in the header as the client's IP address. By doing so, the application relies on the proxy or load balancer to accurately set the "X-Forwarded-For" header with the correct client IP address.

However, it is important to be aware that this configuration also opens up the possibility of bypassing IP-based rate limiting. The rate-limiter system is currently logging this situation as a warning to alert us.

With this PR we are suppressing the warning.

### DIFF
--- a/src/lib/routes/admin-api/user-admin.ts
+++ b/src/lib/routes/admin-api/user-admin.ts
@@ -289,6 +289,7 @@ export default class UserAdminController extends Controller {
                 rateLimit({
                     windowMs: minutesToMilliseconds(1),
                     max: 20,
+                    validate: false,
                     standardHeaders: true,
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):
Looks good to me! But just for clarity: the only difference this makes is that we'll no longer log the warning. The behavior is still the same as before?

sjaanus:
> Looks good to me! But just for clarity: the only difference this makes is that we'll no longer log the warning. The behavior is still the same as before?

Yes, behaviour is still same.

PR #4397, opened 2023-08-03
fix: environment id missing bug

sjaanus:
When editing change request, there are no query params set, so need to pass in as props.

image


kwasniew (approved):
Nice catch!

PR #4405, opened 2023-08-03
chore: reduce build time

gastonfournier:
## About the changes
~~Test if max-workers can speed up test runs~~ No, this didn't help.
Disabling the job that takes longer should help mitigate. We lose the coverage report but we still test our code: https://github.com/Unleash/unleash/actions/runs/5751945832/job/15591813034?pr=4405#step:7:2396

This cuts the build time by 50% and moving this to a daily schedule

### DIFF
--- a/.github/workflows/build_prs_jest_report.yaml
+++ b/.github/workflows/build_prs_jest_report.yaml
@@ -1,10 +1,11 @@
 name: PR -> Test & Coverage
 
 on:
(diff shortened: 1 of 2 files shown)


sjaanus (approved):

gardleopard (approved):
Great, maybe adjust the timeout to 20 minutes as well or something if possible

PR #4408, opened 2023-08-03
feat(#4205): add segment to drafts

thomasheartman:
This PR sends segments to CR drafts when you use the `add to draft` button.

It also adds the logic for which environment to pick.

Segments sent to the API are added to drafts, but not displayed in the UI yet. CRs with only segment changes are also not listed in the table. This'll be covered in upcoming work.

### DIFF
--- a/frontend/src/component/segments/EditSegment/EditSegment.tsx
+++ b/frontend/src/component/segments/EditSegment/EditSegment.tsx
@@ -20,6 +20,7 @@ import { SEGMENT_SAVE_BTN_ID } from 'utils/testIds';
 import { useSegmentLimits } from 'hooks/api/getters/useSegmentLimits/useSegmentLimits';
 import { useOptionalPathParam } from 'hooks/useOptionalPathParam';
 import { useChangeRequestsEnabled } from 'hooks/useChangeRequestsEnabled';
+import { useChangeRequestApi } from 'hooks/api/actions/useChangeRequestApi/useChangeRequestApi';
 
 interface IEditSegmentProps {
     modal?: boolean;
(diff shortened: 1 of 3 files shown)


frontend/src/component/segments/EditSegment/EditSegment.tsx
```diff
@@ -78,28 +72,42 @@ export const EditSegment = ({ modal }: IEditSegmentProps) => {
...
     };
 
+    const { highestPermissionChangeRequestEnv } = useChangeRequestsEnabled(
+        segment?.project || ''
+    );
+    const changeRequestEnv = highestPermissionChangeRequestEnv();
+    const activateSegmentChangeRequests =
+        uiConfig?.flags?.segmentChangeRequests && changeRequestEnv;
+    const { addChange } = useChangeRequestApi();
+
```
thomasheartman, line 82: Moved these variables closer to where they're being used.

frontend/src/component/segments/EditSegment/EditSegment.tsx
```diff
@@ -78,28 +72,42 @@ export const EditSegment = ({ modal }: IEditSegmentProps) => {
...
-                    } else {
-                        navigate('/segments/');
-                    }
-                    setToastData({
-                        title: 'Segment updated',
-                        type: 'success',
-                    });
                 }
+                refetchSegments();
+                if (projectId) {
```
thomasheartman, line 98: I think `projectId` often is `null` or `undefined` here, even if `segment.project` isn't. Why is that? Shouldn't they be the same? Or is it only set if you navigate there from the projects screen, maybe?

kwasniew: because it's read from the hook that has optional query param in the name. Query params are not guaranteed to be there.

PR #4409, opened 2023-08-03
Update PRO plan menu

Tymek:
## About the changes
Update menu for Pro customers - show enterprise options with plan upgrade suggestion page.

image


frontend/src/component/menu/Header/Header.tsx
```diff
@@ -132,12 +132,23 @@ const Header: VFC = () => {
...
-            (mode.includes('enterprise') && isEnterprise())
-        );
+        const { mode, showEnterpriseBadge } = route.menu;
+
+        if (!mode) return true;
+
+        if (isPro()) {
+            return (
+                mode.includes('pro') ||
+                (mode.includes('enterprise') && showEnterpriseBadge) ||
```
thomasheartman, line 142: what exactly is `mode`? what does it mean for mode to include enterprise && showEnterpriseBadge? Is showEnterpriseBadge only true on pro subscriptions? Is it equivalent to `isEnterprise`?

thomasheartman (approved):
Yeah, this looks good to me 😄 I don't quite understand the mode checking, so I've added a question there, but I trust you've got it right. Though it could maybe be worth validating with a test or two?

frontend/src/component/menu/Header/Header.tsx
Tymek: `showEnterpriseBadge` is a flag

Tymek: I'll simplify this. It decoupled showing the badge from showing the route. I don't think we need it

PR #4417, opened 2023-08-04
feat: protect segment operations for change requests

kwasniew:
## About the changes

Update segment and delete segment now have protected mode which means that with CR enabled you can't change them willy-nilly. You either need skip permission for any project environment or go through a change request.

There's a corresponding enterprise change where I have tests and fix breaking changes.

## Discussion points

I noticed that we have interface for the segment service. Since this is the only service with an interface it looks odd. I'm tempted to remove it since there's only one impl of the interface and TS is structurally typed. @gastonfournier since you added the interface WDYT?

### DIFF
--- a/src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
+++ b/src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
@@ -32,7 +32,7 @@ export class ChangeRequestAccessReadModel
                 : Promise.resolve(false),
             this.isChangeRequestsEnabled(project, environment),
         ]);
-        return !(changeRequestEnabled && !canSkipChangeRequest);
+        return canSkipChangeRequest || !changeRequestEnabled;
     }
 
     public async canBypassChangeRequestForProject(
@@ -49,7 +49,7 @@ export class ChangeRequestAccessReadModel
                 : Promise.resolve(false),
             this.isChangeRequestsEnabledForProject(project),
         ]);
-        return !(changeRequestEnabled && !canSkipChangeRequest);
+        return canSkipChangeRequest || !changeRequestEnabled;
     }
 
     public async isChangeRequestsEnabled(
--- a/src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
+++ b/src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
@@ -35,6 +35,23 @@ export class ChangeRequestAccessReadModel
         return !(changeRequestEnabled && !canSkipChangeRequest);
     }
 
+    public async canBypassChangeRequestForProject(
+        project: string,
+        user?: User,
+    ): Promise<boolean> {
+        const [canSkipChangeRequest, changeRequestEnabled] = await Promise.all([
+            user
+                ? this.accessService.hasPermission(
+                      user,
+                      SKIP_CHANGE_REQUEST,
+                      project,
+                  )
+                : Promise.resolve(false),
+            this.isChangeRequestsEnabledForProject(project),
+        ]);
+        return !(changeRequestEnabled && !canSkipChangeRequest);
+    }
+
     public async isChangeRequestsEnabled(
         project: string,
         environment: string,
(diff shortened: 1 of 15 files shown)


src/test/fixtures/fake-project-store.ts
```diff
@@ -177,7 +177,7 @@ export default class FakeProjectStore implements IProjectStore {
         projectId: string,
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
         environment: string,
-    ): Promise<CreateFeatureStrategySchema | undefined> {
+    ): Promise<CreateFeatureStrategySchema | null> {
```
kwasniew, line 180: TS error fix

src/lib/services/segment-service.ts
```diff
@@ -108,13 +130,26 @@ export class SegmentService implements ISegmentService {
 
         await this.eventStore.store({
             type: SEGMENT_UPDATED,
-            createdBy: user.email || user.username,
+            createdBy: user.email || user.username || 'unknown',
```
kwasniew, line 133: TS error fix

src/lib/services/segment-service.ts
```diff
@@ -82,17 +96,25 @@ export class SegmentService implements ISegmentService {
...
             data: segment,
         });
 
         return segment;
     }
 
-    async update(
+    async update(id: number, data: unknown, user: User): Promise<void> {
+        if (this.flagResolver.isEnabled('segmentChangeRequests')) {
```
kwasniew, line 107: we have to be careful not to enable it before the UI is ready

src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
```diff
@@ -35,6 +35,23 @@ export class ChangeRequestAccessReadModel
         return !(changeRequestEnabled && !canSkipChangeRequest);
     }
 
+    public async canBypassChangeRequestForProject(
```
kwasniew, line 38: no check for environment. If a user can skip change requests in any env we're good for now. We can tweak this logic later. For now need something cheap to experiment

src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
```diff
@@ -35,6 +35,23 @@ export class ChangeRequestAccessReadModel
...
+            user
+                ? this.accessService.hasPermission(
+                      user,
+                      SKIP_CHANGE_REQUEST,
+                      project,
+                  )
+                : Promise.resolve(false),
+            this.isChangeRequestsEnabledForProject(project),
+        ]);
+        return !(changeRequestEnabled && !canSkipChangeRequest);
```
thomasheartman, line 52: Could we express this without the double negation? I'm struggling to parse what we're saying here.

src/lib/services/segment-service.ts
thomasheartman: Fair. It's currently enabled in dev, but we can turn that off

thomasheartman (approved):
Nice! My only comment is the double negation. It'd be nice if we could express that a bit clearer, but I'll leave it up to you.

src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts
kwasniew: great catch, changing it to  `return canSkipChangeRequest || !changeRequestEnabled;`

gastonfournier:
> I noticed that we have interface for the segment service. Since this is the only service with an interface it looks odd. I'm tempted to remove it since there's only one impl of the interface and TS is structurally typed. @gastonfournier since you added the interface WDYT?

Looking at https://github.com/Unleash/unleash/pull/3315 I think the plan was to use the interface to establish the contract as enterprise and  OSS had different implementations. I don't remember the context but we were moving things that were on OSS to the enterprise repo and this was probably in the expansion phase of the migration and it just stayed... If it doesn't make more sense and you want to remove it, I vote we follow our conventions (and this is out of our code conventions)

PR #4421, opened 2023-08-04
chore: increase max app names to 1000

gastonfournier:
## About the changes
Looking at our metrics this number should be a good threshold, giving us more functionality than just limiting it to 100.


sighphyre (approved):
LGTM!

PR #4431, opened 2023-08-07
chore: remove customRootRoles flag in favor of killswitch

nunogois:
https://linear.app/unleash/issue/2-1303/adapt-existing-customrootroles-flag-to-a-customrootroleskillswitch

Removes the existing `customRootRoles` flag in favor of a `customRootRolesKillSwitch` which should follow the same logic but negated/inverted.

Once released, this will effectively make custom root roles GA, and we can use [customRootRolesKillSwitch](https://app.unleash-hosted.com/hosted/projects/eg/features/customRootRolesKillSwitch) to disable the feature if needed.


chriswk (approved):
Nice. Looking forward to see this GA

PR #4436, opened 2023-08-07
docs: update strategy variants sdks and add sidebar link

kwasniew:
## About the changes

Added link to strategy variants in sidebar.

Updated supported SDKs.


thomasheartman (commented):
Nice one! 😄 Before merging, could you move the sidebar entry up to before the feature toggle entries, so that the list stays alphabetically sorted?

thomasheartman (approved):

PR #4440, opened 2023-08-08
feat: pointer to strategy variants

kwasniew:
## About the changes

Feature environment variants have a warning nudging for the strategy variant usage (behind a flag):
Screenshot 2023-08-08 at 09 58 38

Added variant info to strategy variants, it's a shared component with the feature environment variants that explains how to use variants in the SDK:
Screenshot 2023-08-08 at 09 57 24


ivarconr:
Do we need both info boxes, particular for the new experience? I feel like it kind of obvious that a methods called "isEnabled()" cannot return a variant, when the interface says "boolean"?

PR #4441, opened 2023-08-08
fix: deletion validation didnt account for groups

daveleek:
## About the changes

Fixes an issue where project role deletion validation didn't validate against project roles being connected to groups

### DIFF
--- a/src/lib/services/access-service.ts
+++ b/src/lib/services/access-service.ts
@@ -10,7 +10,7 @@ import {
     IUserRole,
 } from '../types/stores/access-store';
 import { Logger } from '../logger';
-import { IAccountStore, IUnleashStores } from '../types/stores';
+import { IAccountStore, IGroupStore, IUnleashStores } from '../types/stores';
 import {
     IAvailablePermissions,
     ICustomRole,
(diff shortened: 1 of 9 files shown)


src/lib/services/access-service.ts
```diff
@@ -578,10 +594,11 @@ export class AccessService {
         await this.validateRoleIsNotBuiltIn(id);
 
         const roleUsers = await this.getUsersForRole(id);
+        const roleGroups = await this.getGroupsForRole(id);
 
-        if (roleUsers.length > 0) {
+        if (roleUsers.length > 0 || roleGroups.length > 0) {
             throw new RoleInUseError(
-                'Role is in use by more than one user. You cannot delete a role that is in use without first removing the role from the users.',
+                'Role is in use by one or more users or groups. You cannot delete a role that is in use without first removing the role from the users and groups.',
```
gardleopard, line 601: two ifs here and different error messages would probably make this easier to debug for the user

gardleopard: Or just a simple users(0) or groups(23)

gardleopard (approved):
Great, added a comment about the message that can potentially share more information, but approving anyway:)

src/lib/services/access-service.ts
daveleek: Thanks @gardleopard! good feedback! I like the simple solution as it gives you overview of both regardless of which was the trigger. This is also part of an improvement of the UI that gives you an overview of usage before you attempt deleting

PR #4442, opened 2023-08-08
fix: Added cursor pagination to slackapp conversations query

chriswk:
We've been struggling with not getting hold of all channels. Reading the documentation I found the support for next_cursor in the response_metadata, as long as the response has a next_cursor, there are responses left, so this PR adds a loop for querying until the next cursor or the channels list is undefined or `''`. This should solve the issue Wayfair had as well.

### DIFF
--- a/src/lib/addons/slack-app.ts
+++ b/src/lib/addons/slack-app.ts
@@ -87,8 +87,39 @@ export default class SlackAppAddon extends Addon {
                 const slackConversationsList =
                     await this.slackClient.conversations.list({
                         types: 'public_channel,private_channel',
+                        exclude_archived: true,
+                        limit: 200,
                     });
                 this.slackChannels = slackConversationsList.channels || [];
+                let nextCursor =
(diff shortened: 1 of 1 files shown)


src/lib/addons/slack-app.ts
```diff
@@ -87,8 +87,39 @@ export default class SlackAppAddon extends Addon {
...
+                    if (channels === undefined) {
+                        this.logger.debug(
+                            'Channels list was empty, breaking pagination',
+                        );
+                        nextCursor = undefined;
+                        break;
+                    }
+                    nextCursor = moreChannels.response_metadata?.next_cursor;
+                    this.logger.debug(
+                        `This page had ${channels.length} channels`,
```
gardleopard, line 115: maybe introducting a metric of number of slack channels found could be helpfull?

nunogois (approved):
LGTM

src/lib/addons/slack-app.ts
chriswk: Good point, separate PR.

PR #4444, opened 2023-08-08
feat: added gauge for slack channel count

chriswk:
As pointed out in #4442, we should have a gauge just to keep track of how many channels we saw the last time we requested them. Here I've also added a label based on the first 8 chars of the token. I'm assuming enough entropy in Slack's app key generator. So if you configure multiple addons with different tokens, we should be able to gauge them all. If you create multiple addon configs with the same token, they will be counted as the same. This should be fine, since the channel count really shouldn't vary if we use the same token.


chriswk:
Since we changed the way to do this, this PR is dead. Closing

PR #4004, opened 2023-06-19
feat: change CR strategy title and name behaviour

sjaanus:
image

PR #4449, opened 2023-08-08
fix: CR strategy name changes code

thomasheartman:
This change addresses two things that were done in
https://github.com/Unleash/unleash/pull/4004 and that I believe to be bugs.

1. It shows the previous strategy name also if there was no previous
title. So if there was no previous title, it'll show the strategy name
with a strikethrough and then the new title (see the discussion section).
2. It changes a `span` component to a [`del` component](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del). I believe the
span was erroneously changed from a `s` component (strikethrough
component) in the linked PR (based on a comment on the PR). This
caused the strikethrough to not be there anymore. However, the `del`
component is semantically more correct and reintroduces the
strikethrough, so it is a better change.
3. It uses [`ins` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ins) for names that have changed. 

Finally, it removes a redundant pair of curly braces.

How it looks now:

image

## Discussion

Regarding point 1: It might be that we don't want to show a strikethrough through the name of the strategy if there was no previous title. In that case, the changes related to the first point should be removed. If we do that, it looks like this:

image

It makes it harder (impossible, actually) to see when a custom title was added, but that might be what we want. 

But maybe the solution is to also use `ins` elements for new data. That way the difference is visible (and semantically correct): 
image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.test.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.test.tsx
@@ -13,7 +13,7 @@ test.each(['', undefined])(
 
         // expect no del elements
         expect(
-            screen.getByText(previousTitle || '', { selector: 'del' })
+            screen.queryByText(previousTitle || '', { selector: 'del' })
         ).toBeNull();
 
         // expect ins element with new strategy name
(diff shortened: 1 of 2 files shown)


kwasniew (approved):
I like the last option you suggested with ins the most. 
Since we have lots of "if this then that" dilemmas and this code is at the core of Unleash I'd add a unit test for the tooltip describing the 3 cases you put as screenshots. I know it takes more time to finish the work but in this case it seems to be justified.

PR #4451, opened 2023-08-08
docs: custom root roles

nunogois:
https://linear.app/unleash/issue/2-1136/custom-root-roles-documentation

 - [Adds documentation referencing custom root roles](https://unleash-docs-git-docs-custom-root-roles-unleash-team.vercel.app/reference/rbac);
 - [Adds a "How to create and assign custom root roles" how-to guide](https://unleash-docs-git-docs-custom-root-roles-unleash-team.vercel.app/how-to/how-to-create-and-assign-custom-root-roles);
 - Standardizes "global" roles to "root" roles;
 - Standardizes "standard" roles to "predefined" roles to better reflect their behavior and what is shown in our UI;
 - Updates predefined role descriptions and makes them consistent;
 - Updates the side panel description of the user form;
 - Includes some boy scouting with some tiny fixes of things identified along the way (e.g. the role form was persisting old data when closed and re-opened);
 
 Questions:

 - Is it worth expanding the "Assigning custom root roles" section in the "How to create and assign custom root roles" guide to include the steps for assigning a root role for each entity (user, service account, group)?
 - Should this PR include an update to the existing "How to create and assign custom project roles" guide? We've since updated the UI;


website/docs/reference/rbac.md
```diff
@@ -9,7 +9,7 @@ This document forms the specifications for [Role-Based Access Control](https://e
 
 Unleash has two levels in its hierarchy of resources:
 
-1. **Global resources** - Everything that lives across the entire Unleash instance. Examples of this include:
+1. **Root resources** - Everything that lives across the entire Unleash instance. Examples of this include:
```
gastonfournier, line 12: Do we want to switch from Global resources to root resources? I think switching from Global Roles to Root Roles makes sense, root resources feel a bit weird IMO. But, maybe it is just me

website/docs/topics/the-anatomy-of-unleash.mdx
```diff
@@ -8,17 +8,17 @@ This guide's purpose is to give you a conceptual overview of how Unleash works.
 The end of this guide presents a [short use case, explaining how you might configure Unleash](#use-case) to start working with feature toggles.
 
 
-## The global level
+## The root level
```
gastonfournier, line 11: Ok, back to my other comment, maybe because of this concept of "root level" makes sense to also change to "root resources"

frontend/src/component/admin/users/EditUser/EditUser.tsx
```diff
@@ -77,9 +77,8 @@ const EditUser = () => {
         <FormTemplate
             loading={loading}
             title="Edit user"
-            description="In order to get access to Unleash needs to have an Unleash root role as Admin, Editor or Viewer.
-            You can also add the user to projects as member or owner in the specific projects."
-            documentationLink="https://docs.getunleash.io/reference/rbac#standard-roles"
+            description="In order to manage user access to Unleash, the user needs to be assigned a root role, such as Viewer, Editor, or Admin."
```
thomasheartman, line 80: Same suggestion as above, but maybe we want to tailor the text more towards editing here? 🤔 

```suggestion
            description="In order for a user to get access to Unleash, they need to be assigned a root role, such as Viewer, Editor, or Admin."
```

website/docs/reference/rbac.md
```diff
@@ -19,22 +19,160 @@ Unleash has two levels in its hierarchy of resources:
...
+#### Tag type permissions
+
+- **Update tag types**
+
+  Lets the user update tag types.
+
+- **Delete tag types**
+
+  Lets the user delete tag types.
+
```
thomasheartman, line 175: Non-blocking, but we might want to consider using [description lists](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) or tables here. That would create stronger semantic links between the permission name and its definition. 

Unfortunately, markdown doesn't have a syntax for description lists, so we'd have to drop down to raw html for it. That makes editing it a little trickier, but I think it's a fair tradeoff.

What do you think?

website/docs/how-to/how-to-create-and-assign-custom-root-roles.md
```diff
@@ -0,0 +1,36 @@
...
+---
+
+:::info availability
+
+Custom root roles were introduced in **Unleash 5.4** and are only available in Unleash Enterprise.
+
+:::
+
+
+This guide takes you through [how to create](#creating-custom-root-roles "how to create custom root roles") and [assign](#assigning-custom-root-roles "how to assign custom root roles") [custom root roles](../reference/rbac.md#custom-root-roles). Custom root roles allow you to fine-tune access rights and permissions to root resources in your Unleash instance.
```
thomasheartman, line 12: Nice use of link titles 👏🏼

website/docs/how-to/how-to-create-and-assign-custom-root-roles.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Step 3: Fill in the root role form {#create-step-3}
+
+Give the root role a name, a description, and the set of permissions you'd like it to have. For a full overview of all the options, consult the [custom root roles reference documentation](../reference/rbac.md#custom-root-roles).
+    
+![The root role form filled with some example data, and the "add role" button highlighted at the bottom.](/img/create-crr-step-3.png)
+
+## Assigning custom root roles
+
+You may assign custom root roles just like you would assign any other [predefined root role](../reference/rbac.md#predefined-roles). Besides assigning root roles to users, you can also assign them to [service accounts](../reference/service-accounts.md) and [groups](../reference/rbac.md#user-groups).
```
thomasheartman, line 36: Minor simplification suggestions.

```suggestion
You can assign custom root roles just like you would assign any other [predefined root role](../reference/rbac.md#predefined-roles). Root roles can be assigned to users, [service accounts](../reference/service-accounts.md), and [groups](../reference/rbac.md#user-groups).
```

thomasheartman (approved):
Very nice work on this 🚒  I have a few minor change suggestions and comments, but not much. There's a lot of work in here, so thanks for going through and doing it all 🙏🏼

website/docs/reference/rbac.md
thomasheartman: Yeah, I had the exact same thought, but I think I did the same thing you did

website/docs/reference/rbac.md
thomasheartman: Yeah, that's very nice! 😄  At least it looks very nice. I'm a little worried about the semantics for people who can't see the table. How do screen readers treat empty columns? Probably not as 'repeat the last non-empty entry in this column'. So how about splitting it up into multiple tables, each under its own heading? (keeping the heading structure from before)

nunogois: Addressed in https://github.com/Unleash/unleash/pull/4451/commits/f9e601be7af170d3371d71259f6aff49162d1102

PR #4452, opened 2023-08-08
Update admin navigation

Tymek:
Cleaner tabs navigation for admin tabs

image

Closes https://linear.app/unleash/issue/1-1104/improved-menu-oss


Tymek:
Moved to https://github.com/Unleash/unleash/pull/4458
in order to test Linear automation

PR #4456, opened 2023-08-09
fix: proper aggregation of strategies

gastonfournier:
## About the changes
Open API is creating 2 resources under the same URL path, we want them aggregated:
```shell
$ curl -s https://us.app.unleash-hosted.com/ushosted/docs/openapi.json | jq '.paths."/api/admin/strategies/{name}" | keys'
[
  "delete",
  "get"
]
gaston@gaston-Summit-E16Flip-A12UCT: ~/poc/terraform-provider-unleash on main [!$]
$ curl -s https://us.app.unleash-hosted.com/ushosted/docs/openapi.json | jq '.paths."/api/admin/strategies/{strategyName}" | keys'
[
  "put"
]
```

Note one is under `{name}` while the other is under `{strategyName}`

### DIFF
--- a/src/lib/openapi/spec/playground-strategy-schema.ts
+++ b/src/lib/openapi/spec/playground-strategy-schema.ts
@@ -28,7 +28,7 @@ export const strategyEvaluationResults = {
                     description:
                         "Whether this strategy resolves to `false` or if it might resolve to `true`. Because Unleash can't evaluate the strategy, it can't say for certain whether it will be `true`, but if you have failing constraints or segments, it _can_ determine that your strategy would be `false`.",
                     anyOf: [
-                        { type: 'boolean', enum: [false] },
+                        { type: 'boolean', default: false },
                         {
                             type: 'string',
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):
The name/strategyName fix is very welcome! 💯 I wasn't aware that booleans couldn't use enums. Do you have the docs for that anywhere? I'd ideally like to keep that as is because it accurately describes that state of the response, but if it causes issues, then changing it isn't really a big deal 😄

gastonfournier:
> The name/strategyName fix is very welcome! 100 I wasn't aware that booleans couldn't use enums. Do you have the docs for that anywhere? I'd ideally like to keep that as is because it accurately describes that state of the response, but if it causes issues, then changing it isn't really a big deal smile

I confirmed that the boolean is fine for openapi spec 3.0 (we're on 3.0.3). Version 3.1 introduces `const` which would be more meaningful in this case, but I'm keeping it as is for now

PR #4457, opened 2023-08-09
Demo for strategy variants

kwasniew:
## About the changes

https://github.com/Unleash/unleash/assets/1394682/b87f41ef-6bad-4644-aada-e0faf5c8e0a9

* Variants in demo show new strategy variants instead of the feature variants
* added cleanup code for accumulating strategies to get better perf and correctness. We remove all unconstrained strategies that were breaking demo for other users and we leave only last 30 constrained strategies.


frontend/src/component/demo/demo-setup.ts
```diff
@@ -67,75 +68,48 @@ export const gradualRollout = async () => {
...
-                            payload: {
-                                type: 'string',
-                                value: 'blue',
-                            },
-                        },
-                    },
-                ]),
-            }
+    const tooGenericStrategies = results.filter(
+        (strategy: { constraints: Array<unknown>; segments: Array<unknown> }) =>
```
nunogois, line 87: Don't we have the proper types for this instead of `unknown`?

frontend/src/component/demo/demo-setup.ts
```diff
@@ -67,75 +68,48 @@ export const gradualRollout = async () => {
...
-                            overrides: [],
-                            stickiness: 'default',
-                            payload: {
-                                type: 'string',
-                                value: 'blue',
-                            },
-                        },
-                    },
-                ]),
-            }
```
nunogois, line 138: This means we no longer have one of red, green, blue on the demo app on a clean slate, right? Will it default to black? I honestly don't remember.

I'm OK with it, but it's something we should be aware of and maybe align with @EliseBrevald and @nicolaesocaciu

nunogois (approved):
Haven't tested it but at a glance this looks good 👍

I suggest going through the flow manually and also try to go back in some of the steps to see if it doesn't break.

Aligning with UX would also be a good idea just to be extra safe that this is what we intend.

frontend/src/component/demo/demo-setup.ts
kwasniew: we don't care about those types so I left it as unknown. We care about the array lenght

frontend/src/component/demo/demo-setup.ts
kwasniew: I didn't know the context of those so glad your brought it up. I can bring those back as fallback. It shouldn't change much for the strategy variants.

PR #4459, opened 2023-08-09
Add name with change info to segment changes (1-1230-update-name-change)

thomasheartman:
Extract and reuse the component that we use for strategy name changes.

## Discussion points:

This impl only shows the new name in the second field (editing segment). Should we show it in both? Why do we have both? Do we have UI designs for this?

image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/NameWithChangeInfo/NameWithChangeInfo.test.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/NameWithChangeInfo/NameWithChangeInfo.test.tsx
@@ -4,74 +4,68 @@ import React from 'react';
 import { NameWithChangeInfo } from './NameWithChangeInfo';
 
 test.each(['', undefined])(
-    'Should render only the new title if the previous title was %s',
-    async previousTitle => {
-        const newTitle = 'new title';
+    'Should render only the new name if the previous name was %s',
+    async previousName => {
+        const newName = 'new name';
         render(
-            <NameWithChangeInfo
-                newName={newTitle}
-                previousName={previousTitle}
-            />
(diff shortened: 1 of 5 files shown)


frontend/src/component/changeRequest/ChangeRequest/NameWithChangeInfo/NameWithChangeInfo.test.tsx
thomasheartman, line 1: This is just the same test file moved and updated with new arg names.

frontend/src/component/changeRequest/ChangeRequest/NameWithChangeInfo/NameWithChangeInfo.tsx
thomasheartman, line 1: The extracted component. I see we use `Truncated` a number of places around change requests. Maybe it's worth defining once and used as a dependency?

frontend/src/component/changeRequest/ChangeRequest/SegmentTooltipLink.tsx
```diff
@@ -70,9 +71,10 @@ export const SegmentTooltipLink: FC<IStrategyTooltipLinkProps> = ({
                     maxHeight: 600,
                 }}
             >
-                <Typography component="span">
-                    {formatStrategyName(change.payload.name)}
```
thomasheartman, line 74: Not sure why we were calling `formatStrategyName` here, but I suspect it was a copy-paste error that was left in.

frontend/src/component/changeRequest/changeRequest.types.ts
thomasheartman, line 1: Update types to indicate that name is included.

kwasniew (approved):

PR #4461, opened 2023-08-09
fix: Remove lastSeenAt from useCollaborateData.tsx staleness check

andreas-unleash:
Remove the lastSeenAt property when checking for stale data


frontend/src/hooks/useCollaborateData.tsx
```diff
@@ -39,6 +39,21 @@ interface IStaleNotificationOptions {
     afterSubmitAction: () => void;
 }
 
+const omitUnwantedData = (obj: any, field: string): any => {
+    if (Array.isArray(obj)) {
+        return obj.map(value => omitUnwantedData(value, field));
+    } else if (typeof obj === 'object' && obj !== null) {
+        const result: any = {};
```
prabodhmeshram, line 46: Any particular reason we're declaring the result object as const, as even though const Object and Arrays are mutable, but ideally should be declared as let/var when we're going to update/modify them as we do with other datatypes

andreas-unleash: When you declare a variable with const, it means that the variable itself cannot be reassigned to a new value. However, if the variable refers to an object or array, you can still modify the properties or elements within that object or array. The const keyword doesn't make the object itself immutable, just the variable reference.

Using const is often encouraged when a variable is not going to be reassigned because it communicates the intent more clearly and helps to prevent unintentional reassignments that could lead to bugs.

frontend/src/hooks/useCollaborateData.tsx
```diff
@@ -39,6 +39,21 @@ interface IStaleNotificationOptions {
     afterSubmitAction: () => void;
 }
 
+const omitUnwantedData = (obj: any, field: string): any => {
```
kwasniew, line 42: need to be properly typed since any looses all type information. 

chatgpt suggests sth like this :) 
```ts
type DeepOmit<T, K extends string> = {
  [P in keyof T]: P extends K ? never
                : T[P] extends object ? DeepOmit<T[P], K>
                : T[P]
};

function deepOmit<T, K extends string>(obj: T, key: K): DeepOmit<T, K> {
  if (typeof obj !== 'object' || obj === null) {
    return obj as any;
  }

  const output: any = Array.isArray(obj) ? [] : {};

  for (const k in obj) {
    if (k === key) continue;

    if (typeof obj[k] === 'object') {
      output[k] = deepOmit(obj[k], key);
    } else {
      output[k] = obj[k];
    }
  }

  return output;
}
```

Another option is to use a microlib doing that. The operation you're trying to do seems to be a deepOmit

kwasniew: Also if you write your own deep omit it needs proper unit tests since it's a nontrivial function

frontend/src/hooks/useCollaborateData.tsx
```diff
@@ -74,7 +89,12 @@ export const useCollaborateData = <Type,>(
     useEffect(() => {
         if (!cache || !data) return;
 
-        const equal = dequal(formatDequalData(cache), formatDequalData(data));
+        const cleanedCache = omitUnwantedData(cache, 'lastSeenAt');
```
kwasniew, line 92: I'd test this entire hook. If you want a template for testing hooks you can check e.g. useChangeRequestsEnabled.test.tsx. I usually create a test component for the hook and check the test component.

kwasniew (changes requested):
We need to add types and tests

kwasniew (approved):
nice

PR #4463, opened 2023-08-10
fix: multiline textarea resizing for json input

spgandhi:
### Background
https://github.com/orgs/Unleash/discussions/4450

### Problem
The textarea when adding JSON payload was of fixed size and did not allow resizing. Although MUI knows to render [TextareaAutosize](https://mui.com/material-ui/react-textarea-autosize/) when `multiline=true` as suggested in their docs - https://mui.com/material-ui/react-text-field/#multiline, but even in their examples in the docs, the resizing is not present. So perhaps some bug on their end  

### Solution
With this PR we are adding `resize: vertical` styling if the `Input` component is rendering the native `textarea`. 
This solution will make both `json` and `csv` payload type inputs resizable since they both render as `textarea`

https://github.com/Unleash/unleash/assets/5814269/a484c457-25fc-47bc-bf92-5de4770d16cd

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
@@ -86,9 +86,10 @@ const StyledFieldColumn = styled('div')(({ theme }) => ({
     },
 }));
 
-const StyledInput = styled(Input)(() => ({
+const StyledInput = styled(Input)(({theme}) => ({
     width: '100%',
     '& textarea': {
+        minHeight: theme.spacing(3),
         resize: 'vertical',
     },
 }));
--- a/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
@@ -88,6 +88,9 @@ const StyledFieldColumn = styled('div')(({ theme }) => ({
 
 const StyledInput = styled(Input)(() => ({
     width: '100%',
+    '& textarea': {
+        resize: 'vertical',
+    },
 }));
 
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
```diff
@@ -88,6 +88,9 @@ const StyledFieldColumn = styled('div')(({ theme }) => ({
 
 const StyledInput = styled(Input)(() => ({
     width: '100%',
+    '& textarea': {
+        resize: 'vertical',
```
nunogois, line 92: I think setting a minimum height makes sense here, otherwise we can resize the input to be smaller than its original height.

Using `theme.spacing` in the `StyledInput`:

```TS
const StyledInput = styled(Input)(({ theme }) => ({
```

```suggestion
        minHeight: theme.spacing(3),
        resize: 'vertical',
```

nunogois (changes requested):
Thank you for your contribution, I think this is a positive step forward!

My only suggestion is to set a minimum height in the textarea so that we cannot resize the field to be smaller than the other input fields. Once we address that, this should be good to go 🙂

spgandhi:
@nunogois thanks for the feedback. I have added the minHeight and ready for review again

spgandhi:
@nunogois just checking in on this PR ^^ :)

frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
```diff
@@ -86,8 +86,12 @@ const StyledFieldColumn = styled('div')(({ theme }) => ({
     },
 }));
 
-const StyledInput = styled(Input)(() => ({
+const StyledInput = styled(Input)(({theme}) => ({
```
nunogois, line 89: Tiny styling / formatting nitpick.

```suggestion
const StyledInput = styled(Input)(({ theme }) => ({
```

nunogois:
Thanks for pinging me @spgandhi - I was away for holidays and missed this.

Seems like there's only a small formatting / styling issue that should have been caught by Prettier, but other than that, this looks ready to merge 👍

PR #4466, opened 2023-08-10
feat: cr sidebar segments count

kwasniew:
## About the changes

Show segments count in the change request sidebar

Screenshot 2023-08-10 at 10 23 39


andreas-unleash (approved):
LG

PR #4471, opened 2023-08-10
chore: type our path parameters

gastonfournier:
Path types in our openapi are inferred as string (which is a sensible default). But we can be more specific and provide the right type for each parameter. This is one example of how we can do that

@thomasheartman WDYT?

### DIFF
--- a/src/lib/routes/admin-api/user-admin.ts
+++ b/src/lib/routes/admin-api/user-admin.ts
@@ -363,6 +363,15 @@ export default class UserAdminController extends Controller {
                     description:
                         'Only the explicitly specified fields get updated.',
                     requestBody: createRequestSchema('updateUserSchema'),
+                    parameters: [
+                        {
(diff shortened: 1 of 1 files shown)


thomasheartman (commented):
Nice! I hadn't thought of doing that (even if we do it for query parameters). I like the intention behind it, but I'm wondering how much work it'd be? As for the suggestion you've got here, I think we should also add a description at least. We could constrain the ID (it's probably at least 0, right?), but it has less value here, I suppose.

But yeah, I like it! Just wondering how much time and effort it would take 🤔

ivarconr (approved):

gastonfournier:
Ok, we've decided that this only makes sense when parameters are numbers, in all other cases it's fair to assume they're strings. So, we'll only add them when needed (like in this case) and assume that when not present they're strings.
I'll update this PR and merge it later

PR #4473, opened 2023-08-10
Fix/constraint with legal value that has been deleted

andreas-unleash:
When editing a constraint that uses a context field with legal values:

if the contraint has a value that has been deleted from the legal values of the context field:
- Show the value and mark it as disabled
- On any change -> 'cleans'/removed the deleted legal values from the constraint values

Closes: [1-1209](https://linear.app/unleash/issue/1-1209/if-i-modified-the-legal-values-of-a-used-context-field)

### DIFF
--- a/frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
+++ b/frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
@@ -34,15 +34,14 @@ const resolveLegalValues = (
     values: IConstraint['values'],
     legalValues: IUnleashContextDefinition['legalValues']
 ): { legalValues: ILegalValue[]; deletedLegalValues: ILegalValue[] } => {
-    const deletedLegalValues =
-        (values || [])
-            .filter(
-                value =>
-                    !(legalValues || []).some(
-                        ({ value: legalValue }) => legalValue === value
-                    )
-            )
-            .map(v => ({ value: v, description: '' })) || [];
+    const deletedLegalValues = (values || [])
+        .filter(
+            value =>
+                !(legalValues || []).some(
+                    ({ value: legalValue }) => legalValue === value
+                )
+        )
+        .map(v => ({ value: v, description: '' }));
 
     return {
         legalValues: legalValues || [],
--- a/frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
+++ b/frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
@@ -31,24 +31,21 @@ interface IResolveInputProps {
 }
 
 const resolveLegalValues = (
-    constraint: IConstraint,
-    contextDefinition: IUnleashContextDefinition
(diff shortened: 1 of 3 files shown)


frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
```diff
@@ -30,6 +30,29 @@ interface IResolveInputProps {
     error: string;
 }
 
+const resolveLegalValues = (
+    constraint: IConstraint,
```
kwasniew, line 34: we don't need to expect the whole constraint because we only need its values. I'd use values type in the signature and pass only what we need w/o extra fluff

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
```diff
@@ -30,6 +30,29 @@ interface IResolveInputProps {
...
+
+    const deletedLegalValues =
+        constraintValues
+            .filter(
+                value =>
+                    !legalValues.some(
+                        ({ value: legalValue }) => legalValue === value
+                    )
+            )
+            .map(v => ({ value: v, description: '' })) || [];
```
kwasniew, line 48: we probably don't need the extra || [] at the end

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
kwasniew: all missing unit test

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
```diff
@@ -33,13 +36,14 @@ const createValuesMap = (values: string[]): IValuesMap => {
 };
 
 export const RestrictiveLegalValues = ({
```
kwasniew, line 38: when we're fixing a bug I'd add a test showing that we actually fixed a bug. This component seems to be quite easy to test.

kwasniew (changes requested):

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
andreas-unleash: added

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
kwasniew: filter always returns an array

kwasniew (approved):
2 more comments but overall nice change

PR #4474, opened 2023-08-10
Update quickstart.md

dgorton:
Introduce 'Feature Flag' language and replace most instances of 'Feature Toggle'

## About the changes
Introduce 'Feature Flag' language and replace most instances of 'toggle' with 'flag' on Quickstart

### DIFF
--- a/website/docs/tutorials/quickstart.md
+++ b/website/docs/tutorials/quickstart.md
@@ -4,9 +4,9 @@ title: Quickstart
 
 In this section we will attempt to guide you in order to get started with Unleash easily. There are multiple options to get started with Unleash, browse the headings to find the method that works best for you.
 
-:::info Feature Flags or Feature Toggles?
+:::info Feature flags or feature toggles?
 
-This document uses Feature Flags and Feature Toggles interchangeably. Some people prefer flag; others prefer toggle. We use both - they are synonyms for us.
+This document uses **feature flags** and **feature toggles** interchangeably. Some people prefer flag; others prefer toggle. We use both - they are synonyms for us.
 
 :::
 
--- a/website/docs/tutorials/quickstart.md
+++ b/website/docs/tutorials/quickstart.md
@@ -4,19 +4,25 @@ title: Quickstart
 
 In this section we will attempt to guide you in order to get started with Unleash easily. There are multiple options to get started with Unleash, browse the headings to find the method that works best for you.
 
-## I just want to get started creating toggles without much setup
+:::info Feature Flags or Feature Toggles?
+
+This document uses Feature Flags and Feature Toggles interchangeably. Some people prefer flag; others prefer toggle. We use both - they are synonyms for us.
+
+:::
+
+## I just want to try creating flags without much setup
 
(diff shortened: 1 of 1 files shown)


dgorton:
cc also @JarrodSC for visibility

website/docs/tutorials/quickstart.md
```diff
@@ -4,19 +4,25 @@ title: Quickstart
 
 In this section we will attempt to guide you in order to get started with Unleash easily. There are multiple options to get started with Unleash, browse the headings to find the method that works best for you.
 
-## I just want to get started creating toggles without much setup
+:::info Feature Flags or Feature Toggles?
+
+This document uses Feature Flags and Feature Toggles interchangeably. Some people prefer flag; others prefer toggle. We use both - they are synonyms for us.
```
thomasheartman, line 9: As a discussion topic: why have you capitalized "Feature Flag" and "Feature Toggle" here? They're normal nouns, so I don't see why they should be capitalized. I think we could call attention to them as terms either by **bolding them** or by "surrounding them with quotes".

However, I'm very open to hearing if you have a different perspective on this.

I'd suggest something like this: 

```suggestion
:::info Feature flags or feature toggles?

This document uses **feature flags** and **feature toggles** interchangeably. Some people prefer flag; others prefer toggle. We use both - they are synonyms for us.
```

website/docs/tutorials/quickstart.md
```diff
@@ -306,7 +312,7 @@ unleash.on('synchronized', () => {
   // Unleash is ready to serve updated feature toggles.
 
   // Check a feature flag
-  const isEnabled = unleash.isEnabled('some-toggle');
+  const isEnabled = unleash.isEnabled('some-flag');
 
   // Check the variant
   const variant = unleash.getVariant('app.ToggleY');
```
thomasheartman, line 318: We should probably update this one too, right? I don't know whether we want to use the same flag (as in my suggestion) or we'd rather stick to using a different flag name (such as `some-other-flag`). What do you think?

```suggestion
  const variant = unleash.getVariant('some-flag');
```

thomasheartman (approved):
Looks great! 😄 I've got two comments inline: one discussion point regarding formatting (non-blocking, but probably good to align on), and one instance where I think you missed a usage of "toggle". 

I trust your judgment on what to do with these points, so I'll approve the PR now and you can make any changes you want before merging it 🚀

thomasheartman:
Oh, and regarding the workflows: it says it's waiting for the requirements to be met. `build (18.x)` doesn't run for doc changes, so you can `merge without waiting for requirements` as long as the docs tasks pass. This was a change made recently to reduce build times, I think. I think the plan is to remove it from the docs workflow, but that hasn't been done yet, so for now we have this slightly weird workflow.

website/docs/tutorials/quickstart.md
dgorton: Toggle isn't a forbidden word. I left this as an example where the mental model of a toggle (binary, on/off) might be better than a flag. So I left it as-is.

PR #4480, opened 2023-08-11
fix: diff no changes

kwasniew:
## About the changes

* omit createdAt and createdBy in segments
* rename variables for segments (they were copied from strategies)
* fixed our diff component as it was never showing (no changes) and was showing the entire new code on no changes.
* added missing tests for event diff

before:
Screenshot 2023-08-11 at 11 17 51

after:
Screenshot 2023-08-11 at 11 59 52

### DIFF
--- a/frontend/src/component/events/EventDiff/EventDiff.tsx
+++ b/frontend/src/component/events/EventDiff/EventDiff.tsx
@@ -96,14 +96,14 @@ const EventDiff = ({
         };
     };
 
-    let changes;
+    let changes: any[] = [];
 
     if (diffs) {
         changes = diffs
             .map(buildDiff)
             .sort(sort)
             .map(({ value }) => value);
-    } else {
+    } else if (entry.data == null || entry.preData == null) {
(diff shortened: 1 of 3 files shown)


frontend/src/component/events/EventDiff/EventDiff.tsx
```diff
@@ -96,14 +96,14 @@ const EventDiff = ({
...
-    let changes;
+    let changes: any[] = [];
 
     if (diffs) {
         changes = diffs
             .map(buildDiff)
             .sort(sort)
             .map(({ value }) => value);
-    } else {
+    } else if (entry.data == null || entry.preData == null) {
```
kwasniew, line 106: you have to expand a few lines below to see how changes.length was never 0

kwasniew: the entire new or old code should be shown only when one of the data or preData is null o rundefined. If both are provided we should not go in here

sjaanus (approved):

PR #4490, opened 2023-08-15
fix: Change slackapp to using scheduleMessage

chriswk:
### What
This PR changes the slack-app addon to use slack-api's scheduleMessage instead of postMessage.

### Why
When using postMessage we had to find the channel id in order to be able to post the message to the channel. scheduleMessage allows using the channel name instead of the id, which saves the entire struggle of finding the channel name. It did mean that we had to move to defining blocks of content instead of the easier formatting we did with postMessage. 

### Message look
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     ]
   },
   "dependencies": {
-    "@slack/web-api": "^6.8.1",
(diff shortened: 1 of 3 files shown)


sighphyre (approved):
LGTM

PR #4507, opened 2023-08-16
feat: get api tokens by name

kwasniew:
## About the changes

* Get API tokens by name
* There can be 0, 1 or many tokens. I decided to always return 200 with either empty list, one element list, multi-element list. It feels the most idiomatic approach with a list returned. Originally the plan was to return 404 when the token doesn't exist but for the list/collection based APIs this feels wrong. 
* Current I'm adding in-memory filtering so that we don't have to over-invest in DB access code. Once the capability is found to be useful we can follow-up with more perf optimized solution.
* Also the existing code for tokens was mostly implemented in the controller layer instead of the service layer. I don't want to do massive refactoring and move everything to the service layer in this PR

### DIFF
--- a/src/lib/routes/admin-api/api-token.ts
+++ b/src/lib/routes/admin-api/api-token.ts
@@ -182,7 +182,7 @@ export class ApiTokenController extends Controller {
                     operationId: 'getApiTokensByName',
                     summary: 'Get API tokens by name',
                     description:
-                        'Retrieves all API tokens that match a given token name. Token names are not unique so 0, 1 or many tokens can be returned.',
+                        'Retrieves all API tokens that match a given token name. Because token names are not unique, this endpoint will always return a list. If no tokens with the provided name exist, the list will be empty. Otherwise, it will contain all the tokens with the given name. 
                     responses: {
                         200: createResponseSchema('apiTokensSchema'),
                         ...getStandardResponses(401, 403),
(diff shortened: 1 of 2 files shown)


src/lib/routes/admin-api/api-token.ts
```diff
@@ -168,6 +171,26 @@ export class ApiTokenController extends Controller {
...
+            path: '/:name',
+            handler: this.getApiTokensByName,
+            permission: [ADMIN, READ_CLIENT_API_TOKEN, READ_FRONTEND_API_TOKEN],
+            middleware: [
+                openApiService.validPath({
+                    tags: ['API tokens'],
+                    operationId: 'getApiTokensByName',
+                    summary: 'Get API tokens by name',
+                    description:
+                        'Retrieves all API tokens that match a given token name. Token names are not unique so 0, 1 or many tokens can be returned.',
```
thomasheartman, line 185: Small rewording suggestion. I think this is a little bit clearer, but I'll leave it up to you. 

```suggestion
                        'Retrieves all API tokens that match a given token name. Because token names are not unique, this endpoint will always return a list. If no tokens with the provided name exist, the list will be empty. Otherwise, it will contain all the tokens with the given name. 
```

If you want to keep your wording, you should add a comma between "unique" and "so": 
> Token names are not unique, so 0, 1 or many tokens can be returned.'

thomasheartman (approved):
Looks good to me 😄

PR #4508, opened 2023-08-16
chore: remove additionalProperterties annotation.

thomasheartman:
Unless you set it to `false`, additional properties are always
allowed. By explicitly setting it to `true` the generated
examples contain additional properties, e.g.:

```json
{
  "tokens": [
    "aproject:development.randomstring",
    "[]:production.randomstring"
  ],
  "additionalProp1": {}
}
```

By removing the explicit annotation, we still allow additional
properties, but we don't get them in examples:

```json
{
  "tokens": [
    "aproject:development.randomstring",
    "[]:production.randomstring"
  ]
}
```


kwasniew (approved):
Nice finding. Didn't know that

PR #4513, opened 2023-08-16
feat: review buttons makeover

kwasniew:
## About the changes

Screenshot 2023-08-16 at 13 40 40

* review button has changes count
* added icons
* approve/reject buttons poppers has a width matching the review button width


sjaanus (approved):
Atleast something for Nico :smile:

PR #4520, opened 2023-08-17
feat: slack-app can now post to both tagged and default channel

chriswk:
Currently the slack-app addon only posts to either the tagged channel for the feature or the default channels. This PR adds a new field that will allow you to configure the addon to post to both the default channels and the tagged channel (s)


daveleek (approved):
LGTM!

PR #4521, opened 2023-08-17
feat: create client_applications_usage table migration

andreas-unleash:
Creates client_applications_usage table

### DIFF
--- a/src/migrations/20230817095805-client-applications-usage-table.js
+++ b/src/migrations/20230817095805-client-applications-usage-table.js
@@ -5,8 +5,8 @@ exports.up = function (db, callback) {
         `
           CREATE TABLE IF NOT EXISTS client_applications_usage (
                app_name VARCHAR(255) REFERENCES client_applications(app_name) ON DELETE CASCADE,
-               project VARCHAR(255) REFERENCES projects(id) NOT NULL default 'default',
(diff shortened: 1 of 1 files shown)


src/migrations/20230817095805-client-applications-usage-table.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+          CREATE TABLE IF NOT EXISTS client_applications_usage (
+               app_name VARCHAR(255) REFERENCES client_applications(app_name) ON DELETE CASCADE,
+               project VARCHAR(255) REFERENCES projects(id) NOT NULL default 'default',
```
sjaanus, line 8: I think we can not put foreign key on this, because we also have *. Same for environment

sjaanus: Also, can there be instances where default project does not exist. I am not sure, but looks potentially dangerous.

andreas-unleash: fair enough - default project and environment are created by our migrations.  Good point on the `*`

sjaanus: Actually we could preserve the FK, but keep it nullable. And then null means all projects.

sjaanus (approved):

PR #4527, opened 2023-08-17
Integrations - frontend adjustments

Tymek:
## About the changes
  - [x] Create a feature flag
  - [x] Rename page title
  - [x] Rename menu item
  - [x] Update frontend URL (add redirect from old one)

https://linear.app/unleash/issue/1-1263/integrations-frontend-adjustments


frontend/src/component/integrations/CreateIntegration/CreateIntegration.tsx
```diff
@@ -14,7 +14,7 @@ export const DEFAULT_DATA = {
     environments: [],
 } as unknown as IAddon; // TODO: improve type
 
-export const CreateAddon = () => {
+export const CreateIntegration = () => {
     const providerId = useRequiredPathParam('providerId');
     const { providers, refetchAddons } = useAddons();
```
sjaanus, line 19: Should we update the hook also so it is consistently named, to integration?

Tymek: Everything on backend is still called _Addon_. I suggest leaving it like this, otherwise we need to refactor backend too. Reason for renaming frontend files is it will be easier to find it if files are named in the same way as menu item

sjaanus (approved):
LG. We just need to think how to make the flag work in future.

PR #4542, opened 2023-08-21
feat: more powerful project search

kwasniew:
## About the changes

Closes #

### DIFF
--- a/frontend/src/component/common/Search/SearchSuggestions/SearchSuggestions.tsx
+++ b/frontend/src/component/common/Search/SearchSuggestions/SearchSuggestions.tsx
@@ -72,9 +72,9 @@ export const SearchSuggestions: VFC<SearchSuggestionsProps> = ({
             return {
                 name: column.filterName,
                 header: column.Header ?? column.filterName,
-                options: [...new Set(filterOptions)].sort((a, b) =>
-                    a.localeCompare(b)
-                ),
+                options: [...new Set(filterOptions)]
+                    .filter((it: unknown) => it)
+                    .sort((a, b) => a.localeCompare(b)),
                 suggestedOption:
                     filterOptions[randomRow] ?? `example-${column.filterName}`,
                 values: getFilterValues(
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/Search/SearchSuggestions/SearchSuggestions.tsx
```diff
@@ -72,9 +72,9 @@ export const SearchSuggestions: VFC<SearchSuggestionsProps> = ({
             return {
                 name: column.filterName,
                 header: column.Header ?? column.filterName,
-                options: [...new Set(filterOptions)].sort((a, b) =>
-                    a.localeCompare(b)
-                ),
+                options: [...new Set(filterOptions)]
+                    .filter((it: unknown) => it)
```
kwasniew, line 76: filter out empty suggestions

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -265,6 +266,20 @@ export const ProjectFeatureToggles = ({
                           Cell: FeatureTagCell,
                           width: 80,
                           searchable: true,
+                          filterName: 'tags',
+                          filterBy(
+                              row: IFeatureToggleListItem,
+                              values: string[]
+                          ) {
+                              return includesFilter(
```
kwasniew, line 274: Example:
`tag1:val1
tag2:val2
tag3:val3`
We search if any of the lines includes e.g. `tag2:val2`

frontend/src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -265,6 +266,20 @@ export const ProjectFeatureToggles = ({
...
+                                  getColumnValues(this, row),
+                                  values
+                              );
+                          },
+                          filterParsing(value: string) {
+                              // only first tag from the list is added to search suggestions
+                              return value.split('\n')[0];
```
kwasniew, line 281: when returning filtering hints we return only first tag if we have more than one

frontend/src/hooks/useSearch.ts
```diff
@@ -12,32 +12,37 @@ type IUseSearchOutput<T extends any> = {
     getSearchContext: () => IGetSearchContextOutput<T>;
 };
 
+const normalizeSearchValue = (value: string) =>
```
kwasniew, line 15: we normalize search and filter data before any processing. This normalization allows to add spaces around commas

Tymek (approved):

frontend/src/component/common/Search/SearchSuggestions/SearchSuggestions.tsx
nunogois: Is this the same as `.filter(Boolean)`? Maybe that would be cleaner.

nunogois (approved):
Bit unsure about the behavior from a user-perspective but maybe it's just me. 
I guess the natural progression of this for spaced-values would be to wrap them in double quotes. Something like `tag:tag1, tag2, "My Value With Spaces", another` 🤔 
The code looks solid on a first glance 👍

PR #4549, opened 2023-08-23
Add condition for getting max revision id from store

pangorgo:
## About the changes
In our staging setup, we create ad-hoc environments and import Unleash state from production. After unleash is deployed on such environment, the import job kicks in and feeds Unleash instance with feature flags from production.

Between Unleash being up and running and the import job running, some applications start polling Unleash. They get an empty feature toggle list with `meta.revisionId=0`. Then apps use this as part of `eTag` header in subsequent requests. Even though after import Unleash server finally has toggles to serve, it doesn't because it calculates _max revision id_ based on toggle updates (not null `feature_name` column in query) or `SEGMENT_UPDATED`.

This change adds an extra condition to query so feature toggles import is considered something that should invalidate the cache.

### DIFF
--- a/src/lib/db/event-store.ts
+++ b/src/lib/db/event-store.ts
@@ -3,6 +3,7 @@ import {
     IBaseEvent,
     SEGMENT_UPDATED,
     FEATURE_IMPORT,
+    FEATURES_IMPORTED,
     IEventType,
 } from '../types/events';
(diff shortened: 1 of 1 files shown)


pangorgo:
@sjaanus @chriswk @ivarconr is there anything I can do to get this change reviewed and merged? Are any further explanations required?

PR #4552, opened 2023-08-23
fix: diff component for ordering

kwasniew:
## About the changes

Let the feature data load asynchronously before we use it


FredrikOseberg (approved):
LGTM

PR #4555, opened 2023-08-23
fix: do not allow creation/update of feature toggle with invalid stra…

sjaanus:


src/test/e2e/helpers/database.json
```diff
@@ -14,6 +14,30 @@
           "type": "string"
         }
       ]
+    },
+    {
+      "name": "flexibleRollout",
```
sjaanus, line 19: Also add flexibleRollout for our e2e tests

src/lib/services/feature-toggle-service.ts
```diff
@@ -225,24 +231,24 @@ class FeatureToggleService {
 
     validateUpdatedProperties(
         { featureName, projectId }: IFeatureContext,
-        strategy: IFeatureStrategy,
+        existingStrategy: IFeatureStrategy,
```
sjaanus, line 234: Refactoring to make it more clear

src/lib/services/feature-toggle-service.ts
```diff
@@ -640,13 +655,15 @@ class FeatureToggleService {
...
-        updates: Partial<IFeatureStrategy>,
+        updates: Partial<IStrategyConfig>,
         context: IFeatureStrategyContext,
         userName: string,
     ): Promise<Saved<IStrategyConfig>> {
         const { projectId, environment, featureName } = context;
         const existingStrategy = await this.featureStrategiesStore.get(id);
+
         this.validateUpdatedProperties(context, existingStrategy);
+        await this.validateStrategyType(updates.name);
```
sjaanus, line 666: Validate for updating

src/lib/services/feature-toggle-service.ts
```diff
@@ -271,6 +277,14 @@ class FeatureToggleService {
         }
     }
 
+    async validateStrategyType(
+        strategyName: string | undefined,
+    ): Promise<void> {
+        if (strategyName !== undefined) {
+            await this.strategyStore.get(strategyName);
```
kwasniew, line 284: is it depending on store throwing an error? 
I'd rather call exists method on store and explicitly throw and error BadDataError. Those DB errors get propagated to the UI which is not nice. It's better to wrap it into some HTTP friendly error

kwasniew (approved):
One comment about verification of strategy

src/lib/services/feature-toggle-service.ts
sjaanus: Updated

PR #4564, opened 2023-08-24
Fix/legal values deletion

FredrikOseberg:
This PR fixes a bug reported from a customer where deleting a legal value that was used in a strategy constraint would make it impossible to edit the constraint. 

[The bug was introduced here](https://github.com/Unleash/unleash/pull/4473)

The core of the problem introduced was that the values used to calculate illegal values was based on changing state. On the first render it would display correct state as it would match the legal values coming from the context definition with the legal values currently used in the constraint as values. However, when you triggered the onClick method for the checkboxes the state would be changed because we would remove the illegal values from the valueset and only insert current legal values in the state. This would trigger a re-render of the component, and now the data used to identify the illegal values would no longer be correct, because the bad values had been cleaned from the state. This would cause the UI for constraints to display incorrectly.

Changed the flow to now give you a warning if you have illegal values, and that if you make changes and save the strategy these values will be removed from the constraint: 

Skjermbilde 2023-08-25 kl  08 56 02

Also amended this to apply to the single legal value constraints.

Skjermbilde 2023-08-25 kl  08 57 40

### DIFF
--- a/frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
+++ b/frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
-import { Checkbox } from '@mui/material';
+import { Alert, Checkbox } from '@mui/material';
 import { useThemeStyles } from 'themes/themeStyles';
 import { ConstraintValueSearch } from 'component/common/ConstraintAccordion/ConstraintValueSearch/ConstraintValueSearch';
 import { ConstraintFormHeader } from '../ConstraintFormHeader/ConstraintFormHeader';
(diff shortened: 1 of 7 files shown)


frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
```diff
@@ -35,35 +36,59 @@ const createValuesMap = (values: string[]): IValuesMap => {
...
     };
 
+    const illegalValues = getIllegalValues(
+        constraintValues,
+        deletedLegalValues
+    );
+
+    useEffect(() => {
+        setValuesMap(createValuesMap(values));
+    }, [values, setValuesMap]);
```
kwasniew, line 84: should createValuesMap also be here?

kwasniew (approved):

frontend/src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
FredrikOseberg: This was just moved around, was here from before. Now that you pointed it out, I will take a look.

Issue #4565, opened 2023-08-24
Type safe schema generator for the client


Tymek:
### Describe the feature request

Based on community feedback https://unleash-community.slack.com/archives/C03GZBSJV9A/p1689704431836109 

### Background

Can we generate available flags directly with all possible variants from Unleash instance, and use this to strictly type client code?

PR #4566, opened 2023-08-24
feat: useUiFlag shorthand hook

Tymek:
## About the changes
Instead of this:
```ts
const myFlag = Boolean(uiConfig?.flags?.myFlag)
```
we can have this:
```ts
const myFlag = useUiFlag("myFlag")
```
With the same type safety, less verbose and more purposeful code.

### Important files
- `frontend/src/hooks/useUiFlag.ts`

## Discussion points
Can we in the future share flags between frontend and backend? Right now adding a new flag has to be done in 4 different places (backend flag keys list, backend flags defaults config, backend experimental server options, frontend type).

Most ergonomic option is to pull config directly from Unleash. 
Issue, based on previous user feedback: https://github.com/Unleash/unleash/issues/4565 
Internal feature request document: [docs.google.com/document/d/1Sx0q...](https://docs.google.com/document/d/1Sx0qKZXUVUCjuY5F4MOh1ieOM1A2_jE58zEA7jaM_1g/edit?usp=sharing)


nunogois (approved):
I like this 👍 Maybe it would be clear enough to have it as `useFlag`.

FredrikOseberg:
> I like this 👍 Maybe it would be clear enough to have it as `useFlag`.

useFlag is already used in our React SDK. Probably best to not conflate the terminology. Also, it fits with the current domain model, I don't need to map useFlag to uiFlags in the config mentally, because they are both using the same name for the same concept.

PR #4571, opened 2023-08-25
chore: deps update unleash client

kwasniew:
## About the changes

updating unleash-client that correctly reports non-beta version to our metrics


Tymek (approved):

gardleopard (approved):
5/5

PR #4575, opened 2023-08-25
feat: update breakpoint from 1260 to 1280

sjaanus:


kwasniew (approved):
How do we know that everything looks good now? :)

PR #4579, opened 2023-08-28
Fixes hangs when sending incorrect parameters to deleteProjectApiToken endpoint

daveleek:
## About the changes
Returns either 400 or 404 when token isn't found or doesn't match single project must be provided projectId criteria

Closes #
Linear 2-1003

## Discussion points

Is projects.length > 1 a 400?


nunogois (approved):
LGTM

PR #4587, opened 2023-08-30
feat: total metrics count

nunogois:
Adds total metrics count report to feature toggle "seen" information with the total all-time count of toggle evaluations for that feature.

This sits behind a `totalMetricsCount` feature flag.

image


nunogois:
Doesn't seem to be much interest in this feature at this stage, so I'm closing this for now.

PR #4591, opened 2023-08-30
feat: feature naming patterns

sjaanus:


andreas-unleash (approved):

andreas-unleash:
Figuring out that we have a 2nd EditProject.tsx file took a while - do we need this (seems to me) duplication.  I think better to reuse

thomasheartman (approved):

PR #4592, opened 2023-08-30
docs: troubleshooting guides

gastonfournier:
## About the changes
Adds troubleshooting guides for some common problems under the how-to area (following the [diataxis framework](https://diataxis.fr/)).

This is part of Unleash 2023 offsite challenge


website/docs/how-to/how-to-troubleshoot-flag-exposure.md
```diff
@@ -0,0 +1,9 @@
+---
+title: My feature flag is enabled but all/some of our users are not exposed to it
+---
+
+To confirm how users will have flags resolved, follow these steps:
+1. The Unleash Playground was developed with this particular use case in mind. An access token can be used along with context values (passed in via the UI) to see how a flag will be resolved. Check our [Unleash Playground documentation](../reference/playground.mdx) for more info.
```
thomasheartman, line 6: Minor suggestion: move the playground link:

```suggestion
1. The [Unleash Playground](../reference/playground.mdx Playground reference documentation) was developed with this particular use case in mind. An access token can be used along with context values (passed in via the UI) to see how a flag will be resolved. 
```

website/docs/how-to/how-to-troubleshoot-flag-exposure.md
```diff
@@ -0,0 +1,9 @@
+---
+title: My feature flag is enabled but all/some of our users are not exposed to it
+---
+
+To confirm how users will have flags resolved, follow these steps:
+1. The Unleash Playground was developed with this particular use case in mind. An access token can be used along with context values (passed in via the UI) to see how a flag will be resolved. Check our [Unleash Playground documentation](../reference/playground.mdx) for more info. 
+2. When using a **gradual rollout** strategy, be mindful of the **"stickiness"** value. The stickiness value is a context value used when determining where a context will fall in a gradual rollout. It is a **required field** for the strategy; satisfying the constraints and strategy requirements alone (without the stickiness value) will result in the strategy not being satisfied causing the Unleash platform to defer to other strategies (if possible) or as a disabled state if no other strategies are satisfied.
+    - Sometimes, different implementations of a feature (for example, front-end and back-end) may be unable to provide the same information or define the information differently leading to different outcomes from the platform.
```
thomasheartman, line 8: This is a little unclear to me. Are we saying it's mandatory when defining the strategy? As far as I remember, you can't set the stickiness value to **no** context field. And if it's 100%, shouldn't it still evaluate to true regardless of the context? I think some SDKs might treat this differently, but I think that's the goal.

website/docs/how-to/how-to-troubleshoot-flag-exposure.md
```diff
@@ -0,0 +1,9 @@
...
+3. If using the Unleash Edge or Proxy, consider changing the API endpoint being used by the SDK instance. The SDK and platform may be configured correctly while the intermediary (Edge or Proxy) is somehow malfunctioning.
```
thomasheartman, line 9:
```suggestion
3. If using [Unleash Edge](../../generated/unleash-edge.md) or the [Proxy](../../generated/unleash-proxy.md), consider changing the API endpoint being used by the SDK instance. The SDK and platform may be configured correctly while the intermediary (Edge or Proxy) is somehow malfunctioning. 
```

website/docs/how-to/how-to-troubleshoot-flag-not-returned.md
```diff
@@ -0,0 +1,14 @@
+---
+title: My feature flag is not returned in the Frontend API/Edge/Proxy
```
thomasheartman, line 2: This is only when using flags in the client SDK / front-end SDKs, right? If using Edge with a back-end SDK, you'll still get disabled flags

website/docs/how-to/how-to-troubleshoot-flag-not-returned.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+By default, these endpoints will not return features that are not enabled. This is mainly to save on bandwidth but it can be confusing when you expect your customers to be exposed to a feature but they’re not. What features are accessible to you is mainly determined by the token you’re using and the type of access it has.
```
thomasheartman, line 5: I know we're inconsistent about this other places in the docs, but are they features or feature flags? Might not be worth changing other places, but it may provide additional clarity here? 🤔 

```suggestion
By default, these endpoints will not return feature flags that are not enabled. This is mainly to save on bandwidth but it can be confusing when you expect your customers to be exposed to a feature but they’re not. What features are accessible to you is mainly determined by the token you’re using and the type of access it has.
```

website/docs/how-to/how-to-troubleshoot-flag-exposure.md
gastonfournier: Yeah, a bit unclear to me as well, let me clarify some stuff and post a suggestion.

> you can't set the stickiness value to **no** context field

You can use the `random` strategy for no context field at all, but because the behavior is random, our `default` strategy will try to use some context fields, if available:
image

But you can still pick random. All other strategies have to define a context field, but random is a predefined/special one that comes with Unleash

I think it's better to keep this sentence short and link to the reference documentation to avoid these two pieces to drift apart from each other
```suggestion
2. When using a **gradual rollout** strategy, be mindful of the **[stickiness](../reference/stickiness)** value. When evaluating a flag, if the provided context does not include the field used in the stickiness configuration, the gradual rollout strategy will be evaluated to false.
```

website/docs/how-to/how-to-troubleshoot-flag-not-returned.md
gastonfournier: Yes, but then you will not have this problem :)

The reasoning is that if you are here, it is because the feature is not being returned, not because your feature is being returned as disabled. So, I'm not sure it is worth clarifying that this is only in client/frontend SDKs...

gastonfournier (commented):
Thanks @thomasheartman!! great insights as usual :hugs:

website/docs/how-to/how-to-troubleshoot-flag-exposure.md
thomasheartman: Wait, so does `random` use any stickiness fields?

I like your suggestion, maybe something like this (requires a little more work):
```suggestion
. When using a **gradual rollout** strategy, be mindful of the **[stickiness](../reference/stickiness)** value. When evaluating a flag, if the provided context does not include the field used in the stickiness configuration, the gradual rollout strategy will be evaluated to false. Note that the `random` stickiness is a special case that [... what?]. This also means that if you have defined a context field called `random`, you can't use that for stickiness [is that true?].
```

website/docs/how-to/how-to-troubleshoot-flag-not-returned.md
thomasheartman: Ah, I see your reasoning. Maybe add a note in the text that this only applies to front-end/client-side SDKs/endpoints?

website/docs/how-to/how-to-troubleshoot-flag-exposure.md
gastonfournier: > Wait, so does random use any stickiness fields?

No, random is the only one that does not rely on the context fields at all. So if you're using random as stickiness (or default that ultimately evaluates as random), you will not be facing this issue. Therefore, I think not mentioning these strategies is better to avoid confusion 

> This also means that if you have defined a context field called `random`, you can't use that for stickiness [is that true?]

It shouldn't be true, and if it is, then it's probably a bug. The strategy name, should be used to first try to load a strategy implementation by name, and if no strategy is found, then we use the name as the field we want to be sticky on. But this means that you can't create a context field `random` and be sticky on it (you can but it won't work because random strategy would take precedence)... In any case, I don't think this should be part of this issue as it sounds very edgy (eventually we'll receive a bug report for this :shrug:)

I'm sticking with my suggestion if you're ok:
```suggestion
2. When using a **gradual rollout** strategy, be mindful of the **[stickiness](../reference/stickiness)** value. When evaluating a flag, if the provided context does not include the field used in the stickiness configuration, the gradual rollout strategy will be evaluated to false.```

File: website/docs/how-to/how-to-troubleshoot-flag-not-returned.md (hunk 4 above)
Comment (gastonfournier) [reply website/docs/how-to/how-to-troubleshoot-flag-not-returned.md:2]: I think a note might be confusing, if you have this problem using a server-side endpoint (and I believe you shouldn't have it), we should update this guide. It would be an unknown unknown.

Comment (thomasheartman) [reply website/docs/how-to/how-to-troubleshoot-flag-not-returned.md:2]: Got it. I'll leave it up to you ☺️

Comment (gastonfournier) [reply website/docs/how-to/how-to-troubleshoot-flag-not-returned.md:2]: I believe this troubleshooting guide will be a bit different than regular documentation:
1. More brief and linking to the source of truth in our docs
2. Generic, on the assumption that people landing on this page have the problem, and if we don't have the solution it's probably something we didn't know.

I'm no expert, so I'll keep relying on you :hugs:

Comment (thomasheartman) [reply website/docs/how-to/how-to-troubleshoot-flag-not-returned.md:2]: Agreed! Thanks again for the contribution 🙏🏼

File: website/docs/how-to/how-to-troubleshoot-flag-exposure.md [hunk 6]
```diff
@@ -0,0 +1,8 @@
+---
+title: My feature flag is enabled but all/some of our users are not exposed to it
+---
+
+To confirm how users will have flags resolved, follow these steps:
+1. The [Unleash Playground](../reference/playground.mdx Playground reference documentation) was developed with this particular use case in mind. An access token can be used along with context values (passed in via the UI) to see how a flag will be resolved. 
+2. When using a **gradual rollout** strategy, be mindful of the **[stickiness](../reference/stickiness)** value. When evaluating a flag, if the provided context does not include the field used in the stickiness configuration, the gradual rollout strategy will be evaluated to false.```
+3. If using [Unleash Edge](../reference/unleash-edge) or the [Proxy](../reference/unleash-proxy), consider changing the API endpoint being used by the SDK instance. The SDK and platform may be configured correctly while the intermediary (Edge or Proxy) is somehow malfunctioning.
```
gastonfournier, line 8: @mvittiglio I was reviewing this and I'm not sure I understand:
1. Changing the API Endpoint to what value?
2. Saying that the intermediate is somehow malfunctioning doesn't help troubleshoot, and most likely it can be a misconfiguration more than a malfunction. 

So, maybe you're trying to say to use the [Frontend API](https://docs.getunleash.io/reference/front-end-api) from the SDK, to be able to tell if the problem is on the SDK or on the configuration of Edge/proxy, can that be?

For now, I'll remove this sentence, and we can add it back in another PR once clarified

PR #4580, opened 2023-08-28
chore: Remove duplicated tags

gastonfournier:
## About the changes
At https://github.com/Unleash/unleash/pull/4432 we've introduced the same tags twice which causes issues when generating the api clients.

Closes #2-1350-fix-openapi-duplicated-tags

PR #4595, opened 2023-09-01
openapi: sort tags file

thomasheartman:
This change sorts the tags in the tags file and tests that the list is sorted alphabetically. This makes it easier to
find tags in the file. 

#4580 already introduced a test to check that we have no duplicate
tags, so this isn't as necessary anymore, but it's still nice to have.

It also removes the previous auto-sorting before exporting. This is to ensure that entries are sorted in the source list. This might seem like a regression, but it makes it easier to spot near-duplicate tags:

> Despite having the test that validates there are no duplicates, you can always have Notifications and Notification API by mistake (tags that mean the same but are different). Keeping the list alphabetically sorted might help to prevent this before pushing the change to prod. In this case, we will eventually find out and fix it, so this could be a good reason to have the list sorted.

### DIFF
--- a/src/lib/openapi/util/openapi-tags.test.ts
+++ b/src/lib/openapi/util/openapi-tags.test.ts
@@ -7,7 +7,7 @@ test('no duplicate tags', () => {
     }, []);
 });
 
-test('tags are sorted', () => {
+test('The list of OpenAPI tags is sorted', () => {
     const tags = openApiTags.map((tag) => tag.name);
     const sorted = [...tags].sort((a, b) => a.localeCompare(b));
     console.log(tags, sorted);
--- a/src/lib/openapi/util/openapi-tags.test.ts
+++ b/src/lib/openapi/util/openapi-tags.test.ts
@@ -10,7 +10,6 @@ test('no duplicate tags', () => {
 test('The list of OpenAPI tags is sorted', () => {
     const tags = openApiTags.map((tag) => tag.name);
     const sorted = [...tags].sort((a, b) => a.localeCompare(b));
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
Looks great! We could also add a test to validate this is sorted, so we don't break it in the future

thomasheartman:
Ooh, I like that. Give me a bit to whip it up 😄

thomasheartman:
On the other hand, we already sort it before exporting it, so it's never going to be an issue in the output. I'm not sure how I feel about exporting the raw constant and checking that. It seems like an arbitrary thing to cause test failure over. What do you think?

gastonfournier:
> On the other hand, we already sort it before exporting it, so it's never going to be an issue in the output. I'm not sure how I feel about exporting the raw constant and checking that. It seems like an arbitrary thing to cause test failure over. What do you think?

:thinking: The question I'd ask myself is **What's the value of having it sorted in this file if we're already sorting it later?**, and 2 things come to my mind:
1.  Performance: We could remove the sort afterward to improve performance (provided the test guarantees the list is already sorted). But this feels like a very soft reason.
2. Correction: Despite having the test that validates there are no duplicates, you can always have `Notifications` and `Notification API` by mistake (tags that mean the same but are different). Keeping the list alphabetically sorted might help to prevent this before pushing the change to prod. In this case, we will eventually find out and fix it, so this could be a good reason to have the list sorted.

thomasheartman:
Yeah, okay, I like your reasoning here. You've convinced me. I'll fix.

PR #4596, opened 2023-09-01
docs: fix an issue where it goes into an infinite loop when building

thomasheartman:
This change sets `noRuntimeDownloads` to `true` for
`plugin-remote-content'` to avoid the infinite loop issue mentioned in
https://github.com/rdilweb/docusaurus-plugin-remote-content/issues/51,
as described in
https://github.com/rdilweb/docusaurus-plugin-remote-content#noruntimedownloads.

Note that this will require you to generate the files before
running the doc build, which may be more an issue, so it's a tradeoff.

To generate the files before building run:

```bash
yarn docusaurus download-remote-content-external
yarn docusaurus download-remote-content-sdks
```

We can of course write a simpler command that does this for us (or even make it part of the build step). In fact, I think we'll need to do exactly that. It seems to be failing now because the files are not generated by default by the build steps.

Creating this PR mostly to get feedback from @dgorton. What do you think?

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -7,9 +7,10 @@
   },
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
-    "build": "yarn generate && docusaurus build",
+    "start": "yarn fetch-remote-content && docusaurus start",
+    "build": "yarn generate && yarn fetch-remote-content && docusaurus build",
     "swizzle": "docusaurus swizzle",
+    "fetch-remote-content": "docusaurus download-remote-content-external && docusaurus download-remote-content-sdks",
     "generate": "docusaurus gen-api-docs all && node clean-generated-docs.js",
     "deploy": "yarn generate && docusaurus deploy",
     "clear": "docusaurus clear",
--- a/website/package.json
+++ b/website/package.json
@@ -12,7 +12,7 @@
     "swizzle": "docusaurus swizzle",
     "fetch-remote-content": "docusaurus download-remote-content-external && docusaurus download-remote-content-sdks",
     "generate": "docusaurus gen-api-docs all && node clean-generated-docs.js",
-    "deploy": "yarn generate && docusaurus deploy",
+    "deploy": "yarn generate && yarn fetch-remote-content && docusaurus deploy",
     "clear": "docusaurus clear",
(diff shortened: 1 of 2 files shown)


thomasheartman:
@dgorton Have you had a chance to have a look at this? Have any thoughts?

dgorton (approved):
Apologies for the delay on this - missed this notification.

dgorton:
Fixing infinite loops is obviously a good thing. It would be nice not to add more steps to the build process, however - especially as this will likely catch people by surprise. 

Are you comfortable adding these commands as part of the build step, per your comment? 

> To generate the files before building run:
> 
```
yarn docusaurus download-remote-content-external
yarn docusaurus download-remote-content-sdks
```
> We can of course write a simpler command that does this for us (or even make it part of the build step). In fact, I think we'll need to do exactly that. It seems to be failing now because the files are not generated by default by the build steps.

thomasheartman:
Yeah, good points. I think it should be pretty straightforward. Should look something like d0d9da888bdefb8d3706c5d4fc70a544919f36da. Let's see if that works.

website/package.json
```diff
@@ -7,9 +7,10 @@
...
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
-    "build": "yarn generate && docusaurus build",
+    "start": "yarn fetch-remote-content && docusaurus start",
+    "build": "yarn generate && yarn fetch-remote-content && docusaurus build",
     "swizzle": "docusaurus swizzle",
+    "fetch-remote-content": "docusaurus download-remote-content-external && docusaurus download-remote-content-sdks",
     "generate": "docusaurus gen-api-docs all && node clean-generated-docs.js",
     "deploy": "yarn generate && docusaurus deploy",
```
Tymek, line 15: we probably also need it in `deploy`, since it's the script used in `generate-docs.yaml` CI file

Tymek (changes requested):

website/package.json
thomasheartman: Ah, very good catch!

```suggestion
    "deploy": "yarn generate && yarn fetch-remote-content && docusaurus deploy",
```

Tymek (approved):

PR #4601, opened 2023-09-04
Change request flags removed

kwasniew:
## About the changes

Removing flags for change request reject and change request segments so that the features can go into GA.


thomasheartman (approved):
🚀

PR #4606, opened 2023-09-04
fix: reset selected toggle after archive or revive

Tymek:
## About the changes
When archiving or reviving feature toggles, when toggles disappear from table, actions bar should also disappear.

Closes https://linear.app/unleash/issue/1-1293/bulk-revive-modal-doesnt-go-away


kwasniew (approved):
Very nice. I'm wondering how difficult would it be to show with a RTL test that we had a bug and not that it's gone.

PR #4608, opened 2023-09-05
add descriptions in form

thomasheartman:
This PR adds descriptions of the various feature flag naming input fields in the project settings form. It also wraps the feature naming fields in a fieldset to group them semantically.

image

### DIFF
--- a/frontend/src/component/project/Project/ProjectForm/ProjectForm.tsx
+++ b/frontend/src/component/project/Project/ProjectForm/ProjectForm.tsx
@@ -311,7 +311,14 @@ const ProjectForm: React.FC<IProjectForm> = ({
                                     naming pattern.
                                 </p>
                             </StyledSubtitle>
-                            <StyledInputContainer>
+                            <StyledInputContainer
+                                sx={{
+                                    flexDirection: 'column',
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/Project/ProjectForm/ProjectForm.tsx
```diff
@@ -285,17 +290,40 @@ const ProjectForm: React.FC<IProjectForm> = ({
...
+                                    used to enforce feature flag names within
+                                    this project.
+                                </p>
+                                <p>
+                                    Leave it empty if you don’t want to add a
+                                    naming pattern.
+                                </p>
                             </StyledSubtitle>
-                            <StyledInputContainer>
+                            <StyledInputContainer
```
sjaanus, line 314: Ideally we should not use SX at all or only when one line of css. In this case, a styled component should be created.

sjaanus (approved):
Small comment, but nonblocking approval.

frontend/src/component/project/Project/ProjectForm/ProjectForm.tsx
thomasheartman: Got it; will fix!

PR #4617, opened 2023-09-06
1-1334: show short error message when validation fails

thomasheartman:
The error message only tells the user that the name doesn't match the
pattern. Because we already show the pattern above the input, we don't
need to repeat it in the error message. This makes for a shorter and
more concise message and better UX.

At the same time, for API users, we can keep the more detailed message
that includes info about the pattern, the example, and the description.

image

### DIFF
--- a/frontend/src/component/feature/CreateFeature/CreateFeature.tsx
+++ b/frontend/src/component/feature/CreateFeature/CreateFeature.tsx
@@ -137,6 +137,7 @@ const CreateFeature = () => {
                 handleCancel={handleCancel}
                 mode="Create"
                 clearErrors={clearErrors}
+                featureNaming={projectInfo.featureNaming}
             >
                 <CreateButton
                     name="feature toggle"
--- a/frontend/src/component/feature/FeatureForm/FeatureForm.tsx
+++ b/frontend/src/component/feature/FeatureForm/FeatureForm.tsx
@@ -82,6 +82,11 @@ const StyledTypeDescription = styled('p')(({ theme }) => ({
     position: 'relative',
 }));
 
+const StyledFlagNamingInfo = styled('div')(({ theme }) => ({
+    fontSize: theme.fontSizes.smallBody,
+    color: theme.palette.text.secondary,
+}));
+
 const StyledButtonContainer = styled('div')({
     marginTop: 'auto',
     display: 'flex',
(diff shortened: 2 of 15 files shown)


sjaanus (approved):

frontend/src/component/feature/CreateFeature/CreateFeature.tsx
```diff
@@ -137,6 +137,7 @@ const CreateFeature = () => {
                 handleCancel={handleCancel}
                 mode="Create"
                 clearErrors={clearErrors}
+                featureNaming={projectInfo.featureNaming}
```
FredrikOseberg: Will this always exists? According to the type in featureForm it can be undefined.

frontend/src/component/feature/FeatureForm/FeatureForm.tsx
```diff
@@ -128,16 +136,52 @@ const FeatureForm: React.FC<IFeatureToggleForm> = ({
...
+                            <p>
+                                This project has feature flag naming patterns
+                                enabled.
+                            </p>
+                            <dl id="feature-naming-pattern-info">
+                                <dt>Pattern</dt>
+                                <dd>
+                                    <code>{featureNaming?.pattern}</code>
+                                </dd>
+                                {Boolean(featureNaming?.example) && (
```
FredrikOseberg: Could use conditional render component for readability

frontend/src/component/feature/FeatureForm/FeatureForm.tsx
```diff
@@ -128,16 +136,52 @@ const FeatureForm: React.FC<IFeatureToggleForm> = ({
...
+                                    <>
+                                        <dt>Example</dt>
+                                        <dd>{featureNaming?.example}</dd>
+                                    </>
+                                )}
+                                {Boolean(featureNaming?.description) && (
```
FredrikOseberg: Could also be used here

frontend/src/component/feature/CreateFeature/CreateFeature.tsx
thomasheartman: Yes. So we send it down. If we have it, all good, otherwise featureForm handles that too. Am I missing something that makes this not work in some cases?

frontend/src/component/feature/FeatureForm/FeatureForm.tsx
thomasheartman: Yeah, I thought about that, but decided against it because of the layers of nesting. Felt like it would take up a lot of extra vertical space for not a lot of extra value. However, I do see your point and would be happy to switch to the component instead if you think that's better ☺️

PR #4618, opened 2023-09-06
Clickable search filter options

kwasniew:
## About the changes

https://github.com/Unleash/unleash/assets/1394682/c75392de-b2c7-419f-9a59-893395825230

Main changes:
* filter and search hints are clickable
* combined filter and search hint is taken from the first data table item. It guarantees that we get a result. 
* we left only one filter hint in a combined search and filter hint to keep it simple


frontend/src/component/common/Search/Search.tsx
```diff
@@ -148,7 +149,8 @@ export const Search = ({
                             <Tooltip title="Clear search query" arrow>
                                 <IconButton
                                     size="small"
-                                    onClick={() => {
+                                    onClick={e => {
+                                        e.stopPropagation(); // prevent outside click from the lazily added element
```
kwasniew, line 153: search container ref couldn't see the icon button that is added in a conditional render. As a result the IconButton is not considered to be part of the searchContainerRef. It means that it was considered as a click outside of the searchContainerRef which closes the search. We don't want that so we stop event propagation.

frontend/src/component/common/Search/Search.tsx
```diff
@@ -164,11 +166,13 @@ export const Search = ({
             <ConditionallyRender
                 condition={Boolean(hasFilters) && showSuggestions}
                 show={
-                    <div ref={suggestionsRef}>
-                        <SearchSuggestions
-                            getSearchContext={getSearchContext!}
-                        />
-                    </div>
+                    <SearchSuggestions
+                        onSuggestion={suggestion => {
```
kwasniew, line 170: when people click on a suggestion we change the input and auto focus the input

frontend/src/component/common/Search/SearchSuggestions/SearchSuggestions.tsx
```diff
@@ -57,14 +59,10 @@ const randomIndex = (arr: any[]) => Math.floor(Math.random() * arr.length);
 
 export const SearchSuggestions: VFC<SearchSuggestionsProps> = ({
     getSearchContext,
+    onSuggestion,
 }) => {
     const searchContext = getSearchContext();
 
-    const randomRow = useMemo(
```
kwasniew, line 63: this was generating combined filter and search hints that returned 0 results

FredrikOseberg (approved):
LGTM

PR #4619, opened 2023-09-06
fix: prevent 404 on auth settings hook

nunogois:
Since `auth/{auth_type}/settings` is an Enterprise route, this prevents 404s when we try to use the hook to fetch auth settings in non-Enterprise instances by using the conditional `useEnterpriseSWR` hook.


sighphyre (approved):
Amazing! LGTM

PR #4621, opened 2023-09-06
refactor: clean up no longer used PAT methods

nunogois:
This cleans up some PAT-related methods that are no longer used.


gastonfournier (approved):
:broom:

PR #4623, opened 2023-09-06
1-1320: allow you to update example with no pattern + update state better

thomasheartman:
While having a pattern when you have no example doesn't make a lot of sense, it's a problem that you can't delete the example after deleting the pattern: you previously had to remove the example before the pattern.

This PR fixes that by always allowing you to update the example, even if there is no pattern. Our server doesn't currently accept submitting an example with no pattern, but we could allow that if we want to (and probably just discard it on the back-end).

This PR also updates the validation of the example and the regex. There were more unhandled edge cases previously where the validation would disappear or be wrong. This should be fixed now. The new logic is that, whenever you update the either the pattern or the example, we check:
- if you have an error in your pattern, no pattern, or no example, then delete the example error if it exists
- have a well-formed pattern and an example then check if the example matches the pattern and add/delete an error accordingly

This does have some consequences: editing the pattern can render your example invalid. You'll also get immediate feedback instead of when you switch focus. I think this is often a bad pattern (giving the user too much negative feedback), but in terms of working with regexes, I think it might be a good thing. We also give immediate feedback today, so I don't think this is a regression.

Any thoughts are welcome.

### DIFF
--- a/frontend/src/component/project/Project/ProjectForm/FeatureFlagNamingTooltip.tsx
+++ b/frontend/src/component/project/Project/ProjectForm/FeatureFlagNamingTooltip.tsx
@@ -14,7 +14,7 @@ export const FeatureFlagNamingTooltip: FC = () => {
                 <Box>
                     <h3>Enforce a naming convention for feature flags</h3>
                     <hr />
-                    <p>{`eg. ^[A - Za - z0 - 9]{2}[.][a-z]{4,12}$ matches 'a1.project'`}</p>
+                    <p>{`eg. ^[A-Za-z0-9]{2}[.][a-z]{4,12}$ matches 'a1.project'`}</p>
                     <div className="scrollable">
                         <h3>Brackets:</h3>
                         <table>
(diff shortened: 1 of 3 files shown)


frontend/src/component/project/Project/ProjectForm/FeatureFlagNamingTooltip.tsx
thomasheartman, line 1: This pattern was invalid, so this is just a cleanup.

frontend/src/component/project/Project/ProjectForm/ProjectForm.tsx
```diff
@@ -190,7 +236,9 @@ const ProjectForm: React.FC<IProjectForm> = ({
                     onChange={e => setProjectName(e.target.value)}
                     error={Boolean(errors.name)}
                     errorText={errors.name}
-                    onFocus={() => clearErrors()}
+                    onFocus={() => {
+                        delete errors.name;
+                    }}
```
thomasheartman, line 241: previously, focusing the project name would clear all errors, including the ones for the naming pattern. this is a bit overzealous, so we'll only clear the errors.name property here.

sjaanus (approved):

PR #4624, opened 2023-09-06
feat: persistent search queries

kwasniew:
## About the changes

We remember last search/filter. 

Screenshot 2023-09-06 at 13 54 33

Details:
* if your search has unique id you store your saved query in local storage
* if your search doesn't have unique id, then it stores the saved query in memory state 
* saved queries are clickable and can populate the input
* only non-empty queries are persisted

### DIFF
--- a/frontend/src/component/common/Search/SearchSuggestions/SearchInstructions/SearchInstructions.tsx
+++ b/frontend/src/component/common/Search/SearchSuggestions/SearchInstructions/SearchInstructions.tsx
@@ -7,12 +7,16 @@ const StyledHeader = styled('span')(({ theme }) => ({
     color: theme.palette.text.primary,
 }));
 
-const StyledCode = styled('span')(({ theme }) => ({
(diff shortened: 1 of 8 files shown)


frontend/src/component/common/Search/SearchSuggestions/SearchInstructions/SearchInstructions.tsx
```diff
@@ -49,11 +53,18 @@ export const SearchInstructions: VFC<ISearchInstructionsProps> = ({
             {filters.map(filter => (
                 <StyledFilterHint key={filter.name}>
                     {filter.header}:{' '}
-                    <StyledCode
-                        onClick={() => onClick(firstFilterOption(filter))}
-                    >
-                        {firstFilterOption(filter)}
-                    </StyledCode>
+                    <ConditionallyRender
```
kwasniew, line 56: don't show `environment:undefined` for empty tables

FredrikOseberg (approved):

PR #4626, opened 2023-09-06
docs: update compatibility matrix with strategy variants

sighphyre:
Adds a line item to the SDK compatibility matrix to reflect strategy variants and marks custom stickiness as no longer in beta, since that's been stable for over a year.


gastonfournier (approved):
:+1:

thomasheartman (approved):
Love it ❤️

PR #4627, opened 2023-09-07
feat: jira plugin page

sjaanus:
image

image


frontend/src/component/menu/routes.ts
```diff
@@ -337,6 +338,14 @@ export const routes: IRoute[] = [
         type: 'protected',
         menu: {},
     },
+    {
+        path: '/integrations/view/:providerId',
```
sjaanus, line 342: Currently this serves only static page on Jira, but if more pages come, there might be possibility to make it generic.

frontend/src/component/common/FormTemplate/FormTemplate.tsx
```diff
@@ -165,22 +165,44 @@ const FormTemplate: React.FC<ICreateProps> = ({
...
+                        API Command{' '}
+                        <Tooltip title="Copy command" arrow>
+                            <IconButton onClick={copyCommand} size="large">
+                                <StyledIcon />
+                            </IconButton>
+                        </Tooltip>
+                    </StyledSubtitle>
+                    <Codebox text={formatApiCode!()} />{' '}
+                </>
+            );
```
thomasheartman, line 204: This will surely work just fine for now, but is there a world wherein the `apiDisabled` variable gets somehow uncoupled from `formatApiCode`, so that the function tries to force `formatApiCode` when it's undefined? Would it be better to just check for that value directly within the component?

Something like:
```tsx
if (!formatApiCode) return null
else return fullcomponent
```

Absolutely not a blocker, though! Just a thought I had.

frontend/src/component/integrations/JiraIntegration/JiraImageContainer.tsx
```diff
@@ -0,0 +1,40 @@
...
+    description,
+    src,
+}) => {
+    return (
+        <StyledContainer>
+            <StyledTitle>{title}</StyledTitle>
+            <Typography>{description}</Typography>
+            <StyledImg src={formatAssetPath(src)} />
+        </StyledContainer>
+    );
```
thomasheartman, line 39: What HTML elements is this rendered as (and what kind of a thing is it)? From the context, it sounds as if it might be an image with a caption? In that case, should we use the `figure` and `caption` elements? And should the `title` be an `h3` or something similar (depending on its position in the page)? If you wanna be very nitty gritty, the StyledContainer could probably be an `article` instead of a `div`, too, but that's not very important.

Also, the images don't have any captions or alt text, which is an accessibility issue. Based on the usage in the new code, I think the typography should be put in a `figcaption` component, linked to the images `figure` component.

frontend/src/component/integrations/JiraIntegration/JiraIntegration.tsx
```diff
@@ -0,0 +1,114 @@
...
+            title={`${displayName}`}
+            description={description || ''}
+            documentationLink={documentationUrl}
+            documentationLinkLabel="Jira documentation"
+        >
+            <StyledContainer>
+                <StyledGrayContainer>
+                    <StyledIconLine>
+                        <IntegrationIcon name={name} /> How does it work?
+                    </StyledIconLine>
```
thomasheartman, line 59: Is this nesting `<p>` tags? I don't think that's valid HTML (and it's quite possible your console is screaming at you - I noticed mine was shouting at me yesterday). 

Putting block-level elements, such as `ul` within `p` tags is also not valid HTML, according to [MDN's article on `p`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p)

frontend/src/component/integrations/JiraIntegration/JiraIntegration.tsx
```diff
@@ -0,0 +1,114 @@
...
+                        View plugin on Atlassian marketplace
+                    </StyledLink>
+                </StyledGrayContainer>
+                <Divider />
+                <JiraImageContainer
+                    title={'Manage your feature flags for each environment'}
+                    description={
+                        'View your feature flag status for each of your environments. Quickly turn features on and off directly within Jira.'
+                    }
+                    src={manage}
```
thomasheartman, line 93: Yeah, based on my understanding of the image container now, I think this would be better served as a [figure with a figcaption element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure). That fig caption takes the place of the alt text in this case too, so you don't need to think about that.

thomasheartman (approved):
No blockers, happy for you to merge this if it gives you more momentum 😄 However, I do have a few points that I think should be addressed either before merging or in a follow-up. Musings on whether variables will be present are completely optional and just thoughts I have. But the points on valid HTML structure and accessibility I think should be taken into account and fixed. 

Happy to take any questions or hear thoughts ☺️

frontend/src/component/integrations/JiraIntegration/JiraIntegration.tsx
sjaanus: @thomasheartman changed to figure. I am not sure if title and description both should belong to figcaption, but based on googleing I think yes.

frontend/src/component/integrations/JiraIntegration/JiraIntegration.tsx
sjaanus: Fixed

frontend/src/component/integrations/JiraIntegration/JiraImageContainer.tsx
sjaanus: Done

thomasheartman: Did you push the changes? For some reason, I don't think I'm seeing them? 🤔

frontend/src/component/integrations/JiraIntegration/JiraIntegration.tsx
thomasheartman: Hmm, title in the figcaption? Yeah, it's permitted content, so I think that makes sense 😄

frontend/src/component/integrations/JiraIntegration/JiraIntegration.tsx
thomasheartman: Same with this: I'm not seeing the changes. But maybe github is slow? Or maybe you haven't pushed yet? 😅

frontend/src/component/integrations/JiraIntegration/JiraImageContainer.tsx
```diff
@@ -0,0 +1,48 @@
...
+    gap: theme.spacing(2),
+    flexDirection: 'column',
+}));
+
+export const StyledFigCaption = styled('figcaption')(({ theme }) => ({
+    display: 'flex',
+    gap: theme.spacing(2),
+    flexDirection: 'column',
+}));
+export const StyledTitle = styled(Typography)({
```
thomasheartman, line 17: Based on the usage (and my understanding from the screenies you posted), it feels like this should be an `h2` element if I'm reading it right? But that's a minor thing, and I'll leave it up to you, whether or not to handle that.

```suggestion
export const StyledTitle = styled('h2')({
```

frontend/src/component/integrations/JiraIntegration/JiraIntegration.tsx
thomasheartman: Yeah, there they are!

thomasheartman (approved):
Super; much better 😄 🙌🏼

PR #4633, opened 2023-09-07
fix: api token schema

gastonfournier:
## About the changes
Create api token schema can either provide the field `project` or its plural: `projects` so the joi validation makes them optional: https://github.com/Unleash/unleash/blob/2be77fb55ee85e18648afa455fbcee6b4d568f9d/src/lib/schema/api-token-schema.ts#L20-L24

This means that when returning the token, the response will either contain `project` or `projects` depending on what was provided as input. Therefore we need to make both fields optional in the response

### DIFF
--- a/src/lib/openapi/spec/__snapshots__/api-token-schema.test.ts.snap
+++ b/src/lib/openapi/spec/__snapshots__/api-token-schema.test.ts.snap
@@ -33,24 +33,6 @@ exports[`apiTokenSchema empty 1`] = `
     {
       "instancePath": "",
       "keyword": "required",
-      "message": "must have required property 'project'",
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):

gastonfournier:
I made the wrong assumption. The token (response) always has the [project](https://github.com/Unleash/unleash/blob/e97859af911b3c9749c67ada08d4b4bf9eb34465/src/lib/types/models/api-token.ts#L44) and [projects](https://github.com/Unleash/unleash/blob/e97859af911b3c9749c67ada08d4b4bf9eb34465/src/lib/types/models/api-token.ts#L32) fields

PR #4635, opened 2023-09-07
chore: return 404 when adding environment to project that doesnt exist

daveleek:
## About the changes

Returns a 404 response instead of 500 when trying to set an environment on a project that doesn't exist


nunogois (approved):
Perfect 👍

PR #4646, opened 2023-09-08
fix: multi project roles UI improvements

nunogois:
https://linear.app/unleash/issue/2-1373/small-ui-fixes-and-improvements

 - Update group root role `HelpIcon` tooltip to reflect the new behavior;
 - Fixes a crash on groups search where `description` could be `undefined`;
 - Improves `RoleDescription` design;
 - Fixes the role label in `ProjectGroupView`;
 
image

### DIFF
--- a/frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
+++ b/frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
@@ -98,6 +98,11 @@ export const ProjectAccessTable: VFC = () => {
     const [groupOpen, setGroupOpen] = useState(false);
     const [selectedRow, setSelectedRow] = useState<IProjectAccess>();
 
+    const roleText = (roles: number[]): string =>
+        roles.length > 1
+            ? `${roles.length} roles`
+            : access?.roles.find(({ id }) => id === roles[0])?.name || '';
+
     const columns = useMemo(
         () => [
             {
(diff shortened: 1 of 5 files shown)


frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
```diff
@@ -490,11 +488,17 @@ export const ProjectAccessTable: VFC = () => {
...
-                subtitle={`Role: ${
-                    access?.roles.find(
-                        ({ id }) => id === selectedRow?.entity.roleId
-                    )?.name
-                }`}
+                subtitle={
+                    <>
+                        {selectedRow && selectedRow.entity.roles.length > 1
+                            ? 'Roles:'
+                            : 'Role:'}
```
daveleek, line 495: Will there always be >= 1 roles? Or could selectedRow.entity.roles.length be 0 and should it then say "Roles:" if there are 0 roles in selectedRow.entity.roles ?

daveleek (approved):
LGTM! Single question but you can ignore

frontend/src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
nunogois: I think this is safe. This is shown inside project access and only when you click in a group that has access, which means they should have a role. Worst case scenario it would display "Role:      ", which doesn't sound super bad.

PR #4650, opened 2023-09-08
fix: API improvements aligning the types to our schemas

gastonfournier:
Some of our types in OSS have drifted apart from our OpenAPI schemas. This will help them be aligned again


src/lib/openapi/spec/create-strategy-schema.ts
```diff
@@ -6,12 +6,18 @@ export const createStrategySchema = {
     description:
         'The data required to create a strategy type. Refer to the docs on [custom strategy types](https://docs.getunleash.io/reference/custom-activation-strategies) for more information.',
     required: ['name', 'parameters'],
+    additionalProperties: false,
```
gastonfournier, line 9: Important one so we have to declare all properties in the schema. This is input, so I'm not sure how specific we want to be with input data. Maybe we don't need this

@thomasheartman maybe you remember some decisions around additional properties. My take is that for input is not so important, as customers might want to extend Unleash, but I'm not 100% sure... I see in some cases we're doing it so should be fine: https://github.com/Unleash/unleash/blob/b60e581893587159029a240c894df34286660e97/src/lib/openapi/spec/create-user-schema.ts#L3-L6

thomasheartman: the discussion we had previously was: 
always allow additional properties on inputs, never on outputs. 

Following the old adage to be liberal in what you accept, but strict in what you return, we thought it best to allow any additional properties coming in. We didn't want to break a user's experience because they added a field that we don't recognize.

On the output front, we want to be very clear about what we expect and we want the types to be accurate. For that to be true, we needed conversions to fail if they contained extra data. 

---

Based on that, I don't think we should have additionalProperties: false here. Did you do it for a reason?

On the other hand, one drawback of never using additionalProperties: false on input data, is that it makes creating union types harder (or almost impossible without a discriminator).

But that said, if we require a property to be supplied by the user, then we can use the `required` property.

Could you elaborate a bit on this? I'm not sure I understand.
> Important one so we have to declare all properties in the schema.
 

Anyway, my suggestion is to remove this line.

```suggestion
```

src/lib/types/model.ts
```diff
@@ -402,6 +402,14 @@ export interface IImportData extends ImportCommon {
     data: any;
 }
 
+// Create project aligns with #/components/schemas/createProjectSchema
+// joi is providing default values when the optional inputs are not provided
+// const data = await projectSchema.validateAsync(newProject);
+export type CreateProject = Pick<IProject, 'id' | 'name'> & {
+    mode?: ProjectMode;
+    defaultStickiness?: string;
+};
```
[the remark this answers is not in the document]
thomasheartman: Nice. Could we use OpenAPI to set defaults instead? It has a `default` property you can use. I think that's better than using Joi. Honestly, I was hoping we could get rid of Joi now that we've openapi'd everything. Having two layers of API validation seems excessive (not to mention it can get confusing when you're not sure where an error is coming from).

thomasheartman (commented):
Nice work! I want to figure out the discussion points before approving this, but it's looking very good 😄

chriswk (approved):
This seems fair. I'd like to see us move away from the id for Permission and just use the name as the primary key, since we do require the name to be unique anyway. It can come as a separate PR though, so I'm inclined to give a :+1: on this.

gastonfournier:
> I'd like to see us move away from the id for Permission and just use the name as the primary key

+1 maybe I can mark the type as deprecated

PR #4651, opened 2023-09-11
feat: keyboard navigation in search

kwasniew:
## About the changes

Screenshot 2023-09-11 at 11 45 51

* tab navigation inside search suggestions and outside search suggestions
* when tab moves away from the search suggestions, the suggestions are hidden
* also supports escape key from the search suggestions
* added onUseBlur reusable hook that handles element loosing focus
* fix search and filter to filter our filters if they don't exists (previously it was undefined)

## Discussion points

For now the elements we tab between are spans even though button would be more correct. Buttons add too many styles that we'd need to remove.

### DIFF
--- a/frontend/src/component/common/Search/Search.tsx
+++ b/frontend/src/component/common/Search/Search.tsx
@@ -9,7 +9,7 @@ import { useKeyboardShortcut } from 'hooks/useKeyboardShortcut';
 import { SEARCH_INPUT } from 'utils/testIds';
 import { useOnClickOutside } from 'hooks/useOnClickOutside';
 import { useSavedQuery } from './useSavedQuery';
-import { useOnBlur } from '../../../hooks/useOnBlur';
+import { useOnBlur } from 'hooks/useOnBlur';
 
 interface ISearchProps {
     id?: string;
--- a/frontend/src/component/common/Search/Search.tsx
+++ b/frontend/src/component/common/Search/Search.tsx
@@ -9,6 +9,7 @@ import { useKeyboardShortcut } from 'hooks/useKeyboardShortcut';
 import { SEARCH_INPUT } from 'utils/testIds';
 import { useOnClickOutside } from 'hooks/useOnClickOutside';
 import { useSavedQuery } from './useSavedQuery';
+import { useOnBlur } from '../../../hooks/useOnBlur';
 
 interface ISearchProps {
     id?: string;
@@ -111,7 +112,10 @@ export const Search = ({
         }
     );
     useKeyboardShortcut({ key: 'Escape' }, () => {
-        if (document.activeElement === searchInputRef.current) {
+        if (
+            document.activeElement === searchInputRef.current ||
+            searchContainerRef.current?.contains(document.activeElement)
+        ) {
             searchInputRef.current?.blur();
             hideSuggestions();
         }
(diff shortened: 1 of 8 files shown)


frontend/src/component/common/Search/Search.tsx
```diff
@@ -111,14 +112,15 @@ export const Search = ({
         }
     );
     useKeyboardShortcut({ key: 'Escape' }, () => {
-        if (document.activeElement === searchInputRef.current) {
+        if (searchContainerRef.current?.contains(document.activeElement)) {
```
kwasniew, line 115: escape anywhere in the search area should hide suggestions

frontend/src/component/common/Search/SearchSuggestions/SearchInstructions/SearchInstructions.tsx
```diff
@@ -57,6 +61,10 @@ export const SearchInstructions: VFC<ISearchInstructionsProps> = ({
                         condition={filter.options.length > 0}
                         show={
                             <StyledCode
+                                tabIndex={0}
```
kwasniew, line 64: previously we only supported clicks, now we support on  enter when tab index is selected

frontend/src/component/common/Search/SearchSuggestions/SearchSuggestions.test.tsx
```diff
@@ -106,3 +136,22 @@ test('displays search and filter instructions when filter value is provided', ()
...
+            getSearchContext={() => searchContextWithoutFilters}
+        />
+    );
+
+    expect(
+        screen.getByText(/Start typing to search in Title, Environment/i)
+    ).toBeInTheDocument();
+
+    screen.getByText(/Title A/i).click();
+    expect(recordedSuggestion).toBe('Title A');
```
kwasniew, line 156: previously undefined from filters was added

frontend/src/component/common/Search/SearchSuggestions/SearchSuggestions.tsx
```diff
@@ -103,9 +104,37 @@ export const SearchSuggestions: VFC<SearchSuggestionsProps> = ({
...
 
-    const selectedFilter = filters.map(
-        filter => `${filter.name}:${filter.suggestedOption}`
-    )[0];
+    const selectedFilter =
+        filters.length === 0
+            ? ''
+            : filters.map(
+                  filter => `${filter.name}:${filter.suggestedOption}`
+              )[0];
```
kwasniew, line 112: select index only when filters are present

frontend/src/component/common/Search/SearchSuggestions/SearchSuggestions.tsx
```diff
@@ -103,9 +104,37 @@ export const SearchSuggestions: VFC<SearchSuggestionsProps> = ({
...
+
+    const onFilter = (suggestion: string) => {
```
kwasniew, line 114: extracted existing methods since we use them in click and key down handlers

frontend/src/hooks/useOnBlur.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect } from 'react';
+
+export const useOnBlur = (
```
kwasniew, line 3: reusable hook for navigation away from the element

Tymek (approved):
LGTM

PR #4652, opened 2023-09-11
feat: make import/export work with project patterns

thomasheartman:
This PR adds feature name pattern validation to the import validation step. When errors occur, they are rendered with all the offending features, the pattern to match, plus the pattern's description and example if available.

image

To achieve this I've added an extra method to the feature toggle service that checks feature names without throwing errors (because catching `n` async errors in a loop became tricky and hard to grasp). This method is also reused in the existing feature name validation method and handles the feature enabled chcek.

In doing so, I've also added tests to check that the pattern is applied.


kwasniew (approved):
I'd test those scenarios if you haven't done so:
* what happens when we call import API directly (bypassing validation stage we force in the UI)
* what happens when we use the import from the enterprise version (it's using OSS import but I'd check anyway)

thomasheartman:
@kwasniew Good suggestions; will do! I actually started on the first suggestion, but it threw some seemingly unrelated errors, but I'll revisit it. As for the second one: do you mean to add a test for this in the enterprise repo as well?

thomasheartman:
When you call the API directly, bypassing the validation, you get a 400 saying that the feature (named X) doesn't match the project's pattern Y. So I'd say that's as expected. If you supply multiple features that don't match, only the first one is returned in the error. This is probably due to how import/export works (does one at a time, exiting on error?), so I don't think changing that is within the scope of this change.

kwasniew:
@thomasheartman As long as we stop people from a direct import with any error message it's fine. Validation is how you get detailed errors. Direct import should just fail when you do anything illegal and it should fail with the first error.

PR #4655, opened 2023-09-11
Fix: integrations form

Tymek:
## About the changes
Fix submitting integrations form


frontend/src/component/integrations/IntegrationList/AvailableIntegrations/AvailableIntegrations.tsx
```diff
@@ -140,15 +134,15 @@ export const AvailableIntegrations: VFC<IAvailableIntegrationsProps> = ({
                             </a>
                         </Typography>
                     </div>
-                    <div>
+                    <Box sx={theme => ({ marginTop: theme.spacing(2) })}>
```
sjaanus, line 137: Better as styled components, since you use it also twice.

sjaanus (approved):
One things could be changed

PR #4659, opened 2023-09-12
fix: remove last seen by env

andreas-unleash:
## About the changes

Closes #


gardleopard (changes requested):
This is the wrong branch. customer is running 5.4.2

PR #4660, opened 2023-09-12
fix: remove last seen by env

andreas-unleash:
## About the changes

Closes #


gardleopard (changes requested):
I dont think we should get this out on the 5.4 branch, this is a specific mitigation we are doing.

andreas-unleash:
outdated - closing

PR #4664, opened 2023-09-12
feat: add implicit surrounding `^` and `$` to patterns

thomasheartman:
This PR updates the back-end handling of feature naming patterns to add implicit leading `^`s and trailing `$`s to the regexes when comparing them.

It also adds tests for the new behavior, both for new flag names and for examples.

## Discussion points

Regarding stripping incoming ^ and $: We don't actually need to strip incoming `^`s and `$`s: it appears that `^^^^^x$$$$$` is just as valid as `^x$`. As such, we can leave that in. However, if we think it's better to strip, we can do that too.

Second, I'm considering moving the flag naming validation into a dedicated module to encapsulate everything a little better. Not sure if this is the time or where it would live, but open to hearing suggestions.


src/lib/services/project-service.ts
```diff
@@ -1043,3 +1012,42 @@ export default class ProjectService {
...
+
+    if (featureNaming.pattern && !featureNaming.example) {
+        featureNaming.example = null;
+    }
+    if (featureNaming.pattern && !featureNaming.description) {
+        featureNaming.description = null;
+    }
+
+    return featureNaming;
+};
```
thomasheartman: I've opted for mutating the input object here instead of creating a copy. That seems to be what we usually do, but I'd very happy to use a copy instead.

src/lib/services/project-service.ts
```diff
@@ -61,6 +61,44 @@ import { ProjectDoraMetricsSchema } from 'lib/openapi';
...
```
thomasheartman, line 100: This has been moved out of the project service class to make it easier to unit test.

thomasheartman, line 100: I've opted for mutating the input object here instead of creating a copy. That seems to be what we usually do, but I'd very happy to use a copy instead.

src/lib/services/project-service.ts
```diff
@@ -61,6 +61,42 @@ import { ProjectDoraMetricsSchema } from 'lib/openapi';
...
+
```
thomasheartman, line 99: This has been moved out of the project service class to make it easier to unit test.

src/lib/services/project-service.ts
```diff
@@ -61,6 +61,42 @@ import { ProjectDoraMetricsSchema } from 'lib/openapi';
...
+    featureNaming: IFeatureNaming,
+): IFeatureNaming => {
+    validateFlagNaming(featureNaming);
+
+    if (featureNaming.pattern && !featureNaming.example) {
+        featureNaming.example = null;
+    }
+    if (featureNaming.pattern && !featureNaming.description) {
+        featureNaming.description = null;
+    }
```
thomasheartman, line 95: Null fields out if people update a pattern without touching the description or example.

kwasniew (approved):
I'd extract name validation code into a separate file based on the following principle: "things that change together should be kept together". If you feel like you keep changing the same name validation functions over and over again independently from the other code in the feature-toggle-service and project-service then you have a candidate for extraction.

thomasheartman:
@kwasniew More lines, but it's been extracted into a separate feature directory (feature-naming-pattern) and most tests have been moved with it. It feels like a better way to do it to me (even if it meant changing some more files), so I appreciate the push. Now all the actual validation of input data and of feature names live in the same file.

PR #4667, opened 2023-09-12
feat: update UI to add hints about implicit ^ and $

thomasheartman:
This PR updates the UI to reflect the changes to the implicit ^ and $ that we now add. The changes are:

1. Show input adornments for ^ and $ when you create a pattern.
2. Mention that ^ and $ are added implicitly in description.
3. Checks the example you provide against the pattern with added ^ and $ + adds a test for that. 

~~Points 1 and 2:~~
1 and 2

**Update** Points 1 and 2:
image

## Discussion point:

I have not touched the information about the pattern yet as the PR that updates that is still in review (#4656), but it would be prudent to also update that info to make it clearer. I can address that in a follow-up PR.


kwasniew (approved):
The description and the input field feel a bit "crowded".

thomasheartman:
@kwasniew Agree. We could probably help that a bit with some more spacing. I'll loop @nicolaesocaciu in, and we can have a look at it.

PR #4672, opened 2023-09-12
fix: post global events even when filtering by env

nunogois:
https://linear.app/unleash/issue/2-1276/deal-with-events-without-an-environment

This makes it so that global events (events not tied to a specific environment) are always received, no matter the configured environment filter.

It also includes a respective test and small sentence on the UI explaining the behavior.

image


gastonfournier (approved):
Looks fine! Great with the test!

PR #4674, opened 2023-09-12
feat: Add active users statistics to metrics

Tymek:
## About the changes
  - `getActiveUsers` is using multiple stores, so it is refactored into read-model
  - Refactored Instance stats service into `features` to co-locate related code

Closes https://linear.app/unleash/issue/UNL-230/active-users-prometheus

### Important files
`src/lib/features/instance-stats/getActiveUsers.ts`

## Discussion points
`getActiveUsers` is coded less _class-based_ then previous similar read-models. In one file instead of 3 (read-model interface, fake read model, sql read model). I find types and functions way more readable, but I'm ready to refactor it to interfaces and classes if consistency is more important.


src/lib/db/user-store.ts
```diff
@@ -204,24 +204,47 @@ class UserStore implements IUserStore {
...
+        const result = await this.db
+            .with('Combined', combinedQuery)
+            .select({
+                last_week: this.db.raw(
+                    "COUNT(DISTINCT CASE WHEN seen_at > NOW() - INTERVAL '1 week' THEN user_id END)",
+                ),
+                last_month: this.db.raw(
+                    "COUNT(DISTINCT CASE WHEN seen_at > NOW() - INTERVAL '1 month' THEN user_id END)",
+                ),
+                last_two_months: this.db.raw(
```
ivarconr, line 228: would love to see tests for this query,

Tymek:
Performance: Average over 100 runs, all stats endpoint queries.
```
prior to change: 6.15ms
10 users, 30 tokens: 6.36ms
100 users, 300 tokens: 6.91ms
1000 users, 3000 tokens: 7.75ms
10000 users, 30000 tokens: 32.33ms
```
I didn't see any excessive memory use.

src/lib/features/instance-stats/instance-stats-service.ts
Tymek, line 1: File moved to `features`, in order to co-locate injected `getActiveUsers`

src/lib/features/instance-stats/getActiveUsers.ts
Tymek, line 1: Less _class-based_. In one file instead of 3 (read-model interface, fake read model, sql read model). I find types and functions way more readable, but I'm ready to refactor it to interfaces and classes if consistency is more important.

PR #4675, opened 2023-09-13
UI: make project settings / creation form full-width

thomasheartman:
This PR makes the project settings / creation forms grow to take up the full available width, so that it is in line with the new designs.


nicolaesocaciu (approved):
LGTM!

PR #4681, opened 2023-09-13
feat: stop regexes with whitespace

thomasheartman:
This PR stops regexes with whitespace from being added as feature naming patterns.


andreas-unleash (approved):
Nice!

PR #4683, opened 2023-09-13
fix: integration multiselector

Tymek:
Input should use state set outside and derive "all selected" from it. This was not the case causing issues when loading a form with "wildcard" pre-selected.


nunogois (approved):
LGTM

PR #4686, opened 2023-09-13
feat: playground custom properties are nested

kwasniew:
## About the changes

Screenshot 2023-09-13 at 15 22 50

Properties in playground should be copy-paste friendly for the SDK. That's why custom properties are added in nested properties as in the SDKs. The auto-formatting/normalization works when we're adding values with add/replace button.

## Discussion points

Backend code can generate context field combinations on flat and nested property fields. So our input is more forgiving now. The API response stays the same for now with a flat list of properties since I don't want to change the UI table results.

So even if you send
```
{appName: app1, properties: {customProp: val1}}
```
In the response you get
```
{appName:app1, customProp: val1}
```

### DIFF
--- a/frontend/src/component/playground/Playground/PlaygroundForm/PlaygroundCodeFieldset/PlaygroundCodeFieldset.tsx
+++ b/frontend/src/component/playground/Playground/PlaygroundForm/PlaygroundCodeFieldset/PlaygroundCodeFieldset.tsx
@@ -27,7 +27,10 @@ import { formatUnknownError } from 'utils/formatUnknownError';
 import useToast from 'hooks/useToast';
 import { PlaygroundEditor } from './PlaygroundEditor/PlaygroundEditor';
 import { parseDateValue, parseValidDate } from 'component/common/util';
-import { isStringOrStringArray } from '../../playground.utils';
+import {
+    isStringOrStringArray,
+    normalizeCustomContextProperties,
+} from '../../playground.utils';
 interface IPlaygroundCodeFieldsetProps {
     context: string | undefined;
     setContext: Dispatch<SetStateAction<string | undefined>>;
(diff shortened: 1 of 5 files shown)


frontend/src/component/playground/Playground/PlaygroundForm/PlaygroundCodeFieldset/PlaygroundCodeFieldset.tsx
```diff
@@ -58,7 +61,10 @@ export const PlaygroundCodeFieldset: VFC<IPlaygroundCodeFieldsetProps> = ({
                 try {
                     const contextValue = JSON.parse(input);
 
-                    setFieldExist(contextValue[contextField] !== undefined);
+                    setFieldExist(
+                        contextValue[contextField] !== undefined ||
```
kwasniew, line 65: we need to also check nested properties now

src/lib/features/playground/generateObjectCombinations.ts
```diff
@@ -1,14 +1,19 @@
-type Dict<T> = { [K in keyof T]: string[] };
+type Dict<T> = { [K in keyof T]: (string | number)[] };
 
 export const splitByComma = <T extends Record<string, unknown>>(
```
kwasniew, line 3: it had to change to support nested properties

src/lib/features/playground/generateObjectCombinations.test.ts
```diff
@@ -37,3 +37,28 @@ test('should generate all combinations correctly when only one combination', ()
 
     expect(actualCombinations).toEqual(expectedCombinations);
 });
+
+test('should generate combinations with nested properties', () => {
+    const obj = {
+        sessionId: '1,2',
+        nonString: 2,
+        properties: {
+            nonString: 1,
```
kwasniew, line 46: properly nested custom properties take precedence

thomasheartman (approved):
Looks good to me 😄 One question though: what happens if you provide a custom property (in the properties object) with the same name as one of the standard properties?

For instance:
```js
appName: "top-level:
properties: {
  appName: "nested"
}
```

I know that we can do that with SDKs today, so the SDK handling should be standardized (I hope? 🤔), but is it in line with how you flatten it for the return? What **would** this be returned as? 

```js
appName: "top-level"
```

or 

```js
appName: "nested"
```

kwasniew:
@thomasheartman raised a slack discussion about it. Currently custom properties always win.

PR #4688, opened 2023-09-13
fix: use postmessage in slack app addon

nunogois:
https://linear.app/unleash/issue/2-1392/fix-flaky-unhandled-events

 - Reverts part of https://github.com/Unleash/unleash/pull/4490 and uses `postMessage` instead. This prevents an error where scheduling the message was flaky and some events ended up not being handled at all (see screenshot below). As a bonus, this simplifies our code and prevents having a delay. It seems like this method still works with the channel name instead of needing an id, which was the main motivation towards the changes in the aforementioned PR;
 - Removes `FEATURE_UPDATED` from the event options for the Slack App addon, as [this event is deprecated](https://docs.getunleash.io/reference/environments#addons);
 - Adds support for events without a specific project, including a test, similar to https://github.com/Unleash/unleash/pull/4672
 - Misc cleanups;

image

### DIFF
--- a/src/lib/addons/slack-app-definition.ts
+++ b/src/lib/addons/slack-app-definition.ts
@@ -1,6 +1,5 @@
 import {
     FEATURE_CREATED,
-    FEATURE_UPDATED,
     FEATURE_ARCHIVED,
     FEATURE_REVIVED,
     FEATURE_STALE_ON,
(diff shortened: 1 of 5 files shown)


chriswk (approved):
Nice. I could not find this method which supported the channel argument when I did this. 100% support the change.

PR #4689, opened 2023-09-13
chore: improve access service

gastonfournier:
## About the changes
This enables us to use names instead of permission ids across all our APIs at the computational cost of searching for the ids in the DB but improving the API user experience

## Open topics
We're using methods that are test-only and circumvent our business logic. This makes our test to rely on assumptions that are not always true because these assumptions are not validated frequently.

i.e. We are expecting that after removing a permission it's no longer there, but to test this, the permission has to be there before:
https://github.com/Unleash/unleash/blob/78273e4ff30d7012aa91f6549115587dc31a6e0b/src/test/e2e/services/access-service.e2e.test.ts#L367-L375

But it seems that's not the case.

We'll look into improving this later.


nunogois (dismissed):
Makes sense to me 🤷

src/test/e2e/services/access-service.e2e.test.ts
```diff
@@ -363,6 +363,9 @@ test('should remove CREATE_FEATURE on default environment', async () => {
         permissions.CREATE_FEATURE,
         '*',
     );
+    expect(
+        await accessService.hasPermission(user, CREATE_FEATURE, 'some-project'),
```
nunogois, line 367: If these are global permissions, perhaps we should check `default` project instead.

```suggestion
        await accessService.hasPermission(user, CREATE_FEATURE, 'default'),
```

nunogois: Alternatively, perhaps 'some-project' does not exist in the context of the test, so permissions for this project always return false.

chriswk (dismissed):
Seems reasonable.

src/lib/services/access-service.ts
```diff
@@ -630,11 +634,13 @@ export class AccessService {
...
         if (rolePermissions) {
             if (roleType === CUSTOM_ROOT_ROLE_TYPE) {
+                // this branch uses named permissions
                 await this.store.addPermissionsToRole(
                     newRole.id,
-                    rolePermissions.map((p: NamePermissionRef) => p.name),
+                    rolePermissions,
                 );
             } else {
+                // this branch uses id permissions
```
gastonfournier, line 643: This is the reason why we're making this change, we want to be able to use names instead of ids for permissions

src/lib/schema/role-schema.ts
```diff
@@ -3,9 +3,11 @@ import joi from 'joi';
 export const permissionRoleSchema = joi
     .object()
     .keys({
-        id: joi.number().required(),
+        id: joi.number(),
+        name: joi.string(),
```
gastonfournier, line 7: This validation now expects that either you provide the id or the name (or both)

src/lib/db/access-store.ts
```diff
@@ -700,18 +772,16 @@ export class AccessStore implements IAccessStore {
 
     async addPermissionsToRole(
         role_id: number,
-        permissions: string[],
+        permissions: Omit<PermissionRef, 'environment'>[],
```
gastonfournier, line 775: Omit the environment just to avoid mistakes, because all permissions should be part of the environment passed as a second parameter.

src/lib/db/access-store.ts
```diff
@@ -46,13 +50,79 @@ interface IPermissionRow {
...
+        return permissions.map((permission) => {
+            if (this.isIdPermissionRef(permission)) {
+                return permission as ResolvedPermission;
+            } else {
+                return idPermissionsFromNamed.find(
+                    (p) => p.name === (permission as NamePermissionRef).name,
+                )!;
+            }
+        });
+    };
```
gastonfournier, line 124: I'll add some tests for this

src/lib/db/access-store.ts
```diff
@@ -46,13 +50,79 @@ interface IPermissionRow {
...
+            .select('id as permissionId', 'permission')
+            .from(T.PERMISSIONS)
+            .whereIn('permission', permissionNames);
+
+        return rows.map((row) => ({
+            id: row.permissionId as number,
+            name: row.permission as string,
+            environment: permissions.find(
+                (p) => p.name === (row.permission as string),
+            )?.environment,
```
gastonfournier, line 92: I don't like the performance of this... a find inside a map is not ideal

sighphyre: Is it a problem? O(n*m) isn't an issue when n and m are small. Should be easy enough to build a Map earlier in the function and get O(1) lookups

gastonfournier: Exactly what I had in mind

chriswk: Assume positive intentions here, I'm struggling with formulating the question. 
Why are we fetching ids again? Wasn't the goal to change to names as the primary key?

gastonfournier: > Wasn't the goal to change to names as the primary key?

Yes, but our OpenAPI spec states that ids are required while names are optional. Adding the name as required is a breaking change having to wait until the next major release. **I think** because of this reason we can't drop support for id's (customers might be using them already).

src/test/fixtures/no-logger.ts
```diff
@@ -1,23 +1,27 @@
...
-noLoggerProvider.setMuteError = (mute) => {
+noLoggerProvider.setMuteError = (mute: boolean) => {
     muteError = mute;
 };
 
+// use for debugging only, try not to commit tests with verbose set to true
+noLoggerProvider.setVerbose = (beVerbose: boolean) => {
+    verbose = beVerbose;
+};
+
```
gastonfournier, line 25: Unrelated, but I thought it might be useful

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -175,33 +176,32 @@ test('Token-admin should be allowed to create token', async () => {
...
-            );
             await accessService.addUserToRole(
                 user.id,
                 createClientApiTokenRole.id,
                 'default',
             );
-            req.user = await userService.createUser({
-                email: 'someguyinplaces@example.com',
-                rootRole: role.id,
-            });
```
gastonfournier, line 204: I've refactored this method, not sure what we were trying to do here:
1. **Get Viewer built-in role**
2. Create a user A
3. Create a new role "token creator" without permissions
4. **Add CREATE_PROJECT permission to the Viewer role (shouldn't be possible to add permissions programmatically)**
5. Add user A to the "token creator role" in project 'default' (now A has the root role Viewer and "token creator role" in 'default')
6. **Create a new user B with root role Viewer (that now has the new permission)**
7. **Associate req.user to the newly created user B**

Then the test tries to create an API token using the newly created user B.

Why did we take all those efforts to setup user A if we're not using it in the test? Apparently the only important steps are the ones in **bold**. At least taking into account the title of the test: `A role with only CREATE_PROJECT_API_TOKEN can create project tokens`

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -175,33 +176,32 @@ test('Token-admin should be allowed to create token', async () => {
...
+            req.user = user;
```
gastonfournier, line 204: On this side:
1. Get viewer role
2. Create a user with Viewer role as root role
3. Create a new role with CREATE_PROJECT_API_TOKEN permission
4. Add the user to the role created in `default` project
5. Associate `req.user` to the user created

Later we validate if this user can create an API token in the project where they have the role with that specific permission

src/lib/db/access-store.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    args([
+        { id: 1, environment: 'development' },
+        { id: 2, name: 'ignore this name' },
+    ]),
+])(
+    'resolvePermissions with permission ids (%o) returns the list unmodified',
+    async (permissions) => {
+        const access = db.stores.accessStore as AccessStore;
+        const result = await access.resolvePermissions(permissions);
+        expect(result).toStrictEqual(permissions);
```
gastonfournier, line 56: I'd love to say: no query to the DB here

src/lib/db/access-store.ts
```diff
@@ -46,13 +50,84 @@ interface IPermissionRow {
...
+            return [];
+        }
+        const permissionNames = permissions
+            .filter((p) => p.name !== undefined)
+            .map((p) => p.name);
+
+        const rows = await this.db
+            .select('id', 'permission')
+            .from(T.PERMISSIONS)
+            .whereIn('permission', permissionNames);
```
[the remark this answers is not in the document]
gastonfournier: [d5c3402](https://github.com/Unleash/unleash/pull/4689/commits/d5c34022f614c212228d8362b2eaa4f3d90335fd)

src/lib/db/access-store.ts
```diff
@@ -63,6 +73,75 @@ export class AccessStore implements IAccessStore {
...
+        if (permissionNames.length === 0) {
+            return [];
+        }
+
+        const stopTimer = this.timer('permissionNamesToIds');
+
+        const rows = await this.db
+            .select('id', 'permission')
+            .from(T.PERMISSIONS)
+            .whereIn('permission', permissionNames);
```
gastonfournier, line 96: Same query as here: https://github.com/Unleash/unleash/pull/4689/files#diff-5cb4eca71e6a03ec3d123986d657368a3386f80e986dfef2e3feb0abb0673d87L706-L709 but now it can be used in all scenarios of adding permissions to a role

src/lib/db/access-store.ts
sighphyre: I think this is correct, we can trade off some performance for correctness for a while. When we make a breaking change we can clean this up

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
sighphyre: Guessing copy paste bug. Something something don't trust tests you haven't seen fail

sighphyre (approved):
LGTM, I think with the tests and timers this is good enough to try

PR #4706, opened 2023-09-14
fix: integrations text review

Tymek:
## About the changes
Update copy


frontend/src/component/integrations/IntegrationList/AvailableIntegrations/SDKs.ts
```diff
@@ -58,8 +58,8 @@ export const OFFICIAL_SDKS: Sdk[] = [
     },
     {
         name: 'dotnet',
-        displayName: '.Net SDK',
-        description: 'Officially Unleash Client for .Net',
+        displayName: '.NET SDK',
+        description: 'Officially Unleash Client for .NET',
```
thomasheartman, line 62:
```suggestion
        description: 'Official Unleash Client for .NET',
```

I know you fixed some of these in the previous PR, but there are still more.  All the `officially` in this file can be replaced with `official`.

thomasheartman (approved):
Looks good, but just make sure that all the "officially" in the available integrations file are updated to be "official" instead 😄

PR #4708, opened 2023-09-14
feat: add service method to retrieve group and project access for all users

sighphyre:
This allows us to get the project and group level access for all users in the system. This is built to support an enterprise feature


chriswk (commented):
Let's add timing to this query, just in case it's too heavy.

chriswk (approved):
:+1:

sighphyre:
> Let's add timing to this query, just in case it's too heavy.

Excellent point! Thank you, added

PR #4711, opened 2023-09-14
fix: include tags in variants event

nunogois:
https://linear.app/unleash/issue/2-1396/fix-variants-event-should-include-the-respective-feature-flag-tags

This fixes an issue where the `feature-environment-variants-updated` event would not post to a tagged Slack channel, since it would not take into consideration the feature flag tags.

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -1959,6 +1959,8 @@ class FeatureToggleService {
             ).variants ||
             [];
 
+        const tags = await this.tagStore.getAllTagsForFeature(featureName);
+
         await this.eventStore.store(
             new EnvironmentVariantEvent({
                 featureName,
@@ -1967,6 +1969,7 @@ class FeatureToggleService {
                 createdBy: user,
                 oldVariants: theOldVariants,
                 newVariants: fixedVariants,
+                tags,
             }),
         );
         await this.featureEnvironmentStore.setVariantsToFeatureEnvironments(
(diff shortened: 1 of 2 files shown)


gastonfournier (commented):
I think this solves the problem, but... where should we add this query for tags? 

We're adding a side effect on the creation of the event to add something that's unrelated with the event itself. Should we do this query instead in the SlackApp addon when we receive an event with a `featureName` and `tags === undefined`?

chriswk (approved):
Nice catch.

nunogois:
> I think this solves the problem, but... where should we add this query for tags?
> 
> We're adding a side effect on the creation of the event to add something that's unrelated with the event itself. Should we do this query instead in the SlackApp addon when we receive an event with a `featureName` and `tags === undefined`?

Good point, however this seems like our approach for other events as well, so at least it's consistent. There are many examples of this exact flow in `feature-toggle-service`. Perhaps we should evaluate whether there's a better way of fetching the feature tags, but I think it falls outside of the scope for this fix.

gastonfournier:
> Good point, however this seems like our approach for other events as well, so at least it's consistent. There are many examples of this exact flow in `feature-toggle-service`. Perhaps we should evaluate whether there's a better way of fetching the feature tags, but I think it falls outside of the scope for this fix.

Ok, standardization is a good reasoning to continue with this idea, I'm just thinking that by following my proposal, we can remove all this duplicated code, and we may even catch other events that are in the same situation. After all, not all listeners of the event need to know about the tags, we're adding this information just because of the way we built Slack Addon, therefore we're coupling our code to one particular solution.

But I agree, we can consider it out of the scope, especially if this involves changing other unrelated parts of the codebase.

PR #4712, opened 2023-09-14
fix: env variants event changelog

nunogois:
https://linear.app/unleash/issue/2-1397/fix-environment-variants-change-event-does-not-include-changelogs

By running `applyPatch` without cloning the `oldVariants`, `applyPatch` would patch the `oldVariants` by reference, effectively making them the same as the `newVariants`. This fix `deepClone`s the oldVariants when sending them in as an `applyPatch` parameter so that the `oldVariants` variable is left untouched.

image

### DIFF
--- a/src/lib/services/feature-toggle-service.ts
+++ b/src/lib/services/feature-toggle-service.ts
@@ -1962,9 +1962,6 @@ class FeatureToggleService {
             ).variants ||
             [];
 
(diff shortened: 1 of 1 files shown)


chriswk (approved):
LGTM

gastonfournier (approved):

PR #4719, opened 2023-09-14
fix: Variant type number duplicate options

andreas-unleash:
Fixes a bug where the `number` option was shown multiple times


frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
```diff
@@ -199,7 +199,10 @@ export const VariantForm = ({
     const variantTypeNumber = useUiFlag('variantTypeNumber');
 
     useEffect(() => {
-        if (variantTypeNumber) {
+        if (
+            variantTypeNumber &&
+            !payloadOptions.includes({ key: 'number', label: 'number' })
```
kwasniew, line 204: is include using equality check? what if you have 2 identical objects that are 2 different memory instances

kwasniew (approved):

PR #4729, opened 2023-09-14
refactor: drop tags in events

nunogois:
https://linear.app/unleash/issue/2-1398/refactor-consider-only-fetching-tags-for-events-when-needed

Addresses the discussion in https://github.com/Unleash/unleash/pull/4711#issuecomment-1719279670

This drops tags in events in favor of fetching the feature tags when needed (e.g. when the event is handled), saving us an extra database query for multiple actions and simplifying the model and flow.

I noticed we introduced [**searching events by tags**](https://github.com/Unleash/unleash/pull/4604) (cc @kwasniew) recently. Should we still support searching for events by tags? If so, I suggest we adapt the query to perform a join by `featureName` instead and fetch the current tags. 
If you agree, I can introduce this in this PR.

We can also introduce a migration to drop the `tags` column in the `events` table in a [later PR](https://linear.app/unleash/issue/2-1399/write-a-migration-to-drop-the-tags-column-in-the-events-table).


nunogois:
Closed after internal discussion.

PR #4738, opened 2023-09-15
feat: sdk flat context schema

kwasniew:
## About the changes

Playground response flattens all context fields to make it easy to display them in a table. SDK type reflects that and separates response context from the nested request context with custom properties.


thomasheartman (approved):
Love it 😄

PR #4743, opened 2023-09-15
Feat/project private mode

andreas-unleash:
Adds `private` option to project mode
Update schemas and models to accept and persist

Closes # [1-1366](https://linear.app/unleash/issue/1-1366/introduce-private-in-collaboration-mode)


frontend/src/component/project/Project/ProjectForm/CollaborationModeTooltip.tsx
```diff
@@ -1,6 +1,8 @@
 import { Box, styled, Typography } from '@mui/material';
 import { FC } from 'react';
 import { HelpIcon } from 'component/common/HelpIcon/HelpIcon';
+import { useUiFlag } from '../../../../hooks/useUiFlag';
```
sjaanus, line 4: Shorten

frontend/src/component/project/Project/ProjectForm/CollaborationModeTooltip.tsx
```diff
@@ -11,25 +13,40 @@ const StyledDescription = styled(Typography)(({ theme }) => ({
...
+            htmlTooltip
+            tooltip={
+                <>
+                    <Box>
+                        <StyledTitle>open: </StyledTitle>
+                        <StyledDescription>
+                            everyone can submit change requests
+                        </StyledDescription>
+                    </Box>
+                    <Box sx={{ mt: 2 }}>
```
sjaanus, line 29: This could be styled component. You use the same styles also in bottom.

sjaanus (approved):

frontend/src/component/project/Project/ProjectForm/CollaborationModeTooltip.tsx
andreas-unleash: Box is already a styled 'div' - SxProps are for overrides - I think it is ok as is - but happy to change it if you think it would be better

PR #4744, opened 2023-09-15
1-1385: hide display of pattern info behind a flag

thomasheartman:
Previously, the front end would show info about the pattern if it
exists, regardless of whether the feature is active or not. The
pattern wouldn't be enforced, but it's confusing anyway, so let's hide
it.


kwasniew (approved):
It looks like there's unused import for useUiConfig

PR #4745, opened 2023-09-15
refactor: remove check for feature naming data object

thomasheartman:
We already check for its presence before doing anything. We don't need to
check it twice (we're not Santa Claus, after all).

### DIFF
--- a/src/lib/features/feature-naming-pattern/feature-naming-validation.ts
+++ b/src/lib/features/feature-naming-pattern/feature-naming-validation.ts
@@ -2,6 +2,18 @@ import { IFeatureNaming } from '../../types/model';
 
 const compileRegex = (pattern: string) => new RegExp(`^${pattern}$`);
 
+const disallowedStrings = [' ', '\\t', '\\s', '\\n', '\\r', '\\f', '\\v'];
+
+// Helper functions for error messages
+const whitespaceError = (pattern: string) =>
+    `Feature flag names cannot contain whitespace. You've provided a feature flag naming pattern that contains a whitespace character: "${pattern}". Remove any whitespace characters from your pattern.`;
+
+const exampleMismatchError = (example: string, pattern: string) =>
+    `You've provided a feature flag naming example ("${example}") that doesn't match your feature flag naming pattern ("${pattern}").`;
+
+const invalidValueError = (valueName: string) =>
+    `You've provided a feature flag naming ${valueName}, but no feature flag naming pattern. You must specify a pattern to use a ${valueName}.`;
+
 export const checkFeatureNamingData = (
     featureNaming: IFeatureNaming,
 ):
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-naming-pattern/feature-naming-validation.ts
```diff
@@ -3,53 +3,43 @@ import { IFeatureNaming } from '../../types/model';
 const compileRegex = (pattern: string) => new RegExp(`^${pattern}$`);
```
kwasniew, line 3: I played around with some minor refactoring. WDYT?
```ts
import { IFeatureNaming } from '../../types/model';

const compileRegex = (pattern: string) => new RegExp(`^${pattern}$`);
const disallowedStrings = [' ', '\\t', '\\s', '\\n', '\\r', '\\f', '\\v'];

// Helper functions for error messages
const whitespaceError = (pattern: string) =>
    `Feature flag names cannot contain whitespace. You've provided a feature flag naming pattern that contains a whitespace character: "${pattern}". Remove any whitespace characters from your pattern.`;

const exampleMismatchError = (example: string, pattern: string) =>
    `You've provided a feature flag naming example ("${example}") that doesn't match your feature flag naming pattern ("${pattern}").`;

const noPatternError = (type: string) =>
    `You've provided a feature flag naming ${type}, but no feature flag naming pattern. You must specify a pattern to use a ${type}.`;

export const checkFeatureNamingData = (featureNaming: IFeatureNaming) => {
    const { pattern, example, description } = featureNaming;
    const errors: string[] = [];

    if (disallowedStrings.some((str) => pattern?.includes(str))) {
        errors.push(whitespaceError(pattern as string));
    }

    if (pattern && example && !compileRegex(pattern).test(example)) {
        errors.push(exampleMismatchError(example, pattern));
    }

    if (!pattern && example) {
        errors.push(noPatternError("example"));
    }

    if (!pattern && description) {
        errors.push(noPatternError("description"));
    }

    return errors.length ? { state: 'invalid', reasons: errors } : { state: 'valid' };
};

export type FeatureNameCheckResult =
    | { state: 'valid' }
    | {
          state: 'invalid';
          invalidNames: Set<string>;
      };

export const checkFeatureFlagNamesAgainstPattern = (featureNames: string[], pattern?: string): FeatureNameCheckResult => {
    if (pattern) {
        const mismatchedNames = featureNames.filter((name) => !compileRegex(pattern).test(name));
        return mismatchedNames.length ? { state: 'invalid', invalidNames: new Set(mismatchedNames) } : { state: 'valid' };
    }
    return { state: 'valid' };
};

```

kwasniew (approved):
Love the Santa reference! :D

src/lib/features/feature-naming-pattern/feature-naming-validation.ts
thomasheartman: I like it! But it does have one issue: the first two checks in `checkFeatureNamingData` are mutually incompatible. But I can touch that up.

thomasheartman: Also, there's some typescript type issues with the return in the `checkFeatureNamingData` function.

Second, I'm not sure I agree about adding the extra ternary return in `checkFeatureFlagNamesAgainstPattern`, so I'll leave that off if you don't mind.

Implemented in 05f46f13d944a4520a1121b9455d5d0037adde40

PR #4747, opened 2023-09-15
fix: sort toggleNames before updating last seen

andreas-unleash:
Seems like when 2 pods are trying to POST lastSeen metrics, the db gets into a deadlock state. 

This is an attempt to fix the deadlock by sorting the toggleNames before the update.

The hypothesis is that sorted toggle names will reduce the chance of working on the same row at the same exact time

Closes # [1-1382](https://linear.app/unleash/issue/1-1382/order-data-before-updating-the-lastseen-to-reduce-change-of-deadlock)


sjaanus (approved):
Looks good, but I guess this is just theory. Also we need to release hotfix for 5.4.

PR #4749, opened 2023-09-15
feat: migration for dependent flags

kwasniew:
## About the changes

Adding dependencies column for strategies

### DIFF
--- a/src/migrations/20230915094540-dependent-flags.js
+++ b/src/migrations/20230915094540-dependent-flags.js
@@ -1,0 +1,90 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+          ALTER TABLE feature_strategies ADD COLUMN IF NOT EXISTS dependencies JSONB DEFAULT '[]'::jsonb NOT NULL;
+
+          CREATE VIEW features_view AS
+          SELECT
+              features.name as name,
+              features.description as description,
+              features.type as type,
+              features.project as project,
+              features.stale as stale,
+              features.impression_data as impression_data,
+              features.created_at as created_at,
+              features.archived_at as archived_at,
+              features.last_seen_at as last_seen_at,
+              feature_environments.last_seen_at as env_last_seen_at,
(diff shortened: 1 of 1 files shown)


src/migrations/20230915094540-dependent-flags.js
```diff
@@ -0,0 +1,90 @@
...
+              feature_strategies.id as strategy_id,
+              feature_strategies.strategy_name as strategy_name,
+              feature_strategies.parameters as parameters,
+              feature_strategies.constraints as constraints,
+              feature_strategies.sort_order as sort_order,
+              fss.segment_id as segments,
+              feature_strategies.title as strategy_title,
+              feature_strategies.disabled as strategy_disabled,
+              feature_strategies.variants as strategy_variants,
+              feature_strategies.dependencies as dependencies,
```
kwasniew, line 35: this line is added all the other code copied from: 20230802141830-add-feature-and-environment-last-seen-at-to-features-view

src/migrations/20230915094540-dependent-flags.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+          ALTER TABLE feature_strategies ADD COLUMN IF NOT EXISTS dependencies JSONB DEFAULT '[]'::jsonb NOT NULL;
```
ivarconr, line 6: I am not sure if we want the dependencies on the strategies themselves. Is that what you plan to do here?

kwasniew: this is still a draft pr (updated the status to avoid confusion). I'm experimenting with different models and this one can be either on the strategy or feature_environment or feature. We can have a chat on Monday.

kwasniew: BTW I have full working e2e spike on my machine I can demo to you next week. With dependencies on strategies you can express: if parent flag resolved to variantA resolve child flag to variantB, if parent flag resolved to variant C resolve child flag to variant D.

PR #4751, opened 2023-09-15
fix: sort toggleNames before updating last seen (#4747)

andreas-unleash:
Seems like when 2 pods are trying to POST lastSeen metrics, the db gets into a deadlock state.

This is an attempt to fix the deadlock by sorting the toggleNames before the update.

The hypothesis is that sorted toggle names will reduce the chance of working on the same row at the same exact time

Closes #
[1-1382](https://linear.app/unleash/issue/1-1382/order-data-before-updating-the-lastseen-to-reduce-change-of-deadlock)

(cherry picked from commit 76a834ca91ca821d7fa1a0fb1794754a9767c8a9)

## About the changes

Closes #


sjaanus (approved):

andreas-unleash:
Resolved in a different way - closing

PR #4752, opened 2023-09-15
feat: implement optional json payload and template

daveleek:
## About the changes

Adds optional support for specifying JSON templates for datadog message payload

image

### Important files

`frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterEnableWithDropdown.tsx` - a new component comprising of a text field and a dropdown menu
`src/lib/addons/datadog.ts` - Where the integration is taking place

## Discussion points

- Should I have implemented the new component type as a specifiable addon parameter type in definitions? Felt a bit YAGNI/Premature
- Would like input on naming and the new component etc


frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameter.tsx
```diff
@@ -28,41 +18,34 @@ export const IntegrationParameter = ({
...
 }: IIntegrationParameterProps) => {
-    const value = config.parameters[definition?.name] || '';
-    const type = resolveType(
-        definition,
-        typeof value === 'string' ? value : ''
-    );
-    const error = parametersErrors[definition.name];
+    if (config.provider === 'datadog' && definition.name === 'bodyTemplate') {
+        const datadogJson = useUiFlag('datadogJsonTemplate');
+        console.log(`Datadog json is ${datadogJson}`);
```
sjaanus, line 23: Is this needed?

daveleek: Nope! THANK YOU

src/lib/addons/datadog.ts
```diff
@@ -26,12 +30,18 @@ interface DDRequestBody {
 export default class DatadogAddon extends Addon {
     private msgFormatter: FeatureEventFormatter;
 
-    constructor(config: IAddonConfig) {
+    private flagResolver: IFlagResolver;
+
+    constructor(
+        config: IAddonConfig,
+        { flagResolver }: Pick<IUnleashConfig, 'flagResolver'>,
```
nunogois, line 37: Maybe you have a reason but this feels a bit convoluted. In order to better express the DatadogAddon dependencies I think I would extend `IAddonConfig` to include a flag resolver.

```ts
interface IDatadogAddonConfig extends IAddonConfig {
  flagResolver: IFlagResolver;
}

// ...

this.flagResolver = config.flagResolver
```

Alternatively, simply accept an IFlagResolver here:

```ts
constructor(config: IAddonConfig, flagResolver: IFlagResolver)
```

src/lib/addons/index.ts
```diff
@@ -29,7 +29,7 @@ export const getAddons: (args: {
         new Webhook({ getLogger }),
         slackAddon,
         new TeamsAddon({ getLogger, unleashUrl }),
-        new DatadogAddon({ getLogger, unleashUrl }),
+        new DatadogAddon({ getLogger, unleashUrl }, { flagResolver }),
```
nunogois, line 32: I think you can simplify this with https://github.com/Unleash/unleash/pull/4752/files#r1328576551

src/lib/addons/datadog-definition.ts
```diff
@@ -62,6 +62,21 @@ const dataDogDefinition: IAddonDefinition = {
...
             type: 'textfield',
         },
+        {
+            name: 'bodyTemplate',
+            displayName: 'Body template',
+            placeholder: `{
+"event": "{{event.type}}",
+"createdBy": "{{event.createdBy}}",
+"featureToggle": "{{event.data.name}}",
+"timestamp": "{{event.data.createdAt}}"
```
nunogois, line 72: You can add 2 spaces for each line here to display the value with indentation, like in `webhook-definition`.

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterTextField.tsx
```diff
@@ -0,0 +1,70 @@
...
+export interface IIntegrationParameterTextFieldProps {
+    parametersErrors: Record<string, string>;
+    definition: AddonParameterSchema;
+    setParameterValue: (param: string) => ChangeEventHandler<HTMLInputElement>;
+    config: AddonSchema;
+}
+
+const StyledTextField = styled(TextField)(({}) => ({
+    width: '100%',
+})) as typeof TextField;
```
nunogois, line 27: I think this can be simplified as:

```ts
const StyledTextField = styled(TextField)({
    width: '100%',
});
```

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameter.tsx
```diff
@@ -28,41 +18,32 @@ export const IntegrationParameter = ({
...
+
+        return (
+            <IntegrationParameterEnableWithDropdown
+                config={config}
+                definition={definition}
+                parametersErrors={parametersErrors}
+                setParameterValue={setParameterValue}
+                enableTitle="Data format"
+                disableOptionText="Standard"
+                enableOptionText="Custom template (JSON)"
```
nunogois, line 35: This feels too static. If we want to go with this route, perhaps these labels should be parameterized as well in the addon definition?

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterEnableWithDropdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+}: IIntegrationParameterEnableWithDropdownProps) => {
+    const value = config.parameters[definition?.name] || '';
+    const renderProjectOptions = () => [
+        { key: 'disabled', label: disableOptionText },
+        { key: 'enabled', label: enableOptionText },
+    ];
+
+    const enabledFromStart = value !== '';
+    const [enabled, setEnabled] = useState(
+        enabledFromStart ? 'enabled' : 'disabled'
```
nunogois, line 36: I feel like this should be a boolean. Something like:

```ts
const [enabled, setEnabled] = useState(Boolean(value))
```

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterEnableWithDropdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const onChange = (key: string) => {
+        setEnabled(key);
+        if (key === 'disabled') {
+            const fakeEvent = {
+                preventDefault: () => {},
+                target: {
+                    value: '',
+                },
+            };
+            setParameterValue(definition.name)(fakeEvent as any);
```
nunogois, line 48: IMO it's kinda crazy that we need to do this, I think `setParameterValue` is due for a refactor if we want to go this route.

Either way, you can use a better type here than `any`: `ChangeEvent<HTMLInputElement>`

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterEnableWithDropdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+    setParameterValue,
+    enableTitle,
+    enableOptionText,
+    disableOptionText,
+}: IIntegrationParameterEnableWithDropdownProps) => {
+    const value = config.parameters[definition?.name] || '';
+    const renderProjectOptions = () => [
+        { key: 'disabled', label: disableOptionText },
+        { key: 'enabled', label: enableOptionText },
+    ];
```
nunogois, line 32: Any reason why this is a function instead of simply an array?
 
```ts
const renderProjectOptions = [
  { key: 'disabled', label: disableOptionText },
  { key: 'enabled', label: enableOptionText },
];
```

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterEnableWithDropdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+            setParameterValue(definition.name)(fakeEvent as any);
+        }
+    };
+
+    return (
+        <>
+            <StyledAddonParameterContainer>
+                <GeneralSelect
+                    label={enableTitle}
+                    options={renderProjectOptions()}
```
nunogois, line 57: Can be simplified to `options={options}` after https://github.com/Unleash/unleash/pull/4752/files#r1328605778

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterEnableWithDropdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    value={enabled}
```
nunogois, line 58: Oh, maybe this is the reason why we are using a string instead of boolean in https://github.com/Unleash/unleash/pull/4752/files#r1328594031 - Gotcha

nunogois (commented):
My general thought is that this feels slightly overkill. If the body template is optional, do we really need the dropdown that shows/hides it? Because it feels like we could just leave it there, similar to what we do for the webhook addon. We could still hide it behind the feature flag if we wanted.

I feel like this would simplify this PR and feature drastically.

In the other hand, if we really want this dropdown visibility behavior, perhaps it should be a different field type, other than textfield?

Other than that this feels fine, but I would still like us to think a bit about this before merging.

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterEnableWithDropdown.tsx
daveleek: that's a good point!

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameterEnableWithDropdown.tsx
daveleek: yup :)

src/lib/addons/datadog.ts
daveleek: Went with your first suggestion, thanks!

frontend/src/component/integrations/IntegrationForm/IntegrationParameters/IntegrationParameter/IntegrationParameter.tsx
daveleek: As per discussions I've removed this part to simplify the UI

nunogois (approved):
LGTM

PR #4756, opened 2023-09-15
fix: simplify channels logic in slack app integration

nunogois:
https://linear.app/unleash/issue/2-1393/drop-the-always-post-to-default-channels-field

This drops the "Always post to default channels" field in the Slack App integration in favor of always posting to the configured channels. This should simplify the configuration of this integration.

Here's a breakdown of the logic with this change:
 - Always post to the configured Slack channels, regardless of tags;
 - Tags are still respected. E.g. if we have a configured channel "channel-1" and a tag for "channel-2", then we post to both channels;
 - As channels are optional, if you would like to skip default channels for certain events and handle everything through tags, you can just create a new configuration without any default channels;

This also updates the labels and changes the tests to better reflect the intended behavior.

image

### DIFF
--- a/src/lib/addons/slack-app-definition.ts
+++ b/src/lib/addons/slack-app-definition.ts
@@ -51,17 +51,9 @@ const slackAppDefinition: IAddonDefinition = {
         },
         {
             name: 'defaultChannels',
-            displayName: 'Default channels',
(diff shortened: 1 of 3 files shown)


gastonfournier (approved):
LGTM!

PR #4762, opened 2023-09-18
fix: prevent blur when selecting text

kwasniew:
## About the changes

Selecting text in search suggestions was triggering blur and hiding all the suggestions:
Screenshot 2023-09-18 at 13 51 07

### DIFF
--- a/frontend/src/component/common/Search/Search.tsx
+++ b/frontend/src/component/common/Search/Search.tsx
@@ -114,7 +114,6 @@ export const Search = ({
     useKeyboardShortcut({ key: 'Escape' }, () => {
         if (searchContainerRef.current?.contains(document.activeElement)) {
             searchInputRef.current?.blur();
-            hideSuggestions();
         }
     });
     const placeholder = `${customPlaceholder ?? 'Search'} (${hotkey})`;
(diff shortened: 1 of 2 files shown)


frontend/src/component/common/Search/Search.tsx
```diff
@@ -114,7 +114,6 @@ export const Search = ({
     useKeyboardShortcut({ key: 'Escape' }, () => {
         if (searchContainerRef.current?.contains(document.activeElement)) {
             searchInputRef.current?.blur();
-            hideSuggestions();
```
kwasniew, line 117: blur effect handler will handle it automatically

frontend/src/hooks/useOnBlur.ts
```diff
@@ -5,24 +5,35 @@ export const useOnBlur = (
...
+
+        const handleMouseDown = (event: MouseEvent) => {
+            mouseDownInside = containerRef.current?.contains(event.target as Node) || false;
+        };
+
         const handleBlur = (event: FocusEvent) => {
-            // setTimeout is used because activeElement might not immediately be the new focused element after a blur event
             setTimeout(() => {
                 if (
+                    !mouseDownInside &&
```
kwasniew, line 17: mouse down inside suggestions should not close them

FredrikOseberg (approved):
LGTM

PR #4765, opened 2023-09-18
chore: GA (remove flag) for Slack App integration

nunogois:
https://linear.app/unleash/issue/2-1405/remove-slackappaddon-feature-flag-and-make-this-ga

GA's the new Slack App integration by removing the feature flag 🚀


src/lib/addons/addon-schema.ts
```diff
@@ -9,7 +9,7 @@ export const addonDefinitionSchema = joi.object().keys({
     documentationUrl: joi.string().uri({ scheme: [/https?/] }),
     description: joi.string().allow(''),
     howTo: joi.string().optional().allow(''),
-    deprecated: joi.boolean().optional().default(false),
+    deprecated: joi.string().optional().allow(''),
```
gastonfournier, line 12: Is this change needed? Sounds like deprecated should be a boolean. In OpenAPI there's `deprecated` + `deprecatedDescription` IIRC

gastonfournier (commented):
Just a minor comment maybe, but I'm curious about the change

src/lib/addons/addon-schema.ts
nunogois: This fixes and adapts the addon schema to how the `deprecated` property really works right now: It acts as both the boolean that tells us whether an addon/integration is deprecated and the message why.

This feels pretty clear to me, as you won't specify a `deprecation` property for non-deprecated addons. But if you feel it's clearer to have 2 separate properties for example:
 - `deprecated` - Boolean
 - `deprecatedDescription` - String

I can also do the change.

WDYT?

daveleek: Chiming in that I like the approach with two properties - one for a boolean and the other with a description

gastonfournier: I don't mind either way, it's clear to me if you say `deprecated: "reason"` that the feature is deprecated (so no need to split). 

But I didn't spot anything in the PR that was deprecating something with a string, and therefore these changes seem unrelated to the PR. Also, if unrelated, why are we not getting compilation errors? Maybe because of javascript?

nunogois: This is where you can find it: https://github.com/Unleash/unleash/pull/4765/files#diff-b224eb9a2333908c270924576587147061c787791521514cffe3f275b3308871R26

gastonfournier: Ok, and because this is the first time we deprecate an addon, this is still greenfield and it's not something we expose. https://docs.getunleash.io/reference/api/unleash/update-addon

PR #4766, opened 2023-09-18
feat: simpler integration filters

nunogois:
https://linear.app/unleash/issue/2-1407/remove-the-all-checkboxes-from-project-and-environment-filters

Simplifies integration event filters by removing the "ALL" checkboxes from these components. Whether you opted to check the "ALL" checkbox, or not to filter at all, the result is the same - The selected options would act as a filter.

Includes some refactoring and clean up.


frontend/src/component/integrations/IntegrationForm/IntegrationMultiSelector/IntegrationMultiSelector.tsx
```diff
@@ -51,45 +42,30 @@ export const IntegrationMultiSelector: VFC<IIntegrationMultiSelectorProps> = ({
...
     description,
+    note,
     required,
 }) => {
     const renderInput = (params: AutocompleteRenderInputParams) => (
         <TextField
             {...params}
             error={Boolean(error)}
-            helperText={error}
+            helperText={error || note}
```
gastonfournier, line 53: If this is error or note, maybe we should have a single text and another property of type `style: 'error'  || 'note'`? 
Personally, I don't like positional parameters having different meanings, and they seem to be mutually exclusive (i.e. when error is undefined you can have a note, when there's an error, it takes precedence over the note, so they can't be rendered together...

gastonfournier (commented):
Happy to give it an approval tomorrow after looking at the comment. I'm not approving right now because of the auto-merge

gastonfournier (approved):

PR #4767, opened 2023-09-18
feat: move middleware to enterprise

sjaanus:
1. Moved middleware to enterprise
2. Admin tokens are ID of -1, also give respect to that when returning projects.


src/lib/features/private-project/privateProjectStore.ts
```diff
@@ -15,26 +15,29 @@ class PrivateProjectStore implements IPrivateProjectStore {
     destroy(): void {}
 
     async getUserAccessibleProjects(userId: number): Promise<string[]> {
-        const isNotViewer = await this.db('role_user')
+        // -1 user is ADMIN api token
+        if (userId === -1) {
```
FredrikOseberg, line 19: Could we change this to something more semantic? IE: `const ADMIN_TOKEN = -1`

FredrikOseberg (approved):
Approved with minor comment

PR #4771, opened 2023-09-19
refactor: feature oriented architecture for feature dependencies

kwasniew:
## About the changes

Since feature toggle controller and service are already too big I decided to split feature dependencies write model into a separate directory. This change only moves things around, no new code is added.


FredrikOseberg (approved):
LGTM

PR #4772, opened 2023-09-19
feat: Persist dependent features

kwasniew:
## About the changes

Persisting dependent features toggles.

Screenshot 2023-09-19 at 12 03 42

parent name, child name, whether parent is enabled or disabled and parent variants that we want to match on

## Discussion points

Database Table Design Decisions:

* Both 'parent' and 'child' are set up as foreign key relationships. This allows us to easily display these relationships in the admin user interface and prevents invalid bindings.

* While the 'variants' could technically use a foreign key relationship to enforce valid values, it might be overcomplicated. Given that both strategy variants and feature variants are also stored as JSONB, introducing such constraints could be redundant. Moreover, establishing foreign keys to multiple tables presents its challenges. Using JSONB for 'variants' also provides flexibility for potential operator support in the future.

* I've decided to address the comprehensive feature-with-dependencies read model/view in a separate PR. This is because introducing it would be a more intricate and potentially riskier change compared to merely saving dependencies.


src/migrations/20230919104006-dependent-features.js
```diff
@@ -0,0 +1,28 @@
...
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS dependent_features
+            (
+                parent     varchar(255) NOT NULL,
+                child      varchar(255) NOT NULL,
+                enabled    boolean DEFAULT true NOT NULL,
+                variants   JSONB DEFAULT '[]'::jsonb NOT NULL,
+                PRIMARY KEY (parent, child),
+                FOREIGN KEY (parent) REFERENCES features (name) ON DELETE RESTRICT,
```
kwasniew, line 13: should not be able to delete parent flag when child flags still exist

src/migrations/20230919104006-dependent-features.js
```diff
@@ -0,0 +1,28 @@
...
+                FOREIGN KEY (child) REFERENCES features (name) ON DELETE CASCADE
```
kwasniew, line 14: deleting a child flag should remove the dependency

src/lib/features/dependent-features/dependent-feature-store.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    async upsert(featureDependency: FeatureDependency): Promise<void> {
+        const serializableFeatureDependency: SerializableFeatureDependency = {
+            parent: featureDependency.parent,
+            child: featureDependency.child,
+            enabled: featureDependency.enabled,
+        };
+        if ('variants' in featureDependency) {
+            serializableFeatureDependency.variants = JSON.stringify(
```
kwasniew, line 21: arrays need to be stringified

src/lib/features/dependent-features/dependent-feature-store.ts
```diff
@@ -0,0 +1,30 @@
...
+                featureDependency.variants,
+            );
+        }
+        await this.db('dependent_features')
+            .insert(serializableFeatureDependency)
+            .onConflict(['parent', 'child'])
```
kwasniew, line 27: this gives us upsert semantics

FredrikOseberg (approved):
LGTM. I think it makes sense to use foreign keys for this.

PR #4775, opened 2023-09-19
fix: decouple metrics from ofter-read tables

andreas-unleash:
Decouples persisting `last_seen_at` from `feature_environments` and `features` tables.

This comes about recent occasional deadlock errors we have been seeing.  

Create a `feature_environments_metrics` table 
Copy all the existing data from `feature_environments`
Rebuild the feature view reading from the new table
Restructure `getFeatureOverview` query

Removes the `last_seen_at` column from queries unless it is for the `admin`
No UI changes needed


src/migrations/20230919103955-create-feature-environments-metrics-table.js
```diff
@@ -0,0 +1,110 @@
...
+               feature_strategies.id             as strategy_id,
+               feature_strategies.strategy_name  as strategy_name,
+               feature_strategies.parameters     as parameters,
+               feature_strategies.constraints    as constraints,
+               feature_strategies.sort_order     as sort_order,
+               fss.segment_id                    as segments,
+               feature_strategies.title          as strategy_title,
+               feature_strategies.disabled       as strategy_disabled,
+               feature_strategies.variants       as strategy_variants
+        FROM features
```
FredrikOseberg, line 54: What is all this needed for?

andreas-unleash: drops and recreates the featureView (I found out the hard way that this is much easier than modifying_ - changing where the env_last_seen is being read.

andreas-unleash: ``` feature_environments_metrics.last_seen_at as env_last_seen_at,```

src/migrations/20230919103955-create-feature-environments-metrics-table.js
```diff
@@ -0,0 +1,110 @@
...
+            feature_strategies.variants as strategy_variants
+        FROM
+            features
+                LEFT JOIN feature_environments ON feature_environments.feature_name = features.name
+                LEFT JOIN feature_strategies ON feature_strategies.feature_name = feature_environments.feature_name
+                and feature_strategies.environment = feature_environments.environment
+                LEFT JOIN environments ON feature_environments.environment = environments.name
+                LEFT JOIN feature_strategy_segment as fss ON fss.feature_strategy_id = feature_strategies.id;
+
+          DROP TABLE IF EXISTS feature_environments_metrics
```
kwasniew, line 106: missing ; at the end of line

src/lib/db/feature-toggle-store.ts
```diff
@@ -171,34 +173,22 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
-                    );
+        const enhancedData = data.map((value) => ({
+            feature_name: value.featureName,
+            last_seen_at: now,
+            environment: value.environment,
+        }));
+        const dataToPersist: typeof enhancedData = [];
+        for (const input of enhancedData) {
+            if (await this.exists(input.feature_name)) {
+                dataToPersist.push(input);
```
andreas-unleash, line 184: This is not ideal.  I would much rather have this in the SQL query somehow

src/lib/db/feature-toggle-store.ts
```diff
@@ -171,34 +173,22 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
             }
+        }
+        try {
+            await this.db(FEATURE_ENVIRONMENTS_METRICS_TABLE)
+                .insert(dataToPersist)
+                .onConflict(['feature_name', 'environment'])
+                .merge(['last_seen_at']);
```
andreas-unleash, line 191: essentially upsert - could not find a way to satisfy the constraint that the `name` exists in `features` table.  Because I am using onConflict to merge the last_seen_at column with the new value, I added the `EXISTS` check above.  The logic(assumption) is that the EXISTS operation in one of the least expensive operations.

src/test/e2e/services/last-seen-service.e2e.test.ts
```diff
@@ -129,3 +129,46 @@ test('Should not update anything for 0 toggles', async () => {
 
     service.destroy();
 });
+
+test('Should handle 1000 toggle updates', async () => {
```
andreas-unleash, line 133: ideally I would like to simulate the scenario that caused the deadlock

src/lib/db/feature-toggle-store.ts
```diff
@@ -482,6 +474,21 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
 
         return result?.potentially_stale ?? false;
     }
+
+    private columnsWithMetrics = () => {
```
sjaanus, line 478: Do we need metrics on all of these store methods? This is quite expensive, making a new database call for each feature.

andreas-unleash: I tried to add the column only for UI - admin - where do you think it is not needed?

src/lib/db/feature-toggle-store.ts
sjaanus: But you are doing exist for each datapoint, which means 10s or 100s of requests and roundtrips. Doesn't the onconflict/merge combo work without doing exists?

andreas-unleash: It works for the conflict on the primary key - but not on the constraint to feature name of the features table.

andreas-unleash: So if we try to persist metrics on a feature that does not exist  - the query fails

andreas-unleash: I can try to bake the ON CONFLICT ON CONSTRAINT DO NOTHING to the table - if that works it would eliminate the need on all the EXISTS queries.

src/lib/db/feature-toggle-store.ts
ivarconr: we cannot do this. For customers with thousands of toggles it will overload the DB.

src/lib/db/feature-toggle-store.ts
ivarconr: why not skip on conflict?

src/migrations/20230919103955-create-feature-environments-metrics-table.js
```diff
@@ -0,0 +1,119 @@
...
+               feature_strategies.sort_order     as sort_order,
+               fss.segment_id                    as segments,
+               feature_strategies.title          as strategy_title,
+               feature_strategies.disabled       as strategy_disabled,
+               feature_strategies.variants       as strategy_variants
+        FROM features
+                 LEFT JOIN feature_environments ON feature_environments.feature_name = features.name
+                 LEFT JOIN feature_strategies ON feature_strategies.feature_name = feature_environments.feature_name
+                   AND feature_strategies.environment = feature_environments.environment
+                 LEFT JOIN feature_environments_metrics ON feature_environments_metrics.feature_name = feature_environments.feature_name
```
ivarconr, line 65: do we really need this? The UI only need the max per flag across all environments and more details can be loaded lazily when use clicks in the UI? Not sure about the cost here.

src/lib/db/feature-toggle-store.ts
ivarconr: because of the foreign key for feature_name, I think this can be dealt with in the conflict handling of the insert query itself.

src/lib/db/feature-toggle-store.ts
```diff
@@ -171,34 +173,40 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
-                        this.db(TABLE)
-                            .select('name')
-                            .whereIn('name', toggleNames)
-                            .forUpdate()
-                            .skipLocked(),
+                        enhancedData.map((entry) => entry.feature_name),
                     );
-            }
+
+                // Insert only the rows that exist in the CTE (subquery)
```
andreas-unleash, line 192: If the features table is only used in the CTE subquery, the likelihood of deadlocks occurring directly due to this table is significantly reduced.

src/lib/db/feature-toggle-store.ts
andreas-unleash: unfortunately not - we can either have ON CONFLICT DO NOTHING (which is needed for the constraint violation to `features` or ON CONFLICT (pkey) DO UPDATE (which is needed for upsert)

src/lib/db/feature-toggle-store.ts
```diff
@@ -171,34 +173,40 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
+            environment: value.environment,
+        }));
 
-                // Updating the toggle's last_seen_at also for backwards compatibility
-                await this.db(TABLE)
-                    .update({ last_seen_at: now })
+        try {
+            await this.db.transaction(async (trx) => {
+                // Create a CTE with the data to be inserted
+                const cte = trx('features')
```
ivarconr, line 185: Is this adding a transaction on the features table?

src/migrations/20230919103955-create-feature-environments-metrics-table.js
```diff
@@ -0,0 +1,119 @@
...
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS feature_environments_metrics(
+           environment  varchar(100) not null
+               references environments(name)
+                   ON DELETE CASCADE,
+           feature_name varchar(255) not null
+               references features(name)
+                   ON DELETE CASCADE,
+           last_seen_at timestamp with time zone,
```
ivarconr, line 13: could have a default to now().

src/migrations/20230919103955-create-feature-environments-metrics-table.js
```diff
@@ -0,0 +1,119 @@
+'use strict';
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS feature_environments_metrics(
```
ivarconr, line 6: I am thinking that having foreign keys on feature_name and environment actually helps us. We could instead have something that cleans up unknown flags and environments every now and then. By dropping the constraints the insert operation would not need to validate against other tables.

This is also the approach we took for `client_metrics_env`.

src/lib/db/feature-toggle-store.ts
```diff
@@ -171,34 +173,40 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
-                            .select('name')
-                            .whereIn('name', toggleNames)
-                            .forUpdate()
-                            .skipLocked(),
+                        enhancedData.map((entry) => entry.feature_name),
                     );
-            }
+
+                // Insert only the rows that exist in the CTE (subquery)
+                await trx.raw(
```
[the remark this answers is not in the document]
ivarconr: this will crash if the environment does not exists also, because of the foreign key.

src/migrations/20230919103955-create-feature-environments-metrics-table.js
ivarconr: I also question the name. This is usage metrics and not feature toggle configurations. It is not an obvious difference between `client_metrics_env` and `feature_environments_metrics`

src/lib/db/feature-toggle-store.ts
```diff
@@ -171,34 +173,40 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
 
     async setLastSeen(data: LastSeenInput[]): Promise<void> {
```
ivarconr, line 174: We should move the setLastSeen out of this store. We already have a separate store updating the `client_metrics_env` which is mostly the same challenges (`ClientMetricsStoreV2`)

src/lib/db/feature-toggle-store.ts
andreas-unleash: When I asked ChatGPT 
The features table is read very often - how can I mitigate deadlocks for transactions with more than 1000 rows
`If the features table is only used in the CTE subquery and is not being updated, the likelihood of deadlocks occurring directly due to this table is significantly reduced. In such cases, deadlocks are more likely to be related to the FEATURE_ENVIRONMENTS_METRICS_TABLE, where you are performing updates.`

src/migrations/20230919103955-create-feature-environments-metrics-table.js
andreas-unleash: This is a much better approach I agree.  I did think about it too - but was held back when I saw that we had tests checking exactly this: that we wouldn't store metrics for unknown features.

andreas-unleash: dropping the constraints would solve most problems - we can still create an index on feature_name so we can easily query it when we need it

src/lib/db/feature-toggle-store.ts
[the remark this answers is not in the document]
andreas-unleash: I am trying this out - bard is messing up a bit :)

andreas-unleash:
Went a different direction - closing

PR #4777, opened 2023-09-19
feat: dependent features client api

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/db/feature-toggle-client-store.ts
+++ b/src/lib/db/feature-toggle-client-store.ts
@@ -67,6 +67,8 @@ export default class FeatureToggleClientStore
         const isPlayground = requestType === 'playground';
         const environment = featureQuery?.environment || DEFAULT_ENV;
         const stopTimer = this.timer('getFeatureAdmin');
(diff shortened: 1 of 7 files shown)


src/lib/db/feature-toggle-client-store.ts
```diff
@@ -195,6 +201,14 @@ export default class FeatureToggleClientStore
             ) {
                 this.addSegmentIdsToStrategy(feature, r);
             }
+            if (r.parent && !isAdmin && dependentFeaturesEnabled) {
```
kwasniew, line 204: admin api doesn't need it for now + added a flag guard here

src/lib/features/dependent-features/dependent-features-service.ts
```diff
@@ -17,20 +17,20 @@ export class DependentFeaturesService {
     }
 
     async upsertFeatureDependency(
-        parentFeature: string,
+        childFeature: string,
```
kwasniew, line 20: there was a mistake in the previous version. child should have a dependency on parent (dependent feature)

PR #4784, opened 2023-09-20
feat: add json editor

spgandhi:
## About the changes

In this PR we are adding a JSON Editor that will be used for `json` variant payload type. Its using `vanilla-jsoneditor` package https://www.npmjs.com/package/vanilla-jsoneditor and is lazy loaded only when the `json` payload type is selected in the dropdown UI. 

Feature preview - https://recordit.co/p4Px9DNUIC

Closes #4450

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import React, { Suspense, useEffect, useState } from 'react';
 import Input from 'component/common/Input/Input';
 import { HelpIcon } from 'component/common/HelpIcon/HelpIcon';
 import SelectMenu from 'component/common/select';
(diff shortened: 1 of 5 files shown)


frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
```diff
@@ -441,34 +443,50 @@ export const VariantForm = ({
...
+                            }
+                            value={payload.value}
+                            onChange={e => {
+                                clearError(ErrorField.PAYLOAD);
+                                setPayload(payload => ({
+                                    ...payload,
+                                    value: e.target.value,
+                                }));
+                            }}
+                            placeholder={''}
```
spgandhi, line 484: Here we have removed the JSON placeholder because this block won't be used for JSON payload anymore

FredrikOseberg, line 484: We use a ConditionallyRender component to help with readability for conditional renders. Could be used here.

FredrikOseberg (commented):
This is looking good @spgandhi. Could you also make it match our theme colors in the dark mode? Currently it looks like this which isn't ideal: 

Skjermbilde 2023-09-26 kl  09 29 45

frontend/src/component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsModal/VariantForm/VariantForm.tsx
spgandhi: Done

spgandhi:
@FredrikOseberg made some changes to make the editor more theme compatible. For the hover and focus effect, I see that for inputs, the border width increases from 1px -> 2px on focus. For the JSON Editor, doing this causes the width to change and so looks shaky.So for now I have only updating the colors but not the width

Screenshot 2023-09-26 at 11 55 01 PM

Screenshot 2023-09-26 at 11 54 55 PM

PR #4785, opened 2023-09-20
Initial commit- 10 principles of feature flags

ardeche07:
## About the changes

This is the first commit of the 10 principles of feature flag systems.  I'm moving over pages edits are finalized.

I need help getting these pages into the navigation.

The IA should be:

Topic Guides
   - Feature Flags
      - Principle 1
      - Principle 2
      - Principle 3
      - etc

### DIFF
--- a/website/docs/topics/feature-flags.md
+++ b/website/docs/topics/feature-flags.md
@@ -1,0 +1,32 @@
+## 10 Principles for building and scaling feature flag systems
+
+Feature flags, sometimes called feature toggles or feature switches, are a software development technique that allows engineering teams to decouple the release of new functionality from software deployments. With feature flags, developers can turn specific features or code segments on or off at runtime, without the need for a code deployment or rollback.d Organizations who adopt feature flags see improvements in all key operational metrics for DevOps: Lead time to changes, mean-time-to-recovery, deployment frequency, and change failure rate.
+
+There are 10 principles for building a large-scale feature flag system. These principles have their roots in distributed systems architecture and pay particular attention to security, privacy, and scale that is required by most enterprise systems. If you follow these principles, your feature flag system is less likely to break under load and will be easier to evolve and maintain.
+
+These principles are:
+
+1. Enable run-time control. Control flags dynamically, not using config files.
+2. Design for failure. Favor availability over consistency.
+3. Use unique names across all applications. Enforce naming conventions.
+4. Evaluate locally. Evaluate flags as close to the user as possible.	
+5. Make feature flags short-lived. Do not confuse flags with application configuration.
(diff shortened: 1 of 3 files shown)


website/docs/topics/feature-flags.md
```diff
@@ -0,0 +1,32 @@
+## 10 Principles for building and scaling feature flag systems
```
thomasheartman, line 1:
```suggestion
---
title: 10 Principles for building and scaling feature flag systems
---
```

To stay consistent with how we do it in the rest of the docs, we should use a YAML annotation for the title. 

We **could** use just a markdown title (but in that case it should be just a single `#`), but I think we should try to be consistent here.

website/docs/topics/feature-flags.md
```diff
@@ -0,0 +1,32 @@
...
+3. Use unique names across all applications. Enforce naming conventions.
+4. Evaluate locally. Evaluate flags as close to the user as possible.	
+5. Make feature flags short-lived. Do not confuse flags with application configuration.
+6. Scale horizontally. Decouple writing and reading flags.
+7. Limit payloads. Feature flag payload should be as small as possible.
+8. Choose open by default. Democratize feature flag access.	
+9. Do no harm. Prioritize consistent user experience.
+10. Enable traceability. Make it easy to understand flag evaluation.
+
+#### Background
```
thomasheartman, line 20: Headings should always follow in logical order. 1 -> 2 -> 3 -> 4. Skipping from 1 or 2 to 4 causes issue with the hierarchy.

```suggestion
## Background
```

website/docs/topics/feature-flags/feature-flag-availability-over-consistency.md
```diff
@@ -0,0 +1,19 @@
+## 2. Design for failure. Favor availability over consistency.
```
thomasheartman, line 1: This might have to be quoted because it contains full stops. It might be alright, but it's safer with quotes. (The worst thing is that the build will fail with some inscrutable error message).

```suggestion
---
title: "2. Design for failure. Favor availability over consistency."
---
```

website/docs/topics/feature-flags/feature-flag-run-time-control.md
```diff
@@ -0,0 +1,21 @@
+## Enable run-time control. Control flags dynamically, not using config files.
```
thomasheartman, line 1:
```suggestion
---
title: "Enable run-time control. Control flags dynamically, not using config files."
---
```

thomasheartman (commented):
Nice work! Just added a few small corrections on the structure level of the docs to ensure they comply with the rest of the docs 😄

PR #4786, opened 2023-09-20
feat: make application usage private through project

sjaanus:


src/lib/services/client-metrics/instance-service.ts
```diff
@@ -162,8 +174,27 @@ export default class ClientInstanceService {
 
     async getApplications(
         query: IApplicationQuery,
+        userId: number,
     ): Promise<IClientApplication[]> {
-        return this.clientApplicationsStore.getAppsForStrategy(query);
+        const applications =
+            await this.clientApplicationsStore.getAppsForStrategy(query);
+        if (this.flagResolver.isEnabled('privateProjects') && userId) {
```
FredrikOseberg, line 181: Can userId be 0?

FredrikOseberg (changes requested):
Looks good, but are we missing a test for this?

src/lib/services/client-metrics/instance-service.ts
sjaanus: That is good question. So far I have discovered we can have -1 userId, but have not seen 0.

FredrikOseberg (approved):
LGTM

PR #4791, opened 2023-09-20
feat: private projects handle in playground

sjaanus:


src/lib/features/playground/playground-service.ts
```diff
@@ -66,27 +67,58 @@ export class PlaygroundService {
...
+                await this.privateProjectChecker.getUserAccessibleProjects(
+                    userId,
+                );
+            filteredProjects =
+                projects === ALL
+                    ? accessibleProjects
+                    : projects.filter((project) =>
+                          accessibleProjects.includes(project),
+                      );
+            console.log(accessibleProjects);
```
kwasniew, line 115: leftover

src/lib/features/playground/advanced-playground.test.ts
```diff
@@ -19,6 +19,7 @@ beforeAll(async () => {
                     advancedPlayground: true,
                     strictSchemaValidation: true,
                     strategyVariant: true,
+                    privateProjects: true,
```
kwasniew, line 22: do we have a test showing that private projects are not showing up?

kwasniew (approved):

Issue #4380, opened 2023-08-01
🔥 Unleash v6


ivarconr:
**Decided:**
-

- [x]  Upgrade to Node v20 @chriswk 
      - #7095
- [x] Dropping support for postgres v10, v11, and v12. 
  - Make unleash fail to start with a error message if you try to connect to an old postgres version. This to future proof our ability to take advantage of v13+ capabilites. https://www.postgresql.org/support/versioning/ @chriswk 
   - #7139 
- [x] Remove the passport libs from the official Open Source Docker distribution. (We have a [community image](https://github.com/Unleash/unleash-docker-community) which adds these dependencies). This is huge, because it allow us to drop the "docker" folder with added complexity to avoid those dependencies leaking in to the enterprise version.  @chriswk 
    - https://github.com/Unleash/unleash/pull/7159
    - https://github.com/Unleash/helm-charts/pull/148
    - https://github.com/Unleash/unleash-docker-community/pull/148
- [x] Remove `/edge/metrics` endpoint @chriswk
   - https://github.com/Unleash/unleash/pull/7161
- [x] Deprecate custom strategies (Notice in the UI, docs and the API specification. We recommend constraints instead). Also need to update the Open API spec to state the endpoint to create new ones deprecated. @sighphyre 
- [x] Deprecate "feature variants" in favor of "strategy variants". The goal is to "hide" it for all new feature flags being created. @kwasniew 
- [x] Remove legacy Google SSO provider. Use SAML 2.0 or OpenID Connect instead. @ivarconr 
- [x] Remove legacy "/api/feature" endpoint. Has been deprecated since Unleash v4.4. @daveleek 
- [x] Unleash Enterprise will make "License Key" mandatory. @ivarconr 
- [x] New global UI navigation  (sidebar) @kwasniew 
- [x] Rename "feature toggle" to "feature flag" in the UI (will not change API schema and definition or SDKs to keep breaking changes to a minimum) @sjaanus 
- [x] Error types to be standardized on the `message` property (removing the `description` property) @thomasheartman 
- [x] Some deprecated properties in the OpenAPI spec will be removed @sighphyre 
- [x] Remove deprecated old import (state service). Also need to update docs for this.  @daveleek 
  - PS! We will drop support for file based import at statup (#6306) 
  - WE have plans to extend the current import capability to support this in the future (#7128)

Todo:
* Make sure our documentation reflects v6 requirements and capabilities. 

**Dropped (did not make it for the v6 plans)**
[...]

PR #4794, opened 2023-09-20
chore/remove deprecated feature update

daveleek:
## About the changes

Removes the deprecated feature-update event that was scheduled for removal in 4.3.
Also removes some related endpoint methods that no longer have connected endpoints/references

## Discussion points

Is it safe to remove an event like this? What happens if past events exist in the db and you fetch them? We don't seem to have anything currently that actually produces these events so the scenario needs some manual hands on to test


website/docs/reference/event-types.mdx
```diff
@@ -66,41 +66,6 @@ This event fires when you create a feature. The `data` property contains the det
...
-    "impressionData": true
-  },
-  "preData": null,
-  "tags": [],
-  "featureName": "new-feature",
-  "project": "my-project",
-  "environment": null
-}
-```
-
```
thomasheartman, line 103: I think we should keep this, but change the admonition from a deprecation notice to an obsolescence notice, saying that the event was deprecated in 5.3 and removed in 5.6? Maybe mention that it has not been emitted since ... whenever?

The reason I think we should keep it, is if someone on v4.2 wants to find information on it, then it should still be available.

src/lib/routes/admin-api/feature.ts
thomasheartman, line 1: Wow, this is a lot of deletion! 👀 so all this deleted code has just not been used since 4.3? I'm assuming that this is not a breaking change, right? It's just old code that hasn't been cleaned up?

thomasheartman (commented):
This looks like the right thing to do, but I have a few comments before approving this; hope you don't mind 😄

ivarconr:
This needs to be part of Unleash v6 (#4380)

PR #4801, opened 2023-09-20
docs: Unleash Slack App integration

nunogois:
https://linear.app/unleash/issue/2-1235/docs-slack-app-integration-documentation

This adds a new reference doc for the new Unleash Slack App integration and marks the previous Slack integration as deprecated.

As a side-effect this PR also fixes an issue where we wouldn't be able to delete tags with special characters.


website/docs/reference/integrations/slack-app.md
```diff
@@ -0,0 +1,64 @@
+---
+id: slack-app
```
thomasheartman, line 2: When the id is the same as the file name, you don't need the id. In fact, most of our files don't have this, so I'd suggest removing it for consistency.
```suggestion
```

website/docs/reference/integrations/slack-app.md
```diff
@@ -0,0 +1,64 @@
...
+title: Slack App
+---
+
+:::info Availability
+
+The Slack App integration was introduced in **Unleash 5.5**.
+
+:::
+
+The Slack App integration posts messages to the selected channels in your Slack workspace.
```
thomasheartman, line 12: Minor suggestion. Could we maybe also elaborate a bit? Don't know if this is correct, but something like that?

```suggestion
The Slack App integration posts messages to a specified set of channels in your Slack workspace. The channels can be public or private, and can be specified on a per-flag basis.
```

website/docs/reference/integrations/slack-app.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Installation {#installation}
+
+To install the Slack App integration, follow these steps:
+
+1. Navigate to the *integrations* page in the Unleash admin UI (available at the URL `/integrations`) and click "configure" on the Slack App integration.
+2. On the integration configuration form, click the "install & connect" button.
```
thomasheartman, line 19: small nit: click is something you only do with a mouse, generally. With touch and keyboard, you don't really 'click', so I suggest using words that don't rely on the medium.

```suggestion
1. Navigate to the *integrations* page in the Unleash admin UI (available at the URL `/integrations`) and select "configure" on the Slack App integration.
2. On the integration configuration form, use the "install & connect" button.
```

website/docs/reference/integrations/slack-app.md
```diff
@@ -0,0 +1,64 @@
...
+3. A new tab will open, asking you to select the Slack workspace where you'd like to install the app.
+4. After successful installation of the Unleash Slack App in your chosen Slack workspace, you'll be automatically redirected to a page displaying a newly generated access token.
+5. Copy this access token and paste it into the `Access token` field within the integration settings.
+
+By default, the Unleash Slack App is granted access to public channels. If you want the app to post messages to private channels, you'll need to manually invite it to each of those channels.
+
+## Configuration {#configuration}
+
+The configuration settings allow you to choose the events you're interested in and whether you want to filter them by projects and environments. You can configure a comma-separated list of channels to post the configured events to. These channels are always notified, regardless of the event type or the presence of tags.
```
thomasheartman, line 28: What does the 'presence of tags' mean here? I'm assuming it's because you can use tags to add extra channels, but that could be made clearer by linking to the "tags" section and maybe giving a brief explanation here.

Something like:

```suggestion
The configuration settings allow you to choose the events you're interested in and whether you want to filter them by projects and environments. You can configure a comma-separated list of channels to post the configured events to. These channels are always notified, regardless of the event type or the presence of [slack tags](#tags).
```

website/docs/reference/integrations/slack-app.md
thomasheartman, line 1: As a general note, you don't need to use the `{#id}` suffix (e.g. `Tags {#tags}`) if the id you want is just the title with spaces replaced with hyphens. However, it's perfectly fine to keep it in if you prefer, as it makes linking to those elements a little more robust.

website/docs/reference/integrations/slack-app.md
```diff
@@ -0,0 +1,64 @@
...
+#### Parameters {#parameters}
+
+The Unleash Slack App integration takes the following parameters.
+
+- **Access token** - This is the only required property. After successful installation of the Unleash Slack App in your chosen Slack workspace, you'll be automatically redirected to a page displaying a newly generated access token. You should copy this access token and paste it into this field.
+- **Channels** - A comma-separated list of channels to post the configured events to. These channels are always notified, regardless of the event type or the presence of a slack tag.
+
+## Tags {#tags}
+
+Besides the configured channels, you have the flexibility to choose to notify other channels by tagging your feature toggles with Slack-specific tags. For instance, if you want the Unleash Slack App to send notifications to the `#general` channel, simply configure your feature toggle with the "slack:general" tag. This will ensure that any configured events related to that feature toggle will notify the tagged channel in addition to any configured channels.
```
thomasheartman, line 58: Another minor thing: I think we're moving to using "feature flags" instead of "feature toggles", so I think that in new documents, we should try and stick to that. But maybe @dgorton has a different opinion?

Also, more importantly, I think we should explain tags a little better. "slack:general" doesn't tell me explicitly that it's a tag of type "slack" with the value "general". Also, does it accept both "#general" and "general"? I know that I forget which one is right every time.

Also, "any configured channels" means configured for the entire instance, right? Maybe that could be a little clearer too.

Maybe something like:

```suggestion
Besides the configured channels, you can choose to notify other channels by tagging your feature flags with Slack-specific tags. For instance, if you want the Unleash Slack App to send notifications to the `#general` channel, add a slack-type tag with the value "general" (or "#general"; both will work) to your flag. This will ensure that any configured events related to that feature flag will notify the tagged channel in addition to any channels configured on the integration-level.
```

website/docs/reference/integrations/slack-app.md
```diff
@@ -0,0 +1,64 @@
...
+
+To exclusively use tags for determining notification channels, you can leave the "channels" field blank in the integration configuration. Since you can have multiple configurations for the integration, you're free to mix and match settings to meet your precise needs. Before posting a message, all channels for that event, both configured and tagged, are combined and duplicates are removed.
+
+![Slack Tags](/img/slack_addon_tags.png)
+
+In the picture you can see we have defined two slack tags for the "new-payment-system" toggle. In this example Unleash will post updates to the **#notifications** and **#random** channel, along with any channels defined in the integration configuration.
```
thomasheartman, line 64: The alt text here could be better, but I think this would be even better as a `Figure` component, where the text in the paragraph is the caption. (Note: I'm not sure whether the markdown formatting will work in the caption text, but we can remove that if need be).

```suggestion
<!-- this import should probably go up top --> 
import Figure from '@site/src/components/Figure/Figure.tsx' 

<Figure caption="We have defined two slack tags for the \"new-payment-system\" flag. In this example Unleash will post updates to the **#notifications** and **#random** channels, along with any channels defined in the integration configuration." img="/img/slack-addon-tags.png"/>
```

thomasheartman (approved):
Nice one! A couple minor suggestions, but I'm generally pretty happy 😄

website/docs/reference/integrations/slack-app.md
```diff
@@ -0,0 +1,64 @@
...
+- **Channels** - A comma-separated list of channels to post the configured events to. These channels are always notified, regardless of the event type or the presence of a slack tag.
+
+## Tags {#tags}
+
+Besides the configured channels, you have the flexibility to choose to notify other channels by tagging your feature toggles with Slack-specific tags. For instance, if you want the Unleash Slack App to send notifications to the `#general` channel, simply configure your feature toggle with the "slack:general" tag. This will ensure that any configured events related to that feature toggle will notify the tagged channel in addition to any configured channels.
+
+To exclusively use tags for determining notification channels, you can leave the "channels" field blank in the integration configuration. Since you can have multiple configurations for the integration, you're free to mix and match settings to meet your precise needs. Before posting a message, all channels for that event, both configured and tagged, are combined and duplicates are removed.
+
+![Slack Tags](/img/slack_addon_tags.png)
```
thomasheartman, line 62: Does this file already exist or should it be added with this PR? Regardless, could I ask you to rename it to use hyphens instead of underscores? That's more consistent with most of the other images (should be all, but some anomalies might have snuck in).

website/docs/reference/integrations/slack-app.md
nunogois: Oh right, good point!

website/docs/reference/integrations/slack-app.md
nunogois: Oh, I like this! I went to double-check whether `#channels` would work on both the integration configuration and tags, and good news is that they do!

However I caught a bug where it would not be possible to delete such a tag (e.g. "slack:#general") so as a slight side-effect of this PR I fixed it [here](https://github.com/Unleash/unleash/pull/4801/commits/328defd7fe305e700fb7a157fa5996f31a622087).

website/docs/reference/integrations/slack-app.md
nunogois: It already exists, renamed in https://github.com/Unleash/unleash/pull/4801/commits/f1603fea791f6bb7a50de07f52bc2ace209f6e2a

website/docs/reference/integrations/slack-app.md
nunogois: Thanks, addressed in https://github.com/Unleash/unleash/pull/4801/commits/ed4eb358cc270feb830fecd6ff887ea4bac12e47

PR #4805, opened 2023-09-21
task: Added workflow for calling update-version-action

chriswk:
## What
This adds a workflow which automatically triggers on published releases to update the version number for our version checker. In addition it adds a workflow dispatch, in case the version number ended up wrong after an automatic run (for instance when patching an earlier released minor). 

## Observations
Currently the version checker which receives the update only validates that the new version is also valid semver, it does not however check/verify that we're not suddenly telling it an older version is the newest (improvement available in the version updater repo).

### DIFF
--- a/.github/workflows/update_version_for_version_checker.yml
+++ b/.github/workflows/update_version_for_version_checker.yml
@@ -1,0 +1,35 @@
+name: Update version for version checker
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Which version should we set OSS to"
+        type: 'string'
+        required: true
+
+jobs:
+  update:
+    if: ${{ !(contains(github.ref, 'beta') || contains(github.ref, 'alpha')) }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


gastonfournier (commented):
When is this triggering? If it is with new enterprise releases, then I think this is perfectly fine.

Not sure if we want to keep the `-main` postfix: https://github.com/Unleash/unleash/blob/b4742df8be57417585d6b443f5160eb38451275f/package.json#L4 the main reason for this was to express that the version is in continuous change because we didn't have a mechanism for bumping it on formal releases, and bumping it on every change was time-consuming and could also create an infinite cycle with enterprise that continuously updates OSS on every change

.github/workflows/update_version_for_version_checker.yml
```diff
@@ -0,0 +1,35 @@
+name: Update version for version checker
+
+on:
+  release:
+    types: [published]
```
gastonfournier, line 5: Because of this I believe this will work just fine :+1:

gastonfournier (approved):

PR #4808, opened 2023-09-21
fix: string-width issue when running docker container

kwasniew:
## About the changes

I run `yarn install --production=true` inside docker directory using yarn 1.22.19 (same as alpine version we use in docker).

I built a docker container, went inside it and checked the string-width version. It's 4.2.3 now (correct), previously it was 5.x.x (incorrect). 

If we have a similar problem in future:
* `docker build . -t localbuild:local` - pick whatever name you like
* `docker run -it localbuild:local /bin/sh` - go inside the container
* `cat node_modules/string-width/package.json` - check version if it matches your local node_modules version


gardleopard (approved):
LGTM, great @kwasniew :)

PR #4812, opened 2023-09-22
feat: optimize private projects for enterprise

sjaanus:
1. Implement isEnterprise flag on backend
2. projectAccess now returns **ALL** project mode, so no filter will be applied to increase performance

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -481,6 +481,10 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
     const clientFeatureCaching = loadClientCachingOptions(options);
 
     const prometheusApi = options.prometheusApi || process.env.PROMETHEUS_API;
+
+    const isEnterprise =
+        Boolean(options.enterpriseVersion) && ui.name?.toLowerCase() !== 'pro';
+
     return {
         db,
         session,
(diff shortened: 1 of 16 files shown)


src/lib/create-config.ts
```diff
@@ -481,6 +481,11 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
     const clientFeatureCaching = loadClientCachingOptions(options);
 
     const prometheusApi = options.prometheusApi || process.env.PROMETHEUS_API;
+
+    const isEnterprise =
```
sjaanus, line 485: This is following exactly the same logic as `isEnterprise` in frontend.

src/lib/features/playground/playground-service.ts
```diff
@@ -102,17 +102,21 @@ export class PlaygroundService {
...
-                projects === ALL
-                    ? accessibleProjects
-                    : projects.filter((project) =>
-                          accessibleProjects.includes(project),
-                      );
-            console.log(accessibleProjects);
+            if (projectAccess.mode === 'all') {
+                filteredProjects = projects;
+            } else {
+                if (projects === ALL) {
```
kwasniew, line 112: if you flatten those to if/else if/else I think TS will still work out the details.

kwasniew (approved):
1 comment inside

PR #4815, opened 2023-09-22
feat: add more events in integrations

nunogois:
https://linear.app/unleash/issue/2-1253/add-support-for-more-events-in-the-slack-app-integration

Adds support for a lot more events in our integrations. Here is how the full list looks like:

- ADDON_CONFIG_CREATED
- ADDON_CONFIG_DELETED
- ADDON_CONFIG_UPDATED
- API_TOKEN_CREATED
- API_TOKEN_DELETED
- CHANGE_ADDED
- CHANGE_DISCARDED
- CHANGE_EDITED
- CHANGE_REQUEST_APPLIED
- CHANGE_REQUEST_APPROVAL_ADDED
- CHANGE_REQUEST_APPROVED
- CHANGE_REQUEST_CANCELLED
- CHANGE_REQUEST_CREATED
- CHANGE_REQUEST_DISCARDED
- CHANGE_REQUEST_REJECTED
- CHANGE_REQUEST_SENT_TO_REVIEW
- CONTEXT_FIELD_CREATED
- CONTEXT_FIELD_DELETED
- CONTEXT_FIELD_UPDATED
- FEATURE_ARCHIVED
- FEATURE_CREATED
- FEATURE_DELETED
- FEATURE_ENVIRONMENT_DISABLED
- FEATURE_ENVIRONMENT_ENABLED
- FEATURE_ENVIRONMENT_VARIANTS_UPDATED
- FEATURE_METADATA_UPDATED
- FEATURE_POTENTIALLY_STALE_ON
- FEATURE_PROJECT_CHANGE
- FEATURE_REVIVED
- FEATURE_STALE_OFF
- FEATURE_STALE_ON
- FEATURE_STRATEGY_ADD
- FEATURE_STRATEGY_REMOVE
- FEATURE_STRATEGY_UPDATE
- FEATURE_TAGGED
- FEATURE_UNTAGGED
- GROUP_CREATED
- GROUP_DELETED
- GROUP_UPDATED
- PROJECT_CREATED
- PROJECT_DELETED
- SEGMENT_CREATED
- SEGMENT_DELETED
- SEGMENT_UPDATED
- SERVICE_ACCOUNT_CREATED
- SERVICE_ACCOUNT_DELETED
- SERVICE_ACCOUNT_UPDATED
- USER_CREATED
- USER_DELETED
- USER_UPDATED

I added the events that I thought were relevant based on my own discretion. Know of any event we should add? Let me know and I'll add it 🙂 

For now I only added these events to the new Slack App integration, but we can add them to the other integrations as well since they are now supported.

The event formatter was refactored and changed quite a bit in order to make it easier to maintain and add new events in the future. As a result, events are now posted with different text. Do we consider this a breaking change? If so, I can keep the old event formatter around, create a new one and only use it for the new Slack App integration.

I noticed we don't have good 404 behaviors in the UI for things that are deleted in the meantime, that's why I avoided some links to specific resources (like feature strategies, integration configurations, etc), but we could add them later if we improve this.

This PR also tries to add some consistency to the the way we log events.


frontend/src/component/integrations/IntegrationForm/IntegrationForm.tsx
```diff
@@ -71,10 +71,12 @@ export const IntegrationForm: VFC<IntegrationFormProps> = ({
...
-    const selectableEvents = provider?.events?.map(event => ({
-        value: event,
-        label: event,
-    }));
+    const selectableEvents = provider?.events
+        ?.map(event => ({
+            value: event,
+            label: event,
+        }))
+        .sort((a, b) => a.label.localeCompare(b.label));
```
nunogois, line 79: We now have a bunch of events, so it's better if we sort them. Ideally they would be grouped, but maybe that can come in a later PR.

src/lib/addons/feature-event-formatter-md.ts
```diff
@@ -279,93 +525,35 @@ export class FeatureEventFormatterMd implements FeatureEventFormatter {
...
-                return 'created';
-            case FEATURE_UPDATED:
-                return 'updated';
-            case FEATURE_VARIANTS_UPDATED:
-                return 'updated variants for';
-            default:
-                return type;
-        }
-    }
+        Mustache.escape = (text) => text;
```
nunogois, line 547: [Globally disables Mustache escaping](https://github.com/janl/mustache.js#variables) so we don't need to specify link variables with `{{{link}}}`, which would be unintuitive.

chriswk (approved):
Wow. That's a lot of extra events to support!

Nice!

src/lib/services/segment-service.ts
```diff
@@ -171,7 +171,7 @@ export class SegmentService implements ISegmentService {
         await this.eventService.storeEvent({
             type: SEGMENT_DELETED,
             createdBy: user.email || user.username,
-            data: segment,
+            preData: segment,
```
gastonfournier, line 174: Nice discovery!

gastonfournier (approved):
LGTM!

PR #4816, opened 2023-09-22
feat: add group-deleted event

nunogois:
Adds a missing `group-deleted` event.


src/lib/services/group-service.ts
```diff
@@ -170,8 +170,18 @@ export class GroupService {
...
+
+        await this.groupStore.delete(id);
+
+        if (userName) {
+            await this.eventStore.store({
+                type: GROUP_DELETED,
+                createdBy: userName,
+                data: group,
+            });
+        }
```
nunogois, line 184: This uses expand contract so we don't break Enterprise, that's why you see:
 - `userName?`
 - `if (userName)`

This will get cleaned up after we merge the Enterprise PR.

gastonfournier (approved):
LGTM!

PR #4817, opened 2023-09-22
refactor: contract event group deleted after #4816

nunogois:
As mentioned in https://github.com/Unleash/unleash/pull/4816#discussion_r1334162284


chriswk (approved):
Nice

PR #4818, opened 2023-09-22
feat: enforce no transitive parents

kwasniew:
## About the changes

* checking if we can add dependency to a feature. If the feature has any children then it can have dependencies to any parents. 
* added transaction support for dependent features


andreas-unleash (approved):
LG

PR #4820, opened 2023-09-22
chore: Doc only should skip build

gastonfournier:
## About the changes
This fakes the build on docs-only PRs to be able to have a green build.
Touching a doc to validate it works

### DIFF
--- a/.github/workflows/build_doc_prs.yaml
+++ b/.github/workflows/build_doc_prs.yaml
@@ -23,3 +23,12 @@ jobs:
           sed -i 's/chalk\(\w\|\.\)\+//g' node_modules/@docusaurus/core/lib/client/serverEntry.js
           echo "Chalk removed"
           yarn build
+  build:
+    runs-on: ubuntu-latest
+    name: build
+
+    strategy:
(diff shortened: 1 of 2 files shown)


dgorton (commented):
I still don't know these build processes. This LGTM, but @thomasheartman has a better chance of spotting any potential issues.

gastonfournier:
> I still don't know these build processes. This LGTM, but @thomasheartman has a better chance of spotting any potential issues.

We'll have to try it with some documentation PR. The fact that this PR modifies an action, triggers the real build process. The good news is, that among the `9 successful checks`, none is the fake build process. This suggests that it'll trigger if and only if the PR is **only** changing documentation (actually this is my main concern: to trigger a fake build for docs that modify both code and docs)

gastonfournier:
This could be tested by merging this PR and then rebasing on https://github.com/Unleash/unleash/pull/4821

ivarconr (approved):
Nice 👍🙂

thomasheartman (approved):
I like it a lot 😄

PR #4821, opened 2023-09-24
Update more instances of "toggle" with "flag"

dgorton:
Landing page + sidebar nav + main 'How to create flags' page


website/docs/how-to/how-to-create-feature-toggles.md
```diff
@@ -1,23 +1,29 @@
 ---
-title: How to create a feature toggle
+title: How to create a feature flag
```
thomasheartman, line 2: Ooh, you wanna be very careful with this one. I think this will cause a URL change, so we should also add a redirect rule for this.

thomasheartman (changes requested):
I think this is a good change (and an important article to change), but I think that before this can be merged we need to:
1. Check whether this changes the URL of the how-to doc and if so:
2. Either:
  i. Use an explicit slug or id to keep the URL the same
  ii. Add a redirect rule that redirects `/how-to/how-to-create-feature-toggles` to `/how-to/how-to-create-feature-flags`

Personally I lean towards adding a redirect rule. But that said, if the URL doesn't change, then I think this is good to go.

website/docs/how-to/how-to-create-feature-toggles.md
dgorton: It didn't on my local - just went back and checked again and it's ok. 
I think the paths all live in sidebar.js and I was deliberate about not changing this one.

thomasheartman: Got it. In that case, it probably uses the file name instead. And you're right; the url is the same on the preview too. Thanks for checking.

thomasheartman (approved):
All good here 😄

PR #4824, opened 2023-09-25
feat: Delete dependency api

kwasniew:
## About the changes

API for deleting feature dependency

### DIFF
--- a/src/lib/features/dependent-features/dependent-features-controller.ts
+++ b/src/lib/features/dependent-features/dependent-features-controller.ts
@@ -2,10 +2,10 @@ import { Response } from 'express';
 import Controller from '../../routes/controller';
 import { OpenApiService } from '../../services';
 import {
-    CREATE_FEATURE,
     IFlagResolver,
     IUnleashConfig,
     IUnleashServices,
+    UPDATE_FEATURE,
 } from '../../types';
 import { Logger } from '../../logger';
 import {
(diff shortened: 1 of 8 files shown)


src/lib/features/dependent-features/dependent-features-controller.ts
```diff
@@ -49,13 +59,15 @@ export default class DependentFeaturesController extends Controller {
...
             transactionalDependentFeaturesService,
+            dependentFeaturesService,
             openApiService,
         }: DependentFeaturesServices,
         startTransaction: TransactionCreator<UnleashTransaction>,
     ) {
         super(config);
         this.transactionalDependentFeaturesService =
             transactionalDependentFeaturesService;
+        this.dependentFeaturesService = dependentFeaturesService;
```
kwasniew, line 70: deletion does not have to be transactional as there's only one DB operation involved

src/lib/features/dependent-features/dependent-features-service.ts
```diff
@@ -2,6 +2,7 @@ import { InvalidOperationError } from '../../error';
 import { CreateDependentFeatureSchema } from '../../openapi';
 import { IDependentFeaturesStore } from './dependent-features-store-type';
 
+export type FeatureDependencyId = { parent: string; child: string };
```
chriswk, line 5: Isn't this a circular dependency? dependent-features-store-type ends up depending on dependent-feature-service, and dependent-feature-service depends on dependent-features-store-type. Bringing in the dependent-features-store causes another circle since dependent-features-service calls out to the store, but the store depends on the service?

kwasniew: solved

PR #4826, opened 2023-09-25
feat: context/segment usage private

sjaanus:
1. GET single context field strategies will now hide private projects
2. GET single segment strategies will now hide private projects

This is just reusing the private project checker method, which is heavily tested in enterprise.

### DIFF
--- a/frontend/src/component/segments/SegmentProjectAlert.tsx
+++ b/frontend/src/component/segments/SegmentProjectAlert.tsx
@@ -38,7 +38,6 @@ export const SegmentProjectAlert = ({
             },
         });
     };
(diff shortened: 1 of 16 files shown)


frontend/src/component/segments/SegmentProjectAlert.tsx
```diff
@@ -78,11 +78,19 @@ export const SegmentProjectAlert = ({
         </StyledUl>
     );
 
-    if (projectsUsed.length > 1) {
+    if (projectsUsed.length > 0) {
```
sjaanus, line 81: The project list should also be shown, when just single project is using it.

kwasniew (approved):

PR #4829, opened 2023-09-25
feat: add kill switch for client metrics

FredrikOseberg:
This PR adds a killswitch for client metrics that we can use to disable metrics in our cloud offering.

Discussion points:
* Is it ok for us to change the response code to 204 when we disable the metrics? Is there a better way?


kwasniew (approved):
I understand that we distinguish 204 and 202 so that client can tell whether the feature is enabled or disabled right?

FredrikOseberg:
> I understand that we distinguish 204 and 202 so that client can tell whether the feature is enabled or disabled right?

That's the idea, though the clients don't support this for now. It would be a way for us in the future to distinguish the responses. Though this implementation detail is up for discussion.

PR #4830, opened 2023-09-25
refactor: prefer eventService.storeEvent methods

nunogois:
https://linear.app/unleash/issue/2-1403/consider-refactoring-the-way-tags-are-fetched-for-the-events

This adds 2 methods to `EventService`:
 - `storeEvent`;
 - `storeEvents`;

This allows us to run event-specific logic inside these methods. In the case of this PR, this means fetching the feature tags in case the event contains a `featureName` and there are no tags specified in the event.

This prevents us from having to remember to fetch the tags in order to store feature-related events except for very specific cases, like the deletion of a feature - You can't fetch tags for a feature that no longer exists, so in that case we need to pre-fetch the tags before deleting the feature.

This also allows us to do any event-specific post-processing to the event before reaching the DB layer.
In general I think it's also nicer that we reference the event service instead of the event store directly.

There's a lot of changes and a lot of files touched, but most of it is boilerplate to inject the `eventService` where needed instead of using the `eventStore` directly.

Hopefully this will be a better approach than https://github.com/Unleash/unleash/pull/4729


src/lib/services/event-service.ts
```diff
@@ -35,4 +41,44 @@ export default class EventService {
...
+                    event.tags ||
+                    tags
+                        .filter((tag) => tag.featureName === event.featureName)
+                        .map((tag) => ({
+                            value: tag.tagValue,
+                            type: tag.tagType,
+                        })),
+            }));
+
+            return this.eventStore.batchStore(eventsWithTags);
```
gastonfournier, line 80: This code might be better if we process it once and convert it to a `Map<string, string[]>` with the key being the feature name and the value the list of tags.

We're also iterating the list of events twice and I think you can do it in one go. I'd do something like:

```typescript
for (const enhancer of enhancers) {
    events = await enhance(events);
}
return this.eventStore.batchStore(events);
```
where enhance can be "add tags" or decorate with username (might be handy for one of our next tasks in Q4)

src/lib/services/event-service.ts
```diff
@@ -35,4 +41,44 @@ export default class EventService {
...
+
+    async storeEvent(event: IBaseEvent): Promise<void> {
+        if (event.featureName && !event.tags) {
+            const tags = await this.featureTagStore.getAllTagsForFeature(
+                event.featureName,
+            );
+            return this.eventStore.store({ ...event, tags });
+        }
+        return this.eventStore.store(event);
+    }
```
gastonfournier, line 53: I think we should just
```suggestion
    async storeEvent(event: IBaseEvent): Promise<void> {
        return this.storeEvents([event]);
    }
```

src/lib/types/events.ts
```diff
@@ -338,14 +326,12 @@ export class FeatureEnvironmentEvent extends BaseEvent {
         featureName: string;
         environment: string;
         createdBy: string | IUser;
-        tags: ITag[];
```
gastonfournier, line 341: I think it's fine to leave them as optional just in case some of these are used in Enterprise repository, it **could** make the transition easier.

But it might be also we move faster by breaking the contract and fixing it quickly, maybe we can check this by linking your branch into enterprise and having the change ready

src/test/e2e/api/admin/event.e2e.test.ts
```diff
@@ -22,6 +24,7 @@ beforeAll(async () => {
         },
     });
     eventStore = db.stores.eventStore;
+    eventService = new EventService(db.stores, { getLogger });
```
gastonfournier, line 27: Maybe it's safe to remove `eventStore` variable above

gastonfournier (commented):
This is looking really nice, I just pointed out a few details that we can check tomorrow, sorry for the late review!

src/lib/features/access/createAccessService.ts
```diff
@@ -24,9 +26,15 @@ export const createAccessService = (
...
     const accessStore = new AccessStore(db, eventBus, getLogger);
+    const featureTagStore = new FeatureTagStore(db, eventBus, getLogger);
+    const eventService = new EventService(
+        { eventStore, featureTagStore },
+        config,
+    );
     const groupService = new GroupService(
         { groupStore, eventStore, accountStore },
         { getLogger },
+        eventService,
```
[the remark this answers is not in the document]
nunogois: Nice catch, addressed in https://github.com/Unleash/unleash/pull/4830/commits/ac64a6be6e38db431a8a24adf640c105391f8cc8

src/lib/features/group/createGroupService.ts
```diff
@@ -1,21 +1,28 @@
...
-    const { getLogger } = config;
+    const { getLogger, eventBus } = config;
     const groupStore = new GroupStore(db);
     const accountStore = new AccountStore(db, getLogger);
     const eventStore = new EventStore(db, getLogger);
+    const featureTagStore = new FeatureTagStore(db, eventBus, getLogger);
+    const eventService = new EventService(
+        { eventStore, featureTagStore },
+        { getLogger },
+    );
```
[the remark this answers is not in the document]
nunogois: I disagree with this one due to the fact that other `createXService` methods follow the same approach: They create the needed services inside the methods instead of accepting them as arguments.

If we decide to refactor this I would prefer to target all of those methods and do it in a separate PR.

src/lib/services/addon-service.ts
```diff
@@ -55,13 +58,15 @@ export default class AddonService {
             flagResolver,
         }: Pick<IUnleashConfig, 'getLogger' | 'server' | 'flagResolver'>,
         tagTypeService: TagTypeService,
+        eventService: EventService,
```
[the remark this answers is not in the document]
nunogois: Yep, 100% correct. Addressed in https://github.com/Unleash/unleash/pull/4830/commits/58c1442b8dde1e7e6dc95d54c785306302e7e492

src/test/e2e/api/admin/event.e2e.test.ts
nunogois: Went with [this](https://github.com/Unleash/unleash/pull/4830/commits/9266c7aaf1d9279515c6f9ddce3ece9fd73daadc) approach. I don't think the `deleteAll` method should be exposed on the service layer, but if you disagree I can add this proxy.

src/lib/types/events.ts
nunogois: I think this is definitely a case where we move faster by breaking the contract and fixing it quickly. I have the Enterprise PR ready to go if you'd like to have a look at it.

src/lib/services/event-service.ts
nunogois: What do you think of https://github.com/Unleash/unleash/pull/4830/commits/03976efb1815b4f55d550650af24347e1bcad9f0?

Alternatively, we could also fetch all tags, build a Set out of them, and only then map them. Something like this:

```ts
const featureTagsMap: Map<string, ITag[]> = new Map();
const allTags = await this.featureTagStore.getAll();

for (const tag of allTags) {
  const featureTags = featureTagsMap.get(tag.featureName) || [];
  featureTags.push({ value: tag.tagValue, type: tag.tagType });
  featureTagsMap.set(tag.featureName, featureTags);
}
```

But if we consider the total amount of events we're storing vs total amount of tags we could be fetching, I think the approach in the commit should be more performant, along with the advantage of being more readable. But I'm open to suggestions.

src/lib/features/group/createGroupService.ts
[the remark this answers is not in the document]
kwasniew: I can explain the details behind createXService methods. Those are "composition roots" for each service. So they hide the information how each service is created. Some of our services depend on other services (which is often the consequence of not have read models). If creating a dependent service is difficult I usually create another composition root (createXService) method for the dependent service and call it from my composition root. On the other hand some services are trivial to instantiate or I didn't have time to migrate them to compositon roots. So my recommendation is to decide on a case by case basis. If you feel the pain of constructing a complex graph of objects: create a compositon root for a dependent service.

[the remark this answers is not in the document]
gastonfournier: > they hide the information how each service is created

I think this implies that they should not know how other services are created other than themselves. It acts as a Dependency Injection layer, where services get services injected rather than creating them on each composition root, which might lead to maintainability issues when having to modify how a service is created...

> On the other hand some services are trivial to instantiate or I didn't have time to migrate them to composition roots. 

I believe this one is simple, but I'm ok with whatever Nuno decides to do here, we'll have a chat and move this forward.

src/test/e2e/api/admin/event.e2e.test.ts
gastonfournier: I agree! And I like your approach :+1:

src/lib/services/event-service.ts
gastonfournier: I love it, I'm thinking about your alternative or just keeping the previous implementation (because now that I look at the new code I'm thinking would it be faster to iterate twice a list in memory or to query the DB multiple times per each feature? I believe it depends on the type of event, but in most cases, 2 loops in memory will be faster). In any case, I think it's fine to go with this option because it's currently the one we have in the code (before each event we query the DB), but also, feel free to change it

src/lib/features/group/createGroupService.ts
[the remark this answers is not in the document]
gastonfournier: Also, in other places we prefer injecting the service: https://github.com/Unleash/unleash/pull/4830/files#diff-0f55a0a1460b7ad5fd21da38762c259c1580e877e4b6b746e0032ad27776c886R93 but as Mateusz said, it can be considered tech debt for later

gastonfournier (approved):
Left some comments, but I believe this one is good to go :rocket:

src/lib/services/event-service.ts
nunogois: True, was questioning this myself as well. What do you think of something like this instead?

```ts
private async enhanceEventsWithTags(
    events: IBaseEvent[],
): Promise<IBaseEvent[]> {
    const featureNamesSet = new Set<string>();
    for (const event of events) {
        if (event.featureName && !event.tags) {
            featureNamesSet.add(event.featureName);
        }
    }

    const featureTagsMap: Map<string, ITag[]> = new Map();
    const allTagsInFeatures = await this.featureTagStore.getAllByFeatures(
        Array.from(featureNamesSet),
    );

    for (const tag of allTagsInFeatures) {
        const featureTags = featureTagsMap.get(tag.featureName) || [];
        featureTags.push({ value: tag.tagValue, type: tag.tagType });
        featureTagsMap.set(tag.featureName, featureTags);
    }

    for (const event of events) {
        if (event.featureName && !event.tags) {
            event.tags = featureTagsMap.get(event.featureName);
        }
    }

    return events;
}
```

gastonfournier: I like this approach :+1:

nunogois: Thanks, addressed in https://github.com/Unleash/unleash/pull/4830/commits/c2cbad3063f9768ce1b29ac60e5935769fcc6f5f

PR #4833, opened 2023-09-26
feat: Api to list available parent options

kwasniew:
## About the changes

We need to choose only from dependent features that are in the same project and don't have any parents themselves.
This API returns a list of feature names that your feature can depend on. It will be used by the UI dropdown to populate the values.

### Important files

The interesting part is the DB query


src/lib/features/dependent-features/dependent-features-store.ts
```diff
@@ -38,6 +38,23 @@ export class DependentFeaturesStore implements IDependentFeaturesStore {
         return rows.map((row) => row.child);
     }
 
+    async getParentOptions(child: string): Promise<string[]> {
```
kwasniew, line 41: algorithms:
* select your feature's project
* select all features from this project that are not children themselves

sjaanus: Is the naming deliberate? What are the options? getPossibleParentFeatures?

src/lib/features/dependent-features/dependent-features-store.ts
```diff
@@ -38,6 +38,23 @@ export class DependentFeaturesStore implements IDependentFeaturesStore {
...
+            return [];
+        }
+        const rows = await this.db('features as f')
+            .leftJoin('dependent_features as df', 'f.name', 'df.child')
+            .where('f.project', result[0].project)
+            .andWhere('f.name', '!=', child)
+            .andWhere('df.child', null)
+            .select('f.name');
+
+        return rows.map((item) => item.name);
```
sjaanus, line 55: If the feature is already parent of child, should it still be returned?

src/lib/features/dependent-features/dependent-features-store.ts
kwasniew: yup, I was thinking that parent options is a more elegant way of saying possible parent features :)

src/lib/features/dependent-features/dependent-features-store.ts
kwasniew: if the parent is already a child then it shouldn't (` .andWhere('df.child', null)`). If the parent is already a parent it's a non-issue.

sjaanus: I am not fully sure when this query is being ran, but if child is feature A, if feature B is parent of A, should the query still return B?

kwasniew: if feature B is not a child of any other parent then yes. If your parent is a child of any other feature it should never be returned.

PR #4835, opened 2023-09-26
feat: Enforce one dependency

kwasniew:
## About the changes

Even though our DB supports multiple parent dependencies we want to start experimenting only with one. I'm removing all deps for a feature before creating a new one. Everything is part of the DB Tx started in the controller.


src/test/e2e/api/client/feature.e2e.test.ts
```diff
@@ -60,11 +60,6 @@ beforeAll(async () => {
         'featureY',
         { feature: 'featureX', variants: ['featureXVariant'] },
     );
-    // depend on parent being disabled
```
kwasniew, line 63: currently we only support only one dependency in admin. Will bring this back when we support more than one.

Tymek (approved):

PR #4836, opened 2023-09-26
chore: update DATABASE_URL to use the database created via POSTGRES_D…

snosratiershad:
…B in compose

## About the changes
I've noticed that all migrations are applied to the `postgres` database along we've the variable POSTGRES_DB=db in docker compose.
The reason is the DATABASE_URL variable still refers to `postgres` instead `db`. I updated the URL and now it works with `db`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ services:
       - "4242:4242"
     environment:
       # This points Unleash to its backing database (defined in the `db` section below)
-      DATABASE_URL: "postgres://postgres:unleash@db/postgres"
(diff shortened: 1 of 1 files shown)


Tymek (approved):

thomasheartman:
Hey, @snosratiershad 😄 Yeah, this looks like an oversight on our part. Funny how it went undetected for so long 🤔 I'll just do some extra double checking, but I think this is good to go. Thanks for your contribution 🙏🏼

thomasheartman (approved):

PR #4841, opened 2023-09-27
Feat/reset pnps

FredrikOseberg:
* This PR resets pnps prompts by removing entries in the user_feedback table

### DIFF
--- a/src/migrations/20230927071830-reset-pnps-feedback.js
+++ b/src/migrations/20230927071830-reset-pnps-feedback.js
@@ -3,7 +3,7 @@
 exports.up = function (db, callback) {
     db.runSql(
         `
-       DELETE FROM user_feedback WHERE feedback_id = 'pnps';
+       DELETE FROM user_feedback WHERE feedback_id = 'pnps' AND given < NOW() - INTERVAL '3 months';
         `,
         callback(),
     );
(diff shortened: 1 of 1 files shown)


ivarconr (approved):

src/migrations/20230927071830-reset-pnps-feedback.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+       DELETE FROM user_feedback WHERE feedback_id = 'pnps';
```
ivarconr, line 6: Maybe only when it is older than 3 months?

ivarconr (changes requested):

src/migrations/20230927071830-reset-pnps-feedback.js
FredrikOseberg: Unfortunately, this table does not have any timestamp columns.

sjaanus: There is column `given`, with timestamp.

FredrikOseberg: > There is column `given`, with timestamp.

Yes, but that is only populated if you gave feedback. If you hide the feedback form it won't be populated.

FredrikOseberg: So two options: 
* We clear all entries
* We clear only the entries that have previously given feedback

What do you prefer?

ivarconr (approved):

PR #4842, opened 2023-09-27
fix: empty object playground

kwasniew:
## About the changes
Screenshot 2023-09-27 at 09 47 35

With empty object in code field and custom property selected JSON parser should not return any errors


frontend/src/component/playground/Playground/PlaygroundForm/PlaygroundCodeFieldset/PlaygroundCodeFieldset.tsx
```diff
@@ -63,7 +63,8 @@ export const PlaygroundCodeFieldset: VFC<IPlaygroundCodeFieldsetProps> = ({
 
                     setFieldExist(
                         contextValue[contextField] !== undefined ||
-                            contextValue?.properties[contextField] !== undefined
+                            contextValue?.properties?.[contextField] !==
```
kwasniew, line 66: properties may be empty

thomasheartman (approved):
Nice! Was this all there was to it? Seems pretty simple 😄  I'm assuming this works whenever the `properties` property is missing? Super 👏🏼

PR #4843, opened 2023-09-27
fix: project features table initial state

Tymek:
When loading don't keep initial state, but recalculate if after proper data is present

https://linear.app/unleash/issue/1-1432/project-overview-filtering-is-not-saved-correctly-4789


FredrikOseberg (approved):
LGTM

PR #4845, opened 2023-09-27
chore: limit the amount of unannounced events we announce

gastonfournier:
## About the changes
When the events table is large we might be doing a full table scan searching for unannounced events. We spotted it due to a performance alert and confirmed in AWS performance insights
image

The proposal is to limit this operation to 500 events (rule of thumb) per round https://github.com/Unleash/unleash/blob/f82ae354ebe6d3b0a3ddf15a051d6c8ed995c10e/src/lib/services/index.ts#L141-L147 and also ignore the events older than a day (because it seems reasonable)

## Discussion points
**Idea**: split the `events` table into `recent_events` and `historical_events`. Recent can be anything from a day/week/month. This would help with recurrent queries that rely on recent data from the event's table such as optimal 304 calculation or event this scheduled task that sends unannounced events.

### DIFF
--- a/src/lib/db/event-store.ts
+++ b/src/lib/db/event-store.ts
@@ -412,7 +412,9 @@ class EventStore implements IEventStore {
             .update({ announced: true })
             .where('announced', false)
             .whereNotNull('announced')
-            .returning(EVENT_COLUMNS);
+            .where('created_at', '>', new Date(Date.now() - 1000 * 60 * 60))
+            .returning(EVENT_COLUMNS)
+            .limit(500);
 
         return rows.map(this.rowToEvent);
     }
--- a/src/lib/db/event-store.ts
+++ b/src/lib/db/event-store.ts
@@ -408,7 +408,7 @@ class EventStore implements IEventStore {
         return this.eventEmitter.off(eventName, listener);
     }
 
-    private async setUnannouncedToAnnounced(): Promise<IEvent[]> {
+    async setUnannouncedToAnnounced(): Promise<IEvent[]> {
         const rows = await this.db(TABLE)
             .update({ announced: true })
             .where('announced', false)
--- a/src/lib/db/event-store.ts
+++ b/src/lib/db/event-store.ts
@@ -12,6 +12,7 @@ import { sharedEventEmitter } from '../util/anyEventEmitter';
 import { Db } from './db';
 import { Knex } from 'knex';
 import EventEmitter from 'events';
+import { subDays } from 'date-fns';
 
 const EVENT_COLUMNS = [
     'id',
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):
This sounds like a sensible change to me! It also only deals with events in the last hour, right? 

As for the discussion point: yeah, that might be a good idea, actually. I'm in favor if it'll have net positive effects.

src/lib/db/event-store.ts
```diff
@@ -412,7 +412,9 @@ class EventStore implements IEventStore {
             .update({ announced: true })
             .where('announced', false)
             .whereNotNull('announced')
-            .returning(EVENT_COLUMNS);
+            .where('created_at', '>', new Date(Date.now() - 1000 * 60 * 60))
```
gastonfournier, line 415:
```suggestion
            .where('created_at', '>', new Date(Date.now() - 24 * 1000 * 60 * 60))
```

gastonfournier:
> This sounds like a sensible change to me! It also only deals with events in the last hour, right?
> 
> As for the discussion point: yeah, that might be a good idea, actually. I'm in favor if it'll have net positive effects.

Actually, I wanted to limit this to last 24hs :P good spot: https://github.com/Unleash/unleash/pull/4845/files#r1338329830

src/lib/db/event-store.ts
kwasniew: I'd use: `import { addDays, subDays } from 'date-fns';`

thomasheartman: Ah, yeah, that would work as well! Which ever you think is better 💁🏼  Probably safer to go for 24.

gastonfournier: Thanks! I added subDays but I'm not 100% confident in my querying skills so I'll write a test

ivarconr (approved):

FredrikOseberg (approved):
LGTM

Regarding discussion point: 

The idea is solid, but it will also introduce query complexity. We should investigate whether we can solve the same problem in a different way before we decide on implementation.

gastonfournier:
> Regarding discussion point:
> 
> The idea is solid, but it will also introduce query complexity. We should investigate whether we can solve the same problem in a different way before we decide on implementation.

Another alternative by @gardleopard using partial indexes to index on the announced column (which is boolean but after doing a quick read it doesn't seem to be a problem for partial indexes) https://www.postgresql.org/docs/current/indexes-partial.html

gastonfournier:
We couldn't notice an improvement so we're trying this instead: https://github.com/Unleash/unleash/pull/4856

PR #4846, opened 2023-09-27
feat: read model for dependent features

kwasniew:
## About the changes

Introducing read model separate from the write model using store. The read model can be used by the feature overview to get parents/children without exposing write DB operations.

Screenshot 2023-09-27 at 11 33 15

### DIFF
--- a/src/lib/features/dependent-features/dependent-features-read-model.ts
+++ b/src/lib/features/dependent-features/dependent-features-read-model.ts
@@ -1,0 +1,42 @@
+import { Db } from '../../db/db';
+import { IDependentFeaturesReadModel } from './dependent-features-read-model-type';
+
+export class DependentFeaturesReadModel implements IDependentFeaturesReadModel {
+    private db: Db;
+
+    constructor(db: Db) {
+        this.db = db;
+    }
+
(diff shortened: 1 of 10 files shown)


src/lib/features/dependent-features/dependent-features-read-model.ts
```diff
@@ -0,0 +1,42 @@
...
+        const rows = await this.db('features as f')
+            .leftJoin('dependent_features as df', 'f.name', 'df.child')
+            .where('f.project', result[0].project)
+            .andWhere('f.name', '!=', child)
+            .andWhere('df.child', null)
+            .select('f.name');
+
+        return rows.map((item) => item.name);
+    }
+}
```
FredrikOseberg, line 42: Would prefer `features as feature` to help with scanning.

FredrikOseberg (approved):
Lovely. I commented on a small nit.

src/lib/features/dependent-features/dependent-features-read-model.ts
kwasniew: Will update in the next PR

PR #4853, opened 2023-09-27
feat: biome lint

chriswk:
This PR changes our linter/formatter to biome (https://biomejs.dev/)
Causing our prehook to run almost instantly, and our "yarn lint" task to run in sub 100ms.

Some trade-offs:
* Biome isn't quite as well established as ESLint
* Are we ready to install a different vscode plugin (the biome plugin) instead of the prettier plugin

biome results
ESLint results

The configuration set for biome also has a set of recommended rules, this is turned on by default, in order to get to something that was mergeable I have turned off a couple the rules we seemed to violate the most, that we also explicitly told eslint to ignore.


chriswk:
If we're happy with this, we could eventually apply it to frontend as well.

sighphyre:
> Biome isn't quite as well established as ESLint
> Are we ready to install a different vscode plugin (the biome plugin) instead of the prettier plugin

Both of these seem like really low risk, high value trade-offs to me.

thomasheartman:
I love the idea of this! But wait, it replaces both eslint and prettier? I thought it was just a linter. But I don't mind 😄

src/lib/services/feature-toggle-service.ts
```diff
@@ -1179,7 +1178,7 @@ class FeatureToggleService {
         featureName: string,
         projectId: string,
         newFeatureName: string,
-        replaceGroupId: boolean = true, // eslint-disable-line
```
gastonfournier, line 1182: Not sure here... I believe this is setting `replaceGroupId` to true, but because the next param `userName` does not have a default, it doesn't matter: the compiler would force us to provide this value in all calls to this method. What worries me a bit is that comments like `// @ts-ignore` can void the compiler hint.

If we don't set it to true... then undefined is falsy by default which might change the behavior. 

Maybe it checked all usages of the method to validate the value is always provided?

src/lib/services/state-service.ts
```diff
@@ -633,7 +633,7 @@ export default class StateService {
     async importTagTypes(
         tagTypes: ITagType[],
         keepExisting: boolean,
-        oldTagTypes: ITagType[] = [], // eslint-disable-line
```
gastonfournier, line 636: I believe same comment as in the other parameter with a default value and // eslint-disable-line

gastonfournier: Oh, I'm very intrigued now... this was added by renovate bot! https://github.com/Unleash/unleash/pull/1147/files#diff-8757be4693afd352b1e0507343d2de88066e2d8ec2f453a11be93920b9d0b182L599

gastonfournier (commented):
Overall, I like the recommendations, I love the speed boost, and I'm curious why frontend didn't get any recommendations.

I think this is a reversible decision (despite the many lines of code changed), so I'd in favor of trying it out

src/lib/error/from-legacy-error.ts
```diff
@@ -23,8 +23,6 @@ const getStatusCode = (errorName: string): number => {
             return 400;
         case 'InvalidTokenError':
             return 401;
-        case 'NoAccessError':
-            return 403;
```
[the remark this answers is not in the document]
chriswk: This was removed due to it having a duplicate

src/lib/services/feature-toggle-service.ts
chriswk: Oh, this was manual, we turned this off explicitly with the eslint-disable-line that was in place here. It's an incorrect usage of default arguments, since default arguments must come at the end of the parameter list in order to guarantee the non-default argument.

src/lib/services/state-service.ts
chriswk: That's very odd.

src/lib/error/from-legacy-error.ts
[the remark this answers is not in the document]
chriswk: See https://github.com/Unleash/unleash/blob/64899713028c6e12e4f7e174dd9aad816ff7e843/src/lib/error/from-legacy-error.ts#L56

chriswk:
> I love the idea of this! But wait, it replaces both eslint and prettier? I thought it was just a linter. But I don't mind 😄

Yeah, reading the history of Rome from Meta, before it became biome, it seems like they're also planning on a jest compatible test runner, but the buy in for now is for formatter and linter, though using the config we can turn off parts

chriswk:
So, @gastonfournier frontend did not get any changes because the base setup in root folder only touches src folder, frontend has a separate setup in its package json. Different PR. Enough changes here already

src/lib/services/feature-toggle-service.ts
chriswk: Turned out we had a test that was failing due to this. I rearranged the arguments so the default argument indeed comes last, there were only 5 places (1 in code, 4 in tests) that actually used this, so was manageable. I also checked that enterprise didn't use the method.

src/lib/error/from-legacy-error.ts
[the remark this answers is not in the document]
gastonfournier: Oh! Gotcha!

gastonfournier:
> So, @gastonfournier frontend did not get any changes because the base setup in root folder only touches src folder, frontend has a separate setup in its package json. Different PR. Enough changes here already

I can imagine! It crossed my mind that this could be a reason :+1:

gastonfournier (approved):

PR #4856, opened 2023-09-27
fix: partial index on events announced

gastonfournier:
## About the changes
Add partial index on events by announced. This should help avoid `Seq Scan on events` when the majority of events are announced=true

---
Co-authored-by: Ivar Østhus <ivar@getunleash.io>

### DIFF
--- a/src/lib/db/event-store.test.ts
+++ b/src/lib/db/event-store.test.ts
@@ -33,7 +33,8 @@ test('Trying to get events by name if db fails should yield empty list', async (
     await db.destroy();
 });
 
-test('Find unnanounced events is capped to 500', async () => {
+// We might want to cap this to 500 and this test can help checking that
+test('Find unannounced events returns all events', async () => { 
     const db = await dbInit('events_test', getLogger);
     const type = 'application-created' as const;
 
--- a/src/lib/db/event-store.test.ts
+++ b/src/lib/db/event-store.test.ts
@@ -34,7 +34,7 @@ test('Trying to get events by name if db fails should yield empty list', async (
 });
 
 // We might want to cap this to 500 and this test can help checking that
-test('Find unannounced events returns all events', async () => { 
+test('Find unannounced events returns all events', async () => {
     const db = await dbInit('events_test', getLogger);
     const type = 'application-created' as const;
 
--- a/src/lib/db/event-store.test.ts
+++ b/src/lib/db/event-store.test.ts
@@ -19,6 +19,7 @@ test('Trying to get events if db fails should yield empty list', async () => {
     const store = new EventStore(db, getLogger);
     const events = await store.getEvents();
     expect(events.length).toBe(0);
+    await db.destroy();
 });
 
 test('Trying to get events by name if db fails should yield empty list', async () => {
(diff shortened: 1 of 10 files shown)


src/lib/db/event-store.ts
```diff
@@ -412,11 +411,7 @@ class EventStore implements IEventStore {
         const rows = await this.db(TABLE)
             .update({ announced: true })
             .where('announced', false)
-            .whereNotNull('announced')
-            .where('created_at', '>', subDays(Date.now(), 1))
-            .returning(EVENT_COLUMNS)
-            .limit(500);
```
gastonfournier, line 418: Note: I've tested that limit does not work with update. There are alternatives but may require using native queries: 
- https://stackoverflow.com/questions/11432155/limiting-postgresql-update-command
- https://www.crunchydata.com/blog/simulating-update-or-delete-with-limit-in-postgres-ctes-to-the-rescue

src/lib/db/event-store.ts
```diff
@@ -412,11 +411,7 @@ class EventStore implements IEventStore {
         const rows = await this.db(TABLE)
             .update({ announced: true })
             .where('announced', false)
-            .whereNotNull('announced')
```
gastonfournier, line 415: Where not null no longer needed due to the migration

src/lib/db/event-store.test.ts
```diff
@@ -29,39 +30,24 @@ test('Trying to get events by name if db fails should yield empty list', async (
...
-        const insertQuery = db.rawDatabase('events');
-        await insertQuery
-            .insert({
-                type,
-                created_at: createdAt,
-                created_by: 'a test',
-                data: { name: 'test', createdAt },
-            })
-            .returning(['id']);
+test('Find unnanounced events is capped to 500', async () => {
```
sighphyre, line 36: Does this test name match the content of the test now? Based on the expect I don't think so?

sighphyre (approved):
I would wait for someone who can actually sanity check the SQL properly but yeah this does LGTM

chriswk (commented):
This should work, but can be heavy for instances with millions of events already in place.

src/lib/db/event-store.test.ts
gastonfournier: Oh, no it doesn't :)

chriswk:
To clarify, I'm not worried about the index, I'm pretty sure that will be fine. I'm a bit worried with the initial setup, setting all NULL announced to false, though I do think we mostly have remembered to set it, so the impact should be minimal.

chriswk (approved):
Adding a thumbsup

gastonfournier:
> To clarify, I'm not worried about the index, I'm pretty sure that will be fine. I'm a bit worried with the initial setup, setting all NULL announced to false, though I do think we mostly have remembered to set it, so the impact should be minimal.

We have DEFAULT false, but it was still nullable. I guess this was during a migration, but with the job marking unannounced events (both false or null) to true, we should have zero unannounced set to null. But I'm just doubling down on DEFAULT false just in case

sjaanus (approved):
LG

src/lib/db/event-store.test.ts
gastonfournier:
```suggestion
// We might want to cap this to 500 and this test can help checking that
test('Find unannounced events returns all events', async () => { 
```

src/migrations/20230927172930-events-announced-index.js
```diff
@@ -0,0 +1,17 @@
...
+        UPDATE events set announced = false where announced IS NULL;
+        ALTER TABLE events ALTER COLUMN announced SET NOT NULL;
+        ALTER TABLE events ALTER COLUMN announced SET DEFAULT false;
+        CREATE INDEX events_unannounced_idx ON events(announced) WHERE announced = false;
+        `,
+        callback(),
+    );
+};
+
+exports.down = function (db, callback) {
```
sjaanus, line 15: I think a down function would be beneficial.

gastonfournier: I considered that but down migrations are not executed automatically, so if we want to make changes we're probably going to move forward rather than backward

sighphyre: @sjaanus I know we're in a weird place with the down migrations and what we should do with them. I've never used them though. Are you using them? I don't remember any recommendations to users to consider down migrations either, our response has always been get latest version where we fixed the issue

sjaanus: @sighphyre I have not used, but I know customers have backed down to previous versions and in some cases if we do not provide down migration, it may make their instance unusable.

For example, if we currently do not allow nulls and previously somehow they were entered, we would probably just have errors in logs, but in other cases, if we change something related to projects/strategies/segments, it might break UI.

sighphyre: Oh yeah, if there are actually folks using them then it seems important

gastonfournier: > I know customers have backed down to previous versions and in some cases if we do not provide down migration, it may make their instance unusable.

Indeed, but even in these cases, there's no command for the customer to execute the down migration other than by hand. They'd have to either go one by one, or identify the one they need to execute. 

In such a scenario, the down migration does not update the migrations table (because we're not using the tool for this), and therefore the framework doing the up migration will still believe that the migration is applied and will never try to apply it again... this leaves the customer in a very weird situation. 

I know because we faced this same problem with our demo instance and we learned that down migrations should not be applied manually because they don't include the change in the migrations table (and they shouldn't! The right solution is to use the tool's command for doing it)

PR #4858, opened 2023-09-28
Fix open handles in tests

gastonfournier:
## About the changes
This fixes a bunch of openHandles from our tests

I've used this script to find out the ones that leave them:
`find src -name "*.test.ts" -printf "%f\n" | xargs -i sh -c "echo ===== {} && yarn test {}"`

If there's an issue, the script will halt and the last filename will be the one that has to be fixed. 

Each commit fixes one problem so it's easy to review


nunogois (approved):
LGTM

thomasheartman (approved):

PR #4861, opened 2023-09-28
fix: permissions in the role payload

nunogois:
Fixes the role payload to include only the needed properties from permissions. Fixes `400` (oneOf schema validation error) in project role creation.


gastonfournier (approved):
Thanks @nunogois!

PR #4862, opened 2023-09-28
test: makes overview spec less flaky by doing 2 step search

nunogois:
Makes the Cypress project overview e2e spec less flaky by making the search action a 2 step process.
This test is now passing for me locally when before it wasn't.


gastonfournier (approved):
LGTM!

PR #4864, opened 2023-09-29
feat: events for dependencies

kwasniew:
## About the changes

Dependency added, dependency removed and all dependencies removed have corresponding event entries
Screenshot 2023-09-29 at 12 54 43

Also updated node client dependency so that frontend API can support dependent features automatically

### DIFF
--- a/frontend/src/hooks/api/actions/useDependentFeaturesApi/useDependentFeaturesApi.ts
+++ b/frontend/src/hooks/api/actions/useDependentFeaturesApi/useDependentFeaturesApi.ts
@@ -82,6 +82,7 @@ export const useDependentFeaturesApi = (project: string) => {
         makeRequest,
         setToastData,
         formatUnknownError,
+        project,
     ];
     return {
         addDependency: useCallback(addDependency, callbackDeps),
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -161,7 +161,7 @@
     "stoppable": "^1.1.0",
     "ts-toolbelt": "^9.6.0",
     "type-is": "^1.6.18",
-    "unleash-client": "4.1.1",
+    "unleash-client": "4.2.0-beta.0",
```
kwasniew, line 164: frontend API support for dependent features

frontend/src/hooks/api/actions/useDependentFeaturesApi/useDependentFeaturesApi.ts
```diff
@@ -82,6 +82,7 @@ export const useDependentFeaturesApi = (project: string) => {
         makeRequest,
         setToastData,
         formatUnknownError,
+        project,
```
kwasniew, line 85: we had missing hook dependency

Tymek (approved):

PR #4900, opened 2023-10-02
chore: add workspace .vscode settings

nunogois:
Adds VSCode workspace settings `.vscode/settings.json` with a default formatter: `"editor.defaultFormatter": "biomejs.biome"`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,6 @@ unleash-server.tar.gz
 # Visual Studio Code
 jsconfig.json
 typings
-.vscode
 .nyc_output
 
 # We use yarn.lock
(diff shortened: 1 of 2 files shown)


chriswk (approved):
Going to be a problem with frontend for now? Since frontend still uses eslint/prettier ?

nunogois:
> Going to be a problem with frontend for now? Since frontend still uses eslint/prettier ?

Seems like this was working fine before but not now, so I think you're right, we should wait until we at least remove "frontend" from the Biome formatter ignore list. What's currently stopping us from doing that?

nunogois:
Closed in favor of https://github.com/Unleash/unleash/pull/4903

PR #4902, opened 2023-10-02
feat: generate orval types with dependent features

kwasniew:
## About the changes

Generated orval types and actual usage of the type in dependent features hook.

Fix incorrect type generated by orval for flat sdk context schema.


src/lib/openapi/spec/sdk-flat-context-schema.ts
```diff
@@ -7,7 +7,7 @@ export const sdkFlatContextSchema = {
     description: 'The Unleash context with flattened properties',
     type: 'object',
     required: ['appName'],
-    additionalProperties: { type: 'string' },
+    additionalProperties: true,
```
kwasniew, line 10: orval was generating incorrect type with type: 'string' on additional properties

Tymek (approved):

PR #4903, opened 2023-10-02
feat: biome lint frontend

nunogois:
Follows up on https://github.com/Unleash/unleash/pull/4853 to add Biome to the frontend as well.

image

Added a few `biome-ignore` to speed up the process but we may want to check and fix them in the future.


chriswk (approved):
I love it. Does it work? 
Some thoughts:
* No changes to package.json in frontend, 
   doesn't that mean that applying `yarn lint` in the frontend folder will still call eslint?
* I see we're applying backend's preference for single quotes to frontend as well. Is React JSX/TSX fine with that?

nunogois:
> I love it. Does it work? Some thoughts:
> 
> * No changes to package.json in frontend,
>   doesn't that mean that applying `yarn lint` in the frontend folder will still call eslint?
> * I see we're applying backend's preference for single quotes to frontend as well. Is React JSX/TSX fine with that?

Thank you for those good questions @chriswk 

>Does it work?

Seems to work fine now and I finally fixed the failing tests. Mind testing locally on your machine just to be sure?

>No changes to package.json in frontend,
doesn't that mean that applying yarn lint in the frontend folder will still call eslint?

You're right. This was addressed in https://github.com/Unleash/unleash/pull/4903/commits/f0f33c17b17b827b4c2e36e70ea14d065a66ea6d

It's not a big deal since it is super fast, but it would be great if we could drop these commands in frontend altogether and rely only on the top level ones, since those cover everything.

It's also weird that they act differently in backend and frontend. 
 - Backend: `yarn lint` will check your files and `yarn lint:fix` will attempt to fix them. This check covers both linting and formatting for both frontend and backend;
 - Frontend: `yarn lint` will attempt to lint your frontend files and `yarn lint:check` will lint-check your frontend files. `yarn fmt` will attempt to format your frontend files and `yarn fmt:check` will format-check your frontend files.

>I see we're applying backend's preference for single quotes to frontend as well. Is React JSX/TSX fine with that?

AFAIK it's totally fine.

chriswk:
Ran tests locally, everything seems to be green. Don't worry about Codacy at the moment. I think we either need to have a close look at the config or just drop it.

PR #4905, opened 2023-10-02
fix: update potentially-stale status dynamically

andreas-unleash:
Fixes 2 bugs:

- project-health-service keeping the feature types as an instance variable and only updating it once was preventing real calculation to happen if the lifetime value changed for a feature toggle type
- the ui was reading from a predefined map for the lifetime values so they would never reflect the BE change

Closes # [SR-66](https://linear.app/unleash/issue/SR-66/slack-question-around-potentially-stale-and-its-uses)

Screenshot 2023-10-02 at 14 37 17
Screenshot 2023-10-02 at 14 37 06


daveleek (approved):
LGTM!

andreas-unleash:
Ignoring overview.spec.ts cypress test failure - known issue

PR #4906, opened 2023-10-02
refactor: bubble promise instead of return await

nunogois:
Tiny refactor that bubbles promises instead of using `return await`. Should be more consistent with the rest of the changes in https://github.com/Unleash/unleash/pull/4903


chriswk (approved):
Seems fair.

PR #4907, opened 2023-10-02
Task/cliff changelog

chriswk:
### What
Initial effort at automating the task of updating CHANGELOG when doing releases.

This initial PR sets up the tool https://git-cliff.org/ and adds our missing updates since 4.2.0

The tool should be explored further, we could have the CHANGELOG linking linear issues, not just github issues, as well as ideally each commit displayed with it's short sha. However, rather than not having anything like this for the last 20+ releases, this will at least have our conventional commits be included in the CHANGELOG file.  

Next PR will setup up an automated task to run along with the release tasks to ensure we keep it up to date when releasing.


gastonfournier (approved):
I like how it looks! I think we'll get better at conventional commits over time now that we get to see the output of changelog :+1:

PR #4913, opened 2023-10-03
feat: prevent delete and archive on parent feature

kwasniew:
## About the changes

Prevent delete and archive of a parent feature.
We have both UI protection and backend API protection. 

In the UI we protect against single parent archive:
Screenshot 2023-10-03 at 12 13 25

For the bulk archive and delete we rely on the backend validation for now:

Screenshot 2023-10-03 at 11 42 40

Once we come up with a performant solution for dependencies info in bulk items/lists we can improve bulk archive UI.

### DIFF
--- a/src/lib/features/dependent-features/dependent-features-read-model-type.ts
+++ b/src/lib/features/dependent-features/dependent-features-read-model-type.ts
@@ -1,7 +1,7 @@
 import { IDependency } from '../../types';
 
 export interface IDependentFeaturesReadModel {
-    getChildren(parent: string): Promise<string[]>;
+    getChildren(parents: string[]): Promise<string[]>;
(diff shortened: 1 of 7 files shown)


src/lib/features/dependent-features/dependent-features-read-model-type.ts
```diff
@@ -1,7 +1,7 @@
 import { IDependency } from '../../types';
 
 export interface IDependentFeaturesReadModel {
-    getChildren(parent: string): Promise<string[]>;
+    getChildren(parents: string[]): Promise<string[]>;
```
kwasniew, line 4: changing this signature opens options for bulk operations and also works for a single parent

FredrikOseberg (approved):
LGTM

PR #4914, opened 2023-10-03
chore: automate changelog generation on release branch

gastonfournier:
## About the changes
This generates the changelog in the branch taking into account latest release (it's conditional to wether the release ends with .0 or not)

### DIFF
--- a/.github/workflows/publish-new-version.yaml
+++ b/.github/workflows/publish-new-version.yaml
@@ -22,7 +22,7 @@ jobs:
         node-version: [18.x]
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Set up git-cliff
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-new-version.yaml
```diff
@@ -21,7 +24,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          token: ${{ secrets.GH_PUSH_TOKEN }}
```
gastonfournier, line 24: Hopefully with permissions: contents: write we don't need this anymore

chriswk:
Having tested this, it seems to work

chriswk (approved):
LGTM. But I was part of the work so we might want to wait for another approval

gastonfournier:
> LGTM. But I was part of the work so we might want to wait for another approval

We'll test it out with @andreas-unleash in 5.5.4 patch release

PR #4915, opened 2023-10-03
feat: change project with feature dependencies

kwasniew:
## About the changes

UI and backend prevent changing a project that has either parent or child dependencies:
Screenshot 2023-10-03 at 16 21 44


FredrikOseberg (approved):
LGTM

PR #4916, opened 2023-10-04
chore: revamp transactional impl

gastonfournier:
## About the changes
This transactional implementation decorates a service with a transactional method that removes the need to start transactions in the method using the service. 

This is a gradual rollout with a feature toggle, just because transactions are not easy.


src/lib/features/export-import-toggles/createExportImportService.ts
```diff
@@ -127,109 +127,121 @@ export const createFakeExportImportTogglesService = (
     return exportImportService;
 };
 
-export const createExportImportTogglesService = (
-    db: Db,
+export const unboundExportImportTogglesService = (
```
kwasniew, line 130: the unbound is a little bit unclear to me. maybe lezy/deferred?

src/lib/__snapshots__/create-config.test.ts.snap
```diff
@@ -100,6 +100,7 @@ exports[`should create default config 1`] = `
       },
       "migrationLock": true,
       "multipleRoles": false,
+      "newTransactional": false,
```
kwasniew, line 103: I like to be precise in flag names. maybe transactionalDecorator?

src/lib/features/export-import-toggles/createExportImportService.ts
gastonfournier: I like deferred :+1:

src/lib/__snapshots__/create-config.test.ts.snap
gastonfournier: absolutely!

src/lib/features/export-import-toggles/createExportImportService.ts
gastonfournier: aad3caf

src/lib/__snapshots__/create-config.test.ts.snap
gastonfournier: [f61c572](https://github.com/Unleash/unleash/pull/4916/commits/f61c5722c87dfc2bafc741be9ce994e352c13aed)

PR #4918, opened 2023-10-04
feat: copy feature with parent

kwasniew:
## About the changes

When copying a feature with parent dependencies they are copied over.

Also refactored our usage of composition roots in tests to significantly reduce the size of the tests setup.


FredrikOseberg (approved):
LGTM

PR #4929, opened 2023-10-05
fix: only delete SSO-synced group membership where membership was added by SSO sync

daveleek:
## About the changes

Fixes an issue where SSO group sync would delete a syncable group that a user was manually added to

## Discussion points

Is this the longterm fix for this? Or would we want another column in the mapping table for future-proofing this?


chriswk (approved):
Is the group already tagged with `created_by: SSO` ?

daveleek:
> Is the group already tagged with `created_by: SSO` ?

@chriswk Yeah it has been that way for over 10 months. created_by on group_users gets set as SSO during sync, just that the unit tests didn't so had to update them too

PR #4933, opened 2023-10-05
fix: make cypress list length checks more relaxed

andreas-unleash:
This PR fixes the overview.spec by relaxing the expectation on row count.
This expectation does not seem relevant enough for some tests


gastonfournier (approved):
Looks good to me. It's a shared instance so I believe this is fine

PR #4937, opened 2023-10-05
fix: Remove the info from the variants page

andreas-unleash:
Remove the info from the variants page

Closes # [UNL-231](https://linear.app/unleash/issue/UNL-231/remove-the-info-from-the-variants-page)
Screenshot 2023-10-05 at 15 11 33


daveleek (approved):
LGTM!

PR #4938, opened 2023-10-05
refactor: custom render should provide container

kwasniew:
## About the changes

Move boilerplate provider to the render method.

`<UIProviderContainer>` is around all your tested component so you don't have to type it yourself.

This is 2 LOC change + updated all tests that used it.


frontend/src/utils/testRenderer.tsx
```diff
@@ -21,15 +22,17 @@ export const render = (
...
-        <SWRConfig value={{ provider: () => new Map() }}>
-            <AccessProviderMock permissions={permissions}>
-                <ThemeProvider>
-                    <AnnouncerProvider>
-                        <BrowserRouter>{children}</BrowserRouter>
-                    </AnnouncerProvider>
-                </ThemeProvider>
-            </AccessProviderMock>
-        </SWRConfig>
+        <UIProviderContainer>
```
kwasniew, line 25: this is where we moved it

Tymek (approved):

PR #4939, opened 2023-10-05
Fix/last seen at by environment

FredrikOseberg:
This is a draft of the new architecture for last seen at. Apologies for size, but I initially explored triggers and couldn't merge that before it was done. I changed my approach mid-way and now I'd like to verify the direction and merge before doing the remaining endpoints.

## Discussion
Today there are 5 pages accessing this information:
* Project health
* Project overview (DONE)
* Feature overview
* Feature toggle list
* Archived features

The last seen at property is in it's current form being added to the data structure of the environments property that exists on the feature data structure on some of these views. Since some of the views don't have it, its implementation varies across views. I have chosen not to change the data structure in this iteration, which has influenced the design decisions of reading the data.

It might it would be better if the data was added as an additional field directly on the feature data structure itself, only for the read operations in the admin panel: 

```
{ name: 'my-feature ..., lastSeenAtPerEnvironment: [ { 'development': { 'lastSeenAt': '2023-10-03 13:08:16.263,'  } } ] }
```

Other discussion points:
* Went with batch inserts instead of triggers. Reasons:
  - Can be feature flagged
  - Triggers run on every row, and application will exponentially increase the amount of rows we need to insert LastSeenAt because it's a dimension we don't care about in the new table
  - Would have to shut down releasing to production while testing trigger performance
 * Foreign key on environments to maintain data integrity when deleting environment
 
 Follow up actions
 * Implement remaining endpoints
 * Increase test coverage
 * Land on data structure
 * Test with load
 * Add job to clean up (IE: Feature flags that do not exist)


src/lib/services/client-metrics/last-seen/createLastSeenService.ts
```diff
@@ -0,0 +1,34 @@
+import FakeFeatureToggleStore from '../../../../test/fixtures/fake-feature-toggle-store';
+import FeatureToggleStore from '../../../db/feature-toggle-store';
+import { Db, IUnleashConfig } from '../../../server-impl';
+import { FakeLastSeenStore } from './fake-last-seen-store';
+import { LastSeenService } from './last-seen-service';
+import LastSeenStore from './last-seen-store';
+
+export const createLastSeenService = (
```
FredrikOseberg, line 8: Use composite root pattern for LastSeenService

src/lib/services/client-metrics/last-seen/last-seen-mapper.ts
```diff
@@ -0,0 +1,40 @@
+import { Logger } from '../../../logger';
+import { IFeatureOverview } from '../../../types';
+import { IFeatureLastSeenResults } from './last-seen-read-model';
+
+export class LastSeenMapper {
+    mapToFeatures(
```
FredrikOseberg, line 6: Map results of reading into environments of incoming features to maintain todays data structure, might change in follow up iteration

src/lib/services/client-metrics/last-seen/last-seen-read-model.ts
```diff
@@ -0,0 +1,41 @@
...
+const TABLE = 'last_seen_at_metrics';
+
+export interface IFeatureLastSeenResults {
+    [featureName: string]: {
+        [environment: string]: {
+            lastSeen: string;
+        };
+    };
+}
+export class LastSeenAtReadModel implements ILastSeenReadModel {
```
FredrikOseberg, line 13: retrieve last seen at results per environment for an array of features

src/lib/services/client-metrics/last-seen/last-seen-service.ts
```diff
@@ -42,7 +51,12 @@ export class LastSeenService {
             this.logger.debug(
                 `Updating last seen for ${lastSeenToggles.length} toggles`,
             );
-            await this.featureToggleStore.setLastSeen(lastSeenToggles);
+
+            if (this.config.flagResolver.isEnabled('useLastSeenRefactor')) {
+                await this.lastSeenStore.setLastSeen(lastSeenToggles);
+            } else {
+                await this.featureToggleStore.setLastSeen(lastSeenToggles);
+            }
```
FredrikOseberg, line 59: maintain previous implementation

src/lib/services/project-service.ts
```diff
@@ -1069,6 +1075,24 @@ export default class ProjectService {
...
+                await this.lastSeenReadModel.getForFeature(featureNames);
+
+            console.log(decoratedFeatures, lastSeenAtPerEnvironment);
+
+            decoratedFeatures = mapper.mapToFeatures(
+                decoratedFeatures,
+                lastSeenAtPerEnvironment,
+                this.logger,
+            );
+        }
```
FredrikOseberg, line 1094: Decorate features with lastSeenData

src/migrations/20231003113443-last-seen-at-metrics-table.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+      CREATE TABLE last_seen_at_metrics (
```
FredrikOseberg, line 6: new table, foreign key on environments

andreas-unleash:
Looks good.  Only suggestion is we could add the batch size as a variant in the flag.  Design decisions make sense to me

src/lib/services/client-metrics/last-seen/last-seen-store.ts
```diff
@@ -0,0 +1,75 @@
...
+                return {
+                    feature_name: item.featureName,
+                    environment: item.environment,
+                    last_seen_at: now,
+                };
+            });
+
+            const batchSize = 1000;
+
+            await this.db.transaction(async (trx) => {
```
kwasniew, line 58: I'd do the transaction around the service (in the controller) as we do for all new code.

kwasniew: Since it's a non critical data I'm wondering if having transactions even makes sense here. I'd consider sacrificing correctness for performance.

src/lib/services/client-metrics/last-seen/last-seen-service.ts
```diff
@@ -42,7 +51,12 @@ export class LastSeenService {
             this.logger.debug(
                 `Updating last seen for ${lastSeenToggles.length} toggles`,
             );
-            await this.featureToggleStore.setLastSeen(lastSeenToggles);
+
+            if (this.config.flagResolver.isEnabled('useLastSeenRefactor')) {
+                await this.lastSeenStore.setLastSeen(lastSeenToggles);
```
kwasniew, line 56: is it critical that we wait on the last seen being saved or can we return 202 Accepted and process it in the background

src/lib/services/client-metrics/last-seen/last-seen-service.ts
```diff
@@ -17,17 +17,26 @@ export class LastSeenService {
...
     ) {
+        this.lastSeenStore = lastSeenStore;
         this.featureToggleStore = featureToggleStore;
         this.logger = config.getLogger(
             '/services/client-metrics/last-seen-service.ts',
         );
+        this.config = config;
 
         this.timers.push(
             setInterval(() => this.store(), lastSeenInterval).unref(),
```
kwasniew, line 42: shouldn't this be part of the scheduler?

FredrikOseberg: I will look into separating this in a follow-up

src/lib/services/client-metrics/last-seen/last-seen-store.ts
FredrikOseberg: Agree, done

src/migrations/20231003113443-last-seen-at-metrics-table.js
```diff
@@ -0,0 +1,22 @@
...
+    );
+
+      CREATE INDEX idx_feature_name
+      ON last_seen_at_metrics (feature_name);`,
+        callback(),
+    );
+};
+
+exports.down = function (db, callback) {
+    callback();
```
kwasniew, line 21: missing down migration

kwasniew (approved):
I'd run clinic.js with both last seen variants to compare the performance characteristics

Issue #4912, opened 2023-10-03
[Slack App] Notifications in Slack don't mention segment constraint


mathroc:
### Describe the bug

A feature flag was enabled for a few user IDs and someone change it to have a segment constraint. A notification was pushed to Slack to inform us that the feature flag was updated that said:

> [Someone] updated [MyFeatureFlag](https://unleash..../projects/my-project/features/MyFeatureFlag) in project my-project by updating strategy flexibleRollout in production constraints from [userId is one of (x, y, z)] to empty set of constraints

Reading the message make it sounds like it's now enabled for everyone

### Steps to reproduce the bug

1. Add a Slack integration
2. Enabled a feature flag with some constraint on production
3. Remove the manual constraints and add a segment constraint

### Expected behavior

The notification on Slack should be something like:

> [Someone] updated [MyFeatureFlag](https://unleash..../projects/my-project/features/MyFeatureFlag) in project my-project by updating strategy flexibleRollout in production constraints from [userId is one of (x, y, z)] to [segment: my-segment]

### Unleash version

Unleash 5.4.1

### Subscription type

Don't know

### Hosting type

Self-hosted

daveleek:
Thank you for the bug report @mathroc, we'll have a look and get back to you

daveleek:
@mathroc apologies for late reply. We have looked at this and made an improvement, it's merged and available in Open Source Unleash right now. it will be part of the next Unleash Enterprise version release.

It will list changes in referenced segments list with lists of segment ids in slack messages, similar to how userids are represented

mathroc:
that's great, thank you :pray:

PR #4950, opened 2023-10-06
feat: render segments changes in feature strategy update event messages

daveleek:
## About the changes

Segment changes in predata and data columns were both showing the new segments list

Adds formatting of what's changed with segments to feature strategy update events, so when a user changes the strategy from using constraints, to using segments instead, it's communicated in event updates

results in: 

admin updated [sample-toggle](http://localhost/projects/default/features/sample-toggle) in project [default](http://localhost/projects/default) by updating strategy Sample Strategy in development constraints from [userId is one of (1,2,3)] to empty set of constraints; segments from empty set of segments to (1)

Closes #
#4912 

### Important files

- `src/lib/services/feature-toggle-service.ts` - Segment changes in preData and data
- `src/lib/addons/feature-event-formatter-md.ts` - Formatting segments

## Discussion points

This is an SR least effort PR - we should plan a task where we look at how to render this list of segments in a more comprehensible way (it's just rendering ids now)

### DIFF
--- a/src/lib/addons/feature-event-formatter-md.ts
+++ b/src/lib/addons/feature-event-formatter-md.ts
@@ -415,7 +415,11 @@ export class FeatureEventFormatterMd implements FeatureEventFormatter {
             preData?.constraints,
             data?.constraints,
         );
-        const strategySpecificText = [usersText, constraintText]
+        const segmentsText = this.segmentsChangeText(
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
LGTM! I'm worried about very large segments, but let's leave that for later

PR #4952, opened 2023-10-06
feat: allow to delete dependencies when no orphans

kwasniew:
## About the changes

When bulk archiving/deleting parent and child and no other features depend on the parent then we should allow to archive/delete. However when we detect that archiving/deleting a parent would leave some orphan children then we don't want to allow it.


FredrikOseberg (approved):
LGTM

PR #4953, opened 2023-10-06
chore: handle transactions already started at the controller layer

gastonfournier:
## About the changes
This PR adds a method to safeguard us from opening a new transaction while inside another transaction, resulting in two isolated transactions that will not be atomic (if one fails, the other might still complete successfully).

https://github.com/knex/knex/blob/bbbe4d4637b3838e4a297a457460cd2c76a700d5/lib/knex-builder/make-knex.js#L143C5-L144C88

We're currently opening transactions at the controller layer 
https://github.com/Unleash/unleash/blob/2746bd151766f8afbbaa2f640e8ebee6f4f98086/src/lib/features/export-import-toggles/export-import-controller.ts#L206-L208

but in some other places, we do it at the store level:
https://github.com/Unleash/unleash/blob/2746bd151766f8afbbaa2f640e8ebee6f4f98086/src/lib/db/access-store.ts#L577

## Alternative
We can remove store-level transactions and move them to the controller following this approach:

https://github.com/Unleash/unleash/blob/cb034976b93abc799df774858d716a49f645d669/src/lib/services/index.ts#L282-L284

https://github.com/Unleash/unleash/blob/cb034976b93abc799df774858d716a49f645d669/src/lib/features/export-import-toggles/export-import-controller.ts#L206-L208

This option is more expensive because we have to:
1. Write the factory methods that propagate the transaction to the stores (therefore creating the store factory methods as well)
2. Identify the methods for creating the transactions at the store level and backtrack the calls until the controller layer


src/lib/types/option.ts
```diff
@@ -104,7 +104,7 @@ export interface IUnleashOptions {
     versionCheck?: Partial<IVersionOption>;
     telemetry?: boolean;
     authentication?: Partial<IAuthOption>;
-    ui?: object;
+    ui?: IUIConfig;
```
gastonfournier, line 107: This is nice to have... probably I can remove it from this PR

kwasniew (approved):
Very elegant solution. I'd like us to handle transactions around the service boundaries but it's great for the transitional period.

gastonfournier:
> Very elegant solution. I'd like us to handle transactions around the service boundaries but it's great for the transitional period.

Yes, I've added a comment in the method we're using

src/lib/db/transaction.ts
```diff
@@ -26,13 +26,39 @@ export type WithTransactional<S> = S & {
     transactional: <R>(fn: (service: S) => R) => Promise<R>;
 };
 
+/**
+ * @deprecated this is a temporal solution to deal with transactions at the store level.
```
chriswk, line 30:
```suggestion
 * @deprecated this is a temporary solution to deal with transactions at the store level.
```

gastonfournier: Added here: https://github.com/Unleash/unleash/pull/4959

PR #4956, opened 2023-10-06
fix: Fail when format or lint is incorrect

gastonfournier:
## About the changes
Format check was not throwing a failure when checking: https://github.com/Unleash/unleash/actions/runs/6419038883/job/17428046648#step:6:10


FredrikOseberg (approved):
LGTM

PR #4958, opened 2023-10-06
feat: feature changes counted in new table

chriswk:
As part of more telemetry on the usage of Unleash. 

This PR adds a new `stat_` prefixed table as well as a trigger on the events table trigger on each insert to increment a counter per environment per day.

The trigger will trigger on every insert into the events base, but will filter and only increment the counter for events that actually have the environment set. (there are events, like user-created, that does not relate to a specific environment).

Bit wary on this, but since we truncate down to row per (day, environment) combo, finding conflict and incrementing shouldn't take too long here.

@ivarconr was it something like this you were considering?

### DIFF
--- a/src/lib/features/instance-stats/getProductionChanges.ts
+++ b/src/lib/features/instance-stats/getProductionChanges.ts
@@ -12,19 +12,22 @@ export const createGetProductionChanges =
         const productionChanges = await db
             .select({
                 last_month: db.raw(
-                    "SUM(DISTINCT CASE WHEN day > NOW() - INTERVAL '1 month' THEN updates END)",
+                    "SUM(CASE WHEN day > NOW() - INTERVAL '1 month' THEN updates END)",
                 ),
                 last_two_months: db.raw(
-                    "SUM(DISTINCT CASE WHEN day > NOW() - INTERVAL '2 months' THEN updates END)",
+                    "SUM(CASE WHEN day > NOW() - INTERVAL '60 days' THEN updates END)",
                 ),
                 last_quarter: db.raw(
-                    "SUM(DISTINCT CASE WHEN day > NOW() - INTERVAL '3 months' THEN updates END)",
+                    "SUM(CASE WHEN day > NOW() - INTERVAL '90 days' THEN updates END)",
                 ),
             })
             .from('stat_environment_updates');
         return {
             last30: parseInt(productionChanges?.[0]?.last_month || '0', 10),
-            last60: parseInt(productionChanges?.[0]?.last_month || '0', 10),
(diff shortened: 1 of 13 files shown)


src/lib/features/instance-stats/getProductionChanges.ts
```diff
@@ -0,0 +1,44 @@
...
+    last90: number;
+}>;
+
+export const createGetProductionChanges =
+    (db: Db): GetProductionChanges =>
+    async () => {
+        const productionChanges = await db
+            .select({
+                last_month: db.raw(
+                    "SUM(CASE WHEN day > NOW() - INTERVAL '1 month' THEN updates END)",
```
chriswk, line 16: Should we change this to 30 days as well, to make it consistent with the other two cases?

src/lib/services/version-service.ts
```diff
@@ -194,25 +216,29 @@ export default class VersionService {
...
-                    headers: { 'Content-Type': 'application/json' },
-                });
-                if (res.ok) {
-                    const data = (await res.json()) as IVersionResponse;
-                    this.latest = {
-                        oss: data.versions.oss,
-                        enterprise: data.versions.enterprise,
-                    };
-                    this.isLatest = data.latest;
+                if (this.versionCheckUrl) {
```
gastonfournier, line 222: I think this change is unrelated but it's worth having it independently of this PR

gastonfournier (approved):
This looks good, I feel we're probably close to considering CDC, but I don't think that'll help with self-hosted. :+1:

ivarconr:
> @ivarconr was it something like this you were considering?

yes, this looks spot on! We might peek at environment type in the future when taking out the counters, but that can be improved later.

src/lib/features/instance-stats/getProductionChanges.ts
```diff
@@ -0,0 +1,44 @@
...
+                    "SUM(CASE WHEN day > NOW() - INTERVAL '30 days' THEN updates END)",
+                ),
+                last_two_months: db.raw(
+                    "SUM(CASE WHEN day > NOW() - INTERVAL '60 days' THEN updates END)",
+                ),
+                last_quarter: db.raw(
+                    "SUM(CASE WHEN day > NOW() - INTERVAL '90 days' THEN updates END)",
+                ),
+            })
+            .from('stat_environment_updates');
```
ivarconr, line 25: I would expect a where clause here, to only pick production envs

chriswk: Totally agree. thanks for the reminder

chriswk: I also updated it to look at the environment type and sum across all environments of type 'production'.

PR #4959, opened 2023-10-06
chore: Improve UI Config type

gastonfournier:
Fix link types

### DIFF
--- a/src/lib/types/option.ts
+++ b/src/lib/types/option.ts
@@ -148,14 +148,12 @@ export interface IUIConfig {
     environment?: string;
     slogan?: string;
     name?: string;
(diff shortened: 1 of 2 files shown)


src/lib/types/option.ts
```diff
@@ -148,14 +148,12 @@ export interface IUIConfig {
...
-            href: string;
-            title: string;
-        },
-    ];
+    links?: {
+        value: string;
+        icon?: string;
+        href: string;
+        title: string;
+    }[];
```
gastonfournier, line 156: Previous type was a single element array, while the new type is an array of objects

kwasniew (approved):

PR #4960, opened 2023-10-07
fix: typo in enabled event

nunogois:
Fixes a typo where we would send `disabled` for `enabled` events.

https://unleash-community.slack.com/archives/C03GWTN7XMG/p1696631381409369


gastonfournier (approved):
G!

PR #4964, opened 2023-10-09
Fix: copy feature alert when change requests enabled in any env

andreas-unleash:
Adds an alert if change requests are enabled in any env and you try to copy a feature.

Currently the functionality was disabled but with no explanation

Closes # [SR-79](https://linear.app/unleash/issue/SR-79/when-change-request-is-enable-the-copy-functionality-its-disabled-but)


sighphyre (approved):
LGTM

PR #4969, opened 2023-10-09
chore: Rename validate step

gastonfournier:
This is to bypass branch protection rules

### DIFF
--- a/.github/workflows/build_doc_prs.yaml
+++ b/.github/workflows/build_doc_prs.yaml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   validate:
+    name: build (18.x) # temporary solution to trick branch protection rules
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


thomasheartman (approved):

thomasheartman:
So for clarity (and as a little summary):

Gaston and I just spent a little time looking at this, and it appears that all jobs with a name that matches the name of a required build step must pass before a PR gets merged.

For that reason, renaming the docs' validate step to `build (18.x)` makes the step required, which means doc builds now need to pass before being merged (assuming there are doc changes). This should fix issues such as [renovate bot merging stuff that breaks the docs](https://github.com/Unleash/unleash/pull/4895) going forward.

As a potential future improvement here, we might want to change the name of the required step to be just "build", so that it's more generic. 

Doing this would require us to change the build matrix such that the name of the build step is not `build (18.x)` but instead just `build`.

PR #4971, opened 2023-10-09
feat: do not allow to manage dependencies directly with cr enabled

kwasniew:
## About the changes

When Change Requests are enabled do not allow to add and remove dependencies directly. 

It will be tested as part of the Change Request flow.

This will be a breaking change in change requests since they need to switch to unprotected method calls.

### DIFF
--- a/frontend/src/hooks/api/actions/useChangeRequestApi/useChangeRequestApi.ts
+++ b/frontend/src/hooks/api/actions/useChangeRequestApi/useChangeRequestApi.ts
@@ -13,7 +13,7 @@ export interface IChangeSchema {
         | 'archiveFeature'
         | 'updateSegment'
         | 'addDependency'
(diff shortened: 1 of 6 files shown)


frontend/src/hooks/api/actions/useChangeRequestApi/useChangeRequestApi.ts
```diff
@@ -13,7 +13,7 @@ export interface IChangeSchema {
         | 'archiveFeature'
         | 'updateSegment'
         | 'addDependency'
-        | 'deleteDependencies';
+        | 'deleteDependency';
```
kwasniew, line 16: rename to method actually used by backend code. I'd like those types to be taken from orval to prevent this type of errors

FredrikOseberg (approved):
LGTM

PR #4976, opened 2023-10-10
meta: add note to generate openapi docs before starting local dev

thomasheartman:
This PR adds a small note to generate docs before running the docs locally.


dgorton (approved):
Thanks!

PR #4980, opened 2023-10-10
docs: make videos bigger

thomasheartman:
This PR changes the `VideoContent` component to:
- remove the extra text; keeps only videos
- makes videos take up the full article width
- multiple videos stack vertically (this may or may not be the best way to handle it, but we don't have any instances of using multiple videos as of right now, so we shouldn't touch this until we do).

By chance, it also removes a lot of trailing whitespace in files. I suggest checking out the diff with whitespace hidden.

Before (single video):
image

Before (if there were multiple videos): 
image

After (single video):
image

After (if there are multiple videos): 
image


sebastian-bury (approved):
looks good to me!

PR #4982, opened 2023-10-10
chore: avoid building frontend if not needed

gastonfournier:
## About the changes
Looks like we're building twice: https://github.com/Unleash/unleash/actions/runs/6468079228/job/17559391393#step:5:136


sighphyre (approved):
G

PR #4987, opened 2023-10-10
feat: prevent adding dependency to archived or removed parent

kwasniew:
## About the changes

Prevents adding a dependency to a parent that was removed or archived or doesn't exist.

## Discussion points

I'm not sure if we should fail those cases or just ignore. In the context of a change request we may end up in a situation where parent was deleted as the CR review was pending. So maybe it's better to show warning in CR and proceed without taking any action in the dependency service itself.

### DIFF
--- a/src/lib/features/feature-toggle/features-read-model-type.ts
+++ b/src/lib/features/feature-toggle/features-read-model-type.ts
@@ -1,0 +1,3 @@
+export interface IFeaturesReadModel {
+    featureExists(parent: string): Promise<boolean>;
+}
--- a/src/lib/features/feature-toggle/features-read-model.ts
+++ b/src/lib/features/feature-toggle/features-read-model.ts
@@ -1,0 +1,19 @@
+import { Db } from '../../db/db';
+import { IFeaturesReadModel } from './features-read-model-type';
+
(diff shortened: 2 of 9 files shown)


src/lib/features/feature-toggle/features-read-model-type.ts
```diff
@@ -0,0 +1,3 @@
+export interface IFeaturesReadModel {
```
kwasniew, line 1: introduced new read model in feature-toggle submodule so that we don't have to depend on feature-toggle-service with all write model operations.

src/lib/features/feature-toggle/features-read-model.ts
```diff
@@ -0,0 +1,19 @@
...
+        this.db = db;
+    }
+
+    async featureExists(parent: string): Promise<boolean> {
+        const rows = await this.db('features')
+            .where('name', parent)
+            .andWhere('archived_at', null)
+            .select('name');
+
+        return rows.length > 0;
```
kwasniew, line 17: I find it more readable than inline `select exists`

FredrikOseberg (approved):
LGTM. Perhaps we warn you AND throw an error on the change request? Or we could not let you apply the change request before you deal with the conflicts.

PR #4993, opened 2023-10-10
docs: updated sidebars and added missing doc ID

markunl:
Updated `sidebars.js` to incorporate `Getting Started` as top level doc, with Unleash Academy in child structure (order as discussed with @dgorton ). See screenshot.

Also added missing `id` to `getting-started.md`

image


dgorton (approved):
Looking good!

PR #4994, opened 2023-10-11
Refactor/move features to feature oriented architecture

FredrikOseberg:
This PR gathers feature related files in the same folder.

Discussion points: 

* I left out variants controller, but it's also related to features. Should we move it into the same directory? (Legacy variants per environment).


kwasniew (approved):
Finally someone brave enough to touch features :) 
I wouldn't worry about the remaining controller. we can move it any time

PR #4995, opened 2023-10-11
feat: re-order message banners

nunogois:
https://linear.app/unleash/issue/2-1494/re-order-message-banners

- Re-orders message banners to fit into this logic:

>1. Maintenance banner
>2. External message banner(s) - Most likely coming from Unleash
>3. Internal message banner(s)

- Renames the feature flag to better reflect the feature behavior;
- Lays a basic skeleton structure for this new feature;


daveleek (approved):
LGTM!

PR #4999, opened 2023-10-11
fix: disable all environments when reviving a feature

andreas-unleash:
Disable all environments when reviving a feature

Closes # [SR-93](https://linear.app/unleash/issue/SR-93/disable-all-environments-when-reviving-a-feature)


src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -362,6 +362,11 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             .where({ name })
             .update({ archived_at: null })
             .returning(FEATURE_COLUMNS);
+
+        await this.db(FEATURE_ENVIRONMENTS_TABLE)
```
kwasniew, line 366: I think this should be part of the archive method, not revive.
Also I'm not sure whether it should be done at the store level or service level.
Finally since we have write operations on the write model we need to wrap the whole service method in a transaction. If you decide to do it as part of archive and bulk archive then I'd wrap it in a tx in a controller doing the archive.

kwasniew (changes requested):

src/lib/features/feature-toggle/feature-toggle-store.ts
andreas-unleash: The problem right now with doing it at archive is that already archived features will not get it - We could have a migration that does disables the environments for already archived features, but I would rather not.

andreas-unleash: Good point on the transaction wrapping

kwasniew: so maybe both? in the long term it makes more sense to me to do it as part of archive. We'll probably also clean the dependencies in the archive so having one "cleanup" location would be nice. 
what's the problem with a db migration?

andreas-unleash:
Moved the functionality to the service layer and wrapped in `disableAllEnvsOnRevive` flag

src/test/e2e/api/admin/feature-archive.e2e.test.ts
```diff
@@ -214,7 +219,11 @@ test('can bulk revive features', async () => {
     for (const feature of features) {
         await app.request
             .get(`/api/admin/projects/default/features/${feature}`)
-            .expect(200);
+            .expect(200)
+            .expect((res) => {
+                const feature = res.body;
+                return feature.environments.every((env) => !env.enabled);
+            });
```
kwasniew, line 226: I'd follow established pattern and do 
```
const {body} = await.app().expect(200);
expect(expectationHere)
```

src/test/e2e/api/admin/archive.test.ts
```diff
@@ -142,6 +147,57 @@ test('Should be able to revive toggle', async () => {
...
+        .send({})
+        .expect(200);
+
+    await app.request
+        .get(
+            '/api/admin/projects/default/features/feat-proj-1?variantEnvironments=true',
+        )
+        .expect((res) => {
+            const feature = res.body;
+            return feature.environments.every((env) => !env.enabled);
```
kwasniew, line 197: same here, I'd use expect from jest instead of writing expect in supertest. I normally use supertest expect for HTTP stuff and regular expect for regular assertions since they provide better reporting capabilities.

kwasniew (approved):

src/test/e2e/api/admin/archive.test.ts
andreas-unleash: makes sense

PR #5002, opened 2023-10-11
Docs/academy video player

markunl:
Added explanative text to three courses on embedded playlist video player usage


dgorton (approved):
Approved

PR #5003, opened 2023-10-11
Protect archive feature

kwasniew:
## About the changes

Introduced protected archive/bulk archive operations that take change requests into account


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1492,12 +1513,27 @@ class FeatureToggleService {
     }
 
     async archiveToggles(
+        featureNames: string[],
+        user: User,
+        projectId: string,
+    ): Promise<void> {
+        this.stopWhenChangeRequestsEnabled(projectId, undefined, user);
```
FredrikOseberg, line 1520: missing await?

FredrikOseberg (changes requested):

FredrikOseberg (approved):

PR #5004, opened 2023-10-11
chore: split interfaces for import and export

gastonfournier:
## About the changes
This splits the interfaces for import and export, especially because the import functionality has to be replaced in enterprise repo.

This is a breaking change because of the service renames, but I'll have the PR for the other repository ready so we reduce the time to fix. I intentionally avoided doing it backward compatible because of time.

### DIFF
--- a/src/lib/types/services.ts
+++ b/src/lib/types/services.ts
@@ -39,7 +39,9 @@ import MaintenanceService from '../services/maintenance-service';
 import { AccountService } from '../services/account-service';
 import { SchedulerService } from '../services/scheduler-service';
 import { Knex } from 'knex';
-import ExportImportService from '../features/export-import-toggles/export-import-service';
+import ExportImportService, {
+    IExportService,
+} from '../features/export-import-toggles/export-import-service';
 import { ISegmentService } from '../segments/segment-service-interface';
 import ConfigurationRevisionService from '../features/feature-toggle/configuration-revision-service';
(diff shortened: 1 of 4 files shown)


src/lib/types/services.ts
```diff
@@ -89,9 +91,8 @@ export interface IUnleashServices {
     instanceStatsService: InstanceStatsService;
     favoritesService: FavoritesService;
     maintenanceService: MaintenanceService;
-    /** @deprecated prefer exportImportServiceV2, we're doing a gradual rollout */
-    exportImportService: ExportImportService;
-    exportImportServiceV2: WithTransactional<ExportImportService>;
+    exportService: IExportService;
+    importService: WithTransactional<ExportImportService>;
```
kwasniew, line 95: Why not IExportService after you split them?

kwasniew (approved):
One important comment inside

src/lib/types/services.ts
gastonfournier: I believe you mean, why not IImportService, right? Now, export doesn't have to be transactional and we can use the same and we only need import to be transactional

gastonfournier: Let me know if [6950fc1](https://github.com/Unleash/unleash/pull/5004/commits/6950fc1962255f15b0f246a1c481915f480f6e39) is what you're expecting ;)

kwasniew: correct :)

PR #5006, opened 2023-10-12
Refactor/separate client and admin store

FredrikOseberg:
This PR is the first step in separating the client and admin stores. Currently our feature toggle services uses the client store to serve multiple purposes. The architecture looks like this: 

Skjermbilde 2023-10-12 kl  08 16 07

Admin API uses the feature toggle service to serve both the feature toggle list and playground features, while the client API uses the feature toggle service to serve client features. The admin API can change often and have very different requirements than the client API, which changes infrequently and generally keeps the same stable structure for long periods of time. This architecture is error prone, because when you need to make changes to the admin API, you can very easily affect the client API.

I aim to put up a stone wall between the two APIs. Complete separation between the two APIs, at the cost of some duplication. The new architecture looks like this: 

Skjermbilde 2023-10-12 kl  08 20 37

In this PR I have created a feature oriented architecture for client features and disconnected the client API from the feature toggle service. It now goes through it's own service to it's own store. For feature toggle service I have duplicated and replaced the functionality that serves /api/admin/features, I have kept a lot of the ugliness in the code and haven't removed anything in order to avoid breaking changes. 

Next steps: 
* Move playground to admin API
* Remove client-feature-toggle-store from feature-toggle-service


src/lib/features/client-feature-toggles/client-feature-toggle-service.ts
```diff
@@ -0,0 +1,61 @@
...
+    IFeatureToggleQuery,
+    IUnleashConfig,
+    IUnleashStores,
+} from '../../types';
+
+import { Logger } from '../../logger';
+
+import { FeatureConfigurationClient } from '../feature-toggle/types/feature-toggle-strategies-store-type';
+
+export class ClientFeatureToggleService {
```
FredrikOseberg, line 13: New service that is completely disconnected from the admin API

src/lib/features/client-feature-toggles/createClientFeatureToggleService.ts
```diff
@@ -0,0 +1,45 @@
+import FeatureToggleClientStore from '../client-feature-toggles/client-feature-toggle-store';
+import { Db } from '../../db/db';
+import { IUnleashConfig } from '../../types';
+import FakeClientFeatureToggleStore from '../client-feature-toggles/fakes/fake-client-feature-toggle-store';
+import { ClientFeatureToggleService } from './client-feature-toggle-service';
+
+export const createClientFeatureToggleService = (
```
FredrikOseberg, line 7: composition root for the new service

src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -44,6 +56,81 @@ interface VariantDTO {
...
+    const newTag = rowToTag(row);
+    feature.tags = [...tags, newTag];
+};
+
+const rowToTag = (row: Record<string, any>): ITag => {
+    return {
+        value: row.tag_value,
+        type: row.tag_type,
+    };
+};
```
FredrikOseberg, line 132: Helper functions taken from client-feature-toggle-store

src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -91,6 +178,132 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             .then(this.rowToFeature);
     }
 
+    async getFeatureToggleList(
```
FredrikOseberg, line 181: Copied from feature-toggle-client-store

src/lib/features/feature-toggle/legacy/feature-toggle-legacy-controller.ts
```diff
@@ -196,7 +196,7 @@ class FeatureController extends Controller {
         namePrefix,
     }: any): Promise<IFeatureToggleQuery> {
         if (!tag && !project && !namePrefix) {
-            return null;
+            return {};
```
FredrikOseberg, line 199: Caused type error

src/test/e2e/api/client/feature.e2e.test.ts
```diff
@@ -355,3 +355,30 @@ test('returns a feature toggles impression data for a different project', async
             expect(projectToggle.impressionData).toBe(true);
         });
 });
+
+test('should return correct data structure for /api/admin/features', async () => {
```
FredrikOseberg, line 359: tests the data structure for /api/admin/features

kwasniew:
To be on the safe side I'd leave FeatureToggleClientStore as it is now.
I'd create new client store and admin store on the side with a new code.
With a flag we can direct traffic to either the old store or 2 new stores. Additionally in a transitional period we can run both and compare the results which should be exactly the same.

PR #5007, opened 2023-10-12
feat: export dependent feature toggles

sjaanus:


src/lib/features/dependent-features/dependent-features-read-model-type.ts
```diff
@@ -1,11 +1,15 @@
 import { IDependency } from '../../types';
+import { FeatureDependency } from './dependent-features';
 
 export interface IDependentFeaturesReadModel {
     getChildren(parents: string[]): Promise<string[]>;
     // given a list of parents and children verifies if some children would be orphaned after deletion
     // we're interested in the list of parents, not orphans
     getOrphanParents(parentsAndChildren: string[]): Promise<string[]>;
     getParents(child: string): Promise<IDependency[]>;
+    getFeatureDependenciesByChildren(
```
kwasniew, line 10: The naming is inconsistent with other methods. Non of the other methods uses Feature in the method name. I think you can modify getParents to getParent(children: string[]) to make it more generic without introducing a new method

src/lib/features/dependent-features/dependent-features-read-model.ts
```diff
@@ -42,6 +43,21 @@ export class DependentFeaturesReadModel implements IDependentFeaturesReadModel {
...
+    async getFeatureDependenciesByChildren(
+        children: string[],
+    ): Promise<FeatureDependency[]> {
+        const rows = await this.db('dependent_features').whereIn(
+            'child',
+            children,
+        );
+
+        return rows.map((row) => ({
+            child: row.child,
```
kwasniew, line 55: you can add this to the IDependency and use getParents method instead without growing API surface area

src/lib/features/export-import-toggles/export-import.e2e.test.ts
```diff
@@ -86,6 +94,19 @@ const createToggle = async (
     );
 };
 
+const addFeatureDependency = async (
```
kwasniew, line 97: Please use `await app.addDependency(child, parent);` instead since this code is already encapsulated there. You can check feature-toggles.e2e.test.ts to see how it's used

src/lib/features/export-import-toggles/export-import.e2e.test.ts
```diff
@@ -276,6 +307,11 @@ test('exports features', async () => {
         },
         strategy,
     );
+
+    await addFeatureDependency(defaultFeatureName, {
```
kwasniew, line 311: await app.addDependency(defaultFeatureName, 'second_feature');

src/lib/openapi/spec/feature-dependency-schema.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const featureDependencySchema = {
+    $id: '#/components/schemas/featureDependencySchema',
+    type: 'object',
+    description:
+        'Feature dependency connection between a child feature and its parents',
+    required: ['child', 'parents'],
+    additionalProperties: false,
+    properties: {
+        child: {
```
kwasniew, line 13: we don't use child and parent in the openapi. We use feature and dependencies instead. I'd stick to consistent naming.

src/lib/openapi/spec/export-result-schema.ts
```diff
@@ -166,6 +168,14 @@ export const exportResultSchema = {
                 $ref: '#/components/schemas/tagTypeSchema',
             },
         },
+        featureDependencies: {
```
kwasniew, line 171: I'd call this dependencies since it opens a way for other types of dependencies (e.g. strategy dependencies)

src/lib/features/export-import-toggles/export-import.e2e.test.ts
```diff
@@ -306,14 +342,28 @@ test('exports features', async () => {
                 name: segmentName,
             },
         ],
+        featureDependencies: [
```
kwasniew, line 345: we call it dependencies in other parts of the openapi

src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -690,6 +703,30 @@ export default class ExportImportService {
         const filteredTagTypes = tagTypes.filter((tagType) =>
             featureTags.map((tag) => tag.tagType).includes(tagType.name),
         );
+
+        const reducedFeatureDependencies = Object.values(
```
kwasniew, line 707: can we use lodash groupBy to better express the intent without having to read the algorithm? One example is in the playground-service.ts

kwasniew (changes requested):
Some comments inside

kwasniew (approved):

PR #5008, opened 2023-10-12
feat: dynamic icons by adding material symbols font

nunogois:
https://linear.app/unleash/issue/2-1502/add-support-for-custom-dynamic-icons-mui-icon-component

Adds support for custom dynamic icons by adding the [Material Symbols Outlined font](https://fonts.google.com/icons) and setting the MUI Icon component base class. See: https://mui.com/material-ui/icons/#icon-font-icons

Message banner use case: This will not only enable us to set custom icons for external message banners, but will also let users configure their desired icon from the set of options in the font.

### DIFF
--- a/frontend/src/themes/dark-theme.ts
+++ b/frontend/src/themes/dark-theme.ts
@@ -618,7 +618,6 @@ export default createTheme({
             },
         },
 
-        // Default Icon to Material Symbols Outlined font
         MuiIcon: {
             defaultProps: {
                 baseClassName: 'material-symbols-outlined',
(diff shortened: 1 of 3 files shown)


frontend/src/themes/dark-theme.ts
```diff
@@ -617,5 +617,12 @@ export default createTheme({
                 }),
             },
         },
+
+        // Default Icon to Material Symbols Outlined font
```
sighphyre, line 621: Very personal opinion but this comment doesn't help me much. The code suggests this is a default prop for `material-symbols-outlined`, the comment just reiterates that

Totally fine for it to stick around

sighphyre (approved):
G. Optional comment around comments

PR #5009, opened 2023-10-12
feat: message banners table migration

nunogois:
https://linear.app/unleash/issue/2-1485/db-create-migration-for-a-new-internal-message-banners-table

Adds a DB migration for a new `message_banners` table.

### DIFF
--- a/src/migrations/20231012082537-message-banners.js
+++ b/src/migrations/20231012082537-message-banners.js
@@ -1,0 +1,32 @@
+'use strict';
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS message_banners
+            (
+                id SERIAL PRIMARY KEY NOT NULL,
+                enabled BOOLEAN DEFAULT true NOT NULL,
+                message TEXT NOT NULL,
+                variant TEXT,
+                sticky BOOLEAN DEFAULT false,
+                icon TEXT,
+                link TEXT,
+                link_text TEXT,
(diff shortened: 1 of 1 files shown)


src/migrations/20231012082537-message-banners.js
```diff
@@ -0,0 +1,32 @@
...
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS message_banners
+            (
+                id SERIAL PRIMARY KEY NOT NULL,
+                enabled BOOLEAN DEFAULT true NOT NULL,
+                message TEXT NOT NULL,
+                variant TEXT,
```
daveleek, line 11: Is this meant as type of message? Like warning, info, error etc?

daveleek (approved):
LGTM!

src/migrations/20231012082537-message-banners.js
nunogois: Yep, exactly! Do you think we should have a different type here?

daveleek: I was just pondering the use case, as the name was so close to a part of our domain (variants and features/toggles). We could call it message_type or banner_type? Just suggestions though

nunogois: That's a fair point 🤔 I think I prefer to proceed like this to ensure existing external message banner payloads don't break, but it's something I'll take into account and get back to if we identify some confusion.

PR #5013, opened 2023-10-12
fix: feature flag playground features in new store

FredrikOseberg:
Have playground use the method in the feature-toggle-service instead of asking the client-feature-toggle-store


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1049,9 +1050,16 @@ class FeatureToggleService {
...
     ): Promise<FeatureConfigurationClient[]> {
-        const result = await this.clientFeatureToggleStore.getPlayground(
+        let result = await this.clientFeatureToggleStore.getPlayground(
             query || {},
         );
+
+        if (this.flagResolver.isEnabled('separateAdminClientApi')) {
+            result = (await this.getFeatureToggles(
+                query,
+            )) as IFeatureToggleClient[];
```
FredrikOseberg, line 1060: will clean this up when we remove the feature flag

kwasniew: shouldn't we do one query or the other? Here we're fetching getPlayground first and then override it with another query

FredrikOseberg: If we wanted to optimise this maximally, we could do that. But since this is the admin API and this feature flag should be removed asap to avoid holding all this context, I think it's fine.

kwasniew (approved):
So maybe run both in parallel (Promise.all) and console.warn when the results are different. Then we'll have extra value from those 2 queries.

src/lib/features/feature-toggle/types/feature-toggle-store-type.ts
```diff
@@ -36,6 +37,10 @@ export interface IFeatureToggleStore extends Store<FeatureToggle, string> {
         userId?: number,
         archived?: boolean,
     ): Promise<FeatureToggle[]>;
+    getPlaygroundFeatures(
+        dependentFeaturesEnabled: boolean,
```
kwasniew, line 41: why are we passing it as args instead of passing flag in a constructor as we did before in the previous client store?

src/lib/features/feature-toggle/feature-toggle-utils.ts
```diff
@@ -0,0 +1,279 @@
...
+};
+
+interface Difference {
+    index: (string | number)[];
+    reason: string;
+    valueA: any;
+    valueB: any;
+}
+
+export function deepDiff(arr1: any[], arr2: any[]): Difference[] | null {
```
kwasniew, line 221: I don't see this function in the old code. Is it new?

kwasniew: does it have tests? what's the advantage over using lodash isEqual?

FredrikOseberg: I should have noted that I plan to follow up with some more test coverage in the next PR. The advantage is that it will actually list the two objects that differ, so that you can see what is the difference instead of just getting a boolean value.

src/lib/features/feature-toggle/types/feature-toggle-store-type.ts
FredrikOseberg: I didn't want to introduce it in the feature toggle store, because the feature toggle store does not currently receive the config. I decided to go this route for now, since I expect the feature flag to be removed shortly.

PR #5016, opened 2023-10-12
feat: make maintenance banner sticky

nunogois:
After checking with @nicolaesocaciu we came to the conclusion that the maintenance banner should be sticky.

Co-authored-by Nicolae <nicolae@getunleash.ai>


nicolaesocaciu (approved):
LGTM

PR #5017, opened 2023-10-12
fix: import segment test and fix

sjaanus:
Now OSS can import segments properly. Also updated tests.


kwasniew (approved):

kwasniew:
@sjaanus have you moved validation for segments already?

sjaanus:
> @sjaanus have you moved validation for segments already?

Yes, you approved it yesterday.

PR #5018, opened 2023-10-13
feat: make maintenance-related 503s more intuitive

nunogois:
This makes maintenance-related 503s more intuitive on our UI by mentioning that maintenance banner is currently enabled.

image

### DIFF
--- a/frontend/src/hooks/api/actions/useApi/useApi.ts
+++ b/frontend/src/hooks/api/actions/useApi/useApi.ts
@@ -17,7 +17,6 @@ import {
 } from 'utils/apiUtils';
 import { formatApiPath } from 'utils/formatPath';
 import { ACCESS_DENIED_TEXT } from 'utils/formatAccessText';
-import { useMaintenance } from 'hooks/api/getters/useMaintenance/useMaintenance';
 
 type ApiErrorHandler = (
     setErrors: Dispatch<SetStateAction<{}>>,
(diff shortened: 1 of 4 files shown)


frontend/src/hooks/api/actions/useApi/useApi.ts
```diff
@@ -104,6 +110,28 @@ const useAPI = ({
...
+                    setErrors((prev) => ({
+                        ...prev,
+                        unavailable: 'This operation is unavailable',
+                    }));
+                }
+
+                if (propagateErrors) {
+                    if (maintenanceMode) {
+                        throw new Error(
+                            'Operation unavailable: Maintenance mode is enabled',
```
gastonfournier, line 126: Should we set this message in the backend instead? If you're using an API I believe you'd like to know as well...

chriswk (approved):
Nice.

frontend/src/hooks/api/actions/useApi/useApi.ts
nunogois: Fair point. How about this? https://github.com/Unleash/unleash/pull/5018/commits/d9ea8a6c4ffde382e39195a6a2adfe24c56d6544

gastonfournier: Nice, should we just get the message sent from the backend?

gastonfournier: I mean, that way we only change it on one side

nunogois: How about this then? https://github.com/Unleash/unleash/pull/5018/commits/a1323c994889541fd8946f4d0fddd5cbbd385bf4

gastonfournier (approved):
Looks nice!!

PR #5021, opened 2023-10-13
fix: maintenance banner should show right away when toggled

nunogois:
Makes the maintenance banner show right away when toggled. Includes some misc cleaning up.

Co-authored-by: Nicolae <nicolae@getunleash.ai>


nicolaesocaciu (approved):
LGTM

PR #5022, opened 2023-10-13
refactor: feature toggle list query

FredrikOseberg:
Attempt to clean up huge query method


chriswk (approved):
I like it, it definitely reads a lot better from the store side now. Cool builder pattern for the query.

kwasniew (approved):

PR #5023, opened 2023-10-13
feat: dependencies import validation

sjaanus:


src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -459,6 +464,32 @@ export default class ExportImportService
...
+    private async getUnsupportedDependencies(
+        dto: ImportTogglesSchema,
+    ): Promise<string[]> {
+        const dependentFeatures =
+            dto.data.dependencies?.flatMap((dependency) =>
+                dependency.dependencies.map((d) => d.feature),
+            ) || [];
+        const importedFeatures = dto.data.features.map((f) => f.name);
+
+        const missingFromImported = dependentFeatures.filter(
```
sjaanus, line 476: First find if parent might exist already in payload.

src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -459,6 +464,32 @@ export default class ExportImportService
...
+            (feature) => !importedFeatures.includes(feature),
+        );
+
+        let missingFeatures: string[] = [];
+
+        if (missingFromImported.length) {
+            const featuresFromStore = (
```
sjaanus, line 483: If they are not in payload, go check if they exist already in Unleash.

src/lib/features/export-import-toggles/import-validation-messages.ts
```diff
@@ -116,6 +118,14 @@ export class ImportValidationMessages {
             });
         }
 
+        if (dependencies.length > 0) {
+            errors.push({
+                message:
+                    'We detected the following dependencies in the import file that need to be created first:',
```
kwasniew, line 124: we can remove "in the import file" since some people may not upload a file but paste content or use API

src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -459,6 +464,32 @@ export default class ExportImportService
             : [];
     }
 
+    private async getUnsupportedDependencies(
```
kwasniew, line 467: maybe getMissingDependencies to be consistent with the names you use inside this method?

src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -468,6 +499,19 @@ export default class ExportImportService
         }
     }
 
+    private async verifyDependencies(dto: ImportTogglesSchema) {
+        const unsupportedDependencies = await this.getUnsupportedDependencies(
+            dto,
+        );
+        if (unsupportedDependencies.length > 0) {
+            throw new BadDataError(
+                `Following dependent features are missing from system: ${unsupportedDependencies.join(
```
kwasniew, line 508: I'd just say: 'The following dependent features are missing'

kwasniew (approved):
Why are you only updating permission test? Don't we have a separate test for validation called export-import.e2e.test.ts? The permission test should be focused on permissions.

PR #5040, opened 2023-10-16
feat: remove dependency on archive

kwasniew:
## About the changes

When archiving/batch archiving features we should remove all parent dependencies

Additional changes:
* added transaction support for archive and batch archive so that we can remove dependencies inside a transaction


andreas-unleash (approved):
LGTM!

PR #5041, opened 2023-10-16
feat: track add and remove dependencies

kwasniew:
## About the changes

Track when adding/removing dependencies in regular mode and change request mode.

### DIFF
--- a/frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
+++ b/frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
@@ -12,6 +12,7 @@ import { useChangeRequestApi } from 'hooks/api/actions/useChangeRequestApi/useCh
 import { usePendingChangeRequests } from 'hooks/api/getters/usePendingChangeRequests/usePendingChangeRequests';
 import useToast from 'hooks/useToast';
 import { formatUnknownError } from 'utils/formatUnknownError';
+import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
 
 interface IAddDependencyDialogueProps {
     project: string;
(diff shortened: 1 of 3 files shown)


frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
```diff
@@ -110,16 +117,34 @@ const useManageDependency = (
...
-                        : 'addDependency',
-                );
+                        : 'addDependency';
+                await handleAddChange(actionType);
+                trackEvent('dependent_features', {
+                    props: {
+                        eventType:
+                            actionType === 'addDependency'
+                                ? 'add dependency added to change request'
+                                : 'delete dependency added to change request',
```
thomasheartman, line 130: Very minor thing, but these two are a little confusing at first. Maybe we can rephrase them a bit?

Something like "dependency added in change request" and "dependency removed in change request"? Or "dependency added (change request)"/"dependency removed (change request)". Or we could use multiple props? Something like this:

```js
eventType: 'dependency added',
where: 'change request'
```

Primarily to avoid the "add x added" or "delete x added". But eh, as long as we understand what we're talking about I think it's fine 😄

thomasheartman (approved):
Looks good to me! One minor note about the change request events, but that's entirely optional.

frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
kwasniew: I was thinking about it myself. Originally I wanted to say "dependency added in change requests" but it's actually not true. We're only adding "add dependency" change to a change request. We'll know that dependency was added in change request once we apply the change request which may never happen.

PR #5047, opened 2023-10-16
chore: update docusaurus openapi and theme

thomasheartman:
Related to #5038. I think we'll need to update both the theme and the package in one go.


thomasheartman:
... obviously that didn't work. 🤔

PR #5050, opened 2023-10-16
refactor: create builder class for converting rows to avoid duplication

FredrikOseberg:
Create a builder for creating the data structures for feature toggle list and playground api


andreas-unleash (approved):
LGTM

PR #5053, opened 2023-10-16
ui: add splash screen for oss segments

thomasheartman:
This PR adds a splash screen for segments being open-sourced. 

It looks like this:

image

## About the changes

I've more or less wholesale copied the demo dialog that @nunogois implemented. I've put it in the `splash` directory for now (because that's where it seemed most appropriate). The reason for straight copying it instead of extending existing functionality is primarily that this should be short-lived and deleted after the next release or so. So isolating all the changes into a single directory seems like a good idea.

## Discussion points

Because OSS installations don't connect to Unleash, we can't use feature flags to control the rollout here. Instead, we must just assume that OSS users will want to see it. If there is a better way we can control this, that'd be great. I'd love to be able to use time constraints to not show this after a certain date, for instance, but I don't think that's something we can do right now?

The splash is also set to display on any page you're at when you first load unleash. However, closing the dialog (either by closing or by asking to see segments) will store that in localstorage, and you won't be shown the dialog again.


frontend/src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -74,13 +78,38 @@ const MainLayoutContentContainer = styled('div')(({ theme }) => ({
     zIndex: 200,
 }));
 
+const SegmentsSplash: React.FC = () => {
+    const { value: localStorageState, setValue: setLocalStorageState } =
+        createLocalStorage('OssSegmentsSplashScreen:v1', { shown: false });
+
+    const [showSegmentSplash, setShowSegmentSplash] = React.useState(true);
```
kwasniew, line 85: why do we need both local storage and memory state? can't we solve it with only local storage state?

thomasheartman: Potentially. I can try it out. Just afraid that we'll end up in a sitch where it doesn't close. But good point 💁🏼

thomasheartman: Yeah, don't know if I'm doing something wrong, but that just leads to me not being able to close it. Seems innocuous enough to leave it in 🤷🏼

kwasniew: ok, so let's keep it as it is :)

kwasniew (approved):

frontend/src/component/splash/SegmentsSplashScreen/SegmentsSplashScreen.tsx
```diff
@@ -0,0 +1,87 @@
...
+            </StyledHeader>
+            <Typography color='textSecondary' sx={{ mt: 2 }}>
+                We are excited to announce that we are releasing an enterprise
+                feature for our open source community.
+            </Typography>
+            <StyledImage
+                src={formatAssetPath(ossSegmentsImage)}
+                alt='The segment creation screen, showing an example segment consisting of three constraints.'
+            />
+            <StyledLink href={'https://docs.getunleash.io/reference/segments'}>
```
nunogois, line 64: Can be simplified:
```suggestion
            <StyledLink href='https://docs.getunleash.io/reference/segments'>
```

frontend/src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -74,9 +78,33 @@ const MainLayoutContentContainer = styled('div')(({ theme }) => ({
...
+        <SegmentsSplashScreen
+            open={showSegmentSplash && !localStorageState.shown}
+            onClose={closeSegmentsSplash}
+            showSegments={() => {
+                closeSegmentsSplash();
+                navigate(`/segments`);
+            }}
+        />
+    );
+};
```
nunogois, line 103: I feel like this belongs in a separate file and not part of MainLayout. WDYT?

frontend/src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -74,9 +78,33 @@ const MainLayoutContentContainer = styled('div')(({ theme }) => ({
...
+    };
+
+    return (
+        <SegmentsSplashScreen
+            open={showSegmentSplash && !localStorageState.shown}
+            onClose={closeSegmentsSplash}
+            showSegments={() => {
+                closeSegmentsSplash();
+                navigate(`/segments`);
+            }}
```
nunogois, line 100: Do we want to add a Plausible event here?

frontend/src/component/layout/MainLayout/MainLayout.tsx
nunogois: Yeah, I also have the feeling that this could potentially be slightly simplified. Since I think it would be cool to see more announcements like these in the future, perhaps it's something we could do for a reusable component in a later PR.

nunogois (approved):
LGTM! Great job with this 👍

frontend/src/component/layout/MainLayout/MainLayout.tsx
thomasheartman: Ah, that's a good point. Yeah, we could definitely do that. What do you think we should track? Just the how many people see it?

frontend/src/component/layout/MainLayout/MainLayout.tsx
thomasheartman: Yeah, that's fair. I put it here just for now, but I could pop it in the SegmentsSplashScreen directory. Should probably be part of the main exported component, I reckon. What do you think?

frontend/src/component/layout/MainLayout/MainLayout.tsx
thomasheartman: I think that what's going on is that it doesn't re-render when you update localStorage, so even if you set the localStorage key, that's not picked up until the re-render. Hence the need for an extra variable. 

But if you have suggestions on how to simplify it, I'd be happy to listen 😄 

As for more announcements like these in the future, yeah, maybe. I'm personally not a big fan of in-your-face updates like this, but it _does_ get the point across, I guess. But I suppose it's not up to me to decide for all of us 💁🏼

frontend/src/component/layout/MainLayout/MainLayout.tsx
nunogois: I think that could work, definitely better than being a part of `MainLayout` 👍

frontend/src/component/layout/MainLayout/MainLayout.tsx
nunogois: Yeah, tracking an event here would let us see how many people click it, but it depends on whether we consider it's worth tracking that 🤷 

So, something similar to the `trackEvent`s you'll find in `Demo.tsx` for example.

frontend/src/component/layout/MainLayout/MainLayout.tsx
nunogois: I think this is totally fine for now 🙂 

Yeah, understandable. Maybe a more out-of-the-way message banner could be a better option in the future.

thomasheartman:
Found a visual glitch in non-oss versions (it's in the vercel preview). Seems like the splash gets rendered initially before getting turned off. Could it be that the `isOss` function doesn't return the right value immediately? It might also be related to localstorage, but that's further down the road, so I'd expect it to be stopped before then 🤔

thomasheartman:
Interestingly, I'm not able to replicate it locally, so this'll be a little tricky.

thomasheartman:
But yeah, this line is probably why:

```ts
    const isOss = useCallback(() => {
        return !data?.versionInfo?.current?.enterprise;
    }, [data]);
```

So before we get data, I'm assuming we're counting everything as OSS. I wonder if that's the right approach? Probably fine 💁🏼

PR #5056, opened 2023-10-16
chore: add enterprise event

gastonfournier:
This just adds a new event type


nunogois (approved):
LGTM! 
I'll leave it up to you whether you consider it's worth it adding this to the event formatter and the list of available events in the Slack App definition.

nunogois:
> > LGTM! I'll leave it up to you whether you consider it's worth it adding this to the event formatter and the list of available events in the Slack App definition.
> 
> I think this is a different thread, but shouldn't SlackApp have all events? I feel it's difficult to keep track of every place we need to add the new event to... I rather have slackApp filter out some events but by default have all event types available. WDYT?

I agree. In fact, I think **all** integrations should have **all** events available to configure. However we need to take into account that some events may not have a message template in the event formatter (even though there's a fallback) and that some events are deprecated (these could be filtered out manually). Maybe something to try out in a different PR.

PR #5059, opened 2023-10-16
feat: add option to return disabled strategies

andreas-unleash:
Adds the option to include disabled strategies (behind the playgroundImprovements flag

Closes # [1-1505](https://linear.app/unleash/issue/1-1505/return-disabled-strategies-in-the-playground-features-request)

### DIFF
--- a/src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
+++ b/src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
@@ -151,14 +151,6 @@ export class FeatureToggleRowConverter {
 
             feature = this.createBaseFeature(r, feature, featureQuery);
 
-            if (
-                this.isUnseenStrategyRow(feature, r) &&
-                includeDisabledStrategies &&
-                r.strategy_disabled
-            ) {
-                feature.strategies?.push(this.rowToStrategy(r));
-            }
-
             feature.createdAt = r.created_at;
             feature.favorite = r.favorite;
 
--- a/src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
+++ b/src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
@@ -112,7 +112,6 @@ export class FeatureToggleRowConverter {
         row: any,
         feature: PartialDeep<IFeatureToggleClient>,
         featureQuery?: IFeatureToggleQuery,
-        includeDisabledStrategies?: boolean,
     ) => {
         feature.impressionData = row.impression_data;
         feature.enabled = !!row.enabled;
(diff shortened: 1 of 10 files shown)


src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
```diff
@@ -123,7 +125,10 @@ export class FeatureToggleRowConverter {
         feature.variants = row.variants || [];
         feature.project = row.project;
 
-        if (this.isUnseenStrategyRow(feature, row) && !row.strategy_disabled) {
+        if (
+            this.isUnseenStrategyRow(feature, row) &&
+            (includeDisabledStrategies ? true : !row.strategy_disabled)
+        ) {
```
FredrikOseberg, line 131: I don't think this is the correct place to put this. createBaseFeature should only care about what is common between the two API endpoints. Instead you should take this check and move into the respective methods: `buildFeatureToggleListFromRows` and `buildPlaygroundFeaturesFromRows` and have each method have different checks for which strategies they allow

FredrikOseberg (changes requested):
We should improve the architecture slightly before merging

src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
andreas-unleash: Ok makes sense

src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
```diff
@@ -149,6 +151,14 @@ export class FeatureToggleRowConverter {
...
             feature = this.createBaseFeature(r, feature, featureQuery);
 
+            if (
+                this.isUnseenStrategyRow(feature, r) &&
+                includeDisabledStrategies &&
+                r.strategy_disabled
+            ) {
+                feature.strategies?.push(this.rowToStrategy(r));
+            }
+
```
FredrikOseberg, line 161: Why does the feature toggle list need disabled strategies?

FredrikOseberg (changes requested):
I think this is only relevant to the playground API

src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
andreas-unleash: Following to the frontend - the disabled strategies are not needed

src/lib/features/feature-toggle/converters/feature-toggle-row-converter.ts
andreas-unleash: Turns out that both of those functions are only used for the Admin API - for which we (everywhere) return the disabled strategies

andreas-unleash:
Not only relevant to playground. To be more specific - this is needed for the admin api

FredrikOseberg (approved):
LGTM

PR #5061, opened 2023-10-17
fix: extract username from user should not return undefined

nunogois:
This fixes a return type error by changing the logic of `extractUsernameFromUser` to never return undefined.

In the previous code, `user` could be truthy, but that doesn't mean `email` or `username` were defined. This assumes we always fallback to "unknown" in those scenarios.

### DIFF
--- a/src/lib/util/extract-user.ts
+++ b/src/lib/util/extract-user.ts
@@ -1,7 +1,7 @@
 import { IAuthRequest, IUser } from '../server-impl';
 
 export function extractUsernameFromUser(user: IUser): string {
-    return user ? user.email || user.username : 'unknown';
+    return user?.email || user?.username || 'unknown';
 }
 
 export function extractUsername(req: IAuthRequest): string {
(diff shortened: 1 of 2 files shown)


src/lib/util/extract-user.ts
```diff
@@ -1,7 +1,7 @@
 import { IAuthRequest, IUser } from '../server-impl';
 
 export function extractUsernameFromUser(user: IUser): string {
-    return user ? user.email || user.username : 'unknown';
```
gastonfournier, line 4: Isn't this the same? Can you unit-test where this fails?

gastonfournier: Oh, I think I know what you said in the description, something like:
```typescript
user: {}
```
would lead to undefined. Still, a unit test would be good in this case

gastonfournier (approved):
LGTM

src/lib/util/extract-user.ts
nunogois: Added some tests in https://github.com/Unleash/unleash/pull/5061/commits/19804ba03b483465c8e0cb6dd4ac7a8212142772.
Before this change, the 3rd test "Should return "unknown" if neither email nor username exists" would fail. We would also have a type error on the method, since it is specifying that we should return `string` instead of `string | undefined`.

PR #5065, opened 2023-10-17
feat: warn about sdk update with feature dependencies

kwasniew:
## About the changes

Screenshot 2023-10-17 at 11 34 52

### DIFF
--- a/frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
+++ b/frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
@@ -190,8 +190,8 @@ export const AddDependencyDialogue = ({
             secondaryButtonText='Cancel'
         >
             <Box>
-                <DependenciesUpgradeAlert/>
-                <Box sx={{mt: 2, mb: 4}}>
+                <DependenciesUpgradeAlert />
+                <Box sx={{ mt: 2, mb: 4 }}>
                     Your feature will be evaluated only when the selected parent
                     feature is enabled in the same environment.
                 </Box>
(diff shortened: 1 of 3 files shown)


frontend/src/component/feature/Dependencies/DependenciesUpgradeAlert.tsx
```diff
@@ -0,0 +1,22 @@
...
+};
+
+const DependenciesDocsLink = () => {
+    return (
+        <a
+            href='https://docs.getunleash.io/reference/dependent-features#client-sdk-support'
+            target='_blank'
+            rel='noreferrer'
+        >
+            Read more
```
thomasheartman, line 19: This link text doesn't say anything about what the link points to. It should be possible to understand what a link points to without any context. Can we do something like this instead?
```suggestion
            Client SDK support for feature dependencies
```

Potentially preface the link (outside of the `a` tag) with "Read more about"?

frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
```diff
@@ -189,6 +190,8 @@ export const AddDependencyDialogue = ({
             secondaryButtonText='Cancel'
         >
             <Box>
+                <DependenciesUpgradeAlert/>
+                <br />
```
thomasheartman, line 194: Why do we use the `br` tag? Shouldn't spacing between elements be handled by styling?

thomasheartman (approved):
It's probably fine to put this in, but I have a few comments. The important one is around accessibility. The other one is more a nit, but I see we've already done that.

frontend/src/component/feature/Dependencies/DependenciesUpgradeAlert.tsx
kwasniew: fixing

PR #5069, opened 2023-10-17
chore: generate orval types

andreas-unleash:
Generate new types


frontend/src/openapi/index.ts
```diff
@@ -1 +1,29 @@
 export * from './models';
+export * from './apis/operational';
```
kwasniew, line 2: those shouldn't be added

kwasniew (changes requested):

PR #5072, opened 2023-10-17
fix: log diff

FredrikOseberg:
Add a deep diff function to investigate diff in logs

### DIFF
--- a/src/lib/features/feature-toggle/deep-diff.ts
+++ b/src/lib/features/feature-toggle/deep-diff.ts
@@ -1,0 +1,66 @@
+interface Difference {
+    index: (string | number)[];
+    reason: string;
+    valueA: any;
+    valueB: any;
+}
+
+export function deepDiff(arr1: any[], arr2: any[]): Difference[] | null {
+    const diff: Difference[] = [];
+
+    function compare(a: any, b: any, parentIndex: (string | number)[]): void {
+        if (Array.isArray(a) && Array.isArray(b)) {
+            if (a.length !== b.length) {
+                diff.push({
+                    index: parentIndex,
+                    reason: 'Different lengths',
+                    valueA: a,
+                    valueB: b,
+                });
+            } else {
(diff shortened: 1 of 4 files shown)


src/lib/features/feature-toggle/deep-diff.ts
```diff
@@ -0,0 +1,66 @@
...
+            diff.push({
+                index: parentIndex,
+                reason: 'Different values',
+                valueA: a,
+                valueB: b,
+            });
+        }
+    }
+
+    function arraysEqual(a: any[], b: any[]): boolean {
```
kwasniew, line 56: can we move this function out?

gardleopard (approved):
LGTM

src/lib/features/feature-toggle/deep-diff.ts
```diff
@@ -0,0 +1,66 @@
+interface Difference {
+    index: (string | number)[];
+    reason: string;
+    valueA: any;
+    valueB: any;
+}
+
+export function deepDiff(arr1: any[], arr2: any[]): Difference[] | null {
```
kwasniew, line 8: is there relationship between the arr1 and arr2 types? 
`deepDiff<T>(arr1: T[], arr2: T[]): Difference[] | null`

src/lib/features/feature-toggle/deep-diff.ts
kwasniew: I'd expect array equals signature to by typed: `function arraysEqual<T>(a: T[], b: T[]): boolean`

kwasniew: also it's not really array equality since we sort the elements. It's checking if same elements exist in both arrays which can be done with Sets

src/lib/features/feature-toggle/deep-diff.ts
FredrikOseberg: In this case they are different types.

PR #5073, opened 2023-10-17
feat: dependent features use new transaction mechanism

kwasniew:
## About the changes

Dependent features use new transaction mechanism

Added missing transactions in delete dependencies operations.

### DIFF
--- a/src/lib/services/index.ts
+++ b/src/lib/services/index.ts
@@ -268,10 +268,11 @@ export const createServices = (
     );
 
     const dependentFeaturesService = db
-        ? createDependentFeaturesService(db, config)
+        ? createDependentFeaturesService(config)(db)
         : createFakeDependentFeaturesService(config);
-    const transactionalDependentFeaturesService = (txDb: Knex.Transaction) =>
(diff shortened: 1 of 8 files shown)


src/lib/services/index.ts
```diff
@@ -268,10 +268,11 @@ export const createServices = (
     );
 
     const dependentFeaturesService = db
-        ? createDependentFeaturesService(db, config)
+        ? createDependentFeaturesService(config)(db)
         : createFakeDependentFeaturesService(config);
-    const transactionalDependentFeaturesService = (txDb: Knex.Transaction) =>
-        createDependentFeaturesService(txDb, config);
+    const transactionalDependentFeaturesService = db
```
gastonfournier, line 273: Do you need both dependentFeatureService and transactionalDependentFeaturesService? I think you can just 
```typescript
const dependentFeaturesService = transactionalDependentFeaturesService
```

gastonfournier (approved):
LG!

src/lib/services/index.ts
kwasniew: you're right, we don't need both.

PR #5075, opened 2023-10-17
feat: Adds rate limiting to metric POST endpoints

chriswk:
### What
The heaviest requests we serve are the register and metrics POSTs from our SDKs/clients. 
This PR adds ratelimiting to /api/client/register, /api/client/metrics, /api/frontend/register and /api/frontend/metrics with a default set to 6000 requests per minute (or 100 rps) for each of the endpoints. 

It will be overrideable by the environment variables documented.

### Points of discussion
@kwasniew already suggested using featuretoggles with variants to control the rate per clientId. I struggled to see if we could dynamically update the middleware after initialisation, so this attempt will need a restart of the pod to update the request limit.


src/lib/routes/util.ts
```diff
@@ -29,14 +29,10 @@ export const handleErrors: (
...
-                // @ts-expect-error http errors all have statuses, but there are no
-                // types provided
-                .status(error.status ?? 400)
-                .json({ message: error.message })
-                .end()
-        );
+        return res
+            .status(error.status ?? 400)
+            .json({ message: error.message })
+            .end();
```
kwasniew, line 35: we don't need .end() to close the write stream since json() will do that.

kwasniew (approved):

src/lib/create-config.ts
```diff
@@ -99,6 +100,38 @@ function loadClientCachingOptions(
...
+        6000,
+    );
+    const frontendRegisterMax = parseEnvVarNumber(
+        process.env.REGISTER_FRONTEND_RATE_LIMIT,
+        6000,
+    );
+    const frontendMetricsMax = parseEnvVarNumber(
+        process.env.FRONTEND_METRICS_RATE_LIMIT,
+        6000,
+    );
```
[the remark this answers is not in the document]
chriswk: Agreed, it's better

src/lib/routes/util.ts
chriswk: heh, fair shout, I can update it, but it wasn't actually related to this PR. tsc just pointed out that we no longer got a tsc-error here, so had to remove the @ts-expect-error comment

src/lib/create-config.ts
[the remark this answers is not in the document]
chriswk: Handled in 9499986

src/lib/routes/util.ts
chriswk: And I see we still got that error, because the build failed, it was just the ts-expect-error had ended up in the wrong place. Updated to remove the end() call and readd the ts-expect-error with an explanation in 9499986

PR #5076, opened 2023-10-18
refactor: move message banner interface to common file

nunogois:
https://linear.app/unleash/issue/2-1527/set-up-message-banner-interfaces

Tiny refactor to move the message banner interface to its own file, since it will be used in multiple places.


gastonfournier (approved):
LG!

PR #5077, opened 2023-10-18
fix: server-side request forgery in @cypress/request@2.88.12

sjaanus:
Fixing https://github.com/Unleash/unleash/security/dependabot/84


sighphyre (approved):
G

PR #5078, opened 2023-10-18
feat: add message banner API hooks

nunogois:
https://linear.app/unleash/issue/2-1510/create-message-banner-hooks-that-connect-to-the-new-api-endpoints

Adds new message banner API hooks that will allow us to do CRUD operations for message banners in the UI.

### DIFF
--- a/frontend/src/hooks/api/actions/useMessageBannersApi/useMessageBannersApi.ts
+++ b/frontend/src/hooks/api/actions/useMessageBannersApi/useMessageBannersApi.ts
@@ -1,0 +1,61 @@
+import { IMessageBanner } from 'interfaces/messageBanner';
+import useAPI from '../useApi/useApi';
+
+const ENDPOINT = 'api/admin/message-banners';
+
+export const useMessageBannersApi = () => {
+    const { loading, makeRequest, createRequest, errors } = useAPI({
+        propagateErrors: true,
+    });
+
+    const addMessageBanner = async (messageBanner: IMessageBanner) => {
+        const requestId = 'addMessageBanner';
+        const req = createRequest(
+            ENDPOINT,
+            {
+                method: 'POST',
+                body: JSON.stringify(messageBanner),
+            },
+            requestId,
+        );
+
+        const response = await makeRequest(req.caller, req.id);
+        return response.json();
+    };
+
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/api/actions/useMessageBannersApi/useMessageBannersApi.ts
```diff
@@ -0,0 +1,68 @@
+import { IInternalMessageBanner } from 'interfaces/messageBanner';
+import useAPI from '../useApi/useApi';
+
+const ENDPOINT = 'api/admin/message-banners';
```
gastonfournier, line 4: Just wondering, should we just call them "banners" instead of "message-banners"? I feel we might regret in the future for not giving it a more general name

frontend/src/hooks/api/getters/useMessageBanners/useMessageBanners.ts
```diff
@@ -0,0 +1,38 @@
+import { useMemo } from 'react';
+import { formatApiPath } from 'utils/formatPath';
+import handleErrorResponses from '../httpErrorResponseHandler';
+import { useConditionalSWR } from '../useConditionalSWR/useConditionalSWR';
+import useUiConfig from '../useUiConfig/useUiConfig';
+import { useUiFlag } from 'hooks/useUiFlag';
+import { IInternalMessageBanner } from 'interfaces/messageBanner';
+
+const ENDPOINT = 'api/admin/message-banners';
```
gastonfournier, line 9: This is re-defined here, I believe you can just export this const in the API definition

gastonfournier (approved):
LGTM, I feel some changes are also in another PR you just missed rebasing master, but that's fine

frontend/src/hooks/api/actions/useMessageBannersApi/useMessageBannersApi.ts
gastonfournier: Well, you know more than I do, so I trust your judgment. It was just an observation because message-banner seems weird... a banner is by definition a message you want to deliver... the fact that it's a "message banner" type is an implementation detail IMO. But again, take it for what it is: opinions without a lot of context :D

PR #5080, opened 2023-10-18
fix: force deletion of archived toggles when deleting a project

sighphyre:
This forces the project service to delete archived toggles when a project is deleted. Previously, when deleting a project, you were blocked if you had one or more active toggles but once they're archived the project can be deleted. Those archived toggles would persist indefinitely and when a new project was created that matched the old project name, they would immediately be bound to that project in an archived state. This was really confusing for users across teams, since they could create a project that someone else had created, used and then deleted and they would get a bunch of toggles in their project that they had never touched

### DIFF
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -390,6 +390,14 @@ export default class ProjectService {
             );
         }
 
+        const archivedToggles = await this.featureToggleStore.getAll({
+            project: id,
+            archived: true,
+        });
+
+        await this.featureToggleStore.batchDelete(
+            archivedToggles.map((toggleName) => toggleName.name),
+        );
         await this.projectStore.delete(id);
 
         await this.eventService.storeEvent({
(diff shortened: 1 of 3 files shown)


sjaanus (approved):
LGTM

src/lib/services/project-service.ts
```diff
@@ -390,6 +390,14 @@ export default class ProjectService {
             );
         }
 
+        const archivedToggles = await this.featureToggleStore.getAll({
+            project: id,
+            archived: true,
+        });
+
+        await this.featureToggleStore.batchDelete(
```
sjaanus, line 398: We should also post events for deletions

sjaanus (changes requested):

src/lib/services/project-service.ts
sighphyre: If we're deleting the project, where will these go?

sjaanus: Global event list.

sighphyre: Update - we have a global event log. That's where this will go

sighphyre: Good catch, I'll fix, thanks!

sjaanus (approved):
The code looks good, if you can get tests working.

PR #5085, opened 2023-10-18
feat: ensure at least one owner on remove user/group access

gastonfournier:
## About the changes
This makes sure that projects have at least one owner, either a group or a user. This is to prevent accidentally losing access to a project.

We check this when removing a user/group or when changing  the role of a user/group

**Note**: We can still leave a group empty as the only owner of the project, but that's okay because we can still add more users to the group

### DIFF
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -98,8 +98,6 @@ export default class ProjectService {
 
     private featureToggleStore: IFeatureToggleStore;
 
-    private featureTypeStore: IFeatureTypeStore;
-
     private featureEnvironmentStore: IFeatureEnvironmentStore;
 
     private environmentStore: IEnvironmentStore;
@@ -120,8 +118,6 @@ export default class ProjectService {
 
     private projectStatsStore: IProjectStatsStore;
 
-    private lastSeenReadModel: ILastSeenReadModel;
-
     private flagResolver: IFlagResolver;
 
     private isEnterprise: boolean;
@@ -131,7 +127,6 @@ export default class ProjectService {
             projectStore,
             eventStore,
             featureToggleStore,
-            featureTypeStore,
             environmentStore,
             featureEnvironmentStore,
             accountStore,
@@ -141,7 +136,6 @@ export default class ProjectService {
             | 'projectStore'
             | 'eventStore'
             | 'featureToggleStore'
-            | 'featureTypeStore'
             | 'environmentStore'
             | 'featureEnvironmentStore'
             | 'accountStore'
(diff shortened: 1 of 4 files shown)


src/lib/services/project-service.ts
```diff
@@ -671,6 +671,19 @@ export default class ProjectService {
             projectId,
             userId,
         );
+
+        const ownerRole = await this.accessService.getRoleByName(
+            RoleName.OWNER,
+        );
+
+        if (
```
sighphyre, line 679: It's early and I don't have a lot of coffee in my blood yet but gosh this if statement is hard to read. It's also duplicated. 

Can we pull this out into a function and give some of the contents some names? ChatGPT suggests something like this:

```
const hasExistingOwnerRole = existingRoles.includes(ownerRole.id);
const isRemovingOwnerRole = !roles.includes(ownerRole.id);

if (hasExistingOwnerRole && isRemovingOwnerRole) {
    await this.validateAtLeastOneOwner(projectId, ownerRole);
}

```

sighphyre (commented):
I think this wants a test. I'm guessing this originally regressed because there wasn't one

src/lib/services/project-service.ts
gastonfournier: What I don't like about adding constants is that despite the short-circuit `&&` we have to execute both finds, but I know what to do

sighphyre: Is that actually a concern? `find` executes in about a millisecond on 100K elements (how many roles do we have?). Piping stuff back from the database is going to dwarf this completely. 

If it offends you, go for it, but I don't think it's necessary to trade off readability for performance

gastonfournier: Yeah, when I was making the change I had the same thought and I'm leaning towards readability on this one ;)

src/test/e2e/services/project-service.e2e.test.ts
```diff
@@ -491,31 +499,6 @@ test('should remove user from the project', async () => {
     expect(memberUsers).toHaveLength(0);
 });
 
-test('should not remove user from the project', async () => {
```
gastonfournier, line 494: Moved to line 1006

sighphyre (approved):
LGTM

src/lib/services/project-service.ts
```diff
@@ -508,6 +500,11 @@ export default class ProjectService {
             userId,
         );
 
+        const ownerRole = await this.accessService.getRoleByName(
+            RoleName.OWNER,
```
kwasniew, line 504: this seems to be a bug since we're getting a role for the owner not the user we're deleting

PR #5090, opened 2023-10-19
feat: prevent self dependencies

kwasniew:
## About the changes

Should not allow features depending on themselves

### DIFF
--- a/src/lib/features/dependent-features/dependent-features-service.ts
+++ b/src/lib/features/dependent-features/dependent-features-service.ts
@@ -90,6 +90,10 @@ export class DependentFeaturesService {
     ): Promise<void> {
         const { enabled, feature: parent, variants } = dependentFeature;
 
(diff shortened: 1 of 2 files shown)


src/lib/features/dependent-features/dependent-features-service.ts
```diff
@@ -90,6 +90,10 @@ export class DependentFeaturesService {
     ): Promise<void> {
         const { enabled, feature: parent, variants } = dependentFeature;
 
+        if (child === parent) {
+            throw new InvalidOperationError('Self dependency not allowed.');
```
thomasheartman, line 94: Wonder if we can make this slightly clearer, like this? But it's very minor.

```suggestion
            throw new InvalidOperationError('A feature flag cannot depend on itself.');
```

thomasheartman (approved):
Nice one! 🔥

PR #5091, opened 2023-10-19
chore: remove storybook

sjaanus:
We have vulnerability from storybook https://github.com/Unleash/unleash/security/dependabot/89.
Talked with Thomas that we are not actually using it, so removing it.


package.json
```diff
@@ -215,6 +215,6 @@
   },
   "lint-staged": {
     "*.{js,ts}": ["biome check --apply"],
-    "*.json": ["biome format --write"]
+    "*.json": ["biome format --write --no-errors-on-unmatched"]
```
thomasheartman, line 218: Oh, this is a nice catch! I've been wondering why I've been getting that error sometimes 😄

thomasheartman (approved):
Ooh, that's a nice ratio! But yeah, this looks great to me; thanks for sorting it out 🙌🏼

PR #5092, opened 2023-10-19
chore: force tough-cookie to 4.1.3 due to vulnerability

sjaanus:
Solves this https://github.com/Unleash/unleash/security/dependabot/76


nunogois (approved):
LGTM

PR #5093, opened 2023-10-19
feat: check if child and parent are in the same project

kwasniew:
## About the changes

Verify that child and parent features are in the same project.


FredrikOseberg (approved):
LGTM

PR #5095, opened 2023-10-19
feat: make all internal rate limits configurable

chriswk:
### What
This PR makes the rate limit for user creation and simple login (our password based login) configurable in the same way you can do metricsRateLimiting.

### Worth noting
In addition this PR adds a `rate_limit{endpoint, method}` prometheus gauge, which gets the data from the UnleashConfig.


src/lib/create-config.ts
```diff
@@ -19,6 +19,7 @@ import {
     ICspDomainOptions,
     IClientCachingOption,
     IMetricsRateLimiting,
+    IRateLimiting,
```
thomasheartman, line 22: Irate limiting? 😄

src/lib/create-config.ts
```diff
@@ -132,6 +133,23 @@ function loadMetricsRateLimitingConfig(
...
 }
 
+function loadRateLimitingConfig(options: IUnleashOptions): IRateLimiting {
+    const createUserMaxPerMinute = parseEnvVarNumber(
+        process.env.CREATE_USER_RATE_LIMIT_PER_MINUTE,
+        20,
+    );
+    const simpleLoginMaxPerMinute = parseEnvVarNumber(
+        process.env.SIMPLE_LOGIN_LIMIT_PER_MINUTE,
+        10,
```
thomasheartman, line 143: Curious as to why one is called "x_RATE_LIMIT_PER_MINUTE" and the other is just called "x_LIMIT_PER_MINUTE"? Not a big deal, but feels like they should have similar names to make it easier to remember.

src/lib/metrics.ts
```diff
@@ -259,6 +265,42 @@ export default class MetricsMonitor {
                         .labels({ range: clientStat.range })
                         .set(clientStat.count),
                 );
+
+                rateLimits.reset();
```
thomasheartman, line 269: Any particular reason we need to reset it here? Will we have gotten any inputs before this point that we need to get rid of?

src/lib/metrics.ts
```diff
@@ -259,6 +265,42 @@ export default class MetricsMonitor {
...
+                        config.metricsRateLimiting.frontendMetricsMaxPerMinute,
+                    );
+                rateLimits
+                    .labels({
+                        endpoint: '/api/frontend/register',
+                        method: 'POST',
+                    })
+                    .set(
+                        config.metricsRateLimiting.frontendRegisterMaxPerMinute,
+                    );
```
thomasheartman, line 294: Out of curiosity: what were all of these before? If these options are old, how were they applied? Did we do it somewhere else?

thomasheartman (approved):
This looks pretty good to me! Have a couple of questions and suggestions, but nothing blocking 😄

src/lib/metrics.ts
chriswk: I'm still not sure, I just followed the patterns for the rest of our static metrics.

chriswk: In my mind since it's a gauge there shouldn't really be a need to reset it.

src/lib/create-config.ts
chriswk: indeed, I was quite IRate while writing this :P

src/lib/metrics.ts
chriswk: there were no limits for register and metrics earlier besides what's configured on the loadbalancer side.
For the user creation and login it was 10 and 20.

PR #5096, opened 2023-10-19
alt test using each

gastonfournier:
We have different ways of changing access to a project, this pr covers tests with different functions

### DIFF
--- a/src/test/e2e/services/project-service.e2e.test.ts
+++ b/src/test/e2e/services/project-service.e2e.test.ts
@@ -17,7 +17,8 @@ import {
     createFeatureToggleService,
     createProjectService,
 } from '../../../lib/features';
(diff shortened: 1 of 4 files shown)


src/test/e2e/services/project-service.e2e.test.ts
```diff
@@ -1026,54 +1026,45 @@ describe('ensure project has at least one owner', () => {
...
-                [memberRole.id],
-                'test',
+            expect(() =>
+                fn(project.id, memberRole.id, user.id),
+            ).rejects.toThrowError(
+                new Error('A project must have at least one owner'),
             );
-        }).rejects.toThrowError(
-            new Error('A project must have at least one owner'),
-        );
```
gastonfournier, line 1075: Instead of making 2 assertions, this splits into 2 tests using `test.each`

PR #5098, opened 2023-10-19
refactor: rename message banners to banners

nunogois:
https://linear.app/unleash/issue/2-1531/rename-message-banners-to-banners

This renames "message banners" to "banners".

I also added support for external banners coming from a `banner` flag instead of only `messageBanner` flag, so we can eventually migrate to the new one in the future if we want.


chriswk (approved):
LGTM

PR #5099, opened 2023-10-19
fix: add project filter to feature-toggle-list-builder

andreas-unleash:
Adds `forProject` filter to FeatureToggleQueryBuilder
Filters playground results 

Closes # [1-1525](https://linear.app/unleash/issue/1-1525/getting-results-for-all-projects-when-defining-a-single-project-in-the)

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-store.ts
@@ -198,6 +198,10 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             builder.addSelectColumn('df.enabled as parent_enabled');
         }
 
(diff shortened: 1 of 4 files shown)


kwasniew (approved):
Nice. Question: so this bug got introduced when we started splitting the stores?

andreas-unleash:
> Nice. Question: so this bug got introduced when we started splitting the stores?

Yes, looks like it - tried with flag off and the clientFeatureToggleStore returns correct results

FredrikOseberg:
> > Nice. Question: so this bug got introduced when we started splitting the stores?
> 
> Yes, looks like it - tried with flag off and the clientFeatureToggleStore returns correct results

Interesting, thanks for fixing.

PR #5103, opened 2023-10-19
chore: remove ts-ignore and adapt tests

gastonfournier:
## About the changes
Just using proper types and adapting a few tests

### DIFF
--- a/src/test/e2e/services/project-service.e2e.test.ts
+++ b/src/test/e2e/services/project-service.e2e.test.ts
@@ -171,10 +171,6 @@ test('should not be able to delete project with toggles', async () => {
     await projectService.createProject(project, user);
     await stores.featureToggleStore.create(project.id, {
         name: 'test-project-delete',
-        // @ts-ignore project does not exist in type FeatureToggleDTO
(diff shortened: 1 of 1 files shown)


src/test/e2e/services/project-service.e2e.test.ts
```diff
@@ -1343,11 +1354,13 @@ test('Should allow permutations of roles, groups and users when adding a new acc
     const { users, groups } = await projectService.getAccessToProject(
         project.id,
     );
+    const ownerRole = await stores.roleStore.getRoleByName(RoleName.OWNER);
 
-    expect(users).toHaveLength(2);
+    expect(users).toHaveLength(3); // the 2 added plus the one that created the project
```
gastonfournier, line 1359: FYI @chriswk as you're the last to edit this test, it was 2 because when creating a project with an invalid user (one without an id), the code will just ignore adding the user as owner without failing: https://github.com/Unleash/unleash/blob/1335da63661d476fac0c912c7076b93d38606e5b/src/lib/services/access-service.ts#L561-L567

PR #5104, opened 2023-10-19
feat: show warning about dependencies removed on archive

kwasniew:
## About the changes

Users should know that dependencies will be removed
Screenshot 2023-10-19 at 16 39 07

The most important change is that our read model returns: 
```
    {
        hasDeletedDependencies: true,
        parentsWithChildFeatures: ['parentA', 'parentB'],
    }
```
which covers our previous use case (parent that would orphan child features) and our new use case indicating if any dependencies will be deleted as part of the archive operation.


frontend/src/component/common/FeatureArchiveDialog/FeatureArchiveDialog.test.tsx
```diff
@@ -38,7 +38,10 @@ const setupArchiveValidation = (orphanParents: string[]) => {
     testServerRoute(
         server,
         '/api/admin/projects/projectId/archive/validate',
-        orphanParents,
+        {
```
kwasniew, line 41: new structure returned from backend

frontend/src/component/common/FeatureArchiveDialog/FeatureArchiveDialog.test.tsx
```diff
@@ -144,6 +153,11 @@ test('Show error message when multiple parents of orphaned children are archived
     await screen.findByText(
         'have child features that depend on them and are not part of the archive operation. These parent features can not be archived:',
     );
+    expect(
```
kwasniew, line 156: we should not show the warning if there's already an error about orphaned children. Error and warning would confuse people

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1587,8 +1587,22 @@ class FeatureToggleService {
...
+        const hasDeletedDependencies =
+            await this.dependentFeaturesReadModel.haveDependencies(
+                featureNames,
+            );
+        const parentsWithChildFeatures =
+            await this.dependentFeaturesReadModel.getOrphanParents(
+                featureNames,
+            );
+        return {
+            hasDeletedDependencies,
```
kwasniew, line 1603: added this to our archive validation. It now indicates if any dependencies will be removed

FredrikOseberg (approved):

PR #5107, opened 2023-10-20
fix: remove docusaurus from main package json

kwasniew:
## About the changes

Removing docusaurus as it breaks the build. Also it shouldn't be part of the app code and should stay in the website part.


chriswk (approved):
Nice catch

PR #5108, opened 2023-10-20
feat: promise timeout on lock

kwasniew:
## About the changes

* Adding Promise timeout in DB lock. Once the lock is acquired the function that we're locking (e.g. DB migration) has 60 seconds (default value) to do its job. After the timeout elapsed we unlock the lock.
* This change may also remove some flakiness in the tests since I noticed that the previous incorrect implementation of the timeout did not close the client in the timeout test

## Discussion points

Is 60 second enough for the DB migrations to run?
for now I increase it to 30 minutes to be on the safe side.

### DIFF
--- a/src/lib/util/db-lock.test.ts
+++ b/src/lib/util/db-lock.test.ts
@@ -46,7 +46,7 @@ test('should await other actions on lock', async () => {
 });
 
 test('should handle lock timeout', async () => {
-    const timeoutMs = 1;
+    const timeoutMs = 10;
     let loggedError = '';
     const lock = withDbLock(getDbConfig() as IDBOption, {
         lockKey: 1,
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
This changes db-lock for production, not only for test. I believe (correct me if I'm wrong) that db-lock is only used during migration process, not at runtime.

kwasniew:
@gastonfournier It will affect the production because previously the timeout didn't work. Now it will time out after 60 seconds.

gastonfournier:
> @gastonfournier It will affect the production because previously the timeout didn't work. Now it will time out after 60 seconds.

But only for the import, right? I'm worried that otherwise it may impact other read or write operations by increasing the timeout and perhaps causing a different issue. If it only applies to import, I think it's fine

PR #5109, opened 2023-10-20
feat: Make multiple roles per group/user GA by removing the flag

chriswk:
To prepare for 5.6 GA, 
I've done a find through both Frontend and Backend here to remove the usages of the flag. Seems like the flag was only in use in the frontend. @nunogois can you confirm?


nunogois (approved):
I confirm, that's the only place where it was evaluated.

We can also clean it up from:
 - `frontend/cypress/integration/projects/access.spec.ts`
 - `frontend/src/interfaces/uiConfig.ts`

GJ 👍

chriswk:
will do

PR #5114, opened 2023-10-20
feat: add job that cleans last seen every 24 hours

FredrikOseberg:
This PR adds a cleanup job that removes unknown feature flags from last_seen_at_metrics table every 24 hours since we no longer have a foreign key on the name column in the features table.

### DIFF
--- a/src/lib/services/client-metrics/last-seen/tests/last-seen-service.e2e.test.ts
+++ b/src/lib/services/client-metrics/last-seen/tests/last-seen-service.e2e.test.ts
@@ -1,0 +1,82 @@
+import dbInit, { ITestDb } from '../../../../../test/e2e/helpers/database-init';
+import {
+    IUnleashTest,
+    setupAppWithCustomConfig,
+} from '../../../../../test/e2e/helpers/test-helper';
+import getLogger from '../../../../../test/fixtures/no-logger';
+
+let app: IUnleashTest;
+let db: ITestDb;
+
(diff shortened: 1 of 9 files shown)


src/lib/services/client-metrics/last-seen/tests/last-seen-service.e2e.test.ts
```diff
@@ -0,0 +1,82 @@
...
+            appName: 'test',
+            timestamp: new Date(),
+        };
+    });
+
+    lastSeenService.updateLastSeen(inserts);
+    await lastSeenService.store();
+
+    // We have no method to get these from the last seen service or any other service or store
+    let stored = await db.rawDatabase.raw(
```
FredrikOseberg, line 64: We have no get method for this table

src/lib/services/index.ts
```diff
@@ -110,12 +116,27 @@ export const scheduleServices = async (
...
     } = services;
 
     if (await maintenanceService.isMaintenanceMode()) {
         schedulerService.pause();
     }
 
+    if (flagResolver.isEnabled('useLastSeenRefactor')) {
+        schedulerService.schedule(
+            lastSeenService.cleanLastSeen.bind(lastSeenService),
+            hoursToMilliseconds(24),
```
kwasniew, line 129: Let's change to 1h so that we can run it outside of the app startup

kwasniew (approved):

PR #5118, opened 2023-10-23
fix: don't clean up settings when optional data is not present

gastonfournier:
## About the changes
This fixes a bug updating a project, when optional data (defaultStickiness and featureLimit are not part of the payload).

Example:
```shell
curl -XPUT http://localhost:4242/api/admin/projects/default -H 'Authorization: *:*.unleash-insecure-admin-api-token' -H 'Content-Type: application/json' -d '{"description":"test description","name":"RenamedToThisString"}'
```

Then if you try to get the project with a PAT, the API will return a 403. The reason is that the **project mode** gets cleared (mode=null) because of the following update statement that does not check if defaultStickiness or featureLimit are present, resulting in an update without fields that clears the table https://github.com/Unleash/unleash/blob/main/src/lib/db/project-store.ts#L268-L281

Note that the project still exists, and this can be checked using the initial admin token:
```shell
curl http://localhost:4242/api/admin/projects/default -H 'Authorization: *:*.unleash-insecure-admin-api-token'
```

The problem happens due to:
1. ProjectController does not use the type: UpdateProjectSchema for the request body (will be addressed in another PR in unleash-enterprise)
2. Project Store interface does not match UpdateProjectSchema (but it relies on accepting `additional properties: true`, which is what we agreed on for input)
3. Feature limit is not defined in UpdateProjectSchema (also addressed in the other PR)

### DIFF
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -39,6 +39,7 @@ import {
     ProjectAccessUserRolesDeleted,
     IFeatureNaming,
     CreateProject,
+    IProjectUpdate,
 } from '../types';
 import {
     IProjectQuery,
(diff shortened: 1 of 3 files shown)


src/lib/services/project-service.ts
```diff
@@ -259,16 +260,22 @@ export default class ProjectService {
...
+    async updateProject(
+        updatedProject: IProjectUpdate,
+        user: IUser,
+    ): Promise<void> {
         const preData = await this.projectStore.get(updatedProject.id);
 
         await this.projectStore.update(updatedProject);
 
+        // updated project contains instructions to update the project but it may not represent a whole project
+        const afterData = await this.projectStore.get(updatedProject.id);
```
gastonfournier, line 272: Without this, our diff is a bit misleading:
image (24)

chriswk (approved):
Nice.

PR #5126, opened 2023-10-23
Client instance service

kwasniew:
## About the changes

Moved client instance scheduling to a scheduler service + simplified testing


src/lib/services/index.ts
```diff
@@ -165,6 +165,18 @@ export const scheduleServices = async (
...
+        clientInstanceService.bulkAdd.bind(clientInstanceService),
+        secondsToMilliseconds(5),
+        'bulkAddInstances',
+    );
+
+    schedulerService.schedule(
+        clientInstanceService.announceUnannounced.bind(clientInstanceService),
+        minutesToMilliseconds(5),
+        'announceUnannounced',
+    );
```
FredrikOseberg, line 178: Where these previously configurable by environment variables? Or were the parameters in instance-service.ts never used?

FredrikOseberg (approved):
LGTM, one minor question.

src/lib/services/index.ts
kwasniew: They were default parameters on the constructor that never got overwritten in prod, only in tests

Issue #4475, opened 2023-08-10
FIPS Compatibility - gravatar-url removal / update / replacement to correct md5 usage


TGPSKI:
### Describe the feature request

The imported `gravatar-url` package uses md5 hash method (unsupported in FIPS environments), which prevents Unleash from running where FIPS is enforced. 

```
# package.json

"gravatar-url": "^3.1.0"
```

There's a set of nested dependencies through `sindresorhus` repos, having to follow unleash -> gravatar-url -> md5-hex to actually get a crypto module change.

### Background

https://en.gravatar.com/site/implement/hash/
https://github.com/sindresorhus/md5-hex/blob/main/index.js
https://github.com/aws/constructs/issues/272

This problem is easy to fix in python. I can't find the same option to specify non-cryptographic use cases in node.

### Solution suggestions

* https://www.npmjs.com/package/md5 & replace `gravatar-url`
* remove gravatar support (probably not popular?)

ivarconr:
Thanks for bringing this to our attention. 

I see the easiest path forward to inline the "generateGravatar" function in to the unleash code base. There is not compelling reason to use a library for this, as it basically is a simple md5 of the email. It feels a bit like an unnecessary workaround to please the "FIPS" requirements. I am not sure if there are any way at all to flag this module as safe, as it does not contribute to any cryptographic activities. 

**Next steps:** 
1. Inline the `generateGravatar(email)` as a util function, using the md5 package instead of node built in module ([src](https://github.com/sindresorhus/gravatar-url/blob/main/index.js))
2. Write a few [tests](https://github.com/sindresorhus/gravatar-url/blob/main/index.test-d.ts) for it. 

Any takers?

mahimairaja:
Hi, Is the issue still open?

ivarconr:
Yes, @mahimairaja

mahimairaja:
That great! I would like to work on this issue. Can you please assign this issue to me

mahimairaja:
> **Next steps:**
> 
> 1. Inline the `generateGravatar(email)` as a util function, using the md5 package instead of node built in module ([src](https://github.com/sindresorhus/gravatar-url/blob/main/index.js))
> 2. Write a few [tests](https://github.com/sindresorhus/gravatar-url/blob/main/index.test-d.ts) for it.
> 
> Any takers?


So Can you please guide me. I could understand that I have to implement md5 hashing 
1. but where should I implement it?
2. And what are the test cases are expected to be implemented?

[...]

PR #5128, opened 2023-10-23
feat: replace gravatar-url with inline function

chriswk:
As #4475 says, MD5 is not available in secure places anymore. This PR swaps out gravatar-url with an inline function using crypto:sha256 which is FIPS-140-2 compliant. Since we only used this method for generating avatar URLs the extra customization wasn't needed and we could hard code the URL parameters.
 
fixes: Linear https://linear.app/unleash/issue/SR-112/gh-support-swap-out-gravatar-url-lib
closes: #4475

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -25,7 +25,8 @@
       "suspicious": {
         "noExplicitAny": "off",
         "noExtraNonNullAssertion": "off",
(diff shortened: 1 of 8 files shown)


sighphyre (approved):
LGTM!

PR #5133, opened 2023-10-24
refactor: slight clean up after GAing multiple roles

nunogois:
Follow up cleanup to https://github.com/Unleash/unleash/pull/5109
Based on https://github.com/Unleash/unleash/pull/5109#pullrequestreview-1689535863


chriswk (approved):
Thank you. I thought I caught all of these. :)

PR #5136, opened 2023-10-24
fix: last seen at rendering logic

FredrikOseberg:
This PR fixes a bug where the rendering in the frontend would only render the last seen component if feature.lastSeenAt was set, the new changes considers whether or not environments last seen at is present and takes precedent over the legacy last seen at field.

### DIFF
--- a/frontend/src/timeago.d.ts
+++ b/frontend/src/timeago.d.ts
@@ -1,15 +1,7 @@
 declare module 'react-timeago' {
     import React from 'react';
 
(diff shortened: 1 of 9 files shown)


frontend/src/timeago.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module 'react-timeago' {
```
kwasniew, line 1: can you explain this change?

kwasniew (approved):

PR #5143, opened 2023-10-25
feat: feature search stub

kwasniew:
## About the changes

Adding a stub API for the feature search. A decision is to have a separate subdomain for the feature search for now. In future we may consider merging searches for different parts of the domain.


FredrikOseberg (approved):
Pair programmed

PR #5145, opened 2023-10-25
fix: project mode can not be set to null anymore

sjaanus:
Currently, when there was no entry in project settings table, the project_mode was left null, which broke some logic.
Now we will set it and enforce database level restrictions on it.


chriswk (approved):
LGTM :)

gastonfournier:
Nice!

PR #5146, opened 2023-10-25
Chore/bump dependencies

chriswk:
We're getting security advisories against the node version we're using. This PR bumps everything to 18.18.2/18.x again. And we'll trust our excellent monitoring on detecting performance regressions again


gastonfournier (approved):
LGTM! :safety_vest:

PR #5149, opened 2023-10-25
Feat/add feature search service

FredrikOseberg:
Add feature-search-service increment

### DIFF
--- a/src/lib/features/feature-search/feature-search-service.ts
+++ b/src/lib/features/feature-search/feature-search-service.ts
@@ -1,0 +1,31 @@
+import { Logger } from '../../logger';
+import {
+    IFeatureStrategiesStore,
+    IUnleashConfig,
+    IUnleashStores,
+} from '../../types';
+
+export class FeatureSearchService {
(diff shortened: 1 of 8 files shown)


src/lib/features/feature-search/feature-search-service.ts
```diff
@@ -0,0 +1,31 @@
...
+        this.featureStrategiesStore = featureStrategiesStore;
+        this.logger = getLogger('services/feature-search-service.ts');
+    }
+
+    async search(query: string, tags: string[]) {
+        const features = await this.featureStrategiesStore.getFeatureOverview({
+            projectId: 'default',
+        });
+
+        console.log(features);
```
kwasniew, line 26: can we remove it?

src/lib/features/feature-search/feature-search-service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+        return features;
+        // Search for features
```
kwasniew, line 29: can be removed

kwasniew (approved):
legit

PR #5150, opened 2023-10-25
feat: feature search basic functionality

kwasniew:
## About the changes

Basic search

### DIFF
--- a/src/lib/features/feature-search/feature-search-controller.ts
+++ b/src/lib/features/feature-search/feature-search-controller.ts
@@ -11,11 +11,10 @@ import { Logger } from '../../logger';
 import { createResponseSchema, getStandardResponses } from '../../openapi';
 import { IAuthRequest } from '../../routes/unleash-types';
 import { InvalidOperationError } from '../../error';
-
-interface ISearchQueryParams {
-    query: string;
-    tags: string[];
-}
+import {
+    FeatureSearchQueryParameters,
+    featureSearchQueryParameters,
+} from '../../openapi/spec/feature-search-query-parameters';
 
 const PATH = '/features';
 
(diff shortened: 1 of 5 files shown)


src/lib/openapi/spec/feature-search-query-parameters.ts
```diff
@@ -0,0 +1,28 @@
+import { FromQueryParams } from '../util/from-query-params';
+
+export const featureSearchQueryParameters = [
+    {
+        name: 'query',
+        schema: {
+            default: '',
```
kwasniew, line 7: we need to investigate why the default is not set

src/lib/openapi/spec/feature-search-query-parameters.ts
```diff
@@ -0,0 +1,28 @@
...
+            type: 'string' as const,
+            example: 'feature_a',
+        },
+        description: 'The search query for the feature or tag',
+        in: 'query',
+    },
+    {
+        name: 'projectId',
+        schema: {
+            default: '',
```
kwasniew, line 17: store should do * or %

FredrikOseberg (approved):
LGTM

src/lib/features/feature-search/feature-search-controller.ts
```diff
@@ -56,6 +55,8 @@ export default class FeatureSearchController extends Controller {
                     summary: 'Search and filter features',
                     description: 'Search and filter by selected fields.',
                     operationId: 'searchFeatures',
+                    // TODO: fix the type
+                    parameters: featureSearchQueryParameters as any,
```
kwasniew, line 59: @thomasheartman can you help here? Do we have a standard method for externalizing query params to a type without type assertions?

PR #5153, opened 2023-10-26
refactor: type query params

kwasniew:
## About the changes

Type query params in a way so that we don't need to use type assertions (as X)


thomasheartman (approved):
Looks great! Would you be willing to go ahead and update the other places we cast it too? I think it's only in the old state service 🤔  (A follow-up PR is totally fine).

src/lib/openapi/spec/feature-search-query-parameters.ts
thomasheartman, line 1: Any reason you removed the example in this file? Query parameters should still have examples, I think.

thomasheartman (approved):

PR #5154, opened 2023-10-26
feat: search features without query params

kwasniew:
## About the changes

Can search features without query params. Default values will be used instead.


src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -521,7 +521,10 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
         tag,
         namePrefix,
     }: IFeatureProjectUserParams): Promise<IFeatureOverview[]> {
-        let query = this.db('features').where({ project: projectId });
+        let query = this.db('features');
+        if (projectId) {
```
kwasniew, line 525: this is only used by one existing endpoint. A breaking change is that now if projectId is empty we return all features.

PR #5158, opened 2023-10-26
refactor: optimize queries

FredrikOseberg:
This commit will change the join to only get results for either archived or non-archived features based on the query


andreas-unleash (approved):
LGTM!

PR #5163, opened 2023-10-26
feat: filter by tags

kwasniew:
## About the changes

Filter by tags

### DIFF
--- a/src/lib/features/feature-search/feature-search-controller.ts
+++ b/src/lib/features/feature-search/feature-search-controller.ts
@@ -75,8 +75,8 @@ export default class FeatureSearchController extends Controller {
             const { query, projectId, type, tag } = req.query;
             const userId = req.user.id;
             const normalizedTag = tag
-                ?.map((t) => t.split(':'))
-                .filter((t) => t.length === 2);
+                ?.map((tag) => tag.split(':'))
+                .filter((tag) => tag.length === 2);
             const features = await this.featureSearchService.search({
                 query,
                 projectId,
--- a/src/lib/features/feature-search/feature-search-controller.ts
+++ b/src/lib/features/feature-search/feature-search-controller.ts
@@ -3,6 +3,7 @@ import Controller from '../../routes/controller';
 import { FeatureSearchService, OpenApiService } from '../../services';
 import {
     IFlagResolver,
+    ITag,
     IUnleashConfig,
     IUnleashServices,
     NONE,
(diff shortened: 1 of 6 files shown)


src/lib/features/feature-search/feature-search-controller.ts
```diff
@@ -71,13 +72,17 @@ export default class FeatureSearchController extends Controller {
         res: Response,
     ): Promise<void> {
         if (this.config.flagResolver.isEnabled('featureSearchAPI')) {
-            const { query, projectId, type } = req.query;
+            const { query, projectId, type, tag } = req.query;
             const userId = req.user.id;
+            const normalizedTag = tag
+                ?.map((t) => t.split(':'))
+                .filter((t) => t.length === 2);
             const features = await this.featureSearchService.search({
```
FredrikOseberg, line 80:
```suggestion
                ?.map((tag) => tag.split(':'))
                .filter((tag) => tag.length === 2);
            const features = await this.featureSearchService.search({
```

src/lib/features/feature-search/feature-search-controller.ts
```diff
@@ -71,13 +72,17 @@ export default class FeatureSearchController extends Controller {
         res: Response,
     ): Promise<void> {
         if (this.config.flagResolver.isEnabled('featureSearchAPI')) {
-            const { query, projectId, type } = req.query;
+            const { query, projectId, type, tag } = req.query;
             const userId = req.user.id;
+            const normalizedTag = tag
```
FredrikOseberg, line 77: Is this actually `tags` since it's an array?

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -542,6 +543,13 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
                 .whereILike('features.name', `%${queryString}%`)
                 .orWhereIn('features.name', tagQuery);
         }
+        if (tag && tag.length > 0) {
```
FredrikOseberg, line 546: Again, plural here?

FredrikOseberg (approved):
Small nits, otherwise LG.

src/lib/features/feature-search/feature-search-controller.ts
kwasniew: If we decide to make plural then I'll adjust all query params in one go. For now I prefer consistency

PR #5165, opened 2023-10-26
feat: filter by environment status

kwasniew:
## About the changes

Filter by environment status

* supports many environment statuses
* design is similar to tags
* status is enabled/disabled as in the existing search as opposed to enabled=true|false

`tag[]=simple:tag1&tag[]=simple:tag2` vs `status[]=production:enabled&status[]=development:enabled`

### DIFF
--- a/src/test/e2e/helpers/test-helper.ts
+++ b/src/test/e2e/helpers/test-helper.ts
@@ -47,6 +47,13 @@ export interface IUnleashHttpAPI {
         expectedResponseCode?: number,
     ): supertest.Test;
 
+    enableFeature(
+        feature: string,
+        environment: string,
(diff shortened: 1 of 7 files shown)


src/test/e2e/helpers/test-helper.ts
```diff
@@ -219,6 +226,19 @@ function httpApis(
                 .set('Content-Type', 'application/json')
                 .expect(expectedResponseCode);
         },
+
+        enableFeature(
```
kwasniew, line 230: added helper method to easily enable toggle in a given env

andreas-unleash (approved):
Nice. LGTM!

PR #5167, opened 2023-10-27
docs: update availability due to 5.6 release

nunogois:
Maybe we should update availability in our docs, now that 5.6 is released.


thomasheartman (approved):
Great catch! 😄

PR #5174, opened 2023-10-27
feat: cursor based pagination in search

kwasniew:
## About the changes

Cursor based pagination in feature search

* add millis trick to exclude the previous cursor result
* currently only sorting by created_at is supported. We will extend this to other fields

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
@@ -26,6 +26,7 @@ import { IFeatureProjectUserParams } from './feature-toggle-controller';
 import { Db } from '../../db/db';
 import Raw = Knex.Raw;
 import { IFeatureSearchParams } from './types/feature-toggle-strategies-store-type';
+import { addMilliseconds, format, formatISO, parseISO } from 'date-fns';
 
 const COLUMNS = [
     'id',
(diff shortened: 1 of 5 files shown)


src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -540,9 +543,11 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
                     `%${queryString}%`,
                 ]);
 
-            query = query
-                .whereILike('features.name', `%${queryString}%`)
-                .orWhereIn('features.name', tagQuery);
+            query = query.where((builder) => {
```
kwasniew, line 546: the or query was leaking out so added a nested one

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -571,6 +576,21 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
             });
         }
 
+        // workaround for imprecise timestamp that was including the cursor itself
+        const addMillisecond = (cursor: string) =>
```
kwasniew, line 580: I want to keep it here for now before we learn if it was a good idea

andreas-unleash (approved):
Nice!

PR #5203, opened 2023-10-30
feat: add created_by and status columns to change_request_schedule

andreas-unleash:
Adds 2 columns to change_request_schedule

### DIFF
--- a/src/migrations/20231030091931-add-created-by-and-status-change-request-schedule.js
+++ b/src/migrations/20231030091931-add-created-by-and-status-change-request-schedule.js
@@ -1,0 +1,22 @@
+'use strict';
+
+exports.up = function (db, cb) {
+  db.runSql(`
+        ALTER TABLE change_request_schedule
+            ADD COLUMN IF NOT EXISTS created_by integer REFERENCES users(id);
+
+        ALTER TABLE change_request_schedule
+            ADD COLUMN IF NOT EXISTS status text default 'pending';
+
+    `, cb);
+};
+
+exports.down = function (db, cb) {
+  db.runSql(`
+      ALTER TABLE change_request_schedule
+          DROP COLUMN IF EXISTS created_by;
+
+      ALTER TABLE change_request_schedule
+          DROP COLUMN IF EXISTS status;
+    `, cb);
+};
--- a/src/migrations/20231030091931-add-created-by-and-status-change-request-schedule.js
+++ b/src/migrations/20231030091931-add-created-by-and-status-change-request-schedule.js
@@ -2,11 +2,13 @@
 
 exports.up = function (db, cb) {
   db.runSql(`
+        CREATE TYPE change_request_schedule_status AS ENUM ('pending', 'failed');
+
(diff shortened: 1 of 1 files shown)


src/migrations/20231030091931-add-created-by-and-status-change-request-schedule.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+exports.up = function (db, cb) {
+  db.runSql(`
+        ALTER TABLE change_request_schedule
+            ADD COLUMN IF NOT EXISTS created_by integer REFERENCES users(id);
+
+        ALTER TABLE change_request_schedule
+            ADD COLUMN IF NOT EXISTS status text default 'pending';
```
thomasheartman, line 9: Would we wanna use enum types for this or is text preferable? How do we do it in other places?

src/migrations/20231030091931-add-created-by-and-status-change-request-schedule.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+exports.up = function (db, cb) {
+  db.runSql(`
+        ALTER TABLE change_request_schedule
+            ADD COLUMN IF NOT EXISTS created_by integer REFERENCES users(id);
```
thomasheartman, line 6: Would `scheduled_by` be a better column name than `created_by` in this instance?

```suggestion
            ADD COLUMN IF NOT EXISTS scheduled_by integer REFERENCES users(id);
```

andreas-unleash: either way is fine with me - row was created by someone is what matters

thomasheartman: Got it! I'd lean towards "scheduled_by", but I'll let it be up to you in the end ☺️

src/migrations/20231030091931-add-created-by-and-status-change-request-schedule.js
andreas-unleash: makes sense

thomasheartman: Sure. I think the most important bit is to stay consistent with how we do it other places, so I'd check for that first, and do what we do there. If we have no precedents for this, I think an enum makes sense. But don't trust me blindly on that 😅

thomasheartman (approved):
I'll approve this 😄  It's a fine change, and I'm expecting you to handle the comments in a sensible way. You've seen them, so it's up to you what to do with them. Thanks and good work 🙌🏼

src/migrations/20231030091931-add-created-by-and-status-change-request-schedule.js
ivarconr: https://github.com/Unleash/unleash/commit/6875daf15768d026a53bbc1916fb5c4219fe5680#commitcomment-131584329

PR #5214, opened 2023-10-30
fix: make sure we are still pg10 compatible.

ivarconr:
`EXECUTE FUNCTION` was introduced in Postgres v11. In Postgres v10 the syntax was `EXECUTE PROCEDURE`. This fix changes the syntax to `EXECUTE PROCEDURE`, which is perfectly fine sense our function does not return anything.


chriswk (approved):
LGTM - Tested on PG14 and PG16 and confirmed to work there as well

PR #5230, opened 2023-10-31
feat: Cursor based hateoas

kwasniew:
## About the changes

Generates link to the next page and allows for easy HATEOAS based navigation. 

Changes:
* added utility for calculating next page (eventually we'll move it up the directory tree if needed)
* changed cursor calculation logic to take the cursor of the next item instead of the cursor of the last used item

## Discussion points

Currently only Link to next page is provided and it's easy to parse. We may add extra rel=next later to the header but it complicates parsing

### DIFF
--- a/src/lib/features/feature-search/feature-search-service.ts
+++ b/src/lib/features/feature-search/feature-search-service.ts
@@ -3,6 +3,7 @@ import {
     IFeatureStrategiesStore,
     IUnleashConfig,
     IUnleashStores,
+    serializeDates,
 } from '../../types';
 import { IFeatureSearchParams } from '../feature-toggle/types/feature-toggle-strategies-store-type';
 
@@ -24,6 +25,11 @@ export class FeatureSearchService {
             params,
         );
 
-        return features;
+        const nextCursor =
+            features.length > 0
+                ? features[features.length - 1].createdAt.toJSON()
+                : undefined;
+
+        return { features, nextCursor };
     }
 }
--- a/src/lib/features/feature-search/feature-search-service.ts
+++ b/src/lib/features/feature-search/feature-search-service.ts
@@ -21,15 +21,24 @@ export class FeatureSearchService {
     }
 
     async search(params: IFeatureSearchParams) {
-        const features = await this.featureStrategiesStore.searchFeatures(
-            params,
-        );
+        // fetch one more item than needed to get a cursor of the next item
+        const features = await this.featureStrategiesStore.searchFeatures({
+            ...params,
+            limit: params.limit + 1,
+        });
 
         const nextCursor =
-            features.length > 0
+            features.length > params.limit
                 ? features[features.length - 1].createdAt.toJSON()
                 : undefined;
 
-        return { features, nextCursor };
+        // do not return the items with the next cursor
+        return {
+            features:
+                features.length > params.limit
+                    ? features.slice(0, -1)
+                    : features,
+            nextCursor,
+        };
     }
 }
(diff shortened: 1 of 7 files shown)


src/lib/openapi/spec/feature-search-query-parameters.ts
```diff
@@ -61,17 +61,17 @@ export const featureSearchQueryParameters = [
...
+            'The next feature created at date the client has not seen. Used for cursor-based pagination. Empty if starting from the beginning.',
         in: 'query',
     },
     {
         name: 'limit',
         schema: {
-            type: 'number',
-            example: 10,
+            type: 'string',
+            example: '10',
```
kwasniew, line 74: query params have to be strings

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -533,7 +533,7 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
         cursor,
         limit,
     }: IFeatureSearchParams): Promise<IFeatureOverview[]> {
-        let query = this.db('features');
+        let query = this.db('features').limit(limit);
```
kwasniew, line 536: this may fix our problem with limit not reflecting number of features but a number of feature env rows

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -582,20 +582,10 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
-                addMilliseconds(parseISO(cursor), 1),
-                "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
-            );
         if (cursor) {
-            query = query.where(
-                'features.created_at',
-                '>',
-                addMillisecond(cursor),
-            );
+            query = query.where('features.created_at', '>=', cursor);
```
kwasniew, line 586: in this version we use cursor from the next feature that was not part of the previous list

src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -59,6 +59,10 @@ const searchFeaturesWithCursor = async (
         .expect(expectedCode);
 };
 
+const getPage = async (url: string, expectedCode = 200) => {
```
kwasniew, line 62: used for hateoas based navigation instead of client crafting URLs and calling "endpoints"

src/lib/features/feature-search/feature-search-service.ts
```diff
@@ -20,10 +21,24 @@ export class FeatureSearchService {
     }
 
     async search(params: IFeatureSearchParams) {
-        const features = await this.featureStrategiesStore.searchFeatures(
-            params,
-        );
+        // fetch one more item than needed to get a cursor of the next item
+        const features = await this.featureStrategiesStore.searchFeatures({
+            ...params,
+            limit: params.limit + 1,
```
kwasniew, line 27: we're searching one more feature than we need

src/lib/features/feature-search/feature-search-service.ts
```diff
@@ -20,10 +21,24 @@ export class FeatureSearchService {
...
+        });
 
-        return features;
+        const nextCursor =
```
kwasniew, line 30: we use the last feature to get the cursor if the feature exists

src/lib/features/feature-search/feature-search-service.ts
```diff
@@ -20,10 +21,24 @@ export class FeatureSearchService {
...
+            features.length > params.limit
+                ? features[features.length - 1].createdAt.toJSON()
+                : undefined;
+
+        // do not return the items with the next cursor
+        return {
+            features:
+                features.length > params.limit
+                    ? features.slice(0, -1)
```
kwasniew, line 39: we exclude the next feature that was used for the next cursor calculation

FredrikOseberg (approved):
LGTM

PR #5231, opened 2023-10-31
chore: expose instanceId so it can be used from addons

gastonfournier:
## About the changes
A bit of boy scouting trying to expose the instanceId

### DIFF
--- a/src/lib/services/setting-service.ts
+++ b/src/lib/services/setting-service.ts
@@ -29,8 +29,16 @@ export default class SettingService {
         this.eventService = eventService;
     }
 
-    async get<T>(id: string, defaultValue?: T): Promise<T> {
-        const value = await this.settingStore.get(id);
+    /**
+     * @deprecated use getWithDefault instead
+     */
+    async get<T>(id: string, defaultValue?: T): Promise<T | undefined> {
+        const value = await this.settingStore.get<T>(id);
+        return value || defaultValue;
+    }
+
+    async getWithDefault<T>(id: string, defaultValue: T): Promise<T> {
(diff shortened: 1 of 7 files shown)


src/lib/services/setting-service.ts
```diff
@@ -29,8 +29,16 @@ export default class SettingService {
         this.eventService = eventService;
     }
 
-    async get<T>(id: string, defaultValue?: T): Promise<T> {
```
gastonfournier, line 32: Type was wrong, so we added a new method with the right type called getWithDefault to avoid making this a breaking change

src/lib/services/version-service.ts
```diff
@@ -356,6 +362,7 @@ export default class VersionService {
             current: this.current,
             latest: this.latest || {},
             isLatest: this.isLatest,
+            // @ts-expect-error instance id should always be defined
```
gastonfournier, line 365:
```suggestion
            // @ts-expect-error instance id can be undefined but not on the version. What should we do is still unclear.
```

nunogois (approved):

src/lib/types/stores/settings-store.ts
```diff
@@ -8,4 +8,5 @@ export interface ISettingInsert {
 export interface ISettingStore extends Store<any, string> {
     insert<T>(name: string, content: T): Promise<void>;
     updateRow(name: string, content: any): Promise<void>;
+    get<T>(name: string): Promise<T | undefined>;
```
gastonfournier, line 11: Make type better

PR #5235, opened 2023-11-01
feat: total count in search results

kwasniew:
## About the changes

Return total count of search and filter. In the UI table we need to show all matching results.

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
@@ -532,14 +532,17 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
         status,
         cursor,
         limit,
-    }: IFeatureSearchParams): Promise<IFeatureOverview[]> {
-        let query = this.db('features').limit(limit);
+    }: IFeatureSearchParams): Promise<{
+        features: IFeatureOverview[];
+        total: number;
+    }> {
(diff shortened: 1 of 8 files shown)


src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -670,14 +670,24 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
             ];
         }
 
-        query = query.select(selectColumns);
+        if (cursor) {
```
kwasniew, line 673: moved where and order by to match logical order of SQL queries and exclude them from the total count

src/lib/openapi/spec/feature-search-query-parameters.ts
```diff
@@ -71,10 +71,10 @@ export const featureSearchQueryParameters = [
         name: 'limit',
         schema: {
             type: 'string',
-            example: '10',
+            example: '50',
         },
         description:
-            'The number of results to return in a page. By default it is set to 50',
+            'The number of feature environments to return in a page. By default it is set to 50.',
```
kwasniew, line 77: We're explicit about it now. We can fix the limit to reflect the number of features not feature environments but:
* it's not trivial
* it's nice to not return too many results when someone has 10+ environments. With the current approach we have a natural limit for envs.

FredrikOseberg (approved):
LGTM

PR #5236, opened 2023-11-01
fix: do not allow segment deletion when used in private projects

sjaanus:
Splitted the `getStrategies` into `getVisibleStrategies` and `getAllStrategies`, which will handle the permissions.
If we are deleting a segment, `getAllStrategies` will be used and private projects will not be counted.

### DIFF
--- a/src/lib/segments/segment-service-interface.ts
+++ b/src/lib/segments/segment-service-interface.ts
@@ -19,7 +19,7 @@ export interface ISegmentService {
      * For most use cases, use `getVisibleStrategies`
      * @param id segment id
      */
-    getStrategies(id: number): Promise<IFeatureStrategy[]>;
+    getAllStrategies(id: number): Promise<IFeatureStrategy[]>;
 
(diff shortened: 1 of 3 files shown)


src/lib/segments/segment-service-interface.ts
```diff
@@ -13,7 +13,18 @@ export interface ISegmentService {
 
     get(id: number): Promise<ISegment>;
 
-    getStrategies(id: number, userId: number): Promise<IFeatureStrategy[]>;
+    /**
+     * Gets all strategies for a segment
+     * This is NOT considering the private projects
+     * For most use cases, use `getVisibleStrategies`
+     * @param id segment id
```
kwasniew, line 20: do we need this @param is we have typescript?

kwasniew (approved):

src/lib/segments/segment-service-interface.ts
sjaanus: Sure, no.

PR #5237, opened 2023-11-01
feat: connect project overview table to search api

kwasniew:
## About the changes

Connecting project overview features list to the search API. Currently we only return the first results with no cursor increments. You can see that even though I have hundreds of feature we only show 26 results.
Screenshot 2023-11-01 at 11 05 32

To make it work I had to refactor existing component for a table to allow for configurable onChange. We can either refresh project API or search API in this onChange handler.

The code is based on a new flag:
Screenshot 2023-11-01 at 11 19 09

Next PRs:
* add infinite loader and cursor changes
* translate search and filter into query params
* sort server side instead of client side
* add tests

### DIFF
--- a/frontend/src/component/project/Project/ProjectOverview.tsx
+++ b/frontend/src/component/project/Project/ProjectOverview.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import useProject, {
     useProjectNameOrId,
 } from 'hooks/api/getters/useProject/useProject';
@@ -12,6 +12,7 @@ import { useLastViewedProject } from 'hooks/useLastViewedProject';
 import { ProjectStats } from './ProjectStats/ProjectStats';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 import { useUiFlag } from 'hooks/useUiFlag';
+import { useFeatureSearch } from '../../../hooks/api/getters/useFeatureSearch/useFeatureSearch';
 
 const refreshInterval = 15 * 1000;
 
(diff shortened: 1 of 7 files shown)


frontend/src/component/project/Project/ProjectOverview.tsx
```diff
@@ -34,22 +35,62 @@ const StyledContentContainer = styled(Box)(() => ({
...
+                description={description}
+                memberCount={members}
+                health={health}
+                features={features}
+                stats={stats}
+            />
+            <StyledContentContainer>
+                <ProjectStats stats={project.stats} />
+                <StyledProjectToggles>
+                    <ProjectFeatureToggles
```
kwasniew, line 61: this is the same as previous code but now we can inject search features, search loading state and search refetch to refresh search after toggle flip

frontend/src/component/project/Project/ProjectOverview.tsx
```diff
@@ -34,22 +35,62 @@ const StyledContentContainer = styled(Box)(() => ({
...
     usePageTitle(`Project overview – ${projectName}`);
     const { setLastViewed } = useLastViewedProject();
     const featureSwitchRefactor = useUiFlag('featureSwitchRefactor');
+    const featureSearchFrontend = useUiFlag('featureSearchFrontend');
 
     useEffect(() => {
         setLastViewed(projectId);
     }, [projectId, setLastViewed]);
 
+    if(featureSearchFrontend) return <InfiniteProjectOverview />;
```
kwasniew, line 92: This is a technique I decided to use to early exit with new code when the flag is on. Otherwise the old code continues.

FredrikOseberg (approved):
LGTM

PR #5240, opened 2023-11-01
docs: Update recommended postgres to 14

chriswk:
We've been caught out here, we added a migration that used PG>=11 syntax, but our docs still state that we support v10, when it reached EoL Nov 2022. This updates our docs to state v14 and newer, v14 came out in 2021 and as such won't be EoL until 2026 according to [Postgres 5 year versioning scheme support](https://www.postgresql.org/support/versioning/)

### DIFF
--- a/website/docs/contributing/backend/overview.md
+++ b/website/docs/contributing/backend/overview.md
@@ -15,7 +15,7 @@ We have created a set of ADRs to help guide the development of the backend:
 
 Before developing on this project you will need two things:
 
-- PostgreSQL 12.x or newer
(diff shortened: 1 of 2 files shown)


chriswk:
We won't check, and we've fixed our one known issue with v10 for now, as such this is just a documentation upgrade to save trouble later down the line when we release v6 of Unleash and actually formally drop support for anything older than v14.

thomasheartman (approved):
Good catch!

ivarconr (approved):

ivarconr (approved):

PR #5241, opened 2023-11-01
refactor: memoize heavy components in project overview

FredrikOseberg:
This PR memoizes some of the heavier components in our project overview table


kwasniew (approved):
Nice. I'd still verify memory usage in browser dumps. Memoization trades off rendering speed for more memory usage and I'd like to understand how many more MBs do we use for a table with e.g. 500 toggles.

FredrikOseberg:
> Nice. I'd still verify memory usage in browser dumps. Memoization trades off rendering speed for more memory usage and I'd like to understand how many more MBs do we use for a table with e.g. 500 toggles.

Skjermbilde 2023-11-02 kl  13 10 26

Here's a heap comparison. I did some more testing and it does increase a bit as you scroll the list, but we are well within the parameters of acceptable memory usage in the application total. A window.performance.memory dump shows that we are using around 10% of available memory for our entire application on the most expensive pages.

PR #5242, opened 2023-11-01
feat: infinite scroll API trigger

kwasniew:
## About the changes

When scrolling to the bottom of the features table, we trigger the next request based on the recorded cursor

Also fixes:
* limit respects that features belong to many environments
* displaying total items in the table results

## Discussion points

still haven't decided on cursor pagination patterns:
* we can have next link and use it as a key
* we can have next cursors from the API (header or data field)

### DIFF
--- a/frontend/src/component/project/Project/ProjectOverview.tsx
+++ b/frontend/src/component/project/Project/ProjectOverview.tsx
@@ -12,9 +12,9 @@ import { useLastViewedProject } from 'hooks/useLastViewedProject';
 import { ProjectStats } from './ProjectStats/ProjectStats';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 import { useUiFlag } from 'hooks/useUiFlag';
-import { useFeatureSearch } from '../../../hooks/api/getters/useFeatureSearch/useFeatureSearch';
-import { useOnVisible } from '../../../hooks/useOnVisible';
-import { IFeatureToggleListItem } from '../../../interfaces/featureToggle';
+import { useFeatureSearch } from 'hooks/api/getters/useFeatureSearch/useFeatureSearch';
+import { useOnVisible } from 'hooks/useOnVisible';
+import { IFeatureToggleListItem } from 'interfaces/featureToggle';
 
 const refreshInterval = 15 * 1000;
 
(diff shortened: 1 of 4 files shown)


frontend/src/component/project/Project/ProjectOverview.tsx
```diff
@@ -40,14 +42,27 @@ const InfiniteProjectOverview = () => {
...
+    } = useFeatureSearch(cursor, projectId, { refreshInterval });
+
     const { members, features, health, description, environments, stats } =
         project;
+    const fetchNextPageRef = useOnVisible<HTMLDivElement>(() => {
+        if(!loading && nextCursor !== cursor && nextCursor !== '') {
+            setCursor(nextCursor);
+        }
+    });
+    const [dataList, setDataList] = useState<IFeatureToggleListItem[]>([]);
```
kwasniew, line 61: this is where we accumulate data items we got so far

frontend/src/component/project/Project/ProjectOverview.tsx
```diff
@@ -40,14 +42,27 @@ const InfiniteProjectOverview = () => {
...
+        nextCursor,
+        total,
         refetch,
         loading,
-    } = useFeatureSearch(nextCursor, projectId, { refreshInterval });
+    } = useFeatureSearch(cursor, projectId, { refreshInterval });
+
     const { members, features, health, description, environments, stats } =
         project;
+    const fetchNextPageRef = useOnVisible<HTMLDivElement>(() => {
```
kwasniew, line 56: when the load more marker is reached we update the cursor. We also need to prevent multiple requests so checking for the cursor change

frontend/src/component/project/Project/ProjectOverview.tsx
```diff
@@ -63,12 +78,18 @@ const InfiniteProjectOverview = () => {
...
-                        features={searchFeatures.features}
+                        key={
+                            loading && dataList.length === 0
+                                ? 'loading'
+                                : 'ready'
+                        }
+                        features={dataList}
                         environments={environments}
-                        loading={loading}
+                        loading={loading && dataList.length === 0}
```
FredrikOseberg, line 88: What if the project does not have any features?

FredrikOseberg (approved):
Small question

frontend/src/component/project/Project/ProjectOverview.tsx
kwasniew: No feature toggles available screen as before

PR #5244, opened 2023-11-01
Docs: Client-side feature flag querying behavior update

markunl:
Added a detailed explanation into how frontend / client-side SDKs behave with regards to fetching multiple feature flags and network connectivity.

Content has been reviewed/approved by engineering rotation team (Simon and Chris).

### DIFF
--- a/website/docs/reference/sdks/index.md
+++ b/website/docs/reference/sdks/index.md
@@ -137,11 +137,24 @@ If you can't find an SDK that fits your need, you can also develop your own SDK.
 - [Unleash Client Specifications](https://github.com/Unleash/client-specification) - Used by all official SDKs to make sure they behave correctly across different language implementations. This lets us verify that a gradual rollout to 10% of the users would affect the same users regardless of which SDK you're using.
 - [Client SDK overview](../client-specification) - A brief, overall guide of the _Unleash Architecture_ and important aspects of the SDK role in it all.
 
+
+## Client-side SDK Behavior
+
+The following section details the behavior of frontend / client-side SDKs when initializing and fetching flags with respect to network connectivity.  
+
+When the SDK is initialized in the application, an in memory repository is setup and synchronized against the frontent API using the configured token and context. Note that the frontend API is hosted by either the Unleash Proxy/Edge or the upstream Unleash instance directly.  
+
+1. All feature flag evaluation is performed by the Proxy/Edge or Unleash instance. A payload of all enabled flags and their variants (if applicable) is returned as a single request. Disabled flags are not included.
+
+2. When a page inside the application requests a feature flag, the SDK will return the flag state from memory. No network connection to the frontend API is performed.
+   
+3. The SDK periodically syncs with the frontend API to retrieve the latest set of enabled toggles
+  
 ## Working offline
 
 Once they have been initialized, all Unleash clients will continue to work perfectly well without an internet connection or in the event that the Unleash Server has an outage.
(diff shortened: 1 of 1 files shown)


website/docs/reference/sdks/index.md
```diff
@@ -137,11 +137,24 @@ If you can't find an SDK that fits your need, you can also develop your own SDK.
...
 - [Client SDK overview](../client-specification) - A brief, overall guide of the _Unleash Architecture_ and important aspects of the SDK role in it all.
 
+
+## Client-side SDK behavior
+
+The following section details the behavior of frontend / client-side SDKs when initializing and fetching flags with respect to network connectivity.  
+
+When the SDK is initialized in the application, an in memory repository is setup and synchronized against the frontent API using the configured token and context. Note that the frontend API is hosted by either the Unleash Proxy/Edge or the upstream Unleash instance directly.  
+
+1. All feature flag evaluation is performed by the Proxy/Edge or Unleash instance. A payload of all enabled flags and their variants (if applicable) is returned as a single request. Disabled flags are not included.
```
sighphyre, line 147: I think what you have is great here, this is a very specific detail so feel free to ignore if you think what you already have is clearer

```suggestion
1. All feature flag evaluation is performed by the Proxy/Edge or Unleash instance. A payload of all enabled flags and their resolved variant (if applicable) is returned as a single request. Disabled flags are not included.
```

sighphyre (approved):
This is fantastic! Thank you for doing this!

website/docs/reference/sdks/index.md
thomasheartman: I agree with @sighphyre here. I think that makes it clearer (as long as the reader understands what "resolved" means. But I might pluralize the word "variant" as well, turning it into "resolved variants":

```suggestion
1. All feature flag evaluation is performed by the Proxy/Edge or Unleash instance. A payload of all enabled flags and their resolved variants (if applicable) is returned as a single request. Disabled flags are not included.
```

But yeah, that's very nitpicky, and I don't know if my suggestion is better. Just wanted to put it out there 💁🏼

thomasheartman (approved):
This is a great clarification 🙌🏼

website/docs/reference/sdks/index.md
sighphyre: The reason I say "variants" rather than "variant" is because the response will only ever return a single variant. You can't use this to check what variants may be available, you just get the calculated single value and never a list

thomasheartman: Yeah, no, I figured, but you might get several variants spread over different flags. Pluralizing variants does make it slightly ambiguous, maybe, but you could also misread the singular variant as something like "a set of flags can only resolve to a single variant together". Maybe less likely 💁🏼  But I'm honestly happy with either, just wanted to present a suggestion. Either way, I think adding "resolved" is a fantastic shout, so good catch there 👏🏼 👏🏼 👏🏼

markunl: Thanks both! I went with Simon's suggestion since we're already pluralizing the term "flags". An individual flag may only have one resolved variant, so not pluralizing variants may have slightly less potential for misinterpretation

markunl (commented):
Thanks both! I went with Simon's suggestion since we're already pluralizing the term "flags". An individual flag may only have one resolved variant, so not pluralizing variants may have slightly less potential for misinterpretation

PR #5251, opened 2023-11-02
chore: expose types so we can use them properly

gastonfournier:
Expose types to be used in enterprise and cloud addons

### DIFF
--- a/src/lib/middleware/api-token-middleware.test.ts
+++ b/src/lib/middleware/api-token-middleware.test.ts
@@ -6,24 +6,25 @@ import { ALL, ApiTokenType } from '../types/models/api-token';
 import apiTokenMiddleware, {
     TOKEN_TYPE_ERROR_MESSAGE,
 } from './api-token-middleware';
+import { ApiTokenService } from 'lib/services';
+import { IAuthType, IUnleashConfig } from 'lib/types';
 
-let config: any;
+let config: IUnleashConfig;
 
 beforeEach(() => {
-    config = {
+    config = createTestConfig({
         getLogger,
         authentication: {
             enableApiToken: true,
         },
-    };
+    });
 });
 
 test('should not do anything if request does not contain a authorization', async () => {
     const apiTokenService = {
         getUserForToken: jest.fn(),
-    };
+    } as unknown as ApiTokenService;
 
-    // @ts-expect-error missing methods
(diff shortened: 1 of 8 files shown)


src/lib/types/api-user.ts
```diff
@@ -45,7 +53,7 @@ export default class ApiUser {
         if (projects && projects.length > 0) {
             this.projects = projects;
         } else {
-            this.projects = [project];
+            this.projects = project ? [project] : [];
```
gastonfournier, line 56: This is slightly better, previously we were returning a list with a single undefined element

src/lib/middleware/api-token-middleware.test.ts
```diff
@@ -228,6 +235,7 @@ test('should call next if apiTokenService throws x2', async () => {
         },
     };
 
+    // @ts-expect-error missing methods
```
gastonfournier, line 238: After properly defining the type at `{ apiTokenService }: any,` we get compiler errors here because we're just defining a single method

src/lib/middleware/api-token-middleware.ts
```diff
@@ -42,14 +43,14 @@ const apiAccessMiddleware = (
         return (req, res, next) => next();
     }
 
-    return (req: IAuthRequest, res, next) => {
+    return (req: IAuthRequest | IApiRequest, res, next) => {
```
gastonfournier, line 46: No we need to define either Auth request or Api request

src/lib/routes/unleash-types.ts
```diff
@@ -1,13 +1,25 @@
...
+    session: any;
+}
+
+export interface IApiRequest<
+    PARAM = any,
+    ResBody = any,
+    ReqBody = any,
+    ReqQuery = any,
+> extends Request<PARAM, ResBody, ReqBody, ReqQuery> {
+    user: IApiUser;
```
gastonfournier, line 22: Define another interface. The main difference is that some endpoints use the `user.id` to associate the request to a user, but tokens don't have an id associated. This is the part that I think we should look at carefully...

src/lib/middleware/api-token-middleware.test.ts
gastonfournier: Solved this in a more elegant way

chriswk (approved):
This looks sane to me.

PR #5252, opened 2023-11-02
chore: add schedule option to approved change requests

thomasheartman:
The button doesn't do anything at the moment, but it's there visually.

Because this uses the same button as the dual-function button for approve/reject, I extracted that component into a reusable "multi-action" button. I could have copied the code wholesale, but it's a complex component, so I thought this would be a better solution.

I'll add the dialog in a follow-up PR. This one already has a lot of changes.

Visual:

image


frontend/src/component/changeRequest/ChangeRequestOverview/ApplyButton/ApplyButton.tsx
thomasheartman, line 1: This is the new apply button that also has the schedule option. It uses the extract multi-action button.

andreas-unleash (approved):
LGTM

PR #5259, opened 2023-11-03
refactor: switch to upstream express-openapi

thomasheartman:
Switch the express-openapi implementation from our internal fork to the upstream version. We have upstreamed our changes and a new version has been released, so this should be the last step before we can retire our fork.

Because some of the dependencies have been updated since our internal fork, we also need to update some of our error handling to reflect this.

### DIFF
--- a/src/lib/error/bad-data-error.ts
+++ b/src/lib/error/bad-data-error.ts
@@ -129,7 +129,9 @@ export const fromOpenApiValidationError =
     (requestBody: object) =>
     (validationError: ActualErrorObject): ValidationErrorDescription => {
         const { instancePath, params, message, dataPath } = validationError;
-        const propertyName = dataPath?.substring('.body.'.length) ?? '';
+        const propertyName =
+            dataPath?.substring('.body.'.length) ??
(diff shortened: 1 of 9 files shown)


src/lib/openapi/spec/patch-schema.ts
thomasheartman, line 1: The updated deps were complaining that you can't use a `nullable` attribute without a type. But if you have no type, `null` is already a valid value, so that's all good!

src/lib/error/bad-data-error.ts
```diff
@@ -101,32 +101,62 @@ const oneOfMessage = (
...
+
+    return {
+        description,
+        message: description,
+        path: propertyName,
+    };
+};
+
+// Sometimes, the error object contains a dataPath, even if it's not
+type ActualErrorObject = ErrorObject & { dataPath?: string };
```
thomasheartman, line 126: I think it's possible that the data that previously resided in `dataPath` is now available in `instancePath`, so I'll try to make a follow-up PR where I fix that.

andreas-unleash (approved):
LG

PR #5261, opened 2023-11-03
feat: scheduled change request state

andreas-unleash:
Adds the scheduled state to ChangeRequestOverview.tsx

Screenshot 2023-11-03 at 12 52 07

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ApplyButton/ApplyButton.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ApplyButton/ApplyButton.tsx
@@ -9,7 +9,8 @@ export const ApplyButton: FC<{
     disabled: boolean;
     onSchedule: () => void;
     onApply: () => void;
-}> = ({ disabled, onSchedule, onApply, children }) => (
+    variant?: 'create' | 'update';
+}> = ({ disabled, onSchedule, onApply, variant = 'create', children }) => (
     <MultiActionButton
         permission={APPLY_CHANGE_REQUEST}
         disabled={disabled}
@@ -20,7 +21,10 @@ export const ApplyButton: FC<{
                 icon: <CheckBox fontSize='small' />,
             },
             {
-                label: 'Schedule changes',
+                label:
+                    variant === 'create'
+                        ? 'Schedule changes'
+                        : 'Update schedule',
                 onSelect: onSchedule,
                 icon: <Today fontSize='small' />,
             },
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
@@ -356,7 +356,7 @@ export const ChangeRequestOverview: FC = () => {
                                         variant='outlined'
                                         onClick={onCancel}
                                     >
-                                        {Boolean(changeRequest.schedule)
+                                        {changeRequest.schedule
                                             ? 'Reject'
                                             : 'Cancel'}{' '}
                                         changes
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
@@ -26,7 +26,6 @@ import { ChangeRequestReviewers } from './ChangeRequestReviewers/ChangeRequestRe
 import { ChangeRequestRejectDialogue } from './ChangeRequestRejectDialog/ChangeRequestRejectDialog';
 import { ApplyButton } from './ApplyButton/ApplyButton';
 import { useUiFlag } from 'hooks/useUiFlag';
-import { scheduler } from 'timers/promises';
 
 const StyledAsideBox = styled(Box)(({ theme }) => ({
     width: '30%',
@@ -76,16 +75,15 @@ export const ChangeRequestOverview: FC = () => {
 
     const id = useRequiredPathParam('id');
     const { data: changeRequest, refetchChangeRequest } = useChangeRequest(
-        projectId,
-        id,
+      projectId,
+      id,
     );
     const { changeState, addComment, loading } = useChangeRequestApi();
     const { refetch: refetchChangeRequestOpen } =
-        usePendingChangeRequests(projectId);
+      usePendingChangeRequests(projectId);
     const { setToastData, setToastApiError } = useToast();
     const { isChangeRequestConfiguredForReview } =
-        useChangeRequestsEnabled(projectId);
-
+      useChangeRequestsEnabled(projectId);
     const scheduleChangeRequests = useUiFlag('scheduledConfigurationChanges');
 
     if (!changeRequest) {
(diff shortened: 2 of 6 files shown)


frontend/src/component/changeRequest/ChangeRequestOverview/ApplyButton/ApplyButton.tsx
```diff
@@ -9,7 +9,8 @@ export const ApplyButton: FC<{
     disabled: boolean;
     onSchedule: () => void;
     onApply: () => void;
-}> = ({ disabled, onSchedule, onApply, children }) => (
+    variant?: 'create' | 'update';
+}> = ({ disabled, onSchedule, onApply, variant = 'create', children }) => (
```
thomasheartman, line 13: Hmm. I wonder if this should be handled outside the component? Or maybe a different button entirely? Or maybe the component should be renamed 🤔 Regardless, that can be done in a follow-up PR, so it's not a blocking thing.

But maybe `variant` should be a required prop instead?

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
```diff
@@ -191,7 +189,7 @@ export const ChangeRequestOverview: FC = () => {
         changeRequest.state === 'In review' &&
         !isAdmin;
 
-    const hasApprovedAlready = changeRequest.approvals.some(
+    const hasApprovedAlready = changeRequest.approvals?.some(
```
thomasheartman, line 192: Is `approvals` ever not present? Or has something changed now?

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestReviewStatus/ChangeRequestReviewStatus.tsx
```diff
@@ -109,6 +112,12 @@ const ResolveComponent = ({ changeRequest }: IResolveComponentProps) => {
         return <Rejected />;
     }
 
+    if (state === 'Scheduled') {
+        return (
+            <Scheduled scheduledDate={changeRequest.schedule?.scheduledAt} />
```
thomasheartman, line 117: Is the `?` necessary here? Isn't `schedule` required if the state is `Scheduled`? Or are there cases where it might not be present?

```suggestion
            <Scheduled scheduledDate={changeRequest.schedule.scheduledAt} />
```

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestReviewStatus/ChangeRequestReviewStatus.tsx
```diff
@@ -194,6 +203,69 @@ const Applied = () => {
...
+        if (offset < 0) {
+            sign = '-';
+        }
+
+        // Ensure that hours and minutes are two digits
+        const zeroPaddedHours = hours.toString().padStart(2, '0');
+        const zeroPaddedMinutes = minutes.toString().padStart(2, '0');
+
+        return `UTC${sign}${zeroPaddedHours}:${zeroPaddedMinutes}`;
+    };
```
thomasheartman, line 230: What is this doing? Feels like this should be extracted into a separate function (or even file) and tested. Or if we're just getting the timezone, could we use `Intl.DateTimeFormat().resolvedOptions().timeZone` instead? It won't give the offset, but it will give the time zone, which might be more important. There might also be cases where it's more correct, such as if you're thinking about daylight saving time. The `getBrowserTimezone` won't account for that, but using `Intl.DateTimeFormat().resolvedOptions().timeZone` will, because you're always working with the user's current time zone, and not its offset from UTC (which can change).

thomasheartman (approved):
I think this is great! I've got a few comments, but because this is all behind a flag, we can fix it in future PRs if you'd like to get this in. Most are just minor comments, but I think we should be very thoughtful about how we're dealing with time zones.

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestReviewStatus/ChangeRequestReviewStatus.tsx
thomasheartman: But thinking about it: how do we store times? UTC-times? Regardless, think we should try to avoid doing as much time zone math as possible

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestReviewStatus/ChangeRequestReviewStatus.tsx
andreas-unleash: Because `schedule` can be undefined - the ts compiler is complaining. 
I can either:
- make the `scheduledDate` prop with a `?`
- make `changeRequest.schedule!.scheduledAt`
or leave it as is

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestReviewStatus/ChangeRequestReviewStatus.tsx
andreas-unleash: `Intl.DateTimeFormat().resolvedOptions().timeZone` would give eg. Asia/Pasific or GMT+2

I couldn't find a way to format it as requested

`getTimezoneOffset` returns the offset in minutes from GMT


`hours` calculated from minutes / 60

`minutes` any remainder from the hours calculation (eg. India is UTC+05:30)

Finally the desired format is created by ensuring 2 digits (padding with 0) and putting the different parts together

andreas-unleash: I will move to its own file and write a few tests thanks

thomasheartman: Yeah, so for me, `Intl.DateTimeFormat().resolvedOptions().timeZone` gives "Europe/Oslo". 

You're right in that it doesn't give you a UTC offset. But I think that's better. The scary thing about using offsets like this is that the offset you're calculating here is based on the current time, while the time could be in the future. So if that goes over daylight saving times, I think it would be wrong.

Say you want it to happen at 10:00 UTC-2 in two weeks. But there's a clock change in between, so you're currently at UTC-3. So you think you need to set it to be at 11:00 in two weeks. That way it says 11:00 UTC-3, but it's actually 11:00 UTC-2 because the date picker knows about this. In this case 10:00 Europe/Oslo is more accurate.

andreas-unleash: ```The time-zone offset is the difference, in minutes, between UTC and local time. Note that this means that the offset is positive if the local timezone is behind UTC and negative if it is ahead. For example, for time zones that are ahead of UTC, the value would be negative because local time is added to the UTC to convert it to local time.```

thomasheartman: Sure, I agree with that, but I don't think it adequately handles time changes, such as DST. And I'm afraid that's gonna cause issues down the line.

PR #5264, opened 2023-11-03
fix: last seen deadlocks

FredrikOseberg:
Sort the items before inserting them into the database in order to reduce the chance of deadlocks happening when multiple pods are inserting at the same time.


andreas-unleash (approved):
LGTM

PR #5267, opened 2023-11-03
Feat: scheduled change request dialogs

andreas-unleash:
Creates the Apply and Reject scheduled change request dialogs

Closes # [1-1584](https://linear.app/unleash/issue/1-1584/add-modal-for-apply-now)

Closes # [1-1586](https://linear.app/unleash/issue/1-1586/reject-changes-dialog)

Screenshot 2023-11-03 at 14 43 17
Screenshot 2023-11-03 at 14 43 28

UI e2e tests will be in a follow up PR

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestScheduledDialogs/ChangeRequestScheduledDialogue.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestScheduledDialogs/ChangeRequestScheduledDialogue.tsx
@@ -1,0 +1,59 @@
+import { FC, ReactElement } from 'react';
+import { Alert, styled, Typography } from '@mui/material';
+import { Dialogue } from '../../../common/Dialogue/Dialogue';
+
+export interface ChangeRequestScheduleDialogueProps {
+    title: string;
+    primaryButtonText: string;
+    open: boolean;
+    onConfirm: () => void;
+    onClose: () => void;
+    scheduledTime?: string;
+    message: string;
+    permissionButton?: ReactElement;
+    disabled?: boolean;
+    projectId?: string;
+    environment?: string;
+}
+
+const StyledAlert = styled(Alert)(({ theme }) => ({
+    marginBottom: theme.spacing(2),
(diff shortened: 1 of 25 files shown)


frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestScheduledDialogs/ChangeRequestScheduledDialogue.tsx
```diff
@@ -0,0 +1,59 @@
...
+            secondaryButtonText='Cancel'
+            open={open}
+            onClose={onClose}
+            onClick={() => onConfirm()}
+            fullWidth
+        >
+            <StyledAlert icon={false}>
+                There is a scheduled time to apply these changes set for{' '}
+                <strong>
+                    <br />
```
thomasheartman, line 52: Why are we using the `br` element? If we need to space this out, we should do that with styling, not html, I think.

thomasheartman (approved):
Looks pretty good! I'd prefer to avoid using HTML for styling (so don't use the `br` element), but otherwise good. I've already voiced my concerns for the time handling, but we don't need to address that just yet.

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestScheduledDialogs/ChangeRequestScheduledDialogue.tsx
andreas-unleash: The reason I user `br` is that the designs look to me to require it be in a new line.  What would be a better way to do it?

thomasheartman: Yeah, this sounds like a styling thing to me. The way I've always learnt is that visual-only things should be done with CSS, not HTML. So for instance, if you need the `strong` element to be on its own line, you could use css to give it `display: block` instead. It's not a super important thing, but more of a ... best practices thing. At least as far as I'm aware.

PR #5268, opened 2023-11-03
chore: expose type and more fixes

gastonfournier:
Expose new interface while also getting rid of unneeded compiler ignores

None of the changes should add new security risks, despite this report:
> Code scanning results / CodeQL Failing after 4s — 2 new alerts including 2 high severity security vulnerabilities

Not sure what that means, maybe a removed ignore...

### DIFF
--- a/src/lib/services/proxy-service.ts
+++ b/src/lib/services/proxy-service.ts
@@ -1,7 +1,7 @@
 import { IUnleashConfig, IUnleashServices, IUnleashStores } from '../types';
 import { Logger } from '../logger';
 import { ClientMetricsSchema, ProxyFeatureSchema } from '../openapi';
-import ApiUser from '../types/api-user';
+import ApiUser, { IApiUser } from '../types/api-user';
 import {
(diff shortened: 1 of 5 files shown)


src/lib/services/proxy-service.ts
```diff
@@ -85,15 +85,15 @@ export class ProxyService {
...
+        token: IApiUser,
         metrics: ClientMetricsSchema,
         ip: string,
     ): Promise<void> {
         ProxyService.assertExpectedTokenType(token);
 
         const environment =
             this.services.clientMetricsServiceV2.resolveMetricsEnvironment(
-                token,
+                token as ApiUser,
```
gastonfournier, line 96: Not so easy to change cause we're testing `user instanceof ApiUser` inside resolveMetricsEnvironment. We can't just use isAPI because we also have users who are APIs (service accounts IIRC)

chriswk (approved):
This looks like an improved to QoL, I'm all for it.

PR #5269, opened 2023-11-03
chore: import IUser instead of User for auth

gastonfournier:
## About the changes
A very subtle change in the way we import IUser makes a huge difference because previously, instead of importing IUser interface we were importing User and naming it IUser here:
https://github.com/Unleash/unleash/blob/6f8f21fd488cfbf53a3526ffa138e40c3a43e27c/src/lib/routes/unleash-types.ts#L2
whereas the correct way of importing the interface is:
https://github.com/Unleash/unleash/blob/eec64b119e133813e4f4c4a42e9bdd0a8db65259/src/lib/routes/unleash-types.ts#L2

### DIFF
--- a/src/lib/routes/client-api/register.ts
+++ b/src/lib/routes/client-api/register.ts
@@ -4,7 +4,7 @@ import { IUnleashServices } from '../../types';
 import { IUnleashConfig } from '../../types/option';
 import { Logger } from '../../logger';
 import ClientInstanceService from '../../services/client-metrics/instance-service';
-import { IAuthRequest, IUser, User } from '../../server-impl';
+import { IAuthRequest, IUser } from '../../server-impl';
 import { IClientApp } from '../../types/model';
 import ApiUser, { IApiUser } from '../../types/api-user';
 import { ALL } from '../../types/models/api-token';
--- a/src/lib/routes/client-api/register.ts
+++ b/src/lib/routes/client-api/register.ts
@@ -4,9 +4,9 @@ import { IUnleashServices } from '../../types';
 import { IUnleashConfig } from '../../types/option';
 import { Logger } from '../../logger';
 import ClientInstanceService from '../../services/client-metrics/instance-service';
-import { IAuthRequest, User } from '../../server-impl';
+import { IAuthRequest, IUser, User } from '../../server-impl';
 import { IClientApp } from '../../types/model';
-import ApiUser from '../../types/api-user';
+import ApiUser, { IApiUser } from '../../types/api-user';
 import { ALL } from '../../types/models/api-token';
 import { NONE } from '../../types/permissions';
 import { OpenApiService } from '../../services/openapi-service';
@@ -61,7 +61,10 @@ export default class RegisterController extends Controller {
         });
     }
 
-    private static resolveEnvironment(user: User, data: Partial<IClientApp>) {
+    private static resolveEnvironment(
+        user: IUser | IApiUser,
+        data: Partial<IClientApp>,
+    ) {
         if (user instanceof ApiUser) {
             if (user.environment !== ALL) {
                 return user.environment;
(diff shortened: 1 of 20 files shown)


src/lib/types/user.ts
```diff
@@ -18,9 +18,9 @@ export interface UserData {
 
 export interface IUser {
     id: number;
-    name?: string;
-    username?: string;
-    email?: string;
+    name: string;
+    username: string;
+    email: string;
```
gastonfournier, line 23: Just adapting to the User class definition

src/lib/routes/client-api/register.ts
```diff
@@ -61,7 +61,10 @@ export default class RegisterController extends Controller {
         });
     }
 
-    private static resolveEnvironment(user: User, data: Partial<IClientApp>) {
+    private static resolveEnvironment(
+        user: IUser | IApiUser,
```
gastonfournier, line 65: We're already checking `if (user instanceof ApiUser)` below, so this type is more realistic

src/lib/server-impl.ts
```diff
@@ -207,6 +207,7 @@ export type {
     IUnleashOptions,
     IUnleashConfig,
     IUser,
+    IApiUser,
```
gastonfournier, line 210: Export it!

sighphyre (approved):
LGTM, scary but LGTM

src/lib/routes/client-api/register.ts
thomasheartman: Yes! I feel like we do this loads of places, though 🤔 But you've probably already caught that and fixed it 😄

src/lib/types/user.ts
thomasheartman: Do we really always have all of name, username, and email? I thought (at least) one of these was often unset? Does it default to an empty string?

src/test/fixtures/fake-user-store.ts
```diff
@@ -131,13 +131,13 @@ class UserStoreMock implements IUserStore {
     upsert(user: ICreateUser): Promise<IUser> {
         this.data.splice(this.data.findIndex((u) => u.email === user.email));
         this.data.push({
+            ...(user as IUser),
             id: this.data.length + 1,
             createdAt: new Date(),
             isAPI: false,
             permissions: [],
             loginAttempts: 0,
             imageUrl: '',
```
thomasheartman, line 140: Oh, this is interesting. Doesn't this change the behavior here? Specifically, the `id` would now get overwritten, whereas previously it wouldn't

src/lib/types/user.ts
gastonfournier: It's a fair question and to the best of my knowledge this type is plain wrong... it should be something like: `either have username or email`. I tried doing that but I couldn't. We can try, but ultimately that's what the class definition is (which was already wrong), so I thought this could be a problem for later as we're not changing the interface.

src/test/fixtures/fake-user-store.ts
gastonfournier: image
The thing is that switching the order gives errors because many properties are overridden (id should always be defined in the user object so there's no point in defining it)...

I tried removing the as IUser and I get a different kind of error: 
image

but I found a more elegant way of solving both. Commit incoming

gastonfournier: I believe this fixes your concern: 8b037af924368c13a4c5082b3263e5f386cf977d

Thanks for the challenge :hugs:

thomasheartman: Nice! But wouldn't that still override `id` if it's present on the user object? And in a way that's different from before, where the user's `id` would be overridden by `this.data.length + 1`?

gastonfournier: We're probably not sending the id using ts-ignore or ts-expect error. And if we're already sending it then it should be fine because it was sent before. I believe these values act as a default in case some of the values are not provided in the parameter

gastonfournier: So it's actually same behavior as before

PR #5272, opened 2023-11-06
Chore: remove "dataPath" from data OpenAPI data errors.

thomasheartman:
The `dataPath` was present (but not in the type) in previous versions of the
error library that we use. But with the recent major upgrade, it's
been removed and the `instancePath` property has finally come into use.

This PR removes all the handling for the previous property and
replaces it with `instancePath`. Because the `dataPath` used full
stops and the `instancePath` uses slashes, we need to change a little
bit of the handling too.

### DIFF
--- a/src/lib/error/bad-data-error.ts
+++ b/src/lib/error/bad-data-error.ts
@@ -74,10 +74,10 @@ const genericErrorMessage = (
     propertyName: string,
     errorMessage: string = 'is invalid',
 ) => {
-    const input = getProp(requestBody, propertyName);
+    const input = getProp(requestBody, propertyName.split('/'));
 
     const youSent = JSON.stringify(input);
-    const description = `The \`.${propertyName}\` property ${errorMessage}. You sent ${youSent}.`;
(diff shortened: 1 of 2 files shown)


src/lib/error/bad-data-error.ts
```diff
@@ -74,10 +74,10 @@ const genericErrorMessage = (
     propertyName: string,
     errorMessage: string = 'is invalid',
 ) => {
-    const input = getProp(requestBody, propertyName);
+    const input = getProp(requestBody, propertyName.split('/'));
 
     const youSent = JSON.stringify(input);
-    const description = `The \`.${propertyName}\` property ${errorMessage}. You sent ${youSent}.`;
+    const description = `The \`${propertyName}\` property ${errorMessage}. You sent ${youSent}.`;
```
thomasheartman, line 80: No other property descriptors start with a full stop, and because the new format uses slashes, I'll remove it for now.

sjaanus (approved):
LG

PR #5273, opened 2023-11-06
poc: transaction executor

andreas-unleash:
Thinking about the deadlock pain we have been experiencing, I think this will help.

Creates a transaction executor, able to handle transient errors (like deadlocks) with a configurable number of retries with exponential backoff

### DIFF
--- a/src/lib/db/transaction-executor.ts
+++ b/src/lib/db/transaction-executor.ts
@@ -1,0 +1,69 @@
+import { Knex } from 'knex';
+
+export class TransactionExecutor {
+    private db: Knex;
+    private maxRetries: number;
+
+    constructor(db: Knex, maxRetries: number = 3) {
+        this.db = db;
+        this.maxRetries = maxRetries;
+    }
+
+    /**
+     * Executes a transaction and automatically retries it in case of a transient failure.
+     * @param transactionalWork - The actual work that will be executed in the transaction.
+     * @returns A promise that resolves with the value returned by the transactional work.
+     */
+    public async execute<T>(
+        transactionalWork: (trx: Knex.Transaction) => Promise<T>,
+    ): Promise<T> {
+        let lastError: any = null;
+
+        for (let attempt = 0; attempt < this.maxRetries; attempt++) {
+            try {
+                const result = await this.db.transaction(async (trx) => {
+                    return transactionalWork(trx);
+                });
+
+                return result; // Transaction succeeded
+            } catch (error) {
+                lastError = error;
+                if (!this.isTransientError(error)) {
+                    break; // Non-transient error, do not retry
+                }
+                console.warn(
+                    `Transaction failed with error: ${
+                        error.message
+                    }. Retrying ${attempt + 1}/${this.maxRetries}...`,
+                );
+            }
+        }
+
+        throw lastError; // All retries failed
+    }
+
+    /**
+     * Determines if the error is transient (i.e., likely to succeed if retried).
+     * Logs a human-readable error message and returns true if the error is considered transient; false otherwise.
+     * @param error - The error that occurred.
+     * @returns True if the error is considered transient; false otherwise.
+     */
+    private isTransientError(error: any): boolean {
+        const transientErrors: { [key: string]: string } = {
+            '40001': 'serialization_failure',
+            '40P01': 'deadlock_detected',
+            '55P03': 'lock_not_available',
+        };
+
+        const errorMessage = transientErrors[error.code];
+        if (errorMessage) {
+            console.warn(
+                `Transient Error: ${errorMessage} (${error.code}). Error message: ${error.message}`,
+            );
+            return true;
+        }
+
+        console.error(`Non-transient Error: ${error.message} (${error.code})`);
+        return false;
+    }
+}
--- a/src/lib/db/transaction-executor.ts
+++ b/src/lib/db/transaction-executor.ts
@@ -1,4 +1,5 @@
 import { Knex } from 'knex';
+import delay from '@slack/web-api/dist/helpers';
 
 export class TransactionExecutor {
     private db: Knex;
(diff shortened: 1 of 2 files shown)


src/lib/db/transaction-executor.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Determines if the error is transient (i.e., likely to succeed if retried).
+     * Logs a human-readable error message and returns true if the error is considered transient; false otherwise.
+     * @param error - The error that occurred.
+     * @returns True if the error is considered transient; false otherwise.
+     */
+    private isTransientError(error: any): boolean {
+        const transientErrors: { [key: string]: string } = {
+            '40001': 'serialization_failure',
```
gastonfournier, line 53: Why is this transient?

src/lib/db/transaction-executor.ts
```diff
@@ -0,0 +1,82 @@
...
+            } catch (error) {
+                lastError = error;
+                if (this.isTransientError(error) && attempt < this.maxRetries) {
+                    // Implementing exponential backoff with jitter
+                    const delayMillis =
+                        this.getExponentialBackoffDelay(attempt);
+                    console.error(
+                        `Transaction failed: ${error.message}. Retrying in ${delayMillis} ms.`,
+                    );
+                    await delay(delayMillis);
```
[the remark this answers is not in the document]
andreas-unleash: Makes sense, also adding the exponential backoff to the clients makes instead

andreas-unleash:
It's not my intention to merge, just wanted to start a discussion on better handling some db errors (since we are keeping away from orm).

src/lib/db/transaction-executor.ts
andreas-unleash: Serialization failure is an error condition that arises in scenarios involving concurrent database transactions.

andreas-unleash:
I wonder if our problems would go away if we added ISOLATION LEVEL SERIALIZABLE to (some of) our transactions. (basically forces them to run serially even in concurrent situations)

```
BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
-- Your transaction operations here
COMMIT;
```

There are things to consider however:
- Performance Impact: While providing the highest level of data integrity, the Serializable Isolation level can have a performance impact, especially in a highly concurrent environment, because it can lead to increased transaction rollbacks and retries.

- Handling Rollbacks: Since Serializable transactions are more prone to rollbacks due to serialization failures (error code 40001), applications should be designed to gracefully handle and potentially retry transactions that are safe to retry.

gastonfournier:
> It's not my intention to merge, just wanted to start a discussion on better handling some db errors (since we are keeping away from orm).

:) :+1: great then! We can keep it going, but I'm also wondering how big of a problem is it right now, and whether we identified where this happens

> I wonder if our problems would go away if we added ISOLATION LEVEL SERIALIZABLE to (some of) our transactions (eg the ones that get triggered by a scheduler). Basically forces them to run serially even in concurrent situations.
> 
> ```
> await trx.raw('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE');
> ```
> 
> There are things to consider however:
> 
> * Performance Impact: While providing the highest level of data integrity, the Serializable Isolation level can have a performance impact, especially in a highly concurrent environment, because it can lead to increased transaction rollbacks and retries.
> * Handling Rollbacks: Since Serializable transactions are more prone to rollbacks due to serialization failures (error code 40001), applications should be designed to gracefully handle and potentially retry transactions that are safe to retry.

Maybe it is worth trying something like this... Every time I work with databases I have to do some memory refreshers, cause it's not my area of expertise. But something like this can be tested through load tests to validate there's no performance regression, but it won't be as good as real production testing. I know in some languages like java it could be hard to add a feature flag to it, but in node maybe it's not that difficult. In such case we could test this targeting specifically the customer having issues more frequently...

andreas-unleash:
So, here are my 2 cents about what I think is happening:

> > It's not my intention to merge, just wanted to start a discussion on better handling some db errors (since we are keeping away from orm).
> 
> :) 👍 great then! We can keep it going, but I'm also wondering how big of a problem is it right now, and whether we identified where this happens
> 
> > I wonder if our problems would go away if we added ISOLATION LEVEL SERIALIZABLE to (some of) our transactions (eg the ones that get triggered by a scheduler). Basically forces them to run serially even in concurrent situations.
> > ```
> > await trx.raw('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE');
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There are things to consider however:
> > 
> > * Performance Impact: While providing the highest level of data integrity, the Serializable Isolation level can have a performance impact, especially in a highly concurrent environment, because it can lead to increased transaction rollbacks and retries.
> > * Handling Rollbacks: Since Serializable transactions are more prone to rollbacks due to serialization failures (error code 40001), applications should be designed to gracefully handle and potentially retry transactions that are safe to retry.
> 
> Maybe it is worth trying something like this... Every time I work with databases I have to do some memory refreshers, cause it's not my area of expertise. But something like this can be tested through load tests to validate there's no performance regression, but it won't be as good as real production testing. I know in some languages like java it could be hard to add a feature flag to it, but in node maybe it's not that difficult. In such case we could test this targeting specifically the customer having issues more frequently...

So the only place we get the deadlocks as of now is when trying to commit lastSeen metrics.  My hypothesis is that this is happening when 2 pods are sending the same request at the same time ending up with 2 transactions in deadlock. Sorting the entries helped. (which tells me even 1 sec delay/retry would get them out of deadlock.
This is happening for big-ish users (who are also the most likely to have some sort of clock sync going on in their infrastructure).

Theoretically any scheduled task could end up in this state given enough data to persist without (some) transaction management/error handling.

I will try to simulate the conditions and give it a go.  :)

gastonfournier:
> So the only place we get the deadlocks as of now is when trying to commit lastSeen metrics. My hypothesis is that this is happening when 2 pods are sending the same request at the same time ending up with 2 transactions in deadlock. Sorting the entries helped. (which tells me even 1 sec delay/retry would get them out of deadlock.
> This is happening for big-ish users (who are also the most likely to have some sort of clock sync going on in their infrastructure).
> 
> Theoretically any scheduled task could end up in this state given enough data to persist without (some) transaction management/error handling.

If this is the only case this happens I wouldn't make all operations serializable. I think it's fine to lose some "lastSeen" events, it means the last seen will be eventually consistent. I believe the main issue is errors triggering some alerts. But considering the last seen are eventually consistent, I wonder if we can batch them in memory and then have a single process to persist them on a regular cadence. When returning last-seen metric we can just read from memory and go to the DB on a miss (using memory as a [read-through cache](https://docs.oracle.com/cd/E15357_01/coh.360/e15723/cache_rtwtwbra.htm#COHDG5177)). I can still see 2 of our pods trying to write to the db at the same time, but that can be handled with jitter and maybe here it makes sense to do retries...

> I will try to simulate the conditions and give it a go. :)

Sure, but consider the other option I just shared. If you want we can talk about this problem and do some brainstorming

PR #5274, opened 2023-11-06
feat: introduce offset based search instead of cursor

sjaanus:


chriswk:
Just so you're aware, but offset based search will gradually be slower as you move further into the dataset. It will need to fetch and drop up to your offset before returning the rows you want.

frontend/src/component/project/Project/ProjectOverview.tsx
```diff
@@ -91,10 +88,12 @@ const InfiniteProjectOverview = () => {
                         onChange={refetch}
                         total={total}
                     />
-                    {prevCursors.length > 0 ? (
+                    {currentOffset > 0 ? (
                         <Box onClick={fetchPrevPage}>Prev</Box>
                     ) : null}
-                    {nextCursor && <Box onClick={fetchNextPage}>Next</Box>}
+                    {currentOffset + PAGE_LIMIT < total && (
```
FredrikOseberg, line 94: Could have more readable name for the variable. Should also use ConditionallyRender until we change the ADR. For now I'd use ternaries when you need to access variables that need typecheck and CR otherwise.

FredrikOseberg (approved):
LGTM

PR #5279, opened 2023-11-06
Refactor/project overview api calls

FredrikOseberg:
This PR reduces the overhead of making API calls on pages with heavy renders. We forego loading states and default error handling in favor of more speed by avoiding triggering multiple re-renders from the API call. 

When the page is heavy on elements, such as when you have lot of environments active on the project overview screen, this will reduce the overhead significantly. In my testing it was reduced from between 900-1900ms to 50-80ms on localhost.

Also added a timer that will time API calls in development and notify you via the console when you have a slow API call.

### DIFF
--- a/frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
+++ b/frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
@@ -38,6 +38,7 @@ import { ExportDialog } from './ExportDialog';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 import { focusable } from 'themes/themeStyles';
 import { FeatureEnvironmentSeenCell } from 'component/common/Table/cells/FeatureSeenCell/FeatureEnvironmentSeenCell';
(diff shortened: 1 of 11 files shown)


sjaanus (approved):
Looks fine, but do not fully grasp why doing some `setLoading` and `handleResponses` calls are making things slow. Are you saying one trigger the full UI reload and without them, it does not?

FredrikOseberg:
> Looks fine, but do not fully grasp why doing some `setLoading` and `handleResponses` calls are making things slow. Are you saying one trigger the full UI reload and without them, it does not?

Loading is propagated all the way from the hook down into the component that consumes the hook. In that way you can use the loading state to make decisions on what to do in the component. Our project overview table is memoized but when loading changes, the memoization is busted which means that any change to the loading state triggers a whole re-render of the table. When the table is heavy with a lot of feature toggles and environments, this re-render is blocking for the amount of time that it takes the browser to re-render all the heavy items. Even with memoization on table specific components, this delay is significant. 

Skjermbilde 2023-11-06 kl  15 45 38

PR #5282, opened 2023-11-06
docs: add slack app redirect

nunogois:
Adds a redirect rule for the Slack App integration (fix current broken link).


sighphyre (approved):
Nice catch!

PR #5287, opened 2023-11-06
chore: expose custom-handler-auth type

gastonfournier:
This will help us get type checking on the auth handler function

### DIFF
--- a/src/lib/types/option.ts
+++ b/src/lib/types/option.ts
@@ -55,7 +55,7 @@ export enum IAuthType {
     NONE = 'none',
 }
 
-export type CustomHandlerType = (
+export type CustomAuthHandler = (
     app: Express,
     config: IUnleashConfig,
     services: IUnleashServices,
@@ -64,7 +64,7 @@ export type CustomHandlerType = (
 export interface IAuthOption {
     enableApiToken: boolean;
     type: IAuthType;
-    customAuthHandler?: CustomHandlerType;
+    customAuthHandler?: CustomAuthHandler;
     createAdminUser?: boolean;
     initialAdminUser?: {
         username: string;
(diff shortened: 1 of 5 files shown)


src/lib/types/option.ts
```diff
@@ -53,10 +55,16 @@ export enum IAuthType {
...
+    app: Express,
+    config: IUnleashConfig,
+    services: IUnleashServices,
+) => void;
+
 export interface IAuthOption {
     enableApiToken: boolean;
     type: IAuthType;
-    customAuthHandler?: Function;
+    customAuthHandler?: CustomHandlerType;
```
gastonfournier, line 67: Finally, this loosely typed function is what caused us to miss the removal of the config parameter in this incident: https://unleash.instatus.com/clofjecqc147784bpogyb47nkpz

The type of the function is now tied to the parameter types we send in when we use it: https://github.com/Unleash/unleash/blob/312999066b0f712eecfeebd10f418600a0862c04/src/lib/app.ts#L121

src/lib/types/option.ts
```diff
@@ -53,10 +55,16 @@ export enum IAuthType {
     NONE = 'none',
 }
 
+export type CustomAuthHandler = (
+    app: Express,
+    config: IUnleashConfig,
+    services: IUnleashServices,
```
gastonfournier, line 61:
```suggestion
    config: Partial<IUnleashConfig>,
    services?: IUnleashServices,
```
Maybe making services optional in case you're an open source you can just ignore receiving it. 

While the config we're making it partial so you have to explicitly check for undefined when accessing the config and we can also send `{}` empty object without getting a compilation error

chriswk (approved):
I like this. Might've stopped last weeks small issues with the google sign in.

daveleek (approved):
Awesome!

PR #5296, opened 2023-11-08
Feat/scheduled cr UI tests

andreas-unleash:
Ui tests scheduled change requests

Closes # [1-1598](https://linear.app/unleash/issue/1-1598/e2e-ui-tests)


frontend/src/hooks/useChangeRequestsEnabled.ts
```diff
@@ -38,6 +38,7 @@ export const useChangeRequestsEnabled = (projectId: string) => {
     );
 
     const isChangeRequestConfiguredInAnyEnv = React.useCallback((): boolean => {
+        console.log('isChangeRequestConfiguredInAnyEnv: ');
```
thomasheartman, line 41: This looks like an oversight to me?

```suggestion
```

thomasheartman (approved):
Looks good to me (plus nice catch with the multi-action button key) 😄  I know you talked to Fredrik about testing this, so I'm assuming you've followed his guidance here? If so, I think it's good to go. If not, you should probably have someone with more experience testing the UI at Unleash have a look to make sure things are in order.

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+        feedback: [],
+        splash: {},
+    });
+};
+const setupOtherServerRoutes = () => {
+    uiConfig();
+    changeRequestConfig();
+    user();
+    updateChangeRequestState();
+};
```
FredrikOseberg, line 150: Can we move these into setupHttpRoutes? Since that method is calling this method anyway, I don't think this abstraction is necessary.

FredrikOseberg (approved):
LGTM, one small suggestion (see comment).

PR #5298, opened 2023-11-08
chore: detect if a segment is being used in an active CR

thomasheartman:
This PR adds a way to tell if a specific segment is being used in any active change requests. It's the first step towards preventing segments that are being used in change requests from being deleted.

It does that by checking the db for any unclosed CRs and using those CR ids to look for "addStrategy" and "updateStrategy" events in the cr events table. 

## Discussion points

I've implemented this as part of the change request access read model because it was the closest we had, but it feels kinda dirty. I think I'd like to refactor it out into its own module ("change-request-segment-usage-read-model" or something?), but I suggest doing that in a separate PR (to keep the changes in this PR manageable). I'd much appreciate some thoughts on that.

## Upcoming PRs

This only puts in a way to detect it, but doesn't add that to anything. That'll be in an upcoming iteration.


thomasheartman:
Closing in favor of #5301.

PR #5299, opened 2023-11-08
feat: also allow searching partial tags

sjaanus:
Previously using tags:simple returned nothing. You needed to put tag exactly like tags:simple:mytag.
Now it works.

### DIFF
--- a/src/lib/features/feature-search/feature-search-controller.ts
+++ b/src/lib/features/feature-search/feature-search-controller.ts
@@ -83,9 +83,7 @@ export default class FeatureSearchController extends Controller {
                 sortBy,
             } = req.query;
             const userId = req.user.id;
-            const normalizedTag = tag
-                ?.map((tag) => tag.split(':'))
-                .filter((tag) => tag.length === 2);
+            const normalizedTag = tag?.map((tag) => tag.split(':'));
             const normalizedStatus = status
                 ?.map((tag) => tag.split(':'))
                 .filter(
@@ -95,7 +93,7 @@ export default class FeatureSearchController extends Controller {
                 );
             const normalizedLimit =
                 Number(limit) > 0 && Number(limit) <= 50 ? Number(limit) : 50;
-            const normalizedOffset = Number(offset) > 0 ? Number(limit) : 0;
+            const normalizedOffset = Number(offset) > 0 ? Number(offset) : 0;
             const normalizedSortBy: string = sortBy ? sortBy : 'createdAt';
             const normalizedSortOrder =
                 sortOrder === 'asc' || sortOrder === 'desc' ? sortOrder : 'asc';
--- a/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
@@ -547,6 +547,9 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
         features: IFeatureOverview[];
         total: number;
     }> {
+        const normalizedFullTag = tag?.filter((tag) => tag.length === 2);
+        const normalizedHalfTag = tag?.filter((tag) => tag.length === 1).flat();
+
         let environmentCount = 1;
         if (projectId) {
             const rows = await this.db('project_environments')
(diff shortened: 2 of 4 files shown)


src/lib/features/feature-search/feature-search-controller.ts
```diff
@@ -83,9 +83,7 @@ export default class FeatureSearchController extends Controller {
                 sortBy,
             } = req.query;
             const userId = req.user.id;
-            const normalizedTag = tag
-                ?.map((tag) => tag.split(':'))
-                .filter((tag) => tag.length === 2);
+            const normalizedTag = tag?.map((tag) => tag.split(':'));
```
sjaanus, line 86: Now 1 length tags will also be used. 

**This is the question, should we split these already in controller, or do it in store.
If we do it here, we need to pass those in as 2 parameters. Full tags and half tags.
If we do it in store, we are leaking some normalizing logic in store.**

src/lib/features/feature-search/feature-search-controller.ts
```diff
@@ -95,7 +93,7 @@ export default class FeatureSearchController extends Controller {
                 );
             const normalizedLimit =
                 Number(limit) > 0 && Number(limit) <= 50 ? Number(limit) : 50;
-            const normalizedOffset = Number(offset) > 0 ? Number(limit) : 0;
+            const normalizedOffset = Number(offset) > 0 ? Number(offset) : 0;
```
sjaanus, line 96: Fixed a typo.

thomasheartman: This feels like more than a type 👀

src/lib/features/feature-search/feature-search-controller.ts
thomasheartman: I think I'm a little low on context: what's a 1-length tag? Like "simple:x"? what's a full tag and a half tag?

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -559,29 +561,39 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
-                .select('feature_name')
-                .whereRaw("(?? || ':' || ??) LIKE ?", [
+            if (hasQueryString) {
+                tagQuery.whereRaw("(?? || ':' || ??) ILIKE ?", [
                     'tag_type',
                     'tag_value',
                     `%${queryString}%`,
                 ]);
+            }
+            if (hasHalfTag) {
```
FredrikOseberg, line 576: Conceptually, what is a "half tag"?

src/lib/features/feature-search/feature-search-controller.ts
FredrikOseberg: Maybe it should happen in the service.

thomasheartman (commented):
Looks cool, but I need some more context on full and half tags before I can give a thumbs up, I think 😅

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -559,29 +561,39 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
+                const tagParameter = normalizedHalfTag.map((tag) => `%${tag}%`);
+                tagQuery.orWhereRaw(
+                    `(?? || ':' || ??) ILIKE ANY (ARRAY[${tagParameter
+                        .map(() => '?')
+                        .join(',')}])`,
+                    ['tag_type', 'tag_value', ...tagParameter],
+                );
```
FredrikOseberg, line 583: This is difficult to understand from a glance

src/lib/features/feature-search/feature-search-controller.ts
sjaanus: Then we still need 2 parameters for tags in store method. Which may be fine, but seems a little much.

sjaanus: @thomasheartman our tags consist of **type** and **value**. So a full tag would be the complete tag consisting of type and value. So `simple:mytag`, but previously we had possibility to search just by value or by type, without providing full tag definition. Just by writing simple, we got all `simple` tags.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
sjaanus: @FredrikOseberg  our tags consist of **type** and **value**. So a full tag would be the complete tag consisting of type and value. So `simple:mytag`, but previously we had possibility to search just by value or by type, without providing full tag definition. Just by writing simple, we got all `simple` tags.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
sjaanus: I will be simplifying this in following PR, because of performance reasons.

PR #5301, opened 2023-11-08
refactor: extract segment usage read model

thomasheartman:
This PR adds a way to tell if a specific segment is being used in any active change requests. It's the first step towards preventing segments that are being used in change requests from being deleted.

It does that by checking the db for any unclosed CRs and using those CR ids to look for "addStrategy" and "updateStrategy" events in the cr events table. 

## Upcoming PRs

This only puts in a way to detect it, but doesn't add that to anything. That'll be in an upcoming iteration.


src/lib/features/change-request-segment-usage-read-model.ts/sql-change-request-segment-usage-read-model.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public async isSegmentUsedInActiveChangeRequests(
+        segmentId: number,
+    ): Promise<boolean> {
+        const result = await this.db.raw(
+            `SELECT *
+            FROM change_request_events
+            WHERE change_request_id IN (
+            SELECT id
```
sjaanus, line 20: Generally good to avoid nested queries due to bad performance. Use joins instead.

thomasheartman: Done in 430bb2b77e893081eff3b6f02b48d16159866619

andreas-unleash (approved):
LGTM

PR #5304, opened 2023-11-08
fix: take into account project segments permission

nunogois:
https://linear.app/unleash/issue/SR-164/ticket-1106-user-with-createedit-project-segment-is-not-able-to-edit-a

Fixes a bug where the `UPDATE_PROJECT_SEGMENT` permission is not respected, both on the UI and on the API. The original intention was stated [here](https://github.com/Unleash/unleash/pull/3346#discussion_r1140434517).

This was easy to fix on the UI, since we were simply missing the extra permission on the button permission checks.

Unfortunately the API can be tricky. Our auth middleware tries to grab the `project` information from either the params or body object, but our `DELETE` method does not contain this information. There is no body and the endpoint looks like `/admin/segments/:id`, only including the segment id. 

This means that, in the rbac middleware when we check the permissions, we need to figure out if we're in such a scenario and fetch the project information from the DB, which feels a bit hacky, but it's something we're seemingly already doing for features, so at least it's somewhat consistent.

Ideally what we could do is leave this API alone and create a separate one for project segments, with endpoints where we would have project as a param, like so: `http://localhost:4242/api/admin/projects/:projectId/segments/1`.

This PR opts to go with the quick and hacky solution for now since this is an issue we want to fix quickly, but this is something that we should be aware of. I'm also unsure if we want to create a new API for project segments. If we decide that we want a different solution I don't mind either adapting this PR or creating a follow up.

### DIFF
--- a/frontend/src/component/segments/CreateSegment/CreateSegment.tsx
+++ b/frontend/src/component/segments/CreateSegment/CreateSegment.tsx
@@ -1,7 +1,10 @@
 import React, { useContext } from 'react';
 import { CreateButton } from 'component/common/CreateButton/CreateButton';
 import FormTemplate from 'component/common/FormTemplate/FormTemplate';
-import { CREATE_SEGMENT } from 'component/providers/AccessProvider/permissions';
+import {
+    CREATE_SEGMENT,
+    UPDATE_PROJECT_SEGMENT,
+} from 'component/providers/AccessProvider/permissions';
 import { useSegmentsApi } from 'hooks/api/actions/useSegmentsApi/useSegmentsApi';
 import { useConstraintsValidation } from 'hooks/api/getters/useConstraintsValidation/useConstraintsValidation';
 import { useSegments } from 'hooks/api/getters/useSegments/useSegments';
@@ -112,7 +115,8 @@ export const CreateSegment = ({ modal }: ICreateSegmentProps) => {
             >
                 <CreateButton
                     name='segment'
-                    permission={CREATE_SEGMENT}
+                    permission={[CREATE_SEGMENT, UPDATE_PROJECT_SEGMENT]}
+                    projectId={projectId}
                     disabled={!hasValidConstraints || overSegmentValuesLimit}
                     data-testid={SEGMENT_CREATE_BTN_ID}
                 />
--- a/frontend/src/component/segments/EditSegment/EditSegment.tsx
+++ b/frontend/src/component/segments/EditSegment/EditSegment.tsx
@@ -1,5 +1,8 @@
 import FormTemplate from 'component/common/FormTemplate/FormTemplate';
-import { UPDATE_SEGMENT } from 'component/providers/AccessProvider/permissions';
+import {
+    UPDATE_PROJECT_SEGMENT,
+    UPDATE_SEGMENT,
+} from 'component/providers/AccessProvider/permissions';
 import { useSegmentsApi } from 'hooks/api/actions/useSegmentsApi/useSegmentsApi';
 import { useConstraintsValidation } from 'hooks/api/getters/useConstraintsValidation/useConstraintsValidation';
 import { useSegment } from 'hooks/api/getters/useSegment/useSegment';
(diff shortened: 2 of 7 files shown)


frontend/src/component/segments/CreateSegment/CreateSegment.tsx
```diff
@@ -112,7 +115,8 @@ export const CreateSegment = ({ modal }: ICreateSegmentProps) => {
             >
                 <CreateButton
                     name='segment'
-                    permission={CREATE_SEGMENT}
+                    permission={[CREATE_SEGMENT, UPDATE_PROJECT_SEGMENT]}
+                    projectId={projectId}
```
nunogois, line 119: These 2 lines fix the bug on the UI, where the button was shown as being disabled.

frontend/src/component/segments/EditSegment/EditSegment.tsx
```diff
@@ -135,7 +138,8 @@ export const EditSegment = ({ modal }: IEditSegmentProps) => {
                 mode='edit'
             >
                 <UpdateButton
-                    permission={UPDATE_SEGMENT}
+                    permission={[UPDATE_SEGMENT, UPDATE_PROJECT_SEGMENT]}
+                    projectId={projectId}
```
nunogois, line 142: These 2 lines fix the bug on the UI, where the button was shown as being disabled.

frontend/src/hooks/api/actions/useApi/useApi.ts
```diff
@@ -194,14 +194,14 @@ const useAPI = ({
...
-                requestId || `Uknown request happening on ${apiCaller}`,
+                requestId || `Unknown request happening on ${apiCaller}`,
             );
 
             const res = await requestFunction(apiCaller, requestId, loadingOn);
 
             timeApiCallEnd(
                 start,
-                requestId || `Uknown request happening on ${apiCaller}`,
+                requestId || `Unknown request happening on ${apiCaller}`,
```
nunogois, line 204: Boyscouting: Noticed the typo and fixed it.

src/lib/features/segment/segment-controller.ts
```diff
@@ -225,7 +226,7 @@ export class SegmentsController extends Controller {
             method: 'post',
             path: '',
             handler: this.createSegment,
-            permission: CREATE_SEGMENT,
+            permission: [CREATE_SEGMENT, UPDATE_PROJECT_SEGMENT],
```
nunogois, line 229: These changes add the missing permission on the API level.

src/lib/error/permission-error.ts
```diff
@@ -15,7 +15,7 @@ class PermissionError extends UnleashError {
         const permissionsMessage =
             permissions.length === 1
                 ? `the "${permissions[0]}" permission`
-                : `all of the following permissions: ${permissions
+                : `one of the following permissions: ${permissions
```
nunogois, line 18: Boyscouting: I think "one" instead of "all" is more honest and reflective of the real behavior.

src/lib/middleware/rbac-middleware.test.ts
```diff
@@ -340,7 +392,11 @@ test('should not expect featureName for UPDATE_FEATURE when projectId specified'
         hasPermission: jest.fn(),
     };
 
-    const func = rbacMiddleware(config, { featureToggleStore }, accessService);
+    const func = rbacMiddleware(
+        config,
+        { featureToggleStore, segmentStore },
+        accessService,
+    );
```
nunogois, line 399: Requires injecting a `segmentStore` in the rbacMiddleware.

src/lib/middleware/rbac-middleware.ts
```diff
@@ -87,6 +92,17 @@ const rbacMiddleware = (
...
+            // This is needed to check if the user has the right permissions on a project level
+            if (
+                !projectId &&
+                permissionsArray.includes(UPDATE_PROJECT_SEGMENT)
+            ) {
+                const { id } = params;
+                const { project } = await segmentStore.get(id);
+                projectId = project;
+            }
+
```
nunogois, line 105: This is what I mean by:

>This means that, in the rbac middleware when we check the permissions, we need to figure out if we're in such a scenario and fetch the project information from the DB, which feels a bit hacky, but it's something we're seemingly already doing for features, so at least it's somewhat consistent.

sighphyre (commented):
This is the cleanest hack PR I've ever seen in my life xD

But I do agree, I think this is a hack. I'm okay to approve this because I do think it solves the problem but the cleaner design would be to split this into separate APIs.

Shall we get more eyes than just the 3 of us on this one?

chriswk (approved):
This feels like it looks good, the frontend changes at least was a nice catch. The logic in the segment controller as well seems sane to me.

PR #5305, opened 2023-11-08
fix: project overview and search returning just single tag

sjaanus:
1. Created failing test
2. Fixed uniqueness identifier
3. Test green


src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -108,7 +108,7 @@ function mapInput(input: IFeatureStrategy): IFeatureStrategiesTable {
 const getUniqueRows = (rows: any[]) => {
```
sjaanus, line 108: To be honest, I am not sure what this method is giving. In theory, all rows should be unique enough by default.

FredrikOseberg (approved):
LGTM

PR #5306, opened 2023-11-09
Make compression middleware optional

pedropapa:
## Why
Currently AWS API Gateway doesn't have compression enabled by default, this PR will make it easier to for example deploy Unleash over to AWS Lambda without further configuration in API Gateway, frameworks like Serverless requires a bit more work to set up compression and some times one might not need compression at all.

## How
Create a new config flag called `disableCompression` which will not include `compression` middleware in express' instance when set as true.


ivarconr (approved):

pedropapa:
@ivarconr thanks for the heads up, introduced the env variable and got the docs updated.

ivarconr (approved):

ivarconr (approved):
Nice job 🔥

PR #5308, opened 2023-11-09
feat: API prevents you from deleting segments in crs

thomasheartman:
This PR hooks up the changes introduced in #5301 to the API and puts them behind a feature flag. A new test has been added and the test setup has been slightly tweaked to allow this test.

When the flag is enabled, the API will now not let you delete a segment that's used in any active CRs.


src/lib/features/change-request-segment-usage-service/change-request-segment-usage-read-model.test.ts
thomasheartman, line 1: I'd named the directory holding these files very weirdly. And some of the naming of the models and methods was also off, so these changes fix that.

src/lib/features/segment/segment-controller.ts
```diff
@@ -372,9 +372,16 @@ export class SegmentsController extends Controller {
...
-        const strategies = await this.segmentService.getAllStrategies(id);
 
-        if (strategies.length > 0) {
+        let segmentIsInUse = false;
+        if (this.flagResolver.isEnabled('detectSegmentUsageInChangeRequests')) {
+            segmentIsInUse = await this.segmentService.isInUse(id);
+        } else {
+            const strategies = await this.segmentService.getAllStrategies(id);
+            segmentIsInUse = strategies.length > 0;
+        }
```
thomasheartman, line 382: Really not a big fan of reassigning variables like this, but because `if` is a statement and not an expression in JS, the alternative would be a pretty nasty ternary (see suggestion below). I think this version is cleaner for this use case. Plus, we'll remove it once the flag goes away.

```suggestion
        let segmentIsInUse = this.flagResolver.isEnabled(
            'detectSegmentUsageInChangeRequests',
        )
            ? await this.segmentService.isInUse(id)
            : (await this.segmentService.getAllStrategies(id)).length > 0;
```

src/test/e2e/api/admin/segment.e2e.test.ts
```diff
@@ -108,14 +108,18 @@ const validateSegment = (
...
+        {
+            experimental: {
+                flags: {
+                    anonymiseEventLog: true,
+                    detectSegmentUsageInChangeRequests: true,
+                },
             },
         },
-    });
+        db.rawDatabase,
```
thomasheartman, line 121: We didn't pass the database to the setup before. This meant that we never hit the real services underneath, but used the fake ones instead. For the segment usage read model, that was a problem, so I hooked it up.

andreas-unleash (approved):
LGTM!

PR #5309, opened 2023-11-09
Feat/pagination bar

FredrikOseberg:
Initial implementation of the sticky pagination bar. 

Next steps:
* Consider preloading data for the next page
* Handle what happens when data is loading
* Handle the issue that total is being sent per request and that useSWR refreshes the cache
* Add tests for the pagination


sjaanus (approved):
Nice, some corners could be improved, but can be done later.

PR #5311, opened 2023-11-09
fix: segment project fetch when global

nunogois:
This fixes an edge case not caught originally in https://github.com/Unleash/unleash/pull/5304 - When creating a new segment on the global level:
 - There is no `projectId`, either in the params or body
 - The `UPDATE_PROJECT_SEGMENT` is still a part of the permissions checked on the endpoint
 - There is no `id` on the params

This made it so that we would run `segmentStore.get(id)` with an undefined `id`, causing issues.

The fix was simply checking for the presence of `params.id` before proceeding.


sjaanus (approved):
LG

PR #5316, opened 2023-11-09
docs: remove password note

nunogois:
I think this note may no longer make sense.


gastonfournier (approved):
Indeed! It was [updated with a script](https://github.com/Unleash/unleash/commit/0e80484068e09d70876db17c923cf3b712ab2a60). It made sense when it was passord, not anymore!

PR #5317, opened 2023-11-09
fix: project settings flag limit not properly set

### LINKED ISSUE #5315: Feature flag limit is not correctly updated after reverting it
### Describe the bug

The "feature flag limit" setting in project settings is not correctly updated once you set it and then try to revert it. After clearing the field and saving, alerts about feature flag limits are still visible and the "create" button is disabled.

1. Access the "settings" tab inside a project
2. Set a feature limit (e.g. `1`)
3. Create feature toggles on that project, up to the configured limit
4. Go back to the project settings, clear the field and save
5. Attempt to create a new feature toggle

### Expected behavior

Once the "feature flag limit" setting is cleared, we should be allowed to freely create new feature toggles on that project.

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

5.6.5

### Subscription type

Enterprise

### Hosting type

None

_No response_

nunogois:
https://linear.app/unleash/issue/SR-169/ticket-1107-project-feature-flag-limit-is-not-correctly-updated

Fixes #5315, an issue where it would not be possible to set an empty flag limit. 
This also fixes the UI behavior: Before, when the flag limit field was emptied, it would disappear from the UI.

I'm a bit unsure of the original intent of the `(data.defaultStickiness !== undefined || data.featureLimit !== undefined)` condition. We're in an update method, triggered by a PUT endpoint - I think it's safe to assume that we'll always want to set these values to whatever they come as, we just need to convert them to `null` in case they are not present (i.e. `undefined`).

### DIFF
--- a/src/lib/db/project-store.ts
+++ b/src/lib/db/project-store.ts
@@ -244,8 +244,9 @@ class ProjectStore implements IProjectStore {
         const settingsRow = await this.db(SETTINGS_TABLE)
             .insert({
                 project: project.id,
-                project_mode: project.mode,
                 default_stickiness: project.defaultStickiness,
+                feature_limit: project.featureLimit,
+                project_mode: project.mode,
             })
             .returning('*');
         return this.mapRow({ ...row[0], ...settingsRow[0] });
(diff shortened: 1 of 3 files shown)


andreas-unleash (approved):
LGTM!

src/lib/db/project-store.ts
```diff
@@ -265,25 +266,21 @@ class ProjectStore implements IProjectStore {
...
-                        default_stickiness: data.defaultStickiness,
-                        feature_limit: data.featureLimit,
-                        project_mode: 'open',
+
+            if (await this.hasProjectSettings(data.id)) {
+                await this.db(SETTINGS_TABLE)
+                    .where({ project: data.id })
+                    .update({
+                        default_stickiness: data.defaultStickiness || null,
+                        feature_limit: data.featureLimit || null,
```
gastonfournier, line 275: We might risk overriding a value here... if one of these is not in the payload we should leave it undefined and not touch it

nunogois: Addressed in https://github.com/Unleash/unleash/pull/5317/commits/6803bce9925689988579f39ab8c635a25d835ac7 - Let me know what you think of it.

src/lib/db/project-store.ts
```diff
@@ -265,25 +266,21 @@ class ProjectStore implements IProjectStore {
...
-                        .where({ project: data.id })
-                        .update({
-                            default_stickiness: data.defaultStickiness,
-                            feature_limit: data.featureLimit,
-                        });
-                } else {
-                    await this.db(SETTINGS_TABLE).insert({
-                        project: data.id,
+
+            if (await this.hasProjectSettings(data.id)) {
```
gastonfournier, line 270: I think now we need to 
```
data.defaultStickiness !== undefined ||
                data.featureLimit !== undefined
```

gastonfournier (changes requested):

src/lib/db/project-store.ts
nunogois: https://github.com/Unleash/unleash/pull/5317/commits/0475ac3e4ad6b0d67df47c74c9dc2216b9442658

PR #5318, opened 2023-11-09
Fix/scheduled request UI

andreas-unleash:
Change timezone format
Fixes a bug where the Edit button on hover being elongated

Before:
Screenshot 2023-11-09 at 21 36 01

After:
Screenshot 2023-11-09 at 22 09 26

Adds the missed onClick to the edit button to show the dialog

Fixes a bug with ScheduleChangesDialog onClose

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestScheduledDialogs/ScheduleChangeRequestDialog.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestScheduledDialogs/ScheduleChangeRequestDialog.tsx
@@ -4,7 +4,7 @@ import { Dialogue } from 'component/common/Dialogue/Dialogue';
 import { APPLY_CHANGE_REQUEST } from 'component/providers/AccessProvider/permissions';
 import PermissionButton from 'component/common/PermissionButton/PermissionButton';
 import { DateTimePicker } from 'component/common/DateTimePicker/DateTimePicker';
-import { getBrowserTimezone } from "../ChangeRequestReviewStatus/utils";
+import { getBrowserTimezone } from '../ChangeRequestReviewStatus/utils';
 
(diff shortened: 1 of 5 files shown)


src/lib/types/events.ts
```diff
@@ -248,6 +250,7 @@ export const IEventTypes = [
     CHANGE_REQUEST_CANCELLED,
     CHANGE_REQUEST_SENT_TO_REVIEW,
     CHANGE_REQUEST_APPLIED,
+    SCHEDULED_CHANGE_REQUEST_APPLIED,
     CHANGE_REQUEST_SCHEDULED,
```
andreas-unleash, line 254: This helps the enterprise notification service create a more meaningful message.  
Considered doing it by keeping the same event type CHANGE_REQUEST_APPLIED and adding `scheduled` param in ChangeRequestApplied event in enterprise.
Decided that having a separate event made more sense
(eg one difference is that CHANGE_REQUEST_APPLIED event does not create a notification for the 'applier'   whereas SCHEDULED_CHANGE_REQUEST_APPLIED will)

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestScheduledDialogs/ScheduleChangeRequestDialog.tsx
```diff
@@ -42,15 +42,15 @@ export const ScheduleChangeRequestDialog: FC<ScheduleChangeRequestDialogProps> =
...
+        const timezone = getBrowserTimezone();
 
         return (
             <Dialogue
                 title={title}
                 primaryButtonText={primaryButtonText}
                 secondaryButtonText='Cancel'
                 open={open}
-                onClose={onClose}
+                onClose={() => onClose()}
```
thomasheartman, line 53: Is this necessary? Shouldn't be as far as I can tell?

PR #5319, opened 2023-11-10
docs: POST and PUT body handling

gastonfournier:
## About the changes
While working on Terraform I identified some issues with how our backend handled some requests. This happened because the Terraform client was unaware of new fields and was not sending them. This resulted in bad behavior as some of those missing fields were treated in the backend as `null` and removed existing data from the DB.

This ADR aims to shed some light on the problem and specifies how our server should handle these requests.

### DIFF
--- a/website/docs/contributing/ADRs/ADRs.md
+++ b/website/docs/contributing/ADRs/ADRs.md
@@ -23,7 +23,7 @@ We are in the process of defining ADRs for the back end. At the time of writing 
 * [Naming](./back-end/naming.md)
 * [Preferred export](./back-end/preferred-export.md)
 * [Breaking DB changes](./back-end/breaking-db-changes.md)
-* [Preferred export](./back-end/POST-PUT-api-payload.md)
+* [POST/PUT API payload](./back-end/POST-PUT-api-payload.md)
 
 ## Front-end ADRs
 
--- a/website/docs/contributing/ADRs/ADRs.md
+++ b/website/docs/contributing/ADRs/ADRs.md
@@ -22,6 +22,8 @@ We are in the process of defining ADRs for the back end. At the time of writing 
 
 * [Naming](./back-end/naming.md)
 * [Preferred export](./back-end/preferred-export.md)
+* [Preferred export](./back-end/breaking-db-changes.md)
+* [Preferred export](./back-end/POST-PUT-api-payload.md)
 
 ## Front-end ADRs
 
--- a/website/docs/contributing/ADRs/ADRs.md
+++ b/website/docs/contributing/ADRs/ADRs.md
@@ -22,7 +22,7 @@ We are in the process of defining ADRs for the back end. At the time of writing 
 
 * [Naming](./back-end/naming.md)
 * [Preferred export](./back-end/preferred-export.md)
-* [Preferred export](./back-end/breaking-db-changes.md)
+* [Breaking DB changes](./back-end/breaking-db-changes.md)
 * [Preferred export](./back-end/POST-PUT-api-payload.md)
 
 ## Front-end ADRs
--- a/website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
+++ b/website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
@@ -6,7 +6,7 @@ title: "ADR: POST/PUT API payload"
 
 Whenever we receive a payload in our backend for POST or PUT request we need to take into account backward compatibility. When we add a new field to an existing API payload, clients using the previous version of the payload will not know about that new field. This means that we need to make sure that the new field is optional. If we make the field required, clients using the previous version of the payload will result in overriding the value of the new field with an empty value or null.
 
-### Example: adding new settings to project settings
+### Example: adding new setting field to project settings
 
 Project settings on unleash 5.3:
 ```shell
@@ -37,7 +37,9 @@ curl --location --request PUT 'http://localhost:4242/api/admin/projects/default'
 }'
 ```
 
-Pay attention to the new field feature limit. If a customer updates their server to 5.6 but their integration still does not send that field. If we treat that field as null then our API will override the value of the new field with an empty value.
+Pay attention to the new field feature limit. If a customer updates their server to 5.6 but their integration still does not send that field, it may result in the unwanted behavior of setting that field to empty in the database (in case the server assumes that not sending the field is tring to set it to empty).
+
+This bug can easily be an oversight but can be prevented by following some rules when designing the API payload.
 
 ## Decision
 
--- a/website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
+++ b/website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
@@ -1,0 +1,49 @@
+---
+title: "ADR: POST/PUT API payload"
+---
+
+## Background
+
+Whenever we receive a payload in our backend for POST or PUT request we need to take into account backward compatibility. When we add a new field to an existing API payload, clients using the previous version of the payload will not know about that new field. This means that we need to make sure that the new field is optional. If we make the field required, clients using the previous version of the payload will result in overriding the value of the new field with an empty value or null.
+
+### Example: adding new settings to project settings
+
+Project settings on unleash 5.3:
+```shell
+curl --location --request PUT 'http://localhost:4242/api/admin/projects/default' \
+--header 'Authorization: INSERT_API_KEY' \
+--header 'Content-Type: application/json' \
+--data-raw '{
+  "id": "default",
+  "name": "Default",
+  "description": "Default project",
+  "defaultStickiness": "default",
+  "mode": "open"
+}'
+```
+
+New version of project settings (unleash 5.6):
+
+```shell
+curl --location --request PUT 'http://localhost:4242/api/admin/projects/default' \
+--header 'Authorization: INSERT_API_KEY' \
+--header 'Content-Type: application/json' \
+--data-raw '{
+  "id": "default",
+  "name": "Default",
+  "description": "Default project",
+  "defaultStickiness": "default",
+  "featureLimit": 2
+}'
+```
+
+Pay attention to the new field feature limit. If a customer updates their server to 5.6 but their integration still does not send that field. If we treat that field as null then our API will override the value of the new field with an empty value.
+
+## Decision
+
+When receiving a body from a request we need to take into account 3 possible cases:
+1. The field has a value
+2. The field is `undefined` or not part of the payload
+3. The field is `null`
+
+In case the field has a value, we need to update or set that value in the DB. In case the field is `undefined` or not part of the payload, we need to leave the value in the DB as it is. In case, the field is `null`, we need to remove the value from the DB.
--- a/website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
+++ b/website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
@@ -4,11 +4,11 @@ title: "ADR: POST/PUT API payload"
 
 ## Background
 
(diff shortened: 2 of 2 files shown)


website/docs/contributing/ADRs/ADRs.md
```diff
@@ -22,6 +22,8 @@ We are in the process of defining ADRs for the back end. At the time of writing
 
 * [Naming](./back-end/naming.md)
 * [Preferred export](./back-end/preferred-export.md)
+* [Preferred export](./back-end/breaking-db-changes.md)
```
nunogois, line 25:
```suggestion
* [Breaking DB changes](./back-end/breaking-db-changes.md)
```

website/docs/contributing/ADRs/ADRs.md
```diff
@@ -22,6 +22,8 @@ We are in the process of defining ADRs for the back end. At the time of writing
...
+* [Preferred export](./back-end/POST-PUT-api-payload.md)
```
nunogois, line 26:
```suggestion
* [POST/PUT API payload](./back-end/POST-PUT-api-payload.md)
```

website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
```diff
@@ -0,0 +1,51 @@
+---
+title: "ADR: POST/PUT API payload"
+---
+
+## Background
+
+Whenever we receive a payload in our backend for POST or PUT request we need to take into account backward compatibility. When we add a new field to an existing API payload, clients using the previous version of the payload will not know about that new field. This means that we need to make sure that the new field is optional. If we make the field required, clients using the previous version of the payload will result in overriding the value of the new field with an empty value or null.
```
nunogois, line 7: This one doesn't seem like an issue in POST, only PUT. But I guess it depends on all the ways we're using POST currently.
```suggestion
Whenever we receive a payload in our backend for POST or PUT requests we need to take into account backwards compatibility. When we add a new field to an existing API payload, clients using the previous version of the payload will not know about that new field. This means that we need to make sure that the new field is optional. If we make the field required, clients using the previous version of the payload will override the value of the new field with an empty value or null.
```

website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Example: adding new setting field to project settings
+
+Project settings on unleash 5.3:
```
nunogois, line 11:
```suggestion
Project settings on Unleash 5.3:
```

website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
```diff
@@ -0,0 +1,51 @@
...
+--data-raw '{
+  "id": "default",
+  "name": "Default",
+  "description": "Default project",
+  "defaultStickiness": "default",
+  "mode": "open"
+}'
+```
+
+New version of project settings (unleash 5.6):
```
nunogois, line 25:
```suggestion
New version of project settings (Unleash 5.6):
```

website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
```diff
@@ -0,0 +1,51 @@
...
+--data-raw '{
+  "id": "default",
+  "name": "Default",
+  "description": "Default project",
+  "defaultStickiness": "default",
+  "featureLimit": 2
+}'
+```
+
+Pay attention to the new field feature limit. If a customer updates their server to 5.6 but their integration still does not send that field, it may result in the unwanted behavior of setting that field to empty in the database (in case the server assumes that not sending the field is tring to set it to empty).
```
nunogois, line 40:
```suggestion
Pay attention to the new field feature limit. If a customer updates Unleash to 5.6 but their integration still does not send that field, it may result in the unwanted behavior of setting that field to empty in the database, in case the server assumes that not sending the field means setting it to empty / `null`.
```

website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
```diff
@@ -0,0 +1,51 @@
...
+This bug can easily be an oversight but can be prevented by following some rules when designing the API payload.
+
+## Decision
+
+When receiving a body from a request we need to take into account 3 possible cases:
+1. The field has a value
+2. The field is `undefined` or not part of the payload
+3. The field is `null`
+
+In case the field has a value, we need to update or set that value in the DB. In case the field is `undefined` or not part of the payload, we need to leave the value in the DB as it is. In case, the field is `null`, we need to remove the value from the DB.
```
nunogois, line 51:
```suggestion
In case the field has a value, we need to update or set that value in the DB. In case the field is `undefined` or not part of the payload, we need to leave the value in the DB as it is. In case the field is `null`, we need to remove the value from the DB (set it as `NULL` on the DB).
```

nunogois (approved):
I like it!

andreas-unleash:
I like it. Just as a note for the featureLimit field (same as the defaultStickiness).  It used to be part of the Create/Update project schema as an optional field with null meaning no limit.  In the same way clients with previous versions expecting to update the featureLimit with a call to updateProject will not be satisfied

gastonfournier:
> I like it. Just as a note for the featureLimit field (same as the defaultStickiness). It used to be part of the Create/Update project schema as an optional field with null meaning no limit. In the same way clients with previous versions expecting to update the featureLimit with a call to updateProject will not be satisfied

null meaning no limit is fine. The issue I found and why I'm writing this, is when we assume that `undefined` is the same as `null`.

daveleek (approved):
Good! Nice and clear, and I like the way of dealing with value/null/undefined and how that affects backwards compatability!

website/docs/contributing/ADRs/back-end/POST-PUT-api-payload.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Decision
+
+When receiving a body from a request we need to take into account 3 possible cases:
+1. The field has a value
+2. The field is `undefined` or not part of the payload
+3. The field is `null`
+
+- If the field has a value, we need to update or set that value in the DB.
+- If the field is `undefined` or not part of the payload, we need to leave the value in the DB as it is.
```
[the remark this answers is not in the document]
thomasheartman: Hmm. I .. don't know. Can you flesh out the example you're giving? I think I understand what you're getting at, but I don't have a clear case where this would actually come to play.

My main point here is that a PUT is intended to replace whatever resource exists. However, we already don't obey this and often treat it more as a PATCH even if we call it a PUT, so maybe it doesn't matter much. But it still feels kinda like bending the rules a bit.

At the same time, this line from [the PUT spec](https://www.rfc-editor.org/rfc/rfc9110#PUT) might allow it:

>  When a PUT representation is inconsistent with the target resource, the origin server SHOULD either make them consistent, by transforming the representation or changing the resource configuration, or respond with an appropriate error message containing sufficient information to explain why the representation is unsuitable. 

However, the example it gives is if it has wrong data.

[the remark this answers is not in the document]
gastonfournier: > Hmm. I .. don't know. Can you flesh out the example you're giving? 

So following up with the example provided in the ADR: **adding new setting field to project settings**

I can have an integration (a script) that let's say allows users to update the description of the default project by running:
```bash
#!/usr/bin/env bash
NEW_DESCRIPTION=$1
curl --location --request PUT 'http://localhost:4242/api/admin/projects/default' \
--header 'Authorization: INSERT_API_KEY' \
--header 'Content-Type: application/json' \
--data-raw '{
  "id": "default",
  "name": "Default",
  "description": "${NEW_DESCRIPTION}",
  "defaultStickiness": "default",
  "mode": "open"
}'
```

Everyone is happy with this script and serves a quintessential part of their business (:trollface:), but then we introduce the feature limit and they are happy with this feature and start using it. But suddenly the value disappears. They set it back and after a few days, it disappears again. After a few meetings and quite a lot of debugging time, they co-relate the disappearance of the feature limit with the update of the project description. 

Nothing is wrong with their script, it was working before and it started to cause issues when they upgraded Unleash.

I believe this is one of those cases where PUT should _make them consistent, by transforming the representation_, and the way I'd do that is by getting the current value of feature limit and injecting it into the body (server-side so the script doesn't have to change). Semantically speaking, this is the same as ignoring the field when `undefined`.

Following up with this idea, imagine we tell them that they have to fix their script. They won't know what value to put into featureLimit without making a GET request first to the server and then manipulating the response body.

If we just follow this idea of undefined means "don't touch it", the script could even be simplified and it'd avoid a similar problem with other fields such as `mode` or `defaultStickiness`:
```bash
#!/usr/bin/env bash
NEW_DESCRIPTION=$1
curl --location --request PUT 'http://localhost:4242/api/admin/projects/default' \
--header 'Authorization: INSERT_API_KEY' \
--header 'Content-Type: application/json' \
--data-raw '{
  "description": "${NEW_DESCRIPTION}",
}'
```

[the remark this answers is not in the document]
thomasheartman: Got it. I see your point, but I'm not convinced. If they only wanted to update the description, why wouldn't they use a PATCH? (Not sure we offer that, but 💁🏼 )

I think we could also use the same example as a way that not removing undefined fields would be breaking.

Taking the same example, the company used that script with an missing description to remove the description of  the default project and this is now integral to their business. If we then changed it such that the description would remain, they would have to do the same debugging again, right? Is that any better?

Of course, that example would be with changing the behavior, and maybe that's definitely breaking, but if we wanna standardize on something, we'll have to change at least a few endpoints going forward.

This feels to me like it goes against the idea of PUT requests. I agree that the empty value should always be sensible and allowed when we introduce a new field (unless we force update all existing entries?), but I'm not sure this is the way to do it.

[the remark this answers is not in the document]
gastonfournier: > Got it. I see your point, but I'm not convinced. If they only wanted to update the description, why wouldn't they use a PATCH? (Not sure we offer that, but 💁🏼 )

We don't, but even if we did, clients should follow the GET & PUT pattern to interact with PUT, which is ok in some situations, but I think this would add additional complexity to our API

> I think we could also use the same example as a way that not removing undefined fields would be breaking
> 
> Taking the same example, the company used that script with a missing description to remove the description of the default project and this is now integral to their business. If we then changed it such that the description would remain, they would have to do the same debugging again, right? Is that any better?

I don't agree with this because `null` allows you to be explicit about this intention of removing the field. This ADR gives you the 3 alternatives:
1. You want to set a value (value = "some value")
2. You want to delete the value (value = null)
3. You don't want to mess around with that value (`undefined`)

> This feels to me like it goes against the idea of PUT requests. I agree that the empty value should always be sensible and allowed when we introduce a new field (unless we force update all existing entries?), but I'm not sure this is the way to do it.

We need to find a way of making PUT backward compatible, otherwise, almost every new version will be breaking or we should never touch PUT request bodies which would require us to create new endpoints or new versions of the endpoint (e.g. PUT /v1/api/admin/projects/default would not have featureLimit while PUT /v2/api/admin/projects/default should always have it). I've used this approach in the past and it was not maintainable, so we landed in the middle ground stated in this ADR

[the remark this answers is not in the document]
thomasheartman: Great! Appreciate that extra bit of context 🙌 I think you're right: the option is using some sort of versioning (whether in the URL, payload, or a header). 

To be clear, I don't mind the approach you're suggesting. In fact, I think I wish that was how we did PATCH apis (because honestly, JSON patch is pretty unwieldy). I'm still not sure it jives with my regular understanding of PUT, but I don't dislike it. 

So if we decide this is the way to go, I'm on board with it. We've had the discussion, you've heard my thoughts. My only lingering thing is that I think it causes a larger chance for misunderstanding (if you think a PUT should clear everything else), but we can be upfront about that, maybe?

PR #5324, opened 2023-11-13
chore: log unerlying DB error in set user root role

kwasniew:
## About the changes

Log the underlying DB error when changing a root role.


nunogois (approved):
LGTM

PR #5325, opened 2023-11-13
Feat/pagination preload

FredrikOseberg:
This PR makes changes to how the project overview skeleton screen works. Important changes:

- Add skeleton screens to missing elements, creating a more comprehensive loading screen
- Split the page into different loading sections, so that we can load the table when we fetch the next page without affecting the rest of the page.

https://www.loom.com/share/e5d30dc897ac488ea80cfae11ffab646

Next steps:
* Hide bar if total is less than 25
* Add FE testing


sjaanus (approved):
Nice LG!

PR #5326, opened 2023-11-13
feat: sort favorites on the backend

sjaanus:
Now favorites will be always on first page, if pinned.

### DIFF
--- a/src/lib/features/feature-search/feature.search.e2e.test.ts
+++ b/src/lib/features/feature-search/feature.search.e2e.test.ts
@@ -56,7 +56,7 @@ const sortFeatures = async (
         sortBy = '',
         sortOrder = '',
         projectId = 'default',
-        favoritesFirst = false,
+        favoritesFirst = 'false',
     }: FeatureSearchQueryParameters,
     expectedCode = 200,
 ) => {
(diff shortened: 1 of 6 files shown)


src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -240,8 +270,14 @@ test('should return all feature tags', async () => {
...
-                    { type: 'simple', value: 'my_tag' },
-                    { type: 'simple', value: 'second_tag' },
+                    {
+                        type: 'simple',
+                        value: 'my_tag',
+                    },
+                    {
+                        type: 'simple',
+                        value: 'second_tag',
+                    },
```
FredrikOseberg, line 280: Nothing related to this PR, but I'm curious why this re-formatting occurs. Does Biome have different settings per dev environment?

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -730,6 +735,7 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
             .select(selectColumns)
             .limit(limit * environmentCount)
             .offset(offset * environmentCount);
+        console.log(query.toQuery());
```
FredrikOseberg, line 738: This should be removed

FredrikOseberg (approved):
LGTM, but we should remove the stray console.log

PR #5327, opened 2023-11-13
feat: include segment usage in CRs when showing usage in projects and flags

thomasheartman:
This PR updates the segment usage counting to also include segment usage in pending change requests. 

The changes include:
- Updating the schema to explicitly call out that change request usage is included.
- Adding two tests to verify the new features
- Writing an alternate query to count this data

Specifically, it'll update the part of the UI that tells you how many places a segment is used:

image

## Implementation

Implementing this was a little tricky. Previously, we'd just count distinct instances of feature names and project names on the feature_strategy table. However, to merge this with change request data, we can't just count existing usage and change request usage separately, because that could cause duplicates.

Instead of turning this into a complex DB query, I've broken it up into a few separate queries and done the merging in JS. I think that's more readable and it was easier to reason about. 

Here's the breakdown:
1. Get the list of pending change requests. We need their IDs and their project.
2. Get the list of updateStrategy and addStrategy events that have segment data.
3. Take the result from step 2 and turn it into a dictionary of segment id to usage data.
4. Query the feature_strategy_segment and feature_strategies table, to get existing segment usage data
5. Fold that data into the change request data.
6. Perform the preexisting segment query (without counting logic) to get other segment data
7. Enrich the results of the query from step 2 with usage data.

## Discussion points

I feel like this could be done in a nicer way, so any ideas on how to achieve that (whether that's as a db query or just breaking up the code differently) is very welcome.

Second, using multiple queries obviously yields more overhead than just a single one. However, I do not think this is in the hot path, so I don't consider performance to be critical here, but I'm open to hearing opposing thoughts on this of course.

### DIFF
--- a/src/lib/db/segment-store.test.ts
+++ b/src/lib/db/segment-store.test.ts
@@ -45,6 +45,7 @@ describe('usage counting', () => {
 
     afterEach(async () => {
         await db.stores.featureToggleStore.deleteAll();
+        await db.stores.segmentStore.deleteAll();
         await db.rawDatabase.table('change_requests').delete();
     });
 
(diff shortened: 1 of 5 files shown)


src/lib/db/segment-store.test.ts
thomasheartman, line 1: There is a bit of repetition here, but it's subtly different, so I'm not sure I wanna extract it just yet. But open to suggestions!

andreas-unleash (approved):
LGTM!

PR #5329, opened 2023-11-13
chore: remove featureSwitchRefactor flag

Tymek:
Cleanup. This change has been rolled out to significant number of customers already, and we have another parallel version behind a flag.


frontend/src/component/project/Project/ProjectFeatureToggles/FeatureToggleSwitch/useFeatureToggleSwitch.tsx
```diff
@@ -96,7 +96,11 @@ export const useFeatureToggleSwitch: UseFeatureToggleSwitchType = (
             };
 
             const ensureActiveStrategies: Middleware = (next) => {
-                if (!config.hasStrategies || config.hasEnabledStrategies) {
+                if (
+                    newState === false ||
```
Tymek, line 100: Other conditions should never be met with `newState===false`, but adding this makes it more consistent with our actual business logic

FredrikOseberg (approved):

PR #5331, opened 2023-11-14
chore: only return change request data if the unleash instance is an enterprise instance

thomasheartman:
Otherwise, we might accidentally display CR data to open source users.
But more importantly, it might keep them from being able to delete a
segment that's in use by a CR in their database that they can't touch.

So by checking that they're on an enterprise instance, we avoid this
potential blocker.

I've added the `includeChangeRequestUsageData` parameter as a boolean now, but I'm open to other suggestions.


andreas-unleash (approved):
LGTM

PR #5335, opened 2023-11-14
fix: date should sort asc/desc properly

sjaanus:
Now date field will sort correctly, when arrow down, it is descending, and when arrow up, it is ascending.
I suspect this is a bug in react table v7.

Old broken:
image

New:
image


Tymek (commented):
How about other tables? I think we need to keep it consistent

FredrikOseberg:
> How about other tables? I think we need to keep it consistent

We should keep consistency across tables.

sjaanus:
> How about other tables? I think we need to keep it consistent

Yes, added for all tables now

sjaanus:
> > How about other tables? I think we need to keep it consistent
> 
> We should keep consistency across tables.

Added for all tables now

FredrikOseberg (approved):
LGTM

Tymek (approved):

PR #5341, opened 2023-11-14
Updated scripts to use owner

chriswk:
See [Linear issue 2-1627](https://linear.app/unleash/issue/2-1627/db-permissions-update-docs-and-migration-guides).

Since we now use functions, we need more permissions than just GRANT ALL PERMISSIONS ON DATBASE. In addition we need to be allowed to create functions in the schema that's actually being used. This PR takes the easy way out and say that we need OWNER privileges on the database. That guarantees that we can do all we do in our migration scripts.

### Discussion points
We might encounter some pushback on this, if so, we'll need to say that we need
`GRANT ALL PRIVILEGES ON ALL FUNCTIONS IN <schema>` in addition to GRANT ALL PRIVILEGES ON DATABASE, where <schema> is the schema selected in their configuration.

### DIFF
--- a/scripts/resetdatabase.sql
+++ b/scripts/resetdatabase.sql
@@ -1,3 +1,2 @@
 DROP DATABASE unleash;
-CREATE DATABASE unleash;
-GRANT ALL PRIVILEGES ON DATABASE unleash TO unleash_user;
+CREATE DATABASE unleash WITH OWNER unleash_user;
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):
Docs look good to me, though we could maybe add a little paragraph before the code snippet explaining why we use the owner? As for the script itself, that looks sensible from the linear ticket, assuming that's been checked by someone else?

PR #5343, opened 2023-11-15
feat: fix paging in sql by using common table expressions

sjaanus:
Optimizations:

1. Removed extra round trip to database  to count environments
2. Removed extra round trip to database to count features

Fixes:

Currently, we were using a very optimistic query to set correct limit and offset. This breaks as soon we we join tags.

` query = query
            .select(selectColumns)
            .limit(limit * environmentCount)
            .offset(offset * environmentCount);` 

The solution was to use common table expressions, so we could count and rank features.

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
@@ -736,7 +736,6 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
             .joinRaw('CROSS JOIN total_features')
             .whereBetween('rank', [offset + 1, offset + limit]);
 
-        console.log(finalQuery.toQuery());
         const rows = await finalQuery;
 
         if (rows.length > 0) {
--- a/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
@@ -541,207 +541,210 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
         const validatedSortOrder =
             sortOrder === 'asc' || sortOrder === 'desc' ? sortOrder : 'asc';
 
-        let environmentCount = 1;
-        if (projectId) {
-            const rows = await this.db('project_environments')
-                .count('* as environmentCount')
-                .where('project_id', projectId);
-            environmentCount = Number(rows[0].environmentCount);
-        }
+        const finalQuery = this.db
+            .with('ranked_features', (query) => {
+                query.from('features');
+                if (projectId) {
+                    query.where({ project: projectId });
+                }
+                const hasQueryString = Boolean(queryString?.trim());
+                const hasHalfTag =
+                    normalizedHalfTag && normalizedHalfTag.length > 0;
+                if (hasQueryString || hasHalfTag) {
+                    const tagQuery = this.db
+                        .from('feature_tag')
+                        .select('feature_name');
+                    // todo: we can run a cheaper query when no colon is detected
+                    if (hasQueryString) {
+                        tagQuery.whereRaw("(?? || ':' || ??) ILIKE ?", [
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -541,207 +541,209 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
         const validatedSortOrder =
             sortOrder === 'asc' || sortOrder === 'desc' ? sortOrder : 'asc';
 
-        let environmentCount = 1;
-        if (projectId) {
-            const rows = await this.db('project_environments')
```
sjaanus, line 546: Removed this call to database completely.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -541,207 +541,209 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
-        } else {
-            query = query.orderBy('created_at', validatedSortOrder);
-        }
+                let rankingSql = 'order by ';
+                if (favoritesFirst) {
+                    rankingSql +=
+                        'favorite_features.feature is not null desc, ';
+                }
 
-        const total = await countQuery
```
sjaanus, line 728: Removed this call to database completely.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -541,207 +541,209 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
-            .countDistinct({ total: 'features.name' })
-            .first();
+                if (sortBy.startsWith('environment:')) {
+                    const [, envName] = sortBy.split(':');
+                    rankingSql += this.db
```
sjaanus, line 710: This is the existing sorting part, which we building into `denseRank` query. Knex does not support this function properly, so had to build raw query.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -541,207 +541,209 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
-            .select(selectColumns)
-            .limit(limit * environmentCount)
-            .offset(offset * environmentCount);
+                query
+                    .select(selectColumns)
+                    .denseRank('rank', this.db.raw(rankingSql));
+            })
+            .with(
+                'total_features',
+                this.db.raw('select max(rank) as total from ranked_features'),
```
sjaanus, line 732: Instead of extra query, we can get total features from here.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -541,207 +541,209 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
+            )
+            .select('*')
+            .from('ranked_features')
+            .joinRaw('CROSS JOIN total_features')
+            .whereBetween('rank', [offset + 1, offset + limit]);
```
sjaanus, line 737: Instead of using offset, limit, we can utilize rank to return proper results.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -541,207 +541,209 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
 
-        let environmentCount = 1;
-        if (projectId) {
-            const rows = await this.db('project_environments')
-                .count('* as environmentCount')
-                .where('project_id', projectId);
-            environmentCount = Number(rows[0].environmentCount);
-        }
+        const finalQuery = this.db
+            .with('ranked_features', (query) => {
```
sjaanus, line 545: Using CTE

PR #5344, opened 2023-11-16
feat: new project overview backend

sjaanus:
Adding new project overview endpoint and deprecating the old one.
The new one has extra info about feature types, but does not have features anymore, because features are coming from search endpoint.


kwasniew (approved):
Pair reviewed

PR #5348, opened 2023-11-16
feat(banners): enable customer banners

gastonfournier:
## About the changes
This feature allows our Enterprise customers to configure banners to be displayed on their Unleash instance for all their users to see and interact with. Previously known as "internal message banners".

### DIFF
--- a/frontend/src/component/admin/adminRoutes.ts
+++ b/frontend/src/component/admin/adminRoutes.ts
@@ -77,7 +77,6 @@ export const adminRoutes: INavigationMenuItem[] = [
     {
         path: '/admin/banners',
         title: 'Banners',
-        flag: 'banners',
         menu: { adminSettings: true, mode: ['enterprise'] },
         group: 'instance',
     },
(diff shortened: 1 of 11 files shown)


nunogois (approved):
LGTM. Turns out the tests were just flaky, and it seems like some recent changes may have caused them to be even flakier than usual 🤷‍♂️ 

I still think this approach should be shared with the rest of the team. We should be aware that there's an option to disable this feature, gather feedback about the approach, and that we may want to follow a similar approach for other features in the future.

gastonfournier:
> LGTM. Turns out the tests were just flaky, and it seems like some recent changes may have caused them to be even flakier than usual 🤷‍♂️
> 
> I still think this approach should be shared with the rest of the team. We should be aware that there's an option to disable this feature, gather feedback about the approach, and that we may want to follow a similar approach for other features in the future.

Well after our talk, I lean towards keeping it. It's an opt-in feature (just a menu item) that can be ignored if the user doesn't want it. So for the sake of simplicity I'm keeping them enabled by default

PR #5352, opened 2023-11-16
fix: take into account project segments permission in form

nunogois:
https://linear.app/unleash/issue/SR-184/ticket-1106-users-with-createedit-project-segment-dont-see-all-the

https://github.com/Unleash/unleash/pull/5304 did not take into account permissions further into the Segment form.

This PR fixes the remaining permission checks to take into consideration the project-level permission: `UPDATE_PROJECT_SEGMENT`.


kwasniew (approved):
We can later add a test proving that it actually fixed the problem

PR #5353, opened 2023-11-16
fix: Only show strategy variant changes if there is a diff in the variants

andreas-unleash:
What it says on the box

Closes # [1-1652](https://linear.app/unleash/issue/1-1652/remove-the-variants-from-change-request-page-when-not-modified)

Screenshot 2023-11-16 at 11 26 05

Screenshot 2023-11-16 at 11 25 46


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -127,20 +127,19 @@ export const StrategyChange: VFC<{
...
         change.payload.variants &&
-        change.payload.variants.length > 0 ? (
-            <StyledBox>
-                <StyledTypography>
-                    Updating feature variants to:
-                </StyledTypography>
-                <EnvironmentVariantsTable variants={change.payload.variants} />
-            </StyledBox>
-        ) : null;
+        change.payload.variants.length > 0 &&
```
thomasheartman, line 141: Will this fail if the new strategy has no variants but the old one did?

andreas-unleash: Good catch

andreas-unleash: Fixed

thomasheartman (approved):
Looks better 😄  I'll give this an approval now, but I'd love to see the case you just fixed tested. Maybe I didn't read the tests right, but I didn't see it listed as one of the test cases?

PR #5356, opened 2023-11-16
Add redirects from 404 logs

dgorton:
Add more redirects from the docs refactor

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -431,6 +431,7 @@ module.exports = {
                         from: [
                             '/user_guide/important-concepts',
                             '/tutorials/important-concepts',
+                            '/reference/concepts/',
                         ],
                         to: '/reference',
                     },
@@ -532,7 +533,10 @@ module.exports = {
                         to: '/',
                     },
                     {
-                        from: '/topics/feature-flags/tutorials',
+                        from: [
+                            '/topics/feature-flags/tutorials',
+                            '/tutorials',
+                        ],
                         to: '/feature-flag-tutorials',
                     },
                     {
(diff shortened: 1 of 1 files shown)


website/docusaurus.config.js
```diff
@@ -532,7 +533,10 @@ module.exports = {
                         to: '/',
                     },
                     {
-                        from: '/topics/feature-flags/tutorials',
+                        from: [
+                            '/topics/feature-flags/tutorials',
+                            '/tutorials',
```
thomasheartman, line 538: Are you sure we want this redirect? Will we ever have tutorials that aren't related to feature flags? (Or at least that shouldn't be under that url?)

website/docusaurus.config.js
```diff
@@ -574,19 +578,31 @@ module.exports = {
...
+                        from: [
+                            '/tutorials/managing-constraints',
+                            '/topics/managing-constraints',
+                        ],
                         to: '/understanding-unleash/managing-constraints',
                     },
                     {
-                        from: '/tutorials/the-anatomy-of-unleash',
+                        from: [
+                            '/tutorials/the-anatomy-of-unleash',
```
thomasheartman, line 589: Wait. actually: was this page ever under "tutorials"? That feels off to me. The new redirect looks right though. And that actually goes for all the others you've fixed now too. I don't think they were ever under tutorials? 🤔

thomasheartman (approved):
Couple questions that we can use to clean up, but nothing blocking. If you're happy to merge, go ahead 😄

website/docusaurus.config.js
dgorton: Yeah - it must have been there for at least a while - I got this from Google's 404 report. I don't know where the link is coming from yet, but I wanted to fix this in the meantime.

website/docusaurus.config.js
dgorton: We will probably have others over time, but for now this is the right place. We are planning on building enough of these to merit it's own category. We may add other categories in the future as we test and iterate, but I think this is the best choice now.

PR #5359, opened 2023-11-17
chore: remove segments oss splash

thomasheartman:
This change removes the oss segments splash screen that was introduced
in https://github.com/Unleash/unleash/pull/5053.


nunogois (approved):
LGTM

PR #5362, opened 2023-11-17
feat: useTableState hook

Tymek:
## About the changes
Simplified logic for handling interaction between URL (query), table state and localstorage.

### Important files
Well, `useTableState.test.ts` - >2x LOC in tests

## Discussion points
What do we do with new filters?


Tymek:
Next up: https://github.com/Unleash/unleash/pull/5371

FredrikOseberg (approved):
Nice!

PR #5363, opened 2023-11-17
feat: scheduler runtime control

nunogois:
https://linear.app/unleash/issue/2-1645/address-post-mortem-action-point-all-flags-should-be-runtime

Refactor with the goal of ensuring that flags are runtime controllable, mostly focused on the current scheduler logic.

This includes the following changes:
 - Moves scheduler into its own "scheduler" feature folder
 - Reverts dependency: SchedulerService takes in the MaintenanceService, not the other way around
 - Scheduler now evaluates maintenance mode at runtime instead of relying on its mode state (active / paused)
 - Favors flag checks to happen inside the scheduled methods, instead of controlling whether the method is scheduled at all (favor runtime over startup)
 - Updates tests accordingly
 - Boyscouting

Scheduler mode is currently not being used, along with its respective methods, except in tests. Should we remove it in favor of only controlling the scheduler state through maintenance mode?

Are there any other tests we should add?

### DIFF
--- a/src/lib/features/scheduler/scheduleServices.ts
+++ b/src/lib/features/scheduler/scheduleServices.ts
@@ -1,0 +1,140 @@
+import {
+    hoursToMilliseconds,
+    minutesToMilliseconds,
+    secondsToMilliseconds,
+} from 'date-fns';
+import { IUnleashServices } from '../../server-impl';
+
+export const scheduleServices = async (
+    services: IUnleashServices,
+): Promise<void> => {
+    const {
+        schedulerService,
+        apiTokenService,
+        instanceStatsService,
+        clientInstanceService,
+        projectService,
+        projectHealthService,
+        configurationRevisionService,
+        eventAnnouncerService,
+        featureToggleService,
+        versionService,
+        lastSeenService,
+        proxyService,
+        clientMetricsServiceV2,
+    } = services;
+
+    schedulerService.schedule(
+        lastSeenService.cleanLastSeen.bind(lastSeenService),
+        hoursToMilliseconds(1),
+        'cleanLastSeen',
+    );
+
+    schedulerService.schedule(
+        lastSeenService.store.bind(lastSeenService),
+        secondsToMilliseconds(30),
+        'storeLastSeen',
+    );
+
+    schedulerService.schedule(
+        apiTokenService.fetchActiveTokens.bind(apiTokenService),
+        minutesToMilliseconds(1),
+        'fetchActiveTokens',
+    );
+
+    schedulerService.schedule(
+        apiTokenService.updateLastSeen.bind(apiTokenService),
+        minutesToMilliseconds(3),
+        'updateLastSeen',
+    );
+
+    schedulerService.schedule(
+        instanceStatsService.refreshStatsSnapshot.bind(instanceStatsService),
+        minutesToMilliseconds(5),
+        'refreshStatsSnapshot',
+    );
+
+    schedulerService.schedule(
+        clientInstanceService.removeInstancesOlderThanTwoDays.bind(
+            clientInstanceService,
+        ),
+        hoursToMilliseconds(24),
+        'removeInstancesOlderThanTwoDays',
+    );
+
+    schedulerService.schedule(
+        clientInstanceService.bulkAdd.bind(clientInstanceService),
+        secondsToMilliseconds(5),
+        'bulkAddInstances',
+    );
+
+    schedulerService.schedule(
+        clientInstanceService.announceUnannounced.bind(clientInstanceService),
+        minutesToMilliseconds(5),
+        'announceUnannounced',
+    );
+
+    schedulerService.schedule(
+        projectService.statusJob.bind(projectService),
+        hoursToMilliseconds(24),
+        'statusJob',
+    );
+
+    schedulerService.schedule(
+        projectHealthService.setHealthRating.bind(projectHealthService),
+        hoursToMilliseconds(1),
+        'setHealthRating',
+    );
+
+    schedulerService.schedule(
+        configurationRevisionService.updateMaxRevisionId.bind(
+            configurationRevisionService,
+        ),
+        secondsToMilliseconds(1),
+        'updateMaxRevisionId',
+    );
+
+    schedulerService.schedule(
+        eventAnnouncerService.publishUnannouncedEvents.bind(
+            eventAnnouncerService,
+        ),
+        secondsToMilliseconds(1),
+        'publishUnannouncedEvents',
+    );
+
+    schedulerService.schedule(
+        featureToggleService.updatePotentiallyStaleFeatures.bind(
+            featureToggleService,
+        ),
+        minutesToMilliseconds(1),
+        'updatePotentiallyStaleFeatures',
+    );
+
+    schedulerService.schedule(
+        versionService.checkLatestVersion.bind(versionService),
+        hoursToMilliseconds(48),
+        'checkLatestVersion',
+    );
+
+    schedulerService.schedule(
+        proxyService.fetchFrontendSettings.bind(proxyService),
+        minutesToMilliseconds(2),
+        'fetchFrontendSettings',
+    );
+
+    schedulerService.schedule(
+        () => {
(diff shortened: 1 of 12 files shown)


FredrikOseberg:
> https://linear.app/unleash/issue/2-1645/address-post-mortem-action-point-all-flags-should-be-runtime
> 
> Refactor with the goal of ensuring that flags are runtime controllable, mostly focused on the current scheduler logic.
> 
> This includes the following changes:
> 
> * Moves scheduler into its own "scheduler" feature folder
> * Reverts dependency: SchedulerService takes in the MaintenanceService, not the other way around
> * Scheduler now evaluates maintenance mode at runtime instead of relying on its mode state (active / paused)
> * Favors flag checks to happen inside the scheduled methods, instead of controlling whether the method is scheduled at all (favor runtime over startup)
> * Moved account last seen update to scheduler
> * Updates tests accordingly
> * Boyscouting
> 
> Scheduler `mode` (active / paused) is currently not being used, along with its respective methods, except in tests. I think this could be a potential footgun. Should we remove it in favor of only controlling the scheduler state through maintenance mode?
> 
> The config property `disableScheduler` is still a startup configuration, but perhaps that makes sense to leave as is?
> 
> Are there any other tests we should add?
> 
> Identified some `setInterval` and `setTimeout` that may make sense to leave as-is instead of moving over to the scheduler service:
> 
> * `src/lib/metrics` - This is currently considered a `MetricsMonitor`. Should this be refactored to a service instead and adapt these setIntervals to use the scheduler instead? Is there anything special with this we need to take into account? @chriswk @ivarconr
> * `src/lib/proxy/proxy-repository.ts` - This seems to have a complex and specific logic currently. Perhaps we should leave it alone for now? @FredrikOseberg
> * `src/lib/services/user-service.ts` - This one also seems to be a bit more specific, where we generate new timeouts for each receiver id. Might not belong in the scheduler service. @Tymek



> https://linear.app/unleash/issue/2-1645/address-post-mortem-action-point-all-flags-should-be-runtime
> 
> Refactor with the goal of ensuring that flags are runtime controllable, mostly focused on the current scheduler logic.
> 
> This includes the following changes:
> 
> * Moves scheduler into its own "scheduler" feature folder
> * Reverts dependency: SchedulerService takes in the MaintenanceService, not the other way around
> * Scheduler now evaluates maintenance mode at runtime instead of relying on its mode state (active / paused)
> * Favors flag checks to happen inside the scheduled methods, instead of controlling whether the method is scheduled at all (favor runtime over startup)
> * Moved account last seen update to scheduler
> * Updates tests accordingly
> * Boyscouting
> 
> Scheduler `mode` (active / paused) is currently not being used, along with its respective methods, except in tests. I think this could be a potential footgun. Should we remove it in favor of only controlling the scheduler state through maintenance mode?
> 
> The config property `disableScheduler` is still a startup configuration, but perhaps that makes sense to leave as is?
> 
> Are there any other tests we should add?
> 
> Identified some `setInterval` and `setTimeout` that may make sense to leave as-is instead of moving over to the scheduler service:
> 
> * `src/lib/metrics` - This is currently considered a `MetricsMonitor`. Should this be refactored to a service instead and adapt these setIntervals to use the scheduler instead? Is there anything special with this we need to take into account? @chriswk @ivarconr
> * `src/lib/proxy/proxy-repository.ts` - This seems to have a complex and specific logic currently. Perhaps we should leave it alone for now? @FredrikOseberg
> * `src/lib/services/user-service.ts` - This one also seems to be a bit more specific, where we generate new timeouts for each receiver id. Might not belong in the scheduler service. @Tymek

Great questions. 

I was planning to add a Gauge that basically records whether or not the scheduler is on for observability. But if it can never be turned off, we have no way of determining this state. However, if the scheduler is always running but it's just the execution of the functions that's being controlled by maintenance mode, maybe it's enough to just have observability on the individual scheduled functions. 

Though this approach has a drawback. It's harder to create alerting rules, and there are different intervals for different scheduled calls. We'll need to more carefully calibrate Grafana alert treshold based on individual schedule frequency.

Note: 
* disableScheduler property in the config was created for us to be able to disable the scheduler in tests if I remember correctly
* the setTimeouts in `proxy-repository.ts` is specialized because it is injected as a repository into the unleash client as part of the frontend API where we are using the node-js client for evaluations. I think it's best if we leave this for now.

I'd like to discuss what we do with the schedule service states before I proceed with the observability enhancements.

src/lib/features/scheduler/scheduleServices.ts
```diff
@@ -0,0 +1,147 @@
+import {
+    hoursToMilliseconds,
+    minutesToMilliseconds,
+    secondsToMilliseconds,
+} from 'date-fns';
+import { IUnleashServices } from '../../server-impl';
+
```
FredrikOseberg, line 7: Would like to see a comment here

src/lib/features/scheduler/scheduler-service.ts
```diff
@@ -9,8 +10,14 @@ export class SchedulerService {
...
 
-    constructor(getLogger: LogProvider) {
+    private maintenanceService: MaintenanceService;
+
+    constructor(
+        getLogger: LogProvider,
+        maintenanceService: MaintenanceService,
+    ) {
         this.logger = getLogger('/services/scheduler-service.ts');
+        this.maintenanceService = maintenanceService;
```
FredrikOseberg, line 20: Maybe this should be a read model instead of exposing the entire maintenance service. I'd like it to happen in a follow up though PR if we go that route since I need this merged.

src/lib/features/scheduler/scheduler-service.ts
```diff
@@ -22,7 +29,9 @@ export class SchedulerService {
         this.intervalIds.push(
             setInterval(async () => {
                 try {
-                    if (this.mode === 'active') {
+                    const maintenanceMode =
+                        await this.maintenanceService.isMaintenanceMode();
+                    if (this.mode === 'active' && !maintenanceMode) {
```
FredrikOseberg, line 34: I think we decided to remove these states

FredrikOseberg (changes requested):
Nice. There's a few minor things left to clean up that we discussed yesterday, otherwise this looks good to me!

ivarconr:
> src/lib/metrics - This is currently considered a MetricsMonitor. Should this be refactored to a service instead and adapt these setIntervals to use the scheduler instead? Is there anything special with this we need to take into account? @chriswk @ivarconr

This could be a scheduled job as well. I am unsure if we want to pause it when we are entering maintenance mode though, as it will stop some of our observability in to the running instance. On the other hand that might be ok, depending on the data being collected (E.g. there will not be any change in the number of toggles during maintenance mode)

ivarconr (approved):
Looks great!

src/lib/features/scheduler/scheduleServices.ts
nunogois: Explored creating a custom linting rule in [Biome](https://github.com/biomejs/biome) and, as far as I could tell, I don't think that's supported yet. Added a comment in https://github.com/Unleash/unleash/pull/5363/commits/dcaaefd00058579bd2546cfe7990f1c1eec05d59

src/lib/features/scheduler/scheduler-service.ts
nunogois: Addressed in https://github.com/Unleash/unleash/pull/5363/commits/7c52e3f63826fe9a7c572b959252fc042a0a0cf2

nunogois:
> > src/lib/metrics - This is currently considered a MetricsMonitor. Should this be refactored to a service instead and adapt these setIntervals to use the scheduler instead? Is there anything special with this we need to take into account? @chriswk @ivarconr
> 
> This could be a scheduled job as well. I am unsure if we want to pause it when we are entering maintenance mode though, as it will stop some of our observability in to the running instance. On the other hand that might be ok, depending on the data being collected (E.g. there will not be any change in the number of toggles during maintenance mode)

Maybe we could have a different schedule method (or pass in a different value that specifies this) for methods that should not pause when maintenance mode is enabled.

Something to think about and perhaps implement in a future PR.

FredrikOseberg (approved):
Great job!

PR #5365, opened 2023-11-20
chore: find segment strategies in CRs

thomasheartman:
This PR adds the ability to detect which strategies use a specific segment in active change requests.

It does not wire this functionality up to anything just yet. Follow-up PRs will integrate this with the segment service and eventually with the front end.

### DIFF
--- a/src/lib/features/change-request-segment-usage-service/sql-change-request-segment-usage-read-model.ts
+++ b/src/lib/features/change-request-segment-usage-service/sql-change-request-segment-usage-read-model.ts
@@ -1,5 +1,8 @@
 import { Db } from '../../db/db';
-import { IChangeRequestSegmentUsageReadModel } from './change-request-segment-usage-read-model';
+import {
+    ChangeRequestStrategy,
+    IChangeRequestSegmentUsageReadModel,
+} from './change-request-segment-usage-read-model';
 
 export class ChangeRequestSegmentUsageReadModel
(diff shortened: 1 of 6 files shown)


src/lib/features/change-request-segment-usage-service/sql-change-request-segment-usage-read-model.ts
```diff
@@ -1,5 +1,8 @@
 import { Db } from '../../db/db';
-import { IChangeRequestSegmentUsageReadModel } from './change-request-segment-usage-read-model';
+import {
+    ChangeRequestStrategy,
+    IChangeRequestSegmentUsageReadModel,
+} from './change-request-segment-usage-read-model';
 
 export class ChangeRequestSegmentUsageReadModel
```
andreas-unleash, line 7: I would name the file as `change-request-segment-usage-read-model`

andreas-unleash (approved):
LGTM. Minor comment, maybe we don't need the `sql-change...` file

src/lib/features/change-request-segment-usage-service/sql-change-request-segment-usage-read-model.ts
thomasheartman: Good input, and I would usually do the same. But I'm following the precedent set by the change request access read model (`src/lib/features/change-request-access-service/sql-change-request-access-read-model.ts`), so I think I'd like to keep it that way for now.

PR #5367, opened 2023-11-20
refactor: remove feature flag for Dora

FredrikOseberg:
Discussion point:
* Is frontend protection enough for this feature?

### DIFF
--- a/frontend/src/component/common/ExperimentalFeedback/ExperimentalFeedback.tsx
+++ b/frontend/src/component/common/ExperimentalFeedback/ExperimentalFeedback.tsx
@@ -68,12 +68,14 @@ interface IExperimentalFeedbackProps {
     sketchURL: string;
 }
 
-export const ExperimentalFeedback = ({ trackerKey, eventKey, description, sketchURL }) => {
+export const ExperimentalFeedback = ({
+    trackerKey,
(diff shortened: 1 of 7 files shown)


frontend/src/component/common/ExperimentalFeedback/ExperimentalFeedback.tsx
```diff
@@ -61,10 +61,17 @@ const StyledLink = styled('a')(({ theme }) => ({
...
 
-export const ProjectDoraFeedback = () => {
+interface IExperimentalFeedbackProps {
+    trackerKey: string;
+    eventKey: string;
+    description: string;
+    sketchURL: string;
+}
+
+export const ExperimentalFeedback = ({ trackerKey, eventKey, description, sketchURL }) => {
```
FredrikOseberg, line 71: Refactor the feedback component to a general component instead of deleting it

thomasheartman (approved):
Looks sensible to me. Would it be bad if people used the API for this, you mean?

FredrikOseberg:
> Looks sensible to me. Would it be bad if non-enterprise people used the API for this, you mean?

I don't think it's a big deal. I'll merge this for now and we can always course-correct.

PR #5368, opened 2023-11-20
fix: handle check against non existing projects

thomasheartman:
Instead of throwing an error when the project doesn't exist, we say that the names are valid, because we have nothing to say that they're not. Presumably there is already something in place to prevent you from importing into a non-existent project.

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-service.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-service.ts
@@ -1168,19 +1168,26 @@ class FeatureToggleService {
         projectId: string,
         featureNames: string[],
     ): Promise<FeatureNameCheckResultWithFeaturePattern> {
-        const project = await this.projectStore.get(projectId);
-        const patternData = project.featureNaming;
-        const namingPattern = patternData?.pattern;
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1168,19 +1168,26 @@ class FeatureToggleService {
...
+                );
+
+                if (result.state === 'invalid') {
+                    return { ...result, featureNaming: patternData };
+                }
             }
+        } catch {
+            // the project doesn't exist, so there's nothing to
+            // validate against
+            return { state: 'valid' };
```
gastonfournier, line 1190: This means any error, not necessarily that the project doesn't exist, maybe adding an info log with the actual error would be ok

gastonfournier (approved):
LG! Just a minor comment that can be disregarded

src/lib/features/feature-toggle/feature-toggle-service.ts
thomasheartman: Good point!

PR #5374, opened 2023-11-21
refactor: maintenance read model

nunogois:
https://linear.app/unleash/issue/2-1655/refactor-maintenance-to-feature-folder-and-add-a-read-model-for-status

This PR does 2 things:
 - Moves all "maintenance" files to a `maintenance` features folder
 - Adds a `IMaintenanceStatus` read model that only includes `isMaintenanceMode()`, so we can use this interface in SchedulerService and expose only the `isMaintenanceMode()` method instead of the entire `MaintenanceService`

Is this what you meant in https://github.com/Unleash/unleash/pull/5363#discussion_r1400170835 @FredrikOseberg?

### DIFF
--- a/src/lib/features/scheduler/scheduler-service.ts
+++ b/src/lib/features/scheduler/scheduler-service.ts
@@ -1,6 +1,6 @@
 import EventEmitter from 'events';
 import { Logger, LogProvider } from '../../logger';
-import MaintenanceService from '../../services/maintenance-service';
+import { IMaintenanceStatus } from '../maintenance/maintenance-service';
 import { SCHEDULER_JOB_TIME } from '../../metric-events';
 
 export class SchedulerService {
@@ -8,17 +8,17 @@ export class SchedulerService {
 
     private logger: Logger;
 
-    private maintenanceService: MaintenanceService;
+    private maintenanceStatus: IMaintenanceStatus;
 
     private eventBus: EventEmitter;
 
     constructor(
         getLogger: LogProvider,
-        maintenanceService: MaintenanceService,
+        maintenanceStatus: IMaintenanceStatus,
         eventBus: EventEmitter,
     ) {
         this.logger = getLogger('/services/scheduler-service.ts');
-        this.maintenanceService = maintenanceService;
+        this.maintenanceStatus = maintenanceStatus;
         this.eventBus = eventBus;
     }
 
@@ -46,7 +46,7 @@ export class SchedulerService {
             setInterval(async () => {
                 try {
                     const maintenanceMode =
-                        await this.maintenanceService.isMaintenanceMode();
+                        await this.maintenanceStatus.isMaintenanceMode();
                     if (!maintenanceMode) {
                         await runScheduledFunctionWithEvent();
                     }
@@ -59,7 +59,7 @@ export class SchedulerService {
         );
         try {
             const maintenanceMode =
-                await this.maintenanceService.isMaintenanceMode();
+                await this.maintenanceStatus.isMaintenanceMode();
             if (!maintenanceMode) {
                 await runScheduledFunctionWithEvent();
             }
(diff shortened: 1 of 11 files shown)


FredrikOseberg:
> https://linear.app/unleash/issue/2-1655/refactor-maintenance-to-feature-folder-and-add-a-read-model-for-status
> 
> This PR does 2 things:
> 
> * Moves all "maintenance" files to a `maintenance` features folder
> * Adds a `IMaintenanceStatus` read model that only includes `isMaintenanceMode()`, so we can use this interface in SchedulerService and expose only the `isMaintenanceMode()` method instead of the entire `MaintenanceService`
> 
> Is this what you meant in [#5363 (comment)](https://github.com/Unleash/unleash/pull/5363#discussion_r1400170835) @FredrikOseberg?

This actually looks quite elegant, though it's different from how we usually do read models. I usually favor consistency, but I like the simplicity of this. I'll need to think about it, and let others weigh in in the meantime.

src/lib/features/scheduler/scheduler-service.ts
```diff
@@ -1,19 +1,16 @@
...
+    private maintenanceStatus: IMaintenanceStatus;
 
-    constructor(
-        getLogger: LogProvider,
-        maintenanceService: MaintenanceService,
-    ) {
+    constructor(getLogger: LogProvider, maintenanceStatus: IMaintenanceStatus) {
         this.logger = getLogger('/services/scheduler-service.ts');
-        this.maintenanceService = maintenanceService;
+        this.maintenanceStatus = maintenanceStatus;
```
FredrikOseberg, line 13: Will this restrict you from using other methods on the maintainance service? If I read it correctly the typechecker would at least complain?

nunogois:
> > https://linear.app/unleash/issue/2-1655/refactor-maintenance-to-feature-folder-and-add-a-read-model-for-status
> > This PR does 2 things:
> > 
> > * Moves all "maintenance" files to a `maintenance` features folder
> > * Adds a `IMaintenanceStatus` read model that only includes `isMaintenanceMode()`, so we can use this interface in SchedulerService and expose only the `isMaintenanceMode()` method instead of the entire `MaintenanceService`
> > 
> > Is this what you meant in [#5363 (comment)](https://github.com/Unleash/unleash/pull/5363#discussion_r1400170835) @FredrikOseberg?
> 
> This actually looks quite elegant, though it's different from how we usually do read models. I usually favor consistency, but I like the simplicity of this. I'll need to think about it, and let others weigh in in the meantime.

Can you share me an example of an existing read model so I can have a look?

FredrikOseberg:
> > > https://linear.app/unleash/issue/2-1655/refactor-maintenance-to-feature-folder-and-add-a-read-model-for-status
> > > This PR does 2 things:
> > > 
> > > * Moves all "maintenance" files to a `maintenance` features folder
> > > * Adds a `IMaintenanceStatus` read model that only includes `isMaintenanceMode()`, so we can use this interface in SchedulerService and expose only the `isMaintenanceMode()` method instead of the entire `MaintenanceService`
> > > 
> > > Is this what you meant in [#5363 (comment)](https://github.com/Unleash/unleash/pull/5363#discussion_r1400170835) @FredrikOseberg?
> > 
> > 
> > This actually looks quite elegant, though it's different from how we usually do read models. I usually favor consistency, but I like the simplicity of this. I'll need to think about it, and let others weigh in in the meantime.
> 
> Can you share an example of an existing read model so I can have a look?

Certainly! It's a bit more involved, and more walled off in terms of the actual code. Here are a few implementations: 
* https://github.com/Unleash/unleash/blob/62dbf8d12ff4a3a5e85094c7e5e9a71ccfa8a324/src/lib/features/feature-toggle/features-read-model.ts#L4
* https://github.com/Unleash/unleash/blob/62dbf8d12ff4a3a5e85094c7e5e9a71ccfa8a324/src/lib/services/client-metrics/last-seen/last-seen-read-model.ts#L4
* https://github.com/Unleash/unleash/blob/62dbf8d12ff4a3a5e85094c7e5e9a71ccfa8a324/src/lib/features/change-request-access-service/change-request-access-read-model.ts#L4

gastonfournier (approved):
For this PR and for the sake of accessing maintenance read only. I think this is good enough. The conversation about read models might require more unpacking of facts about why one approach is better than the other

PR #5375, opened 2023-11-21
chore: use query to get strategies instead of whether a segment is in use

thomasheartman:
This change is in preparation for the upcoming implementation (working
in https://github.com/Unleash/unleash/pull/5369). However, because the
implementation is a lot of lines, this PR is to break it apart and
make it easier to merge.

In this stage, we refactor the segment usage read model. Instead of
checking just whether a segment is in use, we now extract the list of
strategies that use this segment. This is slightly more costly,
perhaps, but it will be necessary for the upcoming implementation.


andreas-unleash (approved):
LGTM.

PR #5376, opened 2023-11-21
feat: sorting by last seen, environments now working properly

sjaanus:
Now calculates final ranks also, if there are some ranks missing from duplicates.

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
@@ -747,7 +747,6 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
             .joinRaw('CROSS JOIN total_features')
             .whereBetween('final_rank', [offset + 1, offset + limit]);
 
-        console.log(finalQuery.toQuery());
         const rows = await finalQuery;
 
         if (rows.length > 0) {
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -727,15 +727,27 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
+            .innerJoin(
+                'final_ranks',
+                'ranked_features.feature_name',
+                'final_ranks.feature_name',
+            )
             .joinRaw('CROSS JOIN total_features')
-            .whereBetween('rank', [offset + 1, offset + limit]);
+            .whereBetween('final_rank', [offset + 1, offset + limit]);
 
+        console.log(finalQuery.toQuery());
```
FredrikOseberg, line 750: Needs to be removed

FredrikOseberg (changes requested):
Console.log must be removed before merging. Otherwise LG.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
sjaanus: Damn, the console always sneaking in. Nice catch.

PR #5377, opened 2023-11-21
feat: scheduled functions observability

FredrikOseberg:
See linear issue: https://linear.app/unleash/issue/1-1656/add-scheduler-observability

As per post mortem actions, we are adding observability to scheduled functions.

This PR adds prometheus observability to our scheduled functions via a summary. In addition to timing these functions with the more accurate process.hrtime, we are getting a counter for free per scheduled job id.

Prometheus example: 

Skjermbilde 2023-11-21 kl  13 36 33


gardleopard:
@FredrikOseberg can you add an example of how the metrics looks in our prometheus endpoint?

gardleopard (approved):
LGTM

PR #5378, opened 2023-11-21
feat: return CR uses of segments when flag is active

thomasheartman:
This PR changes the payload of the strategiesBySegment endpoint when the flag is active. In addition to returning just the strategies, the object will also contain a new property, called `changeRequestStrategies` containing the strategies that are used in change requests.

This PR does not update the schema. That can be done later when the changes go into beta. This also allows us some time to iterate on the payload without changing the public API.

## Discussion points:

Should `strategies` and `changeRequestStrategies` ever contain duplicates? Take this scenario:
- Strategy S uses segment T.
- There is an open change request that updates the list of segments for S to T and a new segment U.
- In this case, strategy S would show up both in `strategies` _and_ in `changeRequestStrategies`. 

We have two options: 
1. Filter the list of change request strategies, so that they don't contain any duplicates (this is currently how it's implemented)
2. Ignore the duplicates and just send both lists as is.

What do you think we should do?


src/lib/features/segment/segment-controller.ts
```diff
@@ -347,25 +347,48 @@ export class SegmentsController extends Controller {
         req: IAuthRequest<{ id: number }>,
         res: Response<SegmentStrategiesSchema>,
     ): Promise<void> {
-        const { id } = req.params;
+        const id = Number(req.params.id);
```
thomasheartman, line 350: This was being passed as a string all the way through to the end, which made the query fail to find anything. However, we type it as `number` everywhere after this, so let's actually do the conversion.

src/lib/services/segment-service.ts
```diff
@@ -90,39 +93,50 @@ export class SegmentService implements ISegmentService {
...
+
+        const strategyIds = new Set(strategies.map((s) => s.id));
+
+        const changeRequestStrategies = (
+            await this.changeRequestSegmentUsageReadModel.getStrategiesUsedInActiveChangeRequests(
+                id,
+            )
+        ).filter(
+            (strategy) => !('id' in strategy && strategyIds.has(strategy.id)),
+        );
```
thomasheartman, line 130: This is the point of contention mentioned in the PR description: should we filter out duplicate IDs or not?

src/lib/services/segment-service.ts
```diff
@@ -300,11 +314,13 @@ export class SegmentService implements ISegmentService {
...
-        const strategies =
-            await this.featureStrategiesStore.getStrategiesBySegment(id);
+        const { strategies, changeRequestStrategies } =
+            await this.getAllStrategies(id);
 
         const projectsUsed = new Set(
-            strategies.map((strategy) => strategy.projectId),
+            [strategies, changeRequestStrategies].flatMap((strats) =>
+                strats.map((strategy) => strategy.projectId),
+            ),
```
thomasheartman, line 323: This prevents you from turning a global segment to a project segment if it is used in CRs in different projects.

src/test/e2e/api/admin/segment.e2e.test.ts
```diff
@@ -234,57 +235,6 @@ test('should not delete segments used by strategies', async () => {
...
-        created_by: user.id,
-    });
-
-    expect((await fetchSegments()).length).toEqual(1);
-
-    await app.request.delete(`${SEGMENTS_BASE_PATH}/${segment.id}`).expect(409);
-
-    expect((await fetchSegments()).length).toEqual(1);
-});
-
```
thomasheartman, line 287: This was moved further down the file to colocate it with the other, related tests.

andreas-unleash (approved):
LGTM!

PR #5381, opened 2023-11-21
chore: avoid duplicates

thomasheartman:
This PR handles the case where a single strategy is used in multiple change requests. Instead of listing the strategy several times in the output, we consolidate the entries and add a new `changeRequestIds` property. This is a non-empty list that points to all the change requests it is used in. 

This is required for us to be able to link back to the change requests from the UI overview.


andreas-unleash (approved):
LGTM

PR #5383, opened 2023-11-21
fix: remove user from project

gastonfournier:
Removing a user from a project was impossible if you only had 1 owner. It worked fine when having more than an owner. This should fix it and we'll add tests later

### DIFF
--- a/src/lib/services/project-service.ts
+++ b/src/lib/services/project-service.ts
@@ -470,6 +470,9 @@ export default class ProjectService {
         );
     }
 
+    /**
+     * @deprecated use removeUserAccess
+     */
     async removeUser(
         projectId: string,
         roleId: number,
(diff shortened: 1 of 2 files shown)


kwasniew (approved):

ivarconr (changes requested):
1. Why is this method now deprecated? Why do we not use the "not deprecated" way?
2. I do not see any tests for the scenario being fixed.

gastonfournier:
> 1. Why is this method now deprecated? Why do we not use the "not deprecated" way?

The deprecated method was deprecated in enterprise it's not being used in our frontend but it's still maintained for backward compatibility. We are using the not deprecated method

> 2. I do not see any tests for the scenario being fixed.

No, we did manual testing because the issue seems to be urgent (is this assumption correct?). We're going to add tests later

ivarconr:
> No, we did manual testing because the issue seems to be urgent (is this assumption correct?). We're going to add tests later

I would argue we should add tests before upgrading all customers, at least before cutting a patch. It might be OK to roll it out to sandbox for parallel manual testing.

gastonfournier:
> > No, we did manual testing because the issue seems to be urgent (is this assumption correct?). We're going to add tests later
> 
> I would argue we should add tests before upgrading all customers, at least before cutting a patch. It might be OK to roll it out to sandbox for parallel manual testing.

https://github.com/Unleash/unleash/pull/5383/files#diff-d13508ab2fa1823f598990ee51788d32e10462c90891d4883fc2e105bb0bc103R1051-R1088 we have tests in place now ;)

PR #5385, opened 2023-11-21
test: add postgres 10 test

nunogois:
https://linear.app/unleash/issue/2-1589/add-a-test-to-validate-we-support-postgres-10

Adds an action that runs `yarn test:report` against a PostgreSQL 10 database, in order to validate that we are currently supporting it.


.github/workflows/build_prs_test_report_postgres10.yml
```diff
@@ -0,0 +1,47 @@
...
+  testreport-postgres10:
+    runs-on: ubuntu-latest
+    services:
+      # Label used to access the service container
+      postgres:
+        # Docker Hub image
+        image: postgres:10 # Specify PostgreSQL version 10
+        # Provide the password for postgres
+        env:
+          POSTGRES_PASSWORD: postgres10
```
gastonfournier, line 20: Can we use the matrix?

nunogois: Do you mean something like this? https://github.com/Unleash/unleash/pull/5385/commits/ab83c7b118ed8f6d063753f125f3561ceef49c65

gastonfournier: Actually, I thought about using the existing workflow

nunogois:
We are not merging this right now since this changes production code to satisfy the unique behavior of earlier PostgreSQL versions that we will be dropping support for in the future anyways. This PR validates that we in fact support these versions, even if the response results are displayed in a different order.

PR #5390, opened 2023-11-22
fix: if a strategy both uses a segment actively and in CRs, list it twice

thomasheartman:
This PR changes the behavior of the API a little bit. Instead of
removing any strategies from `changeRequestStrategies` that are also
in `strategies`, we keep them in instead.

The reason for this is that the overview of where a segment is used is
incomplete if it shows only strategies but not CRs. Imagine this:

You want to delete a segment, but you're told it's only used in strategy S.
So you go and remove it from strategy S, but then you're told it's
suddenly used in CRs A, B, and C. This is now a two-step operation
with a bad surprise. Instead, we could show you immediately that this
segment is used in strategy S and CRs A, B, and C.

### DIFF
--- a/src/lib/services/segment-service.ts
+++ b/src/lib/services/segment-service.ts
@@ -119,15 +119,10 @@ export class SegmentService implements ISegmentService {
         const strategies =
             await this.featureStrategiesStore.getStrategiesBySegment(id);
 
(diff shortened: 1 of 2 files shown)


andreas-unleash (approved):
Makes sense.LG

PR #5393, opened 2023-11-22
fix: check whether a usage data is defined

thomasheartman:
The previous check would return `false` if the value was 0, causing a
bug where the usage data wouldn't be included.

This also adds tests to ensure that usage data for CR segments is propagated correctly because that's where I first encountered the issue.

Before this fix, if the values were 0, the data would display like the bottom element in the screenshot:

image

### DIFF
--- a/src/test/e2e/api/admin/segment.e2e.test.ts
+++ b/src/test/e2e/api/admin/segment.e2e.test.ts
@@ -660,7 +660,8 @@ describe('detect strategy usage in change requests', () => {
                 .then((res) => res.body.segments);
 
         // because they use the same db, we can use the regular app
-        // (through createSegment) to create the segment and the flag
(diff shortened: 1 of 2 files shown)


src/test/e2e/api/admin/segment.e2e.test.ts
```diff
@@ -108,17 +108,19 @@ const validateSegment = (
...
+    const customOptions = {
+        experimental: {
+            flags: {
+                anonymiseEventLog: true,
+                detectSegmentUsageInChangeRequests: true,
             },
         },
+    };
+
+    db = await dbInit('segments_api_serial', getLogger, customOptions);
```
thomasheartman, line 120: If not made aware of the custom options, any flag resolution in the stores will fail.

andreas-unleash (approved):
LGTM

PR #5395, opened 2023-11-22
feat: support multiple terms in search, remove tag support in search

sjaanus:
1. Removing tag support in search
2. Adding multi keyword support for search


FredrikOseberg (approved):
LGTM

PR #5397, opened 2023-11-22
fix: dedupe any duplicate permissions

nunogois:
https://linear.app/unleash/issue/2-1656/create-db-migration-that-ensures-correct-state-of-permissions

This adds a migration that dedupes any duplicate permissions.

### DIFF
--- a/src/migrations/20231122121456-dedupe-any-duplicate-permissions.js
+++ b/src/migrations/20231122121456-dedupe-any-duplicate-permissions.js
@@ -35,7 +35,7 @@ DELETE FROM role_permission
 WHERE ctid IN (
     SELECT ctid
     FROM (
-        SELECT ctid, ROW_NUMBER() OVER (PARTITION BY role_id, permission_id ORDER BY (SELECT NULL)) as rn
+        SELECT ctid, ROW_NUMBER() OVER (PARTITION BY role_id, permission_id, environment ORDER BY created_at) as rn
         FROM role_permission
     ) t
     WHERE t.rn > 1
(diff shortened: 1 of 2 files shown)


gastonfournier (commented):
Did we test this somehow? I'd like to see if we can replicate the issue in a test

nunogois:
> Did we test this somehow? I'd like to see if we can replicate the issue in a test

Added an e2e test here: https://github.com/Unleash/unleash/pull/5397/commits/dc39925a0a58c99a80fffa96c1f917b18016a198
This was a great call, since the test helped iron out the script 👍

chriswk (approved):
With the test added, this must be one of our better tested migrations. Great job Nuno :)

src/migrations/20231122121456-dedupe-any-duplicate-permissions.js
```diff
@@ -0,0 +1,72 @@
...
+        COUNT(*) > 1
+)
+
+DELETE FROM permissions
+WHERE id NOT IN (
+    SELECT min_id FROM Duplicates
+)
+AND permission IN (
+    SELECT permission FROM Duplicates
+);
```
gastonfournier, line 64: Can we also apply a unique constraint to the name?

gastonfournier (changes requested):

src/migrations/20231122121456-dedupe-any-duplicate-permissions.js
nunogois: This is currently part of https://github.com/Unleash/unleash/pull/5398/files#diff-c979c34ca172be9fd07e04898e1256e9e305dd0384f74d888c862809413ed69aR4-R6
Do you think it belongs here instead, since this is the exact time we clean this up?

gastonfournier: No, I'm fine as long as it's considered! Unblocking

gastonfournier (approved):
LG!

PR #5398, opened 2023-11-22
fix: drop permissions id column

nunogois:
https://linear.app/unleash/issue/2-1660/create-db-migration-that-drops-the-id-column-for-permissions-and

Follow up to https://github.com/Unleash/unleash/pull/5397


gastonfournier (commented):
I believe this requires a major release cause it breaks our existing API:
image

One alternative could be to have an in-memory map converting from ids to names... but I don't quite like it

nunogois:
Will close this in favor of https://github.com/Unleash/unleash/pull/5409.

PR #5400, opened 2023-11-22
feat: allow filtering projects with operators

sjaanus:
This is first iteration. When we add more fields to be filterable with operators, we can have more reusable components for this.

### DIFF
--- a/src/lib/features/feature-search/feature-search-service.ts
+++ b/src/lib/features/feature-search/feature-search-service.ts
@@ -5,7 +5,11 @@ import {
     IUnleashStores,
     serializeDates,
 } from '../../types';
-import { IFeatureSearchParams } from '../feature-toggle/types/feature-toggle-strategies-store-type';
+import {
+    IFeatureSearchParams,
+    IQueryOperator,
+    IQueryParam,
+} from '../feature-toggle/types/feature-toggle-strategies-store-type';
 
 export class FeatureSearchService {
     private featureStrategiesStore: IFeatureStrategiesStore;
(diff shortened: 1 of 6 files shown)


src/lib/features/feature-search/feature-search-service.ts
```diff
@@ -21,15 +25,48 @@ export class FeatureSearchService {
...
-            await this.featureStrategiesStore.searchFeatures({
-                ...params,
-                limit: params.limit,
-            });
+            await this.featureStrategiesStore.searchFeatures(
+                {
+                    ...params,
+                    limit: params.limit,
+                },
+                queryParams,
```
sjaanus, line 35: Currently expanding with new parameter, but eventually the top params will be removed.

src/lib/features/feature-search/feature-search-service.ts
```diff
@@ -21,15 +25,48 @@ export class FeatureSearchService {
...
+            );
 
         return {
             features,
             total,
         };
     }
+
+    parseOperatorValue = (field: string, value: string): IQueryParam | null => {
```
sjaanus, line 44: Currently supporting just 4 operators.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -1033,5 +1036,27 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
...
+            case 'IS':
+                query.where(param.field, '=', param.value);
+                break;
+            case 'IS_NOT':
+                query.where(param.field, '!=', param.value);
+                break;
+            case 'IS_ANY_OF':
+                query.whereIn(param.field, param.value as string[]);
+                break;
+            case 'IS_NOT_ANY_OF':
```
Tymek, line 1054: IS_NOT_IN would sound more natural to me. Or even `NOT_IN`. I think we should keep  domain language close to SQL. Same applies to the UI. What do you think @nicolaesocaciu ?

sjaanus: I think we following language based on UI elements and those were taken 1 to 1 from linear. Maybe they have thought about it more than us.

sjaanus: Either way we go, I think frontend should have same texts.

nicolaesocaciu: I understand both expressions but for me it's a bit more clear when is see **"Is not any of"** compared with **"Is not in / Not in"**

I will ask also the team what they think, and we will have at some point some user testing with the filters and i will try to see what is the perception about these expressions

Tymek (approved):

PR #5403, opened 2023-11-23
feat: Handle scheduled request events in addons

andreas-unleash:
- Create 2 new events to replace the SCHEDULED_CHANGE_REQUEST_EXECUTED event
- Handle the 3 events in slack-app and webhook addon definitions

3 events handled:
- CHANGE_REQUEST_SCHEDULED
- CHANGE_REQUEST_SCHEDULED_APPLICATION_SUCCESS
- CHANGE_REQUEST_SCHEDULED_APPLICATION_FAILURE

Closes # [1-1555](https://linear.app/unleash/issue/1-1555/update-change-request-scheduled-and-scheduled-change-request-executed)

Note: SCHEDULED_CHANGE_REQUEST_EXECUTED will be removed in follow up PR not to break current enterprise build


src/lib/addons/feature-event-formatter-md.ts
```diff
@@ -140,6 +142,18 @@ const EVENT_MAP: Record<string, IEventData> = {
         action: '*{{user}}* sent to review change request {{changeRequest}}',
         path: '/projects/{{event.project}}/change-requests/{{event.data.changeRequestId}}',
     },
+    [CHANGE_REQUEST_SCHEDULED]: {
+        action: '*{{user}}* scheduled change request {{changeRequest}} in project *{{event.project}}*',
```
thomasheartman, line 146: Can we add when it was scheduled for? Something like this?
```suggestion
        action: '*{{user}}* scheduled change request {{changeRequest}} to be applied at {{time}} in project *{{event.project}}*',
```

src/lib/addons/feature-event-formatter-md.ts
```diff
@@ -140,6 +142,18 @@ const EVENT_MAP: Record<string, IEventData> = {
...
+        path: '/projects/{{event.project}}/change-requests/{{event.data.changeRequestId}}',
+    },
+    [CHANGE_REQUEST_SCHEDULED_APPLICATION_SUCCESS]: {
+        action: '*{{user}}* successfully applied scheduled change request {{changeRequest}} in project *{{event.project}}*',
```
thomasheartman, line 150: Who is the `user` in this case? The same as the one that's usually used for CR applications?

And if we're changing the wording of the failed event, we could also change this one a bit:

```suggestion
        action: 'Successfully applied the scheduled change request {{changeRequest}} by *{{user}}* in project *{{project}}*.',
```

src/lib/addons/feature-event-formatter-md.ts
```diff
@@ -140,6 +142,18 @@ const EVENT_MAP: Record<string, IEventData> = {
...
+        path: '/projects/{{event.project}}/change-requests/{{event.data.changeRequestId}}',
+    },
+    [CHANGE_REQUEST_SENT_TO_REVIEW]: {
+        action: '*{{user}}* failed to apply scheduled change request {{changeRequest}} in project *{{event.project}}*',
```
thomasheartman, line 154: Because this is an urgent event, maybe we should put the`failed` bit first? Not very important, just a suggestion.


```suggestion
        action: '*Failed* to apply the scheduled change request {{changeRequest}} by *{{user}}* in project *{{project}}*.',
```

thomasheartman (approved):
Looks good! A few minor suggestions that we could use to approve this, but nothing that needs to go in right now

src/lib/addons/feature-event-formatter-md.ts
andreas-unleash: yes, the 'applier'.

PR #5406, opened 2023-11-23
Chore: add strategy sorting algorithm

thomasheartman:
This PR adds a strategy sorting algorithm to be used for the segment deletion dialog. It assumes that you have a list of existing strategies and a list of change request strategies. Based on the content of these two lists, it will create one unified list sorted after a number of criteria (as listed in the test).

# Discussion point: 

This impl does the sorting on the front end, but could we do it on the back end? Instead of adding a new property to the segment data, could we simply fold the change request strategies in with the existing segment strategies and return it using the old property? If the only place we do that is in this view, then that might be a good suggestion.

### DIFF
--- a/frontend/src/component/segments/SegmentDelete/SegmentDeleteUsedSegment/sort-strategies.test.ts
+++ b/frontend/src/component/segments/SegmentDelete/SegmentDeleteUsedSegment/sort-strategies.test.ts
@@ -1,0 +1,137 @@
+import { sortStrategiesByFeature } from './sort-strategies';
+
+describe('sorting strategies by feature', () => {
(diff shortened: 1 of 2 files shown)


andreas-unleash (approved):
LG. It would also make sense to do this in the BE as well. Up to you

thomasheartman:
I'll leave this in the front end for now. The reason is that we can't add change request strategies to the existing `strategies` property of the API payload without it being a breaking change. The OpenAPI schema says that `id` is a required field on a strategy, and that field doesn't exist on strategies that have only been added in change requests, but not yet applied.

Some options would be to add a new endpoint or use a new property `strategies including change requests` that has everything pre-sorted. But I think this is a good way to go for now.

PR #5408, opened 2023-11-23
feat: audit roles

gastonfournier:
## About the changes
Audit changes to roles both root and project roles. The tests only cover one case for the creation.

Created by unknown because we haven't modified the enterprise repository

Deleting a project role displays the permissions and environments that are being deleted
Screenshot from 2023-11-24 10-29-57

Creating and modifying project roles
Screenshot from 2023-11-24 10-29-46

Creating, modifying, and deleting root roles
Screenshot from 2023-11-24 10-29-28


src/lib/services/access-service.ts
```diff
@@ -62,6 +70,7 @@ interface IRoleCreation {
     description: string;
     type?: 'root-custom' | 'custom';
     permissions?: PermissionRef[];
+    createdBy?: string;
```
gastonfournier, line 73: This is optional for backward compatibility with existing enterprise version

src/lib/services/access-service.ts
```diff
@@ -76,6 +85,7 @@ export interface IRoleUpdate {
...
```
gastonfournier, line 88: This is optional for backward compatibility with existing enterprise version

src/lib/services/access-service.ts
```diff
@@ -95,34 +105,30 @@ export class AccessService {
...
         }: Pick<
             IUnleashStores,
-            | 'accessStore'
-            | 'accountStore'
-            | 'roleStore'
-            | 'environmentStore'
-            | 'groupStore'
-        >,
+            'accessStore' | 'accountStore' | 'roleStore' | 'environmentStore'
+        > & { groupStore?: any }, // TODO remove groupStore later, kept for backward compatibility with enterprise
```
gastonfournier, line 125: boy scouting: removing groupStore in favor of groupService to depend only in one of the two

src/lib/services/access-service.ts
```diff
@@ -643,6 +649,17 @@ export class AccessService {
                 );
             }
         }
+        const addedPermissions = await this.store.getPermissionsForRole(
+            newRole.id,
+        );
```
gastonfournier, line 654: Currently, permissions can come in any shape from the UI (i.e. only send ids) and we'll adapt, but we just want the names in the event log. This also applies to other parts of this PR

ivarconr:
Can you please include a few examples on how the audit events will look like?

gastonfournier:
> Can you please include a few examples on how the audit events will look like?

Yes, it's just reporting unknown user because that has to change in enterprise repository. I was planning to add tests to validate the format.

Screenshot from 2023-11-24 10-29-57
Screenshot from 2023-11-24 10-29-46
Screenshot from 2023-11-24 10-29-28

src/lib/features/access/createAccessService.ts
```diff
@@ -65,9 +67,15 @@ export const createFakeAccessService = (
...
+    const accessService = new AccessService(
         { accessStore, accountStore, roleStore, environmentStore, groupStore },
         { getLogger, flagResolver },
         groupService,
+        eventService,
     );
+
+    return {
+        accessService,
+        eventStore,
```
kwasniew, line 79: why do you return eventStore that is private to this accessService?

src/lib/services/email-service.ts
```diff
@@ -101,7 +101,7 @@ export class EmailService {
                 text: bodyText,
             };
             process.nextTick(() => {
-                this.mailer.sendMail(email).then(
+                this.mailer!.sendMail(email).then(
```
kwasniew, line 104: can we solve it without "!"  ?

src/lib/services/access-service.ts
```diff
@@ -662,32 +679,63 @@ export class AccessService {
...
+            const sanitizedEnvironment =
+                environment && environment !== null && environment !== ''
+                    ? environment
+                    : undefined;
+            return { name, environment: sanitizedEnvironment };
+        });
     }
 
-    async deleteRole(id: number): Promise<void> {
+    async deleteRole(id: number, deletedBy = 'unknown'): Promise<void> {
```
kwasniew, line 738: in the enterprise repo we should me this transactional so that we guarantee event storage

src/lib/services/access-service.ts
```diff
@@ -699,7 +747,18 @@ export class AccessService {
...
-        return this.roleStore.delete(id);
+        const existingRole = await this.roleStore.get(id);
+        const existingPermissions = await this.store.getPermissionsForRole(id);
+        await this.roleStore.delete(id);
+        this.eventService.storeEvent({
+            type: ROLE_DELETED,
+            createdBy: deletedBy,
+            preData: {
+                ...existingRole,
+                permissions: this.sanitizePermissions(existingPermissions),
```
kwasniew, line 758: how do we decide when to add extra data. This event duplicates the information from ROLE_CREATED. If we wanted to have "normalized" event we'd only provide minimum necessary information. In this case this is the role being deleted. Is it for convenience that we show the same data in ROLE_CREATED and ROLE_DELETED?

src/lib/services/access-service.ts
```diff
@@ -662,32 +679,63 @@ export class AccessService {
...
+            },
+            preData: {
+                ...existingRole,
+                permissions: this.sanitizePermissions(existingPermissions),
+            },
+        });
+        return updatedRole;
+    }
+
+    sanitizePermissions(
```
kwasniew, line 726: where do we test sanitization logic?

src/lib/services/access-service.ts
```diff
@@ -643,6 +649,17 @@ export class AccessService {
                 );
             }
         }
+        const addedPermissions = await this.store.getPermissionsForRole(
+            newRole.id,
+        );
+        this.eventService.storeEvent({
+            type: ROLE_CREATED,
+            createdBy: role.createdBy || 'unknown',
```
kwasniew, line 657: I see potential for error when we have the same default logic in many places.

src/lib/services/email-service.ts
gastonfournier: I think not, because Typescript does not know we're validating that mailer is not undefined inside `if (this.configured()) {`

kwasniew (approved):

src/lib/services/access-service.ts
gastonfournier: Not sure, because data may have changed, we can trace back what the permissions were by the time of deleting this, but for that matter, we would only need the id when we delete any resource, and we're not doing that currently.

src/lib/services/access-service.ts
gastonfournier: I'll do that as part of passing down the author of the action

src/lib/services/access-service.ts
gastonfournier: https://github.com/Unleash/unleash/pull/5408/files/c1cdc0efb24a500f14e5b64d45f906369e11e818#diff-a6135f07806e2e7dbc233652abfbd18fe122690638252d488eb6f82ae8f02e10R198-R212 covers the sanitization logic. But can have specific UT

src/lib/services/access-service.ts
gastonfournier: Yes, but this logic should go away once we fix enterprise repo

src/lib/services/email-service.ts
kwasniew: ah because it's a class field and can be set to undefined any time

gastonfournier: Do you have any idea how to solve this? I think we can't make it a constant :thinking: can't we?

kwasniew: I tried a few things but I think this is ok for now

gastonfournier: Likewise :P

PR #5409, opened 2023-11-23
refactor: favor permission name over id

nunogois:
https://linear.app/unleash/issue/2-1664/create-db-migration-that-favors-the-name-column-over-id-for

Similar to https://github.com/Unleash/unleash/pull/5398, but non-breaking (semver).
This keeps the permissions `id` column intact, however favors the permission name whenever possible.

### DIFF
--- a/src/migrations/20231123155649-favor-permission-name-over-id.js
+++ b/src/migrations/20231123155649-favor-permission-name-over-id.js
@@ -1,0 +1,253 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+-- STEP 1: Ensure 'permission' column contains unique values
+ALTER TABLE permissions
+ADD CONSTRAINT permission_unique UNIQUE (permission);
+
+-- STEP 2: Add a new column for the new foreign key
+ALTER TABLE role_permission
+ADD COLUMN permission text;
+
+-- STEP 3: Populate the new column with corresponding values from 'permissions'
+UPDATE role_permission rp
+SET permission = p.permission
+FROM permissions p
+WHERE rp.permission_id = p.id;
+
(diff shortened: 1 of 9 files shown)


src/migrations/20231123155649-favor-permission-name-over-id.js
```diff
@@ -0,0 +1,253 @@
...
+ALTER TABLE permissions
+DROP CONSTRAINT permissions_pkey;
+
+-- STEP 5: Add the 'permission' primary key constraint
+ALTER TABLE permissions
+ADD PRIMARY KEY (permission);
+
+-- STEP 6: Drop the old 'permission_id' foreign key column
+ALTER TABLE role_permission
+DROP COLUMN permission_id;
```
gastonfournier, line 28: Don't do this in the same release, this breaks the expand and contract pattern: https://openpracticelibrary.com/practice/expand-and-contract-pattern/ 

I believe we need to wait until the next minor release to run this query.

gastonfournier (changes requested):

src/migrations/20231123155649-favor-permission-name-over-id.js
nunogois: Good catch, addressed in https://github.com/Unleash/unleash/pull/5409/commits/14c1cd9476366430cb1da62e970113d97afe615e and moved to another PR: https://github.com/Unleash/unleash/pull/5413.

gastonfournier (approved):
This type of migrations, always makes me think twice, I read through it and despite being risky, I can't think of a scenario where this would fail for the previous version still running with the previous code because we're just dropping constraints. I think this is good and I'm happy to see this happening :rocket:

PR #5410, opened 2023-11-24
Feat: show change request data on segment project usage page

thomasheartman:
Show usage in change requests if that'd cause you to not be able to move the segment into a project.

- [x] ~Relies on changes from #5407 (and #5405, #5406) to go through first.~

image


frontend/src/component/segments/SegmentProjectAlert.tsx
```diff
@@ -100,10 +95,73 @@ export const SegmentProjectAlert = ({
...
+                    )}
+                    target='_blank'
+                    rel='noopener noreferrer'
+                >
+                    {strategy.featureName!} {formatStrategyNameParens(strategy)}
+                </Link>
+            </li>
+        );
+    }
+};
```
thomasheartman, line 167: This is nearly identical to what we did in the segment delete component, except it uses `Link` instead of `StyledLink`. I wonder if it's worth extracting it or not. It feels weird duplicating it, but at the same time, they **are** two different lists that might not necessarily change together.

andreas-unleash (approved):
Nice! LGTM

PR #5412, opened 2023-11-24
fix: prevent concurrent queries from running out of transaction

kwasniew:
## About the changes

JS promises support Promise.all and Promise.allSettled. Promise.all rejects on the first rejecting promise. Promise.allSettled waits for all promises to settle and returns an array of resolved and rejected promises. I'm adding a new utility that behaves like Promise.all (rejects on the first rejection) but gives all promises a chance to finish. 

This new utility solves an issue where some of the promises were running after a transaction was closed because first rejected promise rejected much sooner than the other ones, the tx was closed and the remaining queries were left w/o a pending tx.

Screenshot 2023-11-24 at 10 41 02

### DIFF
--- a/src/lib/features/export-import-toggles/export-import-service.ts
+++ b/src/lib/features/export-import-toggles/export-import-service.ts
@@ -52,6 +52,7 @@ import { FeatureNameCheckResultWithFeaturePattern } from '../feature-toggle/feat
 import { IDependentFeaturesReadModel } from '../dependent-features/dependent-features-read-model-type';
 import groupBy from 'lodash.groupby';
 import { ISegmentService } from '../../segments/segment-service-interface';
+import { allSettledWithRejection } from '../../util/allSettledWithRejection';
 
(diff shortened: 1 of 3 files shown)


src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -252,7 +253,7 @@ export default class ExportImportService
         user: IUser,
         mode = 'regular' as Mode,
     ): Promise<void> {
-        await Promise.all([
+        await allSettledWithRejection([
             this.verifyStrategies(dto),
             this.verifyContextFields(dto),
             this.importPermissionsService.verifyPermissions(dto, user, mode),
```
kwasniew, line 259: verify permissions was always running outside of the current transaction when segments checked failed. Segment check was one simple DB query whole permission check involved multiple queries.

thomasheartman (approved):
Huh, this is a neat trick!

PR #5415, opened 2023-11-24
feat: feature type updated audit log

kwasniew:
## About the changes

Added audit for feature type update

Short version:
Screenshot 2023-11-24 at 13 56 35

Full event:
Screenshot 2023-11-24 at 14 05 24

Design decisions:
* FEATURE_TYPE_UPDATED event added instead of more granular LIFETIME_UPDATED. We may have other changes soon e.g. name change.
* added getRecordedEvents helper method to test app to easily check if the correct event got recorded
* added extra query in a service to get the preData status of the event type
* I decided to add tx support and refactor towards feature oriented architecture in a separate PR

### DIFF
--- a/src/lib/services/feature-type-service.ts
+++ b/src/lib/services/feature-type-service.ts
@@ -6,18 +6,25 @@ import {
     IFeatureTypeStore,
 } from '../types/stores/feature-type-store';
 import NotFoundError from '../error/notfound-error';
+import EventService from './event-service';
+import { FEATURE_FAVORITED, FEATURE_TYPE_UPDATED, IUser } from '../types';
+import { extractUsernameFromUser } from '../util';
 
 export default class FeatureTypeService {
     private featureTypeStore: IFeatureTypeStore;
+
+    private eventService: EventService;
 
     private logger: Logger;
 
     constructor(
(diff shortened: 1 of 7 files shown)


gastonfournier (approved):
LGTM!

src/lib/services/feature-type-service.ts
```diff
@@ -27,23 +34,33 @@ export default class FeatureTypeService {
...
             id,
             translatedLifetime,
         );
 
-        if (!result) {
+        if (!featureType || !result) {
             throw new NotFoundError(
                 `The feature type you tried to update ("${id}") does not exist.`,
             );
         }
```
thomasheartman, line 55: If this is the only error case here, would it make sense to move it to before trying to update the lifetime and only check whether it exists? 


```suggestion
        if (!featureType) {
            throw new NotFoundError(
                `The feature type you tried to update ("${id}") does not exist.`,
            );
        }
```

thomasheartman (approved):
Nice one! 🙌🏼

src/lib/services/feature-type-service.ts
kwasniew: tried that originally but TS led me towards my current solution. You can try for yourself. If you add your line and move it before update the return statement will not compile

ivarconr:
nice!

PR #5416, opened 2023-11-24
fix: enable toggle with disabled strategies dialog

nunogois:
https://linear.app/unleash/issue/UNL-283/improve-dialog-when-there-are-disabled-strategies-and-we-want-to

Improves the "enable toggle when there are disabled strategies" dialog. Also implements passing in the missing `disabledStrategiesCount` property to the dialog.

image


nicolaesocaciu (approved):
This looks super !

PR #5418, opened 2023-11-24
fix: store get type should reflect nullability

kwasniew:
## About the changes

Closes #


src/lib/types/stores/store.ts
```diff
@@ -1,5 +1,5 @@
 export interface Store<T, K> {
-    get(key: K): Promise<T>;
+    get(key: K): Promise<T | null>;
```
ivarconr, line 2: Some stores will today throw not found exception. I think changing would require us to a lot more changes than just this type fix.

PR #5419, opened 2023-11-27
chore: remove dependent feature flags

kwasniew:
## About the changes

Removing dependent features flag from:
* backend code
* UI code
* tests (flags + API calls)


frontend/src/component/common/FeatureArchiveDialog/FeatureArchiveDialog.test.tsx
```diff
@@ -126,6 +123,7 @@ test('Skip change request', async () => {
     await screen.findByText('Archive feature toggles');
     const button = await screen.findByText('Archive toggles');
 
+    await waitFor(() => expect(button).toBeEnabled());
```
kwasniew, line 126: previously we were accidentally waiting on the async flag resolution. After the flag is gone we need to explicitly wait for the button to become enabled

sjaanus (approved):
LG, seems straightforward.

PR #5421, opened 2023-11-27
chore(update): Updates biome config to 1.3 format

chriswk:
Does what it says on the tin, update to 1.3.x format for biome config


gastonfournier (approved):
G!

PR #5431, opened 2023-11-27
[wip]: move enterprise check further left

thomasheartman:


src/lib/features/segment/segment-controller.ts
thomasheartman, line 1: This is just a refactor to remove duplication.

If we change the payload to always include `changeRequestStrategies`, we can remove the flag check entirely.

src/lib/services/segment-service.ts
thomasheartman, line 1: Only look up change request strategies if it's relevant. Meaning: only if the flag is enabled (for now) and if we're on an enterprise instance.

andreas-unleash (approved):
Nice!

PR #5433, opened 2023-11-27
refactor: isolate tests

FredrikOseberg:
This PR fixes a race condition between e2e tests where bulk archiving all toggles in the default project would delete toggles used for the features e2e tests.

It does by isolating the features.spec and overview.spec to their respective projects, so that they always operate on isolated data.

### Future enhancements: 

I'm not particularly fond of passing the projectName through to all the helper methods. It complicates the tests more than it should. I would like to be able to set the project once per test and have all the helper methods be aware of the context. Something like this should work: 

```
before(() => {
   cy.wrap('projectId').as('project');
})
```

And in the helpers: 

```
export const createFeature_API = (
    featureName: string,
    options?: Partial<Cypress.RequestOptions>,
): Chainable<any> => {
    return cy.get('@project').then((project) => {
        projectName = project || 'default';
        return cy.request({
            url: `${baseUrl}/api/admin/projects/${projectName}/features`,
            method: 'POST',
            body: {
                name: `${featureName}`,
                description: 'hello-world',
                type: 'release',
                impressionData: false,
            },
            ...options,
        });
    });
};
```

### DIFF
--- a/frontend/cypress/global.d.ts
+++ b/frontend/cypress/global.d.ts
@@ -59,7 +59,10 @@ declare namespace Cypress {
         addFlexibleRolloutStrategyToFeature_UI(
             options: AddFlexibleRolloutStrategyOptions,
         ): Chainable;
-        updateFlexibleRolloutStrategy_UI(featureToggleName: string);
+        updateFlexibleRolloutStrategy_UI(
+            featureToggleName: string,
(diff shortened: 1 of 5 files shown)


kwasniew (approved):
Thank you!

PR #5437, opened 2023-11-27
fix: tag type deleted should indicate data deletion

kwasniew:
## About the changes

Before - green indicator for tag name:
Screenshot 2023-11-27 at 14 38 48

After - red indicator for all fields:
Screenshot 2023-11-27 at 14 38 09

## Discussion points

When the new code tries to delete non existent tag it returns 404 and it should be fine since idempotency requirement applies to resource state not the HTTP code: https://stackoverflow.com/questions/24713945/does-idempotency-include-response-codes/24713946#24713946


chriswk (approved):
Nice!

PR #5440, opened 2023-11-27
feat: include more data in setting event

gardleopard:
This adds more data to the setting events, so that its possible to see what has changed

Used to look like:
```
{
  "id": "maintenance.mode"
}
```

Now it looks like this:
```
{
  "id": "maintenance.mode",
  "enabled": false
}
```

### DIFF
--- a/src/lib/services/setting-service.ts
+++ b/src/lib/services/setting-service.ts
@@ -44,7 +44,7 @@ export default class SettingService {
 
     async insert(id: string, value: object, createdBy: string): Promise<void> {
         const exists = await this.settingStore.exists(id);
-        let data = value;
+        let data = { id };
         if (typeof value === 'object') {
             data = { id, ...value };
         } else {
--- a/src/lib/services/setting-service.ts
+++ b/src/lib/services/setting-service.ts
@@ -43,7 +43,8 @@ export default class SettingService {
     }
 
     async insert(id: string, value: object, createdBy: string): Promise<void> {
-        const exists = await this.settingStore.exists(id);
+        const existingSettings = await this.settingStore.get<object>(id);
+
         let data = { id };
         if (typeof value === 'object') {
             data = { id, ...value };
@@ -51,13 +52,16 @@ export default class SettingService {
             this.logger.error('Invalid setting value, source: %s', id);
         }
 
-        if (exists) {
+        if (existingSettings) {
             await this.settingStore.updateRow(id, value);
             await this.eventService.storeEvent(
-                new SettingUpdatedEvent({
-                    createdBy,
-                    data,
-                }),
+                new SettingUpdatedEvent(
+                    {
+                        createdBy,
+                        data,
+                    },
+                    existingSettings,
+                ),
             );
         } else {
             await this.settingStore.insert(id, value);
--- a/src/lib/services/setting-service.ts
+++ b/src/lib/services/setting-service.ts
@@ -44,12 +44,19 @@ export default class SettingService {
 
     async insert(id: string, value: object, createdBy: string): Promise<void> {
         const exists = await this.settingStore.exists(id);
+        let data = value;
+        if (typeof value === 'object') {
+            data = { id, ...value };
+        } else {
+            this.logger.error('Invalid setting value, source: %s', id);
+        }
+
         if (exists) {
             await this.settingStore.updateRow(id, value);
             await this.eventService.storeEvent(
                 new SettingUpdatedEvent({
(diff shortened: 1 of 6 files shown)


src/lib/services/setting-service.ts
```diff
@@ -49,15 +49,15 @@ export default class SettingService {
             await this.eventService.storeEvent(
                 new SettingUpdatedEvent({
                     createdBy,
-                    data: { id },
+                    data: { id, ...value },
```
FredrikOseberg, line 52: Are we guaranteed that value is an object? Can it ever be undefined?

src/lib/services/setting-service.ts
```diff
@@ -49,15 +49,15 @@ export default class SettingService {
...
                 }),
             );
         } else {
             await this.settingStore.insert(id, value);
             await this.eventService.storeEvent(
                 new SettingCreatedEvent({
                     createdBy,
-                    data: { id },
+                    data: { id, ...value },
```
FredrikOseberg, line 60: Same question as above.

sighphyre (approved):
G

FredrikOseberg (approved):
LG, just need to address PR comments and Biome.

src/lib/services/setting-service.ts
gardleopard: The message signature is `async insert(id: string, value: object, createdBy: string):` afaik that only accepts objects?

gardleopard:
@FredrikOseberg in the last commit I have added logging if the data sent in is not of valid type and fallback to old event format in that case.

src/lib/services/setting-service.ts
```diff
@@ -44,20 +44,27 @@ export default class SettingService {
 
     async insert(id: string, value: object, createdBy: string): Promise<void> {
         const exists = await this.settingStore.exists(id);
+        let data = value;
+        if (typeof value === 'object') {
+            data = { id, ...value };
+        } else {
+            this.logger.error('Invalid setting value, source: %s', id);
+        }
```
FredrikOseberg, line 52: I think this can cause subtle bugs. If the incoming data is a string for example we won't add the data in the event. I think I would either:

(A) Check all incoming paths to this code to verify that we are indeed validating that it is an object. If it is, then what we have done here is fine.
(B) If we are not sure whether or not it's an object I'd prefer something like: 

```
 if (typeof value === 'object') {
            data = { id, ...value };
        } else {
            data = { id, value } // this was the original right?
        }
```

FredrikOseberg (changes requested):

src/lib/services/setting-service.ts
gardleopard: The original was just id

gardleopard: but you are right, the code is wrong

gardleopard: patched it now

FredrikOseberg (approved):
LGTM

src/lib/services/setting-service.ts
```diff
@@ -44,20 +44,27 @@ export default class SettingService {
...
+
         if (exists) {
             await this.settingStore.updateRow(id, value);
             await this.eventService.storeEvent(
                 new SettingUpdatedEvent({
                     createdBy,
-                    data: { id },
+                    data,
```
ivarconr, line 59: I think we are missing preData here.

src/lib/services/setting-service.ts
```diff
@@ -44,20 +44,27 @@ export default class SettingService {
...
+        if (typeof value === 'object') {
+            data = { id, ...value };
+        } else {
+            this.logger.error('Invalid setting value, source: %s', id);
+        }
+
         if (exists) {
             await this.settingStore.updateRow(id, value);
             await this.eventService.storeEvent(
                 new SettingUpdatedEvent({
```
ivarconr, line 57: I have sine doubts about this generic settings-updated events. Some things we store in the settings table should not be populated to the audit events, e.g. when we store SAML details, we want to avoid printing the private key in the audit logs.

sighphyre (approved):
G

src/lib/services/setting-service.ts
gardleopard: This has now been fixed, nice catch!

src/lib/services/setting-service.ts
gardleopard: You are correct! Nice catch, fixed now.

PR #5443, opened 2023-11-27
fix: naming pattern update event should keep old project settings

kwasniew:
## About the changes

Before - all untouched settings were shown as deleted
Screenshot 2023-11-27 at 16 19 37

After - only changing fields are shown
Screenshot 2023-11-27 at 16 19 45


FredrikOseberg (approved):
LGTM

PR #5445, opened 2023-11-27
feat: report users on group change

gastonfournier:
## About the changes
Add user ids to group changes. This also modifies the payload of group created to include only the user id:

image

Add users:
image

Remove a user:
image

Delete group:
image

Group creation event changes from:
image

To:
image


src/lib/services/group-service.ts
```diff
@@ -255,6 +262,31 @@ export class GroupService {
                 externalGroups,
             );
             await this.groupStore.deleteUsersFromGroup(oldGroups);
+
+            const events: IBaseEvent[] = [];
+            for (const group of newGroups) {
+                events.push({
+                    type: GROUP_USER_ADDED,
+                    createdBy: createdBy ?? 'unknown',
```
ivarconr, line 270: we should standardize this. I think we should say "system" instead of unknown. You can look at the USER_CREATED event for a more standard way to deal with it.

gastonfournier: I know that in some places we use `system` but that's when we know it's something done by Unleash. When it's something done through our API I believe we fallback to `unknown` (but there might be other cases): https://github.com/Unleash/unleash/blob/0c246dfbf536537db9ec95aa22870a7b13c62efb/src/lib/util/extract-user.ts#L4

I think we should standardize and we have a project to make audit better, but in this case I believe the type should always be defined, but changing the type across the two repos and OpenAPI specs was not easy, so I opted for this solution

PR #5448, opened 2023-11-27
fix: specify permission table

nunogois:
https://linear.app/unleash/issue/2-1672/hotfix-specify-table-in-permission-column-queries

This specifies the table that we are querying the `permission` column from, hopefully preventing ambiguity issues.
You can find the relevant changes here: [#5409](https://github.com/Unleash/unleash/pull/5409/files#diff-5cb4eca71e6a03ec3d123986d657368a3386f80e986dfef2e3feb0abb0673d87)

### DIFF
--- a/src/lib/db/access-store.ts
+++ b/src/lib/db/access-store.ts
@@ -204,7 +204,7 @@ export class AccessStore implements IAccessStore {
         let userPermissionQuery = this.db
             .select(
                 'project',
-                'permission',
+                'p.permission',
                 'environment',
                 'type',
                 'ur.role_id',
@@ -217,7 +217,7 @@ export class AccessStore implements IAccessStore {
         userPermissionQuery = userPermissionQuery.union((db) => {
             db.select(
                 'project',
-                'permission',
+                'p.permission',
                 'environment',
                 'p.type',
                 'gr.role_id',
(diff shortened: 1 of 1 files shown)


src/lib/db/access-store.ts
```diff
@@ -204,7 +204,7 @@ export class AccessStore implements IAccessStore {
         let userPermissionQuery = this.db
             .select(
                 'project',
-                'permission',
+                'p.permission',
```
gastonfournier, line 207: Should we proactively do the same for other columns?

src/lib/db/access-store.ts
```diff
@@ -217,7 +217,7 @@ export class AccessStore implements IAccessStore {
         userPermissionQuery = userPermissionQuery.union((db) => {
             db.select(
                 'project',
-                'permission',
+                'p.permission',
```
gastonfournier, line 220: I guess in 5.6 role_permission table does not have the permission column, right? That's why you chose `p.permission`  instead of `rp.permission`

gastonfournier (approved):
LGTM, I feel tempted to do the same for other select fields because this prevents the issue in 1 case but not in other cases. Anyway, this will be an action from the PM, so :+1: to patch this ASAP

src/lib/db/access-store.ts
nunogois: Correct 👍 The reason why I opted for `rp` in [#5409](https://github.com/Unleash/unleash/pull/5409) was because we may want to eventually drop the join. But since `p.permission` exists in both *before* and *after* the migration, it would be the best option.

src/lib/db/access-store.ts
nunogois: I agree that we can fix this ASAP and then try to address this later as a PM action.

PR #5462, opened 2023-11-28
feat: updated default strategy event

kwasniew:
## About the changes

Capturing default strategy updated events
Screenshot 2023-11-28 at 13 14 00


andreas-unleash (approved):
LGTM

PR #5469, opened 2023-11-28
fix(removeapitokenbutton.tsx): add environment name in delete token modal

### LINKED ISSUE #5444: Include environment in token info for "Confirm deletion"

As you may create different tokens for different environment its good idea to include token environment here:
image

### Background

_No response_

### Solution suggestions

_No response_

srcgrp:
fix #5444

## About the changes

I have added an "environment" row to the file RemoveApiTokenButton.tsx to prompt users to confirm their intention before deleting a token. This enhancement aims to prevent accidental deletions and improve user experience.

Closes #5444 

### Important files

The key file modified in this PR is RemoveApiTokenButton.tsx Reviewers should pay special attention to this file to understand the implementation of the "environment" row.

## Discussion points

I would appreciate feedback on the implementation of the "environment" row and its impact on user interactions. Additionally, if there are alternative approaches or considerations that should be taken into account, please let me know.


FredrikOseberg (approved):
LGTM. Thank you!

PR #5470, opened 2023-11-28
Default strategy event user

kwasniew:
## About the changes
Real user for default strategy event


gardleopard (approved):
Looks good to me:)

PR #5478, opened 2023-11-28
fix: add metrics for service account and api tokens

ivarconr:
This PR adds Prometheus metrics for the number of "service accounts" and "api tokens". 

Service Accounts:
image

Api tokens (label for type):
image

### DIFF
--- a/src/lib/db/api-token-store.ts
+++ b/src/lib/db/api-token-store.ts
@@ -96,10 +96,24 @@ export class ApiTokenStore implements IApiTokenStore {
             });
     }
 
-    count(): Promise<number> {
+    async count(): Promise<number> {
         return this.db(TABLE)
             .count('*')
             .then((res) => Number(res[0].count));
+    }
+
+    async countByType(): Promise<Map<string, number>> {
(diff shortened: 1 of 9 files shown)


gastonfournier (approved):
LGTM!

src/lib/db/api-token-store.ts
```diff
@@ -96,12 +96,26 @@ export class ApiTokenStore implements IApiTokenStore {
...
             .count('*')
             .then((res) => Number(res[0].count));
     }
 
+    async countByType(): Promise<Map<string, number>> {
+        return this.db(TABLE)
+            .select('type')
+            .count('*')
+            .groupBy('type')
+            .then((res) => {
```
nunogois, line 110: We should try to use `async` / `await` instead of `then` when possible.

nunogois (approved):

src/lib/db/api-token-store.ts
ivarconr: In general I agree. This was to follow the semantics in the file.

PR #5489, opened 2023-11-29
refactor: tag type service feature oriented architecture and tx support

kwasniew:
## About the changes

Tag type service in feature-oriented architecture with tx support (so events are stored with tag info in one tx)


chriswk (approved):
Nice. Lgtm

PR #5492, opened 2023-11-29
chore: test migration backward compatibility

gastonfournier:
## About the changes
This PR will validate that our current migrations are backward compatible with the latest stable release of Unleash.
It will do so by starting a database, applying the latest migrations, and then starting a docker container with the last stable unleash release and running UI tests against it.

There's a risk that the current version of UI tests will not work with the previous version of our UI. Because of that we copied the previous version of cypress tests (https://github.com/Unleash/unleash/tree/5.6/frontend/cypress) into oss folder and removed the ones that are enterprise only. We can discuss a better way of doing this to avoid having to maintain this folder always in sync with the previous version of Unleash

This action will only run when there are changes in migrations or to cypress tests.


chriswk (approved):
Pretty cool!

gastonfournier:
After release 5.7.0 I ended up adapting the UI tests to just do one thing: create a flag. I verified the test fail in 5.6.9 and 5.6.10 prevented the issue. I also tested with the latest image (now at 5.7.0 validating it's compatible with 5.6.10)

PR #5509, opened 2023-11-30
feat: improve the format of project-access-added event

daveleek:
## About the changes

Improves the structure/format of the generated diff for project-access-added events.

Before:

Skjermbilde 2023-11-30 kl  08 32 41

After:

Skjermbilde 2023-11-30 kl  08 32 52

## Discussion points
Couldn't find any relevant event formatter to update

### DIFF
--- a/src/lib/routes/admin-api/events.test.ts
+++ b/src/lib/routes/admin-api/events.test.ts
@@ -116,13 +116,26 @@ test('should anonymise any PII fields, no matter the depth', async () => {
         new ProjectAccessAddedEvent({
             createdBy: 'some@email.com',
             data: {
-                groups: [
-                    {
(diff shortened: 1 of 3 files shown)


ivarconr:
I am not sure if this is the best way represent who got or lost access to a project. 

The event name, `project-access-added`, indicates it will tell you who gained access to the project. 

If it is a single person added I would assume the data field would look something like this

````js
{
   userId: 123
   role: 3
}
```
(Project id would live in the event metadata). 

This means we would have a different event when someone loose access to a project. E.g. `project-access-removed`
The pre-data for this event would be: 

````js
{
   groupId: 123
   role: 3
}
```

daveleek:
Replaced by #5581

PR #5512, opened 2023-11-30
fix: UI navigation (reverts #5506)

nunogois:
Latest version had a UI navigation bug where we wouldn't correctly navigate to the tab and instead it would add infinitely to the breadcrumbs / URL:

https://github.com/Unleash/unleash/assets/14320932/509fa528-7f9e-4476-a945-f74393e99dd5

This fixes the UI navigation by reverting https://github.com/Unleash/unleash/pull/5506


kwasniew (approved):
We should also add a test for nested navigation to guard against 3rd party breaking change

PR #5518, opened 2023-12-01
feat: update project overview endpoint

sjaanus:
1. Created new hook for endpoint
2. Start removing useProject hook, when features not needed.


frontend/src/component/project/Project/ProjectOverview.tsx
```diff
@@ -138,14 +146,29 @@ const ProjectOverview = () => {
 
     if (featureSearchFrontend) return <PaginatedProjectOverview />;
 
+    const featureTypeCounts = features.reduce(
```
sjaanus, line 149: This is so the old component will work with refactor. This code will be removed when flag gets removed.

kwasniew (approved):

PR #5519, opened 2023-12-01
feat: new email template for scheduled-execution-failed email 

andreas-unleash:
What it says on the box.

Relates to # [1-1687](https://linear.app/unleash/issue/1-1687/send-an-email-when-the-scheduling-fails)

Screenshot 2023-12-01 at 15 33 08

### DIFF
--- a/src/lib/services/email-service.ts
+++ b/src/lib/services/email-service.ts
@@ -74,6 +74,7 @@ export class EmailService {
         name: string,
         recipient: string,
         changeRequestLink: string,
+        changeRequestTitle: string,
         scheduledAt: string,
         errorMessage: string,
     ): Promise<IEmailEnvelope> {
@@ -95,6 +96,7 @@ export class EmailService {
                 TemplateFormat.PLAIN,
                 {
                     changeRequestLink,
+                    changeRequestTitle,
                     scheduledAt,
                     errorMessage,
                     name,
--- a/src/lib/services/email-service.ts
+++ b/src/lib/services/email-service.ts
@@ -31,6 +31,8 @@ export interface IEmailEnvelope {
 
 const RESET_MAIL_SUBJECT = 'Unleash - Reset your password';
 const GETTING_STARTED_SUBJECT = 'Welcome to Unleash';
+const SCHEDULED_EXECUTION_FAILED_SUBJECT =
+    'Unleash - Scheduled execution failed';
 
 export const MAIL_ACCEPTED = '250 Accepted';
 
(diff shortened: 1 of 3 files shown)


src/lib/services/email-service.ts
```diff
@@ -31,6 +31,8 @@ export interface IEmailEnvelope {
 
 const RESET_MAIL_SUBJECT = 'Unleash - Reset your password';
 const GETTING_STARTED_SUBJECT = 'Welcome to Unleash';
+const SCHEDULED_EXECUTION_FAILED_SUBJECT =
+    'Unleash - Scheduled execution failed';
```
thomasheartman, line 35: I'd make the subject line a little clearer. Scheduled execution of what? How about this?

```suggestion
    'Unleash - Scheduled change request could not be applied';
```

src/mailtemplates/scheduled-execution-failed/scheduled-execution-failed.html.mustache
```diff
@@ -0,0 +1,522 @@
...
+                    <tr>
+                        <td align="center" valign="top">
+                            <!-- BEGIN BODY // -->
+                            <table border="0" cellpadding="0" cellspacing="0" width="100%" id="templateBody">
+                                <tr>
+                                    <td valign="top" class="bodyContent" mc:edit="body_content">
+                                        <h1>Scheduled changes execution failed</h1>
+
+                                        <br />
+                                        Scheduled change request <span><a class="changeRequestLink" href="{{{ changeRequestLink }}}" target="_blank" rel="noopener noreferrer">{{{ changeRequestLink }}}</a></span> failed to apply at {{{ scheduledAt }}} due to {{{ errorMessage }}}.
```
thomasheartman, line 480: Does this use the link to the change request as the title? As in https://hosted.unleash.io/bob/projects/x/change-requests/9? That feels confusing. Can we use the change request id and title (if present)?

thomasheartman (commented):
Got a couple things to sort out before we send this out. I'm assuming you've talked to someone with experience regarding the templates? It seems like a lot of html, but if that's how we do it, then that's how we do it 💁🏼

src/mailtemplates/scheduled-execution-failed/scheduled-execution-failed.html.mustache
andreas-unleash: Changed: if title is present would be `id: title` if not `id`

src/mailtemplates/scheduled-execution-failed/scheduled-execution-failed.plain.mustache
```diff
@@ -0,0 +1,6 @@
+Scheduled changes execution failed
+
+Scheduled change request {{{ changeRequestLink }}} failed to apply at {{{ scheduledAt }}} due to {{{ errorMessage }}}.
```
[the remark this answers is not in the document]
andreas-unleash: This is the plain text one.  This is what we do for the reset-password plain text (show the link)

thomasheartman (approved):

PR #5526, opened 2023-12-04
chore: update plugin-openapi-docs to v2

thomasheartman:
Updating from the previous v2 beta version we used.


sjaanus:
Tried to use latest 2.4.3 with openapi 2.0.3 based on this compatibility table https://docusaurus-openapi.tryingpan.dev/#compatibility-matrix, but still fails.

PR #5541, opened 2023-12-05
feat: add openapi validation for search

sjaanus:


kwasniew (approved):

src/lib/openapi/spec/search-features-schema.ts
```diff
@@ -33,7 +34,7 @@ export const searchFeaturesSchema = {
     },
     components: {
         schemas: {
-            featureSchema,
+            featureSearchResponse: featureSearchResponseSchema,
```
Tymek, line 37: why is it renamed here?

Tymek (approved):

src/lib/openapi/spec/search-features-schema.ts
sjaanus: Happened because of renaming. Fixed

PR #5547, opened 2023-12-05
feat: Scheduled change conflict email templates and function

andreas-unleash:
Creates a new email template for scheduled change conflicts and a function to send it.

Relates to: #[1-1686](https://linear.app/unleash/issue/1-1686/send-an-email-when-the-conflicts-are-detected)

Screenshot 2023-12-05 at 16 55 51


src/lib/services/email-service.ts
```diff
@@ -31,6 +31,8 @@ export interface IEmailEnvelope {
 
 const RESET_MAIL_SUBJECT = 'Unleash - Reset your password';
 const GETTING_STARTED_SUBJECT = 'Welcome to Unleash';
+const SCHEDULED_CHANGE_CONFLICT_SUBJECT =
+    'Unleash - Detected some conflicts with your scheduled changes';
```
thomasheartman, line 35: Maybe this is clearer and more direct? The other one sounds like it could have been you that did something wrong, when in reality it's just that someone else did something.

```suggestion
    'Unleash - Scheduled changes can no longer be applied';
```

src/lib/services/email-service.ts
```diff
@@ -138,6 +140,86 @@ export class EmailService {
...
 
+    async sendScheduledChangeConflictEmail(
+        recipient: string,
+        conflictScope: 'flag' | 'strategy',
+        changeRequests: {
+            id: number;
+            scheduledAt: string;
+            title?: string;
+        }[],
+        strategyIdOrFlagName: string,
```
thomasheartman, line 151: Maybe we could use types to make this clearer? This looks like two different conflict types. What happens if you send `conflictScope: 'strategy', strategyIdOrFlagName: 'my-feature-flag'`? 

It'll be treated as a strategy, even if you send a flag name, right?

We could do something like this

```ts
type StrategyConflict = {
  conflictScope: 'strategy',
  strategyId: string;
}

type FlagConflict = {
  conflictScope: 'flag',
  flagName: string;
}
```

and then change the signature to:

```ts
        recipient: string,
        changeRequests: {
            id: number;
            scheduledAt: string;
            title?: string;
        }[],
        conflictData: StrategyConflict | FlagConflict
```

Or we could use a union type for the whole thing.

It might not be necessary, but it just struck me that these arguments make it easy to get into states that are not what you mean.

src/lib/services/email-service.ts
```diff
@@ -138,6 +140,86 @@ export class EmailService {
...
+    ) {
+        if (this.configured()) {
+            const year = new Date().getFullYear();
+            const conflict =
+                conflictScope === 'flag'
+                    ? `The feature flag ${strategyIdOrFlagName} has been archived.`
```
thomasheartman, line 157: We can't link to a deleted strategy (because that UUID is gone), but should we link to an archived flag? It might have been deleted in the meantime (and a new one might have been created), but if we want to make it as easy as possible for people to fix the conflict (by reviving the flag), then a link to the flag would be very useful.

src/mailtemplates/schedule-change-conflict/scheduled-change-conflict.plain.mustache
```diff
@@ -0,0 +1,9 @@
+Conflict detected in a scheduled change
+
+{{ conflict }} Scheduled change requests that use this {{ conflictScope }} can no longer be applied and their scheduled applications will fail{{#conflictResolution}}{{.}}{{/conflictResolution}}{{^conflictResolution}}.{{/conflictResolution}}
```
thomasheartman, line 3: I'm not entirely sure how the syntax works here, but just make sure there is a space before the conflict resolution text.

src/mailtemplates/schedule-change-conflict/scheduled-change-conflict.plain.mustache
```diff
@@ -0,0 +1,9 @@
...
+
+For you, this concerns change requests:
+
+{{#changeRequests}}
+    - # {{id}} - {{#title}}- {{.}}{{/title}} (scheduled for {{scheduledAt}})
```
thomasheartman, line 8: We'll need to add links in the plain text email too

src/mailtemplates/schedule-change-conflict/scheduled-change-conflict.html.mustache
thomasheartman, line 1: Couple thoughts on this:

1. We don't mention which project the change request is in. We should probably make that clear.
2. If a strategy is deleted, we should probably also mention which feature flag it belonged to. No one goes around memorizing strategy ids, so that information alone isn't very useful. You'll need to navigate to the change request to get all the information, but knowing at a glance which flag it pertains to would probably be useful.

src/lib/services/email-service.ts
andreas-unleash: I think linking to the change requests is enough imo

src/lib/services/email-service.ts
andreas-unleash: I revised it a bit so that we always get a flag name and project

src/mailtemplates/schedule-change-conflict/scheduled-change-conflict.html.mustache
andreas-unleash: addressed

thomasheartman (approved):
Nice work!

src/lib/services/email-service.test.ts
```diff
@@ -29,17 +30,17 @@ test('Can send reset email', async () => {
...
+    const emailService = new EmailService({
+        email: {
             host: 'test',
-            port: 9999,
+            port: 587,
             secure: false,
-            sender: 'noreply@getunleash.ai',
             smtpuser: '',
             smtppass: '',
+            sender: 'noreply@getunleash.ai',
```
thomasheartman, line 40: Why did you change the port and add the sender? That seems unrelated?

andreas-unleash: Port is completely unrelated - some tests where using one or the other, so when I changed the constructor and copied the config from one test to the other

PR #5548, opened 2023-12-05
feat: improve role form validation

nunogois:
https://linear.app/unleash/issue/2-1717/improve-the-ux-when-all-the-required-fields-are-not-filled-in

Improves role form validation behavior.
We may want to look into a form validation library, like [react-hook-form](https://react-hook-form.com/), for future implementations.


nicolaesocaciu (approved):
LGTM!

PR #5551, opened 2023-12-05
hack: external events

nunogois:
https://linear.app/unleash/issue/2-1721/discovery-should-this-be-part-of-our-current-events

## Events, webhooks and actions

Let's consider three concepts:
- **Events**: Events tell us that something has happened. There are two types: internal events, which occur within Unleash, and external events, which occur outside of Unleash.
- **Event webhooks**: Event webhooks allow you to configure webhooks that when hit will log an external event of your choosing.
- **Event actions**: Event actions allow you to configure actions to be triggered when a specific event is emitted. They serve as automated responses to certain events, executing predefined actions when particular events occur.

## Before this PR

```mermaid
graph TD
    A[Unleash] -->|Internal events| B[Events]
```

## After this PR

```mermaid
graph TD
    A[Unleash] -->|Internal events| B[Events]
    C[Event Webhooks] -->|External events| B
    B -->|Trigger| D[Event Actions]
```

## DB changes

This means 3 main changes:
- **Events**: The `events` table has a new `is_external` column that tells us whether an event is external.
- **Event webhooks**: New `event_webhooks` table stores the configured event webhooks:
  - `enabled` - Whether the event webhook is enabled.
  - `name` - A name for the event webhook.
  - `event` - The event name that the event webhook should log. Can be something totally custom.
  - `url` - A UUID that is automatically generated and will be part of the webhook URL.
- **Event actions**: New `event_actions` table stores the configured event actions:
  - `enabled` - Whether the event action is enabled.
  - `name` - A name for the event action.
  - `event` - The event name that should trigger the event action.
  - `action` - The action that is going to run. Currently only supports `FEATURE_TOGGLE` (turns a flag on or off).
  - `parameters` - Any parameters relevant to running the action.

## The new features

These are different features and are not dependent on each other.

Logging external events to Unleash can be useful for multiple reasons, and it doesn't necessarily mean that you'd like them to trigger actions. In that case, you can simply configure the necessary event webhooks to log these events, without ever having to configure actions.

Same thing with event actions. You don't necessarily need to log external events to use event actions, since you can configure your actions to run based on internal events by simply specifying the internal event name. E.g. this could be a way to accomplish dependent feature toggles, by creating an action that enables a flag where the trigger event is that another flag was enabled.

### Creating and triggering an event webhook

https://github.com/Unleash/unleash/assets/14320932/501cd246-25d1-4c72-8b4c-0e7d5d221504

### Creating an action that enables a feature flag when our custom external event occurs

https://github.com/Unleash/unleash/assets/14320932/a17f9560-34a7-4176-bccc-4a42a8ae58d5

## About the PR

Most of the code lives in the Enterprise repo, and I've opted to extend existing components where it made sense.
Some of the code changes are girl scouting tasks or things like slightly adapting the existing code to the changes.

I'm a bit unsure about my usage of transactional services, but that's something that can be improved in future PRs.

This is my suggested approach for the "external events" project. I think this is an elegant solution that offers a lot of flexibility for the future. However I'm obviously biased, so that's where you come in. Please share your thoughts and how you feel about this.

Special credit to @gardleopard for rightfully mentioning that `events` are in fact `events`, pointing me in this direction.

### DIFF
--- a/src/lib/db/event-store.ts
+++ b/src/lib/db/event-store.ts
@@ -85,12 +85,12 @@ export interface IEventTable {
 const TABLE = 'events';
 
 class EventStore implements IEventStore {
-    private db: Db;
+    protected db: Db;
 
     // only one shared event emitter should exist across all event store instances
     private eventEmitter: EventEmitter = sharedEventEmitter;
 
-    private logger: Logger;
+    protected logger: Logger;
 
     // a new DB has to be injected per transaction
     constructor(db: Db, getLogger: LogProvider) {
--- a/src/lib/openapi/spec/event-schema.ts
+++ b/src/lib/openapi/spec/event-schema.ts
@@ -1,6 +1,5 @@
 import { FromSchema } from 'json-schema-to-ts';
 import { tagSchema } from './tag-schema';
-import { IEventTypes } from '../../types';
 import { variantSchema } from './variant-schema';
 
 const eventDataSchema = {
@@ -44,7 +43,6 @@ export const eventSchema = {
             type: 'string',
             description:
                 'What [type](https://docs.getunleash.io/reference/api/legacy/unleash/admin/events#event-type-description) of event this is',
-            enum: IEventTypes,
             example: 'feature-created',
         },
         createdBy: {
--- a/src/lib/services/event-service.ts
+++ b/src/lib/services/event-service.ts
@@ -10,7 +10,7 @@ import { ITag } from '../types';
 export default class EventService {
     private logger: Logger;
 
-    private eventStore: IEventStore;
+    protected eventStore: IEventStore;
 
     private featureTagStore: IFeatureTagStore;
 
@@ -51,7 +51,7 @@ export default class EventService {
         return this.eventStore.on(eventName, listener);
     }
 
-    private async enhanceEventsWithTags(
+    protected async enhanceEventsWithTags(
         events: IBaseEvent[],
     ): Promise<IBaseEvent[]> {
         const featureNamesSet = new Set<string>();
--- a/src/lib/types/events.ts
+++ b/src/lib/types/events.ts
@@ -173,6 +173,14 @@ export const BANNER_CREATED = 'banner-created' as const;
 export const BANNER_UPDATED = 'banner-updated' as const;
 export const BANNER_DELETED = 'banner-deleted' as const;
 
+export const EVENT_WEBHOOK_CREATED = 'event-webhook-created' as const;
+export const EVENT_WEBHOOK_UPDATED = 'event-webhook-updated' as const;
+export const EVENT_WEBHOOK_DELETED = 'event-webhook-deleted' as const;
+
+export const EVENT_ACTION_CREATED = 'event-action-created' as const;
+export const EVENT_ACTION_UPDATED = 'event-action-updated' as const;
+export const EVENT_ACTION_DELETED = 'event-action-deleted' as const;
+
 export const IEventTypes = [
     APPLICATION_CREATED,
     FEATURE_CREATED,
@@ -306,6 +314,12 @@ export const IEventTypes = [
     PROJECT_ENVIRONMENT_ADDED,
     PROJECT_ENVIRONMENT_REMOVED,
     DEFAULT_STRATEGY_UPDATED,
+    EVENT_WEBHOOK_CREATED,
+    EVENT_WEBHOOK_UPDATED,
+    EVENT_WEBHOOK_DELETED,
+    EVENT_ACTION_CREATED,
+    EVENT_ACTION_UPDATED,
+    EVENT_ACTION_DELETED,
 ] as const;
 export type IEventType = (typeof IEventTypes)[number];
 
--- a/src/migrations/20231205122703-external-events.js
+++ b/src/migrations/20231205122703-external-events.js
@@ -1,0 +1,19 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        ALTER table events
+                ADD COLUMN IF NOT EXISTS is_external boolean DEFAULT false
+        `,
+        cb,
+    );
+};
+
+exports.down = function (db, cb) {
+    db.runSql(
+        `
+        ALTER table events
+                DROP COLUMN is_external
+        `,
+        cb,
+    );
+};
--- a/src/migrations/20231205155322-event-webhooks.js
+++ b/src/migrations/20231205155322-event-webhooks.js
@@ -1,0 +1,27 @@
+'use strict';
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS event_webhooks
+            (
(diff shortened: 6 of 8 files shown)


ivarconr:
My biggest concern is that we generalize two different concepts. It is not clear to me the benefits of sharing a table for external and internal events. 

Our current events table serve a very important need for Unleash users, to form a audit log for all changes in Unleash. They should have a clear created by, and we have specific need for the retention.

External events will in most cases have weaker retention, and we will most likely both rate limit them. I also think we will end up with different meta data to describe these events. 

We are also using the event table to drive logic for our optimized 304 logic. Having a flood of external events could potentially impact this logic, if we have incorrect indexing or similar. 

My vote goes to store external events in a seperate database table.

gastonfournier:
Let me provide my thoughts without having looked at the code.

## Important
- Can we have a different retention policy for external events? We currently hold events forever. I'm still inclined to use a different table for these events, maybe not even storing them (perhaps just a log), unless the system reacts to that event, in which case we'll have the audit of what changed (i.e. feature toggle enabled), which can be decorated with the information from the external event that caused our system to react.
- I'm thinking of high-velocity events like errors. If a system has a surge of errors (e.g. 100 in one second), do we want to add this event to the event's table 100 times? What if it's triggering the action of disabling a feature, do we want to change it 100 times? We can of course rate limit, but we'd still be performing this action multiple times. This also makes me think about not storing external events, not because it's not doable but because of the additional complexity of storing all events problem compared to not storing them.
- Who is the actor? Explain this in the PR description. I saw PAT being used, but the "external event" user is not the one "making the change" it's the system that reacts to the notification of an event.
- Can we send events before configuring the webhook? I'm thinking this because of the UUID, what if we use a name instead? What if we use the same url and route based on the event name?

## Details:
- Can the response be 202 Accepted rather than 200 when receiving an event?
- Do we limit payload size?
- What if the payload contains PII information?
- How do we link the payload with the parameters to the event? I'm thinking, what if the project is part of the payload instead of configured in the event?
- What if we want to trigger based on a threshold (e.g. when 5 events of this type are received in 1 minute)? Is this something we want to support? It might be useful to disable a feature when a threshold of errors is breached.

nunogois:
> Let me provide my thoughts without having looked at the code.
> 
> ## Important
> * Can we have a different retention policy for external events? We currently hold events forever. I'm still inclined to use a different table for these events, maybe not even storing them (perhaps just a log), unless the system reacts to that event, in which case we'll have the audit of what changed (i.e. feature toggle enabled), which can be decorated with the information from the external event that caused our system to react.

Sure, there's nothing preventing us from having a scheduled process that cleans up external events and leaves internal ones intact.

Storing them seems valuable from a user perspective IMO, even if it's just a log.

I think we already have the information we need to do this decoration, it just depends on how we want it to look like.

> * I'm thinking of high-velocity events like errors. If a system has a surge of errors (e.g. 100 in one second), do we want to add this event to the event's table 100 times? What if it's triggering the action of disabling a feature, do we want to change it 100 times? We can of course rate limit, but we'd still be performing this action multiple times. This also makes me think about not storing external events, not because it's not doable but because of the additional complexity of storing all events problem compared to not storing them.

I think it's up to user implementation. If the user decides to configure an action that is triggered by "app-error" and then logs "app-error" 100 times, I think it's reasonable to think that the system will trigger the action 100 times.

We can of course try to create some guardrails against this.

> * Who is the actor? Explain this in the PR description. I saw PAT being used, but the "external event" user is not the one "making the change" it's the system that reacts to the notification of an event.

I think it depends on how you look at it. The actor for the webhook is definitely the user/token triggering the webhook IMO. Whether that same actor should be logged on the event actions I'm not so sure, maybe it should be tagged as "system" like you're suggesting. Either way this feels like details, and I already added the `(webhook: <webhook name>)` example to this implementation. Like I mentioned above, I think we have enough information to decorate this however we'd like.

> * Can we send events before configuring the webhook? I'm thinking this because of the UUID, what if we use a name instead? What if we use the same url and route based on the event name?

I'm not sure what you mean but I guess it depends on how we approach permissions. With the current implementation in the PR I guess it wouldn't really matter because we're requesting a token with ADMIN permissions anyways, but it's something we should think carefully if we decide to open it to a lower permission or not request a token at all.

> ## Details:
> * Can the response be 202 Accepted rather than 200 when receiving an event?

Sure, but I decided on `200` after consulting with ChatGPT 😅 The only thing registering an event does is registering the event. The action that is triggered runs on a different process.

> * Do we limit payload size?

Not currently and not specifically, but it's something we could do. I'm guessing Unleash implements generic payload size rules for all requests.

> * What if the payload contains PII information?

Up to user implementation. Are there any issues you're anticipating that wouldn't happen with e.g. adding PII to a feature strategy?

> * How do we link the payload with the parameters to the event? I'm thinking, what if the project is part of the payload instead of configured in the event?

I think that's something we could potentially map in an easy way.

> * What if we want to trigger based on a threshold (e.g. when 5 events of this type are received in 1 minute)? Is this something we want to support? It might be useful to disable a feature when a threshold of errors is breached.

I also think this is up to user implementation and out of scope for a first iteration. Instead of logging 5 events of the same type, why not implement this logic internally and only log an event called "5-or-more-events-in-less-than-1-minute" and then set up an event action for it?

gastonfournier:
> It made sense to me that a configured webhook only registers the configured event for that webhook. If we imagine a scenario where the event is sent as part of the webhook, you would only need to know about one webhook URL to be able to log all kinds of events, which is potentially destructive, as we may be logging false information and triggering actions that we shouldn't be able to.

This is how Datadog does it: https://docs.datadoghq.com/api/latest/events/ just a single endpoint you can send anythingg

sighphyre:
I'm still not sure what the shared abstraction buys us here, seems like the only thing we get for free is reusing the existing event emitter. I see more reasons to break these apart then I do to force them to follow the same code path. Early abstractions  are incredibly expensive to maintain and they only get more painful with time. I'd like to see some actual use cases before we commit to this - I'm okay with us opening doors for the future but then we need to have plans to walk through those doors.

>  E.g. this could be a way to accomplish dependent feature toggles, by creating an action that enables a flag where the trigger event is that another flag was enabled.

Dependent features are much more complex than this, this won't work without some painful trade-offs and locking us out of other doors (like variant dependency which we do already support).

Things off the top of my head that are not obviously shared:

- Our existing events are hard-coded constants, which I think they should be, they're part of our traceability story. My gut feel is that users should be able to respond to whatever they want. If they want to do some actions based on "LABRADOR-ATE-MY-SERVER", they should be free to do that without source code changes.
- Existing events support our Audit log, which is a significant feature for our UPP customers, external events doesn't feel like that to me (although changes that arise from responding to that event will be).
- Database table requirements are likely to be different for these two features
- Presentation of these two features in the UI is likely to be different
- Responding to changes in internal events opens the door for cyclical event changes, which means implementing cycle detection

All of these things are very solvable but I'd need a compelling use case to be eager to do that. I don't see these features converging, I see them diverging

nunogois:
Closing for now in favor of a slightly different approach.

PR #5562, opened 2023-12-06
feat: segments operators

kwasniew:
## About the changes

Segment operators set to expected values and correct enabled state.


frontend/src/component/common/FilterItem/FilterItem.tsx
```diff
@@ -53,7 +53,7 @@ export const FilterItem: FC<IFilterItemProps> = ({
     const currentOperator = state ? state.operator : currentOperators[0];
 
     const onDelete = () => {
-        onChange({ operator: 'IS', values: [] });
+        onChange({ operator: singularOperators[0], values: [] });
```
kwasniew, line 56: default operator is first singular now instead of hardcoding

sjaanus (approved):

PR #5563, opened 2023-12-06
refactor: split filter configuration with filter visibility state

sjaanus:
This PR splits the filter configuration with filter visibility state.
This will simplify adding different filter types in future, for example date filters.


frontend/src/component/feature/FeatureToggleList/FeatureToggleFilters/AddFilterButton.tsx
```diff
@@ -29,15 +29,11 @@ const AddFilterButton = ({
...
-        const filters = availableFilters.map((filter) =>
-            filter.label === label
-                ? {
-                      ...filter,
-                      enabled: true,
-                  }
-                : filter,
-        );
-        setAvailableFilters(filters);
+        const filterVisibility = {
```
sjaanus, line 32: Simplifies state change, since we keeping just a Map, instead of complex objects.

Tymek (approved):

PR #5564, opened 2023-12-06
Add one click deploy to readme

jbgh:
Hey!

We built this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.

After deployment, users receive a link like this, which can then be CNAME-ed to any domain:

https://unleash-1665.dome.tools

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.


thomasheartman:
Hi, @jbgh! Appreciate the suggestion. What is Dome? I checked the website and the github, but there is very little information available? The link points to a signup page, what's the expected user flow for this?

sjaanus:
Closing, because no activity on this and lack of information.

PR #5565, opened 2023-12-07
chore: use better images for inverted operator

sighphyre:
This swaps out the negated operator, which previous displayed as an SVG of "!" or "!" with a strike through to be an SVG of the actual word "NOT". This should improve the experience of non technical folks who use constraints so that it's much more explicit.

Constraints in the accordion now render like so:

image

And constraints during editing now render like (when negated is active):

image

And when negated is inactive:

image


nunogois (changes requested):
After aligning with @nicolaesocaciu this seems to slightly break the original design. Here's a PR with the suggested changes: https://github.com/Unleash/unleash/pull/5567

nunogois (approved):

PR #5567, opened 2023-12-07
fix: UI changes

nunogois:
Includes some fixes to https://github.com/Unleash/unleash/pull/5565 after aligning with @nicolaesocaciu 

 - Keeping the `isPrefix` prop is the correct thing to do here (it's also the only place in the whole codebase where it's being used)
 - We should not add a `paddingRight` to the parent
 - We don't need to set the color on the icon, since it's inherited by `StyledIconWrapperBase`
 - After the changes above, the icon looks a bit tight in its box, so we fix that:
   - We set `width` to `auto` in `StyledPrefixIconWrapper` to override the `width: 24` on `StyledIconWrapperBase`
   - We set `paddingLeft` and `paddingRight` to `theme.spacing(1)` (8px)

image


sighphyre (approved):
G

PR #5568, opened 2023-12-07
Feat/introduce custom unleash auth header

FredrikOseberg:
Based on feedback from users, we've learnt that it might be best if we in the future migrate our clients to authenticate with unleash using a custom unleash header instead of using the `Authorization` header. This PR sets up basic support for both in a transition period and is aimed to spark a discussion around this topic.

**Discussion points**
* Should we consider this approach to support both and build support into edge / clients over time to use the new header?
* Is the header correctly named?
* Are there edge cases we haven't yet considered with this approach?


src/test/e2e/helpers/database-init.ts
```diff
@@ -86,6 +87,12 @@ export default async function init(
     getLogger: LogProvider = noLoggerProvider,
     configOverride: Partial<IUnleashOptions> = {},
 ): Promise<ITestDb> {
+    if (isNotSnakeCase(databaseSchema)) {
+        throw new Error(
+            `db init database schema must be snake case, was: ${databaseSchema}`,
+        );
+    }
```
FredrikOseberg, line 94: I added this because dbInit schemas must be snake case. If you try to call it with hyphons it will fail silently: `dbInit('my-schema')`

Now you will get a clear error message if you make this mistake, and don't spend 30 minutes debugging silly things.

chriswk (approved):
I like it. Makes us able to easier deploy to environments where the Authorization header is already taken.

src/lib/features/client-feature-toggles/tests/client-feature-toggles-auth.e2e.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        .expect(200);
+});
+
+test('should allow requests to /api/client/features with x-unleash-auth header', async () => {
+    await app.request
+        .get('/api/client/features')
+        .set('x-unleash-auth', clientToken.secret)
+        .expect('Content-Type', /json/)
+        .expect(200);
+});
```
gastonfournier, line 59: One edge case to consider:
If we send both headers, which one takes precedence? I believe we should read from `x-unleash-auth` and don't fail if both are present. Maybe we can test it's still 200 when x-unleash-auth has a valid token, having also the Authorization header with an invalid string.

gastonfournier (approved):
I like the initiative, but this also makes me wonder if this is the way we want to authorize in the future. I remember @chriswk mentioning other things such as JWT tokens or something already used in the industry. Not something tailor-made for Unleash.

Because that can be a longer conversation I believe this is good regardless what we decide on the other topic

src/lib/features/client-feature-toggles/tests/client-feature-toggles-auth.e2e.test.ts
FredrikOseberg: Agree. That should be the default behavior, I'll add a test that verifies it.

FredrikOseberg: addressed in [this commit](https://github.com/Unleash/unleash/pull/5568/commits/fccbd18a042cf990b3dfa4102f91a48fe8205d1a)

FredrikOseberg:
> Can you explain why introducing a new header and Introduce a migration is better than trying to align Unleash with the specification? I supper there is a problem you are trying to solve that is not described here.
> 
> I suspect we need to move towards a Bearer token at some point and this would introduce another transition.
> 
> If the problem is that companies use the Authorization header for other purposes, maybe we can make the header used customizable? This is already supported by our proxy SDK: https://github.com/Unleash/unleash-proxy-client-js#available-options

Certainly. This came up as a way to try to kill two birds with one stone. We had two basic problems we were discussing: 

1) Our Authorization header does not strictly support the protocol
2) Some of our users have their own proxies that sit between Edge and Unleash

The rationale for implementing a custom auth header for unleash is that it can solve both problems. We can define our own protocol for the custom header and don't need to adhere to the Authorization protocol, and we could free up the Authorization field to be used by other services so that you don't need to hack it to make it work. We could potentially allow you to control what the header sent from edge is set as (we did indeed discuss this as an option) but eventually the request needs to end up in Unleash, which means that you you need to do some header swapping to make it work. Example: 

Skjermbilde 2023-12-07 kl  20 41 41

Implementing a custom auth header could potentially solve both issues, and it would be fairly cheap to implement. By supporting it in edge and unleash we could achieve good support and migrate on our own time. 

Of course, we might have missed something. Or we might decide that we want to keep the Authorization token and implement the Bearer protocol here for Unleash. Could you elaborate on that solution, and the benefits you see there?

gardleopard (approved):

PR #5570, opened 2023-12-07
chore: unify build and build PRs workflow

gastonfournier:
## About the changes
Our build and build_prs workflows are almost the same with some minor details:

1. **Build** also builds the frontend, but **PR Build** does not by specifying `--ignore-scripts`
1. **Build PR** uploads test results: Upload test report to build. We'll ignore this when building on main
1. **Build PR** just builds the backend (reason why the build is much faster: 50s vs ~5m). We have a workflow for building frontend  PRs, we'd be removing the build frontend from the build on main, but we're already exercising that to upload to dockerhub: 
image
 
These are the main differences:
image

This PR unifies the two workflows into one


nunogois (approved):
Let's try it out!

PR #5576, opened 2023-12-08
feat: test filter date item

sjaanus:


frontend/src/component/common/FilterDateItem/FilterDateItem.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    return recordedChanges;
+};
+
+afterEach(() => {
+    vi.restoreAllMocks();
+});
+
+describe('FilterDateItem Component', () => {
+    it('renders initial state correctly', async () => {
+        vi.mock('hooks/useLocationSettings', () => ({
```
kwasniew, line 32: import overrides are very hacky and magical as they overwrite how native imports work. Can we handle it similar to how we do local storage?

kwasniew (approved):
Overall nice change. I'd only try to do it without import mocking.

frontend/src/component/common/FilterDateItem/FilterDateItem.test.tsx
sjaanus: Actually this was a idea I was working on and left in. Removed.

kwasniew: overall I'd recommend not using jest.someMagic or vi.somMagic. It usually means the code is not testable and requires some testing framework magic.

PR #5577, opened 2023-12-08
docs: document how segment conflicts are handled

thomasheartman:
As discovered in the recent segments breakathon, the docs aren't very clear about how conflicts are handled. This PR better documents when they can't be deleted or moved. It also mentions the edge case where a segment is used in an archived flag (it still can't be deleted, so you must either delete the flag, or revive it and remove the segment)


kwasniew (approved):
Very useful info

PR #5578, opened 2023-12-08
feat: add schema for change request strategies

thomasheartman:
This change adds a property to the segmentStrategiesSchema to make sure that change request strategies are listed in the openapi spec

It also renames the files that contains that schema and its tests from `admin-strategies-schema` to `segment-strategies-schema`.


andreas-unleash (approved):
LGTM

PR #5580, opened 2023-12-11
feat: add reset password to user manager

PauloMiranda98:
## About the changes
This PR aims to add reset password to user manager. This will help those who do not intend to configure the email system.

Screenshot from 2023-12-10 20-41-22

Screenshot from 2023-12-16 16-40-36

Screenshot from 2023-12-16 16-40-48

## Discussion points
I didn't find a test for this part of user management in the frontend. Do I need to do automated testing for this implementation?

### DIFF
--- a/frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
+++ b/frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
@@ -50,7 +50,9 @@ const ResetPassword = ({
             if (token) {
                 setResetLink(token.resetPasswordUrl);
             } else {
-                setToastApiError('Could not reset password');
+                setToastApiError(
+                    'Could not reset password. If you made another request in less than a minute, wait for one minute and try again.',
+                );
             }
         } catch (error) {
             setToastApiError(formatUnknownError(error));
--- a/frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
+++ b/frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
@@ -36,7 +36,9 @@ const ResetPassword = ({
     const submit = async (event: React.SyntheticEvent) => {
         event.preventDefault();
         if (!user.email) {
-            setToastApiError('User without email cannot reset password');
+            setToastApiError(
+                "You can't reset the password of a user who doesn't have an email address.",
+            );
             return;
         }
 
@@ -48,7 +50,7 @@ const ResetPassword = ({
             if (token) {
                 setResetLink(token.resetPasswordUrl);
             } else {
-                setToastApiError("Could not reset password");
+                setToastApiError('Could not reset password');
             }
         } catch (error) {
             setToastApiError(formatUnknownError(error));
@@ -84,7 +86,7 @@ const ResetPassword = ({
                 )}
             >
                 <Typography variant='subtitle1'>
-                    Reseting password for user
+                    Resetting password for user
                 </Typography>
                 <div className={themeStyles.flexRow}>
                     <StyledUserAvatar user={user} variant='rounded' />
@@ -104,13 +106,10 @@ const ResetPassword = ({
                 title='Reset password link created'
             >
                 <Box>
-                    <Typography variant='body1' sx={{ mb: 2 }}>
-                        Using this link the user can reset the password.
-                    </Typography>
                     <Typography variant='body1'>
-                        Provide them with the following link to reset their
-                        password. This will allow them to set up a new password
-                        and get back to using their Unleash account.
+                        The user can use this link to reset their password. You
+                        should not share this link with anyone but the user in
+                        question.
                     </Typography>
                     <LinkField inviteLink={resetLink} />
                 </Box>
--- a/frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
+++ b/frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
@@ -1,0 +1,124 @@
+import React, { useState } from 'react';
+import classnames from 'classnames';
+import { Box, styled, TextField, Typography } from '@mui/material';
+import { modalStyles } from 'component/admin/users/util';
+import { Dialogue } from 'component/common/Dialogue/Dialogue';
+import { useThemeStyles } from 'themes/themeStyles';
+import { IUser } from 'interfaces/user';
+import useAdminUsersApi from 'hooks/api/actions/useAdminUsersApi/useAdminUsersApi';
+import { UserAvatar } from 'component/common/UserAvatar/UserAvatar';
+import useToast from 'hooks/useToast';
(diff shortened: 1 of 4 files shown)


frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,124 @@
...
+            return;
+        }
+
+        try {
+            const token = await resetPassword(user.email).then((res) =>
+                res.ok ? res.json() : undefined,
+            );
+
+            if (token) {
+                console.log(token);
```
FredrikOseberg, line 49: This needs to be removed

frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,124 @@
...
+                setResetLink(token.resetPasswordUrl);
+            } else {
+                setToastApiError("Could not reset password");
+            }
+        } catch (error) {
+            console.warn(error);
```
FredrikOseberg, line 55: I think setting the toast is enough here

frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
PauloMiranda98: Done: https://github.com/Unleash/unleash/pull/5580/commits/2b555473f994f5e4c05f919d1a94a96f069a3298

frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
PauloMiranda98: Done: https://github.com/Unleash/unleash/pull/5580/commits/2b555473f994f5e4c05f919d1a94a96f069a3298

frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,122 @@
...
+}: IChangePasswordProps) => {
+    const { classes: themeStyles } = useThemeStyles();
+    const { resetPassword } = useAdminUsersApi();
+    const { setToastApiError } = useToast();
+    const [resetLink, setResetLink] = useState('');
+
+    const submit = async (event: React.SyntheticEvent) => {
+        event.preventDefault();
+        if (!user.email) {
+            setToastApiError('User without email cannot reset password');
```
thomasheartman, line 39: Small rewording suggestion: it's not the user who's attempting to reset the password, but you, the admin, so this rephrases the text a bit to reflect that.

```suggestion
            setToastApiError('You can't reset the password of a user who doesn't have an email address.');
```

frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,122 @@
...
+        >
+            <form
+                onSubmit={submit}
+                className={classnames(
+                    themeStyles.contentSpacingY,
+                    themeStyles.flexColumn,
+                )}
+            >
+                <Typography variant='subtitle1'>
+                    Reseting password for user
```
thomasheartman, line 87:
```suggestion
                    Resetting password for user
```

frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,122 @@
...
+            >
+                <Box>
+                    <Typography variant='body1' sx={{ mb: 2 }}>
+                        Using this link the user can reset the password.
+                    </Typography>
+                    <Typography variant='body1'>
+                        Provide them with the following link to reset their
+                        password. This will allow them to set up a new password
+                        and get back to using their Unleash account.
+                    </Typography>
```
thomasheartman, line 114: How about something like this? I'm not sure we need both paragraphs (they both say the same), and it might be worth adding a note that it shouldn't be shared with any third parties 💁🏼

```suggestion
                    <Typography variant='body1'>
                        The user can use this link to reset their password. 
                        You should not share this link with anyone but the 
                        user in question.
                    </Typography>
```

thomasheartman (commented):
@PauloMiranda98 Thanks again for your contribution 😄 I think this looks great! I've tested the functionality and everything seems to work as expected. I've added a couple minor suggestions for some of the text that I'd like you to have a look at, but once that's good I'm happy to merge this 🙌🏼

frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
PauloMiranda98: I liked the suggestion

PauloMiranda98:
@thomasheartman @FredrikOseberg, I finished the corrections.

This implementation has a limitation due to a backend characteristic. When you try to reset the same user's password twice in the same minute, it results in the 'Could not reset password' error. Is it ok to continue with this "limitation"?

ivarconr:
First of all, super cool! Thanks for doing this! It indeed improve Unleash. 

>  it results in the 'Could not reset password'

Do you know why this happens? I would prefer to just sort it out before merging.

PauloMiranda98:
> First of all, super cool! Thanks for doing this! It indeed improve Unleash.
> 
> > it results in the 'Could not reset password'
> 
> Do you know why this happens? I would prefer to just sort it out before merging.

@ivarconr, this problem happens because this PR (https://github.com/Unleash/unleash/pull/2076). Using the backend it is not possible to reset the same user's password twice within 1 minute. This raises an exception in the backend. What do you suggest doing in this case?

ivarconr:
> What do you suggest doing in this case?

Thanks for the explanation. I suggest we can show that reason in the error message. It will be informative for the user.

PauloMiranda98:
@ivarconr wdyt about https://github.com/Unleash/unleash/pull/5580/commits/d9b95d9baeda23aa5d28f02492c4a425d705eb93?

frontend/src/component/admin/users/UsersList/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,123 @@
...
+        try {
+            const token = await resetPassword(user.email).then((res) =>
+                res.ok ? res.json() : undefined,
+            );
+
+            if (token) {
+                setResetLink(token.resetPasswordUrl);
+            } else {
+                setToastApiError(
+                    'Could not reset password. If you made another request in less than a minute, wait for one minute and try again.',
```
thomasheartman, line 54: Maybe something like this?

```suggestion
                    'Could not reset password. This may be to prevent too many resets. Try again in a minute.',
```

thomasheartman (approved):
Nice work on this one and thanks for implementing all our suggestions 🔥   I'll go ahead and merge it now 🙌🏼

PR #5581, opened 2023-12-11
chore: list users and groups under each role in projectaccessadded event

daveleek:
## About the changes

Replaces #5509 
Changes the project access added event to list all users and groups added to each role instead of in root event.

Skjermbilde 2023-12-11 kl  08 38 12

## Discussion points
- There are separate events that demonstrate access added, removed, etc for single instances of user and group, but none batched like this.
- Should this event be split up into one per instance of roleId? It kind of loses it's tie to a specific operation..

### DIFF
--- a/src/lib/routes/admin-api/events.test.ts
+++ b/src/lib/routes/admin-api/events.test.ts
@@ -120,6 +120,8 @@ test('should anonymise any PII fields, no matter the depth', async () => {
                     {
                         roleId: 1,
                         groupIds: [1, 2],
+                        // Doesn't reflect the real data structure for event here, normally a number array.
+                        // Testing PII anonymisation
                         users: [{ id: 1, username: testUsername }],
                     },
                 ],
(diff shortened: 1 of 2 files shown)


src/lib/routes/admin-api/events.test.ts
```diff
@@ -116,11 +116,11 @@ test('should anonymise any PII fields, no matter the depth', async () => {
...
             data: {
-                groups: [
+                roles: [
                     {
-                        name: 'test',
-                        project: 'default',
-                        users: [{ username: testUsername }],
+                        roleId: 1,
+                        groupIds: [1, 2],
+                        users: [{ id: 1, username: testUsername }],
```
ivarconr, line 123: should we drop the username? Feels like a user id should be enough?

ivarconr (approved):
I just have a small comment. Other than that I think this improvements looks good.

src/lib/routes/admin-api/events.test.ts
daveleek: This is only in the test, it's actually a PII test that just happens to use that particular event.

daveleek: I can take a look at removing this particular test/making it a more appropriate test.

daveleek: I've added a comment to the test that explains the use case and that it's specific to testing PII and not directly relevant of the event itself

PR #5584, opened 2023-12-11
feat: add tags filter

sjaanus:


frontend/src/component/feature/FeatureToggleList/FeatureToggleFilters/FeatureToggleFilters.tsx
```diff
@@ -49,6 +52,7 @@ export const FeatureToggleFilters: VFC<IFeatureToggleFiltersProps> = ({
 }) => {
     const { projects } = useProjects();
     const { segments } = useSegments();
+    const { tags } = useTags('.');
```
sjaanus, line 55: This will call /api/admin/tags/,  which returns all tags.

frontend/src/utils/serializeQueryParams.ts
```diff
@@ -49,13 +49,18 @@ const decodeFilterItem = (
         return undefined;
     }
 
-    const [operator, values = ''] = input.split(':');
-    if (!operator) return undefined;
+    const pattern =
```
sjaanus, line 52: Make serialization have same logic as backend. Only splitting on operators, not if text has colon in it.

frontend/src/component/feature/FeatureToggleList/FeatureToggleFilters/FeatureToggleFilters.tsx
kwasniew: I'd hide it at the source and expose clean API for clients

kwasniew (approved):

frontend/src/component/feature/FeatureToggleList/FeatureToggleFilters/FeatureToggleFilters.tsx
sjaanus: Added new hook. useAllTags

PR #5588, opened 2023-12-11
chore: migration for features created by

daveleek:
## About the changes

Contains the migration that adds the column created_by to `features`


chriswk:
It feels like you've done an extra fix as part of this PR?

daveleek:
Argh, thank you. I've started this off the wrong branch. was pretty sure I was on main but I guess I wasn't. will get a new one up with just this change. That's a separate PR

daveleek:
replaced by #5593

PR #5590, opened 2023-12-11
chore: add migration for adding created_by to feature_tag table

daveleek:
## About the changes

Contains the migration that adds the column created_by to `feature_tag`


chriswk (approved):
LGTM

PR #5594, opened 2023-12-11
feat: add column created_by to role_user table

chriswk:
As it says in the title. Adds a created_by column to the role_user table to more easily track who's doing what.


daveleek (approved):
LGTM!

PR #5596, opened 2023-12-11
feat: add created_by to api_tokens table

chriswk:
Adds a migration for adding created_by column to the api_tokens table.


daveleek (approved):
LGTM!

PR #5601, opened 2023-12-11
docs: Clarify edge proximity to users

gardleopard:
A benefit of hosting your own edge instance is that it will be as close to end users as you are.


website/docs/understanding-unleash/proxy-hosting.mdx
```diff
@@ -26,7 +26,7 @@ If you want Unleash to host the Frontend API for you, you should be aware of the
 -   This is only available to Pro and Enterprise customers who have signed up for a managed Unleash instance.
 -   We allow short spikes in traffic and our adaptive infrastructure will automatically scale to your needs.
 -   Please check the [Fair Use Policy](https://www.getunleash.io/fair-use-policy) to see the limits of the Unleash-hosted Frontend API.
--   There's no guarantee that it'll be close to your applications. This means you'll get higher response times.
+-   There's no guarantee that it'll be geographically close to your end users, this means your end users might be getting slower response times on feature toggle evaluations.
```
thomasheartman, line 29: I think we're moving the docs towards using "flags" over "toggles", so we should try to use "flag" for new content, I think.

```suggestion
-   There's no guarantee that it'll be geographically close to your end users, this means your end users might be getting slower response times on feature flag evaluations.
```

thomasheartman (approved):
🔥

PR #5603, opened 2023-12-11
update withTableState

Tymek:
## About the changes
Handle column visibility from table state in URL and local storage.

### DIFF
--- a/frontend/src/utils/withTableState.ts
+++ b/frontend/src/utils/withTableState.ts
@@ -3,6 +3,7 @@ import {
     type SortingState,
     type PaginationState,
     type TableOptions,
+    type VisibilityState,
     getCoreRowModel,
 } from '@tanstack/react-table';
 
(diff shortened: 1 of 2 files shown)


frontend/src/utils/withTableState.ts
```diff
@@ -73,6 +74,39 @@ const createOnPaginationChange =
         }
     };
 
+const createOnColumnVisibilityChange =
+    (
+        tableState: {
+            columns?: (string | null)[] | null;
```
kwasniew, line 80: maybe a type alias since we use this type a lot?

kwasniew (commented):
We're at the state that this code should already have a test.

frontend/src/utils/withTableState.ts
```diff
@@ -7,6 +7,8 @@ import {
     getCoreRowModel,
 } from '@tanstack/react-table';
 
+type TableStateColumnsType = (string | null)[] | null;
```
kwasniew, line 10: We can drop the "Type" suffix

kwasniew (approved):

PR #5604, opened 2023-12-11
feat: new columns visibility menu

Tymek:
## About the changes
Simplified columns visibility menu.

`<ColumnsMenu />` is using it's own list of elements instead of inferring it from table columns definition. This component doesn't need it's internal structure to depend on react-table and it's interface.


kwasniew (commented):
Can you add some explanation what was simplified so that we have some "lightweight ADR like description" inside this PR. In the past I was browsing some old commits where I tried to understand the context and longer descriptions were usually helpful.

kwasniew (approved):

PR #5607, opened 2023-12-12
fix: optimize headers we return for API calls.

ivarconr:
Today we include a lot of "secutiry headers" for all API calls. Quite a lot of them are only relevent when we return a HTML document for the browser.

This PR removes and simplify these headers for API calls, so that we do not include unecessary data in the HTTP headers.

Each header have been carfully examied by following best practices from these source:

- https://cheatsheetseries.owasp.org/cheatsheets/REST_Security_Cheat_Sheet.html
- https://owasp.org/www-project-secure-headers/

This feature is protected with feature flag named 'stripHeadersOnAPI'.


gardleopard (approved):
LGTM! Nice work!

PR #5610, opened 2023-12-12
fix: return 400 on invalid POST data to project access endpoint

thomasheartman:
This PR fixes the issue discussed in SR-234, where you would get a 200 OK response even if your POST request to `/api/admin/projects/<project-name>/access` contains invalid data (and nothing is persisted).

### DIFF
--- a/src/lib/db/access-store.test.ts
+++ b/src/lib/db/access-store.test.ts
@@ -119,15 +119,14 @@ describe('addAccessToProject', () => {
                 [property]: [123456789],
             };
 
-            await expect(
-                async () =>
-                    await access.addAccessToProject(
-                        payload.roles,
-                        payload.groups,
-                        payload.users,
-                        'projectId',
-                        'createdBy',
-                    ),
+            await expect(() =>
+                access.addAccessToProject(
+                    payload.roles,
+                    payload.groups,
+                    payload.users,
+                    'projectId',
+                    'createdBy',
+                ),
             ).rejects.toThrow(BadDataError);
         },
     );
--- a/src/lib/db/access-store.test.ts
+++ b/src/lib/db/access-store.test.ts
@@ -131,31 +131,4 @@ describe('addAccessToProject', () => {
             ).rejects.toThrow(BadDataError);
         },
     );
-
-    test.each(['groups', 'users'])(
-        'If the request contains invalid data for %s, nothing is inserted',
-        (property) => {
-            const access = db.stores.accessStore as AccessStore;
-
-            const payload = {
-                roles: [4],
-                groups: [1],
-                users: [1],
-                [property]: [123456789],
-            };
-
-            expect(
-                async () =>
-                    await access.addAccessToProject(
-                        payload.roles,
-                        payload.groups,
-                        payload.users,
-                        'projectId',
-                        'createdBy',
-                    ),
-            ).rejects.toThrow(BadDataError);
-
-            // test that nothing has been inserted
-        },
-    );
 });
--- a/src/lib/db/access-store.test.ts
+++ b/src/lib/db/access-store.test.ts
@@ -109,17 +109,17 @@ test.each([
 describe('addAccessToProject', () => {
     test.each(['roles', 'groups', 'users'])(
         'should throw a bad data error if there is invalid data in the %s property of the addAccessToProject payload',
-        (property) => {
+        async (property) => {
             const access = db.stores.accessStore as AccessStore;
 
             const payload = {
-                roles: [],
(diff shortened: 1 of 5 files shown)


src/lib/db/access-store.ts
```diff
@@ -636,17 +649,54 @@ export class AccessStore implements IAccessStore {
...
+                        }
+                    });
             }
             if (userRows.length > 0) {
                 await tx(T.ROLE_USER)
                     .insert(userRows)
                     .onConflict(['project', 'role_id', 'user_id'])
-                    .merge();
+                    .merge()
+                    .catch((err) => {
```
kwasniew, line 677: I'm wondering if we handle those types of scenarios consistently across our codebase. Another option would be for the service layer to check if user ID exists before calling a store. On the other handle it would require extra DB call prior to calling the store. With you solution we make 1 DB operation and catch FK violations.

src/lib/db/access-store.test.ts
```diff
@@ -104,3 +105,30 @@ test.each([
...
+            const access = db.stores.accessStore as AccessStore;
+
+            const payload = {
+                roles: [4, 5],
+                groups: [],
+                users: [],
+                [property]: [123456789],
+            };
+
+            await expect(
```
kwasniew, line 122:
```
expect(() =>
    access.addAccessToProject(
        payload.roles,
        payload.groups,
        payload.users,
        'projectId',
        'createdBy',
    )
).rejects.toThrow(BadDataError);
```

kwasniew: no need for await when you return a promise

kwasniew (approved):

src/test/e2e/services/access-service.e2e.test.ts
```diff
@@ -1382,13 +1383,15 @@ test('calling add access with invalid project role ids should not assign those r
...
     const adminRootRole = await accessService.getRoleByName(RoleName.ADMIN);
 
-    accessService.addAccessToProject(
-        [adminRootRole.id, 9999],
-        [],
-        [emptyUser.id],
-        projectName,
-        'some-admin-user',
-    );
+    expect(async () =>
```
kwasniew, line 1386: missing `await expect`

kwasniew (changes requested):
missing await expect

src/lib/services/access-service.test.ts
```diff
@@ -265,3 +266,19 @@ test('throws error when trying to delete a project role in use by group', async
         );
     }
 });
+
+describe('addAccessToProject', () => {
+    test('should throw an error when you try add access with an empty list of roles', () => {
+        const { accessService } = getSetup();
+        expect(
```
kwasniew, line 273: we don't await the whole expect

src/lib/db/access-store.ts
thomasheartman: Yeah, I thought about that, but I didn't want to start adding extra dependencies for that service for this. If we think it's preferable, then that's a good way to go, though.

src/lib/db/access-store.test.ts
thomasheartman: Mayyyybe. I'll try. I had some issues where I needed some of them, but maybe I added more than I need. I'll have a look.

src/lib/services/access-service.test.ts
thomasheartman: Yes! Should be handled in 0408cd10c3eb6f5a9f9fa2ad367692ab0b60c7c2

src/lib/db/access-store.test.ts
thomasheartman: Yup, looks right. Done in 0408cd10c3eb6f5a9f9fa2ad367692ab0b60c7c2

src/lib/db/access-store.ts
thomasheartman: So in essence: if you think that's better, I'm happy to do that. With the rest of this in place, it should be pretty simple. And we could just move this test to the service layer instead.

Thoughts?

kwasniew: I'm fine with the current approach

src/test/e2e/services/access-service.e2e.test.ts
kwasniew, line 1386: await expect(() => ...)

thomasheartman: done in d8851f8fcd7174d4c08f4aa71804dda39c3bacbe

kwasniew (approved):

PR #5617, opened 2023-12-12
docs: Compare front-end api to edge, not proxy

gardleopard:
Proxy is superseeded by edge, lets compare to edge instead of the old proxy.

### DIFF
--- a/website/docs/reference/front-end-api.md
+++ b/website/docs/reference/front-end-api.md
@@ -8,7 +8,7 @@ The Unleash front-end API was released in Unleash 4.18. You can read more in the
 
 :::
 
-The **Unleash front-end API** offers a simplified workflow for connecting a client-side (front-end) applications to Unleash. It provides the exact same API as [Unleash edge](https://docs.getunleash.io/reference/unleash-edge) and  [Unleash proxy - deprecated](../generated/unleash-proxy.md). The front-end API is a quick and easy way to add Unleash to single-page applications and mobile apps.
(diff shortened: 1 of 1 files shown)


kwasniew (approved):
We could also mention both Edge and Proxy in comparison. But if we're deprecating proxy then Edge is better to stand on its own.

chriswk (approved):
Looks nice.

gardleopard:
> We could also mention both Edge and Proxy in comparison. But if we're deprecating proxy then Edge is better to stand on its own.

We have that comparison here https://docs.getunleash.io/understanding-unleash/proxy-hosting

website/docs/reference/front-end-api.md
```diff
@@ -8,20 +8,19 @@ The Unleash front-end API was released in Unleash 4.18. You can read more in the
 
 :::
 
-The **Unleash front-end API** offers a simplified workflow for connecting a client-side (front-end) applications to Unleash. It provides the exact same API as the [Unleash proxy](../generated/unleash-proxy.md). The front-end API is a quick and easy way to add Unleash to single-page applications and mobile apps.
+The **Unleash front-end API** offers a simplified workflow for connecting a client-side (front-end) applications to Unleash. It provides the exact same API as [Unleash edge](https://docs.getunleash.io/reference/unleash-edge) and  [Unleash proxy - deprecated](../generated/unleash-proxy.md). The front-end API is a quick and easy way to add Unleash to single-page applications and mobile apps.
```
thomasheartman, line 11:
```suggestion
The **Unleash front-end API** offers a simplified workflow for connecting a client-side (front-end) applications to Unleash. It provides the exact same API as [Unleash edge](https://docs.getunleash.io/reference/unleash-edge) and  the [Unleash proxy - deprecated](../generated/unleash-proxy.md). The front-end API is a quick and easy way to add Unleash to single-page applications and mobile apps.
```

thomasheartman (approved):
Nice one! Just a minor suggestion to add a definite article in front of a proxy mention, but all good 😄

PR #5612, opened 2023-12-12
chore:system user and events created by userid migrations

daveleek:
## About the changes

Migrations for:
- Adds column created_by_user_id to events
- Adds column is_system to users
- Inserts unleash_system_user id -1337 to users 

includes `is_system: false` in the activeUsers where filter

Tested by running:
`
select * into users_pre_check from users where id > -1;
delete from users where id > -1;
`
before starting unleash, then inspecting users table after unleash has started and verifying that an 'admin' user has been created.

PR #5619, opened 2023-12-12
feat: adds created_by_user_id to all events

chriswk:
### What
Adds `createdByUserId` to all events exposed by unleash. In addition this PR updates all tests and usages of the methods in this codebase to include the required number.

### Dependencies
* https://github.com/Unleash/unleash/pull/5612 - needs to be merged.

### Points of discussion
This will break enterprise hard when merged, since all events constructed from enterprise codebase will not have the required `createdByUserId` field. So, we'll need a planned merge, tomorrow (Wednesday) I'll prepare a PR on enterprise using this as a linked dependency to make sure I update all calls in enterprise as well.

That should allow us to merge this, and then merge the PR on enterprise to reduce the broken build time to a minimum.

### DIFF
--- a/src/lib/addons/datadog.test.ts
+++ b/src/lib/addons/datadog.test.ts
@@ -45,6 +45,7 @@ test('Should call datadog webhook', async () => {
         createdAt: new Date(),
         type: FEATURE_CREATED,
         createdBy: 'some@user.com',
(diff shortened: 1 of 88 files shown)


daveleek (approved):
LG! Big one 😅

PR #5623, opened 2023-12-13
test: scrollable filter options and sticky search

kwasniew:
## About the changes

Search always stays on top. Filter options are scrollable. The entire filter item component takes 70% of viewport height so that it fits between the filter and the footer without obscuring anything
Screenshot 2023-12-13 at 08 46 39


sjaanus (approved):
LG

Tymek (approved):

PR #5625, opened 2023-12-13
refactor: filter abstraction

sjaanus:
1. Created Filters component that can be used with any table.
2. Moved bunch of files into same folder


frontend/src/component/filter/Filters.tsx
```diff
@@ -0,0 +1,147 @@
...
+    pluralOperators: [string, ...string[]];
+};
+
+type IDateFilterItem = IBaseFilterItem & {
+    dateOperators: [string, ...string[]];
+};
+
+export type IFilterItem = ITextFilterItem | IDateFilterItem;
+
+export const Filters: VFC<IFilterProps> = ({
```
sjaanus, line 46: This is single reusable component that can be used with every table.

kwasniew: if it's gonna be used a lot we need a test. Is that gonna be a followup PR?

kwasniew (approved):

frontend/src/component/filter/Filters.tsx
sjaanus: Yes

PR #5626, opened 2023-12-13
refactor: project overview using filters component

sjaanus:
1. This is second table using the **Filters** component

Looks neat


kwasniew (approved):
Nice

PR #5629, opened 2023-12-13
fix: optimize cores headers

ivarconr:
This commit enhances two aspects of CORS:

- Always support CORS preflight requests.
- Do not add additional secuity headers for prefligh calls.


FredrikOseberg (approved):
LGTM

PR #5633, opened 2023-12-13
fix: avoid duplicate unselected filters

kwasniew:
## About the changes

The unselected list should never allow for any elements that are already selected
Screenshot 2023-12-13 at 12 47 54

This will need to be tested as part of another PR that adds tests for this component

### DIFF
--- a/frontend/src/component/filter/Filters.tsx
+++ b/frontend/src/component/filter/Filters.tsx
@@ -63,29 +63,31 @@ export const Filters: VFC<IFilterProps> = ({
         firstArray: string[],
         secondArray: string[],
     ): string[] => {
+        const resultArray: string[] = [...firstArray];
         const elementsSet = new Set(firstArray);
 
         secondArray.forEach((element) => {
             if (!elementsSet.has(element)) {
-                firstArray.push(element);
+                resultArray.push(element);
             }
         });
 
-        return firstArray;
+        return resultArray;
     };
 
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


frontend/src/component/filter/Filters.tsx
```diff
@@ -63,29 +63,32 @@ export const Filters: VFC<IFilterProps> = ({
         firstArray: string[],
         secondArray: string[],
     ): string[] => {
+        const resultArray: string[] = [...firstArray];
```
kwasniew, line 66: added this to avoid mutating arguments that are passed here

frontend/src/component/filter/Filters.tsx
```diff
@@ -63,29 +63,32 @@ export const Filters: VFC<IFilterProps> = ({
...
             .map((field) => field.label);
+        const allSelectedFilters = mergeArraysKeepingOrder(
+            selectedFilters,
+            newSelectedFilters,
+        );
+        setSelectedFilters(allSelectedFilters);
+
         const newUnselectedFilters = availableFilters
-            .filter((field) => !state[field.filterKey])
+            .filter((item) => !allSelectedFilters.includes(item.label))
```
kwasniew, line 89: exclude based on available and selected items, not on the URL state

andreas-unleash (approved):
Nice!

sjaanus (approved):
LG

PR #5634, opened 2023-12-13
fix: last seen now will not break

sjaanus:


frontend/src/component/project/Project/ProjectFeatureToggles/PaginatedProjectFeatureToggles.tsx
```diff
@@ -350,9 +350,9 @@ export const PaginatedProjectFeatureToggles = ({
                                 featureEnvironment?.name === env.environment,
                         );
                         return [
-                            typeof env === 'string' ? env : env.environment,
```
sjaanus, line 353: env is never string, its complex object

andreas-unleash (approved):

PR #5636, opened 2023-12-13
test: filter selection avoid duplicates

kwasniew:
## About the changes

Makes sure that when we select filter items they disappear from the the selection list.
Screenshot 2023-12-13 at 15 49 43


sjaanus (approved):
Test looks fine, but I am not sure this is how the duplication appeared in Nicos example.

"Reproduce by selecting empty Project, Then select state with value, then project reappears in list."

PR #5642, opened 2023-12-14
refactor: move search implementation out of strategies store

sjaanus:
This is first step of refactoring. Next steps follow with possibly a query builder, or atleast using some reusable methods.


kwasniew (approved):
I'm assuming it's only moving stuff around

sjaanus:
> I'm assuming it's only moving stuff around

Yes and adding timers.

PR #5644, opened 2023-12-14
feat: pagination reset on feature list

kwasniew:
## About the changes

We should reset pagination on: search, filter, favorite first, sort order change

We need to make the same change on the other pages where we do new tables

### DIFF
--- a/frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
+++ b/frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
@@ -92,6 +92,9 @@ export const FeatureToggleListTable: VFC = () => {
         'features-list-table',
         stateConfig,
     );
+    // offset needs to be first so we can override it
+    const setTableStateWithOffsetReset: typeof setTableState = (data) =>
+        setTableState({ offset: 0, ...data });
 
     const filterState = {
         project: tableState.project,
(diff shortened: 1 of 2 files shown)


frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
```diff
@@ -92,6 +92,9 @@ export const FeatureToggleListTable: VFC = () => {
         'features-list-table',
         stateConfig,
     );
+    // offset needs to be first so we can override it
+    const setTableStateWithOffsetReset: typeof setTableState = (data) =>
```
kwasniew, line 96: this is a new method we use instead of the low level setTableState to never forget about the offset reset

sjaanus (dismissed):

frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
sjaanus: Can we define this method somewhere near state, so we do not have to rewrite it in every component, but just take it from our hook.

kwasniew: we need config definition for this to work and autocomplete the offset field

sjaanus (approved):

PR #5645, opened 2023-12-14
chore: enable manual testing of migrations

gastonfournier:
## About the changes
This adds a Makefile to make it easy to test migrations from one version of Unleash to another. 

The script depends on [docker compose V2](https://docs.docker.com/compose/migrate/)

**Before starting**: make sure you're inside test-migrations folder and run `make clean` to be in a clean state.

We can run 2 versions of Unleash side by side with a shared database (the second version will apply migrations to the DB):
```shell
UNLEASH_DOCKER_IMAGE=unleashorg/unleash-server:5.6.10 make start-unleash # defaults to port 4242
UNLEASH_DOCKER_IMAGE=unleashorg/unleash-server:latest make start-another-unleash # defaults to port 4243
make test # run basic UI tests against port 4242 (first image)
EXPOSED_PORT=4243 make test # run basic UI tests against port 4243
```

This also enables us to test our local repository with our code of Unleash server running at port 4244 (`EXPOSE_PORT=4444 make run-current` if you want to change it):
```shell
UNLEASH_DOCKER_IMAGE=unleashorg/unleash-server:5.6.10 make start-unleash # defaults to port 4242
make run-current # exposes the current backend at 4244
```

You can also connect the latest UI to any of the ports specified above, starting the UI at port 3000:
```shell
EXPOSED_PORT=4242 make run-current-ui # exposed port defaults to 4244 which is the port of the current backend
```

### DIFF
--- a/src/server-dev.ts
+++ b/src/server-dev.ts
@@ -7,16 +7,19 @@ process.nextTick(async () => {
     try {
         await start(
             createConfig({
-                db: {
-                    user: 'unleash_user',
-                    password: 'password',
-                    host: 'localhost',
-                    port: 5432,
-                    database: process.env.UNLEASH_DATABASE_NAME || 'unleash',
-                    schema: process.env.UNLEASH_DATABASE_SCHEMA,
-                    ssl: false,
-                    applicationName: 'unleash',
-                },
+                db: process.env.DATABASE_URL
+                    ? undefined
+                    : {
+                          user: 'unleash_user',
+                          password: 'password',
+                          host: 'localhost',
+                          port: 5432,
+                          database:
+                              process.env.UNLEASH_DATABASE_NAME || 'unleash',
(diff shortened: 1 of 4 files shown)


chriswk (approved):
LGTM

src/server-dev.ts
```diff
@@ -7,16 +7,19 @@ process.nextTick(async () => {
...
-                    password: 'password',
-                    host: 'localhost',
-                    port: 5432,
-                    database: process.env.UNLEASH_DATABASE_NAME || 'unleash',
-                    schema: process.env.UNLEASH_DATABASE_SCHEMA,
-                    ssl: false,
-                    applicationName: 'unleash',
-                },
+                db: process.env.DATABASE_URL
+                    ? undefined
```
gastonfournier, line 11: This way we can send DATABASE_URL instead of what we're doing currently. The problem with our current solution is that this db object always takes precedence

PR #5650, opened 2023-12-15
feat: reset persistent table offset on change

kwasniew:
## About the changes

setTableState resets offset on every change internally so that if works automatically for all pages using the hook

### DIFF
--- a/frontend/src/component/filter/AddFilterButton.tsx
+++ b/frontend/src/component/filter/AddFilterButton.tsx
@@ -33,7 +33,7 @@ interface IAddFilterButtonProps {
     availableFilters: IFilterItem[];
 }
 
-const AddFilterButton = ({
+export const AddFilterButton = ({
     visibleOptions,
     setVisibleOptions,
     hiddenOptions,
@@ -87,5 +87,3 @@ const AddFilterButton = ({
         </div>
     );
 };
-
-export default AddFilterButton;
--- a/frontend/src/hooks/usePersistentTableState.ts
+++ b/frontend/src/hooks/usePersistentTableState.ts
@@ -1,7 +1,7 @@
 import { useEffect } from 'react';
 import { useSearchParams } from 'react-router-dom';
 import { createLocalStorage } from 'utils/createLocalStorage';
-import { useQueryParams, encodeQueryParams } from 'use-query-params';
+import { encodeQueryParams, useQueryParams } from 'use-query-params';
 import { QueryParamConfigMap } from 'serialize-query-params/src/types';
 
 const usePersistentSearchParams = <T extends QueryParamConfigMap>(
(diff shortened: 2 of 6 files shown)


frontend/src/component/filter/AddFilterButton.tsx
```diff
@@ -87,5 +87,3 @@ const AddFilterButton = ({
         </div>
     );
 };
-
-export default AddFilterButton;
```
kwasniew, line 91: sticking to our ADR of not using default exports

frontend/src/hooks/usePersistentTableState.ts
```diff
@@ -39,7 +39,38 @@ export const usePersistentTableState = <T extends QueryParamConfigMap>(
...
+    const [tableState, setTableStateInternal] = useQueryParams(
+        queryParamsDefinition,
+    );
+
+    type SetTableStateInternalParam = Parameters<
+        typeof setTableStateInternal
+    >[0];
+
+    const setTableState = (newState: SetTableStateInternalParam) => {
+        if (!queryParamsDefinition.offset) {
```
kwasniew, line 51: if no decoder provided delegate to original function

frontend/src/hooks/usePersistentTableState.ts
```diff
@@ -39,7 +39,41 @@ export const usePersistentTableState = <T extends QueryParamConfigMap>(
...
-    const [tableState, setTableState] = useQueryParams(queryParamsDefinition);
+    const [tableState, setTableStateInternal] = useQueryParams(
+        queryParamsDefinition,
+    );
+
+    type SetTableStateInternalParam = Parameters<
+        typeof setTableStateInternal
+    >[0];
+
+    const setTableState = useCallback(
```
kwasniew, line 50: make the function identity stable

frontend/src/hooks/usePersistentTableState.ts
```diff
@@ -39,7 +39,41 @@ export const usePersistentTableState = <T extends QueryParamConfigMap>(
...
+        (newState: SetTableStateInternalParam) => {
+            if (!queryParamsDefinition.offset) {
```
kwasniew, line 52: when no URL decoder for offset do nothing

frontend/src/hooks/usePersistentTableState.ts
```diff
@@ -39,7 +39,41 @@ export const usePersistentTableState = <T extends QueryParamConfigMap>(
...
+                return setTableStateInternal(newState);
+            }
+            if (typeof newState === 'function') {
```
kwasniew, line 55: this is sth we don't use but the original library allows for passing fn instead of new state

frontend/src/hooks/usePersistentTableState.ts
```diff
@@ -39,7 +39,41 @@ export const usePersistentTableState = <T extends QueryParamConfigMap>(
...
+                setTableStateInternal((prevState) => {
+                    const updatedState = (newState as Function)(prevState);
+                    return queryParamsDefinition.offset
+                        ? {
+                              offset: queryParamsDefinition.offset.decode('0'),
+                              ...updatedState,
+                          }
+                        : updatedState;
+                });
+            } else {
```
kwasniew, line 65: this is the case we use most of the time decoding '0' with NumberDecoder

frontend/src/hooks/usePersistentTableState.ts
```diff
@@ -39,7 +39,41 @@ export const usePersistentTableState = <T extends QueryParamConfigMap>(
...
+                const updatedState = queryParamsDefinition.offset
+                    ? {
+                          offset: queryParamsDefinition.offset.decode('0'),
```
kwasniew, line 68: offset is first so that we can overwrite offset for prev/next page in the newState

sjaanus (approved):
LG

PR #5654, opened 2023-12-15
fix: reducing of features will not break order anymore

sjaanus:


src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -295,34 +295,43 @@ class FeatureSearchStore implements IFeatureSearchStore {
...
-                    this.addTag(acc[row.feature_name], row);
+                if (
+                    !existingEntry.lastSeenAt ||
+                    new Date(row.env_last_seen_at) >
+                        new Date(existingEntry.last_seen_at)
+                ) {
+                    existingEntry.lastSeenAt = row.env_last_seen_at;
                 }
             } else {
-                acc[row.feature_name] = {
```
sjaanus, line 325: This was the issue. If the key is number, it will be put as first element which breaks the ordering. This means, we need to use actual array instead.

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -295,34 +295,16 @@ class FeatureSearchStore implements IFeatureSearchStore {
...
-                        existingEnvironment.name === row.environment,
-                );
-                if (!environmentExists) {
-                    acc[row.feature_name].environments.push(
-                        FeatureSearchStore.getEnvironment(row),
-                    );
-                }
+    getAggregatedSearchData(rows): IFeatureOverview[] {
+        const entriesMap: Map<string, IFeatureOverview> = new Map();
+        const orderedEntries: IFeatureOverview[] = [];
```
sjaanus, line 300: Keeping the order here.

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -295,34 +295,16 @@ class FeatureSearchStore implements IFeatureSearchStore {
...
-                    (existingEnvironment) =>
-                        existingEnvironment.name === row.environment,
-                );
-                if (!environmentExists) {
-                    acc[row.feature_name].environments.push(
-                        FeatureSearchStore.getEnvironment(row),
-                    );
-                }
+    getAggregatedSearchData(rows): IFeatureOverview[] {
+        const entriesMap: Map<string, IFeatureOverview> = new Map();
```
sjaanus, line 299: Keeping entries here for fast access with complexity O(1).

src/lib/types/model.ts
```diff
@@ -205,6 +205,11 @@ export interface IEnvironmentOverview extends IEnvironmentBase {
 
 export interface IFeatureOverview {
     name: string;
+    description: string;
```
sjaanus, line 208: Fixed type

kwasniew (approved):

PR #5657, opened 2023-12-15
React Tutorial Improvements

nnennandukwe:
## About the changes

We've made some first round updates to the React tutorial:

- making it more SEO-friendly with ordered lists, sequential language, description of the JS library
- Switched over the demo app to point to an open source project: [Cypress Real World App](https://github.com/cypress-io/cypress-realworld-app)
- included best practice considerations for client-side development
- updated URL path to point to `/feature-flag-tutorials/react` for simplification

## Discussion points

Would love feedback on if there's a need for more screenshots? Don't want to be too screenshot-heavy though I imagine.
And need feedback on the descriptions of "Considerations for using feature flags in react" https://github.com/Unleash/unleash/compare/react-improvements?expand=1#diff-96d4956f49f80cd76489a72d4d88c2956ce9dcc695f66fe014ad1185e37cb589R21

Want to make sure that what I described makes sense or if it could use some tweaking to convey the right message clearly.


dgorton:
From our verbal conversation: 

1. Link to React official site in the first few sentences
2. Add a few sentences about Cypress and their app to the intro, along w/ a screenshot of what that looks like. Let’s link to the project, say what it is and why that’s cool and also link to the repo.
3. Directly link the two “Considerations for using …” to the relevant principles. Might involve some tweaks to the sentences so it reads naturally, or might be possible to link as-is: … could potentially expose sensitive data such as API keys … (and make the bold the link)
4. There might actually be too many considerations. Security, performance, payloads, … It might be too much. I really like this section - not sure about that - but I want to make sure we aren’t overwhelming someone with too much up front.
5. Add some more screenshots to the end, showing what turning the flag on/off looks like in Unleash and what that looks like in the app.
6. Add a sentence or two about gradual rollouts to the end, talking about how that’s a common use-case in the real world rather than just ‘on’ vs ‘off’ and link to that place in the docs. (activation-strategies looks like the right link)
7. Let’s add a sentence to the conclusion about how they also maybe got to learn about another cool open source project that they might want to use in their React apps.
8. Oh - and we should probably mention that we use flag and toggle as synonyms. Here’s [an example](https://unleash-docs-git-react-improvements-unleash-team.vercel.app/how-to/how-to-create-feature-toggles) of words I’ve used elsewhere to explain that:

This document uses feature flags and feature toggles interchangeably. Some people prefer flag; others prefer toggle. We use both - they are synonyms for us.

I don’t know if those are the right words here and I definitely don’t want them at the top of the document - but maybe the first time we mention toggles we should add a sentence or two like that.

website/docusaurus.config.js
```diff
@@ -544,7 +544,7 @@ module.exports = {
                     },
                     {
                         from: '/topics/feature-flags/tutorials/react/implementing-feature-flags',
-                        to: '/feature-flag-tutorials/react/implementing-feature-flags',
+                        to: '/feature-flag-tutorials/react/',
```
thomasheartman, line 547: Did the previous `to` URL exist? If so, we should probably redirect from that one too. Something like this:

```suggestion
                        from: [ '/topics/feature-flags/tutorials/react/implementing-feature-flags', '/feature-flag-tutorials/react/implementing-feature-flags'],
                        to: '/feature-flag-tutorials/react/',
```

Not sure why github is saying everything has changed, though 🤔

dgorton: Good catch. I think the only redirect we need is from 
`/feature-flag-tutorials/react/implementing-feature-flags` to `/feature-flag-tutorials/react`

I don't think `/topics/feature-flags/tutorials/react/implementing-feature-flags` ever existed - no need to redirect from that.

Not sure about the trailing slash (`react` vs `react/`) or not - something to be tested.

dgorton (approved):
Looks great! :D

PR #5658, opened 2023-12-15
feat: filter chips icons

kwasniew:
## About the changes

Using different filter chips icons according to selection
Screenshot 2023-12-15 at 14 36 02

Instead of passing icon string through layers I pass ReactNode label


sjaanus (approved):
LG

ivarconr:
This looks great!

PR #5660, opened 2023-12-15
feat: show failure in cr overview

andreas-unleash:
Show failure reason in change request overview

Closes [1-1769](https://linear.app/unleash/issue/1-1769/add-reason-and-icon-to-change-request-overview)
Screenshot 2023-12-15 at 10 37 03

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.test.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.test.tsx
@@ -13,6 +13,7 @@ const server = testServerSetup();
 const mockChangeRequest = (
     featureName: string,
     state: ChangeRequestState,
+    failureReason?: string,
 ): IChangeRequest => {
     const result: IChangeRequest = {
         id: 1,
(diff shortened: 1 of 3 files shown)


ivarconr:
This is cool. Will the failure reason usually be more informative than just "conflicts"? Is it possible for the user to understand where she gets conflicts?

andreas-unleash:
For now, we just say ‘conflicts’ or ‘a system error’ if it gets to this
point (the schedule failing).
  However we do show in the UI, when you are about to archive a toggle or
update/delete a feature strategy that would create a conflict - which
change requests will affect.
We also send out an email to the creator and scheduler of the cr that a
conflicting action has been taken.
So there are 2-3 guard rails before it reaches this point

kwasniew (approved):

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.test.tsx
```diff
@@ -65,6 +66,10 @@ const mockChangeRequest = (
         };
     }
 
+    if (failureReason) {
+        result.schedule!.failureReason = failureReason;
+    }
```
FredrikOseberg, line 71: Why is this rewriting needed?

PR #5662, opened 2023-12-16
fix: use node v18.19.0

ivarconr:


chriswk (approved):
I'm assuming you mean 18.19, not 18.9 :+1:

PR #5664, opened 2023-12-18
fix: last seen now sorts nulls last

sjaanus:
Two changes were needed to sort better

1. Since we are still using `last seen` from `features` table for backwards compatibility, we needed to add it to sort condition.
2. Nulls break the order, so now sorting nulls as last.

### DIFF
--- a/src/lib/features/feature-search/feature-search-store.ts
+++ b/src/lib/features/feature-search/feature-search-store.ts
@@ -227,7 +227,6 @@ class FeatureSearchStore implements IFeatureSearchStore {
                 const sortByMapping = {
                     name: 'features.name',
                     type: 'features.type',
-                    lastSeenAt: lastSeenQuery,
                     stale: 'features.stale',
                 };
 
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -476,6 +479,20 @@ test('should sort features', async () => {
         ],
         total: 3,
     });
+
+    const { body: lastSeenDescSort } = await sortFeatures({
+        sortBy: 'lastSeenAt',
+        sortOrder: 'asc',
```
kwasniew, line 485: this is asc and on line 483 it says desc sort

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -279,6 +284,7 @@ class FeatureSearchStore implements IFeatureSearchStore {
             .joinRaw('CROSS JOIN total_features')
             .whereBetween('final_rank', [offset + 1, offset + limit]);
 
+        console.log(finalQuery.toQuery());
```
kwasniew, line 287: console leftover

kwasniew (approved):

PR #5669, opened 2023-12-18
fix: increase page size in backend to 100

sjaanus:


kwasniew (approved):

src/lib/features/feature-search/feature-search-controller.ts
```diff
@@ -105,7 +105,7 @@ export default class FeatureSearchController extends Controller {
                         ['enabled', 'disabled'].includes(tag[1]),
                 );
             const normalizedLimit =
-                Number(limit) > 0 && Number(limit) <= 50 ? Number(limit) : 50;
+                Number(limit) > 0 && Number(limit) <= 50 ? Number(limit) : 100;
```
kwasniew, line 108: shouldn't we have <= 100

PR #5670, opened 2023-12-18
chore: incoming webhooks and tokens migration

nunogois:
https://linear.app/unleash/issue/2-1699/db-create-migration-for-a-new-incoming-webhooks-table

Adds a migration that creates 2 tables:
 - `incoming_webhooks`
 - `incoming_webhook_tokens`

### DIFF
--- a/src/migrations/20231215105713-incoming-webhooks.js
+++ b/src/migrations/20231215105713-incoming-webhooks.js
@@ -1,0 +1,41 @@
+'use strict';
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS incoming_webhooks
+            (
+                id SERIAL PRIMARY KEY NOT NULL,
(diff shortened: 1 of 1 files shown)


chriswk (commented):
I'm not sure we want to cascade here. 

Unless the thought is that incoming webhooks are strongly bound to the user that created them?

I see both pros and contras here. 
Pros:
- explicit control over what webhooks are created by whom
- automatic cleanup when users are removed from the system

Cons:
- automatic cleanup when users are removed. This is a double-edged sword, and the one that worries me the most

chriswk:
Ah, I forgot about the soft delete thing. May I also suggest an index on the created_by_user_id?

chriswk:
Never mind, I see they are in place already.

chriswk:
You don't need `use strict;` btw, so feel free to remove it :)

gastonfournier (approved):
LGTM!

PR #5673, opened 2023-12-18
feat: up and down arrow navigation for filter items

kwasniew:
## About the changes

Navigation with up and down arrow in addition to tab that we supported before. 

https://github.com/Unleash/unleash/assets/1394682/bf96059f-22ad-433a-b864-3a809bed49c2

### DIFF
--- a/frontend/src/component/filter/FilterItem/FilterItem.tsx
+++ b/frontend/src/component/filter/FilterItem/FilterItem.tsx
@@ -1,15 +1,8 @@
 import { Search } from '@mui/icons-material';
 import { Box, InputAdornment, List, ListItemText } from '@mui/material';
 import { FC, ReactNode, useEffect, useRef, useState } from 'react';
-import {
-    StyledCheckbox,
-    StyledDropdown,
-    StyledListItem,
-    StyledPopover,
-    StyledTextField,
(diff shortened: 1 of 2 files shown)


frontend/src/component/filter/FilterItem/FilterItem.tsx
```diff
@@ -27,6 +20,34 @@ export type FilterItemParams = {
     values: string[];
 };
 
+
+interface UseSelectionManagementProps {
+    options: Array<{ label: string; value: string }>;
+    handleToggle: (value: string) => () => void;
+}
+
+const useSelectionManagement = ({ options, handleToggle }: UseSelectionManagementProps) => {
```
kwasniew: extracted a hook to avoid polluting our main component

frontend/src/component/filter/FilterItem/FilterItem.tsx
```diff
@@ -144,6 +170,10 @@ export const FilterItem: FC<IFilterItemProps> = ({
                                 </InputAdornment>
                             ),
                         }}
+                        inputRef={(el) => {
+                            listRefs.current[0] = el;
```
kwasniew, line 183: search is treated as the first navigation list item

PR #5678, opened 2023-12-18
chore: use semver compatible version

gastonfournier:
## About the changes
Replicate the change made to respect semver


daveleek (approved):
LG, thanks for the walk through and explanation

PR #5679, opened 2023-12-18
chore: rename secret to token in incoming_webhook_tokens

nunogois:
Renames `secret` to `token` in `incoming_webhook_tokens` for consistency. This table is not being used yet, so this should be a very safe change.


gastonfournier (approved):
LG!

PR #5685, opened 2023-12-19
fix: empty paginated table state

Tymek:
Table placeholder should be visible if search query returns no data.


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/TableEmptyState/TableEmptyState.tsx
```diff
@@ -3,12 +3,15 @@ import { Box } from '@mui/material';
...
 
 interface ITablePlaceholderProps {
-    total?: number;
+    show: boolean;
     query?: string;
 }
 
-export const Placeholder: FC<ITablePlaceholderProps> = ({ total, query }) => {
-    if (total !== 0) {
+export const TableEmptyState: FC<ITablePlaceholderProps> = ({
```
kwasniew, line 10: can the parent decide whether to show this component or not instead of the component looking at a boolean param?

kwasniew (commented):
I'd not add show param. Also please update the placeholder text and remove "Get started by adding a new feature toggle." because people may already have toggles and filter just removes them from the list

kwasniew (approved):

PR #5686, opened 2023-12-19
fix: prevent banner from crashing with invalid variant

nunogois:
https://linear.app/unleash/issue/2-1760/prevent-banner-from-crashing-when-set-with-an-invalid-variant

This prevents the banner from crashing when set with an invalid variant. Instead, it default its styles to the default variant, which is `info`.

Also adds tests to validate our assumptions.


gastonfournier (approved):
LGTM!

PR #5690, opened 2023-12-19
fix: remove extra call to features on project load

sjaanus:
Currently EnableEnvironmentDialog was loaded even if no feature was touched. Now it will only load, if feature toggle was selected.

### DIFF
--- a/frontend/src/component/project/Project/ProjectFeatureToggles/FeatureToggleSwitch/useFeatureToggleSwitch.tsx
+++ b/frontend/src/component/project/Project/ProjectFeatureToggles/FeatureToggleSwitch/useFeatureToggleSwitch.tsx
@@ -14,6 +14,7 @@ import {
     OnFeatureToggleSwitchArgs,
     UseFeatureToggleSwitchType,
 } from './FeatureToggleSwitch.types';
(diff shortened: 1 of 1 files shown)


kwasniew (approved):

frontend/src/component/project/Project/ProjectFeatureToggles/FeatureToggleSwitch/useFeatureToggleSwitch.tsx
```diff
@@ -215,7 +216,14 @@ export const useFeatureToggleSwitch: UseFeatureToggleSwitchType = (
     const modals = (
         <>
             <FeatureStrategyProdGuard {...prodGuardModalState} />
-            <EnableEnvironmentDialog {...enableEnvironmentDialogState} />
+            <ConditionallyRender
+                condition={enableEnvironmentDialogState.featureId.length !== 0}
```
kwasniew, line 220: can you add a variable with intention revealing name. It may not be clear why featureId.length is the trigger here

sjaanus: Yes, good

PR #5693, opened 2023-12-19
chore: incoming webhook events

nunogois:
https://linear.app/unleash/issue/2-1748/create-new-incoming-webhook-events

Adds new events for write operations on incoming webhooks and their tokens.

### DIFF
--- a/src/lib/types/events.ts
+++ b/src/lib/types/events.ts
@@ -175,6 +175,17 @@ export const BANNER_CREATED = 'banner-created' as const;
 export const BANNER_UPDATED = 'banner-updated' as const;
 export const BANNER_DELETED = 'banner-deleted' as const;
 
+export const INCOMING_WEBHOOK_CREATED = 'incoming-webhook-created' as const;
+export const INCOMING_WEBHOOK_UPDATED = 'incoming-webhook-updated' as const;
+export const INCOMING_WEBHOOK_DELETED = 'incoming-webhook-deleted' as const;
+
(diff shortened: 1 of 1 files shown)


src/lib/types/events.ts
```diff
@@ -309,6 +320,12 @@ export const IEventTypes = [
     PROJECT_ENVIRONMENT_REMOVED,
     DEFAULT_STRATEGY_UPDATED,
     SEGMENT_IMPORT,
+    INCOMING_WEBHOOK_CREATED,
+    INCOMING_WEBHOOK_UPDATED,
+    INCOMING_WEBHOOK_DELETED,
+    INCOMING_WEBHOOK_TOKEN_CREATED,
+    INCOMING_WEBHOOK_TOKEN_UPDATED,
```
gastonfournier, line 327: Do you see a case where the webhook can be updated? Maybe something like GH tokens where they have an expiration date and you can regenerate the secret?

gastonfournier (approved):
LG! Just a small comment

src/lib/types/events.ts
nunogois: That's a good question. The way I see it:
 - You can update the incoming webhook by updating its enabled state (toggling them)
 - You can *potentially* update the token by updating its name

But no strong feelings about this. I'm also OK with not allowing updating the token names.

gastonfournier: Ok! I think I was not considering the token name as a property we could change

PR #5695, opened 2023-12-20
feat: implement column created_by_user_id in feature_tag

daveleek:
## About the changes

Adds the new nullable column created_by_user_id to the data used by feature-tag-store and feature-tag-service. Also updates openapi schemas.

### DIFF
--- a/src/lib/openapi/spec/feature-tag-schema.ts
+++ b/src/lib/openapi/spec/feature-tag-schema.ts
@@ -35,6 +35,12 @@ export const featureTagSchema = {
             description:
                 'The value of the tag. This property is deprecated and will be removed in a future version of Unleash. Superseded by the `tagValue` property.',
         },
+        createdByUserId: {
+            type: 'number',
+            nullable: true,
+            example: 1,
+            description: 'The id of the user who created this tag',
+        },
     },
     components: {},
 } as const;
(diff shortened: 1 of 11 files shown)


src/lib/openapi/spec/feature-tag-schema.ts
```diff
@@ -35,6 +35,12 @@ export const featureTagSchema = {
             description:
                 'The value of the tag. This property is deprecated and will be removed in a future version of Unleash. Superseded by the `tagValue` property.',
         },
+        createdByUserId: {
+            type: 'number',
+            nullable: true,
```
gastonfournier, line 40: Just a note so we don't forget: maybe we need to make this not nullable, but if this is used as input it might be a breaking change for the API

src/lib/schema/feature-schema.ts
```diff
@@ -159,4 +159,5 @@ export const featureTagSchema = joi.object().keys({
     tagValue: joi.string(),
     type: nameType.optional(),
     value: joi.string().optional(),
+    createdByUserId: joi.number().optional(),
```
gastonfournier, line 162: I think this is used to validate input. And I believe all new input should have a createdById (but maybe it's a breaking change for the API)

gastonfournier (approved):
LGTM! We agreed on Slack that we might be able to find out who was the actor by looking at the events log, so keeping the field null for now seems reasonable, despite not being sure if we're going to embark on that task.

src/lib/openapi/spec/feature-tag-schema.ts
daveleek: Yeah this is part of import/export. So I think it needs to be nullable

PR #5696, opened 2023-12-20
fix: now project overview has skeleton instead of placeholders

sjaanus:
Removed `ref` dependency from `useLoading` hook, it was being overly reactive and breaking skeleton.

### DIFF
--- a/frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
+++ b/frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
@@ -45,6 +45,7 @@ import { FeatureSegmentCell } from 'component/common/Table/cells/FeatureSegmentC
 import { useUiFlag } from 'hooks/useUiFlag';
 import { FeatureToggleListTable as LegacyFeatureToggleListTable } from './LegacyFeatureToggleListTable';
 import { FeatureToggleListActions } from './FeatureToggleListActions/FeatureToggleListActions';
+import useLoading from '../../../hooks/useLoading';
 
 export const featuresPlaceholder = Array(15).fill({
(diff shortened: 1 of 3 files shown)


frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
```diff
@@ -257,7 +259,6 @@ const FeatureToggleListTableComponent: VFC = () => {
 
     return (
         <PageContent
-            isLoading={loading}
```
sjaanus, line 260: Header does not need loading anymore

frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
```diff
@@ -311,7 +312,9 @@ const FeatureToggleListTableComponent: VFC = () => {
                 state={filterState}
             />
             <SearchHighlightProvider value={query || ''}>
-                <PaginatedTable tableInstance={table} totalItems={total} />
+                <div ref={bodyLoadingRef}>
```
sjaanus, line 315: Moving loader only to table.

frontend/src/hooks/useLoading.ts
```diff
@@ -16,7 +16,7 @@ const useLoading = (loading: boolean, selector = '[data-loading=true]') => {
                 }
             });
         }
-    }, [loading, ref]);
+    }, [loading]);
```
sjaanus, line 19: This was the cause, with ref as dependency, it was over reactive and overwriting skeleton, when multiple components were using useLoading.

kwasniew (approved):

PR #5701, opened 2023-12-20
fix: separate concerns for ConstraintAccordionList

FredrikOseberg:
## Problem

The ConstraintAccordionList component was used in multiple places: 
* Playground
* Segment form
* StrategyExecution
* Change requests
* Create strategy
* Edit strategy

This is problematic because some of the views are just pure visual representations, and other views allow you to interact with and edit the constraints. This causes a situation where the visual representation needs to be aware of the implementation details of editing and mutating constraints. In addition the ConstraintAccordionList is not just a pure rendering of the list, it also keeps internal state on when to show the create button and optional headers. This is makes it hard to make changes when stylings need to be subtly different across components. 

## Solution

Taking on the full refactor for this is out of scope, but it's unfortunate that the ConstraintAccordionList needs all this internal state. For now I split out the list into it's own component called ConstraintList. I gathered the functions needed for editing and mutating the constraints in a reusable hook and isolated the version of the list used in the new feature strategy edit / create components into it's own component so that the changes in layout will not affect anything else. 

Ideally we should try to move towards a future where the components don't keep internal state like this but clear boundaries and purposes for the use.

### DIFF
--- a/frontend/src/component/common/ConstraintAccordion/ConstraintAccordionList/ConstraintAccordionList.tsx
+++ b/frontend/src/component/common/ConstraintAccordion/ConstraintAccordionList/ConstraintAccordionList.tsx
@@ -5,8 +5,8 @@ import React, {
     RefObject,
     useImperativeHandle,
 } from 'react';
-import { Box, Button, styled, Tooltip, Typography } from '@mui/material';
-import { Add, HelpOutline } from '@mui/icons-material';
+import { Box, Button, styled, Tooltip } from '@mui/material';
+import { HelpOutline } from '@mui/icons-material';
 import { IConstraint } from 'interfaces/strategy';
 import { ConstraintAccordion } from 'component/common/ConstraintAccordion/ConstraintAccordion';
 import produce from 'immer';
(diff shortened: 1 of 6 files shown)


frontend/src/component/common/ConstraintAccordion/ConstraintAccordionList/ConstraintAccordionList.tsx
```diff
@@ -292,3 +172,78 @@ export const ConstraintAccordionList = forwardRef<
...
+}
+
+export const ConstraintList = forwardRef<
+    IConstraintAccordionListRef | undefined,
+    IConstraintList
+>(({ constraints, setConstraints, state }, ref) => {
+    const { context } = useUnleashContext();
+
+    const onEdit =
+        setConstraints &&
```
FredrikOseberg, line 189: unfortunate that we need this pattern here, but out of scope to change that here

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/FeatureStrategyConstraintAccordionList/FeatureStrategyConstraintAccordionList.tsx
```diff
@@ -0,0 +1,191 @@
+import React, { forwardRef, RefObject } from 'react';
```
FredrikOseberg, line 1: isolated component for the new views

PR #5702, opened 2023-12-20
Feat: allow rescheduling of failed scheduled cr

andreas-unleash:
Allows for rescheduling failed scheduled requests


thomasheartman (approved):
Nice! At first I thought it seemed weird that you should be able to apply a failed scheduled request, but it makes sense. Looks good 😄

PR #5710, opened 2023-12-20
chore: helper methods to extract user data from req

nunogois:
With the recent changes it's common that we'll need both the id and processed username from the auth user in the request, so this PR provides some helper methods to simplify this.


daveleek (approved):
LG! Is extract* there to conform to a specific pattern so we can change things later without changing in every file?

PR #5713, opened 2023-12-21
fix: copy last seen at from env

kwasniew:
## About the changes

Last seen at at the feature level should be the latest from the env level values. The bug was undefined comparison while our DB returned null.


chriswk (approved):
LGTM. don't you just love undefined !== null :)

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -395,7 +394,7 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
                 }
 
                 if (
-                    acc.lastSeenAt === undefined ||
+                    acc.lastSeenAt == null ||
```
ivarconr, line 397: Why not do a falsy value check instead?

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -671,9 +674,11 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
             }
             const featureRow = acc[row.feature_name];
             if (
-                featureRow.lastSeenAt === undefined ||
-                new Date(row.env_last_seen_at) >
-                    new Date(featureRow.last_seen_at)
+                featureRow?.lastSeenAt == null ||
```
ivarconr, line 677: I do not understand the null check. We should not retrieve that column from the db anymore. (We stopped writing it in 5.7). We should only pick the latest from the environment.

ivarconr: Maybe this should be a API concern and not DB concern? We only need it in our APIs to maintain backward compatibility.  Is it put here for some specific reasons? Performance concerns?

kwasniew: I'm trying to remove the ?. It was to test one thing in CI. We only replace previous lastSeenAt from other envs as we're reducing over the env list.

Regarding DB concern I'm adding the smallest change compared to the current code. Also as we do it at the store level all services using this get this for free for now.

ivarconr: I would probably write this as:

```js
if(isAfter(new Date(row.env_last_seen_at), new Date(featureRow.lastSeenAt))) {
   featureRow.lastSeenAt = row.env_last_seen_at;
}
```

this works because:
* `isAfter(new Date(null), new Date(null)) => false`
* `isAfter(new Date(row.env_last_seen_at), new Date(null)) => true`
* `isAfter(new Date(null), new Date(featureRow.lastSeenAt)) => false`

kwasniew: done

PR #5714, opened 2023-12-21
feat: tracking feature buttons clicks

sjaanus:


frontend/src/component/feature/FeatureToggleList/FeatureToggleListActions/FeatureToggleListActions.tsx
```diff
@@ -45,6 +47,11 @@ export const FeatureToggleListActions: FC<IFeatureToggleListActions> = ({
     const open = Boolean(anchorEl);
     const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
         setAnchorEl(event.currentTarget);
+        trackEvent('search-feature-buttons', {
+            props: {
+                clicked: 'group-actions',
```
kwasniew, line 52: I'm wondering if we can come up with a better name than 'clicked' which is desktop vocabulary and doesn't apply to mobile tap. I'd prefer a business oriented name, not physical interface oriented name.

kwasniew (approved):
Not sure about the name 'clicked'. Non blocking approval.

frontend/src/component/feature/FeatureToggleList/FeatureToggleListActions/FeatureToggleListActions.tsx
sjaanus: Done

PR #5717, opened 2023-12-21
fix: last seen cherrypick

kwasniew:
## About the changes

Cherry picking last seen at fixes

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
@@ -402,7 +402,7 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
                     acc.lastSeenAt == null ||
                     isAfter(
                         new Date(r.env_last_seen_at),
-                        new Date(acc[r.feature_name]),
+                        new Date(acc.lastSeenAt),
                     )
                 ) {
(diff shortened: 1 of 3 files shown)


src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -399,8 +399,11 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
                 }
 
                 if (
-                    acc.lastSeenAt === undefined ||
-                    new Date(r.env_last_seen_at) > new Date(acc.lastSeenAt)
+                    acc.lastSeenAt == null ||
```
kwasniew, line 402: this is required as new Date(undefined) doesn't work correctly as new Date(null)

ivarconr: But would acc.lastSeenAt now be `undefined` and not `null` because of the removal on [this line](https://github.com/Unleash/unleash/pull/5717/files#diff-d1551d5f4cc2fa9364c492a00400e4038514d4c2c010aa61325ca0ea72aa8aacL390)?

kwasniew: yes, but the line I removed was always breaking the current acc state

ivarconr (approved):

PR #5723, opened 2023-12-21
chore: support full path schemas

nunogois:
This backwards compatible change allows us to specify a schema `id` (full path) which to me feels a bit better than specifying the schema name as a string, since a literal string is prone to typos.

### Before

```ts
requestBody: createRequestSchema(
    'createResourceSchema',
),
responses: {
    ...getStandardResponses(400, 401, 403, 415),
    201: resourceCreatedResponseSchema(
        'resourceSchema',
    ),
},
```

### After

```ts
requestBody: createRequestSchema(
    createResourceSchema.$id,
),
responses: {
    ...getStandardResponses(400, 401, 403, 415),
    201: resourceCreatedResponseSchema(
        resourceSchema.$id,
    ),
},
```

### DIFF
--- a/src/lib/openapi/util/create-request-schema.ts
+++ b/src/lib/openapi/util/create-request-schema.ts
@@ -9,7 +9,9 @@ export const createRequestSchema = (
         content: {
             'application/json': {
                 schema: {
-                    $ref: `#/components/schemas/${schemaName}`,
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
I like this! This also has another consequence which is we could now organize schemas in folders.

PR #5725, opened 2023-12-22
feat: implement createdByUserId for all features

daveleek:
## About the changes

Implements setting values on the created_by_user_id column on the features table in the db

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-store.ts
@@ -49,6 +49,12 @@ export interface FeaturesTable {
     impression_data: boolean;
     archived?: boolean;
     archived_at?: Date;
+    created_by_user_id?: number;
+}
+
+export interface FeatureToggleInsert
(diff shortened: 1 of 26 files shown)


src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -467,20 +473,39 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             stale: data.stale,
             created_at: data.createdAt,
             impression_data: data.impressionData,
+            created_by_user_id: data.createdByUserId,
```
chriswk, line 476: Ah, that was a double take, it looked like you were excluding the createdByUserId and then using it. And then I spotted the fact that you changed it from optional to required. Nice.

gardleopard (commented):
Looks good to me:)

chriswk (approved):
LGTM. Make sure this doesn't break anything in unleash-enterprise as well

daveleek:
Thank you @chriswk! Yeah I've run it with enterprise locally and have all tests passing. I'll keep an eye on the build

PR #5731, opened 2023-12-27
Fix: strategy import

Tymek:
## About the changes
Some optional properties of `strategy` where not included after importing.

Internal issue: [SR-284](https://linear.app/unleash/issue/SR-284/wayfair-import-didnt-copy-title-of-strategies)


kwasniew (approved):
Also would be nice to add a test with the things that were not imported before

PR #5733, opened 2023-12-27
Doc: C++ client supports now feature toggle variants.

aruizs:
Hello,

I'd like to add a reference of the SDK for C++ to the supported Client list for Feature Toggle variants.

Thanks!


Tymek (approved):
Thank you! I'll merge it for now, and @dgorton or @thomasheartman can modify this list later, in case we need to, for example, have official vs community SDK separation on this list.

PR #5735, opened 2023-12-28
Added a check that allows posting edge bulk metrics with a client token

chriswk:
This allows bulk metrics posted with a Client token to be accepted. Previously you needed an admin token to have bulk metrics accepted


gastonfournier (approved):
LG!

PR #5738, opened 2023-12-29
feat: make local storage work and make feedback url configurable

sjaanus:
Make storage work react way.
Make feedback url configurable by env variable.

### DIFF
--- a/frontend/src/hooks/api/actions/useUserFeedbackApi/useUserFeedbackApi.ts
+++ b/frontend/src/hooks/api/actions/useUserFeedbackApi/useUserFeedbackApi.ts
@@ -1,7 +1,6 @@
-import { IInternalBanner } from 'interfaces/banner';
 import useAPI from '../useApi/useApi';
 import { ProvideFeedbackSchema } from '../../../../openapi';
-import useUiConfig from '../../getters/useUiConfig/useUiConfig';
+import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 
 const ENDPOINT = 'feedback';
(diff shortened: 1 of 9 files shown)


frontend/src/hooks/api/actions/useUserFeedbackApi/useUserFeedbackApi.ts
```diff
@@ -1,40 +1,41 @@
...
         propagateErrors: true,
     });
 
     const addFeedback = async (feedbackSchema: ProvideFeedbackSchema) => {
-        const requestId = 'addFeedback';
-        const req = createRequest(
-            ENDPOINT,
-            {
+        if (uiConfig.feedbackUriPath !== undefined) {
+            await fetch(uiConfig.feedbackUriPath, {
```
sjaanus, line 16: If feedback uri is set, it will make call directly there.

frontend/src/hooks/api/actions/useUserFeedbackApi/useUserFeedbackApi.ts
```diff
@@ -1,40 +1,41 @@
...
                 method: 'POST',
+                headers: { 'Content-Type': 'application/json' },
                 body: JSON.stringify(feedbackSchema),
-            },
-            requestId,
-        );
+            });
+        } else {
+            const requestId = 'addFeedback';
+            const req = createRequest(
```
sjaanus, line 23: If it is not set, it will build the path from baseurl.

kwasniew (approved):

PR #5739, opened 2024-01-02
chore: make security optional for /edge/validate endpoint

chriswk:
We've had a couple of misunderstandings from people surprised that Unleash allows posts against the `/edge/validate` endpoint without an API key. It is intentional that this endpoint does not require an Authorization header, so this PR updates our OpenAPI spec to clarify that there is no security required for `/edge/validate`


thomasheartman (approved):
Nice!

PR #5744, opened 2024-01-02
feat: support null created_by_user_id

gastonfournier:
## About the changes
Creating an incoming webhook with an admin token means we can't correlate the action with a real user. In this case we should support null.

### DIFF
--- a/src/migrations/20240102142100-incoming-webhooks-created-by.js
+++ b/src/migrations/20240102142100-incoming-webhooks-created-by.js
@@ -1,0 +1,13 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        ALTER TABLE incoming_webhooks ALTER COLUMN created_by_user_id DROP NOT NULL;
+        ALTER TABLE incoming_webhook_tokens ALTER COLUMN created_by_user_id DROP NOT NULL;
+        `,
+        cb,
+    );
+};
+
+exports.down = function (db, callback) {
+    callback();
+};


gastonfournier:
> Is there a created_by column here as well?

Nope, but that's the same we're doing in the features table which led to this question in Slack: https://unleash-internal.slack.com/archives/C046LV85N3C/p1704202383586369?thread_ts=1704194758.009959&cid=C046LV85N3C

src/migrations/20240102142100-incoming-webhooks-created-by.js
```diff
@@ -0,0 +1,13 @@
...
+        ALTER TABLE incoming_webhooks ALTER COLUMN created_by_user_id DROP NOT NULL;
+        ALTER TABLE incoming_webhook_tokens ALTER COLUMN created_by_user_id DROP NOT NULL;
+        `,
+        cb,
+    );
+};
+
+exports.down = function (db, callback) {
+    callback();
+};
```
daveleek, line 13: What happens if you move forwards, then backwards, and then forwards again? Will this migration fail because there's no NOT NULL to drop? or will it just silently continue with life?

daveleek (commented):
LG, follows the pattern of created_by_user_id for events and features, but I have a migration corner case question

src/migrations/20240102142100-incoming-webhooks-created-by.js
gastonfournier: Because we don't apply down migrations, the column will remain nullable. I don't see a point of manually making it non-nullable as it's more restrictive than being nullable.

nunogois (approved):
This moves us forward for now so I think it is the way to go. We can try to come back to this and improve it at a later stage once we have answers.

PR #5747, opened 2024-01-03
chore: format schedule information according to user preferences

thomasheartman:
This pr uses the user's preferred timezone to display the scheduled times. If the user has no preferences, the default will be used.

With norwegian locale set as preference: 

image

With nothing set (falls back to my system setting):
image


FredrikOseberg (approved):
LGTM

PR #5749, opened 2024-01-03
chore: observable events db migration

nunogois:
https://linear.app/unleash/issue/2-1773/db-create-migration-for-a-new-observable-events-table

Adds a new DB migration to create a new `observable_events` table. Even though we are thinking long term with the `source` columns, the short term purpose of this table will be to store incoming webhook calls.


gastonfournier (approved):
LG! I'm not sure the downmigration is needed, as it may lead to data loss, but we're not executing them anyway....

PR #5750, opened 2024-01-03
feat: add private collaboration mode docs

sjaanus:


website/docs/reference/project-collaboration-mode.md
```diff
@@ -22,6 +23,12 @@ The open collaboration mode is the default in Unleash and is how all projects wo
 
 Only admins and project members can submit change requests in a **protected** project. Other users can not submit change requests.
 
+## **Private** collaboration mode
+
+Private collaboration mode makes the project invisible to all **viewers** that are not project members.
```
FredrikOseberg, line 28: Maybe add some detail here about what gets hidden. IE: The project itself, all associated feature toggles ect.

FredrikOseberg (approved):
Looks good, but maybe we should add some more detail about what is hidden from the user when you set the mode to private.

PR #5753, opened 2024-01-04
feat: add dont ask me again button

sjaanus:


FredrikOseberg (changes requested):
I'm think this should just be visible if we force you into the flow. If you trigger the feedback form yourself we don't need this button.

FredrikOseberg (approved):
LGTM

PR #5758, opened 2024-01-04
task: add kill-switch for edge bulk metrics

chriswk:
As the title says. This adds a kill switch for the processing of Edge bulk metrics.

I was wondering if we should even add a kill switch fully around the router configuration as well, so that if the kill switch is already on when you start up you don't even map the route. WDYT?

### DIFF
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -81,6 +81,7 @@ exports[`should create default config 1`] = `
       "disableBulkToggle": false,
       "disableMetrics": false,
       "disableNotifications": false,
+      "edgeBulkMetricsKillSwitch": false,
       "embedProxy": true,
       "embedProxyFrontend": true,
       "enableLicense": false,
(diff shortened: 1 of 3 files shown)


gastonfournier (approved):
The problem about adding this flag to the endpoint, is that it's mounted at startup of the application, so switching the flag may require a restart. 

I think this is fine, but I'm wondering if returning 202 is good or if we should return 410 (Gone) so users will realize something's wrong

chriswk:
Hmm, Do we want to Edge to log error messages if it posts to an Unleash instance where the endpoint is killed?

gastonfournier:
> Hmm, Do we want to Edge to log error messages if it posts to an Unleash instance where the endpoint is killed?

This should be a misconfiguration or that customers overlooked our request to upgrade edge. I think this shouldn't happen, but if it does, it's something that someone has to take care of rather than silently dropping metrics

chriswk:
Hmm, ok, let's do 404 then, 410 when we might be planning to turn the kill switch off again seems semantically wrong.

gastonfournier (approved):
Looks nice! good message!

PR #5759, opened 2024-01-04
feat: encrypt emails at rest for demo login

gastonfournier:
## About the changes
This allows us to encrypt emails at signup for demo users to further secure our demo instance. Currently, emails are anonymized before displaying events performed by demo users. But this means that emails are stored at rest in our DB. By encrypting the emails at login, we're adding another layer of protection.

This can be enabled with a flag and requires the encryption key and the initialization vector (IV for short) to be present as environment variables.


src/lib/util/anonymise.ts
```diff
@@ -1,4 +1,22 @@
-import { createHash } from 'crypto';
+import { createCipheriv, createHash } from 'crypto';
+
+export function encrypt(s?: string): string {
+    const key = process.env.UNLEASH_ENCRYPTION_KEY;
+    const iv = process.env.UNLEASH_ENCRYPTION_IV;
```
gastonfournier, line 5: These will have to be added to demo

chriswk (approved):
LGTM

PR #5762, opened 2024-01-04
fix: adapts react markdown to open links in new tab

nunogois:
Adapts our current implementations of `ReactMarkdown` to always open markdown links in new tabs.


gastonfournier (approved):
LGTM!

PR #5764, opened 2024-01-04
feat: set apiTokenMiddleware for all AuthTypes.

daveleek:
## About the changes

Fixes an issue where setting AuthType=None on the instance caused /frontend to crash regardless if there's a correct token or not.

This is a redo of this PR: https://github.com/Unleash/unleash/pull/3031
Adds an end 2 end test to verify

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -109,29 +109,26 @@ export default async function getApp(
     );
 
     app.use(baseUriPath, patMiddleware(config, services));
+    app.use(baseUriPath, apiTokenMiddleware(config, services));
 
     switch (config.authentication.type) {
         case IAuthType.OPEN_SOURCE: {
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):
Love it 🙌🙌🙌

thomasheartman:
But for context: this turns out not to be as easy as expected. A number of our internal tests rely on AuthType.None allowing you to fetch flags and post metrics even if this isn't how we think it should be. As such, the fix to the problems described in #3005 might need a more surgical approach.

daveleek:
indeed @thomasheartman. I'll close this for now, the task morphs into a bug fix on the /frontend api

thomasheartman:
Great! Do you have a github issue for the bug that we can link?

PR #5766, opened 2024-01-04
fix: add checks to remove warnings on project settings

FredrikOseberg:
This PR removes warning on the project settings by always making sure we have the project before rendering the settings and do a conditional check on the featureLimit number field.


sjaanus (approved):
LG

PR #5768, opened 2024-01-04
fix: properly handle markdown links

nunogois:
Iterates on https://github.com/Unleash/unleash/pull/5762

While on the previous PR we would always open markdown links on a new tab, we still want to navigate on the same tab when a relative link is specified.

This adds a new `Markdown` common component with this logic by default, which should make things a lot simpler and easier to maintain. The logic that was followed is similar to the existing internal/external links logic in our banners.

### DIFF
--- a/frontend/src/component/common/Markdown/Markdown.tsx
+++ b/frontend/src/component/common/Markdown/Markdown.tsx
@@ -1,0 +1,24 @@
+import { Link } from '@mui/material';
+import { AnchorHTMLAttributes, ComponentProps } from 'react';
+import ReactMarkdown from 'react-markdown';
+import { useNavigate } from 'react-router-dom';
+
+const LinkRenderer = ({
+    href = '',
+    children,
+}: AnchorHTMLAttributes<HTMLAnchorElement>) => {
+    const navigate = useNavigate();
+
+    if (href.startsWith('/'))
+        return <Link onClick={() => navigate(href)}>{children}</Link>;
+
+    return (
+        <Link href={href} target='_blank' rel='noreferrer'>
+            {children}
+        </Link>
+    );
+};
+
+export const Markdown = (props: ComponentProps<typeof ReactMarkdown>) => (
+    <ReactMarkdown components={{ a: LinkRenderer }} {...props} />
+);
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/Markdown/Markdown.tsx
```diff
@@ -0,0 +1,24 @@
...
+import ReactMarkdown from 'react-markdown';
+import { useNavigate } from 'react-router-dom';
+
+const LinkRenderer = ({
+    href = '',
+    children,
+}: AnchorHTMLAttributes<HTMLAnchorElement>) => {
+    const navigate = useNavigate();
+
+    if (href.startsWith('/'))
```
gastonfournier, line 12: Relative links can also start without the slash, maybe just:
```suggestion
    if (!href.match(/^https?:\/\//))
```

gastonfournier (approved):
Feel free to merge, but the comment should remove some edge cases

frontend/src/component/common/Markdown/Markdown.tsx
nunogois: Hmm true, but the same can be said about absolute links. `mailto:` should not open a new tab, even though it doesn't start with `http*`. I'm somewhat confident with this for now, since it's the same logic that we're following for the banner link. If we decide to change it, we should change it in both places.

nunogois: More context: https://github.com/Unleash/unleash/pull/5011

gastonfournier: Sure, it makes sense. Feel free to merge it as is but I think we can potentially support other protocols with the regexp as well. This is what I like the most, being explicit about which protocols will open in a new tab:
```typescript
if (href.match(/^(https?:\/\/|mailto:)/))
    return (
        <Link href={href} target='_blank' rel='noreferrer'>
            {children}
        </Link>
    );
return (
        <Link href={href} rel='noreferrer'>{children}</Link>
);
```

ivarconr: In general we know about three scenarios we actually use:

1. Internal links. We want these to not open In a new tab. 
2. External links. We want these to open in a new tab. 
3. Mailto links. We want these to open in a new tab. This is particularly useful if the use has configured Gmail as the default email client. 

The current solution from @nunogois solves all these use cases In a very simple and easy to understand way. I see no reason for complicated regex to support possible scenarios in the future.

gastonfournier: Yes, I can see this handling absolute internal URLs, but if this can be an input from a user they could also be: `href="feature/variants"`. We could also validate and require absolute internal URLs, or just consider this an edge case we don't want to cover (i.e. internal links that are not absolute will open in a new tab). 

In any case, I'm not considering this a blocker for the PR

PR #5774, opened 2024-01-05
task: Add an X-Unleash-Version header to register response

chriswk:
Related to our work for making Edge bulk metrics a 1st class citizen of Unleash, this PR adds an X-Unleash-Version header to the response from client registration.

Based on when we add the new `/api/client/metrics/bulk` endpoint, Edge can use the response header from upstream to decide whether to post metrics to `/edge/metrics` or `/api/client/metrics/bulk`.

### DIFF
--- a/src/lib/routes/client-api/register.ts
+++ b/src/lib/routes/client-api/register.ts
@@ -15,6 +15,7 @@ import { createRequestSchema } from '../../openapi/util/create-request-schema';
 import { ClientApplicationSchema } from '../../openapi/spec/client-application-schema';
 import rateLimit from 'express-rate-limit';
 import { minutesToMilliseconds } from 'date-fns';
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
Looks good!

src/lib/routes/client-api/register.ts
```diff
@@ -93,6 +94,6 @@ export default class RegisterController extends Controller {
         data.environment = RegisterController.resolveEnvironment(user, data);
         data.project = RegisterController.extractProjectFromRequest(req);
         await this.clientInstanceService.registerClient(data, clientIp);
-        res.status(202).end();
+        res.header('X-Unleash-Version', version).status(202).end();
```
ivarconr, line 97: This would be the OSS version, which probably is enough, I think.

PR #5775, opened 2024-01-05
fix: add condition to show variants tab

FredrikOseberg:
Adds a conditional check so we don't display this tab for strategies that don't support it.


thomasheartman (approved):
Nice one! As long as that's the right condition (and it probably is), then this looks good to me.

PR #5776, opened 2024-01-05
fix: demo banner zIndex, display on top

nunogois:
This PR does 2 things:
- Fixes the `DemoBanner` zIndex to be the same as the sticky banners (no longer displays on top of modals)
- Moves the `Demo` wrapper to `App` instead of `MainLayout`, always displaying the demo banner before other banners

image


FredrikOseberg (approved):
LGTM

PR #5777, opened 2024-01-05
chore: show new change request diagram for users who are exposed to scheduled CRs

thomasheartman:
This PR adds the new CR diagram (with schedule info) conditionally. It'll only render for users who are exposed to scheduled CRs.

It also updates the text of the old version to make it all consistent and adds a `aria-details` attribute to the SVG so that assistive tech knows what it is.

New version
image


nicolaesocaciu (approved):
LGTM!

PR #5778, opened 2024-01-05
docs: add image zoom to docs

thomasheartman:
This PR adds a zoom feature to the docs.

Interact with the images to blow them up to fill the screen. Still leaves the header for now, so some larger images are a bit wonky, but we can probably get around that.


ivarconr (approved):
Looks great.

PR #5779, opened 2024-01-05
feat: Added bulk metrics support under /api/client/metrics/bulk path

chriswk:
This adds a bulk endpoint under `/api/client/metrics`. Accessible under `/api/client/metrics/bulk`.

This allows us to piggyback on the need for an API user with access. This PR mostly copies the behaviour from our `/edge/metrics` endpoint, but it filters metrics to only include the environment that the token has access to.

So a client token that has access to the `production` will not be allowed to report metrics for the `development` environment. More importantly, a `development` token will not be allowed to post metrics for the `production` environment.

### DIFF
--- a/src/lib/routes/client-api/metrics.ts
+++ b/src/lib/routes/client-api/metrics.ts
@@ -17,6 +17,7 @@ import { minutesToMilliseconds } from 'date-fns';
 import { BulkMetricsSchema } from '../../openapi/spec/bulk-metrics-schema';
 import { clientMetricsEnvBulkSchema } from '../../services/client-metrics/schema';
 import { IClientMetricsEnv } from '../../types/stores/client-metrics-store-v2';
+import ApiUser from '../../types/api-user';
 
 export default class ClientMetricsController extends Controller {
     logger: Logger;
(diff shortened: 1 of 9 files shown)


src/lib/routes/client-api/metrics.ts
```diff
@@ -75,6 +78,26 @@ export default class ClientMetricsController extends Controller {
                 }),
             ],
         });
+
+        this.route({
+            method: 'post',
+            path: '/bulk',
+            handler: this.bulkMetrics,
+            permission: NONE, // should have a permission but not bound to any environment
```
gastonfournier, line 86: Should we now consider the permission? Or should we just remove/change the comment

src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -229,4 +229,11 @@ export default class ClientMetricsServiceV2 {
         }
         return 'default';
     }
+
+    resolveUserEnvironment(user: IUser | IApiUser): string {
```
gastonfournier, line 233: Can we post metrics as a regular user?

src/lib/routes/client-api/metrics.ts
ivarconr: Remove comment please. We could have a more descriptive name of the case of client API. In our setup there is a separation of authentication ("do i have access") and authorization ("Can i do this action"). Permission is simply saying there is no specific permissions to perform this action, because the handler needs to implement the filtering itself.

src/lib/services/client-metrics/metrics-service-v2.ts
ivarconr: No, you need a Server Token.

src/lib/routes/client-api/metrics.ts
chriswk: Yeah, dangers of copy-pasting. Comment removed.

src/lib/services/client-metrics/metrics-service-v2.ts
gastonfournier: Yes, not that Typescript will care much about, but at least we'd be more clear.

chriswk: Made it clearer and it only accepts IApiUser now.

chriswk: Hmm, I'm going to keep the `IUser | IApiUser` here, we don't have a good setup for testing with custom api tokens at the moment, and this allows the test to verify that we do filter by environment token has access to, since if we don't have an API token we can only have access to 'default'

gastonfournier (approved):
LGTM!

PR #5788, opened 2024-01-08
chore: incoming webhook hooks

nunogois:
https://linear.app/unleash/issue/2-1813/create-new-incoming-webhook-hooks-on-the-frontend

Adds incoming webhook hooks to help us with CRUD operations on the frontend.


gastonfournier (approved):
LG!

PR #5791, opened 2024-01-08
feat: daily metrics db migration

kwasniew:
## About the changes

Business problem: we want to store up to 3 months of metrics data. Hourly storage is not feasible, so we go for the daily storage above 2 days.

Technical solution:
Adding 2 new tables to simulate time series DB in postgres. The idea is to store data in high resolution initially (the original tables) and then aggregate it to a lower resolution as it ages (2 new tables). 
The original metrics tables contain hourly data, the new tables will contain daily data. The structure is the same except from the timestamp column changing to date column.

Calculations for 90 days of data: https://gist.github.com/kwasniew/864bb6e5111b7de516b6e4454963d2d4#data-requirements-for-the-extended-90-days-scenario. New table should take twice as much space as the hourly table.

In the next PR I will create a scheduler that uses those tables for aggregation: https://github.com/Unleash/unleash/pull/5804

### DIFF
--- a/src/migrations/20240108151652-add-daily-metrics.js
+++ b/src/migrations/20240108151652-add-daily-metrics.js
@@ -1,7 +1,7 @@
 exports.up = function (db, cb) {
     db.runSql(
         `
-      CREATE TABLE IF NOT EXISTS client_metrics_env_daily(
+      CREATE TABLE IF NOT EXISTS client_metrics_env_daily (
                feature_name  VARCHAR(255),
                app_name      VARCHAR(255),
                environment   VARCHAR(100),
--- a/src/migrations/20240108151652-add-daily-metrics.js
+++ b/src/migrations/20240108151652-add-daily-metrics.js
@@ -1,0 +1,45 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+      CREATE TABLE IF NOT EXISTS client_metrics_env_daily(
+               feature_name  VARCHAR(255),
+               app_name      VARCHAR(255),
+               environment   VARCHAR(100),
+               timestamp     TIMESTAMP WITH TIME ZONE,
+               yes           INTEGER DEFAULT 0,
+               no            INTEGER DEFAULT 0,
+               PRIMARY KEY (feature_name, app_name, environment, timestamp)
+      );
+      CREATE TABLE IF NOT EXISTS client_metrics_env_variants_daily (
+               feature_name  VARCHAR(255),
+               app_name      VARCHAR(255),
+               environment   VARCHAR(100),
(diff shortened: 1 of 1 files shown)


src/migrations/20240108151652-add-daily-metrics.js
```diff
@@ -0,0 +1,45 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+      CREATE TABLE IF NOT EXISTS client_metrics_env_daily (
+               feature_name  VARCHAR(255),
+               app_name      VARCHAR(255),
+               environment   VARCHAR(100),
+               timestamp     TIMESTAMP WITH TIME ZONE,
```
kwasniew, line 8: I'm thinking about using date here since we don't want the time section

src/migrations/20240108151652-add-daily-metrics.js
```diff
@@ -0,0 +1,45 @@
...
+               yes           INTEGER DEFAULT 0,
+               no            INTEGER DEFAULT 0,
+               PRIMARY KEY (feature_name, app_name, environment, timestamp)
+      );
+      CREATE TABLE IF NOT EXISTS client_metrics_env_variants_daily (
+               feature_name  VARCHAR(255),
+               app_name      VARCHAR(255),
+               environment   VARCHAR(100),
+               timestamp     TIMESTAMP WITH TIME ZONE,
```
kwasniew, line 17: same here, date not time

src/migrations/20240108151652-add-daily-metrics.js
```diff
@@ -0,0 +1,45 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+      CREATE TABLE IF NOT EXISTS client_metrics_env_daily (
+               feature_name  VARCHAR(255),
+               app_name      VARCHAR(255),
+               environment   VARCHAR(100),
+               date          DATE,
```
kwasniew, line 8: this is deliberately date not timestamp as the original hourly table

src/migrations/20240108151652-add-daily-metrics.js
```diff
@@ -0,0 +1,45 @@
...
+               yes           INTEGER DEFAULT 0,
+               no            INTEGER DEFAULT 0,
+               PRIMARY KEY (feature_name, app_name, environment, date)
+      );
+      CREATE TABLE IF NOT EXISTS client_metrics_env_variants_daily (
```
kwasniew, line 13: copy of metrics_env_variants but with date instead of timestamp

andreas-unleash (approved):
LGTM

PR #5800, opened 2024-01-09
fix: upgrade unleash-client to v5.3.0

Tymek:
Update Unleash Node client

https://linear.app/unleash/issue/1-1705/upgrade-node-client-to-53-in-unleash


src/lib/features/playground/feature-evaluator/client.ts
```diff
@@ -176,7 +183,7 @@ export default class UnleashClient {
 
                 return {
                     name: strategySelector.name,
-                    id: strategySelector.id,
+                    id: strategySelector.id || randomId(),
```
Tymek, line 186: fix for TS: id can't be undefined

kwasniew (approved):

PR #5803, opened 2024-01-09
chore: incoming webhooks description migration

nunogois:
https://linear.app/unleash/issue/2-1822/incoming-webhooks-add-description

A description field will help us better distinguish incoming webhooks and their purpose.


gastonfournier (approved):
LGTM!

PR #5808, opened 2024-01-09
feat: add db metrics for environment-store

gardleopard:
This will give us prometheus metrics on environment lookups in the database


gastonfournier (approved):
LG

PR #5809, opened 2024-01-09
feat: include environment type label in feature_toggle_update metrics

gardleopard:
This is needed in order to identify what type of an environment a toggle is updated in. This can be test, development, pre-production or production.

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -1,4 +1,5 @@
 import client from 'prom-client';
+import memoizee from 'memoizee';
 import EventEmitter from 'events';
 import { Knex } from 'knex';
 import * as events from './metric-events';
@@ -19,11 +20,12 @@ import {
     CLIENT_REGISTER,
 } from './types/events';
 import { IUnleashConfig } from './types/option';
-import { IUnleashStores } from './types/stores';
+import { IEnvironmentStore, IUnleashStores } from './types/stores';
 import { hoursToMilliseconds, minutesToMilliseconds } from 'date-fns';
 import Timer = NodeJS.Timer;
 import { InstanceStatsService } from './features/instance-stats/instance-stats-service';
 import { ValidatedClientMetrics } from './services/client-metrics/schema';
+import { IEnvironment } from './types';
 
 export default class MetricsMonitor {
     timer?: Timer;
@@ -47,7 +49,15 @@ export default class MetricsMonitor {
             return Promise.resolve();
         }
 
-        const { eventStore } = stores;
+        const { eventStore, environmentStore } = stores;
+
+        const cachedEnvironments: () => Promise<IEnvironment[]> = memoizee(
+            async () => environmentStore.getAll(),
+            {
+                promise: true,
+                maxAge: hoursToMilliseconds(1),
+            },
+        );
 
         client.collectDefaultMetrics();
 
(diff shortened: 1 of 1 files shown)


gastonfournier (approved):
Looks good!

src/lib/metrics.ts
```diff
@@ -360,57 +370,89 @@ export default class MetricsMonitor {
...
         eventStore.on(FEATURE_METADATA_UPDATED, ({ featureName, project }) => {
-            featureToggleUpdateTotal.labels(featureName, project, 'n/a').inc();
+            featureToggleUpdateTotal
+                .labels(featureName, project, 'n/a', 'n/a')
+                .inc();
         });
         eventStore.on(FEATURE_UPDATED, ({ featureName, project }) => {
             featureToggleUpdateTotal
-                .labels(featureName, project, 'default')
+                .labels(featureName, project, 'default', 'n/a')
```
ivarconr, line 389: I think default env here must be a bug?

src/lib/metrics.ts
```diff
@@ -360,57 +370,89 @@ export default class MetricsMonitor {
...
+                );
                 featureToggleUpdateTotal
-                    .labels(featureName, project, environment)
+                    .labels(featureName, project, environment, environmentType)
                     .inc();
             },
         );
         eventStore.on(
             FEATURE_ENVIRONMENT_ENABLED,
             ({ featureName, project, environment }) => {
```
kwasniew, line 442: should be async for consistency with previous listeners

src/lib/metrics.ts
```diff
@@ -360,57 +370,89 @@ export default class MetricsMonitor {
...
-                featureToggleUpdateTotal
-                    .labels(featureName, project, environment)
-                    .inc();
+                this.resolveEnvironmentType(
+                    environment,
+                    cachedEnvironments,
+                ).then((environmentType) => {
```
kwasniew, line 446: we can refactor similar to previous await cases

src/lib/metrics.test.ts
```diff
@@ -93,7 +99,32 @@ test('should collect metrics for updated toggles', async () => {
...
+    stores.eventStore.emit(FEATURE_ENVIRONMENT_ENABLED, {
+        featureName: 'TestToggle',
+        project: 'default',
+        environment: 'testEnvironment',
+        data: { name: 'TestToggle' },
+    });
+
+    // Wait for event to be processed, not nice, but it works.
+    await new Promise((done) => {
+        setTimeout(done, 10);
```
kwasniew, line 122: does it also work with setTimeout(done, 1)?

kwasniew: also you can search for `await ms(100);` this pattern in our codebase that better reveals the intent

kwasniew (approved):
4/5

src/lib/metrics.ts
gardleopard: fixed

src/lib/metrics.ts
gardleopard: fixed

src/lib/metrics.test.ts
gardleopard: fixed

gardleopard: ops was wrong i have not fixed

src/lib/metrics.ts
gardleopard: fixed

src/lib/metrics.test.ts
gardleopard: on my machine 1ms is enough

PR #5810, opened 2024-01-09
fix: remove unsued metric 'optimal304DiffingCounter'

ivarconr:
This metric was used while developing the optimal304 feature. The feature flag has been removed and this data is not longer being collected and this will remove the metric from Prometheus.


gardleopard (approved):
LGTM

PR #5811, opened 2024-01-09
fix: import test wait should be between import and page load

gardleopard:
the wait was after fetching the feature page, should be before probably.


gastonfournier (approved):
I think you're correct

PR #5820, opened 2024-01-10
chore: add schedule suspended event type

thomasheartman:
To facilitate emitting this event in Enterprise


andreas-unleash (approved):
LGTM

PR #5821, opened 2024-01-10
feat: allow slack-app and webhook to listen to schedule suspended events

thomasheartman:
This PR adds the schedule suspended event to the slack-app and webhook definitions.

It also slightly tweaks the markdown formatting of change requests to add a definite article. This means the snapshot also needs to be updated.


nunogois (approved):
LGTM 👍

PR #5824, opened 2024-01-10
refactor: move plausible context on top of feedback

sjaanus:
We were unable to send plausible events from feedback. Now we can.

### DIFF
--- a/frontend/src/component/App.tsx
+++ b/frontend/src/component/App.tsx
@@ -51,72 +51,64 @@ export const App = () => {
 
     return (
         <ErrorBoundary FallbackComponent={LayoutError}>
-            <ErrorBoundary FallbackComponent={LayoutError}>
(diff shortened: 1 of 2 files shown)


frontend/src/component/App.tsx
```diff
@@ -52,76 +51,72 @@ export const App = () => {
...
-                                                                key={route.path}
-                                                                path={
-                                                                    route.path
-                                                                }
-                                                                element={
-                                                                    <LayoutPicker
-                                                                        isStandalone={
-                                                                            route.isStandalone ===
-                                                                            true
+            <ErrorBoundary FallbackComponent={LayoutError}>
```
kwasniew, line 54: duplicate error boundary

sjaanus: Interesting, we had it already before. I will remove

PR #5826, opened 2024-01-10
chore: remove redundant linting check, lint before test

daveleek:
## About the changes

- Removes redundant lint check
- Runs linting before test


.github/workflows/build_frontend_prs.yml
```diff
@@ -21,7 +21,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: yarn --frozen-lockfile
-      - run: yarn run test
-      - run: yarn run fmt:check
```
daveleek, line 25: fmt:check and lint:check both run `biome check src`

chriswk (approved):
Nice catch.

PR #5828, opened 2024-01-10
chore: adapt integrations layout for incoming webhooks

nunogois:
https://linear.app/unleash/issue/2-1823/adapt-integrations-page-to-incoming-webhooks-tab-layout

Adapts the current integrations page to the incoming webhooks feature, which includes things like:
 - Displaying both configured and available integrations in a single "page block"
 - Implement tabs
 - Add "Incoming Webhooks" integration card
 - Adapt the existing `IntegrationCard` component to support `onClick`

This also includes a small girl scouting fix: Some tabs (like on the roles page) did not correctly reflect the active tab.

### `incomingWebhooks` disabled

image

### `incomingWebhooks` enabled

Notice the new "Incoming webhooks" tab and integration card.

image

### DIFF
--- a/frontend/src/component/admin/roles/RolesPage.tsx
+++ b/frontend/src/component/admin/roles/RolesPage.tsx
@@ -15,6 +15,7 @@ import { PageHeader } from 'component/common/PageHeader/PageHeader';
 import { Add } from '@mui/icons-material';
 import ResponsiveButton from 'component/common/ResponsiveButton/ResponsiveButton';
 import { IRole } from 'interfaces/role';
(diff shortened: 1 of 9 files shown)


gastonfournier (approved):
I think the idea is to move all configured integrations to tabs, but this is a good starting point. Please, follow it up and validate in production that nothing breaks in the UI. Screenshots look great

nunogois:
> I think the idea is to move all configured integrations to tabs, but this is a good starting point. Please, follow it up and validate in production that nothing breaks in the UI. Screenshots look great

After aligning with @nicolaesocaciu we decided to leave that part for later. Thanks, will do!

PR #5830, opened 2024-01-10
feat: Metrics api returning daily data

kwasniew:
## About the changes

Metrics API switches to daily metrics above 48 hours

Detailed changes:
* hours back max value validation dependent on the flag
* utility to calculate default/empty daily buckets starting from the end of the previous day

Flow:
* controller checks which max hours back is supported depending on a flag
* service either goes to the new metrics or old metrics based on a flag
* empty buckets for missing metrics are generated

## Discussion points

We set hourly buckets to the beginning of the previous hour. E.g. metrics at 13:42 is reported in 13:00 bucket. But for the daily metrics we set the whole day of metrics to 23:59 of the previous day. Maybe we should set it to 00:00 of the previous day?


src/lib/services/client-metrics/metrics-service-v2.ts
```diff
@@ -183,17 +187,26 @@ export default class ClientMetricsServiceV2 {
...
-              )
-            : await this.clientMetricsStoreV2.getMetricsForFeatureToggle(
-                  featureName,
-                  hoursBack,
-              );
-
-        const hours = generateHourBuckets(hoursBack);
+        let hours: HourBucket[];
+        let metrics: IClientMetricsEnv[];
+        if (this.flagResolver.isEnabled('extendedUsageMetrics')) {
```
kwasniew, line 192: new and old path clearly separated to have an easy mechanism to turn new metrics on and off

src/lib/util/time-utils.test.ts
```diff
@@ -1,7 +1,16 @@
...
 
 test('generateHourBuckets', () => {
     const result = generateHourBuckets(24);
 
     expect(result).toHaveLength(24);
 });
+
+test('generateDayBuckets', () => {
+    const result = generateDayBuckets(7);
+    const endOfDayYesterday = endOfDay(subDays(new Date(), 1));
```
kwasniew, line 12: it doesn't make sense to have a daily bucket for a pending day

src/lib/services/client-metrics/metrics-service-v2.test.ts
```diff
@@ -120,3 +122,115 @@ test('process metrics properly even when some names are not url friendly, with d
...
+    const service = new ClientMetricsServiceV2(
+        { clientMetricsStoreV2 },
+        config,
+        lastSeenService,
+    );
+
+    const metrics = await service.getClientMetricsForToggle('feature', 3 * 24);
+
+    expect(metrics).toMatchObject([
+        { ...baseData, timestamp: endOfDay(threeDaysAgo) },
```
kwasniew, line 171: fill in missing data for the days without metrics

FredrikOseberg (approved):
LGTM

PR #5834, opened 2024-01-10
task: Upgrade to biome 1.5, moving files list to joined config variable

chriswk:
Like the title says. This simplifies our files list to one list instead of having to maintain separate for linting and formatting (this could in some cases be desirable, but I don't think it is for us).


nunogois (approved):
LGTM

chriswk:
abandoned due to being superseded by a different PR.

PR #5836, opened 2024-01-10
feat: daily metrics display

kwasniew:
## About the changes

Screenshot 2024-01-10 at 16 24 09

* Displaying in the last x days instead of last x hours when > 48 hours
* Daily dates are formatted differently from hourly dates
* Daily dates are displayed in UTC timezone as opposed to hourly dates that are using local timezone
* Added missing tests for the stats display, the char with canvas is too difficult to test

Depends on https://github.com/Unleash/unleash/pull/5830


FredrikOseberg (approved):
LGTM

PR #5846, opened 2024-01-11
docs: remove white background from images + zoomed images

thomasheartman:
This PR does the following:
1. Replaces all images in the anatomy of Unleash doc with new versions that don't have text on transparent backgrounds. This makes the images work on both light and dark backgrounds.
2. Removes the white background from images in the docs. This was only there to make sure the anatomy images worked in dark mode too.
3. Uses the general docs background color (based on your preferred theme) when zooming in on images.


nicolaesocaciu (approved):
LGTM!

PR #5847, opened 2024-01-11
fix(import): making all imports relative and removing baseUrl

bassrock:
## About the changes

I use Unleash in a project as a module of a larger express app, I had recently been on Unleash 4.8 and started trying to upgrade, but noticed that I got a ton of type errors when upgrading to anything above ~4.12 .  Upon further investigation it seems to have been caused by Unleash's type declarations around that time starting to mix between relative and fix'd path resolution depending on a baseURL in the tsconfig. I was able to get around this in my repo where I use unleash by adding `skipLibCheck: true` to my tsconfig.

This change will change unleash to use relative imports and removes the `baseUrl` in the tsconfig so that IDEs and the build command will fail if a relative path import is not used.

### Important files
No file is particularly important, this changes all imports that weren't relative, and modifies the tsconfig.

## Discussion points
Nope!


sighphyre:
Hey @bassrock, thank you for the contribution! 

I've chatted with the team, we think this makes sense so we're happy to accept it if you can take a look at why the changes are causing the build to fail on `tsc`

bassrock:
@sighphyre Great! Just pushed an amended merge commit to fix the build issue, looks like I missed a few imports on merge from main.

sighphyre (approved):
Hey @bassrock, thank you so much for doing this! It's hugely appreciated!

I've had to push a merge commit to your fork, I hope you don't mind - lots of changes lately and the GH UI gave up during resolving the conflict!

bassrock:
@sighphyre Not a problem, glad to see this merged!

PR #5848, opened 2024-01-11
chore: validate import data using api

gardleopard:
Use the admin api to validate the feature toggle that has been imported.

This hopefully makes the test less fragile as we do not depend on the frontend to validate the import.


ivarconr (approved):
I changed the title to "chore: ..". Other than that this looks good.

PR #5850, opened 2024-01-11
fix: remove empty variants when changing tabs

FredrikOseberg:
This PR makes a change to how variants work in the new setup. Variants will now:

* Be removed if you change tab or unmount the component and it has no name
* Moved StrategyVariants into a separate component to isolate this change
* Add error handling around onSubmit and only trigger feedback if it's successful

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyForm/NewFeatureStrategyForm.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyForm/NewFeatureStrategyForm.tsx
@@ -337,7 +337,9 @@ export const NewFeatureStrategyForm = ({
             if (newStrategyConfigurationFeedback && !hasSubmittedFeedback) {
                 createFeedbackContext();
             }
-        } catch (e) {}
+        } catch (e) {
+            console.error(e);
+        }
     };
 
     const handleChange = (event: React.ChangeEvent<{}>, newValue: number) => {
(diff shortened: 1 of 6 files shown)


frontend/src/component/feature/FeatureStrategy/NewFeatureStrategyCreate/NewFeatureStrategyCreate.test.tsx
```diff
@@ -199,20 +199,45 @@ describe('NewFeatureStrategyCreate', () => {
...
+        const targetingEl = await screen.findByText('Targeting');
+        fireEvent.click(targetingEl);
+
+        const addConstraintEl = await screen.findByText('Add constraint');
+        expect(addConstraintEl).toBeInTheDocument();
+
+        fireEvent.click(variantsEl);
+
+        const variants = screen.queryAllByTestId('VARIANT');
+        expect(variants.length).toBe(0);
```
kwasniew, line 240: could we add a check before this one that shows that the length was non 0 before?

frontend/src/component/feature/FeatureStrategy/FeatureStrategyForm/NewFeatureStrategyForm.tsx
```diff
@@ -330,11 +331,13 @@ export const NewFeatureStrategyForm = ({
...
+        try {
+            await onSubmit();
 
-        if (newStrategyConfigurationFeedback && !hasSubmittedFeedback) {
-            createFeedbackContext();
-        }
+            if (newStrategyConfigurationFeedback && !hasSubmittedFeedback) {
+                createFeedbackContext();
+            }
+        } catch (e) {}
```
kwasniew, line 340: do we want to console.error?

kwasniew (approved):

PR #5851, opened 2024-01-11
chore: wait for post to finish instead of waiting a fixed time

gardleopard:
This will hopefully make the test faster and more stable

### DIFF
--- a/frontend/cypress/integration/import/import.spec.ts
+++ b/frontend/cypress/integration/import/import.spec.ts
@@ -115,9 +115,10 @@ describe('imports', () => {
             });
         cy.get("[data-testid='VALIDATE_BUTTON']").click();
         cy.get("[data-testid='IMPORT_CONFIGURATION_BUTTON']").click();
-        // cy.contains('Import completed');
 
         cy.wait('@featureImported');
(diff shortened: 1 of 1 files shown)


gastonfournier (approved):
Lovely!

frontend/cypress/integration/import/import.spec.ts
```diff
@@ -113,8 +117,7 @@ describe('imports', () => {
         cy.get("[data-testid='IMPORT_CONFIGURATION_BUTTON']").click();
         // cy.contains('Import completed');
```
kwasniew, line 118: I'm wondering why this commented out code was commented out. Ideally UI tests should wait on the UI changes not HTTP API interception

kwasniew: However if your change fixes the issue then thumbs up

gardleopard: I dont know

gardleopard: Added this back now @kwasniew, good spot. Now we wait for the POST to finish and then validate ui.

PR #5855, opened 2024-01-11
chore: add no unused imports biome rule

nunogois:
Adds a Biome rule for "no unused imports", which is something we sometimes have trouble catching.

We're adding this as a warning for now. It is safely and easily fixable with `yarn lint:fix`.

image
image


biome.json
```diff
@@ -4,9 +4,6 @@
     "enabled": true,
     "rules": {
       "recommended": true,
-      "performance": {
-        "noDelete": "off"
-      },
```
nunogois, line 9: This was a duplicate property. See `performance` below. I kept the one that has the most options set.

sighphyre (approved):
Phew that's a lot of used imports! Nice one!

chriswk (approved):
Nice, is unusedImport now a safe fix? It didn't use to be.

nunogois:
> Nice, is unusedImport now a safe fix? It didn't use to be.

The fix itself is safe, but it's still part of the nursery set of rules: https://biomejs.dev/linter/rules/no-unused-imports/

PR #5858, opened 2024-01-11
chore: update orval schemas

thomasheartman:
This PR updates Orval schemas. The motivation is to get the new schemas for schedule suspension, but as always, we get everything else too.


andreas-unleash (approved):
Nice! LGTM

PR #5859, opened 2024-01-11
feat: show suspended schedules in ui

thomasheartman:
This PR updates the UI to display suspended schedules correctly. 

It:

- uses a standard "pause" icon for the new suspended state
- uses a grey / disabled color for icons text
- mentions why the CR was suspended in the review box at the bottom (and when hovering over the badge / icon)
- updates the timeline to say "schedule suspended" (or "failed") when it is in that state.

In doing this, it also does a little bit of refactoring to use types that are more accurate and to avoid too many "maybe undefined" variables.

It looks a bit like this now:

image

### DIFF
--- a/frontend/src/component/changeRequest/changeRequest.types.ts
+++ b/frontend/src/component/changeRequest/changeRequest.types.ts
@@ -18,10 +18,10 @@ export interface IChangeRequest {
     comments: IChangeRequestComment[];
     conflict?: string;
     state: ChangeRequestState;
-    schedule?: IChangeRequestSchedule;
+    schedule?: ChangeRequestSchedule;
 }
 
-export type IChangeRequestSchedule =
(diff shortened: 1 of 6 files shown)


frontend/src/component/changeRequest/changeRequest.types.ts
```diff
@@ -18,14 +18,25 @@ export interface IChangeRequest {
...
+          status: 'failed';
+          scheduledAt: string;
+          failureReason?: string | null;
+          reason: string;
+      }
+    | {
+          status: 'suspended';
+          scheduledAt: string;
+          reason: string;
+      };
```
thomasheartman, line 39: I think this is better expressed as a proper union type

thomasheartman:
Closed in favor of #5871, #5872, and #5873

PR #5867, opened 2024-01-11
Biome1.5.1

chriswk:
Lots of work here, mostly because I didn't want to turn off the `noImplicitAnyLet` lint. This PR tries its best to type all the untyped lets biome complained about (Don't ask me how many hours that took or how many lints that was >200...), which in the future will force test authors to actually type their global variables setup in `beforeAll`.

### DIFF
--- a/src/lib/addons/slack-app.test.ts
+++ b/src/lib/addons/slack-app.test.ts
@@ -26,7 +26,7 @@ jest.mock('@slack/web-api', () => ({
 }));
 
 describe('SlackAppAddon', () => {
-    let addon;
+    let addon: SlackAppAddon;
     const accessToken = 'test-access-token';
     const loggerMock = {
         debug: jest.fn(),
@@ -102,7 +102,10 @@ describe('SlackAppAddon', () => {
             ],
         };
 
-        await addon.handleEvent(eventWith2Tags, { accessToken });
+        await addon.handleEvent(eventWith2Tags, {
+            accessToken,
+            defaultChannels: '',
+        });
 
         expect(slackApiCalls.length).toBe(2);
         expect(slackApiCalls[0].channel).toBe('general');
@@ -132,6 +135,7 @@ describe('SlackAppAddon', () => {
     it('should not post a message if there are no tagged channels and no defaultChannels', async () => {
         await addon.handleEvent(event, {
             accessToken,
+            defaultChannels: '',
         });
 
         expect(slackApiCalls.length).toBe(0);
@@ -167,7 +171,10 @@ describe('SlackAppAddon', () => {
             .mockResolvedValueOnce({ ok: true })
             .mockRejectedValueOnce(mockError);
 
-        await addon.handleEvent(eventWith3Tags, { accessToken });
+        await addon.handleEvent(eventWith3Tags, {
+            accessToken,
+            defaultChannels: '',
+        });
 
         expect(postMessage).toHaveBeenCalledTimes(3);
         expect(loggerMock.warn).toHaveBeenCalledWith(
--- a/src/lib/routes/admin-api/state.ts
+++ b/src/lib/routes/admin-api/state.ts
@@ -102,6 +102,7 @@ class StateController extends Controller {
         const userName = extractUsername(req);
         const { drop, keep } = req.query;
         // TODO: Should override request type so file is a type on request
+        // biome-ignore lint/suspicious/noImplicitAnyLet: <explanation>
         let data;
         // @ts-expect-error
         if (req.file) {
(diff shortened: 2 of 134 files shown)


nunogois (approved):
Wow.

src/lib/addons/slack-app.test.ts
```diff
@@ -26,7 +26,7 @@ jest.mock('@slack/web-api', () => ({
 }));
 
 describe('SlackAppAddon', () => {
-    let addon;
+    let addon: SlackAppAddon;
```
gastonfournier, line 29: Shouldn't the type be `| undefined`? because you're not assigning it a value on declaration

src/lib/addons/slack-app.test.ts
```diff
@@ -102,7 +102,10 @@ describe('SlackAppAddon', () => {
             ],
         };
 
-        await addon.handleEvent(eventWith2Tags, { accessToken });
+        await addon.handleEvent(eventWith2Tags, {
+            accessToken,
+            defaultChannels: '',
```
gastonfournier, line 107: I guess default channels should be allowed to be undefined... maybe the type definition is wrong?

src/lib/routes/admin-api/state.ts
```diff
@@ -102,6 +102,7 @@ class StateController extends Controller {
         const userName = extractUsername(req);
         const { drop, keep } = req.query;
         // TODO: Should override request type so file is a type on request
+        // biome-ignore lint/suspicious/noImplicitAnyLet: <explanation>
```
gastonfournier, line 105: This is prod code, maybe we can add an explanation, or just don't know why this is ignored rather than <explanation>

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -51,18 +52,27 @@ test('editor users should only get client or frontend tokens', async () => {
     const { request, destroy } = await setupAppWithCustomAuth(stores, preHook);
 
     await stores.apiTokenStore.insert({
+        environment: '',
+        projects: [],
+        tokenName: '',
```
gastonfournier, line 57: I guess this is fine, but unsure if this ends up being the same as sending undefined. I trust it because the test fails, but I'm not sure if we're actually testing the output of this insert

src/test/e2e/api/admin/tags.e2e.test.ts
```diff
@@ -128,7 +131,7 @@ test('Can tag features', async () => {
     });
 
     await db.stores.tagStore.createTag(removedTag);
-    await db.stores.featureTagStore.tagFeature(featureName, removedTag);
+    await db.stores.featureTagStore.tagFeature(featureName, removedTag, -1337);
```
gastonfournier, line 134: Shouldn't this default to -1337 if we don't send it? Just wondering whether adding this is needed or just makes it more explicit

src/test/e2e/stores/client-application-store.e2e.test.ts
```diff
@@ -125,15 +132,18 @@ test('Merge keeps value for single row in database', async () => {
         appName: clientRegistration.appName,
         description: 'new description',
     });
-    const stored = await clientApplicationsStore.getApplication(
+    const stored = await clientApplicationsStore.get(
```
gastonfournier, line 135: getApplication and get methods are the same, should we remove one of them?

src/test/fixtures/fake-strategies-store.ts
```diff
@@ -24,13 +24,13 @@ export default class FakeStrategiesStore implements IStrategyStore {
...
     async createStrategy(update: IMinimalStrategy): Promise<void> {
-        let params;
+        let params: object[];
         if (
             typeof update.parameters === 'string' ||
             typeof update.parameters === 'number'
         ) {
             if (update.parameters === '') {
-                params = {};
+                params = [{}];
```
gastonfournier, line 33: Maybe it does not affect anything but I believe this should be:
```suggestion
                params = [];
```

gastonfournier (approved):
Looks good! Overall I noticed inconsistencies between `let variable: X;` and `let variable: X | undefined` but I think it doesn't hurt. Also, a few comments can be notes for later. As long as tests pass I'm ok with this one, it's a great improvement

src/test/e2e/stores/client-application-store.e2e.test.ts
chriswk: getApplication doesn't exist on the store interface. it was written only on the fake store, but yes, we could remove it.

src/test/e2e/api/admin/tags.e2e.test.ts
gastonfournier: Yes, I was unsure whether this was required by the linter or compiler or just QoL improvement.

src/test/e2e/stores/client-application-store.e2e.test.ts
gastonfournier: https://github.com/Unleash/unleash/blob/e27a578fab648c395f80fccc5d59636d3704db21/src/lib/db/client-applications-store.ts#L157-L169 isn't it this one?

PR #5870, opened 2024-01-12
refactor: use union types for change request types

thomasheartman:
This changes the two interfaces IChangeRequest and IChangeRequestSchedule to be union types instead of interfaces. It also extracts the constituents of those union types into proper types themselves (so that they can be used in function type signatures etc). It also updates the type names.

This turned out to be more work than I had imagined, but I think the end result pays off, giving us more type safety and control. 

I wanted to use just `ChangeRequest` for the IChangeRequest type, but that caused issues due to naming collisions with the `ChangeRequest` component that we have, causing tests to fail. I've named it `ChangeRequestType` as a potential solution, but suggestions are welcome.

The relevant changes are in `frontend/src/component/changeRequest/changeRequest.types.ts`. Everything else is updated references and some necessary refactoring to respect the new types.

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
@@ -111,7 +111,9 @@ export const ChangeRequestOverview: FC = () => {
         changeRequest.environment,
     );
 
-    const hasSchedule = Boolean('schedule' in changeRequest && changeRequest.schedule?.scheduledAt);
+    const hasSchedule = Boolean(
+        'schedule' in changeRequest && changeRequest.schedule?.scheduledAt,
+    );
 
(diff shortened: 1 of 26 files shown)


frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
thomasheartman, line 1: This file is a little messy right now, but I'll clean it up in a follow-up that handles suspended schedules too.

frontend/src/component/changeRequest/ChangeRequestStatusBadge/ChangeRequestStatusBadge.tsx
thomasheartman, line 1: This file will be cleaned up in a follow up that handles suspended states properly.

andreas-unleash (approved):
LGTM!

PR #5874, opened 2024-01-12
docs: add a note about suspended change request schedules

thomasheartman:
This PR describes in closer detail what happens when the user who scheduled a change request is deleted. It mentions the new suspended state and what the user can do to get out of it.


andreas-unleash (approved):
LGTM!

PR #5876, opened 2024-01-12
feat: aggregate metrics for the same timestamp

kwasniew:
## About the changes

We want to add/aggregate metrics for the same timestamp instead of showing x different lines on top of each other
Screenshot 2024-01-12 at 10 23 18

Also updated info about time settings:

* local time for hours
Screenshot 2024-01-12 at 10 42 21

* UTC time for days
Screenshot 2024-01-12 at 10 42 30

### DIFF
--- a/frontend/src/interfaces/featureToggle.ts
+++ b/frontend/src/interfaces/featureToggle.ts
@@ -107,5 +107,5 @@ export interface IFeatureMetricsRaw {
     timestamp: string;
     yes: number;
     no: number;
-    variants: Record<string, number>;
+    variants: Record<string, number> | undefined;
 }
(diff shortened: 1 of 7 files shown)


frontend/src/interfaces/featureToggle.ts
```diff
@@ -107,5 +107,5 @@ export interface IFeatureMetricsRaw {
     timestamp: string;
     yes: number;
     no: number;
-    variants: Record<string, number>;
+    variants: Record<string, number> | undefined;
```
kwasniew, line 110: reflecting actual backend data

FredrikOseberg (approved):

PR #5878, opened 2024-01-12
feat: Prometheus last day metrics

kwasniew:
## About the changes

We can check how many metrics we accumulate from the instance stats API and from prometheus.


src/lib/metrics.ts
```diff
@@ -235,6 +243,15 @@ export default class MetricsMonitor {
...
                 }
 
+                enabledMetricsBucketsPreviousDay.reset();
+                enabledMetricsBucketsPreviousDay.set(
+                    stats.previousDayMetricsBucketsCount.enabledCount,
+                );
+                variantMetricsBucketsPreviousDay.reset();
+                variantMetricsBucketsPreviousDay.set(
+                    stats.previousDayMetricsBucketsCount.variantCount,
+                );
```
FredrikOseberg, line 253: Does this need to be reset and set?

src/lib/features/instance-stats/instance-stats-service.ts
```diff
@@ -244,6 +255,7 @@ export class InstanceStatsService {
             this.eventStore.filteredCount({ type: FEATURES_EXPORTED }),
             this.eventStore.filteredCount({ type: FEATURES_IMPORTED }),
             this.getProductionChanges(),
+            this.clientMetricsStore.countPreviousDayMetricsBuckets(),
```
FredrikOseberg, line 258: Does it make sense to have this in instance stats?

FredrikOseberg (commented):
I'm wondering if this could be a static counter outside of instance stats, or if there is some reason we need to store it and reset and set it.

src/lib/metrics.ts
kwasniew: I though that's what we do for all the other metrics. We need to update this one once per day. Other resets are not necessary. On the other hand it's just a count call under the hood.

src/lib/features/instance-stats/instance-stats-service.ts
kwasniew: I wanted to have debug ability both in the API and prometheus. Do you have any example of a metric that is not either on instance stats or calculated after event emitter?

FredrikOseberg (approved):
LGTM

PR #5888, opened 2024-01-12
fix: role form sluggishness

nunogois:
This seems to improve the performance in the role form while still maintaining the same validation logic.

A big factor was the memoization of the categories calculation and respective elements, which is especially impactful when there are many environments.


frontend/src/component/admin/roles/RoleForm/useRoleForm.ts
```diff
@@ -79,7 +57,7 @@ export const useRoleForm = (
     });
 
     const isNameUnique = (name: string) => {
-        return !roles.some(
+        return ![...roles, ...projectRoles].some(
```
nunogois, line 60: Girlscouting: Only root roles were being checked for uniqueness. This now checks both.

sjaanus (approved):
LG

PR #5893, opened 2024-01-15
chore: align with system user in db

thomasheartman:
This change removes the system user's email from the definition, instead setting it to `null`. It also changes the name to "Unleash System".

The IUser interface doesn't allow `null` email addresses, so we change the type definition of the system user to get around it. However, using `null` (instead of just removing the property entirely) is useful because when you get the system user from the DB, it's email value will be null (after it has been nulled out).

As of today, there is nowhere in the Unleash system (OSS or Enterprise) where we use the system user as an IUser (we only use username and ID). So this change shouldn't break anything.

This should follow https://github.com/Unleash/unleash/pull/5849.


daveleek (approved):
Yup, Looks good!

PR #5896, opened 2024-01-15
fix: differentiated interval and initial schedule call

chriswk:
Differentiate log lines so we can see if it happens on every call, or just on the initial call.


gardleopard (approved):
LGTM

PR #5898, opened 2024-01-15
fix: include custom root roles in user access overview

nunogois:
https://linear.app/unleash/issue/2-1844/fix-add-custom-root-roles-to-user-access-overview

I noticed our user access overview method did not take into account custom root roles, which meant only users with default root roles were being returned.

This changes the query to check for `IN ('root', 'root-custom')` instead, including both "root" and "custom root" roles.

image

### DIFF
--- a/src/lib/db/access-store.ts
+++ b/src/lib/db/access-store.ts
@@ -918,7 +918,9 @@ export class AccessStore implements IAccessStore {
                     SELECT r.name
                     FROM   role_user ru
                     INNER JOIN roles r on ru.role_id = r.id
-                    WHERE ru.user_id = u.id and r.type='root'
+                    WHERE ru.user_id = u.id and r.type IN (${ROOT_ROLE_TYPES.map(
+                        (type) => `'${type}'`,
+                    ).join(',')})
                 ) r, LATERAL (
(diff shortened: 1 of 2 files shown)


src/lib/db/access-store.ts
```diff
@@ -917,7 +918,9 @@ export class AccessStore implements IAccessStore {
                     SELECT r.name
                     FROM   role_user ru
                     INNER JOIN roles r on ru.role_id = r.id
-                    WHERE ru.user_id = u.id and r.type='root'
+                    WHERE ru.user_id = u.id and r.type IN (${ROOT_ROLE_TYPES.map(
+                        (type) => `'${type}'`,
+                    ).join(',')})
```
nunogois, line 923: This is our own constant, so it should be safe to inject it like this, but I'm open to suggestions.

sighphyre (commented):
Any chance of a test for this?

src/lib/db/access-store.ts
sighphyre: Seems sane to me, since we own the type

sighphyre (approved):
Awesome, thank you for fixing this!

PR #5903, opened 2024-01-16
chore: add tokens column to incoming webhooks table

nunogois:
https://linear.app/unleash/issue/2-1827/add-remaining-columns-to-the-incoming-webhooks-table

Adds the "tokens" column to the incoming webhooks table.

Also includes some slight adjustments to the table, including a fix to center the actions button.

image


chriswk (approved):
LGTM

PR #5905, opened 2024-01-16
feat: added killswitch for admin tokens

chriswk:
Since we've now added PAT's we really do recommend switching to those, or for enterprises, we recommend using service accounts. 

Admin tokens have an obvious disadvantage in that they're not connected to any user, so actions performed by them are harder to audit.

This PR adds a killswitch for turning it off, in preparation for deprecating them and ultimately removing them in the future.


gastonfournier (approved):
LGTM

PR #5906, opened 2024-01-16
feat: prevent double daily metrics insert

kwasniew:
## About the changes

Prevent repeated calculation of previous day daily metrics. E.g. when multiple pods start calculation only the first pod should calculate the metrics.

Details:
* adding new method to count how many daily metrics buckets we have for the previous day
* heuristic: if we have more than count of hourly buckets / 24 of daily buckets then we do not proceed with daily metrics calculation
* added 3 test scenarios: too many hourly buckets, daily metrics already calculated and happy path calculate new daily metrics

### DIFF
--- a/src/lib/db/client-metrics-store-v2.test.ts
+++ b/src/lib/db/client-metrics-store-v2.test.ts
@@ -216,7 +216,8 @@ test('count previous day metrics', async () => {
         },
     ]);
 
-    const result = await clientMetricsStore.countPreviousDayMetricsBuckets();
+    const result =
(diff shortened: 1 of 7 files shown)


kwasniew:
@FredrikOseberg This is fine. Currently this is the case. All pods will calculate the same result and overwrite the previous one with the same value. We want to avoid calculating the same thing twice and updating it.
```
          ON CONFLICT (feature_name, app_name, environment, date, variant)
          DO UPDATE SET count = EXCLUDED.count;
```

PR #5909, opened 2024-01-16
docs: add deprecation notice to Admin tokens

chriswk:
We want to deprecate and eventually remove the option for admin tokens. This PR updates the docs to inform readers about the deprecation as well as recommending what they can use instead.


daveleek (approved):
LG! I like the explanation for the deprecation!

PR #5910, opened 2024-01-16
chore: centralize events service creation

gastonfournier:
## About the changes
EventsService is a dependency in most of our services. This creates helper methods to create them easily and replace a few places where we're creating them manually


kwasniew (approved):
Yesss!

PR #5912, opened 2024-01-16
fix: should not import archived child and parent

kwasniew:
## About the changes

Do not import archived parent or archived child.


sjaanus (approved):

sjaanus:
LG

PR #5914, opened 2024-01-16
Dashboard stats db tables

Tymek:


src/migrations/20240116150504-create-stat-users-table.js
```diff
@@ -0,0 +1,18 @@
+exports.up = function(db, cb) {
+    db.runSql(`
+        CREATE TABLE stat_users(
+                created_at TIMESTAMP NOT NULL PRIMARY KEY,
```
sjaanus, line 4: Not sure how we are inserting data, but if data is added in same batch or transaction, you might get same timestamp and that breaks primary key. Any specific reason why primary key is created_at?

PR #5923, opened 2024-01-17
fix: metrics apps should be selected from the longest time span

kwasniew:
## About the changes

Select applications from the longest time span. Previously it was 48 hours. Now it's 3 months. Without this fix we're not showing all app names for longer time spans.


FredrikOseberg (approved):
LGTM

PR #5924, opened 2024-01-17
feat: admin token calls get an admin token user

gastonfournier:
## About the changes
Whenever we get a call from an admin token we want to associate it with the [admin token user](https://github.com/Unleash/unleash/blob/4d42093a07a86dcff1766425c8429b20dade8e0d/src/lib/types/core.ts#L34-L41). This should give us the needed audit for this type of calls that currently were lacking a user id (we only stored a string with the token name in the event log).

We consciously decided not to use `id` as the property to prevent any unforeseen side effects. The reason is that only `IUser` type has an id and adding an id to `IApiUser` might lead to confusion.


src/lib/routes/admin-api/api-token.ts
```diff
@@ -312,6 +312,7 @@ export class ApiTokenController extends Controller {
             const token = await this.apiTokenService.createApiToken(
                 createToken,
                 extractUsername(req),
+                extractUserId(req),
```
gastonfournier, line 315: Don't fall back to the system user. I haven't found this being used elsewhere but the default value set to system user is a risk because it's not considering that an admin token should get the admin token user instead

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -190,6 +190,7 @@ export class ProjectApiTokenController extends Controller {
             const token = await this.apiTokenService.createApiToken(
                 createToken,
                 extractUsername(req),
+                extractUserId(req),
```
gastonfournier, line 193: Don't fallback to system user

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -190,6 +204,35 @@ test('Token-admin should be allowed to create token', async () => {
     await destroy();
 });
 
+test('An admin token should be allowed to create a token', async () => {
```
gastonfournier, line 207: @chriswk adding this additional case, which is different from the setup of the other tests, to use the authorization header instead of the preHook that injects a user (not a token). Let me know what you think about this, maybe the test should be in a different test-suite

src/lib/services/api-token-service.ts
```diff
@@ -221,8 +232,19 @@ export class ApiTokenService {
 
     public async createApiTokenWithProjects(
         newToken: Omit<IApiTokenCreate, 'secret'>,
-        createdBy: string = SYSTEM_USER.username,
-        createdByUserId: number = SYSTEM_USER.id,
+        createdBy: IApiUser | IUser,
```
gastonfournier, line 235: This is a breaking change on a public method but was just added in 5.8 https://github.com/Unleash/unleash/commit/bfa82d79bf245058612081062345e0ca7c38b7b4

Alternatively, I could do it in a compatible way with this signature and validate that the field is of type string
```suggestion
        createdBy: string | IApiUser | IUser,
```

gastonfournier: NVM, I had to support this in [869f4ba](https://github.com/Unleash/unleash/pull/5924/commits/869f4ba4cca2ef5b54e00a26ea36de1c18fa4651) due to failing tests. Maybe this can be rewritten but it's also a public method that can be used to extend Unleash.

src/lib/util/extract-user.test.ts
```diff
@@ -22,11 +23,11 @@ describe('extractUsernameFromUser', () => {
     test('Should return "unknown" if neither email nor username exists', () => {
```
daveleek, line 23: I guess the title of the test is slightly incorrect after this change :)

src/lib/util/extract-user.test.ts
```diff
@@ -22,11 +23,11 @@ describe('extractUsernameFromUser', () => {
...
         const user = {} as IUser;
 
-        expect(extractUsernameFromUser(user)).toBe('unknown');
+        expect(extractUsernameFromUser(user)).toBe(SYSTEM_USER.username);
     });
 
     test('Should return "unknown" if user is null', () => {
```
daveleek, line 29: Should return system user?

daveleek (approved):
Nice, I think this is a great improvement! Just pointed out a couple of unit tests whose names could probably be improved after this fix :)

src/lib/util/extract-user.test.ts
gastonfournier: Spot on!

src/lib/util/extract-user.test.ts
gastonfournier: Just the username

gastonfournier:
On top of this one there's https://github.com/Unleash/unleash/pull/5933

PR #5927, opened 2024-01-17
fix: select default application based on timespan

kwasniew:
## About the changes

We cannot select apps from the longest timespan (3 months) since we may have new app names in the last 24h that are not present in the longer timespan yet. So I decided to recalculate default application when the list of available applications changes.

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureMetrics.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureMetrics.tsx
@@ -3,7 +3,6 @@ import { PageContent } from 'component/common/PageContent/PageContent';
 import { useEffect, useMemo, useState } from 'react';
 import {
     FEATURE_METRIC_HOURS_BACK_DEFAULT,
-    FEATURE_METRIC_HOURS_BACK_MAX,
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureMetrics.tsx
```diff
@@ -22,26 +21,41 @@ import {
     withDefault,
 } from 'use-query-params';
 import { aggregateFeatureMetrics } from './aggregateFeatureMetrics';
-import { useExtendedFeatureMetrics } from './useExtendedFeatureMetrics';
 
 export const FeatureMetrics = () => {
     const projectId = useRequiredPathParam('projectId');
     const featureId = useRequiredPathParam('featureId');
     const environments = useFeatureMetricsEnvironments(projectId, featureId);
-    const applications = useFeatureMetricsApplications(featureId);
```
kwasniew, line 31: applications we get depend on time span now so we can't select the longest anymore. daily timespans are aggregated with up to 24h delay so we may not have some apps in the daily timespans

FredrikOseberg (approved):
LGTM

PR #5928, opened 2024-01-17
feat: metrics periods expressed in days

kwasniew:
## About the changes

Switching week, month and 3 months into days metrics to be consistent and have a simpler calculation model. 
Screenshot 2024-01-17 at 11 01 47


src/lib/create-config.ts
```diff
@@ -553,7 +553,7 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
 
     const dailyMetricsStorageDays = Math.min(
         parseEnvVarNumber(process.env.DAILY_METRICS_STORAGE_DAYS, 31),
-        92,
+        91,
```
kwasniew, line 556: store one more day t

nicolaesocaciu (commented):
I think this makes sense 👍

FredrikOseberg (approved):

PR #5929, opened 2024-01-17
fix: store redirect to localStorage to avoid loss of redirect

andreas-unleash:
Stores `redirect` param in localStorage in the Authentication component.
Retrieves the `redirect` param from localStorage at the Login screen if it is not there in the url.

This will solve losing the redirect information all provider logins

Closes # [1-1890](https://linear.app/unleash/issue/1-1890/capture-path-before-logging-in-and-redirect-to-it-if-there-and-custom)


frontend/src/component/user/Login/Login.tsx
```diff
@@ -26,7 +27,8 @@ const Login = () => {
     const query = useQueryParams();
     const resetPassword = query.get('reset') === 'true';
     const invited = query.get('invited') === 'true';
-    const redirect = query.get('redirect') || '/';
+    const redirect =
```
ivarconr, line 30: I believe we need to clear the parameter after a successful login. Session storage might be more suited as it is cleared on logout. Should we also consider a TTL?

andreas-unleash: we clear the local storage on log out and we use it over session pretty consistently.  A short TTL makes sense since this is essentially aimed at holding this info until we get the callback from the auth providers

andreas-unleash: Implemented both clear on successful login and ttl of 10'

FredrikOseberg: @andreas-unleash I think session storage is more correct here, the redirect will only apply for the session the user initialises with the url. With local storage we might get into weird scenarios even with a 10 minute TTL. For example what happens if you load a page with the redirect url and then you close the tab? If you open a new tab and go to unleash I expect you'll read local storage and be redirected.

andreas-unleash: fair. I will change it

frontend/src/component/user/Login/Login.tsx
```diff
@@ -26,9 +27,11 @@ const Login = () => {
...
+        query.get('redirect') || getSessionStorageItem('login-redirect') || '/';
 
     if (user) {
+        setSessionStorageItem('login-redirect', undefined);
```
FredrikOseberg, line 34: Could we default to undefined so we don't need to pass the second argument here?

frontend/src/utils/storage.test.ts
```diff
@@ -0,0 +1,120 @@
+import {
+    setLocalStorageItem,
+    getLocalStorageItem,
+    setSessionStorageItem,
+    getSessionStorageItem,
+} from './storage';
+import { vi } from 'vitest';
+
+// Mocking the global localStorage
+const localStorageMock = (() => {
```
FredrikOseberg, line 10: Isn't this obsolete now?

FredrikOseberg (approved):
LGTM

PR #5931, opened 2024-01-17
fix: make the conflcit email template button conditionally render

andreas-unleash:
Makes the schedule conflict email button conditional to having the link present.

Closes # [1-1950](https://linear.app/unleash/issue/1-1950/modify-the-template-to-make-the-new-button-conditional)

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/FeatureStrategyConstraints.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/FeatureStrategyConstraints.tsx
@@ -1,5 +1,5 @@
 import { IConstraint, IFeatureStrategy } from 'interfaces/strategy';
-import React, { useEffect, useMemo } from 'react';
+import React, { useEffect } from 'react';
 import {
     UPDATE_FEATURE_STRATEGY,
     CREATE_FEATURE_STRATEGY,
--- a/src/lib/services/email-service.ts
+++ b/src/lib/services/email-service.ts
@@ -146,7 +146,7 @@ export class EmailService {
     async sendScheduledChangeConflictEmail(
         recipient: string,
         conflictScope: 'flag' | 'strategy',
-        conflictingChangeRequestId: number,
+        conflictingChangeRequestId: number | undefined,
         changeRequests: {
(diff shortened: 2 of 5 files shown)


src/lib/services/email-service.ts
```diff
@@ -173,7 +173,9 @@ export class EmailService {
                 ? `${this.config.server.unleashUrl}/projects/${project}/archive?sort=archivedAt&search=${flagName}`
                 : false;
 
-            const conflictingChangeRequestLink = `${this.config.server.unleashUrl}/projects/${project}/change-requests/${conflictingChangeRequestId}`;
+            const conflictingChangeRequestLink = conflictingChangeRequestId
+                ? `${this.config.server.unleashUrl}/projects/${project}/change-requests/${conflictingChangeRequestId}`
+                : false;
```
kwasniew, line 178: maybe null instead to avoid string | boolean type?

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/FeatureStrategyConstraints.tsx
```diff
@@ -1,5 +1,5 @@
 import { IConstraint, IFeatureStrategy } from 'interfaces/strategy';
-import React, { useEffect, useMemo } from 'react';
+import React, { useEffect } from 'react';
```
kwasniew, line 2: why are we adding it if it's not used?

src/lib/services/email-service.ts
andreas-unleash: This is needed for the mustashe template conditional check

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/FeatureStrategyConstraints.tsx
andreas-unleash: this is from biome removing it

src/lib/services/email-service.ts
andreas-unleash: The tags in mustashe will not render when `false` . Not sure if null will act the same, I think it doesn't

PR #5932, opened 2024-01-17
feat: added more granular project permissions

chriswk:
### What
Adds Read and Write permissions for project administration settings (user access, change request settings, default strategy, other).

### Why
On request from two large customers that wanted our RBAC controls to be more granulated to easier be able to limit the access they granted their users.


gastonfournier (approved):
LG!

PR #5933, opened 2024-01-17
chore: change event service method

gastonfournier:
## About the changes
The new method makes it easy to store events without having to know the details about how we define the username or userId

### DIFF
--- a/src/lib/services/event-service.ts
+++ b/src/lib/services/event-service.ts
@@ -2,10 +2,11 @@ import { IUnleashConfig } from '../types/option';
 import { IFeatureTagStore, IUnleashStores } from '../types/stores';
 import { Logger } from '../logger';
 import { IEventStore } from '../types/stores/event-store';
-import { IBaseEvent, IEventList } from '../types/events';
+import { IBaseEvent, IEventList, IUserEvent } from '../types/events';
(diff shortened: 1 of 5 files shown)


src/lib/services/event-service.ts
```diff
@@ -81,15 +82,61 @@ export default class EventService {
         return events;
     }
 
+    isAPI(user: IUser | IApiUser): boolean {
+        return (user as IApiUser)?.type === ApiTokenType.ADMIN;
```
daveleek, line 86: This will return false for frontend and backend tokens, which I expect is intentional, but I wonder if we could reflect that with the name perhaps

daveleek (approved):
Nice! 👍

src/lib/services/event-service.ts
```diff
@@ -81,15 +82,60 @@ export default class EventService {
...
+
+    getUserDetails(user: IUser | IApiUser): {
+        createdBy: string;
+        createdByUserId: number;
+    } {
+        return {
+            createdBy:
+                (user as IUser)?.email ||
+                user?.username ||
+                (this.isAdminToken(user)
```
ivarconr, line 97: Why is this or clause needed. Won't a ApiUser always have a username?

PR #5934, opened 2024-01-17
chore : upgrade jsdom to 23

sjaanus:


sjaanus:
Bumping to 23.1.0 because 23.2.0 has big performance issues https://github.com/jsdom/jsdom/issues/3659

sighphyre (approved):
LGTM

PR #5937, opened 2024-01-17
fix: table loader

kwasniew:
## About the changes

Bug: loaders in the table were cancelled before data arrived

Root cause:
Screenshot 2024-01-17 at 16 14 38

Solution:
Give a loader at the top level (in Project) a very specific identifier for things it should load so that very deeply nested loaders are not affected and can have their local loading hooks.


nunogois (approved):
LGTM

PR #5938, opened 2024-01-17
Chore inc webhooks modal form

nunogois:
https://linear.app/unleash/issue/2-1818/ui-create-incoming-webhook-newedit-modal

Adds the incoming webhooks modal form, which allows users to create and edit incoming webhooks, along with their respective tokens.

Follows a logic similar to service accounts and their tokens, and tries to use the newest form validation flow that we implemented in the roles form.

image

### DIFF
--- a/frontend/src/component/admin/banners/BannerModal/BannerForm.tsx
+++ b/frontend/src/component/admin/banners/BannerModal/BannerForm.tsx
@@ -10,7 +10,7 @@ import { ChangeEvent, Dispatch, SetStateAction, useState } from 'react';
 import { Visibility } from '@mui/icons-material';
 import { BannerDialog } from 'component/banners/Banner/BannerDialog/BannerDialog';
 
-const StyledForm = styled('form')(({ theme }) => ({
+const StyledForm = styled('div')(({ theme }) => ({
     display: 'flex',
     flexDirection: 'column',
     gap: theme.spacing(4),
--- a/frontend/src/component/admin/serviceAccounts/ServiceAccountsTable/ServiceAccountModal/ServiceAccountTokens/ServiceAccountTokens.tsx
+++ b/frontend/src/component/admin/serviceAccounts/ServiceAccountsTable/ServiceAccountModal/ServiceAccountTokens/ServiceAccountTokens.tsx
@@ -218,7 +218,7 @@ export const ServiceAccountTokens = ({
 
     const { headerGroups, rows, prepareRow, setHiddenColumns } = useTable(
         {
-            columns,
+            columns: columns as any[],
             data,
             initialState,
             sortTypes,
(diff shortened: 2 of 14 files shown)


frontend/src/component/admin/banners/BannerModal/BannerForm.tsx
```diff
@@ -10,7 +10,7 @@ import { ChangeEvent, Dispatch, SetStateAction, useState } from 'react';
 import { Visibility } from '@mui/icons-material';
 import { BannerDialog } from 'component/banners/Banner/BannerDialog/BannerDialog';
 
-const StyledForm = styled('form')(({ theme }) => ({
+const StyledForm = styled('div')(({ theme }) => ({
```
nunogois, line 13: Scouting: This meant that a `form` would be inside another `form`, which caused a console error.

frontend/src/component/admin/serviceAccounts/ServiceAccountsTable/ServiceAccountModal/ServiceAccountTokens/ServiceAccountTokens.tsx
```diff
@@ -228,7 +218,7 @@ export const ServiceAccountTokens = ({
 
     const { headerGroups, rows, prepareRow, setHiddenColumns } = useTable(
         {
-            columns,
+            columns: columns as any[],
```
nunogois, line 221: Scouting: This had a TS error. The diff above is caused by the removal of `as Column<IPersonalAPIToken>[]` - It should otherwise be the same. `columns: columns as any[]` is more consistent with what we usually do anyways.

frontend/src/component/incomingWebhooks/IncomingWebhooksModal/IncomingWebhooksForm/useIncomingWebhooksForm.ts
```diff
@@ -0,0 +1,136 @@
+import { useIncomingWebhooks } from 'hooks/api/getters/useIncomingWebhooks/useIncomingWebhooks';
+import { IIncomingWebhook } from 'interfaces/incomingWebhook';
+import { useEffect, useState } from 'react';
+
+const INCOMING_WEBHOOK_NAME_REGEX = /^[A-Za-z0-9\-_]*$/;
```
gastonfournier, line 5: Do we have this validation in the backend? It's nice we do these validations in the frontend but we should also have them in the backend. In general, I don't like to repeat validations both in frontend and backend, but this gives a quick response to the user. I'd consider server validation over frontend validation.

gastonfournier (approved):
Looks good to me, just a minor comment around validations but we can address it later

frontend/src/component/incomingWebhooks/IncomingWebhooksModal/IncomingWebhooksForm/useIncomingWebhooksForm.ts
nunogois: We do have it, shared 1on1. 

The server validation is actually less strict than this one, since it only checks for url safeness. Surprisingly, some less common characters are considered URL safe. 

This one is stricter, only checking for alphanumeric characters, `-` and `_`, which seemed a bit more sane to me. I like the fact that the UI _suggests_ this set of characters while the server technically accepts anything that is URL safe, but we can make them act the same way if you think that's a better approach.

nunogois: Regarding repeating them, this feels saner to me than pinging the server in _real-time_ every time the value changes.

Issue #4303, opened 2023-07-21
Migrations fail in upgrade from 4.x to 5.x with replicated database


william00179:
### Describe the bug

When the automatic migrations run on boot after upgrading from 4.x to 5.8.2 the following error is observed on startup.

```
[2023-07-21T03:14:10.526] [ERROR] server-impl.js - Failed to migrate db error: cannot delete from table "role_permission" because it does not have a replica identity and publishes deletes
    at Parser.parseErrorMessage (/unleash/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/unleash/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/unleash/node_modules/pg-protocol/dist/parser.js:39:38)
    at TLSSocket.<anonymous> (/unleash/node_modules/pg-protocol/dist/index.js:11:42)
    at TLSSocket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:297:9)
    at Readable.push (node:internal/streams/readable:234:10)
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:190:23) {
  length: 261,
  severity: 'ERROR',
  code: '55000',
  detail: undefined,
  hint: 'To enable deleting from the table, set REPLICA IDENTITY using ALTER TABLE.',
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'execReplication.c',
  line: '641',
  routine: 'CheckCmdReplicaIdentity'
}
[ERROR] error: cannot delete from table "role_permission" because it does not have a replica identity and publishes deletes
    at Parser.parseErrorMessage (/unleash/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/unleash/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/unleash/node_modules/pg-protocol/dist/parser.js:39:38)
    at TLSSocket.<anonymous> (/unleash/node_modules/pg-protocol/dist/index.js:11:42)
    at TLSSocket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:297:9)
    at Readable.push (node:internal/streams/readable:234:10)
[...]

PR #5943, opened 2024-01-18
feat: add primary keys to missing tables

sjaanus:
Follow up of https://github.com/Unleash/unleash/issues/4303

We are adding primary keys to all tables missing them, currently **role_permission**, **api_token_project**, and **project_stats**. 
By adding primary keys, the issue with migrations failing during upgrades in replicated database setups will be resolved.

### DIFF
--- a/src/migrations/20240118093611-missing-primary-keys.js
+++ b/src/migrations/20240118093611-missing-primary-keys.js
@@ -4,7 +4,7 @@ exports.up = function (db, callback) {
     db.runSql(
         `
         ALTER TABLE project_stats ADD PRIMARY KEY (project);
-        ALTER TABLE api_token_project ADD PRIMARY KEY (secret);
+        ALTER TABLE api_token_project ADD COLUMN id SERIAL PRIMARY KEY;
         ALTER TABLE role_permission ADD COLUMN id SERIAL PRIMARY KEY;
         `,
         callback,
@@ -16,6 +16,7 @@ exports.down = function (db, callback) {
         `
         ALTER TABLE project_stats DROP CONSTRAINT project_stats_pkey;
         ALTER TABLE api_token_project DROP CONSTRAINT api_token_project_pkey;
+        ALTER TABLE api_token_project DROP COLUMN id;
         ALTER TABLE role_permission DROP CONSTRAINT role_permission_pkey;
         ALTER TABLE role_permission DROP COLUMN id;
         `,
--- a/src/migrations/20240118093611-missing-primary-keys.js
+++ b/src/migrations/20240118093611-missing-primary-keys.js
@@ -1,0 +1,24 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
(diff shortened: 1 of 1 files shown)


src/migrations/20240118093611-missing-primary-keys.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+        ALTER TABLE project_stats ADD PRIMARY KEY (project);
```
sjaanus, line 6: Adding on unique existing column

src/migrations/20240118093611-missing-primary-keys.js
```diff
@@ -0,0 +1,24 @@
...
+        ALTER TABLE api_token_project ADD PRIMARY KEY (secret);
```
sjaanus, line 7: Adding on unique existing column

src/migrations/20240118093611-missing-primary-keys.js
```diff
@@ -0,0 +1,24 @@
...
+        ALTER TABLE role_permission ADD COLUMN id SERIAL PRIMARY KEY;
```
sjaanus, line 8: No good column to add, so creating id column.

sighphyre (approved):
LGTM

src/migrations/20240118093611-missing-primary-keys.js
gastonfournier: Can we have a combined key? role_id + permission

sjaanus: Nope, there is also environment column. So it would have to be triple column primary key.

src/migrations/20240118093611-missing-primary-keys.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+        ALTER TABLE project_stats ADD PRIMARY KEY (project);
+        ALTER TABLE api_token_project ADD COLUMN id SERIAL PRIMARY KEY;
```
sjaanus, line 7: @gastonfournier in this case I would use secret and project here also, if we using combined keys.

src/migrations/20240118093611-missing-primary-keys.js
sjaanus: And no after checking, we can not use it, because environment is nullable and so is permission. We can not make environment not nullable, because root roles have no environment. So adding id seems only solution. @gastonfournier

gastonfournier: ok, I appreciate you checking this!

gastonfournier: Will it populate past rows?

sjaanus: Yes, serial will auto populate it

gastonfournier: > Yes, serial will auto populate it

TIL!

PR #5947, opened 2024-01-18
chore: re use extract user methods

gastonfournier:
## About the changes
1. Re-use existing methods in extract-user.ts: https://github.com/Unleash/unleash/blob/70f6a07f2c2ad1c39ee7bcc1d4cf35d39c285b67/src/lib/features/events/event-service.ts#L93-L101
2. Move event-service and event-store to features/event
3. Add export default in previous paths for backward compatibility: https://github.com/Unleash/unleash/blob/70f6a07f2c2ad1c39ee7bcc1d4cf35d39c285b67/src/lib/services/event-service.ts#L1-L4 and https://github.com/Unleash/unleash/blob/70f6a07f2c2ad1c39ee7bcc1d4cf35d39c285b67/src/lib/db/event-store.ts#L1-L4


nunogois (approved):
LGTM - Looking forward to see the cleanup once we adapt Enterprise to the new files!

gastonfournier:
> LGTM - Looking forward to see the cleanup once we adapt Enterprise to the new files!

https://github.com/ivarconr/unleash-enterprise/pull/1003

PR #5953, opened 2024-01-18
fix: normalize hours back to prevent freezing UI

kwasniew:
## About the changes

If user types invalid hoursBack into the URL it always goes back to a default value of 48 hours. It prevents UI from freezing when someone tinkers with the URL.
Screenshot 2024-01-18 at 13 55 34


FredrikOseberg (approved):
LGTM

PR #5956, opened 2024-01-18
docs/improve website README.md with local docs server installation step

markunl:
Minor addition to set correct local dir context in terminal when setting up a local docs instance

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -4,12 +4,10 @@ This website is built using [Docusaurus 2](https://docusaurus.io/), a modern sta
 
 ## Installation
 
-In a terminal, cd into the website folder of the locally cloned unleash repository. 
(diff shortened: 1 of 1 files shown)


nnennandukwe (approved):

website/README.md
```diff
@@ -4,6 +4,8 @@ This website is built using [Docusaurus 2](https://docusaurus.io/), a modern sta
 
 ## Installation
 
+In a terminal, cd into the website folder of the locally cloned unleash repository. Then execute:
```
thomasheartman, line 7: Maybe we should add the command for changing directories too? 

Something like this, maybe?

```suggestion
In a terminal, cd into the website folder of the locally cloned unleash repository. 

```console
cd unleash/website
```

Then install the dependencies:
```

thomasheartman (approved):

PR #5960, opened 2024-01-18
fix: admin token should be passed forward from controllers 

sjaanus:
We were sending `user.id` to the service, but if an admin token is used, there is no `user.id.` Instead, there is `user.internalAdminTokenUserId`. so we need to use the special method `extractUserIdFromUser`.

This PR adds this implementation, and now the service correctly retrieves the appropriate ID for admins.

Related to: https://github.com/Unleash/unleash/pull/5924


gastonfournier (approved):
LGTM!

PR #5963, opened 2024-01-19
feat: add algorithm to detect what strategy changes would be overwritten by applying a CR

thomasheartman:
This change adds an algorithm with tests for detecting what changes
would be overwritten by applying a CR.

Test cases:
- It compares strategies regardless of order of keys in the objects.
This ensures that two strategies with the same content but different
order of keys are compared correctly.
- It treats `undefined` or missing segments in old config as equal to `[]` in change
- It treats `undefined` or missing strategy variants in old config and
change as equal to `[]`
- It lists changes in a sorted list with the correct values
- It ignores object order on nested objects. Similar to the first
point, this does order-insensitive comparison for nested objects (such
as params and constraints).


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
```diff
@@ -0,0 +1,82 @@
+import { IChangeRequestUpdateStrategy } from 'component/changeRequest/changeRequest.types';
+import { IFeatureStrategy } from 'interfaces/strategy';
+import isEqual from 'lodash.isequal';
+
+const hasJsonDiff = (object: unknown, objectToCompare: unknown) =>
+    JSON.stringify(object) !== JSON.stringify(objectToCompare);
```
thomasheartman, line 6: This function also exists as `hasDiff` in the StrategyChange.tsx component, but I'm not sure they'll always change at the same rate, so I'm copying it in here slightly renamed.

andreas-unleash: Why is there ported to the BE?

andreas-unleash (approved):

PR #5964, opened 2024-01-19
feat: show changes that would be overwritten in change request overview UI

thomasheartman:
This PR adds a first, rough iteration of what it could look like to show changes that would be overwritten by applying a PR.

The changes are listed in a table (semantically; looks more like a list visually) and show the property, the current live value and the version that you have in your changes. The changes are hidden by default, but can be shown by expanding a details element.

@nicolaesocaciu Suggested that we merge this version for now and iterate on the design later.

Here's what it looks like closed:
image

Here's what it looks like with a typical change load:
image

Here's what it looks like if you change more or less every property changed:
image


andreas-unleash (approved):
LG

PR #5969, opened 2024-01-19
refactor: add typesafe wrappers for prom client metrics

nunogois:
https://linear.app/unleash/issue/2-1856/add-typesafe-wrappers-over-prom-clients-metrics

As discussed on the latest knowledge sharing session, this adds typesafe wrappers over prom client's metrics, requiring us to specify all the configured labels for each metric.

This uses a functional approach and only exposes the methods that are currently relevant to us, while also exposing the underlying instance of the metric for an easy access if needed.

Since we often chain `labels` with `inc` in counters, this adds a convenience `increment` method for counters which does both in a single call.

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -1,4 +1,4 @@
-import client from 'prom-client';
+import { collectDefaultMetrics } from 'prom-client';
 import memoizee from 'memoizee';
 import EventEmitter from 'events';
 import { Knex } from 'knex';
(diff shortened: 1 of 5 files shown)


sighphyre (commented):
Ay I really like this!

src/lib/metrics.ts
```diff
@@ -372,28 +375,40 @@ export default class MetricsMonitor {
...
         });
         eventStore.on(FEATURE_UPDATED, ({ featureName, project }) => {
-            featureToggleUpdateTotal
-                .labels(featureName, project, 'default', 'production')
-                .inc();
+            featureToggleUpdateTotal.increment({
+                toggle: featureName,
+                project,
+                environment: 'default',
+                environmentType: 'production',
```
nunogois, line 410: I assume these `environment` and `environmentType` values are on purpose? I found it weird that these ones are specified, but not the others.

PR #5973, opened 2024-01-19
fix: incoming webhooks form UI adjustments

nunogois:
Small UI improvements in the new incoming webhooks form after aligning with @nicolaesocaciu.

image


nicolaesocaciu (approved):
LGTM!

PR #5974, opened 2024-01-19
task: Make project permissions more fine grained

chriswk:
Connected to [#5932](https://github.com/Unleash/unleash/pull/5932) - This starts using the new permissions in addition to the old UPDATE_PROJECT permission. That way, if you're happy with UPDATE_PROJECT, you don't need to change.

However, you can now add more fine grained permissions for both READ and WRITE operations.


nunogois (approved):
LGTM!

PR #5982, opened 2024-01-22
fix: unauthorized disable feature

gastonfournier:
## About the changes
This was spotted while testing automated actions. Steps to reproduce:

1. Add an editor user
2. Get a PAT for the editor user
3. As Admin create a feature in a project where the editor user is not a member and enable the feature
4. Try using the editor's PAT to modify the feature: ```$ curl -XPOST -H 'Authorization: user:editor-pat' https://sandbox.getunleash.io/enterprise/api/admin/projects/gaston-editor/features/editor-feature/environments/development/off | jq
{
  "id": "444d157c-dd85-4953-afec-781cde9d029f",
  "name": "PermissionError",
  "message": "You don't have the required permissions to perform this operation. To perform this action, you need the \"UPDATE_FEATURE_ENVIRONMENT\" permission.",
  "details": [
    {
      "message": "You don't have the required permissions to perform this operation. To perform this action, you need the \"UPDATE_FEATURE_ENVIRONMENT\" permission.",
      "description": "You don't have the required permissions to perform this operation. To perform this action, you need the \"UPDATE_FEATURE_ENVIRONMENT\" permission."
    }
  ],
  "permissions": [
    "UPDATE_FEATURE_ENVIRONMENT"
  ]
}```

5. As the editor create a project (you'd be made owner) and try the same request but just change the project name for the new project just created (don't change anything else): ```$ curl -XPOST -I -H 'Authorization: editor-pat' https://sandbox.getunleash.io/enterprise/api/admin/projects/myown/features/editor-feature/environments/development/off 
HTTP/1.1 200 OK```

This does not happen when trying to turn on a flag because during the turn-on process we do validate if the feature belongs to project when we call updateStrategy: https://github.com/Unleash/unleash/blob/c18a7c0dc226f6b32a601a8d002dcd8f94d3c9af/src/lib/features/feature-toggle/feature-toggle-service.ts#L1751-L1764

### DIFF
--- a/src/lib/services/access-service.ts
+++ b/src/lib/services/access-service.ts
@@ -65,6 +65,9 @@ export type IdPermissionRef = Pick<IPermission, 'id' | 'environment'>;
 export type NamePermissionRef = Pick<IPermission, 'name' | 'environment'>;
 export type PermissionRef = IdPermissionRef | NamePermissionRef;
 
+type APIUser = Pick<IUser, 'id' | 'permissions'> & { isAPI: true };
+type NonAPIUser = Pick<IUser, 'id'> & { isAPI?: false };
+
 export interface IRoleCreation {
     name: string;
     description: string;
@@ -150,7 +153,7 @@ export class AccessService {
      * @param projectId
      */
     async hasPermission(
-        user: Pick<IUser, 'id' | 'permissions' | 'isAPI'>,
+        user: APIUser | NonAPIUser,
         permission: string | string[],
         projectId?: string,
         environment?: string,
@@ -198,7 +201,7 @@ export class AccessService {
     }
 
     async getPermissionsForUser(
-        user: Pick<IUser, 'id' | 'isAPI' | 'permissions'>,
+        user: APIUser | NonAPIUser,
     ): Promise<IUserPermission[]> {
         if (user.isAPI) {
             return user.permissions?.map((p) => ({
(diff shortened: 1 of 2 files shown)


src/lib/services/access-service.ts
```diff
@@ -65,6 +65,9 @@ export type IdPermissionRef = Pick<IPermission, 'id' | 'environment'>;
 export type NamePermissionRef = Pick<IPermission, 'name' | 'environment'>;
 export type PermissionRef = IdPermissionRef | NamePermissionRef;
 
+type APIUser = Pick<IUser, 'id' | 'permissions'> & { isAPI: true };
+type NonAPIUser = Pick<IUser, 'id'> & { isAPI?: false };
```
gastonfournier, line 69: This is unrelated but it's a QoL improvement. When using an API user this service reads the permissions from the user (they should be injected before this method is called), but when called with a non API user, we just ignore the permissions in the object and fetch them again from the DB

nunogois (approved):
Nice catch!

gardleopard (commented):
Is it possible to test this behaviour?

gastonfournier:
> Is it possible to test this behaviour?

It's possible to create an e2e test, but I think the manual test is good enough because this is a scouting thing. I don't want to tax this into the task I'm doing.

gastonfournier:
Merging because we can add the test later if we think it's valuable

PR #5986, opened 2024-01-22
feat: add option to use variants with feedback

FredrikOseberg:
This PR will allow us to use a feature flag with variants to control whether or not we should show the comments field of the feedback form. This will allow us to see whether we can increase feedback collection if we reduce the load on the customer.

### DIFF
--- a/frontend/src/component/feedbackNew/FeedbackComponent.tsx
+++ b/frontend/src/component/feedbackNew/FeedbackComponent.tsx
@@ -20,6 +20,7 @@ import { IToast } from 'interfaces/toast';
 import { useTheme } from '@mui/material/styles';
 import { FeedbackData, FeedbackMode } from './FeedbackContext';
 import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
+import { useUiFlag } from 'hooks/useUiFlag';
 
 export const ParentContainer = styled('div')(({ theme }) => ({
     position: 'relative',
@@ -206,6 +207,7 @@ export const FeedbackComponent = ({
     const { setHasSubmittedFeedback } = useUserSubmittedFeedback(
         feedbackData.category,
     );
+    const feedbackComments = useUiFlag('feedbackComments');
 
     function isProvideFeedbackSchema(data: any): data is ProvideFeedbackSchema {
         data.difficultyScore = data.difficultyScore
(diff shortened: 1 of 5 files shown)


frontend/src/component/feedbackNew/FeedbackComponent.tsx
```diff
@@ -349,50 +351,103 @@ export const FeedbackComponent = ({
...
-                                            }}
-                                            variant='outlined'
-                                            size='small'
-                                        />
-                                    </Box>
+
+                                    {feedbackComments !== false &&
+                                    feedbackComments.enabled &&
+                                    feedbackComments.name ===
+                                        'withoutTextfield' ? (
```
kwasniew, line 358: why is field lower case? Can we also keep the variant names more business centric (the role the component plays) not UI centric so that it's not coupled to whether we use text field/text area/dropdown?

src/lib/types/experimental.ts
```diff
@@ -208,6 +209,19 @@ const flags: IFlags = {
         process.env.UNLEASH_EXPERIMENTAL_EXECUTIVE_DASHBOARD,
         false,
     ),
+    feedbackComments: {
+        name: 'feedbackComments',
+        enabled: parseEnvVarBoolean(
+            process.env.UNLEASH_EXPERIMENTAL_FEEDBACK_COMMENTS,
+            false,
+        ),
+        payload: {
```
kwasniew, line 218: nice. is it a first time we do sth like this in our code?

kwasniew (approved):

src/lib/types/experimental.ts
FredrikOseberg: No, we also do this for the message banner. I took the pattern from there.

frontend/src/component/feedbackNew/FeedbackComponent.tsx
FredrikOseberg: I don't have strong feelings about this. I chose this name because in the context of the code it made perfect sense in order to explain what the condition does. What would prefer instead?

kwasniew: maybe withComment and withoutComment variants

PR #5989, opened 2024-01-22
fix: show the updated value instead of the snapshot value

thomasheartman:
This PR fixes a bug in the displayed value of the conflict list so that it shows the value it would update to instead of the snapshot value.

In doing so, it updates the logic of the algorithm to:

1. if the snapshot value and the current value are the same, it's not a conflict (it's an intended change)
2. If the snapshot value differs from the current value, it is a conflict if and only if the value in the change differs from the current value. Otherwise, it's not a conflict.

The new test cases are:
- it shows a diff for a property if the snapshot and live version differ for that property and the changed value is different from the live version
- it does not show a diff for a property if the live version and the change have the same value, even if the snapshot differs from the live version
- it does not show a diff for a property if the snapshot and the live version are the same

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
@@ -8,17 +8,17 @@ import omit from 'lodash.omit';
 
 type JsonDiffProps = {
     snapshotValue: unknown;
-    liveValue: unknown;
+    currentValue: unknown;
     changeValue: unknown;
     fallback?: unknown;
 };
 const hasJsonDiff = ({
     snapshotValue,
-    liveValue,
+    currentValue,
     changeValue,
     fallback,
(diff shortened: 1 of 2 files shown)


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
```diff
@@ -1,80 +1,124 @@
...
-            return hasJsonDiff(a, b);
-        };
 
+    if (snapshot && currentStrategyConfig) {
         // compare each property in the snapshot. The property order
         // might differ, so using JSON.stringify to compare them
         // doesn't work.
         const changes: ChangesThatWouldBeOverwritten = Object.entries(
-            currentStrategyConfig,
+            omit(currentStrategyConfig, 'strategyName'),
```
thomasheartman, line 62: `strategyName` both isn't interested *and* doesn't exist on the change payload.

sjaanus (approved):

PR #5994, opened 2024-01-23
chore: implement created_by_user_id in features

daveleek:
## About the changes

implement updating of features table created_by_user_id in feature-toggle-store.ts

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-service.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-service.ts
@@ -2402,6 +2402,10 @@ class FeatureToggleService {
             );
         }
     }
+
+    async setFeatureCreatedByUserIdFromEvents(): Promise<void> {
+        await this.featureToggleStore.setCreatedByUserId(500);
+    }
 }
 
 export default FeatureToggleService;
--- a/src/lib/features/feature-toggle/feature-toggle-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-store.ts
@@ -18,7 +18,11 @@ import { DEFAULT_ENV } from '../../../lib/util';
 
 import { FeatureToggleListBuilder } from './query-builders/feature-toggle-list-builder';
 import { FeatureConfigurationClient } from './types/feature-toggle-strategies-store-type';
-import { IFeatureTypeCount, IFlagResolver } from '../../../lib/types';
+import {
+    ADMIN_TOKEN_USER,
+    IFeatureTypeCount,
+    IFlagResolver,
+} from '../../../lib/types';
 import { FeatureToggleRowConverter } from './converters/feature-toggle-row-converter';
 import { IFeatureProjectUserParams } from './feature-toggle-controller';
 
@@ -74,6 +78,9 @@ const commonSelectColumns = [
 
 const TABLE = 'features';
 const FEATURE_ENVIRONMENTS_TABLE = 'feature_environments';
+const EVENTS_TABLE = 'events';
+const USERS_TABLE = 'users';
+const API_TOKEN_TABLE = 'api_tokens';
 
 export default class FeatureToggleStore implements IFeatureToggleStore {
     private db: Db;
@@ -717,6 +724,47 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             .where({ name: featureName });
 
         return result?.potentially_stale ?? false;
+    }
+
+    async setCreatedByUserIdWithSQL(): Promise<void> {
+        const query = this.db.raw(
+            `UPDATE features
+             SET created_by_user_id = u.id
+             FROM events AS e
+             JOIN users AS u ON u.email = e.created_by OR u.email = e.created_by
+             WHERE features.created_by_user_id IS null AND
+             e.type = 'feature-created' AND
+             e.data ->>'name' = features.name`,
+        );
+
+        await query;
+    }
+
+    async setCreatedByUserId(batchSize: number): Promise<void> {
+        const toUpdate = await this.db(`${TABLE} as f`)
+            .joinRaw(`JOIN ${EVENTS_TABLE} AS ev ON ev.data ->>'name' = f.name`)
+            .joinRaw(
+                `LEFT OUTER JOIN ${USERS_TABLE} AS u on ev.created_by = u.username OR ev.created_by = u.email`,
+            )
+            .joinRaw(
+                `LEFT OUTER JOIN ${API_TOKEN_TABLE} AS t on ev.created_by = t.username`,
+            )
+            .whereRaw(
+                `f.created_by_user_id IS null AND ev.type = 'feature-created'`,
+            )
+            .orderBy('f.created_at', 'asc')
+            .limit(batchSize)
+            .select(['f.*', 'ev.created_by', 'u.id', 't.username']);
+
+        toUpdate
+            .filter((row) => row.id || row.username)
+            .forEach(async (row) => {
+                const id = row.id || ADMIN_TOKEN_USER.id;
+
+                await this.db(TABLE)
+                    .update({ created_by_user_id: id })
+                    .where({ name: row.name });
+            });
     }
 }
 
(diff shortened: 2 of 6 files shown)


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -2402,6 +2402,10 @@ class FeatureToggleService {
             );
         }
     }
+
+    async setFeatureCreatedByUserIdFromEvents(): Promise<void> {
+        await this.featureToggleStore.setCreatedByUserId(500);
```
daveleek, line 2407: What's a reasonable ratio of features per run?

src/lib/features/scheduler/schedule-services.ts
```diff
@@ -150,4 +150,12 @@ export const scheduleServices = async (
         minutesToMilliseconds(3),
         'updateAccountLastSeen',
     );
+
+    schedulerService.schedule(
+        featureToggleService.setFeatureCreatedByUserIdFromEvents.bind(
+            featureToggleService,
+        ),
+        secondsToMilliseconds(5),
```
daveleek, line 158: Every 5 seconds? More often/seldom? If we have a few to get through we should probably make sure they happen at a decent pace

src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -718,6 +725,47 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
+             e.type = 'feature-created' AND
+             e.data ->>'name' = features.name`,
+        );
+
+        await query;
+    }
+
+    async setCreatedByUserId(batchSize: number): Promise<void> {
+        const toUpdate = await this.db(`${TABLE} as f`)
+            .joinRaw(`JOIN ${EVENTS_TABLE} AS ev ON ev.data ->>'name' = f.name`)
```
daveleek, line 745: Just realized there's a feature_name column on events. So I'll change this to use that

src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -718,6 +725,36 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
+        const toUpdate = await this.db(`${TABLE} as f`)
+            .joinRaw(`JOIN ${EVENTS_TABLE} AS ev ON ev.feature_name = f.name`)
+            .joinRaw(
+                `LEFT OUTER JOIN ${USERS_TABLE} AS u on ev.created_by = u.username OR ev.created_by = u.email`,
+            )
+            .joinRaw(
+                `LEFT OUTER JOIN ${API_TOKEN_TABLE} AS t on ev.created_by = t.username`,
+            )
+            .whereRaw(
+                `f.created_by_user_id IS null AND ev.type = 'feature-created'`,
```
gastonfournier, line 742: Maybe `f.created_by_user_id = -1337 AND f.created_by != 'system-user'`. I don't think this is as simple, but the idea to fix the other situation

daveleek: Thank you @gastonfournier! Yeah that is related to migration in `events` and there's a separate task for that! Thanks for the info, good help!👍

src/lib/__snapshots__/create-config.test.ts.snap
```diff
@@ -79,6 +79,10 @@ exports[`should create default config 1`] = `
       "caseInsensitiveInOperators": false,
       "celebrateUnleash": false,
       "changeRequestConflictHandling": false,
+      "createdByUserIdDataMigration": {
+        "enabled": false,
+        "name": "createdByUserIdDataMigration",
+      },
```
gastonfournier, line 85: This looks weird, all `experiments` are `[key: string]: boolean`, but this one breaks that. It doesn't mean it's wrong, but I think we have to be careful here

src/lib/features/feature-toggle/feature-toggle-service.ts
gastonfournier: I'd say 50... we have an average of ~50 toggles per customer. But this batch size is not about the number of feature toggles, it's more about the rows returned from the query

src/lib/features/feature-toggle/types/feature-toggle-store-type.ts
```diff
@@ -103,4 +103,6 @@ export interface IFeatureToggleStore extends Store<FeatureToggle, string> {
     getFeatureTypeCounts(
         params: IFeatureProjectUserParams,
     ): Promise<IFeatureTypeCount[]>;
+
+    setCreatedByUserId(batchSize: number): Promise<void>;
```
gastonfournier, line 107: Should we split this into a different service? Or else we should have a task to clean it up. I'm worried about the noise we're adding into feature-toggle-store which is now +700 lines

src/lib/features/scheduler/schedule-services.ts
gastonfournier: I think we can do this every half an hour or so. It shouldn't take that long.

gastonfournier (commented):
I think this looks mostly fine, I will take a deeper look at the query

src/lib/__snapshots__/create-config.test.ts.snap
daveleek: Thank you for pointing that out, it does look a bit strange indeed! I'll take a look at it

src/lib/types/experimental.ts
```diff
@@ -222,6 +223,13 @@ const flags: IFlags = {
                 '',
         },
     },
+    createdByUserIdDataMigration: {
+        name: 'createdByUserIdDataMigration',
```
gastonfournier, line 227: I think this is not needed
```suggestion
```

src/lib/features/feature-toggle/feature-toggle-service.ts
gastonfournier: 100 every 15 minutes after looking at real data might be a good thing. This should most likely finish the job in 10 days even for large tables

gastonfournier (approved):
We discussed this in person, there are a few things to change but I'm granting the approval

src/lib/types/experimental.ts
daveleek: Needed to change the whole thing to
`createdByUserIdDataMigration: parseEnvVarBoolean(
     process.env.CREATED_BY_USERID_DATA_MIGRATION,
     false,
),
`

src/lib/features/feature-toggle/types/feature-toggle-store-type.ts
daveleek: As discussed I moved the table reference constants into the method so it's self contained

src/lib/features/scheduler/schedule-services.ts
daveleek: As discussed I set it to every 15mins

src/lib/features/feature-toggle/feature-toggle-service.ts
daveleek: Has been set to 100 every 15mins as discussed!

src/lib/types/experimental.ts
```diff
@@ -222,6 +223,10 @@ const flags: IFlags = {
                 '',
         },
     },
+    createdByUserIdDataMigration: parseEnvVarBoolean(
+        process.env.CREATED_BY_USERID_DATA_MIGRATION,
+        false,
```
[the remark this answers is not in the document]
daveleek: Good thinking, thanks @gastonfournier!

src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -718,6 +722,40 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
```
daveleek, line 743: .whereRaw('u.id is not null or t.username is not null')
potential improvement so we don't get the same 100 every time

src/lib/types/experimental.ts
```diff
@@ -222,6 +223,10 @@ const flags: IFlags = {
                 '',
         },
     },
+    createdByUserIdDataMigration: parseEnvVarBoolean(
+        process.env.CREATED_BY_USERID_DATA_MIGRATION,
+        true,
```
daveleek, line 228: should be opt-in, change to false

PR #5998, opened 2024-01-23
test: variant edit corner cases

kwasniew:
## About the changes

Checking if variants don't add meta info. 
Checking if stickiness in variants is propagated from the strategy settings.


FredrikOseberg (approved):
LGTM. Paired.

PR #5999, opened 2024-01-23
chore: replace systemUser -1 with new -1337 system user

daveleek:
## About the changes
This PR replaces the old systemUser -1 in user-service.ts with the new SYSTEM_USER -1337 and adds a migration to move events created_by = -1 to -1337

## Discussion points
Does it make sense to do both of these things? Or should we skip the migration? How would this behave in a large system with hundreds of thousands of events, should this be split up?


src/lib/services/user-service.ts
```diff
@@ -244,7 +244,12 @@ class UserService {
         return userCreated;
     }
 
-    private getCreatedBy(updatedBy: IUser = systemUser) {
+    private getCreatedBy(
+        updatedBy: IUser = new User({
+            id: SYSTEM_USER.id,
+            username: SYSTEM_USER.username,
+        }),
```
nunogois, line 251: I'm curious, is there any reason why we don't simply do `updatedBy: IUser = SYSTEM_USER` like before?

nunogois (approved):
LGTM 👍

src/lib/services/user-service.ts
thomasheartman: Huh, I wonder why it doesn't work. Considering the previous version also only has id and username. Maybe if we wrapped the system user in a `new User`? 🤷🏼

daveleek: Good point, should've done that instead! 👌

thomasheartman: If that works, can we also type it as an IUser? 🤔

daveleek: It's an Omit<IUser, 'email'> so that doesn't work. The constructor takes a UserData object where `email?: string` which I guess is why it's ok to pass as parameter

thomasheartman: What if you type the system user as an IUser and use the user constructor to assign it? Would that work? Otherwise, it's probably fine 💁🏼

daveleek: I think I did something similar in the change I implemented?
I didn't change the underlying type of SYSTEM_USER though, it's still Omit<IUser, 'email'>
`private getCreatedBy(updatedBy: IUser = new User(SYSTEM_USER)): string {`

thomasheartman: Sure! I didn't look at the changes. Was only curious if we could make the system user an IUser again (which some functions require for their args).

This seems to work, but it requires you to remove `permissions` and `isAPI`. Not sure what the knock-on effects of that are, though.
```ts
export const SYSTEM_USER: IUser = new User({
    id: -1337,
    imageUrl: '',
    name: 'Unleash System',
    username: 'unleash_system_user',
});
```

thomasheartman: Looking at the constructor, `isAPI` defaults to false, while there is no default value set for `permissions`. So should probably be fine 🤷🏼  But I'll leave it up to you. The current impl looks good too.

daveleek: Thanks @thomasheartman! I'm going to keep it as it is now then. Let this one sit for a while and see if we should improve anything regarding the system user

PR #6000, opened 2024-01-23
docs: Added the new more fine-grained project permissions to the docs

chriswk:
This adds documentation for the 8 new permissions added last week, in preparation for including them in the 5.9.0 release.
Related
* https://github.com/Unleash/unleash/pull/5932
* https://github.com/Unleash/unleash/pull/5974
* https://github.com/ivarconr/unleash-enterprise/pull/1000


website/docs/reference/rbac.md
```diff
@@ -3,37 +3,45 @@ id: rbac
 title: Role-based Access control
 ---
 
-This document forms the specifications for [Role-Based Access Control](https://en.wikipedia.org/wiki/Role-based_access_control) which was introduced as part of the **Unleash v4 release**.
+This document forms the specifications
+for [Role-Based Access Control](https://en.wikipedia.org/wiki/Role-based_access_control) which was introduced as part of
```
gastonfournier, line 7: I think it's usually nice to be explicit when we introduce an acronym (because it's used below)
```suggestion
for [Role-Based Access Control](https://en.wikipedia.org/wiki/Role-based_access_control) (RBAC) which was introduced as part of
```

gastonfournier (approved):
LGTM! Just a minor suggestion

website/docs/reference/rbac.md
chriswk: I'll add this, but it has nothing to do with my PR. I see the autoformatting here has touched more than it should have.

PR #6002, opened 2024-01-23
fix: add unleash to default email sender string

andreas-unleash:
Adds "Unleash" to the "noreply@getunleash.io" for default email sender


sighphyre (approved):
LGTM

PR #6003, opened 2024-01-23
task: Make /edge/metrics toggleable by toggle

chriswk:
Previously we used a killswitch and returned 404 if the feature was enabled. This flips that to a default disabled toggle, that has to be turned on to handle old Edge (pre 17.0.0) posting bulk metrics


gastonfournier (commented):
Missing snapshot update but looks god

PR #6004, opened 2024-01-23
chore: This helps output the migrations being applied

gastonfournier:
## About the changes
This outputs to console.log but should be indexable:
image


gardleopard (approved):
LGTM

gardleopard (changes requested):
lets think about it, might be to much information

gastonfournier:
> Might be to noisy maybe we should not??

Sure, this is the type of thing to consider. You only get the migrations that are applied, if your DB is up-to-date with the migration you get this instead:
image

So we should only see the log of the migrations applied by the pod running the migrations. The others should just print something like above

gardleopard (approved):
then its good for me

PR #6008, opened 2024-01-23
chore: add support for project and environment in PermissionGuard

nunogois:
I noticed some manual `hasAccess` usages in permission guards due to the fact that `PermissionGuard` does not accept `project` and `environment`. This PR adds this support to `PermissionGuard` so we can adapt these `hasAccess` checks to use it instead, adding consistency and cleaning things up.

This PR does not include these adaptations however, it only adds the optional properties to the component. We can address these at a later point.


gastonfournier (approved):
LGTM!

PR #6016, opened 2024-01-24
Feature Flags with Python Tutorial

nnennandukwe:
## About the changes

A new tutorial to get developers to implement feature flags into a Python Flask app.
We have steps that guide them through the process using a simple full-stack survey app.

Based off of a reviewed [Google Doc here](https://docs.google.com/document/d/1YlG-TyQcMEyUrMjiWlYrd-J5DXog5ErbvSKAyuzKAnU/edit#heading=h.ryvmv14764i3)

The tutorial can go live in the docs nav in a new Python section, where the Python examples will follow.

### Important files

`website/docs/feature-flag-tutorials/python/implementing-feature-flags.md`


website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+
+[Flask Surveys Container App](https://github.com/pamelafox/flask-surveys-container-app) is an example Python application using [Flask](https://flask.palletsprojects.com/en/3.0.x/) and [SQLAlchemy](https://www.sqlalchemy.org/) to create and store surveys. Flask is a Python framework that provides out-of-the-box configurations to get the shell of a full-stack web application up and running, which includes [Jinja](https://jinja.palletsprojects.com/en/3.1.x/) for building web page HTML templates served using Python routing. This sample app is containerized, meaning you won’t have to rely on your system’s configurations of different libraries and tools used to allow the app to run successfully. It’s wrapped in a Docker container that will install and run the app in isolation.
```
sighphyre, line 10: We don't own this repo. It very much looks to be a sample repository to me. What happens in the future if this owner archives or deletes this project? Should we consider forking this if we want to use it in our examples?

sighphyre: Nit: 

> It’s wrapped in a Docker container that will install and run the app in isolation

It's quite unusual to talk about installing something with a container. I would just say run

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+- Docker
+- (Optional) a code editor like Visual Studio Code
+
+
+![An architectural diagram of our Python app using Unleash feature flags](/img/python-flask-unleash-architecture.png)
+
+
+This architecture diagram breaks down how the Python app works with Unleash to control feature flags.
+
+Your Python app is an **Evaluation Service**, which is meant to evaluate feature flags and pass evaluated results down to the SDK in the client application. In this context, the results will be passed to the Python SDK.
```
sighphyre, line 43: This doesn't make sense to me at a glance. I think the arch diagram is good but the description doesn't match and seems incorrect to me. 

The SDK is the Evaluation Service, it receives raw data about toggles from Unleash, evaluates them and returns that result to the host application.

(really like the name Evaluation Service btw! I use Flag Engine in my own internal model but we don't have a consistent name for this idea yet)

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+![An architectural diagram of our Python app using Unleash feature flags](/img/python-flask-unleash-architecture.png)
+
+
+This architecture diagram breaks down how the Python app works with Unleash to control feature flags.
+
+Your Python app is an **Evaluation Service**, which is meant to evaluate feature flags and pass evaluated results down to the SDK in the client application. In this context, the results will be passed to the Python SDK. 
+
+The Unleash Server is a **Feature Flag Control Service**, which is a service that manages your feature flags and is used to retrieve flag data from (and send data to, especially when not using a UI). The Unleash server has a UI for creating and managing projects and feature flags. There are also [API commands available in our documentation](https://docs.getunleash.io/reference/api/unleash) to perform the same actions straight from your CLI or server-side app.
+
+An Evaluation Service and a Feature Flag Control Service work hand in hand for you to leverage feature flags. For this tutorial, we will be using both to toggle extended functionality for surveys hosted in the Python app database.
```
sighphyre, line 47: I'm not actually sure here what Evaluation Service encompasses here. Is this just the SDK? Because then I would just say "Feature Flag SDK" here

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+
+
+## 1. Unleash best practice for backend apps
```
sighphyre, line 50: This section feels a bit awkward to me. I'm not sure it fits well here. Everything else seems to be pretty concrete advice targeted towards beginners using Unleash. This is largely about scaling feature flag systems from an expert perspective. These are definitely things that someone getting a crash course in Unleash doesn't need to know. If you want this stuff, personally add it as a further reading section or addendum

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+Since Python is a backend language, there are special security considerations to plan around when implementing feature flags.
+
+Most importantly, you must:
+
+- Limit feature flag payloads for scalability, security, and efficiency
+- Improve architectural resiliency with graceful degradation
+
+As your application scales, performance and resiliency become more critical and costly if not addressed. A feature flagging system should not be the reason your app slows down or fails. That’s why we recommend you account for this by reducing the size of your feature flag payloads. For example, instead of making one large call to retrieve flag statuses for all users as part of your configuration, group your users by specific attributes as part of your targeting rules that would be most relevant for your application.
+
+Additionally, you can cache your feature flag configuration to help reduce network round trips and dependency on external services. You can rely on the cache if your Feature Flag Control Service is not available, which will mitigate potential failure in your application.
```
sighphyre, line 61: Unleash SDKs already very much do this without any further configuration

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+
+Use Git to clone the Unleash repository and Docker to build and run it. Open a terminal window and run the following commands:
+
+```
+git clone git@github.com:Unleash/unleash.git
+cd unleash
+docker compose up -d
+```
+
+You will now have Unleash installed onto your machine and running in the background. You can access this instance in your web browser at http://localhost:4242
```
sighphyre, line 79: Install is odd in this context. I'd just say running in a container

(side note, this spins up postgres too, I don't know if you feel that's relevant for this)

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+
+From your project view on the platform, go to "Project Settings" and then "API Access".
+
+Select the ‘New API token’ button.
+
+
+![Image of the API token button in API Access view](/img/python-tutorial-create-api-token.png)
+
+
+Name the API token and select the “Server-side SDK” token type, since we’ll be doing our flag evaluation on the server using the Python SDK. You can read more about [Unleash API tokens in our documentation](https://docs.getunleash.io/reference/api-tokens-and-client-keys#client-tokens).
```
sighphyre, line 124: Maybe this link makes better sense in the **Note** section on line 114?

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+
+```py
+from UnleashClient import UnleashClient
+```
+
+
+In the same file, call the Unleash client for initialization when the app runs with this code snippet:
+
+
+```py
```
sighphyre, line 178: A few things about this snippet:

1) The url you've chosen here won't work on Linux machines - this is kinda a little hack to bridge two compose runs
2) The example has mixed tabs and spaces, which makes this horrid to copy paste, since a Python interpreter will panic

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+
+
+![Image of Surveys app loaded in browser](/img/python-tutorial-surveys-app-loaded.png)
+
+
+Create 1 or more new surveys so they’re populated in your database!
+
+## 5. Use a feature flag to roll out a delete method
+
+In a real-world use case for your feature flag, you can gradually roll out new features to a percentage of users by configuring the flag's strategy.
```
sighphyre, line 211: This paragraph and the one below it don't follow smoothly for me. How does gradual rollout in prod segue into controlling our delete case?

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+The modified return statement in this method will now look like this:
+
+```py
+return render_template("surveys_list.html", surveys=surveys, client=client)
+```
+
+In the same file, we will create a new route and a ‘delete’ method with this code snippet:
+
+```py
+@bp.route("/surveys/<int:survey_id>/delete", methods=["GET", "POST", "DELETE"])
```
sighphyre, line 241: This is a bit of a non standard way of doing an REST endpoint. I'd expect the operation to be expressed in the HTTP verb and not the URL. At minimum, we should definitely not have a GET for a delete operation. Your sample may need a bit of fiddling if you take this suggestion though

```suggestion
@bp.route("/surveys/<int:survey_id>", methods=["DELETE"])
```

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+
+In `src/backend/templates/surveys_list.html`, add the following code to your survey table underneath the “Survey Page” button:
+
+```html
+{% if client.is_enabled('delete_survey_flag') %}
+<td class="text-end"><a href="{{ url_for('surveys.delete_survey', survey_id=survey.id) }}" class="btn btn-sm btn-danger">Delete</a></td>
+{% endif %}
+```
+
+This code wraps a Delete button in a conditional statement that checks whether or not the feature flag is enabled. This button has a link that points to the `delete_survey` method we created, which will pull in the survey using an ID to search the database, find the matching survey, and delete it from the database session.
```
sighphyre, line 262:
```suggestion
This code wraps a delete button in a conditional statement that checks whether or not the feature flag is enabled. This button has a link that points to the `delete_survey` method we created, which will pull in the survey using an ID to search the database, find the matching survey, and delete it from the database session.
```

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+## 6. Verify the toggle experience
+
+
+Now that we’ve added in new functionality and connected it to our feature flag, we can verify that if you disable the flag in your development environment, your Python app will no longer serve an HTML page with the option to delete surveys you’ve created.
+
+In your local Unleash instance, turn off the feature flag by disabling it in the development environment.
+
+![Image of feature flag with a disabled environment](/img/python-tutorial-disabled-flag.png)
+
+Next, return to your Survey app and refresh the browser. With the flag disabled, the Delete button will no longer be visible.
```
sighphyre, line 282: Default update on the Python SDK is around 30 seconds, useful to remind folks they may have to wait a bit!

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,346 @@
...
+![Screenshot of app in browser without delete buttons for surveys](/img/python-tutorial-surveys-without-delete.png)
+
+
+## 7. Improve a feature flag implementation with error handling
+
+If you turn the feature flag off, you won’t be able to use the delete button to remove a survey from your list. However, if a user wanted to bypass the UI to delete a survey, they could still use the URL from the delete method route to target a survey and delete it.
+
+You could do this by entering in a URL like this to the browser:
+
+[http://localhost:50505/surveys/2/delete](http://localhost:50505/surveys/2/delete)
```
sighphyre, line 293: Do you want to provide a linkable example that will open a browser and make db changes to a running application? Sounds a bit dangerous to me!

Also note - if we drop the GET verb I suggested earlier, this won't work!

sighphyre (approved):
Very cool! I've left quite a lot of comments but I absolutely don't want to block this so I have put an approve on it.

I do think we should at least address the GET/POST/DELETE issue though

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
nnennandukwe: I tried to base this off of what I could find in the docs. I don't think we have these terms defined anywhere but we are currently using them so this is what I came up with. I'd like to simplify it to two short sentences what evaluation service and feature flag control service is in this particular context.

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
nnennandukwe: I could make it a note instead and define the terms evaluation service and FF control service, but yes we don't need too much detail in this tutorial if we want the reader up and running quickly.

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
nnennandukwe: I originally thought to only use "DELETE" as well, but it caused issues in this sample repo.

As for Docker, I'm not aware of a less hacky way to have the app connect to unleash server.

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
sighphyre: That'd be awesome! I'm happy to help walk you through the arch!

If we have docs that are suggesting this is the correct data flow, please point them to me, I think they're either misleading or incorrect!

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
sighphyre: Ball is in your court! Just an opinion of mine, you're the Word Smith!

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
sighphyre: I thought it might be a thing to work around a problem! This makes absolute sense but I do think we should aim to fix before we publish.

The "right" way to link the composes is probably to setup an external docker network and bind them both to that external network. Needs a code change in those compose files though.

Happy to help if you need!

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
nunogois: Agreed. 
 - **Unleash** is the **Evaluation Service** / **Flag Engine**
 - **Unleash Python SDK** is what will help you connect to Unleash from your Python app
 - ...and your Python app is whatever you want it to be 🙂

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,331 @@
...
+
+This code wraps a delete button in a conditional statement that checks whether or not the feature flag is enabled. This button has a link that points to the `delete_survey` method we created, which will pull in the survey using an ID to search the database, find the matching survey, and delete it from the database session.
+
+Your surveys page will now look something like this:
+
+
+![Screenshot of app in browser with delete buttons in survey table](/img/python-tutorial-surveys-with-delete.png)
+
+
+Test the new functionality by deleting one of your surveys. The page should refresh without the survey you deleted.
```
nunogois, line 261: If I understood correctly, this only hides the button in the UI. Might be worth adding the `if client.is_enabled('delete_survey_flag')` check to the route as well, so that users are not able to delete things by directly hitting the endpoint, and I think it would prove to be more interesting for a tutorial like this, where the focus is on a backend SDK.

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,331 @@
...
+
+In your local Unleash instance, turn off the feature flag by disabling it in the development environment.
+
+![Image of feature flag with a disabled environment](/img/python-tutorial-disabled-flag.png)
+
+Next, return to your Survey app and refresh the browser. With the flag disabled, the delete button will no longer be visible.
+
+> **Note:** An update on the Python SDK may take around 30 seconds.
+
+![Screenshot of app in browser without delete buttons for surveys](/img/python-tutorial-surveys-without-delete.png)
```
nunogois, line 277: If I understood correctly, this only hides the button in the UI. Might be worth adding the if client.is_enabled('delete_survey_flag') check to the route as well, so that users are not able to delete things by directly hitting the endpoint, and I think it would prove to be more interesting for a tutorial like this, where the focus is on a backend SDK.

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,331 @@
...
+@bp.route("/surveys/<int:survey_id>/delete", methods=["GET", "POST", "DELETE"])
+def delete_survey(survey_id):
+   if client.is_enabled('delete_survey_flag'):
+       survey = db.get_or_404(Survey, survey_id)
+       db.session.delete(survey)
+       db.session.commit()
+
+       return redirect(url_for("surveys.surveys_list_page"))
+   else:
+       abort(404, description="Resource not found")
```
nunogois, line 320: I would suggest the opposite if possible. Guard the route with a simple check: "If not enabled, abort". Otherwise, proceed with the normal behavior.

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,331 @@
...
+```
+
+Now, if you turn off the flag in your Unleash instance and attempt to delete a survey directly with a URL, the 404 error will return.
+
+![Screenshot of 404 error rendering in browser](/img/python-tutorial-404.png)
+
+Learn more about [Flask Blueprint error handling](https://flask.palletsprojects.com/en/3.0.x/errorhandling/#blueprint-error-handlers).
```
nunogois, line 327: I'm assuming you see the 404 on your browser because you try to navigate to that endpoint? This would not be possible if the route was decorated with the correct methods: only `DELETE`. In that case, I would expect to see a Postman / Insomnia / cURL screenshot with the DELETE request instead.

website/docs/feature-flag-tutorials/python/implementing-feature-flags.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Conclusion
+
+In this tutorial, we installed Unleash locally, created a new feature flag, installed Unleash into a Python Flask app, and toggled new functionality that altered a database with a containerized project!
```
nunogois, line 331: Unsure about the usage of "install" here. Similar to https://github.com/Unleash/unleash/pull/6016/files#r1464460774

PR #5239, opened 2023-11-01
wip: license checker for self-hosted

ivarconr:
Show banner for enterprise self-hosted if they violate their license.

PR #6018, opened 2024-01-24
task: Add banner encouraging edge upgrade

chriswk:
Only triggers if there is any rows in client instances that have

    sdk_version: unleash-edge with version < 17.0.0

The function that checks this memoizes the check for 10 minutes to avoid scanning the client instances table too often.

### DIFF
--- a/frontend/src/component/banners/EdgeUpgradeBanner/EdgeUpgradeBanner.tsx
+++ b/frontend/src/component/banners/EdgeUpgradeBanner/EdgeUpgradeBanner.tsx
@@ -1,0 +1,22 @@
+import { useUiFlag } from '../../../hooks/useUiFlag';
+import { ConditionallyRender } from '../../common/ConditionallyRender/ConditionallyRender';
+import { Banner } from '../Banner/Banner';
+import { IBanner } from '../../../interfaces/banner';
+
+export const EdgeUpgradeBanner = () => {
+    const displayUpgradeEdgeBanner = useUiFlag('displayUpgradeEdgeBanner');
+    const upgradeEdgeBanner: IBanner = {
+        message: `We noticed that you're using an outdated Unleash Edge. To ensure you continue to receive metrics, we recommend upgrading to v17.0.0 or later.`,
+        link: 'https://github.com/Unleash/unleash-edge',
+        linkText: 'Get latest',
+        variant: 'warning',
+    };
+    return (
+        <>
+            <ConditionallyRender
+                condition={displayUpgradeEdgeBanner}
+                show={<Banner key={'upgradeEdge'} banner={upgradeEdgeBanner} />}
+            />
+        </>
+    );
+};
--- a/frontend/src/component/banners/internalBanners/InternalBanners.tsx
+++ b/frontend/src/component/banners/internalBanners/InternalBanners.tsx
@@ -1,19 +1,8 @@
 import { Banner } from 'component/banners/Banner/Banner';
 import { useBanners } from 'hooks/api/getters/useBanners/useBanners';
-import { useUiFlag } from '../../../hooks/useUiFlag';
-import { ConditionallyRender } from '../../common/ConditionallyRender/ConditionallyRender';
-import { IBanner } from '../../../interfaces/banner';
 
 export const InternalBanners = () => {
     const { banners } = useBanners();
-    const displayUpgradeEdgeBanner = useUiFlag('displayUpgradeEdgeBanner');
-
-    const upgradeEdgeBanner: IBanner = {
-        message: `We noticed that you're using an outdated Unleash Edge. To ensure you continue to receive metrics, we recommend upgrading to v17.0.0 or later.`,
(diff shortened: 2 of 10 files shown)


chriswk:
Don't approve yet, I'm fixing the possible SQL Injection first.

src/lib/features/metrics/instance/instance-service.ts
```diff
@@ -224,4 +225,20 @@ export default class ClientInstanceService {
...
+        const semver = new SemVer(sdkVersion);
+        const instancesOfSdk =
+            await this.clientInstanceStore.getBySdkName(sdkName);
+        return instancesOfSdk.some((instance) => {
+            if (instance.sdkVersion) {
+                const [_sdkName, sdkVersion] = instance.sdkVersion.split(':');
+                const instanceUsedSemver = new SemVer(sdkVersion);
+                return instanceUsedSemver < semver;
+            }
+        });
```
gastonfournier, line 242: I was just checking that SemVer can throw some exceptions: https://github.com/npm/node-semver/blob/main/classes/semver.js and we have a helper class that does the try catch: https://github.com/Unleash/unleash/blob/chore/update-features-created-by-user-id/src/lib/util/semver.ts maybe we should just try catch here

frontend/src/component/banners/internalBanners/InternalBanners.tsx
```diff
@@ -11,6 +22,10 @@ export const InternalBanners = () => {
                 .map((banner) => (
                     <Banner key={banner.id} banner={banner} />
                 ))}
+            <ConditionallyRender
+                condition={displayUpgradeEdgeBanner}
+                show={<Banner key={'upgradeEdge'} banner={upgradeEdgeBanner} />}
+            />
```
nunogois, line 28: Similarly to https://github.com/Unleash/unleash/pull/5239#discussion_r1413730809 - I would recommend we create a new `EdgeUpgradeBanner` (or similar) component and add it directly to `App.tsx`. I don't think it makes much sense to add it here, especially as it is not an [internal banner](https://docs.getunleash.io/reference/banners).

gastonfournier (approved):
Looks good just a small thing around handling SemVer I think it's a safety check that's not hurtful to add. But Granting the approval as I don't expect that to happen, but just being extra cautious could be a good thing here

frontend/src/interfaces/uiConfig.ts
```diff
@@ -79,6 +79,7 @@ export type UiFlags = {
     executiveDashboard?: boolean;
     changeRequestConflictHandling?: boolean;
     feedbackComments?: Variant;
+    displayUpgradeEdgeBanner?: boolean;
```
nunogois, line 82: Might be worth considering making this a `Variant` flag instead. That way we can control not only whether the banner is shown, but also its contents through the flag.

nunogois (changes requested):
LG, but like mentioned in https://github.com/Unleash/unleash/pull/6018/files#r1464683982, I don't think it belongs inside `InternalBanners`.

src/lib/features/metrics/instance/instance-service.ts
chriswk: Thanks. Good point. I'll have a look

frontend/src/component/banners/internalBanners/InternalBanners.tsx
chriswk: Agreed. Will fix

frontend/src/interfaces/uiConfig.ts
chriswk: This is something we'd like to have in OSS deploys as well, and hopefully not for very long.

chriswk: So, I think having it as a normal flag is fine.

chriswk:
> LG, but like mentioned in https://github.com/Unleash/unleash/pull/6018/files#r1464683982, I don't think it belongs inside `InternalBanners`.

It's no longer in InternalBanners.

frontend/src/component/banners/EdgeUpgradeBanner/EdgeUpgradeBanner.tsx
```diff
@@ -0,0 +1,22 @@
...
+        message: `We noticed that you're using an outdated Unleash Edge. To ensure you continue to receive metrics, we recommend upgrading to v17.0.0 or later.`,
+        link: 'https://github.com/Unleash/unleash-edge',
+        linkText: 'Get latest',
+        variant: 'warning',
+    };
+    return (
+        <>
+            <ConditionallyRender
+                condition={displayUpgradeEdgeBanner}
+                show={<Banner key={'upgradeEdge'} banner={upgradeEdgeBanner} />}
```
nunogois, line 18: No need for key here, since it's only 1 element.

nunogois (approved):
LGTM! Thanks for addressing my comment 🙌

PR #6020, opened 2024-01-24
fix: flag resolver get variant

nunogois:
https://linear.app/unleash/issue/2-1880/fix-flag-resolver-getvariant-behavior

Fixes the flag resolver `getVariant` behavior when there's a variant object set in `experimental` - The flag resolver should call the external resolver `getVariant` when not overridden to be true, even if set as variant object in `experimental`.

Related: https://github.com/Unleash/unleash/pull/3808


FredrikOseberg (approved):
Yep, this looks good!

PR #6022, opened 2024-01-24
chore: add metrics for conflict creation detection

thomasheartman:
This PR adds a 'change-request-conflict-created' event whenever someone save a strategy update for a strategy that's used in either pending or scheduled change requests.

Data for pending change requests will only be sent if change requests are enabled. Data for scheduled change requests will be sent regardless.

Getting this data is somewhat involved, so I've extracted as much of the logic into a separate file as possible.

The event re-uses the existing `change_request` metric and sends the following data for each change request that we discover conflicts on:

```ts 
{
  state: ChangeRequestState,
  changeRequest: string, // <unleash identifier>#<change request id>
  action: 'edit-strategy',
  eventType: 'conflict-created'
}
```

There's only one action for this for now, but we could expand this event to things such as strategy deletion, feature archival, in the future. That said, I'd be happy to take it out.

### Has the strategy actually been updated?

This does not check whether a strategy has actually changed before emitting the event, only that you save your strategy changes.

This assumes that most people will simply close the modal by clicking/tapping outside it or using the escape key instead of pressing save.

However, it will likely lead to some false positives. If we think that is an issue, I would suggest adding a check that something in the strategy has actually changed in a follow-up PR.


frontend/src/utils/unique-change-request-id.ts
thomasheartman, line 1: The implementation here was wrong previously. Must've slipped by a review.

andreas-unleash (approved):
Nice! Makes a lot of sense

PR #6027, opened 2024-01-24
chore: events created by userid migration

daveleek:
## About the changes

Schedules a best-effort task setting the value of events.created_by_user_id based on what is found in the created_by column and if it's capable of resolving that to a userid/a system id. The process is executed in the events-store, it takes a chunk of events that haven't been processed yet, attempts to join users and api_tokens tables on created_by = username/email, loops through and tries to figure out an id to set. Then updates the record.

## Discussion points

1. There's potentially a lot of events to process so this can go on for quite some time. We don't need this specifically for anything, so that should be fine. Currently set at 100 records every 2mins, open to suggestions on better frequency/update counts.

2. unknown, migration, init-api-tokens get translated to system user

3. Events with `created_by = null` get filtered out in the query. Should we address these?

### DIFF
--- a/src/lib/features/events/event-store.ts
+++ b/src/lib/features/events/event-store.ts
@@ -435,6 +435,11 @@ class EventStore implements IEventStore {
 
     async setCreatedByUserId(batchSize: number): Promise<void> {
         const API_TOKEN_TABLE = 'api_tokens';
+
+        if (!this.flagResolver.isEnabled('createdByUserIdDataMigration')) {
+            return;
+        }
+
         const toUpdate = await this.db(`${TABLE} as e`)
             .joinRaw(
                 `LEFT OUTER JOIN users AS u ON e.created_by = u.username OR e.created_by = u.email`,
--- a/src/lib/features/events/event-store.ts
+++ b/src/lib/features/events/event-store.ts
@@ -477,6 +477,9 @@ class EventStore implements IEventStore {
                 return this.db(TABLE)
                     .update({ created_by_user_id: ADMIN_TOKEN_USER.id })
                     .where({ id: row.id });
+            } else {
+                this.logger.warn(`Could not find user for event ${row.id}`);
+                return Promise.resolve();
             }
         });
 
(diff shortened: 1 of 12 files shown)


src/lib/features/events/event-store.ts
```diff
@@ -428,6 +432,49 @@ class EventStore implements IEventStore {
...
+        toUpdate.forEach(async (row) => {
+            if (
+                row.created_by === 'unknown' ||
+                row.created_by === 'migration' ||
+                (row.created_by === 'init-api-tokens' &&
+                    row.type === 'api-token-created')
+            ) {
+                await this.db(TABLE)
+                    .update({ created_by_user_id: SYSTEM_USER_ID })
+                    .where({ id: row.id });
```
daveleek, line 466: update all where created_by in (unknown/migration/init-api-tokens) instead

src/lib/features/events/event-store.ts
```diff
@@ -428,6 +432,49 @@ class EventStore implements IEventStore {
...
+            .joinRaw(
+                `LEFT OUTER JOIN users AS u ON e.created_by = u.username OR e.created_by = u.email`,
+            )
+            .joinRaw(
+                `LEFT OUTER JOIN ${API_TOKEN_TABLE} AS t on e.created_by = t.username`,
+            )
+            .whereRaw(
+                `e.created_by_user_id IS null AND e.created_by IS NOT null`,
+            )
+            .orderBy('e.created_at', 'asc')
```
daveleek, line 453: order by id desc

src/lib/features/events/event-store.ts
```diff
@@ -428,6 +432,56 @@ class EventStore implements IEventStore {
 
         events.forEach((e) => this.eventEmitter.emit(e.type, e));
     }
+
+    async setCreatedByUserId(batchSize: number): Promise<void> {
```
gastonfournier, line 436: Maybe it'd be better to extract this method into a different function, same as the other code from the other task to assign user ids. But for this iteration it seems fine

gastonfournier (approved):
LGTM!

PR #6031, opened 2024-01-25
chore: change users image_url column to text

andreas-unleash:
Change the image_url column to text

Closes # https://github.com/orgs/Unleash/projects/8/views/1?pane=issue&itemId=51021534


chriswk (approved):
LGTM

Small nit: `increase` not `increate` nor `increae`, but since that's only in filename and PR title, no need for changing it. :)

PR #6032, opened 2024-01-25
chore: add a migration that removes incorrectly set sysuserid

daveleek:
## About the changes

We had a bug that would set created_by_user_id = -1337 even for some events that had an actual user, and the created_by would be the users email. To fix the data this PR introduces a migration that simply sets created_by_user_id to null when created_by is not the unleash system user but the created_by_user_id is. The other scheduled jobs should pick this up and assign the correct user id

## Discussion points

The EXPLAIN of the previous version of this query with the created_by filter looking like this: `created_by NOT IN ('system_user', 'unleash_system_user', 'system@getunleash.io')` indicated that it would scan indexes and be fairly efficient


chriswk (approved):
LGTM.
I do think we have an index both on created_by and on created_by_user_id. Postgres query planner might still find for some instances that it's faster to do a full table scan. ;)

*Edit*
Double checked, we have on created_by_user_id, not on created_by.

PR #6035, opened 2024-01-25
feat: add CR id to plausible events

sjaanus:
Added conflict count to CR metrics and CR id.

Something to think about:
There was idea that we can aggregate this data based on CR id, but CR id is just a number from 0 to x. So it will not be unique across instances.


thomasheartman:
Regarding the uniqueness of CR ids across instances: this can be handled by using the instance's baseUriPath or its instanceId. I've made that change in dc0a1980cb982d6abe7b045676d538cc6f062b47

thomasheartman (approved):

PR #6041, opened 2024-01-26
fix: Use knex whereLike instead of whereRaw to fix escaping

chriswk:
This escape with `??` double escaped the LIKE query causing no results. This updates to using whereLike, which does the correct escaping for string query.


gastonfournier (approved):
LG!

PR #6043, opened 2024-01-26
feat: stale flags count per project

kwasniew:
## About the changes

/projects returns not just each project feature count but also stale and potentially stale count. This resource is used in the projects screen but the underlying store will be used for trends calculation that also requires stale info.

This is the UI using the existing resource. 
Screenshot 2024-01-26 at 10 45 06

Explain plan comparison before and after:

Screenshot 2024-01-26 at 11 50 03

### DIFF
--- a/src/lib/db/project-store.ts
+++ b/src/lib/db/project-store.ts
@@ -141,7 +141,10 @@ class ProjectStore implements IProjectStore {
 
         let selectColumns = [
             this.db.raw(
-                'projects.id, projects.name, projects.description, projects.health, projects.updated_at, projects.created_at, count(features.name) FILTER (WHERE features.archived_at is null) AS number_of_features',
+                'projects.id, projects.name, projects.description, projects.health, projects.updated_at, projects.created_at, ' +
+                    'count(features.name) FILTER (WHERE features.archived_at is null) AS number_of_features' +
+                    'count(features.name) FILTER (WHERE features.stale IS TRUE) AS stale_feature_count, ' +
+                    'count(features.name) FILTER (WHERE features.potentially_stale IS TRUE) AS potentially_stale_feature_count',
             ),
             'project_settings.default_stickiness',
             'project_settings.project_mode',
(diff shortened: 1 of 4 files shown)


src/lib/db/project-store.ts
```diff
@@ -141,7 +141,10 @@ class ProjectStore implements IProjectStore {
 
         let selectColumns = [
             this.db.raw(
-                'projects.id, projects.name, projects.description, projects.health, projects.updated_at, projects.created_at, count(features.name) FILTER (WHERE features.archived_at is null) AS number_of_features',
+                'projects.id, projects.name, projects.description, projects.health, projects.updated_at, projects.created_at, ' +
+                    'count(features.name) FILTER (WHERE features.archived_at is null) AS number_of_features, ' +
+                    'count(features.name) FILTER (WHERE features.stale IS TRUE) AS stale_feature_count, ' +
+                    'count(features.name) FILTER (WHERE features.potentially_stale IS TRUE) AS potentially_stale_feature_count',
```
gastonfournier, line 147: How efficient is this? I'd expect these aggregated metrics to be memoized with a different query, maybe even kept in memory (similar to prometheus metric stats)

kwasniew: we already have the first count in our codebase for a long time. Before and after with 500 toggles is negligible (20 ms server response for the entire query).

gastonfournier: But this brings this to 3x of the previous time, right?

Is this executed on every request? In which paths?

kwasniew: I think DB query engine is much smarter than that so it shouldn't be 3 times but same order of magnitude.
This query is currently executed when you open all projects view. I will also use it for trends calculation once a week.

gastonfournier: > I think DB query engine is much smarter than that so it shouldn't be 3 times but same order of magnitude.

I don't know much about this, but I'm more concerned about creating a pattern that keeps replicating and suddenly we end up with 10 of these queries... also the impact on large installations might be noticeable. I'm not sure, but I don't take aggregation queries lightly as I've been hit by them multiple times.

Can we instead read this value from trends? Maybe calculate trends every other hour to keep them updated?

I'm not against this solution, just trying to think of alternatives

Tymek (approved):

gastonfournier (approved):
Nice it does only 1 loop! Still, let's keep an eye on the metrics when releasing that, at least for the exercise

PR #6045, opened 2024-01-26
docs: update nav bar

thomasheartman:
Update the Unleash docs navbar according to the new spec:

Changing "estimated time burden" to "estimated time" (or "estimated reading time"), is just a matter of changing the text in the files. There's nothing special making it say burden other than plain text. That can be done in a follow-up PR.

The new nav bar includes the Unleash academy and certification dropdowns as well as the docs and quickstart items. On smaller screen sizes (between 997 and 1150 px), we hide the word "Unleash" from "Unleash academy" and "Unleash certification".

The narrowest nav bar:
image

When Unleash pops back in: 
image


markunl:
Updated logo asset using CSS:

image

markunl (approved):

PR #6046, opened 2024-01-26
chore: introduce a timeout to allow indexes to update

daveleek:
## About the changes

Adds a timout of 100ms to feature created by userid migration tests

### DIFF
--- a/src/test/e2e/features-created-by-user-id-data-migration.e2e.test.ts
+++ b/src/test/e2e/features-created-by-user-id-data-migration.e2e.test.ts
@@ -102,6 +102,10 @@ test('should set created_by_user_id on features', async () => {
 
     await stores.featureToggleStore.setCreatedByUserId(200);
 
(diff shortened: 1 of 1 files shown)


gastonfournier (changes requested):
I don't think it's flaky test. It's this code: https://github.com/Unleash/unleash/blob/4a025a4b4b1fff7fc231cea2b19a8a3e063160e3/src/lib/features/feature-toggle/feature-toggle-store.ts#L749-L757

If fires a lot of async tasks and moves on. You should map all promises to an array and do await Promise.all before finishing. Here's an example: https://github.com/Unleash/unleash/blob/4a025a4b4b1fff7fc231cea2b19a8a3e063160e3/src/lib/db/api-token-store.ts#L163-L173

daveleek:
Thank you @gastonfournier! Addressed in https://github.com/Unleash/unleash/pull/6048 so I'll close this!

PR #6048, opened 2024-01-26
chore: default the data migration flag for createdbyuserid to false

daveleek:
## About the changes

Sets data migration of features and events created_by_user_id to disabled by default


gastonfournier (approved):
LGTM!

PR #6049, opened 2024-01-26
fix: return 400 when enabling env of archived toggle

andreas-unleash:
Creates a new ArchivedFeatureError.
Throw this error when trying to toggle a feature environment for an archived feature.  

Closes https://github.com/orgs/Unleash/projects/8/views/1?pane=issue&itemId=51242922


sighphyre (approved):
LGTM

PR #6057, opened 2024-01-29
chore: actions modal form

nunogois:
https://linear.app/unleash/issue/2-1882/ui-add-actions-modal-and-form

Adds actions modal and form, allowing users to create and edit actions.

The main thing that is missing is adding the remaining fields, which will be included in a later PR.


gastonfournier (approved):
LGTM!

PR #6058, opened 2024-01-29
chore: mark change requests as "willOverwriteStrategyConfig" if they have any conflicts

thomasheartman:
This PR has allows us to mark change requests as "willOverwriteStrategyConfig" if they have any conflicts.

The current implementation uses react's contexts, which is probably the right way to go. However, I haven't been able to make it work correctly 😢

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChangeOverwriteWarning.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChangeOverwriteWarning.tsx
@@ -71,7 +71,8 @@ const OverwriteTable = styled('table')(({ theme }) => ({
 export const ChangesToOverwrite: React.FC<{
     currentStrategy?: IFeatureStrategy;
     change: IChangeRequestUpdateStrategy;
-}> = ({ change, currentStrategy }) => {
+    markAsConflictedChange?: () => void;
+}> = ({ change, currentStrategy, markAsConflictedChange }) => {
     const checkForChanges = useUiFlag('changeRequestConflictHandling');
     const changesThatWouldBeOverwritten = checkForChanges
         ? getChangesThatWouldBeOverwritten(currentStrategy, change)
@@ -80,6 +81,8 @@ export const ChangesToOverwrite: React.FC<{
     if (!changesThatWouldBeOverwritten) {
         return null;
     }
+
+    markAsConflictedChange?.();
 
     return (
         <ChangesToOverwriteWarning>
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChangeOverwriteWarning.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChangeOverwriteWarning.tsx
@@ -74,7 +74,7 @@ export const ChangesToOverwrite: React.FC<{
     change: IChangeRequestUpdateStrategy;
 }> = ({ change, currentStrategy }) => {
     const checkForChanges = useUiFlag('changeRequestConflictHandling');
-    const { registerWillOverwriteStrategyChanges: registerConflicts } =
+    const { registerWillOverwriteStrategyChanges } =
         useChangeRequestPlausibleContext();
     const changesThatWouldBeOverwritten = checkForChanges
         ? getChangesThatWouldBeOverwritten(currentStrategy, change)
@@ -84,9 +84,7 @@ export const ChangesToOverwrite: React.FC<{
         return null;
     }
 
-    console.log('found conflicts. registering.');
-
-    registerConflicts();
+    registerWillOverwriteStrategyChanges();
 
     return (
         <ChangesToOverwriteWarning>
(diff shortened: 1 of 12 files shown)


frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
```diff
@@ -280,25 +279,6 @@ export const ChangeRequestOverview: FC = () => {
...
-                    (changeRequest.schedule.reason ||
-                        changeRequest.schedule.failureReason) ??
-                    undefined
-                );
-            case 'suspended':
-                return changeRequest.schedule.reason;
-            default:
-                return undefined;
-        }
-    })();
```
thomasheartman, line 300: This was apparently unused, so I removed it.

frontend/src/component/changeRequest/ChangeRequestSidebar/ChangeRequestSidebar.tsx
thomasheartman, line 1: If I understand react's useContext correctly, we _should_ be able to use the cr plausible context here and just fetch the `willOverwriteStrategyChanges` prop directly, instead of passing it into the changeState function. That would allow us to not have to add extra parameters to the function only for use in plausible. However, I was unable to make it work.

But if we _can_ make it work, then I would suggest adding all the plausible data (including the previous state of the CR) into the plausible context instead of passing it as parameters. That way, the changeState function can stick with just accepting the values it needs to actually change the state and we can instead pass anything else via the context.

Optionally, if we can't make that work, we can instead make it accept a single parameter: `PlausibleContext` instead of many smaller ones.

frontend/src/component/changeRequest/ChangeRequestSidebar/EnvironmentChangeRequest/EnvironmentChangeRequest.tsx
thomasheartman, line 1: Almost all changes in this file are only indentation changes due to wrapping the whole component in the plausible context provider.

frontend/src/component/changeRequest/ChangeRequestSidebar/EnvironmentChangeRequest/EnvironmentChangeRequest.tsx
```diff
@@ -56,146 +61,165 @@ const ChangeRequestContent = styled(Box)(({ theme }) => ({
 export const EnvironmentChangeRequest: FC<{
     environmentChangeRequest: ChangeRequestType;
     onClose: () => void;
-    onReview: (id: number, comment?: string) => void;
+    onReview: (
+        id: number,
+        willOverwriteStrategyConfig: boolean,
```
thomasheartman, line 66: I also think we should be able to remove this parameter and instead access the data via plausible as mentioned in a previous comment. But again, that assumes that we can make the useContext call work within the useChangeRequestApi code.

frontend/src/component/changeRequest/changeRequest.types.ts
```diff
@@ -100,7 +100,7 @@ export interface IChangeRequestChangeBase {
     };
 }
 
-export type PlausibleChangeRequestState =
+export type PlausibleChangeRequestPreviousState =
```
thomasheartman, line 103: As a further refactoring, if we move the previous state into the plausible context, I think we should also move this type definition to that file.

frontend/src/hooks/api/actions/useChangeRequestApi/useChangeRequestApi.ts
```diff
@@ -75,9 +75,9 @@ export const useChangeRequestApi = () => {
         trackEvent('change_request', {
             props: {
                 eventType: payload.state,
-                previousState,
+                ...(previousState ? { previousState } : {}),
```
thomasheartman, line 78: There is no previous state for new drafts, so let's leave that field out in that case.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChangeOverwriteWarning.tsx
```diff
@@ -73,10 +75,18 @@ export const ChangesToOverwrite: React.FC<{
...
+        useChangeRequestPlausibleContext();
     const changesThatWouldBeOverwritten = checkForChanges
         ? getChangesThatWouldBeOverwritten(currentStrategy, change)
         : null;
 
+    useEffect(() => {
+        if (changesThatWouldBeOverwritten) {
+            registerWillOverwriteStrategyChanges();
+        }
+    }, [changesThatWouldBeOverwritten]);
```
thomasheartman, line 88: This needs to be wrapped in `useEffect` to avoid the issue where react complains you cannot updated a component while rendering another component. Refer to [this blog post](https://bobbyhadz.com/blog/react-cannot-update-component-while-rendering-different-component).

There may also be other ways of solving this, but this was the first one I found.

frontend/src/component/changeRequest/ChangeRequestSidebar/ChangeRequestSidebar.tsx
```diff
@@ -76,15 +77,18 @@ export const ChangeRequestSidebar: VFC<IChangeRequestSidebarProps> = ({
...
-    const onReview = async (draftId: number, comment?: string) => {
+    const onReview = async (
+        changeState: (project: string) => Promise<void>,
+    ) => {
         try {
-            await changeState(project, draftId, 'Draft', {
-                state: 'In review',
-                comment,
-            });
+            await changeState(project);
```
thomasheartman, line 91: This is a trick: we need to call the `useChangeRequestApi` hook from a component that's within the provider context. When we call it here, it won't get the right state. As such, we instead move the hook call into the child component, keeping the rest of the logic here.

PR #6061, opened 2024-01-29
chore: filter on token username and user id in SQL instead

daveleek:
## About the changes

Change the sorting of features to migrate created_by_user_id for, and filter out unresolvable feature/users

Query tested manually in enterprise

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-store.ts
@@ -740,21 +740,21 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
                 `LEFT OUTER JOIN ${API_TOKEN_TABLE} AS t on ev.created_by = t.username`,
             )
             .whereRaw(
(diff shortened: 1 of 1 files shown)


src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -740,21 +740,21 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
                 `LEFT OUTER JOIN ${API_TOKEN_TABLE} AS t on ev.created_by = t.username`,
             )
             .whereRaw(
-                `f.created_by_user_id IS null AND ev.type = 'feature-created'`,
+                `f.created_by_user_id IS null AND
+                ev.type = 'feature-created' AND
+                (u.id IS NOT null OR t.username IS NOT null)`,
```
gastonfournier, line 745: We still need to figure out what to do when we can't resolve the user, we've identified many cases in enterprise and demo instances. 

Any suggestion @chriswk @ivarconr?

gastonfournier (approved):
LGTM!

PR #6071, opened 2024-01-30
feat: scheduler init jitter

kwasniew:
## About the changes

Adding scheduler init jitter to avoid too many DB operations running at once. Subsequent runs are not affected.

Design decisions:
* default jitter calculated so that we don't need to change code using scheduler
* default jitter is 2s-30s
* if any scheduler runs more often than max jitter time (30 seconds) then no jitter is applied
* you can explicitly opt-out of jitter by passing 0ms jitter


gastonfournier (approved):
LGTM! Let's try it out!

PR #6082, opened 2024-01-31
feat: scheduler overrun protection

kwasniew:
## About the changes

Adding overrun protection. When there's the same job id already pending we wait for the next interval. 

I decided to keep pending jobs ids instead of replacing setInterval with setTimeout since it's less invasive.


gastonfournier (approved):
I like the approach!

PR #6085, opened 2024-01-31
feat: A/B test search feedback variants

sjaanus:
Search was not getting any feedback. We introduced 3 different variants to compare conversion rate.

image
image
image


kwasniew (approved):

FredrikOseberg (approved):
Nice!

PR #6086, opened 2024-01-31
refactor: take chatgpt's suggestions for diff calc algorithm

thomasheartman:
We had to make some updates to let the compiler know about the types and fix an issue with nested objects not being compared as objects (instead as strings), but this saves us a few lines and is hopefully more readable.

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
@@ -33,11 +33,15 @@ const hasChanged = (
     snapshotValue: unknown,
     currentValue: unknown,
     changeValue: unknown,
-) =>
-    typeof snapshotValue === 'object'
-        ? !isEqual(snapshotValue, currentValue) ||
-          !isEqual(currentValue, changeValue)
-        : hasJsonDiff({ snapshotValue, currentValue, changeValue });
+) => {
+    if (typeof snapshotValue === 'object') {
+        return (
+            !isEqual(snapshotValue, currentValue) &&
+            !isEqual(currentValue, changeValue)
+        );
+    }
+    return hasJsonDiff({ snapshotValue, currentValue, changeValue });
+};
 
 type DataToOverwrite<Prop extends keyof ChangeRequestEditStrategy> = {
     property: Prop;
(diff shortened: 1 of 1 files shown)


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
```diff
@@ -53,80 +58,48 @@ export const getChangesThatWouldBeOverwritten = (
...
-            omit(currentStrategyConfig, 'strategyName'),
-        )
-            .map(([key, currentValue]: [string, unknown]) => {
-                const snapshotValue = snapshot[key as keyof IFeatureStrategy];
-                const changeValue =
-                    change.payload[key as keyof ChangeRequestEditStrategy];
+    const changes: ChangesThatWouldBeOverwritten = Object.entries(
+        omit(currentStrategyConfig, 'strategyName'),
+    )
+        .map(([key, currentValue]: [string, unknown]) => {
```
kwasniew, line 66: can we map the function so that it's possible to understand the high level steps of the algorithm without reading the contents of the mapping function?

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
```diff
@@ -53,80 +58,48 @@ export const getChangesThatWouldBeOverwritten = (
...
+                    fallback,
+                })
+                    ? changeInfo
+                    : undefined;
+            } else if (hasChanged(snapshotValue, currentValue, changeValue)) {
+                return changeInfo;
+            }
+        })
+        .filter(
+            (
```
kwasniew, line 94: again I'd name this function to express the intent

kwasniew (approved):

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
thomasheartman: For sure. I was considering that. Will look into it.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
thomasheartman: Yeah, that's fair. I'll fix.

thomasheartman: 30d8a2cc3123780d7529b5fb1420dbda76e0b8a0

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/strategy-change-diff-calculation.ts
thomasheartman: The typing there gets tricky, but we can use unknown types to get around it. Addressed in 42b70ecebf5a01207007ccf71f9891592d0c4d61

PR #6089, opened 2024-01-31
chore: scheduled created-by migrations metrics

daveleek:
## About the changes

the created_by_user_id data migration from resolving events.created_by (for both events and features) now emits events on how many rows were updated.

Adds listeners for these events that records these metrics with prometheus

image

### DIFF
--- a/src/lib/features/events/event-store.ts
+++ b/src/lib/features/events/event-store.ts
@@ -433,16 +433,16 @@ class EventStore implements IEventStore {
         events.forEach((e) => this.eventEmitter.emit(e.type, e));
     }
 
-    async setCreatedByUserId(batchSize: number): Promise<void> {
+    async setCreatedByUserId(batchSize: number): Promise<number> {
         const API_TOKEN_TABLE = 'api_tokens';
 
         if (!this.flagResolver.isEnabled('createdByUserIdDataMigration')) {
-            return;
+            return -1;
         }
 
         const toUpdate = await this.db(`${TABLE} as e`)
             .joinRaw(
-                `LEFT OUTER JOIN users AS u ON e.created_by = u.username OR e.created_by = u.email`,
+                'LEFT OUTER JOIN users AS u ON e.created_by = u.username OR e.created_by = u.email',
             )
             .joinRaw(
                 `LEFT OUTER JOIN ${API_TOKEN_TABLE} AS t on e.created_by = t.username`,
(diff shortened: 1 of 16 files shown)


src/lib/features/events/event-store.ts
```diff
@@ -433,16 +433,16 @@ class EventStore implements IEventStore {
         events.forEach((e) => this.eventEmitter.emit(e.type, e));
     }
 
-    async setCreatedByUserId(batchSize: number): Promise<void> {
+    async setCreatedByUserId(batchSize: number): Promise<number> {
         const API_TOKEN_TABLE = 'api_tokens';
 
         if (!this.flagResolver.isEnabled('createdByUserIdDataMigration')) {
-            return;
+            return -1;
```
gastonfournier, line 440: Find it weird that we return -1 here while we return Promise.resolve() down below... maybe this function should always return the number of rows updated and zero is meaningful, but -1 is not so much.

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -2444,7 +2450,12 @@ class FeatureToggleService {
     }
 
     async setFeatureCreatedByUserIdFromEvents(): Promise<void> {
-        await this.featureToggleStore.setCreatedByUserId(100);
+        const updated = await this.featureToggleStore.setCreatedByUserId(100);
+        if (updated > -1) {
+            this.eventBus.emit(FEATURES_CREATED_BY_PROCESSED, {
+                updated,
+            });
+        }
```
gastonfournier, line 2458: I think it makes sense if we say that zero were updated to do this:
```suggestion
        if (updated > 0) {
            this.eventBus.emit(FEATURES_CREATED_BY_PROCESSED, {
                updated,
            });
        }
```

src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -723,13 +723,13 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
 
-    async setCreatedByUserId(batchSize: number): Promise<void> {
+    async setCreatedByUserId(batchSize: number): Promise<number> {
         const EVENTS_TABLE = 'events';
         const USERS_TABLE = 'users';
         const API_TOKEN_TABLE = 'api_tokens';
 
         if (!this.flagResolver.isEnabled('createdByUserIdDataMigration')) {
-            return;
+            return -1;
```
gastonfournier, line 732: Same suggestion
```suggestion
            return 0;
```

daveleek: I'll return undefined here instead as we discussed

src/lib/features/feature-toggle/feature-toggle-service.ts
daveleek: I'll change this to check for undefined

gastonfournier (approved):
LGTM!

PR #6090, opened 2024-01-31
fix: freezing navigation

kwasniew:
## About the changes

Renovate upgraded react router again breaking navigation (however in a subtle way). To reproduce the error go to projects/default/metrics and manually update the end of url to /logs at the end of the URL. It should freeze and the process should start increasing memory usage. With the downgrade of router I can't reproduce it anymore.

We downgraded router in the past: https://github.com/Unleash/unleash/pull/5512

To prevent the same mistake I'm excluding router from the renovate bot.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -106,7 +106,7 @@
     "react-joyride": "^2.5.3",
     "react-linkify": "^1.0.0-alpha",
     "react-markdown": "^8.0.4",
-    "react-router-dom": "6.21.2",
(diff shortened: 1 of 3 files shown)


FredrikOseberg (approved):
Nice. Are there any issues about this on the RR repo?

kwasniew:
@FredrikOseberg thia was related to previous bug we had with nested navigation routing: https://github.com/remix-run/react-router/issues/11052

PR #6100, opened 2024-02-01
feat: time to production chart

kwasniew:
## About the changes

Adding new chart
Screenshot 2024-02-01 at 12 08 42

Also:
* removing duplication by extracting project data hook
* fixed initial undefined data error

### DIFF
--- a/frontend/src/component/executiveDashboard/useProjectChartData.ts
+++ b/frontend/src/component/executiveDashboard/useProjectChartData.ts
@@ -1,0 +1,55 @@
+import { useMemo } from 'react';
+import {
+    ExecutiveSummarySchema,
+    ExecutiveSummarySchemaProjectFlagTrendsItem,
+} from '../../openapi';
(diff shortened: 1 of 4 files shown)


frontend/src/component/executiveDashboard/useProjectChartData.ts
```diff
@@ -0,0 +1,55 @@
...
+                    borderColor: color,
+                    backgroundColor: color,
+                    fill: true,
+                };
+            },
+        );
+
+        const objectKeys = Object.keys(groupedFlagTrends);
+
+        const firstElementTrends = groupedFlagTrends[objectKeys[0]] || [];
```
kwasniew, line 45: fallback to [] to fix map on undefined error

sjaanus:
Is there a way to add units? Currently 0.1 to 1 does not tell alot.

kwasniew:
@sjaanus we'll think about it once we get real data

PR #6101, opened 2024-02-01
refactor: PATs

nunogois:
https://linear.app/unleash/issue/SR-379/refactor-pats

This PR refactors PATs.

 - Adds a new `createPatSchema`, which better aligns with https://docs.getunleash.io/contributing/ADRs/overarching/separation-request-response-schemas
 - Drops the model type and class in favor of using the schema types directly, which is more consistent with the rest of the codebase and easier to maintain
 - Misc scouting, improvement and fixes

This breaks Enterprise temporarily, but it's faster to move forward this way.


andreas-unleash (approved):
LGTM!

PR #6103, opened 2024-02-01
docs: Add troubleshooting note about missing Edge metrics

chriswk:
So we have a place in the docs to refer users struggling with getting metrics from Edge after upgrade to Unleash v5.9.0, this PR adds a small troubleshooting note about two fixes. 
1. Upgrade Edge
2. Use environmental flag to re-enable old endpoint


daveleek (approved):
LG!

PR #6109, opened 2024-02-02
chore: track metrics for how many CRs are moved into next state with conflicts

thomasheartman:
Use React's context to track how many CRs are moved into their next state with conflicts present.

This PR wraps environment change requests and change request overviews in a change request plausible context that contains a `willOverwriteStrategyChanges` property. This property is updated by the diff calculation if there are any conflicts and then read by the `changeState` function in the `useChangeRequestApi` hook. 

As long as at least one of the strategies in the CR contain conflicts, it will be marked as overwriting changes.


sjaanus (approved):
While it looks good, I am not entirely sure that building an entirely new context just to track a super specific metric is reasonable, or at least it does not align with anything else we have previously in the codebase. This approach would only make sense if we specifically agree that this metric is a game-changer.

Just my 2 cents.

PR #6112, opened 2024-02-02
fix: add new steps to demo guide

FredrikOseberg:
This PR adds the necessary steps for the demo guidance to work properly with the new form


nunogois (approved):
LGTM! I assume we're enabling `newStrategyConfiguration` right away in `demo` so we don't break the flow 🙂

PR #6114, opened 2024-02-02
Feat/inactive user

chriswk:
Adds the necessary foundation for listing inactive users and supporting bulk deletion of the same users.

### DIFF
--- a/frontend/src/component/admin/users/InactiveUsersList/DeleteUser/DeleteUser.tsx
+++ b/frontend/src/component/admin/users/InactiveUsersList/DeleteUser/DeleteUser.tsx
@@ -5,7 +5,6 @@ import { Alert, styled } from '@mui/material';
 import useLoading from 'hooks/useLoading';
 import { Typography } from '@mui/material';
 import { useThemeStyles } from 'themes/themeStyles';
-import { IUser } from 'interfaces/user';
 import { UserAvatar } from 'component/common/UserAvatar/UserAvatar';
 import { IInactiveUser } from '../../../../../hooks/api/getters/useInactiveUsers/useInactiveUsers';
 
--- a/frontend/src/component/admin/users/InactiveUsersList/DeleteUser/DeleteUser.tsx
+++ b/frontend/src/component/admin/users/InactiveUsersList/DeleteUser/DeleteUser.tsx
@@ -70,12 +70,7 @@ const DeleteUser = ({
                     style={{ marginTop: '1rem' }}
                 >
                     Are you sure you want to delete{' '}
-                    {user
-                        ? `${user.name || 'user'} (${
-                              user.email || user.username
-                          })`
-                        : ''}
-                    ?
+                    {`${user.name || 'user'} (${user.email || user.username})`}
                 </Typography>
             </div>
         </Dialogue>
--- a/frontend/src/component/admin/users/InactiveUsersList/DeleteUser/DeleteUser.tsx
+++ b/frontend/src/component/admin/users/InactiveUsersList/DeleteUser/DeleteUser.tsx
@@ -1,0 +1,86 @@
+import { Dialogue } from 'component/common/Dialogue/Dialogue';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
+import { REMOVE_USER_ERROR } from 'hooks/api/actions/useAdminUsersApi/useAdminUsersApi';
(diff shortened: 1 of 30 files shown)


chriswk:
image
Link added to /admin/users

image
New page with inactive users

src/lib/users/inactive/inactive-users-controller.ts
```diff
@@ -0,0 +1,102 @@
...
+            method: 'post',
+            path: '/delete',
+            handler: this.deleteInactiveUsers,
+            permission: ADMIN,
+            middleware: [
+                openApiService.validPath({
+                    operationId: 'deleteInactiveUsers',
+                    summary: 'Deletes inactive users',
+                    description:
+                        'Deletes all inactive users. An inactive user is a user that has not logged in in the last 180 days',
```
nunogois, line 65: Maybe `180` should be a constant somewhere, in case we decide to later change it.

src/lib/users/inactive/inactive-users-service.ts
```diff
@@ -0,0 +1,56 @@
...
+            userService: UserService;
+        },
+    ) {
+        this.logger = getLogger('services/client-feature-toggle-service.ts');
+        this.inactiveUsersStore = inactiveUsersStore;
+        this.userService = services.userService;
+    }
+
+    async getInactiveUsers(): Promise<InactiveUserSchema[]> {
+        const users = await this.inactiveUsersStore.getInactiveUsers(180);
```
nunogois, line 29: Same as https://github.com/Unleash/unleash/pull/6114/files#r1476472808

frontend/src/component/admin/users/InactiveUsersList/InactiveUsersList.tsx
```diff
@@ -0,0 +1,265 @@
...
+                    title={`Inactive users (${rows.length})`}
+                    actions={
+                        <>
+                            <Button
+                                variant='contained'
+                                color='primary'
+                                onClick={openDelInactiveDialog}
+                            >
+                                Delete all inactive users
+                            </Button>
```
nunogois, line 220: I think we should disable this if there are no inactive users.

nunogois (approved):
From a quick glance this looks good to me 👍

src/lib/users/inactive/inactive-users-service.ts
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    async deleteInactiveUsers(
+        calledByUser: IUser,
+        userIds: number[],
+    ): Promise<void> {
+        this.logger.info('Deleting inactive users');
+        for (const userid of userIds) {
+            await this.userService.deleteUser(userid, calledByUser);
```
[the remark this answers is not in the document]
chriswk: We want this to send the USER_DELETED events, so for now, I think it's better to call it one by one.

src/lib/users/inactive/inactive-users-controller.ts
chriswk: Added as a constant to the create-inactive-users-service.ts file for now. We can move it around later. Will update the schema as well.

frontend/src/component/admin/users/InactiveUsersList/InactiveUsersList.tsx
chriswk: Sure.

frontend/src/component/admin/users/InactiveUsersList/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    data-loading
+                    variant='body1'
+                    style={{ marginTop: '1rem' }}
+                >
+                    Are you sure you want to delete{' '}
+                    {user
+                        ? `${user.name || 'user'} (${
+                              user.email || user.username
+                          })`
+                        : ''}
```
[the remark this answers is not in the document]
chriswk: There shouldn't be. Thanks for the catch. I'll update

src/lib/users/inactive/inactive-users-service.ts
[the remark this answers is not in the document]
nunogois: I think that's probably fine, but for future reference, there are also bulk event methods:
https://github.com/Unleash/unleash/blob/1da59abb2dc57cfa5dead45333106741d698f075/src/lib/features/events/event-service.ts#L129-L140

PR #6118, opened 2024-02-02
Fix/check maintenance flag once per minute

ivarconr:
## About the changes
Every schedule job will now check if maintenance is enabled. This ends up querying the settings table in the db at least once per second per running unleash instance. This small fix caches this query for 60 seconds to reduce the load somewhat. 

We should reconsider this solution for the long term, but this will be a great improvement on the short term. 

**Logs after this fix running locally.** 
We can observe that we resolve settings from the DB once per minute. 
image

Also we should consider giving a warning in section where you enable maintenance mode that it can take up to a minute to propagate.


gastonfournier (approved):
Looks good! I think it's a fair trade-off. Maintenance mode is not a widely used feature and 1 minute should not cause any major issue.

PR #6119, opened 2024-02-02
chore: split metrics by request type

gastonfournier:
## About the changes
Queries on client-feature-toggle store have many purposes depending on the requestType, making the query more complex or not depending on the use case. Also, each use case has different frequencies (i.e. playground is expected to be used rarely).

The name for the store metrics was wrong, copy&pasted from:
https://github.com/Unleash/unleash/blob/7b04db054793a03e12bd92a53c24d5feb40c561d/src/lib/features/feature-toggle/feature-toggle-store.ts#L107

Which was also present in feature-tag metrics:
https://github.com/Unleash/unleash/blob/7b04db054793a03e12bd92a53c24d5feb40c561d/src/lib/db/feature-tag-store.ts#L37

With this, we'll have more granularity to understand the execution time and frequency of each


ivarconr (approved):
yes please!

PR #6120, opened 2024-02-03
fix: maintenance mode should assume disable if db call fails.

ivarconr:
Usually maintenance mode is disabled. If the call throws, which we see a lot of when a unleash instance is in terminating state, we should return a default value.

By having it throw inside of the memoizee function, the response is not cached, and it will trigger new calls until it return a cachable result.

### DIFF
--- a/src/lib/features/maintenance/maintenance-service.ts
+++ b/src/lib/features/maintenance/maintenance-service.ts
@@ -33,19 +33,22 @@ export default class MaintenanceService implements IMaintenanceStatus {
     }
 
     async isMaintenanceMode(): Promise<boolean> {
-        return (
-            this.config.flagResolver.isEnabled('maintenanceMode') ||
-            (await this.resolveMaintenance())
-        );
+        try {
+            return (
+                this.config.flagResolver.isEnabled('maintenanceMode') ||
+                (await this.resolveMaintenance())
+            );
+        } catch (e) {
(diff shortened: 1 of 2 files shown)


src/lib/features/maintenance/maintenance-service.ts
```diff
@@ -33,19 +33,22 @@ export default class MaintenanceService implements IMaintenanceStatus {
...
 
     async isMaintenanceMode(): Promise<boolean> {
-        return (
-            this.config.flagResolver.isEnabled('maintenanceMode') ||
-            (await this.resolveMaintenance())
-        );
+        try {
+            return (
+                this.config.flagResolver.isEnabled('maintenanceMode') ||
+                (await this.resolveMaintenance())
```
ivarconr, line 39: by allowing the momizee function to throw, we unsure it will not be cached.

src/lib/services/version-service.ts
```diff
@@ -193,7 +193,7 @@ export default class VersionService {
             )) ?? { id: undefined };
             this.instanceId = id;
         } catch (err) {
-            this.logger.warn('Could not find instanceInfo');
+            this.logger.warn('Could not find instanceInfo', err);
```
gastonfournier, line 196: :1st_place_medal:

gastonfournier (approved):
LG!

PR #6132, opened 2024-02-05
feat: make edge explicit in UI

kwasniew:
## About the changes

Users should not guess that proxy and edge tokens are the same thing:

Screenshot 2024-02-05 at 13 18 40

Screenshot 2024-02-05 at 13 18 46


sighphyre (approved):
Yes please!

nunogois (approved):
Makes sense to me. However, you can also use frontend tokens to connect _to_ Edge, so I hope this is not confusing.

PR #6133, opened 2024-02-05
fix: inactive users query was too wide

chriswk:
Knex wasn't formatting the query like I expected. This changes the query to use more AND, less ORs


sighphyre (approved):
LGTM

PR #6135, opened 2024-02-05
chore: memoizee active tokens

gastonfournier:
## About the changes
getAllActive from api-tokens store is the second most frequent query
image

To prevent starving our db connections, we can cache this data that rarely changes and clear the cache when we see changes. Because we will only clear changes in the node receiving the change we're only caching the data for 1 minute.

This should give us some room to test if this solution will work

### DIFF
--- a/src/lib/services/api-token-service.ts
+++ b/src/lib/services/api-token-service.ts
@@ -74,14 +74,16 @@ export class ApiTokenService {
             apiTokenStore,
             environmentStore,
         }: Pick<IUnleashStores, 'apiTokenStore' | 'environmentStore'>,
-        config: Pick<IUnleashConfig, 'getLogger' | 'authentication'>,
+        config: Pick<
+            IUnleashConfig,
+            'getLogger' | 'authentication' | 'flagResolver'
+        >,
         eventService: EventService,
-        flagResolver: IFlagResolver,
     ) {
         this.store = apiTokenStore;
         this.eventService = eventService;
         this.environmentStore = environmentStore;
-        this.flagResolver = flagResolver;
+        this.flagResolver = config.flagResolver;
         this.logger = config.getLogger('/services/api-token-service.ts');
         if (!this.flagResolver.isEnabled('useMemoizedActiveTokens')) {
             // This is probably not needed because the scheduler will run it
--- a/src/lib/services/api-token-service.ts
+++ b/src/lib/services/api-token-service.ts
@@ -24,6 +24,7 @@ import {
     ApiTokenCreatedEvent,
     ApiTokenDeletedEvent,
     ApiTokenUpdatedEvent,
+    IFlagResolver,
     IUser,
     SYSTEM_USER,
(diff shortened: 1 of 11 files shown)


src/lib/services/api-token-service.ts
```diff
@@ -76,7 +78,6 @@ export class ApiTokenService {
         this.eventService = eventService;
         this.environmentStore = environmentStore;
         this.logger = config.getLogger('/services/api-token-service.ts');
-        this.fetchActiveTokens();
```
gastonfournier, line 79: No need to call it here, probably updateLastSeen below is also pointless because it probably does nothing:
https://github.com/Unleash/unleash/blob/3bea6bbd204098ec04cecd8000dfa060a9e5314f/src/lib/services/api-token-service.ts#L101-L102

src/lib/services/api-token-service.ts
```diff
@@ -111,7 +116,12 @@ export class ApiTokenService {
     }
 
     public async getAllActiveTokens(): Promise<IApiToken[]> {
-        return this.store.getAllActive();
+        if (!this.initialized) {
+            // unlikely this will happen but nice to have a fail safe
+            this.logger.info('Fetching active tokens before initialized');
+            await this.fetchActiveTokens();
+        }
+        return this.activeTokens;
```
gastonfournier, line 124: This method now returns from memory, but if not yet initialized will attempt to load the tokens. This is just a safeguard

src/lib/services/edge-service.ts
```diff
@@ -1,25 +1,25 @@
...
         { getLogger }: Pick<IUnleashConfig, 'getLogger'>,
     ) {
         this.logger = getLogger('lib/services/edge-service.ts');
-        this.apiTokenStore = apiTokenStore;
+        this.apiTokenService = apiTokenService;
     }
 
     async getValidTokens(tokens: string[]): Promise<ValidatedEdgeTokensSchema> {
-        const activeTokens = await this.apiTokenStore.getAllActive();
+        const activeTokens = await this.apiTokenService.getAllActiveTokens();
```
gastonfournier, line 22: Using the service prevents going to the DB every time we getValidTokens

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -207,28 +207,32 @@ test('Token-admin should be allowed to create token', async () => {
...
+        await services.apiTokenService.createApiTokenWithProjects(
+            {
+                tokenName: 'default-admin',
+                type: ApiTokenType.ADMIN,
+                projects: ['*'],
+                environment: '*',
+            },
+            'test',
+            1,
+        );
```
gastonfournier, line 222: This uses the service logic that also pushes the newly created token in the memory cache

ivarconr:
Can we unleash this chang with a flag?

kwasniew (approved):

PR #6136, opened 2024-02-06
chore: type improvement for schedulers

gastonfournier:
## About the changes

Closes #

### DIFF
--- a/src/lib/features/scheduler/schedule-services.ts
+++ b/src/lib/features/scheduler/schedule-services.ts
@@ -48,6 +48,7 @@ export const scheduleServices = async (
         apiTokenService.fetchActiveTokens.bind(apiTokenService),
         minutesToMilliseconds(1),
         'fetchActiveTokens',
+        0, // no jitter -> run immediately
     );
 
     schedulerService.schedule(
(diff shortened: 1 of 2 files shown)


src/lib/features/scheduler/schedule-services.ts
```diff
@@ -131,16 +132,16 @@ export const scheduleServices = async (
     );
 
     schedulerService.schedule(
-        () => {
-            clientMetricsServiceV2.bulkAdd().catch(console.error);
+        async () => {
+            await clientMetricsServiceV2.bulkAdd().catch(console.error);
```
kwasniew, line 136: `() => clientMetricsServiceV2.bulkAdd().catch(console.error);` is exactly the same since the last statement returns Promise

src/lib/features/scheduler/schedule-services.ts
```diff
@@ -131,16 +132,16 @@ export const scheduleServices = async (
...
         },
         secondsToMilliseconds(5),
         'bulkAddMetrics',
     );
 
     schedulerService.schedule(
-        () => {
-            clientMetricsServiceV2.clearMetrics(48).catch(console.error);
+        async () => {
+            await clientMetricsServiceV2.clearMetrics(48).catch(console.error);
```
kwasniew, line 144: same here

src/lib/features/scheduler/schedule-services.ts
gastonfournier: but there's no return... I could add one, right?

kwasniew: In general it's more idiomatic to do  `() => Promise.resolve()` over `async () => { await Promise.resolve(); }`

kwasniew (approved):
Is there any particular problem you're trying to solve here?

src/lib/features/scheduler/schedule-services.ts
kwasniew: () => Promise.resolve() returns the promise when you don't add {} around it

kwasniew: `() => Promise.resolve()` is more idiomatic than `() => { return Promise.resolve() }`

gastonfournier:
> Is there any particular problem you're trying to solve here?

I saw we were making an await on a function that didn't return a promise, so I thought maybe there's something off that's causing our scheduled jobs to pile up. So I wanted to change the type to find out that, but I don't think there's something...

PR #6139, opened 2024-02-06
feat: Make storing to local storage robust

andreas-unleash:
Handles non serializable objects like Map and Set and makes the serialisation/deserialisation process more robust generally.

Closes # [UNL-322](https://linear.app/unleash/issue/UNL-322/add-types-to-local-storage-and-force-simple-types)


FredrikOseberg (approved):
I am wary of this change since it touches everything going in and out of localstorage. In addition, while governing what goes in and out does help, I think that the core of the issue is that we expected the data type to be one thing and received another. I'm not sure we have covered that use-case here, though this is better than nothing.

PR #6141, opened 2024-02-06
fix: disable double start

kwasniew:
## About the changes

We found that proxy repository was started twice because unleash client has auto start enabled by default.


sjaanus (approved):

chriswk (approved):
Great catch

PR #6145, opened 2024-02-07
chore: update schedule emails to handle suspensions and segment updates

thomasheartman:
This PR updates the change request email sending method to handle the recent changes we have made. That means that the email now:
- says that change requests have been suspended instead of saying that application will fail.
- handles cases where segments or strategies have been updated causing potential conflicts.

I have updated the email templates and made some adjustments to the email sending method. To make the transition from one to the other easier, I have kept the original method as an interim solution until enterprise has switched over.

### DIFF
--- a/src/lib/services/email-service.ts
+++ b/src/lib/services/email-service.ts
@@ -145,7 +145,22 @@ export class EmailService {
 
     async sendScheduledChangeConflictEmail(
         recipient: string,
-        conflictScope: 'flag' | 'strategy',
+        conflictData:
+            | { cause: 'flag archived'; flagName: string }
+            | {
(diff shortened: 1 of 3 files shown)


src/lib/services/email-service.ts
```diff
@@ -196,8 +266,10 @@ export class EmailService {
                 {
                     conflict,
                     conflictScope,
-                    conflictResolution,
-                    conflictResolutionLink,
+                    canBeRescheduled,
+                    flagArchived,
+                    flagLink,
+                    conflictingChangeRequestLink,
```
thomasheartman, line 272: I think this was left out by mistake previously?

PR #6153, opened 2024-02-07
docs: unexpected results in abn test troubleshooting guide

nunogois:
https://linear.app/unleash/issue/SR-374/troubleshooting-guide-my-feature-flag-is-returning-unexpected-results

Adds a quick troubleshooting guide that covers some common pitfalls when running A/B tests.


website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
```diff
@@ -0,0 +1,33 @@
...
+
+3. Double check that your code is correctly handling the feature flag variant response. You can refer to your SDK documentation for more information on how to handle feature flag variants. For example, if you're using the [Unleash React SDK](/reference/sdks/react), you can follow the [check variants](/reference/sdks/react#check-variants) section of the documentation. Given the example variants above, this could look like the following:
+
+```tsx
+import { useVariant } from '@unleash/proxy-client-react';
+
+export const TestComponent = () => {
+  const variant = useVariant('ab-test-flag');
+
+  if (variant.enabled && variant.name === 'A') {
```
sighphyre, line 24: Ooo, it's late and I'm tired, remind me what the `variant.enabled` check catches? My tired brain wants to say that should always be true if the toggle is true and a variant returned that isn't the default

website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
```diff
@@ -0,0 +1,33 @@
...
+title: My feature flag is returning unexpected results in my A/B/n test
+---
+
+Occasionally, users might come across a situation where a feature flag is returning unexpected results when running an A/B/n test.
+
+The first thing to look into is your feature configuration.
+
+If you're unsure whether the feature flag is being properly returned, you can go through the steps in this troubleshooting guide before proceeding: [My feature flag is not returned in the Frontend API/Edge/Proxy](/using-unleash/troubleshooting/flag-not-returned.md).
+
+1. If you're using a [gradual rollout activation strategy](/reference/activation-strategies#gradual-rollout) be sure to set the rollout percentage to the percentage of your user base you would like to target for the test. If you would like to target 100% of your user base on your A/B/n test, confirm that your rollout percentage is set to 100%. This adjustment guarantees that all users participate in the A/B/n test, irrespective of the division ratio, be it an even 50-50 split, for a typical A/B test with 50% for Group A and 50% for Group B, or any other distribution that you configure in your variants.
```
sighphyre, line 11: Up to you if you want to include this but part of my debugging strategy is often to throw away gradual rollout and use the standard strategy. Can't be off when your logic is `true === true`;

Other things that might be useful to point out - know what your stickiness is. The default, you can use `userId` or `sessionId` but one of those _must_ be on your context to get consistent resolutions.

Last thing (phew), if you use custom stickiness and don't pass that property, your toggle will always be off (and you'll get the default variant)

website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
```diff
@@ -0,0 +1,33 @@
...
+2. Ensure that your variants are correctly configured. You can refer to [feature toggle variants](/reference/feature-toggle-variants). For example, if you would like to run a simple 50-50 A/B test, then your variants should look similar to this: 
+
+![An example of variants configured for an A/B test](/img/troubleshooting-flag-abn-test-unexpected-result-variants.png)
+
+3. Double check that your code is correctly handling the feature flag variant response. You can refer to your SDK documentation for more information on how to handle feature flag variants. For example, if you're using the [Unleash React SDK](/reference/sdks/react), you can follow the [check variants](/reference/sdks/react#check-variants) section of the documentation. Given the example variants above, this could look like the following:
+
+```tsx
+import { useVariant } from '@unleash/proxy-client-react';
```
sighphyre, line 19: It may be easier to use the NodeSDK here instead so we can include the context, null contexts are a common one when folks ask why their toggles are off

website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
```diff
@@ -0,0 +1,33 @@
...
+  if (variant.enabled && variant.name === 'A') {
+    return <AComponent />;
+  } else if (variant.enabled && variant.name === 'B') {
+    return <BComponent />;
+  }
+  return <DefaultComponent />;
+};
+```
+
+Feature activation strategies may have constraints, segments, and rules that can be combined in different ways that can lead to complex scenarios. If you're still not seeing the results you expect, try using the [Playground](/reference/playground.mdx) to verify that the feature is properly configured and responding as expected.
```
sighphyre, line 33: Rules? Are there more than constraints and the strategy eval code?

sighphyre (approved):
I really like this. Left some ideas but 100% optional discussion pints

website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
nunogois: Straight copy/paste from the React SDK docs. I think you're right and the `variant.name` check is enough, since that will only be either `A` or `B` if the variant is enabled anyways. Simplified in https://github.com/Unleash/unleash/pull/6153/commits/147a3b751cffc6a6fb91a1f6acbe8aade60ed50a

website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
nunogois: Overrides maybe? This was copied from the last point here: https://docs.getunleash.io/using-unleash/troubleshooting/flag-not-returned

nunogois: How about we simplify this to:

>Feature activation strategies may be combined in different ways that can lead to complex scenarios. If you're still not seeing the results you expect, try using the [Playground](/reference/playground.mdx) to verify that the feature is properly configured and responding as expected.

sighphyre: Ooo, forgot about overrides! Yep, that'd do it

I do like the simplified version though. It's enough to say "STUFF" can go wrong but here's a good path to fixing it

website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
nunogois: Can you show me an example of what you have in mind? This is a basic example scenario, so I'm not sure if we would pass context in `getVariant` anyways, right?

website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
sighphyre: Awesome! Love the change!

website/docs/using-unleash/troubleshooting/flag-abn-test-unexpected-result.md
nunogois: Addressed in https://github.com/Unleash/unleash/pull/6153/commits/72dfa139bd9d934eb2bb9539595e8832a616f793, which also simplifies the point in the other guide.

ivarconr:
really nice work @nunogois 🔥

PR #6166, opened 2024-02-08
feat: add partial index on client_applications annouced FALSE

sjaanus:
We have customers with tens or hundreds of thousands of applications, and we have a scheduler running that sets application fields to `announced` as true. However, every time it runs, it queries the entire table, which is slow and causes database connection acquisition issues. To make it faster, we added a partial index to the table.


gardleopard (approved):
Fantastic!

PR #6169, opened 2024-02-08
Feat/UI error observability

FredrikOseberg:
This PR adds an endpoint to Unleash that accepts an error message and option error stack and logs it as an error. This allows us to leverage errors in logs observability to catch UI errors consistently.

Considered a test, but this endpoint only accepts and logs input, so I'm not sure how useful it would be.


frontend/src/index.tsx
```diff
@@ -20,35 +20,59 @@ import { FeedbackProvider } from 'component/feedbackNew/FeedbackProvider';
...
+    };
+
+    return (
+        <UIProviderContainer>
+            <AccessProvider>
+                <BrowserRouter basename={basePath}>
+                    <QueryParamProvider adapter={ReactRouter6Adapter}>
+                        <ThemeProvider>
+                            <AnnouncerProvider>
+                                <ErrorBoundary
```
FredrikOseberg, line 52: This error boundary is higher up than the one in App.tsx and is identical

gardleopard (approved):
7. LGTM

PR #6170, opened 2024-02-08
feat: add index on events for faster revision id

sjaanus:
On all pods and instances, we run the same revision update query every second. It is relatively fast when the application has started. This is the single most ran query in unleash.
Benchmarks:

1. Running pod with existing revisionID:
  - old 5.5ms
  - new 0.028ms

2. New pod without existing revisionID
 - old 9.329ms
 - new 0.033ms

This query is getting optimized

https://github.com/Unleash/unleash/blob/7e66a79f9f6f2c6f2475bfd723ea938bbd4fb0f6/src/lib/features/events/event-store.ts#L161


src/migrations/20240208130439-events-revision-id-index.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+exports.up = function(db, cb) {
+    db.runSql(
+        `
+            CREATE INDEX IF NOT EXISTS idx_events_feature_type_id ON events (id)
+                WHERE feature_name IS NOT NULL
+                    AND type IN ('segment-updated', 'feature_import', 'features-imported');
```
chriswk, line 8: The query is feature_name IS NOT NULL OR type IN (...), should the index also use OR here?

gardleopard (approved):
7. LGTM

src/migrations/20240208130439-events-revision-id-index.js
sjaanus: Good catch, was testing different variants and typo appeared.

PR #6171, opened 2024-02-08
chore: add migration for traffic data collection

daveleek:
## About the changes

Adds migration for creating table `stat_traffic_usage`.
This table primary-keys on day, traffic_group, and status_code_series.

Traffic group is the grouping for API endpoints for which traffic is counted.
status_code_series is 200/202 etc = 200, 304 etc = 300

### DIFF
--- a/src/migrations/20240208123212-create-stat-traffic-usage-table.js
+++ b/src/migrations/20240208123212-create-stat-traffic-usage-table.js
@@ -9,6 +9,8 @@ exports.up = (db, callback) => {
             count BIGINT NOT NULL DEFAULT 0,
             PRIMARY KEY(day, traffic_group, status_code_series)
             );
+            CREATE INDEX IF NOT EXISTS stat_traffic_usage_traffic_group_idx ON stat_traffic_usage (traffic_group);
+            CREATE INDEX IF NOT EXISTS  stat_traffic_usage_status_code_series_idx ON stat_traffic_usage (status_code_series);
         `,
         callback,
     );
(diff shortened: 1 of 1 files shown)


src/migrations/20240208123212-create-stat-traffic-usage-table.js
```diff
@@ -0,0 +1,24 @@
+
+exports.up = (db, callback) => {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS stat_traffic_usage(
+            day DATE NOT NULL,
+            traffic_group TEXT NOT NULL,
```
daveleek, line 7: ended up naming this traffic_group as the `group` name we initially came up with was a reserved keyword in Postgresql

src/migrations/20240208123212-create-stat-traffic-usage-table.js
```diff
@@ -0,0 +1,24 @@
...
+            status_code_series INT NOT NULL,
```
daveleek, line 8: appended a _series to make it clear that it's 200/300 etc, not 200/201/202/304 etc

chriswk:
@gardleopard , @sjaanus do you think this would do for storing 400 days of data for visitation stats?

chriswk:
I think creating subindexes on day, traffic_group and status_code_series could help. the PK index will be a thruple of (day, tg, scs), so if we want to do where status_code_series = 200 it will be a table scan.

daveleek:
Thanks @chriswk - I've added the suggested indexes

chriswk (approved):
LGTM

PR #6176, opened 2024-02-08
fix: prevent infinite rerenders in useIncomingWebhooks

nunogois:
React can sometimes be non-intuitive and behave erratically due to the way it detects changes in hook dependencies.

This prevents infinite re-renders from `useIncomingWebhooks` by using a static `DEFAULT_DATA` constant, so that its reference is always the same, so no changes are detected when there are none.

Unrelated scouting, but this PR also removes an unneeded dependency in the memoized columns in `ProjectActionsTable`.


gastonfournier (approved):
LG!

PR #6185, opened 2024-02-09
docs: Delete old preferred-styles-import-placement

gastonfournier:
Having old ADRs makes it harder for newcomers, that have to navigate through many ADRs and spot the ones that were superseded by others. We can save the mental burden by just moving the old ones to a deprecated folder


chriswk (approved):
Should we add a link to the one that supersedes it?

gastonfournier:
> Should we add a link to the one that supersedes it?

It already has one, but I just had to fix it: [747b984](https://github.com/Unleash/unleash/pull/6185/commits/747b9847177c37026a803b009f8e16fcb449e5a1)

PR #6190, opened 2024-02-09
feat: implement a store for stat_traffic_data

daveleek:
## About the changes

Implements a new store for collected traffic data usage that connects to the new table `stat_traffic_data` primary key'd on [day, trafficGroup, status_code_series].

Day being a date
Traffic group being which endpoint is being counted for, ie /api/admin, /api/frontend etc
Status code series grouping 2xx status responses and 304 into their respective 200 / 300 series.

No service here, this is for pro/enterprise

### DIFF
--- a/src/lib/db/index.ts
+++ b/src/lib/db/index.ts
@@ -41,7 +41,7 @@ import { DependentFeaturesStore } from '../features/dependent-features/dependent
 import LastSeenStore from '../features/metrics/last-seen/last-seen-store';
 import FeatureSearchStore from '../features/feature-search/feature-search-store';
 import { InactiveUsersStore } from '../users/inactive/inactive-users-store';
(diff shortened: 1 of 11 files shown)


chriswk:
May I suggest creating this as a package by feature? This seems like a good candidate for a clear single feature.

daveleek:
Yup, it felt a bit small while building it, but the rest of this feature will be packaged by feature so probably good for consistency to move it here as well

daveleek:
I've moved this into a separate folder

chriswk:
The folder structure seems good, but I think we're running into the overloaded concepts issue again. the features folder are for things related to feature (toggle). I'd move this to `src/lib/traffic-data-usage`. This doesn't directly relate to features.

chriswk (approved):
Still giving a :+1: . I think it would be better to have traffic-data-usage on lib/ but I don't want that to block us.

PR #6191, opened 2024-02-09
chore: Move actions inside projects

gastonfournier:
## About the changes
Action should be relative to a project (in general).

### DIFF
--- a/frontend/src/hooks/api/actions/useActionsApi/useActionsApi.ts
+++ b/frontend/src/hooks/api/actions/useActionsApi/useActionsApi.ts
@@ -13,13 +13,11 @@ export type ActionSetPayload = Omit<
     actions: ActionPayload[];
 };
 
-export const useActionsApi = (project?: string) => {
+export const useActionsApi = (project: string) => {
     const { loading, makeRequest, createRequest, errors } = useAPI({
         propagateErrors: true,
     });
-    const endpoint = project
-        ? `api/admin/projects/${project}/actions`
-        : 'api/admin/actions';
+    const endpoint = `api/admin/projects/${project}/actions`;
 
     const addActionSet = async (actionSet: ActionSetPayload) => {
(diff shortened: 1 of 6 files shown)


frontend/src/hooks/api/actions/useActionsApi/useActionsApi.ts
```diff
@@ -15,15 +13,18 @@ export type ActionSetPayload = Omit<
...
 };
 
-export const useActionsApi = () => {
+export const useActionsApi = (project?: string) => {
     const { loading, makeRequest, createRequest, errors } = useAPI({
         propagateErrors: true,
     });
+    const endpoint = project
+        ? `api/admin/projects/${project}/actions`
+        : 'api/admin/actions';
```
nunogois, line 22: Do we want this? 🤔

frontend/src/hooks/api/getters/useActions/useActions.ts
```diff
@@ -8,14 +8,18 @@ import { useUiFlag } from 'hooks/useUiFlag';
...
 
-    const { data, error, mutate } = useConditionalSWR(
+    const { data, error, mutate } = useConditionalSWR<{
+        actions: IActionSet[];
+    }>(
         isEnterprise() && actionsEnabled,
         { actions: [] },
-        formatApiPath(ENDPOINT),
+        formatApiPath(
+            project ? `api/admin/projects/${project}/actions` : ENDPOINT,
```
nunogois, line 21: Same here. Do we still want global level actions?

nunogois (approved):

frontend/src/hooks/api/actions/useActionsApi/useActionsApi.ts
gastonfournier: Not sure, I was hesitant to remove this, I think it's not needed given that I removed the endpoints :D

PR #6203, opened 2024-02-12
refactor: project actions

nunogois:
https://linear.app/unleash/issue/2-1938/refactor-project-actions

Refactors project actions to not include the project in the payload.

Includes other misc scouting.

### DIFF
--- a/frontend/src/component/incomingWebhooks/IncomingWebhooksModal/IncomingWebhooksForm/useIncomingWebhooksForm.ts
+++ b/frontend/src/component/incomingWebhooks/IncomingWebhooksModal/IncomingWebhooksForm/useIncomingWebhooksForm.ts
@@ -99,8 +99,6 @@ export const useIncomingWebhooksForm = (incomingWebhook?: IIncomingWebhook) => {
             return false;
         }
 
-        // TODO call backend to check if token name is unique
-
         clearError(ErrorField.TOKEN_NAME);
         return true;
     };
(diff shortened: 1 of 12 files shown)


frontend/src/component/incomingWebhooks/IncomingWebhooksModal/IncomingWebhooksForm/useIncomingWebhooksForm.ts
```diff
@@ -99,8 +99,6 @@ export const useIncomingWebhooksForm = (incomingWebhook?: IIncomingWebhook) => {
             return false;
         }
 
-        // TODO call backend to check if token name is unique
-
```
nunogois, line 103: This is not needed since:
 - We only hit this when creating a new incoming webhook, so the token name will always be unique for that webhook (there are no other tokens);
 - Other times are covered by a real-time check here: https://github.com/Unleash/unleash/blob/c224d7dc4cf2659179531f6fed6c90444b0346fd/frontend/src/component/incomingWebhooks/IncomingWebhooksModal/IncomingWebhooksForm/IncomingWebhooksTokens/IncomingWebhooksTokensCreateDialog.tsx#L44-L45

frontend/src/component/project/Project/ProjectSettings/ProjectActions/ProjectActionsTable/ProjectActionsModal/ProjectActionsForm/useProjectActionsForm.ts
```diff
@@ -41,31 +41,26 @@ export const useProjectActionsForm = (action?: IActionSet) => {
...
+        setActorId(action?.actorId ?? 0);
+        setActions(
+            action?.actions?.map((action) => ({
+                id: uuidv4(),
+                action: action.action,
+                sortOrder: action.sortOrder,
+                executionParams: action.executionParams,
+            })) ?? [],
+        );
+        setValidated(false);
```
nunogois, line 63: Previous logic meant that we would not reset actions and filters, since all of their logic was inside the `if`s.
E.g. we create a new action, and then try creating a new action, we'll see the same actions and filters as the previous action.

gastonfournier (approved):
Looks great! :rocket:

PR #6216, opened 2024-02-13
feat: sdk reporting flag and e2e test

sjaanus:
1. Add flag
2. Add e2e test with more complete example
3. Some bug fixes

### DIFF
--- a/src/lib/features/project/project-store.ts
+++ b/src/lib/features/project/project-store.ts
@@ -742,7 +742,7 @@ class ProjectStore implements IProjectStore {
                     instances: [row.instance_id],
                     sdks: [getSdk(sdkParts)],
                 };
-                entriesMap.set(row.feature_name, entry);
+                entriesMap.set(row.app_name, entry);
                 orderedEntries.push(entry);
             }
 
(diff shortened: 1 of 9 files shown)


src/lib/features/project/project-store.ts
```diff
@@ -742,15 +742,17 @@ class ProjectStore implements IProjectStore {
                     instances: [row.instance_id],
                     sdks: [getSdk(sdkParts)],
                 };
-                entriesMap.set(row.feature_name, entry);
+                entriesMap.set(row.app_name, entry);
```
FredrikOseberg, line 745: This change seems significant. I assume it's intended.

src/lib/features/project/project-store.ts
```diff
@@ -742,15 +742,17 @@ class ProjectStore implements IProjectStore {
...
                 orderedEntries.push(entry);
             }
 
             const sdk = entry.sdks.find((sdk) => sdk.name === sdkParts[0]);
             if (!sdk) {
                 entry.sdks.push(getSdk(sdkParts));
             } else {
-                sdk.versions.push(sdkParts[1]);
+                if (!sdk.versions.includes(sdkParts[1])) {
```
FredrikOseberg, line 753: We never check if sdkParts exists after splitting, can there be a scenario where we don't have it?

FredrikOseberg (approved):
Approved, but some questions remain

src/lib/features/project/project-store.ts
sjaanus: Yes, this was typo basically, which was caught by test.

src/lib/features/project/project-store.ts
sjaanus: That is good question, also we might not have the value at all. Probably we need to add more validations in future.

PR #6223, opened 2024-02-13
feat: permission matrix (PoC)

gastonfournier:
## About the changes
This is a rough initial version as a PoC for a permission matrix. 

This is only available after enabling the flag `userAccessUIEnabled` that is set to true by default in local development.

The access was added to the users' admin page but could be embedded in different contexts (e.g. when assigning a role to a user):
image

This is how the matrix looks like
screencapture-localhost-3000-admin-users-3-access-2024-02-13-12_15_44


frontend/src/component/common/ConditionallyRender/ConditionallyRender.tsx
```diff
@@ -15,11 +15,12 @@ type TargetElement =
 
 type RenderFunc = () => JSX.Element;
 
-export const ConditionallyRender = ({
+export const ConditionallyRender: React.FC<IConditionallyRenderProps> = ({
+    children,
```
gastonfournier, line 19: This change is unrelated but allows us to use this syntax:
```typescript
            <ConditionallyRender condition={...}>
                <MyComponent />
            </ConditionallyRender>
```

Instead of this one which I find more awkward:
```typescript
            <ConditionallyRender condition={...} show={<MyComponent />} />
```

I don't have strong opinions on this change, so I'm happy to revert this (cc @FredrikOseberg as you may have some insights on whether this is a good change or not)

chriswk:
But I would love to have the matrix, One thing I see is the struggle with hierarchical permissions due to the 8 sub permissions of update project being added.

The matrix will list you as having UPDATE_PROJECT, but not PROJECT_USER_ACCESS_WRITE, but with UPDATE_PROJECT you can do what PROJECT_USER_ACCESS_WRITE can. So this makes it clear that we either should deprecate/remove the UPDATE_PROJECT permission, or find out how to model hierarchical permissions.

frontend/src/component/common/ConditionallyRender/ConditionallyRender.tsx
gastonfournier: Yes, it is. The else case can't be modeled easily, so this gives you some flexibility, you either use `show` with `elseShow` or you can also do:

```typescript
<ConditionallyRender condition={...} elseShow={<AnotherComponent />}>
    <MyComponent />
</ConditionallyRender>
```

Which I don't like that much. But for the sake of this PR I will split this apart.

gastonfournier:
> But I would love to have the matrix, One thing I see is the struggle with hierarchical permissions due to the 8 sub permissions of update project being added.
> 
> The matrix will list you as having UPDATE_PROJECT, but not PROJECT_USER_ACCESS_WRITE, but with UPDATE_PROJECT you can do what PROJECT_USER_ACCESS_WRITE can. So this makes it clear that we either should deprecate/remove the UPDATE_PROJECT permission, or find out how to model hierarchical permissions.

I'd expect our permission checker to take this hierarchy into account, so the access-service would be the single source of truth to validate whether you have access or not to perform an action. Maybe that way we can also specify:
```typescript
permission: [PROJECT_USER_ACCESS_WRITE], // knowing that UPDATE_PROJECT grants PROJECT_USER_ACCESS_WRITE
```
rather than
```typescript
permission: [UPDATE_PROJECT, PROJECT_USER_ACCESS_WRITE],
```

WDYT?

nunogois (approved):
Thank you for considering my suggestions!

frontend/src/component/common/ConditionallyRender/ConditionallyRender.tsx
FredrikOseberg: @gastonfournier 

I think we should keep consistency in the patterns we use, passing components as props for composition is quite common to do in React. I don't think we gain that much more clarity by adding another way to do the `show` part. If we do this I'd like to revisit how we do the other ConditionallyRenders in other parts of the application as well, but that is a much larger task.

gastonfournier: For an if (condition) -> component I like https://github.com/Unleash/unleash/pull/6223#discussion_r1487670255 better, but for else, I don't so much. I think it makes things more readable, getting read of the property and using children, is more html-like. But I reverted this change for the sake of this feature :)

PR #6225, opened 2024-02-13
feat: dashboard lead time gauge

Tymek:
image

### DIFF
--- a/frontend/src/component/executiveDashboard/Gauge/Gauge.tsx
+++ b/frontend/src/component/executiveDashboard/Gauge/Gauge.tsx
@@ -1,4 +1,6 @@
+import { VFC } from 'react';
 import { Box, useTheme } from '@mui/material';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 
 const polarToCartesian = (
     centerX: number,
@@ -113,7 +115,13 @@ const GaugeText = () => {
     );
 };
 
-export const Gauge = ({ value = 100, min = 0, max = 100 }) => {
+type GaugeProps = {
+    value?: number;
+    min?: number;
+    max?: number;
+};
+
+export const Gauge: VFC<GaugeProps> = ({ value, min = 0, max = 100 }) => {
     const theme = useTheme();
     const radius = 1;
     const lineWidth = 0.25;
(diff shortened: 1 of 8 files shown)


frontend/src/component/executiveDashboard/Gauge/Gauge.tsx
```diff
@@ -0,0 +1,195 @@
...
+const degreesToRadians = (degrees: number) => degrees * (Math.PI / 180);
+
+const normalizeValue = (value: number, min: number, max: number) =>
+    (value - min) / (max - min);
+
+const describeArc = (radius: number, startAngle: number, endAngle: number) => {
+    const start = polarToCartesian(0, 0, radius, endAngle);
+    const end = polarToCartesian(0, 0, radius, startAngle);
+    const largeArcFlag = endAngle - startAngle <= Math.PI ? '0' : '1';
+    const d = [
```
FredrikOseberg, line 26: This variable name is very obscure. Can we improve the naming?

frontend/src/component/executiveDashboard/Gauge/Gauge.tsx
```diff
@@ -0,0 +1,195 @@
...
+    const angles = [180 + 50, 180 + 90 + 40];
+
+    return (
+        <>
+            {angles.map(degreesToRadians).map((angle) => {
+                const start = polarToCartesian(0, 0, startRadius, angle);
+                const end = polarToCartesian(0, 0, endRadius, angle);
+
+                return (
+                    <>
```
FredrikOseberg, line 58: Don't we need to encapsulate these with svg?

EDIT: I see they are encapsulated further down. But maybe a comment that this needs to be used in the context of an svg tag.

frontend/src/themes/dark-theme.ts
```diff
@@ -278,6 +278,19 @@ const theme = {
...
+        /**
+         * Dashboard and charts
+         */
+        charts: {
+            gauge: {
+                gradientStart: '#4C4992',
+                gradientEnd: '#9792ED',
+                background: '#39384C',
+                sectionLine: '#8c89bf',
+                text: colors.grey[800],
```
FredrikOseberg, line 291: These should be coming from the theme? Or align with Nicolae which theme colors to use instead.

FredrikOseberg (approved):
Approved with some comments. This seems to be a very complex component, I wish there was an easier way to do this as it will add a lot of cognitive load to future work in this area. But, I trust that you use your judgment to find the simplest solution and it may be that the graphics of the widget require this approach.

frontend/src/themes/dark-theme.ts
Tymek: That's the approach in entire dark theme, unfortunately

Tymek: I aligned with Nicolae. We already have some of the colors used in theme, but in different context, and I didn't want to mix it. Using `palette.secondary.main` next to `palette.spotlight.border` is confusing. I don't want other components to influence colors on this chart.

FredrikOseberg: We don't want hard coded colors. Period. So we need to align and add these colors to theme as more appropriate labels. This is technical debt accumulation.

PR #6235, opened 2024-02-14
feat: show info on what would be deleted

thomasheartman:
This PR updates the way we show deleted strategies in the CR UI. Instead of showing just the strategy name and a diff on hover, we show the same strategy config as we do for new and updated strategies.

This makes it easier to see what you have deleted.

In doing so, it also fixes two issues:
1. inconsistent border radius for segment changes listed. Due to an override in `frontend/src/themes/theme.ts`, these would get a border radius of `theme.shape.borderRadiusLarge` instead of `theme.shape.borderRadiusMedium`. It does this by adding a class and making the selector more specific.
2. The background was unset for the strategy rollout box and constraint item boxes.

It looks like this:

image

image

Or with more kinds of strategies:
image

Note: I'm happy to isolate the color changes to a separate PR if that's preferable.

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
@@ -66,6 +66,13 @@ const InlineList = styled('ul')(({ theme }) => ({
     },
 }));
 
+const ChangeBox = styled(Box)(({ theme }) => ({
+    padding: theme.spacing(3),
+    '&:has(.delete-strategy)': {
+        backgroundColor: theme.palette.error.light,
+    },
+}));
+
 export const FeatureChange: FC<{
     actions: ReactNode;
     index: number;
@@ -134,14 +141,7 @@ export const FeatureChange: FC<{
                 }
             />
 
-            <Box
-                sx={(theme) => ({
-                    padding: theme.spacing(3),
-                    '&:has(.delete-strategy)': {
-                        backgroundColor: theme.palette.error.light,
-                    },
-                })}
-            >
+            <ChangeBox>
                 {(change.action === 'addDependency' ||
                     change.action === 'deleteDependency') && (
                     <DependencyChange
@@ -190,7 +190,7 @@ export const FeatureChange: FC<{
                         actions={actions}
                     />
                 )}
-            </Box>
+            </ChangeBox>
         </StyledSingleChangeBox>
     );
 };
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
@@ -134,7 +134,14 @@ export const FeatureChange: FC<{
                 }
             />
 
-            <Box sx={(theme) => ({ padding: theme.spacing(3) })}>
+            <Box
+                sx={(theme) => ({
+                    padding: theme.spacing(3),
+                    '&:has(.delete-strategy)': {
+                        backgroundColor: theme.palette.error.light,
+                    },
+                })}
+            >
                 {(change.action === 'addDependency' ||
                     change.action === 'deleteDependency') && (
                     <DependencyChange
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
@@ -66,7 +66,7 @@ const InlineList = styled('ul')(({ theme }) => ({
     },
 }));
 
-const ChangeBox = styled(Box)(({ theme }) => ({
+const ChangeInnerBox = styled(Box)(({ theme }) => ({
     padding: theme.spacing(3),
     '&:has(.delete-strategy)': {
         backgroundColor: theme.palette.error.light,
@@ -141,7 +141,7 @@ export const FeatureChange: FC<{
                 }
             />
 
-            <ChangeBox>
+            <ChangeInnerBox>
                 {(change.action === 'addDependency' ||
                     change.action === 'deleteDependency') && (
                     <DependencyChange
@@ -190,7 +190,7 @@ export const FeatureChange: FC<{
                         actions={actions}
                     />
                 )}
-            </ChangeBox>
+            </ChangeInnerBox>
         </StyledSingleChangeBox>
     );
 };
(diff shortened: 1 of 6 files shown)


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
```diff
@@ -134,7 +134,14 @@ export const FeatureChange: FC<{
...
 
-            <Box sx={(theme) => ({ padding: theme.spacing(3) })}>
+            <Box
+                sx={(theme) => ({
+                    padding: theme.spacing(3),
+                    '&:has(.delete-strategy)': {
+                        backgroundColor: theme.palette.error.light,
+                    },
+                })}
+            >
```
thomasheartman, line 144: Because the container is made the same way regardless of what content it will have, we need some way to check whether it is a box for deleting strategies. Luckily, the `has` selector will do that for us. In unsupported browsers, it'll just fall back to using the regular (theme) background color.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -25,7 +25,7 @@ export const ChangeItemWrapper = styled(Box)({
     alignItems: 'center',
 });
 
-const ChangeItemCreateEditWrapper = styled(Box)(({ theme }) => ({
+const ChangeItemCreateEditDeleteWrapper = styled(Box)(({ theme }) => ({
```
thomasheartman, line 28: Rename the component to make it clearer that it's being used also for deleted items now.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -187,34 +187,48 @@ export const StrategyChange: VFC<{
...
+                        condition={Boolean(currentStrategy)}
+                        show={
+                            <Typography>
+                                {
+                                    <StrategyExecution
+                                        strategy={currentStrategy!}
+                                    />
+                                }
+                            </Typography>
+                        }
```
thomasheartman, line 221: This is the real meat of this PR (and the only actually necessary thing).

frontend/src/component/common/SegmentItem/SegmentItem.tsx
```diff
@@ -25,7 +25,9 @@ const StyledAccordion = styled(Accordion, {
     shouldForwardProp: (prop) => prop !== 'isDisabled',
 })<{ isDisabled: boolean | null }>(({ theme, isDisabled }) => ({
     border: `1px solid ${theme.palette.divider}`,
-    borderRadius: theme.shape.borderRadiusMedium,
+    '&.segment-accordion': {
+        borderRadius: theme.shape.borderRadiusMedium,
+    },
```
thomasheartman, line 30: Increase the specificity of this selector to override the default for mui accordions (mentioned in the PR description).

We could also use the [`:is` specificity trick](https://css-tricks.com/where-has-a-cool-specificity-trick-too/) and do something like this to avoid adding the class:
```ts
    '&:is(*,#selector-for-specificity)': {
        borderRadius: theme.shape.borderRadiusMedium,
    },
```

But I think using a class is clearer.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
nicolaesocaciu: I think this is fine, its a good compromise

frontend/src/component/common/SegmentItem/SegmentItem.tsx
nicolaesocaciu: Yes i agree, i think its clearer with a class

nicolaesocaciu (approved):
Everything looks good for me!

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
thomasheartman: Oh, and for the record, caniuse reports that [91.96% of browsers support `:has`](https://caniuse.com/?search=%3Ahas).

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
```diff
@@ -134,7 +134,14 @@ export const FeatureChange: FC<{
                 }
             />
 
-            <Box sx={(theme) => ({ padding: theme.spacing(3) })}>
+            <Box
```
FredrikOseberg, line 137: Should avoid SX when we can and use styled components, especially for complex styles that clutter the render function.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -187,34 +187,48 @@ export const StrategyChange: VFC<{
...
-                                    change={change}
-                                    currentStrategy={currentStrategy}
-                                />
-                            </StrategyTooltipLink>
-                        )}
-                    </ChangeItemInfo>
-                    <div>{actions}</div>
-                </ChangeItemWrapper>
+                <>
+                    <ChangeItemCreateEditDeleteWrapper className='delete-strategy'>
```
FredrikOseberg, line 191: Can we have a more specific className? I can see us creating a similar className in the future and causing conflicts.

FredrikOseberg (approved):
LGTM, minor comments.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
thomasheartman: Happy to switch to that, yeah!

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
thomasheartman: Sure can! Let me see what I can cook up

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/FeatureChange.tsx
thomasheartman: Addressed in e7807eed1fe76524ce6eb00b56170bce58c92f3e.

I know we've said not to use `sx`, but I still see it all over the code base. Is that an artefact from before we switched to styled components or is it acceptable in some cases?

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
thomasheartman: Addressed in 938d21663ee9faed3f9487012c4c9e5906853bb4

PR #6236, opened 2024-02-14
feat: project applications server side paging and sorting and filtering

sjaanus:
Uses exactly same pattern as search-store. Nothing too crazy here.
Most code is in tests.

### DIFF
--- a/src/lib/features/project/project-applications.e2e.test.ts
+++ b/src/lib/features/project/project-applications.e2e.test.ts
@@ -331,6 +331,14 @@ test('should show correct number of total', async () => {
         started: Date.now(),
         interval: 10,
     });
+    await app.request.post('/api/client/register').send({
+        appName: metrics.appName,
+        instanceId: 'another-instance',
+        strategies: ['default'],
+        sdkVersion: 'unleash-client-test',
+        started: Date.now(),
+        interval: 10,
+    });
     const secondApp = 'second-app';
     await app.request.post('/api/client/register').send({
         appName: secondApp,
--- a/src/lib/features/project/project-applications.e2e.test.ts
+++ b/src/lib/features/project/project-applications.e2e.test.ts
@@ -89,19 +89,22 @@ test('should return applications', async () => {
         .expect('Content-Type', /json/)
         .expect(200);
 
-    expect(body).toMatchObject([
-        {
-            environments: ['default'],
-            instances: ['instanceId'],
-            name: 'appName',
-            sdks: [
-                {
-                    name: 'unleash-client-test',
-                    versions: ['1.2'],
-                },
-            ],
-        },
-    ]);
+    expect(body).toMatchObject({
+        applications: [
+            {
+                environments: ['default'],
+                instances: ['instanceId'],
+                name: 'appName',
+                sdks: [
+                    {
+                        name: 'unleash-client-test',
+                        versions: ['1.2'],
+                    },
+                ],
+            },
+        ],
+        total: 1,
+    });
 });
 
 test('should return applications if sdk was not in database', async () => {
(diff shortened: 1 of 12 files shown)


src/lib/features/project/project-store.ts
```diff
@@ -579,32 +581,75 @@ class ProjectStore implements IProjectStore {
...
+            .with('ranked', (qb) => {
+                const hasSearchParams = searchParams?.length;
+                if (hasSearchParams) {
+                    const sqlParameters = searchParams.map(
+                        (item) => `%${item}%`,
+                    );
+                    const sqlQueryParameters = sqlParameters
+                        .map(() => '?')
+                        .join(',');
+                    const columns = [
```
sjaanus, line 607: Filtering can be done on all columns

src/lib/features/project/project-applications.e2e.test.ts
```diff
@@ -163,17 +169,222 @@ test('should return application without version if sdk has just name', async ()
...
+                        versions: [],
+                    },
+                ],
+            },
+        ],
+        total: 1,
+    });
+});
+
+test('should show correct number of total', async () => {
```
FredrikOseberg, line 323: There is a lot of duplication of the same calls in these tests, I think you have an opportunity to refactor here to make the tests less verbose and more succinct.

src/lib/features/project/project-store.ts
```diff
@@ -579,32 +581,84 @@ class ProjectStore implements IProjectStore {
     }
 
     async getApplicationsByProject(
-        projectId: string,
-    ): Promise<IProjectApplication[]> {
+        params: IProjectApplicationsSearchParams,
+    ): Promise<IProjectApplications> {
+        const { project, limit, sortOrder, sortBy, searchParams, offset } =
+            params;
+        const validatedSortOrder =
```
FredrikOseberg, line 588: Maybe this could be done in the schema level and reject requests with 400 bad request the contents are not 'asc' or 'desc' ?

src/lib/features/project/project-store.ts
```diff
@@ -579,32 +581,84 @@ class ProjectStore implements IProjectStore {
...
+            sortOrder === 'asc' || sortOrder === 'desc' ? sortOrder : 'asc';
         const query = this.db
             .with('applications', (qb) => {
                 qb.select('project', 'app_name', 'environment')
                     .distinct()
                     .from('client_metrics_env as cme')
                     .leftJoin('features as f', 'cme.feature_name', 'f.name')
-                    .where('project', projectId);
+                    .where('project', project);
```
FredrikOseberg, line 596: We should decide on a standard here throughout the codebase, I think it's potentially confusing that we sometimes use `project` and sometimes `projectId`.

src/lib/features/project/project-store.ts
```diff
@@ -579,32 +581,84 @@ class ProjectStore implements IProjectStore {
...
-                'a.app_name',
-                'a.environment',
-                'ci.instance_id',
-                'ci.sdk_version',
-            )
-            .from('applications as a')
-            .leftJoin('client_instances as ci', function () {
-                this.on('ci.app_name', 'a.app_name').andOn(
-                    'ci.environment',
+            .with('ranked', (qb) => {
```
FredrikOseberg, line 598: This query is intimidating, are there good patterns we can use to make it more succinct and understandable?

FredrikOseberg (approved):
Leaving a tentative approval, but I've outlined some areas that I think can be improved.

src/lib/features/project/project-store.ts
sjaanus: I want to be absolutely sure also on store level, because this can be called from codebase and is potential SQL injection hole.

src/lib/features/project/project-store.ts
sjaanus: Refactored part of it into shared util. applySearchFilters

src/lib/features/project/project-store.ts
sjaanus: The feature search uses project in its filters, so followed that.

PR #6240, opened 2024-02-14
fix: remove jitter from time sensitive scheduled jobs

kwasniew:
## About the changes

Removing jitter from time sensitive jobs that need to run straight away.


gastonfournier (approved):
LGTM!

PR #6246, opened 2024-02-15
feat: graceful shutdown hooks

chriswk:
Adds a IGracefulShutdown interface to UnleashConfig.

Allows services/stores to register a shutdown hook which will be run before our Knex connection is destroyed, so perfect for flushing data.

Initially thought to be used for traffic metrics to avoid losing up to 14 minutes of data when shutting down, was also able to reuse it for addons and scheduler service, instead of our custom unleashStop() method having to know that it needs to stop exactly these two services, our services can now register a shutdown hook themself.

### DIFF
--- a/src/lib/error/unleash-error.ts
+++ b/src/lib/error/unleash-error.ts
@@ -28,9 +28,9 @@ export const UnleashApiErrorTypes = [
     'InvalidTokenError',
     'OwaspValidationError',
     'ForbiddenError',
-
     // server errors; not the end user's fault
     'InternalError',
+    'DuplicateShutdownHook',
 ] as const;
 
 export type UnleashApiErrorName = (typeof UnleashApiErrorTypes)[number];
--- a/src/lib/error/unleash-error.ts
+++ b/src/lib/error/unleash-error.ts
@@ -30,7 +30,7 @@ export const UnleashApiErrorTypes = [
     'ForbiddenError',
     // server errors; not the end user's fault
     'InternalError',
-    'DuplicateShutdownHook',
+    'DuplicateShutdownHookError',
 ] as const;
 
 export type UnleashApiErrorName = (typeof UnleashApiErrorTypes)[number];
--- a/src/lib/features/scheduler/scheduler-service.ts
+++ b/src/lib/features/scheduler/scheduler-service.ts
@@ -1,8 +1,8 @@
 import EventEmitter from 'events';
-import { Logger, LogProvider } from '../../logger';
+import { Logger } from '../../logger';
 import { IMaintenanceStatus } from '../maintenance/maintenance-service';
 import { SCHEDULER_JOB_TIME } from '../../metric-events';
-import { registerGracefulShutdownHook } from '../../util';
+import { IUnleashConfig } from '../../types';
 
 // returns between min and max seconds in ms
 // when schedule interval is smaller than max jitter then no jitter
@@ -29,15 +29,17 @@ export class SchedulerService {
     private executingSchedulers: Set<string> = new Set();
 
     constructor(
-        getLogger: LogProvider,
+        {
+            getLogger,
+            gracefulShutdown,
+            eventBus,
+        }: Pick<IUnleashConfig, 'getLogger' | 'gracefulShutdown' | 'eventBus'>,
         maintenanceStatus: IMaintenanceStatus,
-        eventBus: EventEmitter,
     ) {
         this.logger = getLogger('/services/scheduler-service.ts');
         this.maintenanceStatus = maintenanceStatus;
         this.eventBus = eventBus;
-        registerGracefulShutdownHook(
-            this.logger,
+        gracefulShutdown.registerGracefulShutdownHook(
             'scheduler-service',
             async () => {
                 this.logger.info('Shutting down');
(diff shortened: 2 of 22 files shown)


src/lib/features/scheduler/scheduler-service.ts
```diff
@@ -28,13 +29,23 @@ export class SchedulerService {
...
         maintenanceStatus: IMaintenanceStatus,
-        eventBus: EventEmitter,
     ) {
         this.logger = getLogger('/services/scheduler-service.ts');
         this.maintenanceStatus = maintenanceStatus;
         this.eventBus = eventBus;
+        gracefulShutdown.registerGracefulShutdownHook(
+            'scheduler-service',
+            async () => {
+                this.logger.info('Shutting down');
```
chriswk, line 45: will remove this line, it was here to see that we did log with the correct logger during implementation phase.

src/lib/error/unleash-error.ts
```diff
@@ -28,9 +28,9 @@ export const UnleashApiErrorTypes = [
     'InvalidTokenError',
     'OwaspValidationError',
     'ForbiddenError',
-
     // server errors; not the end user's fault
     'InternalError',
+    'DuplicateShutdownHook',
```
nunogois, line 33:
```suggestion
    'DuplicateShutdownHookError',
```

src/lib/util/graceful-shutdown.ts
```diff
@@ -1,7 +1,60 @@
-import { Logger } from '../logger';
+import { Logger, LogProvider } from '../logger';
 import { IUnleash } from '../types/core';
+import { GenericUnleashError } from '../error/unleash-error';
 
-function registerGracefulShutdown(unleash: IUnleash, logger: Logger): void {
+export interface IGracefulShutdown {
+    registerGracefulShutdownHook(
+        serviceName: string,
+        hook: () => Promise<void>,
```
nunogois, line 8: Maybe this helps in case we want to register a non-promise shutdown hook. I noticed a lot of `async` shutdown hooks that did not need to be `async` since they did not `await` anything.
```suggestion
        hook: () => void | Promise<void>,
```

src/lib/util/graceful-shutdown.ts
```diff
@@ -1,7 +1,60 @@
...
+    ): void;
+    resetShutdownHooks(): void;
+    executeShutdownHooks(): Promise<void>;
+}
+
+export class GracefulShutdownHookManager implements IGracefulShutdown {
+    private shutdownHooks: Map<string, () => Promise<void>>;
+    private logger: Logger;
+    constructor(getLogger: LogProvider) {
+        this.shutdownHooks = new Map<string, () => Promise<void>>();
```
nunogois, line 18: I think it's worth extracting the shutdown hook type to something like:
```ts
type ShutdownHook = () => void | Promise<void>
```

src/lib/util/graceful-shutdown.ts
```diff
@@ -1,7 +1,60 @@
...
+        serviceName: string,
+        hook: () => Promise<void>,
+    ): void {
+        this.logger.info(`Registering graceful shutdown for ${serviceName}`);
+        if (this.shutdownHooks.has(serviceName)) {
+            throw new GenericUnleashError({
+                name: 'DuplicateShutdownHook',
+                message: `${serviceName} already registered for graceful shutdown`,
+                statusCode: 500,
+            });
```
nunogois, line 45: Might be cleaner to create a new `src/lib/error/duplicate-shutdown-hook-error.ts`

src/lib/util/graceful-shutdown.ts
```diff
@@ -1,7 +1,60 @@
-import { Logger } from '../logger';
+import { Logger, LogProvider } from '../logger';
 import { IUnleash } from '../types/core';
+import { GenericUnleashError } from '../error/unleash-error';
 
-function registerGracefulShutdown(unleash: IUnleash, logger: Logger): void {
+export interface IGracefulShutdown {
```
nunogois, line 5: This is not an interface for the shutdown itself but for a manager of shutdowns.
```suggestion
export interface IGracefulShutdownHookManager {
```

nunogois (approved):
Neat!

src/lib/util/graceful-shutdown.ts
chriswk: Agreed, added in 939054192. As well as a test proving that we can add and execute synchronous hooks as well

src/lib/util/graceful-shutdown.ts
chriswk: hmm, I'm not sure if we want the manager in the interface name. What do you think of `IGracefulShutdownHookRegistry` ? I think I like that better for the class as well.

chriswk: Fixed in cb54915dd

src/lib/util/graceful-shutdown.ts
chriswk: Yeah, fixed in cb54915dd

src/lib/util/graceful-shutdown.ts
chriswk: Yeah, I like it

chriswk: handled in b492e8033

chriswk:
Superseded by #6585

Issue #5785, opened 2024-01-08
Endpoint /api/frontend throws UnknownError when configured with AuthType.None


daveleek:
### Describe the bug

When configuring Unleash with AuthType.None the endpoint /api/frontend breaks both with and without token.
It throws an "UnknownError" with the error message:
> Cannot destructure property 'type' of 'undefined' as it is undefined.

The desired behavior is shown in the endpoint /api/client/features which works both with and without token when Unleash is configured with AuthType.None

### Steps to reproduce the bug

1. export AUTH_TYPE=none
2. Start Unleash locally
3. Attempt to GET http://localhost:4242/api/frontend without specifying an Authorization header

### Expected behavior

Should return all toggles belonging to default project and the result of their evaluation.
If no token is supplied, the api should return all toggles belonging to default project and the result of their evaluation
If a token is supplied, the token's environment and project data should be respected.

### Logs, error output, etc.

```bash
[dev:backend] [2024-01-08T12:12:36.630] [DEBUG] controller/ProxyController - I encountered an error that wasn't an instance of the `UnleashError` type. The original error was: {}. It was mapped to {
[dev:backend]   "id": "85701fbb-1e14-4078-acf0-2db0e8126ea1",
[dev:backend]   "name": "UnknownError",
[dev:backend]   "message": "Cannot destructure property 'type' of 'undefined' as it is undefined.",
[dev:backend]   "details": [
[dev:backend]     {
[dev:backend]       "message": "Cannot destructure property 'type' of 'undefined' as it is undefined.",
[dev:backend]       "description": "Cannot destructure property 'type' of 'undefined' as it is undefined."
[dev:backend]     }
[dev:backend]   ]
[dev:backend] }
[dev:backend] [2024-01-08T12:12:36.630] [ERROR] controller/ProxyController - Server failed executing request: {} TypeError: Cannot destructure property 'type' of 'undefined' as it is undefined.
[dev:backend]     at ProxyService.assertExpectedTokenType (/Users/davidleek/Documents/code/unleash/dist/lib/services/proxy-service.js:79:38)
[dev:backend]     at ProxyService.clientForProxyToken (/Users/davidleek/Documents/code/unleash/dist/lib/services/proxy-service.js:45:22)
[dev:backend]     at ProxyService.getProxyFeatures (/Users/davidleek/Documents/code/unleash/dist/lib/services/proxy-service.js:24:35)
[...]

PR #6247, opened 2024-02-15
fix: Allow AuthType None to use valid API tokens

### LINKED ISSUE #5799: When AuthType is `none`, `api/client/features` ignores tokens
### Describe the bug

If you set `AuthType.None`, the `api/client/features` endpoint will ignore any auth token you pass to it. This results in it always returning data for all projects and the `default` environment. 

In other words, it is **not possible** to get features for a different environment or to get a payload that only includes a subset of all the projects that exist.

Somewhat related to #5785.

1. Start Unleash with AuthType=None (either through env variables or through the Unleash auth options)
2. Have a project with some flags in it. Define configuration for at least one of the flags in the `development` environment (or any non-`default` environment) and enable it for that environment. Make sure the flag is not enabled for the `default` env (shouldn't be possible unless you reactivate the env)
3. Create a token for that environment (or re-use an existing one)
4. Using the token you created, request flags from the `api/client/features` endpoint.
5. Inspect the payload, notice that:
  i. The flag has: `enabled: false`
  ii. The flag has an empty list of strategies. No configuration.

### Expected behavior

When you provide an auth token to a client features request, the auth token's env and project settings should be respected.

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

Appears to be because we don't use 

```js
    app.use(baseUriPath, apiTokenMiddleware(config, services));
```

when you use auth type none, so we don't pass the token's data to the client request handler. Adding that line to the handler would fix this one specific issue, but it breaks a lot of tests because it disables the option to call the endpoint with no auth at all. This was attempted in #5764.

The fact that you can call this endpoint without the auth token may not be intended, but it *is* how it works today, so removing it could be considered breaking (especially when we have about 40 tests that rely on that behavior).

### Unleash version

Tip of main a couple days ago.

### Subscription type

None

### Hosting type

None

_No response_

ivarconr:
Fixes #5799 and #5785

When you do not provide a token we should resolve to the "default" environment to maintain backward compatibility. If you actually provide a token we should prefer that and even block the request if it is not valid. 

An interesting fact is that "default" environment is not available on a fresh installation of Unleash. This means that you need to provide a token to actually get access to toggle configurations. 

**Todo:**

- [x] Add a few tests to showcase how it is intended to work. 
- [x] Add deprecation warnings in logs and documentation.


src/lib/middleware/no-authentication.ts
```diff
@@ -12,4 +14,38 @@ function noneAuthentication(basePath: string, app: Application): void {
         next();
     });
 }
+
+export function noApiToken(baseUriPath: string, app: Application) {
+    app.use(`${baseUriPath}/api/frontend`, (req, res, next) => {
+        console.log(req.path);
```
ivarconr, line 20: remove

src/lib/openapi/spec/user-schema.ts
```diff
@@ -11,7 +11,6 @@ export const userSchema = {
         id: {
             description: 'The user id',
             type: 'integer',
-            minimum: 0,
```
ivarconr, line 14: This breaks other assumptions.

daveleek: Good point!

daveleek (approved):
LGTM!

src/lib/app.ts
```diff
@@ -147,6 +147,8 @@ export default async function getApp(
             break;
         }
         case IAuthType.NONE: {
+            noApiToken(baseUriPath, app);
+            app.use(baseUriPath, apiTokenMiddleware(config, services));
```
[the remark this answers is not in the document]
ivarconr: I agree that app.ts is complex, but I do not feel this change is making it worse as it follow the same pattern as already established.

website/docs/using-unleash/deploy/configuring-unleash.md
```diff
@@ -63,8 +63,8 @@ unleash.start(unleashOptions);
   - `type` / `AUTH_TYPE`: `string` —  What kind of authentication to use. Possible values
     - `open-source` - Sign in with username and password. This is the default value.
     - `custom` - If implementing your own authentication hook, use this
-    - `none` - Turn off authentication all together
     - `demo` - Only requires an email to sign in (was default in v3)
+    - `none` - _Deprecated_ Turn off authentication all together. If no API token is proved towards /api/client or /api/frontend you will receive configuration for the "default" environment. We generally recommend to use the `demo` type for simple unsecure usage of Unleash. This auth-type has many known limitations, particularly related to personalized capabilities such as "favourites" and "notifications.
```
kwasniew, line 67: typo: proved -> provided

kwasniew: in code we use favorites not favourites so I'd stay consistent

src/server-dev.ts
```diff
@@ -53,6 +53,7 @@ process.nextTick(async () => {
                     },
                 },
                 authentication: {
+                    // type: IAuthType.NONE,
```
kwasniew, line 56: deliberate?

src/test/e2e/api/client/feature.auth-none.e2e.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        {
+            tokenName: 'test',
+            type: ApiTokenType.CLIENT,
+            environment: DEFAULT_ENV,
+            projects: ['default'],
+        },
+    );
+    clientSecret = token.secret;
+
+    const fToken =
```
kwasniew, line 73: fToken -> frontendToken

kwasniew (approved):
Nice!

src/server-dev.ts
ivarconr: we should not use none developing locally.

src/lib/middleware/no-authentication.ts
```diff
@@ -12,4 +15,37 @@ function noneAuthentication(basePath: string, app: Application): void {
         next();
     });
 }
+
+export function noApiToken(baseUriPath: string, app: Application) {
+    app.use(`${baseUriPath}/api/frontend`, (req, res, next) => {
+        //@ts-ignore
```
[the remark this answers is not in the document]
ivarconr: yes, requires us to specify the type of the request to potentially have a user.

src/test/e2e/api/client/feature.auth-none.e2e.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        .expect(401);
+});
+
+test('returns 3 toggles for correct api token', async () => {
+    return app.request
+        .get('/api/client/features')
+        .set('Authorization', clientSecret)
+        .expect('Content-Type', /json/)
+        .expect(200);
+});
```
thomasheartman, line 112: looks like this is missing a check for how many features are returned

website/docs/using-unleash/deploy/configuring-unleash.md
```diff
@@ -63,8 +63,8 @@ unleash.start(unleashOptions);
   - `type` / `AUTH_TYPE`: `string` —  What kind of authentication to use. Possible values
     - `open-source` - Sign in with username and password. This is the default value.
     - `custom` - If implementing your own authentication hook, use this
-    - `none` - Turn off authentication all together
     - `demo` - Only requires an email to sign in (was default in v3)
+    - `none` - _Deprecated_ Turn off authentication all together. If no API token is provided towards /api/client or /api/frontend you will receive configuration for the "default" environment. We generally recommend to use the `demo` type for simple unsecure usage of Unleash. This auth-type has many known limitations, particularly related to personalized capabilities such as "favorites" and "notifications.
```
thomasheartman, line 67: Good update. Got a few small suggestions:

```suggestion
    - `none` - _Deprecated_ Turn off authentication completely. If no API token is provided towards /`api/client` or `/api/frontend` you will receive configuration for the "default" environment. We generally recommend you use the `demo` type for simple, insecure usage of Unleash. This auth type has many known limitations, particularly related to personalized capabilities such as favorites and [notifications](../../reference/notifications.md). 
```

thomasheartman (commented):
This looks good to me, but I'm on board with Mateusz's comments about removing the ts-ignores (or at least trading them for ts-expects with a reason for expecting the error) and the `//authType: 'none'` comment.

src/server-dev.ts
ivarconr, line 56:
```suggestion
```

src/test/e2e/api/client/feature.auth-none.e2e.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+test('returns 401 for incorrect api token', async () => {
+    return app.request
+        .get('/api/client/features')
+        .set('Authorization', 'some-invalid-token')
+        .expect('Content-Type', /json/)
+        .expect(401);
+});
+
+test('returns 3 toggles for correct api token', async () => {
```
ivarconr, line 106:
```suggestion
test('returns success for correct api token', async () => {
```

PR #6248, opened 2024-02-15
chore: rate limit calling inc webhooks

nunogois:
https://linear.app/unleash/issue/2-1942/rate-limit-incoming-webhooks-call-endpoint

Adds a configurable rate limit to calling incoming webhooks. We're setting a 1RPS limit for now, but I'm open to suggestions.


chriswk (approved):
Looks reasonable enough to me.

PR #6249, opened 2024-02-15
docs: add ADR for logging with error object

chriswk:
I've held an internal knowledge sharing session on this already. If someone can think of a better phrasing for the background, I'm all ears. I think it's just nice to have this documented, so people remember that our logging framework already has a good way to format errors when you use the API `logger.error("<message>", e)`

### DIFF
--- a/website/docs/contributing/ADRs/overarching/logging.md
+++ b/website/docs/contributing/ADRs/overarching/logging.md
@@ -4,12 +4,12 @@ title: "ADR: Logging errors"
 
 ## Background
 
-After debugging multiple errors over the last few years, we've consistently found that when something
+After debugging multiple errors over the last few years, we've consistently found that when something goes wrong, we
+would like as much context as possible to debug faster.
 
 ## Decision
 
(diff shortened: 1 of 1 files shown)


website/docs/contributing/ADRs/overarching/logging.md
```diff
@@ -0,0 +1,40 @@
+---
+title: "ADR: Logging errors"
+---
+
+## Background
+
+After debugging multiple errors over the last few years, we've consistently found that when something
```
kwasniew, line 7: missing statement

kwasniew (commented):
I think there are two decisions hiding in here:
* how to use the error method correctly (2 arg version)
* how to decide what an error is

chriswk:
Yeah, I think that's a fair point. I already mentioned that I think a one on log levels are justified. Should we make another one, and then I'll strip out the wording about what an error is from this one?

kwasniew:
I'm fine with both, no strong opinion here

ivarconr (approved):
Love it!

PR #6251, opened 2024-02-15
chore: emit PROXY_REPOSITORY_CREATED event when creating new repo

daveleek:
## About the changes

- Emits a new event on the eventBus when Proxy-service creates a new repository for a frontend token
- Adds a prometheus metrics counter for created proxy-repositories

image

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -227,6 +227,10 @@ export default class MetricsMonitor {
             name: 'event_created_by_migration_count',
             help: 'Event createdBy migration count',
         });
+        const proxyRepositoriesCreated = createCounter({
+            name: 'proxy_repositories_created',
+            help: 'Proxy repositories created',
+        });
 
         async function collectStaticCounters() {
             try {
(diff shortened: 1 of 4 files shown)


FredrikOseberg (approved):
LGTM

src/lib/services/proxy-service.ts
```diff
@@ -109,6 +110,7 @@ export class ProxyService {
         if (!client) {
             client = this.createClientForProxyToken(token);
             this.clients.set(token.secret, client);
+            this.config.eventBus.emit(PROXY_REPOSITORY_CREATED, {});
```
FredrikOseberg, line 113: Is the second argument needed?

src/lib/metrics.ts
```diff
@@ -227,6 +227,10 @@ export default class MetricsMonitor {
             name: 'event_created_by_migration_count',
             help: 'Event createdBy migration count',
         });
+        const proxyRepositoriesCreated = createCounter({
+            name: 'proxy_repositories_created',
+            help: 'Proxy repositories created',
+        });
```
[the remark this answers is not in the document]
daveleek: I guess this one should be expected, but is also out of scope to address right now

src/lib/services/proxy-service.ts
daveleek: I didn't have it there initially, but I couldn't find example of code that didn't provide something. And I wondered if that had anything to do with snapshot suddenly failing. But I'll remove it

PR #6252, opened 2024-02-15
chore: add call incoming webhook rate limit metric

nunogois:
Follow up to https://github.com/Unleash/unleash/pull/6248 - Monitor this rate limit in metrics.

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -351,6 +351,12 @@ export default class MetricsMonitor {
                 rateLimits
                     .labels({ endpoint: '/auth/simple', method: 'POST' })
                     .set(config.rateLimiting.simpleLoginMaxPerMinute);
+                rateLimits
+                    .labels({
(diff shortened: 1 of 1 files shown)


src/lib/metrics.ts
```diff
@@ -351,6 +351,15 @@ export default class MetricsMonitor {
...
                     .labels({ endpoint: '/auth/simple', method: 'POST' })
                     .set(config.rateLimiting.simpleLoginMaxPerMinute);
+                rateLimits
+                    .labels({
+                        endpoint: '/api/incoming-webhook/:name',
+                        method: 'POST',
+                    })
+                    .set(
+                        config.rateLimiting.callIncomingWebhookMaxPerSecond *
+                            60,
```
nunogois, line 361: Current rate limit unit in metrics is "per minute", so we `* 60` the configured RPS value to convert to RPM.

gastonfournier (approved):
Looks good!

PR #6260, opened 2024-02-16
feat: project applications UI

sjaanus:
image


frontend/src/component/common/Table/cells/StringArrayCell.tsx
```diff
@@ -0,0 +1,58 @@
...
+    fontSize: theme.fontSizes.smallerBody,
+}));
+
+interface IArrayFieldCellProps<T> {
+    row: T;
+    field: keyof T;
+    singularLabel: string;
+}
+
+export const StringArrayCell: VFC<IArrayFieldCellProps<any>> = ({
```
sjaanus, line 18: A generic component to display list of strings in table.

src/lib/features/project/project-store.ts
```diff
@@ -801,6 +808,15 @@ class ProjectStore implements IProjectStore {
             }
         });
 
+        entriesMap.forEach((entry) => {
```
sjaanus, line 811: Add sorting

src/lib/features/project/project-store.ts
```diff
@@ -637,10 +637,17 @@ class ProjectStore implements IProjectStore {
...
+                applications,
+                total: Number(rows[0].total) || 0,
+            };
+        }
+
         return {
-            applications,
-            total: Number(rows[0].total) || 0,
+            applications: [],
+            total: 0,
```
thomasheartman, line 650: Interesting. Did this somehow not work before or are you just saving a DB call when we don't need it?

frontend/src/component/common/Table/cells/StringArrayCell.tsx
```diff
@@ -0,0 +1,58 @@
...
+    field,
+    singularLabel,
+}) => {
+    const { searchQuery } = useSearchHighlightContext();
+    const fieldValue = row[field];
+
+    if (!Array.isArray(fieldValue) || fieldValue.length === 0)
+        return <TextCell />;
+
+    const pluralLabel = `${singularLabel}s`;
```
thomasheartman, line 29: This is an edge case, I guess, but not all nouns pluralize by adding an "s" at the end. "Shrimp", "sheep", and "fish" are some examples. Additionally, some words change vowels (e.g. "hobby" becomes "hobbies"). It might be okay in our current system, but it might also be incorrect sometimes. I would suggest that you either have to supply a pluralized label too or that you *can* provide that label, and if you don't the system falls back to just adding the s.

frontend/src/component/common/Table/cells/StringArrayCell.tsx
thomasheartman, line 1: Does this one allow you to see the content of the cell (the strings) without using a mouse? That is, do I have access to all the same information using just my keyboard (or just a tapping finger on touch screens) as I do with the mouse?

frontend/src/component/project/ProjectApplications/ProjectApplications.tsx
```diff
@@ -0,0 +1,204 @@
...
+                            <ConditionallyRender
+                                condition={!isSmallScreen}
+                                show={
+                                    <>
+                                        <Search
+                                            placeholder='Search'
+                                            expandable
+                                            initialValue={query || ''}
+                                            onChange={setSearchValue}
+                                            id='globalFeatureToggles'
```
thomasheartman, line 152: Is this the right search id? I'm not familiar with the global feature toggle search, but it feels like a different feature from this.

frontend/src/hooks/api/getters/useProjectApplications/useProjectApplications.ts
```diff
@@ -0,0 +1,108 @@
...
+
+const getProjectApplicationsFetcher = (
+    projectId: string,
+    params: GetProjectApplicationsParams,
+) => {
+    const urlSearchParams = new URLSearchParams(
+        Array.from(
+            Object.entries(params)
+                .filter(([_, value]) => !!value)
+                .map(([key, value]) => [key, value.toString()]), // TODO: parsing non-string parameters
```
thomasheartman, line 91: regarding the todo-comment here: do we not need that for this one? why is it left as a todo?

frontend/src/component/project/Project/Project.tsx
```diff
@@ -314,6 +321,7 @@ export const Project = () => {
                 />
                 <Route path='settings/*' element={<ProjectSettings />} />
                 <Route path='metrics' element={<ProjectDoraMetrics />} />
+                <Route path='applications' element={<ProjectApplications />} />
```
thomasheartman, line 324: So it's available to everyone, but it's just hidden unless the flag is turned on, yeah?

frontend/src/hooks/api/getters/useProjectApplications/useProjectApplications.ts
```diff
@@ -0,0 +1,108 @@
...
+            mutate();
+        }, [mutate]);
+
+        const returnData = data || fallbackData;
+        return {
+            ...returnData,
+            loading: isLoading,
+            error,
+            refetch,
+            initialLoad: isLoading,
```
thomasheartman, line 74: Why are `loading` and `initialLoad` the same here? What are we using them for? Judging from this bit of the useSWR docs, [`isLoading` is true on initial load, when there is no value present.](https://swr.vercel.app/docs/advanced/understanding) From then on, `isLoading` is always false (unless you change the key), but `isValidating` gets set to true when it revalidates with the server.

With that in mind, do we need the `initialLoad` in the cache and all that?

thomasheartman (approved):
This looks like it does what it should do to me, so I'll give it an approval for that. However, I've got a number of comments that I'd expect to have answered or addressed before it's merged. I'm assuming that none of them will present an issue (hence the approval), but we'll see ☺️

src/lib/features/project/project-store.ts
sjaanus: If rows count is 0 it will fail, and making it more readable

frontend/src/component/common/Table/cells/StringArrayCell.tsx
sjaanus: Yes, I thought about that, but thought to fix it when we run into that case, but I can make it right now.

frontend/src/component/common/Table/cells/StringArrayCell.tsx
sjaanus: Should be, yes. It works same as all other components in Unleash.

frontend/src/component/project/ProjectApplications/ProjectApplications.tsx
sjaanus: Good catch

frontend/src/component/project/Project/Project.tsx
sjaanus: Yes

frontend/src/hooks/api/getters/useProjectApplications/useProjectApplications.ts
sjaanus: Fixed, initialLoad was not needed

PR #6267, opened 2024-02-19
feat: add tabs

thomasheartman:
This PR adds a new file "Application.tsx", which is analogous to the existing Project.tsx file in that it contains the base layout for the application page, as well as tabs pointing to sub pages. Currently, the overview tab uses a paragraph with some fallback text, while the connected instances table displays the instances table. 

I have mostly copied the existing ApplicationEdit component and used that as a base, assuming that we'll delete that component when we're done with this.

image

### DIFF
--- a/frontend/src/component/application/Application.tsx
+++ b/frontend/src/component/application/Application.tsx
@@ -29,6 +29,8 @@ import PermissionButton from 'component/common/PermissionButton/PermissionButton
 import { formatDateYMD } from 'utils/formatDate';
 import { formatUnknownError } from 'utils/formatUnknownError';
 import { useRequiredPathParam } from 'hooks/useRequiredPathParam';
+import { useUiFlag } from 'hooks/useUiFlag';
+import { ApplicationEdit } from './ApplicationEdit/ApplicationEdit';
 
 type Tab = {
     title: string;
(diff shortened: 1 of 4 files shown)


frontend/src/component/application/Application.tsx
thomasheartman, line 1: This is pretty much a direct port of the old application screen, except I have ripped out the content. So what remains is the header and the page layout. 

If the sdkReporting flag is turned off, we'll show the old screen instead.

frontend/src/component/menu/routes.ts
```diff
@@ -163,12 +163,13 @@ export const routes: IRoute[] = [
 
     // Applications
     {
-        path: '/applications/:name',
+        path: '/applications/:name/*',
```
thomasheartman, line 166: We need to render for all subpaths of applications/:name too.

frontend/src/component/application/Application.tsx
thomasheartman: I know the sketches for the header are slightly different than the existing design, but I suggest we do the updates for that in a separate PR.

sjaanus (approved):
Approving for now, we can change UX stuff later.

PR #6270, opened 2024-02-19
feat: Adding Project access requires same role 

chriswk:
In order to prevent users from being able to assign roles/permissions they don't have, this PR adds a check that the user performing the action either is Admin, Project owner or has the same role they are trying to grant/add.

This addAccess method is only used from Enterprise, so there will be a separate PR there, updating how we return the roles list for a user, so that our frontend can only present the roles a user is actually allowed to grant.

This adds the validation to the backend to ensure that even if the frontend thinks we're allowed to add any role to any user here, the backend can be smart enough to stop it.

We should still update frontend as well, so that it doesn't look like we can add roles we won't be allowed to.

### DIFF
--- a/frontend/src/component/project/ProjectAccess/ProjectAccessEditGroup/ProjectAccessEditGroup.tsx
+++ b/frontend/src/component/project/ProjectAccess/ProjectAccessEditGroup/ProjectAccessEditGroup.tsx
@@ -1,8 +1,10 @@
-import {ProjectAccessAssign} from '../ProjectAccessAssign/ProjectAccessAssign';
-import {useRequiredPathParam} from 'hooks/useRequiredPathParam';
-import useProjectAccess, {ENTITY_TYPE,} from 'hooks/api/getters/useProjectAccess/useProjectAccess';
-import {useAccess} from 'hooks/api/getters/useAccess/useAccess';
-import {useUserProjectRoles} from "../../../../hooks/api/getters/useUserProjectRoles/useUserProjectRoles";
+import { ProjectAccessAssign } from '../ProjectAccessAssign/ProjectAccessAssign';
+import { useRequiredPathParam } from 'hooks/useRequiredPathParam';
+import useProjectAccess, {
+    ENTITY_TYPE,
+} from 'hooks/api/getters/useProjectAccess/useProjectAccess';
+import { useAccess } from 'hooks/api/getters/useAccess/useAccess';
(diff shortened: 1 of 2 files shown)


src/lib/features/metrics/instance/metrics.test.ts
chriswk, line 1: This can be ignored. It just annoyed me that I apparently managed to forget to remove console.log statements, so I boyscouted.

src/lib/features/project/project-service.ts
```diff
@@ -708,30 +740,38 @@ export default class ProjectService {
...
+                                roleId,
+                                groupIds: groups,
+                                userIds: users,
+                            };
+                        }),
+                    },
+                }),
+            );
+        } else {
+            throw new InvalidOperationError(
```
sighphyre, line 771: Might be slightly cleaner as a guard clause

src/lib/routes/admin-api/user/user.ts
```diff
@@ -131,8 +132,61 @@ class UserController extends Controller {
...
+                200,
+                res,
+                rolesSchema.$id,
+                {
+                    version: 1,
+                    roles,
+                },
+            );
+        } else {
+        }
```
sighphyre, line 188: I feel like this wants some code in the else :P

src/lib/services/access-service.ts
```diff
@@ -215,6 +216,19 @@ export class AccessService {
...
     }
 
+    /**
+     * Returns all roles the user has in the project.
+     * Including roles via groups.
+     * In addition it includes root roles
+     * @param userId user to find roles for
+     * @param project project to find roles for
+     */
+    async getAllProjectRolesForUser(
```
sighphyre, line 226: Name still good? I read the function name as not returning root roles but the doc string suggests otherwise

sighphyre (approved):
Some very small details, but LGTM

src/lib/services/access-service.ts
chriswk: It does return root roles. The other name `getAllProjectRolesAndRootRolesForUser` was already taken but only returned numbers, might want to look into usage of that other function. For now, keeping this.

src/lib/routes/admin-api/user/user.ts
chriswk: Very true. Added `res.status(400).end()`

PR #6271, opened 2024-02-19
chore: avoid printing out warnings from known frontend proxies

gastonfournier:
## About the changes
Our frontend API creates new instances of unleash-client-proxy. Because this is by-design, we don't want to log a warning that was designed to warn users about potential misconfiguration of Unleash Proxy.

As an extra, I'm renaming ProxyController to FrontendAPIController to better reflect the intent of this controller.

### DIFF
--- a/src/lib/services/proxy-service.ts
+++ b/src/lib/services/proxy-service.ts
@@ -125,7 +125,7 @@ export class ProxyService {
         );
 
         const client = new Unleash({
-            appName: `frontend-proxy-${token.username ?? 'unknown'}`,
+            appName: 'proxy',
             url: 'unused',
             storageProvider: new InMemStorageProvider(),
             disableMetrics: true,
--- a/src/lib/services/proxy-service.ts
+++ b/src/lib/services/proxy-service.ts
@@ -125,12 +125,13 @@ export class ProxyService {
         );
 
         const client = new Unleash({
-            appName: 'proxy',
+            appName: `frontend-proxy-${token.username ?? 'unknown'}`,
             url: 'unused',
             storageProvider: new InMemStorageProvider(),
             disableMetrics: true,
             repository,
             disableAutoStart: true,
+            skipInstanceCountWarning: true,
         });
 
         client.on(UnleashEvents.Error, (error) => {
@@ -188,7 +189,7 @@ export class ProxyService {
         } catch (error) {
             this.logger.debug('Unable to fetch frontend settings', error);
         }
-        return this.cachedFrontendSettings;
+        return this.cachedFrontendSettings!;
     }
 
     async getFrontendSettings(useCache = true): Promise<FrontendSettings> {
--- a/src/lib/services/proxy-service.ts
+++ b/src/lib/services/proxy-service.ts
@@ -189,7 +189,7 @@ export class ProxyService {
         } catch (error) {
             this.logger.debug('Unable to fetch frontend settings', error);
         }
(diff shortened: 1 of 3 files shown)


src/lib/services/proxy-service.ts
```diff
@@ -188,7 +189,7 @@ export class ProxyService {
         } catch (error) {
             this.logger.debug('Unable to fetch frontend settings', error);
         }
-        return this.cachedFrontendSettings;
+        return this.cachedFrontendSettings!;
```
ivarconr, line 192: Why did you add not (!) here?

gastonfournier: It's just telling the compiler that this expression will not be null or undefined. I'm not sure that's entirely true, but we are assuming that fetchFrontendSettings always returns a value. In getUIConfig this line https://github.com/Unleash/unleash/blob/2553bf18e55c0730e997c33529fd815301b0fdd8/src/lib/routes/admin-api/config.ts#L169 would fail.

**Another option** is to change the signature of the method to be 
```typescript
async fetchFrontendSettings(): Promise<FrontendSettings | undefined>
```
and handle this case on the clients.

**Third option**: just leave the compiler error alone (note it's actually an error, but we're ignoring null-checks. This is one of those cases). I think this is a viable option because this line is unrelated to this PR change (just something I found while working on this file).

nunogois (approved):
I like this 👍

src/lib/services/proxy-service.ts
ivarconr: but returning `!undefined` will be "true" which in my book is not the same as `FrontendSettings`? Would it not be better to deal with the possible null/undefined and return a default value instead?

ivarconr: Another option: Not try to catch the error. We would have to control what the memoziee function should do then.

nunogois: Like Gastón said, this is a [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) in TS, which basically is "just telling the compiler that this expression will not be null or undefined". It's not the same as `!this.cachedFrontendSettings`.

I think it's good scouting since we're keeping the behavior we had previously. The difference is that we're "fixing" the TS error by assuring that we're passing a non-null value.

If we want to fix this for good I would look into why `cachedFrontendSettings` is possibly null/undefined and fix the types there, or possibly add a logic branch for when it is, like `if (!this.cachedFrontendSettings) throw new Error('No frontend settings found.')` before.

As it stands I'm happy with merging this as is, since, like mentioned before, "we're keeping the behavior we had previously."

FredrikOseberg: @ivarconr `!` in front of the expression is a javascript operater and will negate the value of the expression. `!` after the expression is way of telling the typescript compiler that the value will always exist: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

FredrikOseberg (approved):
LGTM. It would be fantastic if we could migrate this to a feature oriented architecture, and have all the folder and file names match what the name is controller. So we'd have:

* frontend-api-controller
* frontend-api-service

src/lib/services/proxy-service.ts
gastonfournier: Leaving this change out just to move this PR forward

PR #6291, opened 2024-02-21
chore: adds killswitch to scheduled per sec caches

daveleek:
## About the changes

Adds killswitches to update revision id and publish unpublished events


chriswk (approved):
Sure, let's try. it's not something we can do anywhere else, since this will break ETag caching, if you always return 0, our logic will think it never changes.

PR #6293, opened 2024-02-21
fix: adjust padding of problem detected icon

thomasheartman:
This PR adjusts the vertical position of the problem detected icon in the environment selector.

image


nicolaesocaciu (approved):
The position looks perfect
- maybe the formula is a bit to complex but i trust you

PR #6294, opened 2024-02-21
fix: generalize multi action button

FredrikOseberg:
This PR moves the CR specific logic out of the MultiActionButton and generalises so that we can re-use it across the application. The CR specific logic is moved into:

* ApplyButton.tsx
* ReviewButton.tsx

This fixes a bug where multi action button would be disabled if you tried to apply an approved change request that you had created yourself.

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ReviewButton/ReviewButton.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ReviewButton/ReviewButton.tsx
@@ -1,31 +1,49 @@
-import { FC } from 'react';
+import { FC, useContext } from 'react';
 
 import CheckBox from '@mui/icons-material/Check';
 import Clear from '@mui/icons-material/Clear';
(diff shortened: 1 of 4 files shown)


frontend/src/component/changeRequest/ChangeRequestOverview/ReviewButton/ReviewButton.tsx
```diff
@@ -1,31 +1,49 @@
...
-);
+}> = ({ disabled, onReject, onApprove, children }) => {
+    const { isAdmin } = useContext(AccessContext);
+    const projectId = useRequiredPathParam('projectId');
+    const id = useRequiredPathParam('id');
+    const { user } = useAuthUser();
+    const { data } = useChangeRequest(projectId, id);
+
+    const approverIsCreator = data?.createdBy.id === user?.id;
+    const canApprove = disabled || (approverIsCreator && !isAdmin);
```
kwasniew, line 24: should't it be canNotApprove? or disableApprove?

kwasniew (approved):
One comment about a potentially misleading name

PR #6298, opened 2024-02-21
chore: action_set_events db table migration

nunogois:
https://linear.app/unleash/issue/2-1962/implement-new-action-events-logic

Adds a new `action_set_events` table for the new action events logic.

PR #6300, opened 2024-02-21
chore: db drop action_states

nunogois:
Follow up to: https://github.com/Unleash/unleash/pull/6298

We no longer need this table, since it was superseded by `action_events` and is no longer used.
I believe it's safe to drop this table right away since the feature is not being used yet.


gastonfournier (approved):
Validate that the new code is working and go :rocket:

PR #6303, opened 2024-02-21
feat: application overview backend

sjaanus:


src/test/fixtures/fake-client-applications-store.ts
```diff
@@ -78,4 +79,8 @@ export default class FakeClientApplicationsStore
         await this.delete(details.appName);
         return this.bulkUpsert([details]);
     }
+
+    getApplicationOverview(appName: string): Promise<IApplicationOverview> {
```
kwasniew, line 83: if we had this implemented we could base our testing strategy on service level tests without db which are faster and write only 1 e2e tests that is slow

src/test/e2e/api/admin/applications.e2e.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        await app.services.apiTokenService.createApiTokenWithProjects({
+            type: ApiTokenType.CLIENT,
+            projects: ['default'],
+            environment: 'default',
+            tokenName: 'tester',
+        });
+});
+
+afterEach(async () => {
+    await db.stores.clientMetricsStoreV2.deleteAll();
```
kwasniew, line 76: I've never noticed this before but maybe Promise.all is possible here

src/lib/features/metrics/instance/instance-service.ts
```diff
@@ -212,6 +212,15 @@ export default class ClientInstanceService {
         };
     }
 
+    async getApplicationOverview(
+        appName: string,
+    ): Promise<IApplicationOverview> {
+        const application =
+            this.clientApplicationsStore.getApplicationOverview(appName);
```
kwasniew, line 219: I though it's missing await but it's a one liner. If we add extra line we may introduce a bug.
maybe return value without creating a variable?

kwasniew (approved):
legit. some minor comments inside

src/test/fixtures/fake-client-applications-store.ts
sjaanus: I feel like all of the complexity is in database query. If we are using fake store, we are not testing that complexity.

src/test/e2e/api/admin/applications.e2e.test.ts
sjaanus: I used promise all and the benefit is not existant. But in theory it will help, so I added this change.

PR #6307, opened 2024-02-22
feat: combine services enterprise

FredrikOseberg:
More detailed information can be found in the unleash-enterprise repository.


src/lib/app.ts
```diff
@@ -35,7 +35,9 @@ export default async function getApp(
     services: IUnleashServices,
     unleashSession?: RequestHandler,
     db?: Knex,
-): Promise<Application> {
+): Promise<{ app: Application; services: unknown; stores: unknown }> {
```
gastonfournier, line 38: Can we have these as parametric types? Something like `<EServices extends IUnleashServices, EStores extends IUnleashStores>`

```suggestion
): Promise<{ app: Application; services: EServices; stores: EStores }> {
```

src/lib/app.ts
```diff
@@ -179,7 +181,11 @@ export default async function getApp(
     );
 
     if (typeof config.preRouterHook === 'function') {
-        config.preRouterHook(app, config, services, stores, db);
+        const { services: enterpriseServices, stores: enterpriseStores } =
+            config.preRouterHook(app, config, services, stores, db);
+
+        combinedServices = enterpriseServices;
+        combinedStores = enterpriseStores;
```
gastonfournier, line 188: Do we need to do this? I think the "beauty" of the getApp method is that it just return an app, and the preHook just modifies it

If we have to do that, I wouldn't call them enterpriseServices, I'd do something like:
```typescript
const { services: extendedServices, stores: extendedStores } =
            config.preRouterHook(app, config, services, stores, db);
```

gastonfournier (commented):
I think this makes sense

PR #6309, opened 2024-02-22
feat: schema for paginated applications

kwasniew:
## About the changes

* Enhancing existing request schema to support query parameters
* Adding total field currently reflecting all features returned (will need to change after pagination)
* Making fields in response schema required (not sure why they were optional before)

Next PRs:
* generate orval types
* switch frontend to use the query params

### DIFF
--- a/src/lib/openapi/spec/applications-query-parameters.ts
+++ b/src/lib/openapi/spec/applications-query-parameters.ts
@@ -1,0 +1,57 @@
+import { FromQueryParams } from '../util/from-query-params';
+
+export const applicationsQueryParameters = [
+    {
+        name: 'query',
+        schema: {
+            type: 'string',
+            example: 'first_app',
+        },
+        description: 'The search query for the application name',
+        in: 'query',
+    },
+    {
+        name: 'offset',
+        schema: {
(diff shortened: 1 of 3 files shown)


src/lib/openapi/spec/applications-query-parameters.ts
```diff
@@ -0,0 +1,57 @@
...
+        in: 'query',
+    },
+    {
+        name: 'limit',
+        schema: {
+            type: 'string',
+            example: '50',
+        },
+        description:
+            'The number of applications to return in a page. By default it is set to 1000.',
```
kwasniew, line 30: 1000 is important here

src/lib/openapi/spec/applications-schema.ts
```diff
@@ -7,8 +7,14 @@ export const applicationsSchema = {
     additionalProperties: false,
     description:
         'An object containing a list of applications that have connected to Unleash via an SDK.',
+    required: ['total', 'applications'],
```
kwasniew, line 10: making it more strict so non-breaking change

sjaanus (approved):

PR #6311, opened 2024-02-22
chore: remove redundant check for project names in db

daveleek:
## About the changes

skips the fetching project names from db part in the proxy-repository that was done when the token had ALL_PROJECTS


chriswk (approved):
LGTM

chriswk (approved):
I like it :)

PR #6312, opened 2024-02-22
feat: project applications paging backend

sjaanus:


src/lib/routes/admin-api/metrics.ts
```diff
@@ -188,15 +188,31 @@ class MetricsController extends Controller {
...
+        const normalizedLimit =
+            Number(limit) > 0 && Number(limit) <= 1000 ? Number(limit) : 1000;
+        const normalizedOffset = Number(offset) > 0 ? Number(offset) : 0;
+        const normalizedSortBy: string = sortBy ? sortBy : 'appName';
+        const normalizedSortOrder =
+            sortOrder === 'asc' || sortOrder === 'desc' ? sortOrder : 'asc';
+
         const applications = await this.clientInstanceService.getApplications(
-            query,
+            {
```
kwasniew, line 206: I'd extract the normalization as a method that returns this object. As a result the main controller method is not concerned with low level details

kwasniew (approved):

PR #6317, opened 2024-02-22
chore: incoming webhook events UI

nunogois:
https://linear.app/unleash/issue/2-1937/incoming-webhook-events-ui

This PR implements the UI for incoming webhook events.

We're also introducing a new `SidePanelList` component that we'll be able to reuse when we tackle action set events. This PR also promotes `ReactJSONEditor` to a common component and adapts it slightly for this use case.

image

image

We're refreshing the events view every 5s, so if you're monitoring events for a specific incoming webhook you can see the latest ones coming in.
We load 20 (configurable through the hook) events by default. Everytime you reach the end of the list you can load 20 more events until you reach the end of the event list.

image

### DIFF
--- a/frontend/src/component/common/ReactJSONEditor/ReactJSONEditor.tsx
+++ b/frontend/src/component/common/ReactJSONEditor/ReactJSONEditor.tsx
@@ -4,7 +4,11 @@ import 'vanilla-jsoneditor/themes/jse-theme-dark.css';
 import { styled } from '@mui/material';
 import UIContext from 'contexts/UIContext';
 
-const JSONEditorThemeWrapper = styled('div')(({ theme }) => ({
+type EditorStyle = 'default' | 'sidePanel';
+
+const JSONEditorThemeWrapper = styled('div', {
+    shouldForwardProp: (prop) => prop !== 'editorStyle',
+})<{ editorStyle?: EditorStyle }>(({ theme, editorStyle = 'default' }) => ({
     '&.jse-theme-dark': {
         '--jse-background-color': theme.palette.background.default,
         '--jse-panel-background': theme.palette.background.default,
@@ -24,9 +28,34 @@ const JSONEditorThemeWrapper = styled('div')(({ theme }) => ({
         borderBottomLeftRadius: theme.shape.borderRadius,
         borderBottomRightRadius: theme.shape.borderRadius,
     },
+    ...(editorStyle === 'sidePanel' && {
+        '&&&': {
+            '--jse-main-border': 0,
+            '& > div': {
+                height: '100%',
+            },
+            '& .jse-focus': {
+                '--jse-main-border': 0,
+            },
+            '& .cm-gutters': {
+                '--jse-panel-background': 'transparent',
+                '--jse-panel-border': 'transparent',
+            },
+            '& .cm-gutter-lint': {
+                width: 0,
+            },
+            '& .jse-text-mode': {
+                borderBottomRightRadius: theme.shape.borderRadiusMedium,
+            },
+        },
+    }),
 }));
 
-const VanillaJSONEditor: React.FC<JSONEditorPropsOptional> = (props) => {
+interface IReactJSONEditorProps extends JSONEditorPropsOptional {
+    editorStyle?: EditorStyle;
+}
+
+const VanillaJSONEditor: React.FC<IReactJSONEditorProps> = (props) => {
     const refContainer = useRef<HTMLDivElement | null>(null);
     const refEditor = useRef<JSONEditor | null>(null);
 
@@ -58,11 +87,12 @@ const VanillaJSONEditor: React.FC<JSONEditorPropsOptional> = (props) => {
     return <div ref={refContainer} />;
 };
 
-const ReactJSONEditor: React.FC<JSONEditorPropsOptional> = (props) => {
+const ReactJSONEditor: React.FC<IReactJSONEditorProps> = (props) => {
     const { themeMode } = useContext(UIContext);
     return (
         <JSONEditorThemeWrapper
             className={themeMode === 'dark' ? 'jse-theme-dark' : ''}
+            editorStyle={props.editorStyle}
         >
             <VanillaJSONEditor
                 mainMenuBar={false}
--- a/frontend/src/component/common/ReactJSONEditor/ReactJSONEditor.tsx
+++ b/frontend/src/component/common/ReactJSONEditor/ReactJSONEditor.tsx
@@ -30,6 +30,9 @@ const JSONEditorThemeWrapper = styled('div', {
     },
     ...(editorStyle === 'sidePanel' && {
         '&&&': {
+            '& .jse-main': {
+                minHeight: 0,
+            },
             '--jse-main-border': 0,
             '& > div': {
                 height: '100%',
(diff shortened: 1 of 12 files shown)


frontend/src/component/common/ReactJSONEditor/ReactJSONEditor.tsx
```diff
@@ -24,9 +28,40 @@ const JSONEditorThemeWrapper = styled('div')(({ theme }) => ({
...
+                width: 0,
+            },
+            '& .jse-text-mode': {
+                borderBottomRightRadius: theme.shape.borderRadiusMedium,
+            },
+            '& .cm-scroller': {
+                '--jse-delimiter-color': theme.palette.text.primary,
+            },
+        },
+    }),
```
daveleek, line 57: This felt a tad noisy, but I guess this is where it belongs

frontend/src/component/incomingWebhooks/IncomingWebhooksEvents/IncomingWebhooksEventsModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+import FormTemplate from 'component/common/FormTemplate/FormTemplate';
+import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
+import { SidePanelList } from 'component/common/SidePanelList/SidePanelList';
+import { formatDateYMDHMS } from 'utils/formatDate';
+import { useLocationSettings } from 'hooks/useLocationSettings';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
+
+const LazyReactJSONEditor = lazy(
+    () => import('component/common/ReactJSONEditor/ReactJSONEditor'),
+);
```
daveleek, line 15: Could the component be exported this way? so it's always lazily imported? Or is that an implementation detail more than what we should do every time?

frontend/src/component/incomingWebhooks/IncomingWebhooksModal/IncomingWebhooksModal.tsx
```diff
@@ -137,12 +151,16 @@ export const IncomingWebhooksModal = ({
             <FormTemplate
                 loading={loading}
                 modal
-                title={title}
+                title=''
```
daveleek, line 154: Could we make this property/parameter optional? Or did you want to show that it's intentionally omitted?

frontend/src/component/incomingWebhooks/IncomingWebhooksEvents/IncomingWebhooksEventsModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+    onOpenConfiguration,
+}: IIncomingWebhooksEventsModalProps) => {
+    const { uiConfig } = useUiConfig();
+    const { locationSettings } = useLocationSettings();
+    const { incomingWebhookEvents, hasMore, loadMore, loading } =
+        useIncomingWebhookEvents(incomingWebhook?.id, 20, {
+            refreshInterval: 5000,
+        });
+
+    if (!incomingWebhook) {
```
daveleek, line 78: under which circumstances would this occur?

frontend/src/hooks/api/getters/useIncomingWebhookEvents/useIncomingWebhookEvents.ts
```diff
@@ -0,0 +1,79 @@
...
+) => {
+    const { isEnterprise } = useUiConfig();
+    const incomingWebhooksEnabled = useUiFlag('incomingWebhooks');
+
+    const getKey: SWRInfiniteKeyLoader = (
+        pageIndex: number,
+        previousPageData: IncomingWebhookEventsResponse,
+    ) => {
+        // Does not meet conditions
+        if (!incomingWebhookId || !isEnterprise || !incomingWebhooksEnabled)
```
daveleek, line 36: `!incomingWebhookId`
Is this a UI that can be opened without having an existing inc.webhook?

daveleek (approved):
Couple of questions around things like potentially undefined parameters meaning we return null etc. But apart from that LGTM!

frontend/src/component/incomingWebhooks/IncomingWebhooksEvents/IncomingWebhooksEventsModal.tsx
nunogois: If I understand correctly you're asking whether we can export the lazy import, so that it is always lazy imported. I think it should always be up to the parent to decide how to import the component, but we can discuss this in the team.

daveleek: Yes exactly. I saw it being lazily imported twice in the same PR, so figured it was worth asking if this should be the standard way of doing this

frontend/src/component/incomingWebhooks/IncomingWebhooksModal/IncomingWebhooksModal.tsx
nunogois: Seems like it's already optional so we can just omit it. Nice catch! https://github.com/Unleash/unleash/pull/6317/commits/f35add8c1545efc11cba516126b63b8cd956f581

frontend/src/hooks/api/getters/useIncomingWebhookEvents/useIncomingWebhookEvents.ts
nunogois: I think this is related to: https://github.com/Unleash/unleash/pull/6317#discussion_r1500470878

The behavior we want for this hook is that you can call it whenever, but if you're not passing an `incomingWebhookId`, it should not fetch data.

Since you cannot conditionally call hooks in React, this matches the previous comment well, because you can just call this hook every time and it will gracefully handle undefined cases.

frontend/src/component/incomingWebhooks/IncomingWebhooksEvents/IncomingWebhooksEventsModal.tsx
daveleek: Cool, thanks for the explanation!

daveleek: What would the drawbacks of conditionally rendering the Modal be?

nunogois: That could also work, but if we're using `ConditionallyRender`, the component should be returned in a callback, so something like this:

```tsx
<ConditionallyRender
    condition={Boolean(selectedIncomingWebhook)}
    show={() => (
        <IncomingWebhooksModal
            incomingWebhook={selectedIncomingWebhook!}
            open={modalOpen}
            setOpen={setModalOpen}
            newToken={(token: string) => {
                setNewToken(token);
                setTokenDialog(true);
            }}
            onOpenEvents={() => {
                setModalOpen(false);
                setEventsModalOpen(true);
            }}
        />
    )}
/>
```

TypeScript also doesn't know that `show` only happens when the condition is `true`, so you also need to add the `!` in `selectedIncomingWebhook!`. But if you do this, I believe it's safe to assume that the `incomingWebhook` prop is no longer optional since it's always going to be present.

While this could be a valid option and sometimes easier, the main drawbacks for me at this stage are noisy code at the parent level and less consistency with the rest of our code base.

Effectively, what we have now, checking for `if (!incomingWebhook) return null` inside the component, is the same as a conditional render, but shifting the responsibility of handling this gracefully to the component itself and not the parent.

PR #6318, opened 2024-02-22
fix: only show visible environments for that project in actions

nunogois:
https://linear.app/unleash/issue/2-1969/actions-form-only-show-environments-scoped-to-the-current-project

Filters the environment options in actions to only show environments that are currently visible for that project, safeguarding users from configuring actions for invalid environments.

image


daveleek (approved):
LGTM!

PR #6320, opened 2024-02-22
chore: remove mermaid in favor of react archer

nunogois:
https://linear.app/unleash/issue/2-1968/use-react-archer-instead-of-mermaid-in-the-network-overview

This PR removes [mermaid](https://github.com/mermaid-js/mermaid) in favor of [react-archer](https://github.com/pierpo/react-archer), which meant adapting `NetworkOverview` to use `react-archer` instead, since that was the only remaining feature using `mermaid`.

Opted not to use RPS labels in the lines since it was very noisy visually. Took the liberty to make some visual changes, but we can align afterwards with @nicolaesocaciu if needed.

### Example with 1 app

image

### Example with 3 apps

image

### Example with 30 apps

image

### Example in light theme

image


frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
```diff
@@ -84,37 +115,71 @@ const toGraphData = (metrics?: RequestsPerSecondSchema) => {
...
-                        )}") -- ${reqs} req/s<br>${type} --> Unleash`,
-                )
-                .join('\n')}
-        end
-    `;
+    if (apps.length === 0) {
+        return <Alert severity='warning'>No data available.</Alert>;
+    }
+
+    const mockApps = new Array(30).fill(1).map((_, i) => ({
```
kwasniew, line 128: is the mock apps intended?

kwasniew (approved):
I am a big fan of what you did here

frontend/src/component/admin/network/NetworkOverview/NetworkOverview.tsx
nunogois: Oops, you're right. I thought I had reverted this but it slipped by. Nice catch! Cleaned up in https://github.com/Unleash/unleash/pull/6320/commits/b2221f7d96b959b9902a5687750b49a2c1696cc4

PR #6323, opened 2024-02-22
feat: capped at limit strategy in application usage

kwasniew:
## About the changes

Info when strategy in application usage exceeds 1000. When apps < 1000 then we don't show this info.
Screenshot 2024-02-22 at 20 20 28

### DIFF
--- a/frontend/src/component/strategies/StrategyView/StrategyDetails/StrategyDetails.tsx
+++ b/frontend/src/component/strategies/StrategyView/StrategyDetails/StrategyDetails.tsx
@@ -84,7 +84,7 @@ export const StrategyDetails = ({
                 </Grid>
 
                 <Grid item sm={12} md={toggles.length > 0 ? 6 : 12}>
-                    <h6>Applications using this strategy</h6>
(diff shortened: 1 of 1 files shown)


frontend/src/component/strategies/StrategyView/StrategyDetails/StrategyDetails.tsx
```diff
@@ -84,7 +84,7 @@ export const StrategyDetails = ({
                 </Grid>
 
                 <Grid item sm={12} md={toggles.length > 0 ? 6 : 12}>
-                    <h6>Applications using this strategy</h6>
+                    <h6>Applications using this strategy {applications.length >= 1000 && '(Capped at 1000)'}</h6>
```
sjaanus, line 87: Is equal needed?

sjaanus (approved):
LG with comment

frontend/src/component/strategies/StrategyView/StrategyDetails/StrategyDetails.tsx
kwasniew: we get 1000 max so we don't know if there's more. So If I get max then I'm assuming there may be more and add the info

PR #6330, opened 2024-02-23
feat: application missing features backend

sjaanus:
This PR adds a property issues to application schema, and also adds all the missing features that have been reported by SDK, but do not exist in Unleash.

### DIFF
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -305,7 +305,6 @@ export default class ClientApplicationsStore
                 );
             })
             .where('a.app_name', appName);
-
         const rows = await query;
         if (!rows.length) {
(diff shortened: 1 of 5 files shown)


src/lib/db/client-applications-store.ts
```diff
@@ -316,22 +316,41 @@ export default class ClientApplicationsStore
...
+            if (!project && feature_name) {
+                missingFeatures.add(feature_name);
+            }
+
             let env = acc.find((e) => e.name === environment);
             if (!env) {
                 env = {
                     name: environment,
-                    instanceCount: 1,
+                    instanceCount: instance_id ? 1 : 0,
```
sjaanus, line 341: Sometimes instance can be missing, so it should not be 1 by default

kwasniew (approved):

PR #6331, opened 2024-02-23
fix: better arrow design in network overview

nunogois:
Improves the arrow design in the "Network overview" after aligning with @nicolaesocaciu


nicolaesocaciu (approved):
Perfect!

PR #6333, opened 2024-02-26
refactor: paginated data hook

kwasniew:
## About the changes

Extracted pagination to a common hook so that user code doesn't have to remember:
* to pass query params
* to clear the SWR cache
* to propagate total and data separately

This hook is now used by project applications and applications.

There's still useFeatureSearch that hasn't been migrated since it adds some total caching. It can be done separately by creating a new hook that composes the pagination hook with a caching behavior.


sjaanus (approved):
LG

PR #6334, opened 2024-02-26
feat: application missing strategies

sjaanus:
Now also showing missing strategies, that SDK sends, but do not exist in Unleash.

### DIFF
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -27,6 +27,12 @@ const TABLE = 'client_applications';
 
 const TABLE_USAGE = 'client_applications_usage';
 
+const DEPRECATED_STRATEGIES = [
+    'gradualRolloutRandom',
+    'gradualRolloutSessionId',
+    'gradualRolloutUserId',
+];
+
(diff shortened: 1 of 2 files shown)


src/lib/db/client-applications-store.ts
```diff
@@ -334,6 +350,15 @@ export default class ClientApplicationsStore
                 missingFeatures.add(feature_name);
             }
 
+            strategies.forEach((strategy) => {
+                if (
+                    !DEPRECATED_STRATEGIES.includes(strategy) &&
```
sjaanus, line 355: SDKs need to backwards compatible, so they are sending some very old strategy types that have been removed from newer versions of Unelash. These should not be reported as missing types.

src/lib/db/client-applications-store.ts
```diff
@@ -294,6 +300,7 @@ export default class ClientApplicationsStore
                 'ci.instance_id',
                 'ci.sdk_version',
                 'ci.last_seen',
+                'strategies',
```
kwasniew, line 303: we had a bug due to missing explicit table name. I'd add one here

PR #6335, opened 2024-02-26
chore: remove new strategy configuration flag

kwasniew:
## About the changes

Removing flag responsible for new strategy configuration.

Next steps:
* find components with the New prefix that are replacing components that are no longer used and remove the old components/rename new ones (next PR)
* find components with the New prefix that are replacing components that are still used and require call site change (next week)


sjaanus (approved):
LG

PR #6337, opened 2024-02-26
fix: track conflicts in new strategy screen

andreas-unleash:
Fixes a bug where the conflict tracking was only in the old feature strategy edit screen. 
- Ports the conflict tracking to the NewFeatureStrategyEdit screen

Closes # [1-2093](https://linear.app/unleash/issue/1-2093/cr-conflict-detection-in-new-strategy-edit-screen)


kwasniew (approved):
Thanks!

PR #6342, opened 2024-02-26
feat: simplified insights data fetching

Tymek:
Simplify insights data fetching - let charts pick data instead of preparing labels in hook.

This also **allows access other data-point related properties for use in a tooltip**

### DIFF
--- a/frontend/src/component/executiveDashboard/LineChart/LineChartComponent.tsx
+++ b/frontend/src/component/executiveDashboard/LineChart/LineChartComponent.tsx
@@ -1,4 +1,10 @@
-import { type ReactNode, useMemo, useState, type VFC } from 'react';
+import {
+    type ReactNode,
+    useMemo,
+    useState,
+    type VFC,
+    ComponentProps,
+} from 'react';
 import {
     CategoryScale,
     LinearScale,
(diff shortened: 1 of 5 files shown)


frontend/src/component/executiveDashboard/LineChart/LineChartComponent.tsx
```diff
@@ -223,28 +228,38 @@ const customHighlightPlugin = {
 };
 
 const LineChartComponent: VFC<{
-    data: ChartData<'line', (number | ScatterDataPoint | null)[], unknown>;
+    data: ChartData<any, unknown>;
```
kwasniew, line 231: can we do it w/o loosing type information?

frontend/src/component/executiveDashboard/LineChart/LineChartComponent.tsx
```diff
@@ -223,28 +228,38 @@ const customHighlightPlugin = {
...
     aspectRatio?: number;
     cover?: ReactNode;
     isLocalTooltip?: boolean;
+    overrideOptions?: ComponentProps<typeof Line>['options'];
```
kwasniew, line 235: isn't there a named/explicit type in chartjs for that so that we don't have to advanced type level operation?

PR #6344, opened 2024-02-26
feat: allow api token middleware to fetch from db

gastonfournier:
## About the changes
When edge is configured to automatically generate tokens, it requires the token to be present in all unleash instances. 
It's behind a flag which enables us to turn it on on a case by case scenario.

The risk of this implementation is that we'd be adding load to the database in the middleware that evaluates tokens (which are present in mostly all our API calls. We only query when the token is missing but because the /client and /frontend endpoints which will be the affected ones are high throughput, we want to be extra careful to avoid DDoSing ourselves

## Alternatives:
One alternative would be that we merge the two endpoints into one. Currently, Edge does the following:
If the token is not valid, it tries to create a token using a service account token and /api/admin/create-token endpoint. Then it uses the token generated (which is returned from the prior endpoint) to query /api/frontend. What if we could call /api/frontend with the same service account we use to create the token? It may sound risky but if the same application holding the service account token with permission to create a token, can call /api/frontend via the generated token, shouldn't it be able to call the endpoint directly?

The purpose of the token is authentication and authorization. With the two tokens we are authenticating the same app with 2 different authorization scopes, but because it's the same app we are authenticating, can't we just use one token and assume that the app has both scopes?

If the service account already has permissions to create a token and then use that token for further actions, allowing it to directly call /api/frontend does not necessarily introduce new security risks. The only risk is allowing the app to generate new tokens. Which leads to the third alternative: should we just remove this option from edge?


gastonfournier:
> Worth arguing that if you have an SA token, maybe you should be allowed to query client features endpoint.
> 
> We'd need to define project/environment, maybe just let edge query project: * and the environment it needs, thus reducing the number of caches edge will need to keep.

Indeed, I think this is key. We recently discussed about users being able to grant permissions they don't have and how we are moving away from that. In the same theme, we are allowing a service account, with permissions to create a token, grant permissions to pull from /client/features. We can argue that then the service account should have that permission to be able to grant it, and then if it already has the permission why do we need another token?

I wonder how different is that from just adding to edge the tokens you need. 

In any case, this PR would allow us to mitigate the problem without having to upgrade edge, but I do consider this just that: a mitigation.

gastonfournier:
A few things we want to include:
1. Limit this to /client/features
2. Try to identify if the request comes from edge

website/docs/reference/front-end-api.md
```diff
@@ -42,7 +42,7 @@ The client needs to point to the correct API endpoint. The front-end API is avai
 
 ### API token
 
-You can create appropriate token, with type `FRONTEND` on `<YOUR_UNLEASH_URL>/admin/api/create-token` page or with a request to `/api/admin/api-tokens`. See our guide on [how to create API tokens](../how-to/how-to-create-api-tokens.mdx) for more details.
+You can create appropriate token, with type `FRONTEND` on `<YOUR_UNLEASH_URL>/api/admin/create-token` page or with a request to `/api/admin/api-tokens`. See our guide on [how to create API tokens](../how-to/how-to-create-api-tokens.mdx) for more details.
```
gastonfournier, line 45: This is something I've spotted while reading the docs

gastonfournier:
> A few things we want to include:
> 
>     1. Limit this to /client/features
> 
>     2. Try to identify if the request comes from edge

These are two options to limit the scope of this PR: https://github.com/Unleash/unleash/pull/6354 and https://github.com/Unleash/unleash/pull/6356

src/lib/services/api-token-service.ts
```diff
@@ -160,7 +160,9 @@ export class ApiTokenService {
         }
     }
 
-    public getUserForToken(secret: string): IApiUser | undefined {
+    public async getUserForToken(
+        secret: string,
+    ): Promise<IApiUser | undefined> {
```
sighphyre, line 165: We're marking this async, doesn't this want resolves on its non promise return values now?

gastonfournier: Not sure I understand your question. All async methods needs to return a Promise as far as I know, which is why now we have to do awaits every time we call this method. E.g.: https://github.com/Unleash/unleash/blob/654a880b4f1584432bc3a9a553c711d67d87ef85/src/lib/routes/auth/reset-password-controller.ts#L167

gastonfournier:
> Just verified that all requests from Edge comes with its user-agent header set to `unleash-edge-<edge.version>` as well as the `UNLEASH-APPNAME` header set to `unleash-edge`

Awesome, I've added https://github.com/Unleash/unleash/pull/6354/commits/93b1253038be6621ab3af98140cca04fd6fa831b so we can toggle on app name as well. The code on how to get the app name was copied from https://github.com/Unleash/unleash/blob/9101c39eb790754490acb0dba5f0d3900bc3b788/src/lib/middleware/response-time-metrics.ts#L27

src/lib/services/api-token-service.ts
sighphyre: You're absolutely right, I forget that JS has this syntactic cancer

chriswk (approved):
Nice. I think this should work

PR #6356, opened 2024-02-27
Api token consistency scope to edge

gastonfournier:
## About the changes
Alternative to https://github.com/Unleash/unleash/pull/6354 this is more specific and less flexible. The reason why this is valid is because we don't want this to become a feature. A least not because of the problem with edge, as we **may** have documentation explaining that tokens are eventually consistent.


gastonfournier:
Leaning towards https://github.com/Unleash/unleash/pull/6354

PR #6358, opened 2024-02-27
chore: action events UI

nunogois:
https://linear.app/unleash/issue/2-1936/action-events-ui

Implements the UI for action events.

image

image

### DIFF
--- a/frontend/src/component/common/SidePanelList/SidePanelList.tsx
+++ b/frontend/src/component/common/SidePanelList/SidePanelList.tsx
@@ -21,13 +21,14 @@ const StyledSidePanelHalf = styled('div')({
 });
 
 const StyledSidePanelHalfLeft = styled(StyledSidePanelHalf, {
-    shouldForwardProp: (prop) => prop !== 'height',
-})<{ height?: number }>(({ theme, height }) => ({
+    shouldForwardProp: (prop) => prop !== 'height' && prop !== 'maxWidth',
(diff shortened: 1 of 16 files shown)


frontend/src/component/common/SidePanelList/SidePanelList.tsx
```diff
@@ -43,12 +44,14 @@ export const StyledSidePanelListColumn = styled('div', {
...
 })<{ maxWidth?: number; align?: ColumnAlignment }>(
     ({ theme, maxWidth, align = 'start' }) => ({
+        display: 'flex',
         flex: 1,
         padding: theme.spacing(2),
         fontSize: theme.fontSizes.smallBody,
         justifyContent: align,
         ...(maxWidth && { maxWidth }),
         textAlign: align,
+        alignItems: 'center',
```
gastonfournier, line 54: Will this and the display change above affect other UI components? I'm just worried this might cause unnoticed changes a few days prior a release

gastonfournier (approved):
I've seen this working and looks nice, my only concern is the changes on common components that might cause side effects. Other than that I'd iterate and improve based on this

frontend/src/component/common/SidePanelList/SidePanelList.tsx
nunogois: `SidePanelList` is a new component introduced in https://github.com/Unleash/unleash/pull/6317 so I wouldn't worry about it.

The only other place where we're using it, Incoming webhook events, also looks correct with this change.

PR #6359, opened 2024-02-27
feat: improved health chart tooltip

Tymek:
image

Also: refactored element common with user stats
image


andreas-unleash (approved):
Very nice!

PR #6361, opened 2024-02-27
chore: add a flag for scim

sighphyre:
Adds a flag for SCIM so we can build incrementally without breaking other things


chriswk (approved):
SCIM FTW

PR #6363, opened 2024-02-27
task: banner now displayed based on if flag is enabled

chriswk:
So, since our assumption about client instances ended up being wrong; this PR moves the EdgeUpgradeBanner to be displayed if the featureflag edgeBulkMetrics is enabled. That way, if customers comes back and says they have upgraded but still get the banner, we can remove them from the segment.


src/lib/routes/admin-api/config.ts
```diff
@@ -153,7 +139,8 @@ class ConfigController extends Controller {
         const flags = {
             ...this.config.ui.flags,
             ...expFlags,
-            displayUpgradeEdgeBanner: usesOldEdge,
+            displayUpgradeEdgeBanner:
+                this.flagResolver.isEnabled('edgeBulkMetrics'),
```
gastonfournier, line 143: Should this just be:
```suggestion
                this.flagResolver.isEnabled('upgradeEdgeBanner'),
```

And can't we just use the existing banner functionality?

ivarconr: I was thinking the same thing as Gaston. We should leave this as is (mostly for OSS users).

chriswk: Hmm, for OSS we've already disabled the endpoint, so old edge won't have given you metrics, but it should give the banner. But after the last couple of days with testing the usesSdkOlderThan doesn't seem to work as well in practice as it did in our testing.

With regards to using existing banner functionality, the assignment back in 5.9 was to write a specific banner component so we could show it to OSS users/self-hosted as well.

gastonfournier: > But after the last couple of days with testing the usesSdkOlderThan doesn't seem to work as well in practice as it did in our testing.

I experienced the banner in enterprise while playing around with edge about the token creation. It does have an odd behavior but it shows up. Maybe we can look at this together.

> With regards to using existing banner functionality, the assignment back in 5.9 was to write a specific banner component so we could show it to OSS users/self-hosted as well.

I assume now we are changing to showing this to only hosted instances (not OSS nor self-hosted). At least, that's what I understand from this PR. But maybe it's 2 fold: we keep the existing functionality for OSS and self-hosted and we have a toggle for hosted, but if we're doing that for hosted, we can use the existing banner functionality with the flag: messageBanner

chriswk: Yeah, reverted the change that removed the usage of checking for client instances. Now the PR literally is just adding the possibility to toggle the banner on explicitly through the `displayEdgeBanner` flag

src/lib/types/experimental.ts
```diff
@@ -252,6 +253,10 @@ const flags: IFlags = {
         process.env.UNLEASH_EXPERIMENTAL_SCIM_API,
         false,
     ),
+    displayEdgeBanner: parseEnvVarBoolean(
+        process.env.UNLEASH_EXPERIMENTAL_DISABLE_SCHEDULED_CACHES,
```
gastonfournier, line 257:
```suggestion
        process.env.UNLEASH_EXPERIMENTAL_DISPLAY_EDGE_BANNER,
```

although I don't think we need the env variable for this feature in particular

gastonfournier (changes requested):
Looks good, just a copy paste mistake that I'm not sure it's needed but in both cases we need to change it ;)

gastonfournier (approved):
:+1:

PR #6371, opened 2024-02-28
feat: application overview ux improvements

sjaanus:
1. Added navigation from environments to instances
2. Last seen is now shown as TimeAgo
3. Added icons for total environments and features
4. Fixed schema

image


src/lib/openapi/spec/application-overview-environment-schema.ts
```diff
@@ -28,6 +28,7 @@ export const applicationOverviewEnvironmentSchema = {
         },
         lastSeen: {
             type: 'string',
+            nullable: true,
```
kwasniew, line 31: how do we decide on undefined vs nullable in our schemas?

frontend/src/component/application/ApplicationChart.tsx
```diff
@@ -163,6 +181,18 @@ export const ApplicationChart = ({ data }: IApplicationChartProps) => {
                             >
                                 {applicationName}
                             </Typography>
+                            <StyledIconRow>
```
kwasniew, line 184: would it be more readable if we listened to codescene and avoided large function/large JSX which is also functions?

PR #6380, opened 2024-02-28
fix: setRolesForUser and setRolesForGroup role check

chriswk:
In order to stop privilege escalation via `/api/admin/projects/:project/users/:userId/roles` and `/api/admin/projects/:project/groups/:groupId/roles` this PR adds the same check we added to setAccess methods to the methods updating access for these two methods.

Also adds tests that verify that we throw an exception if you try to assign roles you do not have.  

Thank you @nunogois for spotting this during testing.

### DIFF
--- a/src/test/e2e/services/project-service.e2e.test.ts
+++ b/src/test/e2e/services/project-service.e2e.test.ts
@@ -621,6 +621,71 @@ describe('Managing Project access', () => {
             ),
         ).resolves.not.toThrow();
     });
+    test('Users can not assign roles they do not have to a user through explicit roles endpoint', async () => {
+        const project = {
+            id: 'user_fail_assign_to_user',
+            name: 'user_fail_assign_to_user',
+            description: '',
+            mode: 'open' as const,
+            defaultStickiness: 'clientId',
+        };
+        await projectService.createProject(project, user);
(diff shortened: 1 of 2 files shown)


src/test/e2e/services/project-service.e2e.test.ts
```diff
@@ -621,6 +621,71 @@ describe('Managing Project access', () => {
...
+        });
+        expect(
+            projectService.setRolesForGroup(
+                project.id,
+                group.id,
+                [customRole.id],
+                projectUser.username,
+                projectUser.id,
+            ),
+        ).rejects.toThrow();
```
nunogois, line 687: I think I would feel a bit more confident if we specifically check for the error we're throwing. Otherwise, if the methods 500 for example, these tests still pass, right?
```suggestion
        ).rejects.toThrow(new InvalidOperationError('User tried to assign a role they did not have access to'));
```

nunogois (approved):
Non-blocking approval 👍

src/test/e2e/services/project-service.e2e.test.ts
chriswk: Agreed. Will update. Thanks.

PR #6388, opened 2024-02-29
fix: appliction was rendered too early

sjaanus:


frontend/src/component/application/ApplicationEdit/ApplicationEdit.tsx
```diff
@@ -75,7 +76,7 @@ export const ApplicationEdit = () => {
     const tabData = [
         {
             label: 'Application overview',
-            component: <ApplicationEdit />,
+            component: <ApplicationView />,
```
kwasniew, line 79: I understand the other change but I'm not sure about this one

kwasniew (approved):

PR #6392, opened 2024-02-29
Fix/segments can not be deleted when archived

FredrikOseberg:
This PR makes a change to the logic of segments: 

(1) You can now delete segments that are being referenced in strategies on archived toggles
(2) We will no longer show "used in project" and "used in feature" when the toggle referencing it is archived

Skjermbilde 2024-02-29 kl  13 58 55

Also added a warning when deleting: 
Skjermbilde 2024-02-29 kl  13 59 21

Discussion points: 
* I have added tests to verify the behavior, and we should only be touching the usage metrics with these added queries. But I see that one of the methods that calls getAll is used in the proxy repository, which makes me wary of making this change.

I've tried to map out the impact of my changes here:
Skjermbilde 2024-02-29 kl  14 21 06

### DIFF
--- a/src/lib/db/segment-store.ts
+++ b/src/lib/db/segment-store.ts
@@ -15,6 +15,7 @@ import { IFlagResolver } from '../types';
 import { isDefined } from '../util';
 
 const T = {
+    features: 'features',
     segments: 'segments',
     featureStrategies: 'feature_strategies',
     featureStrategySegment: 'feature_strategy_segment',
@@ -174,7 +175,13 @@ export default class SegmentStore implements ISegmentStore {
                     T.featureStrategies,
                     `${T.featureStrategies}.id`,
                     `${T.featureStrategySegment}.feature_strategy_id`,
-                );
+                )
+                .leftJoin(
+                    T.features,
+                    `${T.featureStrategies}.feature_name`,
+                    `${T.features}.name`,
+                )
+                .where(`${T.features}.archived_at`, 'IS', null);
 
             currentSegmentUsage.forEach(
                 ({ segmentId, featureName, projectName }) => {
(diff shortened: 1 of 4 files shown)


src/lib/db/segment-store.ts
```diff
@@ -174,7 +175,13 @@ export default class SegmentStore implements ISegmentStore {
                     T.featureStrategies,
                     `${T.featureStrategies}.id`,
                     `${T.featureStrategySegment}.feature_strategy_id`,
-                );
+                )
+                .leftJoin(
```
FredrikOseberg, line 179: Here, the idea is to not returnstrategies if the archived at is null.

sjaanus (approved):
Nice!

src/lib/db/segment-store.ts
FredrikOseberg, line 179: This should only affect usage count

src/lib/db/segment-store.ts
```diff
@@ -229,11 +236,15 @@ export default class SegmentStore implements ISegmentStore {
                     'used_in_projects',
                     'used_in_features',
                 )
-                .countDistinct(
-                    `${T.featureStrategies}.project_name AS used_in_projects`,
+                .select(
```
FredrikOseberg, line 239: Should only affect usage count

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -814,7 +814,14 @@ class FeatureStrategiesStore implements IFeatureStrategiesStore {
                 `${T.featureStrategySegment}.feature_strategy_id`,
                 `${T.featureStrategies}.id`,
             )
-            .where(`${T.featureStrategySegment}.segment_id`, '=', segmentId);
+            .join(
```
FredrikOseberg, line 817: Here, we don't want to return strategies that exist on archived toggles

PR #6400, opened 2024-03-01
fix: path metric labels

gastonfournier:
## About the changes
Some of our metrics are not labeled correctly, one example is `<base-path>/api/frontend/client/metrics` is labeled as `/client/metrics`. We can see that in internal-backstage/prometheus:
image

This issue affects all endpoints that fail to validate the request body. Also, endpoints that are rejected by the authorization-middleware or the api-token-middleware are reported as `(hidden)`. 

To gain more insights on our api usage but being protective of metrics cardinality we're prefixing `(hidden)` with some well known base urls: https://github.com/Unleash/unleash/pull/6400/files#diff-1ed998ca46ffc97c9c0d5d400bfd982dbffdb3004b78a230a8a38e7644eee9b6R17-R33

## How to reproduce:
Make an invalid call to metrics (e.g. stop set to null), then check /internal-backstage/prometheus and find the 400 error. Expected to be at `path="/api/client/metrics"` but will have `path=""`:
```shell
curl -H"Authorization: *:development.unleash-insecure-client-api-token" -H'Content-type: application/json' localhost:4242/api/client/metrics -d '{
  "appName": "bash-test",
  "instanceId": "application-name-dacb1234",
  "environment": "development",
  "bucket": {
    "start": "2023-07-27T11:23:44Z",
    "stop": null,
    "toggles": {
      "myCoolToggle": {
        "yes": 25,
        "no": 42,
        "variants": {
          "blue": 6,
          "green": 15,
          "red": 46
        }
      },
      "myOtherToggle": {
        "yes": 0,
        "no": 100
      }
    }
  }
}'
```


chriswk (dismissed):
This looks good, but probably don't include the console.log in the thing deployed to our servers? Still give you a :+1: here, but please remember to remove the console.log

src/lib/middleware/response-time-metrics.ts
```diff
@@ -34,6 +34,7 @@ export function responseTimeMetrics(
             time,
             appName,
         };
+        console.log('timingInfo', timingInfo);
```
gastonfournier, line 37:
```suggestion
```

src/lib/middleware/response-time-metrics.ts
```diff
@@ -16,7 +16,7 @@ export function responseTimeMetrics(
 ): any {
     return _responseTime((req, res, time) => {
         const { statusCode } = res;
-        const pathname = req.route ? req.baseUrl + req.route.path : '(hidden)';
+        const pathname = req.route ? req.path : '(hidden)';
```
gastonfournier, line 19: When we have a route, they may not include a parent controller's path, i.e.:
https://github.com/Unleash/unleash/blob/d74bee472f016cf00b5b7cb0f62767f2aefe13f6/src/lib/routes/index.ts#L48-L51
And the internal route is what we get reported:
https://github.com/Unleash/unleash/blob/d74bee472f016cf00b5b7cb0f62767f2aefe13f6/src/lib/routes/proxy-api/index.ts#L87-L89

chriswk:
From the console log, I assume you have actually tested this and this is a case where the docs and actual implementation doesn't match?

gastonfournier:
This happens only when OpenAPI validation fails because it uses middlewares that don't preserve the parent path. Looking into this specifically

src/test/e2e/helpers/test-helper.ts
```diff
@@ -321,8 +320,6 @@ async function createApp(
     });
     const services = createServices(stores, config, db);
     const unleashSession = sessionDb(config, undefined);
-    const emitter = new EventEmitter();
-    emitter.setMaxListeners(0);
```
gastonfournier, line 325: Was this code doing anything? Config contains an event emitter and this variable is not used at all

chriswk: doesn't look like it was no. Good catch

src/lib/routes/controller.ts
```diff
@@ -206,12 +209,11 @@ export default class Controller {
         this.app.use(path, router);
     }
 
-    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
     useWithMiddleware(path: string, router: IRouter, middleware: any): void {
         this.app.use(path, middleware, router);
     }
 
-    get router(): any {
+    get router(): IRouter {
```
chriswk, line 216: Nice catch, this has been annoying me as well.

chriswk (approved):
Nice work, will be nice to get the full paths. :)

src/test/e2e/helpers/test-helper.ts
gastonfournier: I've seen some errors in logs which I believe are prevented by this line. I'll re-add it later with a comment

PR #6402, opened 2024-03-01
refactor: improve link field customizability

Tymek:
Add ability to reuse "copy link" dialogue in other places

### DIFF
--- a/frontend/src/component/admin/users/LinkField/LinkField.tsx
+++ b/frontend/src/component/admin/users/LinkField/LinkField.tsx
@@ -5,10 +5,23 @@ import useToast from 'hooks/useToast';
 interface ILinkFieldProps {
     inviteLink: string;
     small?: boolean;
+    successTitle?: string;
+    errorTitle?: string;
 }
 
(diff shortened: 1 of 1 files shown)


frontend/src/component/admin/users/LinkField/LinkField.tsx
```diff
@@ -5,19 +5,32 @@ import useToast from 'hooks/useToast';
 interface ILinkFieldProps {
     inviteLink: string;
     small?: boolean;
+    successTitle?: string;
+    errorTitle?: string;
 }
 
-export const LinkField = ({ inviteLink, small }: ILinkFieldProps) => {
+export const LinkField = ({
```
kwasniew, line 12: Sth is off here. We have a generically named component LinkField that has invite link specific props and default values

Tymek: Yes. We have wrong tooltips on for example password reset. Nobody noticed. I'll have a fix in another PR

kwasniew (approved):

PR #6403, opened 2024-03-01
feat: insights project filtering hooks with tests

Tymek:
Hooks to filter and aggregate insights data on frontend


frontend/src/component/executiveDashboard/hooks/useFilteredTrends.test.ts
```diff
@@ -0,0 +1,173 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { useFilteredTrends } from './useFilteredTrends';
+
+const mockProjectFlagTrends = [
```
kwasniew, line 4: are all the fields relevant from the perspective of this tests? Or is it the minimum set of that to show what the tests does without any noise?

Tymek: Almost, except date. I opted for this instead of type-gymnastics like "Omit<...> & Required<..>"

andreas-unleash (approved):
LGTM!

PR #6405, opened 2024-03-01
fix: Project select bug with duplicate values

andreas-unleash:
Project select fix for Executive Dashboard and Playground

Extract the select to it's own component and move to `common`
Re-use for both Dashboard and Playground
Adds the id in parenthesis when there are duplicate names

Screenshot 2024-03-01 at 12 04 22
Screenshot 2024-03-01 at 12 36 46


daveleek (approved):
LGTM!

PR #6413, opened 2024-03-04
chore: rename incoming webhooks to signals (DB)

nunogois:
https://linear.app/unleash/issue/2-1997/rename-feature-in-the-db

Renames the **incoming webhooks** feature to **signals** in the DB.

Also adds a `description` column to `action_sets`, so we can match the new designs.


gastonfournier (approved):
LGTM! There are breaking changes but this functionality is still not exposed to users.

PR #6423, opened 2024-03-04
refactor: segment-store getAll refactor, move mappers out of main method.

sjaanus:
For `getAll` method, kept only sql queries and moved mappers into separate functions.
No change in logic, just refactoring.


kwasniew (approved):
nice.
btw the semantic commit name for this is `refactor:` not `chore:`

PR #6424, opened 2024-03-04
refactor: playground uses segment read model

kwasniew:
## About the changes

Playground service uses segment read model since we don't need write model.

Screenshot 2024-03-04 at 16 26 37

In the next refactoring I will do the same with the getActiveForClient and remove getActive and getActiveForClient from the store.

### DIFF
--- a/src/lib/features/segment/segment-read-model.ts
+++ b/src/lib/features/segment/segment-read-model.ts
@@ -1,4 +1,9 @@
-import { IConstraint, IFeatureStrategySegment, ISegment } from '../../types';
+import {
+    IClientSegment,
+    IConstraint,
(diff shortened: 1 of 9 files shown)


src/lib/features/segment/segment-read-model.ts
```diff
@@ -75,4 +80,28 @@ export class SegmentReadModel implements ISegmentReadModel {
             segmentId: row.segment_id,
         }));
     }
+
+    async getActive(): Promise<ISegment[]> {
```
kwasniew, line 84: copied from the store, will delete from the store after we migrate all usages (next PR)

sjaanus (approved):

PR #6427, opened 2024-03-04
chore: add new signals icon instead of webhook icon

nunogois:
https://linear.app/unleash/issue/2-2002/add-new-signals-icon-instead-of-using-the-webhooks-icon

Adds a new icon for signals, instead of using the same icon as webhooks.

Includes some slight refactoring.

image

image


ivarconr (approved):

chriswk (approved):
LGTM

PR #6431, opened 2024-03-05
chore: remove archived column from features table

sjaanus:
This column has not been used for 1.5 years and was replace by **archived_at** column and people still get confused of why this is not working as name suggests. Removing this column to remove technical debt.

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-store.ts
@@ -300,10 +300,11 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             .count('type')
             .groupBy('type');
 
-        query.where({
-            project: projectId,
-            archived,
-        });
+        query
+            .where({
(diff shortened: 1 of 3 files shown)


src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -300,10 +300,11 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
...
 
-        query.where({
-            project: projectId,
-            archived,
-        });
+        query
+            .where({
+                project: projectId,
+            })
+            .modify(FeatureToggleStore.filterByArchived, archived);
```
kwasniew, line 307: why do we keep a query on a filed that we're dropping?

kwasniew: ah I see there's a different column with _at there

kwasniew (approved):

PR #6434, opened 2024-03-05
refactor: segment in feature oriented architecture

kwasniew:
## About the changes

Finished segment migration to feature oriented architecture. Will need to change the enteprise repo with breaking changes.
Manually dragged store and service for segments into the feature directory. All other directories are auto updated.


FredrikOseberg (commented):
Could we move the e2e tests also?

sjaanus (approved):

PR #6439, opened 2024-03-05
task: added scim id to user

chriswk:
SCIM synchronizations requires a stable id no matter how many changes are made to username and email (our other unique fields). In addition, exposing internal incremented database ids to an external service (our current id field) feels insecure. Our plan is to create either a uuidv7 or ulid when scim operations are performed against the user, so the external scim provisioner has a stable globally unique id to use to refer to the users they're modifying.


sighphyre (approved):
G

PR #6446, opened 2024-03-06
feat: frontend api observability

kwasniew:
## About the changes

We want to measure execution time of the frontend API evaluation. Similar measurement to scheduler execution and store execution time. 

Design decisions:
* high level event FRONTEND_API_TIME with one lower level action evaluateEnabledFeatures. Adding space for more actions in the FRONTEND_API_TIME timeline. 

## Discussion points

I wanted to correlate the time with the number of all features and enabled features after filtering. However our abstraction doesn't allow for this. Also I'm not sure if it's a good idea to mix time metrics with number of results metrics. For other time based metrics we only add actual time not the number of results in a query.

### DIFF
--- a/src/lib/metric-events.ts
+++ b/src/lib/metric-events.ts
@@ -1,6 +1,7 @@
 const REQUEST_TIME = 'request_time';
 const DB_TIME = 'db_time';
 const SCHEDULER_JOB_TIME = 'scheduler_job_time';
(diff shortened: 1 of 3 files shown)


kwasniew:
may not be needed since we already have endpoint observability

kwasniew:
We can achieve the same result by tracing /api/frontend endpoint

PR #6450, opened 2024-03-06
fix: scim id was not unique

chriswk:
This PR drops the previous PRs scim_id idx and replaces it with an actual UNIQUE index


sighphyre (approved):
LGTM but this is being burned by a test right now apparently

PR #6451, opened 2024-03-06
fix: align privacy policy link to the center

nunogois:
Aligns the privacy policy to the center in the profile dropdown, just like we do for the profile settings.

image


nicolaesocaciu (approved):
Perfect

PR #6453, opened 2024-03-06
feat: add support for markdown in toggle descriptions

### LINKED ISSUE #6399: feat: Markdown support in flag description

Adding support for simple syntax like **bold**, _italic_ and links would help. Unsure about "heading" etc due to limited space for descriptions. 

### Background

Adding links to descriptions is not rendering nice. 

### Solution suggestions

Reuse the `<Markdown>` component in the rendering.

ivarconr:
This is a small PoC on supporting limited markdown in the feature flag descriptions. I think it make sense to only support a very basic set of markdown formatting initially, as it can easily mess up the UI to support to much. Current support is for 
the following allowed html elements: "a", "b", "p", "strong" and "em", anything else is automatically stripped away. 

You enter Markdown when defining a toggle:
image

Markdown is used on the feature toggle view:
image

Markdown is converted to plain text in the toggle list:
image

Markdown is used in the html tooltip:
image

**Concerns:** 
- I am concerned with how this implementation will try to strip the description for all toggles at render time. I am unsure of the performance of this. 
- Can we loose control on the rendering by supporting markdown? 

**Other possible solutions.** 
* We could store a plain text description and a markdown formatted version when we create / update a feature toggle. This would remove some of the rendering cost on the UI, but would require changes in the response format in the  API and lead to more data being exchanged. 

Closes #6399

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -2,7 +2,10 @@
   "name": "unleash-frontend-local",
   "version": "0.0.0",
   "private": true,
-  "files": ["index.js", "build"],
+  "files": [
+    "index.js",
+    "build"
+  ],
   "engines": {
     "node": ">=18"
   },
@@ -142,11 +145,18 @@
     }
   },
   "browserslist": {
-    "production": [">0.2%", "not dead", "not op_mini all"],
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not op_mini all"
+    ],
     "development": [
       "last 1 chrome version",
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "dependencies": {
+    "remove-markdown": "^0.5.0"
   }
 }
--- a/frontend/src/component/common/Markdown/Markdown.tsx
+++ b/frontend/src/component/common/Markdown/Markdown.tsx
@@ -22,3 +22,7 @@ const LinkRenderer = ({
 export const Markdown = (props: ComponentProps<typeof ReactMarkdown>) => (
     <ReactMarkdown components={{ a: LinkRenderer }} {...props} />
 );
+
+export const SimpleMarkdown = (props: ComponentProps<typeof ReactMarkdown>) => (
+    <ReactMarkdown components={{ a: LinkRenderer }} skipHtml allowedElements={['a', 'p', 'strong', 'em']} {...props} />
+);
--- a/frontend/src/component/common/Table/cells/LinkCell/LinkCell.tsx
+++ b/frontend/src/component/common/Table/cells/LinkCell/LinkCell.tsx
@@ -12,6 +12,10 @@ import {
     StyledDescription,
 } from './LinkCell.styles';
 
+//@ts-ignore
+import removeMd from 'remove-markdown';
+import { SimpleMarkdown } from 'component/common/Markdown/Markdown';
+
 interface ILinkCellProps {
     title?: string;
     to?: string;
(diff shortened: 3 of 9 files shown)


frontend/package.json
```diff
@@ -142,11 +145,18 @@
...
+      "not op_mini all"
+    ],
     "development": [
       "last 1 chrome version",
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "dependencies": {
+    "remove-markdown": "^0.5.0"
```
nunogois, line 160: Due to the way vite bundles dependencies, and for consistency sake, I think this belongs in `devDependencies` instead.

frontend/src/component/common/Markdown/Markdown.tsx
```diff
@@ -22,3 +22,7 @@ const LinkRenderer = ({
 export const Markdown = (props: ComponentProps<typeof ReactMarkdown>) => (
     <ReactMarkdown components={{ a: LinkRenderer }} {...props} />
 );
+
+export const SimpleMarkdown = (props: ComponentProps<typeof ReactMarkdown>) => (
+    <ReactMarkdown components={{ a: LinkRenderer }} skipHtml allowedElements={['a', 'p', 'strong', 'em']} {...props} />
+);
```
nunogois, line 28: Not sure if it's better but I wonder if you could skip the new dependency by creating a PlainTextMarkdown component that allows/renders no elements, just the plain text.

frontend/src/component/common/Table/cells/LinkCell/LinkCell.tsx
```diff
@@ -12,6 +12,10 @@ import {
     StyledDescription,
 } from './LinkCell.styles';
 
+//@ts-ignore
```
nunogois, line 15: Just curious - Why is this needed?

frontend/src/component/common/Table/cells/LinkCell/LinkCell.tsx
```diff
@@ -26,23 +30,25 @@ export const LinkCell: React.FC<ILinkCellProps> = ({
     subtitle,
     children,
 }) => {
+    const subTitleClean = removeMd(subtitle);
```
nunogois, line 33: I think this casing is more consistent, since it matches the original name.

Alternatively, if it's more readable, we could use something like `plainSubtitle` or `plainTextSubtitle`.

```suggestion
    const subtitleClean = removeMd(subtitle);
```

frontend/src/component/common/Markdown/Markdown.tsx
nunogois: Nvm, I see how the plain text is useful for the highlighting, so I prefer your approach.

nunogois (approved):
This feels like a reasonable approach to me and if we detect any issues it's good to know that it's behind a flag 👍

nunogois:
I also have some concerns about the links styling inside the description panel in the feature view, since both are purple currently.

Maybe @nicolaesocaciu can help with a suggestion that has better contrast.

frontend/package.json
ivarconr: yes, this was a lazy mistake by me.

frontend/src/component/common/Table/cells/LinkCell/LinkCell.tsx
ivarconr: because `remove-markdown` does not have typescript definitions. I should add one instead of ignoring, just did not want to do it for the PoC.

ivarconr:
> When we have links on a purple background they should be "white" similar with the links in the purple sidepanel

Agreed. I did not consider any styles yet, but this needs to be addressed if we move forward with markdown and link support in feature flag descriptions.

Issue #6408, opened 2024-03-01
fix: Starting unleash without DB details gives silly error message. 


ivarconr:
### Describe the bug

When starting Unleash without any DB details you get a silly error message, Unleash does not exit, but does not start either. It crashes after a few minutes. 

### Steps to reproduce the bug

1. clone repo
2. yarn start
3. Observe error message:

```
➜  unleash git:(main) ✗ yarn start
yarn run v1.22.21
$ TZ=UTC node ./dist/server.js
[2024-03-01T11:46:35.410] [ERROR] server-impl.js - Locking error: SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string
```

### Expected behavior

Unleash should require DB details and should exit hard if none are provided. 

### Logs, error output, etc.

```bash
➜  unleash git:(main) ✗ yarn start
yarn run v1.22.21
$ TZ=UTC node ./dist/server.js
[2024-03-01T11:46:35.410] [ERROR] server-impl.js - Locking error: SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string
```

### Unleash version

main

### Subscription type

Open source

### Hosting type

Self-hosted

PR #6454, opened 2024-03-06
fix: only release migration lock if acquired

ivarconr:
We should not try to release the migration lock if where unable to acquire it. By trying to close it when we have not successfully connected to the database we end up hanging for a while before the process is eventually killed. 

I did not add a better error-message, as Unleash now gives a better error stack and crashes immediate if you start without a database password. We should still consider if you need to specify db credentials or not. Technically it is possible to have a postgres without a password (but it is likely not common).

Closes: #6408


gastonfournier (approved):
Looks good!

PR #6460, opened 2024-03-07
feat: create global repository for frontend repositories

sjaanus:


kwasniew (approved):
This new code is WIP and is not enabled yet so we can stabilize it in the next PR

PR #6462, opened 2024-03-07
chore: consider execution limits per minute and actions limit per

gastonfournier:
## About the changes
Define a schema that works both for the frontend and the backend to define soft limits in the resource usage.


nunogois (approved):
Very cool!

PR #6463, opened 2024-03-07
chore: automate openapi schema list

gastonfournier:
## About the changes
This PR automates the generation of exported open api schemas on pre-commit, removing some manual steps and also standardizing the process. The schema list definition now looks way simpler:
https://github.com/Unleash/unleash/blob/b6f3877296389fcecd344d1c9295027ac6c6d52a/src/lib/openapi/index.ts#L37-L49

Also added https://github.com/Unleash/unleash/blob/2817e66b29ecf5bc908e340931cbdc0281192dea/src/lib/openapi/spec/index.ts#L1-L4 for devs


src/lib/openapi/index.ts
```diff
@@ -233,207 +34,19 @@ interface OpenAPIV3DocumentWithServers extends OpenAPIV3.Document {
...
+/*
+ * All schemas in `openapi/spec` should be listed here.
+ * Instead of listing them all maunally, exclude those that are not schemas (maybe they should be moved elsewhere)
+ */
+import * as importedSchemas from './spec';
+const {
+    constraintSchemaBase,
+    unknownFeatureEvaluationResult,
+    playgroundStrategyEvaluation,
+    strategyEvaluationResults,
```
gastonfournier, line 46: These are not schemas, but they are exported inside schema files. I don't think it's a problem but at least it posts a question on whether they should be there or not.

src/lib/openapi/spec/client-metrics-env-schema.ts
```diff
@@ -61,4 +61,4 @@ export const clientMetricsEnvSchema = {
     },
 } as const;
 
-export type ClientMetricsSchema = FromSchema<typeof clientMetricsEnvSchema>;
+export type ClientMetricsEnvSchema = FromSchema<typeof clientMetricsEnvSchema>;
```
gastonfournier, line 64: This one was a bad copy paste that was spotted thanks to automating the exports

src/lib/services/tag-schema.ts
```diff
@@ -1,8 +1,9 @@
 import Joi from 'joi';
 
 import { customJoi } from '../routes/util';
-import { TAG_MAX_LENGTH, TAG_MIN_LENGTH } from '../../lib/openapi';
 
+export const TAG_MIN_LENGTH = 2;
+export const TAG_MAX_LENGTH = 50;
```
gastonfournier, line 6: These two were moved instead of excluding them from the schemas

.husky/pre-commit
```diff
@@ -2,3 +2,10 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 npx lint-staged
+
+node .husky/update-openapi-spec-list.js
```
gastonfournier, line 6: Using node instead of bash, because it should be available in local dev

src/lib/openapi/index.ts
thomasheartman: Good catch. I think I'm responsible for all of these, actually 😅 I think some of them make sense to keep where they are, but others maybe not. I don't think `unknownFeatureEvaluationResult` is actually used anywhere 🤔

thomasheartman (approved):
This looks good to me! 😄 While I'm generally not a big fan of pre-commit hooks, I think automating this is great. Is there a way to do it that's less invasive?

.husky/update-openapi-spec-list.js
thomasheartman, line 1: Does this *always* write the file? How long does it take?

gastonfournier: > I'm generally not a big fan of pre-commit hooks

That makes 2 of us, but this is a quick way to find out whether this is a good solution or not. There are other alternatives, some less invasive, but this will help us understand what are our needs.

> Does this always write the file?

Yes, only on commit, but it checks if the file was modified, and only then includes it in the change.

> How long does it take?

Around 30ms

PR #6464, opened 2024-03-07
docs: add note about upgrading to version 5.7 and above

thomasheartman:
Upgrading to 5.7 and above from earlier versions is a little problematic. Let's document it.

### DIFF
--- a/website/docs/using-unleash/deploy/upgrading-unleash.md
+++ b/website/docs/using-unleash/deploy/upgrading-unleash.md
@@ -11,7 +11,7 @@ When running on high-availability (multiple Unleash instances), upgrading from v
 
 If you rather want to avoid that, you should first update to 5.6.11 and ensure all your Unleash instances are running 5.6.11 before upgrading to 5.7 or later. Any Unleash instances left on earlier versions after the backing database has been updated will continue to work, but you will not be able to log in to them, but the SDKs will continue working fine.
 
-Once you have upgraded to version 5.7 and above, you can no longer downgrade to versions before 5.7.
+Once you have upgraded to version 5.7 and above, you can no longer downgrade to versions before 5.7 without manual changes to the database.
 
 ## Upgrading directly from v3.x to v5.x
 
--- a/website/docs/using-unleash/deploy/upgrading-unleash.md
+++ b/website/docs/using-unleash/deploy/upgrading-unleash.md
@@ -7,7 +7,7 @@ Generally, the intention is that `unleash-server` should always provide support 
 
 ## Upgrading to 5.7 and later from versions < 5.6.11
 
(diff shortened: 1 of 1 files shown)


website/docs/using-unleash/deploy/upgrading-unleash.md
```diff
@@ -5,6 +5,14 @@ import VideoContent from '@site/src/components/VideoContent.jsx'
 
 Generally, the intention is that `unleash-server` should always provide support for clients one major version lower than the current one. This should make it possible to upgrade `unleash` gradually.
 
+## Upgrading to 5.7 and later from versions < 5.6.11
+
+When running on high-availability (multiple Unleash instances), upgrading from versions lower than 5.6.11 to major will cause a temporary UI unavailability while old versions and new versions are both serving traffic, due to a compatibility issue. If you can afford having a small period of time with the UI unavailable (note the SDKs will not be affected), then  you can safely upgrade.
```
thomasheartman, line 10: @gastonfournier , I'm making this little change here:

```suggestion
When running on high-availability (multiple Unleash instances), upgrading from versions lower than 5.6.11 to version 5.7 or higher will cause a temporary UI unavailability while old versions and new versions are both serving traffic, due to a compatibility issue. If you can afford having a small period of time with the UI unavailable (note the SDKs will not be affected), then  you can safely upgrade.
```

website/docs/using-unleash/deploy/upgrading-unleash.md
```diff
@@ -5,6 +5,14 @@ import VideoContent from '@site/src/components/VideoContent.jsx'
 
 Generally, the intention is that `unleash-server` should always provide support for clients one major version lower than the current one. This should make it possible to upgrade `unleash` gradually.
 
+## Upgrading to 5.7 and later from versions < 5.6.11
+
+When running on high-availability (multiple Unleash instances), upgrading from versions lower than 5.6.11 to version 5.7 or higher will cause a temporary UI unavailability while old versions and new versions are both serving traffic, due to a compatibility issue. If you can afford having a small period of time with the UI unavailable (note the SDKs will not be affected), then  you can safely upgrade.
+
+If you rather want to avoid that, you should first update to 5.6.11 and ensure all your Unleash instances are running 5.6.11 before upgrading to 5.7 or later. Any Unleash instances left on earlier versions after the backing database has been updated will continue to work, but you will not be able to log in to them, but the SDKs will continue working fine.
+
+Once you have upgraded to version 5.7 and above, you can no longer downgrade to versions before 5.7.
```
gastonfournier, line 14:
```suggestion
Once you have upgraded to version 5.7 and above, you can no longer downgrade to versions before 5.7 without manual changes to the database.
```

gastonfournier (approved):
Looks nice just one more suggestion to reflect that it's still possible but discouraged (maybe you can stress that even further)

thomasheartman:
Thanks! I like that suggestion, so I'll keep it as is. I don't necessarily think we need to stress that last point further. Thanks for your help here 😄

gastonfournier:
Looks nice then!

PR #6467, opened 2024-03-07
chore: add actions warning to delete project

nunogois:
https://linear.app/unleash/issue/2-1831/what-happens-with-automated-actions-if-the-project-is-archived

Adds an actions-related warning to the project deletion.

image

### DIFF
--- a/frontend/src/component/project/Project/ProjectSettings/Settings/DeleteProject.tsx
+++ b/frontend/src/component/project/Project/ProjectSettings/Settings/DeleteProject.tsx
@@ -4,18 +4,17 @@ import PermissionButton from 'component/common/PermissionButton/PermissionButton
 import { DeleteProjectDialogue } from '../../DeleteProject/DeleteProjectDialogue';
 import { useState } from 'react';
 import { useNavigate } from 'react-router';
+import { useUiFlag } from 'hooks/useUiFlag';
+import { useActions } from 'hooks/api/getters/useActions/useActions';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
+import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 
 const StyledContainer = styled('div')(({ theme }) => ({
+    display: 'flex',
+    flexDirection: 'column',
     borderTop: `1px solid ${theme.palette.divider}`,
-}));
-
-const StyledTitle = styled('div')(({ theme }) => ({
     paddingTop: theme.spacing(4),
-    lineHeight: 2,
-}));
-
-const StyledCounter = styled('div')(({ theme }) => ({
-    paddingTop: theme.spacing(3),
+    gap: theme.spacing(2),
 }));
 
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/Project/ProjectSettings/Settings/DeleteProject.tsx
```diff
@@ -33,21 +32,45 @@ export const DeleteProject = ({
     projectId,
     featureCount,
 }: IDeleteProjectProps) => {
+    const { isEnterprise } = useUiConfig();
+    const automatedActionsEnabled = useUiFlag('automatedActions');
+    const { actions } = useActions(projectId);
```
gastonfournier, line 37: Just thinking if it makes sense for reusability, and in particular because I see we are repeating the same `condition={isEnterprise() && automatedActionsEnabled}` in many places:
```suggestion
    const { actions, automatedActionsEnabled } = useActions(projectId);
```

So we can encapsulate the logic of isEnabled inside the useActions and just expose a boolean from it

gastonfournier (approved):
LGTM! Just an idea to consider but can be done as another PR

frontend/src/component/project/Project/ProjectSettings/Settings/DeleteProject.tsx
nunogois: I understand where you're going with this suggestion, and part of me likes it. However I think these are different responsibilities and if we think about the way we'll clean up this flag in the future, I think this is the better approach for now.

gastonfournier: well, I don't know, because eventually we will still have ways of disabling this feature (say a customer who does not want it because of reasons). If we have it inside the `useActions` you can just change in one place the `useUiFlag('automatedActions');` for a `true` and you'd be done. But I think this turns into a more philosophical conversation on how do we handle enabling/disabling new features in the UI

nunogois: Yes but I think it's important that we're being specific about this requiring both Enterprise and the feature flag being on instead of that information being encapsulated inside the API getter hook. 

Later on when we remove the flag, we go from `isEnterprise() && isFlagOn" to only "isEnterprise()", extremely close to where the impact is, which I like in this case, since it's important for whoever looks at this code to see at a glance that those messages are only shown for enterprise customers.

I guess in the future it could be nice if we could add these conditions to the feature somehow in a central manner. However I don't think it:
 - Belongs with the flag, since flags should be short-lived;
 - Belongs in the API hooks, since those have their own specific responsibility;

gastonfournier: I still believe that the responsibility of whether it should be enabled or not is central to the useActions component (or whatever is called), because it's a central place that defines whether it's an enterprise feature or not, and when to display it or not if the feature is beta rather than having these conditions scattered across multiple files. Still I don't have a strong attachment to one option vs the other, so let's agree to disagree for now

PR #6468, opened 2024-03-07
fix: project tokens path typo

nunogois:
Noticed a small typo in the path shown in the project API access side panel.


daveleek (approved):
Oh wow, good catch! LGTM!

PR #6474, opened 2024-03-08
feat: adapted CRUD store from enterprise into OSS

gastonfournier:
## About the changes
This ports the CRUD store into OSS which is an abstraction to reduce the amount of boilerplate code we have to write in stores.

By extending CRUDStore, the store becomes simply the type definition:
```typescript
 type ActionModel = { 
     actionSetId: number; 
     action: string; 
     executionParams: Record<string, unknown>; 
     createdByUserId: number; 
     sortOrder: number; 
 }; 
  
 export class ActionStore extends CRUDStore< 
     ActionModel & { id: number; createdAt: Date }, 
     ActionModel 
 > { 
}
```

And eventually specific mappings between those types can be provided (if the mapping is more complex than camelCase -> snake_case):
```typescript
 toRow: ({ project, name, actor, match, createdByUserId }) => ({ 
     created_by_user_id: createdByUserId, 
     project, 
     name, 
     actor_id: actor, 
     source: match.source, 
     source_id: match.sourceId, 
     payload: match.payload, 
 }), 
 fromRow: ({ 
     id, 
     created_at, 
     created_by_user_id, 
     project, 
     name, 
     actor_id, 
     source, 
     source_id, 
     payload, 
 }) => ({ 
     id, 
     createdAt: created_at, 
     createdByUserId: created_by_user_id, 
     project, 
     name, 
     actor: actor_id, 
     match: { 
         source, 
         sourceId: source_id, 
         payload, 
     }, 
 }), 
```
Stores can also be extended to include additional functionality in case you need to join with another table or do an aggregation, but it significantly reduces the amount of boilerplate code needed to create a basic store

### DIFF
--- a/src/lib/db/crud/crud-store.ts
+++ b/src/lib/db/crud/crud-store.ts
@@ -1,0 +1,132 @@
+import { NotFoundError } from '../../error';
+import { DB_TIME } from '../../metric-events';
+import { Db, IUnleashConfig } from '../../server-impl';
+import { Store } from '../../types/stores/store';
+import metricsHelper from '../../util/metrics-helper';
+import { defaultFromRow, defaultToRow } from './default-mappings';
+import { Row } from './row-type';
+
+export type CrudStoreConfig = Pick<IUnleashConfig, 'eventBus'>;
+
+/**
+ * This abstract class defines the basic operations for a CRUD store
+ *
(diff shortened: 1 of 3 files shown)


nunogois (approved):
🚀

PR #6475, opened 2024-03-08
fix: mapper function should be partial

gastonfournier:
Small fix to make mapper function partial as it should be


src/lib/db/crud/crud-store.ts
```diff
@@ -64,14 +64,14 @@ export abstract class CRUDStore<
             allQuery = allQuery.where(this.toRow(query) as Record<string, any>);
         }
         const items = await allQuery;
-        return items.map(this.fromRow);
+        return items.map(this.fromRow) as ReadModel[];
```
gastonfournier, line 67: In all these internal methods **we know** we're fetching all data

nunogois (approved):

PR #6476, opened 2024-03-08
Compare old results with new frontend api

kwasniew:
## About the changes

Running old frontend API and new frontend API side by side and comparing results.

Details:
* comparison only on when flag enabled
* always returning old results and warning if new results are different
* global cache is updated only when the flag is on
* copy pasted most of the old code with minor tweak so it's easy to delete old code after the migration
* sticking to the existing naming (separate PR will remove proxy and replace it with frontend API naming)
* next PR will create a composition root for the proxy service

### DIFF
--- a/src/lib/services/index.ts
+++ b/src/lib/services/index.ts
@@ -111,6 +111,9 @@ import {
 import { InactiveUsersService } from '../users/inactive/inactive-users-service';
 import { SegmentReadModel } from '../features/segment/segment-read-model';
 import { FakeSegmentReadModel } from '../features/segment/fake-segment-read-model';
+import { GlobalFrontendApiCache } from '../proxy/global-frontend-api-cache';
+import ClientFeatureToggleReadModel from '../proxy/client-feature-toggle-read-model';
+import FakeClientFeatureToggleReadModel from '../proxy/fake-client-feature-toggle-read-model';
 
 export const createServices = (
     stores: IUnleashStores,
@@ -293,12 +296,27 @@ export const createServices = (
         ? createClientFeatureToggleService(db, config)
         : createFakeClientFeatureToggleService(config);
 
-    const proxyService = new ProxyService(config, stores, {
-        featureToggleServiceV2,
-        clientMetricsServiceV2,
-        settingService,
+    const clientFeatureToggleReadModel = db
+        ? new ClientFeatureToggleReadModel(db, config.eventBus)
+        : new FakeClientFeatureToggleReadModel();
+    const globalFrontendApiCache = new GlobalFrontendApiCache(
+        config,
+        segmentReadModel,
+        clientFeatureToggleReadModel,
         configurationRevisionService,
-    });
+    );
+
+    const proxyService = new ProxyService(
+        config,
+        stores,
+        {
+            featureToggleServiceV2,
+            clientMetricsServiceV2,
+            settingService,
+            configurationRevisionService,
+        },
+        globalFrontendApiCache,
+    );
 
     const edgeService = new EdgeService({ apiTokenService }, config);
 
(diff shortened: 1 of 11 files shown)


src/lib/services/index.ts
```diff
@@ -293,12 +296,27 @@ export const createServices = (
         ? createClientFeatureToggleService(db, config)
         : createFakeClientFeatureToggleService(config);
 
-    const proxyService = new ProxyService(config, stores, {
-        featureToggleServiceV2,
-        clientMetricsServiceV2,
-        settingService,
+    const clientFeatureToggleReadModel = db
```
kwasniew, line 299: this will be composition root in next pr

src/lib/services/proxy-service.ts
```diff
@@ -85,6 +97,33 @@ export class ProxyService {
             }));
     }
 
+    async getNewProxyFeatures(
+        token: IApiUser,
+        context: Context,
+    ): Promise<ProxyFeatureSchema[]> {
+        const client = await this.newClientForProxyToken(token);
```
kwasniew, line 104: copy paste of old code with one line modification

src/lib/services/proxy-service.ts
```diff
@@ -143,6 +196,33 @@ export class ProxyService {
         return client;
     }
 
+    private async createNewClientForProxyToken(
+        token: IApiUser,
+    ): Promise<Unleash> {
+        const repository = new FrontendApiRepository(
```
kwasniew, line 202: copy paste of old code with this line replaced to use new repository

src/lib/types/experimental.ts
```diff
@@ -262,6 +263,10 @@ const flags: IFlags = {
         process.env.UNLEASH_EXPERIMENTAL_RESPONSE_TIME_METRICS_FIX,
         false,
     ),
+    globalFrontendApiCache: parseEnvVarBoolean(
```
kwasniew, line 266: new flag for safe rollout on the new frontend api

src/lib/services/proxy-service.ts
```diff
@@ -117,6 +156,20 @@ export class ProxyService {
...
 
+    private async newClientForProxyToken(token: IApiUser): Promise<Unleash> {
+        ProxyService.assertExpectedTokenType(token);
+
+        let newClient = this.newClients.get(token.secret);
+        if (!newClient) {
+            newClient = this.createNewClientForProxyToken(token);
+            this.newClients.set(token.secret, newClient);
+            // TODO: do we need this twice?
+            // this.config.eventBus.emit(PROXY_REPOSITORY_CREATED);
```
kwasniew, line 167: we will uncomment this when the old client is gone

src/lib/services/proxy-service.ts
```diff
@@ -45,21 +47,31 @@ export class ProxyService {
...
+
     /**
-     * This is intentionally a Promise becasue we want to be able to await
+     * This is intentionally a Promise because we want to be able to await
      * until the client (which might be being created by a different request) is ready
      * Check this test that fails if we don't use a Promise: src/test/e2e/api/proxy/proxy.concurrency.e2e.test.ts
      */
     private readonly clients: Map<ApiUser['secret'], Promise<Unleash>> =
         new Map();
+    private readonly newClients: Map<ApiUser['secret'], Promise<Unleash>> =
```
kwasniew, line 59: we store new and old clients side by side. when flag is disabled new clients will be empty.

src/lib/services/proxy-service.ts
```diff
@@ -17,18 +17,20 @@ import { validateOrigins } from '../util';
 import { BadDataError, InvalidTokenError } from '../error';
 import { PROXY_REPOSITORY_CREATED } from '../metric-events';
 import { ProxyRepository } from '../proxy';
+import { FrontendApiRepository } from '../proxy/frontend-api-repository';
+import { GlobalFrontendApiCache } from '../proxy/global-frontend-api-cache';
 
-type Config = Pick<
+export type Config = Pick<
```
kwasniew, line 23: tests need those types

sjaanus (approved):
LG

PR #6477, opened 2024-03-08
feat: make frontend api complexity O(n) instead of O(n2)

sjaanus:
Now frontend api complexity for evaluation is down from O(n²) to O(n). Because previous getToggle was filtering from the features pool, now we have hashmap, so we can find by key.

### DIFF
--- a/src/lib/proxy/client-feature-toggle-read-model.ts
+++ b/src/lib/proxy/client-feature-toggle-read-model.ts
@@ -1,11 +1,5 @@
 import { Knex } from 'knex';
-import {
-    IFeatureToggleClient,
-    IFeatureToggleQuery,
-    IStrategyConfig,
-    ITag,
-    PartialDeep,
-} from '../types';
+import { IFeatureToggleClient, IStrategyConfig, PartialDeep } from '../types';
 import { ensureStringValue, mapValues } from '../util';
 import { Db } from '../db/db';
 import FeatureToggleStore from '../features/feature-toggle/feature-toggle-store';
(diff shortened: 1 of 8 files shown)


src/lib/proxy/client-feature-toggle-read-model.ts
```diff
@@ -52,7 +42,6 @@ export default class ClientFeatureToggleReadModel
             'fe.environment as environment',
             'fs.id as strategy_id',
             'fs.strategy_name as strategy_name',
-            'fs.title as strategy_title',
```
sjaanus, line 55: Title is not needed in frontend api

src/lib/proxy/client-feature-toggle-read-model.ts
```diff
@@ -208,30 +181,9 @@ export default class ClientFeatureToggleReadModel
         );
     }
 
-    private addTag(
```
sjaanus, line 211: Tags are not needed in frontend api

src/lib/proxy/frontend-api-repository.ts
```diff
@@ -50,9 +50,7 @@ export class FrontendApiRepository
 
     getToggle(name: string): FeatureInterface {
         //@ts-ignore (we must update the node SDK to allow undefined)
-        return this.getToggles(this.token).find(
-            (feature) => feature.name === name,
-        );
+        return this.globalFrontendApiCache.getToggle(name, this.token);
```
sjaanus, line 53: This is the method that will be called n times (number of features) . This must be fastest!

src/lib/proxy/frontend-api-repository.ts
```diff
@@ -50,9 +50,7 @@ export class FrontendApiRepository
...
     }
 
     getToggles(): FeatureInterface[] {
```
sjaanus, line 56: This method is called just once.

kwasniew (approved):

PR #6480, opened 2024-03-08
feat: insights share events

Tymek:
Figure out how many people are using and sharing insights in Beta.

https://linear.app/unleash/issue/1-2145/track-number-of-opened-share-links


andreas-unleash (approved):
LG

PR #6481, opened 2024-03-08
fix: add actions validation

nunogois:
https://linear.app/unleash/issue/2-2022/improve-actions-validation

Improves our current actions form validation. 

Empty actions are now ignored on the payload and we get errors in actions where any of the required fields are empty.

Also refactored our current actions into a constant map that can be shared across frontend and backend.


gastonfournier (approved):
Looks fine, but we still have to add the backend validation, otherwise someone using the API will be able to bypass this frontend validation

nunogois:
> Looks fine, but we still have to add the backend validation, otherwise someone using the API will be able to bypass this frontend validation

Agreed, will be tackled in a follow up task.

PR #6486, opened 2024-03-11
feat: compare feture definitions not evaluations

kwasniew:
## About the changes

Comparing feature definitions since evaluation may differ when random stickiness is used.


sjaanus (approved):
I think this covers 90% of issues, so it is fine. But isEnabled call is using our proxy, so it does not testing that proxy is returning correct data, when getToggle is called. So we might need something for that.

PR #6487, opened 2024-03-11
refactor: proxy/frontend api in feature oriented architecture

kwasniew:
## About the changes

Proxy/Frontend API in one directory to get feature-oriented architecture. Move service and controller. Next PR will move it into feature directory.

### DIFF
--- a/src/lib/proxy/proxy-controller.ts
+++ b/src/lib/proxy/proxy-controller.ts
@@ -1,8 +1,8 @@
 import { Request, Response } from 'express';
-import Controller from '../controller';
-import { IUnleashConfig, IUnleashServices, NONE } from '../../types';
-import { Logger } from '../../logger';
-import { IApiUser } from '../../types/api-user';
+import Controller from '../routes/controller';
+import { IUnleashConfig, IUnleashServices, NONE } from '../types';
+import { Logger } from '../logger';
+import { IApiUser } from '../types/api-user';
 import {
     ClientMetricsSchema,
     createRequestSchema,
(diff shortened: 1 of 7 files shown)


sjaanus (approved):

src/lib/proxy/proxy-controller.ts
```diff
@@ -1,8 +1,8 @@
 import { Request, Response } from 'express';
-import Controller from '../controller';
-import { IUnleashConfig, IUnleashServices, NONE } from '../../types';
-import { Logger } from '../../logger';
-import { IApiUser } from '../../types/api-user';
+import Controller from '../routes/controller';
```
FredrikOseberg, line 2: I'd like the file to be named "frontend-api-controller" to be more consistent with what we refer to it as internally. This should also be the controller name.

src/lib/proxy/proxy-service.ts
```diff
@@ -16,9 +16,9 @@ import {
 import { validateOrigins } from '../util';
 import { BadDataError, InvalidTokenError } from '../error';
 import { PROXY_REPOSITORY_CREATED } from '../metric-events';
-import { ProxyRepository } from '../proxy';
-import { FrontendApiRepository } from '../proxy/frontend-api-repository';
-import { GlobalFrontendApiCache } from '../proxy/global-frontend-api-cache';
+import { ProxyRepository } from './index';
```
FredrikOseberg, line 19: Would be great if we could call this FrontendApiService and "frontend-api-service.ts"

FredrikOseberg (commented):
In general, the term proxy is very confusing and I'd love it if we can reduce the cognitive load by referring to the grouping, filenames and names of the actual instances as frontend api.

PR #6489, opened 2024-03-11
feat: frontend api clean up not needed columns

sjaanus:
1. Clean up not used fields
2. Fix a bug where segment was not attached

### DIFF
--- a/src/lib/proxy/client-feature-toggle-read-model.ts
+++ b/src/lib/proxy/client-feature-toggle-read-model.ts
@@ -36,7 +36,6 @@ export default class ClientFeatureToggleReadModel
             'features.project as project',
             'features.stale as stale',
             'features.impression_data as impression_data',
-            'features.created_at as created_at',
             'fe.variants as variants',
             'fe.enabled as enabled',
             'fe.environment as environment',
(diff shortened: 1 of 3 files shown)


src/lib/proxy/client-feature-toggle-read-model.ts
```diff
@@ -142,6 +140,9 @@ export default class ClientFeatureToggleReadModel
                 feature.strategies = feature.strategies || [];
                 feature.strategies.push(this.rowToStrategy(row));
             }
+            if (row.segment_id) {
```
sjaanus, line 143: This was missing from new query

src/lib/proxy/client-feature-toggle-read-model.ts
```diff
@@ -118,7 +117,6 @@ export default class ClientFeatureToggleReadModel
                     project: row.project,
                     stale: row.stale,
                     type: row.type,
-                    dependencies: [],
```
sjaanus, line 121: This is same as in old frontend api.

kwasniew (commented):
We need a test to show a bug we fixed

kwasniew (approved):

PR #6495, opened 2024-03-11
feat: count frontend api requests 

sjaanus:
Now frontend API requests will be counted separately under getAllByfrontend. We are already tracking new FE db calls, so we can build grafana dashboard.

### DIFF
--- a/src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
+++ b/src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
@@ -25,7 +25,7 @@ import Raw = Knex.Raw;
 export interface IGetAllFeatures {
     featureQuery?: IFeatureToggleQuery;
     archived: boolean;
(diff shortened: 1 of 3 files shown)


src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
```diff
@@ -71,7 +71,7 @@ export default class FeatureToggleClientStore
         const isAdmin = requestType === 'admin';
         const isPlayground = requestType === 'playground';
         const environment = featureQuery?.environment || DEFAULT_ENV;
-        const stopTimer = this.timer(`getFeatureAdmin${requestType}`);
+        const stopTimer = this.timer(`getAllBy${requestType}`);
```
sjaanus, line 74: Renaming to be not confusing.

kwasniew (approved):

PR #6496, opened 2024-03-11
fix: compare definitions

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/proxy/proxy-service.ts
+++ b/src/lib/proxy/proxy-service.ts
@@ -102,8 +102,8 @@ export class ProxyService {
         const oldClient = await this.clientForProxyToken(token);
         const oldDefinitions = oldClient.getFeatureToggleDefinitions() || [];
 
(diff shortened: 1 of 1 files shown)


src/lib/proxy/proxy-service.ts
```diff
@@ -102,8 +102,8 @@ export class ProxyService {
         const oldClient = await this.clientForProxyToken(token);
         const oldDefinitions = oldClient.getFeatureToggleDefinitions() || [];
 
-        const newClient = await this.newClientForProxyToken(token);
-        const newDefinitions = newClient.getFeatureToggleDefinitions() || [];
+        const newDefinitions =
+            this.globalFrontendApiCache.getToggles(token) || [];
```
kwasniew, line 106: we can get all definitions directly from cache since it already does project and env filtering

sjaanus (approved):

PR #6498, opened 2024-03-11
fix(insights): easier chart tooltip

Tymek:
Tooltip on charts should show when hovering anywhere on Y-axis.


Tymek:
Done in another PR

PR #6499, opened 2024-03-11
fix: icon imports

FredrikOseberg:
Based on this article: https://mui.com/material-ui/guides/minimizing-bundle-size/ importing from `'@mui/icons-material'` instead of specifying the actual path to the icon like `import Delete from '@mui/icons-material/Delete';` can be up to six time slower. This change changes all named imports in Unleash referencing the `@mui/icons-material` to default imports. 

This reduced the amount of modules we had to process when building the frontend from 15206 to 4746

Before:
Skjermbilde 2024-03-11 kl  14 19 58

After:
Skjermbilde 2024-03-11 kl  14 20 32

Build time locally decreased by around 50%

Before:
Skjermbilde 2024-03-11 kl  14 31 45

After:
Skjermbilde 2024-03-11 kl  14 27 00


ivarconr (approved):
cool!

kwasniew (approved):

PR #6500, opened 2024-03-11
chore: rename parametric types to make them clearer

gastonfournier:
## About the changes
Just renaming to bring more clarity based on https://github.com/Unleash/unleash/pull/6436#discussion_r1517334796


nunogois (approved):
I'm on board with this. My only comment is that I would like to see `Input` and `Output` come first in the names 🙂

PR #6501, opened 2024-03-11
fix: admin token requests does not automatically have id

chriswk:
To check that users do indeed have permissions to update the roles from project-service, we've been depending on req.user.id. 
We had one error on Friday March 8th, where we managed to send undefined/null to a method that requires a number. This PR assumes that if we have an API token, and we have admin permissions and userId is not set we're a legacy admin token. As such we do not have a real user id, so this sets the userId on the request to SYSTEM_USER_ID (-1337), to credit someone with the operation.

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -44,6 +44,7 @@ import {
     ProjectUserUpdateRoleEvent,
     RoleName,
     SYSTEM_USER,
+    SYSTEM_USER_ID,
 } from '../../types';
 import {
     IProjectAccessModel,
@@ -701,8 +702,11 @@ export default class ProjectService {
         );
     }
 
-    private isAdmin(roles: IRoleWithProject[]): boolean {
-        return roles.some((r) => r.name === RoleName.ADMIN);
+    private isAdmin(userId: number, roles: IRoleWithProject[]): boolean {
+        return (
+            userId === SYSTEM_USER_ID ||
+            roles.some((r) => r.name === RoleName.ADMIN)
+        );
     }
 
     private isProjectOwner(
@@ -723,7 +727,7 @@ export default class ProjectService {
             projectId,
         );
         if (
-            this.isAdmin(userRoles) ||
+            this.isAdmin(userAddingAccess, userRoles) ||
             this.isProjectOwner(userRoles, projectId)
         ) {
             return true;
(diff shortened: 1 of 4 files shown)


thomasheartman (approved):
Looks alright to me. I'm not entirely convinced that using the system user is the right way to go, but I'm assuming you've got a good reason for that (besides, we don't really have a special "general admin user id", do we?).

src/lib/features/project/project-service.ts
```diff
@@ -701,8 +702,11 @@ export default class ProjectService {
         );
     }
 
-    private isAdmin(roles: IRoleWithProject[]): boolean {
-        return roles.some((r) => r.name === RoleName.ADMIN);
+    private isAdmin(userId: number, roles: IRoleWithProject[]): boolean {
+        return (
+            userId === SYSTEM_USER_ID ||
```
daveleek, line 707: What @thomasheartman mentioned made me think, maybe we should be using ADMIN_TOKEN_USER instead? (-42)

daveleek (approved):
Looks good! Tiny question that builds on @thomasheartman s question though, maybe use the token admin user for consistency

src/lib/features/project/project-service.ts
sighphyre: What problem does that solve?

chriswk:
> Looks alright to me. I'm not entirely convinced that using the system user is the right way to go, but I'm assuming you've got a good reason for that (besides, we don't really have a special "general admin user id", do we?).

Exactly, there's no user id for admin tokens, and as such, admin tokens does kinda belong to the system user

chriswk:
> Any chance of a test for this? You're free to tell me this doesn't need one but it looks like it does at a glance, if for nothing else but to keep it from back again
Added two tests for the rbac middleware. Thanks for reminding me. My mind went to "I've got lots of test for the permission work already". But I realised that I wasn't actually testing that I get a `req.user.id` for admin tokens. Now in place.

chriswk:
> > Looks alright to me. I'm not entirely convinced that using the system user is the right way to go, but I'm assuming you've got a good reason for that (besides, we don't really have a special "general admin user id", do we?).
> 
> Exactly, there's no user id for admin tokens, and as such, admin tokens does kinda belong to the system user

I found extractUserId, so if we have the api token middleware, we have a separate admin_token_user with an id of -42, which will be used before we use -1337 if we can't find the data put there from the apiTokenMiddleware.

PR #6503, opened 2024-03-11
fix: check for project in the action permissions check

nunogois:
Fixes the action permission check by also considering the project in the check.


gastonfournier (approved):
LG!

PR #6514, opened 2024-03-12
feat: start tracking operation duration

sjaanus:


src/lib/features/frontend-api/frontend-api-service.ts
```diff
@@ -61,6 +63,8 @@ export class FrontendApiService {
 
     private cachedFrontendSettings?: FrontendSettings;
 
+    private timer: (string) => any;
```
kwasniew, line 66: I think we use Function type in other places

PR #6515, opened 2024-03-12
test: compare frontend api implementations

kwasniew:
## About the changes

Service test that compares frontend API implementations. We check returned features and events emitted.

I tried to cover interesting cases with multiple projects and multiple environments.


sjaanus (approved):
LG

PR #6517, opened 2024-03-12
feat: start monitoring total time to update cache

sjaanus:


src/lib/metrics.ts
```diff
@@ -85,10 +85,10 @@ export default class MetricsMonitor {
             maxAgeSeconds: 600,
             ageBuckets: 5,
         });
-        const operationDuration = createSummary({
-            name: 'operation_duration_seconds',
-            help: 'Operation duration time',
-            labelNames: ['operationId'],
+        const methodDuration = createSummary({
```
kwasniew, line 88: why do you rename it to method? Methods exist on objects. If we wanted to time a regular function method wouldn't make sense anymore.

kwasniew (approved):

PR #6518, opened 2024-03-12
fix: path metric labels (#6400)

gastonfournier:
## About the changes
Some of our metrics are not labeled correctly, one example is `<base-path>/api/frontend/client/metrics` is labeled as `/client/metrics`. We can see that in internal-backstage/prometheus:

image

This issue affects all endpoints that fail to validate the request body. Also, endpoints that are rejected by the authorization-middleware or the api-token-middleware are reported as `(hidden)`.

To gain more insights on our api usage but being protective of metrics cardinality we're prefixing `(hidden)` with some well known base urls: https://github.com/Unleash/unleash/pull/6400/files#diff-1ed998ca46ffc97c9c0d5d400bfd982dbffdb3004b78a230a8a38e7644eee9b6R17-R33

## How to reproduce:
Make an invalid call to metrics (e.g. stop set to null), then check /internal-backstage/prometheus and find the 400 error. Expected to be at `path="/api/client/metrics"` but will have `path=""`:
```shell
curl -H"Authorization: *:development.unleash-insecure-client-api-token" -H'Content-type: application/json' localhost:4242/api/client/metrics -d '{
  "appName": "bash-test",
  "instanceId": "application-name-dacb1234",
  "environment": "development",
  "bucket": {
    "start": "2023-07-27T11:23:44Z",
    "stop": null,
    "toggles": {
      "myCoolToggle": {
        "yes": 25,
        "no": 42,
        "variants": {
          "blue": 6,
          "green": 15,
          "red": 46
        }
      },
      "myOtherToggle": {
        "yes": 0,
        "no": 100
      }
    }
  }
}'
```

## About the changes

Closes #

### DIFF
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -132,6 +132,7 @@ exports[`should create default config 1`] = `
       "personalAccessTokensKillSwitch": false,
       "proPlanAutoCharge": false,
       "queryMissingTokens": false,
+      "responseTimeMetricsFix": false,
       "responseTimeWithAppNameKillSwitch": false,
       "scimApi": false,
       "sdkReporting": false,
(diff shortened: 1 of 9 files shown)


chriswk (approved):
Nice. LGTM

PR #6519, opened 2024-03-12
task: Make Inactive users an enterprise feature (#6510)

chriswk:
Ivar pointed out to me that this was intended as an enterprise only feature. So this PR makes it an enterprise only feature. Conditionally render the link in the normal user table, and use premium feature component if you happen to hit the route and not be running on the enterprise plan.


gastonfournier (approved):
LGTM!

PR #6522, opened 2024-03-12
chore: re-add max listeners

gastonfournier:
## About the changes
Some tests are reporting this error:

`MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 UPDATE_REVISION listeners added to [ConfigurationRevisionService]. Use emitter.setMaxListeners() to increase limit`

I thought it's because of this change https://github.com/Unleash/unleash/pull/6400#discussion_r1511494201 but it was not.

I've managed to get a trace of the issue following this https://stackoverflow.com/q/62897235/239613 and now we can identify `ConfigurationRevisionService` as one explanation. The reason is that it's a singleton that should be cleaned up after tests.

E.g.: https://github.com/Unleash/unleash/actions/runs/8248332132/job/22558419656?pr=6517#step:8:15

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "prebuild:watch": "yarn run clean",
     "build:watch": "tsc -w --strictNullChecks false",
     "prebuild": "yarn run clean",
-    "test": "NODE_ENV=test PORT=4243 jest",
+    "test": "NODE_ENV=test PORT=4243 node --trace-warnings node_modules/.bin/jest",
     "test:unit": "NODE_ENV=test PORT=4243 jest --testPathIgnorePatterns=src/test/e2e --testPathIgnorePatterns=dist",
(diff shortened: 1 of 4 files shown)


kwasniew (approved):

package.json
```diff
@@ -45,7 +45,7 @@
     "prebuild:watch": "yarn run clean",
     "build:watch": "tsc -w --strictNullChecks false",
     "prebuild": "yarn run clean",
-    "test": "NODE_ENV=test PORT=4243 jest",
+    "test": "NODE_ENV=test PORT=4243 node --trace-warnings node_modules/.bin/jest",
```
gastonfournier, line 48: This gives us the following trace in dev:

Screenshot from 2024-03-12 15-53-47

PR #6523, opened 2024-03-12
Apply biome formatting to the docs website

alvinometric:
The docs were ignored by biome, this fixes it. 
With this change
- The docs are no longer ignored by the formatter and the linter
- Custom components in the docs were updated to satisfy the linter
- All the files in the docs were formatted.

There is no new feature or change.


website/global.js
thomasheartman, line 1: Hmm, I'm a little unsure about this one. I'm guessing this is sent down the wire to users, so I'd actually like to keep this one the way it was and added to the ignore file. It's not a massive file, but this isn't something any of us should be touching anyway, so leaving it all on one line is also an indication that this is not to be touched.

thomasheartman (approved):

biome.json
```diff
@@ -68,22 +68,14 @@
       "node_modules",
       "docker",
       "bundle.js",
-      "website/blog",
-      "website/build",
```
thomasheartman, line 72: I'd actually keep this one, if you don't mind. There's no point running biome on the generated output, I think.  Even if that's an edge case.

biome.json
thomasheartman, line 1: Huh,. out of all of these deleted ignores, isn't `website/docs` the **only** directory we actually have (except for build, which exists locally)? Weird that the rest of these ignores are added 🤔

biome.json
alvinometric: Good catch, for some reason I thought it would build to the `.docusaurus` folder

website/global.js
alvinometric: Good call, I agree 🙏

biome.json
alvinometric: There are a bunch of config files that are `.js`, and there are a bunch of custom components too

thomasheartman: Oh, yeah, sorry, I meant specifically for that first block of deletes (from website/blog to website). I just moved my comment and forgot to update that 😅

PR #6531, opened 2024-03-13
chore: flatten payload util

nunogois:
https://linear.app/unleash/issue/2-2029/support-filtering-on-nested-properties

This will allow us to support and suggest payload subproperties in the action filters.


gastonfournier (approved):
Looks good! Let's test it out

PR #6533, opened 2024-03-13
chore: suggest nested properties in action filters

nunogois:
https://linear.app/unleash/issue/2-2029/support-filtering-on-nested-properties

Suggests nested properties in action filters. Also sorts them alphabetically.

Follow up to https://github.com/Unleash/unleash/pull/6531

image


gastonfournier (approved):
LG!

PR #6537, opened 2024-03-13
Feat: sticky insights header and widget tooltip icon

andreas-unleash:
What it says on the tin

Closes # [1-2182](https://linear.app/unleash/issue/1-2182/sticky-header-on-scroll)

Screenshot 2024-03-14 at 10 47 32

https://github.com/Unleash/unleash/assets/104830839/f5249fd6-a4cc-4e52-9b01-89ef3cbeb47c

### DIFF
--- a/frontend/src/component/executiveDashboard/ExecutiveDashboard.tsx
+++ b/frontend/src/component/executiveDashboard/ExecutiveDashboard.tsx
@@ -49,8 +49,7 @@ const StickyWrapper = styled(Box)(({ theme }) => ({
     position: 'sticky',
     top: 0,
     zIndex: 1000,
-    display: 'block',
-    padding: theme.spacing(1,0,4),
+    padding: theme.spacing(2,0,4),
     background: theme.palette.background.application,
 }));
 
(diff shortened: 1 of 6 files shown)


frontend/src/component/executiveDashboard/ExecutiveDashboard.tsx
```diff
@@ -45,7 +45,19 @@ const ChartWidget = styled(Widget)(({ theme }) => ({
     },
 }));
 
+const StickyWrapper = styled(Box)<{ scrolled?: boolean }>(
+    ({ theme, scrolled }) => ({
+        position: 'sticky',
+        top: 0,
+        zIndex: 1000,
+        padding: scrolled ? theme.spacing(2, 0) : theme.spacing(0, 0, 2),
```
FredrikOseberg, line 53: I'm curious what this does in practice? Why does the padding need to be different for the header once we scroll?

FredrikOseberg (approved):
LGTM, but I'm curious about the complexity needed to record the scrolling

frontend/src/component/executiveDashboard/ExecutiveDashboard.tsx
andreas-unleash: When this gets pinned to to the top it looks weird with 0 padding (the search input label is almost half visible).  When you are at the top - even the 2 px padding makes it inconsistent with other pages with empty breadcrumbs.  This with the small transition makes it look ok. I added a video of how it looks

PR #6539, opened 2024-03-13
feat: outdated sdks api

kwasniew:
## About the changes

API for returning outdated SDKs and apps using those SDKs

### DIFF
--- a/src/lib/db/client-instance-store.ts
+++ b/src/lib/db/client-instance-store.ts
@@ -205,6 +205,20 @@ export default class ClientInstanceStore implements IClientInstanceStore {
         return rows.map(mapRow);
     }
 
+    async groupApplicationsBySdk(): Promise<
+        { sdkVersion: string; applications: string[] }[]
+    > {
+        const rows = await this.db
+            .select([
+                'sdk_version as sdkVersion',
+                this.db.raw('ARRAY_AGG(DISTINCT app_name) as applications'),
+            ])
+            .from(TABLE)
+            .groupBy('sdk_version');
+
+        return rows;
+    }
+
     async getDistinctApplications(): Promise<string[]> {
         const rows = await this.db
             .distinct('app_name')
--- a/src/lib/features/metrics/instance/findOutdatedSdks.ts
+++ b/src/lib/features/metrics/instance/findOutdatedSdks.ts
@@ -14,16 +14,18 @@ const config: SDKConfig = {
     'unleash-client-php': '1.13.0',
 };
 
+export const isOutdatedSdk = (sdkVersion: string) => {
+    const result = sdkVersion.split(':');
+    if (result.length !== 2) return false;
+    const [sdkName, version] = result;
+    const minVersion = config[sdkName];
+    if (!minVersion) return false;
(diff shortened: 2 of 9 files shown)


src/lib/db/client-instance-store.ts
```diff
@@ -205,6 +205,20 @@ export default class ClientInstanceStore implements IClientInstanceStore {
         return rows.map(mapRow);
     }
 
+    async groupApplicationsBySdk(): Promise<
+        { sdkVersion: string; applications: string[] }[]
+    > {
+        const rows = await this.db
+            .select([
+                'sdk_version as sdkVersion',
```
kwasniew, line 213: I prefer to use alias to mapRow

src/test/fixtures/fake-client-instance-store.ts
```diff
@@ -32,10 +33,19 @@ export default class FakeClientInstanceStore implements IClientInstanceStore {
     }
 
     async getBySdkName(sdkName: string): Promise<IClientInstance[]> {
-        return Promise.resolve(
-            this.instances.filter((instance) =>
-                instance.sdkVersion?.startsWith(sdkName),
-            ),
+        return this.instances.filter((instance) =>
```
kwasniew, line 36: removing unnecessary promise since async already adds one

src/lib/types/stores/client-instance-store.ts
```diff
@@ -26,6 +26,9 @@ export interface IClientInstanceStore
         environment: string,
     ): Promise<IClientInstance[]>;
     getBySdkName(sdkName: string): Promise<IClientInstance[]>;
+    groupApplicationsBySdk(): Promise<
```
kwasniew, line 29: following existing pattern. We can consider splitting into read models when we move to feature oriented architecture

src/lib/db/client-instance-store.ts
```diff
@@ -205,6 +205,20 @@ export default class ClientInstanceStore implements IClientInstanceStore {
...
+                this.db.raw('ARRAY_AGG(DISTINCT app_name) as applications'),
```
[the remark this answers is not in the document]
sjaanus: Array agg is historically a slow db function, but if the tradeoff is good, it is fine. It will only perform bad with too much data.

src/lib/features/metrics/instance/findOutdatedSdks.ts
```diff
@@ -14,16 +14,18 @@ const config: SDKConfig = {
     'unleash-client-php': '1.13.0',
 };
 
+export const isOutdatedSdk = (sdkVersion: string) => {
```
sjaanus, line 17: usually our files are not camel case

sjaanus (approved):

src/lib/features/metrics/instance/findOutdatedSdks.ts
kwasniew: this is a good point. I usually use - when we use classes and camel case when the file only has functions

kwasniew: but maybe we should move all files in backend code to use dashes

PR #6540, opened 2024-03-13
feat: making banner more composable

kwasniew:
## About the changes

Allowing ReactNode children in banner dialog so that we can style not just from the UI markdown but also from the application code. My current use case is listing all outdated sdks with applications using those SDKs. So I will be building double nested list with links to applications. I want to use regular styled React nodes for those.


frontend/src/component/banners/OutdatedSdksBanner/OutdatedSdksBanner.tsx
```diff
@@ -0,0 +1,25 @@
...
+        linkText: 'Please update those versions',
+        dialogTitle: 'Outdated SDKs',
+        dialog: <h1>Outdated SDKs</h1>,
+    };
+    return (
+        <>
+            <ConditionallyRender
+                condition={displayOutdatedSdksBanner}
+                show={
+                    <Banner key={'upgradeSdks'} banner={outdatedSdksBanner} />
```
nunogois, line 20: We probably don't need a key here 🤔

frontend/src/component/admin/banners/BannerModal/BannerModal.tsx
```diff
@@ -61,7 +61,9 @@ export const BannerModal = ({ banner, open, setOpen }: IBannerModalProps) => {
         setLink(banner?.link || '');
         setLinkText(banner?.linkText || '');
         setDialogTitle(banner?.dialogTitle || '');
-        setDialog(banner?.dialog || '');
+        if (typeof banner?.dialog === 'string') {
+            setDialog(banner?.dialog || '');
+        }
```
nunogois, line 66: Maybe `setDialog(typeof banner?.dialog === 'string' ? banner?.dialog : '');` instead? Otherwise I think we're not really emptying this value when creating a new banner.

frontend/src/component/banners/OutdatedSdksBanner/OutdatedSdksBanner.tsx
kwasniew: copy paste from another banner :)

nunogois (approved):
Pretty cool 🙌

PR #6541, opened 2024-03-13
feat: show outdated sdks banner

kwasniew:
## About the changes

Added banner with a list of outdated SDKs and applications using those SDKs

Screenshot 2024-03-13 at 16 37 51
Screenshot 2024-03-13 at 16 55 43

Next actions:
* I'd improve the stylings of the list
* When clicking a link the modal doesn't disappear

## Discussion points

Probably we need a separate flag from the sdkReporting since we already exposed sdkReporting flag to some users.

I found we have similar banner that was created for edge. Do we need to keep both?


frontend/src/hooks/api/getters/useOutdatedSdks/useOutdatedSdks.ts
```diff
@@ -0,0 +1,14 @@
+import { fetcher, useApiGetter } from '../useApiGetter/useApiGetter';
+import { OutdatedSdksSchema } from '../../../../openapi';
+
+const PATH = 'api/admin/metrics/sdks/outdated';
+
+export const useOutdatedSdks = () => {
+    const { data, refetch, loading, error } = useApiGetter<OutdatedSdksSchema>(
```
kwasniew, line 7: I found this hook that saves some typing

frontend/src/hooks/api/getters/useApiGetter/useApiGetter.ts
```diff
@@ -26,3 +27,9 @@ export const useApiGetter = <T>(
         loading: !error && !data,
     };
 };
+
+export const fetcher = (path: string, errorTarget: string) => {
```
kwasniew, line 31: added generic fetcher so that we don't need to repeat it

PR #6545, opened 2024-03-14
refactor: allow user events to take in only what they need

sighphyre:
This changes the user event system type to ingest a partial IUserWithRootRole interface rather than the whole interface. 

This also undeprecates the storeEvent(s) methods.

**Why the new interface**

Usually when creating users we can just sling around the entire User object. That's absolutely fine in cases where we have the entire object including all its permissions. But we certainly don't have all those permissions in an easily accessible way in the SSO/SCIM code.

Luckily, that event object doesn't need or care about those properties. It explicitly extracts only the fields the new interface provides and discards everything else. So we're depending on a wider type than we need. 

**Why the undeprecation**

Some things don't get made by real users. Some things get made by the system user, which both doesn't and probably shouldn't have an email. I can't find the breadcrumbs for why we deprecated that method. There's no explanation in the deprecation warning and the PR that makes the change has no reasoning either

The old deprecated method is used extensively throughout the code and it's hard to know if that code is now broken or why it should be considered broken if it is

Completely open to being told this is wrong and bad, but then I think we need that deprecation warning to have useful details on what danger it causes

### DIFF
--- a/src/lib/features/events/event-service.ts
+++ b/src/lib/features/events/event-service.ts
@@ -105,14 +105,14 @@ export default class EventService {
     }
 
     /**
-     * @deprecated use storeUserEvent instead
+     * For events arising from user actions you should prefer to use storeUserEvents instead
      */
     async storeEvent(event: IBaseEvent): Promise<void> {
         return this.storeEvents([event]);
     }
 
     /**
-     * @deprecated use storeUserEvents instead
+     * For events arising from user actions you should prefer to use storeUserEvents instead
      */
     async storeEvents(events: IBaseEvent[]): Promise<void> {
         let enhancedEvents = events;
--- a/src/lib/types/events.ts
+++ b/src/lib/types/events.ts
@@ -1304,11 +1304,11 @@ export class PotentiallyStaleOnEvent extends BaseEvent {
 }
 
 export class UserCreatedEvent extends BaseEvent {
-    readonly data: IUserWithRootRole;
+    readonly data: IUserEventData;
 
     constructor(eventData: {
         createdBy: string | IUser;
-        userCreated: IUserWithRootRole;
+        userCreated: IUserEventData;
         createdByUserId: number;
     }) {
         super(USER_CREATED, eventData.createdBy, eventData.createdByUserId);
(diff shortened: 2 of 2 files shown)


src/lib/features/events/event-service.ts
```diff
@@ -105,14 +105,14 @@ export default class EventService {
     }
 
     /**
-     * @deprecated use storeUserEvent instead
+     * For events arising from user actions you should prefer to use storeUserEvents instead
```
gastonfournier, line 108: So this means this one is for system action? If we consider system as a user, then shouldn't we use storeUserEvents? I think we want to identify all events with a user

src/lib/types/events.ts
```diff
@@ -1333,19 +1333,27 @@ export class UserUpdatedEvent extends BaseEvent {
...
         this.preData = mapUserToData(eventData.deletedUser);
     }
 }
 
-function mapUserToData(user: IUserWithRootRole): any {
+interface IUserEventData extends Partial<IUserWithRootRole> {
+    id: number;
+    name: string;
+    username: string;
+    email: string;
```
gastonfournier, line 1352: - name vs username, what's the difference?
- email is sometimes optional, I believe the username might be as well, so the type I have in mind is something like `Either<UserWithUsername, UserWithEmail>`

gastonfournier (commented):
Two ideas to talk after standup

sighphyre:
Moving this to a draft because @gastonfournier and I have spoken, we think we can do something better with the methods that replaced the deprecated methods and probably make better

sighphyre:
We spoke more! This is massively hard to do in a better way today because of what we're potentially storing nulls in the db.

This seems like the best incremental change we can make today

src/lib/features/events/event-service.ts
sighphyre: Yessir. This boils down to maybe kinda sorta not having an email. Our types say it always exists. Our database disagrees.

System user is explicitly Omit<IUser, "email"> and this flow requires an email.

Hard to fix because of the db

gastonfournier (approved):
This looks nice. We've identified a few tasks we'd like to address: 

- https://linear.app/unleash/issue/2-2044/we-need-to-get-rid-of-createuserevent-from-event-service
- https://linear.app/unleash/issue/2-2043/username-name-and-email-can-be-null-we-have-to-be-honest-and-make-them

src/lib/types/events.ts
sighphyre: > name vs username, what's the difference?

Good question. No idea. It depends on how the user got into the system and when they got into the system. Legacy flows are slightly different. SSO is slightly different (also depending on provider and setup, you might get name but not username or vice versa). 

I don't believe there _is_ a meaningful difference, only us doing terrible things to the types 

> email is sometimes optional, I believe the username might be as well, so the type I have in mind is something like Either<UserWithUsername, UserWithEmail>

We spoke already. This is for anyone trying to find bread crumbs at a future date (hello future person!). This doesn't work. There's cases where none of these fields are set, even thought the type system says they should be

PR #6548, opened 2024-03-14
refactor: optimize applications overview

sjaanus:
There was no need to join the entire metrics table, as it is a huge table. We only needed all combinations of app_name, environment, and feature_name. The new query retrieves all this data, which will then be joined into the main query.

### DIFF
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -294,7 +294,6 @@ export default class ClientApplicationsStore
         const query = this.db
             .with('latest_metrics', (qb) => {
                 qb.select('cme.app_name', 'cme.environment', 'cme.feature_name')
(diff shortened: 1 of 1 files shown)


src/lib/db/client-applications-store.ts
```diff
@@ -292,6 +292,15 @@ export default class ClientApplicationsStore
         appName: string,
     ): Promise<IApplicationOverview> {
         const query = this.db
+            .with('latest_metrics', (qb) => {
+                qb.select('cme.app_name', 'cme.environment', 'cme.feature_name')
+                    .from('client_metrics_env as cme')
+                    .groupBy(
```
kwasniew, line 298: distinct?

kwasniew (approved):

Issue #6542, opened 2024-03-13
Context properties types


vrshgv:
### Describe the feature request

It's not good that it's impossible to update unleash custom context fields with anything except string. Not even a number. Why would I need all these operators for constraints, when there is no chance to use them with my context fields.

thomasheartman:
Hi, @vrshgv!

Good question. Unleash only accepts strings for context values, but it will try to convert the string to the appropriate value when checking a flag's constraints, so if you use a numeric greater than operator for the value `"5"`, it will convert that value to the number 5 before doing the conversion. If the value can't be converted, the constraint will fail. 

Does that answer your question?

vrshgv:
Hello! Yes, it does, thank you. I guess it was not obvious for me in documentation.

thomasheartman:
Great! And that's great feedback about the documentation. I'll see if we can't update that 😄 

It seems your issue has been resolved, so I'll go ahead and close the issue now. Feel free to open it again if you feel there's more to be done.

PR #6551, opened 2024-03-14
docs: add notes about how Unleash deals with string context values and conversion

thomasheartman:
This change addresses the question in #6542 about using non-string
operators on string values by clarifying that Unleash SDK converts the
string values before performing the comparison.


thomasheartman:
@vrshgv Would this change to the docs have made things clearer for you?

sighphyre (approved):
Thank you, very nice!

PR #6553, opened 2024-03-14
fix: now issue count under application will show correctly

sjaanus:
Previously if multiple environments had same missing SDK, it was counted multiple times. Now it will take only unique.


frontend/src/component/application/ApplicationIssues/ApplicationIssues.tsx
```diff
@@ -230,7 +230,10 @@ export const ApplicationIssues = ({ application }: IApplicationIssuesProps) => {
             ),
         ),
     ];
-    const issueCount = mode.issueCount;
```
kwasniew, line 233: what about the code in the getApplicationIssueMode that was giving incorrect results? Should we delete it or move the new fixed code there?

kwasniew (approved):

PR #6582, opened 2024-03-15
fix: action select value

nunogois:
This should fix the action select bug where it shows as empty when the form loads.

### DIFF
--- a/frontend/src/component/project/Project/ProjectSettings/ProjectActions/ProjectActionsTable/ProjectActionsModal/ProjectActionsForm/ProjectActionsFormStep/ProjectActionsFormStepActions/ProjectActionsActionSelect.tsx
+++ b/frontend/src/component/project/Project/ProjectSettings/ProjectActions/ProjectActionsTable/ProjectActionsModal/ProjectActionsForm/ProjectActionsFormStep/ProjectActionsFormStepActions/ProjectActionsActionSelect.tsx
@@ -45,7 +45,7 @@ export const ProjectActionsActionSelect = ({
             options={actionOptions}
             autoHighlight
             autoSelect
-            value={actionOptions.find(({ key }) => key === value)}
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/Project/ProjectSettings/ProjectActions/ProjectActionsTable/ProjectActionsModal/ProjectActionsForm/ProjectActionsFormStep/ProjectActionsFormStepActions/ProjectActionsActionSelect.tsx
```diff
@@ -45,7 +45,7 @@ export const ProjectActionsActionSelect = ({
             options={actionOptions}
             autoHighlight
             autoSelect
-            value={actionOptions.find(({ key }) => key === value)}
+            value={actionOptions.find(({ key }) => key === value) || null}
```
gastonfournier, line 48: Is this fixing the issue? I thought that the value was unset and we're not able to set it

gastonfournier (approved):
LG but I'm not sure it fixes it

frontend/src/component/project/Project/ProjectSettings/ProjectActions/ProjectActionsTable/ProjectActionsModal/ProjectActionsForm/ProjectActionsFormStep/ProjectActionsFormStepActions/ProjectActionsActionSelect.tsx
nunogois: Should fix the issue from some quick tests. The issue was MUI being weird to the value being `undefined`. It's OK with `null`s however, since that just means that the field is empty.

PR #6586, opened 2024-03-15
fix: outdated sdk version handles non semver versions

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/features/metrics/instance/findOutdatedSdks.ts
+++ b/src/lib/features/metrics/instance/findOutdatedSdks.ts
@@ -21,6 +21,7 @@ export const isOutdatedSdk = (sdkVersion: string | null) => {
     const [sdkName, version] = result;
     const minVersion = config[sdkName];
     if (!minVersion) return false;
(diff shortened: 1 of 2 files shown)


sjaanus (approved):

src/lib/features/metrics/instance/findOutdatedSdks.ts
```diff
@@ -21,6 +21,7 @@ export const isOutdatedSdk = (sdkVersion: string | null) => {
     const [sdkName, version] = result;
     const minVersion = config[sdkName];
     if (!minVersion) return false;
+    if (!semver.valid(version)) return false;
     if (semver.lt(version, minVersion)) return true;
     return false;
```
chriswk, line 26: Just for my understanding, this is basically
`return minVersion && semver.valid(version) && semver.lt(version, minVersion)`
Just with explicit returns for each condition?

kwasniew: spot on. Another PR then :)

PR #6588, opened 2024-03-15
chore: redesign action configurations

nunogois:
https://linear.app/unleash/issue/2-2050/redesign-action-definitions-oss

Depends on a PR on Enterprise that will provide a new endpoint with action configurations.

We should also clean up the now unnecessary constants and hook.


gastonfournier (approved):
Looks nice :)

PR #6589, opened 2024-03-18
chore: Bump biome and configure husky

chriswk:
Upgrades biome to 1.6.1, and updates husky pre-commit hook.

Most changes here are making type imports explicit.


biome.json
```diff
@@ -20,21 +20,21 @@
...
+        "noConfusingVoidType": "off",
+        "noArrayIndexKey": "off",
+        "noThenProperty": "off"
       },
       "performance": {
         "noAccumulatingSpread": "off",
         "noDelete": "off"
-      },
-      "nursery": {
-        "noUnusedImports": "warn"
```
nunogois, line 37: Did you mean to remove this? Maybe you meant to move it instead?

nunogois (approved):
Other than the fact that this PR killed my Chrome a dozen times, I think it's great 😄

PR #6590, opened 2024-03-18
fix: project settings selected tab bug

nunogois:
https://linear.app/unleash/issue/2-2034/project-settings-url-bug

This prevents a bug where the wrong tab would show as selected, only present in an edge case where the project has the name of a settings tab.

### DIFF
--- a/frontend/src/component/project/Project/ProjectSettings/ProjectSettings.tsx
+++ b/frontend/src/component/project/Project/ProjectSettings/ProjectSettings.tsx
@@ -93,7 +93,8 @@ export const ProjectSettings = () => {
             tabs={tabs}
             value={
                 tabs.find(
-                    ({ id }) => id && location.pathname?.includes(`/${id}`),
+                    ({ id }) =>
+                        id && location.pathname?.includes(`/settings/${id}`),
                 )?.id || tabs[0].id
             }
             onChange={onChange}


frontend/src/component/project/Project/ProjectSettings/ProjectSettings.tsx
```diff
@@ -93,7 +93,8 @@ export const ProjectSettings = () => {
             tabs={tabs}
             value={
                 tabs.find(
-                    ({ id }) => id && location.pathname?.includes(`/${id}`),
+                    ({ id }) =>
+                        id && location.pathname?.includes(`/settings/${id}`),
```
gastonfournier, line 97: Maybe ends with would be more accurate?

gastonfournier (approved):
Looks fine but can we be even more precise?

frontend/src/component/project/Project/ProjectSettings/ProjectSettings.tsx
nunogois: That would be tricky since some of the settings pages have deeper pathing like sub-pages. The more precise solution would be using regex. Think it's worth a try?

gastonfournier: No, then let's use this, after all including the / in the matching this issue should be solved for good, because we don't allow feature names or projects to have slashes

gastonfournier: Yes, confirmed: 
image

PR #6593, opened 2024-03-18
fix: invalid HTML and react errors in insights pages

thomasheartman:
This PR fixes these errors (that were showing up in the dev console) in the insights pages:
- nesting a div within a p in the count header (flags, environments, apps); instead flip the relationship and nest the p within the div
- missing keys in mapped components
- passing a boolean "scrolled" value to the underlying component (a div) is invalid: instead, make it so that that prop is not passed

The only one of these that could have a visual impact is the first one (p>div -> div>p), but it appears to be the same to me.

Here's before the change:
image

And here's after:
image


thomasheartman:
Err, minor update: found that I might've attacked the wrong boolean prop? I'll check

thomasheartman:
Nope! Just solved it in the wrong way.

PR #6594, opened 2024-03-18
feat: project health chart

kwasniew:
## About the changes

Screenshot 2024-03-18 at 14 32 31

A circle chart for active, stale and potentially stale flags.

Important points:
* uses svg circles
* adds artificial gap between active and stale
* potentially stale flags are subset of active

Next PR will add a legend and the whole tile.


sjaanus (approved):
LG

PR #6601, opened 2024-03-19
fix: set log level to error for db migrate

FredrikOseberg:
This PR sets the log level of db-migrate to error, because it interferes with the indexing of our logs in Loki when the logs are not in JSON format.


chriswk (approved):
LGTM

gastonfournier (approved):
:+1:

PR #6603, opened 2024-03-19
feat: add configurable event for banner goto action

sjaanus:
We added banner to get attention of customer, but we are not tracking its usage.
Added a way to track the **goto action** button.


frontend/src/component/banners/OutdatedSdksBanner/OutdatedSdksBanner.tsx
```diff
@@ -27,7 +46,11 @@ export const OutdatedSdksBanner = () => {
                         <span>{item.sdkVersion}</span>
                         <StyledList>
                             {item.applications.map((application) => (
-                                <li key={application}>
+                                <li
+                                    key={application}
+                                    onClick={bannerClicked}
+                                    onKeyDown={bannerClicked}
```
sjaanus, line 52: for keyboard navigation

kwasniew (approved):
good catch

PR #6609, opened 2024-03-19
fix: authorization header typo

gastonfournier:
## About the changes
There seems to be a typo in the authorization header. We're keeping the old typo as preferred just in case, but if not present we'll default to the authorization header (not authorisation).

Not sure about the impact of this bug, as all registrations might be using default project.


andreas-unleash (approved):
LG

PR #6611, opened 2024-03-19
chore: call signal endpoint example

nunogois:
Adds a "call signal endpoint" example to the token dialog.

image


gastonfournier (approved):
LGTM!

PR #6623, opened 2024-03-20
fix: use defaults when creating gradualRollout strategies

chriswk:
Via the API you can currently create gradualRollout strategies without any parameters set, when visiting the UI afterwards, you can edit this, because the UI reads the parameter list from the database and sees that some parameters are required, and refuses to accept the data. This PR adds defaults for gradualRollout strategies created from the API, making sure gradual rollout strategies always have `rollout`, `groupId` and `stickiness` set.

My one worry here is if we should take the time to check if default stickiness is disabled. I think it makes sense to default to default here; but I'm open to suggestions.

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
@@ -24,8 +24,9 @@ import FeatureToggleStore from './feature-toggle-store';
 import { ensureStringValue, mapValues } from '../../util';
 import type { IFeatureProjectUserParams } from './feature-toggle-controller';
 import type { Db } from '../../db/db';
-import Raw = Knex.Raw;
 import { isAfter } from 'date-fns';
+import merge from 'deepmerge';
+import Raw = Knex.Raw;
 
 const COLUMNS = [
     'id',
@@ -119,6 +120,10 @@ const sortEnvironments = (overview: IFeatureOverview) => {
     }));
 };
 
+function mergeAll<T>(objects: Partial<T>[]): T {
+    return merge.all<T>(objects.filter((i) => i));
+}
+
 interface StrategyUpdate {
     strategy_name: string;
     parameters: object;
@@ -148,6 +153,19 @@ function mapStrategyUpdate(
     update.variants = JSON.stringify(input.variants || []);
     return update;
 }
+
+const defaultParameters = (featureName: string, strategyType: string) => {
+    if (strategyType === 'gradualRollout') {
+        return {
+            rollout: '100',
+            stickiness: 'default',
+            groupId: featureName,
+        };
+    } else {
+        /// We don't really have good defaults for the other kinds of known strategies, so return an empty map.
+        return {};
+    }
+};
 
 class FeatureStrategiesStore implements IFeatureStrategiesStore {
     private db: Db;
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -119,6 +120,10 @@ const sortEnvironments = (overview: IFeatureOverview) => {
     }));
 };
 
+function mergeAll<T>(objects: Partial<T>[]): T {
+    return merge.all<T>(objects.filter((i) => i));
+}
+
```
chriswk, line 126: hmm, we already have this in create-config as well. Should we extract it to a helper? Where to put it, types/index ? Create a merger.ts file?

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
```diff
@@ -149,6 +150,30 @@ function mapStrategyUpdate(
...
+    params: PartialSome<IFeatureStrategy, 'id' | 'createdAt'>,
+) => {
+    if (params.strategyName === 'gradualRollout') {
+        return {
+            rollout: '100',
+            stickiness: 'default',
+            groupId: params.featureName,
+        };
+    } else {
+        /// We don't really have good defaults for the other kinds of known strategies, so return an empty map.
```
sighphyre, line 167: Honestly it might be worth returning an error at this point. There's no way you can provide sane defaults for something like userWithIdStrategy, but that is definitely broken without config

I think this probably covered 99% of stuff though. Gut feeling is most people use either the default gradual rollout for most things and standard for some edge cases (last 1% is weird things like IP address strategies but those should fail to create without config)

sighphyre (approved):
LGTM

I think we should consider making borked strategies other than gradualrollout a failure rather than a quiet ingestion of broken data but this does improve the current situation a lot!

thomasheartman:
Just chiming in with some context:
I think solving this for gradual rollouts is great. But I agree with @sighphyre that we should also cover other strategies where we can. Whether we just flat out fail or fall back to defaults is up for discussion (though I'm certainly in favor of defaulting where we can to keep things working).

That said, that is probably better solved in a later PR (or multiple).

thomasheartman:
And yes, we definitely should check whether the `default` stickiness is enabled.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
chriswk: Is `{ "userIds": "" }` any better than it just being null, how does the frontend handle that?

chriswk:
> And yes, we definitely should check whether the `default` stickiness is enabled.

I'll see how I can do that before merging this then.

src/lib/features/feature-toggle/feature-toggle-strategies-store.ts
thomasheartman: No idea how the front end handles parameters being undefined for user id strategy, but having no user IDs specified is totally valid 💁🏼

Let me see what happens if I create it via the API.

thomasheartman: The UI doesn't appear to care in this case. It treats an empty parameters the same as having an empty list of user ids. This is good in the functional sense, but we could still fix the data for the strategy before inserting it 🤷🏼  I dunno. Doesn't matter much for this, anyway (because it works), but it feels like a better thing(TM) to store something with the required configuration data than to just assume that any clients will auto-fill it later.

Issue #4630, opened 2023-09-07
Authorization header use not conforming to HTTP standards


vkrizan:
### Describe the bug

`Authorization` header use with API tokens is out of HTTP standards of [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#section-11.4). The header lacks `auth-scheme`.

### Expected behavior

`auth-scheme` used in `Authoriztion` header. For API keys this could be `Bearer`.

### Additional context

* https://www.rfc-editor.org/rfc/rfc9110#section-11.4
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

### Unleash version

5.4.2

### Subscription type

Open source

### Hosting type

Self-hosted

nunogois:
Hi @vkrizan, thank you for raising this issue. I'll share this with the rest and then we'll get back to you.

sighphyre:
Hey @vkrizan, our auth protocol looks a lot like RFC 9110 but it isn't, it's a custom protocol we use. This isn't something that's likely to change in the near future since it's pretty deeply baked in at the moment. Is this actually causing problems for you?

vkrizan:
Thank you for your response.

There are no problems considering the non-standard Authorization header used at the moment. I've raised it, as my assumption about the use of API keys (with the use of "Bearer") against Unleash were incorrect. Documentation about the use is only to be found in the [Unleash API spec](https://docs.getunleash.io/reference/api/unleash/unleash-api#authentication) and on some examples.

I'm a bit concerned that not using the standard can lead to problems when implementing custom authentication methods, or at least create a tech-debt. Our folks have re-implemented a part of the toggle APIs with the Bearer support, with a fallback to using the header value as is: https://github.com/app-sre/unleash/blob/1a232f991ff224b634fd75b9da375b83e65d438f/index.js#L183.

I'd maybe suggest to start migrate to the standard, and have the fallback to header value to not break existing clients.
If you think that this is not such a big concern, feel free to close.

sighphyre:
Hey @vkrizan, gotta be honest, I was kind of hoping you'd point out a bunch of places that this caused huge problems. It's always felt a little weird to me that we do this but I've never had any real argument to change it and it always felt like a big, fragile change to everything that needs to auth to Unleash. 

The code you've linked is actually a pretty neat idea, that would make this safe, I think.

[...]

PR #6624, opened 2024-03-20
chore: bearer token middleware

nunogois:
Adds a bearer token middleware that adds support for tokens prefixed with "Bearer" scheme. Prefixing with "Bearer" is optional and the old way of authenticating still works, so we now support both ways.

Also, added as part of our OpenAPI spec which now displays authorization as follows:
image

Related to #4630. Doesn't fully close the issue as we're still using some invalid characters for the RFC, in particular `*` and `[]`

For safety reasons this is behind a feature flag

### DIFF
--- a/src/lib/middleware/bearer-token-middleware.ts
+++ b/src/lib/middleware/bearer-token-middleware.ts
@@ -1,15 +1,24 @@
 import type { Request, Response, NextFunction } from 'express';
+import type { IUnleashConfig } from '../types';
 
-export const bearerTokenMiddleware = (
-    req: Request,
-    _: Response,
-    next: NextFunction,
-) => {
-    const authHeader = req.headers.authorization;
+export const bearerTokenMiddleware = ({
+    getLogger,
+    flagResolver,
+}: Pick<IUnleashConfig, 'getLogger' | 'flagResolver'>) => {
+    const logger = getLogger('/middleware/bearer-token-middleware.ts');
+    logger.debug('Enabling bearer token middleware');
 
-    if (authHeader) {
-        req.headers.authorization = authHeader.replace(/^Bearer\s+/i, '');
-    }
+    return (req: Request, _: Response, next: NextFunction) => {
+        if (flagResolver.isEnabled('bearerTokenMiddleware')) {
+            const authHeader = req.headers.authorization;
 
-    next();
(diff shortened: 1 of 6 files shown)


gastonfournier (approved):
@nunogois this looks great! I'm confident cause we have a toggle switch it off at will. So I'd say go for it after rebasing (just in case).

src/lib/middleware/bearer-token-middleware.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const bearerTokenMiddleware = ({
+    getLogger,
+    flagResolver,
+}: Pick<IUnleashConfig, 'getLogger' | 'flagResolver'>) => {
+    const logger = getLogger('/middleware/bearer-token-middleware.ts');
+    logger.debug('Enabling bearer token middleware');
+
+    return (req: Request, _: Response, next: NextFunction) => {
+        if (flagResolver.isEnabled('bearerTokenMiddleware')) {
```
gastonfournier, line 12: We should have it enabled for signals as it's an experimental endpoint and we do want to support Bearer tokens in this endpoint, while the others we want to have it as a release flag. Should there be any security related issue with Bearer tokens in signals we can just kill the whole functionality, and that should be fine because it's experimental

src/lib/middleware/bearer-token-middleware.ts
```diff
@@ -0,0 +1,24 @@
+import type { Request, Response, NextFunction } from 'express';
+import type { IUnleashConfig } from '../types';
+
+export const bearerTokenMiddleware = ({
+    getLogger,
+    flagResolver,
+}: Pick<IUnleashConfig, 'getLogger' | 'flagResolver'>) => {
+    const logger = getLogger('/middleware/bearer-token-middleware.ts');
+    logger.debug('Enabling bearer token middleware');
```
gastonfournier, line 9: Maybe we can remove this line
```suggestion
```

src/lib/middleware/bearer-token-middleware.ts
nunogois: Addressed in https://github.com/Unleash/unleash/pull/6624/commits/9a558dffd6b76baa51191ba18510c0cc2e0cc2cf

PR #6626, opened 2024-03-20
fix: ttp stat show latest date

andreas-unleash:
Makes the time to production stats show the current value (latest date).  This aligns it with other stats on the page.

Closes # [1-2203](https://linear.app/unleash/issue/1-2203/make-the-time-to-production-stats-show-the-latest-current-data)

### DIFF
--- a/frontend/src/component/executiveDashboard/hooks/useAvgTimeToProduction.ts
+++ b/frontend/src/component/executiveDashboard/hooks/useAvgTimeToProduction.ts
@@ -1,6 +1,12 @@
 import { useMemo } from 'react';
-import type { ExecutiveSummarySchema } from 'openapi';
+import type {
+    ExecutiveSummarySchema,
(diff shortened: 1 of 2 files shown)


frontend/src/component/executiveDashboard/hooks/useAvgTimeToProduction.ts
thomasheartman, line 1: Is the calculation in this file correct? On line 45, we divide the total avg time to production with total projects. But some of the total projects might be filtered out because of missing latest trends, right? Looks like this is why it's failing too.

Would it work to do something like adding all the valid trend values to a list and then summing it and dividing the total by the number of elements in the list?

andreas-unleash: that is correct - invalid trends should reduce the number of total projects - fixed

thomasheartman (approved):

PR #6627, opened 2024-03-20
chore: adapted IUser to our DB schema

gastonfournier:
## About the changes
Our schema allows us to have null username or email. With this change at least we'll get a compiler warning when using either of these properties

Screenshot from 2024-03-20 12-20-45


frontend/src/interfaces/user.ts
```diff
@@ -1,26 +0,0 @@
-export const AccountTypes = ['User', 'Service Account'] as const;
-type AccountType = (typeof AccountTypes)[number];
-
-export interface IUser {
```
gastonfournier, line 4: This can be replaced with the server type from `@server/types`

gastonfournier: This didn't work well because the frontend interface has other fields and it was not easy to fix

sjaanus (approved):

PR #6633, opened 2024-03-20
fix: render values that are N/A

thomasheartman:
This change makes the tooltip still render values and headers that are
`N/A` (instead of not rendering them at all).

This makes the tooltip more consistent and predictable. At least to
me, it was confusing that some of the values were just hidden sometimes.

I've also added a test to make sure that the tooltip renders the N/A values.

This is what it looks like now:

image


andreas-unleash (approved):
LGTM!

PR #6637, opened 2024-03-20
chore: fix react errors

andreas-unleash:
Fix hook usage react error (console)

### DIFF
--- a/frontend/src/component/executiveDashboard/ExecutiveDashboard.tsx
+++ b/frontend/src/component/executiveDashboard/ExecutiveDashboard.tsx
@@ -11,20 +11,19 @@ import { DashboardHeader } from './components/DashboardHeader/DashboardHeader';
 import { useDashboardData } from './hooks/useDashboardData';
 import { Charts } from './Charts';
 
-const StickyWrapper = styled(Box)<{ scrolled?: string }>(
-    ({ theme, scrolled }) => ({
-        position: 'sticky',
-        top: 0,
-        zIndex: 1000,
-        padding:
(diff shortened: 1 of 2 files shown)


frontend/src/component/executiveDashboard/ExecutiveDashboard.tsx
thomasheartman, line 1: This probably isn't the right way to fix this. Instead, you should look at how I solved this in https://github.com/Unleash/unleash/pull/6593/files. I think it's the exact same component (`StickyWrapper`) being fixed. The issue is that `scrolled` is being passed onto the underlying html, so a boolean doesn't work. Instead, we should exclude it by using:
```js
shouldForwardProp: (prop) => prop !== 'scrolled',
```

That way we can keep the boolean and still have valid HTML.

Putting it all together, that means the `StickyWrapper` component should be changed to this:

```ts
const StickyWrapper = styled(Box, {
    shouldForwardProp: (prop) => prop !== 'scrolled',
})<{ scrolled?: boolean }>(({ theme, scrolled }) => ({
    position: 'sticky',
    top: 0,
    zIndex: 1000,
    padding: scrolled ? theme.spacing(2, 0) : theme.spacing(0, 0, 2),
    background: theme.palette.background.application,
    transition: 'padding 0.3s ease',
}));
```

and that you can leave the rest of the file alone.

thomasheartman (approved):
The second file here looks good (`useDashboardData`), but I think the first one has a better solution. Refer to that comment for why I think it's not the right solution and for a link that shows you exactly what you can do.

However, I'll give you an approval so that you can fix it in your own time and merge when it's ready. Thanks for fixing it 😄

PR #6639, opened 2024-03-20
fix: do not show flagsPerUser when calculation results to NaN

andreas-unleash:
What it says on the tin

Closes # [1-2209](https://linear.app/unleash/issue/1-2209/total-flag-widget-says-nan-when-no-data)


frontend/src/component/executiveDashboard/componentsStat/FlagStats/FlagStats.tsx
```diff
@@ -83,7 +83,11 @@ export const FlagStats: React.FC<IFlagStatsProps> = ({
             </StyledRingContainer>
 
             <ConditionallyRender
-                condition={flagsPerUser !== undefined && flagsPerUser !== ''}
+                condition={
+                    flagsPerUser !== undefined &&
+                    flagsPerUser !== '' &&
+                    flagsPerUser !== 'NaN'
```
thomasheartman, line 89: Wait, are you saying that the value we store here the string "NaN"? Or is it the value `NaN`?

I would expect this comparison to be done using the [Number.isNan](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) function, but I don't know what the value we get from the API is in this case? If it is the string, then we should fix this on the API level, right?

frontend/src/component/executiveDashboard/componentsStat/FlagStats/FlagStats.tsx
```diff
@@ -102,7 +106,7 @@ export const FlagStats: React.FC<IFlagStatsProps> = ({
                             </Typography>
                         </StyledTextContainer>
                         <StyledFlagCountPerUser>
-                            {flagsPerUser}
+                            {flagsPerUser ?? 'N/A'}
```
thomasheartman, line 109: So the only time this'll evaluate to `N/A` is if flagsPerUser is `null`, right? Is that a value that we expect some times?

How is flags per user calculated? When will this be null? If there are no flags, we can just say it's 0, can't we?

thomasheartman (commented):
I'm a little unsure about the approach here. I'm not sure it's actually gonna fix the issue (refer to the inline comments). 

Also, might be worth adding a test to check that it renders correctly when given the expected data?

frontend/src/component/executiveDashboard/componentsStat/FlagStats/FlagStats.tsx
andreas-unleash: The value is coming from a calculation parsed to a string. So yes the value that gets here is  'NaN' and because of the other string comparisons here, I added it here for consistency
```  <FlagStats
                        count={summary.total}
                        flagsPerUser={
                            showAllProjects
                                ? (summary.total / users.total).toFixed(2)
                                : ''
                        }
                    />```

File: frontend/src/component/executiveDashboard/componentsStat/FlagStats/FlagStats.tsx (hunk 2 above)
Comment (thomasheartman) [reply frontend/src/component/executiveDashboard/componentsStat/FlagStats/FlagStats.tsx:105]: If `flagsPerUser` has been changed to always being a string now, can we remove the fallback? It will never happen, so leaving it in is a little misleading, I think.
```suggestion
                            {flagsPerUser}
```

frontend/src/component/executiveDashboard/hooks/useFilteredFlagsSummary.ts
```diff
@@ -38,8 +42,14 @@ export const useFilteredFlagsSummary = (
             },
         );
 
+        const flagsPerUser = users?.total ? sum.total / users.total : 0;
+
         return {
             ...sum,
+            flagsPerUser:
+                Number.isNaN(flagsPerUser) || flagsPerUser === 0
```
thomasheartman, line 50: Why are we using "N/A" if flags per user is 0? If there are 0 flags, then 0 flags per user is an entirely valid value, isn't it?

frontend/src/component/executiveDashboard/Charts.tsx
```diff
@@ -124,9 +125,7 @@ export const Charts: VFC<IChartsProps> = ({
                     <FlagStats
                         count={summary.total}
                         flagsPerUser={
-                            showAllProjects
-                                ? (summary.total / users.total).toFixed(2)
```
thomasheartman, line 128: Sorry, this has taken me a while to realize, but **this** is what caused us to store the string 'NaN', right? 

Couldn't we have gotten around all the other changes here by instead doing something like:

```ts
const averageFlagsPerUserCalculation = summary.total / users.total
const averageFlagsPerUser = averageFlagsPerUserCalculation.isNan() ? 'N/A' : averageFlagsPerUserCalculation.toFixed(2)
```

and then using that value as the prop instead?

That way, we don't need to update any other files, we don't need to touch the hook, and we don't need to touch any tests, etc.

andreas-unleash: I think it makes more sense to move the calculation to the summary hook inline with the rest of the calculations.

thomasheartman: Sure. You know the system better than I do, so I'll trust you on that.

PR #6642, opened 2024-03-20
feat(integrations): ability to add stringified event

Tymek:
You might want to include entire event, but as a different field.


chriswk (approved):
Seems fair. 😀

PR #6651, opened 2024-03-21
feat: read change requests for insights

kwasniew:
## About the changes

* introducing read model for project insights
* adding one method with change request counts
* removing unused code in project service

Next PR:
* add more tests with non zero values
* consider moving some of the code to the new store instead of coupling project insights service to stores and services from other subdomains

### DIFF
--- a/src/lib/features/project-insights/project-insights-read-model.ts
+++ b/src/lib/features/project-insights/project-insights-read-model.ts
@@ -1,0 +1,59 @@
+import type {
+    ChangeRequestCounts,
+    IProjectInsightsReadModel,
+} from './project-insights-read-model-type';
+import type { Db } from '../../db/db';
+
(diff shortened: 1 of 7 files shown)


src/lib/features/project-insights/project-insights-read-model.ts
```diff
@@ -0,0 +1,59 @@
...
+        };
+
+        const rows: Array<{ state: ChangeRequestDBState; count: string }> =
+            await this.db('change_requests')
+                .select('state')
+                .count('* as count')
+                .where('project', '=', projectId)
+                .groupBy('state');
+
+        return rows.reduce((acc, current) => {
```
kwasniew, line 38: I consider removing some duplication here in next PR

sjaanus (approved):

PR #6652, opened 2024-03-21
fix: default strategy ui update fix

FredrikOseberg:
This PR fixes a bug where editing the default strategy would not refresh the resource it was depending on to display the data. This also surfaces another issue, which is that project settings is using data from the getProjectOverview hook to display the default strategies in each environment. This should be it's own resource, but that is beyond the scope of this PR.


frontend/src/component/project/Project/ProjectSettings/ProjectDefaultStrategySettings/ProjectEnvironment/ProjectEnvironmentDefaultStrategy/EditDefaultStrategy.tsx
```diff
@@ -102,7 +104,8 @@ const EditDefaultStrategy = () => {
             },
         });
 
-        await refetchSavedStrategySegments();
```
FredrikOseberg, line 105: This has no effect.

gardleopard (approved):
LGTM, you made this one look easy!

PR #6653, opened 2024-03-21
chore: create new flag to hide insights ui (#6638)

andreas-unleash:
Creates a new flag to control the executive dashboard ui

Closes #
[1-2208](https://linear.app/unleash/issue/1-2208/create-separate-ui-flag-decoupled-from-the-backend-flag)

---------

(cherry picked from commit 84707e2bf3aa2241a60c92aa4fd40f5b6b6e34c2)

## About the changes

Closes #

### DIFF
--- a/frontend/src/component/menu/routes.ts
+++ b/frontend/src/component/menu/routes.ts
@@ -66,8 +66,8 @@ export const routes: IRoute[] = [
         title: 'Insights',
         component: ExecutiveDashboard,
         type: 'protected',
-        menu: { mobile: true, advanced: true },
-        flag: 'executiveDashboard',
+        menu: { mobile: true },
+        flag: 'executiveDashboardUI',
         enterprise: false,
     },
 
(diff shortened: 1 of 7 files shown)


frontend/src/component/menu/routes.ts
```diff
@@ -66,8 +66,8 @@ export const routes: IRoute[] = [
         title: 'Insights',
         component: ExecutiveDashboard,
         type: 'protected',
-        menu: { mobile: true, advanced: true },
```
thomasheartman, line 69: Why are we removing `advanced: true` here? What does that mean?

andreas-unleash: Advanced is the dropdown menu `Configuration`. So it was on the nav bar but also the dropdown

thomasheartman: Right, so it was a duplicate, and it's not supposed to be?

thomasheartman (approved):

PR #6656, opened 2024-03-21
feat: ttp for all projects should show median value per day

andreas-unleash:
Separates out the calculation of the median to its own file and tested

Closes # [1-2212](https://linear.app/unleash/issue/1-2212/timetoproduction-for-all-projects-should-show-the-change-in-the-median)

### DIFF
--- a/frontend/src/component/executiveDashboard/chart-info.ts
+++ b/frontend/src/component/executiveDashboard/chart-info.ts
@@ -51,19 +51,19 @@ export const chartInfo = {
             'How the overall health changes over time for the selected projects.',
     },
     averageTimeToProduction: {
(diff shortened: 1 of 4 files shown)


frontend/src/component/executiveDashboard/chart-info.ts
```diff
@@ -51,19 +51,19 @@ export const chartInfo = {
             'How the overall health changes over time for the selected projects.',
     },
     averageTimeToProduction: {
```
thomasheartman, line 53: Can we change the name of the key in this object too? Just so it matches the content.

thomasheartman (approved):
Nice one! 🔥

PR #6659, opened 2024-03-21
feat: traffic visibility UI and store

daveleek:
## About the changes

This PR provides store method for retrieving traffic usage data based on period parameter, and UI + ui hook with the new chart for displaying traffic usage data spread out over selectable month.

Skjermbilde 2024-03-21 kl  12 40 38

In this PR we copied and adapted a plugin written by DX for highlighting a column in the chart:

image

There are some minor improvements planned which will come in a separate PR, reversing the order in legend and tooltip so the colors go from light to dark, and adding a month -sum below the legend

## Discussion points

- Should any of this be extracted as a separate reusable component?

PR #6673, opened 2024-03-22
chore: upgrade typescript in frontend to 5.4.2

nunogois:
Follow up to https://github.com/Unleash/unleash/pull/6605

This upgrades TypeScript in frontend to `5.4.2`, matching the version we have on the server.
Only 2 things broke with this upgrade, so the changes are related to fixing the types in those places.

This fixes https://github.com/Unleash/unleash/pull/6659

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyEdit/change-request-conflict-data.ts
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyEdit/change-request-conflict-data.ts
@@ -1,8 +1,6 @@
 import type {
     ChangeRequestState,
     ChangeRequestType,
-    IChangeRequestFeature,
-    IFeatureChange,
 } from 'component/changeRequest/changeRequest.types';
 import type { ScheduledChangeRequestViewModel } from 'hooks/api/getters/useScheduledChangeRequestsWithStrategy/useScheduledChangeRequestsWithStrategy';
 import type { IUiConfig } from 'interfaces/uiConfig';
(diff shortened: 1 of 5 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyEdit/change-request-conflict-data.ts
```diff
@@ -14,18 +12,7 @@ type ChangeRequestConflictCreatedData = {
...
-              id: ChangeRequestType['id'];
-              features: {
-                  name: IChangeRequestFeature['name'];
-                  changes: (Pick<IFeatureChange, 'action'> & {
-                      payload: { id?: number | string };
-                  })[];
-              }[];
-          }[]
-        | undefined,
+    changeRequests: ChangeRequestType[] | undefined,
```
nunogois, line 15: @thomasheartman can you please double check that this is sane? From what I've seen the only reason for this semi-type was to satisfy the types in tests. I think we should try to adapt our tests to our code and not the other way around, so this seems fine to me, but still want to double check.

daveleek (approved):
This looks sane but yeah let's see if @thomasheartman agrees! Great to get matching typescript versions though, nice work!

frontend/src/component/feature/FeatureStrategy/FeatureStrategyEdit/change-request-conflict-data.ts
thomasheartman: You're right! The only reason for it was to make it easier to test (so you don't have to generate a full change request for each test). 

I'm happy for you to revert it as long as the tests still pass 💁🏼  

I ... think I agree with adapting tests to code, but in this case, it also made be more clear about **what** the function needs instead of just saying "change request". So I'm undecided. Anyway, yeah, feel free to change it 😄

nunogois: I think I understand what you mean, since strictly speaking it may only need a subset of a change request 🤔 maybe this could be tackled in the future by using sub/base types that are extended to the full type only when needed.

However in general I'm leaning towards just specifying the full type every time, since that's easier, more maintainable, and it could just be the case that there are no current real use cases other than tests, like in this case.

Thank you for your input 🙌

PR #6678, opened 2024-03-22
Fixed syntax highlighting in the docs

alvinometric:
Adds support for Svelte syntax highlighting and lets us add support for additional languages in the future


website/yarn.lock
```diff
@@ -11336,14 +11326,6 @@ universalify@^2.0.0:
   resolved "https://registry.yarnpkg.com/universalify/-/universalify-2.0.0.tgz#75a4984efedc4b08975c5aeb73f530d02df25717"
   integrity sha512-hAZsKq7Yy11Zu1DE0OzWjw7nnLZmJZYTDZZyEFHZdUhV8FkH5MCfoU1XMaxXovpyW5nq5scPqq0ZDP9Zyl04oQ==
 
-unleash-proxy-client@2.5.0:
```
nnennandukwe, line 11339: Is there a particular reason this gets removed?

alvinometric: As far as I could find, it's not used anywhere in the (website/docs) codebase

nnennandukwe (approved):
LGTM! 🚀

ivarconr (approved):
Looks great!

website/yarn.lock
thomasheartman: Yeah, this looks right. Just to give some context: we **did** use Unleash to control a feedback form previously, but that was .. probably years ago now, and we haven't used it since. I guess we (or rather: I) just forgot to remove the dependency.

Thanks for cleaning up after me!

thomasheartman (approved):

website/src/theme/prism-include-languages.js
thomasheartman, line 1: Actually, on re-reviewing this, do you need to add this file? Isn't it enough to add this to the list of `prism.additionalLanguages` in `website/docusaurus.config.js`? Is there any reason why svelte is different?

The current list looks like this, by the way:

```js
        prism: {
            theme: require('prism-react-renderer/themes/oceanicNext'),
            additionalLanguages: [
                'csharp',
                'dart',
                'http',
                'java',
                'kotlin',
                'php',
                'ruby',
                'rust',
                'swift',
            ],
        },
```

website/yarn.lock
alvinometric: Thanks for the context @thomasheartman 🙏

website/src/theme/prism-include-languages.js
alvinometric: These strings of `additionalLanguages` all map to a language that already exists in prism. Svelte doesn't, (like Vue, etc) that's what I need to add `prism-svelte` as a dependency and load it manually.
https://docusaurus.io/docs/markdown-features/code-blocks#supported-languages

thomasheartman: Aha! Got it. Makes sense, then. Thanks for the clarification 😄

PR #6682, opened 2024-03-25
chore: add scim id migration for groups

sighphyre:
Adds a scim id to groups. Seemed like we could get away with the name here but unfortunately our names are not unique


chriswk (approved):
LGTM

PR #6684, opened 2024-03-25
feat: dependant flag on feature search

sjaanus:


src/lib/openapi/spec/feature-search-response-schema.ts
```diff
@@ -129,47 +143,6 @@ export const featureSearchResponseSchema = {
             nullable: true,
             description: 'The list of feature tags',
         },
-        children: {
```
sjaanus, line 132: Clean up schema. It is weird this even is here.

kwasniew (approved):
How is the perf/query plan impacted?

sjaanus:
> How is the perf/query plan impacted?

64k 8.3ms  to cost 70k /  9.9ms

not terrible

PR #6695, opened 2024-03-26
fix: Add metrics for old proxy forward

ivarconr:
This change adds a new prometheus counter to all us to capture when we automatically forward traffic from old /proxy paths to the /api/frontend path.

image


gardleopard (approved):
LGTM

PR #6696, opened 2024-03-26
feat: add index on group name

chriswk:
As the title says, adds an index on the group name. Scim (in particular Azure) uses group names to find correct group to sync.


sighphyre (approved):
Nice!

PR #6699, opened 2024-03-26
chore: generate orval types

sjaanus:


frontend/src/openapi/models/sdkContextSchema.ts
```diff
@@ -26,5 +26,5 @@ export interface SdkContextSchema {
     sessionId?: string;
     /** An identifier for the current user */
     userId?: string;
-    [key: string]: any;
+    [key: string]: string;
```
kwasniew, line 29: you have to fix it manually because we added a new backend type that breaks orval

kwasniew (approved):

PR #6725, opened 2024-03-28
fix: move outdated sdks check up

kwasniew:
## About the changes

Moving outdated sdks check one level up so that we avoid making backend API calls when the flag is disabled.


sjaanus (approved):
Nice, I noticed 404s also at one point, but I thought it was just fluke.

PR #6727, opened 2024-03-28
feat: keep parent selection in edit dependency dialogue

kwasniew:
## About the changes

Before: remove dependency option always selected by default
Screenshot 2024-03-28 at 10 29 51

After: if there's existing parent dependency it gets auto-selected by default
Screenshot 2024-03-28 at 10 29 42


sjaanus (approved):
Nice, I thought this was bug.

PR #6731, opened 2024-03-28
feat: disabled feature dependency

kwasniew:
## About the changes

Feature dependency on a disabled feature (e.g. killswitch)

### Important files

https://github.com/Unleash/unleash/assets/1394682/0d39285f-af4a-4ac7-a327-9d1a262f4f2a

Details:
* new flag that allows to select feature status: enabled/disabled
* new dropdown for status
* new display value in the feature overview (shows only when parent is disabled)
* works with regular add dependency and with add dependency in change request (separate PR in the enterprise)
* tests verifying new capability

### DIFF
--- a/frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
+++ b/frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
@@ -14,11 +14,13 @@ import useToast from 'hooks/useToast';
 import { formatUnknownError } from 'utils/formatUnknownError';
 import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
 import { DependenciesUpgradeAlert } from './DependenciesUpgradeAlert';
+import { useUiFlag } from 'hooks/useUiFlag';
 
 interface IAddDependencyDialogueProps {
     project: string;
     featureId: string;
     parentFeatureId?: string;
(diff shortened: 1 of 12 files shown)


frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
```diff
@@ -197,21 +230,58 @@ export const AddDependencyDialogue = ({
...
                 <ConditionallyRender
                     condition={showDependencyDialogue}
                     show={
                         <LazyOptions
                             project={project}
                             featureId={featureId}
                             parent={parent}
-                            onSelect={setParent}
+                            onSelect={(status) => {
+                                setParentValue({ status: 'enabled' });
```
sjaanus, line 253: Maybe it should not reset the bottom configuration.

sjaanus (approved):

frontend/src/component/feature/Dependencies/AddDependencyDialogue.tsx
kwasniew: I don't have a strong opinion here so let's gather feedback. I started w/o reset and it felt weird but it's just one opinion. For now it's hidden behind a flag anyway.

PR #6737, opened 2024-03-29
fix: revert archived column for backwards migration compatibility

kwasniew:
## About the changes

When downgrading to 5.10 we had one query impacted because of the missing legacy archived column (used in getFeatureTypeCounts). Bringing it back so that we keep our rule of being backwards compat for 2 versions ([from our db changes ADR](https://github.com/Unleash/unleash/blob/main/website/docs/contributing/ADRs/back-end/breaking-db-changes.md)).

It reverts this change: https://github.com/Unleash/unleash/pull/6431

### DIFF
--- a/src/migrations/20240329064629-revert-feature-archived.js
+++ b/src/migrations/20240329064629-revert-feature-archived.js
@@ -1,0 +1,18 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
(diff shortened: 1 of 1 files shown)


ivarconr (approved):

ivarconr (approved):
Looks sane to me.

sjaanus (approved):

PR #6754, opened 2024-04-02
fix: upgrade mime library

gastonfournier:
After upgrading some dependencies, the dependency resolution in unleash-enterprise is resolving mime@4 which introduces some breaking changes: https://github.com/broofa/mime/issues/295 rather than pinning to version 3, this PR upgrades unleash to use mime@4 which should fix the build in unleash-enterprise

https://github.com/broofa/mime/blob/b4c1deeb51e52ae8f89833069f5cf865b882dc4d/README.md?plain=1#L16-L23


gastonfournier:
Closed because of the following hotfix https://github.com/ivarconr/unleash-enterprise/pull/1190

PR #6755, opened 2024-04-02
chore: Better randomness

gastonfournier:
## About the changes
This change is irrelevant as it doesn't pose a security risk, but there's no reason for us not to use a different type of random generation for the `sessionId`

**Note:** the magic number 18 was picked because that's the length of the string we get from `String(Math.random())`

Closes https://github.com/Unleash/unleash/security/code-scanning/68 and https://github.com/Unleash/unleash/security/code-scanning/69


sighphyre (approved):
Sure, seems okay to me

sighphyre:
> Note: the magic number 18 was picked because that's the length of the string we get from String(Math.random())

`Math.random()` produces a float so that's true... sometimes xD

PR #6764, opened 2024-04-03
chore: stop using deprecated method

gastonfournier:
`storeUserEvent` from event-service was deprecated. We stop using it and remove the method completely


src/lib/features/events/event-service.test.ts
```diff
@@ -1,20 +0,0 @@
...
-        secret: '',
-        username: 'admin-token-username',
-        permissions: [],
-        internalAdminTokenUserId: ADMIN_TOKEN_USER.id,
-    };
-    const eventService = createFakeEventsService(createTestConfig());
-    const userDetails = eventService.getUserDetails(adminToken);
-    expect(userDetails.createdBy).toBe('admin-token-username');
-    expect(userDetails.createdByUserId).toBe(ADMIN_TOKEN_USER.id);
-});
```
nunogois, line 20: I guess this is no longer relevant, which is why it's getting deleted.

nunogois (approved):

sighphyre (approved):
YES!

src/lib/features/events/event-service.test.ts
gastonfournier: Indeed, this was only testing `getUserDetails` that was only added to serve these new methods. There was no other test (which is sad).

PR #6767, opened 2024-04-03
chore: fix bearer token middleware signal endpoint logic

nunogois:
This should make it so that the `signal-endpoint` route match is slightly less strict.

### DIFF
--- a/src/lib/middleware/bearer-token-middleware.ts
+++ b/src/lib/middleware/bearer-token-middleware.ts
@@ -10,7 +10,7 @@ export const bearerTokenMiddleware = ({
 
     return (req: Request, _: Response, next: NextFunction) => {
         if (
-            req.path.startsWith('/api/signal-endpoint/') ||
+            req.path.includes('/signal-endpoint/') ||
             flagResolver.isEnabled('bearerTokenMiddleware')
         ) {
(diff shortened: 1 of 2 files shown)


src/lib/middleware/bearer-token-middleware.ts
```diff
@@ -10,7 +10,7 @@ export const bearerTokenMiddleware = ({
 
     return (req: Request, _: Response, next: NextFunction) => {
         if (
-            req.path.startsWith('/api/signal-endpoint/') ||
+            req.path.includes('/signal-endpoint/') ||
```
gastonfournier, line 13: we could also use `const baseUriPath = config.server.baseUriPath || '';` that could be more precise. The config is available in the constructor so it should be doable, but I leave it up to you

gastonfournier (approved):
LG! With a suggestion

src/lib/middleware/bearer-token-middleware.ts
nunogois: Something like this? https://github.com/Unleash/unleash/pull/6767/commits/b647c8e6e3e3b474e3f4985f5940f5e4765c8dc5

gastonfournier (approved):

PR #6768, opened 2024-04-03
fix: tabs alignment

kwasniew:
## About the changes

Screenshot 2024-04-03 at 11 01 07

### DIFF
--- a/frontend/src/component/project/Project/Project.tsx
+++ b/frontend/src/component/project/Project/Project.tsx
@@ -272,20 +272,23 @@ export const Project = () => {
                                         tab.isEnterprise ? 'end' : undefined
                                     }
                                     icon={
-                                        <span>
+                                        <>
                                             <ConditionallyRender
                                                 condition={Boolean(tab.new)}
                                                 show={
-                                                    <StyledBadge color='success'>
-                                                        Beta
-                                                    </StyledBadge>
+                                                    // extra span to avoid badge getting color override from the overly specific parent component
+                                                    <span>
+                                                        <StyledBadge color='success'>
+                                                            Beta
+                                                        </StyledBadge>
+                                                    </span>
                                                 }
                                             />
                                             {(tab.isEnterprise &&
                                                 isPro() &&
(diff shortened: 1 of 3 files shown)


frontend/src/component/project/Project/Project.tsx
```diff
@@ -272,20 +272,23 @@ export const Project = () => {
                                         tab.isEnterprise ? 'end' : undefined
                                     }
                                     icon={
-                                        <span>
```
kwasniew, line 275: this span was breaking alignment and it was added recently

frontend/src/component/project/Project/Project.tsx
```diff
@@ -272,20 +272,23 @@ export const Project = () => {
...
+                                        <>
                                             <ConditionallyRender
                                                 condition={Boolean(tab.new)}
                                                 show={
-                                                    <StyledBadge color='success'>
-                                                        Beta
-                                                    </StyledBadge>
+                                                    // extra span to avoid badge getting color override from the overly specific parent component
+                                                    <span>
```
kwasniew, line 280: moving this span around the styled badge to escape css rule with direct tab span child selector that gets the purple color

thomasheartman: Would it be sensible to update the styling on the tab span instead of adding the extra span? Feels like this fixes the symptom but not the cause. We might need to do the same thing again later? 💁🏼

thomasheartman: Would probably also make codescene happier

thomasheartman (approved):

frontend/src/component/project/Project/Project.tsx
kwasniew: Tab is a material UI component with a very specific rule for its direct span child. I'd need to overrider Material UI internals which I didn't want to do.

thomasheartman: Oh, right. Seems weird that they'd add that, but yeah, happy with your reasoning there.

PR #6771, opened 2024-04-03
chore: cache query misses to protect against DDoS

gastonfournier:
## About the changes
This PR establishes a simple yet effective mechanism to avoid DDoS against our DB while also protecting against memory leaks.

This will enable us to release the flag `queryMissingTokens` to make our token validation consistent across different nodes

### DIFF
--- a/src/lib/services/api-token-service.test.ts
+++ b/src/lib/services/api-token-service.test.ts
@@ -1,11 +1,6 @@
 import { ApiTokenService } from './api-token-service';
 import { createTestConfig } from '../../test/config/test-config';
-import type {
-    IApiUser,
-    IUnleashConfig,
-    IUnleashOptions,
-    IUser,
-} from '../server-impl';
+import type { IUnleashConfig, IUnleashOptions, IUser } from '../server-impl';
 import { ApiTokenType, type IApiTokenCreate } from '../types/models/api-token';
 import FakeApiTokenStore from '../../test/fixtures/fake-api-token-store';
 import FakeEnvironmentStore from '../features/project-environments/fake-environment-store';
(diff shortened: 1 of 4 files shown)


src/lib/services/api-token-service.ts
```diff
@@ -185,10 +188,19 @@ export class ApiTokenService {
             );
         }
 
+        const nextAllowedQuery = this.queryAfter.get(secret) ?? 0;
         if (
             !token &&
+            isPast(nextAllowedQuery) &&
```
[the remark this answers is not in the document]
gastonfournier: One is a feature flag that will be removed later, we should consider that when defining these thresholds :/

src/lib/services/api-token-service.test.ts
```diff
@@ -243,4 +248,44 @@ describe('When token is added by another instance', () => {
...
+            try {
+                found = await apiTokenService.getUserForToken(
+                    `${invalidToken}-${i % 2}`,
+                );
+            } catch (e) {
+                // ignore
+            }
+            expect(found).toBeUndefined();
+        }
+        expect(spy).toHaveBeenCalledTimes(2);
```
nunogois, line 275: Not super readable at a glance the way it's written, but I guess it's called twice because it's called once for the even token and another for the odd token?

I wonder if it would be better to simplify this, even if it costs DRYness.

src/lib/services/api-token-service.ts
```diff
@@ -185,10 +188,19 @@ export class ApiTokenService {
...
             this.flagResolver.isEnabled('queryMissingTokens', flagContext)
         ) {
+            if (this.queryAfter.size > 1000) {
+                // establish a max limit to query after size to prevent memory leak
```
nunogois, line 198: I think it's more readable this way
```suggestion
                // establish a max limit for queryAfter size to prevent memory leak
```

nunogois (approved):

src/lib/services/api-token-service.test.ts
gastonfournier: Sure, I think it makes sense, I was just playing around to make sure it worked properly but I realize it was a bit odd especially because of the try catch but also not very clear to see we have 2 tokens. I'll make a small change on it

src/test/fixtures/fake-api-token-store.ts
```diff
@@ -39,11 +38,8 @@ export default class FakeApiTokenStore
     }
 
     async get(key: string): Promise<IApiToken> {
-        const token = this.tokens.find((t) => t.secret === key);
-        if (token) {
-            return token;
-        }
-        throw new NotFoundError(`Could not find token with secret ${key}`);
+        // @ts-expect-error get can return undefined. See api-token-store.e2e.test.ts
+        return this.tokens.find((t) => t.secret === key);
```
gastonfournier, line 42: @nunogois just FYI while looking into this, I also found that this fake store was wrongly defined. I've tested how the real one behaves in src/test/e2e/stores/api-token-store.e2e.test.ts

src/lib/services/api-token-service.test.ts
gastonfournier: Addressed here: [f323e68](https://github.com/Unleash/unleash/pull/6771/commits/f323e686dbd12e3060dd7ff0df3ea83c60b75a22)

PR #6773, opened 2024-04-03
fix: handle objects in top-level context in playground

thomasheartman:
Don't include invalid context properties in the contexts that we evaluate. 

This PR removes any non-`properties` fields that have a non-string value. 

This prevents the front end from crashing when trying to render an object.

Expect follow-up PRs to include more warnings/diagnostics we can show to the end user to inform them of what fields have been removed and why.

### DIFF
--- a/src/lib/features/playground/playground-api.e2e.test.ts
+++ b/src/lib/features/playground/playground-api.e2e.test.ts
@@ -31,16 +31,12 @@ afterAll(async () => {
 });
 
 test('strips invalid context properties from input before using it', async () => {
-    const invalidData = {
-        invalid: {},
-    };
-
     const validData = {
         appName: 'test',
     };
 
     const inputContext = {
-        ...invalidData,
+        invalid: {},
         ...validData,
     };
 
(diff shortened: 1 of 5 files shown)


src/lib/features/playground/playground-service.ts
```diff
@@ -124,7 +124,20 @@ export class PlaygroundService {
                 this.resolveFeatures(filteredProjects, env),
             ),
         );
-        const contexts = generateObjectCombinations(context);
+
+        const { appName, properties, ...otherContextFields } = context;
+
+        const cleanedContextFields = Object.fromEntries(
```
kwasniew, line 130: can we extract a function with a descriptive name and test it for all the corner cases so that we don't need to test for corner cases in the e2e test. Rationale: E2E tests are pretty slow and I'd leave them for the happy path.

kwasniew (changes requested):
Another comment other than the inline one: can we move the e2e test to the feature oriented architecture? For some reason it's still stuck in the shared directory.

src/lib/features/playground/playground-service.ts
thomasheartman: Done! I've left the e2e test because it still tests two important things:
1. We actually perform the cleaning of the context
2. We still return as `input` **exactly** what you passed in.

Extracted the function into its own file, added tests that check the edge cases I could think of.

thomasheartman:
Moving the e2e test was also fixed. It wasn't stuck, it just didn't exist before. So without thinking, I created it there because that's where most of the e2e tests live. So thanks for pointing that out!

src/lib/features/playground/playground-api.e2e.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        .post('/api/admin/playground/advanced')
+        .send({
+            context: inputContext,
+            environments: ['production'],
+            projects: '*',
+        })
+        .expect(200);
+
+    expect(body.input.context).toMatchObject(inputContext);
+});
```
thomasheartman, line 78: We could put these back into a single test if we want to, but splitting it out into two makes it clearer what they're each testing. Happy to do either.

kwasniew (approved):

PR #6774, opened 2024-04-04
fix: strategy targeting numeric also check value field

daveleek:
## About the changes

API returns both value and values fields. Empty values array causes ui to think constraint doesnt have a value

This PR checks if value field exists and is empty before returning check on values and length

## Discussion points

Included a check on the value of the field in addition to checking if the field exists, I'll check the Open API spec but can we guarantee it doesn't exist without also having a value?
Should we check if the field is undefined? Feels like if it exists it's returned either with a value or as empty.


sighphyre (approved):
LGTM

PR #6784, opened 2024-04-05
feat: playground api returns removed context values under a new `warnings` property

thomasheartman:
This PR expands upon #6773 by returning the list of removed properties in the API response. To achieve this, I added a new top-level `warnings` key to the API response and added an `invalidContextProperties` property under it. This is a list with the keys that were removed.

## Discussion points

**Should we return the type of each removed key's value?** We could expand upon this by also returning the type that was considered invalid for the property, e.g. `invalidProp: 'object'`. This would give us more information that we could display to the user. However, I'm not sure it's useful? We already return the input as-is, so you can always cross-check. And the only type we allow for non-`properties` top-level properties is `string`. Does it give any useful info? I think if we want to display this in the UI, we might be better off cross-referencing with the input?

**Can properties be invalid for any other reason?** As far as I can tell, that's the only reason properties can be invalid for the context. OpenAPI will prevent you from using a type other than string for the context fields we have defined and does not let you add non-string properties to the `properties` object. So all we have to deal with are top-level properties. And as long as they are strings, then they should be valid.

**Should we instead infer the diff when creating the model?** In this first approach, I've amended the `clean-context` function to also return the list of context fields it has removed. The downside to this approach is that we need to thread it through a few more hoops. Another approach would be to compare the input context with the context used to evaluate one of the features when we create the view model and derive the missing keys from that. This would probably work in 98 percent of cases. However, if your result contains no flags, then we can't calculate the diff. But maybe that's alright? It would likely be fewer lines of code (but might require additional testing), although picking an environment from feels hacky.

### DIFF
--- a/src/lib/features/playground/clean-context.test.ts
+++ b/src/lib/features/playground/clean-context.test.ts
@@ -1,15 +1,15 @@
 import { cleanContext } from './clean-context';
 
-test('strips invalid context properties from the context', async () => {
-    const invalidJsonTypes = {
-        object: {},
-        array: [],
-        true: true,
-        false: false,
-        number: 123,
-        null: null,
-    };
+const invalidJsonTypes = {
+    object: {},
+    array: [],
+    true: true,
+    false: false,
+    number: 123,
+    null: null,
+};
 
+test('strips invalid context properties from the context', async () => {
     const validValues = {
         appName: 'test',
     };
@@ -19,7 +19,7 @@ test('strips invalid context properties from the context', async () => {
         ...validValues,
     };
 
-    const cleanedContext = cleanContext(inputContext);
+    const { context: cleanedContext } = cleanContext(inputContext);
 
     expect(cleanedContext).toStrictEqual(validValues);
 });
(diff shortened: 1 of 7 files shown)


src/lib/features/playground/clean-context.test.ts
```diff
@@ -29,7 +29,25 @@ test("doesn't add non-existing properties", async () => {
...
+    const { removedProperties } = cleanContext({
+        appName: 'test',
+        ...invalidJsonTypes,
+    });
+
+    const invalidProperties = Object.keys(invalidJsonTypes);
+
+    // verify that the two lists contain all the same elements
+    expect(removedProperties).toEqual(
+        expect.arrayContaining(invalidProperties),
```
kwasniew, line 47: is it because the order is not predictable? for predictable order I'd do simple equal

kwasniew (approved):
do you plan to use the warnings in the UI?

src/lib/features/playground/clean-context.test.ts
thomasheartman: I haven't checked the order, to be honest. I could, but it's not something that I think we can intrinsically rely on; we might want to sort it alphabetically or something later 🤷🏼 I don't think that should cause the tests to fail. So this just seemed like a more robust way to check it.

thomasheartman:
Yes, that's the plan. Create a follow-up PR that exposes this data in the UI if it's present.

PR #6786, opened 2024-04-05
feat: feature lifecycle usage behind a flag

kwasniew:
## About the changes

* introducing featureLifecycle flag that can disable event listening at runtime
* making some of the methods private in the lifecycle service
* setting up fake lifecycle service in the initial service bootstrap so that we can test with the in-memory based persistence

When get get DB migrations and SQL repository we'll switch to the real store. 

In the next PR I'll expose the in-memory recorded lifecycle stages over API.

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
@@ -5,6 +5,8 @@ import {
     FEATURE_CREATED,
     type IEnvironmentStore,
     type IEventStore,
+    type IFlagResolver,
+    type IUnleashConfig,
 } from '../../types';
 import type {
     FeatureLifecycleView,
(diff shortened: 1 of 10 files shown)


src/lib/services/index.ts
```diff
@@ -347,6 +349,9 @@ export const createServices = (
     const inactiveUsersService = new InactiveUsersService(stores, config, {
         userService,
     });
+    const { featureLifecycleService } =
+        createFakeFeatureLifecycleService(config);
```
kwasniew, line 353: we'll switch this to prod impl when we're done with controller and testing the in-memory version

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -18,47 +20,71 @@ export class FeatureLifecycleService {
...
+        },
+        { flagResolver }: Pick<IUnleashConfig, 'flagResolver'>,
+    ) {
         this.eventStore = eventStore;
         this.featureLifecycleStore = featureLifecycleStore;
         this.environmentStore = environmentStore;
+        this.flagResolver = flagResolver;
+    }
+
+    private async checkEnabled(fn: () => Promise<void>) {
```
kwasniew, line 43: each event needs to be wrapped with a check because we want runtime control. If we put this around the listen method we'd only have startup control.

PR #6787, opened 2024-04-05
fix: total flags and flags per user when all projects

andreas-unleash:
Fixes the flag stats flagsPerUser calculation and display value. 
Previously the calculation depended on project data which might not be there with the permission changes

Closes # [1-2255](https://linear.app/unleash/issue/1-2255/unable-to-render-total-number-of-flags-when-there-are-no-project-data)

### DIFF
--- a/frontend/src/component/insights/InsightsCharts.tsx
+++ b/frontend/src/component/insights/InsightsCharts.tsx
@@ -12,7 +12,11 @@ import { TimeToProduction } from './componentsStat/TimeToProduction/TimeToProduc
 import { TimeToProductionChart } from './componentsChart/TimeToProductionChart/TimeToProductionChart';
 import { MetricsSummaryChart } from './componentsChart/MetricsSummaryChart/MetricsSummaryChart';
 import { UpdatesPerEnvironmentTypeChart } from './componentsChart/UpdatesPerEnvironmentTypeChart/UpdatesPerEnvironmentTypeChart';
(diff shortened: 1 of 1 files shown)


frontend/src/component/insights/InsightsCharts.tsx
```diff
@@ -78,6 +84,16 @@ export const InsightsCharts: VFC<IChartsProps> = ({
     const showAllProjects = projects[0] === allOption.id;
     const isOneProjectSelected = projects.length === 1;
 
+    function getFlagsPerUser(
+        flags: InstanceInsightsSchemaFlags,
+        users: InstanceInsightsSchemaUsers,
+    ) {
+        const flagsPerUserCalculation = flags.total / users.total;
```
Tymek, line 91: This function can receive more primitive type, and thus be simplified. I would pass `flags.total` and `users.total` instead of entire objects. Also `formatFlagsPerUser` would be more descriptive to me.

Tymek (approved):

frontend/src/component/insights/InsightsCharts.tsx
andreas-unleash: Not sure I understand what you mean here

Tymek: This function doesn't need to have complex arguments. It can probably work with `? number | null` arguments instead, and be called as `formatFlagsPerUser(flags.total, users.total)

PR #6790, opened 2024-04-05
feat: Feature lifecycle sql store

kwasniew:
## About the changes

* db migration for lifecycle stages
* sql store based on the new migration and tested through the e2e test (if there's error in any query the tests catches it)
* making feature lifecycle service an event emitter itself so that we can wait to stage change events in tests instead of `await ms(randomTime)`

### DIFF
--- a/src/lib/features/feature-lifecycle/fake-feature-lifecycle-store.ts
+++ b/src/lib/features/feature-lifecycle/fake-feature-lifecycle-store.ts
@@ -8,9 +8,12 @@ export class FakeFeatureLifecycleStore implements IFeatureLifecycleStore {
     private lifecycles: Record<string, FeatureLifecycleView> = {};
 
     async insert(featureLifecycleStage: FeatureLifecycleStage): Promise<void> {
-        const existing = await this.get(featureLifecycleStage.feature);
+        if (await this.stageExists(featureLifecycleStage)) {
+            return;
+        }
+        const existingStages = await this.get(featureLifecycleStage.feature);
         this.lifecycles[featureLifecycleStage.feature] = [
-            ...existing,
+            ...existingStages,
             {
                 stage: featureLifecycleStage.stage,
                 enteredStageAt: new Date(),
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-store.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-store.ts
@@ -1,0 +1,52 @@
+import type {
+    FeatureLifecycleStage,
+    IFeatureLifecycleStore,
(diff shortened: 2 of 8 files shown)


src/lib/features/feature-lifecycle/feature-lifecycle-store.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    async get(feature: string): Promise<FeatureLifecycleView> {
+        const results = await this.db('feature_lifecycles')
+            .where({ feature })
+            .orderBy('created_at', 'asc');
+
+        return results.map(({ stage, created_at }: DBType) => ({
+            stage,
+            enteredStageAt: created_at,
```
kwasniew, line 40: will probably rename enteredStageAt to createdAt in the service and openapi

src/lib/features/feature-lifecycle/fake-feature-lifecycle-store.ts
```diff
@@ -8,9 +8,12 @@ export class FakeFeatureLifecycleStore implements IFeatureLifecycleStore {
     private lifecycles: Record<string, FeatureLifecycleView> = {};
 
     async insert(featureLifecycleStage: FeatureLifecycleStage): Promise<void> {
-        const existing = await this.get(featureLifecycleStage.feature);
+        if (await this.stageExists(featureLifecycleStage)) {
```
kwasniew, line 11: prevent double inserts in fake impl

src/migrations/20240405120422-add-feature-lifecycles.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+exports.up = function(db, cb) {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS feature_lifecycles (
+            feature VARCHAR(255) NOT NULL REFERENCES features(name) ON DELETE CASCADE,
```
kwasniew, line 7: we want to delete feature lifecycle when feature is removed

src/migrations/20240405120422-add-feature-lifecycles.js
```diff
@@ -0,0 +1,18 @@
...
+            stage VARCHAR(255) NULL,
+            created_at TIMESTAMP WITH TIME ZONE default now(),
```
kwasniew, line 9: sticking to generic name instead of specific stage_entered_at

kwasniew: using timestamp with timezone to have the same type as features table

andreas-unleash (approved):
LGTM

PR #6794, opened 2024-04-08
feat: allow schedulers to run in a single node

gastonfournier:
## About the changes
This PR provides a service that allows a scheduled function to run in a single instance. It's currently not in use but tests show how to wrap a function to make it single-instance:
https://github.com/Unleash/unleash/blob/65b7080e0532ffdcbd8c2aa2593d0aff78e632fb/src/lib/features/scheduler/job-service.test.ts#L26-L32

The key `'test'` is used to identify the group and most likely should have the same name as the scheduled job.


src/migrations/20240405174629-jobs.js
```diff
@@ -0,0 +1,26 @@
...
+        `
+        CREATE TABLE IF NOT EXISTS jobs (
+            id SERIAL PRIMARY KEY NOT NULL,
+            name VARCHAR(255) NOT NULL,
+            last_execution TIMESTAMP NOT NULL,
+            checkpoint VARCHAR(255) NOT NULL,
+            UNIQUE (name)
+        );
+
+        CREATE INDEX IF NOT EXISTS idx_job_names ON jobs(name);
```
gastonfournier, line 12:
```suggestion
        CREATE UNIQUE INDEX IF NOT EXISTS idx_job_names ON jobs(name);
```

src/migrations/20240405174629-jobs.js
```diff
@@ -0,0 +1,26 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS jobs (
+            id SERIAL PRIMARY KEY NOT NULL,
+            name VARCHAR(255) NOT NULL,
```
gastonfournier, line 6:
```suggestion
            name TEXT NOT NULL,
```
VARCHAR(255) does not have value anymore since postgres ~10 and with postgres 14 indexing is better with TEXT

src/migrations/20240405174629-jobs.js
```diff
@@ -0,0 +1,26 @@
...
+            last_execution TIMESTAMP NOT NULL,
+            checkpoint VARCHAR(255) NOT NULL,
+            UNIQUE (name)
```
gastonfournier, line 9:
```suggestion
```

gastonfournier:
cc @kwasniew @sjaanus just FYI cause it's related to schedulers. Maybe you have a candidate for this functionality, but first we're going to test it in enterprise customers with the feature we are developing. Maybe we can wait until it's actually tested to use it elsewhere but just keep it in mind ;)

src/migrations/20240405174629-jobs.js
```diff
@@ -0,0 +1,35 @@
...
+        `,
+        cb,
+    );
+};
+
+exports.down = function (db, cb) {
+    db.runSql(
+        `
+            DROP TABLE IF EXISTS jobs;
+            DROP INDEX IF EXISTS idx_job_names;
```
chriswk, line 31:
```suggestion
            DROP INDEX IF EXISTS idx_job_finished;
            DROP INDEX IF EXISTS idx_job_stage;
            DROP TABLE IF EXISTS jobs;
            
```

chriswk (approved):
Other than the slight mismatch in our down migration, this looks good to me.

PR #6795, opened 2024-04-08
feat: handle invalid metrics

kwasniew:
## About the changes

Closes #


src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -64,6 +71,7 @@ export class FeatureLifecycleService extends EventEmitter {
             );
         });
         this.eventBus.on(CLIENT_METRICS, async (event) => {
+            if (!event.featureName || !event.environment) return;
```
kwasniew, line 74: if any data is missing we don't even proceed

PR #6800, opened 2024-04-08
chore: SCIM settings UI

nunogois:
https://linear.app/unleash/issue/2-1981/ui-allow-users-to-set-up-scim

Adds UI for Provisioning (SCIM) settings.

image

image

image

### DIFF
--- a/frontend/src/component/admin/auth/ScimSettings/ScimSettings.tsx
+++ b/frontend/src/component/admin/auth/ScimSettings/ScimSettings.tsx
@@ -1,0 +1,144 @@
+import type React from 'react';
+import { useEffect, useState } from 'react';
+import { Button, FormControlLabel, Grid, Switch } from '@mui/material';
+import { Alert } from '@mui/material';
+import useToast from 'hooks/useToast';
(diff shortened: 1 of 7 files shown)


frontend/src/component/admin/auth/ScimSettings/ScimSettings.tsx
```diff
@@ -0,0 +1,144 @@
...
+    };
+
+    return (
+        <>
+            <Grid container sx={{ mb: 3 }}>
+                <Grid item md={12}>
+                    <Alert severity='info'>
+                        Please read the{' '}
+                        <a
+                            href='https://docs.getunleash.io/reference/scim'
```
sighphyre, line 70: Thanks! Now I know where to put the docs xD

sighphyre (approved):
LGTM! But I'd wait for someone with good Ueyes before hitting merge

FredrikOseberg (commented):
I think this is fine for now. The entire SSO section needs to be revamped at some point but that is another story and beyond the scope of this. I feel like this keeps the theme of what we have so far pretty well.

PR #6808, opened 2024-04-09
Feat/allow admin login using demo auth

### LINKED ISSUE #6398: Initial/default admin user cannot be accessed when using `demo` authentication type

00Chaotic:
## About the changes
The `admin` user currently cannot be accessed in `demo` authentication mode, as the auth mode requires only an email to log in, and the admin user is not created with an email. This change allows for logging in as the admin user only if an `AUTH_DEMO_ALLOW_ADMIN_LOGIN` is set to `true` (or the corresponding `authDemoAllowAdminLogin` config is enabled).

Closes #6398 

### Important files
[demo-authentication.ts](https://github.com/Unleash/unleash/compare/main...00Chaotic:unleash:feat/allow_admin_login_using_demo_auth?expand=1#diff-c166f00f0a8ca4425236b3bcba40a8a3bd07a98d067495a0a092eec26866c9f1R25)

## Discussion points
Can continue discussion of [this comment](https://github.com/Unleash/unleash/pull/6447#issuecomment-2042405647) in this PR.

### DIFF
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -15,7 +15,7 @@ exports[`should create default config 1`] = `
     "styleSrc": [],
   },
   "authentication": {
-    "authDemoAllowAdminLogin": false,
+    "demoAllowAdminLogin": false,
     "createAdminUser": true,
     "customAuthHandler": [Function],
     "enableApiToken": true,
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -15,9 +15,9 @@ exports[`should create default config 1`] = `
     "styleSrc": [],
   },
   "authentication": {
-    "demoAllowAdminLogin": false,
     "createAdminUser": true,
     "customAuthHandler": [Function],
+    "demoAllowAdminLogin": false,
     "enableApiToken": true,
     "initApiTokens": [],
     "initialAdminUser": {
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -15,6 +15,7 @@ exports[`should create default config 1`] = `
     "styleSrc": [],
   },
   "authentication": {
+    "authDemoAllowAdminLogin": false,
     "createAdminUser": true,
     "customAuthHandler": [Function],
     "enableApiToken": true,
--- a/src/lib/create-config.test.ts
+++ b/src/lib/create-config.test.ts
@@ -231,13 +231,13 @@ test('should load demo admin login flag from env var', async () => {
 
     const config = createConfig({});
 
-    expect(config.authentication.authDemoAllowAdminLogin).toBeTruthy();
+    expect(config.authentication.demoAllowAdminLogin).toBeTruthy();
     delete process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN;
 });
 
 test('should default demo admin login to false', async () => {
     const config = createConfig({});
-    expect(config.authentication.authDemoAllowAdminLogin).toBeFalsy();
+    expect(config.authentication.demoAllowAdminLogin).toBeFalsy();
 });
 
 test('should load environment overrides from env var', async () => {
--- a/src/lib/create-config.test.ts
+++ b/src/lib/create-config.test.ts
@@ -226,6 +226,20 @@ test('should handle cases where no env var specified for tokens', async () => {
     expect(config.authentication.initApiTokens).toHaveLength(1);
 });
 
+test('should load demo admin login flag from env var', async () => {
+    process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN = 'true';
+
+    const config = createConfig({});
+
+    expect(config.authentication.authDemoAllowAdminLogin).toBeTruthy();
+    delete process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN;
+});
+
+test('should default demo admin login to false', async () => {
+    const config = createConfig({});
+    expect(config.authentication.authDemoAllowAdminLogin).toBeFalsy();
+});
+
 test('should load environment overrides from env var', async () => {
     process.env.ENABLED_ENVIRONMENTS = 'default,production';
 
--- a/src/lib/create-config.test.ts
+++ b/src/lib/create-config.test.ts
@@ -226,15 +226,6 @@ test('should handle cases where no env var specified for tokens', async () => {
     expect(config.authentication.initApiTokens).toHaveLength(1);
 });
 
-test('should load demo admin login flag from env var', async () => {
-    process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN = 'true';
-
-    const config = createConfig({});
-
-    expect(config.authentication.demoAllowAdminLogin).toBeTruthy();
-    delete process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN;
-});
-
 test('should default demo admin login to false', async () => {
     const config = createConfig({});
     expect(config.authentication.demoAllowAdminLogin).toBeFalsy();
--- a/src/lib/middleware/demo-authentication.e2e.test.ts
+++ b/src/lib/middleware/demo-authentication.e2e.test.ts
@@ -1,0 +1,73 @@
+import dbInit from '../../test/e2e/helpers/database-init';
+import { IAuthType } from '../server-impl';
+import { setupAppWithCustomAuth } from '../../test/e2e/helpers/test-helper';
+import type { ITestDb } from '../../test/e2e/helpers/database-init';
+import type { IUnleashStores } from '../types';
+
+let db: ITestDb;
+let stores: IUnleashStores;
+
+beforeAll(async () => {
+    db = await dbInit('demo_auth_serial');
+    stores = db.stores;
+});
+
+afterAll(async () => {
+    await db?.destroy();
+});
+
+const getApp = (adminLoginEnabled: boolean) =>
+    setupAppWithCustomAuth(stores, () => {}, {
+        authentication: {
+            authDemoAllowAdminLogin: adminLoginEnabled,
+            type: IAuthType.DEMO,
+            createAdminUser: true,
+        },
+    });
+
+test('the authDemoAllowAdminLogin flag should not affect regular user login/creation', async () => {
+    const app = await getApp(true);
+    return app.request
+        .post(`/auth/demo/login`)
+        .send({ email: 'test@example.com' })
+        .expect(200)
+        .expect((res) => {
+            expect(res.body.email).toBe('test@example.com');
+            expect(res.body.id).not.toBe(1);
+        });
+});
+
+test('if the authDemoAllowAdminLogin flag is disabled, using `admin` should have the same result as any other invalid email', async () => {
+    const app = await getApp(false);
+
+    const nonAdminUsername = 'not-an-email';
+    const adminUsername = 'admin';
+
+    const nonAdminUser = await app.request
+        .post(`/auth/demo/login`)
+        .send({ email: nonAdminUsername });
+
+    const adminUser = await app.request
+        .post(`/auth/demo/login`)
+        .send({ email: adminUsername });
+
+    expect(nonAdminUser.status).toBe(adminUser.status);
+
+    for (const user of [nonAdminUser, adminUser]) {
+        expect(user.body).toMatchObject({
+            error: expect.stringMatching(/^Could not sign in with /),
+        });
+    }
+});
+
+test('should allow you to login as admin if the authDemoAllowAdminLogin flag enabled', async () => {
+    const app = await getApp(true);
+    return app.request
+        .post(`/auth/demo/login`)
+        .send({ email: 'admin' })
+        .expect(200)
+        .expect((res) => {
+            expect(res.body.id).toBe(1);
+            expect(res.body.username).toBe('admin');
+        });
+});
--- a/src/lib/middleware/demo-authentication.e2e.test.ts
+++ b/src/lib/middleware/demo-authentication.e2e.test.ts
@@ -19,13 +19,13 @@ afterAll(async () => {
 const getApp = (adminLoginEnabled: boolean) =>
     setupAppWithCustomAuth(stores, () => {}, {
         authentication: {
-            authDemoAllowAdminLogin: adminLoginEnabled,
+            demoAllowAdminLogin: adminLoginEnabled,
             type: IAuthType.DEMO,
             createAdminUser: true,
         },
     });
 
-test('the authDemoAllowAdminLogin flag should not affect regular user login/creation', async () => {
+test('the demoAllowAdminLogin flag should not affect regular user login/creation', async () => {
     const app = await getApp(true);
     return app.request
         .post(`/auth/demo/login`)
@@ -37,7 +37,7 @@ test('the authDemoAllowAdminLogin flag should not affect regular user login/crea
         });
 });
 
-test('if the authDemoAllowAdminLogin flag is disabled, using `admin` should have the same result as any other invalid email', async () => {
+test('if the demoAllowAdminLogin flag is disabled, using `admin` should have the same result as any other invalid email', async () => {
     const app = await getApp(false);
 
     const nonAdminUsername = 'not-an-email';
@@ -60,7 +60,7 @@ test('if the authDemoAllowAdminLogin flag is disabled, using `admin` should have
     }
 });
 
-test('should allow you to login as admin if the authDemoAllowAdminLogin flag enabled', async () => {
+test('should allow you to login as admin if the demoAllowAdminLogin flag enabled', async () => {
     const app = await getApp(true);
     return app.request
         .post(`/auth/demo/login`)
--- a/src/lib/middleware/demo-authentication.ts
+++ b/src/lib/middleware/demo-authentication.ts
@@ -4,7 +4,7 @@ import type { IUnleashServices } from '../types/services';
 import type { IUnleashConfig } from '../types/option';
 import ApiUser from '../types/api-user';
 import { ApiTokenType } from '../types/models/api-token';
-import type { IAuthRequest } from '../server-impl';
+import type { IAuthRequest, IUser } from '../server-impl';
 import type { IApiRequest } from '../routes/unleash-types';
 import { encrypt } from '../util';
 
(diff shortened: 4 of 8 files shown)


src/lib/middleware/demo-authentication.test.ts
thomasheartman, line 1: Small request (and I know this is my fault): can we rename this file to `demo-authentication.e2e.test.ts`? Just to make it clearer that this is an end-to-end test.

src/lib/middleware/demo-authentication.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        });
+});
+
+test('should return 403 for admin user if flag disabled', async () => {
+    const app = await getApp(false);
+    return app.request
+        .post(`/auth/demo/login`)
+        .send({ email: 'admin' })
+        .expect(403);
+});
```
thomasheartman, line 45: As discussed in the other PR thread, I think the right thing to check for here is that using `admin` when the flag is disabled gives you the same response as any other invalid email. I suggest replacing this with a test that compares the admin attempt with a different invalid attempt.

Even if you disagree here, I think 403 is the wrong status code, and that 401 would be better if we were to go with that.

```suggestion
test('if the authDemoAllowAdminLogin flag is disabled, using `admin` should have the same result as any other invalid email', async () => {
    const app = await getApp(false);

    const nonAdminUsername = 'not-an-email'
    const adminUsername = 'admin'

    const nonAdminUser = await app.request
        .post(`/auth/demo/login`)
        .send({ email: nonAdminUsername })

    const adminUser = await app.request
        .post(`/auth/demo/login`)
        .send({ email: adminUsername })

    // verify that the status is the same
    expect(nonAdminUser.status).toBe(adminUser.status);

    for (const user of [nonAdminUser, adminUser]) {
        // verify that the body has the same structure
        expect(user.body).toMatchObject({
            error: expect.stringMatching(/^Could not sign in with /)
        })
    }
});
```

src/lib/middleware/demo-authentication.ts
```diff
@@ -19,14 +19,26 @@ function demoAuthentication(
...
         try {
-            const user = await userService.loginUserWithoutPassword(
-                email,
-                true,
-            );
+            if (!authentication.authDemoAllowAdminLogin && email === 'admin') {
+                return res.status(403)
+                    .json({ error: `Admin login not enabled for demo auth type`})
+                    .end;
+            } else if (authentication.authDemoAllowAdminLogin && email === 'admin') {
```
thomasheartman, line 29: This is the other part of the auth status change that we've discussed previously.

```suggestion
            if (authentication.authDemoAllowAdminLogin && email === 'admin') {
```

src/lib/middleware/demo-authentication.test.ts
```diff
@@ -0,0 +1,57 @@
...
+const getApp = (adminLoginEnabled: boolean) =>
+    setupAppWithCustomAuth(stores, () => {}, {
+        authentication: {
+            authDemoAllowAdminLogin: adminLoginEnabled,
+            type: IAuthType.DEMO,
+            createAdminUser: true,
+        },
+    });
+
+test('should create regular user with flag enabled', async () => {
```
thomasheartman, line 27:
```suggestion
test('the authDemoAllowAdminLogin flag should not affect regular user login/creation', async () => {
```

src/lib/middleware/demo-authentication.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+test('should return 403 for admin user if flag disabled', async () => {
+    const app = await getApp(false);
+    return app.request
+        .post(`/auth/demo/login`)
+        .send({ email: 'admin' })
+        .expect(403);
+});
+
+test('should allow login with admin user if flag enabled', async () => {
```
thomasheartman, line 47:
```suggestion
test('should allow you to login as admin if the authDemoAllowAdminLogin flag enabled', async () => {
```

thomasheartman (changes requested):
Thanks for taking the time to work out a solution and implementing it 🙌🏼  This looks pretty good to me, but I've requested changes due to the admin return code that we've discussed before. Once that's changed (or if you convince me that this really *is* the right way), then I'd be happy to approve it.

As for why I think not returning a special case for admin when the flag is disabled, refer to my [last comment on the other PR](https://github.com/Unleash/unleash/pull/6447#issuecomment-2046821993). (I'm sure you've seen it, but it's also for anyone else reading this later 💁🏼 )

I've also added some small suggestions to the test titles to make it clearer what exactly they are testing.

00Chaotic:
Thanks for the review, I've made the requested changes to the test case naming.

Regarding the error code:
- Outside of this change, we return a `400` which suggests the request is malformed or otherwise incorrect in some way ([docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400)) - this is a good enough "generic" error, although it does not account for any server-side errors that may occur, but that's a separate discussion
- If we return a `401`, this would suggest that the request lacks validation credentials, which I understand to mean that the request would have been fulfilled if it contained the relevant credentials (which is not the case here, as the request would be denied as long as the flag is disabled, regardless of any credentials) ([docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401))
- A `403` suggests that the server understands the request but refuses to authorize it ([docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403)). I feel this is most appropriate here, as there is nothing wrong with the request itself, but the server refuses to authorize it because the flag is disabled

thomasheartman:
> Thanks for the review, I've made the requested changes to the test case naming.

Of course! Thanks for making the changes 😄 

> Regarding the error code: [...]

Yup, these are good points. 

On 401 vs 403:  I did go and read up on when to use 401 vs 403, and I got a sense that people generally recommended 403 for areas you're not allowed to enter with your credentials. In other words, "I know who you are, but you don't have access here". On the other hand, 401 was recommended for when the user isn't authenticated. You could say that this case is "special", because in a sense, the credentials are correct, but "you can't come in here right now". On the other hand, we also don't let you authenticate, so the server doesn't really know who you are 💁🏼 

To quote [MDN's doc on 401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) (with added emphasis):
> [the] request has not been completed because it lacks **valid** authentication credentials

And if that flag isn't enabled, then those authentication credentials aren't valid. 

Which brings us back to the 400. If the flag isn't enabled, then 'admin' isn't any different (and thus shouldn't be treated any differently, I think) from any other non-email address. And the established pattern is that Unleash gives you a 400 in that case. I think you could make an argument that it should be 401, but the 400 makes sense because you've not provided the data that Unleash asked for.

But then again, I'm not an expert on HTTP status codes, and this is all just my understanding. 

However, as the arbiter here, I'm going to say I remain unconvinced that we should create a special case for 'admin' when the flag is disabled. With the points you provided, I think 400, 401, and (maybe to a lesser extent?) 403 are all valid options. But Unleash has an established precedent of returning 400 if the email isn't an email, and I don't see this case as being so different that it warrants special-casing. This is both to keep the interface consistent (people who don't use the flag shouldn't need to care whether someone provides "admin" or "bogus"; they're both invalid) and to keep the code simpler.

So my request still stands. I think that with the flag disabled, nothing should change for existing users and people who don't use the flag. 'admin' should be treated exactly as any other invalid email address. Does that sound agreeable to you?

00Chaotic:
Yep, at the end of the day it _is_ your product so if you feel the `400` is the best way forward then we can go with that.

I've pushed the changes along with some linting fixes, unfortunately locally it messes up with the line endings and suggests changes to 2000+ files so I'll just see what the PR check says

thomasheartman:
Super! Thanks for your hard work!

> I've pushed the changes along with some linting fixes, unfortunately locally it messes up with the line endings and suggests changes to 2000+ files so I'll just see what the PR check says

Yeah, linting appears to be failing. Weird that it messes up line endings locally for you. Unfortunately, we don't have automated lint fixing in CI, but there's not that many errors. Maybe try just running the formatter on the three specific files that fail? You can probably target specific files with the[ CLI](https://biomejs.dev/formatter/#use-the-formatter-with-the-cli), as mentioned in the docs?

00Chaotic:
@thomasheartman not sure if the build check is hanging, seems to still be waiting for a status after ~24hrs since the last change

thomasheartman:
Thanks for reporting in! I think you're right there. Running it now 🙋🏼

src/lib/middleware/demo-authentication.e2e.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const adminUser = await app.request
+        .post(`/auth/demo/login`)
+        .send({ email: adminUsername });
+
+    expect(nonAdminUser.status).toBe(adminUser.status);
+
+    for (const user of [nonAdminUser, adminUser]) {
+        expect(user.body).toMatchObject({
+            error: expect.stringMatching(/^Could not sign in with /)
```
thomasheartman, line 58:
```suggestion
            error: expect.stringMatching(/^Could not sign in with /),
```

sighphyre (commented):
I think this looks good to me. I'd like to see green tests here before I can put a formal approve on this but I'm happy with the ideas here!

src/lib/create-config.test.ts
```diff
@@ -226,6 +226,20 @@ test('should handle cases where no env var specified for tokens', async () => {
     expect(config.authentication.initApiTokens).toHaveLength(1);
 });
 
+test('should load demo admin login flag from env var', async () => {
+    process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN = 'true';
+
+    const config = createConfig({});
+
+    expect(config.authentication.authDemoAllowAdminLogin).toBeTruthy();
```
thomasheartman, line 234: This test is failing for some reason. Does it run correctly locally for you?

src/lib/__snapshots__/create-config.test.ts.snap
```diff
@@ -15,6 +15,7 @@ exports[`should create default config 1`] = `
     "styleSrc": [],
   },
   "authentication": {
+    "authDemoAllowAdminLogin": false,
```
thomasheartman, line 18: I know this is late, but what do you think about changing this to just `demoAllowAdminLogin`? The env name would stay the same (`AUTH_DEMO_ALLOW_ADMIN_LOGIN`), but the idea is that the `AUTH_` prefix indicates that it is part of the `authentication` object. This is the same way that `AUTH_TYPE` (which becomes `authentication.type`) works. I think that was my intention earlier, but I guess I wasn't clear enough and didn't think about it when we got this far 😅

00Chaotic:
Haven't had a chance to look at this for a few days but just saw the email for the failing linting issue (apologies, rebased and force pushed which probably wiped your changes).

I agree with the name change and will get that pushed, and look into the failing test 👍

src/lib/create-config.test.ts
00Chaotic: Okay this is weird, the `process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN` isn't being evaluated correctly. It evaluates to `'true'` if I print it in this test case but if I add a `console.log()` in `parseEnvVarBoolean()` which is used [here](https://github.com/Unleash/unleash/pull/6808/files#diff-a38cc7ae9f963264333b6535498970a33e046dba50a6af554f438343a6edd2ddR274), it prints `undefined`.

Strange because a lot of the other tests in this file do the exact same thing and yet it correctly parses those env vars 🤔 

Can't really be a timing/race issue either since those operations aren't async.

Is there any behind-the-scenes env var processing I'm not aware of that may be inadvertently affecting this variable?

thomasheartman: Hmm, so this is a little tricky, but it _is_ actually kind of a timing issue.

The thing is, most of the other tests in that file have corresponding functions in `createConfig`, but `defaultAuthentication` is defined as a variable. So it does the parsing of the env variables when that file first gets loaded, which happens before you set the process variable.

You can try this out by setting any of the other auth variables to a different value (boolean or otherwise).

So this isn't really an issue for Unleash itself. It makes testing a little trickier, but only in this specific case.

We **could** make it so that the auth config is also loaded through a function, but I'm a little hesitant to do that. It might have further security repercussions if you can dynamically change the environment and thus the auth options of Unleash after it starts. It seems a little far-fetched, but I'd rather not change it.

So how do we deal with this, then? It feels a little weird, but I think I'd actually just remove that test.

But I'd actually like to hear @sighphyre's take here: is it safe to load auth options dynamically or does it pose potential risks? Unleash ever only does it on startup, so I don't see how it's really gonna be an issue, but I'm not a security expert. And if not, is this test low-stakes enough that leaving it out is fine?

00Chaotic: Happy to wait for their input.

I'd be hesitant to remove the test since this might be our only way of knowing if the `true` case ever stops working in future, whereas for the `false` case we still have the default config case that will include this flag.

I don't really have a better suggestion though so if you both agree that this is the best way forward then we can take this route.

00Chaotic:
> (Oh, don't worry about commit history, by the way; we'll squash it down before merging anyway).

As a side-note unrelated to this PR, the squash vs merge debate seems to be a long-standing one, so I'm curious as to your reasoning, if any, for using squash merges (if that's what you meant as opposed to manually squashing the commits and then merging).

There are multiple articles arguing for both sides but [this one](https://timwise.co.uk/2021/03/15/github-rebase-and-squash-considered-harmful/) is one I've come across a couple times. At my work we're moving away from squash merging and more towards manually merging individual commits (if needed), and then doing a normal merge into the target branch.

Not that I'm pushing for it to be done here, but just food for thought 🙂

thomasheartman:
> I'm curious as to your reasoning

Ah, yes; one of the ever-present developer debates! Honestly, it makes things easy and quick for us, and that's **my** main motivation, at least. I remember there was a pretty clear reasoning when we set that as the default, but I can't remember it exactly.

I read the article, and while I think the author has some good points, it also does exactly what it says closer to the conclusion:

> As always extremes are bad and anyone who fast-talks you into believing one extreme is a panacea is glossing over important nuance and interesting counter-examples.

I think it very much comes down to your team's workflow. PRs here are getting smaller and smaller, and being able to merge the entire content with one clear message is nice 🤷🏼 Sure, we could rebase, rewrite commit messages, etc, but I'm not convinced it's worth it, personally.

That said, I'd love it if github made it easier to split a PR up into multiple PRs or to structure changes into a series of commits. Similarly, in git, if you have a series of patches, I'd love to be able to say "actually, take lines 57 through 68 in file A and lines 15, 17, and 25 in file B and but them in a different branch/commit/(or indeed github's PR)" even when those changes have been made in three different commits. I'm sure you _can_ do it, but I'm not sure it'll be quick or easy.

So eh 🤷🏼 I used to believe strongly in one direction, but then someone convinced me otherwise. Now I think all (most?) approaches have their merit, and when it comes down to which is the right one for your team: the answer is the ol' adage "it depends".

src/lib/create-config.test.ts
thomasheartman: You're right: we won't have a test that checks whether you can set it via **environment variables** (which it also seems like we don't explicitly for a lot of the other config variables). But we **do** test that you can turn it on via config in the e2e test file, so it's not _untested_.  

So the only thing that test is actually doing is checking that we parse the env var correctly. Which definitely isn't worthless, but feels like it's unlikely to break. If we break env var parsing, then the other tests that load env vars dynamically will yell at us. If we get the name wrong, well, we'd have to keep supporting the wrong name, then. But a test won't really protect us against that unless we go and update the exposed env var directly, which would be a breaking change and thus not likely to happen.

So in short: I think it'd be nice if it worked, and if loading it dynamically is cool, then that's great! However, I'm also not too bothered if its' better to leave this one case out.

sighphyre: > But I'd actually like to hear @sighphyre's take here: is it safe to load auth options dynamically or does it pose potential risks? Unleash ever only does it on startup, so I don't see how it's really gonna be an issue, but I'm not a security expert. And if not, is this test low-stakes enough that leaving it out is fine?

I think if you've lost control of your environment and an attacker can inject something in there then you have much, much bigger problems on your hands. I don't see any harm in moving that to being loaded later in the startup sequence though, if only for testing purposes. It should still only be loaded once when this is invoked.

So weirdly enough I don't really have strong feelings about what we do here. Mostly because demo auth isn't a core feature and the most dangerous thing that can happen here is that is gets accidentally turned **on**. So I have strong feelings about adjacent things - that demo auth remains off by default and that that test is clear but that seems to be covered by the test below the one in question.

I think @thomasheartman is right here, this is covered enough by the e2e tests. So it largely depends on how strongly you feel about it, @00Chaotic. I'd lean in the direction of removing the test here and getting this merged because gosh this conversation has gone on for a long time and PRs that stay open for a long time get worse and worse to try merge. I think this is good enough to merge in it's current state.

I don't see harm in restructuring the way this gets loaded to later if you think the quality would be improved by a more isolated test though.

src/lib/create-config.test.ts
```diff
@@ -226,6 +226,20 @@ test('should handle cases where no env var specified for tokens', async () => {
...
 
+test('should load demo admin login flag from env var', async () => {
+    process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN = 'true';
+
+    const config = createConfig({});
+
+    expect(config.authentication.demoAllowAdminLogin).toBeTruthy();
+    delete process.env.AUTH_DEMO_ALLOW_ADMIN_LOGIN;
+});
+
```
thomasheartman, line 237: Removing this as discussed in [this thread](https://github.com/Unleash/unleash/pull/6808#discussion_r1565376845).

```suggestion
```

thomasheartman (approved):

thomasheartman:
@00Chaotic I've made the change to the tests that we discussed in the thread (and removed that one test) and merged the PR. Thanks a whole bunch for your contribution and for the discussions we had along the way 🙌🏼

src/lib/create-config.test.ts
00Chaotic: Thanks for your feedback, I'm happy to remove the tests. I agree about just trying to get it merged at this point, it has definitely been messy with long discussions in multiple PRs haha.

I'll go ahead and remove the test case, and we can merge it in once all the checks pass.

00Chaotic:
Ah thanks @thomasheartman, just saw your latest comment. Glad to have the issue closed. I really enjoyed working together, thanks for your patience and perseverance with this haha. Wishing you all the best with Unleash 🙏

PR #6812, opened 2024-04-10
fix: show all envs in project tables unless you've explicitly hidden some

thomasheartman:
This PR changes the behavior of the project tables' environment columns based on input from customers.

Up until now, you have been shown either the first project or the first three projects in the list of the project's environment. The decision on whether to show one or three is based on screen size. The breakpoint appears to be about 1280px. Above that you get three, below it you get one.

With this PR, we'll show you *all* environments by default, regardless of screen size. However, that's just for the default values. If you manually change column visibility, those changes will of course be respected.

I've used a new package, `css-mediaquery`, to test that all screen sizes show all envs.


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/hooks/useDefaultColumnVisibility.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+            addEventListener: () => {},
+            removeEventListener: () => {},
+            dispatchEvent: () => true,
+        };
+    };
+}
+
+function resizeScreen(width: number) {
+    window.matchMedia = createMatchMedia(width);
+}
```
thomasheartman, line 25: This bit allows us to resize the screen in the tests (based on [this article](https://www.js-howto.com/test-responsive-design-using-jest-and-react-testing-library/) that I found). I'm wondering whether we should extract it into a separate file or leave it here for now.

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/hooks/useDefaultColumnVisibility.ts
thomasheartman, line 1: This is the actual fix to the problem.

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/hooks/useDefaultColumnVisibility.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+    );
+};
+
+test.each(Object.keys(theme.breakpoints.values))(
+    'it renders all envs on %s screens',
+    (screenSize) => {
+        resizeScreen(
+            theme.breakpoints.values[
+                screenSize as keyof typeof theme.breakpoints.values
+            ] + 1,
```
Tymek, line 62: I suspect we have an off-by-one issue here. If you cut a stick in 2 places, how many sticks you will end up with? It's the same with breakpoints.

Tymek (changes requested):
Since it's the first time we're doing this I think it's OK to not have a 'central/unified' utility for it. But please double-check if we are testing all relevant states of the screen, or "all greater then breakpoints"

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/hooks/useDefaultColumnVisibility.test.tsx
thomasheartman: That's a fair point. But I had in fact thought of this. The thing is, assuming I'm looking at the right file (`frontend/src/themes/theme.ts`), then breakpoints is defined as such:

```ts
    breakpoints: {
        values: {
            xs: 0,
            sm: 600,
            md: 960,
            lg: 1280,
            xl: 1536,
        },
    },
```

So in this case, the smallest size we set for the screen is 1 pixel. The way I read this, `xs` is from 0 px and up, `sm` from 600 px and up etc. Thus, using one px above their lower threshold should put us safely within their zone.

Am I making sense?

thomasheartman: Or, the breakpoint values are defined by their lower bound.

Tymek: O, perfect, thanks for clarification

Tymek (approved):

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/hooks/useDefaultColumnVisibility.test.tsx
thomasheartman: In theory, we could probably skip the +1, but that just means we avoid any weird edge cases if we use strictly vs greater than or equal to

PR #6817, opened 2024-04-10
fix: show loading features only when loading is happening

FredrikOseberg:
Previously, the dummy data would persist when there is no data coming from the API. This causes us to display dummy data in the dora metrics table which is not correct. This PR fixes that by only showing the loading features when we are actually loading.


frontend/src/component/project/Project/ProjectSettings/ProjectDefaultStrategySettings/ProjectDefaultStrategySettings.test.tsx
```diff
@@ -0,0 +1,85 @@
+import { screen, fireEvent } from '@testing-library/react';
```
FredrikOseberg, line 1: Biome decided to format this file as well. It has not been touched, but had wrong formatting for some reason.

sjaanus (approved):
LG

PR #6818, opened 2024-04-10
chore: job-service handle exceptions

gastonfournier:
## About the changes
This allows failed functions to record that they've failed when wrapped with a job.

### DIFF
--- a/src/lib/features/scheduler/job-service.test.ts
+++ b/src/lib/features/scheduler/job-service.test.ts
@@ -8,6 +8,8 @@ let store: JobStore;
 const config = createTestConfig();
 beforeAll(async () => {
     db = await dbInit('job_service_serial', config.getLogger);
+    // @ts-ignore setMuteError is not part of getLogger interface
+    config.getLogger.setMuteError(true);
     store = new JobStore(db.rawDatabase, config);
 });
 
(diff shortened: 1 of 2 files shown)


src/lib/features/scheduler/job-service.test.ts
```diff
@@ -8,6 +8,8 @@ let store: JobStore;
 const config = createTestConfig();
 beforeAll(async () => {
     db = await dbInit('job_service_serial', config.getLogger);
+    // @ts-ignore setMuteError is not part of getLogger interface
+    config.getLogger.setMuteError(true);
```
chriswk, line 12: is the setMuteError really on the LogProvider, and not the Logger?

chriswk (approved):
Nice

src/lib/features/scheduler/job-service.test.ts
gastonfournier: The provider: https://github.com/Unleash/unleash/blob/3f8336cd19fbaca8916bf0b5b671770892eba193/src/test/fixtures/no-logger.ts#L17-L19 because of how the function is defined: https://github.com/Unleash/unleash/blob/3f8336cd19fbaca8916bf0b5b671770892eba193/src/test/fixtures/no-logger.ts#L12

I don't like it much, but I also didn't want to spend much time on it

PR #6821, opened 2024-04-10
Feat/configure scheduled created by migration

daveleek:
## About the changes

- Removes the feature flag for the created_by migrations.
- Adds a configuration option in IServerOption for `ENABLE_SCHEDULED_CREATED_BY_MIGRATION` that defaults to `false`
- the new configuration option when set on startup enables scheduling of the two created_by migration services (features+events)
- Removes the dependency on flag provider in EventStore as it's no longer needed
- Adds a brief description of the new configuration option in `configuring-unleash.md`
- Sets the events created_by migration interval to 15 minutes, up from 2.


website/docs/using-unleash/deploy/configuring-unleash.md
```diff
@@ -121,6 +121,7 @@ unleash.start(unleashOptions);
   - _unleashUrl_ (string) - Used to specify the official URL this instance of Unleash can be accessed at for an end user. Can also be configured through the environment variable `UNLEASH_URL`.
   - _gracefulShutdownEnable_: (boolean) - Used to control if Unleash should shutdown gracefully (close connections, stop tasks,). Defaults to true. `GRACEFUL_SHUTDOWN_ENABLE`
   - _gracefulShutdownTimeout_: (number) - Used to control the timeout, in milliseconds, for shutdown Unleash gracefully. Will kill all connections regardless if this timeout is exceeded. Defaults to 1000ms `GRACEFUL_SHUTDOWN_TIMEOUT`
+  - _enableScheduledCreatedByMigration_: (boolean) - Schedules migrations that resolves created_by for events and features. Defaults to false. `ENABLE_SCHEDULED_CREATED_BY_MIGRATION`
```
gastonfournier, line 124: I think we need more details here as it doesn't explain much, as a customer I wouldn't know if I want this enabled or not... Maybe something along these lines, although please validate what I wrote:
```suggestion
  - _enableScheduledCreatedByMigration_: (boolean) - Schedules migrations that fills the field created_by_user_id for past events and features. It does it gradually in chunks and runs every hour so it doesn't affect the application's performance. Defaults to false. `ENABLE_SCHEDULED_CREATED_BY_MIGRATION`
```

gastonfournier (approved):
Code looks good, docs need some love, but we can also do another iteration

website/docs/using-unleash/deploy/configuring-unleash.md
daveleek: Thank you @gastonfournier! One of them runs every 2mins and the other runs every 15 mins, so I guess I can rephrase that part. except for that it looks like a good improvement!

PR #6822, opened 2024-04-10
fix: replace true on important create screens

FredrikOseberg:
This adds replace: true to navigate on the create feature toggle screen and create project screen. This will make it so you don't go back to the form after you have created the resource, replacing the entry in the history with the new url. We can do this in more places, but some of them require a bit more thought. For example when creating a user, you navigate from the admin screen to the user page, and then back to the same screen. Adding `{ replace: true }` in this context makes it so that when you press back you end up on the same screen, because it's recorded twice in history. 

Another discussion point: 
* Would you expect the edit screens to also replace the history?


thomasheartman (approved):
This looks good to me and I think this is the right way to go in these cases. Thanks 😄

PR #6831, opened 2024-04-11
feat: add "my projects" filter to projects list

thomasheartman:
This PR adds the buttons (only UI, no functionality) to show either "all projects" or "my projects". 

The buttons use a styled button group and are hidden behind the new `projectListFilterMyProjects` flag.

The button placement breaks with the previously established page header pattern of having all actions moved to the right. To accommodate this new placement, I created a new flex container in the header called `leftActions`, which is essentially just a mirror of the normal actions. I went with `leftActions` instead of `inlineStartActions` or something similar because I think it's clearer, and I don't see us adapting Unleash for different writing directions right now. We can always change it later.

I have also slightly increased the end margin of the page header to accommodate the new designs and to adjust the spacing before the buttons. I adjusted the margin of the text instead of the padding of the left actions because this will keep the spacing to the page header the same on every page. Without it, we could end up in situations where the spacing changes from page to page based on whether it has left actions or not, which is probably undesirable.

image

### Hover colors

~~Find out what the right hover color variable is. I'm using the light mode hover color for now, which works well in both light and dark modes (looks nice and is AAccessible), but it's not the same as the hover color for other buttons in dark mode.~~

Fixed ☝🏼 

### Small windows

Also worth noting: at around 500px, the layout shift starts to cause problems and we end up with overlapping elements. How do we want to deal with narrower screens? Today, the UI is pretty functional until we reach about 250px. It would be nice to not increase that size.

The new version breaking at about 500px:
image

The old version breaking at about 250px:
image

### Margins

We also need to figure out how much space we want on smaller windows:

image


frontend/src/component/project/ProjectList/ProjectList.tsx
```diff
@@ -49,6 +55,29 @@ const StyledCardLink = styled(Link)(({ theme }) => ({
     pointer: 'cursor',
 }));
 
+const StyledButtonGroup = styled(ToggleButtonGroup)(({ theme }) => ({
+    'html[data-theme="light"] button': {
+        '--hover-background-color': '#615BC2',
+    },
+    'html[data-theme="dark"] button': {
+        '--hover-background-color': '#34325E',
+    },
```
thomasheartman, line 64: No, you're right, this doesn't work. We don't use `data-theme` in Unleash. But this is just a placeholder with the correct values for now. It should be removed once we find the right variable to use for the actual hover background color.

frontend/src/component/project/ProjectList/ProjectList.tsx
```diff
@@ -172,6 +205,36 @@ export const ProjectListNew = () => {
             header={
                 <PageHeader
                     title={`Projects (${projectCount})`}
+                    leftActions={
+                        <ConditionallyRender
```
Tymek, line 209: This looks like it could be a separate component.

frontend/src/component/project/ProjectList/ProjectList.tsx
Tymek: Let's note it in code. `// FIXME: find hover background color and remove placeholder`?

FredrikOseberg: Couldn't you simply do `&:hover { }` and set it to a variable from the theme? Or are these values missing from the theme?

thomasheartman: Yeah, just using `:hover` should work, but no, I can't figure out what the values in the theme are. Do you know?

frontend/src/component/project/ProjectList/ProjectList.tsx
thomasheartman: I'm not entirely sure I understand. What do you mean?

frontend/src/component/tags/TagTypeList/__tests__/__snapshots__/TagTypeList.test.tsx.snap
thomasheartman, line 1: This test was failing because the class name had changed. I'm assuming this is somehow caused by some of the other changes in this PR. However, it doesn't appear to have a visual effect. I'm guessing some input args to some algorithm changed which is reflected here. I think this is a safe change to make, but I'd love for you to tell me if you disagree.

Tymek (approved):

PR #6833, opened 2024-04-11
Docs: Fixed missing values in webhooks

alvinometric:
Fixed missing values in the webhook docs, the list of values is from [the type definitions](https://github.com/Unleash/unleash/blob/d3eac6caf83b6bdff25f218d0be761f7a390a2b3/src/lib/addons/webhook-definition.ts#L100-L129)

Credit goes to @gastonfournier for finding these 🙏


gastonfournier (approved):
LGTM!

PR #6836, opened 2024-04-11
chore: SCIM guard for users

nunogois:
https://linear.app/unleash/issue/2-2093/api-should-not-allow-manual-management-of-scim-managed-users-in

Introduces a SCIM guard for SCIM users. SCIM users should be managed exclusively by the SCIM client, not Unleash.

We decided to be restrictive for now, completely covering all of the write methods, but may fine-tune some of this at a later stage.

Will eventually be followed up by a UI-centric PR.

### DIFF
--- a/frontend/src/component/admin/users/UsersList/ChangePassword/ChangePassword.tsx
+++ b/frontend/src/component/admin/users/UsersList/ChangePassword/ChangePassword.tsx
@@ -13,6 +13,7 @@ import PasswordMatcher from 'component/user/common/ResetPasswordForm/PasswordMat
 import type { IUser } from 'interfaces/user';
 import useAdminUsersApi from 'hooks/api/actions/useAdminUsersApi/useAdminUsersApi';
 import { UserAvatar } from 'component/common/UserAvatar/UserAvatar';
+import useToast from 'hooks/useToast';
 
 const StyledUserAvatar = styled(UserAvatar)(({ theme }) => ({
     width: theme.spacing(5),
@@ -36,6 +37,7 @@ const ChangePassword = ({
     const [validPassword, setValidPassword] = useState(false);
     const { classes: themeStyles } = useThemeStyles();
     const { changePassword } = useAdminUsersApi();
+    const { setToastData } = useToast();
 
     const updateField: React.ChangeEventHandler<HTMLInputElement> = (event) => {
         setError(undefined);
@@ -58,6 +60,11 @@ const ChangePassword = ({
             await changePassword(user.id, data.password);
             setData({});
             closeDialog();
+            setToastData({
+                title: 'Password changed successfully',
+                text: 'The user can now sign in using the new password.',
+                type: 'success',
+            });
         } catch (error: unknown) {
             console.warn(error);
             setError(PASSWORD_FORMAT_MESSAGE);
--- a/frontend/src/hooks/api/actions/useAdminUsersApi/useAdminUsersApi.ts
+++ b/frontend/src/hooks/api/actions/useAdminUsersApi/useAdminUsersApi.ts
@@ -1,10 +1,4 @@
 import useAPI from '../useApi/useApi';
-import {
-    handleBadRequest,
-    handleForbidden,
-    handleNotFound,
-    handleUnauthorized,
-} from './errorHandlers';
 
 interface IUserPayload {
     name: string;
@@ -16,10 +10,7 @@ export const REMOVE_USER_ERROR = 'removeUser';
 
 const useAdminUsersApi = () => {
     const { loading, makeRequest, createRequest, errors } = useAPI({
-        handleBadRequest,
-        handleNotFound,
-        handleUnauthorized,
-        handleForbidden,
+        propagateErrors: true,
     });
 
     const addUser = async (user: IUserPayload) => {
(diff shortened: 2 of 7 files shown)


src/test/e2e/api/admin/user-admin.scim.e2e.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    await db.destroy();
+});
+
+test('fetching a SCIM user should include scimId', async () => {
+    const { body } = await app.request
+        .get(`/api/admin/user-admin/${scimUserId}`)
+        .expect(200);
+
+    expect(body.email).toBe(scimUser.email);
+    expect(body.scimId).toBe('some-random-scim-id');
```
nunogois, line 64: Including the `scimId` in the response type will allow us to adapt the UI as well. I think this is fine however if, for any reason, we would like not to expose this, we could expose a `isScim` boolean property instead.

frontend/src/hooks/api/actions/useAdminUsersApi/useAdminUsersApi.ts
```diff
@@ -16,10 +10,7 @@ export const REMOVE_USER_ERROR = 'removeUser';
 
 const useAdminUsersApi = () => {
     const { loading, makeRequest, createRequest, errors } = useAPI({
-        handleBadRequest,
-        handleNotFound,
-        handleUnauthorized,
-        handleForbidden,
+        propagateErrors: true,
```
nunogois, line 13: The reason for this change was to have proper errors in the UI toast, otherwise we would get `cannot read property "msg" of undefined` errors. I think this is a safe change. Can you please confirm @FredrikOseberg and @Tymek?

frontend/src/component/admin/users/UsersList/ChangePassword/ChangePassword.tsx
```diff
@@ -58,6 +60,11 @@ const ChangePassword = ({
             await changePassword(user.id, data.password);
             setData({});
             closeDialog();
+            setToastData({
+                title: 'Password changed successfully',
+                text: 'The user can now sign in using the new password.',
+                type: 'success',
+            });
```
nunogois, line 67: Pure scouting. I noticed we didn't have a toast when we changed a user's password. I think this is nicer.

sighphyre: Ye that's nice!

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -716,4 +725,28 @@ export default class UserAdminController extends Controller {
             projectRoles,
         });
     }
+
+    async throwIfScimUser({
+        id,
+        scimId,
+    }: Pick<IUser, 'id' | 'scimId'>): Promise<void> {
+        if (this.isEnterprise && this.flagResolver.isEnabled('scimApi')) {
```
sighphyre, line 733: Gosh that's a lot of nested ifs! How would you feel about flattening this one a little?

I'm also tempted to move this check:

```
                const isScim =
                    Boolean(scimId)
```

Up in the function call. That looks super cheap. It'd be cool if we can early exit without making an API call!

src/test/e2e/api/admin/user-admin.scim.e2e.test.ts
sighphyre: Yeah, I think this is okay too!

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -716,4 +725,28 @@ export default class UserAdminController extends Controller {
...
+            );
+
+            if (enabled) {
+                const isScim =
+                    Boolean(scimId) ||
+                    Boolean((await this.userService.getUser(id)).scimId);
+
+                if (isScim) {
+                    throw new ForbiddenError(
+                        'Cannot perform this operation on SCIM users',
```
sighphyre, line 746: Tempted to make this a little... dumber. We're knee deep in SCIM so we know what this is. But an arbitrary user may not have all the context (they probably should since they were probably the one that set it up but that's life).

What about "This user is managed by your SCIM provider and cannot be changed manually"?

Not going to put a suggestion because that's gonna set fire to your tests O.O

sighphyre (approved):
I like it. It'd suggest double checking my comment on moving the boolean check to avoid an API call before you merge but this looks quite nice!

src/lib/routes/admin-api/user-admin.ts
nunogois: Agreed, not a fan either. Tried addressing it here: https://github.com/Unleash/unleash/pull/6836/commits/9b88eb26e08bfae867c52ac4b0b8a50360e84cb5

sighphyre: Yeah that's better!

src/lib/routes/admin-api/user-admin.ts
sighphyre: Legend!

frontend/src/hooks/api/actions/useAdminUsersApi/useAdminUsersApi.ts
nunogois: Double checked with @FredrikOseberg and this should be safe as long as the calls to this hook's methods are wrapped in a `try`/`catch`, which they seem to be.

PR #6844, opened 2024-04-12
feat: display removed context props in the UI

thomasheartman:
This PR shows context warnings in the UI if they exist.

That means that if you provide invalid properties on the top level, you'll get the result back, but we'll also tell you that we didn't take these properties into account.

Screenie:

image


frontend/src/component/playground/Playground/AdvancedPlayground.test.tsx
```diff
@@ -88,7 +88,95 @@ test('should display error on submit', async () => {
 
     const user = userEvent.setup();
     const submitButton = screen.getByText('Submit');
-    user.click(submitButton);
+    await user.click(submitButton);
```
thomasheartman, line 91: This await appears to have been missing 🤷🏼

frontend/src/component/playground/Playground/AdvancedPlayground.tsx
```diff
@@ -28,6 +28,56 @@ const StyledAlert = styled(Alert)(({ theme }) => ({
...
 }));
 
+const GenerateWarningMessages: React.FC<{
+    response?: AdvancedPlaygroundResponseSchema;
+}> = ({ response }) => {
+    const invalidContextProperties =
+        response?.warnings?.invalidContextProperties;
+
+    if (invalidContextProperties && invalidContextProperties.length > 0) {
+        invalidContextProperties.sort;
```
gastonfournier, line 38: Should this be `invalidContextProperties.sort()`?

frontend/src/component/playground/Playground/AdvancedPlayground.tsx
```diff
@@ -304,11 +354,16 @@ export const AdvancedPlayground: VFC<{
...
                                     show={
-                                        <AdvancedPlaygroundResultsTable
-                                            loading={loading}
-                                            features={results?.features}
-                                            input={results?.input}
-                                        />
+                                        <>
+                                            <GenerateWarningMessages
+                                                response={results}
+                                            />
```
gastonfournier, line 360: This component will be `null` when no warning, I'm just wondering if this is ok (i.e. react knows how to handle null components). Even if react supports this, I remember we were using conditional rendering rather than returning null.

gastonfournier (approved):
Looks nice, just a few comments/questions but non-blockers

frontend/src/component/playground/Playground/AdvancedPlayground.tsx
thomasheartman: Yes... yes it should. Good point. I .. should probably test that it does that too 😅

frontend/src/component/playground/Playground/AdvancedPlayground.tsx
thomasheartman: I'm pretty sure that react knows how to handle nulls, yeah. I think it's even preferable to, say, a, fragment. `ConditionallyRender` also returns `null` if it's not rendering anything.

frontend/src/component/playground/Playground/AdvancedPlayground.tsx
thomasheartman: Done in 74026919882db9a6ab0fd9e05854b905b8a4989d

frontend/src/component/playground/Playground/AdvancedPlayground.tsx
gastonfournier: Sounds reasonable! I kind of like this approach because it requires less boilerplate code (I mean, tags are boilerplate in a way). So I think it's good and something for my TIL list :)

PR #6845, opened 2024-04-12
chore: SCIM guard for groups

nunogois:
https://linear.app/unleash/issue/2-2111/api-should-not-allow-manual-management-of-scim-managed-groups-in

Introduces a SCIM guard for SCIM groups. SCIM groups should be managed exclusively by the SCIM client, not Unleash.

We decided to be restrictive for now, completely covering all of the write methods, but may fine-tune some of this at a later stage.

Will eventually be followed up by a UI-centric PR.


sighphyre (approved):
LGTM!

PR #6847, opened 2024-04-12
fix: insights menu point alignment

andreas-unleash:
What is says on the tin

Closes [1-2288](https://linear.app/unleash/issue/1-2288/misalignment-insights-beta-menu-item)

Screenshot 2024-04-12 at 14 42 28

### DIFF
--- a/frontend/src/component/menu/Header/Header.tsx
+++ b/frontend/src/component/menu/Header/Header.tsx
@@ -9,7 +9,6 @@ import {
     styled,
     type Theme,
     Box,
-    Typography,
 } from '@mui/material';
 import MenuIcon from '@mui/icons-material/Menu';
 import SettingsIcon from '@mui/icons-material/Settings';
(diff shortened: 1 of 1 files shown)


frontend/src/component/menu/Header/Header.tsx
```diff
@@ -122,9 +123,16 @@ const StyledLinkWithBetaBadge = ({
...
 }: { title: string; to: string }) => (
     <StyledLink to={to} sx={{ margin: 0 }}>
-        <div>
+        <Typography
+            component='div'
+            sx={(theme) => ({
+                display: 'flex',
+                alignItems: 'center',
+                gap: theme.spacing(1),
+            })}
```
thomasheartman, line 132: Don't we recommend not using the `sx` prop, instead preferring styled components (for performance reasons, I think)?

If so, I think we should just quickly drag this out into a separate component to do just that.

andreas-unleash: No worries, changed to styled component

thomasheartman (approved):

PR #6848, opened 2024-04-12
chore(docs): improve documentation with note about initial version and not needing flag

daveleek:
## About the changes

Adds a note about new installations starting with version where most events and features should have a created_by userid set and thus not need this configuration option/corresponding scheduled data migration task


alvinometric:
Thanks 🙌

alvinometric (approved):

PR #6856, opened 2024-04-15
feat: start extracting project from session object

sjaanus:
Previously, we were extracting the project from the token, but now we will retrieve it from the session, which contains the full list of projects.

This change also resolves an issue we encountered when the token was a multi-project token, formatted as []:dev:token. Previously, it was unable to display the exact list of projects. Now, it will show the exact project names.

### DIFF
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -423,7 +423,6 @@ export default class ClientApplicationsStore
     }
 
     private remapUsageRow = (input) => {
-        console.log(input);
         if (this.flagResolver.isEnabled('parseProjectFromSession')) {
             if (!input.projects || input.projects.length === 0) {
                 return [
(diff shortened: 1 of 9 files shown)


src/lib/db/client-applications-store.ts
```diff
@@ -110,39 +111,39 @@ const remapRow = (input) => {
...
-            .insert(usageRow)
+            .insert(usageRows)
             .onConflict(['app_name', 'project', 'environment'])
             .merge();
     }
 
     async bulkUpsert(apps: Partial<IClientApplication>[]): Promise<void> {
         const rows = apps.map(remapRow);
-        const usageRows = apps.map(remapUsageRow);
+        const usageRows = apps.flatMap(this.remapUsageRow);
```
sjaanus, line 146: Since now it was array in array, we will flatmap it.

src/lib/db/client-applications-store.ts
```diff
@@ -420,4 +421,30 @@ export default class ClientApplicationsStore
             },
         };
     }
+
+    private remapUsageRow = (input) => {
```
sjaanus, line 425: Added case, when input is list of **projects** instead of just single **project**.

FredrikOseberg (commented):
Looks good, but are there existing tests covering this functionality since no new coverage was added?

sjaanus:
> Looks good, but are there existing tests covering this functionality since no new coverage was added?

Yes there was some coverage and added another test.

PR #6859, opened 2024-04-15
chore: UI SCIM guard for users

nunogois:
https://linear.app/unleash/issue/2-2092/ui-should-not-allow-manual-management-of-scim-managed-users-in-unleash

Adds a UI SCIM guard when trying to manage users.

The condition for the guard is:
 - Enterprise
 - SCIM flag enabled
 - SCIM setting enabled
 - SCIM user

image


sighphyre (approved):
LGTM

PR #6860, opened 2024-04-15
feat: validate that id is number or we throw our our source code

sjaanus:
Previously, we were not validating that the ID was a number, which sometimes resulted in returning our database queries (source code) to the frontend. Now, we have validation middleware.


FredrikOseberg (approved):
LGTM. The refactoring looks simple enough. Were there tests testing this functionality in the places we changed from the controller level validation to the middleware?

PR #6861, opened 2024-04-15
feat: project list frontend filter

Tymek:
## About the changes
Persist filter state in URL. This change also modifies how search phrase is used. Before we didn't persist it between visits to `/projects`. This is consistent with behavior on other pages, but it's a change that isn't behind a flag.

Internal ticket https://linear.app/unleash/issue/1-2286/front-end-store-filter-selection

### DIFF
--- a/frontend/cypress/integration/projects/list.spec.ts
+++ b/frontend/cypress/integration/projects/list.spec.ts
@@ -1,0 +1,36 @@
+///<reference path="../../global.d.ts" />
+import {
+    SEARCH_INPUT,
+    //@ts-ignore
+} from '../../../src/utils/testIds';
+
+describe('project overview', () => {
+    const randomId = String(Math.random()).split('.')[1];
+    const projectName = `unleash-e2e-projects-list-${randomId}`;
+
+    before(() => {
+        cy.runBefore();
+        cy.login_UI();
+        cy.createProject_API(projectName);
+    });
+
+    after(() => {
+        cy.deleteProject_API(projectName);
+    });
+
+    it('allows for searching and persists search query', () => {
+        cy.login_UI();
+        cy.visit(`/projects`);
+        cy.get(`[data-testid="${SEARCH_INPUT}"]`).as('search').click();
+        cy.get('@search').type(projectName);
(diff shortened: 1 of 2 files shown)


frontend/cypress/integration/projects/list.spec.ts
```diff
@@ -0,0 +1,36 @@
+///<reference path="../../global.d.ts" />
+import {
+    SEARCH_INPUT,
+    //@ts-ignore
```
thomasheartman, line 4: Out of curiosity: what are we ignoring here and why?

frontend/src/component/project/ProjectList/ProjectList.tsx
```diff
@@ -131,32 +131,34 @@ export const ProjectListNew = () => {
...
+            value: true,
+        },
+    ];
 
-        setSearchParams(tableState, {
-            replace: true,
-        });
-    }, [searchValue, setSearchParams]);
+    const stateConfig = {
+        myProjects: withDefault(BooleanParam, false),
```
thomasheartman, line 150: I'm not convinced this should be a boolean. Is there a reason you didn't use an enum? E.g. `projectFilter: 'all projects' | 'my projects'` or `projectFilter: undefined | 'my projects'`?

I'm thinking about this primarily from two perspectives:

1. From a future extension point of view. If we decide we want to add more filters, then tacking on another enum value is easy. But if we use a boolean we need a new property and then need to consider how `myProjects` interacts with the new property.

2. For code clarity: it was not clear to me why `filterOptions` had values `true` and `false`. I at first assumed it had to do with the default value or something.

thomasheartman (commented):
Nice one! 🙌🏼  I like the general direction here, but I'm not convinced about the use of a boolean for storing the filter state. Check out the comment I made on line 150 and let's have a discussion on it 😄

frontend/cypress/integration/projects/list.spec.ts
Tymek: Same as other E2E tests. Maybe it's related to previous Node or TS version and can be removed. I'll try

PR #6872, opened 2024-04-16
Task/use audit info in events

chriswk:
So, this is the extension to the auditInfo middleware.

I've tried to use/add the audit info to all events I could see/find. This makes this PR necessarily huge, because we do store quite a few events here and there. I realise it might not be complete yet, but tests run green, and I think we now have a pattern to follow for other events. Merging this will probably break enterprise quite hard, so will prepare a PR on enterprise tomorrow which is linked against this.


src/lib/features/dependent-features/dependent-features-service.ts
```diff
@@ -148,82 +149,81 @@ export class DependentFeaturesService {
...
+                },
+            }),
+        );
     }
 
     async deleteFeatureDependency(
         dependency: FeatureDependencyId,
         projectId: string,
         user: IUser,
+        auditUser: IAuditUser,
```
gastonfournier, line 170: For the future, we can consider merging `auditUser` with `user` as they should always represent the same, but user has permissions and auditUser don't, but for now, let's do this as a gradual improvement.

src/lib/middleware/audit-middleware.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        let audit: IAuditUser | undefined;
+        app.get('/api/admin/test', (req: IAuthRequest, res) => {
+            audit = req.audit;
+            res.status(200).end();
+        });
+        const request = supertest(app);
+        await request.get('/api/admin/test').expect(200);
+        expect(audit).toBeDefined();
+        expect(audit!.id).toBe(-1);
+        expect(audit!.username).toBe('unknown');
```
gastonfournier, line 26: Just a note: this comes from `new NoAuthUser().id` and `new NoAuthUser().username` which is injected by the noAuthentication function above

src/lib/middleware/audit-middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+import { extractAuditInfo } from '../util';
+
+export const auditAccessMiddleware = ({
+    getLogger,
+}: Pick<IUnleashConfig, 'getLogger'>): any => {
+    const logger = getLogger('/middleware/audit-middleware.ts');
+    return (req: IAuthRequest | IApiRequest, _res, next) => {
+        if (!req.user) {
+            logger.info('Could not find user');
+        } else {
```
gastonfournier, line 12: Do we want to log everytime we don't find a user?
```suggestion
        if (req.user) {
```

gastonfournier (approved):
This looks good, just a minor comment, but no blocker

src/lib/features/dependent-features/dependent-features-service.ts
chriswk: yes, agreed, the thing missing was having some way from audit user to UserWithPermission. so the one thing audituser has that most other user implementations do not have is the ip.

src/lib/middleware/audit-middleware.ts
chriswk: Since this middleware is only mounted on `/api/admin` not finding a user might actually be an error.

PR #6881, opened 2024-04-17
chore: add enterprise extension point to enable change requests on project creation

thomasheartman:
This PR adds an optional function parameter to the `createProject` function that is intended to enable change requests for the newly created project.

The assumption is that all the logic within will be decided in the enterprise impl. The only thing we want to verify here is that it is called after the project has been created.


src/test/fixtures/fake-role-store.ts
thomasheartman, line 1: This change is to make the added test pass. Without it, it would throw an error because the fake role store couldn't find the owner role.

There is an optional workaround that also works: we can **not** provide an ID on the creating user and ts-ignore that it doesn't conform to the interface. But this felt like a reasonable improvement to me.

Tymek (approved):

PR #6883, opened 2024-04-17
feat: optimize applications overview query

sjaanus:
We encountered an issue with a customer because this query was returning 3 million rows. The problem arose from each instance reporting approximately 100 features, with a total of 30,000 instances. The query was joining these, thus multiplying the data. This approach was fine for a reasonable number of instances, but in this extreme case, it did not perform well.

This PR modifies the logic; instead of performing outright joins, we are now grouping features by environment into an array, resulting in just one row returned per instance.

I tested locally with the same dataset. Previously, loading this large instance took about 21 seconds; now it has reduced to 2 seconds. Although this is still significant, the dataset is extensive.

### DIFF
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -298,51 +298,6 @@ export default class ClientApplicationsStore
         return mapRow(row);
     }
 
-    async getOldApplicationOverview(
-        appName: string,
-    ): Promise<IApplicationOverview> {
-        const stopTimer = this.timer('getApplicationOverviewOld');
-        const query = this.db
-            .with('metrics', (qb) => {
-                qb.distinct(
-                    'cme.app_name',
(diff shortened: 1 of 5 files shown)


src/lib/db/client-applications-store.ts
```diff
@@ -294,33 +294,39 @@ export default class ClientApplicationsStore
...
         const query = this.db
             .with('metrics', (qb) => {
-                qb.distinct(
+                qb.select([
                     'cme.app_name',
                     'cme.environment',
-                    'cme.feature_name',
-                ).from('client_metrics_env as cme');
+                    'f.project',
+                    this.db.raw(
```
sjaanus, line 301: Group features into array, grouped by environment.

FredrikOseberg (approved):
LGTM

PR #6886, opened 2024-04-18
feat: split projects view into "my projects" and "other projects"

thomasheartman:
This PR removes the previous "my projects" filter in favor always splitting projects, but showing both on the main screen.

To make it a bit easier to work with, it also moves the project group component into its own file, causing some extra lines of code change. My apologies 🙇🏼


frontend/src/component/project/ProjectList/ProjectList.tsx
```diff
@@ -176,7 +154,26 @@ export const ProjectListNew = () => {
...
+        for (const project of filteredProjects) {
+            console.log('handling', project);
+            if (shouldDisplayInMyProjects(myProjects)(project)) {
+                my.push(project);
+            } else {
+                other.push(project);
+            }
+        }
+        return { my, other };
+    }, [filteredProjects, myProjects, splitProjectList]);
```
thomasheartman, line 176: I wanna extract this into its own function so that I can test it properly.

frontend/src/component/project/ProjectList/ProjectList.tsx
```diff
@@ -222,42 +207,104 @@ export const ProjectListNew = () => {
...
+                                        )}
+                                    </>
+                                )}
+                            />
+                        </StyledItemsContainer>
+                    }
+                />
+            </StyledProjectGroupContainer>
+        );
+    };
```
thomasheartman, line 300: This is just extracting the component that was already there, so that we can use it in two places. However, it feels pretty gnarly. 

I'm unsure about moving it out into its own file, though, because then we'd have to add five extra parameters to the function:
- `loading`
- `searchValue`
- `CardComponent`
- `ContainerComponent`
- `handleHover`

Still, it might be a neater approach. What do you think?

thomasheartman: Update: I went ahead with extracting this into its own file. I think it's the right thing to do.

Tymek (approved):

PR #6887, opened 2024-04-18
chore: Remove unused type

gastonfournier:
This type was added to try to consolidate events now we're moving in a different direction and this type is not used.


chriswk (approved):
LGTM

PR #6892, opened 2024-04-19
feat: Make Database SSL Configurable through files

### LINKED ISSUE #6718: Easier DB CA certificate configuration

I'd like to have an ability to configure path to CA certificate instead of having to put file contents as an environment variable. Currently it's only possible by creating my own Docker image.

### Background

When using AWS Helm chart it's hard to interpolate variables, like reading file contents with `cat`. Shell is not available. Also, CA files can be longer then allowed secret length.

### Solution suggestions

In addition to `DATABASE_SSL` if we add an environment variable like `DATABASE_CA` we can point it to a file mounted in RDS.

```
volumeMounts:
        - name: rds-pem
          mountPath: /certificates
env:
          DATABASE_HOST: dfdsfdsfdfdfdf.rds.amazonaws.com
          DATABASE_NAME: unleash
          DATABASE_CA: /certificates/regional-bundle.pem
```

### Other informations
Request source: https://unleash-community.slack.com/archives/C03GWTN7XMG/p1711446883483999
Internal ticket: https://linear.app/unleash/issue/SR-509/easier-db-ca-certificate-configuration

chriswk:
This makes it configurable either through a single JSON file with all three certificates as separate keys or via separate files per ca/cert/key key.

fixes #6718

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -1,25 +1,26 @@
 import { parse } from 'pg-connection-string';
 import merge from 'deepmerge';
 import * as fs from 'fs';
+import { readFileSync } from 'fs';
 import {
-    type IUnleashOptions,
-    type IUnleashConfig,
-    type IDBOption,
-    type ISessionOption,
-    type IServerOption,
-    type IVersionOption,
     type IAuthOption,
     IAuthType,
-    type IImportOption,
-    type IEmailOption,
-    type IListeningPipe,
-    type IListeningHost,
-    type IUIConfig,
+    type IClientCachingOption,
     type ICspDomainConfig,
     type ICspDomainOptions,
-    type IClientCachingOption,
+    type IDBOption,
+    type IEmailOption,
+    type IImportOption,
+    type IListeningHost,
+    type IListeningPipe,
     type IMetricsRateLimiting,
     type IRateLimiting,
+    type IServerOption,
+    type ISessionOption,
+    type IUIConfig,
+    type IUnleashConfig,
+    type IUnleashOptions,
+    type IVersionOption,
 } from './types/option';
 import { getDefaultLogProvider, LogLevel, validateLogProvider } from './logger';
 import { defaultCustomAuthDenyAll } from './default-custom-auth-deny-all';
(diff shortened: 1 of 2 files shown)


src/lib/create-config.ts
```diff
@@ -182,17 +183,63 @@ const dateHandlingCallback = (connection, callback) => {
...
+const filesExists = (...filePath: string[]): boolean => {
+    return filePath.every((file) => fs.existsSync(file));
+};
+
+const databaseSsl = () => {
+    if (process.env.DATABASE_SSL != null) {
+        return JSON.parse(process.env.DATABASE_SSL);
+    } else if (
+        process.env.DATABASE_SSL_CA_CONFIG != null &&
+        filesExists(process.env.DATABASE_SSL_CA_CONFIG)
```
sighphyre, line 203: Not convinced about the file existence checks here. Is there a place where we want the file check to return true but return an empty object if it's unreadable or contains broken JSON?

Maybe it's the Python dev in me but I'd lean on just reading the thing and trapping the exception case

src/lib/create-config.ts
```diff
@@ -182,17 +183,63 @@ const dateHandlingCallback = (connection, callback) => {
     });
 };
 
+const databaseSslFromFile = (caFilePath: string) => {
+    try {
+        return readFileSync(caFilePath).toJSON();
+    } catch (e) {
+        return {};
```
sighphyre, line 190: I'd be tempted to just bubble here. If you asked for this to happen and it didn't, I'd be rather surprised to find a quiet fail without a log

chriswk: Agreed. Let's fail hard, we do in other paths for config.

src/lib/create-config.ts
chriswk: I think that's a fair point, I think failing on non-existing files if you use the env vars is fair enough

alvinometric:
Just to confirm, this doesn't change the default username and passwords that are setup when self-hosting, right? It just lets you configure it.

chriswk:
> Just to confirm, this doesn't change the default username and passwords that are setup when self-hosting, right? It just lets you configure it.

It doesn't touch anything for default username and password, it's concerned with configuring a secure connection to the database.
I see the autoformatter has touched quite a bit of the configuring-unleash file that did not receive any actual eyeballs on it. Just to confirm that the new keys are
DATABASE_SSL_CA_FILE: sets `ssl: { ca }`
DATABASE_SSL_KEY_FILE: sets `ssl: { key }`
DATABASE_SSL_CERT_FILE: sets `ssl: { cert }`
--
DATABASE_SSL_CA_CONFIG: expects a json file with `{ rejectUnauthorized: bool, ca: string, key: string, cert: string }`
--
DATABASE_SSL_REJECT_UNAUTHORIZED: sets `ssl: { rejectUnauthorized }`

alvinometric:
Perfect, thanks

sighphyre (approved):
Ye! Code looks good, so I'm gonna tack on an approve. I'll put comments on the docs if I spot anything

Wow, IntelliJ got a little intense with the whitespace here but the docs look good at a glance

src/lib/create-config.ts
```diff
@@ -182,17 +183,53 @@ const dateHandlingCallback = (connection, callback) => {
     });
 };
 
+const databaseSsl = () => {
+    if (process.env.DATABASE_SSL != null) {
+        return JSON.parse(process.env.DATABASE_SSL);
+    } else if (process.env.DATABASE_SSL_CA_CONFIG != null) {
+        return readFileSync(process.env.DATABASE_SSL_CA_CONFIG).toJSON();
+    } else if (
+        process.env.DATABASE_SSL_KEY_FILE != null &&
```
egor-xyz, line 192: Why is the DATABASE_SSL_KEY_FILE mandatory? For example, AWS provides only a CA file without a KEY file.

egor-xyz:
> > Just to confirm, this doesn't change the default username and passwords that are setup when self-hosting, right? It just lets you configure it.
> 
> It doesn't touch anything for default username and password, it's concerned with configuring a secure connection to the database. I see the autoformatter has touched quite a bit of the configuring-unleash file that did not receive any actual eyeballs on it. Just to confirm that the new keys are
> 
> * `DATABASE_SSL_CA_FILE`: sets `ssl: { ca }`
> * `DATABASE_SSL_KEY_FILE`: sets `ssl: { key }`
> * `DATABASE_SSL_CERT_FILE`: sets `ssl: { cert }`
> 
> * DATABASE_SSL_CA_CONFIG: expects a path to a json file with `{ rejectUnauthorized: bool, ca: string, key: string, cert: string }`, currently, the `DATABASE_SSL` option expects a stringified json object with the same keys, set in an environment variable. This is often hard to manage in various kubernetes setups. it's better to be able to write an actual json file to disk, and then read this in to memory in Unleash.
> 
> * DATABASE_SSL_REJECT_UNAUTHORIZED: sets `ssl: { rejectUnauthorized }`

Can I set only the CA file path without `DATABASE_SSL_KEY_FILE` and `DATABASE_SSL_CERT_FILE`? AWS provides me only one CA file, and it's enough for connection to DB

PR #6893, opened 2024-04-19
feat: make edge use token's cache

gastonfournier:
## About the changes
This PR removes the feature flag `queryMissingTokens` that was fully rolled out.
It introduces a new way of checking edgeValidTokens controlled by the flag `checkEdgeValidTokensFromCache` that relies in the cached data but hits the DB if needed. 

The assumption is that most of the times edge will find tokens in the cache, except for a few cases in which a new token is queried. From all tokens we expect at most one to hit the DB and in this case querying a single token should be better than querying all the tokens.

### DIFF
--- a/src/lib/services/api-token-service.test.ts
+++ b/src/lib/services/api-token-service.test.ts
@@ -18,31 +18,6 @@ import FakeFeatureTagStore from '../../test/fixtures/fake-feature-tag-store';
 import { createFakeEventsService } from '../../lib/features';
 import { extractAuditInfoFromUser } from '../util';
 
-const token: IApiTokenCreate = {
-    environment: 'default',
-    projects: ['*'],
-    secret: '*:*:some-random-string',
-    type: ApiTokenType.CLIENT,
-    tokenName: 'new-token-by-another-instance',
-    expiresAt: undefined,
-};
-
-const setup = (options?: IUnleashOptions) => {
-    const config: IUnleashConfig = createTestConfig(options);
-    const apiTokenStore = new FakeApiTokenStore();
-    const environmentStore = new FakeEnvironmentStore();
-
-    const apiTokenService = new ApiTokenService(
-        { apiTokenStore, environmentStore },
-        config,
-        createFakeEventsService(config),
-    );
-    return {
-        apiTokenService,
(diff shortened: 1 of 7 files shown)


src/lib/services/api-token-service.test.ts
```diff
@@ -220,60 +220,43 @@ describe('When token is added by another instance', () => {
...
-        });
+        const { apiTokenService, apiTokenStore } = setup();
         const apiTokenStoreGet = jest.spyOn(apiTokenStore, 'get');
 
         const invalidToken = 'invalid-token';
-        for (let i = 0; i < 10; i++) {
+        for (let i = 0; i < 5; i++) {
             expect(
-                await apiTokenService.getUserForToken(invalidToken),
+                await apiTokenService.getTokenWithCache(invalidToken),
```
gastonfournier, line 250: Reducing the scope of the test to just fetching the token

src/lib/services/api-token-service.ts
```diff
@@ -97,66 +96,21 @@ export class ApiTokenService {
...
-        } catch (e) {
-            this.logger.error('Unable to create initial Admin API tokens');
-        }
-    }
-
-    public async getUserForToken(
-        secret: string,
-        flagContext?: IFlagContext, // temporarily added, expected from the middleware
-    ): Promise<IApiUser | undefined> {
+    async getTokenWithCache(secret: string): Promise<IApiToken | undefined> {
```
gastonfournier, line 113: This method is extracted from `getUserForToken` it was moved up to be close to `getToken` method above

src/lib/services/edge-service.ts
```diff
@@ -10,30 +10,60 @@ export default class EdgeService {
...
+            for (const token of tokens) {
+                const found =
+                    await this.apiTokenService.getTokenWithCache(token);
+                if (found) {
+                    validatedTokens.push({
+                        token: token,
+                        type: found.type,
+                        projects: found.projects,
+                    });
+                }
```
gastonfournier, line 42: New behavior checking the tokens one by one, hitting the cache unless one is not present. Invalid tokens are also cached

PR #6894, opened 2024-04-19
Correctly applied styles to `Figure` component in docs

alvinometric:
It turns out the styles that we had in `style.modules.css` were overridden by browser defaults.

## Before

image

## After

image

Now taking into account all the styles that were in the CSS files (notice margin and border)


thomasheartman (approved):
Looks good to me 👍🏼

PR #6900, opened 2024-04-22
Adds contributors to docs

alvinometric:
Hello!

We wanted to make the docs less impersonal, so we decided to add contributors. Now each doc page that has an `editUrl` (i.e, isn't generated) shows a list of everyone that contributed to it.

This list is generated by:
1. Running `swizzle` on the `DocItem/Footer` in Docusaurus.
2. Grabbing metadata for the current file using an internal docusaurus API (Thank you to @homotechsual for the help there)
3. Getting the commits to the file in question with the GitHub API

image

<details>
 <summary> Here's the command I ran, for posterity </summary>
 <code>npm run swizzle @docusaurus/theme-classic DocItem/Footer  -- --wrap</code>
</details>

## Discussion points

1. Design. What do you think of the layout?
2. Right now I'm hardcoding the info of Unleash team members. This creates a small maintenance burden, but it's something we wanted to add.


thomasheartman (commented):
This is cool 😄 Left a couple comments on the implementation in the code, but overall I think it's nice 👏🏼 

Now, on your discussion points:

1. I think the design is fine. It feels a bit wonky in that if you have a job title, then that pushes your username up etc., but it's not a big deal. If you want some design input, how about talking to Fredrik?

2. Hard coding the Unleash members. I .. don't love it, but I think it's fine if we think that it's good for visibility. I guess we could fetch that data from github somehow too, but not sure it's worth the complexity. I'm not convinced we should do titles, though. How about just showing a small Unleash logo next to Unleash members or something? Not a show stopper, but just an idea. However, if you do go ahead: those titles aren't correct, and I don't think we should use titles that aren't. 

More importantly: the SDK overview doesn't work anymore: https://unleash-docs-git-alvin-docs-contributors-unleash-team.vercel.app/reference/sdks, so I think there's some cases you're not catching in the code 👀

website/src/theme/DocItem/Footer/contributors.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+    if (!contributors.length) {
+        return <h3>Fetching contributors...</h3>;
+    }
+
+    return (
+        <div className={styles.contributors}>
+            <h3>Contributors</h3>
+
+            <div className={styles.wrapper}>
```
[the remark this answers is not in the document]
alvinometric: Good call

website/src/theme/DocItem/Footer/contributors.jsx
```diff
@@ -0,0 +1,87 @@
...
+                {contributors?.map((contributor) => {
+                    return (
+                        <figure className={styles.contributor}>
+                            <a href={contributor.html_url}>
+                                <img
+                                    src={contributor.avatar_url}
+                                    alt={contributor.login}
```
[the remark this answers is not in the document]
alvinometric: I think this could imply that they work at GitHub, maybe we could do @alvinometric for example?

website/src/theme/DocItem/Footer/contributors.jsx
```diff
@@ -0,0 +1,87 @@
...
+                                    width={70}
+                                    style={{ borderRadius: '100%' }}
```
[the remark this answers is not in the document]
alvinometric: No reason, moved it

website/src/theme/DocItem/Footer/contributors.jsx
```diff
@@ -0,0 +1,87 @@
...
+    if (!contributors.length) {
+        return <h3>Fetching contributors...</h3>;
+    }
+
+    return (
+        <div className={styles.contributors}>
+            <h3>Contributors</h3>
+
+            <div className={styles.wrapper}>
+                {contributors?.map((contributor) => {
```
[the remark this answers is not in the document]
alvinometric: It's sorted alphabetically now, with priority given to Unleash team members

website/src/theme/DocItem/Footer/contributors.jsx
```diff
@@ -0,0 +1,87 @@
...
+                    console.error(error);
+                    setContributors([]);
+                });
+        };
+
+        fetchFileContributors();
+    }, []);
+
+    if (!contributors.length) {
+        return <h3>Fetching contributors...</h3>;
```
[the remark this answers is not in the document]
alvinometric: The problem is then I have to nest everything else in an if. I'm not sure it justifies the UX benefits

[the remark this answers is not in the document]
thomasheartman: Huh? Not sure I understand what you mean here. You could just do this instead, right?

```jsx
        <div className={styles.contributors}>
            <h3>Contributors</h3>
            <p>Fetching contributors ...</p>
        </div>
```

That feels more semantically correct to me. Changing headers dynamically seems a bit strange to me due to landmark navigation.

thomasheartman:
It seems you fixed the crashing page error. Did you find out what was wrong there? For the future, we might want to consider adding feature flags to the docs again, so that we can disable new features like this if they slip by. It was only by accident that I caught the crash 🤔  But that's for later, for sure.

Regarding the design, it looks kinda messy when you have lots of contributors:
image

Should we consider formatting it all vertically instead? Or how about just showing the images? That way we know that all images will be the same size, so it looks neat 💁🏼  Or we give each one a set size, and cut text if it overflows?

thomasheartman:
I suspect that using just the avatars will be enough. It's also consistent with how we (and many other people) do it on github, for instance

thomasheartman:
Oh, and we should also find a way to deal with SDK pages (such as https://unleash-docs-git-alvin-docs-contributors-unleash-team.vercel.app/reference/sdks/flutter). They currently only show the header and never find the right data:

image

This is because these docs aren't hosted in the Unleash repo itself, but on the various SDKs/Edge/Proxy repos. These docs are essentially just the readme for those repos. However, the edit page link still works and takes you to the right repo.

You can find more information about this in the `remote-content` directory. The docusaurus config contains the following imports:

```js
const { sdks } = require('./remote-content/sdks');
const { docs: edgeAndProxy } = require('./remote-content/edge-proxy');
```

Those files contain all the SDK links, where they get sourced from, etc.

website/src/theme/DocItem/Footer/contributors.jsx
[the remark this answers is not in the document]
thomasheartman: I guess it could? Doesn't really read that way to me, but I'm not particularly bothered about this. My goal here is to make it clear to the end user where the link goes, but I think you can inspect the link for it, so I'm easy.

alvinometric:
Hah, you reviewed before I could comment, so to address the initial points:

1. Yes, @FredrikOseberg  any inputs on the design?
2. For sure, I'll rectify the titles once we're good on the code side.

Re: the SDKs, I should be able to retrieve the contributors for these too, I just need to tweak the code a bit.

FredrikOseberg:
> Hah, you reviewed before I could comment, so to address the initial points:
> 
> 1. Yes, @FredrikOseberg  any inputs on the design?
> 2. For sure, I'll rectify the titles once we're good on the code side.
> 
> Re: the SDKs, I should be able to retrieve the contributors for these too, I just need to tweak the code a bit.

I think the easiest path forward is to only display avatars and to display more information on hover/focus. The key to keep it looking good and consistent is to make sure everything has the same size, so when we include user input this quickly gets hard to manage because titles and names will differ in size leading to the perceived disorganized layout. This is similar to how for example github does it: 

Skjermbilde 2024-04-23 kl  13 56 19

alvinometric:
What about adding a logo for Unleash team members?

alvinometric:
https://github.com/Unleash/unleash/assets/107407814/5a0ce8db-5534-47e1-ac75-f9a17edc58ea

Here's what it looks like now with icons and more info on hover

website/src/theme/DocItem/Footer/contributors.module.scss
```diff
@@ -0,0 +1,63 @@
...
+    right: -5px;
+    width: 30px;
+    height: 30px;
+    background-image: url(/img/logo.svg);
+}
+
+.contributor {
+    margin: 5px 10px 5px 0;
+    display: flex;
+    gap: 10px;
```
thomasheartman, line 36: Why does the contributor have display flex and/or gap? There's only one item, right? And instead of setting the margin on the contributor, wouldn't it be easier to use the `gap` property on the `.wrapper`?

Doesn't really matter too much, though, just curious 💁🏼

website/src/theme/DocItem/Footer/contributors.module.scss
```diff
@@ -0,0 +1,63 @@
...
+
+    img{
+        border-radius: 100%;
+        border: 2px solid var(--ifm-color-emphasis-300);
```
thomasheartman, line 40: I think there's an ifm border color variable that we use for most other things. That may be a better variable to use here.

website/src/theme/DocItem/Footer/contributors.module.scss
```diff
@@ -0,0 +1,63 @@
...
+    background-image: url(/img/logo.svg);
+}
+
+.contributor {
+    margin: 5px 10px 5px 0;
+    display: flex;
+    gap: 10px;
+
+    img{
+        border-radius: 100%;
```
thomasheartman, line 39: Any reason why we don't set this on the parent `a` tag instead? Would probably help the keyboard focus outline, at least.

website/static/img/logo.svg
thomasheartman, line 1: Changing this also affects other parts of the page where we use it, most notably the footer:

image

You might want to add a separate image instead.

thomasheartman (commented):
Nice, yeah, that looks pretty cool! 

A few things we could fix (though not urgent, so happy to have that done in a follow-up:

1. The keyboard focus. The outline looks misaligned (I'm guessing that's because you set border radius on the image rather than the link) and there's no effect when you focus it (though whether we want it to pop up on keyboard focus or not is a decision to make).
image

2. It's subtle, but the focus outline doesn't match the shape of the avatar when it zooms in. Notice the white edge on the lower right area of this avatar:
image

website/static/img/logo.svg
alvinometric: I think the new one looks better?

website/src/theme/DocItem/Footer/contributors.module.scss
alvinometric: Good call, less code!

website/src/theme/DocItem/Footer/contributors.module.scss
alvinometric: It doesn't work, it would only work if the image was a background

website/static/img/logo.svg
alvinometric: Oh wait I had it flipped, yeah that looks terrible, I'll fix it. 
The reason I updated it is the previous image had a ton of whitespace around it for no reason.

website/src/theme/DocItem/Footer/contributors.module.scss
thomasheartman: What if you set `overflow: hidden`?

website/static/img/logo.svg
thomasheartman: Yeah, that makes sense. Just adjust the size in the footer and you're golden.

alvinometric:
I think I've addressed everything, but to come back:

1. I think the focus is fixed now that I've removed the margins in favour of the `gap`
2. I can't reproduce it, again it's possible that it's a result of relying on `gap`

nnennandukwe (approved):
👍 lgtm

website/src/theme/DocItem/Footer/GitHubContributors.jsx
```diff
@@ -0,0 +1,93 @@
...
+    const url = `https://api.github.com/repos/${owner}/${repo}/commits?path=${filePath}`;
+
+    useEffect(() => {
+        const fetchFileContributors = () => {
+            fetch(url)
+                .then((response) => response.json())
+                .then((commits) => {
+                    console.log(commits);
+                    const contributors = getContributors(commits);
+                    console.log(contributors);
```
thomasheartman, line 35: Oh, we probably wanna remove these logs, though:
```suggestion
                    const contributors = getContributors(commits);
```

thomasheartman (commented):
Nice, yeah! I'm happy for this to go in now, but I think there's a few things we should update (but please, make it another PR, so we can get this in):

1. I'm still not convinced we should have titles for Unleash people, especially not when they're not correct titles. I'd just go with "Unleash employee" or something (maybe grab a special one for Ivar who can never not be the founder 💁🏼 )
2. The keyboard focus styles are still weird. I'd be happy to look into this after you've merged if you want? It also acts differently for Unleash people than it does for outside contributors:
  image
  image


Dynamically updating a header still feels iffy to me. Feels like it'll be bad for accessibility in terms of navigation etc, but I can't point to anything specific, so I might just be overly worried. I also think it'd make more sense to update the content than the header, but .. eh, it's not the end of the world.

alvinometric:
Thanks again, so

1. Yeah, let's make this another PR
2. Thanks for keeping me honest with this, I've updated it with the right titles now
3. I'm just retuning `null` now, better than having an empty header

PR #6903, opened 2024-04-23
feat: Stage timeline styling

kwasniew:
## About the changes

Screenshot 2024-04-23 at 08 55 48

Styling stage timeline.

Design decisions: 
* stages and lines modelled as centered flexbox elements
* stage box has :after content so that we can easily add active/inactive text below without affecting stages alignment
* active stage modelled as a speech bubble with a CSS drawn rectangle. Originally I tried to use the MUI tooltip but it opens only on hover


sjaanus (approved):
LG. I was thinking of making these with css, but svg is nice.

PR #6905, opened 2024-04-23
chore: add a bunch of logs to validate api token validation behavior

gastonfournier:
This change is meant to test something in sandbox. It will be reverted after the investigation.

### DIFF
--- a/src/lib/services/api-token-service.ts
+++ b/src/lib/services/api-token-service.ts
@@ -108,6 +108,9 @@ export class ApiTokenService {
     async fetchActiveTokens(): Promise<void> {
         try {
             this.activeTokens = await this.store.getAllActive();
+            this.logger.info(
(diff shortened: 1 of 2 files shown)


src/lib/services/api-token-service.ts
```diff
@@ -193,6 +213,7 @@ export class ApiTokenService {
         secret: string,
     ): Promise<IApiUser | undefined> {
         const token = await this.getTokenWithCache(secret);
+        this.logger.info(`getUserForToken ${secret} found: ${token}`);
```
sighphyre, line 216: That's probably gotta come out so just a reminder that we gotta revert this one when we're done

sighphyre (approved):
LGTM

src/lib/services/api-token-service.ts
gastonfournier: absolutely

PR #6906, opened 2024-04-23
chore: remove new frontend api feature flag

sjaanus:
The flag has been 100% for a bit now, we need to prepare for GA.


src/lib/features/frontend-api/frontend-api-service.e2e.test.ts
```diff
@@ -1,155 +0,0 @@
-import {
```
sjaanus, line 1: This was only testing the diff and had no tests left.

src/lib/features/frontend-api/frontend-api-service.ts
```diff
@@ -170,56 +136,13 @@ export class FrontendApiService {
...
-            this.logger.error('We found an event error', error);
-        });
-
-        await client.start();
-
         return client;
     }
 
-    private async createNewClientForFrontendApiToken(
+    private async createClientForFrontendApiToken(
```
sjaanus, line 145: Looking at lines 172 and 182 in this service, we did not do this for new clients during beta phase.
I renamed the `newClients` to `clients` and it auto-resolved, but I think it is correct to still have delete and destroy functions on the new clients.

src/lib/features/frontend-api/frontend-api-controller.ts
```diff
@@ -186,48 +183,11 @@ export default class FrontendAPIController extends Controller {
...
-                        req.user.environment
-                    }, diff ${JSON.stringify(
-                        diff(sortedToggles, sortedNewToggles),
-                    )}`,
-                );
-            }
-        } else if (
-            this.config.flagResolver.isEnabled('returnGlobalFrontendApiCache')
-        ) {
-            toggles =
```
kwasniew, line 219: just to make sure, is this the path that we return after the flag removal?

src/lib/features/frontend-api/frontend-api-controller.ts
```diff
@@ -186,48 +183,11 @@ export default class FrontendAPIController extends Controller {
...
-                await this.services.frontendApiService.getNewFrontendApiFeatures(
```
kwasniew, line 220: I'm assuming you renamed this to getFrontendApiFeatures so that we don't have old/new naming

kwasniew (approved):
To be on the safe side I'd manually test this before merging

src/lib/features/frontend-api/frontend-api-controller.ts
sjaanus: yes

src/lib/features/frontend-api/frontend-api-controller.ts
sjaanus: yes

PR #6907, opened 2024-04-23
chore: remove logs for secret and change invalid token query logic

gastonfournier:
## About the changes
What's going on is the following:
1. When a token is not found in the token's cache we try to find it in the db
2. To prevent a denial of service attack using invalid tokens, we cache the invalid tokens so we don't hit the db.
3. The issue is that we stored this token in the cache regardless we found it or not. And if the token was valid the first time we'd add a timestamp to avoid querying this token again the next time. 
4. The next iteration the token should be in the cache: https://github.com/Unleash/unleash/blob/54383a6578c221bdacdd1b9f14a108a1eb256e7c/src/lib/services/api-token-service.ts#L162 but for some reason it is not and therefore we have to make a query. But this is where the query prevention mechanism kicks in because it finds the token in the cache and kicks us out. This PR fixes this by only storing in the cache for misses if not found: https://github.com/Unleash/unleash/blob/54383a6578c221bdacdd1b9f14a108a1eb256e7c/src/lib/services/api-token-service.ts#L164-L165

The token was added to the cache because we were not checking if it had expired. Now we added a check and we also have a log for expired tokens. Some improvement opportunities:
- I don't think we display that a token has expired in the UI which probably led to this issue
- When a token expired we don't display a specific error message or error response saying that which is not very helpful for users

### DIFF
--- a/src/lib/services/api-token-service.test.ts
+++ b/src/lib/services/api-token-service.test.ts
@@ -12,7 +12,7 @@ import {
     API_TOKEN_UPDATED,
     TEST_AUDIT_USER,
 } from '../types';
-import { addDays, minutesToMilliseconds } from 'date-fns';
+import { addDays, minutesToMilliseconds, subDays } from 'date-fns';
 import EventService from '../features/events/event-service';
 import FakeFeatureTagStore from '../../test/fixtures/fake-feature-tag-store';
 import { createFakeEventsService } from '../../lib/features';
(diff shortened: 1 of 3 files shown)


src/lib/services/api-token-service.ts
```diff
@@ -145,36 +145,41 @@ export class ApiTokenService {
...
+                        token = undefined;
+                    } else {
+                        this.activeTokens.push(token);
+                    }
+                } else {
+                    // prevent querying the same invalid secret multiple times. Expire after 5 minutes
+                    this.queryAfter.set(secret, addMinutes(new Date(), 5));
+                }
+                stopCacheTimer();
+            } else {
```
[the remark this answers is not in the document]
sighphyre: Code scene kinda has a point here imo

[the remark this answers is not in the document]
sighphyre: And this one too

chriswk (approved):
Nice, though, do we really want our tests to write to console.log?

src/lib/services/api-token-service.test.ts
```diff
@@ -261,4 +261,26 @@ describe('API token getTokenWithCache', () => {
         }
         expect(apiTokenStoreGet).toHaveBeenCalledTimes(2);
     });
+
+    test('should not return the token if it has expired and shoud perform only one db query', async () => {
+        const { apiTokenService, apiTokenStore } = setup({
+            getLogger: () => ({
+                debug: console.log,
```
sighphyre, line 268: This makes tests very noisy, do we actually want this?

gastonfournier: Nope! I did that for testing only, removing!

src/lib/services/api-token-service.test.ts
```diff
@@ -261,4 +261,26 @@ describe('API token getTokenWithCache', () => {
...
+                fatal: console.log,
+                info: console.log,
+                warn: console.log,
+                error: console.log,
+            }),
+        });
```
gastonfournier, line 274:
```suggestion
        const { apiTokenService, apiTokenStore } = setup();
```

gastonfournier:
> LGTM comments optional

Yes, I'm aware of the additional complexity but for the sake of time and moving on to other important tasks I'll let this one slip

PR #6911, opened 2024-04-23
chore: remove log

gastonfournier:
This log is also not needed because we have the API status code 401 that shows the same data


sighphyre (approved):
LGTM

PR #6915, opened 2024-04-24
fix: now metrics in search will be aggregated across applications

sjaanus:
We found a bug indicating that the search was not aggregating metrics properly. The issue was that if there were multiple applications posting metrics, only the metrics from the first one were used.

Fixes implemented in this PR:

Do not join metrics into the main query, as they are not needed for filtering and paging.
Determine 'yes' and 'no' counts outside of the main query, within a CTE.


src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -212,23 +210,6 @@ class FeatureSearchStore implements IFeatureSearchStore {
                             '=',
                             'features.name',
                         );
-                    })
```
sjaanus, line 215: Now not joining in into main query, because not neede for filtering, paging

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -264,13 +245,48 @@ class FeatureSearchStore implements IFeatureSearchStore {
...
+                    ]);
+            })
             .select('*')
             .from('ranked_features')
             .innerJoin(
                 'final_ranks',
                 'ranked_features.feature_name',
                 'final_ranks.feature_name',
             )
+            .leftJoin('metrics', (qb) => {
```
sjaanus, line 279: Join it later, just to get the data.

kwasniew (approved):

src/lib/features/feature-search/feature-search-store.ts
LukasHeimann: I think this change might have introduced a bug -- when you join here, fields of `metrics` are also in the final result, duplicating the column names `feature_name` and `environment`. When there are no metrics, the fields can be `null` here (leftJoin, after all), causing problems with the aggregation of the results.

https://github.com/Unleash/unleash/issues/6988

PR #6916, opened 2024-04-24
Project owners read model - db read

Tymek:
## About the changes
Implementation of the logic for fetching project owners.

Based on pattern documented in https://docs.getunleash.io/contributing/ADRs/back-end/write-model-vs-read-models

### Next up
- Create and test function combining projects list result with project owners read model results.
- Create Schema update
- Add project owners to the response when the feature flag is enabled

Internal ticket https://linear.app/unleash/issue/1-2318/get-project-owner-users-and-groups-from-db

### DIFF
--- a/src/lib/features/project/project-owners-read-model.test.ts
+++ b/src/lib/features/project/project-owners-read-model.test.ts
@@ -6,6 +6,9 @@ import { ProjectOwnersReadModel } from './project-owners-read-model';
 
 describe('unit tests', () => {
     test('maps owners to projects', () => {});
+    test('returns "system" when a project has no owners', async () => {
+        // this is a mapping test; not an integration test
+    });
 });
 
 let db: ITestDb;
@@ -107,18 +110,6 @@ describe('integration tests', () => {
         });
     });
 
-    test('returns "system" when a project has no owners', async () => {
-        const projectId = randomId();
-
-        await db.stores.projectStore.create({ id: projectId, name: projectId });
-
-        // fetch project owners
-        const owners = await readModel.getAllProjectOwners();
-
-        expect(owners).toMatchObject({
-            [projectId]: [{ type: 'system' }],
-        });
-    });
     test('gets project group owners', async () => {});
     test('users are listed before groups', async () => {});
     test('owners (users and groups) are sorted by when they were added; oldest first', async () => {});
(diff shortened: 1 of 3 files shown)


src/lib/features/project/project-owners-read-model.ts
```diff
@@ -0,0 +1,151 @@
...
+                groupsDict[project] = [...groupsDict[project], data];
+            } else {
+                groupsDict[project] = [data];
+            }
+        });
+
+        return groupsDict;
+    }
+
+    async getAllProjectOwners(): Promise<ProjectOwnersDictionary> {
```
Tymek, line 120: Splitting it in 2 private methods makes the main one more readable to me, so that's what I did

src/lib/features/project/project-owners-read-model.test.ts
```diff
@@ -0,0 +1,175 @@
...
+let member: IUser;
+let group: IGroup;
+
+beforeAll(async () => {
+    db = await dbInit('project_owners_read_model_serial', getLogger);
+    readModel = new ProjectOwnersReadModel(db.rawDatabase, db.stores.roleStore);
+    ownerRoleId = (await db.stores.roleStore.getRoleByName(RoleName.OWNER)).id;
+
+    const ownerData = {
+        name: 'Owner User',
```
thomasheartman, line 28: Suggest changing this to "Owner Name" to be consistent with the member. (would also need to update the test on line 80.) But up to you 💁🏼 

```suggestion
        name: 'Owner Name',
```

src/lib/features/project/project-owners-read-model.ts
```diff
@@ -0,0 +1,151 @@
...
+            projects.map((project) => {
+                return [
+                    project,
+                    [
+                        ...(usersDict[project] || []),
+                        ...(groupsDict[project] || []),
+                    ],
+                ];
+            }),
+        );
```
thomasheartman, line 139: This is optional, but there's a lot of spreading here. We're not necessarily doing it in a loop, but it's a lot more iterations than we need. How about doing something like this instead (coded in github, so might not work directly 😅 ):

```suggestion
        for (const [project, groups] of Object.entries(groupsDict) {
          if (project in usersDict) {
            usersDict[project] = usersDict[project].concat(groups)
          } else {
            usersDict[project] = groups
          }
        }
```

and then return the `usersDict` instead? That'll save us a lot of iteration and feels more readable to me 🤷🏼

thomasheartman (approved):

src/lib/features/project/project-owners-read-model.test.ts
Tymek: done in next PR

src/lib/features/project/project-owners-read-model.ts
Tymek: done in next PR

PR #6917, opened 2024-04-24
feat: Completed stage UI

kwasniew:
## About the changes

Completed but active traffic in prod for the last 2 days:
Screenshot 2024-04-24 at 10 31 43

Completed and safe to archive since no traffic in the last 2 days:
Screenshot 2024-04-24 at 10 31 30

Details:
* currently a decision if it's safe to archive is done frontend side but we may consider moving it server side


FredrikOseberg:
Archive button text seems to be a lot larger than the usual button size. Could we use the default size?

kwasniew:
@FredrikOseberg I set the size to the smallest one we have in the sizes. The only way to achieve the small font is to !important fontSize since explicit fontSize is overwritten by the permission button component.

sjaanus (approved):

PR #6920, opened 2024-04-24
feat: return lifecycle state in feature overview

kwasniew:
## About the changes

Exposing current lifecycle stage in the feature API.

Screenshot 2024-04-24 at 13 54 11

Design decisions:
* exposing read model for reading a current stage and injecting it to the feature toggle service
* stages are ordered by fixed rules and even when they were created at mixed order we respect the fixed stage order
* lifecycle is added to feature in a similar fashion to dependencies
* lifecycle can be undefined when no stages were recorded


sjaanus (approved):
LG

PR #6921, opened 2024-04-24
feat: connect feature lifecycle to real API

kwasniew:
## About the changes

* reading lifecycle data from the feature API
* if no lifecycle data so not show lifecycle UI
* each stage gets corresponding environment metrics info from dev or prod (data combined from lifecycle and env metrics)


frontend/src/interfaces/featureToggle.ts
```diff
@@ -49,6 +49,10 @@ export interface IFeatureToggle {
     impressionData: boolean;
     strategies?: IFeatureStrategy[];
     dependencies: Array<IDependency>;
+    lifecycle?: {
```
kwasniew, line 52: we don't use orval here

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/FeatureOverviewMetaData.tsx
```diff
@@ -82,17 +82,7 @@ const FeatureOverviewMetaData = () => {
...
-        status: 'kept',
-        environments: [
-            { name: 'production', lastSeenAt: new Date().toISOString() },
-            {
-                name: 'staging',
-                lastSeenAt: new Date().toISOString(),
-            },
-        ],
-    };
+    const currentStage = populateCurrentStage(feature);
```
kwasniew, line 85: read real feature instead of hardcoded data

sjaanus (approved):

PR #6928, opened 2024-04-25
feat: map project owners to projects list

Tymek:
## About the changes
Follow-up for https://github.com/Unleash/unleash/pull/6916
- Combining list of projects with owners
- Additional tests and checks 

Internal ticket: https://linear.app/unleash/issue/1-2320/project-owners-mapping

### DIFF
--- a/src/lib/features/project/project-owners-read-model.test.ts
+++ b/src/lib/features/project/project-owners-read-model.test.ts
@@ -118,7 +118,7 @@ afterEach(async () => {
 });
 
 describe('integration tests', () => {
-    test('returns an empty list if there are no projects', async () => {
+    test('returns an empty object if there are no projects', async () => {
         const owners = await readModel.getAllProjectOwners();
 
         expect(owners).toStrictEqual({});
@@ -254,6 +254,7 @@ describe('integration tests', () => {
         const projectId = randomId();
         await db.stores.projectStore.create({ id: projectId, name: projectId });
 
+        // Raw query in order to set the created_at date
         await db.rawDatabase('role_user').insert({
             user_id: owner2.id,
             role_id: ownerRoleId,
@@ -261,6 +262,7 @@ describe('integration tests', () => {
             created_at: new Date('2024-01-01T00:00:00.000Z'),
         });
 
+        // Raw query in order to set the created_at date
         await db.rawDatabase('group_role').insert({
             group_id: group2.id,
             role_id: ownerRoleId,
@@ -309,7 +311,7 @@ describe('integration tests', () => {
         });
     });
 
-    test('enriches fully', async () => {
+    test('does not modify an empty array', async () => {
         const projectsWithOwners = await readModel.addOwners([]);
 
         expect(projectsWithOwners).toStrictEqual([]);
--- a/src/lib/features/project/project-owners-read-model.test.ts
+++ b/src/lib/features/project/project-owners-read-model.test.ts
@@ -4,10 +4,53 @@ import { type IUser, RoleName, type IGroup } from '../../types';
 import { randomId } from '../../util';
 import { ProjectOwnersReadModel } from './project-owners-read-model';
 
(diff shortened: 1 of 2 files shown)


src/lib/features/project/project-owners-read-model.ts
```diff
@@ -122,30 +129,31 @@ export class ProjectOwnersReadModel {
...
+            dict[project] = usersDict[project];
+        });
+
+        Object.keys(groupsDict).forEach((project) => {
+            if (project in dict) {
+                dict[project] = [...dict[project], ...groupsDict[project]];
+            } else {
+                dict[project] = groupsDict[project];
+            }
+        });
```
thomasheartman, line 147: Nice! Still a few small things that I think can make this faster:
1. Why do we create a new dict? If we reuse the users dictionary, then we skip an iteration of those keys and save on some memory space. Yes, mutation can be iffy, but this feels like a controlled place.
2. I think `concat` is faster than spreading and two arrays into a new one (at least from my limited reading), so I'd suggest going with that instead.

Thoughts?

src/lib/features/project/project-owners-read-model.test.ts
```diff
@@ -4,10 +4,53 @@ import { type IUser, RoleName, type IGroup } from '../../types';
...
+            projects,
+            owners,
+        );
+
+        expect(projectsWithOwners).toMatchObject([
+            { name: 'project1', owners: [{ ownerType: 'system' }] },
+            { name: 'project2', owners: [{ ownerType: 'system' }] },
+        ]);
     });
 });
```
thomasheartman, line 55: Can be done later, but I think we should put the unit tests in a separate file? So we'll probably have `project-owners-read-model.test.ts` and `project-owners-read-model.integration.test.ts`. Feels like it jives better with what we're doing most the time.

src/lib/features/project/project-owners-read-model.test.ts
```diff
@@ -64,6 +118,12 @@ afterEach(async () => {
...
 
 describe('integration tests', () => {
+    test('returns an empty list if there are no projects', async () => {
```
thomasheartman, line 121: We changed the return type here, so let's update the test name to match

```suggestion
    test('returns an empty object if there are no projects', async () => {
```

src/lib/features/project/project-owners-read-model.test.ts
```diff
@@ -155,21 +215,132 @@ describe('integration tests', () => {
...
+            project: projectId,
+            created_at: new Date('2024-01-01T00:00:00.000Z'),
+        });
+
+        await db.rawDatabase('group_role').insert({
+            group_id: group2.id,
+            role_id: ownerRoleId,
+            project: projectId,
+            created_at: new Date('2024-01-01T00:00:00.000Z'),
+        });
```
thomasheartman, line 269: This one is using the raw database instead of the store to manually add the date, right? Can we mention that in a small comment, just so that the next person doesn't have to wonder?

src/lib/features/project/project-owners-read-model.test.ts
```diff
@@ -155,21 +215,132 @@ describe('integration tests', () => {
...
+                },
+                {
+                    name: 'Group Name',
+                    ownerType: 'group',
+                },
+            ],
+        });
     });
 
     test('enriches fully', async () => {
```
thomasheartman, line 312: Can do it later, but we gotta update this test name 😅

thomasheartman (approved):

src/lib/features/project/project-owners-read-model.ts
Tymek: Ok, I'll update it. I find it more readable in a previous form. I don't think the optimization is worth it

src/lib/features/project/project-owners-read-model.test.ts
thomasheartman: Fair enough 💁🏼

PR #6930, opened 2024-04-25
fix: flag trends should support bigint for yes and no evaluations

sjaanus:
Recently we see some pods failing with inserting yes, no values that were over int. Increasing type to bigint.


kwasniew (approved):
what about other columns in this table?

PR #6931, opened 2024-04-25
feat: include CR envs enabled on creation in event and update validation

thomasheartman:
This PR improves the handling of change request enables on project creation in two ways:

1. We now verify that the envs you try to enable CRs for exist before passing them on to the enterprise functionality.
2. We include data about environments and change request environments in the project created events.

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -262,7 +262,7 @@ export default class ProjectService {
 
         if (invalidEnvs.length > 0) {
             throw new BadDataError(
-                `These environments do not exist and can not be selected for the project: ${invalidEnvs
+                `These environments do not exist: ${invalidEnvs
                     .map((env) => `'${env}'`)
                     .join(', ')}.`,
             );
@@ -288,10 +288,9 @@ export default class ProjectService {
         newProject: CreateProject,
         user: IUser,
         auditUser: IAuditUser,
-        enableChangeRequestsForSpecifiedEnvironments: (args: {
-            environments: CreateProject['changeRequestEnvironments'];
-            validateEnvironments: (environments: string[]) => Promise<void>;
-        }) => Promise<
+        enableChangeRequestsForSpecifiedEnvironments: (
+            environments: CreateProject['changeRequestEnvironments'],
+        ) => Promise<
             void | ProjectCreated['changeRequestEnvironments']
         > = async () => {
             return;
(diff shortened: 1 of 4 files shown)


src/lib/features/project/project-service.ts
```diff
@@ -1357,11 +1385,12 @@ export default class ProjectService {
     }
 
     // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-    removeModeForNonEnterprise(data): any {
+    removePropertiesForNonEnterprise(data): any {
         if (this.isEnterprise) {
             return data;
         }
-        const { mode, ...proData } = data;
+
```
thomasheartman, line 1392:
```suggestion
```

src/lib/features/project/project-service.ts
```diff
@@ -310,7 +320,25 @@ export default class ProjectService {
...
+                    await enableChangeRequestsForSpecifiedEnvironments(
+                        data.changeRequestEnvironments,
+                    );
+
+                data.changeRequestEnvironments =
+                    changeRequestEnvironments ?? [];
+            } else {
+                data.changeRequestEnvironments = [];
+            }
+        }
```
thomasheartman, line 341: This is the meat of the change. I've done a few things here:

1. Only call this method if we're on enterprise (and if the flag is enabled)
2. Validate the data in OSS before passing it to the function. That keeps all validation in one place and we can safely assume that all envs passed into the function are valid.
3. Assume that the function will return the list of environments and their required approvals. 
4. If we're on enterprise and you didn't provide any change request environments, then we'll return an empty list to indicate that no envs were set up.

This means that when we call this in enterprise, there's less worrying about logic and that the data is valid. Instead, it's a function that takes an input and yields an output. In addition to putting all the heavy logic in one place, it also makes it easier to test all of this in one spot.

Tymek (approved):
LGTM. I didn't see a test for trying to add a project with no environments, but I'm not sure if we need it

thomasheartman:
Right! Can you explain what the case is; I'm not sure I understand. No environments or no change request environments?

Tymek:
> Right! Can you explain what the case is; I'm not sure I understand. No environments or no change request environments?

No environments. Not related directly to this PR

thomasheartman:
Okay! Are you asking about missing `environments` or an empty list? Either way, I'm pretty sure they're covered in the e2e tests in enterprise.

PR #6934, opened 2024-04-25
feat: remove bearer token middleware flag

gastonfournier:
## About the changes
Removes the bearer token middleware flag for GA


chriswk (approved):
LGTM

PR #6935, opened 2024-04-25
feat: project owners in project service

Tymek:
## About the changes
Schema and integrating into service and controller for project owners

### DIFF
--- a/src/lib/features/project/fake-project-owners-read-model.ts
+++ b/src/lib/features/project/fake-project-owners-read-model.ts
@@ -2,27 +2,15 @@ import type { IProjectWithCount } from '../../types';
 import type {
     IProjectOwnersReadModel,
     IProjectWithCountAndOwners,
-    ProjectOwnersDictionary,
 } from './project-owners-read-model.type';
 
 export class FakeProjectOwnersReadModel implements IProjectOwnersReadModel {
-    static addOwnerData(
+    async addOwners(
         projects: IProjectWithCount[],
-        _owners: ProjectOwnersDictionary,
-    ): IProjectWithCountAndOwners[] {
+    ): Promise<IProjectWithCountAndOwners[]> {
         return projects.map((project) => ({
             ...project,
             owners: [{ ownerType: 'system' }],
         }));
-    }
-
-    async getAllProjectOwners(): Promise<ProjectOwnersDictionary> {
-        return {};
-    }
-
-    async addOwners(
-        projects: IProjectWithCount[],
-    ): Promise<IProjectWithCountAndOwners[]> {
-        return FakeProjectOwnersReadModel.addOwnerData(projects, {});
     }
 }
--- a/src/lib/features/project/fake-project-owners-read-model.ts
+++ b/src/lib/features/project/fake-project-owners-read-model.ts
@@ -1,0 +1,35 @@
+import type { IProjectWithCount } from '../../types';
+import type {
+    IProjectOwnersReadModel,
+    IProjectWithCountAndOwners,
+    ProjectOwnersDictionary,
+} from './project-owners-read-model.type';
+
+const T = {
+    ROLE_USER: 'role_user',
+    GROUP_ROLE: 'group_role',
+    ROLES: 'roles',
+    USERS: 'users',
+};
+
+export class FakeProjectOwnersReadModel implements IProjectOwnersReadModel {
+    static addOwnerData(
+        projects: IProjectWithCount[],
+        _owners: ProjectOwnersDictionary,
+    ): IProjectWithCountAndOwners[] {
+        return projects.map((project) => ({
+            ...project,
+            owners: [{ ownerType: 'system' }],
+        }));
+    }
+
+    async getAllProjectOwners(): Promise<ProjectOwnersDictionary> {
+        return {};
+    }
+
+    async addOwners(
+        projects: IProjectWithCount[],
+    ): Promise<IProjectWithCountAndOwners[]> {
+        return FakeProjectOwnersReadModel.addOwnerData(projects, {});
+    }
+}
--- a/src/lib/features/project/fake-project-owners-read-model.ts
+++ b/src/lib/features/project/fake-project-owners-read-model.ts
@@ -5,13 +5,6 @@ import type {
     ProjectOwnersDictionary,
 } from './project-owners-read-model.type';
 
-const T = {
-    ROLE_USER: 'role_user',
-    GROUP_ROLE: 'group_role',
-    ROLES: 'roles',
-    USERS: 'users',
-};
-
 export class FakeProjectOwnersReadModel implements IProjectOwnersReadModel {
     static addOwnerData(
         projects: IProjectWithCount[],
--- a/src/lib/features/project/project-controller.ts
+++ b/src/lib/features/project/project-controller.ts
@@ -197,9 +197,19 @@ export default class ProjectController extends Controller {
             user.id,
         );
 
-        // if (this.flagResolver.isEnabled('projectsListNewCards')) {
-        //   TODO: get project owners and add to response
-        // }
+        if (this.flagResolver.isEnabled('projectsListNewCards')) {
+            const projectsWithOwners =
+                await this.projectService.addOwnersToProjects(projects);
+
+            this.openApiService.respondWithValidation(
+                200,
+                res,
+                projectsSchema.$id,
+                { version: 1, projects: serializeDates(projectsWithOwners) },
+            );
+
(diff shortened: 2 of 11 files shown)


src/lib/features/project/fake-project-owners-read-model.ts
```diff
@@ -0,0 +1,35 @@
...
+    ProjectOwnersDictionary,
+} from './project-owners-read-model.type';
+
+const T = {
+    ROLE_USER: 'role_user',
+    GROUP_ROLE: 'group_role',
+    ROLES: 'roles',
+    USERS: 'users',
+};
+
```
thomasheartman, line 14: Was this left in by mistake?
```suggestion
```

src/lib/features/project/project-owners-read-model.type.ts
```diff
@@ -0,0 +1,29 @@
...
+    | Array<UserProjectOwner | GroupProjectOwner>;
+
+export type ProjectOwnersDictionary = Record<string, ProjectOwners>;
+
+export type IProjectWithCountAndOwners = IProjectWithCount & {
+    owners: ProjectOwners;
+};
+
+export interface IProjectOwnersReadModel {
+    getAllProjectOwners(): Promise<ProjectOwnersDictionary>;
```
thomasheartman, line 25: Why is this part of the interface? I don't think this should ever be called by an external module?

src/lib/features/project/fake-project-owners-read-model.ts
```diff
@@ -0,0 +1,35 @@
...
+export class FakeProjectOwnersReadModel implements IProjectOwnersReadModel {
+    static addOwnerData(
+        projects: IProjectWithCount[],
+        _owners: ProjectOwnersDictionary,
```
thomasheartman, line 18: can we just leave this out if we don't use it?
```suggestion
```

src/lib/features/project/fake-project-owners-read-model.ts
```diff
@@ -0,0 +1,35 @@
...
+        return projects.map((project) => ({
+            ...project,
+            owners: [{ ownerType: 'system' }],
+        }));
+    }
+
+    async getAllProjectOwners(): Promise<ProjectOwnersDictionary> {
+        return {};
+    }
+
```
thomasheartman, line 29: If we never use this, let's remove it?
```suggestion
```

src/lib/features/project/fake-project-owners-read-model.ts
```diff
@@ -0,0 +1,35 @@
...
+const T = {
+    ROLE_USER: 'role_user',
+    GROUP_ROLE: 'group_role',
+    ROLES: 'roles',
+    USERS: 'users',
+};
+
+export class FakeProjectOwnersReadModel implements IProjectOwnersReadModel {
+    static addOwnerData(
```
thomasheartman, line 16: I feel like we can remove this method from the fake model as well? If we're only ever doing the same thing, might as well just put that directly into addOwners?

src/lib/features/project/project-controller.ts
```diff
@@ -197,9 +197,19 @@ export default class ProjectController extends Controller {
...
+
+            this.openApiService.respondWithValidation(
+                200,
+                res,
+                projectsSchema.$id,
+                { version: 1, projects: serializeDates(projectsWithOwners) },
+            );
+
+            return;
+        }
```
thomasheartman, line 212: Would it be better to make this an if/else? For clarity? You might also not need the extra return 🤷🏼  No very strong feelings here.

src/lib/features/project/project-owners-read-model.test.ts
```diff
@@ -107,14 +124,7 @@ afterAll(async () => {
...
 afterEach(async () => {
-    if (db) {
-        const projects = await db.stores.projectStore.getAll();
-        for (const project of projects) {
-            // Clean only project roles, not all roles
-            await db.stores.roleStore.removeRolesForProject(project.id);
-        }
-        await db.stores.projectStore.deleteAll();
-    }
+    db.stores.roleStore;
```
thomasheartman, line 127: This looks weird? Did you forget to change something here?

src/lib/features/project/project-owners-read-model.ts
```diff
@@ -44,7 +28,7 @@ export class ProjectOwnersReadModel {
     ): IProjectWithCountAndOwners[] {
         return projects.map((project) => ({
             ...project,
-            owners: owners[project.name] || [{ ownerType: 'system' }],
+            owners: owners[project.id] || [{ ownerType: 'system' }],
```
thomasheartman, line 31: Good catch here. Did we really use project name? 😬

src/lib/features/project/project-service.ts
```diff
@@ -218,6 +222,12 @@ export default class ProjectService {
         return projects;
     }
 
+    async addOwnersToProjects(
+        projects: IProjectWithCount[],
+    ): Promise<IProjectWithCount[]> {
+        return this.projectOwnersReadModel.addOwners(projects);
+    }
```
thomasheartman, line 229: Why are we adding this to the service? I thought the controller was gonna call the read model directly? Maybe I've misunderstood? 🤔

src/lib/openapi/spec/project-schema.ts
```diff
@@ -89,6 +89,71 @@ export const projectSchema = {
             description:
                 'The average time from when a feature was created to when it was enabled in the "production" environment during the current window',
         },
+        owners: {
+            description: 'The owners of this project, both users and groups.',
```
thomasheartman, line 93:
```suggestion
            description: 'The users and/or groups that have the "owner" role in this project. If no such users or groups exist, the list will contain the "system" owner instead.',
```

src/lib/openapi/spec/project-schema.ts
```diff
@@ -89,6 +89,71 @@ export const projectSchema = {
...
+            oneOf: [
+                {
+                    type: 'array',
```
thomasheartman, line 96: We must always have at least one item, right?

```suggestion
                    type: 'array',
                    minItems: 1,
```

src/lib/openapi/spec/project-schema.ts
```diff
@@ -89,6 +89,71 @@ export const projectSchema = {
...
+                                type: 'object',
+                                required: ['ownerType', 'name'],
+                                properties: {
+                                    ownerType: {
+                                        type: 'string',
+                                        enum: ['group'],
+                                    },
+                                    name: {
+                                        type: 'string',
+                                        nullable: true,
```
thomasheartman, line 134: Why is this nullable? A group must have a name, right?
```suggestion
```

src/lib/openapi/spec/project-schema.ts
```diff
@@ -89,6 +89,71 @@ export const projectSchema = {
...
+                                        example: 'Group Name',
+                                    },
+                                },
+                            },
+                        ],
+                    },
+                },
+                {
+                    type: 'array',
```
thomasheartman, line 143: We can specify the length here. Or we could maybe add an enum? That'd be ideal, but I'm not sure what works.

```suggestion
                    type: 'array',
                    minItems: 1,
                    maxItems: 1,
                    enum: ['{ownerType: 'system'}]
```

src/lib/features/project/project-service.ts
Tymek: I don't think we have this pattern anywhere. Controllers use stores always with the help of services

src/lib/openapi/spec/project-schema.ts
Tymek: I'll try, but I had issues with this failing before. OpenAPI didn't handle oneOf + maxItems in a correct way

src/lib/features/project/project-owners-read-model.ts
Tymek: yup. it took me couple of minutes to debug xD

src/lib/features/project/project-service.ts
thomasheartman: The playground controller (`src/lib/features/playground/playground.ts`) calls the view model directly, so that looks like the pattern to me? Is it different in enterprise or something?

src/lib/openapi/spec/project-schema.ts
thomasheartman: Interesting. Not sure what the deal is. My preference would be to just set the type to array and provide the enum, but I'm not sure how it handles that

thomasheartman (approved):

PR #6937, opened 2024-04-25
fix: disable SCIM for OIDC, nothing to test it on

daveleek:
disable SCIM for OIDC, nothing to test it on
Also checks scim enabled before saving


chriswk (changes requested):
I think you were supposed to do this in OidcAuth, not SamlAuth ?

daveleek:
Indeed, thanks @chriswk! Fixed now

sighphyre (approved):
The change looks okay, I think there's some fixes to the flow that we'd need to do in another PR though

PR #6939, opened 2024-04-25
fix: don't apply bearer token middleware to root

chriswk:
Conflicts with scim which assumes Bearer <token>, and is located under /scim, with no /api prefix


sighphyre (approved):
LGTM - pending green tests

PR #6941, opened 2024-04-25
fix: Data usage metrics chart included per day annotion

ivarconr:
Added a small annotation to the chart on how much data is included in the pro plan on average per day to make it easier to spot when a customer average above the included threshold.

image


daveleek (approved):
Nice!

PR #6942, opened 2024-04-26
chore: remove workaround

thomasheartman:
This PR removes the workaround introduced in https://github.com/Unleash/unleash/pull/6931. After https://github.com/ivarconr/unleash-enterprise/pull/1268 has been merged, this should be safe to apply.

Notably, this PR: 
- tightens up the type for the enable change request function, so we can use that to inform the code
- skips trying to do anything with an empty array

The last point is less important than it might seem because both the env validation and the current implementation of the callback is essentially a no-op when there are no envs. However, that's hard to enforce. If we just exit out early, then at least we know nothing happens.

Optionally, we could do something like this instead, but I'm not sure it's better or worse. Happy to take input.
```ts
            const crEnvs = newProject.changeRequestEnvironments ?? []
            await this.validateEnvironmentsExist(crEnvs.map((env) => env.name));
            const changeRequestEnvironments =
                await enableChangeRequestsForSpecifiedEnvironments(crEnvs,);

            data.changeRequestEnvironments = changeRequestEnvironments;
```

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -291,9 +291,9 @@ export default class ProjectService {
         enableChangeRequestsForSpecifiedEnvironments: (
             environments: CreateProject['changeRequestEnvironments'],
         ) => Promise<
-            void | ProjectCreated['changeRequestEnvironments']
+            ProjectCreated['changeRequestEnvironments']
         > = async () => {
-            return;
+            return [];
         },
     ): Promise<ProjectCreated> {
         await this.validateProjectEnvironments(newProject.environments);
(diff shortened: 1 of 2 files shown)


src/lib/features/project/project-service.ts
```diff
@@ -324,17 +324,19 @@ export default class ProjectService {
...
+                );
+                const changeRequestEnvironments =
+                    await enableChangeRequestsForSpecifiedEnvironments(
+                        newProject.changeRequestEnvironments,
+                    );
+
+                data.changeRequestEnvironments = changeRequestEnvironments;
+            } else {
+                data.changeRequestEnvironments = [];
+            }
```
thomasheartman, line 339: As mentioned in the description, this could be this instead (essentially the same as before, but without the comment and without extra set of `??`s. Happy to hear your thoughts. 

```suggestion
            const crEnvs = newProject.changeRequestEnvironments ?? []
            await this.validateEnvironmentsExist(crEnvs.map((env) => env.name));
            const changeRequestEnvironments =
                await enableChangeRequestsForSpecifiedEnvironments(crEnvs,);

            data.changeRequestEnvironments = changeRequestEnvironments;
```

I kinda like this version, because it avoids all the nesting and it reads cleaner to me 🤷🏼

Tymek: More verbose version looks more readable

Tymek (approved):

PR #6944, opened 2024-04-26
feat: revived feature goes to initial lifecycle stage

kwasniew:
## About the changes

Revived feature goes back to the initial stage. This is a new initial stage with a date matching the revive time.
Screenshot 2024-04-26 at 09 41 34

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
@@ -3,6 +3,7 @@ import {
     FEATURE_ARCHIVED,
     FEATURE_COMPLETED,
     FEATURE_CREATED,
+    FEATURE_REVIVED,
     type IEnvironmentStore,
     type IEventStore,
     type IFlagResolver,
@@ -93,6 +94,11 @@ export class FeatureLifecycleService extends EventEmitter {
                 this.featureArchived(event.featureName),
             );
         });
+        this.eventStore.on(FEATURE_REVIVED, async (event) => {
+            await this.checkEnabled(() =>
+                this.featureRevived(event.featureName),
+            );
+        });
     }
 
     async getFeatureLifecycle(feature: string): Promise<FeatureLifecycleView> {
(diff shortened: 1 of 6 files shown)


src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -155,4 +161,9 @@ export class FeatureLifecycleService extends EventEmitter {
         ]);
         this.emit(STAGE_ENTERED, { stage: 'archived' });
     }
+
+    private async featureRevived(feature: string) {
+        await this.featureLifecycleStore.delete(feature);
```
kwasniew, line 166: delete all previous stages

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -155,4 +161,9 @@ export class FeatureLifecycleService extends EventEmitter {
...
+        await this.featureInitialized(feature);
```
kwasniew, line 167: re-initialize with a new date

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
sjaanus: LG, but I was thinking about it. If we want to measure how long we were in each stage, this makes it not countable, the first stage. Unless we take it from feature created date.
But if we use created date, then after revive, the created date could have been 2 years ago.

sjaanus (approved):
LG, but some thoughts added.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
kwasniew: We need an opinion here. One option would be to delete all events expect from the initial event. So when you revive you only erase later stages.

kwasniew: Since I don't know which is better I'll merge it now with an easy option to change it later

PR #6946, opened 2024-04-26
feat: pre-live is non production not just dev

kwasniew:
## About the changes

Pre-live is not just development but also anything else that is not production.

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
@@ -134,7 +134,7 @@ export class FeatureLifecycleService extends EventEmitter {
             }
             if (env.type === 'production') {
                 await this.stageReceivedMetrics(features, 'live');
-            } else if (env.type === 'development') {
+            } else if (env.type !== 'production') {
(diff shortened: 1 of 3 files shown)


src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -134,7 +134,7 @@ export class FeatureLifecycleService extends EventEmitter {
             }
             if (env.type === 'production') {
                 await this.stageReceivedMetrics(features, 'live');
-            } else if (env.type === 'development') {
+            } else if (env.type !== 'production') {
```
sjaanus, line 137: This is weird, this is just else clause.

Also I think pre-live is also production, but toggle turned off.

So I would do

```
await this.stageReceivedMetrics(features, 'pre-live');
if (env.type === 'production') {
    await this.stageReceivedMetrics(features, 'live');
} 
```

sjaanus: This means, any metrics will put it to pre-live, any metrics in production puts it in production.

kwasniew: nice idea

PR #6947, opened 2024-04-26
feat: feature completed connected to backend

sjaanus:
Connects feature complete to backend


kwasniew (approved):
We can also add a test for the button click but if we plan to add more steps (completion status selection) I'd defer it

PR #6948, opened 2024-04-26
fix: fix UI issues with SCIM new token button

daveleek:
Makes the regenerate token button in SCIM settings behave like the rest of the grid (remove margin left, add margin bottom)

image


chriswk (approved):
:hammer: :hourglass:

PR #6960, opened 2024-04-29
rough layout of the new form

thomasheartman:
This PR adds a very rough first implementation of the look of the new project form. *It is not final and does not work yet*. 

The important part here is that the layout is roughly right (we'll adjust spacing etc later) and that we've got all the basic elements present.

I'll hook it up to actually work in an upcoming PR. 

Note: I know the use of css nesting can be contentious. Please consider this code to be more of a proof of concept. Stubbing things out in CSS is faster than creating styled components everywhere. Happy to refactor it later, but I think the important part now is to get it in.

image

The missing icon, text alignment, etc, will also be solved later.


frontend/src/component/project/Project/CreateProject/NewProjectForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+    background: theme.palette.background.default,
+
+    '> * + *': {
+        borderBlockStart: `1px solid ${theme.palette.divider}`,
+    },
+
+    '> *': {
+        padding: theme.spacing(7),
+    },
+}));
```
Tymek, line 24: In JSX I prefer local styles over ones defined by ancestor. If I expect that in React "UI is the function of state" it's harder to debug when same component can have a different look depending on where it's placed. When there are styled applied to each item it's easier to have 2 components:

```
<StyledContainer>
  <StyledSection>...</StyledSection>
  <StyledSection>...</StyledSection>
</StyledContainer>
```
In current case, we can use styled `section` or `fieldgroup`

Tymek (approved):

frontend/src/component/project/Project/CreateProject/NewProjectForm.tsx
thomasheartman: I've fixed these now.  I stuck with `div`s for the form sections because:
1. [According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section), 'Sections should always have a heading, with very few exceptions.' These don't have headings and it doesn't really make sense for them to have them. The grouping is mostly visual.
2. There is (or was?) a longstanding bug in Chrome, where `fieldset` elements couldn't use `display: flex`. Don't know if this has been fixed yet. But also, I'm not sure they're all fieldsets. The template probably wouldn't be part of the same set as the name and description, for instance, and I don't think (though I may be wrong) that form action buttons are usually in a fieldset. 

So without a better choice (though I'd love to hear suggestions if you have any!) I went with the generic block-level element: the `div`.

PR #6961, opened 2024-04-29
feat: enable project creation through new form

thomasheartman:
This PR allows very simple project creation. You can add a name and description. The ID is generated for you at the moment (we'll fix this later). Nothing else works, but the project is created successfully.


frontend/src/component/project/Project/CreateProject/NewProjectForm.tsx
```diff
@@ -75,15 +72,66 @@ const FormActions = styled(StyledFormSection)(({ theme }) => ({
...
+    setProjectName: React.Dispatch<React.SetStateAction<string>>;
+    setProjectDesc: React.Dispatch<React.SetStateAction<string>>;
+    setFeatureLimit?: React.Dispatch<React.SetStateAction<string>>;
+    setProjectMode?: React.Dispatch<React.SetStateAction<ProjectMode>>;
+    handleSubmit: (e: any) => void;
+    errors: { [key: string]: string };
+    mode: 'Create' | 'Edit';
+    clearErrors: () => void;
+    validateProjectId: () => void;
+};
```
thomasheartman, line 94: These are all the props from the old form. I've put them here as a template. We'll use the ones we need and remove the ones we don't as we go.

frontend/src/component/project/Project/CreateProject/CreateProject.tsx
thomasheartman, line 1: Just shove in all the props that we send to the existing form for now, so that we have all the stuff we need.

Tymek (approved):

PR #6962, opened 2024-04-29
refactor: badge icon spacing

Tymek:
Simplified after previous modifications

### DIFF
--- a/frontend/src/component/common/Badge/Badge.tsx
+++ b/frontend/src/component/common/Badge/Badge.tsx
@@ -40,6 +40,7 @@ const StyledBadge = styled('span')<IBadgeProps>(
     ({ theme, color = 'neutral', icon }) => ({
         display: 'inline-flex',
         alignItems: 'center',
+        gap: theme.spacing(0.5),
         padding: theme.spacing(icon ? 0.375 : 0.625, 1),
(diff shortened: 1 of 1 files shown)


frontend/src/component/common/Badge/Badge.tsx
```diff
@@ -121,12 +110,15 @@ export const Badge: FC<IBadgeProps> = forwardRef(
         >
             <ConditionallyRender
                 condition={Boolean(icon) && !iconRight}
-                show={BadgeIcon(color, icon!, false, Boolean(children))}
+                show={BadgeIcon(color, icon!)}
+            />
+            <ConditionallyRender
+                condition={Boolean(children)}
+                show={<div>{children}</div>}
```
thomasheartman, line 117: Why are we putting the children within a div here? Couldn't this whole component be replaced with `{children}`? Is it in case we have several children and we need to group them?

thomasheartman (approved):

frontend/src/component/common/Badge/Badge.tsx
Tymek: Because `children` can be <Fragment/> containing multiple elements, and without wrapping it we can run into issues where flexbox adds spacing between every `child` item

PR #6964, opened 2024-04-29
feat: do not insert into database stages that already exist

sjaanus:
Previously when we had thousands of metrics coming in, we were trying to write them all to database and running into on conflict


andreas-unleash (approved):
LGTM

PR #6967, opened 2024-04-30
wip: select envs in new project creation form

thomasheartman:


frontend/src/component/project/Project/CreateProject/SelectionButton.styles.tsx
thomasheartman, line 1: This is a lift and shift of the `FilterItem.styles.tsx` file that belongs to the filters. 

We might need to go and clean this up a bit, but I'll wait until we have all the dropdowns sorted out.

frontend/src/component/project/Project/CreateProject/SelectionButton.tsx
thomasheartman, line 1: Similar to the style file, this is mostly a copy paste of the "FilterItem" component, but with some changes to better work for this case (like how we don't need operators etc). Like the styles, I'll update this as I go along.

PR #6968, opened 2024-04-30
feat: add completed event as webhook event

sjaanus:
Now feature completed event can be used with webhook.


alvinometric:
Hello! 
Do we have docs elsewhere for what `completed` means and how to trigger it?

andreas-unleash (approved):

sjaanus:
> Hello! Do we have docs elsewhere for what `completed` means and how to trigger it?

@alvinometric this is a beta feature we are working on. By rule, we usually create docs before it goes GA.

PR #6971, opened 2024-05-02
docs: update feature availability

nunogois:
Updates feature availability. 

Banners is a feature that is already GA'd since Unleash 5.7, so beta availability seems irrelevant now.

We removed the mention of planning to GA Signals & Actions in 5.12, since that's not happening and we would prefer not to include these mentions in the future.


chriswk (approved):
LGTM

PR #6973, opened 2024-05-02
feat: duration in stage, add feature lifecycle prometheus metrics

sjaanus:
Introduce a new concept. Duration in stage.
Also add it into prometheus metric.


kwasniew (approved):
Let's add a flag in a separate PR

PR #6979, opened 2024-05-06
feat: configure CRs when creating projects

thomasheartman:
This PR adds the ability to configure CRs when creating a project.

The design is unfinished, and the code certainly needs cleanup, but I'd like to get this into sandbox so we can look at it.

Things that still need to be done:

1. What do we do about this button when the user has no environments selected? As a rough draft, I've disabled it. However, we should make it possible to navigate to and give you an explanation why it was disabled, e.g. "You have no project environments selected. Please select at least one project environment.".
2. The form design is not done: the width should be constant and not jumpy the way it is now. Also, the search field is too wide.
3. I've made the desicion that if you deselect a project env, we also remove that env from your CR config it it's in there.
4. Potential improvement: if you enable and then disable CRs for an env, we *could* probably store the data in between, so that if you set required approvers 5 and then disabled it, it'd still be 5 when you re-enabled it. That sounds like a good user experience. We should also be able to extend that to adding/removing environments from project envs.

### DIFF
--- a/frontend/src/component/project/Project/CreateProject/SelectionButton.tsx
+++ b/frontend/src/component/project/Project/CreateProject/SelectionButton.tsx
@@ -288,14 +288,17 @@ export const TableSelect: FC<TableSelectProps> = ({
         Record<
             string,
             {
-                enabled: boolean;
+                changeRequestEnabled: boolean;
                 requiredApprovals: number;
             }
         >
     >(
         Object.fromEntries(
             Object.entries(projectChangeRequestConfiguration).map(
-                ([name, config]) => [name, { ...config, enabled: true }],
+                ([name, config]) => [
+                    name,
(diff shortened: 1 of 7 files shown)


frontend/src/component/project/Project/CreateProject/SelectionButton.tsx
```diff
@@ -262,3 +263,151 @@ type SingleSelectListProps = Pick<
...
+    >;
+};
+export const TableSelect: FC<TableSelectProps> = ({
+    button,
+    search,
+    onChange,
+    projectChangeRequestConfiguration,
+    updateProjectChangeRequestConfiguration,
+    activeEnvironments,
+}) => {
```
thomasheartman, line 291: This is mostly a duplicate of the general component, so we should probably combine them. For later, though.

PR #6983, opened 2024-05-06
feat: create initial stages for features

sjaanus:
We are getting questions from engineers, why I do not see lifecycle. The same will happen with our customers. Now customers will see lifecycle component unified across features.

### DIFF
--- a/src/migrations/20240506141345-lifecycle-initial-stage.js
+++ b/src/migrations/20240506141345-lifecycle-initial-stage.js
@@ -1,0 +1,22 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+            INSERT INTO feature_lifecycles (feature, stage, created_at)
+            SELECT name, 'initial', NOW() AS created_at
+            FROM features
+            WHERE name NOT IN (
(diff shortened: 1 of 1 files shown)


src/migrations/20240506141345-lifecycle-initial-stage.js
```diff
@@ -0,0 +1,19 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+            INSERT INTO feature_lifecycles (feature, stage, created_at)
+            SELECT name, 'initial', features.created_at
```
kwasniew, line 5: can we table namespace the name as we did created_at?

src/migrations/20240506141345-lifecycle-initial-stage.js
```diff
@@ -0,0 +1,19 @@
...
+            FROM features
+                     LEFT JOIN feature_lifecycles ON features.name = feature_lifecycles.feature
```
kwasniew, line 7: too many spaces before left join

kwasniew (approved):

PR #6985, opened 2024-05-06
feat: Allow to use CA certificate file path for DB

egor-xyz:
Regarding ticket #6892:

I would like to enable the use of a CA certificate without requiring other certificates. This would be useful for AWS Helm, as AWS only provides a single PEM file for DB connections.

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -183,11 +183,14 @@ const dateHandlingCallback = (connection, callback) => {
     });
 };
 
-const addSSLOption = (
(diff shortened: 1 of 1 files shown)


egor-xyz:
FYI @chriswk @sighphyre

chriswk (approved):
Nice. I like it. @ivarconr @sighphyre can you see anything wrong with applying Egor's patch? This allows adding only a CA, which I think is a good way to go.

Thank you @egor-xyz

Issue #6988, opened 2024-05-06
Feature Search API misbehaves when features have no metrics within the last hour


LukasHeimann:
### Describe the bug

https://github.com/Unleash/unleash/pull/6915 slightly reworked how metrics are joined to the feature search api. Crucially, as the final result is a `select *`, this means that all columns of `metrics` are now exposed next to all columns of `ranked_features`, instead of just `yes` and `no` counts as before. Crucially, there is a duplication of the columns `environment` and `feature_name`.

As long as there are metrics for the toggle (in the last hour, as per the filter), this is no problem, as this is the join criteria after all. However, if there are no metrics, these fields are `NULL` when coming from the `metrics` subselect.

![image](https://github.com/Unleash/unleash/assets/48723560/07dd9f24-da67-4587-8bff-3863014417ea)
_(reproduced the query in dbeaver and executed against my db, only slightly rearranged the columns in the final output)_

This later on messes up result aggregation, and all toggles are cobbled up under a `null`-named feature. This example shows the UI:
![image](https://github.com/Unleash/unleash/assets/48723560/44fd7cc2-bbf6-49a3-935b-02fcaf92bb0e)

As the feature name is `null`, trying to activate or deactivate this entry will result in errors.

### Steps to reproduce the bug

1. Create feature toggles
2. Don't have metrics for them (in the last hour)
3. Execute the search api and see that the results are messed up

### Expected behavior

The list of feature toggles for my project should be complete, even if I don't have any metrics.

I guess a final fix would be to replace the `select *` with a list of select columns that should be used, so it can be controlled that `feature_name` and `environment` are taken from `ranked_features` and not `metrics`. A slightly worse, but simpler alternative might be to rename the columns in the `metrics` subselect to something else, so that there is no name-clash.

### Unleash version

5.12.0

sveisvei:
I am seeing the same thing after upgrading to `5.12.0`

ivarconr:
Thanks @LukasHeimann! 
We have addressed the issue with a fix in `Unleash 5.12.1` and `Unleash Enterprise 5.12.2`.

LukasHeimann:
Thank you for the fix, the problem is gone

kwasniew:
@LukasHeimann massive thank you for the detailed bug report.

PR #6989, opened 2024-05-06
fix: duplicate column name in search query

kwasniew:
## About the changes

Avoid duplicate column names in search query

Fixes: https://github.com/Unleash/unleash/issues/6988


ivarconr (approved):

sighphyre (approved):
Okay, wow, hard to reproduce with an automated test but we've both confirmed this works with local testing and the tests pass so let's do it!

PR #6992, opened 2024-05-06
fix: Add appropriate response headers to SPA entry point HTML response

zapplebee:
## About the changes

Current state, when returning the HTML entry point from the server, there are no headers attached. We encountered an issue with a deployment and this had an impact for us.

A brief description:

1. We deployed the most recent version. Noticed an unrelated issue.
2. Users tried to use the most recent version and due to their client cache, requested assets that did not exist in the newest version.
3. Our cache layer cached the assets that were not there with the HTML response. It had to infer the type based on the filename because there was no attached `Content-Type` header. This cache was very sticky.
4. After rolling back we saw the HTML response (from the cache) instead of the appropriate response from the upstream Unleash application.

This PR does a few things.

1. When responding with the HTML entry point, it adds header (`Content-Type: text/html`). 
2. When the client is requesting an asset (a path that ends with an extension), it also instructs the resource not to be cached (`Cache-Control: no-cache`) and returns a 404. This will prevent misses from getting cached.

Closes #

No open issue.

## Discussion points

To me, there doesn't seem to be a lot of test infra on serving the SPA application. If that is an error, please indicate where that is and an appropriate test can be added.

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -216,6 +216,13 @@ export default async function getApp(
 
     app.get(`${baseUriPath}/*`, (req, res) => {
         res.set('Content-Type', 'text/html');
+        const requestPath = path.parse(req.url);
+        // appropriately return 404 requests for assets with an extension (js, css, etc)
+        if (requestPath.ext !== '' && requestPath.ext !== 'html') {
+            res.set('Cache-Control', 'no-cache');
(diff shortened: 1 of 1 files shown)


chriswk (approved):
Thank you for adding this. I'm going to merge this, and it will be part of our next release.

PR #7001, opened 2024-05-08
chore: add automatic ID generation algorithm

thomasheartman:
This PR adds a function to automatically generate a project ID on creation. Using this when the id is missing will be handled in following PRs. 

The function uses the existing `slug` package to create a slug, and then takes the 12 characters of a uuidv4 string to generate an ID.

The included tests check that the 12 character hash is added and that the resulting string is url friendly (by checking that `encodeURIComponent` doesn't change it).

We could also test a lot of edge cases (such as dealing with double spaces, trimming the string, etc), but I think that's better handled by the library itself (but you can check out what I removed in https://github.com/Unleash/unleash/pull/7001/commits/2d9bcb6390eb2e2712535591ca84f4e583e481f0 for an idea).


andreas-unleash (approved):
Nice!

PR #7002, opened 2024-05-08
fix: optimize table placeholder loading

kwasniew:
## About the changes

Performance issues before: we were loading placeholder data (page size * number of environments) with real switch toggles that are doing full permission checks and contain many DOM elements to render. The placeholder data was hidden behind a loading indicator. As a result when clicking on the sorting column there was a significant delay.

After this optimization we're rendering simple divs hidden behind a loading indicator.

https://github.com/Unleash/unleash/assets/1394682/ee42e22a-efe9-4e96-bef5-f302968cf76a

### DIFF
--- a/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/FeatureToggleCell/FeatureToggleCell.tsx
+++ b/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/FeatureToggleCell/FeatureToggleCell.tsx
@@ -84,3 +84,9 @@ const FeatureToggleCellComponent = ({
 };
 
 export const FeatureToggleCell = React.memo(FeatureToggleCellComponent);
+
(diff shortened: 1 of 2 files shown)


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/FeatureToggleCell/FeatureToggleCell.tsx
```diff
@@ -84,3 +84,9 @@ const FeatureToggleCellComponent = ({
 };
 
 export const FeatureToggleCell = React.memo(FeatureToggleCellComponent);
+
+export const PlaceholderFeatureToggleCell = () => (
+    <StyledSwitchContainer>
+        <div data-loading>toggle</div>
```
kwasniew, line 90: the text is selected to match the width of the switch

sjaanus (approved):

PR #7014, opened 2024-05-08
feat: add status fields for feature lifecycle table

sjaanus:


src/lib/features/feature-lifecycle/feature-lifecycle-store.ts
```diff
@@ -10,6 +10,8 @@ import type { StageName } from '../../types';
 type DBType = {
     stage: StageName;
     created_at: string;
+    status?: string;
```
kwasniew, line 13: normally we have a rule of thumb that we don't add anything in migration prs but I think the optional one should be fine.

kwasniew (approved):

PR #7016, opened 2024-05-08
fix: only send cr flag config when the flag is turned on

thomasheartman:
I realized that, in an oversight, the form now includes CR config, even if the new form isn't active. The API should just ignore it if it doesn't understand it, so it's not very harmful, but it's better if we don't send it at all. This PR does that.


thomasheartman:
Superseded by #7102

PR #7017, opened 2024-05-08
feat: expose lifecycle stage in project overview search

kwasniew:
## About the changes

Each feature in project overview search will have a lifecycle with current stage and entered date similarly to single feature query.

I'm adding this new capability behind a flag so that we can safely roll it out.

### DIFF
--- a/src/lib/features/feature-search/feature-search-store.ts
+++ b/src/lib/features/feature-search/feature-search-store.ts
@@ -7,6 +7,7 @@ import type {
     IFeatureOverview,
     IFeatureSearchOverview,
     IFeatureSearchStore,
+    IFlagResolver,
     ITag,
 } from '../../types';
 import FeatureToggleStore from '../feature-toggle/feature-toggle-store';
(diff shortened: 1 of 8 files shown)


src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -224,6 +253,14 @@ class FeatureSearchStore implements IFeatureSearchStore {
                     );
                 });
 
+                if (featureLifecycleEnabled) {
+                    query.leftJoin(
```
sjaanus, line 257: I would not join it into original query, but rather into final query, where the paging has been done already.

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -86,6 +95,19 @@ class FeatureSearchStore implements IFeatureSearchStore {
         const validatedSortOrder =
             sortOrder === 'asc' || sortOrder === 'desc' ? sortOrder : 'asc';
 
+        const featureLifecycleEnabled =
+            this.flagResolver.isEnabled('featureLifecycle');
+        const latestLifecycleStageQuery = this.db
```
sjaanus, line 100: Also can we look into opportunities to pull it out as a method, and put all selects, joins and query there

sjaanus (approved):
LG!

PR #7021, opened 2024-05-09
feat: feature lifecycle completed schema

sjaanus:
1. Added new schema and tests
2. Controller also accepts the data
3. Also sending fake data from frontend currently

Next steps, implement service/store layer and frontend

### DIFF
--- a/src/lib/openapi/spec/feature-lifecycle-completed-schema.ts
+++ b/src/lib/openapi/spec/feature-lifecycle-completed-schema.ts
@@ -1,0 +1,30 @@
+import type { FromSchema } from 'json-schema-to-ts';
+
+export const featureLifecycleCompletedSchema = {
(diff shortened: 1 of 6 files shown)


src/lib/openapi/spec/feature-lifecycle-completed-schema.ts
```diff
@@ -0,0 +1,30 @@
...
+    required: ['status'],
+    properties: {
+        status: {
+            type: 'string',
+            enum: ['kept', 'discarded'],
+            example: 'kept',
+            description:
+                'The status of the feature after it has been marked as completed',
+        },
+        statusValue: {
```
kwasniew, line 17: do we want to allow to send discarded with statusValue? In frontend type modelling I made this scenario impossible. But maybe openapi can be more forgiving.

kwasniew (approved):

PR #7022, opened 2024-05-09
feat: add completed status backend

sjaanus:
1. Implemented controller, service, store for status saving


src/lib/features/feature-lifecycle/feature-lifecycle-store.ts
```diff
@@ -29,28 +29,31 @@ export class FeatureLifecycleStore implements IFeatureLifecycleStore {
     async insert(
         featureLifecycleStages: FeatureLifecycleStage[],
     ): Promise<void> {
-        const joinedLifecycleStages = featureLifecycleStages
-            .map((stage) => `('${stage.feature}', '${stage.stage}')`)
-            .join(', ');
+        const existingFeatures = await this.db('features')
```
sjaanus, line 32: Brought back the old insert function @kwasniew created, because the previous one has SQL injection possible vulnerability and it did not handle undefined values well

kwasniew (approved):

PR #7024, opened 2024-05-09
feat: Lifecycle in project overview

kwasniew:
## About the changes

Displaying lifecycle next to last seen
Screenshot 2024-05-09 at 12 48 26

Lifecycle also supports transitions in this view
Screenshot 2024-05-09 at 12 48 31

Details:
* fixing search openapi + orval to make environment type required (it is always returned)
* flag controls whether to show old last seen or add a new lifecycle in the same table cell


src/lib/openapi/spec/feature-search-environment-schema.ts
```diff
@@ -10,7 +10,7 @@ export const featureSearchEnvironmentSchema = {
     $id: '#/components/schemas/featureSearchEnvironmentSchema',
     type: 'object',
     additionalProperties: false,
-    required: ['name', 'enabled'],
+    required: ['name', 'enabled', 'type'],
```
kwasniew, line 13: making env type required in search

frontend/src/interfaces/featureToggle.ts
```diff
@@ -32,6 +32,11 @@ export type ILastSeenEnvironments = Pick<
     'name' | 'enabled' | 'lastSeenAt' | 'yes' | 'no'
 >;
 
+export type Lifecycle = {
```
kwasniew, line 35: used in another place so extract type refactoring

sjaanus (approved):

PR #7026, opened 2024-05-09
fix: Adjust meta data icons

kwasniew:
## About the changes

Aligning icons
Screenshot 2024-05-10 at 11 38 18
Screenshot 2024-05-10 at 11 38 03


sjaanus:
Created seems to have too much space around.

sjaanus:
Also Project, lifecycle gap is smaller than  Dependency, dependency value. Perhaps can be unified.

sjaanus (approved):

PR #7032, opened 2024-05-10
feat: lifecycle prometheus metrics per project

sjaanus:
When we pushed metrics per feature, it had too many datapoints and grafana could not handle it. Now I am taking median for a project.

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
@@ -237,7 +237,6 @@ export class FeatureLifecycleService extends EventEmitter {
                 acc[key] = [];
             }
 
-            // Find the next different stage for the same feature
             const nextItem = array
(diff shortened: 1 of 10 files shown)


src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -224,41 +225,52 @@ export class FeatureLifecycleService extends EventEmitter {
...
+                stage: stage as StageName,
+                duration,
             });
         });
-        return times;
+
+        return medians;
+    }
+
+    private calculateMedian(numbers: number[]): number {
```
kwasniew, line 267: I'd move it to median.ts or math.ts in utils and test it separately as a pure function not attached to a class as a private method

sjaanus: Done

kwasniew (approved):

PR #7041, opened 2024-05-13
feat: expose postgres version

chriswk:
Adds a postgres_version gauge to allow us to see postgres_version in prometheus and to post it upstream when version checking. Depends on https://github.com/bricks-software/version-function/pull/20 to be merged first to ensure our version-function doesn't crash when given the postgres-version data.


gardleopard (approved):
Great stuff :)

PR #7043, opened 2024-05-13
feat: webhook data for completed

sjaanus:
Now we post status and statusValue with webhook.


src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -183,6 +183,7 @@ export class FeatureLifecycleService extends EventEmitter {
         await this.eventService.storeEvent(
             new FeatureCompletedEvent({
                 featureName: feature,
+                data: status,
```
kwasniew, line 186: is it string or object? if string I'd allow data to be an object to make it more extensible

PR #7045, opened 2024-05-13
feat: kept and discarded read model

kwasniew:
## About the changes

* Adding status information to stages
* Modifying single feature read model and search query
* Modelled field as optional but when the field is present it's not null 
* Making lifecycle UI display when lifecycle is present (previously we had as short window where lifecycle had no value after new feature creation)

Screenshot 2024-05-13 at 12 44 46
Screenshot 2024-05-13 at 12 44 52

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/populateCurrentStage.ts
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/populateCurrentStage.ts
@@ -43,7 +43,10 @@ export const populateCurrentStage = (
         case 'completed':
             return {
                 name: 'completed',
-                status: 'kept',
+                status:
+                    feature.lifecycle.status === 'discarded'
+                        ? 'discarded'
+                        : 'kept',
                 environments: getFilteredEnvironments(() => true),
                 enteredStageAt,
             };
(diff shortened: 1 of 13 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/populateCurrentStage.ts
```diff
@@ -43,7 +43,10 @@ export const populateCurrentStage = (
         case 'completed':
             return {
                 name: 'completed',
-                status: 'kept',
+                status:
+                    feature.lifecycle.status === 'discarded'
```
kwasniew, line 47: special case for discarded, other states use green icon

src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
```diff
@@ -62,6 +66,11 @@ const getFeatureLifecycle = async (featureName: string, expectedCode = 200) => {
         .get(`/api/admin/projects/default/features/${featureName}/lifecycle`)
         .expect(expectedCode);
 };
+
+const getCurrentStage = async (featureName: string) => {
```
kwasniew, line 70: added new method to test the read model directly inside this module

sjaanus (approved):

PR #7052, opened 2024-05-14
fix: omit yes no from stale data comparison

kwasniew:
## About the changes

When computing stale data e.g. on strategy edit we need to omit yes/no counts same way we do it for lastSeenAt.

Expanded deepOmit function to support multiple keys.


FredrikOseberg (approved):
LGTM

PR #7054, opened 2024-05-14
feat: hide feature environment variants for new users and those who dont need it

kwasniew:
## About the changes

Variants tab will be hidden for new instances and those who have never used it before. You can still create feature env variants in the API and UI will respect it and show the UI. This is to encourage the usage of strategy variants.

Variants in this tab will be gone for some users
Screenshot 2024-05-14 at 15 29 26

Backend:
* backend returns `hideFeatureEnvironmentVariants` UI flag based on the DB check if the instance is not using feature environment variants
* feature environment variant check usage is optimized with the `SELECT EXISTS` and extra memoization every 10 minutes (similar solution we already for `displayUpgradeEdgeBanner`)
* `hideFeatureEnvironmentVariants` is added to `config.ts` similar to other instance configs
* I'm injecting `featureEnvironmentStore` to `config.ts` so that we can ask for the existence of the feature environment variants

Frontend:
* In the UI we're only hiding the tab with feature environment variants based on the backend flag. We may consider another flag to control this manually but for now I'd like to minimize the number of flags

Docs:
* I'm adding deprecation notice in the availability section and in the title. Maybe we need to make it even more salient?

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureView.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureView.tsx
@@ -48,7 +48,7 @@ import { ReactComponent as ParentLinkIcon } from 'assets/icons/link-parent.svg';
 import { ChildrenTooltip } from './FeatureOverview/FeatureOverviewMetaData/ChildrenTooltip';
 import copy from 'copy-to-clipboard';
 import useToast from 'hooks/useToast';
-import { useUiFlag } from '../../../hooks/useUiFlag';
+import { useUiFlag } from 'hooks/useUiFlag';
 
 const StyledHeader = styled('div')(({ theme }) => ({
     backgroundColor: theme.palette.background.paper,
--- a/frontend/src/component/feature/FeatureView/FeatureView.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureView.tsx
@@ -48,6 +48,7 @@ import { ReactComponent as ParentLinkIcon } from 'assets/icons/link-parent.svg';
 import { ChildrenTooltip } from './FeatureOverview/FeatureOverviewMetaData/ChildrenTooltip';
 import copy from 'copy-to-clipboard';
 import useToast from 'hooks/useToast';
+import { useUiFlag } from '../../../hooks/useUiFlag';
 
 const StyledHeader = styled('div')(({ theme }) => ({
     backgroundColor: theme.palette.background.paper,
@@ -145,6 +146,9 @@ export const FeatureView = () => {
     const [openStaleDialog, setOpenStaleDialog] = useState(false);
     const [isFeatureNameCopied, setIsFeatureNameCopied] = useState(false);
     const smallScreen = useMediaQuery(`(max-width:${500}px)`);
+    const hideFeatureEnvironmentVariants = useUiFlag(
+        'hideFeatureEnvironmentVariants',
+    );
 
     const { feature, loading, error, status } = useFeature(
         projectId,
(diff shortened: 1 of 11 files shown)


src/test/e2e/api/admin/config.e2e.test.ts
```diff
@@ -98,3 +98,12 @@ test('sets ui config with frontendSettings', async () => {
             expect(res.body.frontendApiOrigins).toEqual(frontendApiOrigins),
         );
 });
+
+test('ui config should have displayFeatureEnvironmentVariants flag disabled if no env variants are used', async () => {
```
kwasniew, line 102: e2e test to test real sql

src/lib/routes/admin-api/config.ts
```diff
@@ -86,6 +88,14 @@ class ConfigController extends Controller {
                 maxAge: minutesToMilliseconds(10),
             },
         );
+        this.usesFeatureEnvironmentVariants = memoizee(
```
kwasniew, line 91: following the same pattern as the code above with memoization.

src/lib/routes/admin-api/config.test.ts
```diff
@@ -98,3 +98,39 @@ describe('displayUpgradeEdgeBanner', () => {
         expect(body.flags.displayUpgradeEdgeBanner).toEqual(false);
     });
 });
+
+describe('displayFeatureEnvironmentVariants', () => {
```
kwasniew, line 102: in-memory tests that verify the logic

frontend/src/component/feature/FeatureView/FeatureView.tsx
```diff
@@ -145,6 +146,9 @@ export const FeatureView = () => {
     const [openStaleDialog, setOpenStaleDialog] = useState(false);
     const [isFeatureNameCopied, setIsFeatureNameCopied] = useState(false);
     const smallScreen = useMediaQuery(`(max-width:${500}px)`);
+    const hideFeatureEnvironmentVariants = useUiFlag(
```
kwasniew, line 149: another option is to to a flag called show/display

src/lib/routes/admin-api/config.test.ts
```diff
@@ -98,3 +98,39 @@ describe('displayUpgradeEdgeBanner', () => {
         expect(body.flags.displayUpgradeEdgeBanner).toEqual(false);
     });
 });
+
+describe('hideFeatureEnvironmentVariants', () => {
```
kwasniew, line 102: fast in memory tests that test logic and memoization

src/test/e2e/api/admin/config.e2e.test.ts
```diff
@@ -98,3 +98,12 @@ test('sets ui config with frontendSettings', async () => {
             expect(res.body.frontendApiOrigins).toEqual(frontendApiOrigins),
         );
 });
+
+test('ui config should have hideFeatureEnvironmentVariants flag enabled if no env variants are used', async () => {
```
kwasniew, line 102: one slower test that verifies DB and integration of prod components

website/docs/reference/feature-toggle-variants.md
```diff
@@ -1,13 +1,18 @@
...
-title: Feature Toggle Variants
+title: Feature Toggle Variants (deprecated)
 ---
 :::info Availability
 
 **Feature toggle variants** were first introduced in Unleash 3.2.
+
 In Unleash 4.21, variants were updated to be **environment-dependent**, meaning the same feature could have different variant configurations in different environments.
 
+Unleash 6.0 deprecates environment-dependent variants in favor of the [strategy variants](./strategy-variants.md). New Unleash instances and those who never used environment variants will
```
sjaanus, line 10: I would pull this higher and put it into a warning-like box or something. So that is first thing that people see. Basically we want all new users to know that this is not worth reading, you should not use it.

frontend/src/component/feature/FeatureView/FeatureView.tsx
ivarconr: I think the name should be driven by what our desired value should be if we are unable to evaluate it from the feature flag system (I'm case it's down or the flag is accidentally removed). 

In the scenario of sunseting a feature the default action should be to not display the feature. With that rationale the active action would be to enable the feature, which should only be possible if the flag is correctly set. I would cast my vote for "enableLegacyFeatureVariant" to make it clear that you actively turn on a deprecated feature. 

(You could argue that hideX with a default value of true serves the same purpose and does look somewhat like a reversed kill switch. Personally it feels a bit twisted in my head)

ivarconr (approved):

PR #7055, opened 2024-05-14
feat: move SCIM config into separate tab

daveleek:
## About the changes

Moves SCIM config back into separate tab

image


daveleek:
We need to think about good texts/descriptions for this

daveleek:
> We need to think about good texts/descriptions for this

I have a suggestion added now

sighphyre (approved):
Sure, LGTM

PR #7057, opened 2024-05-15
feat: backfill current stage on startup

kwasniew:
## About the changes

When the application starts we run the backfill missing initial stage migration. This is the same one we had in the migrations but run from the code. It solves the issue for the newly created features after the migration and before the feature was enabled.


FredrikOseberg (approved):
LGTM

PR #7064, opened 2024-05-16
chore: update documentation for feature toggle variants

### LINKED ISSUE #7056: Chore: Is variant type 'number' supported?
The variant type represents payload parsing hints for your application, but the value transmitted over the wire will always be a string. We will only perform validation, not actually parse the value for you. We need to update the documentation to be more clear in how Unleash deals with these scenarios.

### Discussed in https://github.com/orgs/Unleash/discussions/6954

<div type='discussions-op-text'>

<sup>Originally posted by **pc-pallon** April 26, 2024</sup>
In the code such as [here](https://github.com/Unleash/unleash/blob/49e84d3a911a1fc66c374cd97d63a6554092222e/src/lib/types/model.ts#L140) and in the UI such as in the screenshot below, the variant type 'number' seems to be supported.
Screenshot 2024-04-26 at 14 25 01

However, in other code locations such as [here](https://github.com/Unleash/unleash/blob/49e84d3a911a1fc66c374cd97d63a6554092222e/frontend/src/utils/variants.ts#L1) and in the document [here](https://docs.getunleash.io/reference/feature-toggle-variants#variant-payload), there is no type 'number'.

Is it actually supported?</div>

FredrikOseberg:
Add documentation of feature flag variant types

Closes #7056

### DIFF
--- a/website/docs/reference/feature-toggle-variants.md
+++ b/website/docs/reference/feature-toggle-variants.md
@@ -87,6 +87,9 @@ Unleash currently supports these payload types:
 - JSON
 - CSV
 - String
+- Number
+
+The variant payload type provides hints to your application about how to parse the payload. It does not enforce any specific format, so you can use any format you like. The payload is passed to the client as a string, so you will need to parse it on the client side according to the datatype expected.
 
 ### Variant stickiness
 


chriswk (approved):
LGTM

PR #7065, opened 2024-05-16
fix: bearer tokens with base-path

gastonfournier:
## About the changes
We've identified that Bearer token middleware is not working for /enterprise instance.

Looking at a few lines below: https://github.com/Unleash/unleash/blob/88e3b1b79efc0954951a4ee270483c1fd2bf97dc/src/lib/app.ts#L81-L84 we can see that we were missing the basePath in the use definition.


chriswk (approved):
Good catch.

PR #7068, opened 2024-05-16
fix: reached stage should emit feature name

kwasniew:
## About the changes

When we emit event with a state that we reached in a lifecycle a feature name should be included. This should remove flaky test.


gastonfournier (approved):
Nice!

PR #7070, opened 2024-05-17
chore: automate release actions

gastonfournier:
## About the changes
Automate some of our release steps in the same workflow. This gives us the ability to opt-out some of the jobs which allows us to release patches to older versions without setting that patch as the latest version (currently not covering docker publish but if this test works successfully we might include it as well). 

This comes from tests in our enterprise repository.

### DIFF
--- a/.github/workflows/update_version_for_version_checker.yml
+++ b/.github/workflows/update_version_for_version_checker.yml
@@ -25,12 +25,6 @@ jobs:
           workload_identity_provider: 'projects/340004706233/locations/global/workloadIdentityPools/gh-actions-pool/providers/github-actions-oidc-unleash'
           service_account: 'versionUpdateSa@metrics-304612.iam.gserviceaccount.com'
           token_format: 'access_token'
-      - name: Update version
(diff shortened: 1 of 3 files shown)


.github/workflows/update_version_for_version_checker.yml
```diff
@@ -25,12 +25,6 @@ jobs:
           workload_identity_provider: 'projects/340004706233/locations/global/workloadIdentityPools/gh-actions-pool/providers/github-actions-oidc-unleash'
           service_account: 'versionUpdateSa@metrics-304612.iam.gserviceaccount.com'
           token_format: 'access_token'
-      - name: Update version
-        if: ${{ github.event_name == 'release' }}
-        uses: 'Unleash/update-version-action@v0'
-        with:
-          version: ${{ github.event.release.tag_name }}
-          distribution: 'oss'
```
gastonfournier, line 33: This doesn't work and I'm moving it into the publish-new-version workflow

daveleek (approved):
Cool, lets try it out!

PR #7075, opened 2024-05-17
fix: new strategy using default strategy

kwasniew:
## About the changes

First click on add strategy didn't respect the default strategy. Subsequent ones did. The reason for that is useEffect didn't observe the default strategy being updated from no strategy to actual strategy from the API. 

Also, updated useProject to useProjectOverview which is part of the useProject deprecation. They are almost the same. The only difference is project overview doesn't read all project features which we don't need here to read env default strategy. 

## Discussion points

Since we're observing all changes to the default strategy it means that if I'm editing a strategy and the default project environment strategy changes my edits are erased and I have to start from scratch. I'm assuming that default strategy changes are so rare it shouldn't be a problem and more complex logic is not justified.

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
@@ -119,6 +119,7 @@ export const FeatureStrategyCreate = () => {
     }, [
         featureId,
         JSON.stringify(strategyDefinition),
+        JSON.stringify(defaultStrategy),
         shouldUseDefaultStrategy,
     ]);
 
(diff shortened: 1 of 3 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
```diff
@@ -119,6 +119,7 @@ export const FeatureStrategyCreate = () => {
     }, [
         featureId,
         JSON.stringify(strategyDefinition),
+        JSON.stringify(defaultStrategy),
```
kwasniew, line 122: This is a fix for the bug

sjaanus (approved):

PR #7078, opened 2024-05-20
Update app.json

jeremylynch:
Getting this error when deploying to Heroku: 

```
We tried to create heroku-postgresql:essential-0, but received an error from the add-on provider. Try the request again, or log a support ticket and include this message: Unsupported version: 13. We support the following versions: 14, 15, 16
```

This PR upgrades the postgres version to 14

### DIFF
--- a/app.json
+++ b/app.json
@@ -12,7 +12,7 @@
     {
       "plan": "heroku-postgresql",
       "options": {
-        "version": "13"
+        "version": "14"
       }
     }
   ]


sjaanus (approved):

kwasniew (approved):
Looks like Heroku is deprecating version 13 so this change seems to be reasonable: https://devcenter.heroku.com/articles/heroku-postgres-version-support

PR #7080, opened 2024-05-20
feat: outdated sdks project level

sjaanus:
This adds project level endpoint to catch outdated SDKs only for that project.


src/lib/types/stores/client-instance-store.ts
```diff
@@ -29,6 +29,9 @@ export interface IClientInstanceStore
     groupApplicationsBySdk(): Promise<
         { sdkVersion: string; applications: string[] }[]
     >;
+    groupApplicationsBySdkAndProject(
```
kwasniew, line 32: are we grouping by project? I thought we're only filtering/where by project? Maybe we can add optional projectId to the method signature and only keep 1 method?

PR #7087, opened 2024-05-20
refactor: replace useProject with useProjectOverview

kwasniew:
## About the changes

* useProject replaced with useProjectOverview which doesn't fetch all project features
* added client side feature counting based on flag types
* removed project list on hover that was making 3 requests on every tile hover. We don't need to prefetch all this information (especially all the project features)


frontend/src/hooks/api/getters/useProjectOverview/useProjectOverview.ts
```diff
@@ -50,4 +50,12 @@ export const useProjectOverviewNameOrId = (id: string): string => {
     return useProjectOverview(id).project.name || id;
 };
 
+export const featuresCount = (
```
kwasniew, line 53: backend doesn't return all features for the overview, but it already counts features by feature type so we can sum all types

frontend/src/component/project/ProjectList/ProjectGroup.tsx
```diff
@@ -108,9 +107,7 @@ export const ProjectGroup: React.FC<{
                                             to={`/projects/${project.id}`}
                                         >
                                             <ProjectCard
-                                                onHover={() =>
-                                                    handleHover(project.id)
-                                                }
+                                                onHover={() => {}}
```
sjaanus, line 110: Why empty? make it optional or remove?

sjaanus (approved):

frontend/src/component/project/ProjectList/ProjectGroup.tsx
kwasniew: underlying component makes it required and I'd like to remove it in a separate PR

PR #7088, opened 2024-05-20
feat: add global isAdmin method for access service

sjaanus:
This is preparation, so we do not need to define an `isAdmin` method in the change request context.

### DIFF
--- a/src/lib/services/access-service.test.ts
+++ b/src/lib/services/access-service.test.ts
@@ -294,7 +294,7 @@ test('should return true if user has admin role', async () => {
         .fn()
         .mockResolvedValue([{ id: 1, name: 'ADMIN', type: 'custom' }]);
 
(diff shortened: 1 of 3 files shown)


src/lib/services/access-service.test.ts
```diff
@@ -285,3 +285,31 @@ describe('addAccessToProject', () => {
...
     });
 });
+
+test('should return true if user has admin role', async () => {
+    const { accessService, accessStore } = getSetup();
+
+    const userId = 1;
+    accessStore.getRolesForUserId = jest
+        .fn()
+        .mockResolvedValue([{ id: 1, name: 'ADMIN', type: 'custom' }]);
```
kwasniew, line 295: `() => [{ id: 1, name: 'ADMIN', type: 'custom' }]` is simpler

kwasniew (approved):

PR #7089, opened 2024-05-20
chore: workflows call workflows

gastonfournier:
Relying on tags to trigger workflows makes it hard to trace what's happening after a release, currently:
1. We manually trigger a release workflow
2. The release workflow executes and tags the new release in code
3. Several other workflows trigger after matching the tag doing different things: build docker images, tarballs and other things.

This creates a loose dependency between the workflows which are actually part of the same "release workflow" which makes it difficult to spot when one or other dependent workflow fails because the dependency is indirect through the tagging mechanism.

This PR switches to a more direct approach using [workflow calls](https://docs.github.com/en/actions/using-workflows/reusing-workflows). This will create a graph as shown in the following graph:  making it easier to track and identify any problem.

The "drawback" of this approach is that previously we could trigger all dependent workflows at once by creating a tag matching the expected pattern without manually triggering a new release. This limitation can be overcome by adding a manual workflow_dispatch to the workflows using the tag trigger.


.github/workflows/docker_publish.yaml
```diff
@@ -6,14 +6,8 @@ on:
...
       - website/**
-    tags:
-      - 'v*'
+  workflow_call:
   workflow_dispatch:
-    inputs:
-      ignore-push:
-        description: 'Ignore push to dockerhub. If not set the image will be pushed with the sha of the commit as tag'
-        required: false
-        type: boolean
```
gastonfournier, line 16: Removed the ability of ignoring pushes. We never use it, it was mainly for testing this workflow during development

.github/workflows/docker_publish.yaml
```diff
@@ -6,14 +6,8 @@ on:
       - main
     paths-ignore:
       - website/**
-    tags:
-      - 'v*'
+  workflow_call:
```
gastonfournier, line 9: Instead of relying on tags, publish docker from the release workflow: https://github.com/Unleash/unleash/pull/7089/files#diff-c15c3b26808da8f11c4b99dba3911b9ded250104a2c8991073e1993e90444612R111

.github/workflows/docker_publish.yaml
```diff
@@ -35,10 +29,10 @@ jobs:
...
             unleashorg/unleash-server
           tags: |
-            # only enabled for v* tags:
-            type=semver,pattern={{ version }},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
-            type=semver,pattern={{ major.minor }},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
-            type=semver,pattern={{ major }},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+            # only enabled for workflow dispatch except main (assume its a release):
+            type=semver,pattern={{ version }},enable=${{ github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main' }}
+            type=semver,pattern={{ major.minor }},enable=${{ github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main' }}
+            type=semver,pattern={{ major }},enable=${{ github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main' }}
```
gastonfournier, line 35: We can't check that the ref starts with tags because the ref will be the one of the branch we're releasing from. Instead, whenever this workflow is called will tag the image with semver, except when called from main (cause we're not releasing from main)

.github/workflows/release.yaml
```diff
@@ -1,8 +1,6 @@
 name: 'Publish to npm'
 on:
-  push:
-    tags:
-      - 'v*'
+  workflow_call:
```
gastonfournier, line 3: Publishing to npm based on a workflow call. If we want to still publish on tag we can add workflow_dispatch to call this manually

.github/workflows/release_changelog.yml
```diff
@@ -1,4 +1,4 @@
-name: 'Releases'
+name: 'Release changelog'
```
gastonfournier, line 1: Just a rename. Keeping this workflow relying on tags because it uses a deprecated action and relies on tag refs. We should migrate this: https://github.com/Unleash/unleash/actions/runs/9125610662/job/25092161594#step:4:3

chriswk (approved):
I like the idea :)

PR #7093, opened 2024-05-21
test: fix flaky lifecycle test

kwasniew:
## About the changes

Adding feature name check when reaching a certain lifecycle stage to check if we're looking at the right thing.

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
@@ -98,7 +98,8 @@ const uncompleteFeature = async (featureName: string, expectedCode = 200) => {
 function reachedStage(feature: string, stage: StageName) {
     return new Promise((resolve) =>
         featureLifecycleService.on(STAGE_ENTERED, (event) => {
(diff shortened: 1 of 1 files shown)


src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
```diff
@@ -98,7 +98,8 @@ const uncompleteFeature = async (featureName: string, expectedCode = 200) => {
 function reachedStage(feature: string, stage: StageName) {
     return new Promise((resolve) =>
         featureLifecycleService.on(STAGE_ENTERED, (event) => {
-            if (event.stage === stage) resolve(stage);
+            if (event.stage === stage && event.feature === feature)
```
kwasniew, line 101: when you introduce feature to a method signature that would be cool if you used it :)

chriswk (approved):
Nice

src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
gastonfournier: wow! I totally missed it!

PR #7094, opened 2024-05-21
Temp edge cache flag removal

gastonfournier:
## About the changes
EdgeService is the only place where we use active tokens validation in bulk. By switching to validating from the cache, we no longer need a method to return all active tokens from the DB.


chriswk (approved):
Nice. Let's get this out.

PR #7095, opened 2024-05-21
Node20

chriswk:
Upgrades workflows, nvmrc and package.json to use Node 20.


.github/workflows/build_doc_prs.yaml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   validate:
-    name: build (18.x) # temporary solution to trick branch protection rules
+    name: build (20.x) # temporary solution to trick branch protection rules
```
gastonfournier, line 10: Careful with the name, as we need to change the branch protection rule. Maybe a good opportunity to just use `build`. It's tricky when using matrix because it changes the names and the branch protection rules with it, but we're not using matrix build anymore

src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
```diff
@@ -125,7 +125,7 @@ test('should return lifecycle stages', async () => {
         bucket: {
             toggles: {
                 my_feature_a: 'irrelevant',
-                non_existent_feature: 'irrelevent',
+                non_existent_feature: 'irrelevant',
```
gastonfournier, line 128: :detective: :medal_sports:

gastonfournier (approved):
LG!

src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
sighphyre: This change is so meta I don't even know how to respond to this

chriswk: You don't say. :)

.github/workflows/build_doc_prs.yaml
chriswk: I've changed to build now, and once merged, I'll update the protection rules.

PR #7116, opened 2024-05-23
fix: always hide horizontal overflow

thomasheartman:
Set Y scroll to auto in compact and modal modes.

None of our modals should ever be wider than the viewport, so we can always hide horizontal overflow.


thomasheartman:
Closed in favor of #7198

PR #7120, opened 2024-05-23
feat: add ip to state-service and group-service

gastonfournier:
The add ip to two services. Despite state is being deprecated, I think we better get it out of the way.

I've tested with enterprise and tried not to break compatibility


src/lib/services/group-service.ts
```diff
@@ -286,6 +289,52 @@ export class GroupService {
         }
     }
 
+    async syncExternalGroupsWithAudit(
```
gastonfournier, line 292: This method is a replacement of syncExternalGroups. It's only used in enterprise repo and because of that overloading the parameters was more complicated. After updating enterprise I'll remove the other method which is now marked as deprecated

chriswk (approved):
LGTM

PR #7124, opened 2024-05-23
chore: remove e2e that is not needed anymore

sjaanus:
We talked with @nunogois that this test is testing migration from 2023, but time has passed and the migration is working properly, so we think to cut down on test run time, we can remove it.


nunogois (approved):
There's also a snap we can get rid of:
`src/test/e2e/__snapshots__/favor-permission-name-over-id.e2e.test.ts.snap`

Unrelated to your current task, but maybe we should consider cleaning up other similar migration tests in the near future, like:
 - `dedupe-permissions.e2e.test.ts`
 - `system-user-migration.test.ts`

Our reasoning, in case anyone stumbles upon this:

These migrations already ran for some time, so they are already "validated". They are also frozen in time and will not be changed. If anything, we'll create new migrations instead.

So in a way these tests are technical debt. They don't provide us any value currently and only add to our test run time.

PR #7130, opened 2024-05-23
docs: Okta SCIM setup

sighphyre:
Adds docs for Okta/SCIM setup


chriswk (approved):
I think this makes sense.

PR #7132, opened 2024-05-23
feat: Navigation switch

kwasniew:
## About the changes

Closes #


frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationSidebar.tsx
```diff
@@ -121,9 +124,11 @@ const MiniListItem: FC<{ href: string; text: string }> = ({
 
 export const StyledBox = styled(Box)(({ theme }) => ({
     backgroundColor: theme.palette.background.paper,
-    pt: theme.spacing(3),
-    pb: theme.spacing(3),
-    minHeight: '95vh',
+    paddingTop: theme.spacing(2),
+    paddingBottom: theme.spacing(6),
+    height: '100%',
+    position: 'absolute',
```
kwasniew, line 130: removes sidebar from regular flow

FredrikOseberg: I'm not sure about this change, since I'd like the content area to have the same padding on all sides when the sidebar is active. But let's see how it plays out.

FredrikOseberg (approved):

PR #7143, opened 2024-05-24
chore: rename toggle to flags in docs #3

sjaanus:


website/docs/reference/strategy-variants.md
```diff
@@ -77,9 +77,9 @@ Strategy stickiness is calculated on the received user and context, as described
 
 If you would like to reassign users to different variants using existing stickiness parameter then you can change the groupId of the strategy. This will provide different input to the stickiness calculation.
 
-### Strategy variants vs feature toggle variants
+### Strategy variants vs feature flag variants
```
kwasniew, line 80: how do you know that no one references this link?

kwasniew (approved):

website/docs/reference/strategy-variants.md
sjaanus: Good catch, to be honest, most of them I just search one by one. But I think this one is actually being referenced.

nnennandukwe (approved):
Awesome! 💯

PR #7144, opened 2024-05-24
feat: Persist navigation settings

kwasniew:
## About the changes

* can persist settings for the navigation. It will keep mini or full mode according to your preferences.
* added generic drop-in replacement for the useState hook that persists things in localStorage. I don't want the complexity of state and local storage syncing to leak out to app code


sjaanus (approved):
LG

PR #7149, opened 2024-05-24
feat: add remote ip to all events (2)

gastonfournier:
## About the changes
This aligns us with the requirement of having ip in all events. After tackling the enterprise part we will be able to make the ip field mandatory here: https://github.com/Unleash/unleash/blob/2c66a4ace44e3553195f45a8eafa0d19ca26113c/src/lib/types/events.ts#L362

### DIFF
--- a/src/lib/db/group-store.ts
+++ b/src/lib/db/group-store.ts
@@ -33,6 +33,8 @@ const GROUP_COLUMNS = [
     'scim_id',
 ];
 
+export const SSO_SYNC_USER = 'SSO';
+
 const rowToGroup = (row) => {
     if (!row) {
         throw new NotFoundError('No group found');
@@ -332,7 +334,7 @@ export default class GroupStore implements IGroupStore {
                     })
                     .orWhereRaw('jsonb_array_length(mappings_sso) = 0'),
             )
-            .where({ 'gu.user_id': userId, 'gu.created_by': 'SSO' });
+            .where({ 'gu.user_id': userId, 'gu.created_by': SSO_SYNC_USER });
 
         return rows.map(rowToGroupUser);
     }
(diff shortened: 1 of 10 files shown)


src/lib/db/group-store.ts
```diff
@@ -33,6 +33,8 @@ const GROUP_COLUMNS = [
     'scim_id',
 ];
 
+export const SSO_SYNC_USER = 'SSO';
```
gastonfournier, line 36: I'm making this very explicit cause it gave me a headache to realize this username **has to be used** in the sync groups workflow instead of the audit user username (but only to call this store)

src/lib/services/group-service.ts
```diff
@@ -302,7 +303,7 @@ export class GroupService {
             await this.groupStore.addUserToGroups(
                 userId,
                 newGroups.map((g) => g.id),
-                auditUser.username,
+                SSO_SYNC_USER,
```
gastonfournier, line 306: This change fixes the e2e tests. This is needed because the assumption the method `getOldGroupsForExternalUser` does is that all group sync is stored under this username.

src/lib/services/access-service.test.ts
```diff
@@ -201,7 +201,7 @@ test('user with custom root role should get a user root role', async () => {
     expect(role.name).toBe('custom-root-role');
     const events = await eventStore.getEvents();
     expect(events).toHaveLength(1);
-    expect(events[0]).toEqual({
+    expect(events[0]).toMatchObject({
```
gastonfournier, line 204: This is because of the changes in fake event store.

chriswk (approved):
LGTM. I believe SYSTEM_USER has an ip field btw

gastonfournier:
> LGTM. I believe SYSTEM_USER has an ip field btw

Thanks! Unfortunately (for me), the ip is empty: https://github.com/Unleash/unleash/blob/2c66a4ace44e3553195f45a8eafa0d19ca26113c/src/lib/types/core.ts#L44-L48 so I might have to touch a few tests... I'll keep it as is for now :+1:

PR #7157, opened 2024-05-27
chore!: [v6] remove error.description in error messages

thomasheartman:
In preparation for v6, this PR removes usage and references to `error.description` instead favoring `error.message` (as mentioned #4380)

I found no references in the front end, so this might be (I believe it to be) all the required changes.


chriswk (approved):
Looks like you caught them all. Gogo!

PR #7166, opened 2024-05-27
chore: switch insights ui flag to kill switch

sjaanus:
The flag is already set up in our Unleash instance.


daveleek (approved):
Nice, you have some merge issues and snapshot issues on `routes.test.tsx`, but after that 👍

PR #7182, opened 2024-05-28
feat: currently selected nav item

kwasniew:
## About the changes

Adding a marker for the currently selected item. Works for full and mini sidebar.
Screenshot 2024-05-28 at 10 30 07

The tricky part is to set the initial value. I created a hook `useInitialPathname` that takes the URL and extracts the top level navigation part of the URL. E.g. /project/my-project should normalize to /project for the sidebar active item selection.

### DIFF
--- a/frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
+++ b/frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
@@ -1,4 +1,4 @@
-import { type FC, type ReactNode, useCallback } from 'react';
+import { type FC, useCallback } from 'react';
 import type { INavigationMenuItem } from 'interfaces/route';
 import type { NavigationMode } from './NavigationMode';
 import {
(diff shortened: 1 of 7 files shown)


frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
```diff
@@ -121,26 +123,43 @@ export const OtherLinksList = () => {
...
     const DynamicListItem = mode === 'mini' ? MiniListItem : FullListItem;
 
     return (
         <List>
-            <DynamicListItem href='/projects' text='Projects' onClick={onClick}>
+            <DynamicListItem
+                href='/projects'
+                text='Projects'
+                onClick={() => onClick('/projects')}
+                selected={activeItem === '/projects'}
```
daveleek, line 137: Yeah I guess this might be the most elegant way to go about this, to tell each component which route-parent is the current one. 🤔

daveleek (approved):
LGTM!

PR #7184, opened 2024-05-28
chore: remove state service

daveleek:
## About the changes

Removes the deprecated state endpoint, state-service (despite the service itself not having been marked as deprecated), and the file import in server-impl. Leaves a TODO in place of where file import was as traces for a replacement file import based on the new import/export functionality


src/test/e2e/stores/feature-toggle-client-store.e2e.test.ts
```diff
@@ -1,36 +0,0 @@
...
-test('should be able to fetch client toggles', async () => {
-    const response = await app.request
-        .post('/api/admin/state/import?drop=true')
-        .attach('file', 'src/test/examples/exported-segments.json');
-
-    expect(response.status).toBe(202);
-
-    const clientToggles = await clientFeatureToggleStore.getClient({});
-    expect(clientToggles).toHaveLength(1);
-});
```
daveleek, line 36: Only test in the file. The method it actually tests is used in and covered by tests for `client-feature-toggle-service.getClientFeatures`, none of the other methods in client-feature-toggle-store has tests like this.

website/docs/how-to/how-to-import-export.md
```diff
@@ -31,6 +31,12 @@ Be careful when using the `drop` parameter in production environments: cleaning
 
 ### State Service {#state-service}
 
+:::caution Removal notice
+
+State Service has been removed as of Unleash 6.0
```
daveleek, line 36: Is this good enough? Should we point out the fact that Pro will "get this" before v6?

src/lib/services/state-schema.ts
```diff
@@ -1,70 +0,0 @@
...
-});
-
-export const environmentSchema = joi.object().keys({
-    name: nameType,
-    displayName: joi.string().optional().allow(''),
-    type: joi.string().required(),
-    sortOrder: joi.number().optional(),
-    enabled: joi.boolean().optional(),
-    protected: joi.boolean().optional(),
-});
```
daveleek, line 41: Will move this to enterprise as it's referenced there

src/lib/services/state-schema.ts
```diff
@@ -1,70 +0,0 @@
...
-
-export const updateEnvironmentSchema = joi.object().keys({
-    displayName: joi.string().optional().allow(''),
-    type: joi.string().optional(),
-    sortOrder: joi.number().optional(),
-});
```
daveleek, line 47: Moving this to enterprise as well

src/lib/services/state-schema.ts
```diff
@@ -1,70 +0,0 @@
...
-
-export const sortOrderSchema = joi.object().pattern(/^/, joi.number());
```
daveleek, line 49: Was moved to separate file sort-order-schema.ts as it was the only schema still being used after removing state service/endpoints

src/lib/features/project-environments/environment-service.ts
```diff
@@ -16,7 +16,7 @@ import {
 import type { Logger } from '../../logger';
 import { BadDataError, UNIQUE_CONSTRAINT_VIOLATION } from '../../error';
 import NameExistsError from '../../error/name-exists-error';
-import { sortOrderSchema } from '../../services/state-schema';
+import { sortOrderSchema } from '../../services/sort-order-schema';
```
daveleek, line 19: State service/schema being removed, so moving this to separate file

kwasniew (approved):

PR #7186, opened 2024-05-28
chore: deprecate custom strategies

sighphyre:
This marks a whole whack load of things as deprecated when trying to access or use custom strategies:

- When navigating to a flag that has a custom strategy present, you'll get a short text warning that they're deprecated and a link to constraints. This is a small snippet that's intended to raise awareness going forward because generally once strategies are defined users won't interact with them again.
- When adding a custom strategy to a flag, you'll get a warning that they're deprecated and a link to the constraints docs
- When accessing the strategies page (as part of the create custom strategy flow), you'll get a more comprehensive warning telling you that they may be removed in a future release and a warning to not lean on them any more. **You'll also get a suggestion to reach out to us if you really, really need them**
- If you ignore the previous warning, it'll show again in the create strategy UI with yet another link to the constraints.

The text is a little different each time to make it fit better within the context you're operating from.

OpenAPI docs also now mark these as deprecated and suggest using constraints instead.

I'd love to have a comprehensive guide on how to migrate to constraints but I don't think that can exist - you can do anything with custom strategies, including things you shouldn't be able to do, but constraints are more limited.

Warning when viewing a flag with a custom strategy:

image

Warning when trying to add a custom strategy to an existing flag:

image

Warning when accessing custom strategies list:

image

Warning when trying to add a new custom strategy:

image


frontend/src/utils/strategyNames.tsx
```diff
@@ -36,6 +36,17 @@ export const getFeatureStrategyIcon = (strategyName: string) => {
     }
 };
 
+export const BuiltInStrategies = [
```
sighphyre, line 39: Moved, from a previous place

frontend/src/component/strategies/CustomStrategyInfo/CustomStrategyInfo.tsx
```diff
@@ -17,11 +17,8 @@ export const CustomStrategyInfo: FC<{ alert?: boolean }> = ({ alert }) => {
     const content = (
         <>
             <Paragraph>
-                We recommend you to use the predefined strategies like Gradual
-                rollout with constraints instead of creating a custom strategy.
-            </Paragraph>
-            <Paragraph>
-                If you decide to create a custom strategy be aware of:
+                If you decide to create a custom strategy be aware of the
```
sighphyre, line 20: The recommendation is handled by the warnings now instead. No point in duplicating the information

frontend/src/component/strategies/CustomStrategyInfo/CustomStrategyInfo.tsx
```diff
@@ -32,8 +29,8 @@ export const CustomStrategyInfo: FC<{ alert?: boolean }> = ({ alert }) => {
                         to evaluate differently
                     </li>
                     <li>
-                        Requires a lot of configuration in both Unleash admin UI
-                        and the SDK.
+                        Custom strategies require a lot of configuration in both
```
sighphyre, line 32: Just boyscouting

frontend/src/component/feature/FeatureStrategy/FeatureStrategyForm/FeatureStrategyForm.tsx
```diff
@@ -389,6 +390,29 @@ export const FeatureStrategyForm = ({
                         />
                     }
                 />
+
+                <ConditionallyRender
+                    condition={
+                        !BuiltInStrategies.includes(strategy.name || 'default')
```
sighphyre, line 396: In theory the strategy name should never be undefined but that's what our types say. This is probably a very, very legacy hangover from iteration 1 of Unleash when strategies didn't exist. If that's the case then you effectively have a default strategy

FredrikOseberg (approved):
You covered all the bases. Nice!

PR #7191, opened 2024-05-28
feat: last viewed project

kwasniew:
## About the changes

Uses existing hook to fetch recently viewed project (shown only on desktop)
Screenshot 2024-05-28 at 15 19 27

The current hook doesn't update when local storage updated. So it only works after refresh. Will fix in a separate PR. 

Also minor UI updates:
* icons for service account and banners
* expand/hide moved to the bottom with new icon and at the bottom of the screen
* more spacing between sections
* hiding admin icons in mini view (next PR will add a gear icon to open admin in expanded mode)

Screenshot 2024-05-28 at 15 37 46


frontend/src/component/layout/MainLayout/NavigationSidebar/ListItems.tsx
```diff
@@ -19,6 +21,13 @@ const listItemButtonStyle = (theme: Theme) => ({
     },
 });
 
+const CappedText = styled(Typography)({
```
kwasniew, line 24: project name can be long so we want to add ...

PR #7194, opened 2024-05-29
chore: remove unused artillery scripts referencing state api

daveleek:
## About the changes

Should be fairly non-braking.
Found an artillery script that references the state api,  I've never heard of it being in use and it's not changed for 2 years so I'm deleting it


gardleopard (approved):
LGTM

PR #7199, opened 2024-05-29
chore: remove unused and deprecated methods in feature toggle legacy controller and in feature toggle service

daveleek:
## About the changes

Removes unused and deprecated methods in feature toggle legacy controller (methods not having an endpoint)
Removes unused and deprecated methods in feature toggle service (methods not in use in Unleash)


chriswk (approved):
Feels good man

PR #7200, opened 2024-05-29
fix: when finding median time to production, ignore 0s

sjaanus:
We have an issue that if you open up Insights, the Time to Production chart was showing nothing because it was taking the median across all projects. You might have many new or empty projects where the median was 0 (no data).

For example, the median from [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.7, 50.3, 140] was 0.

Now, we will remove the 0 values to have a more reasonable median.


chriswk (approved):
LGTM

PR #7202, opened 2024-05-29
fix: sort segments before comparing in cr diff calculations

thomasheartman:
This change fixes a bug where we would show the list of segments as
changed (causing a conflict) if their order wasn't the same in the
change as in the original.

By sorting the segments before comparing them, we can avoid this issue.

To avoid modifying the objects that are passed in (in case that has knock-on effects anywhere), we copy the objects. And because `toSorted` "only" has about 89% coverage now, I chose to use `sort` and spread the arrays instead.

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeOverwriteWarning/strategy-change-diff-calculation.ts
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeOverwriteWarning/strategy-change-diff-calculation.ts
@@ -149,7 +149,7 @@ export function getStrategyChangesThatWouldBeOverwritten(
                 ...change,
                 payload: {
                     ...change.payload,
-                    segments: [...(change.payload.segments ?? [])].sort(),
+                    segments: [...change.payload.segments].sort(),
                 },
             };
 
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeOverwriteWarning/strategy-change-diff-calculation.ts
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeOverwriteWarning/strategy-change-diff-calculation.ts
@@ -142,6 +142,11 @@ export function getStrategyChangesThatWouldBeOverwritten(
     change: IChangeRequestUpdateStrategy,
 ): ChangesThatWouldBeOverwritten | null {
     const fallbacks = { segments: [], variants: [], title: '' };
+    if (change.payload.segments && currentStrategyConfig?.segments) {
+        change.payload.segments.sort();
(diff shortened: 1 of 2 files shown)


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeOverwriteWarning/strategy-change-diff-calculation.ts
```diff
@@ -142,9 +142,31 @@ export function getStrategyChangesThatWouldBeOverwritten(
...
 ): ChangesThatWouldBeOverwritten | null {
     const fallbacks = { segments: [], variants: [], title: '' };
+
+    const withSortedSegments = (() => {
+        if (change.payload.segments && currentStrategyConfig?.segments) {
+            const changeCopy = {
+                ...change,
+                payload: {
+                    ...change.payload,
+                    segments: [...(change.payload.segments ?? [])].sort(),
```
nunogois, line 152: Do we still need the `??` since we're truthy checking this on line 147?

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeOverwriteWarning/strategy-change-diff-calculation.ts
```diff
@@ -142,9 +142,31 @@ export function getStrategyChangesThatWouldBeOverwritten(
...
+            };
+
+            const currentCopy = {
+                ...currentStrategyConfig,
+                segments: [...currentStrategyConfig.segments].sort(),
+            };
+
+            return { current: currentCopy, change: changeCopy };
+        } else {
+            return { current: currentStrategyConfig, change: change };
```
nunogois, line 163: Nitpicking, but maybe it would look cleaner if we early returned instead.
```ts
if (!change.payload.segments || !currentStrategyConfig?.segments) {
  return { current: currentStrategyConfig, change };
}
// ...
```

nunogois (approved):
Great reasoning 👍

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeOverwriteWarning/strategy-change-diff-calculation.ts
thomasheartman: Nope; good call!

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeOverwriteWarning/strategy-change-diff-calculation.ts
thomasheartman: Yeah, good shout.

PR #7203, opened 2024-05-29
fix: footer overflow

kwasniew:
## About the changes

Footer component overflows the main content making some elements e.g. hide/expand not clickable. Instead of zIndex escalation I prefer to prevent the overflow. 
Screenshot 2024-05-29 at 10 51 48

Root cause:
Footer uses mui grid. The grid has a negative margin. The negative margin comes from the spacing property on the grid. It has this behavior: https://github.com/mui/material-ui/issues/10223. We could also remove the spacing and set it to 0 but the spacing is needed for keeping footer elements apart vertically and we can live with the horizontal impact of the negative margin once we hide it behind overflow hidden.


daveleek (approved):
Interesting, should take care of the problem at least!

PR #7208, opened 2024-05-29
feat: Synced last viewed projects

kwasniew:
## About the changes

When you use this line in multiple places:

`const {lastViewed, setLastViewed} = useLastViewedProject();`

they are all synchronized and lastViewed is always kept in sync through the custom event emitted on document and values fetched from local storage.

Screenshot 2024-05-29 at 12 05 11

### DIFF
--- a/frontend/src/hooks/useCustomEvent.ts
+++ b/frontend/src/hooks/useCustomEvent.ts
@@ -1,0 +1,27 @@
+import { useEffect } from 'react';
+
+/**
+ * A hook that provides methods to emit and listen to custom DOM events.
+ * @param eventName The name of the event to listen for and dispatch.
+ */
+export const useCustomEvent = (
+    eventName: string,
+    handler: (event: CustomEvent) => void,
+) => {
+    const emitEvent = () => {
+        const event = new CustomEvent(eventName);
+        document.dispatchEvent(event);
+    };
+
+    useEffect(() => {
+        const eventListener = (event: Event) => handler(event as CustomEvent);
+
+        document.addEventListener(eventName, eventListener);
+
+        return () => {
+            document.removeEventListener(eventName, eventListener);
+        };
+    }, [eventName, handler]);
+
+    return { emitEvent };
+};
--- a/frontend/src/hooks/useLastViewedProject.ts
+++ b/frontend/src/hooks/useLastViewedProject.ts
@@ -10,7 +10,7 @@ export const useLastViewedProject = () => {
         return getLocalStorageItem(key);
     });
 
-    const { emitEvent } = useCustomEvent('lastViewedUpdated', () => {
+    const { emitEvent } = useCustomEvent('lastViewedProjectUpdated', () => {
         setLastViewed(getLocalStorageItem(key));
     });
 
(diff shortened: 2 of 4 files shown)


frontend/src/hooks/useCustomEvent.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from 'react';
+
+/**
+ * A hook that provides methods to emit and listen to custom DOM events.
+ * @param eventName The name of the event to listen for and dispatch.
+ */
+export const useCustomEvent = (
```
kwasniew, line 7: this hook uses document as event listener that can propagate events across the UI

frontend/src/hooks/useLastViewedProject.ts
```diff
@@ -9,9 +10,14 @@ export const useLastViewedProject = () => {
...
     });
 
+    const { emitEvent } = useCustomEvent('lastViewedUpdated', () => {
+        setLastViewed(getLocalStorageItem(key));
+    });
+
     useEffect(() => {
         if (lastViewed) {
             setLocalStorageItem(key, lastViewed);
+            emitEvent();
```
kwasniew, line 20: every time we update local storage in one hook, all other hooks created from the same fn will get their local state updated

frontend/src/hooks/useLastViewedProject.ts
```diff
@@ -9,9 +10,14 @@ export const useLastViewedProject = () => {
         return getLocalStorageItem(key);
     });
 
+    const { emitEvent } = useCustomEvent('lastViewedUpdated', () => {
+        setLastViewed(getLocalStorageItem(key));
```
kwasniew, line 14: when some other hook updated we update our local last viewed state from the local storage

frontend/src/hooks/useCustomEvent.ts
kwasniew: I decided to extract this hook separately and compose in the useLastViewedProject

frontend/src/hooks/useCustomEvent.ts
```diff
@@ -0,0 +1,27 @@
...
+    eventName: string,
```
thomasheartman, line 8: I think leaving it as string for now is fine, but we could consider using an enum later if we find that we use this a lot, to avoid propagating something that we're not catching. Not that that would guarantee anything to be caught, but it might catch some minor typos or case inconsistencies (are dom events case sensitive?).

thomasheartman (approved):
Nice one! I like this solution 😄

frontend/src/hooks/useCustomEvent.ts
kwasniew: Normal usage pattern for now is to use this new hook only once in the useLastViewedProject so whatever we put there (even with typo) is consistently used by all useLastViewedProject() invocations.

kwasniew: events are case sensitive: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#type

thomasheartman: Ah, good point! And because you're automatically listening for the event you created, it doesn't really matter. Yeah, I get it now 😄

PR #7209, opened 2024-05-29
docs: add more clarification on when to use sx vs styled

thomasheartman:
This change adds a section on when to use `sx` vs `styled`. I'm adding
it because it was unclear to me when we should use one over the other.
Hopefully this clears it up and makes it easier for others going forward.

### DIFF
--- a/website/docs/contributing/ADRs/front-end/preferred-styling-method.md
+++ b/website/docs/contributing/ADRs/front-end/preferred-styling-method.md
@@ -11,14 +11,20 @@ In the codebase, we need to have a uniform way of performing style updates.
 
 ## Decision
 
-We have decided to move away from using makeStyles as it's currently deprecated from @material/ui, and kept alive with an 
+We have decided to move away from using makeStyles as it's currently deprecated from @material/ui, and kept alive with an
 external interop package to maintain compatability with the latest version. The preferred path forward is to use styled components which is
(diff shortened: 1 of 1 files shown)


website/docs/contributing/ADRs/front-end/preferred-styling-method.md
```diff
@@ -11,14 +11,20 @@ In the codebase, we need to have a uniform way of performing style updates.
...
+As with everything else, whether to use styled components or the `sx` prop depends on the context.
+
+Styled components have better performance characteristics, but it's fairly minor. So unless you're rendering something a lot of times, it's not really a big deal. But when in doubt: Use styled components. And when using a styled component feels like too much overhead, consider using the `sx` prop.
+
 ### Consequences: code sharing
 
-With makeStyles it was common to reuse CSS fragments via library utilities. 
-In the styled components approach we use themeable functions and  object literals 
+With makeStyles it was common to reuse CSS fragments via library utilities.
+In the styled components approach we use themeable functions and  object literals
```
nunogois, line 27:
```suggestion
In the styled components approach we use themeable functions and object literals.
```

nunogois (approved):
Yep, sounds about right.

Maybe it would be worth adding a link to https://mui.com/system/getting-started/usage/#performance-tradeoffs as a reference.

thomasheartman:
Yes! That's the link I was looking for but couldn't find! Thanks 😄

PR #7210, opened 2024-05-29
feat: now CLIENT_METRICS event will be emitted with new structure

sjaanus:
1. CLIENT_METRICS event will be emitted with new structure
2. CLIENT_METRICS event will be emitted from bulkMetrics endpoint

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
@@ -20,9 +20,10 @@ import type {
 import EventEmitter from 'events';
 import type { Logger } from '../../logger';
 import type EventService from '../events/event-service';
-import type { ValidatedClientMetrics } from '../metrics/shared/schema';
 import type { FeatureLifecycleCompletedSchema } from '../../openapi';
 import { calculateStageDurations } from './calculate-stage-durations';
+import type { IClientMetricsEnv } from '../metrics/client-metrics/client-metrics-store-v2-type';
+import groupBy from 'lodash.groupby';
 
 export const STAGE_ENTERED = 'STAGE_ENTERED';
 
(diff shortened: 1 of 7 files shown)


src/lib/features/metrics/instance/metrics.ts
```diff
@@ -161,8 +162,10 @@ export default class ClientMetricsController extends Controller {
                     promises.push(
                         this.metricsV2.registerBulkMetrics(filteredData),
                     );
+                    this.config.eventBus.emit(CLIENT_METRICS, data);
```
sjaanus, line 165: Starting to emit CLIENT_METRICS from bulk metrics endpoint

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -171,7 +171,7 @@ export default class ClientMetricsServiceV2 {
             );
             await this.registerBulkMetrics(clientMetrics);
 
-            this.config.eventBus.emit(CLIENT_METRICS, value);
+            this.config.eventBus.emit(CLIENT_METRICS, clientMetrics);
```
sjaanus, line 174: We do not need to create mapper, we already have the mapped version available.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -95,13 +96,20 @@ export class FeatureLifecycleService extends EventEmitter {
...
-            async (event: ValidatedClientMetrics) => {
-                if (event.environment) {
-                    const features = Object.keys(event.bucket.toggles);
-                    const environment = event.environment;
-                    await this.checkEnabled(() =>
-                        this.featuresReceivedMetrics(features, environment),
-                    );
+            async (events: IClientMetricsEnv[]) => {
+                if (events.length > 0) {
+                    const groupedByEnvironment = groupBy(events, 'environment');
```
sjaanus, line 101: Since from bulk endpoint we can have events from different environments, we will group it up by environment for lifecycle

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -95,13 +96,20 @@ export class FeatureLifecycleService extends EventEmitter {
...
+
+                    for (const [environment, metrics] of Object.entries(
+                        groupedByEnvironment,
+                    )) {
+                        const features = metrics.map(
+                            (metric) => metric.featureName,
+                        );
+                        await this.checkEnabled(() =>
```
kwasniew, line 109: I'm assuming it's not gonna be more than a few envs. Because if the loop has 100s of entries then sequential awaits would kill performance

kwasniew (approved):

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
sjaanus: I am not sure how edge is sending metrics, I am not sure it is env by env. So just in case I made the grouping to support more than one env and not assume that every item has same env.

PR #7213, opened 2024-05-30
feat: add tooltips to new project creation form config buttons

thomasheartman:
This PR adds tooltips to the new project creation form buttons to make it clearer what they do. The tooltips tell you what the buttons do and contains the same description that we use for docs.

image

The tooltips will cover other buttons on narrow windows, but I think that's an acceptable tradeoff

image


daveleek (approved):
Awesome!

PR #7214, opened 2024-05-30
chore: sync user groups is a system action

gastonfournier:
## About the changes
After an internal conversation, we concluded that syncExternalGroups is an action that Unleash performs as a system, not something triggered by the user. We keep the method and just write the event log that the action was performed by the system user.


chriswk (approved):
Nice

PR #7215, opened 2024-05-30
fix: handle long owner names for projects

thomasheartman:
This change updates the new project card footer to better handle long
names. Previously, if a name was too long to fit on a single line,
it'd wrap. It didn't cause any layout shift, but it made the card look
off.

So instead of wrapping, we now truncate the name and add an ellipsis.

To achieve this, I changed the layout to grid instead of flexbox and
removed a nested flex container.

Before: 

image

After:

image

Other cards remain the same, as shown here (new code on left, old code on right):

image

### DIFF
--- a/frontend/src/component/project/NewProjectCard/ProjectOwners/ProjectOwners.tsx
+++ b/frontend/src/component/project/NewProjectCard/ProjectOwners/ProjectOwners.tsx
@@ -41,15 +41,13 @@ const useOwnersMap = () => {
     };
 };
 
-const StyledContainer = styled('div')(({ theme }) => ({
-    marginBottom: theme.spacing(1),
-    display: 'flex',
-    alignItems: 'flex-end',
-}));
-
 const StyledUserName = styled('p')(({ theme }) => ({
     fontSize: theme.typography.body1.fontSize,
     margin: theme.spacing(0, 0, 0.5, 0),
+    overflowX: 'hidden',
+    textOverflow: 'ellipsis',
+    textWrap: 'nowrap',
+    alignSelf: 'end',
 }));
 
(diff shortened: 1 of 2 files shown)


frontend/src/component/project/NewProjectCard/ProjectOwners/ProjectOwners.tsx
```diff
@@ -41,23 +41,21 @@ const useOwnersMap = () => {
...
+    alignSelf: 'end',
 }));
 
 export const ProjectOwners: FC<IProjectOwnersProps> = ({ owners = [] }) => {
     const ownersMap = useOwnersMap();
     const users = owners.map(ownersMap);
 
     return (
-        <StyledContainer>
+        <>
```
sjaanus, line 58: Why is empty wrapper needed?

sjaanus (approved):
Added one comment

frontend/src/component/project/NewProjectCard/ProjectOwners/ProjectOwners.tsx
thomasheartman: Because you can't return multiple elements from a component, so you need a wrapper. We want these elements to flow into the wrapping grid container instead of being in their own little box. This makes it so that they're part of the same grid instead of a nested box.

thomasheartman: Ah, but you're right! There aren't multiple items. I'll fix.

thomasheartman: No, wait, github was tricking me with the fold. There *are* two items, so they need to be wrapped like this.

PR #7216, opened 2024-05-30
fix: handle long names in new project card footers

thomasheartman:
This is a redo of  #7215. I missed the case where you have multiple owners, causing it to lay out weird in that case.

This version handles that by returning an empty div for the owners instead to fill up that space.
image

image

There **are** edge cases where the owners wrap:
image

But that is also the case with the current implementation:
image

And only happens at very specific breakpoints.


daveleek (approved):
Nice!

PR #7217, opened 2024-05-30
fix: handle overflowing avatars in the new project card

thomasheartman:
This change prevents the project owner avatars in the new project card
footer from overflowing by making the number of avatars to show
configurable.

It defaults to 9, the old hard-coded number of avatars shown, but can
be configured.

The reason it overflows here is probably that the extra button in the
project card footer (the favorite) makes the footer smaller than what
we have for the group card footer.

Before: 
image

After:

image


daveleek (approved):
LGTM!

Issue #6306, opened 2024-02-22
Import from file will be gone - [Deprecated] Import & Export


nejco:
### Describe the feature request

Previously with [[Deprecated] Import & Export](https://docs.getunleash.io/how-to/how-to-import-export) we had an option to import all the data at the startup from a file. We are using it via `quarkus-unleash` [usage](https://github.com/quarkiverse/quarkus-unleash/blob/625fb0d0a56e15f88d116f5f39817cc1b2e80911/deployment/src/main/java/io/quarkiverse/unleash/devservices/UnleashDevServiceProcessor.java#L288). With new [Environment Import & Export](https://docs.getunleash.io/how-to/how-to-environment-import-export#environment-importexport-vs-the-instance-importexport-api) this feature is missing.

### Background

Another use case we have for this feature is to be able to start clean environment from source for testing. We don't want to manually create all unleash resources every time. Synching from another environment is also a bad option since someone might have tempered it.

Is there any way to achieve this after [[Deprecated] Import & Export](https://docs.getunleash.io/how-to/how-to-import-export) will be deleted (planned for v6)? I know we can do it via API manually with API but this will be hard to maintain and it is very error prone. It will also be very hard to determine when unleash is actually stable with all resources created.

### Solution suggestions

Keep import from file feature planned to be deprecated and make sure it supports all new features.

ivarconr:
Thanks for raising this @nejco! 

This is why we have deprecated this feature a year already, to collect feedback. In general this is not a widely adopted capability among Unleash Users. The "old import" was poorly written hard to maintain, and did not solve important needs for Unleash users. It was best suited for an "init state", but was really poor at synchronizing state over time. 

> We are using it via quarkus-unleash [usage](https://github.com/quarkiverse/quarkus-unleash/blob/625fb0d0a56e15f88d116f5f39817cc1b2e80911/deployment/src/main/java/io/quarkiverse/unleash/devservices/UnleashDevServiceProcessor.java#L288).

Can you explain briefly how this work? Is it to configure Unleash with a set up pre-defined feature flags? Are you setting up more than flags? 

In general I do not see big challenges adding a file-import variant for the new import as well. 

> Another use case we have for this feature is to be able to start clean environment from source for testing. We don't want to manually create all unleash resources every time. 

For automated testing we now generally advice to use [Unleash Edge in *offline mode*](https://github.com/Unleash/unleash-edge?tab=readme-ov-file#offline). 

Does that make sense?

[...]

PR #7219, opened 2024-05-31
Feat/file import

daveleek:
## About the changes

Adds a file import option to Unleash

Removes import options:
- dropBeforeImport
- keepExisting

Adds new import options:
- project
- environment

Adds an additional file-import only verify method that skips checking project/flag creation access

## Discussion points

- Should we put this behind a flag? The feature itself IS opt-in, but reuses the IMPORT_FILE env variable (which itself is a very specific variable that you only use when you specifically want to import something and remove before next restart)
- This only addresses OSS. While the interface changes dictates that enterprise will need to implement the interface as well (separate PR), there's currently no mechanism in place to override which import gets used so OSS will currently always be used regardless of hosted option

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -333,11 +333,8 @@ const defaultAuthentication: IAuthOption = {
 
 const defaultImport: WithOptional<IImportOption, 'file'> = {
     file: process.env.IMPORT_FILE,
-    dropBeforeImport: parseEnvVarBoolean(
-        process.env.IMPORT_DROP_BEFORE_IMPORT,
-        false,
-    ),
-    keepExisting: parseEnvVarBoolean(process.env.IMPORT_KEEP_EXISTING, false),
+    project: process.env.IMPORT_PROJECT ?? 'default',
+    environment: process.env.IMPORT_ENVIRONMENT ?? 'development',
 };
 
 const defaultEmail: IEmailOption = {
--- a/src/lib/features/export-import-toggles/export-import-service.ts
+++ b/src/lib/features/export-import-toggles/export-import-service.ts
@@ -5,6 +5,8 @@ import type { IFeatureStrategiesStore } from '../feature-toggle/types/feature-to
 import {
     FeaturesExportedEvent,
     FeaturesImportedEvent,
+    SYSTEM_USER,
+    SYSTEM_USER_AUDIT,
     type FeatureToggleDTO,
     type IAuditUser,
     type IContextFieldStore,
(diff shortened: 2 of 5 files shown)


src/lib/features/export-import-toggles/import-file-reader.ts
```diff
@@ -0,0 +1,8 @@
+import * as fs from 'fs';
+
+export const readFile: (file: string) => Promise<string> = async (file) =>
```
kwasniew, line 3: we can call it read-file.ts since it's a generic utility. Also we may use more high level promise based API (https://nodejs.org/en/learn/manipulating-files/reading-files-with-nodejs). Just need to check if out node version supports it

src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -281,16 +300,40 @@ export default class ExportImportService
         auditUser: IAuditUser,
     ): Promise<void> {
         const cleanedDto = await this.cleanData(dto);
-
         await this.importVerify(cleanedDto, user);
+        await this.processImport(cleanedDto, user, auditUser);
+    }
+
+    async importFromFile(
```
kwasniew, line 307: we can have one e2e test with test file placed next to the test

src/lib/create-config.ts
```diff
@@ -333,11 +333,8 @@ const defaultAuthentication: IAuthOption = {
 
 const defaultImport: WithOptional<IImportOption, 'file'> = {
     file: process.env.IMPORT_FILE,
-    dropBeforeImport: parseEnvVarBoolean(
-        process.env.IMPORT_DROP_BEFORE_IMPORT,
-        false,
-    ),
-    keepExisting: parseEnvVarBoolean(process.env.IMPORT_KEEP_EXISTING, false),
+    project: process.env.IMPORT_PROJECT ?? 'default',
+    environment: process.env.IMPORT_ENVIRONMENT ?? 'development',
```
kwasniew, line 337: we can ask our users what env they use in the old import most often

kwasniew (approved):
Very nice!

src/lib/create-config.ts
daveleek: I've reached out to the user requesting the feature here:
https://github.com/Unleash/unleash/issues/6306#issuecomment-2141404520

src/lib/features/export-import-toggles/import-file-reader.ts
chriswk: even more clear (and in preparation of esm) `import * as fs from 'node:fs'` I know biome sometimes warns about specfiying import.

daveleek: Yeah I'll add a separate PR with this and a e2e test

PR #7220, opened 2024-05-31
chore: make ip mandatory

gastonfournier:
IP should be mandatory for storing events. Automated events should use 127.0.0.1

### DIFF
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -69,7 +69,6 @@ exports[`should create default config 1`] = `
     "_events": {},
     "_eventsCount": 0,
     "_maxListeners": undefined,
-    Symbol(shapeMode): false,
     Symbol(kCapture): false,
   },
   "feedbackUriPath": undefined,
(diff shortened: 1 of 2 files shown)


src/lib/__snapshots__/create-config.test.ts.snap
```diff
@@ -69,7 +69,6 @@ exports[`should create default config 1`] = `
     "_events": {},
     "_eventsCount": 0,
     "_maxListeners": undefined,
-    Symbol(shapeMode): false,
```
chriswk, line 72: You need to update your node version

src/lib/types/events.ts
```diff
@@ -349,17 +349,15 @@ export const IEventTypes = [
...
+// this rerpresents the write model for events
 export interface IBaseEvent {
     type: IEventType;
     createdBy: string;
     createdByUserId: number;
     project?: string;
     environment?: string;
     featureName?: string;
-    ip?: string;
+    ip: string;
```
chriswk, line 360: yes!

chriswk (approved):
You'll need to bump your node version and rerun test with the `-u` flag.

PR #7221, opened 2024-05-31
chore: bump node 20 version in docker image

chriswk:
As the title says, bumps from 20.13.1 to 20.14.0 to ensure v6 gets the latest and greatest next week


sighphyre (approved):
Nice

PR #7222, opened 2024-05-31
chore: bumped node version to 18.20.3

chriswk:


sighphyre (approved):
The change looks fine but why is the formatter complaining in the build?

chriswk:
> The change looks fine but why is the formatter complaining in the build?

because apparently something happened with the package.json format? even if I didn't touch it. I'll make it go green and then merge.

PR #7223, opened 2024-05-31
fix: correctly align project card info when some cards have multi-line names

thomasheartman:
This change updates the styling of the project card bodies so that
alignment of the project data (number of flags, members, health) stays
at the bottom of the project card body, even if some cards have longer
bodies than others. It also updates the icon placement so that it is consistent even when the title is multi-line.

Before this change, the bottom row would be misaligned if some (but
not all) project cards on a row had names long enough to span multiple lines. The project icon would also be misaligned

Before:
image

Misaligned project icon: 
image

After:

image

Icon alignment: 
image


daveleek (approved):
Cool!

PR #7224, opened 2024-05-31
chore: remove project list split feature flags

thomasheartman:
This PR removes all the feature flags related to the project list split and updates the snapshot.

Now the project list will always contain "my projects" and "other projects"


daveleek (approved):
Awesome! Was this flag also in enterprise or just here?

thomasheartman:
No, this was a UI-only flag (it's just filtering stuff), so this is all the cleanup we need.

thomasheartman:
Well, except it being set in the startup file, but I'll have a PR up for that too

PR #7225, opened 2024-05-31
chore: remove flag for new project cards

thomasheartman:
This PR removes the flag for the new project card design, making it GA.

It also removes deprecated components and updates one reference (in the groups card) to the new components instead.


daveleek (approved):
Nice! Same question here, is this flag referenced/set up in enterprise?

thomasheartman:
Yes, this flag is set up in enterprise. Creating a PR for that too.

thomasheartman:
Thanks for checking 😄

PR #7226, opened 2024-05-31
fix: increase performance of outdated SDK query

sjaanus:
Joining might not always be the best solution. If a table contains too much data, and you later run sorting on top of it, it will be slow.

In this case, we will first reduce the instances table to a minimal version because instances usually share the same SDK versions. Only after that, we join.

Based on some customer data, we reduced query time from 3000ms to 60ms. However, this will vary based on the number of instances the customer has.


gardleopard (approved):
LGTM

PR #7230, opened 2024-05-31
chore: upgrade to yarn v4

chriswk:
Outdated as of 2020

This has been on our TODO list for a long time, I guess using v6 as a linchpin to finally do it isn't too bad. This is an initial attempt just following official step-by-step on yarn's homepage.


website/package.json
```diff
@@ -70,5 +74,6 @@
     "react-router": "6.23.1",
     "replace-in-file": "7.2.0",
     "typescript": "4.8.4"
-  }
+  },
+  "packageManager": "yarn@4.2.2"
```
alvinometric, line 78: Do we want to do `@4` here in case there's a vulnerability and a patch release?

chriswk: It's a fair question, but the regex validator here requires an absolute version. Though, you're allowed to do ^4.2.2 I believe.

.github/workflows/gradual-strict-null-checks.yml
```diff
@@ -50,9 +50,9 @@ jobs:
           YARN_1="yarn --mutex network --cwd ./current"
           YARN_2="yarn --mutex network --cwd ./main"
 
-          $YARN_1 install --ignore-scripts &> /dev/null && $YARN_1 build:backend --strictNullChecks true 2> .stderr-current > .out-current &
+          $YARN_1 install  &> /dev/null && $YARN_1 build:backend --strictNullChecks true 2> .stderr-current > .out-current &
```
gastonfournier, line 53: I think we can ignore this file as it's not working as intended... although we could test if running it with yarn 4 fixes it :thinking:

.yarn/install-state.gz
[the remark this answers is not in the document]
chriswk: These are now expected yes. I was first afraid it was basically the same as node_modules, but it's not, it's sub 1MiB at the moment, and helps with verifying package data.

.github/workflows/dependency-review.yml
```diff
@@ -17,5 +17,5 @@ jobs:
         uses: actions/dependency-review-action@v4
         with:
           fail-on-severity: moderate
-          allow-licenses: Apache-2.0, MIT, BSD-2-Clause, BSD-3-Clause, ISC, CC0-1.0, Unlicense
+          allow-licenses: Apache-2.0, MIT, BSD-2-Clause, BSD-3-Clause, ISC, CC0-1.0, Unlicense, Null, BlueOak-1.0.0
```
nunogois, line 20: Seems like adding `Null` here is making the action fail. Maybe an alternative solution is to add [jackspeak](https://github.com/isaacs/jackspeak) to [allow-dependencies-licenses](https://github.com/actions/dependency-review-action?tab=readme-ov-file#configuration-options).

chriswk: hmm, good catch, it seemed like Null was a license ID, but looking at the license list, that's a non-existing identifier. Will update.

chriswk: Done. Thanks for pointing it out.

chriswk:
I see the dependency review still failing. this is due to the spdx parser not handling packages with dual licensing (`Apache 2.0 OR MIT`, `APACHE 2.0 AND MIT`) even if we allow both licenses. There are PRs happening which will fix this, so hopefully it gets merged soon https://github.com/actions/dependency-review-action/pull/719

nunogois (approved):
Great job with this, I vote we try it out 👍

PR #7231, opened 2024-05-31
fix: Created by on application-created adds the ip as created by

gastonfournier:
## About the changes
This leaks potential pii information in the event log. Instead of that we're using the system user


chriswk (approved):
LGTM

PR #7232, opened 2024-05-31
Prevent jumping content navbar switch

kwasniew:
## About the changes

When a project list is very long and we scroll to the middle of the screen the main content is not jumping anymore.
Previously the scroll position was affected by the navbar size which was changing on mini/full modechange. This PR opts the navigation out of the scroll position setting with https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor

## Discussion points

Another solution that I found was flex row direction set to row-reverse and re-ordering DOM items. I didn't like it due to re-ordered tab navigation and screen reader ordering.


frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationSidebar.tsx
```diff
@@ -45,8 +45,9 @@ export const StretchContainer = styled(Box)(({ theme }) => ({
     alignSelf: 'stretch',
     display: 'flex',
     flexDirection: 'column',
-    gap: theme.spacing(3),
+    gap: theme.spacing(2),
     zIndex: 1,
+    overflowAnchor: 'none',
```
kwasniew, line 50: this is the solution

thomasheartman (approved):
Nice one! I agree with your assessment on not reordering dom items. Good shout!

PR #7241, opened 2024-06-03
chore: clear yarn.lock

gastonfournier:
rm yarn.lock && yarn install


sjaanus (approved):
LG, lets see if it builds

PR #7242, opened 2024-06-03
chore: change deny list to allow list

gastonfournier:
All licenses we allow


sighphyre (approved):
Allow list seems safer so :white_check_mark:  

Is this really the entire list of licenses?

PR #7249, opened 2024-06-03
feat: project health chart now goes from 0 to 100 to give perspective

sjaanus:
image


frontend/src/component/insights/componentsChart/ProjectHealthChart/ProjectHealthChart.tsx
```diff
@@ -113,6 +113,41 @@ export const ProjectHealthChart: VFC<IProjectHealthChartProps> = ({
                     ? {}
                     : {
                           parsing: { yAxisKey: 'health', xAxisKey: 'date' },
+                          scales: {
+                              y: {
+                                  min: 0,
```
sjaanus, line 118: This is main change, but to avoid other scales default options from being overridden, I need to pull it out. Other options is to create 2 new props for min and max, but that does not scale nicely.

daveleek (approved):
Nice!

PR #7250, opened 2024-06-03
feat: adds information about project modes to the project creation form

thomasheartman:
This change adds information about the project modes to the new
project creation form, using the tooltip for project creation modes.

In doing so, it updates the config button tooltip to accept extra
elements and adds styling for them.

What it looks like: 

image

## a11y issues

This solution does present one problem: the popover doesn't get focus, so it's impossible for you to scroll with only a keyboard. However, this is something that's present in Unleash already, and not something that I think would be easily solvable, so I don't think this is when we should solve it.


daveleek (approved):
Great! Love the explanation improvements!

PR #7253, opened 2024-06-03
chore: readded resolutions for our docker package file

chriswk:
Since this isn't inherited from our mother package, we needed to readd resolutions and overrides. I've just copied what we used to have here for now.


gastonfournier (approved):
LG!

ivarconr (approved):

PR #7265, opened 2024-06-04
feat: upgrade from react v17 to v18

sjaanus:
**Upgrade to React v18 for Unleash v6. Here's why I think it's a good time to do it:**
- Command Bar project: We've begun work on the command bar project, and there's a fantastic library we want to use. However, it requires React v18 support.
- Straightforward Upgrade: I took a look at the upgrade guide https://react.dev/blog/2022/03/08/react-18-upgrade-guide and it seems fairly straightforward. In fact, I was able to get React v18 running with minimal changes in just 10 minutes!
- Dropping IE Support: React v18 no longer supports Internet Explorer (IE), which is no longer supported by Microsoft as of June 15, 2022. Upgrading to v18 in v6 would be a good way to align with this change.

TS updates:
* FC children has to be explicit: https://stackoverflow.com/questions/71788254/react-18-typescript-children-fc
* forcing version 18 types in resolutions: https://sentry.io/answers/type-is-not-assignable-to-type-reactnode/

Test updates:
* fixing SWR issue that we have always had but it manifests more in new React (https://github.com/vercel/swr/issues/2373)


alvinometric:
I'm going to open up the discussion: Should we consider React 19? 

On one hand, we're doing a big upgrade anyway, but on the other, I'm not sure if it will break dependencies

sjaanus:
> I'm going to open up the discussion: Should we consider React 19?
> 
> On one hand, we're doing a big upgrade anyway, but on the other, I'm not sure if it will break dependencies

I think it is still beta. We usually do not use beta versions.

alvinometric:
Looking great 🙌

PR #7266, opened 2024-06-04
Docs: Merged 11 principles articles into a single document

alvinometric:
This consolidates the "11 Principles for building and scaling feature flag systems" article into a single document, that you can navigate using the TOC on the right side, rather than the sidebar on the left hand side.

This also updates the redirects to that the old links still work.


website/docusaurus.config.js
```diff
@@ -781,6 +781,22 @@ module.exports = {
...
+                            '/topics/feature-flags/scale-horizontally',
+                            '/topics/feature-flags/limit-payloads',
+                            '/topics/feature-flags/availability-over-consistency',
+                            '/topics/feature-flags/short-lived-feature-flags',
+                            '/topics/feature-flags/unique-names',
+                            '/topics/feature-flags/democratize-feature-flag-access',
+                            '/topics/feature-flags/prioritize-ux',
+                            '/topics/feature-flags/enable-traceability',
+                        ],
+                        to: '/topics/feature-flags/feature-flag-best-practices',
```
ivarconr, line 798: Keep in mind that these redirects are purely client side, which generally is bad for SEO. We have established server side redirects here: https://github.com/Unleash/unleash/blob/main/website/vercel.json. 

(On that note we are getting quite a few redirects in now, we need to review them over time as they add complexity and there might be limits at Vercel that we are not aware of, its not uncommon to have an upper limits on these rules to keep performance under control.)

ivarconr:
> This consolidates the "11 Principles for building and scaling feature flag systems" article into a single document, that you can navigate using the TOC on the right side, rather than the sidebar on the left hand side.

Can you elaborate a bit on the purpose of this change? Is it purely UX optimization, or are we trying to achieve some specific benefits here?

website/docusaurus.config.js
alvinometric: I'll bring this up with @riboflavin so we can run an audit on our existing redirects to see if they get much traffic. Then we can start to trim down the list

alvinometric:
It's both for SEO and UX 😄

PR #7267, opened 2024-06-04
feat: more powerful feature search by type

kwasniew:
## About the changes

Previous search by type was an array and now we can use operators. Also the UI didn't use previous filter by type and now we can re-use one of the filters.


sjaanus (approved):
LG

PR #7268, opened 2024-06-04
fix: disallow invalid tag values

thomasheartman:
This PR fixes how Unleash handles tag values. Specifically, it does these things:
1. Trims leading and trailing whitespace from tag values before inserting them into the database
2. Updates OpenAPI validation to not allow whitespace-only and to ignore leading and trailing whitespace

Additionally, it moves the tag length constants into the constants file from the Joi tag schema file. This is because importing the values previously rendered them as undefined (probably due to a circular dependency somewhere in the system). This means that the previous values were also ignored by OpenAPI.

UI updates reflecting this wil follow.

## Background
When you tag a flag, there's nothing stopping you from using an entirely empty tag or a tag with leading/trailing whitespace.

Empty tags make little sense and leading trailing whitespace differences are incredibly subtle:
image

Additionally, leading and trailing whitespace is not shown in the dropdown list, so you'd have to guess at which is the right one.
image


src/lib/services/tag-service.test.ts
```diff
@@ -0,0 +1,29 @@
+import type { IUnleashConfig } from '../types/option';
+import { createTestConfig } from '../../test/config/test-config';
+import type EventService from '../features/events/event-service';
+import FakeTagStore from '../../test/fixtures/fake-tag-store';
+import TagService from './tag-service';
+
+const config: IUnleashConfig = createTestConfig();
+
+test('should trim tag names before saving them', async () => {
```
kwasniew, line 9: maybe: 'should trim tag values ...'?

src/lib/openapi/spec/tag-schema.ts
```diff
@@ -11,16 +11,16 @@ export const tagSchema = {
     properties: {
         value: {
             type: 'string',
-            minLength: TAG_MIN_LENGTH,
-            maxLength: TAG_MAX_LENGTH,
-            description: 'The value of the tag',
+            pattern: `^\\s*\\S.{${TAG_MIN_LENGTH - 2},${
```
kwasniew, line 14: I'm assuming it has no impact on the orval type

thomasheartman: No, it shouldn't I don't think Orval cares about regexes because they're hard to encode in typescript. The description has changed, though, so we should also regenerate Orval for that.

PR #7269, opened 2024-06-04
fix: trying to create a tag that's too short gives errors

thomasheartman:
1. Only suggest to create a tag value if the input is more than two
characters after trimming.
2. Ignore trailing and leading whitespace when considering which autocomplete options to show

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/ManageTagsDialog/TagsInput.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/ManageTagsDialog/TagsInput.tsx
@@ -108,15 +108,16 @@ export const TagsInput = ({
         options: TagOption[],
         params: FilterOptionsState<TagOption>,
     ) => {
+        const inputValue = params.inputValue.trim();
+
         const filtered = filter(options, {
             ...params,
-            inputValue: params.inputValue.trim(),
+            inputValue,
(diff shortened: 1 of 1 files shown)


chriswk (approved):
LGTM

frontend/src/component/feature/FeatureView/FeatureOverview/ManageTagsDialog/TagsInput.tsx
```diff
@@ -108,14 +108,19 @@ export const TagsInput = ({
...
+        });
 
-        const { inputValue } = params;
         // Suggest the creation of a new value
         const isExisting = options.some(
             (option) => inputValue === option.title,
         );
-        if (inputValue !== '' && !isExisting) {
+
+        if (inputValue.length >= 2 && !isExisting) {
```
daveleek, line 123: We're checking the filtered/trimmed value for existing here, we're certain that the trimmed/filtered value is what ends up getting created afterwards right?

thomasheartman: Ye. We handle that on the back end here: https://github.com/Unleash/unleash/pull/7268

But regardless, if you select an existing option from the dropdown, you get the existing value anyway.

PR #7272, opened 2024-06-04
task: upgraded semver dependency (and biome)

chriswk:
Sorry for the extra noise here, but this seems to be the biome upgrade altering formatting slightly.


gastonfournier (approved):
G!

PR #7287, opened 2024-06-05
fix: involuntarily scrolled to the top when mousing off group/project avatars

thomasheartman:
This PR fixes a bug where if you navigated to the projects page via the menu, scrolled down, and hovered over a project's avatars, you'd be scrolled to the top of the page when you moused off the avatar.

Turns out this issue was also in the group cards. It seems to be that the popover attempts to restore focus back to where you where, which, if you navigated via the menu, is at the top of the page. Because these popovers don't have any focusable content, we can disable that functionality.

Additionally, I've disabled the scroll lock when the popover is open. The scroll lock made it impossible to scroll when one of the popovers is open, which is confusing as a user.


kwasniew (approved):
nice finding

PR #7288, opened 2024-06-05
fix: project settings table overflow

kwasniew:
## About the changes

Stop project settings tables from overflowing. Some tables (e.g. API tokens) in the project settings had px based min width that is overflowing after we added a sidebar. I'm adjusting the breakpoint for mobile view a bit earlier. We could also consider not using px based min width and use % based width instead.

Before (on 1500px, type, created and other columns push the content out of the parent container):
Screenshot 2024-06-05 at 13 47 25

After (on 1500px):
Screenshot 2024-06-05 at 13 48 08


sjaanus (approved):
LG

PR #7289, opened 2024-06-05
chore: make feature.spec and segements.spec more resilient

thomasheartman:
This PR attempts to make the feature.spec and segements.spec test suites more reliable. They have been flaking out a lot recently, and this will hopefully make them less flaky.

The way of handling it is a little different for each test suite.

## feature.spec

Some of the failures we're seeing for the feature/feauture.spec test suite are due to uncaught resize observer issues (possibly triggered by the banners). 

We can ignore these errors as they don't impact functionality, only rendering, and are likely to resolve themselves quickly in real-world scenarios.

On the other hand, it might also ignore actual errors, so I'm not a 100% on this. Would love some input.

However, MDN has some info on [observation errors](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors):

> As long as the error event does not fire indefinitely, resize observer will settle and produce a stable, likely correct, layout. However, visitors may see a flash of broken layout, as a sequence of changes expected to happen in a single frame is instead happening over multiple frames.

Based on that, I think this is a pretty safe error to ignore. 

I'm unsure whether catching this exception is only set in the `after` cleanup or whether it pollutes the cy object for all tests, but I think it's fine either way. But if you have ideas, I'd love to hear them.

## segments.spec

The issue here appears to be that when we first input the segment's name in the form, it takes a little time for the UI to become ready, so the first characters of the string are cut off.  

This is a known [issue that the cypress team are aware](https://github.com/cypress-io/cypress/issues/3817), but that isn't likely to get fixed any time soon because no one can give them a reproducible example.

You can see the effect of this on segments that haven't been cleaned up in the preview:
image

To work around it, we add a 500ms wait before we start filling out the form. Yes, adding [waits in your tests is an antipattern](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting), but it's the easiest way around in this case.

We *could* investigate and find a way not to need that, but that would likely be a much larger project. This appears to mitigate the issue immediately, so is at least a pretty good temporary fix in my opinion. We also already do this in other tests, so there is a precedent for it.

### DIFF
--- a/.github/workflows/e2e.frontend.yaml
+++ b/.github/workflows/e2e.frontend.yaml
@@ -2,7 +2,7 @@ name: e2e:frontend
 on: [deployment_status]
 jobs:
   e2e:
-    if: true
+    if: github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success' && contains(github.event.deployment_status.target_url, 'unleash-monorepo-frontend')
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/e2e.frontend.yaml
+++ b/.github/workflows/e2e.frontend.yaml
@@ -2,7 +2,7 @@ name: e2e:frontend
 on: [deployment_status]
 jobs:
   e2e:
-    if: github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success' && contains(github.event.deployment_status.target_url, 'unleash-monorepo-frontend')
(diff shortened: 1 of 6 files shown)


chriswk (approved):
If this allows me to not have to repeatedly run the action to make it go green. I'm all for it.

PR #7292, opened 2024-06-05
feat: display created by user in search

kwasniew:
## About the changes

Displaying columns with created by user:
Screenshot 2024-06-05 at 17 40 54

Currently we show avatar and user name + id on hover.

Next PR:
* extract and reuse on hover component from the project overview
* add filter by user id backend (new filter in search and list of all users who created features in this project) and UI component

## Discussion points

Should we put it behind a dedicated flag? It's a new table column that's visible by default on large screens.

### DIFF
--- a/frontend/src/openapi/models/featureSearchResponseSchema.ts
+++ b/frontend/src/openapi/models/featureSearchResponseSchema.ts
@@ -28,7 +28,7 @@ export interface FeatureSearchResponseSchema {
      */
     createdAt: string | null;
     /** User who created the feature flag */
-    createdBy?: FeatureSearchResponseSchemaCreatedBy;
+    createdBy: FeatureSearchResponseSchemaCreatedBy;
     /**
      * The type of dependency. 'parent' means that the feature is a parent feature, 'child' means that the feature is a child feature.
      * @nullable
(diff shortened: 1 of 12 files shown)


ivarconr:
Have we considered how it looks for old feature flags created before we supported created_by_user_id? Will it then just say system user?

ivarconr:
> Should we put it behind a dedicated flag? It's a new table column that's visible by default on large screens.

Whats the argument to not flag this?

frontend/src/openapi/models/featureSearchResponseSchema.ts
```diff
@@ -28,7 +28,7 @@ export interface FeatureSearchResponseSchema {
      */
     createdAt: string | null;
     /** User who created the feature flag */
-    createdBy?: FeatureSearchResponseSchemaCreatedBy;
+    createdBy: FeatureSearchResponseSchemaCreatedBy;
```
thomasheartman, line 31: will this always have a created by even in old flags? We did do some sort of patching of pre-existing flags, didn't we?

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -404,6 +405,11 @@ class FeatureSearchStore implements IFeatureSearchStore {
 
             if (!entry) {
                 // Create a new entry
+                const name =
+                    row.user_name ||
+                    row.user_username ||
+                    row.user_email ||
+                    'unknown';
```
thomasheartman, line 412: Is the unknown case where we don't know who created the flag initially? Like Ivar asked, what do we show then?

thomasheartman (approved):
Looks fine to me, but I've got the same questions Ivar has. And like him, I'd lean towards flagging this initially. Probably not a huge deal, but gives us the option to turn it off if something goes wrong (reversible decisions 😉 )

src/lib/features/feature-search/feature-search-store.ts
kwasniew: We show this
Screenshot 2024-06-06 at 09 42 21

PR #7302, opened 2024-06-06
feat: project flag creators api

kwasniew:
## About the changes

Adding a new endpoint that returns all users who created a flag in a given project. It includes those users who are deactivated. This API will be used for feature search based on flag creator (including those who left the project).
As a display name we use name, username with a fallback to email

### DIFF
--- a/src/lib/features/project/project-controller.ts
+++ b/src/lib/features/project/project-controller.ts
@@ -353,7 +353,7 @@ export default class ProjectController extends Controller {
     }
 
     async getProjectFlagCreators(
-        req: IAuthRequest,
+        req: IAuthRequest<IProjectParam>,
         res: Response<ProjectFlagCreatorsSchema>,
     ): Promise<void> {
         const { projectId } = req.params;
--- a/src/lib/features/project/project-controller.ts
+++ b/src/lib/features/project/project-controller.ts
@@ -44,6 +44,10 @@ import { normalizeQueryParams } from '../feature-search/search-utils';
 import ProjectInsightsController from '../project-insights/project-insights-controller';
 import FeatureLifecycleController from '../feature-lifecycle/feature-lifecycle-controller';
 import type ClientInstanceService from '../metrics/instance/instance-service';
+import {
+    projectFlagCreatorsSchema,
+    type ProjectFlagCreatorsSchema,
+} from '../../openapi/spec/project-flag-creators-schema';
 
 export default class ProjectController extends Controller {
     private projectService: ProjectService;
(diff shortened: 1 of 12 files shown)


src/lib/features/project/project-controller.ts
```diff
@@ -327,6 +351,24 @@ export default class ProjectController extends Controller {
             serializeDates(applications),
         );
     }
+
+    async getProjectFlagCreators(
+        req: IAuthRequest,
+        res: Response<ProjectFlagCreatorsSchema>,
+    ): Promise<void> {
+        const { projectId } = req.params;
```
thomasheartman, line 359: Small suggestion, but can we type this in the request body? Right now it's implicit, right? Should be specified both here and in the route description earlier in the file.

src/lib/features/project/project-flag-creators-read-model.ts
```diff
@@ -0,0 +1,31 @@
...
+            .where('features.project', project)
+            .select([
+                'users.id',
+                'users.name',
+                'users.username',
+                'users.email',
+            ]);
+        return result.map((row) => ({
+            id: Number(row.id),
+            name: String(row.name || row.username || row.email),
```
thomasheartman, line 28: Don't we have a standardized function to extract this from users? Do we handle it differently here than in other places?

src/lib/openapi/spec/project-flag-creators-schema.ts
```diff
@@ -0,0 +1,32 @@
...
+        additionalProperties: false,
+        required: ['id', 'name'],
+        properties: {
+            id: {
+                type: 'integer',
+                example: 50,
+                description: 'The user id.',
+            },
+            name: {
+                description: 'Name of the user',
```
thomasheartman, line 18: Can we mention how we use fallbacks here if there is no real name?

```suggestion
                description: 'Name of the user. If the user has no set name, the API falls back to using the user's username (if they have one) or email (if neither name or username is set).',
```

thomasheartman (approved):
Looks sensible! Got a couple of questions and suggestions, but nothing hyper urgent.

src/lib/features/project/project-flag-creators.e2e.test.ts
```diff
@@ -0,0 +1,61 @@
...
+});
+
+test('should return flag creators', async () => {
+    await app.request
+        .post(`/auth/demo/login`)
+        .send({
+            email: 'user1@getunleash.io',
+        })
+        .expect(200);
+    await app.createFeature('toggle-name-1');
```
sjaanus, line 43: We should stop using toggle in code.

src/lib/features/project/project-flag-creators-read-model.ts
kwasniew: I think we do it in a few places. I'm not sure how to generalize it since query can give those fields different names when you use aliases.

PR #7303, opened 2024-06-06
fix: trim sso URL fields (#7301)

chriswk:
What the title says. There are input values that are whitespace sensitive, so this will trim clientId and entity field, preventing the form from sending leading or trailing whitespace. Will make a PR on enterprise as well to trim on the backend as well.


gastonfournier (approved):
:+1:

PR #7304, opened 2024-06-06
chore: test that the tags API still returns tags that you can't create anymore

thomasheartman:
This change adds a test to the tags API to ensure that even if you
can't create tags that are pure whitespace anymore, you'll still
receive pre-existing tags from the API that fit this description.

The test is here to ensure that we don't break this in future versions
of Unleash.

### DIFF
--- a/src/test/e2e/api/admin/tags.e2e.test.ts
+++ b/src/test/e2e/api/admin/tags.e2e.test.ts
@@ -209,3 +209,10 @@ test('Can bulk remove tags', async () => {
         })
         .expect(200);
 });
+
+describe('backward compatibility', () => {
+    test('should return invalid tag names if they exist', async () => {
+        await db.stores.tagStore.createTag({ value: '  ', type: 'simple' });
+        await app.request.get('/api/admin/tags').expect(200);
+    });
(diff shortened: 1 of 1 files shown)


src/test/e2e/api/admin/tags.e2e.test.ts
```diff
@@ -209,3 +209,12 @@ test('Can bulk remove tags', async () => {
         })
         .expect(200);
 });
+
+describe('backward compatibility', () => {
```
kwasniew, line 213: I'd follow one test convention in a file. If we have test(...) multiple times I wouldn't do describe with a test inside.

kwasniew (approved):

src/test/e2e/api/admin/tags.e2e.test.ts
thomasheartman: Sure, I'm happy to unwrap it. It's just the one for now, so it's not a big deal. But I think using describe to group tests related to a single area is useful, and I don't necessarily think that having both root-levels test and describe blocks is bad 🤷🏼

PR #7312, opened 2024-06-06
Remove axios in favour of native in API docs

alvinometric:
Massive thanks to @thomasheartman for digging around GitHub and figuring out how to do this.

Regrettably it doesn't use the native fetch, which is what I thought it would do. It's more code so I'm happy to close it but I still think it's worth the discussion

Before this PR:
image

After this PR:
image

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -194,13 +194,13 @@ module.exports = {
                 variant: 'requests',
             },
             {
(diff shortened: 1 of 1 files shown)


thomasheartman (approved):
Nice! Yeah, it's more code, but it removes the axios dependency, so I'm in favor of it 😄

PR #7319, opened 2024-06-07
feat: add hook for collecting recently visited pages

daveleek:
## About the changes

Adds a hook that captures and makes available a list of 5 most recent visited pages/features/projects

### DIFF
--- a/frontend/src/hooks/useRecentlyVisited.ts
+++ b/frontend/src/hooks/useRecentlyVisited.ts
@@ -1,0 +1,126 @@
+import { useCallback, useEffect, useState } from 'react';
+import { useLocation, useMatch } from 'react-router-dom';
+import { routes } from 'component/menu/routes';
+import { getLocalStorageItem, setLocalStorageItem } from '../utils/storage';
+import { basePath } from 'utils/formatPath';
+import { useCustomEvent } from './useCustomEvent';
+
+const RECENTLY_VISITED_PAGES_UPDATED_EVENT = 'recentlyVisitedPagesUpdated';
+const MAX_ITEMS = 5;
+
+export type LastViewedPage = {
+    featureId?: string;
+    projectId?: string;
+    pathName?: string;
+};
+
+const removeIncorrect = (flags?: any[]): LastViewedPage[] => {
+    if (!Array.isArray(flags)) return [];
+    return flags.filter(
+        (flag) =>
+            (flag.featureId && flag.projectId) ||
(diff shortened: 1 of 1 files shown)


frontend/src/hooks/useRecentlyVisited.ts
```diff
@@ -0,0 +1,126 @@
...
+    const location = useLocation();
+
+    useEffect(() => {
+        if (!location.pathname) return;
+
+        const path = routes.find((r) => r.path === location.pathname);
+        if (path) {
+            setCappedLastVisited([{ pathName: path.path }]);
+        } else {
+            if (featureMatch?.params.featureId) {
```
sjaanus, line 56: We can use just else if

frontend/src/hooks/useRecentlyVisited.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    useEffect(() => {
+        if (lastVisited) {
+            setLocalStorageItem(key, lastVisited);
+            emitEvent();
+        }
+    }, [JSON.stringify(lastVisited), key, emitEvent]);
+
+    const setCappedLastVisited = useCallback(
+        (pages: LastViewedPage[]) => {
```
sjaanus, line 79: This is probably single page.

frontend/src/hooks/useRecentlyVisited.ts
```diff
@@ -0,0 +1,105 @@
...
+            )
+                return;
+            if (
+                lastVisited.find(
+                    (item) => item.pathName && item.pathName === page.pathName,
+                )
+            )
+                return;
+            if (
+                lastVisited.find(
```
sjaanus, line 81: Can we extract a const, so it describes what this actually does?

frontend/src/hooks/useRecentlyVisited.ts
```diff
@@ -0,0 +1,105 @@
...
+        (page: LastViewedPage) => {
+            if (page.featureId && !page.projectId) return;
+            if (
+                lastVisited.find(
+                    (item) =>
+                        item.featureId && item.featureId === page.featureId,
+                )
+            )
+                return;
+            if (
```
sjaanus, line 74: I wonder how difficult would it to bring it back to top1? Instead of ignoring.

daveleek:
Merged as part of Command bar PoC [PR](https://github.com/Unleash/unleash/pull/7350)

PR #7325, opened 2024-06-07
Ignore files in `.yarn` folder

alvinometric:
As per [the official docs](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored), this adds files in the the `.yarn` folder to gitignore.


chriswk:
We should probably remove the already added files inside the .yarn folder, to avoid keeping outdated information.

chriswk (changes requested):
Could you delete the files that are already added first?

alvinometric:
#7333 @thomasheartman beat me to the merge

PR #7338, opened 2024-06-10
docs: move provisioning reference to scim reference

sighphyre:
Moves the docs at refence/provisioning to reference/scim because there's already a reference to /scim in our source code.

### DIFF
--- a/website/sidebars.js
+++ b/website/sidebars.js
@@ -243,7 +243,7 @@ module.exports = {
                         'reference/public-signup',
                         'reference/projects',
                         'reference/project-collaboration-mode',
-                        'reference/provisioning',
+                        'reference/scim',
                         'reference/rbac',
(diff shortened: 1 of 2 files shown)


daveleek (approved):
LGTM!

website/sidebars.js
```diff
@@ -243,7 +243,7 @@ module.exports = {
                         'reference/public-signup',
                         'reference/projects',
                         'reference/project-collaboration-mode',
-                        'reference/provisioning',
+                        'reference/scim',
```
daveleek, line 246: would we need to add a redirect if someone has the old url?

sighphyre: Don't see how they would have the old URL, so I don't think so, no. I did this, rather than patch the code so that all existing references would point to working docs

PR #7344, opened 2024-06-11
feat: add popover to users in flags list

thomasheartman:
This PR adds a popover to the user avatars in the flag list.

The popover is similar to the one used for projects and groups, but it differs in a few ways:
- There's less padding. There's quite a lot of padding in the other popovers, and it felt like too much for this table.
- It only shows one bit of text (the user's name/username/email). The other popovers show email and name/username, but we don't have all that information, so this is a stripped down version.

Flag list popover:
image

Group popover:
image

or if no email
image

### DIFF
--- a/frontend/src/component/common/UserAvatar/UserAvatar.tsx
+++ b/frontend/src/component/common/UserAvatar/UserAvatar.tsx
@@ -4,10 +4,9 @@ import {
     styled,
     type SxProps,
     type Theme,
-    Popover,
 } from '@mui/material';
 import type { IUser } from 'interfaces/user';
-import { useState, type FC } from 'react';
+import type { FC } from 'react';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 
 const StyledAvatar = styled(Avatar, {
@@ -78,7 +77,7 @@ export const UserAvatar: FC<IUserAvatarProps> = ({
             sx={sx}
             {...props}
             data-loading
-            alt='Gravatar'
+            alt={user?.name || user?.email || user?.username || 'Gravatar'}
             src={src}
             title={title}
             onMouseEnter={onMouseEnter}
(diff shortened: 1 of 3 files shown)


frontend/src/component/common/UserAvatar/UserAvatar.tsx
```diff
@@ -77,7 +77,7 @@ export const UserAvatar: FC<IUserAvatarProps> = ({
             sx={sx}
             {...props}
             data-loading
-            alt='Gravatar'
+            alt={user?.name || user?.email || user?.username || 'Gravatar'}
```
thomasheartman, line 80: The previous alt wasn't very descriptive at all, so we'll add the user's name if it's present.

frontend/src/component/common/UserAvatar/UserAvatarWithPopover.tsx
thomasheartman, line 1: This file nabs and modifies the implementation in `frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx` to better suit what we're doing here.

In theory, we *could* probably use the same element in the new group card avatars component too, but I've not done that because the implementation is slightly different. In the group avatars component, all avatars *share* the same popover. Switching to having a popover for each avatar might have performance implications. However, I'd be happy to take a stab at it and see. Additionally, there's the minor styling differences.

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -176,13 +177,15 @@ export const ProjectFeatureToggles = ({
...
+                                  <UserAvatarWithPopover
                                       user={{
                                           id: original.createdBy.id,
                                           name: original.createdBy.name,
                                           imageUrl: original.createdBy.imageUrl,
                                       }}
-                                      avatarWidth={(theme) => theme.spacing(3)}
+                                      avatarWidth={(theme: Theme) =>
+                                          theme.spacing(3)
+                                      }
```
thomasheartman, line 188: My linter was complaining about the "any" type here (probably because of the extra layer of abstraction for the props), so I added it explicitly.

frontend/src/component/common/UserAvatar/UserAvatar.tsx
kwasniew: I wish we could do it only in the backend so that the lookup logic doesn't leak out to frontend. I already collapse all of those to name for the created by field in search results. But I assume other places still expose the whole user that you need to dissect.

kwasniew (approved):

frontend/src/component/common/UserAvatar/UserAvatar.tsx
thomasheartman: Yeah, so this is in the existing user avatar component, which already uses this kind of messy logic everywhere. I don't really want to go look at all the places we use that component and what kind of data we might have gotten from the back end there. So this is just updating that component to show a slightly better alt text based on whatever data you feed it.

So ... yeah.

PR #7347, opened 2024-06-11
wip

thomasheartman:


frontend/src/component/common/SrOnly/SrOnly.tsx
thomasheartman, line 1: I've been doing stuff like this a few places lately, so figured it'd be time to add a specific component that does it for you instead of having to copy it in from the ground up every time. I'll go back and update some uses in a followup PR.

frontend/src/component/common/UserAvatar/UserAvatarWithPopover.tsx
thomasheartman, line 1: This component isn't necessary anymore, so we can remove it.

frontend/src/component/common/SrOnly/SrOnly.tsx
```diff
@@ -0,0 +1,14 @@
+import { styled } from '@mui/material';
+
+// Visually hide content, but make it available to screen readers
+export const SrOnly = styled('div')(() => ({
```
kwasniew, line 4: Can we use full name ScreenReaderOnly?

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/useProjectFeatureSearch.ts
```diff
@@ -76,19 +77,22 @@ export const useProjectFeatureSearchActions = (
 ) => {
     const onAttributeClick = (attribute: Attribute, value: string) => {
         const attributeState = tableState[attribute.key];
+        console.log('using this attribute state', attributeState);
```
kwasniew, line 80: leftover

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/useProjectFeatureSearch.ts
```diff
@@ -76,19 +77,22 @@ export const useProjectFeatureSearchActions = (
...
 
         if (
             attributeState &&
             attributeState.values.length > 0 &&
             !attributeState.values.includes(value)
         ) {
+            console.log('adding to attribute state', attribute, value);
```
kwasniew, line 87: leftover

frontend/src/component/common/UserAvatar/UserAvatar.tsx
```diff
@@ -47,9 +48,10 @@ export const UserAvatar: FC<IUserAvatarProps> = ({
     className,
     sx,
     children,
+    hideTitle,
     ...props
 }) => {
-    if (!title && !onMouseEnter && user) {
+    if (!hideTitle && !title && !onMouseEnter && user) {
```
kwasniew, line 54: do we need both hideTitle and title? Can't we not pass the title to hide it?

kwasniew (approved):
Please check my comments. Minor things mostly. Very cool capability.

frontend/src/component/common/SrOnly/SrOnly.tsx
thomasheartman: Sure can!

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/useProjectFeatureSearch.ts
thomasheartman: Good catch

frontend/src/component/common/UserAvatar/UserAvatar.tsx
thomasheartman: Not the way it is today, no. If you don't pass a title and don't pass a mouseover function, it'll create a title for you. I *could* pass a no-op mouseover function from the avatar cell, but that feels like you have to know too much about the internals to make the component work as you intended.

thomasheartman: That said, I'm happy for you to dispute that and take it from there.

Also, I don't want to modify how we do titles from before, because it might cause issues elsewhere.

But yeah, that component *is* a bit messy, and this doesn't really make it better 💁🏼

kwasniew: Don't want to cause distractions to other work in this PR. Generally when I find myself adding boolean props to my components it means composition is needed (boolean prop is configuration which doesn't scale very well). To achieve composition over configuration I normally pass the actual component I want to compose with and the API surface is ReactNode instead of a boolean flag.

PR #7349, opened 2024-06-11
docs: Updated and formatted license docs.

alvinometric:
The license keys are no longer in beta as of v6.

### DIFF
--- a/website/docs/using-unleash/deploy/license-keys.mdx
+++ b/website/docs/using-unleash/deploy/license-keys.mdx
@@ -4,8 +4,8 @@ title: License Keys for Unleash Enterprise
 
 import VideoContent from "@site/src/components/VideoContent.jsx";
 
-This document describes all aspects and workflows of license key management in Unleash 5.8 and above.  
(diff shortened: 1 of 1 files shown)


website/docs/using-unleash/deploy/license-keys.mdx
```diff
@@ -1,39 +1,41 @@
 ---
-title: License Keys for Unleash Enterprise (BETA)
+title: License Keys for Unleash Enterprise
 ---
 
-import VideoContent from '@site/src/components/VideoContent.jsx'
+import VideoContent from "@site/src/components/VideoContent.jsx";
 
 This document describes all aspects and workflows of license key management in Unleash 5.8 and above.
```
sebastian-bury, line 7: I think we should mention that this is available from 5.8 and required for self hosted enterprise after v6

sebastian-bury (changes requested):

website/docs/using-unleash/deploy/license-keys.mdx
alvinometric: Good call!

sebastian-bury (approved):
LGTM!

PR #7368, opened 2024-06-12
feat: clean up command bar props

sjaanus:
A lot of things were passed in for search, but since we only have single command bar, we do not need them and it was bloating our component.


daveleek (approved):
Great, thank you for cleaning up! <3

PR #7369, opened 2024-06-12
chore: disable filtering for unknown users

thomasheartman:
This PR disables the filtering capability in the front end for unknown users.

Modifying the back end to support filtering for unknown users is not something we want to do yet. It's possible, but it requires adding a lot of special cases to the handling code (refer to [PR #7359](https://github.com/Unleash/unleash/pull/7359)), which we'd like to avoid if possible. To avoid annoying cases where the filtering doesn't work as expected and breaks user expectations, we're disabling the filtering capability for unknown users in the front end.

We can consider whether to enable back-end results for unknown users in the future if we get user feedback that it's important.

This PR works by changing the avatar cell component. When the user has id 0 (and is therefore unknown), we:
- set aria-disabled to true. This alerts users with assistive tech that the button is disabled, but it doesn't take it out of the tab order, so it's not mysteriously missing.
- change the tooltip text, telling users that they can't filter by unknown users.
- disable the avatar callback function, so clicking on the avatar doesn't do anything.

The accompanying tests assert this functionality.

I considered also updating the screen reader text, but I think that would add more confusion or be more information than the user needs. According to MDN's article on the [aria-disabled attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled):

> [the aria-disabled] declaration will inform people using assistive technologies, such as screen readers, that such elements are not meant to be editable or otherwise operable.

### DIFF
--- a/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/AvatarCell.test.tsx
+++ b/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/AvatarCell.test.tsx
@@ -18,25 +18,25 @@ test("the filtering button should have aria-disabled='false' when the user id is
 });
 
 test('the onAvatarClick function should not be called when the user id is 0', async () => {
-    const callback = vi.fn();
-    const Cell = AvatarCell(callback);
+    const onAvatarClick = vi.fn();
+    const Cell = AvatarCell(onAvatarClick);
     render(<Cell row={{ original: { createdBy: { id: 0, name: '' } } }} />);
     const button = await screen.findByRole('button');
 
     fireEvent.click(button);
 
-    expect(callback).not.toHaveBeenCalled();
+    expect(onAvatarClick).not.toHaveBeenCalled();
 });
 
 test('the onAvatarClick function should be called when the user id is not 0', async () => {
-    const callback = vi.fn();
-    const Cell = AvatarCell(callback);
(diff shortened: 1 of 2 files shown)


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/AvatarCell.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+test("the filtering button should have aria-disabled='false' when the user id is not 0", async () => {
+    const Cell = AvatarCell(() => {});
+    render(<Cell row={{ original: { createdBy: { id: 1, name: '' } } }} />);
+    const button = await screen.findByRole('button');
+    expect(button).toHaveAttribute('aria-disabled', 'false');
+});
+
+test('the onAvatarClick function should not be called when the user id is 0', async () => {
+    const callBack = vi.fn();
```
kwasniew, line 21: weird camel casing. callback is more common

kwasniew (approved):

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/AvatarCell.test.tsx
thomasheartman: Yeah, you're right. Don't know why I did that. Will fix 😅

PR #7371, opened 2024-06-12
fix: yarn v4 requires prepack instead of prepare script when building…

chriswk:
What the title says. After the migration to yarn v4, needed a prepack script/action instead of a prepare step.


gastonfournier (approved):
LG!

PR #7372, opened 2024-06-12
feat: extract global feature search

sjaanus:
We need global search for command menu, so extracting into separate hook.


kwasniew (approved):

daveleek (approved):
Cool! This also highlights the question around `⌘K` - should we change that in the command bar search to be a different key? should we keep as is but override it for search bars when they are there?

PR #7373, opened 2024-06-12
Revert "fix: yarn v4 requires prepack instead of prepare script when building…"

gastonfournier:
Reverts Unleash/unleash#7371


chriswk (approved):
Sad approve

PR #7375, opened 2024-06-12
feat: recent project by name

kwasniew:
## About the changes

Previously we had recent projects displayed by id. How we do it by name.
Screenshot 2024-06-12 at 13 25 25

Also checking if recent project was deleted (current limitation is that it only works after refresh)

### DIFF
--- a/frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
+++ b/frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
@@ -22,6 +22,7 @@ import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import FlagIcon from '@mui/icons-material/OutlinedFlag';
 import { useUiFlag } from 'hooks/useUiFlag';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
+import useProjectOverview from 'hooks/api/getters/useProjectOverview/useProjectOverview';
 
 const StyledBadgeContainer = styled('div')(({ theme }) => ({
     paddingLeft: theme.spacing(2),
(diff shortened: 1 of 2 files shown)


Tymek (approved):

frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
```diff
@@ -245,6 +247,10 @@ export const RecentProjectsNavigation: FC<{
     projectId: string;
     onClick: () => void;
 }> = ({ mode, onClick, projectId }) => {
+    const { project, loading } = useProjectOverview(projectId);
+    const projectDeleted = !project.name && !loading;
+
+    if (projectDeleted) return null;
     return (
```
daveleek, line 254: Is there a specific benefit to doing this here rather than where `RecentProjectsNavigation` is referenced? (ie [here](https://github.com/Unleash/unleash/blob/main/frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationSidebar.tsx#L62))

daveleek (approved):
LGTM!

frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
kwasniew: yes, originally I tried to put it one component up the hierarchy. However this component should be conditional. Only when you have projectId should you call useProjectOverview. So by putting useProjectOverview here we make this hook conditional and the condition is one layer up.

daveleek: Sounds good! 👍 :shipit:

PR #7378, opened 2024-06-12
feat: search features from command bar

sjaanus:
Now searching works in command bar

1. Currently piggybacking on the search hook, but I think it is not fast enough, and also it is using the query params as the global search. This causes some weird behaviour in UI. This probably means we will create separate endpoint for this.

image

### DIFF
--- a/frontend/src/component/commandBar/CommandBar.tsx
+++ b/frontend/src/component/commandBar/CommandBar.tsx
@@ -16,6 +16,11 @@ import { useOnClickOutside } from 'hooks/useOnClickOutside';
 import { useOnBlur } from 'hooks/useOnBlur';
 import { RecentlyVisited } from './RecentlyVisited/RecentlyVisited';
 import { useRecentlyVisited } from 'hooks/useRecentlyVisited';
+import { useGlobalFeatureSearch } from '../feature/FeatureToggleList/useGlobalFeatureSearch';
+import {
+    CommandResultGroup,
+    type CommandResultGroupItem,
+} from './RecentlyVisited/CommandResultGroup';
 
 export const CommandResultsPaper = styled(Paper)(({ theme }) => ({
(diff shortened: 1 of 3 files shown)


frontend/src/component/commandBar/RecentlyVisited/CommandResultGroup.tsx
```diff
@@ -0,0 +1,77 @@
...
+    link: string;
+}
+
+interface CommandResultGroupProps {
+    icon: string;
+    groupName: string;
+    items: CommandResultGroupItem[];
+}
+
+export const CommandResultGroup = ({
```
sjaanus, line 46: This can be reused also for projects and we need to see if we can use it also for pages.

frontend/src/component/commandBar/CommandBar.tsx
```diff
@@ -113,6 +119,13 @@ export const CommandBar = () => {
     useOnClickOutside([searchContainerRef], hideSuggestions);
     useOnBlur(searchContainerRef, hideSuggestions);
 
+    const { features, setTableState } = useGlobalFeatureSearch(3);
+
+    const flags: CommandResultGroupItem[] = features.map((feature) => ({
+        name: feature.name,
+        link: `/projects/${feature.project}/features/${feature.name}`,
```
daveleek, line 126: I think there's a route formatter somewhere we might use for this, it's on my mental todo. But it's unlikely to change I suppose

frontend/src/component/commandBar/RecentlyVisited/CommandResultGroup.tsx
daveleek: Yeah I think we can use this for recently visited as well

daveleek (approved):
awesome!

PR #7384, opened 2024-06-13
fix: playground parent deps check

kwasniew:
## About the changes

Sync parent deps check code with Node SDK

### DIFF
--- a/src/lib/features/playground/feature-evaluator/client.ts
+++ b/src/lib/features/playground/feature-evaluator/client.ts
@@ -77,13 +77,10 @@ export default class UnleashClient {
             }
 
             if (parent.enabled !== false) {
-                if (!parentToggle.enabled) {
-                    return false;
-                }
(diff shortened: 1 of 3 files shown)


src/lib/features/playground/feature-evaluator/client.ts
```diff
@@ -77,26 +77,20 @@ export default class UnleashClient {
...
+                    return featureEnabled && parent.variants.includes(name);
                 }
                 return (
                     this.isEnabled(parent.feature, context, () => false)
                         .result === true
                 );
             }
 
-            return (
-                !parentToggle.enabled &&
```
kwasniew, line 95: this was a bug

thomasheartman (approved):
Nice! Are there tests for this or some other way to verify that this is the correct behavior or that it won't drift in the future?

ivarconr (approved):

PR #7385, opened 2024-06-13
feat: optimize search store by removing inline EXISTS

sjaanus:
Instead of running exists on every row, we are joining the exists, which runs the query only once.
This decreased load time on my huge dataset from 2000ms to 200ms.

Also added tests that values still come through as expected.

### DIFF
--- a/src/lib/features/feature-search/feature-search-store.ts
+++ b/src/lib/features/feature-search/feature-search-store.ts
@@ -248,7 +248,7 @@ class FeatureSearchStore implements IFeatureSearchStore {
                         'features.created_by_user_id',
                     );
 
-                this.findFeatureStrategies(query);
+                this.applyStrategiesByEnvironment(query);
 
                 query.leftJoin('last_seen_at_metrics', function () {
                     this.on(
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -245,6 +248,8 @@ class FeatureSearchStore implements IFeatureSearchStore {
                         'features.created_by_user_id',
                     );
 
+                this.findFeatureStrategies(query);
```
daveleek, line 251: This method returns after applying additional query operations on the query, but the return appears to be discarded? Are they saved in the query object passed into the method or should we actually be doing a `query = this.findFeatureStrategies(query);` here?

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -351,6 +356,48 @@ class FeatureSearchStore implements IFeatureSearchStore {
         };
     }
 
+    private findFeatureStrategies(queryBuilder: Knex.QueryBuilder) {
```
daveleek, line 359: Does this need to actually return something or should it be renamed to something like `applyFindFeatureStrategies` to signal that the outcome of calling the method is "stored" in the QueryBuilder argument passed in?

src/lib/features/feature-search/feature-search-store.ts
sjaanus: Yes, removed return. It will be saved.

src/lib/features/feature-search/feature-search-store.ts
sjaanus: Yep, renamed

PR #7388, opened 2024-06-13
feat: command bar search projects

sjaanus:
Now can search for projects.
Also adding debounce to not spam backend with requests. Also the UI is less flickery.

### DIFF
--- a/frontend/src/component/commandBar/CommandBar.tsx
+++ b/frontend/src/component/commandBar/CommandBar.tsx
@@ -22,7 +22,7 @@ import {
     type CommandResultGroupItem,
 } from './RecentlyVisited/CommandResultGroup';
 import { useAsyncDebounce } from 'react-table';
-import useProjects from '../../hooks/api/getters/useProjects/useProjects';
+import useProjects from 'hooks/api/getters/useProjects/useProjects';
 
 export const CommandResultsPaper = styled(Paper)(({ theme }) => ({
     position: 'absolute',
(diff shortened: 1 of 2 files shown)


frontend/src/component/commandBar/CommandBar.tsx
```diff
@@ -181,6 +202,11 @@ export const CommandBar = () => {
                             icon={'flag'}
                             items={flags}
                         />
+                        <CommandResultGroup
+                            groupName={'Projects'}
+                            icon={'flag'}
```
daveleek, line 207: https://github.com/Unleash/unleash/blob/main/frontend/src/component/commandBar/RecentlyVisited/RecentlyVisited.tsx#L53

Here's a way to get the project icon (cellphone coverage bars)

frontend/src/component/commandBar/RecentlyVisited/CommandResultGroup.tsx
```diff
@@ -49,6 +49,10 @@ export const CommandResultGroup = ({
     items,
 }: CommandResultGroupProps) => {
     const slicedItems = items.slice(0, 3);
+
+    if (items.length === 0) {
+        return;
```
daveleek, line 54: Might not be an issue, but don't we usually return null in cases like this?

daveleek (approved):
Awesome! Left a tip on how to get the project icon but non-blocking

PR #7392, opened 2024-06-13
feat: Max strategies metrics

kwasniew:
## About the changes

Collecting metrics for max number of strategies in a feature (across all feature environments) and in a single feature environment with the largest number of strategies.

Design decisions:
* created read model for this data so that we're not coupling ourselves to write operations on the metrics side
* not exposing it to instance stats as other metrics since it seems to specific for the instance stats we send in customer reports
* instead metrics are querying the read model on top of querying the instance stats


src/lib/features/feature-toggle/fake-feature-strategies-read-model.ts
```diff
@@ -0,0 +1,20 @@
+import type { IFeatureStrategiesReadModel } from './types/feature-strategies-read-model-type';
+
+export class FakeFeatureStrategiesReadModel
+    implements IFeatureStrategiesReadModel
```
thomasheartman, line 4: Just wondering, would it be better to call this a `FeatureStrategiesMetrictsReadModel` or someting? Not a hard suggestion, just thought it would be clearer. This read model could be mistaken for something that returns flag strats 💁🏼

thomasheartman (approved):

PR #7394, opened 2024-06-13
feat: optimize search store by removing inline EXISTS

sjaanus:
Instead of running exists on every row, we are joining the exists, which runs the query only once.
This decreased load time on my huge dataset from 2000ms to 200ms.

Also added tests that values still come through as expected.

### DIFF
--- a/src/lib/features/feature-search/feature-search-store.ts
+++ b/src/lib/features/feature-search/feature-search-store.ts
@@ -337,7 +337,9 @@ class FeatureSearchStore implements IFeatureSearchStore {
             total: 0,
         };
     }
-
+    /*
+        This is noncritical data that can should be joined after paging and is not part of filtering/sorting
+     */
     private queryExtraData(queryBuilder: Knex.QueryBuilder) {
         this.queryMetrics(queryBuilder);
         this.queryStrategiesByEnvironment(queryBuilder);
--- a/src/lib/features/feature-search/feature-search-store.ts
+++ b/src/lib/features/feature-search/feature-search-store.ts
@@ -86,10 +86,7 @@ class FeatureSearchStore implements IFeatureSearchStore {
             .distinctOn('stage_feature')
             .orderBy([
                 'stage_feature',
-                {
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -161,12 +164,6 @@ class FeatureSearchStore implements IFeatureSearchStore {
 
                 selectColumns = [
                     ...selectColumns,
-                    this.db.raw(
```
sjaanus, line 164: This is the part we removing

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -351,6 +338,72 @@ class FeatureSearchStore implements IFeatureSearchStore {
...
+        queryBuilder.leftJoin('metrics', (qb) => {
+            qb.on(
+                'metric_environment',
+                '=',
+                'ranked_features.environment',
+            ).andOn('metric_feature_name', '=', 'ranked_features.feature_name');
+        });
+    }
+
+    private queryStrategiesByEnvironment(queryBuilder: Knex.QueryBuilder) {
```
sjaanus, line 356: This is what I am adding

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -351,6 +338,72 @@ class FeatureSearchStore implements IFeatureSearchStore {
         };
     }
 
+    private queryExtraData(queryBuilder: Knex.QueryBuilder) {
+        this.queryMetrics(queryBuilder);
+        this.queryStrategiesByEnvironment(queryBuilder);
+    }
+
+    private queryMetrics(queryBuilder: Knex.QueryBuilder) {
```
sjaanus, line 346: Also very small refactor, moving this also out of main join.

sjaanus: Reducing the big method by 20 lines.

kwasniew (approved):

PR #7395, opened 2024-06-14
fix: lifecycle button permissions

kwasniew:
## About the changes

We were missing projectId on 3 lifecycle buttons: mark completed, revert and archive. 
Screenshot 2024-06-14 at 09 44 30

Also fixed styling for locked buttons. Previously they were growing and moving text in another line. Now we move the second button to the next line.


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/FeatureLifecycleTooltip.tsx
```diff
@@ -319,6 +325,7 @@ const SafeToArchive: FC<{
                 sx={{
                     display: 'flex',
                     flexDirection: 'row',
+                    flexWrap: 'wrap',
```
kwasniew, line 328: when the padlock is added the buttons should not grow in size but go to the next line

sjaanus (approved):

PR #7397, opened 2024-06-14
feat: command bar pages and name resolving

daveleek:
## About the changes

Resolves names and icons etc for pages, features, and projects in recently visited
Adds a Pages suggestion section to command bar (that filters out non-OSS routes for OSS)

Skjermbilde 2024-06-14 kl  10 32 07


frontend/src/component/commandBar/RecentlyVisited/RecentlyVisited.tsx
```diff
@@ -37,55 +39,127 @@ const StyledListItemText = styled(ListItemText)(({ theme }) => ({
...
+            </StyledListItemText>
+        </ListItemButton>
+    );
+};
+
+const RecentlyVisitedProjectButton = ({
+    projectId,
+    key,
+}: { projectId: string; key: string }) => {
+    const { project, loading } = useProjectOverview(projectId);
```
sjaanus, line 126: Hmm downside of this is that every time they open the menu, we will make 1-5 concurrent calls to backend. But for now it is fine.

sjaanus (approved):

frontend/src/component/commandBar/RecentlyVisited/RecentlyVisited.tsx
daveleek: Yup, unfortunately

PR #7398, opened 2024-06-14
chore: add metrics/gauges for "max constraint values" and "max constraints"

thomasheartman:
This PR adds metrics tracking for:
- "maxConstraintValues": the highest number of constraint values that are in use
- "maxConstraintsPerStrategy": the highest number of constraints used on a strategy

It updates the existing feature strategy read model that returns max metrics for other strategy-related things.

It also moves one test into a more fitting describe block.

### DIFF
--- a/src/lib/features/feature-toggle/feature-strategies-read-model.ts
+++ b/src/lib/features/feature-toggle/feature-strategies-read-model.ts
@@ -54,7 +54,7 @@ export class FeatureStrategiesReadModel implements IFeatureStrategiesReadModel {
     } | null> {
         throw new Error('Method not implemented.');
     }
-    getMaxConstraintPerStrategy(): Promise<{
+    getMaxConstraintsPerStrategy(): Promise<{
         feature: string;
         environment: string;
         count: number;
--- a/src/lib/features/feature-toggle/feature-strategies-read-model.ts
+++ b/src/lib/features/feature-toggle/feature-strategies-read-model.ts
@@ -57,10 +57,14 @@ export class FeatureStrategiesReadModel implements IFeatureStrategiesReadModel {
                 'feature_name',
                 'environment',
                 this.db.raw(
-                    "MAX(coalesce(jsonb_array_length(value->'values'), 0)) as max_values_count",
+                    "MAX(coalesce(jsonb_array_length(constraint_value->'values'), 0)) as max_values_count",
                 ),
             )
-            .joinRaw('JOIN jsonb_array_elements(constraints) as value ON true')
+            .from(
+                this.db.raw(
+                    'feature_strategies, jsonb_array_elements(constraints) AS constraint_value',
+                ),
+            )
             .groupBy('feature_name', 'environment')
             .orderBy('max_values_count', 'desc')
             .limit(1);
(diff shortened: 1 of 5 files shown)


src/lib/features/feature-toggle/tests/feature-toggle-strategies-store.e2e.test.ts
```diff
@@ -246,6 +247,9 @@ describe('strategy parameters default to sane defaults', () => {
         });
         expect(strategy.parameters.stickiness).toBe(defaultStickiness);
     });
+});
+
+describe('max metrics collection', () => {
     test('Read feature with max number of strategies', async () => {
```
thomasheartman, line 253: This test was previously in the describe block called "strategy parameters default to sane defaults". This seems like a better fit.

src/lib/features/feature-toggle/feature-strategies-read-model.ts
```diff
@@ -47,4 +46,60 @@ export class FeatureStrategiesReadModel implements IFeatureStrategiesReadModel {
...
+        feature: string;
+        environment: string;
+        count: number;
+    } | null> {
+        const rows = await this.db('feature_strategies')
+            .select(
+                'feature_name',
+                'environment',
+                'constraints',
+                this.db.raw(
```
kwasniew, line 60: where do you use the constraint_count?

src/lib/features/feature-toggle/feature-strategies-read-model.ts
```diff
@@ -47,4 +46,60 @@ export class FeatureStrategiesReadModel implements IFeatureStrategiesReadModel {
...
+                    'MAX(jsonb_array_length(constraints)) as constraint_count',
+                ),
+                this.db.raw(
+                    "MAX(coalesce(jsonb_array_length(value->'values'), 0)) as max_values_count",
+                ),
+            )
+            .joinRaw('JOIN jsonb_array_elements(constraints) as value ON true')
```
kwasniew, line 67: can we do it without join?
```
SELECT 
    feature_name, 
    environment, 
    constraints, 
    MAX(COALESCE(jsonb_array_length(value->'values'), 0)) AS max_values_count
FROM 
    feature_strategies,
    jsonb_array_elements(constraints) AS value
GROUP BY 
    feature_name, 
    environment, 
    constraints
ORDER BY 
    max_values_count DESC
LIMIT 2;
```

src/lib/features/feature-toggle/feature-strategies-read-model.ts
thomasheartman: Good catch: addressed in 02921837a0a83863507960d4e311296377f9b329

kwasniew (approved):
Looks very nice. I'd only think about query simplification

src/lib/features/feature-toggle/feature-strategies-read-model.ts
thomasheartman: Probably can! I'll look into it on monday morning 😄

thomasheartman: Yes we can! 3a39d55114f7bb1f4fbedd7f60d8ede147877503

thomasheartman: 3a39d55114f7bb1f4fbedd7f60d8ede147877503

PR #7399, opened 2024-06-14
feat: menubar is not dependant on query params anymroe

sjaanus:
Previously since query params were changing by global search, and menubar was also altering them, they were conflicting. Menubar does not need query params as state. So using search hook directly.


daveleek (approved):
Nice!

PR #7400, opened 2024-06-14
fix: project to lifecycle events

sjaanus:
We need project for those events to filter it out in webhook configuration.


src/lib/types/events.ts
```diff
@@ -722,29 +722,35 @@ export class FeatureTagImport extends BaseEvent {
...
 
     constructor(p: {
+        project: string;
         featureName: string;
         data: FeatureLifecycleCompletedSchema;
         auditUser: IAuditUser;
     }) {
         super(FEATURE_COMPLETED, p.auditUser);
-        const { featureName, data } = p;
+        const { featureName, data, project } = p;
```
kwasniew, line 734: this p looks weird

kwasniew (approved):

PR #7401, opened 2024-06-14
fix: make default for gradual rollout work on the correct strategy name

sighphyre:
This moves the defaults parameters check for adding strategies to target the correct name for gradual rollout strategies. Previously we checked for `gradualRollout` but actually on the backend/SDKs the strategy name is `flexibleRollout`.

Tests have been updated but tested by checking against the API call the UI suggests and cURLs.

### DIFF
--- a/src/lib/features/client-feature-toggles/tests/client-feature-toggles.e2e.test.ts
+++ b/src/lib/features/client-feature-toggles/tests/client-feature-toggles.e2e.test.ts
@@ -25,7 +25,11 @@ const apiClientResponse = [
             {
                 name: 'flexibleRollout',
                 constraints: [],
-                parameters: { rollout: '100' },
+                parameters: {
+                    rollout: '100',
+                    stickiness: 'default',
+                    groupId: 'test1',
+                },
                 variants: [],
             },
         ],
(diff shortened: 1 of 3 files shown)


src/lib/features/client-feature-toggles/tests/client-feature-toggles.e2e.test.ts
```diff
@@ -169,7 +173,11 @@ test('should return correct data structure from /api/client/features', async ()
         {
             name: 'flexibleRollout',
             constraints: [],
-            parameters: { rollout: '100' },
+            parameters: {
```
sighphyre, line 176: Ironically the test name of ''should return correct data structure from /api/client/features" was wrong but I've updated both the call for the data and the data that we check against. For this test to previously be correct, the data it matched against should have included stickiness and groupId

chriswk (approved):
Seems sensible.

src/lib/features/client-feature-toggles/tests/client-feature-toggles.e2e.test.ts
gastonfournier: Shouldn't the request body remain as before? or maybe empty? Perhaps there's a test for that already

daveleek (approved):
Apparently not so flexible gradual default rollout! LG

nunogois (approved):
LG

src/lib/features/client-feature-toggles/tests/client-feature-toggles.e2e.test.ts
nunogois: Was going to comment that if we don't already have that, we should. Essentially a test that validates that adding a strategy without all the parameters still grabs the default parameters.

PR #7402, opened 2024-06-14
Removed old v3 and v4 docs

alvinometric:
🔥


nunogois (approved):
It's always nice to clean up old stuff!

PR #7404, opened 2024-06-17
chore: adds icons for insights and playground paths to recently visited

daveleek:
## About the changes

Resolving missing icons for insights and playground


daveleek:
Merged in different PR

PR #7407, opened 2024-06-18
docs: replace png with svg for feature flag org methods

thomasheartman:
This is a follow-up to the recent new doc, replacing a png with a svg.


alvinometric (approved):
🙌 Thanks!

PR #7408, opened 2024-06-18
fix: check for permission in group access assignment

nunogois:
Fix project role assignment for users with `ADMIN` permission, even if they don't have the Admin root role. This happens when e.g. users inherit the `ADMIN` permission from a group root role, but are not Admins themselves.


sighphyre (approved):
Nice,  LGTM

PR #7421, opened 2024-06-21
fix: turn off showing usernames and emails in the project cards when the flag is turned on

thomasheartman:
This PR:
- adds a flag to anonymize user emails in the new project cards
- performs the anonymization using the existing `anonymise` function that we have.

It does not anonymize the system user, nor does it anonymize groups. It does, however, leave the gravatar url unchanged, as that is already hashed (but we may want to hide that too). 

This PR also does not affect the user's name or username. Considering the target is the demo instance where the vast majority of users don't have this (and if they do, they've chosen to set it themselves), this seems an appropriate mitigation.

With the flag turned off:
image

With the flag on:
image


FredrikOseberg (approved):
LGTM

PR #7431, opened 2024-06-24
[WIP] Trying to defer kapa script

alvinometric:


ivarconr (approved):

website/docusaurus.config.js
```diff
@@ -876,7 +876,7 @@ module.exports = {
             'data-project-color': '#1A4049',
             'data-project-logo':
                 'https://cdn.getunleash.io/uploads/2022/05/logo.png',
-            async: true,
+            defer: true,
```
ivarconr, line 879: Defer is exactly what we want to avoid the script being executed before the main script is loaded:
https://dev.to/fidalmathew/async-vs-defer-in-javascript-which-is-better-26gm

PR #7432, opened 2024-06-24
feat: Anonimize demo users list flag view

kwasniew:
## About the changes

Allow to anonimize users in 2 views.

Feature view
Screenshot 2024-06-24 at 13 02 45

Project features view
Screenshot 2024-06-24 at 13 02 55

Decisions: 
* piggyback on anonymiseEventLog flag since it's already used for segments and event log. We may consider renaming this variable though in a separate PR
* `anonymiseKeys(data, ['name'])` wouldn't cut it since name is too generic and fn is doing deep search. I'm using `anonymise` fn directly only on the field I know should be anonimized
* updated 2 tests to show that anonimize works
* had to update types in a few places since they were either incorrect or any typed


thomasheartman (approved):
Nice one! 😄 My only comment would be that just turning on the "anonymiseEventLog" flag for all the tests in the related suites sounds a little overzealous. If you wanted to test something that needed the original email for a user, you'd be stuck. Would it make sense to instead create a couple tests that check that anonymization happens when the flag is on (but not otherwise)?

kwasniew:
So when we actually have this problem I'd like to tackle it then. Currently we have one test that verifies if the createdBy user is added. Since anonimization builds on top of it I wanted to include as much code as possible on the critical path so that we don't have to add more e2e tests that are slow. At some point we may want to split those tests that do anonimization but I'm tempted to wait for the evidence from code.

PR #7434, opened 2024-06-24
feat: lifecycle stage count

kwasniew:
## About the changes

Count how many features are in each lifecycle stage both at the instance level and at the project level.

It will be used by prometheus metrics but also has a potential for in-app reporting.

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-read-model.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-read-model.ts
@@ -1,5 +1,9 @@
 import type { Db } from '../../db/db';
-import type { IFeatureLifecycleReadModel } from './feature-lifecycle-read-model-type';
+import type {
+    IFeatureLifecycleReadModel,
+    StageCount,
+    StageCountByProject,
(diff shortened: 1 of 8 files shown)


src/lib/features/feature-lifecycle/feature-lifecycle-read-model.ts
```diff
@@ -17,6 +21,61 @@ export class FeatureLifecycleReadModel implements IFeatureLifecycleReadModel {
...
+                COUNT(*) AS feature_count
+            FROM (
+                SELECT DISTINCT ON (feature)
+                    feature,
+                    stage,
+                    created_at
+                FROM
+                    feature_lifecycles
+                ORDER BY
+                    feature, created_at DESC
```
daveleek, line 37: What role does ordering play here? Does it improve performance in any way?

daveleek: Disregard, reading the rest of the query again I see you're selecting distinct, so you want the most recent status!

daveleek (approved):
Cool!

PR #7440, opened 2024-06-25
fix: add license key notice to upgrade instructions

gastonfournier:
Add missing notice about license keys when upgrading to v6


website/docs/using-unleash/deploy/upgrading-unleash.md
```diff
@@ -9,6 +9,13 @@ Generally, the intention is that `unleash-server` should always provide support
 
 [Unleash v6](https://github.com/Unleash/unleash/issues/4380) was released on June 6th, 2024. **We expect this upgrade to be straightforward**, provided you have followed the previous migration guides on this page. We're removing some features that were deprecated in version 5 and marking some features as deprecated, but these will continue to be supported during version 6. Here's the list of the outstanding changes you need to take into account when migrating from v5 to v6:
 
+### Self-hosted enterprise customers are now required to use a License Key
+
+Before upgrading you should make sure to get a valid license key. More information in [this page](../deploy/license-keys.mdx).
```
daveleek, line 14: Could add a little something about "will display a warning banner"

daveleek (approved):
Suggested perhaps adding something that says it will display a warning banner, but LG!

website/docs/using-unleash/deploy/upgrading-unleash.md
gastonfournier: It's documented in the main page of license keys, the reason why I don't want to add a lot of details here is that these may get outdated, while the other page will always be a valid source of truth.

PR #7441, opened 2024-06-25
feat: stage count by project metric

kwasniew:
## About the changes

Expose prometheus metric for counting how many features are in each of the stages grouped by project.

Details:
* this metric is a gauge
* query behind this metric is feature flagged just in case we find perf problems
* reading data in the metrics.ts from the injected read model. This code is not added to instance stats since they don't need it

Next PRs:
* move lifecycle median duration metrics from instance stats to metrics.ts so that generating instance stats doesn't run unnecessary lifecycle queries
* change duration metric type to gauge since we don't do sampling in median duration metrics

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -285,6 +285,12 @@ export default class MetricsMonitor {
             help: 'Duration of feature lifecycle stages',
         });
 
+        const stageCountByProject = createGauge({
+            name: 'stage_count_by_project',
+            help: 'Count features in a given stage by project id',
+            labelNames: ['stage', 'project_id'],
+        });
+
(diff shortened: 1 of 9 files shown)


src/lib/metrics.ts
```diff
@@ -327,6 +335,16 @@ export default class MetricsMonitor {
                         .observe(stage.duration);
                 });
 
+                stageCountByProject.reset();
+                stageCountByProjectResult.forEach((stageResult) =>
+                    stageCountByProject
+                        .labels({
+                            project_id: stageResult.project,
```
kwasniew, line 342: we could also do it at the instance level but I Decided having per project info is more useful

sjaanus (approved):

PR #7443, opened 2024-06-25
feat: now command bar will not search behind the scene

sjaanus:
Previously it was doing request to search behind to scenes, when noone was using the nav bar. This fixes it.


daveleek (approved):
Wow, I would've expected a variable with that name to behave differently I think but if this works then amazing 👌

PR #7449, opened 2024-06-25
feat: lifecycle stage entered counter

kwasniew:
## About the changes

Lifecycle stage entered is counted in prometheus metrics

Details:
* make lifecycle service emit events onto the event bus so that metrics can pick them up
* lifecycle service is not event emitter itself since now it delegates to the event bus and event bus is used in tests instead
* renamed existing lifecycle metrics to follow similar convention 
* only newly inserted stages cause event emit. If there's conflict because stage already exists we don't do anything

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
@@ -15,17 +15,17 @@ import {
 import type {
     FeatureLifecycleView,
     IFeatureLifecycleStore,
+    NewStage,
 } from './feature-lifecycle-store-type';
-import EventEmitter from 'events';
+import type EventEmitter from 'events';
 import type { Logger } from '../../logger';
 import type EventService from '../events/event-service';
 import type { FeatureLifecycleCompletedSchema } from '../../openapi';
 import type { IClientMetricsEnv } from '../metrics/client-metrics/client-metrics-store-v2-type';
 import groupBy from 'lodash.groupby';
+import { STAGE_ENTERED } from '../../metric-events';
 
-export const STAGE_ENTERED = 'STAGE_ENTERED';
-
-export class FeatureLifecycleService extends EventEmitter {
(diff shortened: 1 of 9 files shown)


src/lib/metrics.ts
```diff
@@ -285,12 +285,18 @@ export default class MetricsMonitor {
             help: 'Duration of feature lifecycle stages',
         });
 
-        const stageCountByProject = createGauge({
-            name: 'stage_count_by_project',
+        const featureLifecycleStageCountByProject = createGauge({
+            name: 'feature_lifecycle_stage_count_by_project',
```
kwasniew, line 289: this will required prometheus query update so I want to do it today because we introduced this metrics only today

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -128,22 +127,26 @@ export class FeatureLifecycleService extends EventEmitter {
...
+        const newlyEnteredStages = await this.featureLifecycleStore.insert(
             features.map((feature) => ({ feature, stage })),
         );
-        features.forEach((feature) =>
-            this.emit(STAGE_ENTERED, { stage, feature }),
-        );
+        this.recordStagesEntered(newlyEnteredStages);
+    }
+
+    private recordStagesEntered(newlyEnteredStages: NewStage[]) {
```
kwasniew, line 146: we use this function for all stage transitions and we emit onto the event bus

src/lib/metrics.ts
```diff
@@ -285,12 +285,18 @@ export default class MetricsMonitor {
...
             help: 'Count features in a given stage by project id',
             labelNames: ['stage', 'project_id'],
         });
 
+        const featureLifecycleStageEnteredCounter = createCounter({
```
kwasniew, line 294: the reason we have a counter in addition to gauges is because we want to track deltas every week

sjaanus (approved):

PR #7450, opened 2024-06-25
feat: search only features when there is search string

sjaanus:
Now the search hook is inside another component, so we do not get searches without search query.
Also we had 2 state variable handing the search query. Removed one of them.


daveleek (approved):
Cool, this is not needed for project and pages search?

PR #7451, opened 2024-06-25
feat: transactional complete/uncomplete feature

kwasniew:
## About the changes

Complete/Uncomplete feature should be transactional since we save lifecycle and event data.

### DIFF
--- a/src/lib/features/feature-lifecycle/createFeatureLifecycle.ts
+++ b/src/lib/features/feature-lifecycle/createFeatureLifecycle.ts
@@ -14,48 +14,41 @@ import { FeatureEnvironmentStore } from '../../db/feature-environment-store';
 import FakeFeatureEnvironmentStore from '../../../test/fixtures/fake-feature-environment-store';
 import EventEmitter from 'events';
 
-export const createFeatureLifecycleService = (
-    db: Db,
-    config: IUnleashConfig,
-) => {
-    const { eventBus, getLogger, flagResolver } = config;
-    const eventStore = new EventStore(db, getLogger);
-    const featureLifecycleStore = new FeatureLifecycleStore(db);
-    const environmentStore = new EnvironmentStore(db, eventBus, getLogger);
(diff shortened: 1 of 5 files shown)


src/lib/services/index.ts
```diff
@@ -362,9 +362,17 @@ export const createServices = (
         config.getLogger,
     );
 
-    const { featureLifecycleService } = db
-        ? createFeatureLifecycleService(db, config)
-        : createFakeFeatureLifecycleService(config);
+    const transactionalFeatureLifecycleService = db
+        ? withTransactional(
+              (db: Db) =>
+                  createFeatureLifecycleService(db, config)
```
kwasniew, line 368: this is slightly different from other code because createFeatureLifecycleService composition root returns more than a service

src/lib/features/feature-lifecycle/createFeatureLifecycle.ts
```diff
@@ -14,48 +14,41 @@ import { FeatureEnvironmentStore } from '../../db/feature-environment-store';
...
-            featureLifecycleStore,
-            environmentStore,
-            featureEnvironmentStore,
-        },
-        {
-            eventService,
-        },
-        config,
-    );
+export const createFeatureLifecycleService =
```
kwasniew, line 17: let's follow the same pattern as other composition roots

src/lib/features/feature-lifecycle/createFeatureLifecycle.ts
```diff
@@ -14,48 +14,41 @@ import { FeatureEnvironmentStore } from '../../db/feature-environment-store';
...
+                environmentStore,
+                featureEnvironmentStore,
+            },
+            {
+                eventService,
+            },
+            config,
+        );
 
-    return {
```
kwasniew, line 52: there's no need to break established pattern and expose all the building blocks other than the service itself

sjaanus (approved):

PR #7456, opened 2024-06-26
chore: change generated project id format to use incrementing numbers instead of hashes

thomasheartman:
This change changes how we generate project IDs. Instead of adding a
12-character hash, we now use an incrementing number when necessary.
We first slugify the project name (as we already do). If no project
with that ID exists, that's what you get. If that ID already exists,
then we add a number to the end of the slugified name, starting at 1.

So if you create a project called "My Project" three times over,
first one will have the ID "my-project", the second one will have the
ID "my-project-1", and the third one will have the ID "my-project-2".

In a list:
- my-project
- my-project-1
- my-project-2

This does not impact any existing projects.

The only potential drawback I can see this having, is that it makes
race conditions more likely. If two people create a project with the
same name at the same time (so that the service gives them the same
ID), the db will error out at one of them. With the random hashes,
that's much less of a concern.

At the same time, though: is that really a problem? If two people
create a project with the same name at the same time, isn't it likely
that they'd prefer it if that was the same project instead of two
different ones?

### DIFF
--- a/src/lib/features/project/project-service.e2e.test.ts
+++ b/src/lib/features/project/project-service.e2e.test.ts
@@ -2650,10 +2650,10 @@ describe('automatic ID generation for create project', () => {
             auditUser,
         );
 
-        expect(project.id).toMatch(/^new-name-/);
+        expect(project.id).toBe('new-name');
     });
 
-    test('two projects with the same name get different ids', async () => {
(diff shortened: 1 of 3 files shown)


ivarconr:
I like this improvement! 👍🏼

FredrikOseberg (approved):
LGTM

PR #7466, opened 2024-06-27
refactor: largest resources queries

kwasniew:
## About the changes

* change to a simpler query without the WITH clause
* using native knex syntax


sjaanus (approved):
LG

PR #7469, opened 2024-06-27
feat: command bar track events

sjaanus:
Start tracking plausible events

1. Log the search keywords that returned 0 results
2. Track all clicks, based on source(search/recents/pages), type etc.

### DIFF
--- a/frontend/src/component/commandBar/CommandBar.tsx
+++ b/frontend/src/component/commandBar/CommandBar.tsx
@@ -25,6 +25,7 @@ import { useRoutes } from 'component/layout/MainLayout/NavigationSidebar/useRout
 import { useAsyncDebounce } from 'react-table';
 import useProjects from 'hooks/api/getters/useProjects/useProjects';
 import { CommandFeatures } from './CommandFeatures';
+import { usePlausibleTracker } from '../../hooks/usePlausibleTracker';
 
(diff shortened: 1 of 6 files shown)


frontend/src/component/commandBar/CommandBar.tsx
```diff
@@ -148,10 +152,25 @@ export const CommandBar = () => {
...
+                    eventType: 'no search results found',
+                    query: query,
+                },
+            });
+        }
     }, 200);
 
     const onSearchChange = (value: string) => {
         debouncedSetSearchState(value);
+        setValue(value);
```
sjaanus, line 173: Brought back the value, because this was actually needed for the UI to be fluid.

frontend/src/component/commandBar/CommandBar.tsx
```diff
@@ -148,10 +152,25 @@ export const CommandBar = () => {
...
+            mappedPages.length === 0 &&
+            searchedFlagCount === 0;
+        if (noResultsFound) {
+            trackEvent('command-bar', {
+                props: {
+                    eventType: 'no search results found',
+                    query: query,
+                },
+            });
+        }
```
daveleek, line 168: For the feedback component (what were you looking for/would you like to request a feature?), this looks like we could reuse(?) but rewritten as a useState?

daveleek (approved):
Awesome!

PR #7470, opened 2024-06-27
feat: introduce large cache for swr

sjaanus:
Previously, clearing the SWR cache cleared all entries. Now you can configure the cache size.

1. This makes the search more fluid. Previously, if you went back and forth on pages, you were always sent to the loading state. 
2. This also solves the issue where the command bar search cleared the cache for all other searches. 
3. Additionally, it addresses the problem where the global search cleared the cache for project search.

### DIFF
--- a/frontend/src/hooks/api/getters/useFeatureSearch/useFeatureSearch.test.tsx
+++ b/frontend/src/hooks/api/getters/useFeatureSearch/useFeatureSearch.test.tsx
@@ -42,26 +42,4 @@ describe('useFeatureSearch', () => {
         await screen.findByText(/Features:/);
         await screen.findByText(/Total:/);
     });
-
-    test('should keep only latest cache entry', async () => {
-        testServerRoute(server, '/api/admin/search/features?project=project1', {
-            features: [{ name: 'Feature1' }],
-            total: 1,
-        });
-        render(<TestComponent params={{ project: 'project1' }} />);
-        await screen.findByText(/Features:/);
-        await screen.findByText(
-            'Cache: api/admin/search/features?project=project1',
-        );
-
-        testServerRoute(server, '/api/admin/search/features?project=project2', {
-            features: [{ name: 'Feature2' }],
-            total: 1,
-        });
-        render(<TestComponent params={{ project: 'project2' }} />);
-        await screen.findByText(/Features:/);
-        await screen.findByText(
-            'Cache: api/admin/search/features?project=project2',
-        );
-    });
 });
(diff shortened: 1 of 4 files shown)


frontend/src/hooks/api/getters/useFeatureSearch/useFeatureSearch.test.tsx
```diff
@@ -42,26 +42,4 @@ describe('useFeatureSearch', () => {
         await screen.findByText(/Features:/);
         await screen.findByText(/Total:/);
     });
-
-    test('should keep only latest cache entry', async () => {
```
kwasniew, line 46: what is this test (hook inside a component usage) replaced with? I only see low level hook tests but usually the most revealing tests for hooks misbehavior are usage in component tests

frontend/src/hooks/useClearSWRCache.ts
```diff
@@ -1,14 +1,38 @@
...
+
+interface Cache<Data = any> {
+    keys(): IterableIterator<string>;
+    get(key: string): State<Data> | undefined;
+    set(key: string, value: State<Data>): void;
+    delete(key: string): void;
+}
+
+export const manageCacheEntries = (
+    cache: Cache,
```
kwasniew, line 11: why do we have generic in Cache while we skip it here and default to any?

frontend/src/hooks/useClearSWRCache.ts
```diff
@@ -1,14 +1,38 @@
...
+import { type State, useSWRConfig } from 'swr';
+
+interface Cache<Data = any> {
+    keys(): IterableIterator<string>;
+    get(key: string): State<Data> | undefined;
+    set(key: string, value: State<Data>): void;
+    delete(key: string): void;
+}
+
+export const manageCacheEntries = (
```
kwasniew, line 10: what are you managing about the cache? manage is a very generic name that is not intention revealing

kwasniew: is it clearCacheEntries?

frontend/src/hooks/api/getters/useFeatureSearch/useFeatureSearch.test.tsx
sjaanus: Pulled this test back, but altered a little.

frontend/src/hooks/useClearSWRCache.ts
sjaanus: Yes

kwasniew (approved):

PR #7472, opened 2024-06-27
fix: prevent project cell overflow on api keys table

Tymek:
image

Instead of listing all projects, that currently stretch the row height, I added a tooltip. This is similar to segments. It will enable virtualization of the table

### DIFF
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
@@ -44,4 +44,34 @@ describe('ProjectsList', () => {
 
         expect(container.textContent).toEqual('*');
     });
+
+    it('should show up to 4 projects', async () => {
+        const { container } = render(
+            <ProjectsList
+                projects={['project1', 'project2', 'project3', 'project4']}
+            />,
+        );
+
+        expect(container.textContent).toContain(
+            'project1, project2, project3, project4',
+        );
+    });
+
+    it('should abbreviate for 5 projects', async () => {
+        const { container } = render(
+            <ProjectsList
+                projects={[
+                    'project1',
+                    'project2',
+                    'project3',
+                    'project4',
+                    'project5',
+                ]}
+            />,
+        );
+
+        expect(container.textContent).toContain(
+            'project1, project2, project3, +2 more',
+        );
+    });
 });
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
@@ -4,19 +4,14 @@ import { ProjectsList } from 'component/admin/apiToken/ProjectsList/ProjectsList
 
 describe('ProjectsList', () => {
     it('should prioritize new "projects" array over deprecated "project"', async () => {
-        render(
+        const { container } = render(
             <ProjectsList
                 project='project'
                 projects={['project1', 'project2']}
             />,
         );
 
-        const links = await screen.findAllByRole('link');
-        expect(links).toHaveLength(2);
-        expect(links[0]).toHaveTextContent('project1');
-        expect(links[1]).toHaveTextContent('project2');
-        expect(links[0]).toHaveAttribute('href', '/projects/project1');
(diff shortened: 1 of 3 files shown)


thomasheartman (commented):
Ah, this feels like a nice improvement to me! I'm happy to approve it (because I think it's better than what we have today), but I don't think it's the optimal solution.

I'd have loved to see a description of the problem and what the solution does in the PR description, though. Now I had to guess and go try it out for myself to see what actually happens. But you can still put that description in the merge commit description, so it's clear what's happened for people who come after.


Anyway, here's what happens when you have too many projects for a single API key at the moment:

image

I think this is what you're trying to solve, right?

**However**, you can still achieve the same thing with only 4 projects:

image

Sure, having 4 projects with 100 character long ids is an edge case, but I'm willing to bet you that most users don't call their projects `project1`, `project2`, etc.

So I think the better solution here would be to cap it to a character length or to a line count. I know the css line-clamp functionality isn't implemented yet, but maybe we could use overflow-ellipsis and add a tooltip to that here instead? 

Or maybe we say "n projects" and you can focus the text to see which projects, like how we do for  dependent features and segments?
image
image

That said, knowing which projects you're talking about *is* more important here.

Oh, and even with a single project, you can get to mulitple lines: 
image


.... soooo I'm not really sure. Is it better?

Tymek:
> Anyway, here's what happens when you have too many projects for a single API key at the moment

Not possible with virtualized table, where all rows need to be of the same height

Thanks for the suggestion of doing it with "used in". I'll do figure something out

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
```diff
@@ -45,33 +34,13 @@ describe('ProjectsList', () => {
         expect(container.textContent).toEqual('*');
     });
 
-    it('should show up to 4 projects', async () => {
+    it('should show number of projects', async () => {
```
thomasheartman, line 37:
```suggestion
    it('should show the number of projects', async () => {
```

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
```diff
@@ -20,75 +20,68 @@ interface IProjectsListProps {
...
+                    </span>
+                </HtmlTooltip>
+            </TextCell>
+        );
+    }
+
+    if (projectsList.length === 1 || project) {
+        const item = project || projectsList[0];
+
+        return <LinkCell to={`/projects/${item}`} title={item} />;
```
thomasheartman, line 83: Does this still render the project's ID if there is only one ID? If so, won't `wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww` still cause problems with the height of the table row?

Tymek: Link cell already has truncation built-in

Tymek:
> I'm also a little concerned about the potential a11y impacts of this change. Previously you could find any link on the page (by tabbing to it or by searching for it). If they're all hidden within tooltips, does that now make them unreachable by keyboard only?

I added "tabIndex" to the element. It will show this tooltip on keyboard tab focus, but unfortunately it is still not possible to tab-enter individual projects. Are you OK with this compromise?

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
thomasheartman: Ah, great! Thanks.

thomasheartman:
> unfortunately it is still not possible to tab-enter individual projects. Are you OK with this compromise?

It's consistent with how we handle other cases like this in the UI, so while I don't like it, I also don't think this is the point where we start rewriting the tooltip component. Yeah, that's fine. 

I'll approve this now, but please put a description of what the PR does in the merge commit. And if you don't mind, please also update the PR description, so that it's easy to understand if anyone stumbles upon this PR later. 

Oh, also: there's a wording suggestion for one of the tests that i'd like you to have a look at and either accept or reject.

thomasheartman (approved):

PR #7473, opened 2024-06-27
feat: strategy limit to 30

kwasniew:
## About the changes

Setting strategy count for feature env (set to 30) in frontend and backend code. 

Hard limit in the UI 
Screenshot 2024-06-27 at 16 15 11

Hard limit in the UI 
Screenshot 2024-06-27 at 16 15 27

This is currently behind a flag

Next PR (since this one is already approaching 200 LOC):
* add configuration for the env strategy limit and share it between frontend and backend


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -358,6 +358,27 @@ class FeatureToggleService {
...
+        featureEnv: {
+            projectId: string;
+            environment: string;
+            featureName: string;
+        },
+        limit: number,
+    ) {
+        if (!this.flagResolver.isEnabled('resourceLimits')) return;
+        const existingCount = (
+            await this.featureStrategiesStore.getStrategiesForFeatureEnv(
```
kwasniew, line 371: Originally I wanted to add a new count query but I decided to use the query for strategies and calculating length. In extreme scenarios the extra data we transfer from DB will be measured in 100s of kBs but this should not be an issue.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
```diff
@@ -131,6 +132,11 @@ const FeatureOverviewEnvironment = ({
     const featureEnvironment = feature?.environments.find(
         (featureEnvironment) => featureEnvironment.name === env.name,
     );
+    const resourceLimitsEnabled = useUiFlag('resourceLimits');
+    const limitReached =
+        resourceLimitsEnabled &&
+        Array.isArray(featureEnvironment?.strategies) &&
+        featureEnvironment?.strategies.length >= 30;
```
ivarconr, line 139: I think it is a terrible idea to hard code these values here. I think at least they should be moved to a central configuration sections sot that we keep controls of UI limits. 

I also find it a bit strange to limit this in the UI but not via the API.

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -624,6 +645,11 @@ class FeatureToggleService {
             strategyConfig.variants = fixedVariants;
         }
 
+        await this.validateStrategyLimit(
+            { featureName, projectId, environment },
+            30,
```
ivarconr, line 650: this is a hard limit. I would prefer the ability to configure the limit, even though we advice against it.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
kwasniew: This PR is still WIP and eventually we'll have one value shared between backend and frontend. I'm currently hardcoding those so that I can finish writing UI tests

src/lib/features/feature-toggle/feature-toggle-service.ts
kwasniew: yup, that's the long term plan. I found IUnleashConfig['resourceLimits'] but want to do it in another step

ivarconr: I find it jumping the gun a bit. I think there is value in going via soft-limits that can be adjusted. We have limited insights into self-hosted unleash instances on on this resource.

kwasniew: Can you elaborate?
My plan for the next PR was to add env var here: https://github.com/Unleash/unleash/blob/main/src/lib/create-config.ts#L625 so that you can adjust it. So the hard limit is configurable.
In addition we gonna have a soft limit when you're approaching the value that will be communicated either via toast or dedicated UI component telling you how close you are to the hard limit.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
```diff
@@ -86,6 +88,10 @@ export const FeatureStrategyMenu = ({
                 variant={variant}
                 size={size}
                 sx={{ minWidth: matchWidth ? '282px' : 'auto' }}
+                disabled={Boolean(disableReason)}
+                tooltipProps={{
+                    title: disableReason ? disableReason : undefined,
+                }}
```
thomasheartman, line 94: I don't think using the disable attribute is right here because it makes the buttons "invisible" to screen readers. As such, unless you can see it and navigate to the button with a mouse, you'll never understand why the button is disabled.

We should instead use the `aria-disabled` attribute, add the correct styling, and set onClick to a no-op function. That way, you don't lose out on anything if you can't mouse over it.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
```diff
@@ -99,8 +105,9 @@ export const FeatureStrategyMenu = ({
                 variant='outlined'
                 size={size}
                 hideLockIcon
+                disabled={Boolean(disableReason)}
                 tooltipProps={{
-                    title: 'More strategies',
+                    title: disableReason ? disableReason : 'More strategies',
```
thomasheartman, line 110: Same thing here regarding accessibility.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
```diff
@@ -179,6 +185,11 @@ const FeatureOverviewEnvironment = ({
                                         environmentId={env.name}
                                         variant='outlined'
                                         size='small'
+                                        disableReason={
+                                            limitReached
+                                                ? 'Limit reached'
```
thomasheartman, line 190: For the reasons, can we be more explicit? Just "limit reached" doesn't tell me what limit or what the limit is. Maybe something like "you can't add more than 30 strategies to this environment" or something? 🤷🏼

src/lib/features/feature-toggle/feature-toggle-service.ts
FredrikOseberg: This looks fine to me. Everything is flagged and we are all aligned that we will not ship anything that can not be configured.

thomasheartman (approved):
I think this is a good start for this feature. As pointed out, there's obviously room for improvement here, but I'm happy for that to be in a later PR as long as it's behind a flag for now.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
kwasniew: I will do this with the config change

PR #7481, opened 2024-06-28
fix: cap project ids to 90 characters (without suffix)

thomasheartman:
This fixes the issue where project names that are 100 characters long
or longer would cause the project creation to fail. This is because
the resulting ID would be longer than the 100 character limit imposed
by the back end.

We solve this by capping the project ID to 90 characters, which leaves
us with 10 characters for the suffix, meaning you can have 1 billion
projects (999,999,999 + 1) that start with the same 90
characters (after slugification) before anything breaks.

It's a little shorter than what it strictly has to be (we could
probably get around with 95 characters), but at this point, you're
reaching into edge case territory anyway, and I'd rather have a little
too much wiggle room here.

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -305,16 +305,16 @@ export default class ProjectService {
         }
     }
     async generateProjectId(name: string): Promise<string> {
-        const generateUniqueId = async (name: string, suffix?: number) => {
-            const slug = createSlug(name);
+        const slug = createSlug(name).slice(0, 90);
+        const generateUniqueId = async (suffix?: number) => {
             const id = suffix ? `${slug}-${suffix}` : slug;
             if (await this.projectStore.hasProject(id)) {
-                return await generateUniqueId(name, (suffix ?? 0) + 1);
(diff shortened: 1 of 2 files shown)


src/lib/features/project/project-service.ts
```diff
@@ -305,16 +305,16 @@ export default class ProjectService {
         }
     }
     async generateProjectId(name: string): Promise<string> {
-        const generateUniqueId = async (name: string, suffix?: number) => {
-            const slug = createSlug(name);
+        const slug = createSlug(name).slice(0, 90);
+        const generateUniqueId = async (suffix?: number) => {
```
thomasheartman, line 309: We don't need to recalculate the slug every time - it'll never change -  so I moved the variable assignment out, which also means we can remove that argument from the inner function.

daveleek (approved):
Interesting, so we don't always set the project id to have a suffix, only when it already exists? LGTM

thomasheartman:
yeah, we changed it recently to use an incrementing counter instead of the 12-character hash. So now the first project for an id will just get the name slugified. Following projects with the same slug will get `-1`, `-2`, `-3`, etc.

PR #7483, opened 2024-06-28
chore: command bar refactor of search result items for consistent styling and icons

daveleek:
Refactors command bar ui so that the styling and icons is consistent among search results, recents, and page suggestions. fixes a couple of snags, removes color from icons


sjaanus (approved):
LG for now..

Ideally we want to have single component for all the same design.

PR #7497, opened 2024-07-01
feat: what's new in Unleash

gastonfournier:
https://linear.app/unleash/issue/2-2354/new-in-unleash-section-in-sidebar

Add a "New in Unleash" section in the side bar and use it to announce signals and actions.

image

Inside signals page we're also including a feedback button to try to collect some insights.

image


nunogois (approved):
LGTM, nice teamwork 🙌

PR #7498, opened 2024-07-01
feat: constraint values limit

kwasniew:
## About the changes

Backend part of constraint values limit checking (250). As any other limit check it's behind a flag. 

Next PRs:
* add frontend validation
* add separate limit for enterprise (1000)

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-service.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-service.ts
@@ -366,15 +366,14 @@ class FeatureToggleService {
         }
     }
 
-    async validateStrategyLimit(
-        featureEnv: {
-            projectId: string;
-            environment: string;
-            featureName: string;
-        },
-        limit: number,
-    ) {
+    async validateStrategyLimit(featureEnv: {
+        projectId: string;
+        environment: string;
+        featureName: string;
+    }) {
         if (!this.flagResolver.isEnabled('resourceLimits')) return;
+
+        const limit = this.resourceLimits.featureEnvironmentStrategies;
         const existingCount = (
             await this.featureStrategiesStore.getStrategiesForFeatureEnv(
                 featureEnv.projectId,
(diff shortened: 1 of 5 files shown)


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -366,15 +366,14 @@ class FeatureToggleService {
...
-        limit: number,
-    ) {
+    async validateStrategyLimit(featureEnv: {
+        projectId: string;
+        environment: string;
+        featureName: string;
+    }) {
         if (!this.flagResolver.isEnabled('resourceLimits')) return;
+
+        const limit = this.resourceLimits.featureEnvironmentStrategies;
```
kwasniew, line 376: refactored so that all limit knowledge is localized in one function

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -387,6 +386,22 @@ class FeatureToggleService {
...
+        const limit = this.resourceLimits.constraintValues;
+        const constraintOverLimit = updatedConstrains.find(
+            (constraint) =>
+                Array.isArray(constraint.values) &&
+                constraint.values?.length > limit,
+        );
+        if (constraintOverLimit) {
+            throw new BadDataError(
+                `Constraint values limit of ${limit} is exceeded for ${constraintOverLimit.contextName}.`,
+            );
```
thomasheartman, line 401: I think we should use the Exceeds Limit error (`src/lib/error/exceeds-limit-error.ts`) for this instead of redoing the same thing ourselves. 

What do you think?

thomasheartman (approved):
Looks good to me! Only note is about the error we're using. If we already have an established limit exceeded error, I think we should stick to that.

src/lib/features/feature-toggle/feature-toggle-service.ts
kwasniew: Oh, didn't know we have it. I will adjust it in another PR. I think we can err on the side of consistency over precise contextual message.

thomasheartman: Yeah, I just discovered it earlier today when I was implementing my own stuff 😅  I still think you can be about as precise in that message, but it'd be something like:
```ts
new ExceedsLimitError(`constraint value for constraint ${contsraintOverLimit.contextName}`, limit)
```

We'll adjust it in that next pr

PR #7499, opened 2024-07-01
API tokens scoped to deleted projects shouldn't give wildcard access

Tymek:
If you have SDK tokens scoped to projects that are deleted, you should not get access to any flags with those.


src/lib/db/api-token-store.ts
```diff
@@ -38,6 +38,15 @@ interface ITokenRow extends ITokenInsert {
 const tokenRowReducer = (acc, tokenRow) => {
     const { project, ...token } = tokenRow;
     if (!acc[tokenRow.secret]) {
+        if (!tokenRow.project) {
+            if (
+                !tokenRow.secret.startsWith('*:') ||
+                tokenRow.secret.startsWith('[]:')
```
daveleek, line 44: This shouldn't be necessary, covered by the first check

daveleek:
We have some test issues here as we're expecting a certain format but haven't been consistent with this is unit tests

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -74,7 +74,7 @@ test('editor users should only get client or frontend tokens', async () => {
         projects: [],
         tokenName: '',
         username: 'test',
-        secret: '1234',
+        secret: '*:1234',
```
sighphyre, line 77: Let's be good citizens here - the token format was previously wrong because it should have been `project:environment.hash`, we've changed it to `project:hash` but let's go all the way to `project:environment.hash` so we're honest about what `secret` should hold

src/test/e2e/api/client/feature.token.deleted.project.e2e.test.ts
```diff
@@ -0,0 +1,153 @@
...
+            name: 'default',
+            constraints: [],
+            parameters: {},
+        },
+        { projectId: project, featureName: feature1, environment: DEFAULT_ENV },
+        TEST_AUDIT_USER,
+    );
+    await featureToggleServiceV2.createStrategy(
+        {
+            name: 'flexibleRollout',
```
sighphyre, line 82:
```suggestion
            name: 'default',
```

Flexible rollout with no parameters is definitely broken. I know it doesn't matter for this PR, but this PR was complex because a test was naughty about a data constraint, so let's just be good to ourselves in future and pick a strategy that's valid without parameters

sighphyre (commented):
Looks promising, let's get the test tokens in the fully correct format this time though so we don't face a similar problem in future

src/lib/db/api-token-store.ts
```diff
@@ -74,8 +86,14 @@ const toRow = (newToken: IApiTokenCreate) => ({
     alias: newToken.alias || null,
 });
 
-const toTokens = (rows: any[]): IApiToken[] => {
-    const tokens = rows.reduce(tokenRowReducer, {});
+const toTokens = (
+    rows: any[],
+    allowOrphanedWildcardTokens?: boolean,
```
sighphyre, line 91: Can we make this non nullable? It's not public so we shouldn't break other things and I really, really want people to choose this behavior intentionally and have the type system scream at them if they don't realize a choice is being made for them

src/lib/types/experimental.ts
```diff
@@ -299,6 +300,10 @@ const flags: IFlags = {
         process.env.UNLEASH_EXPERIMENTAL_RESOURCE_LIMITS,
         false,
     ),
+    tokenWidcardOrphans: parseEnvVarBoolean(
```
sighphyre, line 303: I think `allowOrphanedWildcardTokens` was a really good name. Personally I think you should stick with that. No harm in having a long name I think and the clarity it brings is always nice

src/lib/db/api-token-store.ts
```diff
@@ -35,33 +36,44 @@ interface ITokenRow extends ITokenInsert {
...
-    const currentToken = acc[tokenRow.secret];
-    if (tokenRow.project) {
-        if (isAllProjects(currentToken.projects)) {
-            currentToken.projects = [];
+const createTokenRowReducer =
+    (allowOrphanedWildcardTokens?: boolean) => (acc, tokenRow) => {
+        const { project, ...token } = tokenRow;
+        if (!acc[tokenRow.secret]) {
+            if (
+                !allowOrphanedWildcardTokens &&
```
sighphyre, line 44: Gosh this if statement is hairy. Honestly I'm not sure how to make this clearer so if you got no ideas I won't complain but if you have something clever to improve the readability I won't say no

sighphyre (changes requested):
I think this looks good. I'm only hard requesting changes on the nullable boolean being passed. Other things are up to you

src/lib/db/api-token-store.ts
Tymek: The only improvement I have in mind is not separating createTokenRowReducer into a separate function, so I can keep some variables in scope.

Complexity here can be cleaned up later on, if we decide to remove orphaned keys with migration.

sighphyre: > Complexity here can be cleaned up later on, if we decide to remove orphaned keys with migration.

Yeah that's probably okay then. I'll be quiet, it can stay

PR #7500, opened 2024-07-01
feat: use new environment limit in Unleash UI

thomasheartman:
This PR updates the Unleash UI to use the new environment limit.

As it turns out, we already had an environment limit in the UI, but it was hardcoded (luckily, its value is the same as the new default value 🥳).

In addition to the existing places this limit was used, it also disables the "new environment" button if you've reached the limit. Because this limit already exists, I don't think we need a flag for it. The only change is that you can't click a button (that should be a link!) that takes you to a page you can't do anything on.


src/lib/openapi/spec/resource-limits-schema.ts
```diff
@@ -80,7 +80,7 @@ export const resourceLimitsSchema = {
             type: 'integer',
             minimum: 1,
             example: 50,
-            description: 'The maximum number active environments allowed.',
+            description: 'The maximum number environments allowed.',
```
kwasniew, line 83: Missing preposition. "of environments"

frontend/src/component/environments/CreateEnvironmentButton/CreateEnvironmentButton.tsx
```diff
@@ -3,18 +3,29 @@ import Add from '@mui/icons-material/Add';
...
 import { useNavigate } from 'react-router-dom';
+import { useEnvironments } from 'hooks/api/getters/useEnvironments/useEnvironments';
 
 export const CreateEnvironmentButton = () => {
     const { uiConfig } = useUiConfig();
+    const { environments } = useEnvironments();
+    const environmentLimit = uiConfig.resourceLimits.environments;
     const navigate = useNavigate();
 
+    const atEnvironmentLimit = environments.length >= environmentLimit;
```
kwasniew, line 14: I used limitReached convention in other places. I like both atLimit and limitReached as long as we agree on consistent naming.

frontend/src/component/environments/CreateEnvironmentButton/CreateEnvironmentButton.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+                enabledToggleCount: 0,
+            },
+            {
+                name: 'development',
+                type: 'development',
+                sortOrder: 2,
+                enabled: true,
+                protected: false,
+                projectCount: 49,
+                apiTokenCount: 24353,
```
kwasniew, line 39: are those numbers important? If not I'd keep them identical between the envs and possibly sth like o or 1.

frontend/src/component/environments/CreateEnvironmentButton/CreateEnvironmentButton.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+                enabledToggleCount: 20,
+            },
+        ],
+    });
+};
+
+test('should allow you to create environments when there are fewer environments than the limit', async () => {
+    setupApi(5);
```
kwasniew, line 47: maybe `setupApi({envLimit: 5, currentEnvCount: 2})` to emphasize what values matter and why the rest of the test behaves in a certain way

frontend/src/component/environments/CreateEnvironmentButton/CreateEnvironmentButton.test.tsx
thomasheartman: Good point. No, they're just the numbers that came out on my local machine.

thomasheartman: I think I can just remove all the extra data anyway.

frontend/src/component/environments/CreateEnvironmentButton/CreateEnvironmentButton.test.tsx
thomasheartman: Agreed!

PR #7509, opened 2024-07-02
feat: command bar results key down should exit/refocus input

daveleek:
Refocuses command bar input field if command bar results has focus and user starts to type. Closes if Escape button clicked

### DIFF
--- a/frontend/src/component/commandBar/CommandBar.tsx
+++ b/frontend/src/component/commandBar/CommandBar.tsx
@@ -199,7 +199,7 @@ export const CommandBar = () => {
     const placeholder = `Command bar (${hotkey})`;
 
     useOnClickOutside([searchContainerRef], hideSuggestions);
-    const onKeyDown = (event: any) => {
+    const onKeyDown = (event: React.KeyboardEvent) => {
         if (event.key === 'Escape') {
             setShowSuggestions(false);
         } else if (event.keyCode >= 48 && event.keyCode <= 110) {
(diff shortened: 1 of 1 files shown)


frontend/src/component/commandBar/CommandBar.tsx
```diff
@@ -194,13 +194,18 @@ export const CommandBar = () => {
...
     useKeyboardShortcut({ key: 'Escape' }, () => {
-        if (searchContainerRef.current?.contains(document.activeElement)) {
-            searchInputRef.current?.blur();
-        }
+        setShowSuggestions(false);
     });
     const placeholder = `Command bar (${hotkey})`;
 
     useOnClickOutside([searchContainerRef], hideSuggestions);
+    const onKeyDown = (event: any) => {
```
daveleek, line 202: Couldn't find a good event type for this that let me use event.key and event.keyCode, KeyboardEvent would but isn't compatible with the onKeyDown prop

frontend/src/component/commandBar/CommandBar.tsx
```diff
@@ -194,13 +194,18 @@ export const CommandBar = () => {
...
+        if (event.key === 'Escape') {
+            setShowSuggestions(false);
+        } else if (event.keyCode >= 48 && event.keyCode <= 110) {
```
daveleek, line 205: Anything outside these we should be interested in accepting?

sjaanus: Maybe this event: React.KeyboardEvent<HTMLSpanElement>

sjaanus (approved):

frontend/src/component/commandBar/CommandBar.tsx
daveleek: Cool thanks, I wasn't aware there was a difference between React.KeyboardEvent and KeyboardEvent. That works! I'll push that improvement and merge

PR #7510, opened 2024-07-02
feat: add resource limits for API tokens

thomasheartman:
This PR adds the back end for API token resource limits. 

It adds the limit to the schema and checks the limit in the service.

## Discussion points

The PAT service uses a different service and different store entirely, so I have not included testing any edge cases where PATs are included. However, that could be seen as "knowing too much". We could add tests that check both of the stores in tandem, but I think it's overkill for now.

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -661,6 +661,10 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
             process.env.UNLEASH_ENVIRONMENTS_LIMIT,
             50,
         ),
+        apiTokens: parseEnvVarNumber(
+            process.env.UNLEASH_API_TOKENS_LIMIT,
+            2000,
+        ),
     };
 
     return {
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -663,7 +663,7 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
         ),
         apiTokens: parseEnvVarNumber(
             process.env.UNLEASH_API_TOKENS_LIMIT,
-            options?.resourceLimits?.apiTokens || 2000,
+            2000,
         ),
     };
 
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -663,7 +663,7 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
         ),
         apiTokens: parseEnvVarNumber(
             process.env.UNLEASH_API_TOKENS_LIMIT,
-            2000,
+            options?.resourceLimits?.apiTokens || 2000,
         ),
     };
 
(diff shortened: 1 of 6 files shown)


src/lib/create-config.ts
```diff
@@ -661,6 +661,10 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
             process.env.UNLEASH_ENVIRONMENTS_LIMIT,
             50,
         ),
+        apiTokens: parseEnvVarNumber(
```
kwasniew, line 664: I'd wrap it with `Math.max(1, parse...)` if you want to enforce min 1. But check my next comments about the minimum value to think if we need one.

src/lib/services/api-token-service.limit.test.ts
```diff
@@ -0,0 +1,102 @@
...
+        name: 'production',
+        type: 'production',
+        enabled: true,
+        protected: true,
+        sortOrder: 1,
+    });
+
+    const eventService = createFakeEventsService(config);
+
+    const service = new ApiTokenService(
```
kwasniew, line 31: it looks like you created an implicit `composition root` for the fake ApiTokenService. Since we touch this `not yet ported to the new architecture` code maybe it's worth creating a composition root for a real and fake API Token Service in the next PR?

src/lib/services/api-token-service.ts
```diff
@@ -286,6 +295,14 @@ export class ApiTokenService {
         const environments = await this.environmentStore.getAll();
         validateApiTokenEnvironment(newToken, environments);
 
+        if (this.flagResolver.isEnabled('resourceLimits')) {
```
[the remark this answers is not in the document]
thomasheartman: Yeah, good point; that's an oversight. I'll address this for sure.

src/lib/services/api-token-service.limit.test.ts
thomasheartman: Aye, I could do that. Would you expect it to be in its own feature directory? And if so, would you expect the project-based API token controller and admin-based API token controller to both be there?

src/lib/openapi/spec/resource-limits-schema.ts
```diff
@@ -82,6 +82,13 @@ export const resourceLimitsSchema = {
             example: 50,
             description: 'The maximum number of environments allowed.',
         },
+        apiTokens: {
+            type: 'integer',
+            minimum: 1,
```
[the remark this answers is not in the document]
thomasheartman: Yeah, sure. I have no qualms with setting the minimum to 0.

src/lib/create-config.ts
thomasheartman: Roger.

src/lib/services/api-token-service.limit.test.ts
kwasniew: You can check how we migrated nothing but the composition with createAccessService.ts. This way we don't need to do more than is needed to proceed.

thomasheartman: Addressed in https://github.com/Unleash/unleash/pull/7519

PR #7511, opened 2024-07-02
feat: clear search string and close box after click

sjaanus:
Now when selecting an item from command menu, the search string is cleared and command menu closed.


daveleek (approved):
Great!

PR #7513, opened 2024-07-02
fix: improve menu styling

sjaanus:
Now quick suggestions padding is not bigger than for pages. 

image


frontend/src/component/commandBar/CommandRecent.tsx
```diff
@@ -57,7 +56,7 @@ export const CommandRecent = ({
             groupName='Quick suggestions'
             onClick={onClick}
         >
-            <List>{buttons}</List>
+            {buttons}
```
sjaanus, line 59: This was the issue, buttons were in List here and also in subcomponent. Now it is not double wrapped.

daveleek (approved):
Awesome!

PR #7514, opened 2024-07-02
feat: projects limit

kwasniew:
## About the changes

Limit number of projects in the backend to 500.

Next PR:
* disable add project button when the limit is reached and prevent new projects in the UI

## Discussion points

Enforcing minimum limit of 1 project. This is to prevent configuration mistakes where no projects are allowed.

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -70,7 +70,10 @@ import { calculateAverageTimeToProd } from '../feature-toggle/time-to-production
 import type { IProjectStatsStore } from '../../types/stores/project-stats-store-type';
 import { uniqueByKey } from '../../util/unique';
 import { BadDataError, PermissionError } from '../../error';
-import type { ProjectDoraMetricsSchema } from '../../openapi';
+import type {
+    ProjectDoraMetricsSchema,
+    ResourceLimitsSchema,
+} from '../../openapi';
 import { checkFeatureNamingData } from '../feature-naming-pattern/feature-naming-validation';
 import type { IPrivateProjectChecker } from '../private-project/privateProjectCheckerType';
 import type EventService from '../events/event-service';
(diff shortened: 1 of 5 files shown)


src/lib/openapi/spec/resource-limits-schema.ts
```diff
@@ -82,6 +83,12 @@ export const resourceLimitsSchema = {
             example: 50,
             description: 'The maximum number of environments allowed.',
         },
+        projects: {
+            type: 'integer',
+            minimum: 1,
```
kwasniew, line 88: this is also enforced in the service

src/lib/features/project/project-service.ts
```diff
@@ -304,6 +311,18 @@ export default class ProjectService {
             await this.validateEnvironmentsExist(environments);
         }
     }
+
+    async validateProjectLimit() {
+        if (!this.flagResolver.isEnabled('resourceLimits')) return;
+
+        const limit = Math.max(this.resourceLimits.projects, 1);
```
thomasheartman, line 318: Any particular reason why you wanna put this minimum here and not in the parsing of the value?

src/lib/openapi/spec/resource-limits-schema.ts
thomasheartman: I'm sure you know this, but the limit here really has no bearing on what goes on in Unleash, right? This is just a promise we make to our end users. And in fact, in this case, if you set it to 0, you'll get 0 out. Unleash will say the schema doesn't allow the value, but it won't correct it for you.

Unless I'm missing something?

thomasheartman (approved):
Nice one! Just a few minor comments.

src/lib/features/project/project-service.ts
kwasniew: Will do this in the next PR. I like your idea more.

src/lib/openapi/spec/resource-limits-schema.ts
kwasniew: I wanted to communicate that you'll never get 0 in the GET request for UI config with the resource limits. The parsing (`Math.max(1, parseEnvVarNumber(process.env.UNLEASH_PROJECTS_LIMIT, 500))`) we're doing should provide this guarantee

thomasheartman:
Side note: should we consider setting minimum values in parsing for all the limits? So that you can't set nagitev ilimts etc?

PR #7518, opened 2024-07-02
feat: Project limit UI

kwasniew:
## About the changes

Screenshot 2024-07-02 at 12 32 04

Details:
* block the add project button when project limit exceeded
* since we show the number of projects on the page I don't show this number in the disabled message to keep the component API simpler (we can change it though)
* excluding negative numbers in config (abstraction waiting to be extracted)

### DIFF
--- a/frontend/src/component/project/ProjectList/ProjectList.test.tsx
+++ b/frontend/src/component/project/ProjectList/ProjectList.test.tsx
@@ -1,7 +1,7 @@
 import { render } from 'utils/testRenderer';
 import { ProjectListNew } from './ProjectList';
 import { screen, waitFor } from '@testing-library/react';
-import { testServerRoute, testServerSetup } from '../../../utils/testServer';
+import { testServerRoute, testServerSetup } from 'utils/testServer';
 import { CREATE_PROJECT } from '../../providers/AccessProvider/permissions';
 
 const server = testServerSetup();
--- a/frontend/src/component/project/ProjectList/ProjectList.test.tsx
+++ b/frontend/src/component/project/ProjectList/ProjectList.test.tsx
@@ -1,0 +1,37 @@
+import { render } from 'utils/testRenderer';
+import { ProjectListNew } from './ProjectList';
+import { screen, waitFor } from '@testing-library/react';
+import { testServerRoute, testServerSetup } from '../../../utils/testServer';
+import { CREATE_PROJECT } from '../../providers/AccessProvider/permissions';
+
+const server = testServerSetup();
+
+const setupApi = () => {
+    testServerRoute(server, '/api/admin/ui-config', {
+        flags: {
+            resourceLimits: true,
+        },
+        resourceLimits: {
+            projects: 1,
+        },
+        versionInfo: {
+            current: { enterprise: 'version' },
+        },
+    });
+
+    testServerRoute(server, '/api/admin/projects', {
+        projects: [],
+    });
+};
+
+test('Enabled new project button when limits, version and permission allow for it', async () => {
+    setupApi();
+    render(<ProjectListNew />, {
+        permissions: [{ permission: CREATE_PROJECT }],
+    });
+
+    await waitFor(async () => {
+        const button = await screen.findByText('New project');
+        expect(button).not.toBeDisabled();
+    });
+});
--- a/frontend/src/component/project/ProjectList/ProjectList.test.tsx
+++ b/frontend/src/component/project/ProjectList/ProjectList.test.tsx
@@ -30,6 +30,9 @@ test('Enabled new project button when limits, version and permission allow for i
         permissions: [{ permission: CREATE_PROJECT }],
     });
 
+    const button = await screen.findByText('New project');
+    expect(button).toBeDisabled();
+
     await waitFor(async () => {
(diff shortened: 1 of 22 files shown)


src/lib/features/project/project-service.limit.test.ts
```diff
@@ -28,24 +28,3 @@ test('Should not allow to exceed project limit', async () => {
         "Failed to create project. You can't create more than the established limit of 1.",
     );
 });
-
-test('Should enforce minimum project limit of 1', async () => {
```
kwasniew, line 32: we moved the invalid data checking to the config

frontend/src/component/project/ProjectList/ProjectList.test.tsx
```diff
@@ -0,0 +1,37 @@
+import { render } from 'utils/testRenderer';
+import { ProjectListNew } from './ProjectList';
+import { screen, waitFor } from '@testing-library/react';
+import { testServerRoute, testServerSetup } from '../../../utils/testServer';
+import { CREATE_PROJECT } from '../../providers/AccessProvider/permissions';
+
+const server = testServerSetup();
+
+const setupApi = () => {
+    testServerRoute(server, '/api/admin/ui-config', {
```
daveleek, line 10: Cool, is this how we stub out backends for testing?

kwasniew: yup, we mock HTTP with the msw library wrapped in our own abstraction (https://mswjs.io/)

frontend/src/component/project/ProjectList/ProjectList.test.tsx
thomasheartman, line 1: Is there no test for when it should be disabled?

frontend/src/component/project/ProjectList/ProjectList.tsx
```diff
@@ -86,16 +95,31 @@ function resolveCreateButtonData(
...
-const ProjectCreationButton = () => {
+const useProjectLimit = (projectsLimit: number, projectCount: number) => {
+    const resourceLimitsEnabled = useUiFlag('resourceLimits');
+    const limitReached = resourceLimitsEnabled && projectCount >= projectsLimit;
+
+    return limitReached;
+};
+
+const ProjectCreationButton: FC<{ projectCount: number }> = ({
+    projectCount,
```
thomasheartman, line 106: Question regarding SWR and shared caches etc:

Is there any point in passing the `projectCount` here instead of using `useProjects` directly in the button component? If it's a shared cache, then it doesn't incur any extra costs, right? With all the other stuff this button is doing (getting search params, access data, etc), it feels strange that _that's_ the thing we pass in.

thomasheartman (approved):

frontend/src/hooks/api/getters/useUiConfig/defaultValue.tsx
```diff
@@ -40,5 +40,6 @@ export const defaultValue: IUiConfig = {
         featureEnvironmentStrategies: 30,
         environments: 50,
         constraintValues: 250,
+        projects: 500,
```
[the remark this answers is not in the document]
kwasniew: I think we use it if backend didn't have a value

frontend/src/component/project/ProjectList/ProjectList.tsx
kwasniew: If we move useProjects inside the button then the button is aware of all the details of all the projects while it only needs to know how many projects we have. I introduced this param to limit the surface area of the things the button needs to know about projects.

frontend/src/component/project/ProjectList/ProjectList.test.tsx
thomasheartman: Sure, but this doesn't ensure that the limit condition actually kicks in. I mean, I'll leave it up to you, but because this task is about actually disabling the button when the user has reached the limit, it seems strange to me not to have a test that checks that exact thing 🤷🏼

frontend/src/component/project/ProjectList/ProjectList.tsx
thomasheartman: Yeah. This is a good point, and I'm honestly not sure what we typically do. On one extreme, you have the Elm architecture, right, where everything is passed down as props from above. But with SWR and react hooks, it's more like everything is accessible everywhere, so we grab what we need when we need it.

So in that case, it makes sense to me that this button, which is a highly specialized component already, can just go ahead and grab the projects directly if it needs them.

On the other hand, this kind of rampant implicit reading does make me throw up in my mouth a bit because it goes against a lot of what I have been taught. Also makes it harder to test when you need to set up a bunch of server routes etc. 

So I'm happy to go either way, but I'm not convinced that leaving the project count alone as something we need to pass makes the most sense. But I'm also not opposed to it. I see your reasoning and am happy for you to leave it as is. I also got to hear your thoughts on the matter, so that's great 😄

kwasniew: For the one off component like this one it probably does not matter that much. I'd be more worried if we shared this component in 20 different places then the shape of the API and the corresponding tradeoffs would matter a lot more.

PR #7519, opened 2024-07-02
chore: extract api token service composition root; place it in /features

thomasheartman:
This is a refactoring task, creating an ApiTokenService composition root in /features.


src/lib/services/api-token-service.test.ts
```diff
@@ -37,21 +31,7 @@ test('Should init api token', async () => {
...
-    const eventService = createFakeEventsService(config);
-
-    new ApiTokenService(
-        { apiTokenStore, environmentStore },
-        config,
-        eventService,
-    );
-
-    await insertCalled;
+    const { apiTokenStore } = createFakeApiTokenService(config);
```
thomasheartman, line 34: I'm kinda unsure about this test. I wonder whether the implicit insertion of init api tokens into a fake store by the api token service creation is really telling us anything useful? I guess it's proving that the store gets the data from the config passed to it, though.

Wouldn't it be better to have this in an apiTokenStore test?

kwasniew (approved):

PR #7524, opened 2024-07-03
feat: segments limit

kwasniew:
## About the changes

Limit how many segments we can create in our backend (by default it's 300). 

Details:
* I had to add method to our fake store to get the unit test for limits running
* if we set limit to 0 then we can effectively disable segment creation

### DIFF
--- a/src/test/fixtures/fake-segment-store.ts
+++ b/src/test/fixtures/fake-segment-store.ts
@@ -2,12 +2,18 @@ import type { ISegmentStore } from '../../lib/features/segment/segment-store-typ
 import type { IFeatureStrategySegment, ISegment } from '../../lib/types/model';
 
 export default class FakeSegmentStore implements ISegmentStore {
-    count(): Promise<number> {
-        return Promise.resolve(0);
+    segments: ISegment[] = [];
+    currentId: number = 0;
+
+    async count(): Promise<number> {
+        return this.segments.length;
     }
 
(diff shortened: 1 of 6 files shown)


src/test/fixtures/fake-segment-store.ts
```diff
@@ -2,12 +2,18 @@ import type { ISegmentStore } from '../../lib/features/segment/segment-store-typ
...
+    currentId: number = 0;
+
+    async count(): Promise<number> {
+        return this.segments.length;
     }
 
-    create(): Promise<ISegment> {
-        throw new Error('Method not implemented.');
+    async create(segment: Omit<ISegment, 'id'>): Promise<ISegment> {
+        const newSegment = { ...segment, id: this.currentId++ };
```
thomasheartman, line 13: nit: I'm a little wary of incrementing variables in assignment (ie `id: this.currentId++`). It's not immediately clear to me whether it gets incremented before or after the assignment (but probably before, right? so the first id is 1?).

I'd prefer to see that as a separate statement outside of the assignment. But this is just a test, so it's not a big deal, anyway.

thomasheartman (approved):

src/test/fixtures/fake-segment-store.ts
kwasniew: here it doesn't matter that much, but to avoid confusion I'll separate it into something more obvious

PR #7526, opened 2024-07-03
feat: track interaction with search

sjaanus:
Now we will start tracking how users interact with search box. Whether they open it manually or use CTRL K


frontend/src/component/common/Search/Search.tsx
```diff
@@ -143,6 +145,22 @@ export const Search = ({
             }
         },
     );
+
+    const recordFocusEvent = (e: React.FocusEvent<HTMLInputElement>) => {
+        if (!e.relatedTarget) {
```
daveleek, line 150: Cool, what sets the relatedTarget prop? The hotkey hook?

daveleek (approved):
Great!

frontend/src/component/common/Search/Search.tsx
sellinjaanus: good question, it does not work as I expected

PR #7528, opened 2024-07-03
feat: segments limit ui

kwasniew:
## About the changes

Disable 'New segment' button when limit of 300 reached. Screenshot has manually adjusted limit of 6
Screenshot 2024-07-03 at 11 19 26

Additionally, I decided to make the create segment button disabled when segments are being loaded or config is being loaded.

Modified existing test to avoid adding too many UI tests since they are quite slow.

### DIFF
--- a/frontend/src/component/segments/CreateSegmentButton/CreateSegmentButton.tsx
+++ b/frontend/src/component/segments/CreateSegmentButton/CreateSegmentButton.tsx
@@ -15,7 +15,6 @@ export const CreateSegmentButton: FC<{
     const projectId = useOptionalPathParam('projectId');
     const navigate = useNavigate();
 
-    console.log('***', disabled);
 
     return (
         <PermissionButton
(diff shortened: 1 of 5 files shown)


frontend/src/component/segments/CreateSegmentButton/CreateSegmentButton.tsx
```diff
@@ -6,11 +6,17 @@ import PermissionButton from 'component/common/PermissionButton/PermissionButton
...
 
-export const CreateSegmentButton = () => {
+export const CreateSegmentButton: FC<{
+    disabled: boolean;
+    tooltip?: string;
+}> = ({ disabled, tooltip }) => {
     const projectId = useOptionalPathParam('projectId');
     const navigate = useNavigate();
 
+    console.log('***', disabled);
```
thomasheartman, line 18: think you forgot this one

```suggestion
```

thomasheartman (approved):

PR #7529, opened 2024-07-03
chore: remove unused clone code

sjaanus:
Removing this code, because variants per env are GA and I did not find any usage of it.


gastonfournier (approved):
LGTM!

PR #7530, opened 2024-07-03
feat: use different cache key for command bar

sjaanus:
Now command bar and search do not share same cache, so they wont conflict anymore.

### DIFF
--- a/frontend/src/hooks/api/getters/useFeatureSearch/useFeatureSearch.ts
+++ b/frontend/src/hooks/api/getters/useFeatureSearch/useFeatureSearch.ts
@@ -25,8 +25,8 @@ const fallbackData: SearchFeaturesSchema = {
     total: 0,
 };
 
-const SWR_CACHE_SIZE = 10;
-const PREFIX_KEY = 'api/admin/search/features?';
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/api/getters/useFeatureSearch/useFeatureSearch.ts
```diff
@@ -55,17 +55,19 @@ const createFeatureSearch = () => {
     return (
         params: SearchFeaturesParams,
         options: SWRConfiguration = {},
+        cachePrefix: string = '',
     ): UseFeatureSearchOutput => {
         const { KEY, fetcher } = getFeatureSearchFetcher(params);
```
daveleek, line 60: is the KEY in use here after this change?

sjaanus: Yes, on line 61.

daveleek: Sigh, blink and you miss it 😅

daveleek (approved):
LG!

PR #7531, opened 2024-07-03
API tokens - virtualized table

Tymek:
## About the changes
API tokens table in both main list and project API tokens list can now support more items - it doesn't slow the browser down if there is >500 items.
Adjusted to match current width limit. Downside is that "Type" column description doesn't entirely fit.

image

### DIFF
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
@@ -22,16 +22,16 @@ describe('ProjectsList', () => {
         expect(links[0]).toHaveTextContent('project');
     });
 
-    it('should render no text if no projects are passed', async () => {
+    it('should render "*" if no projects are passed', async () => {
         const { container } = render(<ProjectsList />);
 
-        expect(container.textContent).toEqual('');
+        expect(container.textContent).toEqual('*');
     });
 
-    it('should render no text if empty projects array is passed', async () => {
+    it('should render "*" if empty projects array is passed', async () => {
         const { container } = render(<ProjectsList projects={[]} />);
 
-        expect(container.textContent).toEqual('');
+        expect(container.textContent).toEqual('*');
     });
 
     it('should show the number of projects', async () => {
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
@@ -22,16 +22,16 @@ describe('ProjectsList', () => {
         expect(links[0]).toHaveTextContent('project');
     });
 
-    it('should render asterisk if no projects are passed', async () => {
+    it('should render no text if no projects are passed', async () => {
         const { container } = render(<ProjectsList />);
 
-        expect(container.textContent).toEqual('*');
+        expect(container.textContent).toEqual('');
     });
 
-    it('should render asterisk if empty projects array is passed', async () => {
+    it('should render no text if empty projects array is passed', async () => {
         const { container } = render(<ProjectsList projects={[]} />);
 
-        expect(container.textContent).toEqual('*');
+        expect(container.textContent).toEqual('');
     });
 
     it('should show the number of projects', async () => {
@@ -42,5 +42,17 @@ describe('ProjectsList', () => {
         );
 
         expect(container.textContent).toContain('4 projects');
+    });
+
+    it('should render "*" if project is "*" and no projects are passed', async () => {
+        const { container } = render(<ProjectsList project='*' />);
+
+        expect(container.textContent).toEqual('*');
+    });
+
+    it('should render "*" if projects has only "*"', async () => {
+        const { container } = render(<ProjectsList projects={['*']} />);
+
+        expect(container.textContent).toEqual('*');
     });
 });
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
@@ -60,8 +60,7 @@ export const ProjectsList: FC<IProjectsListProps> = ({ projects, project }) => {
 
     if (
         (projectsList.length === 1 && projectsList[0] === '*') ||
-        project === '*' ||
-        (!project && (!projectsList || projectsList.length === 0))
+        project === '*'
     ) {
         return (
             <TextCell>
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
@@ -58,27 +58,26 @@ export const ProjectsList: FC<IProjectsListProps> = ({ projects, project }) => {
         );
     }
 
-    if ((projects?.length === 1 && projects?.[0] === '*') || project === '*') {
-        return (
-            <TextCell>
-                <HtmlTooltip
-                    title='ALL current and future projects'
-                    placement='bottom'
-                    arrow
-                >
-                    <span>
-                        <Highlighter search={searchQuery}>*</Highlighter>
-                    </span>
-                </HtmlTooltip>
-            </TextCell>
-        );
-    }
-
-    if (projectsList.length === 1 || project) {
+    if (
+        (projectsList.length === 1 && projectsList[0] !== '*') ||
+        (project && project !== '*')
+    ) {
         const item = project || projectsList[0];
 
         return <LinkCell to={`/projects/${item}`} title={item} />;
     }
 
-    return null;
+    return (
+        <TextCell>
+            <HtmlTooltip
+                title='ALL current and future projects'
+                placement='bottom'
+                arrow
+            >
+                <span>
+                    <Highlighter search={searchQuery}>*</Highlighter>
+                </span>
+            </HtmlTooltip>
+        </TextCell>
+    );
 };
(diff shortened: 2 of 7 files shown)


frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
```diff
@@ -22,16 +22,16 @@ describe('ProjectsList', () => {
         expect(links[0]).toHaveTextContent('project');
     });
 
-    it('should render asterisk if no projects are passed', async () => {
+    it('should render no text if no projects are passed', async () => {
```
thomasheartman, line 25: What's the `container` in this case? The component is called `ProjectsList` ...

I was confused by this at first, but the projects list here is the projects list _cell_, right? How come we're changing the behavior when you don't pass anything?

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
```diff
@@ -84,5 +80,5 @@ export const ProjectsList: FC<IProjectsListProps> = ({ projects, project }) => {
         return <LinkCell to={`/projects/${item}`} title={item} />;
     }
 
-    return null;
+    return '';
```
thomasheartman, line 83: why are we changing the return to an empty string here? wouldn't it be better to return `null`?

frontend/src/component/common/ApiTokenTable/ApiTokenTable.tsx
```diff
@@ -27,7 +9,7 @@ import { ConditionallyRender } from 'component/common/ConditionallyRender/Condit
 
 import { useConditionallyHiddenColumns } from 'hooks/useConditionallyHiddenColumns';
 
-const hiddenColumnsSmall = ['Icon', 'createdAt'];
+const hiddenColumnsMedium = ['Icon', 'createdAt', 'seenAt'];
 const hiddenColumnsCompact = ['Icon', 'project', 'seenAt'];
```
thomasheartman, line 13: What's the difference between medium/small screens and compact? Is that how they're rendered? I'm curious because they have almost the same columns hidden, but not quite.

frontend/src/component/common/ApiTokenTable/ApiTokenTable.tsx
```diff
@@ -37,34 +19,27 @@ interface IApiTokenTableProps {
...
     loading,
     rows,
-    getTableProps,
-    getTableBodyProps,
     headerGroups,
     globalFilter,
     prepareRow,
 }: IApiTokenTableProps) => {
-    const isSmallScreen = useMediaQuery(theme.breakpoints.down('md'));
+    const isMediumScreen = useMediaQuery(theme.breakpoints.down('xl'));
```
thomasheartman, line 36: Should this really be `xl`? I'd expect `l` to be more natural?

frontend/src/component/common/ApiTokenTable/useApiTokenTable.tsx
```diff
@@ -22,15 +27,16 @@ export const useApiTokenTable = (
         return [
             {
                 id: 'Icon',
-                width: '1%',
```
thomasheartman, line 25: Out of curiosity, How does `1%` actually work on the table? Because I don't think it is actually 1%, is it?

thomasheartman (commented):
> API tokens table in both main list and project API tokens list can now support more items.

I don't understand what you mean. More items? As in more api keys? More columns? How do they support them?


> Adjusted to match current width limit. Downside is that "Type" column description doesn't entirely fit.

Ah, that _is_ unfortunate. Are we sure we want to truncate it? If you've added new columns, is there anything else you can make smaller? If we're gonna start truncating, I'd double check with UX that that's alright (maybe you've done that already?). Not sure how important that column is to our users, but if it's important, it'd be a shame to truncate it.

frontend/src/component/common/ApiTokenTable/useApiTokenTable.tsx
Tymek: On tables that are not virtualized (actual `display:table`, instead of flexbox) it will indicate that this column should be squeezed to content.

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
Tymek: Does not matter much. This looked more consistent with tests

frontend/src/component/common/ApiTokenTable/ApiTokenTable.tsx
Tymek: "compact" is here to indicate different set of columns for API tokens table inside Project settings. That's how it was done before so I didn't change it.

frontend/src/component/common/ApiTokenTable/ApiTokenTable.tsx
Tymek: xl is still 1536px. With sidebar and on contemporary monitors this is how hiding columns turned out when testing screen width.

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
Tymek: This was the place where I found an issue with keys for removed projects. I'd like this component to show if we have no projects assigned, and not hide that something is wrong.

Tymek:
> Ah, that _is_ unfortunate. Are we sure we want to truncate it? If you've added new columns, is there anything else you can make smaller? If we're gonna start truncating, I'd double check with UX that that's alright (maybe you've done that already?). Not sure how important that column is to our users, but if it's important, it'd be a shame to truncate it.

Current version already had an issue where on some medium-large screens it will display horizontal scroll bar. I don't think we can show more without adjusting breakpoints

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
thomasheartman: Right, so what you're saying is that the case where we have no projects in the output means that the API token doesn't work on any projects? Are you sure about that? 

Because I think that if you have one of those ancient keys from before we introduced the new [API key format](https://docs.getunleash.io/reference/api-tokens-and-client-keys#format) (in Unleash 4.3), then you won't have any project data associated with it, but it'll still work for all projects.

So I'm not sure this is the right assumption.

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
thomasheartman: Right. I don't really know if it matters, but the [docs use null](https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null). Happy for you to keep this one, though. What do we do in other components?

frontend/src/component/common/ApiTokenTable/ApiTokenTable.tsx
thomasheartman: Cool; all good.

frontend/src/component/common/ApiTokenTable/ApiTokenTable.tsx
thomasheartman: Right. But shouldn't it be `isLargeScreen`, then? I'm just confused by the mismatch between the media query and the variable name. It looks like someone made a mistake somewhere.

frontend/src/component/common/ApiTokenTable/useApiTokenTable.tsx
thomasheartman: Cool; thanks!

thomasheartman:
> Current version already had an issue where on some medium-large screens it will display horizontal scroll bar. I don't think we can show more without adjusting breakpoints

Sure. And looking at the screenshot, it appears it's the description of the type that gets cut off, not the actual type name itself, right? Should be fine.

> API tokens table in both main list and project API tokens list can now support more items.

I still don't know what you mean by this statement, by the way. Can you elaborate?

frontend/src/component/common/ApiTokenTable/ApiTokenTable.tsx
Tymek: `isNotExtraLarge`?

Tymek:
> >  API tokens table in both main list and project API tokens list can now support more items.
> I still don't know what you mean by this statement, by the way. Can you elaborate?

As in, it doesn't slow the browser down if there is >500 items

frontend/src/component/common/ApiTokenTable/ApiTokenTable.tsx
thomasheartman: Sure? 🤷🏼

thomasheartman:
> As in, it doesn't slow the browser down if there is >500 items

Great! Can you add that to the description, so it's clearer to the next person who sees this PR?

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.test.tsx
Tymek: Ok, I checked that. Btw, that's Unleash 4.1:

Screenshot from 2024-07-09 11-32-52

thomasheartman: Oh, look, the sidebar is back! Or maybe it's back now, then? 🤔 

Anyway, what did you find when you checked it? That table doesn't show projects, but if you upgrade Unleash to a whenever we introduced the projects column in the table, what happens?

thomasheartman (approved):

PR #7532, opened 2024-07-03
feat: UI limit for API tokens

thomasheartman:
This PR activates the limit for API token creation in both the global API token window and in the project-level API token tab.

Because the same button is used in two places, I encapsulated the fetching of flags and resource limits within the button. I can be convinced to pass the current API token count and the limit as arguments, but I think this is the right solution for this case.

### DIFF
--- a/frontend/src/component/common/ApiTokenTable/CreateApiTokenButton/CreateApiTokenButton.test.tsx
+++ b/frontend/src/component/common/ApiTokenTable/CreateApiTokenButton/CreateApiTokenButton.test.tsx
@@ -1,0 +1,68 @@
+import { screen, waitFor } from '@testing-library/react';
+import { render } from 'utils/testRenderer';
+import { testServerRoute, testServerSetup } from 'utils/testServer';
+import { CreateApiTokenButton } from './CreateApiTokenButton';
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/ApiTokenTable/CreateApiTokenButton/CreateApiTokenButton.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+        tokens: Array.from({ length: apiTokenCount }).map((_, i) => ({
+            secret: 'super-secret',
+            tokenName: `token—name-${i}`,
+            type: 'client',
+        })),
+    });
+};
+
+test('should allow you to create API tokens when there are fewer apiTokens than the limit', async () => {
+    setupApi({ apiTokenLimit: 5, apiTokenCount: 2 });
```
kwasniew, line 32: Nitpick: I usually set the boundary condition to be as close to the actual boundary value as possible. So in this case I'd do 3. But it's just a personal preference. When I see 5 I'm wondering if there's something special about the number.

thomasheartman: Fair enough. Happy to change that 💁🏼

PR #7533, opened 2024-07-03
feat: quick suggestions click close

sjaanus:
Now recently visited projects, features, pages will also close the command bar.


daveleek (approved):
Brilliant!

PR #7536, opened 2024-07-03
Chore: add limits to feature flags

thomasheartman:
This PR adds a feature flag limit to Unleash. It's set up to be overridden in Enterprise, where we turn the limit up. 

I've also fixed a couple bugs in the fake feature flag store.

### DIFF
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -200,7 +200,7 @@ exports[`should create default config 1`] = `
     "constraintValues": 250,
     "environments": 50,
     "featureEnvironmentStrategies": 30,
-    "flags": 250,
+    "flags": 5000,
     "projects": 500,
     "segmentValues": 1000,
     "segments": 300,
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -200,7 +200,7 @@ exports[`should create default config 1`] = `
     "constraintValues": 250,
     "environments": 50,
     "featureEnvironmentStrategies": 30,
-    "flags": 5000,
+    "featureFlags": 5000,
     "projects": 500,
     "segmentValues": 1000,
     "segments": 300,
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -200,6 +200,7 @@ exports[`should create default config 1`] = `
     "constraintValues": 250,
     "environments": 50,
     "featureEnvironmentStrategies": 30,
+    "flags": 250,
     "projects": 500,
     "segmentValues": 1000,
     "segments": 300,
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -679,6 +679,13 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
             0,
             parseEnvVarNumber(process.env.UNLEASH_SEGMENTS_LIMIT, 300),
         ),
+        flags: Math.max(
+            1,
+            parseEnvVarNumber(
+                process.env.UNLEASH_FLAGS_LIMIT,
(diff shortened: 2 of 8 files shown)


src/lib/features/feature-toggle/createFeatureToggleService.ts
thomasheartman, line 1: I've only added the store I needed here because there's sooooo many dependencies. Sure, we could add all of them, but I think adding the ones we need when we need them might be better here.

src/lib/features/feature-toggle/fakes/fake-feature-toggle-store.ts
thomasheartman, line 1: There's two (2.5?) fixes here:

1. Filter queries didn't work for `getAll`. They do now.
2. The filter query didn't work correctly because it didn't differentiate between `query.achived: false` `query.archived: undefined`. It also didn't account for the fact that if  flags don't have an `archived` property, then it should default to `false`. 
3. I've assumed that the same thing goes for `stale`, so I added the same handling there.

src/lib/features/feature-toggle/tests/feature-toggle-service.limit.test.ts
```diff
@@ -14,71 +15,145 @@ const alwaysOnFlagResolver = {
...
-                featureEnvironmentStrategies: LIMIT,
-            },
-        } as unknown as IUnleashConfig);
-
-    const addStrategy = () =>
-        featureToggleService.unprotectedCreateStrategy(
-            { name: 'default', featureName: 'feature' } as IStrategyConfig,
-            { projectId: 'default', featureName: 'feature' } as any,
-            {} as IAuditUser,
+describe('Strategy limits', () => {
```
thomasheartman, line 18: I moved strategy and constraint testing into a shared describe block because they both use strategy methods on the service. This keeps them separate from the flags testing block. We can split this out into two files if we think that's better, but this seems sensible to me for now.

src/lib/openapi/spec/resource-limits-schema.ts
```diff
@@ -103,6 +104,13 @@ export const resourceLimitsSchema = {
             example: 300,
             description: 'The maximum number of segments allowed.',
         },
+        flags: {
```
thomasheartman, line 107: Should we use `flags` or `featureFlags`; got any thoughts? I put `flags` in because it's shorter, but maybe `featureFlags` is more explicit? 🤷🏼

src/lib/__snapshots__/create-config.test.ts.snap
```diff
@@ -200,6 +200,7 @@ exports[`should create default config 1`] = `
     "constraintValues": 250,
     "environments": 50,
     "featureEnvironmentStrategies": 30,
+    "flags": 250,
```
kwasniew, line 203: 10k with 50k override

src/lib/create-config.ts
```diff
@@ -679,6 +679,13 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
             0,
             parseEnvVarNumber(process.env.UNLEASH_SEGMENTS_LIMIT, 300),
         ),
+        flags: Math.max(
+            1,
+            parseEnvVarNumber(
+                process.env.UNLEASH_FLAGS_LIMIT,
+                options?.resourceLimits?.flags || 250,
```
kwasniew, line 686: wrong value

src/lib/features/feature-toggle/createFeatureToggleService.ts
kwasniew: I do the same, expose what you really need

src/lib/openapi/spec/resource-limits-schema.ts
```diff
@@ -103,6 +104,13 @@ export const resourceLimitsSchema = {
...
+            type: 'integer',
+            minimum: 1,
+            example: 5000,
```
kwasniew, line 110: I'd match the example with the actual limit since this is what most people will get anyway

kwasniew (approved):
Nice that you improved the fake store along the way

PR #7538, opened 2024-07-03
feat: limit component

kwasniew:
## About the changes

Approaching limit variant
Screenshot 2024-07-03 at 16 07 17

Limit reached variant
Screenshot 2024-07-03 at 16 07 43

Details:
* there's name and shortName for the things we're limiting. Use case: for strategies we need `strategies in this environment` in the title and just `strategies` in the other two sections
* when limit percentage is below 80% we don't show the component (we can make it configurable if needed or even move the check outside the component)
* the component has optional onClose that is a placeholder for the client code passing a click handler. Most probably we'll allow to local storage persist the closed state when the limit is approaching and force the limit info when the limit is reached without an option to close it. Maybe this logic will even stay in this component but it's up to the next PRs to decide

### DIFF
--- a/frontend/src/component/common/Limit/Limit.test.tsx
+++ b/frontend/src/component/common/Limit/Limit.test.tsx
@@ -1,0 +1,67 @@
+import { screen } from '@testing-library/react';
+import { render } from 'utils/testRenderer';
+import { Limit } from './Limit';
+
+test('Render approaching limit variant', () => {
+    render(
+        <Limit
+            name='strategies in this environment'
+            shortName='strategies'
+            limit={10}
+            currentValue={8}
+        />,
+    );
+
+    screen.getByText(
+        'You are nearing the limit for strategies in this environment',
+    );
+    screen.getByText('80%');
+    screen.getByText('Limit: 10');
+});
+
+test('Render reached limit variant', () => {
+    render(
+        <Limit
+            name='strategies in this environment'
+            shortName='strategies'
+            limit={10}
+            currentValue={10}
+        />,
+    );
+
+    screen.getByText(
+        'You reached the limit for strategies in this environment',
+    );
+    screen.getByText('100%');
+    screen.getByText('Limit: 10');
+});
+
+test('Render exceeded limit variant', () => {
+    render(
+        <Limit
+            name='strategies in this environment'
+            shortName='strategies'
+            limit={10}
+            currentValue={20}
+        />,
+    );
+
+    screen.getByText(
+        'You reached the limit for strategies in this environment',
+    );
+    screen.getByText('200%');
+    screen.getByText('Limit: 10');
+});
+
+test('Do not render any limit below threshold', () => {
+    render(
+        <Limit
+            name='strategies in this environment'
+            shortName='strategies'
+            limit={10}
+            currentValue={7}
+        />,
+    );
+
+    expect(screen.queryByText('Limit: 10')).not.toBeInTheDocument();
+});
--- a/frontend/src/component/common/Limit/Limit.test.tsx
+++ b/frontend/src/component/common/Limit/Limit.test.tsx
@@ -30,7 +30,7 @@ test('Render reached limit variant', () => {
     );
 
     screen.getByText(
-        'You reached the limit for strategies in this environment',
+        'You have reached the limit for strategies in this environment',
     );
     screen.getByText('100%');
     screen.getByText('Limit: 10');
@@ -47,7 +47,7 @@ test('Render exceeded limit variant', () => {
     );
 
     screen.getByText(
-        'You reached the limit for strategies in this environment',
+        'You have reached the limit for strategies in this environment',
     );
     screen.getByText('200%');
     screen.getByText('Limit: 10');
--- a/frontend/src/component/common/Limit/Limit.tsx
+++ b/frontend/src/component/common/Limit/Limit.tsx
@@ -1,7 +1,7 @@
 import { Box, IconButton, styled, Tooltip, Typography } from '@mui/material';
 import LinearProgress from '@mui/material/LinearProgress';
 import { Link } from 'react-router-dom';
-import WarningIcon from '@mui/icons-material/WarningAmber';
+import WarningIcon from '@mui/icons-material/ErrorOutlined';
 import ErrorIcon from '@mui/icons-material/Cancel';
 import CloseIcon from '@mui/icons-material/Close';
 import type { FC } from 'react';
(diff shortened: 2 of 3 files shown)


FredrikOseberg:
> ## About the changes
> Approaching limit variant Screenshot 2024-07-03 at 16 07 17
> 
> Limit reached variant Screenshot 2024-07-03 at 16 07 43
> 
> Details:
> 
> * there's name and shortName for the things we're limiting. Use case: for strategies we need `strategies in this environment` in the title and just `strategies` in the other two sections
> * when limit percentage is below 80% we don't show the component (we can make it configurable if needed or even move the check outside the component)
> * the component has optional onClose that is a placeholder for the client code passing a click handler. Most probably we'll allow to local storage persist the closed state when the limit is approaching and force the limit info when the limit is reached without an option to close it. Maybe this logic will even stay in this component but it's up to the next PRs to decide
> 
> ### Important files
> ## Discussion points

Warning icons look different than in the sketches. Do you plan a follow up?

frontend/src/component/common/Limit/Limit.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <Limit
+            name='strategies in this environment'
+            shortName='strategies'
+            limit={10}
+            currentValue={10}
+        />,
+    );
+
+    screen.getByText(
+        'You reached the limit for strategies in this environment',
```
thomasheartman, line 33: This is a suggestion for the component, but I'll comment here because it was the first place I saw it:

```suggestion
        'You have reached the limit for strategies in this environment',
```

We should add the "have" in at the start.

frontend/src/component/common/Limit/Limit.tsx
```diff
@@ -0,0 +1,144 @@
...
+}));
+
+const Header = styled(Box)(({ theme }) => ({
+    display: 'flex',
+    gap: theme.spacing(1),
+    alignItems: 'center',
+    fontWeight: 'bold',
+    borderBottom: `2px solid ${theme.palette.background.application}`,
+    padding: theme.spacing(3, 4),
+    fontSize: theme.fontSizes.mainHeader,
```
thomasheartman, line 37: Isn't `fontSizes` deprecated? Can we use `theme.typography` instead? something like:

```suggestion
    fontSize: theme.typography.h2.fontSize,
```

frontend/src/component/common/Limit/Limit.tsx
```diff
@@ -0,0 +1,144 @@
...
+                <ConditionallyRender
+                    condition={belowLimit}
+                    show={
+                        <ExpandableBox>
+                            You are nearing the limit for {name}
+                        </ExpandableBox>
+                    }
+                    elseShow={
+                        <ExpandableBox>
+                            You reached the limit for {name}
```
thomasheartman, line 96: Ah, here it is! I'll add a suggestion 😄 

```suggestion
                            You have reached the limit for {name}
```

frontend/src/component/common/Limit/Limit.tsx
```diff
@@ -0,0 +1,144 @@
...
+                                <CloseIcon />
+                            </IconButton>
+                        </Tooltip>
+                    }
+                />
+            </Header>
+            <Main>
+                <LimitStats>
+                    You have added {currentValue} {shortName ?? name}, that is
+                    equivalent to{' '}
```
thomasheartman, line 115: Very nitpicky and more on the UX side of things: but does it really make sense to say "That is equivalent to 100% of the limit" when you're at 100%? It's a small thing, but maybe we could do something like:

> You have added {currentValue} {shortName ?? name}. To add new {shortName ?? name}, you must first remove some of the existing {shortName ?? name}.

But very happy for that not to be in this PR at all 😄

frontend/src/component/common/Limit/Limit.tsx
```diff
@@ -0,0 +1,144 @@
...
+                            <IconButton onClick={onClose}>
+                                <CloseIcon />
+                            </IconButton>
+                        </Tooltip>
+                    }
+                />
+            </Header>
+            <Main>
+                <LimitStats>
+                    You have added {currentValue} {shortName ?? name}, that is
```
thomasheartman, line 114: I think this should either be a full stop or replaced by a "which":

```suggestion
                    You have added {currentValue} {shortName ?? name}. That is
```

```suggestion
                    You have added {currentValue} {shortName ?? name}, which is
```

frontend/src/component/common/Limit/Limit.tsx
```diff
@@ -0,0 +1,144 @@
...
+                        target='_blank'
+                        to={'https://docs.getunleash.io/reference/limits'}
+                    >
+                        Read more about limits
+                    </Link>
+                    <Typography fontWeight='bold'>Limit: {limit}</Typography>
+                </LimitExplanation>
+            </Main>
+            <Footer>
+                If you need more than <b>{limit}</b> {shortName ?? name}, please
```
thomasheartman, line 136: These should probably be `<strong>` instead. According to mdn:
> you should not use `<b>` for styling text or granting importance. If you wish to create boldface text, you should use the CSS [font-weight](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight) property. If you wish to indicate an element is of special importance, you should use the [`<strong>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong) element.

```suggestion
                If you need more than <strong>{limit}</strong> {shortName ?? name}, please
```

frontend/src/component/common/Limit/Limit.tsx
```diff
@@ -0,0 +1,144 @@
...
+                reach out to us at{' '}
+                <a href='mailto:cs@getunleash.io?subject=Increase limit'>
```
thomasheartman, line 138: Nice handling of the mailto link here 🙌🏼

thomasheartman (approved):
Looks fine to me! Got a couple minor nits to have a look at, but nothing blocking (or near blocking). But I think Fredrik's point stands, though: are the designs being updated in a followup?

frontend/src/component/common/Limit/Limit.tsx
kwasniew: BTW, we can also go past the limit if you did it before we introduced this feature so 110% is also possible

kwasniew: We can tweak it in the followup PRs when we have at least a few usages of the component

frontend/src/component/common/Limit/Limit.tsx
kwasniew: didn't know about this one

kwasniew:
Regarding the icons we can adjust them in a followup PR. I used the same error and warning icons as I found in other places.

frontend/src/component/common/Limit/Limit.tsx
thomasheartman: Ah, yeah, you're right. That's interesting. So the limit isn't really a limit in that sense. So yeah, let's keep it.

frontend/src/component/common/Limit/Limit.tsx
thomasheartman: Oh, yeah! I found it a while back when I got sick of seeing that `fontSizes` was deprecated 🤷🏼  Not sure if that's the right size for you, though! But it's probably one of them.

PR #7541, opened 2024-07-04
ui limits for flags

thomasheartman:
This PR disables the "create feature flag" button when you've reached the limits.

This one is a little more complex than the other UI limits, because we also have to take into account the project feature limit. I've tried to touch as little as possible, but I _have_ extracted the calculation of both limits into a single hook.


frontend/src/component/feature/CreateFeatureButton/CreateFeatureButton.tsx
thomasheartman, line 1: This file was apparently never used. So I deleted it.

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureTogglesHeader/ProjectFeatureTogglesHeader.tsx
thomasheartman, line 1: The VFC type is deprecated and equivalent to FC. I came across this and decided to update it.

frontend/src/component/feature/CreateFeature/useFlagLimits.test.ts
```diff
@@ -0,0 +1,71 @@
+import { renderHook } from '@testing-library/react';
+import { useFlagLimits } from './CreateFeature';
+import { vi } from 'vitest';
+
+vi.mock('hooks/useUiFlag', async (importOriginal) => {
```
kwasniew, line 5: this breaks the established pattern where we use msw for mocking HTTP

frontend/src/component/feature/CreateFeature/useFlagLimits.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    const actual = await importOriginal();
+    return {
+        ...(actual as {}),
+        useUiFlag: (flag: string) => flag === 'resourceLimits',
+    };
+});
+
+test('if both global and project-level limits are reached, then the error message shows the message for instance-wide limits', () => {
+    const { result } = renderHook(() =>
```
kwasniew, line 14: it's a mater of personal preferences but I usually create a hand written test component for my hooks to test them. However this one is also fine.

frontend/src/component/feature/CreateFeature/useFlagLimits.test.ts
thomasheartman: Sure! I can see if just mocking the endpoint will do the trick.

frontend/src/component/feature/CreateFeature/useFlagLimits.test.ts
thomasheartman: Right. I'm pretty sure we do this a few other places, so I'm inclined to keep it. But if you write hand written test components for it, you probably have a good reason for it. Why do you prefer those components?

kwasniew (approved):
I like how we improved test coverage here

frontend/src/component/feature/CreateFeature/useFlagLimits.test.ts
kwasniew: Mostly to show hook usage in context of some component. I know that renderHook is like a simulation of an anonymous component but I still prefer the explicitness of my own components. But as you said, we have both examples in the code so feel free to ignore my comment.

frontend/src/component/feature/CreateFeature/useFlagLimits.test.ts
thomasheartman: Actually, I've spent like 90 minutes now trying to get this to work as expected, so I'm gonna leave it in for now. We do use this same pattern two other places in the code. Because it's only a temporary requirement and because it's in tests, I don't want to spend more time on it.

If you have a solution for it that's easy to implement, I'm happy to have a look at that and set it up in a follow-up PR, though.

PR #7542, opened 2024-07-04
feat: limit component used in strategies

kwasniew:
## About the changes

Approaching limit
Screenshot 2024-07-04 at 13 53 06

Limit reached
Screenshot 2024-07-04 at 13 53 18

Flag disabled and limit reached
Screenshot 2024-07-04 at 15 00 37

Details:
* added warning and error limit component to env strategies
* removed disable on add strategy button in feature env overview so that people can get to the last screen and find the limit reached details
* removed irrelevant test for disable button state in feature env overview


thomasheartman (approved):
Nice one! 😄 

But I don't understand what your last image is supposed to show? I think the text says:
> Flex disabled and limit reached

But I'm not sure what I'm supposed to be looking for?

kwasniew:
@thomasheartman it was a typo. Flag disabled. When flag is disabled then the button is not locked/disabled. This is me showing that we don't break existing functionality.

PR #7544, opened 2024-07-04
feat: hide command bar when losing focus

daveleek:
Hides command bar when it loses focus


frontend/src/component/commandBar/CommandBar.tsx
```diff
@@ -213,6 +213,15 @@ export const CommandBar = () => {
             searchInputRef.current?.focus();
         }
     };
+
+    const onBlur = (evt: React.FocusEvent) => {
+        if (evt.relatedTarget === null) {
```
sjaanus, line 218: Maybe just one if statement

sjaanus (approved):

PR #7545, opened 2024-07-05
fix: sidebar update active when navigated via command menu.

sjaanus:
Currently, if the location changes, it does not update the active item. I added useEffect to update the active item.

This is required when the user navigates via the command menu.


daveleek (approved):
LG!

PR #7548, opened 2024-07-05
feat: Limit environments component

kwasniew:
## About the changes

Adding limit indicator to create environment form.

3 interesting cases below (also covered in unit tests)

Approaching limit
Screenshot 2024-07-05 at 09 26 15

Limit reached
Screenshot 2024-07-05 at 09 26 05

Previous limit reached (when flag disabled)
Screenshot 2024-07-05 at 09 26 42

PR #7549, opened 2024-07-05
chore: don't prevent users from entering the env form when they're at the limit

thomasheartman:
This change reverts the changes related to the button in particular
that were introduced in #7500. The button is now always enabled, and
the actual resource warning and creation blocking happens in the form,
courtesy of #7548.


kwasniew (approved):
nice, wasn't aware of this check :) good that you remembered about it

PR #7552, opened 2024-07-05
feat: Disallow repeating last 5 passwords.

chriswk:
We'll store hashes for the last 5 passwords, fetch them all for the user wanting to change their password, and make sure the password does not verify against any of the 5 stored hashes.

### DIFF
--- a/src/lib/db/user-store.ts
+++ b/src/lib/db/user-store.ts
@@ -11,8 +11,10 @@ import type {
     IUserUpdateFields,
 } from '../types/stores/user-store';
 import type { Db } from './db';
+import bcrypt from 'bcryptjs';
 
 const TABLE = 'users';
+const PASSWORD_HASH_TABLE = 'used_passwords';
 
 const USER_COLUMNS_PUBLIC = [
     'id',
@@ -24,6 +26,8 @@ const USER_COLUMNS_PUBLIC = [
     'is_service',
     'scim_id',
 ];
+
+const USED_PASSWORDS = ['user_id', 'password_hash', 'used_at'];
 
 const USER_COLUMNS = [...USER_COLUMNS_PUBLIC, 'login_attempts', 'created_at'];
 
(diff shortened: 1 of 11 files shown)


chriswk:
I'll update the frontend to handle the new PreviouslyUsedPassword Error.

src/lib/db/user-store.ts
```diff
@@ -71,6 +75,35 @@ class UserStore implements IUserStore {
...
+        userId: number,
+        password: string,
+    ): Promise<boolean> {
+        const previouslyUsedPasswords = await this.db(
+            PASSWORD_HASH_TABLE,
+        ).where({ user_id: userId });
+        return previouslyUsedPasswords.some((row) =>
+            bcrypt.compareSync(password, row.password_hash),
+        );
+    }
```
chriswk, line 88: Some argument to be made about moving the compareSync into the service instead, so the store only concerns itself with the database op. Will look at come Monday

nunogois: Agreed, would like to see this on the service layer if it's not too much trouble.

src/lib/db/user-store.ts
```diff
@@ -177,10 +210,24 @@ class UserStore implements IUserStore {
...
-        return this.activeUsers().where('id', userId).update({
+    async setPasswordHash(
+        userId: number,
+        passwordHash: string,
+        disallowNPreviousPasswords: number,
+    ): Promise<void> {
+        await this.activeUsers().where('id', userId).update({
             password_hash: passwordHash,
         });
+        if (passwordHash) {
```
nunogois, line 221: I think we don't need this here. If we do, it should also encompass the update above.

src/lib/error/password-previously-used.ts
```diff
@@ -0,0 +1,11 @@
+import { UnleashError } from './unleash-error';
+
+export class PasswordPreviouslyUsedError extends UnleashError {
+    statusCode = 400;
+
+    constructor(
+        message: string = `You've previously used this password, please use a new password`,
```
nunogois, line 7:
```suggestion
        message: string = `You've previously used this password. Please use a new password.`,
```

src/lib/services/user-service.ts
```diff
@@ -62,6 +63,7 @@ export interface ILoginUserRequest {
 }
 
 const saltRounds = 10;
+const disallowNPreviousPasswords = 5;
```
nunogois, line 66: Should this be configurable instead of a constant?

src/lib/db/user-store.ts
```diff
@@ -177,10 +210,24 @@ class UserStore implements IUserStore {
         return item.password_hash;
     }
 
-    async setPasswordHash(userId: number, passwordHash: string): Promise<void> {
-        return this.activeUsers().where('id', userId).update({
+    async setPasswordHash(
+        userId: number,
+        passwordHash: string,
+        disallowNPreviousPasswords: number,
```
nunogois, line 216: Should this have a default so we don't need to set it every time?

src/test/config/test-config.ts
```diff
@@ -13,6 +13,7 @@ function mergeAll<T>(objects: Partial<T>[]): T {
 }
 
 export function createTestConfig(config?: IUnleashOptions): IUnleashConfig {
+    getLogger.setMuteError(true);
```
nunogois, line 16: I wonder what's this, and whether it should be included in this PR.

nunogois (approved):
LGTM. Left a few comments that we can optionally address.

I suggest manually going through the password screens just to confirm that the new behaviors make sense to you as well.

src/lib/services/user-service.ts
chriswk: I think for now we do the same as we do with salt rounds, just leave it as a constant

src/test/config/test-config.ts
chriswk: It's muting the error logger of our test outputs, so we don't end up with tons of error logging that isn't actually an error when running tests.

PR #7556, opened 2024-07-08
feat: add limit warning for feature flags

thomasheartman:
This PR adds the Limit component to the feature flag creation form.

At the limit: 
image

Approaching the limit:
image

Below the limit threshold (no change):
image

### DIFF
--- a/frontend/src/component/feature/CreateFeature/CreateFeature.test.tsx
+++ b/frontend/src/component/feature/CreateFeature/CreateFeature.test.tsx
@@ -52,30 +52,6 @@ describe('button states', () => {
             expect(button).not.toBeDisabled();
         });
     });
-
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/CreateFeature/CreateFeature.test.tsx
```diff
@@ -28,50 +28,66 @@ const setupApi = ({
...
             name: /create feature flag/i,
         });
-        expect(button).not.toBeDisabled();
+        await waitFor(() => {
+            expect(button).not.toBeDisabled();
+        });
     });
 });
 
-test("should not allow you to create API tokens when you're at the global limit", async () => {
```
thomasheartman, line 55: This test was always true because the button starts off as disabled when we're loading, so I've removed it.

kwasniew (approved):

PR #7560, opened 2024-07-09
Chore: use createApiTokenService instead of newing it up

thomasheartman:
PR #7519 introduced the pattern of using `createApiTokenService` instead of newing it up. This usage was introduced in a concurrent PR (#7503), so we're just cleaning up and making the usage consistent.

### DIFF
--- a/src/lib/features/project/project-service.e2e.test.ts
+++ b/src/lib/features/project/project-service.e2e.test.ts
@@ -9,7 +9,7 @@ import { RoleName } from '../../types/model';
 import { randomId } from '../../util/random-id';
 import EnvironmentService from '../project-environments/environment-service';
 import IncompatibleProjectError from '../../error/incompatible-project-error';
-import { ApiTokenService, EventService } from '../../services';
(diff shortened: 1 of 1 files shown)


kwasniew (approved):

src/lib/features/project/project-service.e2e.test.ts
```diff
@@ -9,7 +9,7 @@ import { RoleName } from '../../types/model';
 import { randomId } from '../../util/random-id';
 import EnvironmentService from '../project-environments/environment-service';
 import IncompatibleProjectError from '../../error/incompatible-project-error';
-import { ApiTokenService, EventService } from '../../services';
+import { type ApiTokenService, EventService } from '../../services';
```
kwasniew, line 12: cool syntax. you don't see it often :)

thomasheartman: this is all biome's auto-fix black magic 🪄  but I'll take it 😄

PR #7561, opened 2024-07-09
feat: API Tokens limit - UI

Tymek:
## About the changes
When approaching limit or limit reached for the number of API tokens, we show a corresponding message.

Screenshot from 2024-07-09 13-55-34


frontend/src/component/admin/apiToken/CreateApiToken/CreateApiToken.tsx
```diff
@@ -22,17 +25,26 @@ import {
...
+import { Limit } from 'component/common/Limit/Limit';
 
 const pageTitle = 'Create API token';
 interface ICreateApiTokenProps {
     modal?: boolean;
 }
+
+const StyledLimit = styled(Limit)(({ theme }) => ({
+    margin: theme.spacing(2, 0, 4),
+}));
```
thomasheartman, line 37: As we discussed after standup yesterday, I don't love that we style the limit component *for* the form _outside_ the form. However, this form also appears to be different than the others (it's got more children and fewer props), so it may be a worthwhile tradeoff.

thomasheartman (commented):
Nice one! But there's a couple things I think we should address before putting this in. Using [Mateusz's PR](https://github.com/Unleash/unleash/pull/7558/files) #7558 as an example:

For consistency, I think we should stick to the `useXLimit` pattern. We've used this in all the other places. In #7558, it's `useProjectLimit`. Here, it'd be `useApiTokenLimit`. That'd also give you a `limitReached` variable that you can use to disable the button. This takes all the logic of calculating that stuff and sticks it into one place.

I'd like some basic tests to make sure that the component appears and renders as expected (`createProjectDialog.test.tsx` in the other PR).

thomasheartman (approved):
Implementation looks good, so I'll give it an approval, but I'd still like you to get some tests up before merging this.

PR #7564, opened 2024-07-09
Rust tutorial

alvinometric:
I'm no Rustacean, so feel free to change things 😄


website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+-   The URL of your Unleash instance’s API. It’s `[http://localhost:4242/api/](http://localhost:4242/api/)` for your local version. You’ll want to replace it with your remote instance.
+-   The API token we created on our Unleash instance, feel free to create another one if you can’t find it.
+
+Let's first install all the SDK and all its dependencies.
+
+```sh
+cargo add enum-map@2.0.3
+cargo add reqwest --features json
+cargo add serde --features derive
+cargo add tokio --features full
```
sighphyre, line 197: What version of tokio did you use? This doesn't seem to work on 1.38 which is latest in the 1.x range

I'd maybe consider using the threaded example as a base for this: https://github.com/Unleash/unleash-client-rust/blob/main/examples/threads.rs

It's conceptually simpler and doesn't need to deal with Tokio's absolute lack of commitment to stability

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+cargo add unleash-api-client --features reqwest
+```
+
+There are a few dependencies, and here's why: We need an HTTP client to make the request, and `serde` to deserialize the Unleash response. Our SDK is constantly polling Unleash to retrieve the value of feature flags, and caches that in memory. We don't want to block your app while this happens, so we need an async runtime. We're picking `tokio`.
```
sighphyre, line 201: Doesn't need to be async, threads work just fine here. I think it's better to say that the user must choose the nature of their concurrency when using the client - we provide async-std, tokio and threaded features

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+use webp::Encoder as WebPEncoder;
+
+#[derive(Debug, Deserialize, Serialize, Enum, Clone)]
+enum Flags {
+    avif,
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+    let client = client::ClientBuilder::default().into_client::<Flags, reqwest::Client>(
```
sighphyre, line 221: Some liberal use of imports would make this easier to read

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+        "unleash-rust-client-example",
+        Some("default:development.unleash-insecure-api-token".to_string()),
+    )?;
+    client.register().await?;
+
+    let (_, _) = tokio::join!(client.poll_for_updates(), async {
+        // Ensure we have features for this demo.
+        sleep(Duration::from_millis(500)).await;
+
+        let avif = client.is_enabled(Flags::avif, None, false);
```
sighphyre, line 233: No context here. I think that's okay for a demo but I would be a much happier ginger with a comment telling users that they should use one for production

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+
+    Ok(())
+}
+```
+
+## 5. Verify the toggle experience
+
+Now that we’ve connected our project to Unleash and grabbed our feature flag, we can verify that if you disable that flag in your development environment, you should only see the WebP conversion.
+
+> **Note:** An update to a feature flag may take 30 seconds to propagate.
```
sighphyre, line 275: This is within your control - you're setting up the client in the builder, you can just set this to a second for this tutorial (I think that's a nice idea anyway so long as you warn folks it's for demo reasons).

The poll interval in the SDK SDK seems to be 15 seconds though, so, updates shouldn't take longer than that

sighphyre (changes requested):
Ye, pretty cool!

I care a lot that example code works as written though so I think we should fix the places that I've raised that will cause a compile error if we follow the tutorial as written. Other than that, some optional details

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
thomasheartman: Agree with Simon's points here. Setting the poll rate lower for a tutorial seems sensible (and makes it more fun to follow along with when you don't have to wait).

Also, this should be a docusaurus admonition again. Maybe something like this (though it feels a bit **too** wordy (even for me 😅 ))


```suggestion
:::info
Your feature flag configuration will only update as often as your SDK polls Unleash. The default polling interval for the Rust SDK is 15 seconds, but you can set whatever interval works best for your use case.
:::
```

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+## Conclusion
+
+All done! Now you know how to add feature flags with Unleash in Rust. You’ve learned how to:
+
+-   Convert images to WebP and Avif using Rust
+-   Install Unleash and create/enable a feature flag
+-   Initialise the Unleash client and provide it with an async runtime
+-   Grab the value of a feature flag with the Rust SDK
+
+Thank you
```
thomasheartman, line 286: Do we want any sort of punctuation for this last line? 

```suggestion
Thank you!
```

website/sidebars.js
```diff
@@ -153,6 +153,15 @@ module.exports = {
...
                 },
+                {
+                    type: 'category',
+                    label: 'Rust',
+                    link: {
+                        type: 'doc',
+                        id: 'feature-flag-tutorials/rust/implementing-feature-flags-rust',
+                    },
+                    items: [],
+                },
```
thomasheartman, line 164: Why is this a category? Shouldn't this just be a doc, like the iOS one above it?

```suggestion
                {
                    type: 'doc',
                    label: 'Rust',
                    id: 'feature-flag-tutorials/rust/implementing-feature-flags-rust',
                },
```

thomasheartman (commented):
Looks like @sighphyre's got you covered when it comes to the code side of things, so I've focused more on the docs side. It's mostly good, but I've got some questions and suggestions for you to check out. As always, if you disagree with something, let's discuss it 😄

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
sighphyre: Okay I have no idea what I did but this does work now...  maybe you can just ignore me on this one

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+
+Run the code
+
+```sh
+cargo run
+```
+
+You should see another image named `output.webp` in your folder. Make sure that the image is the same and that the compression worked correctly.
+
+![Another image named `output.webp` is now in your folder](./webp-image-in-folder.png)
```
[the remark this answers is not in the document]
sighphyre: Confused by the alt text again! Nice correction Thomas (both to me and the text)

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Prerequisites
+
+For this tutorial, you’ll need the following:
+
+-   Rust
+-   Git
+-   Docker and Docker Compose
+
+![architecture diagram for our implementation](./diagram.png)
```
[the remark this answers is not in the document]
alvinometric: Just like you said in your second paragraph, I found local images to be easier to use. But we've not been great at not reusing local images from other tutorials. 

I don't have a solution right now but I'm sure Melinda will, or have more time to think about it.

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+
+```sh
+cargo add image ravif rgb webp
+```
+
+We use the `image`, `webp` and `ravif` crates to convert images to WebP and Avif, respectively. The `unleash_api_client` crate is used to communicate with the Unleash server. The `tokio` and `reqwest` crates are used to make this connection asynchronous.
+
+Final step before we start coding: Download this image or add an image of your own and call it "input.png." Make sure it's in the same folder.
+
+![The Great Wave off Kanagawa, to be used as an input image for our program](input.png)
```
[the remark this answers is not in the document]
alvinometric: And that there's a boat on that image 😄

website/sidebars.js
alvinometric: This is because the examples #7572 were coming shortly after so I thought I'd make our lives easier

website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -0,0 +1,286 @@
...
+
+We love Rust here at Unleash, our own [Unleash Edge](https://docs.getunleash.io/reference/unleash-edge) is built with Rust and it's a core part of our product.
+
+- [Prerequisites](#prerequisites)
+- [1. Install a local feature flag provider](#1-install-a-local-feature-flag-provider)
+- [2. Convert an image to WebP](#2-convert-an-image-to-webp)
+- [3. Add Avif support](#3-add-avif-support)
+- [4. Add Unleash to your Rust app](#4-add-unleash-to-your-rust-app)
+- [5. Verify the toggle experience](#5-verify-the-toggle-experience)
+- [Conclusion](#conclusion)
```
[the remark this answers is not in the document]
alvinometric: I don't have a good answer apart from "we've always done it" but your point is valid. 

I'm happy to remove it, so I'll let @ferrantim decide

website/sidebars.js
thomasheartman: Ah, right! Yeah, that makes sense. Ignore this, then 😄

nnennandukwe (approved):
👍 excited for this to get out. when Alvin returns, we can do any additional updates to the tutorial that he feels would be good to do based on optional suggestions mentioned.

PR #7566, opened 2024-07-10
task: added flag to remove unsafe inline style src header

chriswk:
Our CSP reports that unsafe-inline is not recommended for styleSrc. This PR adds a flag for making it possible to remove this element of our CSP headers. It should allow us to see what (if anything) breaks hard.


nunogois (approved):
LGTM. Just to be clear, this only works at startup, right?

chriswk:
Well, it gets added to a middleware, so it might actually work on the fly

PR #7567, opened 2024-07-10
docs: document resource limits

thomasheartman:
This PR adds the first version of the resource limit documentation. It also corrects the link in the `Limit` component.

In addition to the new limits from the soft limits project, I've also added the limits from the signals project.

## Discussion points

I've set the Pro resource limits to projects and envs to the same value as on the [plans & pricing](https://www.getunleash.io/pricing) page instead of the actual hard soft limit. I think that makes messaging easier and more consistent. 

However, this does present us with a bit of a problem. Because we've been pretty consistent at saying "no, you can't have more projects" for Pro. But we're saying it's easy to change (when in actuality the limit isn't enforced 💁🏼 )

## Screenie

It looks a little bit like this (but check the files or preview for the actual text content)
image


frontend/src/component/common/Limit/Limit.tsx
```diff
@@ -127,7 +127,9 @@ export const Limit: FC<{
                 <LimitExplanation>
                     <Link
                         target='_blank'
-                        to={'https://docs.getunleash.io/reference/limits'}
+                        to={
```
kwasniew, line 130: we can also drop {} since it's a string literal

kwasniew (approved):

frontend/src/component/common/Limit/Limit.tsx
thomasheartman: Ah, yeah, good point. Done in afa00294426d01b82c1a2e2b177cb2b2e744fa9c

PR #7568, opened 2024-07-10
feat: statistics for orphaned tokens

Tymek:
## About the changes

1. **Token Validation Updates:**
   - Modified `createTokenRowReducer` to exclude tokens in v1 format from orphaned tokens. Those where created before the concept of project and environment and are always wildcard tokens.

2. **Metrics Monitoring:**
   - Added `countDeprecatedTokens` to count:
     - `orphanedTokens` - tokens that cause issues, because of wildcard access
     - `activeOrphanedTokens` (within last 3 months)
     - `legacyTokens` - tokens in v1 format - metric added in case we would like to deprecate or migrate it couple of major version in the future
     - `activeLegacyTokens` (within last 3 months)

New Prometheus fields are:
`orphaned_api_tokens_total`, `orphaned_api_tokens_active`, `legacy_api_tokens_total`, `legacy_api_tokens_active`

3. **Feature Flags Adjustment:**
   - Changed `allowOrphanedWildcardTokens` kill switch default to `true`.

### DIFF
--- a/src/lib/db/api-token-store.ts
+++ b/src/lib/db/api-token-store.ts
@@ -278,16 +278,25 @@ export class ApiTokenStore implements IApiTokenStore {
         legacyTokens: number;
         activeLegacyTokens: number;
     }> {
-        const allLegacy = await this.makeTokenProjectQuery()
-            .whereNull('token_project_link.project')
-            .andWhere('tokens.secret', 'NOT LIKE', '%:%');
+        const allLegacyCount = await this.db<ITokenRow>(`${TABLE} as tokens`)
+            .where('tokens.secret', 'NOT LIKE', '%:%')
+            .count()
+            .first()
+            .then((res) => Number(res?.count));
 
-        const activeLegacy = await this.makeTokenProjectQuery()
-            .whereNull('token_project_link.project')
-            .andWhere('tokens.secret', 'NOT LIKE', '%:%')
-            .andWhereRaw("tokens.seen_at > NOW() - INTERVAL '3 MONTH'");
+        const activeLegacyCount = await this.db<ITokenRow>(`${TABLE} as tokens`)
+            .where('tokens.secret', 'NOT LIKE', '%:%')
+            .andWhereRaw("tokens.seen_at > NOW() - INTERVAL '3 MONTH'")
+            .count()
+            .first()
+            .then((res) => Number(res?.count));
 
-        const orphanedTokensQuery = this.makeTokenProjectQuery()
+        const orphanedTokensQuery = this.db<ITokenRow>(`${TABLE} as tokens`)
+            .leftJoin(
+                `${API_LINK_TABLE} as token_project_link`,
+                'tokens.secret',
+                'token_project_link.secret',
+            )
             .whereNull('token_project_link.project')
             .andWhere('tokens.secret', 'NOT LIKE', '*:%') // Exclude intentionally wildcard tokens
             .andWhere('tokens.secret', 'LIKE', '%:%') // Exclude legacy tokens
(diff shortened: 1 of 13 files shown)


src/lib/features/project/project-service.e2e.test.ts
```diff
@@ -82,7 +82,9 @@ beforeAll(async () => {
     const config = createTestConfig({
         getLogger,
         experimental: {
-            flags: {},
+            flags: {
+                cleanApiTokenWhenOrphaned: true,
```
Tymek, line 86: That's fun. Killswitch being disabled by default was masking that tests related to other flag don't pass.

src/lib/db/api-token-store.ts
```diff
@@ -270,4 +271,43 @@ export class ApiTokenStore implements IApiTokenStore {
...
+            });
+
+        const allOrphaned = await orphanedTokensQuery.clone();
+
+        const activeOrphaned = await orphanedTokensQuery
+            .clone()
+            .andWhereRaw("tokens.seen_at > NOW() - INTERVAL '3 MONTH'");
+
+        return {
+            orphanedTokens: allOrphaned.length,
```
kwasniew, line 307: why do we read all tokens into memory and calculating length instead of the count query in SQL?

src/lib/db/api-token-store.ts
```diff
@@ -270,4 +271,43 @@ export class ApiTokenStore implements IApiTokenStore {
...
+        orphanedTokens: number;
+        activeOrphanedTokens: number;
+        legacyTokens: number;
+        activeLegacyTokens: number;
+    }> {
+        const allLegacy = await this.makeTokenProjectQuery()
+            .whereNull('token_project_link.project')
+            .andWhere('tokens.secret', 'NOT LIKE', '%:%');
+
+        const activeLegacy = await this.makeTokenProjectQuery()
```
kwasniew, line 285: can we run those queries in parallel instead of sequentially?

kwasniew: or maybe we can even join all those queries into one count query similar to what you did in createGetActiveUsers with CASE WHEN?

src/test/e2e/stores/api-token-store.e2e.test.ts
```diff
@@ -35,3 +39,146 @@ test('get token returns the token when exists', async () => {
...
 });
+
+describe('count deprecated tokens', () => {
+    test('should return 0 if there is no legacy or orphaned tokens', async () => {
+        await stores.projectStore.create({
+            id: 'test',
+            name: 'test',
+        });
+        await stores.apiTokenStore.insert({
+            secret: '*:*.be44368985f7fb3237c584ef86f3d6bdada42ddbd63a019d26955178',
```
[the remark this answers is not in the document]
Tymek: Almost all fields can count - for example tokens of ADMIN type can't be orphaned, because it doesn't have projects. At the same time it can be in v1 or v2 format. I'm providing a variety of possible _correct_ tokens.

src/lib/openapi/spec/instance-admin-stats-schema.ts
```diff
@@ -227,6 +227,40 @@ export const instanceAdminStatsSchema = {
             example:
                 'b023323477abb1eb145bebf3cdb30a1c2063e3edc1f7ae474ed8ed6c80de9a3b',
         },
+        deprecatedTokens: {
```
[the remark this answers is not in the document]
Tymek: 👍🏻  I'll get metrics from store, skipping instance stats

kwasniew (approved):

PR #7569, opened 2024-07-10
feat: show orphaned API tokens

Tymek:
## About the changes
Add a visual indication that a token was scoped to projects that have been deleted.

image

### DIFF
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
@@ -5,7 +5,7 @@ import { HtmlTooltip } from 'component/common/HtmlTooltip/HtmlTooltip';
 import { LinkCell } from 'component/common/Table/cells/LinkCell/LinkCell';
 import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
 import { useSearchHighlightContext } from 'component/common/Table/SearchHighlightContext/SearchHighlightContext';
-import ErrorIcon from '@mui/icons-material/ReportGmailerrorredRounded';
+import ErrorIcon from '@mui/icons-material/Error';
 import { Link } from 'react-router-dom';
 
 const StyledLink = styled(Link)(({ theme }) => ({
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
@@ -95,10 +95,7 @@ export const ProjectsList: FC<IProjectsListProps> = ({
                 title={
                     isOrphanedToken ? (
                         <>
-                            ALL current and future projects. This is an orphaned
-                            token with it's project deleted. This token gives
-                            more access than it was intended to and should be
-                            deleted.
+                            This is an orphaned token. All of its original projects have been deleted and it now has access to all current and future projects. You should stop using this token and delete it. It will lose access to all projects at a later date.
                         </>
                     ) : (
                         'ALL current and future projects.'
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
@@ -95,7 +95,11 @@ export const ProjectsList: FC<IProjectsListProps> = ({
                 title={
                     isOrphanedToken ? (
                         <>
-                            This is an orphaned token. All of its original projects have been deleted and it now has access to all current and future projects. You should stop using this token and delete it. It will lose access to all projects at a later date.
+                            This is an orphaned token. All of its original
+                            projects have been deleted and it now has access to
+                            all current and future projects. You should stop
+                            using this token and delete it. It will lose access
+                            to all projects at a later date.
                         </>
                     ) : (
                         'ALL current and future projects.'
--- a/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
+++ b/frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
@@ -7,6 +7,7 @@ import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
 import { useSearchHighlightContext } from 'component/common/Table/SearchHighlightContext/SearchHighlightContext';
 import ErrorIcon from '@mui/icons-material/Error';
 import { Link } from 'react-router-dom';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 
 const StyledLink = styled(Link)(({ theme }) => ({
     textDecoration: 'none',
@@ -30,15 +31,13 @@ const StyledContainer = styled('div')({
 interface IProjectsListProps {
     project?: string;
     projects?: string | string[];
-    isWildcard?: boolean;
-    isLegacy?: boolean;
+    secret?: string;
 }
 
 export const ProjectsList: FC<IProjectsListProps> = ({
     projects,
     project,
-    isWildcard,
-    isLegacy,
+    secret,
 }) => {
     const { searchQuery } = useSearchHighlightContext();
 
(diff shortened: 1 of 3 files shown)


frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
```diff
@@ -1,10 +1,11 @@
+import { Fragment, type FC } from 'react';
 import { styled } from '@mui/material';
 import { Highlighter } from 'component/common/Highlighter/Highlighter';
 import { HtmlTooltip } from 'component/common/HtmlTooltip/HtmlTooltip';
 import { LinkCell } from 'component/common/Table/cells/LinkCell/LinkCell';
 import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
 import { useSearchHighlightContext } from 'component/common/Table/SearchHighlightContext/SearchHighlightContext';
-import { Fragment, type FC } from 'react';
+import ErrorIcon from '@mui/icons-material/ReportGmailerrorredRounded';
```
kwasniew, line 8: I've never seen this icons before in our code. Don't we have an established icon for warnings or errors (if we think it's an error)?

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
```diff
@@ -67,16 +86,40 @@ export const ProjectsList: FC<IProjectsListProps> = ({ projects, project }) => {
...
-                title='ALL current and future projects'
-                placement='bottom'
-                arrow
-            >
-                <span>
+            <HtmlTooltip title={getTitle()} placement='bottom' arrow>
+                <StyledContainer>
                     <Highlighter search={searchQuery}>*</Highlighter>
-                </span>
+                    {!isWildcard && !isLegacy && <StyledErrorIcon />}
```
kwasniew, line 121: IMO this deserves a test

frontend/src/component/common/ApiTokenTable/useApiTokenTable.tsx
```diff
@@ -61,6 +61,10 @@ export const useApiTokenTable = (
                     <ProjectsList
                         project={props.row.original.project}
                         projects={props.row.original.projects}
+                        isWildcard={props.row.original.secret?.startsWith?.(
```
kwasniew, line 64: in the backend code we had a concept of orphaned and legacy tokens. Here we use difference vocabulary of legacy and wildcard. What's the reason for that?
Also isX, isY leads to the boolean flags explosion. Can we introduce a union type to represent secret types? e.g. `type SecretType = 'legacy' | 'orphaned'`

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
```diff
@@ -67,16 +86,40 @@ export const ProjectsList: FC<IProjectsListProps> = ({ projects, project }) => {
...
+        }
+
+        if (isLegacy) {
+            return (
+                <>
+                    ALL current and future projects. This token has v1 format.
+                    Read more about{' '}
+                    <a href='https://docs.getunleash.io/reference/api-tokens-and-client-keys#format'>
+                        token formats
+                    </a>
```
thomasheartman, line 107: I'm not sure I understand why we're showing this. Is it important for the user to know? Also, why should they go and read about the formats? The text doesn't seem actionable.

frontend/src/component/common/ApiTokenTable/useApiTokenTable.tsx
thomasheartman: I agree mostly with this assessment. More than just the number of flags, we also gotta account for what all the combinations mean (legacy && wildcard = ?, legacy && !wildcard = v1, !legacy && wildcard = v2 all projects, !legacy && !wildcard = v2 scoped).

But I'm not sure that `secretType` is the right thing here. Because `orphaned` isn't a secret type. But if we put something like `tokenVersion = v1 | v2` in, then we should be able to take it from there, right? 

if it's v1, it's legacy.
If it's v2, check if it's wildcard or not, etc?

But I'll admit, I'm not quite sure I understand what the payload is for orphaned tokens. @Tymek , can you show me what the payload looks like for:
- legacy tokens
- project scoped v2 tokens
- all project v2 tokens
- orphaned tokens

kwasniew: Yup, doesn't have to be SecretType. My main point is that as a reader of the code partially familiar with this part of domain I expect union/list of variants I should be aware of and to have a unit test documenting each of those variants in action.

Tymek:
I decided to move more logic into the component - this makes it way easier to test and understand what's going on.

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
```diff
@@ -1,11 +1,13 @@
+import { Fragment, type FC } from 'react';
 import { styled } from '@mui/material';
 import { Highlighter } from 'component/common/Highlighter/Highlighter';
 import { HtmlTooltip } from 'component/common/HtmlTooltip/HtmlTooltip';
 import { LinkCell } from 'component/common/Table/cells/LinkCell/LinkCell';
 import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
 import { useSearchHighlightContext } from 'component/common/Table/SearchHighlightContext/SearchHighlightContext';
-import { Fragment, type FC } from 'react';
+import ErrorIcon from '@mui/icons-material/Error';
```
thomasheartman, line 8: Is an error icon the right icon to show here? Wouldn't a warning be more appropriate? 🤔

frontend/src/component/admin/apiToken/CreateApiToken/CreateApiToken.tsx
```diff
@@ -24,9 +24,11 @@ import {
 } from '@server/types/permissions';
 
 const pageTitle = 'Create API token';
+
```
kwasniew, line 27: not sure what happened here :)

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
```diff
@@ -67,16 +85,35 @@ export const ProjectsList: FC<IProjectsListProps> = ({ projects, project }) => {
...
         <TextCell>
             <HtmlTooltip
-                title='ALL current and future projects'
+                title={
+                    isOrphanedToken ? (
+                        <>
+                            ALL current and future projects. This is an orphaned
+                            token with it's project deleted. This token gives
+                            more access than it was intended to and should be
+                            deleted.
```
thomasheartman, line 101: I wonder if we can rephrase this somehow. Because from the way it sounds now, it seems like it's a bug in Unleash (which it is), but it also sounds like we're not going to fix it.

I wonder if we can phrase this as a warning of sorts - like "it works now, but it'll stop working in the future"?

I don't know. Something like:

```suggestion
                            This is an orphaned token. All of its original projects have been deleted and it now has access to all current and future projects. You should stop using this token and delete it. It will lose access to all projects at a later date.
```

frontend/src/component/admin/apiToken/CreateApiToken/CreateApiToken.tsx
Tymek: I think I run code formatting on another file, sorry 😅

kwasniew (approved):
Agree, so much cleaner.

frontend/src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
Tymek: I don't want this to be ignored. Tokens like this, in theory, leak data.

thomasheartman: Fair enough. There's nothing preventing you from just ignoring it (for now), though. But I'm easy either way; do whatever you feel is best ☺️

thomasheartman (approved):

PR #7570, opened 2024-07-10
fix: project name ellipsis overflow for long names

kwasniew:
## About the changes

Make long project names look good (ellipsis overflow) on all screens.

Screenshot 2024-07-10 at 15 54 39
Screenshot 2024-07-10 at 15 54 56
Screenshot 2024-07-10 at 15 55 05

### DIFF
--- a/frontend/src/component/project/Project/Project.styles.ts
+++ b/frontend/src/component/project/Project/Project.styles.ts
@@ -1,8 +1,12 @@
 import { styled, Tab } from '@mui/material';
 import { FavoriteIconButton } from 'component/common/FavoriteIconButton/FavoriteIconButton';
 
-export const StyledDiv = styled('div')(() => ({
+export const StyledDiv = styled('div')(({ theme }) => ({
     display: 'flex',
+    maxWidth: '90%',
+    [theme.breakpoints.down('md')]: {
+        maxWidth: '70%',
+    },
 }));
 
 export const StyledTopRow = styled('div')(() => ({
@@ -20,6 +24,7 @@ export const StyledColumn = styled('div')(() => ({
 export const StyledName = styled('span')(({ theme }) => ({
     fontSize: theme.typography.h1.fontSize,
     textOverflow: 'ellipsis',
+    overflow: 'hidden',
     whiteSpace: 'nowrap',
 }));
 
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):
This is probably the safest (and easiest) option  😄 

I'm happy to approve it, but a few thoughts:
- if we shorten the name here, where can you see the full project name in all its glory? Is there anywhere we don't truncate it?
- if not, should we introduce a name length limit in the next major, perhaps?
- As an option, have we considered just letting the project name wrap around to the next line? With edge cases, I guess you could end up with a full screen of just project name, so probably not the safest option.
- As a different option, have we considered making the name scrollable if it's too long? That might be neat, because we can keep it on one line, but you have the option of seeing the full name (if only in pieces)

None of these are hard requirements or anything; I'm just thinking out loud. I think we should merge this for now, and if we want to revisit it later, we can. However, names this long are most likely an edge case, so it's probably not worth spending too much time on.

frontend/src/component/project/Project/Project.styles.ts
```diff
@@ -1,8 +1,12 @@
 import { styled, Tab } from '@mui/material';
 import { FavoriteIconButton } from 'component/common/FavoriteIconButton/FavoriteIconButton';
 
-export const StyledDiv = styled('div')(() => ({
+export const StyledDiv = styled('div')(({ theme }) => ({
     display: 'flex',
+    maxWidth: '90%',
+    [theme.breakpoints.down('md')]: {
+        maxWidth: '70%',
+    },
```
thomasheartman, line 9: Where do you get the percentages from? Have you manually checked with the other elements? Wouldn't it be easier to make the top row a flex container and set its width to 100%? Flex grow the middle and voila?

thomasheartman (approved):

frontend/src/component/project/Project/Project.styles.ts
kwasniew: that's what we had before and text-ellipsis didn't work. I think it always adds the ellipsis when there's width or max-width

kwasniew: I tested this visually to make sure it looks good on all screen sizes and found by trial and error that on larger devices 90% is fine while on smaller ones it's 70%

kwasniew:
@thomasheartman you can see full project id in the breadcrumbs. It contains the name and hash suffix. Also you have full name in the project settings form. I doubt we're optimizing for a very realistic use case here so unless we hear someone complaining about their 80 character project name breaking I'm fine with the ellipsis.

kwasniew:
https://github.com/Unleash/unleash/pull/7575 is more elegant

PR #7572, opened 2024-07-10
Rust examples 🦀

alvinometric:
And now the examples


alvinometric:
The build is breaking because it expects #7564 to exist first

nnennandukwe:
> The build is breaking because it expects #7564 to exist first

Reviewing post-Rust tutorial merge.

nnennandukwe (approved):
Think you hit all the major points here! 👍

PR #7575, opened 2024-07-11
chore: fix project name overflow

thomasheartman:
Potential way to fix it using grid

image

image

image


kwasniew:
I like it more than https://github.com/Unleash/unleash/pull/7570

kwasniew (approved):

PR #7583, opened 2024-07-12
feat: user seats component

kwasniew:
## About the changes

Creating self-contained component for user seats.

Screenshot 2024-07-12 at 14 41 13

Decisions:
* this component is hosted on the insights page but since it's not affected by the date range selection I decided to not add this data to the insights endpoint
* instead this components fetches the data it needs from the existing APIs
* I still need to figure out if this is the right way to fetch the user count and the seats count
* I'm calling it user seats not user licenses since I want consistent terms across all pages


thomasheartman (approved):

Tymek:
🚀

PR #7590, opened 2024-07-15
feat: show all results in the same time

sjaanus:
Previously, when the result box was loading, it returned projects and menu items first. After the feature search response came back, it also showed the features, but this made the component jump around too much. Now, everything is shown when the feature result comes back, reducing the jumping around.

### DIFF
--- a/frontend/src/component/commandBar/CommandBar.tsx
+++ b/frontend/src/component/commandBar/CommandBar.tsx
@@ -97,6 +97,7 @@ export const CommandBar = () => {
     const searchInputRef = useRef<HTMLInputElement>(null);
     const searchContainerRef = useRef<HTMLInputElement>(null);
     const [showSuggestions, setShowSuggestions] = useState(false);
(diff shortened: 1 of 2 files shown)


frontend/src/component/commandBar/CommandBar.tsx
```diff
@@ -215,39 +216,54 @@ export const CommandBar = () => {
...
+            allCommandBarLinks,
+            selectedIndex,
+        };
     };
 
-    useKeyboardShortcut({ key: 'ArrowDown', preventDefault: true }, () => {
-        const itemsAndIndex = findCommandBarLinksAndSelectedIndex();
-        if (!itemsAndIndex) return;
-        const { allCommandBarLinks, selectedIndex } = itemsAndIndex;
+    useKeyboardShortcut(
```
sjaanus, line 225: This whole section seems to be biome. I think biome was updated last week.

thomasheartman (approved):

PR #7597, opened 2024-07-16
feat: make frontend aware that OIDC can be configured through env

chriswk:


frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -21,6 +21,7 @@ import { formatUnknownError } from 'utils/formatUnknownError';
 import { removeEmptyStringFields } from 'utils/removeEmptyStringFields';
 import { SsoGroupSettings } from '../SsoGroupSettings';
 import type { IRole } from 'interfaces/role';
+import { ConditionallyRender } from '../../../common/ConditionallyRender/ConditionallyRender';
```
nunogois, line 24: We usually prefer this:
```suggestion
import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
```

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -45,10 +46,10 @@ type State = typeof initialState & {
 export const OidcAuth = () => {
     const { setToastData, setToastApiError } = useToast();
     const { uiConfig } = useUiConfig();
+    const { oidcLocked } = uiConfig;
     const [data, setData] = useState<State>(initialState);
     const { config } = useAuthSettings('oidc');
     const { updateSettings, errors, loading } = useAuthSettingsApi('oidc');
-
     useEffect(() => {
```
nunogois, line 53: Poor empty line got yeeted, but I think it helps with readability / separation of logic blocks 😄 
```suggestion
    
    useEffect(() => {
```

frontend/src/interfaces/uiConfig.ts
```diff
@@ -30,6 +30,7 @@ export interface IUiConfig {
     strategySegmentsLimit?: number;
     frontendApiOrigins?: string[];
     resourceLimits: ResourceLimitsSchema;
+    oidcLocked?: boolean;
```
nunogois, line 33: Now that I think about it, maybe we should be a bit more specific about this name. It doesn't really tell us why OIDC is locked. How about something like `oidcConfiguredThroughEnv`?

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
```diff
@@ -107,6 +108,24 @@ export const OidcAuth = () => {
         <>
             <Grid container sx={{ mb: 3 }}>
                 <Grid item md={12}>
+                    <ConditionallyRender
+                        condition={oidcLocked === true}
```
nunogois, line 112: No strong feelings either way, just an alternative.
```suggestion
                        condition={Boolean(oidcLocked)}
```

nunogois (approved):
LGTM!

I think maybe we should rethink the variable name to be more explicit / descriptive of what's really going on. 

Other than that, I added some small comments that are not very significant, so feel free to ignore them.

frontend/src/component/admin/auth/OidcAuth/OidcAuth.tsx
chriswk: Yeah, auto imports. will update

frontend/src/interfaces/uiConfig.ts
chriswk: Took your suggestion. It made sense to me. Updated.

PR #7601, opened 2024-07-16
fix: toast error doesn't tell you what the error is

thomasheartman:
This change improves the toast error message for auth settings by
preferring the message from the error's details list if it exists. If
it doesn't it'll still fall back to the original error message.

Before:
image

After:
image

### DIFF
--- a/frontend/src/hooks/api/actions/useAuthSettingsApi/useAuthSettingsApi.ts
+++ b/frontend/src/hooks/api/actions/useAuthSettingsApi/useAuthSettingsApi.ts
@@ -12,8 +12,9 @@ export const handleBadRequest = async (
     if (!setErrors) return;
     if (res) {
         const data = await res.json();
-        setErrors({ message: data.message });
-        throw new Error(data.message);
+        const message = data.details?.[0]?.message ?? data.message;
+        setErrors({ message });
+        throw new Error(message);
     }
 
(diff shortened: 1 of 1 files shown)


frontend/src/hooks/api/actions/useAuthSettingsApi/useAuthSettingsApi.ts
```diff
@@ -12,8 +12,9 @@ export const handleBadRequest = async (
     if (!setErrors) return;
     if (res) {
         const data = await res.json();
-        setErrors({ message: data.message });
-        throw new Error(data.message);
+        const message = data.details?.[0]?.message ?? data.message;
+        setErrors({ message });
+        throw new Error(message);
```
kwasniew, line 17: I know we're keeping this behavior but it's weird that we handle setErrors and then re-throw again

kwasniew (approved):

frontend/src/hooks/api/actions/useAuthSettingsApi/useAuthSettingsApi.ts
thomasheartman: Yeah, I think it's because this one sets the error in the form (not the toast), but then we catch the second error and use that to populate the toast. But yaeh, not looking at that now 😅

PR #7609, opened 2024-07-17
fix: Update OpenAPI error converter to handle query param errors too

thomasheartman:
This PR updates the OpenAPI error converter to also work for errors with query parameters. 
We previously only sent the body of the request along with the error, which meant that query parameter errors would show up incorrectly.

For instance given a query param with the date format and the invalid value `01-2020-01`, you'd previously get the message:
> The `from` value must match format "date". You sent undefined

With this change, you'll get this instead:
> The `from` value must match format "date". You sent "01-2020-01". 

The important changes here are two things:
- passing both request body and query params
- the 3 lines in `fromOpenApiValidationError` that check where we should get the value you sent from.

The rest of it is primarily updating tests to send the right arguments and some slight rewording to more accurately reflect that this can be either request body or query params.

### DIFF
--- a/src/lib/error/bad-data-error.ts
+++ b/src/lib/error/bad-data-error.ts
@@ -44,7 +44,7 @@ const missingRequiredPropertyMessage = (
     missingPropertyName: string,
 ) => {
     const path = constructPath(pathToParentObject, missingPropertyName);
-    const message = `The \`${path}\` property is required. It was not present on the data you sent.`;
+    const message = `The \`${path}\` value is required. It was not present on the data you sent.`;
     return {
         path,
         message,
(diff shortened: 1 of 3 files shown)


src/lib/error/bad-data-error.ts
```diff
@@ -117,10 +117,14 @@ const enumMessage = (
...
-    (requestBody: object) =>
+    (request: { body: object; query: object }) =>
     (validationError: ErrorObject): ValidationErrorDescription => {
         const { instancePath, params, message } = validationError;
-        const propertyName = instancePath.substring('/body/'.length);
+        const [errorSource, substringOffset] = instancePath.startsWith('/body')
+            ? [request.body, '/body/'.length]
+            : [request.query, '/query/'.length];
+
+        const propertyName = instancePath.substring(substringOffset);
```
thomasheartman, line 127: This (along with also passing in the request query) is the real meat of this PR. With this little change, we handle query params as well.

kwasniew (approved):

PR #7610, opened 2024-07-17
Insights layout

Tymek:
## About the changes
Refactored insights page - stats and charts relevant to the same metric are now combined into a single widget.

image


frontend/src/component/insights/components/WidgetTitle/WidgetTitle.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const WidgetTitle: FC<{
+    title: ReactNode;
+    tooltip?: ReactNode;
+}> = ({ title, tooltip }) => (
+    <>
+        <Typography
+            variant='h3'
+            sx={(theme) => ({
+                marginBottom: theme.spacing(3),
```
thomasheartman, line 15: Why does this component control its own margins? Isn't that better handled by the parent?

thomasheartman: I noticed where we use it, the parent also sets up a flag gap. Seems like we're doubling up on spacing for some reason?

frontend/src/component/insights/components/WidgetTitle/WidgetTitle.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Typography } from '@mui/material';
+import { HelpIcon } from 'component/common/HelpIcon/HelpIcon';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
+import InfoOutlined from '@mui/icons-material/InfoOutlined';
+
+export const WidgetTitle: FC<{
+    title: ReactNode;
+    tooltip?: ReactNode;
+}> = ({ title, tooltip }) => (
+    <>
```
thomasheartman, line 11: This is only a single component, so we can get rid of the wrappers


```suggestion
```

frontend/src/component/insights/components/WidgetTitle/WidgetTitle.tsx
```diff
@@ -0,0 +1,32 @@
...
+            <ConditionallyRender
+                condition={Boolean(tooltip)}
+                show={
+                    <HelpIcon htmlTooltip tooltip={tooltip}>
+                        <InfoOutlined />
+                    </HelpIcon>
+                }
+            />
+        </Typography>
+    </>
```
thomasheartman, line 31:
```suggestion
```

thomasheartman (approved):
Nice work! Looking at the screenie you added, I thought this might be a good use case for subgrid, to make all the left-hand stats be the same width. But then I checked the figma sketches and realized that none of them are the same width 😄 

It would be nice at some point to look at how these widgets fold together. Here's Firefox simulating a phone screen. Notice how all the graphs are so small that they're pretty useless.  
image

But I realized we already do this today, so it shouldn't be a blocker.
image

PR #7611, opened 2024-07-17
feat: show info on healthy flags in health tooltip

thomasheartman:
This PR updates the tooltips for the health chart to also include information on how healthy  flags there are. The user could make this calculation themselves before, but it'd require them to subtract the sum of stale and potentially stale flags from the total. This makes it so that they don't have to do the calculation.

I've also included a bar for the healthy flags in the overview, so that it's easier to see how large a portion it is compared to the others.

Also: clean up some uses of the now-deprecated VFC.

image

### DIFF
--- a/frontend/src/component/insights/componentsChart/ProjectHealthChart/ProjectHealthChart.tsx
+++ b/frontend/src/component/insights/componentsChart/ProjectHealthChart/ProjectHealthChart.tsx
@@ -1,5 +1,5 @@
 import 'chartjs-adapter-date-fns';
-import { useMemo, type VFC } from 'react';
+import { type FC, useMemo } from 'react';
 import type { InstanceInsightsSchema } from 'openapi';
 import { HealthTooltip } from './HealthChartTooltip/HealthChartTooltip';
 import { useProjectChartData } from 'component/insights/hooks/useProjectChartData';
@@ -20,7 +20,7 @@ interface IProjectHealthChartProps {
     isLoading?: boolean;
 }
 
-export const ProjectHealthChart: VFC<IProjectHealthChartProps> = ({
+export const ProjectHealthChart: FC<IProjectHealthChartProps> = ({
     projectFlagTrends,
     isAggregate,
     isLoading,
--- a/frontend/src/component/insights/componentsChart/ProjectHealthChart/ProjectHealthChart.tsx
+++ b/frontend/src/component/insights/componentsChart/ProjectHealthChart/ProjectHealthChart.tsx
@@ -46,7 +46,8 @@ export const ProjectHealthChart: FC<IProjectHealthChartProps> = ({
                         (acc, item) => {
                             if (item) {
                                 acc.total += item.total;
-                                acc.stale += item.stale + item.potentiallyStale;
+                                acc.stale += item.stale;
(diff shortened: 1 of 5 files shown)


frontend/src/component/insights/componentsChart/ProjectHealthChart/ProjectHealthChart.tsx
```diff
@@ -73,12 +76,17 @@ export const ProjectHealthChart: VFC<IProjectHealthChartProps> = ({
                     data: weeks.map((item) => ({
                         health: item.total
                             ? (
-                                  ((item.total - item.stale) / item.total) *
+                                  ((item.total -
```
kwasniew, line 79: don't we get this value from the backend? Also is it how we calculate it in the backend?

kwasniew: if we decide to keep it here I'd extract the equation into a small fn so that we don't have those deeply nested calculations that line up vertically due to spacing constraints in our linter/formatter

thomasheartman: Good questions! I don't know about the first one, but I can find out. And yeah, I'm happy to extract it. Was considering that for testing anyway, so

thomasheartman: You're right. We **do** get this value from the backend. Except it's not a required property on the object, so it can be undefined. However, I think that might be an oversight. It appears to build fine locally with the marking and I've submitted a [PR to see if it'll work against all the tests](https://github.com/ivarconr/unleash-enterprise/pull/1457) etc.

It does appear to be the same calculation that we do in the back end, though I haven't been able to find where we calculate it on the project-level when we insert the data (do you know?).

But that said, I don't think we can use the health value here because it doesn't give us the correct result. The items here are projects, right? So imagine this:

You have two projects, A and B. 
- A has 100 flags, of which all are stale. That gives it a health rating of 0.
- B has 1 flag, which is not stale. Its health rating is 100. 

Now, if you were to record each project's health rating, sum it up, and divide that by the number of projects ((0 +100) / 2), you'd get a 50% health rating. That doesn't sound like the correct health for the instance to me. That's the average of the projects

So if you instead tally up the total flags and run the calc we do today, you'd get (101-100) / 101 * 100, which is ~1%. That sounds more correct to me.

The problem is that if we just add the health from each project, we lose information on how many flags are in each project. I remember Andreas talking about this and adjusting the calculation based on that. If we were to use each individual item's health here, we'd have to weight them based on the number of flags in each project.

---

And if you're thinking "but Thomas, if you calculate health as (sum health) / (total flags), then it looks right in your example". And yeah, 100/101 ~= 1% (in fact, it's 0.99, which is also what the other calculation is).

Let's add a project C, then. 100 flags, 25 stale. Health = 75%.

Using health values we get (0 + 100 + 75) / 201 ~= 0.87.

Using the current approach, we get (201 - 100 - 25) / 201 * 100 ~= 37.8

So ... I don't think that works. I'm not about to write a proof of why it doesn't, but I think you get the gist. (Though if you think I'm wrong, please **do** tell me, because I'm definitely invested now 😄 )

PR #7614, opened 2024-07-17
chore: integration events service

nunogois:
https://linear.app/unleash/issue/2-2438/create-new-integration-event-service
https://linear.app/unleash/issue/2-2442/automatically-clean-up-old-integration-events

Adds a new `IntegrationEventsService`.

### DIFF
--- a/src/lib/features/integration-events/integration-events-service.ts
+++ b/src/lib/features/integration-events/integration-events-service.ts
@@ -25,10 +25,6 @@ export class IntegrationEventsService {
         this.logger = getLogger('integration-events-service');
     }
 
-    isEnabled(): boolean {
-        return this.flagResolver.isEnabled('integrationEvents');
-    }
-
     async getPaginatedEvents(
         id: number,
         limit: number,
@@ -43,14 +39,12 @@ export class IntegrationEventsService {
 
     async registerEvent(
         integrationEvent: IntegrationEventWriteModel,
-    ): Promise<IntegrationEventSchema | undefined> {
-        if (!this.isEnabled()) return;
-
+    ): Promise<IntegrationEventSchema> {
         return this.integrationEventsStore.insert(integrationEvent);
     }
 
     async cleanUpEvents(): Promise<void> {
-        if (!this.isEnabled()) return;
+        if (!this.flagResolver.isEnabled('integrationEvents')) return;
 
         await this.integrationEventsStore.cleanUpEvents();
     }
--- a/src/lib/features/integration-events/integration-events-service.ts
+++ b/src/lib/features/integration-events/integration-events-service.ts
@@ -1,0 +1,57 @@
+import type { Logger } from '../../logger';
+import type { IFlagResolver, IUnleashConfig } from '../../types';
+import type {
+    IntegrationEventsStore,
+    IntegrationEventWriteModel,
+} from './integration-events-store';
+import type { IntegrationEventSchema } from '../../openapi/spec/integration-event-schema';
+
+export class IntegrationEventsService {
+    private readonly logger: Logger;
+    private integrationEventsStore: IntegrationEventsStore;
+    private flagResolver: IFlagResolver;
+
+    constructor(
+        {
+            integrationEventsStore,
+        }: { integrationEventsStore: IntegrationEventsStore },
+        {
+            getLogger,
+            flagResolver,
+        }: Pick<IUnleashConfig, 'getLogger' | 'flagResolver'>,
+    ) {
+        this.integrationEventsStore = integrationEventsStore;
+        this.flagResolver = flagResolver;
+        this.logger = getLogger('integration-events-service');
+    }
+
+    isEnabled(): boolean {
+        return this.flagResolver.isEnabled('integrationEvents');
+    }
+
+    async getPaginatedEvents(
+        id: number,
+        limit: number,
+        offset: number,
+    ): Promise<IntegrationEventSchema[]> {
+        return this.integrationEventsStore.getPaginatedEvents(
+            id,
+            limit,
+            offset,
+        );
+    }
+
+    async registerEvent(
+        integrationEvent: IntegrationEventWriteModel,
+    ): Promise<IntegrationEventSchema | undefined> {
+        if (!this.isEnabled()) return;
+
+        return this.integrationEventsStore.insert(integrationEvent);
+    }
+
+    async cleanUpEvents(): Promise<void> {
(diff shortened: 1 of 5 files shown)


src/lib/features/integration-events/integration-events.e2e.test.ts
```diff
@@ -5,14 +5,12 @@ import {
 } from '../../../test/e2e/helpers/test-helper';
 import getLogger from '../../../test/fixtures/no-logger';
 import { TEST_AUDIT_USER } from '../../types';
-import type {
-    IntegrationEventsStore,
-    IntegrationEventWriteModel,
-} from './integration-events-store';
+import type { IntegrationEventsService } from './integration-events-service';
```
sighphyre, line 8: So this gets swapped with the service. Kinda suggests that the store is doing all the work here, is there a plan to have the service own a responsibility in a future PR?

src/lib/features/integration-events/integration-events-service.ts
```diff
@@ -0,0 +1,57 @@
...
+        offset: number,
+    ): Promise<IntegrationEventSchema[]> {
+        return this.integrationEventsStore.getPaginatedEvents(
+            id,
+            limit,
+            offset,
+        );
+    }
+
+    async registerEvent(
```
sighphyre, line 44: I think this function feels weird to me for a few reasons:

1) Because it needs to respect the feature flag, its return type can be undefined (glanced at the store, looks like the store returns an always defined type?). If the flag is the only reason we can return undefined I think that's incorrect
2) It's unclear at a glance why this needs to return the inserted object at all. Is that purely for tests?
3) Handling the flag at all here feels off to me, having functions that quietly do nothing if a flag is off feels incorrect to me unless they're at the outermost layers of the system (and even then I think I'd want logged warning or an error to be raised). This could make sense if this is the common layer where all integrations will hit and there's no other abstraction to handle this

src/lib/features/integration-events/integration-events-service.ts
```diff
@@ -0,0 +1,57 @@
...
+            getLogger,
+            flagResolver,
+        }: Pick<IUnleashConfig, 'getLogger' | 'flagResolver'>,
+    ) {
+        this.integrationEventsStore = integrationEventsStore;
+        this.flagResolver = flagResolver;
+        this.logger = getLogger('integration-events-service');
+    }
+
+    isEnabled(): boolean {
```
sighphyre, line 28: Entirely possible this is purely personal taste but I would leave abstractions off flag resolutions. Abstractions are most useful when they're hooks for future change but flags should only change to go away in the very near future. Won't complain if you want to keep it though

sighphyre (approved):
I think the code is nice and clean but without insight into an overarching design and details on the future plans for handling things like Jira and Terraform it's really difficult for me to know if this is going in a good direction and how much change the abstractions here are going to need to deal with in future. I think the really difficult questions on how weirdos like Terraform and Jira will be handled means we're kinda shooting in the dark on system design here

Going to slap an approve on it because it's entirely possible you have this all worked out in your mind and I see nothing scary here

chriswk:
> I think the code is nice and clean but without insight into an overarching design and details on the future plans for handling things like Jira and Terraform it's really difficult for me to know if this is going in a good direction and how much change the abstractions here are going to need to deal with in future. I think the really difficult questions on how weirdos like Terraform and Jira will be handled means we're kinda shooting in the dark on system design here
> 
> Going to slap an approve on it because it's entirely possible you have this all worked out in your mind and I see nothing scary here

Hmmm. I think Jira and Terraform are the reverse way though, and so far this is what was previously called addons; which is why it doesn't worry about weirdos like Jira and Terraform ;)

chriswk (approved):
I'm giving a thumbs up as well. Simon brings up some fair points about early abstraction, but not a deal breaker.

src/lib/features/integration-events/integration-events.e2e.test.ts
nunogois: Yeah, the idea is that this e2e test grows along with the development of this feature. We only had the store before, so we used that. We now have the service, which should be responsible for the business logic of the feature, so I'm using the service instead. 

The service isn't really doing much. That may change in the future or it may not, but if it does we now have a place for it 🤷 . Most of our services are just calling store methods so this is at least consistent in that sense.

I guess one thing that it is doing for now is controlling the clean up behavior through the feature flag. By checking it in the `cleanUpEvents` method, we can always schedule the method but dynamically control whether we're actually cleaning up events (see: https://github.com/Unleash/unleash/pull/5363). I don't think there's an outer layer that is more suitable for this.

src/lib/features/integration-events/integration-events-service.ts
nunogois: 1. Yep, fixed in https://github.com/Unleash/unleash/pull/7614/commits/278f7c7fe2091607f130cc95386a5f167166ee69 - We'll handle this in `src/lib/addons/addon.ts` at a later stage.
2. Tests and because I think it's generally polite to return to newly created object, like we do for POST requests. It's especially nice when we have follow up actions, like doing something with its id. Might just be personal preference though.
3. Given that this will not be available through the controller layer, this would likely be the common outermost layer of this specific logic, however like we discussed and I mentioned above, we can pass down the `flagResolver` and handle this in the `Addon` base class.

> Handling the flag at all here feels off to me, having functions that quietly do nothing if a flag is off feels incorrect to me unless they're at the outermost layers of the system (and even then I think I'd want logged warning or an error to be raised)

Generally speaking, I'm not so sure I agree and I would like you to fight me here because I want to better understand your arguments. 

I prefer being able to call these new methods where they make sense and not having to worry about it. As the layer usually responsible for the core domain logic of a feature, the service should be smart about it and only actually do the thing if it's enabled. Otherwise, it is the caller's responsibility to ensure that this feature is always behind a feature flag. It's also more work to clean up the flag afterwards.

I also don't like the idea of logging warnings or raising errors because to me code that is flagged off should act the exact same (or as close as possible) as not existing at all: no-op.

However I 100% agree that it's icky to change method's signatures because of flags, which is why I like our new approach 🙂

src/lib/features/integration-events/integration-events-service.ts
nunogois: Agreed, this is unnecessary and potentially even dangerous. Addressed in https://github.com/Unleash/unleash/commit/278f7c7fe2091607f130cc95386a5f167166ee69

PR #7616, opened 2024-07-18
fix: validate patched data with schema

nunogois:
https://linear.app/unleash/issue/2-2453/validate-patched-data-against-schema

This adds schema validation to patched data, fixing potential issues of patching data to an invalid state.

This can be easily reproduced by patching a strategy constraints to be an object (invalid), instead of an array (valid):

```sh
curl -X 'PATCH' \
  'http://localhost:4242/api/admin/projects/default/features/test/environments/development/strategies/8cb3fec6-c40a-45f7-8be0-138c5aaa5263' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '[
  {
    "path": "/constraints",
    "op": "replace",
    "from": "/constraints",
    "value": {}
  }
]'
```

Unleash will accept this because there's no validation that the patched data actually looks like a proper strategy, and we'll start seeing Unleash errors due to the invalid state.

This PR adapts some of our existing logic in the way we handle validation errors to support any dynamic object. This way we can perform schema validation with any object and still get the benefits of our existing validation error handling.

This PR also takes the liberty to expose the full instancePath as propertyName, instead of only the path's last section. We believe this has more upsides than downsides, especially now that we support the validation of any type of object.

image

### DIFF
--- a/src/lib/error/bad-data-error.ts
+++ b/src/lib/error/bad-data-error.ts
@@ -68,13 +68,11 @@ const additionalPropertiesMessage = (
 };
 
 const genericErrorMessage = (
-    requestBody: object,
     propertyName: string,
+    propertyValue: object,
     errorMessage: string = 'is invalid',
 ) => {
-    const input = getProp(requestBody, propertyName.split('/'));
-
-    const youSent = JSON.stringify(input);
+    const youSent = JSON.stringify(propertyValue);
     const message = `The \`${propertyName}\` property ${errorMessage}. You sent ${youSent}.`;
     return {
         message,
@@ -121,24 +119,21 @@ export const fromOpenApiValidationError =
     (data: object) =>
     (validationError: ErrorObject): ValidationErrorDescription => {
         const { instancePath, params, message } = validationError;
-        const [errorSource, substringOffset] = instancePath.startsWith('/body')
-            ? [(data as { body: object }).body, '/body/'.length]
-            : instancePath.startsWith('/query')
-              ? [(data as { query: object }).query, '/query/'.length]
-              : // This is not validating body or query, so we can assume this is manually validating plain data
-                [data, 1];
 
-        const propertyName = instancePath.substring(substringOffset);
+        const propertyValue = getProp(
+            data,
+            instancePath.split('/').filter(Boolean),
+        );
 
         switch (validationError.keyword) {
             case 'required':
                 return missingRequiredPropertyMessage(
-                    propertyName,
+                    instancePath,
                     params.missingProperty,
                 );
             case 'additionalProperties':
                 return additionalPropertiesMessage(
-                    propertyName,
+                    instancePath,
                     params.additionalProperty,
                 );
             case 'enum':
@@ -146,16 +141,17 @@ export const fromOpenApiValidationError =
                     instancePath.substring(instancePath.lastIndexOf('/') + 1),
                     message,
                     params.allowedValues,
-                    getProp(
-                        errorSource,
-                        instancePath.substring(substringOffset).split('/'),
-                    ),
+                    propertyValue,
                 );
 
             case 'oneOf':
-                return oneOfMessage(propertyName, validationError.message);
+                return oneOfMessage(instancePath, validationError.message);
             default:
-                return genericErrorMessage(errorSource, propertyName, message);
+                return genericErrorMessage(
+                    instancePath,
+                    propertyValue,
+                    message,
+                );
         }
     };
 
--- a/src/lib/error/bad-data-error.ts
+++ b/src/lib/error/bad-data-error.ts
@@ -7,6 +7,7 @@ type ValidationErrorDescription = {
     message: string;
     path?: string;
 };
+
 class BadDataError extends UnleashError {
     statusCode = 400;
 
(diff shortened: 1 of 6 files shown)


src/lib/openapi/validate.ts
```diff
@@ -38,3 +39,16 @@ export const validateSchema = <S = SchemaId>(
...
+
+export const throwOnInvalidSchema = <S = SchemaId>(
+    data: object,
+    schema: S,
+): void => {
+    const validationErrors = validateSchema(schema, data);
+    if (validationErrors) {
+        throw fromOpenApiValidationErrors(
+            data,
+            validationErrors.errors as [ErrorObject, ...ErrorObject[]],
```
kwasniew, line 51: More idiomatic pattern is
```
const [firstError, ...otherErrors] = validationErrors.
 throw fromOpenApiValidationErrors(
            data,
            [firstError, ...otherErrors]);
```

src/lib/features/feature-toggle/feature-toggle-controller.ts
```diff
@@ -1060,6 +1062,9 @@ export default class ProjectFeaturesController extends Controller {
         const strategy = await this.featureService.getStrategy(strategyId);
 
         const { newDocument } = applyPatch(strategy, patch);
+
+        throwOnInvalidSchema(newDocument, featureStrategySchema.$id);
```
kwasniew, line 1066: we can call it validateSchema since it's a pattern we follow in other places. Internally those functions return void and throw and error but we describe business purpose (validate) not implementation details (throw) in the method signature

src/lib/error/bad-data-error.ts
```diff
@@ -117,12 +118,15 @@ const enumMessage = (
...
     (validationError: ErrorObject): ValidationErrorDescription => {
         const { instancePath, params, message } = validationError;
         const [errorSource, substringOffset] = instancePath.startsWith('/body')
-            ? [request.body, '/body/'.length]
-            : [request.query, '/query/'.length];
+            ? [(data as { body: object }).body, '/body/'.length]
+            : instancePath.startsWith('/query')
+              ? [(data as { query: object }).query, '/query/'.length]
+              : // This is not validating body or query, so we can assume this is manually validating plain data
+                [data, 1];
```
thomasheartman, line 129: I'm not suuuuper hot on this implementation, to be frank. One thing is that you'll lose autocompletion, but it feels a little loose when you can pass any object. You could send an empty object, but put `/body` in the instance path and it'd blow up. But then again, this is typescript. You can lie to it as much as you want 🤷🏼 

But I'm not gonna die on that hill, if the rest of you think this is a better implementation.

thomasheartman (approved):

src/lib/openapi/validate.ts
nunogois: Thanks, addressed in https://github.com/Unleash/unleash/pull/7616/commits/a056b0a9485b28656731595eaf23e0663c5ef43b

src/lib/features/feature-toggle/feature-toggle-controller.ts
nunogois: We already have a `validateSchema` in this file, but I'm open to suggestions!

src/lib/error/bad-data-error.ts
```diff
@@ -117,12 +118,15 @@ const enumMessage = (
 };
 
 export const fromOpenApiValidationError =
-    (request: { body: object; query: object }) =>
+    (data: object) =>
```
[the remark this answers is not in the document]
nunogois: That could also work, if you think it has benefits. Do you want to explore that idea and we can compare them?

src/lib/error/bad-data-error.ts
nunogois: Hey @thomasheartman, I did a few more changes to make this even more dynamic. Let me know what you think.

>One thing is that you'll lose autocompletion, but it feels a little loose when you can pass any object.

Looking at our code I don't think we're taking advantage of the current type anyways. I agree it feels loose, but I also think it makes sense, if we're allowing you to validate anything against a schema. Why should we be limited to objects with `body` and `query` properties?

>You could send an empty object, but put /body in the instance path and it'd blow up.

I don't think it will (with the recent changes), and I added a test that validates that.

>But I'm not gonna die on that hill, if the rest of you think this is a better implementation.

I'm also open to alternatives, if you think we can do this in a better way!

thomasheartman (approved):
> This PR also takes the liberty to expose the full instancePath as propertyName, instead of only the path's last section. We believe this has more upsides than downsides, especially now that we support the validation of any type of object.

I like this! That's referring to how the `path` property in the output now starts with `/body`, right? Just for some background info, the reason we trimmed it before is that the paths didn't use to start that way, and when it was brought in, I didn't think to add any handling for query params. So I think this is a good way to go. It could've also been nice to add an `in` property to the object, which is either `body` or `query`, but I don't think that's necessary.

I think the `data: object` parameter change makes a lot more sense with the changes the way they are now. I'm still a little hesitant, but I think it's just me desperately clinging on to my prior position 😄 Rationally I think it's a nice improvement now.

kwasniew (approved):

PR #7617, opened 2024-07-18
chore: send prometheus metrics when someone tries to exceed resource limits

thomasheartman:
This PR adds prometheus metrics for when users attempt to exceed the limits for a given resource. 

The implementation sets up a second function exported from the ExceedsLimitError file that records metrics and then throws the error. This could also be a static method on the class, but I'm not sure that'd be better.

### DIFF
--- a/src/lib/error/exceeds-limit-error.ts
+++ b/src/lib/error/exceeds-limit-error.ts
@@ -3,13 +3,13 @@ import { GenericUnleashError } from './unleash-error';
 import { EXCEEDS_LIMIT } from '../metric-events';
 
 export class ExceedsLimitError extends GenericUnleashError {
-    constructor(resource: string, limit: number, eventBus: EventEmitter) {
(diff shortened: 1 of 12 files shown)


src/lib/error/exceeds-limit-error.ts
```diff
@@ -9,3 +11,12 @@ export class ExceedsLimitError extends GenericUnleashError {
         });
     }
 }
+
+export const throwExceedsLimitError = (
```
kwasniew, line 15: I'd put eventBus as the first param since it's a complex dependency and not data. As a rule of thumb I prefer complex dependencies first and data second.
```
class Class {
   constructure(dep) {}
  method(data) {}
}```
becomes
```fn(dep, data) {}```
If you curry this function you can bake deps in and you're left with the data.

There's a great talk on this topic: https://www.youtube.com/watch?v=m3svKOdZijA

kwasniew (approved):
Minor nitpick, otherwise solid work

PR #7622, opened 2024-07-19
feat: allow you to gradually scale back constraint usage

thomasheartman:
This PR updates the limit validation for constraint numbers on a single strategy. In cases where you're already above the limit, it allows you to still update the strategy as long as you don't add any **new** constraints (that is: the number of constraints doesn't increase).

A discussion point: I've only tested this with unit tests of the method directly. I haven't tested that the right parameters are passed in from calling functions. The main reason being that that would involve updating the fake strategy and feature stores to sync their flag lists (or just checking that the thrown error isn't a limit exceeded error), because right now the fake strategy store throws an error when it doesn't find the flag I want to update.

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-service.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-service.ts
@@ -391,16 +391,15 @@ class FeatureToggleService {
     }
 
     validateConstraintsLimit(constraints: {
-        updatedConstraints: IConstraint[];
-        existingConstraints: IConstraint[];
+        updated: IConstraint[];
+        existing: IConstraint[];
     }) {
         if (!this.flagResolver.isEnabled('resourceLimits')) return;
 
         const constraintsLimit = this.resourceLimits.constraints;
         if (
-            constraints.updatedConstraints.length > constraintsLimit &&
-            constraints.updatedConstraints.length >
-                constraints.existingConstraints.length
+            constraints.updated.length > constraintsLimit &&
+            constraints.updated.length > constraints.existing.length
         ) {
             throwExceedsLimitError(this.eventBus, {
                 resource: `constraints`,
@@ -409,7 +408,7 @@ class FeatureToggleService {
         }
 
         const constraintValuesLimit = this.resourceLimits.constraintValues;
-        const constraintOverLimit = constraints.updatedConstraints.find(
+        const constraintOverLimit = constraints.updated.find(
             (constraint) =>
                 Array.isArray(constraint.values) &&
                 constraint.values?.length > constraintValuesLimit,
@@ -669,8 +668,8 @@ class FeatureToggleService {
             strategyConfig.constraints.length > 0
         ) {
             this.validateConstraintsLimit({
-                updatedConstraints: strategyConfig.constraints,
-                existingConstraints: [],
+                updated: strategyConfig.constraints,
+                existing: [],
             });
             strategyConfig.constraints = await this.validateConstraints(
                 strategyConfig.constraints,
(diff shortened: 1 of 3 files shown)


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -390,19 +390,26 @@ class FeatureToggleService {
         }
     }
 
-    validateConstraintsLimit(updatedConstrains: IConstraint[]) {
+    validateConstraintsLimit(constraints: {
+        updatedConstraints: IConstraint[];
+        existingConstraints: IConstraint[];
```
thomasheartman, line 395: Because these parameters have the same type, I want to force people to name them to avoid confusion. That's also the reason why I've marked both as required.

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -390,19 +390,25 @@ class FeatureToggleService {
         }
     }
 
-    validateConstraintsLimit(updatedConstrains: IConstraint[]) {
+    validateConstraintsLimit(constraints: {
+        updated: IConstraint[];
+        existing: IConstraint[];
```
thomasheartman, line 395: Because these parameters have the same type, I want to force people to name them to avoid confusion. That's also the reason why I've marked both as required.

src/lib/features/feature-toggle/tests/feature-toggle-service.limit.test.ts
```diff
@@ -91,6 +91,44 @@ describe('Strategy limits', () => {
         );
     });
 
+    test('Should not throw limit exceeded errors if the new number of constraints is less than or equal to the previous number', () => {
```
kwasniew, line 94: do we also have a test for should throw?

thomasheartman: We already have a number of tests in place that checks that it throws when you reach the limit, right? But I can add another one in if you think that's useful?

thomasheartman: Like I can add a little:

```ts
        expect(() => {
            featureToggleService.validateConstraintsLimit({
                updated: [...constraints, ...constraints],
                existing: constraints,
            });
        }).toThrow(ExceedsLimitError);
```

To the end, for instance?

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -390,19 +390,25 @@ class FeatureToggleService {
         }
     }
 
-    validateConstraintsLimit(updatedConstrains: IConstraint[]) {
+    validateConstraintsLimit(constraints: {
```
kwasniew, line 393: tbh I don't like that our validate functions are not private since it's tempting to call them from outside the service and I'd like to avoid that. On the other hand I understand the testability concern with fake stores setup.

kwasniew (approved):

src/lib/features/feature-toggle/feature-toggle-service.ts
thomasheartman: Addressed in 17ee7675bfa50707bbb8b14f363f77ba00d8cb91

src/lib/features/feature-toggle/tests/feature-toggle-service.limit.test.ts
thomasheartman: Also added in 17ee7675bfa50707bbb8b14f363f77ba00d8cb91

PR #7625, opened 2024-07-19
feat: feature collaborators read model

kwasniew:
## About the changes

Added feature collaborators read model.

Details:
* read model is not connected to feature service yet
* there's a dedicated integration test verifying the query
* events table has user id which is used to find full user info (name/username + email) in the users table
* we ignore created_by field in the events since it's a mix of names and emails. id of the user seems to be more reliable
* the gist of event events query is to select distinct user ids with the users who made most recent changes/events listed first
* currently we're not filtering out any events for the collaborator selection (we may find in future that some events are irrelevant)


thomasheartman (approved):
Nice one! I agree with your assessment outlined in the details, especially with regard to filtering etc.

PR #7621, opened 2024-07-18
chore: register integration events in webhooks

nunogois:
https://linear.app/unleash/issue/2-2450/register-integration-events-webhook

Registers integration events in the **Webhook** integration.

Even though this touches a lot of files, most of it is preparation for the next steps. The only actual implementation of registering integration events is in the **Webhook** integration. The rest will follow on separate PRs.

Here's an example of how this looks like in the database table:

```json
{
  "id": 7,
  "integration_id": 2,
  "created_at": "2024-07-18T18:11:11.376348+01:00",
  "state": "failed",
  "state_details": "Webhook request failed with status code: ECONNREFUSED",
  "event": {
    "id": 130,
    "data": null,
    "tags": [],
    "type": "feature-environment-enabled",
    "preData": null,
    "project": "default",
    "createdAt": "2024-07-18T17:11:10.821Z",
    "createdBy": "admin",
    "environment": "development",
    "featureName": "test",
    "createdByUserId": 1
  },
  "details": {
    "url": "http://localhost:1337",
    "body": "{ \"id\": 130, \"type\": \"feature-environment-enabled\", \"createdBy\": \"admin\", \"createdAt\": \"2024-07-18T17: 11: 10.821Z\", \"createdByUserId\": 1, \"data\": null, \"preData\": null, \"tags\": [], \"featureName\": \"test\", \"project\": \"default\", \"environment\": \"development\" }"
  }
}
```

PR #7626, opened 2024-07-19
chore: register integration events in Slack integration

nunogois:
https://linear.app/unleash/issue/2-2458/register-integration-events-slack

Registers integration events in the **Slack** integration.

Similar to: https://github.com/Unleash/unleash/pull/7621

Also slightly improves the previous work on webhooks.


src/lib/addons/slack.test.ts
```diff
@@ -44,252 +46,298 @@ jest.mock(
...
 
-            async registerEvent(_) {
-                return Promise.resolve();
+            async registerEvent(event) {
+                return registerEventMock(event);
             }
         },
 );
 
-test('Should call slack webhook', async () => {
```
nunogois, line 56: Adding the `describe` messed up the diff, but the only real change is the new `Should call registerEvent` test.

chriswk (approved):
LGTM

PR #7627, opened 2024-07-19
feat: feature collaborators added to API behind a flag

kwasniew:
## About the changes

Exposing feature collaborators in the GET feature API.

Details:
* this is behind a flag and by default people will not see the new collaborators field
* the new collaborators field is deliberately omitted in the openapi spec before UI tells us what is the right structure
* currently collaborators have nested users and a potential total count field can be added later. If we find the total is not needed we can flatten the structure
* we also support anonimization of collaborators in a similar way we anonimize the created by field

### DIFF
--- a/src/lib/features/feature-toggle/feature-collaborators-read-model.ts
+++ b/src/lib/features/feature-toggle/feature-collaborators-read-model.ts
@@ -31,6 +31,8 @@ export class FeatureCollaboratorsReadModel
             .join('users', 'recent_events.created_by_user_id', 'users.id')
             .orderBy('recent_events.max_created_at', 'desc');
 
+        console.log(query.toQuery());
+
         const rows = await query;
 
         return rows.map((row) => {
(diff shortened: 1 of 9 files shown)


src/lib/features/feature-toggle/tests/feature-toggles.auth.e2e.test.ts
```diff
@@ -153,10 +158,14 @@ test('Should read flag creator', async () => {
         TEST_AUDIT_USER,
     );
 
-    await db.stores.featureToggleStore.create('default', {
-        name,
-        createdByUserId: user.id,
-    });
+    await app.services.featureToggleService.createFeatureToggle(
```
kwasniew, line 161: had to switch to a service call so that an event is generated for the collaborators calculation

src/lib/features/feature-toggle/feature-collaborators-read-model.ts
```diff
@@ -31,6 +31,8 @@ export class FeatureCollaboratorsReadModel
             .join('users', 'recent_events.created_by_user_id', 'users.id')
             .orderBy('recent_events.max_created_at', 'desc');
 
+        console.log(query.toQuery());
+
```
thomasheartman: You probably wanna get rid of this.

```suggestion
```

thomasheartman (approved):

src/lib/features/feature-toggle/feature-collaborators-read-model.ts
thomasheartman: Ah, you got there before me 😄

kwasniew: already did as you were writing the comment :D

PR #7641, opened 2024-07-22
chore: integration events hook

nunogois:
https://linear.app/unleash/issue/2-2440/create-new-integration-event-hooks

Adds a frontend hook for integration events, allowing us to easily fetch paginated integration events for a specific integration configuration id.


chriswk (approved):
Nice

PR #7644, opened 2024-07-23
fix: make collaborator component smaller.

thomasheartman:
This PR makes the collaborator component smaller, bringing it more into line with what the sketches show.

To achieve this, we need to modify the group card avatars component a bit to allow you to somehow decide the size of the avatars. I opted to do this by allowing you to override the avatar component that gets passed in. We could also allow you to pass in `avatarDiameter` or something of the sort, but that felt clunkier and like more code.

It's worth noting, though, that you can't affect the avatar overlap from outside at the moment. This works fine when the avatars are the current size, but it might be trouble if you make them tiny. However, there's no reason why we would do that, so I think it's fine for now. We could do something like use CSS custom properties that you have to set, but that feels like it'd break easily, or we could pass in extra react props, but nah.

Updated component in context:
image

Previous iteration:
image

Sketches:
image

### DIFF
--- a/frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx
+++ b/frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx
@@ -55,12 +55,19 @@ export const GroupCardAvatars = ({
     return <GroupCardAvatarsInner AvatarComponent={Avatar} {...props} />;
 };
 
+type GroupCardAvatarsInnerProps = Omit<
+    IGroupCardAvatarsProps,
+    'AvatarComponent'
+> & {
+    AvatarComponent: typeof UserAvatar;
+};
+
 const GroupCardAvatarsInner = ({
     users = [],
     header = null,
     avatarLimit = 9,
     AvatarComponent,
-}: IGroupCardAvatarsProps) => {
+}: GroupCardAvatarsInnerProps) => {
     const shownUsers = useMemo(
         () =>
             users
--- a/frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx
+++ b/frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx
@@ -1,4 +1,4 @@
-import { type Theme, styled } from '@mui/material';
+import { styled } from '@mui/material';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/UserAvatar/UserAvatar.tsx
thomasheartman, line 1: The deleted code here was my previous attempt at styling avatars. Turns out you can just use `styled`, so we don't need all this extra weird code here. That's also the update in `AvatarCell`

thomasheartman:
Huh, this breaks the tooltip for the group cards. Whoops.
image

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx
```diff
@@ -42,12 +43,23 @@ interface IGroupCardAvatarsProps {
...
 export const GroupCardAvatars = ({
+    AvatarComponent,
+    ...props
+}: IGroupCardAvatarsProps) => {
+    const Avatar = StyledAvatar(AvatarComponent ?? UserAvatar);
+
+    return <GroupCardAvatarsInner AvatarComponent={Avatar} {...props} />;
+};
+
+const GroupCardAvatarsInner = ({
```
thomasheartman, line 58: This is an interesting use case that I want to explain.

By allowing you to pass a custom avatar prop, you can add your own custom styling to the avatars used by the group avatar component. This is useful in this instance because we want the avatars to be smaller than the group card avatar size. 

By currying the `StyledAvatar` function and allowing you to partially apply it by passing in the component, we allow you to pass a component in that we then add more styles to. This feels pretty cool to me.

However, I had to extract it into its own component because when you moused over an avatar, that'd cause the component to do some recalculations, leading to a rerendering of the avatar. This had two side effects when trying to open the popover:
1. It would constantly retrigger the onMouseOver function.
2. Because it didn't have a target element anymore, the popover would be rendered in the top left of the screen as shown in this image: 
image

By extracting the Avatar styling to an outside function, we can get around that completely.

frontend/src/component/feature/FeatureView/Collaborators.tsx
```diff
@@ -73,6 +85,7 @@ const Container = styled('article')(({ theme }) => ({
     gap: theme.spacing(10),
     alignItems: 'center',
     justifyContent: 'space-between',
+    fontSize: `calc(0.9 * ${theme.typography.body2.fontSize})`,
```
Tymek:
```suggestion
```

We don't have smaller font size in theme

Tymek (approved):

thomasheartman:
Merged as part of https://github.com/Unleash/unleash/pull/7649

PR #7649, opened 2024-07-24
fix: add workaround for tooltip

thomasheartman:
This PR adds a workaround to the non-working tooltip for the "last modified by" avatar to make it work. I don't know why it's necessary, only that it doesn't work without either an intermediary element or without an extra added space (refer to the next section).

image

## Problems with the regular tooltip

Aside from adding extra components, I could only get the tooltip to work if I added an empty string before it, i.e.

``` jsx
<Tooltip>
    {''}
    <Avatar/>
</Tooltip>
```

But this is apparently invalid, because the tooltip only accepts a single child. I tried to compare it with the tooltip we have in `frontend/src/component/project/Project/PaginatedProjectFeatureToggles/AvatarCell.tsx`, but it didn't really help. Until I realized that that avatar is not the direct descendant of the tooltip.

I'm genuinely quite confused as to why just nesting directly doesn't work, but it is what it is.

## Options

The other option to this workaround, is to just use the regular image title instead. That would allow us to remove the nesting, but it doesn't look as slick, I think.

### DIFF
--- a/frontend/src/component/feature/FeatureView/Collaborators.tsx
+++ b/frontend/src/component/feature/FeatureView/Collaborators.tsx
@@ -1,5 +1,6 @@
 import { styled } from '@mui/material';
-import { GroupCardAvatars } from 'component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars';
+import { GCA } from 'component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars';
+import { HtmlTooltip } from 'component/common/HtmlTooltip/HtmlTooltip';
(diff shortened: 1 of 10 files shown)


frontend/src/component/feature/FeatureView/Collaborators.tsx
```diff
@@ -0,0 +1,96 @@
...
+    );
+};
+
+const Container = styled('article')(({ theme }) => ({
+    display: 'flex',
+    flexDirection: 'row',
+    gap: theme.spacing(10),
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    fontSize: `calc(0.9 * ${theme.typography.body2.fontSize})`,
```
thomasheartman, line 73:
```suggestion
```

PR #7652, opened 2024-07-24
chore: keep latest integration events for each integration configuration

nunogois:
https://linear.app/unleash/issue/2-2469/keep-the-latest-event-for-each-integration-configuration

This makes it so we keep the latest event for each integration configuration, along with the previous logic of keeping the latest 100 events of the last 2 hours. 

This should be a cheap nice-to-have, since now we can always know what the latest integration event looked like. This will tie-in nicely with the next task of making the latest integration event state visible in the integration card.

Also improved the clarity of auto-deletion explanation in the modal.

### DIFF
--- a/src/lib/features/integration-events/integration-events-store.ts
+++ b/src/lib/features/integration-events/integration-events-store.ts
@@ -41,11 +41,21 @@ export class IntegrationEventsStore extends CRUDStore<
                 qb.select('id')
                     .from(this.tableName)
                     .whereRaw(`created_at >= now() - INTERVAL '2 hours'`)
-                    .orderBy('created_at', 'desc')
+                    .orderBy('id', 'desc')
                     .limit(100);
+            })
+            .with('latest_per_integration', (qb) => {
+                qb.select(this.db.raw('DISTINCT ON (integration_id) id'))
+                    .from(this.tableName)
+                    .orderBy('integration_id')
+                    .orderBy('id', 'desc');
             })
             .from(this.tableName)
             .whereNotIn('id', this.db.select('id').from('latest_events'))
+            .whereNotIn(
+                'id',
+                this.db.select('id').from('latest_per_integration'),
+            )
             .delete();
     }
 }
--- a/frontend/src/component/integrations/IntegrationEventsModal/IntegrationEventsModal.tsx
+++ b/frontend/src/component/integrations/IntegrationEventsModal/IntegrationEventsModal.tsx
@@ -105,8 +105,10 @@ export const IntegrationEventsModal = ({
                     </StyledHeaderRow>
                     <StyledHeaderRow>
                         <StyledSubtitle>
-                            All events older than the last 100 or older than the
-                            past 2 hours will be automatically deleted.
+                            Except for the most recent event for each
+                            integration, all events older than the last 100 or
+                            older than the past 2 hours will be automatically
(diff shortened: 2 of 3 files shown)


src/lib/features/integration-events/integration-events-store.ts
```diff
@@ -41,11 +41,21 @@ export class IntegrationEventsStore extends CRUDStore<
                 qb.select('id')
                     .from(this.tableName)
                     .whereRaw(`created_at >= now() - INTERVAL '2 hours'`)
-                    .orderBy('created_at', 'desc')
+                    .orderBy('id', 'desc')
                     .limit(100);
             })
+            .with('latest_per_integration', (qb) => {
+                qb.select(this.db.raw('DISTINCT ON (integration_id) id'))
```
gastonfournier, line 48: Should we add something like max per id or something? This query might be expensive. I'm not sure but maybe  you already considered this

src/lib/features/integration-events/integration-events-store.ts
```diff
@@ -41,11 +41,21 @@ export class IntegrationEventsStore extends CRUDStore<
...
+                    .from(this.tableName)
+                    .orderBy('integration_id')
+                    .orderBy('id', 'desc');
+            })
             .from(this.tableName)
             .whereNotIn('id', this.db.select('id').from('latest_events'))
+            .whereNotIn(
+                'id',
+                this.db.select('id').from('latest_per_integration'),
```
gastonfournier, line 57: How big this can be? I assume this depends on the response of the comment above

gastonfournier (commented):
In general looks good but I have a few doubts about the queries

src/lib/features/integration-events/integration-events-store.ts
nunogois: I don't expect this to be expensive but you may be thinking of something that I missed. 
Let me know if this is better: https://github.com/Unleash/unleash/pull/7652/commits/272cd0fb4094cc29f3837e9527adca4b1c37f86d

src/lib/features/integration-events/integration-events-store.ts
nunogois: This will be as big as the total number of integration configurations in the instance, which I don't expect to be a high number.
Again, maybe this will help: https://github.com/Unleash/unleash/pull/7652/commits/272cd0fb4094cc29f3837e9527adca4b1c37f86d

gastonfournier (approved):
MAX is probably better than distinct as I assume we have an index there. 

Maybe this is a good opportunity to add a timer to these queries, something like: https://github.com/Unleash/unleash/blob/339e92f4eb4e393bb842ea64bdf84a7b33992717/src/lib/db/access-store.ts#L90 but that's unrelated and can be considered just a nice to have

nunogois:
> MAX is probably better than distinct as I assume we have an index there.
> 
> Maybe this is a good opportunity to add a timer to these queries, something like:
> 
> https://github.com/Unleash/unleash/blob/339e92f4eb4e393bb842ea64bdf84a7b33992717/src/lib/db/access-store.ts#L90
> 
> but that's unrelated and can be considered just a nice to have

Good idea! Addressed in https://github.com/Unleash/unleash/pull/7652/commits/75b4b5494a6bba71994914f840a1edcec0848d6a

PR #7656, opened 2024-07-24
chore: show latest integration event on card

nunogois:
https://linear.app/unleash/issue/2-2443/show-the-latest-integration-event-result-on-that-integrations-card

Shows the latest event on the integration card.

Also renames one of the folders to make its contents clearer.

image


gastonfournier (approved):
LG!

PR #7663, opened 2024-07-25
chore: create shared dialog form template

thomasheartman:
This PR extracts the dialog form that we created for the new project form into a shared component in the `common` folder. 

Most of the code has been lifted and shifted, but there's been some minor adjustments along the way. The main file is `frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx`. Everything else is just cleanup.


frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx
```diff
@@ -0,0 +1,112 @@
...
+                            style: { fontSize: theme.typography.h2.fontSize },
+                        }}
+                        InputLabelProps={{
+                            style: { fontSize: theme.typography.h2.fontSize },
+                        }}
+                    />
+                </ProjectDescriptionContainer>
+            </TopGrid>
+
+            <ConfigButtons>{configButtons}</ConfigButtons>
```
thomasheartman, line 103: Discussion point: should we instead use children for the config buttons or is an explicit prop better?

frontend/src/component/project/Project/CreateProject/NewCreateProjectForm/CreateProjectDialog.tsx
thomasheartman, line 1: This file has grown because it has subsumed the parts of `NewProjectForm` that are no longer necessary. It makes this one longer, but it saves us ~60 lines of code because we avoid some weird duplication and unnecessary abstractions.

frontend/src/component/project/Project/CreateProject/NewCreateProjectForm/CreateProjectDialog.test.tsx
thomasheartman, line 1: These updates are necessary because with the new template, the header has the same title as the button. Besides, looking it up by role is what the testing library recommends anyway.

frontend/src/component/project/Project/CreateProject/NewCreateProjectForm/CreateProjectDialog.tsx
thomasheartman, line 1: This file has grown because it has subsumed the parts of NewProjectForm that are no longer necessary. It makes this one longer, but it saves us ~60 lines of code because we avoid some weird duplication and unnecessary abstractions.

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
```diff
@@ -16,16 +15,19 @@ export const StyledFormSection = styled('div')(({ theme }) => ({
 
 export const TopGrid = styled(StyledFormSection)(({ theme }) => ({
     display: 'grid',
-    gridTemplateAreas:
-        '"icon header" "icon project-name" "icon project-description"',
+    gridTemplateAreas: `
+        "icon header"
+        ".    project-name"
```
sjaanus, line 20: This looks weird. Is it out formatting?

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
```diff
@@ -16,16 +15,19 @@ export const StyledFormSection = styled('div')(({ theme }) => ({
...
+        ".    project-description"`,
     gridTemplateColumns: 'auto 1fr',
     gap: theme.spacing(4),
 }));
 
-export const StyledIcon = styled(ProjectIcon)(({ theme }) => ({
-    fill: theme.palette.primary.main,
-    stroke: theme.palette.primary.main,
-}));
+export const styleIcon = (Icon: React.ComponentType) =>
```
sjaanus, line 26: Why such lower case naming and code style change?

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
thomasheartman: Oh, yeah, I formatted it like this to make it visually match what the layout is. We can put it on a single line if you don't like it, but I think this is a useful way to show it.

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx
sjaanus: I think using prop is fine for now, because it is not clear what the children must be. Are they input fields or buttons or something completely different.

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
thomasheartman: Not sure what the code style change you're referring to is, but I lowercased it because it's a function now, and not a component directly. Happy to uppercase it if you think that's better, though 🤷🏼

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx
```diff
@@ -0,0 +1,112 @@
...
+    validateName?: () => void;
+    testIds?: Partial<{
+        nameInput: string;
+        descriptionInput: string;
+        submitButton: string;
+    }>;
+};
+
+export const DialogFormTemplate: React.FC<FormProps> = ({
+    Limit,
```
sjaanus, line 41: why caps

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
thomasheartman: Oh, and it's a function because we don't know what kind of icon to use here, so we need to create this dynamically. This allows us to do that formatting in the template component instead of on the outside.

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx
thomasheartman: To be consistent with other instances of this and because it's a full component on its own. Again, no strong opinions here, so I'll lowercase it if you want. Or name it `limitComponent` etc.

frontend/src/component/project/Project/CreateProject/NewCreateProjectForm/CreateProjectDialog.tsx
```diff
@@ -205,17 +273,159 @@ export const CreateProjectDialog = ({
...
-                </NewProjectForm>
+                    handleSubmit={handleSubmit}
+                    name={projectName}
+                    setName={setProjectName}
+                    description={projectDesc}
+                    setDescription={setProjectDesc}
+                    errors={errors}
+                    icon={StyledProjectIcon}
+                    onClose={onClose}
+                    testIds={{
```
sjaanus, line 284: Could the ids be same not configurable? I feel like when testing there is always just 1 formtemplate open, so no confusion.

thomasheartman: Yeah, maybe? They're used in cypress tests (`frontend/cypress/support/UI.ts`), but cypress should also only have a single form open at the same time. I'll have a look at it.

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
sjaanus: ah yes its a function now, I think it is fine then

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx
sjaanus: ok caps is fine if component

frontend/src/component/project/Project/CreateProject/NewCreateProjectForm/CreateProjectDialog.tsx
thomasheartman: Made some changes now. Let's see whether cypress likes it. Good shout

sjaanus (approved):

PR #7664, opened 2024-07-25
feat: tag feature on creation

sjaanus:
Now it is possible to tag feature on creation.

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-service.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-service.ts
@@ -1256,7 +1256,7 @@ class FeatureToggleService {
         await this.validateName(value.name);
         await this.validateFeatureFlagNameAgainstPattern(value.name, projectId);
 
-        const exists = await this.projectStore.hasProject(projectId);
+        const projectExists = await this.projectStore.hasProject(projectId);
 
         if (await this.projectStore.isFeatureLimitReached(projectId)) {
             throw new InvalidOperationError(
@@ -1266,7 +1266,7 @@ class FeatureToggleService {
 
         await this.validateFeatureFlagLimit();
 
-        if (exists) {
+        if (projectExists) {
             let featureData: FeatureToggleInsert;
             if (isValidated) {
                 featureData = { createdByUserId: auditUser.id, ...value };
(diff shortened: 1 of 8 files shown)


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1256,7 +1256,7 @@ class FeatureToggleService {
         await this.validateName(value.name);
         await this.validateFeatureFlagNameAgainstPattern(value.name, projectId);
 
-        const exists = await this.projectStore.hasProject(projectId);
+        const projectExists = await this.projectStore.hasProject(projectId);
```
sjaanus, line 1259: Make this more clear

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1301,6 +1301,16 @@ class FeatureToggleService {
                 );
             }
 
+            if (value.tags && value.tags.length > 0) {
```
sjaanus, line 1304: If exist in payload, insert tags.

src/lib/openapi/spec/create-feature-schema.ts
```diff
@@ -32,14 +33,17 @@ export const createFeatureSchema = {
         },
         tags: {
             type: 'array',
+            description: 'Tags to add to the feature.',
             items: {
-                type: 'string',
-                example: 'simple:test',
+                $ref: '#/components/schemas/tagSchema',
```
sjaanus, line 38: Fixing the schema to reuse existing schema and make it similar across unleash.

thomasheartman: Oh, wow, did we actually have the possibility to do this before?

thomasheartman (approved):

src/lib/openapi/spec/create-feature-schema.ts
thomasheartman: Wait, does this break the API somehow? As in, is this a breaking change?

sjaanus: Nope, I implemented it yesterday, but after adding backend code, it made sense to use tagSchema.

PR #7668, opened 2024-07-25
fix: validate project names on blur

thomasheartman:
This fix validates the project name when you blur the field in the new project form. The only instances where it'll be wrong is if you have just whitespace or an empty string, but you'll be notified immediately.

Also removes some unused variables and parameters that I found.


frontend/src/component/project/Project/CreateProject/NewCreateProjectForm/CreateProjectDialog.tsx
```diff
@@ -276,6 +275,7 @@ export const CreateProjectDialog = ({
                     errors={errors}
                     icon={StyledProjectIcon}
                     onClose={onClose}
+                    validateName={validateName}
```
thomasheartman, line 278: This was the actual fix. Everything else is just fluff.

sjaanus (approved):

PR #7669, opened 2024-07-25
fix: Hide project selection option in CreateFeatureDialog when OSS

thomasheartman:
This change wraps the project selection option in the
CreateFeatureDialog in a conditional that hides it when Unleash is
OSS.

OSS doesn't have access to the project creation API, so there's no
point in showing this.


sjaanus (approved):
LG, hopefully it works without the input present.

thomasheartman:
Valid concern! Everything appears to work as expected when testing it, at least. It takes the project from the URL anyway, so 💁🏼

PR #7670, opened 2024-07-25
docs: integration events

nunogois:
https://linear.app/unleash/issue/2-2444/write-documentation-on-integration-events

Adds documentation for integration events.

Should be pretty straightforward, so a section in the existing integrations reference doc should be good enough.

See it here: https://unleash-docs-git-docs-integration-events-unleash-team.vercel.app/reference/integrations#integration-events


chriswk (approved):
Looks really good

PR #7672, opened 2024-07-26
fix: project icon sizing and color

thomasheartman:
This PR fixes the project icon sizing and color in the create feature form. However, it includes a little more cleanup work.

1. I've switched out the previous project icon with a new one. The difference? The old one had lots of whitespace around it, which made sizing it awkward.
2. I've added a wrapping Icon component to common. The idea was taken from this thread on the MUI GitHub about how to make imported icons behave consistently with MUI icons.
3. Because I've switched the icon and added a new component, I've also gone and updated the other places we used the old icon, so that we're consistent about the usage.

In create dialog form:
image

Updated icon in:
Project card component
image

Project creation form
image

Sidebar and command bar (it's actually larger here, but maybe that's okay? Previous project was kinda small):
image

(Previous sidebar and command bar):
image

I'd be willing to leave the sidebar and command bar for now if we think it's better to leave them using the same size as previously, but we can talk about that. I think it's better in the sidebar; undecided about the command bar.

Bonus changes: I fixed some typos

### DIFF
--- a/frontend/src/assets/icons/projectIconSmall.svg
+++ b/frontend/src/assets/icons/projectIconSmall.svg
@@ -1,5 +1,5 @@
-<svg xmlns="http://www.w3.org/2000/svg" focusable='false' aria-hidden='true' width="37" height="35" viewBox="0 0 37 35" fill="none">
-    <rect x="8.9864" y="9.93245" width="3.78378" height="13.2432" rx="1.41892" stroke-width="0.945947" />
-    <rect x="16.5542" y="13.7161" width="3.78378" height="9.45946" rx="1.41892" stroke-width="0.945947" />
-    <rect x="24.1217" y="18.4456" width="3.78378" height="4.72973" rx="1.41892" stroke-width="0.945947" />
+<svg width="14" height="10" viewBox="0 0 14 10" fill="none" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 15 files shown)


sjaanus (approved):
If only place that is using the old project icon is command menu, then lets go with the new one, just for consistency and to get rid of old. 
If there are other places using the same way as command menu, lets keep it, because it looked better imo.

thomasheartman:
So right now (in this PR), no places use the old icon.

Previously, it was used by the project tiles, project creation form, sidebar, and command bar (all the screenies shown).

Everywhere except the sidebar and command bar, it was easy to make it look more or less exactly how it did before.

For the sidebar and command bar, it would take some extra work. I'd be willing to look at it, if you think that's worth spending time on, but I'm leaning towards leaving it as is myself 🤷🏼

sjaanus:
> So right now (in this PR), no places use the old icon.
> 
> Previously, it was used by the project tiles, project creation form, sidebar, and command bar (all the screenies shown).
> 
> Everywhere except the sidebar and command bar, it was easy to make it look more or less exactly how it did before.
> 
> For the sidebar and command bar, it would take some extra work. I'd be willing to look at it, if you think that's worth spending time on, but I'm leaning towards leaving it as is myself 🤷🏼

I think it is fine to leave as is

PR #7675, opened 2024-07-26
fix: on project delete with tokens put token deleted in audit log

Tymek:
## About the changes
Use service with audit tracking instead of store directly.

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -22,7 +22,6 @@ import {
     type IFeatureEnvironmentStore,
     type IFeatureNaming,
     type IFeatureToggleStore,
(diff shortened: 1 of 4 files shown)


src/lib/features/project/project-service.ts
```diff
@@ -565,14 +564,14 @@ export default class ProjectService {
             archived: true,
         });
 
-        this.featureToggleService.deleteFeatures(
+        await this.featureToggleService.deleteFeatures(
```
Tymek, line 567: It reminds me that we probably should have a transaction around this.

thomasheartman (approved):
Looks like a good change to me. You'll probably need to update some tests, though 💁🏼

PR #7679, opened 2024-07-26
feat: redirect to new feature flag creation

sjaanus:
Currently found only one instance that is using `create-toggle`. So changed behaviour for that.

### DIFF
--- a/frontend/src/component/application/ApplicationIssues/ApplicationIssues.tsx
+++ b/frontend/src/component/application/ApplicationIssues/ApplicationIssues.tsx
@@ -9,7 +9,7 @@ import {
 } from 'component/providers/AccessProvider/permissions';
 import { useContext } from 'react';
 import AccessContext from 'contexts/AccessContext';
-import { useUiFlag } from '../../../hooks/useUiFlag';
+import { useUiFlag } from 'hooks/useUiFlag';
 
 const WarningContainer = styled(Box)(({ theme }) => ({
     display: 'flex',
--- a/frontend/src/component/application/ApplicationIssues/ApplicationIssues.tsx
+++ b/frontend/src/component/application/ApplicationIssues/ApplicationIssues.tsx
@@ -9,6 +9,7 @@ import {
 } from 'component/providers/AccessProvider/permissions';
 import { useContext } from 'react';
 import AccessContext from 'contexts/AccessContext';
+import { useUiFlag } from '../../../hooks/useUiFlag';
 
 const WarningContainer = styled(Box)(({ theme }) => ({
     display: 'flex',
@@ -108,6 +109,7 @@ type ApplicationIssues =
 
 const FeaturesMissing = ({ features }: IFeaturesMissingProps) => {
     const { hasAccess } = useContext(AccessContext);
+    const improveCreateFlagFlow = useUiFlag('improveCreateFlagFlow');
     const length = features.length;
 
     if (length === 0) {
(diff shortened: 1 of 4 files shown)


frontend/src/component/feature/FeatureForm/FeatureForm.tsx
```diff
@@ -197,6 +197,7 @@ const FeatureForm: React.FC<IFeatureToggleForm> = ({
                     </StyledInputDescription>
                 }
             />
+            // TODO: this can be removed after new create flag flow goes live
             <FeatureProjectSelect
                 value={project}
                 onChange={(projectId) => {
```
sjaanus, line 203: This is using create-toggle, but this is the old creation component that will be removed.

frontend/src/component/common/BreadcrumbNav/BreadcrumbNav.tsx
```diff
@@ -46,6 +46,7 @@ const BreadcrumbNav = () => {
                 item !== 'copy' &&
                 item !== 'features' &&
                 item !== 'features2' &&
+                // TODO: this can be removed after new create flag flow goes live
```
sjaanus, line 49: Currently this will be needed for breadcrumbs, but not in future.

frontend/src/component/application/ApplicationIssues/ApplicationIssues.tsx
```diff
@@ -128,12 +130,25 @@ const FeaturesMissing = ({ features }: IFeaturesMissingProps) => {
...
-                                    to={`/projects/default/create-toggle?name=${feature}`}
-                                >
-                                    Create feature flag
-                                </StyledLink>
+                                <ConditionallyRender
+                                    condition={improveCreateFlagFlow}
+                                    show={
+                                        <StyledLink
+                                            key={feature}
+                                            to={`/projects/default?create=true&name=${feature}`}
```
thomasheartman, line 138: oh, cool! Does the name query param work too? 😄 

I'm guessing we're using the default project because it'll always be there, right? Don't know if there is a better way to select the project, but I also don't know if anyone uses this option 🤷🏼

frontend/src/component/feature/FeatureForm/FeatureForm.tsx
```diff
@@ -197,6 +197,7 @@ const FeatureForm: React.FC<IFeatureToggleForm> = ({
                     </StyledInputDescription>
                 }
             />
+            // TODO: this can be removed after new create flag flow goes live
```
thomasheartman, line 200: Same thing; add the flag name:
```suggestion
            // TODO: this can be removed after new create flag flow goes live. Flag: improveCreateFlagFlow
```

thomasheartman (approved):

frontend/src/component/application/ApplicationIssues/ApplicationIssues.tsx
sjaanus: In the application list, we do not have project info. Ideally we would have it, but it is separate issue.
About query param, probably need to implement now that you say it.

thomasheartman: no, the query param works 😄

thomasheartman: But yeah, I agree. this is a separate issue, which we may or may not wanna tackle later.

thomasheartman: It went to the default project before too, so I think that's fine.

thomasheartman (approved):

frontend/src/component/application/ApplicationIssues/ApplicationIssues.tsx
sjaanus: Ah, it does work.

PR #7680, opened 2024-07-26
Made the Rust tutorial simpler

alvinometric:
This makes the Rust tutorial simpler, by converting to JPEG and Webp, which means
- Compiles on Ubuntu
- Removes the need for the `rgb` and `ravif` crate
- Makes the `process_images` function a lot simpler


website/docs/feature-flag-tutorials/rust/implementing-feature-flags-rust.md
```diff
@@ -200,26 +197,24 @@ cargo add tokio --features full
 cargo add unleash-api-client --features reqwest
 ```
 
-There are a few dependencies, and here's why: We need an HTTP client to make the request, and `serde` to deserialize the Unleash response. Our SDK is constantly polling Unleash to retrieve the value of feature flags, and caches that in memory. 
+There are a few dependencies, and here's why: We need an HTTP client to make the request, and `serde` to deserialize the Unleash response. Our SDK is constantly polling Unleash to retrieve the value of feature flags, performs its own evaluation, and caches that in memory.
```
sighphyre, line 200: Subtle thing here but probably mostly just for your understanding - the flags are cached, the evaluation is done on the fly and the results of that are never cached

sighphyre (approved):
Cheers, LGTM!

PR #7681, opened 2024-07-26
Wip: split out avatar group; use same tooltip for all avatars

thomasheartman:
Extracts the Avatar Group component into a `common` component and adds a standard tooltip to all avatars.

Relates to linear issue 1-2606

This is a suggestion / proof of concept for how we can solve it. While I think we can merge this as is, I'd also be happy to take any discussions on other ways to approach it etc.

## Why are these changes made together?

Because extracting the avatar group without adding the new tooltip data made the existing tooltip misbehave (it'd show up in the top left of the screen, not synced to the avatar in any way).

I probably could have (and still can if you think it's prudent) split it out such that the avatar gets a standardized tooltip first (and disable it for the group card avatars), and split out the avatars in a follow-up. Happy to do that if you think it's better.

## What does this mean? 

It used to be that we had no consistent way of dealing with avatars and tooltips. Some places had them, some places didn't. This change makes it so that all avatars that we can show tooltips for will get the same tooltip.

Previously, we had at least 4 different ways of dealing with tooltips:
- The HTML tooltip (that would be standardized with this PR) in the project flags table
image
- The "title" that you'd get on your user avatar
image
- The group card list tooltip
image
- And sometimes you'd get nothing at all
image

with this change, we'll always show the same kind of tooltip if we can:

image

## What goes in the tooltip? 

We use the `UserAvatar` component for a fair few different things and I didn't want to extract separate components for all the different use cases. Instead, I wanted to get an overview over what we use it for and what is relevant info to show.

I found all the places we used it and tried to form an opinion. 

This tooltip will work with a user's email, name, username, and id. If there is no user (such as for empty avatars and avatars displaying only "+n" for remaining members), we show no tooltip.

Following the example set by the group card avatars, we'll try to use email or username (in that order) as the main bit of text. If the user has an email or a username and also a name, the name will be used as secondary text.

If the user does not have an email or username, but has a name, we'll use the name as the main text.

If the user does not have an email, a username, or a name, we'll try to show "User ID: N" if they have an id.

If they do not have a username, a name, an email, or an ID, we bail out and show nothing.

## Why can you disable the tooltip?

In some cases, you might want to disable the tooltip because you have more information to feed into it. An example of that is in the project flags table, where we want to show more information in cases where the user is 'unknown':

image

## Additional fixes

This PR also adds a few lines of CSS to fix a minor avatar layout bug.

Before:
image

After:
image


frontend/src/interfaces/featureToggle.ts
thomasheartman, line 1: Fixes a typing error (it's been a number all along, just not been marked as such)

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/FeatureOverviewMetaData.tsx
thomasheartman, line 1: we have all the information here, so we can pass the entire user object to get tooltips working.

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx
```diff
@@ -43,101 +23,20 @@ interface IGroupCardAvatarsProps {
...
-                    }
-                />
-                <GroupPopover
-                    open={avatarOpen}
-                    user={popupUser}
-                    anchorEl={anchorEl}
-                    onPopoverClose={onPopoverClose}
-                />
-            </StyledAvatars>
+            <AvatarGroup {...avatarGroupProps} />
```
Tymek, line 39: Feels like this should be tackled by composition, instead of prop inheritance 

```tsx
<GroupCardAvatars header="">
  <AvatarGroup ... />
</GroupCardAvatars>
```

frontend/src/component/common/UserAvatar/UserAvatar.tsx
```diff
@@ -85,4 +97,26 @@ export const UserAvatar: FC<IUserAvatarProps> = ({
             />
         </StyledAvatar>
     );
+
+    const tooltip = disableTooltip ? undefined : tooltipContent(user);
+    if (tooltip) {
```
Tymek, line 102: How about simpler
```suggestion
    if (!disableTooltip) {
      const tooltip = tooltipContent(user);
```

Tymek: Or even more plain, with early return `if (disableTooltip) {return Avatar}`

frontend/src/component/feature/FeatureView/Collaborators.tsx
```diff
@@ -30,11 +29,7 @@ const LastModifiedBy: FC<Collaborator> = ({ id, name, imageUrl }) => {
         <SectionContainer>
             <span>Last modified by</span>
             <LastModifiedByAvatarAndLink>
-                <HtmlTooltip arrow describeChild title={name}>
-                    <span>
-                        <StyledAvatar user={{ id, name, imageUrl }} hideTitle />
-                    </span>
-                </HtmlTooltip>
+                <StyledAvatar user={{ id, name, imageUrl }} />
```
Tymek, line 32: 👏🏻 👍🏻

frontend/src/interfaces/featureToggle.ts
Tymek: Can we use OpenAPI types?

Tymek (approved):
Looks good. Much needed refactoring

frontend/src/component/common/UserAvatar/UserAvatar.tsx
thomasheartman: Well, yes and no. Even if we have a user, there's no guarantee that it has anything we can put into a tooltip. It's defined as 
```tsx
    user?: Partial<
        Pick<IUser, 'id' | 'name' | 'email' | 'username' | 'imageUrl'>
    >;
```

which means that `{}` is also valid, right?

So we can totally do an early return, but we'd also have to check for it later. Another option would be to extract the tooltip into a separate component that returns either a tooltip with the avatar as child or `<>{children}</>` if we don't know how to make a tooltip.

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx
thomasheartman: Right. So in that case, `GroupCardAvatars` doesn't actually provide any avatars; it's just a header? At that point, I don't think calling it `GroupCardAvatars` makes a lot of sense anymore. But maybe it doesn't and we should just delete this whole file?

frontend/src/interfaces/featureToggle.ts
thomasheartman: We *can*, but I don't think this is the time to swap one for the other.

 This interface is marked with:
```js
/**
 * @deprecated use FeatureSchema from openapi
 */
```

However, it's still in use some places, so I don't want to change this one around. Regardless, this is an improvement on the existing (deprecated) type. I changed this because the `FeatureOverviewMetadata` component uses the old interface. 

Fixing this one might be as easy as making a few small updates to `frontend/src/hooks/api/getters/useFeature/useFeature.ts`, but I feel like it'd be better to do that in a separate PR.

thomasheartman: Let's see if this PR passes CI: https://github.com/Unleash/unleash/pull/7698

thomasheartman: As expected, [CI is full of errors](https://github.com/Unleash/unleash/actions/runs/10159620079/job/28094232668?pr=7698) saying that
>  Argument of type 'FeatureSchema' is not assignable to parameter of type 'IFeatureToggle'.
image


I think it may be worth looking at eventually, but I don't want to do that as part of this PR

frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/NewGroupCardAvatars.tsx
Tymek: Yup. It's easier to reason about this when there is 1 less level of abstraction

thomasheartman: Less to reason about, but also more detailed knowledge required elsewhere. I'll have a look at what this'd look like later.

frontend/src/component/common/UserAvatar/UserAvatar.tsx
thomasheartman: Tried a few different ways, but ran into some TS typing errors. I don't want to spend more time debugging that for now, so I'll leave it for now. We can change it later if we want to.

PR #7682, opened 2024-07-26
feat: Use a toggling button for impression data on/off

thomasheartman:
Instead of using a dropdown list for impression data, this PR introduces a new config button that is a switch type button. It will be filled (contained) when impression data is on and will be outlined when impression data is off.

Off:
image

On:
image

### DIFF
--- a/frontend/src/component/common/DialogFormTemplate/ConfigButtons/ToggleConfigButton.tsx
+++ b/frontend/src/component/common/DialogFormTemplate/ConfigButtons/ToggleConfigButton.tsx
@@ -1,26 +1,36 @@
-import type { ConfigButtonProps } from './ConfigButton';
 import { ButtonLabel, StyledTooltipContent } from './ConfigButton.styles';
 import { TooltipResolver } from 'component/common/TooltipResolver/TooltipResolver';
 import { Button } from '@mui/material';
+import type { ReactNode } from 'react';
 
-type ToggleConfigButtonProps = Pick<
-    ConfigButtonProps,
-    'button' | 'description' | 'tooltip'
-> & { onClick: () => void; currentValue: boolean };
+type ToggleConfigButtonProps = {
+    onClick: () => void;
+    currentValue: boolean;
+    label: string;
+    icon: ReactNode;
+    labelWidth?: string;
+    additionalTooltipContent?: ReactNode;
+    tooltip: {
+        header: string;
+        description: string;
+        additionalContent?: ReactNode;
+    };
+};
 
 export function ToggleConfigButton({
     onClick,
     currentValue,
-    button,
+    label,
+    icon,
+    labelWidth,
     tooltip,
-    description,
 }: ToggleConfigButtonProps) {
     return (
         <TooltipResolver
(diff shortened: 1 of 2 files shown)


frontend/src/component/common/DialogFormTemplate/ConfigButtons/ToggleConfigButton.tsx
```diff
@@ -0,0 +1,44 @@
+import type { ConfigButtonProps } from './ConfigButton';
+import { ButtonLabel, StyledTooltipContent } from './ConfigButton.styles';
+import { TooltipResolver } from 'component/common/TooltipResolver/TooltipResolver';
+import { Button } from '@mui/material';
+
+type ToggleConfigButtonProps = Pick<
```
sjaanus, line 6: I feel like you are overusing the ConfigButtonProps props. I feel that you are picking it from a familiar type, but there is actually no reason.

1. It is hard to track to instantly see what types these are.
2. All properties from `button` (for example **additionalTooltipContent**) are not used, so it might be confusing to know.  If a property is added to button, there is no motivation to implement it in all of the files that is using it.

sjaanus (changes requested):
Looks good other than confusing type.

frontend/src/component/common/DialogFormTemplate/ConfigButtons/ToggleConfigButton.tsx
thomasheartman: Ah, good point! I might've missed the additional tooltip content. I'll make sure to add that.

As far as the naming goes, sure. What would you call it? Would you wanna change the entire component name?

thomasheartman: Done in 6531c38038cfedaa1aa816ea2a2e1395061e080f

frontend/src/component/common/DialogFormTemplate/ConfigButtons/ToggleConfigButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Button } from '@mui/material';
+import type { ReactNode } from 'react';
+
+type ToggleConfigButtonProps = {
+    onClick: () => void;
+    currentValue: boolean;
+    label: string;
+    icon: ReactNode;
+    labelWidth?: string;
+    additionalTooltipContent?: ReactNode;
```
sjaanus, line 12: Is this needed?

sjaanus (approved):
LG with comment

frontend/src/component/common/DialogFormTemplate/ConfigButtons/ToggleConfigButton.tsx
thomasheartman, line 12:
```suggestion
```

thomasheartman: No, you're right. Just forgot to delete it. Good catch

PR #7695, opened 2024-07-29
chore: origin middleware

nunogois:
https://linear.app/unleash/issue/2-2489/create-a-first-iteration-of-an-origin-middleware-that-logs-ui-and-api

Small spike around what the first iteration of an "origin middleware" would look like.

No strong feelings all around, so feel free to tell me this is all wrong and we should go a different route. However diving a little bit into it personally helps me wrap my head around it, so it may help you too.

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -30,6 +30,7 @@ import { catchAllErrorHandler } from './middleware/catch-all-error-handler';
 import NotFoundError from './error/notfound-error';
 import { bearerTokenMiddleware } from './middleware/bearer-token-middleware';
 import { auditAccessMiddleware } from './middleware';
+import { originMiddleware } from './middleware/origin-middleware';
 
 export default async function getApp(
     config: IUnleashConfig,
@@ -176,6 +177,8 @@ export default async function getApp(
         baseUriPath,
         rbacMiddleware(config, stores, services.accessService),
     );
+
+    app.use(baseUriPath, originMiddleware(config));
 
     app.use(`${baseUriPath}/api/admin`, auditAccessMiddleware(config));
     app.use(
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -178,7 +178,7 @@ export default async function getApp(
         rbacMiddleware(config, stores, services.accessService),
     );
 
-    app.use(baseUriPath, originMiddleware(config));
+    app.use(`${baseUriPath}/api/admin`, originMiddleware(config));
 
     app.use(`${baseUriPath}/api/admin`, auditAccessMiddleware(config));
     app.use(
--- a/src/lib/middleware/origin-middleware.ts
+++ b/src/lib/middleware/origin-middleware.ts
@@ -1,0 +1,27 @@
+import type { Request, Response, NextFunction } from 'express';
+import type { IUnleashConfig } from '../types';
+
+export const originMiddleware = ({
+    getLogger,
+    flagResolver,
+}: Pick<IUnleashConfig, 'getLogger' | 'flagResolver'>) => {
+    const logger = getLogger('/middleware/origin-middleware.ts');
+    logger.debug('Enabling origin middleware');
+
+    return (req: Request, _: Response, next: NextFunction) => {
+        if (!flagResolver.isEnabled('originMiddleware')) {
+            return next();
+        }
+
+        const isUI = !req.headers.authorization;
+
+        if (isUI) {
+            logger.debug('UI request');
+        } else {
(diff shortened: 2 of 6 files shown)


src/lib/app.ts
```diff
@@ -177,6 +178,8 @@ export default async function getApp(
         rbacMiddleware(config, stores, services.accessService),
     );
 
+    app.use(baseUriPath, originMiddleware(config));
```
nunogois, line 181: Placement of the middleware is important. I think this works nicely (after auth and before most other things), but we may need to move it.

gastonfournier: Should we use 
```suggestion
    app.use("${baseUriPath}/admin", originMiddleware(config));
```?

File: src/lib/middleware/origin-middleware.ts [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+}: Pick<IUnleashConfig, 'getLogger' | 'flagResolver'>) => {
+    const logger = getLogger('/middleware/origin-middleware.ts');
+    logger.debug('Enabling origin middleware');
+
+    return (req: Request, _: Response, next: NextFunction) => {
+        if (!flagResolver.isEnabled('originMiddleware')) {
+            return next();
+        }
+
+        const isUI = !req.headers.authorization;
```
gastonfournier, line 16: Is there something else we'd check? I assume we're already authorized at this point, so probably this is enough

gastonfournier (commented):
Overall it looks good, I think this is what I had in mind as an initial step. Now, should this be OSS available? I'm just thinking that OSS might not need this and neither self-hosted. So I'm just wondering if this should be in our cloud repo

src/lib/app.ts
nunogois: Good question, it depends on which routes we want to cover. This covers everything, but maybe we want to be more selective.

nunogois:
> Overall it looks good, I think this is what I had in mind as an initial step. Now, should this be OSS available? I'm just thinking that OSS might not need this and neither self-hosted. So I'm just wondering if this should be in our cloud repo

I don't think it hurts to have this in OSS, especially since it's behind a flag. This would allow us to easily cover OSS and self-hosted if we decide to do so at a later stage. I think it will depend on how we decide to evolve it.

src/lib/middleware/origin-middleware.ts
nunogois: Yep, I also think it's enough, and I believe the placement means this only runs after authorization.

gastonfournier: > I may be missing some edge cases

If there's any I think this will help us figure that out (somehow)

src/lib/app.ts
sighphyre: I'm with Gaston on this one, I don't see a need to look at anything but what we want, and we know what we want lives on the admin API

gastonfournier: My take is to just cover /api/admin because I looked at our access logs and it's difficult to actually get any information out of them. We can query 1minute in loki but anything else becomes impossible. /admin should have less frequency and more value for our task

src/lib/middleware/origin-middleware.ts
```diff
@@ -0,0 +1,27 @@
...
+
+        if (isUI) {
+            logger.debug('UI request');
```
sighphyre, line 19: That's gonna be a lotta logs! Do we need "UI request" and "API request"?

gastonfournier: Maybe we only log on POST/PUT/DELETE?

src/lib/app.ts
nunogois: Sounds good, addressed in https://github.com/Unleash/unleash/pull/7695/commits/d46d14f7a63a14d4151593125606cede72d849d2

src/lib/middleware/origin-middleware.ts
```diff
@@ -0,0 +1,27 @@
...
+        } else {
+            logger.debug('API request');
+            logger.debug({ userAgent: req.headers['user-agent'] });
```
gastonfournier, line 22: Let's collapse these into one log line

nunogois: Like this? https://github.com/Unleash/unleash/pull/7695/commits/fa6c7155275b97308be7051ce61d002e5767440d

gastonfournier (approved):
I'm happy to try this, I'd just try to log only on not GET requests, cause I believe that's what we're interested on, but also happy to test this out and then remove GET requests

PR #7697, opened 2024-07-30
security: bump ws dependency

gastonfournier:
Should fix https://github.com/Unleash/unleash/security/dependabot/146 and https://github.com/Unleash/unleash/security/dependabot/147


chriswk (approved):
LGTM

PR #7698, opened 2024-07-30
fix: use FeatureSchema instead of deprecated IFeatureToggle in `useFeature`

thomasheartman:
`IFeatureToggle` has been deprecated for a hot minute, so let's stop using it in the `useFeature` hook.


thomasheartman:
this would require a lot of updates. Not touching this now; this was an experiment.

PR #7706, opened 2024-07-31
feat: rollback transaction wrapper

kwasniew:
## About the changes

Adding `withRollback` function that performs a transactional operation on a service and then rolls it back.

Why do we need it?
We're implementing playground for change requests to preview the changes. We gonna apply a change request, run playground and roll back the transaction to simulate what would happen if the change request was applied.

```ts
await this.changeRequestPreviewService.rollback(service => {
            return service.preview(...);
        });
```

Design decisions:
* I decided to use the most strict transaction isolation level of `serializable` but we may consider relaxing it to the postgres default of `read committed` if phantom reads and non-repeatable reads are not a problem
* `withRollback` is modelled after `withTransactional` to have symmetrical API. The only difference is that we always roll back at the end of the transaction

### DIFF
--- a/src/lib/db/transaction.ts
+++ b/src/lib/db/transaction.ts
@@ -36,8 +36,13 @@ export type DeferredServiceFactory<S> = (db: Knex) => S;
 export type ServiceFactory<S> = (
     config: IUnleashConfig,
 ) => DeferredServiceFactory<S>;
+
 export type WithTransactional<S> = S & {
     transactional: <R>(fn: (service: S) => R) => Promise<R>;
+};
+
+export type WithRollback<S> = S & {
+    rollback: <R>(fn: (service: S) => R) => Promise<R>;
 };
 
 /**
@@ -77,6 +82,25 @@ export function withTransactional<S>(
             const transactionalService = serviceFactory(trx);
             return fn(transactionalService);
         });
+
+    return service;
+}
+
+export function withRollback<S>(
+    serviceFactory: (db: Knex) => S,
+    db: Knex,
+): WithRollback<S> {
+    const service = serviceFactory(db) as WithRollback<S>;
+
+    service.rollback = async <R>(fn: (service: S) => R) => {
+        const trx = await db.transaction({ isolationLevel: 'serializable' });
+        try {
+            const transactionService = serviceFactory(trx);
+            return fn(transactionService);
+        } finally {
+            await trx.rollback();
+        }
+    };
 
     return service;
 }


thomasheartman (approved):

src/lib/db/transaction.ts
```diff
@@ -81,6 +86,25 @@ export function withTransactional<S>(
...
 }
 
+export function withRollback<S>(
+    serviceFactory: (db: Knex) => S,
+    db: Knex,
+): WithRollback<S> {
+    const service = serviceFactory(db) as WithRollback<S>;
+
+    service.rollback = async <R>(fn: (service: S) => R) => {
+        const trx = await db.transaction({ isolationLevel: 'serializable' });
```
gastonfournier, line 96: The highest possible isolation level, is good for keeping consistency but it might backfire in terms of performance and resource locking and reducing the concurrency level.

It's true we're not going to use it that frequently, but the moment it's used it might have a negative impact in the system performance overall.

gastonfournier (commented):
I have some concerns about the approach, and I can be wrong or overthinking this, but it'd be good to test this out with some load and identify some metrics that might suggest this is going well or not well

src/lib/db/transaction.ts
kwasniew: that's why I'm considering using 'read committed' which is postgres default. I'm starting with the most strict level for the experimentation purposes. 

I can give the context where it's gonna be used and why the tx may take a few seconds:
* you open transaction
* you apply your change request
* you run playground evaluation (may take a few seconds so bad for other transactions)
* you roll back the change

src/lib/db/transaction.ts
```diff
@@ -81,6 +86,25 @@ export function withTransactional<S>(
...
+        try {
+            const transactionService = serviceFactory(trx);
+            return fn(transactionService);
+        } finally {
+            await trx.rollback();
+        }
+    };
```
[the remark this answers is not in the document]
kwasniew: Thanks for the feedback. I see your point and I will consider renaming it to sth else. I couldn't find a good name.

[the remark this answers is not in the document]
thomasheartman: Maybe even `withTemporaryTransaction` or something? 🤷🏼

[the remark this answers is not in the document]
kwasniew: I have an issue with temporary/transient because all transactions are temporary for the duration of the transaction. I need something that explains that this transaction with be rolled back. `withRolledBackTransaction`?

src/lib/db/transaction.ts
gastonfournier: Spending a few seconds worries me with the current isolation level, but experimenting seems fine, I'd try to see if we suddenly spawn up a lot of new pods when actively hitting this functionality while at the same time we have some load on the client endpoints and maybe the UI.

Perhaps read committed works well because it prevents dirty reads and that's all we really want here, right?

kwasniew: This is sth behind a flag so I'll do internal testing first and probably downgrade the isolation level

PR #7708, opened 2024-07-31
feat: new event search

sjaanus:
This introduces the new event search API, with paging.


src/lib/features/events/event-service.ts
```diff
@@ -40,9 +43,38 @@ export default class EventService {
         };
     }
 
-    async searchEvents(search: SearchEventsSchema): Promise<IEventList> {
-        const totalEvents = await this.eventStore.filteredCount(search);
-        const events = await this.eventStore.searchEvents(search);
+    async deprecatedSearchEvents(
```
sjaanus, line 46: Old method, just renaming. No logic changes here.

src/lib/features/events/event-store.ts
```diff
@@ -147,6 +157,22 @@ class EventStore implements IEventStore {
         }
     }
 
+    async searchEventsCount(
```
sjaanus, line 160: This is method to give us total event count across all pages.

src/lib/features/feature-search/feature-search-service.ts
```diff
@@ -38,27 +38,11 @@ export class FeatureSearchService {
         };
     }
 
-    parseOperatorValue = (field: string, value: string): IQueryParam | null => {
```
sjaanus, line 41: Moving this to search-utils, because it can be reused by other searches.

src/lib/routes/admin-api/event.ts
```diff
@@ -98,6 +108,27 @@ export default class EventController extends Controller {
         this.route({
             method: 'post',
             path: '/search',
+            handler: this.deprecatedSearchEvents,
+            permission: NONE,
+            middleware: [
+                openApiService.validPath({
+                    operationId: 'deprecatedSearchEvents',
+                    tags: ['Events'],
+                    deprecated: true,
```
sjaanus, line 117: Deprecating old search.

src/test/fixtures/fake-event-store.ts
```diff
@@ -79,23 +85,11 @@ class FakeEventStore implements IEventStore {
         return this.events;
     }
 
-    async searchEvents(): Promise<IEvent[]> {
+    async deprecatedSearchEvents(): Promise<IEvent[]> {
         throw new Error('Method not implemented.');
     }
-
-    async getForFeatures(
```
sjaanus, line 86: This method was unused, just cleaning up

src/lib/features/events/event-service.ts
```diff
@@ -110,4 +142,48 @@ export default class EventService {
             });
         }
     }
+
+    convertToQueryParams = (
```
thomasheartman, line 146: Not a big deal, but the name here is confusing, because we're accepting query params but we're also returning query ... params?

Is one of them URL query params, and the other is db query params?

Maybe we could do something like `toDbQuery` or something?

src/lib/features/events/event-store.ts
```diff
@@ -147,6 +157,22 @@ class EventStore implements IEventStore {
...
+        const query = this.buildSearchQuery(params, queryParams);
+        const count = await query.count().first();
+        if (!count) {
+            return 0;
+        }
+        if (typeof count.count === 'string') {
+            return Number.parseInt(count.count, 10);
+        } else {
+            return count.count;
+        }
```
thomasheartman, line 173: Why is this necessary? Does the db sometimes return the count as a string and other times as a number? Maybe it's the TS type that's doing it?

src/lib/features/feature-search/search-utils.ts
```diff
@@ -1,10 +1,13 @@
...
+import type {
+    IQueryOperator,
+    IQueryParam,
+} from '../feature-toggle/types/feature-toggle-strategies-store-type';
 
 export interface NormalizeParamsDefaults {
     limitDefault: number;
     maxLimit?: number; // Optional because you might not always want to enforce a max limit
-    sortByDefault: string;
+    sortByDefault?: string;
```
thomasheartman, line 10: Maybe I'm missing some context here, but why are you making this optional?

src/lib/openapi/spec/event-search-query-parameters.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    {
+        name: 'createdBy',
+        schema: {
+            type: 'string',
+            example: 'IS:2',
+            pattern: '^(IS|IS_ANY_OF):(.*?)(,([a-zA-Z0-9_]+))*$',
+        },
+        description:
+            'The ID of event creator to filter by. The creators can be specified with an operator. The supported operators are IS, IS_ANY_OF.',
```
thomasheartman, line 56: Can we bring this in line with the other is|is_any_of operators?

```suggestion
            'Filter by the ID of the event creator, using supported operators: IS, IS_ANY_OF.',
```

src/lib/openapi/spec/event-search-query-parameters.ts
```diff
@@ -0,0 +1,105 @@
...
+        },
+        description:
+            'The ending date of the creation date range in IS:yyyy-MM-dd format',
+        in: 'query',
+    },
+    {
+        name: 'offset',
+        schema: {
+            type: 'string',
+            example: '50',
```
thomasheartman, line 85: If this has a default value, is there a reason not to put it in the schema?

```suggestion
            example: '50',
            default: '0',
```

src/lib/openapi/spec/event-search-query-parameters.ts
```diff
@@ -0,0 +1,105 @@
...
+        },
+        description:
+            'The number of features to skip when returning a page. By default it is set to 0.',
+        in: 'query',
+    },
+    {
+        name: 'limit',
+        schema: {
+            type: 'string',
+            example: '50',
```
thomasheartman, line 95: Same as above

```suggestion
            example: '50',
            default: '50',
```

src/lib/openapi/spec/event-search-response-schema.ts
```diff
@@ -0,0 +1,34 @@
+import type { FromSchema } from 'json-schema-to-ts';
+import { eventSchema } from './event-schema';
+import { tagSchema } from './tag-schema';
+
+export const eventSearchResponseSchema = {
+    $id: '#/components/schemas/eventsSearchResponseSchema',
+    type: 'object',
+    additionalProperties: false,
+    required: ['events', 'total'],
+    description: 'A list of events that has happened in the system',
```
thomasheartman, line 10: Small suggestion here. The most important thing is changing "has" to "have".

```suggestion
    description: 'A list of events that have been registered by the system',
```

src/lib/routes/admin-api/event.ts
```diff
@@ -98,6 +108,27 @@ export default class EventController extends Controller {
...
+                    summary: 'Search for events (deprecated)',
+                    description:
+                        'Allows searching for events matching the search criteria in the request body',
```
thomasheartman, line 120: Can we tell users what to use instead of this one?

```suggestion
                        'Allows searching for events matching the search criteria in the request body. This operation is deprecated. You should perform a GET request to the same endpoint with your query encoded as query parameters instead.',
```

src/lib/routes/admin-api/event.ts
```diff
@@ -152,7 +187,9 @@ export default class EventController extends Controller {
         res: Response<FeatureEventsSchema>,
     ): Promise<void> {
         const feature = req.params.featureName;
-        const eventList = await this.eventService.searchEvents({ feature });
+        const eventList = await this.eventService.deprecatedSearchEvents({
```
thomasheartman, line 190: Are we planning on updating these method calls too or would that be too much work?

src/lib/openapi/spec/event-search-query-parameters.ts
```diff
@@ -0,0 +1,105 @@
...
+        },
+        description:
+            'The number of feature environments to return in a page. By default it is set to 50.',
```
thomasheartman, line 98: Let's mention the max here too.

```suggestion
            'The number of events to return in a page. By default it is set to 50. The maximum  is 1000.',
```

src/lib/routes/admin-api/event.ts
```diff
@@ -188,4 +227,33 @@ export default class EventController extends Controller {
...
+                maxLimit: 1000,
+            },
+        );
+
+        const { events, totalEvents } = await this.eventService.searchEvents({
+            ...req.body,
+            offset: normalizedOffset,
+            limit: normalizedLimit,
+        });
+
```
thomasheartman, line 248: This bit here is a little hard to follow. What do you get back from `normalizeQueryParams`? If it's the entire object, couldn't you do something like:

```suggestion
        const normalizedQuery = normalizeQueryParams(
            req.query,
            {
                limitDefault: 50,
                maxLimit: 1000,
            },
        );

        const { events, totalEvents } = await this.eventService.searchEvents(normalizedQuery);

```

But maybe it doesn't work like that? 

Anyway, we should be able to deal with the default value in the schema, but because it's a string, there's no way to set a max value, unfortunately.

thomasheartman (approved):
Overall, I think this looks great. I've got some suggestions, especially on the OpenAPI side, as well as some questions I'd love to get an answer to, but nothing blocking.

PR #7709, opened 2024-07-31
chore: request origin prom metrics

nunogois:
https://linear.app/unleash/issue/2-2501/adapt-origin-middleware-to-stop-logging-ui-requests-and-start

This adapts the new origin middleware to stop logging UI requests (too noisy) and adds new Prometheus metrics.

image

This should allow us to better analyze this data. If we see a lot of API requests, we can dive into the logs for that instance and check the logged data, like the user agent.

This PR adds some helper methods to make listening and emitting metric events more strict in terms of types. I think it's a positive change aligned with our scouting principle, but if you think it's complex and does not belong here I'm happy with dropping it.

### DIFF
--- a/src/lib/metric-events.ts
+++ b/src/lib/metric-events.ts
@@ -1,3 +1,5 @@
+import type EventEmitter from 'events';
+
 const REQUEST_TIME = 'request_time';
 const DB_TIME = 'db_time';
 const FUNCTION_TIME = 'function_time';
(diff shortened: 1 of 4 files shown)


src/lib/metric-events.ts
```diff
@@ -9,6 +11,51 @@ const PROXY_REPOSITORY_CREATED = 'proxy_repository_created';
...
+    | typeof FUNCTION_TIME
+    | typeof SCHEDULER_JOB_TIME
+    | typeof FEATURES_CREATED_BY_PROCESSED
+    | typeof EVENTS_CREATED_BY_PROCESSED
+    | typeof FRONTEND_API_REPOSITORY_CREATED
+    | typeof PROXY_REPOSITORY_CREATED
+    | typeof PROXY_FEATURES_FOR_TOKEN_TIME
+    | typeof STAGE_ENTERED
+    | typeof EXCEEDS_LIMIT
+    | typeof REQUEST_ORIGIN;
```
nunogois, line 28: I think it could be nice to simplify this in the future. E.g. for feature flags in `experimental.ts` we have the literals in a single type instead of declaring a constant for each flag name.

src/lib/metrics.ts
```diff
@@ -694,6 +700,16 @@ export default class MetricsMonitor {
...
 
+        events.onMetricEvent(
+            eventBus,
+            events.REQUEST_ORIGIN,
+            ({ type, method }) => {
+                if (flagResolver.isEnabled('originMiddleware')) {
+                    requestOriginCounter.increment({ type, method });
+                }
+            },
+        );
```
nunogois, line 711: This could later be simplified to:

```ts
events.onMetricEvent(eventBus, events.REQUEST_ORIGIN, requestOriginCounter.increment)
``` 

Which would still satisfy our types.

sighphyre (approved):
LGTM

PR #7713, opened 2024-07-31
fix: strategy parameters UI

Tymek:
## About the changes
Old versions of Unleash allow for creating "Gradual Rollout" strategies without `groupId` or `stickiness`. UI will now populate those fields, not getting stuck when editing strategies without said fields.

Internal ticket: [1-2198](https://linear.app/unleash/issue/1-2198/strategy-configuration-auto-add-stickiness-and-groupid-if-missing)

### DIFF
--- a/frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.tsx
+++ b/frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.tsx
@@ -23,7 +23,7 @@ interface IFlexibleStrategyProps {
     updateParameter: (field: string, value: string) => void;
     context: any;
     editable: boolean;
-    errors: IFormErrors;
+    errors?: IFormErrors;
 }
 
 const StyledBox = styled(Box)(({ theme }) => ({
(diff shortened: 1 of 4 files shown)


frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.tsx
```diff
@@ -77,18 +81,18 @@ const FlexibleStrategy = ({
...
 
     if (parameters.stickiness === '') {
         onUpdate('stickiness')(stickiness);
     }
 
     useEffect(() => {
-        if (isDefaultStrategyEdit && !parameters.groupId) {
-            onUpdate('groupId')('');
+        if (!parameters.groupId) {
+            onUpdate('groupId')(isDefaultStrategyEdit ? '' : featureId);
```
thomasheartman, line 93: Does this handle the case where you've emptied the field yourself?

thomasheartman (commented):
Nice one! Can we get some tests in for this? Just looking at the code, it's hard to judge whether it's doing what it should. Can we test that:
1. If the strategy has no stickiness or groupId when opened, those will be set to default values when you save it.
2. If you manually empty the groupId field, we don't save the default value there, but instead give you an error message when you try to save it?

frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.tsx
Tymek: Yes. It will show an error 

image

thomasheartman: Cool!

thomasheartman (approved):
I'd love to see some tests for this behavior before it goes in, but I'll give it an approval so as not to block you. And I guess that even if this didn't work, it wouldn't be a worse experience than before, so 🤷🏼

PR #7718, opened 2024-08-01
feat: create gauges for all resource limits

thomasheartman:
This PR adds Grafana gauges for all the existing resource limits. The primary purpose is to be able to use this in alerting. Secondarily, we can also use it to get better insights into how many customers have increased their limits, as well as how many people are approaching their limit, regdardless of whether it's been increased or not.

### Implementation

The first approach I took (in https://github.com/Unleash/unleash/pull/7718/commits/87528b4c6785899d7afec34721cb9f0df9bf0bfc), was to add a new gauge for each resource limit. However, there's a lot of boilerplate for it. 

I thought doing it like this (the current implementation) would make it easier. We should still be able to use the labelName to collate this in Grafana, as far as I understand? As a bonus, we'd automatically get new resource limits when we add them to the schema.

``` tsx
        const resourceLimit = createGauge({
            name: 'resource_limit',
            help: 'The maximum number of resources allowed.',
            labelNames: ['resource'],
        });

        // ...

        for (const [resource, limit] of Object.entries(config.resourceLimits)) {
            resourceLimit.labels({ resource }).set(limit);
        }
```

That way, when checking the stats, we should be able to do something like this:

``` promql
resource_limit{resource="constraintValues"}
```

### Do we need to reset gauges?

I noticed that we reset gauges before setting values in them all over the place. I don't know if that's necessary. I'd like to get that double clarified before merging this.

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -353,6 +353,71 @@ export default class MetricsMonitor {
             labelNames: ['type', 'method'],
         });
 
+        const resourceLimitSegmentValues = createGauge({
+            name: 'resource_limit_segment_values',
+            help: 'The maximum number of segment values allowed.',
+        });
+        const resourceLimitStrategySegments = createGauge({
+            name: 'resource_limit_strategy_segments',
+            help: 'The maximum number of segments allowed.',
+        });
+        const resourceLimitSignalEndpoins = createGauge({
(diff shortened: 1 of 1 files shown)


gardleopard (approved):
Looks like what we talked about :)

PR #7720, opened 2024-08-01
docs: add subpages of edge documentation

gastonfournier:
When all edge documentation was in a single page, we had all of it here: https://docs.getunleash.io/reference/unleash-edge but after splitting up https://github.com/Unleash/unleash-edge/pull/475 some of that documentation is not present in our docs and we should probably have them as sub-pages

This change allows us to include subpages for external documentation. Note that the key of the subpage needs to match the remote path

Checkout the preview: https://unleash-docs-git-add-edge-subpages-unleash-team.vercel.app/
image

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -16,7 +16,8 @@
     "clear": "docusaurus clear",
     "serve": "docusaurus serve",
     "write-translations": "docusaurus write-translations",
-    "write-heading-ids": "docusaurus write-heading-ids"
+    "write-heading-ids": "docusaurus write-heading-ids",
+    "test": "NODE_ENV=test node --trace-warnings ../node_modules/.bin/jest remote-content"
   },
   "dependencies": {
     "@docusaurus/core": "2.3.1",
--- a/website/remote-content/shared.js
+++ b/website/remote-content/shared.js
@@ -149,5 +149,6 @@ This content was generated on <time dateTime="${generationTime.toISOString()}">$
 
 module.exports.getUrls = (documents) =>
     Object.entries(documents).map(
-        ([repo, { branch }]) => `${repo}/${branch}/README.md`,
+        ([repo, { branch, remotePath }]) =>
+            `${repo}/${branch}/${remotePath ?? 'README.md'}`,
     );
--- a/website/remote-content/shared.js
+++ b/website/remote-content/shared.js
@@ -132,7 +132,7 @@ custom_edit_url: ${data.repoUrl}/edit/${data.branch}/${subpage ? subpageKey : 'R
 ---
 
 :::info Generated content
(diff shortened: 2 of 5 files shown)


website/package.json
```diff
@@ -16,7 +16,8 @@
     "clear": "docusaurus clear",
     "serve": "docusaurus serve",
     "write-translations": "docusaurus write-translations",
-    "write-heading-ids": "docusaurus write-heading-ids"
+    "write-heading-ids": "docusaurus write-heading-ids",
+    "test": "NODE_ENV=test node --trace-warnings ../node_modules/.bin/jest remote-content"
```
gastonfournier, line 20: Ok, this is very cheap of me, I could have added jest as a dependency inside website, but this works well for local testing. Not intended to be used in CI, it just gives me a faster feedback loop than building the contents. I can remove it but I wanted to keep the tests for the review process

website/remote-content/shared.js
```diff
@@ -89,13 +89,16 @@ module.exports.modifyContent =
     }) =>
     (filename, content) => {
         const data = getRepoDataFn(filename);
+        const subpageKey = filename.replace(`${data.name}/${data.branch}/`, '');
+        const subpage = data.subPages?.[subpageKey];
```
gastonfournier, line 93: Try to get a subpage from the filename

website/remote-content/shared.js
```diff
@@ -148,6 +153,9 @@ This content was generated on <time dateTime="${generationTime.toISOString()}">$
...
 
 module.exports.getUrls = (documents) =>
-    Object.entries(documents).map(
-        ([repo, { branch }]) => `${repo}/${branch}/README.md`,
-    );
+    Object.entries(documents).flatMap(([repo, { branch, subPages }]) => [
+        `${repo}/${branch}/README.md`,
+        ...(Object.keys(subPages ?? {}).map(
+            (remotePath) => `${repo}/${branch}/${remotePath}`,
+        ) ?? []),
```
gastonfournier, line 160: Now we not only generate 1 url per object, we also generate urls for the subpages

chriswk (approved):
LGTM

sighphyre (approved):
Wow, cool, tests for docs. LGTM

website/package.json
gastonfournier: yes, I'm not using it during build time, so right now it's a handy method to test locally, but the tests are not executed on CI, maybe we should? For now, from no test to manual tests I'm ok.

thomasheartman: Right. Yeah, I think we could do it in CI, but we'd probably need as local Jest for that , then, yeah? We haven't really had much in the way of tests in the docs before, so we'd have to set it up, but it sounds smart. But then again, how often do we change this? About once every two years? Still, be nice to have it checked when it does change

PR #7730, opened 2024-08-02
feat: copy strategy to current environment

kwasniew:
## About the changes

You can copy strategy not only to other environments but also to the current one
Screenshot 2024-08-02 at 08 57 31

Use case: you configure variants once and want to have a few strategies with the same variants without having to type them manually
Screenshot 2024-08-02 at 08 58 46

Details:
* it's a UI change only, we're adding current environment to the list of envs you can copy to
* current env is the last one on the list and has a text: "Copy to current"
* I checked and it also supports change requests out of the box
* tooltip changes from "copy to another environment" to "copy to environment"

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/CopyStrategyIconMenu/CopyStrategyIconMenu.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/CopyStrategyIconMenu/CopyStrategyIconMenu.tsx
@@ -120,7 +120,7 @@ export const CopyStrategyIconMenu: VFC<ICopyStrategyIconMenuProps> = ({
                 }
             />
             <Tooltip
-                title={`Copy to another environment${
+                title={`Copy to environment${
                     enabled ? '' : ' (Access denied)'
                 }`}
             >
@@ -150,7 +150,7 @@ export const CopyStrategyIconMenu: VFC<ICopyStrategyIconMenuProps> = ({
                     'aria-labelledby': `copy-strategy-icon-menu-${strategy.id}`,
                 }}
             >
-                {environments.map((environment) => {
+                {[...environments, environmentId].map((environment) => {
                     const access = checkAccess(
                         CREATE_FEATURE_STRATEGY,
                         environment,
@@ -179,7 +179,10 @@ export const CopyStrategyIconMenu: VFC<ICopyStrategyIconMenuProps> = ({
                                         }
                                     />
                                     <ListItemText>
-                                        Copy to {environment}
+                                        Copy to{' '}
+                                        {environment === environmentId
+                                            ? 'current'
+                                            : environment}
                                     </ListItemText>
                                 </MenuItem>
                             </div>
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/CopyStrategyIconMenu/CopyStrategyIconMenu.tsx
```diff
@@ -150,7 +150,7 @@ export const CopyStrategyIconMenu: VFC<ICopyStrategyIconMenuProps> = ({
                     'aria-labelledby': `copy-strategy-icon-menu-${strategy.id}`,
                 }}
             >
-                {environments.map((environment) => {
+                {[...environments, environmentId].map((environment) => {
```
thomasheartman, line 153: So the environments that come through as the function parameter are "all other environments", but not this one, yeah? So we're just adding the ID of this one to the list. 

Edit: right, an environment's **name** is also it's ID, right?

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/CopyStrategyIconMenu/CopyStrategyIconMenu.tsx
```diff
@@ -179,7 +179,10 @@ export const CopyStrategyIconMenu: VFC<ICopyStrategyIconMenuProps> = ({
                                         }
                                     />
                                     <ListItemText>
-                                        Copy to {environment}
+                                        Copy to{' '}
+                                        {environment === environmentId
+                                            ? 'current'
+                                            : environment}
```
thomasheartman, line 185: There's an edge case, here, right? 
- If you have a different environment named "current", you'll have two "Copy to current" options.

I don't think it's very important ("current" is a weird name for an env, anyway), but it might be nice to consider if we can distinguish this option from the other options somehow 🤷🏼

thomasheartman (approved):
Nice one 😄 A small question and a suggestion for potential future stuff, but absolutely nothing blocking.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/CopyStrategyIconMenu/CopyStrategyIconMenu.tsx
kwasniew: I tracked the code invocations and we pass current env and other envs as two different props. I merged those two with the current one forced to be the last one on the list. name/id is the same thing for environments

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/CopyStrategyIconMenu/CopyStrategyIconMenu.tsx
kwasniew: really good catch. I was also considering "Duplicate in current" to make it stand out from the list

kwasniew: Screenshot 2024-08-02 at 09 24 18

PR #7732, opened 2024-08-02
chore: use EventSchema instead of IEvent

thomasheartman:
Changes the type used by the useEventSearch hook to be `EventSchema`
from OpenAPI instead. This is more accurate with what we're actually
getting. And crucially for the event log search, it contains the
`createdByUserId` property that we need to filter out events.

It's mostly a straightforward find and replace except for one instance where we need to do some extra fiddling. There's an inline comment explaining that.


frontend/src/component/events/EventDiff/EventDiff.tsx
```diff
@@ -17,7 +16,7 @@ interface IEventDiffResult {
 }
 
 interface IEventDiffProps {
-    entry: Partial<IEvent>;
+    entry: { data: unknown; preData: unknown };
```
thomasheartman, line 19: This one is kinda wonky. We only use `data` and `preData`. However, they're typed as `unknown` at the call site. In `IEvent`, both of these were typed as `any`, but in `EventSchema`, they're typed as `{[key: string] any} | null`.

It's almost the same, but not quite, so TS throws an error if we just sub in `EventSchema` for `IEvent` here.

So what I've done is to specify the fields we actually care about and typed them as `unknown`. We don't know anything more about them than we do about `any` properties, but at least now TS will stop you if you try to access any properties on those objects 🤷🏼

sjaanus (approved):

PR #7733, opened 2024-08-02
feat: add event types to filter button

thomasheartman:
Add event types to the event log filter button.

The event types are taken from the OpenAPI definition.


sjaanus (approved):
Epic

PR #7735, opened 2024-08-02
tests: flexible strategy component stickiness and groupId

Tymek:
Follow-up for #7713

### DIFF
--- a/frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.test.tsx
+++ b/frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.test.tsx
@@ -64,7 +64,7 @@ test('manipulates the rollout slider', async () => {
     expect(groupIdInput).toHaveValue('newGroupId');
 });
 
-test('if stickienss or groupId not present, fill it with defaults', async () => {
+test('if stickiness or groupId not present, fill it with defaults', async () => {
     const updateParameter = vi.fn();
     const Wrapper = () => (
         <Routes>
--- a/frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.test.tsx
+++ b/frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.test.tsx
@@ -1,5 +1,6 @@
+import { vi } from 'vitest';
 import { useState } from 'react';
-import { screen, fireEvent } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.test.tsx
```diff
@@ -61,3 +63,75 @@ test('manipulates the rollout slider', async () => {
     expect(slider).toHaveValue('50');
     expect(groupIdInput).toHaveValue('newGroupId');
 });
+
+test('if stickienss or groupId not present, fill it with defaults', async () => {
```
thomasheartman, line 67: 😄 
```suggestion
test('if stickiness or groupId not present, fill it with defaults', async () => {
```

frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.test.tsx
```diff
@@ -61,3 +63,75 @@ test('manipulates the rollout slider', async () => {
...
+    setupApi();
+
+    render(<Wrapper />, {
+        route: '/projects/default/features/test',
+    });
+
+    await waitFor(() => {
+        expect(updateParameter).toHaveBeenCalledWith('stickiness', 'default');
+        expect(updateParameter).toHaveBeenCalledWith('groupId', 'test');
+    });
```
thomasheartman, line 96: So do we add the default values when they open the form or when they try to save it? I think both can be potentially confusing:

Assuming that the strategy is missing stickiness or groupId.

If we add them when they open, they might get a false impression and think that the strategy *has* these values. If they just close the form without saving, then nothing'll happen, and the configuration will still be invalid.

If we add them on save instead, they'll look empty, but they'll be set without any warning when you save. You'll also be in the position where you can save it with empty values sometimes, but not others. 

Both have drawbacks, so I'm not gonna say which one we should do. I think it's enough of an edge case anyway, that it's probably fine.

thomasheartman (approved):

frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.test.tsx
Tymek: When they open the form. Test has to be written this way, because this component doesn't handle state for those parameters

PR #7739, opened 2024-08-02
feat: event search on new endpoint, first test

sjaanus:
Changed the url of event search to search/events to align with search/features. With that created a search controller to keep all searches under there.
Added first test.

### DIFF
--- a/src/lib/features/events/event-service.ts
+++ b/src/lib/features/events/event-service.ts
@@ -20,7 +20,6 @@ export interface IEventSearchParams {
     type?: string;
     offset: number;
     limit: number;
-    // sortBy?: string;
 }
 
 export default class EventService {
--- a/src/lib/features/events/event-service.ts
+++ b/src/lib/features/events/event-service.ts
@@ -8,9 +8,20 @@ import type EventEmitter from 'events';
 import type { IApiUser, ITag, IUser } from '../../types';
 import { ApiTokenType } from '../../types/models/api-token';
 import { EVENTS_CREATED_BY_PROCESSED } from '../../metric-events';
-import type { EventSearchQueryParameters } from '../../openapi/spec/event-search-query-parameters';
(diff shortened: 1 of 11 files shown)


src/lib/routes/admin-api/event.ts
```diff
@@ -125,26 +115,6 @@ export default class EventController extends Controller {
                 }),
             ],
         });
-
```
sjaanus, line 128: This is just moved to separate controller.

src/lib/features/events/event-store.ts
```diff
@@ -89,8 +89,8 @@ export interface IEventTable {
 
 export interface IEventSearchParams {
     query: string | undefined;
-    offset: string | undefined;
-    limit: string | undefined;
+    offset: number | undefined;
+    limit: number | undefined;
```
thomasheartman, line 93: Why do we have two interfaces with the exact same name now? Did we always? Is it gonna cause confusion?

src/lib/features/events/event-service.ts
```diff
@@ -8,10 +8,21 @@ import type EventEmitter from 'events';
...
+    project?: string;
+    query?: string;
+    createdAtFrom?: string;
+    createdAtTo?: string;
+    createdBy?: string;
+    type?: string;
+    offset: number;
+    limit: number;
+    // sortBy?: string;
+}
```
thomasheartman, line 24: Does the openapi schema not cut it anymore?

thomasheartman (approved):
Nice! Just a little skeptical of using the same name for two interfaces that where one is a subset of the other, but I'll let you be the judge of that 💁🏼

src/lib/features/events/event-store.ts
sjaanus: Good catch, removed one.

PR #7744, opened 2024-08-02
fix: playground parent disabled with strategy

kwasniew:
## About the changes

This bug manifests when you evaluate parent/child relationship where parent is disabled but has some strategies.
I added extra check in the parent deps checking code. It's different from the node SDK but should pass the tests and keep it backwards compatible with the execution plans requirement of the playground.

This solution diverges from the node SDK because of the playground behavior described as a test "considers disabled features with a default strategy to be enabled". I had to add an extra check in parent deps checking function to overcome for the differences in the behavior.

### DIFF
--- a/src/lib/features/playground/advanced-playground.test.ts
+++ b/src/lib/features/playground/advanced-playground.test.ts
@@ -59,6 +59,13 @@ const enableToggle = (featureName: string) =>
         )
         .send({})
         .expect(200);
+const disableToggle = (featureName: string) =>
+    app.request
+        .post(
+            `/api/admin/projects/default/features/${featureName}/environments/default/off`,
+        )
+        .send({})
+        .expect(200);
 
 afterAll(async () => {
     await app.destroy();
@@ -99,6 +106,8 @@ test('advanced playground evaluation with unsatisfied parent dependency', async 
     await createFeatureToggle('test-parent');
     await createFeatureToggle('test-child');
     await enableToggle('test-child');
+    await enableToggle('test-parent');
+    await disableToggle('test-parent');
     await app.addDependency('test-child', 'test-parent');
 
     const { body: result } = await app.request
@@ -116,6 +125,7 @@ test('advanced playground evaluation with unsatisfied parent dependency', async 
     // child is disabled because of the parent
     expect(child.hasUnsatisfiedDependency).toBe(true);
     expect(child.isEnabled).toBe(false);
+    expect(child.strategies.data.length).toBe(1);
     expect(child.isEnabledInCurrentEnvironment).toBe(true);
     expect(child.variant).toEqual({
         name: 'disabled',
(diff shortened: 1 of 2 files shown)


src/lib/features/playground/advanced-playground.test.ts
```diff
@@ -99,6 +106,8 @@ test('advanced playground evaluation with unsatisfied parent dependency', async
     await createFeatureToggle('test-parent');
     await createFeatureToggle('test-child');
     await enableToggle('test-child');
+    await enableToggle('test-parent');
```
kwasniew, line 109: enabling the toggle adds a default strategy

src/lib/features/playground/advanced-playground.test.ts
```diff
@@ -124,6 +134,7 @@ test('advanced playground evaluation with unsatisfied parent dependency', async
     });
     expect(parent.hasUnsatisfiedDependency).toBe(false);
     expect(parent.isEnabled).toBe(false);
+    expect(parent.strategies.data.length).toBe(1);
```
kwasniew, line 137: we want to return strategies even though the parent is disabled

src/lib/features/playground/feature-evaluator/client.ts
```diff
@@ -225,7 +225,7 @@ export default class UnleashClient {
 
         const [result, variants, variant] = overallStrategyResult();
         const evalResults: FeatureStrategiesEvaluationResult = {
-            result,
+            result: feature.enabled ? result : false,
```
kwasniew, line 228: ignore strategies evaluation result when feature is not enabled

src/lib/features/playground/feature-evaluator/client.ts
```diff
@@ -75,6 +75,9 @@ export default class UnleashClient {
             if (parentToggle.dependencies?.length) {
                 return false;
             }
+            if (Boolean(parent.enabled) !== Boolean(parentToggle.enabled)) {
```
kwasniew, line 78: parent is spec for the parent dependency, parentToggle is the actual toggle from the repository

thomasheartman: I'm not entirely sure what you mean by "parent is spec". does this mean that the flag is enabled in the environment, but none of the strategies pass (or vice versa)?

thomasheartman (approved):
I trust you on this one. the change looks small an innocent enough, that I think we should get it in if we can.

But you mention there's a test that checks that:
> "considers disabled features with a default strategy to be enabled"

I can't remember why we want that to be the case. Do you know?  (It's probably on me, but I can't quite remember why that is)

src/lib/features/playground/feature-evaluator/client.ts
kwasniew: parent is a dependent flag prerequisite state definition we get from the child flag definition. parentToggle is the actual parent flag definition with all the strategies.

PR #7752, opened 2024-08-05
feat: playground try configuration mode

kwasniew:
## About the changes

* removing old event `playground_token_input_used` - it was never registered in plausible
* re-using more generic event `playground` with type of `try-configuration` and `mode`
* we have 3 modes: default, api token and change request


frontend/src/component/playground/Playground/AdvancedPlayground.tsx
```diff
@@ -246,14 +246,28 @@ export const AdvancedPlayground: FC<{
...
     };
 
+    const trackTryConfiguration = () => {
+        let mode: 'default' | 'api_token' | 'change_request' = 'default';
+        if (token && token !== '') {
+            mode = 'api_token';
+        } else if (changeRequest) {
+            mode = 'change_request';
+        }
+        console.log('mode', mode);
```
Tymek, line 256:
```suggestion

```

Tymek (approved):

Issue #7751, opened 2024-08-05
Validation for Feature Toggle Type when creating a feature toggle is missing


meldsza:
### Describe the bug

When creating a feature toggle via the API, there is no validation whether the feature toggle type exists or not.

### Steps to reproduce the bug

```sh
curl --location --request POST 'http://unleash:4242/api/admin/projects/default/features' \
    --header 'Authorization: APIKEY' \
    --header 'Content-Type: application/json' \
    --data-raw '{
  "type": "asdadsadasasd",
  "name": "TestFeatureValidation",
  "description": "TestFeatureValidation",
  "impressionData": false
}'
```
Gives
```json
{"name":"TestFeatureValidation","description":"TestFeatureValidation","type":"asdadsadasasd","project":"default","stale":false,"createdAt":"2024-08-05T08:37:02.863Z","lastSeenAt":null,"impressionData":false,"archivedAt":null,"archived":false}
```

Note that the type `asdadsadasasd` does not exist

### Expected behavior

I would expect an error with status code 422 stating the that the feature toggle does not exist

### Unleash version

Unleash 6.0.6

### Subscription type

Open source

### Hosting type

Self-hosted

gastonfournier:
Hi @meldsza we'll look into it. It's been like that since forever, but we think it's an opportunity to make our API better as we expect the type to be one of the known types.

PR #7769, opened 2024-08-06
fix: feature type is now validated

sjaanus:
Previously people were able to send random data to feature type. Now it is validated.

Fixes https://github.com/Unleash/unleash/issues/7751

### DIFF
--- a/src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
+++ b/src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
@@ -2030,7 +2030,7 @@ test('should clone feature flag without strategies', async () => {
     const envName = 'some-env-3';
     const featureName = 'feature.flag.base';
     const cloneName = 'feature.flag.clone';
-    const type = 'eExperiment';
+    const type = 'experiment';
     const description = 'Lorem ipsum...';
 
     // Create environment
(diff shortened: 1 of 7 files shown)


src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
```diff
@@ -3707,3 +3707,13 @@ test('can get evaluation metrics', async () => {
         ],
     });
 });
+
+test('Should not be able to create flag with random type', async () => {
```
thomasheartman, line 3711: Can we be clearer what we mean here? Unleash doesn't have a flag type called random, but it _might_.

```suggestion
test("Should not be able to create flag with a type that doesn't exist", async () => {
```

src/lib/openapi/spec/create-feature-schema.ts
```diff
@@ -13,7 +13,13 @@ export const createFeatureSchema = {
             description: 'Unique feature name',
         },
         type: {
-            type: 'string',
+            enum: [
```
thomasheartman, line 16: Nice! Do we have this defined somewhere else that we could use instead of redefining it? Not a big deal, of course, but might save us from drift in the future 🤷🏼

thomasheartman (approved):
A couple minor suggestions, but this looks good to me 😄

PR #7770, opened 2024-08-06
feat: link to frontend api url

kwasniew:
## About the changes

Showing both CLIENT and FRONTEND API URLs. Also added copy link icon
Screenshot 2024-08-06 at 11 32 50

Details:
* using grid with the first two columns equal to the longest text (auto column + min-content) and the last column taking remaining space (1fr)
* using copy to clipboard that was previously used in signals


sjaanus (approved):
LG and clear

ivarconr (approved):

PR #7775, opened 2024-08-06
feat: make `to` date inclusive

thomasheartman:
Changes the handling of the `to` query parameter in
the API to be inclusive.


sjaanus (approved):
LG, but probably needs changes after createdAt rename is merged.

PR #7791, opened 2024-08-07
fix: don't chain more commands after clear

thomasheartman:
The [docs for `cy.clear()`](https://docs.cypress.io/api/commands/clear) say that:
> It is unsafe to chain further commands that rely on the subject after .clear().

So let's not do that.


thomasheartman:
Nope. Didn't work.

PR #7793, opened 2024-08-07
feat: add paging to event log

thomasheartman:
Adds sticky pagination to the event log:

image

This PR uses the sticky pagination bar that we use on other tables to navigate the event search results.

## Decisions / discussion points

The trickiest issue here is how we calculate the next and previous page offsets. This is tricky because we don't expose the page number to the API, but the raw offset itself. This abstraction makes it possible to set an offset that isn't a multiple of the page size.

Say the page size is 25. If you manually set an offset of 30 (through changing the URL), what do you expect should happen when you:
- load the page? Should you see results 31 to 55? 26 to 50?
- go to the next page? Should your next offset be 55 or 50?
- previous page: should your previous page offset be 5? 25? 0?

The current implementation has taken what I thought would be the easiest way out: If your offset is between two multiples of the page size, we'll consider it to be the lower of the two.
- The next page's offset is the next multiple of the page size that is higher than the current offset (50 in the example above).
- The previous page's offset will be not the nearest lower page size, but the one below. So if you set offset 35 and page size 25, your next page will take you back to 0 (as if the offset was 25).

We could instead update the API to accept `page` instead of offset, but that wouldn't align with how other tables do it.

Comparing to the global flags table, if you set an offset that isn't a multiple of the page size, we force the offset to 0. We can look at handling it like that in a follow-up, though I'd argue that forcing it to be the next lower multiple of the page size would make more sense.

One issue that appears when you can set custom offsets is that the little "showing x-y items out of z" gets out of whack (because it only operates on multiples of the page size (seemingly))
image

## The Event Log as a table

While we haven't used the HTML `table` element to render the event log, I would argue that it _is_ actually a table. It displays tabular data. Each card (row) has an id, a project, etc. 

The current implementation forces the event log search to act as a table state manager, but we could transform the event list into an events table to better align the pagination handling. The best part? We can keep the exact same design too. A table doesn't have to _look_ like a table to be a table.

### DIFF
--- a/frontend/src/component/events/EventLog/EventLog.tsx
+++ b/frontend/src/component/events/EventLog/EventLog.tsx
@@ -55,7 +55,6 @@ const NewEventLog = ({ title, project, feature }: IEventLogProps) => {
         tableState,
         setTableState,
         filterState,
-        limit,
         pagination,
     } = useEventLogSearch(
         project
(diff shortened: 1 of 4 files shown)


sjaanus (approved):
LG, to be honest, I would not bother more with custom offset because it is a very niche use case.

PR #7795, opened 2024-08-07
chore: initial admin email

nunogois:
https://linear.app/unleash/issue/2-2518/figure-out-how-to-create-the-initial-admin-user-in-unleash

~Adds a new `initialAdminEmailUser` auth option, that can also be controlled through these environment variables:~
 - ~`INITIAL_ADMIN_EMAIL_USER_NAME`~
 - ~`INITIAL_ADMIN_EMAIL_USER_EMAIL`~

~If these are set, Unleash will create the initial admin account using the provided name and email address. Instead of setting an initial password during account creation, an email will be sent to the specified address with a link for the new admin user to securely set their password.~

The logic around `initAdminUser` that was introduced in https://github.com/Unleash/unleash/pull/4927 confused me a bit. I wrote new tests with what I assume are our expectations for this feature and refactored the code accordingly, but would like someone to confirm that it makes sense to them as well.

~I also moved the "create user but also handle email logic" from the controller layer (`user-admin`) to the service layer (`user-service`). Not only should this make more sense in general, but it's also more reusable since we can just call this new method from the service.~

The logic was split into 2 different methods: one to get the initial invite link, and another to send a welcome email. Now these two methods are more granular than the previous alternative and can be used independently of creating a new user.


gastonfournier (commented):
I still couldn't review it properly, just a small comment before jumping on a meeting

src/lib/__snapshots__/create-config.test.ts.snap
```diff
@@ -20,10 +20,8 @@ exports[`should create default config 1`] = `
     "demoAllowAdminLogin": false,
     "enableApiToken": true,
     "initApiTokens": [],
-    "initialAdminUser": {
-      "password": "unleash4all",
-      "username": "admin",
-    },
+    "initialAdminEmailUser": undefined,
```
sighphyre, line 23: Isn't this a breaking change? This is part of the public config isn't it?

nunogois: Yes, however the type was already optional before, which was a lie anyways. I just removed the defaults since they were not being used anyways and actually handled this as an optional.

src/lib/services/user-service.ts
```diff
@@ -183,6 +187,25 @@ class UserService {
         }
     }
 
+    async initAdminEmailUser({ name, email }: EmailAdminUser): Promise<void> {
+        const userCount = await this.store.count();
+
+        if (userCount === 0) {
+            try {
+                this.logger.info(
+                    `Creating default admin user, with email '${email}' ${name ? ` and name '${name}'` : ''}`,
```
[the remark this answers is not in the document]
gastonfournier: Agreed, elsewhere we're printing the password, we should probably fix those but not in this PR

src/lib/create-config.ts
```diff
@@ -324,10 +338,8 @@ const defaultAuthentication: IAuthOption = {
     type: authTypeFromString(process.env.AUTH_TYPE),
     customAuthHandler: defaultCustomAuthDenyAll,
     createAdminUser: true,
-    initialAdminUser: {
-        username: process.env.UNLEASH_DEFAULT_ADMIN_USERNAME ?? 'admin',
-        password: process.env.UNLEASH_DEFAULT_ADMIN_PASSWORD ?? 'unleash4all',
-    },
+    initialAdminUser: parseEnvVarInitialAdminUser(),
+    initialAdminEmailUser: parseEnvVarInitialAdminEmailUser(),
```
gastonfournier, line 342: Just a question: should it be one or the other? Just an idea:
```suggestion
    initialAdminUser: parseEnvVarInitialAdminEmailUser() ?? parseEnvVarInitialAdminUser(),
```

Then we can handle this complexity in the method. So an initial admin user is either a name + email or a username + password

The one benefit is that we have less fields in the config file, the second benefit is that we define the preference of email based over password based at an earlier stage.

gastonfournier (approved):
LGTM! we can consider my comment as just cosmetic. I don't think it makes a huge difference but I'll let you decide on it ;)

ivarconr (approved):
Fantastic. Looks sane to me. Love that you keep existing functionality and enhance it with the new.

gastonfournier:
Made some changes:
- removed the ability of setting up a new email because it was not needed (can be done later)
- split the createUserWithEmail into more granular functionality

gastonfournier:
Tested locally to make sure we still get the default user

PR #7809, opened 2024-08-08
feat: event creators

sjaanus:
Adds an endpoint to return all event creators.

An interesting point is that it does not return the user object, but just created_by as a string. This is because we do not store user IDs for events, as they are not strictly bound to a user object, but rather a historical user with the name X.

### DIFF
--- a/src/lib/features/events/event-store.ts
+++ b/src/lib/features/events/event-store.ts
@@ -16,7 +16,7 @@ import { sharedEventEmitter } from '../../util/anyEventEmitter';
 import type { Db } from '../../db/db';
 import type { Knex } from 'knex';
 import type EventEmitter from 'events';
-import { ADMIN_TOKEN_USER, SYSTEM_USER_ID } from '../../types';
+import { ADMIN_TOKEN_USER, SYSTEM_USER, SYSTEM_USER_ID } from '../../types';
 import type { DeprecatedSearchEventsSchema } from '../../openapi';
 import type { IQueryParam } from '../feature-toggle/types/feature-toggle-strategies-store-type';
 import { applyGenericQueryParams } from '../feature-search/search-utils';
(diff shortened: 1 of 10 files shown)


src/lib/openapi/spec/event-creators-schema.ts
thomasheartman, line 1: So when you hit this API, the response will just be:

```
[
 "John",
 "Sally",
 "Mary"
]
```

Do we want to wrap the usernames in objects or something in case we want to expand in the future? I'm not sure when we ever would or how we would do that, though 🤷🏼 probably not?

src/lib/types/stores/event-store.ts
```diff
@@ -43,4 +43,5 @@ export interface IEventStore
     query(operations: IQueryOperations[]): Promise<IEvent[]>;
     queryCount(operations: IQueryOperations[]): Promise<number>;
     setCreatedByUserId(batchSize: number): Promise<number | undefined>;
+    getEventCreators(): Promise<Array<{ id: number; name: string }>>;
```
thomasheartman, line 46: Hang on, if we get the ID here, why do we only have the name for the API?

thomasheartman (commented):
No issues with the implementation in the back, but why do we only use the `createdBy` field? Isn't that what `createdById` is for, so we can bind it to the user object? Wouldn't it be better to get distinct `createdById`s and map them to users?

... but then, what do you do with `-42`, for instance, which might be used by multiple integrations 🤔 

I'm just thinking that if your username is User1, but then you created some events before changing your username to User2 ... how do you ever correlate them? Can you?

In short: tell me why you think using just the name (and no ids) is the way to go here.

sjaanus:
> No issues with the implementation in the back, but why do we only use the `createdBy` field? Isn't that what `createdById` is for, so we can bind it to the user object? Wouldn't it be better to get distinct `createdById`s and map them to users?
> 
> ... but then, what do you do with `-42`, for instance, which might be used by multiple integrations 🤔
> 
> I'm just thinking that if your username is User1, but then you created some events before changing your username to User2 ... how do you ever correlate them? Can you?
> 
> In short: tell me why you think using just the name (and no ids) is the way to go here.

Ah this was oversight from me, we do have ID in the events. I will change it back and include it in payload then.

thomasheartman:
Yeah, I think we do now. But are there also cases where we maybe can't use id? What's the plan: are we just sending the ID? Or the username and ID? What happens in cases where multiple events have the same id but different usernames? I think integrations use -42? 🤔

src/lib/features/events/event-store.ts
```diff
@@ -380,6 +380,32 @@ class EventStore implements IEventStore {
         return query;
     }
 
+    async getEventCreators(): Promise<Array<{ id: number; name: string }>> {
```
sjaanus, line 383: 1. -1337 and -42 get their name automatically from our predefined constants
2. Then looks at users table
3. Finally looks at events table

src/test/e2e/api/admin/event.e2e.test.ts
```diff
@@ -150,3 +151,61 @@ test('can search for events', async () => {
             expect(res.body.events[0].data.id).toEqual(events[1].data.id);
         });
 });
+
+test('event creators - if system user, return system name, else should return name from database if user exists, else from events table', async () => {
```
thomasheartman, line 155: Nice! I'd personally like to see tests for:
- that we have a default for admin token user's name (-42). 
- what happens if multiple events have the same created by user id but different created by names (do we take the first, last, or a random one?)

But I'm happy for that to happen in a follow-up PR. I think the second one is especially interesting, but it's also definitely an edge case. I'll leave that up to you.

thomasheartman (approved):
Cool 😄 Added a small request for some follow-up work, but nothing blocking 🙌🏼

PR #7826, opened 2024-08-09
feat: prevent revive flag/flags in archived project

kwasniew:
## About the changes

Prevent people from reviving features in archived project

Screenshot 2024-08-09 at 16 14 13

Details:
* added a validation method that checks if there's an active project
* covering single feature and batch revive
* currently we don't distinguish between deleted and archived and we always say that active project was not found. We may consider splitting this into not found and invalid operation in future

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-service.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-service.ts
@@ -1235,6 +1235,18 @@ class FeatureToggleService {
         }
     }
 
+    private async validateActiveProject(projectId: string) {
+        if (this.flagResolver.isEnabled('archiveProjects')) {
+            const hasActiveProject =
+                await this.projectStore.hasActiveProject(projectId);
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1235,6 +1235,18 @@ class FeatureToggleService {
         }
     }
 
+    private async validateActiveProject(projectId: string) {
```
Tymek, line 1238: This form sounds like "check something in active project"
```suggestion
    private async validateProjectIsActive(projectId: string) {
```

Tymek (approved):

src/lib/features/feature-toggle/feature-toggle-service.ts
kwasniew: will address in another PR cause we have 2 similar method in project and feature service

PR #7835, opened 2024-08-12
Fix: time to production

Tymek:
## About the changes
- filter out time to production data points of `0 days to production`
- allow for gathering data for quickly enabled feature flags


kwasniew (approved):

src/lib/features/feature-toggle/time-to-production/time-to-production.test.ts
```diff
@@ -23,4 +23,15 @@ describe('calculate average time to production', () => {
 
         expect(timeToProduction).toBe(21);
     });
+
+    test('should return more then 0 if feature was enabled almost instantly', () => {
```
Tymek, line 27:
```suggestion
    test('should return more than 0 if feature was enabled almost instantly', () => {
```

PR #7837, opened 2024-08-12
fix: display feature naming patterns in dialog

thomasheartman:
Updates the dialog form template to include a `namingPattern` prop that can be used to display the naming pattern of whatever the form is used for.

Also updates the create feature dialog to display the naming pattern in the current project.

The naming pattern component re-uses the pattern that we have in place for feature naming patterns, but puts it in an expandable dialog instead. 

Screenies:

Naming pattern closed:
image

image

Naming pattern open:
image

### DIFF
--- a/frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
+++ b/frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
@@ -33,14 +33,14 @@ export const StyledHeader = styled(Typography)({
     fontWeight: 'lighter',
 });
 
-export const ProjectNameContainer = styled('div')(({ theme }) => ({
+export const NameContainer = styled('div')(({ theme }) => ({
     gridArea: 'project-name',
     display: 'flex',
     flexFlow: 'column nowrap',
     gap: theme.spacing(4),
 }));
 
-export const ProjectDescriptionContainer = styled('div')({
+export const DescriptionContainer = styled('div')({
     gridArea: 'project-description',
 });
 
--- a/frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
+++ b/frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
@@ -33,9 +33,12 @@ export const StyledHeader = styled(Typography)({
     fontWeight: 'lighter',
 });
 
-export const ProjectNameContainer = styled('div')({
+export const ProjectNameContainer = styled('div')(({ theme }) => ({
     gridArea: 'project-name',
-});
+    display: 'flex',
+    flexFlow: 'column nowrap',
+    gap: theme.spacing(4),
+}));
 
 export const ProjectDescriptionContainer = styled('div')({
     gridArea: 'project-description',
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
```diff
@@ -33,11 +33,14 @@ export const StyledHeader = styled(Typography)({
...
+export const NameContainer = styled('div')(({ theme }) => ({
     gridArea: 'project-name',
-});
+    display: 'flex',
+    flexFlow: 'column nowrap',
+    gap: theme.spacing(4),
+}));
 
-export const ProjectDescriptionContainer = styled('div')({
+export const DescriptionContainer = styled('div')({
```
thomasheartman, line 43: The old name was a leftover from when this was only used for projects.

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.styles.tsx
```diff
@@ -33,11 +33,14 @@ export const StyledHeader = styled(Typography)({
...
 });
 
-export const ProjectNameContainer = styled('div')({
+export const NameContainer = styled('div')(({ theme }) => ({
     gridArea: 'project-name',
-});
+    display: 'flex',
+    flexFlow: 'column nowrap',
+    gap: theme.spacing(4),
+}));
```
thomasheartman, line 41: The gap only shows up when the naming pattern is there. Without it, it still looks the same.

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx
```diff
@@ -15,6 +15,11 @@ import {
 import { Button } from '@mui/material';
 import { CreateButton } from 'component/common/CreateButton/CreateButton';
 import type { IPermissionButtonProps } from 'component/common/PermissionButton/PermissionButton';
+import type { FeatureNamingType } from 'interfaces/project';
+import { ConditionallyRender } from '../ConditionallyRender/ConditionallyRender';
+import { NamingPatternInfo } from './NamingPatternInfo';
+
+type NamingPattern = FeatureNamingType;
```
thomasheartman, line 22: Just aliasing the type here. We don't have naming patterns for anything but features, but there is no reason why we couldn't, and there's nothing in the type that relates to features specifically, so this is just to make it more "agnostic". Happy to remove the aliasing if we don't like it.

frontend/src/component/common/DialogFormTemplate/NamingPatternInfo.tsx
thomasheartman, line 1: This is mostly copied from `frontend/src/component/feature/FeatureNamingPatternInfo/FeatureNamingPatternInfo.tsx`, but moved into an accordion.

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx
sjaanus: Are you trying to do 

`
import { NamingPatternInfo as NamingPattern } from './NamingPatternInfo';`

sjaanus (approved):

frontend/src/component/common/DialogFormTemplate/DialogFormTemplate.tsx
thomasheartman: No, no. NamingPatternInfo is a component. NamingPattern is a type.

PR #7839, opened 2024-08-12
feat: prevent move feature to archived project

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/features/project/project-service.e2e.test.ts
+++ b/src/lib/features/project/project-service.e2e.test.ts
@@ -962,7 +962,7 @@ test('should not change project if feature flag project does not match current p
     }
 });
 
-test('should return 404 if no project is found with the project id', async () => {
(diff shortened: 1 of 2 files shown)


src/lib/features/project/project-service.e2e.test.ts
```diff
@@ -962,7 +962,7 @@ test('should not change project if feature flag project does not match current p
     }
 });
 
-test('should return 404 if no project is found with the project id', async () => {
+test('should return 404 if no active project is found with the project id', async () => {
```
Tymek, line 965: looks like 2 tests

Tymek (approved):

src/lib/features/project/project-service.e2e.test.ts
kwasniew: I wanted to avoid setting up the same thing twice

PR #7841, opened 2024-08-12
feat: export events as json

sjaanus:
image

### DIFF
--- a/frontend/src/component/events/EventLog/EventActions.tsx
+++ b/frontend/src/component/events/EventLog/EventActions.tsx
@@ -68,8 +68,9 @@ export const EventActions: FC<IEventActions> = ({ events }) => {
             <Tooltip title={'Export'} arrow describeChild>
                 <div>
                     <IconButton
+                        aria-label={'Export'}
                         aria-haspopup='true'
-                        aria-expanded={open ? 'true' : undefined}
+                        aria-expanded={open}
                         onClick={handleClick}
                         type='button'
                     >
@@ -86,22 +87,14 @@ export const EventActions: FC<IEventActions> = ({ events }) => {
                 disableScrollLock={true}
             >
                 <MenuList>
-                    <MenuItem
-                        onClick={() => {
-                            exportCsv();
-                        }}
-                    >
+                    <MenuItem onClick={exportCsv}>
                         <ListItemText>
                             <Typography variant='body2'>
                                 Export as CSV
                             </Typography>
                         </ListItemText>
                     </MenuItem>
-                    <MenuItem
-                        onClick={() => {
-                            exportJson();
-                        }}
-                    >
+                    <MenuItem onClick={exportJson}>
                         <ListItemText>
                             <Typography variant='body2'>
                                 Export as JSON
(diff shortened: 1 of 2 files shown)


frontend/src/component/events/EventLog/EventActions.tsx
```diff
@@ -0,0 +1,115 @@
...
+            <Tooltip title={'Export'} arrow describeChild>
+                <div>
+                    <IconButton
+                        aria-haspopup='true'
+                        aria-expanded={open ? 'true' : undefined}
+                        onClick={handleClick}
+                        type='button'
+                    >
+                        <FileDownload />
+                    </IconButton>
```
thomasheartman, line 77: I think this button needs some sort of text or label so that it's obvious what it does to non-sighted users.

If we don't want any visible text on the button, you can use the `ScreenReaderOnly` component, which'll make the contents visible to screen readers, but not rendered visibly on the screen.

I'm also not familiar with all the aria-stuff you need to make this accessible, but it seems like you've done a bit of research. Are the aria attributes based on something?

Also, according to [mdn on the aria-expanded attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#values), `undefined` indicates that "the element does not own or control a grouping element that is expandable.", while `false` indicates that "the grouping element this element owns or controls is collapsed". So in this case, I think it should just be `aria-expanded={open}`

frontend/src/component/events/EventLog/EventActions.tsx
```diff
@@ -0,0 +1,115 @@
...
+                        <ListItemText>
+                            <Typography variant='body2'>
+                                Export as CSV
+                            </Typography>
+                        </ListItemText>
+                    </MenuItem>
+                    <MenuItem
+                        onClick={() => {
+                            exportJson();
+                        }}
```
thomasheartman, line 103: If you wanna be fancy and skip a function nesting (also applies to the previous one for CSV)

```suggestion
                        onClick={exportJson}}
```

frontend/src/component/events/EventLog/EventActions.tsx
```diff
@@ -0,0 +1,115 @@
...
+        const fileName = `events_${currentDate}.json`;
+
+        const a = document.createElement('a');
+        a.href = url;
+        a.download = fileName;
+        a.click();
+
+        URL.revokeObjectURL(url);
+        setAnchorEl(null);
+    };
```
thomasheartman, line 55: Is it not possible to just point to the API with the current search string and use that to generate it? Not very important, but it might save us a few lines of code 🤷🏼  Might not, though

thomasheartman (approved):

frontend/src/component/events/EventLog/EventActions.tsx
sjaanus: I would rather spend user machine resources that our servers. Also the data is already there.

frontend/src/component/events/EventLog/EventActions.tsx
sjaanus: I think the wrapper has tooltip text that will be read with screenreader.

sjaanus: Or not..

sjaanus: Added aria label.

PR #7860, opened 2024-08-13
feat: exporting events as csv

sjaanus:
Adding csv library that also makes nested properties into columns.

### DIFF
--- a/frontend/src/component/events/EventLog/EventActions.tsx
+++ b/frontend/src/component/events/EventLog/EventActions.tsx
@@ -26,6 +26,27 @@ interface IEventActions {
     events: EventSchema[];
 }
 
+const convertObjectToCsv = (events: any[]) => {
+    const headers = Object.keys(events[0]);
+
+    const csvRows = [
+        headers.join(','),
+        ...events.map((row) =>
+            headers
+                .map((field) => {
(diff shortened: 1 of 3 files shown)


frontend/src/component/events/EventLog/EventActions.tsx
```diff
@@ -26,6 +26,27 @@ interface IEventActions {
     events: EventSchema[];
 }
 
+const convertObjectToCsv = (events: any[]) => {
+    const headers = Object.keys(events[0]);
```
sjaanus, line 30: All the event properties will be headers

frontend/src/component/events/EventLog/EventActions.tsx
```diff
@@ -26,6 +26,27 @@ interface IEventActions {
...
+
+    const csvRows = [
+        headers.join(','),
+        ...events.map((row) =>
+            headers
+                .map((field) => {
+                    const value = JSON.stringify(row[field]) || '';
+                    const escapedValue = value.includes(',')
```
sjaanus, line 38: Add quotes around if comma, so it will be counted into single column.

thomasheartman (approved):

PR #7864, opened 2024-08-13
Update collaboration mode docs

melindafekete:
## About the changes
Testing some new visuals for required plan / version. Update collaboration mode docs.

**[Collaboration mode preview](https://unleash-docs-git-melinda-docs-collaboration-mode-unleash-team.vercel.app/reference/project-collaboration-mode#open-collaboration-mode)**


ivarconr (approved):

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,65 @@
...
+The collaboration mode of a [project](../reference/projects.md) defines who within your Unleash instance can access the project. This setting also determines who can submit [change requests](../reference/change-requests.md)–a feature that provides controlled and secure updates for feature toggles.
 
-The open collaboration mode is the default in Unleash and is how all projects worked in Unleash before the introduction of collaboration modes.
+Unleash supports three collaboration modes: **open**, **protected**, and **private**. Each mode is designed with different user roles and permissions in mind, catering to various security and collaboration requirements. Let’s explore how the three modes work.
 
-## **Protected** collaboration mode
+## Open collaboration mode
 
-Only admins and project members can submit change requests in a **protected** project. Other users can not submit change requests.
+All users of your Unleash instance can access the project and submit change requests. This is the default collaboration mode.
```
melindafekete, line 21: Still need to update this based on final clarification

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,65 @@
...
-**Editors**, **admins**, and users with **custom root roles** can still see private projects.
+All users of your Unleash instance can access the project but only admins and project members can submit change requests.
 
-### Change requests
 
-When you change a project's collaboration mode from open to protected, users who do not have the right permissions will lose the ability to create new change requests.
+## Private collaboration mode
 
-However, existing change requests created by these users will not be deleted. Any users with open change requests will still be able to cancel the change request, but they will **not** be able to update the change request.
+All project members can access the project and submit change requests. Outside of project members, only admins, editors, and users with [custom root roles](../how-to/how-to-create-and-assign-custom-root-roles.md) can access the project and submit change requests.
```
sjaanus, line 31: Just a minor comment. It will read better if there is no negative(outside of) in text. Something like
#
All project members, as well as admins, editors, and users with [custom root roles](https://github.com/Unleash/unleash/pull/how-to/how-to-create-and-assign-custom-root-roles.md), can access the project and submit change requests. Non-members do not have access.

melindafekete: Thanks! What custom root role would allow you to access a private project?

sjaanus: Any custom root role.

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,65 @@
...
 
-## Project collaboration mode setting
 
-The collaboration mode for a project can be set at the time of creation and modified at any subsequent time, found within the 'Enterprise Settings' section.
+| Collaboration Mode | View project    | Submit change requests                                                                                                        |
+|--------------------|-----------------|-------------------------------------------------------------------------------------------------------------------------------|
+| Open               | All users       | All users                                                                                                                     |
+| Protected          | All users       | Project members, Admins, Editors, and users with [custom root roles](../how-to/how-to-create-and-assign-custom-root-roles.md) |
```
kwasniew, line 37: I don't think that Editors can submit change requests in the protected mode.

website/docs/reference/project-collaboration-mode.md
melindafekete: Sorry, I meant what root permission would that role need to have? I can't see any permissions related to viewing private projects or submitting change requests.

kwasniew: There's no permission to submitting change request: you need to be either be an admin or a project member.

website/docs/reference/project-collaboration-mode.md
melindafekete: Verified that Thomas as an Editor cannot submit change requests

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,64 @@
...
-- [**open**](#open-collaboration-mode)
-- [**protected**](#protected-collaboration-mode)
-- [**private**](#private-collaboration-mode)
+# Overview
 
-## **Open** collaboration mode
+Project collaboration modes allow you to manage project visibility and interaction levels, ensuring better control and collaboration within your organization.
 
-Anyone can submit change requests in an **open** project, regardless of their permissions within the project and globally. This is the default collaboration mode for projects.
+The collaboration mode of a [project](projects.md) defines who within your Unleash instance can access the project. This setting also determines who can submit [change requests](change-requests.md)–a feature that provides controlled and secure updates for feature toggles.
```
thomasheartman, line 15: Couple small things:
1. We're working the docs (and everything else) towards "flags" instead of "toggles" (you've probably heard this by now, but just in case you haven't), so let's stick with that going forward.

2. Should we use spaces around hyphens/em-dashes that are used for interjections or not? I think I've always preferred using spaces, especially because people don't tend to differentiate between hyphens and em-dashes, so leaving out the spaces can technically cause some ambiguity (though it's probably not an issue). Honestly no strong feelings on this, just wanna align; your call.

3. I'm not sure about the description of change requests there. Is it necessary? And more importantly: is that the right way to describe them? It kinda sounds like marketing lingo, to me, which feels weird in the docs. But let's talk about it 💁🏼 
 
```suggestion
The collaboration mode of a [project](projects.md) defines who within your Unleash instance can access the project. This setting also determines who can submit [change requests](change-requests.md) – a feature that provides controlled and secure updates for feature flags.
```

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,64 @@
...
 
-The open collaboration mode is the default in Unleash and is how all projects worked in Unleash before the introduction of collaboration modes.
+Unleash supports three collaboration modes: **open**, **protected**, and **private**. The [predefined root and project roles](rbac.md#predefined-roles), as well as [custom root roles](rbac.md#custom-root-roles) of a user, determine what projects they can access.
```
thomasheartman, line 17: Maybe consider restructuring the sentence here, so that "user" shows up earlier? I think it'd make it easier to understand what roles we're talking about (at least I was a little confused when reading this at first).

Maybe something like this? 

```suggestion
Unleash supports three collaboration modes: **open**, **protected**, and **private**. A user's [predefined root and project roles](rbac.md#predefined-roles), as well as their [custom root roles](rbac.md#custom-root-roles), determine what projects they can access.
```

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,64 @@
...
-**Editors**, **admins**, and users with **custom root roles** can still see private projects.
+All users of your Unleash instance can access the project but only Admins and project Members can submit change requests.
 
-### Change requests
 
-When you change a project's collaboration mode from open to protected, users who do not have the right permissions will lose the ability to create new change requests.
+## Private collaboration mode
 
-However, existing change requests created by these users will not be deleted. Any users with open change requests will still be able to cancel the change request, but they will **not** be able to update the change request.
+Only project Members, Admins, Editors, and users with [custom root roles](rbac.md#custom-root-roles) can access the project. Only Admins and project Members can submit change requests.
```
thomasheartman, line 31: Do we need to differentiate between seeing and accessing a project here?

When it says only some users can "access" a project, that doesn't read to me like it's necessarily hidden from other users, which is a core part of the functionality here. Maybe we should clarify that the project is also hidden from viewers?

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,64 @@
...
-When you change a project's collaboration mode from open to protected, users who do not have the right permissions will lose the ability to create new change requests.
+## Private collaboration mode
 
-However, existing change requests created by these users will not be deleted. Any users with open change requests will still be able to cancel the change request, but they will **not** be able to update the change request.
+Only project Members, Admins, Editors, and users with [custom root roles](rbac.md#custom-root-roles) can access the project. Only Admins and project Members can submit change requests. 
 
-## Project collaboration mode setting
 
-The collaboration mode for a project can be set at the time of creation and modified at any subsequent time, found within the 'Enterprise Settings' section.
+|           | View project                                                                                    | Submit change requests     |
```
thomasheartman, line 34: Should we not have a heading for the mode?

```suggestion
| Mode      | View project                                                                                    | Submit change requests     |
```

Also, in preceding paragraphs, we talk about "accessing" the project. Here it's about "viewing" the project. What's the difference? Maybe we should align on one or the other?

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,64 @@
...
 
-![Project creation form with a collaboration mode field and corresponding explanation.](/img/collaboration-mode.png)
+# Set project collaboration mode
 
-Since the collaboration mode is an integral feature of the project, the options to modify it can be found on the 'Project Settings' page.
+To [create a new project](projects.md#create-a-project) with a specific collaboration mode, do the following:
+1. In the Unleash Admin UI, go to **Projects** > **New project**.
+2. Enter **Project name**.
+3. Click **Open** to choose your collaboration mode.
+4. Click **Create project**.
```
thomasheartman, line 46: Here's a question for the style guide: do we use words like "click"? Click is exclusively a mouse gesture. People who navigate by keyboards or touch screens don't "click" things. I believe I've seen style guides that discourages use of mouse-specific terminology because it's not inclusive of other ways to navigate the website. Do you have any thoughts around this?

website/docs/reference/project-collaboration-mode.md
```diff
@@ -2,52 +2,64 @@
...
+To modify the collaboration mode of an existing project, do the following:
+1. In the Unleash Admin UI, go to **Projects** and select the project you want to modify.
+2. Go to **Project settings > Enterprise settings** and use the **Project collaboration mode** list to update your collaboration mode.
+3. Click **Save changes**.
 
-Projects that were created in earlier versions of Unleash (before the release of project collaboration modes) get the **open** mode when they are migrated to a version of Unleash with project collaboration modes.
+### Change collaboration mode to protected
 
-To change the project collaboration mode for an existing project you have to edit the project.
+When you change the collaboration mode of an existing project to protected, all users who do not have sufficient permissions lose the ability to create new change requests. Existing change requests remain in place. Users with insufficient permissions can still cancel their change requests but can no longer update them.
```
thomasheartman, line 57: This is a great clarification 🙌🏼

thomasheartman (approved):

website/docs/reference/project-collaboration-mode.md
melindafekete: Thanks I agree, I'll go for view everywhere

website/docs/reference/project-collaboration-mode.md
melindafekete: I'm not sure I've ever come across a good term that was truly inclusive of all different kinds of devices/ navigation approaches. Do you know of one? For me, one that comes close is _select_, but I prefer to reserve it for dropdowns.

So I think it's okay to continue using _click_ for now. I think it remains a deeply ingrained term that is widely recognizable and understood by most users.

For general UI documentation stuff, I would stick to this [style guide](https://developers.google.com/style/ui-elements), unless we find that it doesn't work well for us.

website/docs/reference/project-collaboration-mode.md
thomasheartman: 1. Of course!
2. Understood! I'll try to keep that in mind for the next time I write something with dashes.
3. That's a good point. I think my issue is that "controlled and secure updates for feature flags" sounds like something that belongs on a marketing website to me. It doesn't really tell me anything about the feature. At least it feels like fluff to **me**. But you might not agree, and it's your decision in the end, so it's entirely up to you. I've voiced my opinion, I'm very happy to be overruled 💁🏼

website/docs/reference/project-collaboration-mode.md
thomasheartman: Nah, I haven't heard any good ones. "Interact" comes up, but "Interact with the create project button" is both long, and ... can be kinda confusing.

But thanks for the link and for picking a style guide. I'm glad to have a definitive answer for this so I don't have to wonder anymore 😄

PR #7866, opened 2024-08-14
feat: archived projects query improved

kwasniew:
## About the changes

When querying projects we get archived at information from the service. Also updated openapi to include optional archivedAt

### DIFF
--- a/src/lib/features/project/project-store.ts
+++ b/src/lib/features/project/project-store.ts
@@ -140,8 +140,8 @@ class ProjectStore implements IProjectStore {
             }
         }
 
-        if (query) {
-            projects = projects.where(query);
+        if (query?.id) {
+            projects = projects.where(`${TABLE}.id`, query.id);
         }
 
         let selectColumns = [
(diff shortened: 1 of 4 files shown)


src/lib/openapi/spec/project-schema.ts
```diff
@@ -64,6 +64,12 @@ export const projectSchema = {
             description: 'When this project was last updated.',
             example: '2023-07-28T12:12:28Z',
         },
+        archivedAt: {
```
kwasniew, line 67: optional API field for project

src/lib/features/project/project-store.ts
```diff
@@ -140,8 +140,8 @@ class ProjectStore implements IProjectStore {
             }
         }
 
-        if (query) {
-            projects = projects.where(query);
+        if (query?.id) {
```
kwasniew, line 143: the old query was only for id and I prefer to have explicit params since previously it was also catching archived boolean which is not a valid column name

Tymek (approved):

PR #7874, opened 2024-08-14
feat: allow editing root role/description on SCIM group

sighphyre:
This relaxes validation on modifying SCIM groups.

From a UI perspective, users are now allowed to access the edit screen for SCIM groups (previously you couldn't click the relevant UI elements and if you navigated manually you'd just get the save button disabled). Now instead updating the name is disabled and modifying the SSO mappings is hidden. The usual places that previously blocked you from changing users are still disabled.

On the backend, validation has been relaxed to allow some editing of groups, so long as you're not changing the name or modifying the users (effectively the only two properties SCIM cares about).

### DIFF
--- a/frontend/src/component/admin/groups/CreateGroup/CreateGroup.tsx
+++ b/frontend/src/component/admin/groups/CreateGroup/CreateGroup.tsx
@@ -106,6 +106,7 @@ export const CreateGroup = () => {
                 handleSubmit={handleSubmit}
                 handleCancel={handleCancel}
                 mode={CREATE}
+                isScimGroup={false}
             >
                 <Button
                     type='submit'
--- a/src/lib/services/group-service.ts
+++ b/src/lib/services/group-service.ts
@@ -231,6 +231,22 @@ export class GroupService {
                 throw new NameExistsError('Group name already exists');
             }
         }
+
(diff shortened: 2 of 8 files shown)


src/lib/services/group-service.ts
```diff
@@ -30,6 +30,10 @@ import type { IUser } from '../types/user';
 import type EventService from '../features/events/event-service';
 import { SSO_SYNC_USER } from '../db/group-store';
 
+const setsAreEqual = (firstSet, secondSet) =>
```
sighphyre, line 33: This is ugly and bloats the call site where it's actually relevant but only really useful in one place so I've pulled it out into a non exported function

frontend/src/component/admin/groups/CreateGroup/CreateGroup.tsx
```diff
@@ -106,6 +106,7 @@ export const CreateGroup = () => {
                 handleSubmit={handleSubmit}
                 handleCancel={handleCancel}
                 mode={CREATE}
+                isScimGroup={false}
```
sighphyre, line 109: Can't be a SCIM group if SCIM didn't make it!

daveleek (approved):
Awesome, this looks good!

PR #7879, opened 2024-08-14
fix: cap number of collaborators displayed

thomasheartman:
Caps the number of collaborators to display to a constant (currently
set to 99).

Above that, it'll always show "+99".

However, we also add a tooltip that shows you a prettified version of
the extra collaborators (using the `millify` package that we already use for metrics)

Screenies:

< 1000 collaborators:
image

1000 - 1M collaborators:
image

1M+ collaborators:
image

Update, it now shows the total number of collaborators instead of the overflow:
image

### DIFF
--- a/frontend/src/component/common/AvatarGroup/AvatarGroup.tsx
+++ b/frontend/src/component/common/AvatarGroup/AvatarGroup.tsx
@@ -88,7 +88,7 @@ const AvatarGroupInner = ({
                 show={
                     <AvatarComponent
                         user={{
-                            username: `+${millify(overflow)}`,
+                            username: `Total: ${millify(users.length)}`,
                         }}
                     >
                         +{Math.min(overflow, MAX_OVERFLOW_DISPLAY_NUMBER)}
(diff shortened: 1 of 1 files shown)


frontend/src/component/common/AvatarGroup/AvatarGroup.tsx
```diff
@@ -73,16 +76,22 @@ const AvatarGroupInner = ({
...
+                condition={overflow > 0}
                 show={
-                    <AvatarComponent>
-                        +{users.length - shownUsers.length}
+                    <AvatarComponent
+                        user={{
+                            username: `+${millify(overflow)}`,
+                        }}
+                    >
+                        +{Math.min(overflow, MAX_OVERFLOW_DISPLAY_NUMBER)}
```
kwasniew, line 94: Personally I was confused by the + sign here. Maybe sth like: `Total: 3452` would be more obvious?
But no strong opinion here so I'm also ok with the current approach.

thomasheartman: Do you mean in the tooltip or in the avatar circle? I was considering doing that for the tooltip, yeah. So if that's what you mean, I'll do that before merging. If it's in the circle, then I don'th think we have that much space 😅

kwasniew: tooltip

thomasheartman: yeah, cool. then we're aligned. I'll update

PR #7880, opened 2024-08-14
fix: always provide empty segments list in feature env strategies

kwasniew:
## About the changes

When reading feature env strategies and there's no segments it returns empty list of segments now. Previously it was undefined leading to overly verbose change request diffs.

Screenshot 2024-08-14 at 16 06 14


ivarconr (approved):
Thanks. 🎉💥

PR #7883, opened 2024-08-15
feat: resolve useragent source and add as source label to metrics

daveleek:
Tries to match known user agent patterns to a label that gets added to metrics, so we can learn more about incoming integrations being used

image

### DIFF
--- a/src/lib/middleware/integration-headers.ts
+++ b/src/lib/middleware/integration-headers.ts
@@ -7,7 +7,6 @@ const userAgentMatches = [
     { label: 'Curl', matcher: /^curl/ },
     { label: 'Go', matcher: /^Go-http-client/ },
     { label: 'Python', matcher: /^python-requests/ },
-    { label: 'Node', matcher: /^node-fetch/ },
     { label: 'Node', matcher: /^node/ },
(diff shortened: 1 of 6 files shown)


ivarconr (approved):

src/lib/middleware/integration-headers.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const ORIGIN = 'origin';
+const httpMatcher = /^https?:\/\//;
+const userAgentMatches = [
+    { label: 'Axios', matcher: /^axios/ },
+    { label: 'Curl', matcher: /^curl/ },
+    { label: 'Go', matcher: /^Go-http-client/ },
+    { label: 'Python', matcher: /^python-requests/ },
+    { label: 'Node', matcher: /^node-fetch/ },
+    { label: 'Node', matcher: /^node/ },
```
gardleopard, line 11: wont this match node-fetch as well?

gardleopard (approved):
LGTM, i think node-fetch regex is redundant.

src/lib/middleware/integration-headers.ts
daveleek: Good point thanks, I'll remove!

PR #7884, opened 2024-08-15
feat: support private projects for event search

sjaanus:
Adds private projects support for event search. Unit test should cover the usecases.

### DIFF
--- a/src/lib/features/events/event-service.test.ts
+++ b/src/lib/features/events/event-service.test.ts
@@ -20,7 +20,7 @@ describe('filterPrivateProjectsFromParams', () => {
             mode: 'all',
         };
 
-        const projectParam = 'IS:project1,IS:project3';
+        const projectParam = 'IS_ANY_OF:project1,IS:project3';
 
         const result = filterAccessibleProjects(projectParam, projectAccess);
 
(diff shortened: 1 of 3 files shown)


src/lib/features/events/event-service.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        const result = filterPrivateProjectsFromParams(
+            projectParam,
+            projectAccess,
+        );
+
+        expect(result).toBe('IS:project1');
+    });
+
+    it('should throw an error if no projects match and not throw an error', () => {
```
thomasheartman, line 56: Huh? What's this test supposed to say?  Something like this?
```suggestion
    it('should throw an error if no projects match', () => {
```

src/lib/features/events/event-service.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        expect(result).toBe('IS_ANY_OF:project1,project2');
+    });
+
+    it('should return the original projectParam when mode is all', () => {
+        const projectAccess: ProjectAccess = {
+            mode: 'all',
+        };
+
+        const projectParam = 'IS:project1,IS:project3';
```
thomasheartman, line 23: Project param the query parameter in the URL, right? If you have multiple projects, it's formatted as an `IS_ANY_OF:`, not as a concatenated list of `IS` (e.g. https://sandbox.getunleash.io/enterprise/history?query=&limit=25&offset=0&project=IS_ANY_OF%3A61ab6557353d%2Candreas_test).

We should probably use the same kinda query string, yeah?

(Side note: how does the API handle multiples of the same query param? E.g. "project=IS_ANY_OF:a,b&project=IS:c"?)

thomasheartman (approved):

src/lib/features/events/event-service.test.ts
sjaanus: Yeah, fixed

src/lib/features/events/event-service.test.ts
sjaanus: Fixed. Side note, not sure actually, but it is not supposed to be used that way, and only doable via API.

PR #7886, opened 2024-08-15
fix: exclude archived features in segments count

kwasniew:
## About the changes

Exclude archived feature flags from the segment usage count calculations

Making sure this list count
Screenshot 2024-08-15 at 10 09 36
matches this strategy values
Screenshot 2024-08-15 at 10 09 20

### DIFF
--- a/src/lib/features/segment/segment-store.test.ts
+++ b/src/lib/features/segment/segment-store.test.ts
@@ -48,7 +48,7 @@ describe('usage counting', () => {
         await db.rawDatabase.table('change_requests').delete();
     });
 
-    test('segment usage in active CRs is counted if we ask for it', async () => {
+    test('segment usage in active CRs is counted iff we ask for it', async () => {
         const CR_ID = 54321;
 
         const flag1 = await db.stores.featureToggleStore.create('default', {
--- a/src/lib/features/segment/segment-store.test.ts
+++ b/src/lib/features/segment/segment-store.test.ts
@@ -48,7 +48,7 @@ describe('usage counting', () => {
         await db.rawDatabase.table('change_requests').delete();
     });
 
-    test('segment usage in active CRs is counted iff we ask for it', async () => {
+    test('segment usage in active CRs is counted if we ask for it', async () => {
         const CR_ID = 54321;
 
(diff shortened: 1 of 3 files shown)


src/lib/features/segment/segment-store.ts
```diff
@@ -123,17 +124,15 @@ export default class SegmentStore implements ISegmentStore {
...
-            )
-            .countDistinct(
-                `${T.featureStrategies}.project_name AS used_in_projects`,
-            )
-            .countDistinct(
-                `${T.featureStrategies}.feature_name AS used_in_features`,
-            )
+            .select([
+                ...this.prefixColumns(),
+                this.db.raw(
```
kwasniew, line 129: The COUNT functions are now wrapped with CASE statements. These CASE statements ensure that only non-archived features (where features.archived_at is NULL) are counted. If a feature is archived, the CASE returns NULL, which does not affect the count, effectively making it 0

src/lib/features/segment/segment-store.ts
```diff
@@ -145,6 +144,11 @@ export default class SegmentStore implements ISegmentStore {
                 `${T.featureStrategies}.id`,
                 `${T.featureStrategySegment}.feature_strategy_id`,
             )
+            .leftJoin(
```
kwasniew, line 147: added a new join to get feature archived info

kwasniew: I also tried the where clause just before groupBy instead of rewriting counts but it didn't work.

src/lib/features/segment/segment-store.test.ts
```diff
@@ -48,7 +48,7 @@ describe('usage counting', () => {
         await db.rawDatabase.table('change_requests').delete();
     });
 
-    test('segment usage in active CRs is counted iff we ask for it', async () => {
+    test('segment usage in active CRs is counted if we ask for it', async () => {
```
thomasheartman, line 51: This wasn't actually a typo, but shorthand for "if and only if", but maybe I should've written it out? 😅

sjaanus (approved):
LG

PR #7899, opened 2024-08-15
feat: Show project archived message

kwasniew:
## About the changes

When you get to the url of an archived project you will be given similar message to archived flag. It has a link to all archived projects list. I deliberately keep UI simple since very few people should end up here.
Screenshot 2024-08-15 at 14 39 26

Aslo fixed time ago component to avoid ticking for the first 60 seconds


Tymek (commented):
LGTM

PR #7901, opened 2024-08-16
fix: orphaned token label only for items without projects

Tymek:
## About the changes
Orphaned token label was incorrectly assigned.

### DIFF
--- a/frontend/src/component/admin/apiToken/ApiTokenIcon/ApiTokenIcon.test.tsx
+++ b/frontend/src/component/admin/apiToken/ApiTokenIcon/ApiTokenIcon.test.tsx
@@ -55,6 +55,12 @@ describe('isOrphanedToken', () => {
                 projects: ['a', 'b'],
             }),
         ).toBe(false);
+          expect(
+              isOrphanedToken({
+                  secret: '[]:development.be7536c3a160ff15e3a92da45de531dd54bc1ae15d8455c0476f086b',
+                  projects: ['a'],
+              }),
+          ).toBe(false);
     });
 });
 
--- a/frontend/src/component/admin/apiToken/ApiTokenIcon/ApiTokenIcon.test.tsx
+++ b/frontend/src/component/admin/apiToken/ApiTokenIcon/ApiTokenIcon.test.tsx
@@ -1,7 +1,62 @@
 import { render } from 'utils/testRenderer';
 import { screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { ApiTokenIcon } from './ApiTokenIcon';
+import { ApiTokenIcon, isOrphanedToken } from './ApiTokenIcon';
+
+describe('isOrphanedToken', () => {
+    it('should be true for wildcard tokens with secret indicating it is orphaned', () => {
+        expect(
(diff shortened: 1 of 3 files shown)


frontend/src/component/admin/apiToken/ApiTokenIcon/ApiTokenIcon.test.tsx
thomasheartman, line 1: Have you tested all invariants here now? What about if you send just `project` without projects? Are there any other cases that're missing?

I wonder if it would be easier to unit test if you extract the `isOrphanedToken` into its own function that only does that 🤷🏼

frontend/src/component/admin/apiToken/ApiTokenIcon/ApiTokenIcon.tsx
```diff
@@ -11,10 +11,30 @@ interface IApiTokenIconProps {
...
+    projects,
+}) => {
     const tokenFormat = secret?.includes(':') ? 'v2' : 'v1'; // see https://docs.getunleash.io/reference/api-tokens-and-client-keys#format
     const isWildcardToken = secret?.startsWith('*:');
-    const isOrphanedToken = tokenFormat === 'v2' && !isWildcardToken;
+    const hasProjects = useMemo(() => {
+        if (
+            projects &&
+            Array.isArray(projects) &&
+            projects.length > 1 &&
```
thomasheartman, line 25: what about the case where you have `projects: ['a']`? That's a potential pitfall here, isn't it? Shouldn't this be `projects.length >= 1`?

frontend/src/component/admin/apiToken/ApiTokenIcon/ApiTokenIcon.tsx
```diff
@@ -11,10 +11,30 @@ interface IApiTokenIconProps {
...
+            projects[0] !== '*'
+        ) {
+            return true;
+        }
+        if (project && project !== '*') {
+            return true;
+        }
```
FredrikOseberg, line 32: Isn't this covered by the isWildCard property?

FredrikOseberg (approved):
Looks good

frontend/src/component/admin/apiToken/ApiTokenIcon/ApiTokenIcon.test.tsx
```diff
@@ -1,7 +1,62 @@
...
+                project: 'test',
+            }),
+        ).toBe(false);
+        expect(
+            isOrphanedToken({
+                secret: '[]:development.be7536c3a160ff15e3a92da45de531dd54bc1ae15d8455c0476f086b',
+                projects: ['a', 'b'],
+            }),
+        ).toBe(false);
+    });
```
thomasheartman, line 58: What if a token only has one project left but was made for multiple tokens?  Will this pass? (I think it will now, but just to be sure)

```suggestion
          expect(
              isOrphanedToken({
                  secret: '[]:development.be7536c3a160ff15e3a92da45de531dd54bc1ae15d8455c0476f086b',
                  projects: ['a'],
              }),
          ).toBe(false);
    });
```

thomasheartman (approved):

PR #7903, opened 2024-08-16
fix: orphaned token label patch

Tymek:
https://github.com/Unleash/unleash/pull/7901


FredrikOseberg (approved):
LG

Tymek:
Updated to match commit on `main` https://github.com/Unleash/unleash/commit/b9ea24be8d2b63a7fdff8d33e81f707465f4322e

PR #7904, opened 2024-08-16
refactor: switch `projectStore.getProjects` with `projectReadModel.getProjectsForAdminUi` in project service

thomasheartman:
Hooks up the new project read model and updates the existing project service to use it instead when the flag is on.

In doing:
- creates a composition root for the read model
- includes it in IUnleashStores
- updates some existing methods to accept either the old or the new model
- updates the OpenAPI schema to deprecate the old properties

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -32,7 +32,6 @@ import {
     type IProjectRoleUsage,
     type IProjectStore,
     type IProjectUpdate,
-    type IProjectWithCount,
     type IUnleashConfig,
     type IUnleashStores,
     MOVE_FEATURE_TOGGLE,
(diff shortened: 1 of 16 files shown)


src/lib/features/project/project-service.ts
```diff
@@ -214,16 +219,18 @@ export default class ProjectService {
...
-        const projects = await this.projectStore.getProjectsWithCounts(
-            query,
-            userId,
-        );
+    ): Promise<TransitionalProjectData[]> {
+        const getProjects = this.flagResolver.isEnabled('useProjectReadModel')
+            ? this.projectReadModel.getProjectsForAdminUi
+            : this.projectStore.getProjectsWithCounts;
+
+        const projects = await getProjects(query, userId);
```
thomasheartman, line 234: This is the main change. With feature flag off, everything will work as before. With flag on, use the new read model instead.

sjaanus (approved):
LG

PR #7905, opened 2024-08-16
chore: add flags `projectListImprovements` and `useProjectReadModel`

thomasheartman:
These are both related to the work on the project list improvements project.

The `projectListImprovements` flag will be used to enable disable the
new project list improvements.

The `useProjectReadModel` flag will be used to enable/disable the use
of the new project read model and is mostly a safety feature.


frontend/src/interfaces/uiConfig.ts
thomasheartman, line 1: The UI never needs to know about which query the backend uses.

sjaanus (approved):

PR #7913, opened 2024-08-19
feat: static link to archived flags in archived project

kwasniew:
## About the changes

A link to archived project archived flags is always there without a number of archived flags (we decided not to expose it for now). 
Screenshot 2024-08-19 at 10 51 59

I decided to choose a link similar to view archive link in the flag search
Screenshot 2024-08-19 at 10 54 08

I also played with the card title link but it  felt like it leads you to full project information that we don't want to show for the archived projects.

### DIFF
--- a/frontend/src/component/project/NewProjectCard/NewProjectCard.styles.ts
+++ b/frontend/src/component/project/NewProjectCard/NewProjectCard.styles.ts
@@ -38,6 +38,7 @@ export const StyledDivHeader = styled('div')(({ theme }) => ({
     ...flexRow,
     width: '100%',
     marginBottom: theme.spacing(2),
+    gap: theme.spacing(1),
 }));
 
 export const StyledCardTitle = styled('h3')(({ theme }) => ({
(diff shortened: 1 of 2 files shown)


frontend/src/component/project/NewProjectCard/NewProjectCard.styles.ts
```diff
@@ -38,6 +38,7 @@ export const StyledDivHeader = styled('div')(({ theme }) => ({
     ...flexRow,
     width: '100%',
     marginBottom: theme.spacing(2),
+    gap: theme.spacing(1),
```
kwasniew, line 41: added this gap so that the private icon doesn't clash with the title

frontend/src/component/project/NewProjectCard/NewProjectCard.styles.ts
```diff
@@ -89,7 +90,7 @@ export const StyledIconBox = styled(Box)(({ theme }) => ({
     display: 'grid',
     placeItems: 'center',
     padding: theme.spacing(0, 0.5, 0, 1),
-    marginRight: theme.spacing(2),
+    marginRight: theme.spacing(1),
```
kwasniew, line 93: reduced by 1 spacing because gap was added

Tymek (approved):

PR #7914, opened 2024-08-19
feat: filter projectless events for normal users

sjaanus:
Now events that do not have project ( for example user creation, segment creation etc), will not be displayed to non root admins.

### DIFF
--- a/src/lib/features/access/access-read-model.ts
+++ b/src/lib/features/access/access-read-model.ts
@@ -1,0 +1,28 @@
+import {
+    ADMIN_TOKEN_USER,
+    type IAccessStore,
+    type IUnleashStores,
+    SYSTEM_USER_ID,
+} from '../../types';
+import type { IAccessReadModel } from './access-read-model-type';
+import * as permissions from '../../types/permissions';
+
+const { ADMIN } = permissions;
+
+export class AccessReadModel implements IAccessReadModel {
+    private store: IAccessStore;
+
+    constructor({ accessStore }: Pick<IUnleashStores, 'accessStore'>) {
+        this.store = accessStore;
+    }
+
+    async isRootAdmin(userId: number): Promise<boolean> {
+        if (userId === SYSTEM_USER_ID || userId === ADMIN_TOKEN_USER.id) {
+            return true;
+        }
+        const roles = await this.store.getRolesForUserId(userId);
+        return roles.some(
+            (role) => role.name.toLowerCase() === ADMIN.toLowerCase(),
+        );
+    }
+}
--- a/src/lib/features/events/event-service.ts
+++ b/src/lib/features/events/event-service.ts
@@ -16,6 +16,7 @@ import { parseSearchOperatorValue } from '../feature-search/search-utils';
 import { endOfDay, formatISO } from 'date-fns';
 import type { IPrivateProjectChecker } from '../private-project/privateProjectCheckerType';
 import type { ProjectAccess } from '../private-project/privateProjectStore';
+import type { IAccessReadModel } from '../access/access-read-model-type';
 
 export default class EventService {
     private logger: Logger;
@@ -23,6 +24,8 @@ export default class EventService {
     private eventStore: IEventStore;
 
     private featureTagStore: IFeatureTagStore;
+
+    private accessReadModel: IAccessReadModel;
 
     private privateProjectChecker: IPrivateProjectChecker;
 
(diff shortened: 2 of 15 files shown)


src/lib/features/access/access-read-model.ts
```diff
@@ -0,0 +1,28 @@
...
+    type IAccessStore,
+    type IUnleashStores,
+    SYSTEM_USER_ID,
+} from '../../types';
+import type { IAccessReadModel } from './access-read-model-type';
+import * as permissions from '../../types/permissions';
+
+const { ADMIN } = permissions;
+
+export class AccessReadModel implements IAccessReadModel {
```
sjaanus, line 12: Pulling this out into read model, because otherwise we had circular dependency.

src/test/e2e/api/admin/event-search.e2e.test.ts
```diff
@@ -390,3 +455,63 @@ test('should filter events by project using IS_ANY_OF', async () => {
         total: 2,
     });
 });
+
+test('should not show user creation events for non-admins', async () => {
```
sjaanus, line 459: This tests the main logic.

src/lib/features/events/event-service.ts
```diff
@@ -222,6 +229,14 @@ export default class EventService {
     async getEventCreators() {
         return this.eventStore.getEventCreators();
     }
+
+    async getProjectFilterForNonAdmins(userId: number): Promise<IQueryParam[]> {
+        const isRootAdmin = await this.accessReadModel.isRootAdmin(userId);
```
sjaanus, line 234: This is the main logic. If you are not root admin, add db query filter to not include projects that have null as value.

src/lib/features/feature-search/search-utils.ts
```diff
@@ -50,14 +50,19 @@ export const applyGenericQueryParams = (
...
                 query.whereIn(param.field, param.values);
                 break;
             case 'IS_NOT':
             case 'IS_NONE_OF':
-                query.whereNotIn(param.field, param.values);
+                if (isSingleParam) {
+                    query.whereNot(param.field, param.values[0]);
+                } else {
+                    query.whereNotIn(param.field, param.values);
+                }
```
thomasheartman, line 65: Is this necessary? A list with a single list is still a lilt, right? So `is not in [a]` and `is not in [a, b]` should both be valid?

src/lib/features/segment/segment-service.ts
```diff
@@ -151,10 +151,11 @@ export class SegmentService implements ISegmentService {
         await this.validateName(input.name);
         const segment = await this.segmentStore.create(input, auditUser);
 
+        const project = segment.project;
         await this.eventService.storeEvent(
             new SegmentCreatedEvent({
                 data: segment,
-                project: segment.project || 'no-project',
+                project: project,
```
thomasheartman, line 158: Not sure why you're extracting the variable here. It's only used once, and I don't think this adds any more clarity?

But if you want to do it, we can at least avoid doubling up:

```suggestion
                project,
```

Your call.

src/lib/types/events.ts
thomasheartman, line 1: Good type clarification here. This is just making the types more accurate to how we actually use them, right?

src/test/e2e/api/admin/event-search.e2e.test.ts
```diff
@@ -182,13 +246,15 @@ test('should filter events by type', async () => {
...
+        project: 'default',
         createdBy: 'admin1@example.com',
         createdByUserId: TEST_USER_ID + 1,
         ip: '127.0.0.1',
     });
 
     await eventService.storeEvent({
         type: FEATURE_CREATED,
+        project: 'default',
         createdBy: 'admin2@example.com',
```
thomasheartman, line 258: Interesting: did this test not work without the project? If so, why not?

src/lib/features/events/event-service.ts
thomasheartman: Oh, right. So effectively, this removes any non-project events for non-admins.

src/lib/features/events/event-service.ts
```diff
@@ -222,6 +229,14 @@ export default class EventService {
     async getEventCreators() {
         return this.eventStore.getEventCreators();
     }
+
+    async getProjectFilterForNonAdmins(userId: number): Promise<IQueryParam[]> {
```
thomasheartman, line 233: Small suggestion here: the naming was a little unclear to me at first, because it doesn't seem like we know whether the user is an admin or not at this point. If you read the method, you'll see that it's checked in there, though.

I'm not sure I have any better suggestions, but maybe `getRoleBasedFilters` would work? I ... dunno 🤷🏼

thomasheartman (approved):
Nike ✅

src/lib/features/feature-search/search-utils.ts
sjaanus: Yes, it is because it is doing following sql query:

not in (x), and if you put in null for project , not in (null), then it will not work, because SQL things.

src/test/e2e/api/admin/event-search.e2e.test.ts
sjaanus: Actually since feature created always has project, wanted to make it legit.

PR #7915, opened 2024-08-19
fix: avoid awkward overlaps of group cards with lots of members, long project names, and small cards

thomasheartman:
This PR updates the styling of the group cards to better handle edge cases where you have a lot of assigned projects, long project names, lots of members, etc.

In particular, it does the following things:
- aligns the avatars along the bottom of the card, so that even if there's a lot of projects, the avatars stay close to the bottom edge
- adds word breaks for the project names, so that long names can break when they need to
- adds some spacing between the two columns in the bottom row, so that even when you they get close, they never quite touch.

Note: there is one more thing I'd like to address in a follow up: as shown in the top row of the after image, there's some extra wrapping of the first "This group has no users", even though it has the room to grow. I'll keep looking into this and make a follow-up.

Before:
image

After:

image

### DIFF
--- a/frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
+++ b/frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
@@ -76,6 +76,7 @@ const ProjectBadgeContainer = styled('div')(({ theme }) => ({
     justifyContent: 'flex-end',
     gap: theme.spacing(0.5),
     flexWrap: 'wrap',
+    wordWrap: 'anywhere',
 }));
 
 const InfoBadgeDescription = styled('span')(({ theme }) => ({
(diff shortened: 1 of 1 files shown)


frontend/src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
```diff
@@ -85,6 +87,10 @@ const InfoBadgeDescription = styled('span')(({ theme }) => ({
     fontSize: theme.fontSizes.smallBody,
 }));
 
+const ProjectNameBadge = styled(Badge)({
+    wordwrap: 'anywhere',
```
kwasniew, line 91: I haven't seen this anywhere else in our codebase. How would it work with 'break-word'?

thomasheartman: looks like styled has some issues with the word-wrap property (`wordWrap` would render correctly as `word-wrap`, but linter told me the property was invalid; `wordwrap` was accepted by the linter, but rendered as `wordwrap` so it didn't work), so I'll go with `word-break: break-word` instead.

PR #7919, opened 2024-08-19
feat: add service name to webhooks parameters and log when set

daveleek:
Adds a field for "integration service name" to the webhooks integration, and logs it to info when it is set during creation/update

Skjermbilde 2024-08-19 kl  14 30 00

Skjermbilde 2024-08-19 kl  14 30 28

Skjermbilde 2024-08-19 kl  11 20 06


sighphyre (approved):
Doesn't seem harmful but it does seem weird to me. If this field goes into the database, surely we can just read it from there rather than hoping it appears in our logs? And why is only the logging behind a flag? What are we hoping to be able to do there?

gastonfournier (changes requested):
We do have a section with a description
image

I know that name might make our users be more precise than having a description, but I wonder if we should are more information. If it's optional, how can we trust it? Can we instead try to guess based on the URL they're connecting to? It's an open question, not that I know the answer, but might be more efficient than asking for another optional input. We know about laziness... 

Maybe it's also worthy to include UX on this one. I'm requesting changes so we have the opportunity to talk about this, nothing worrying about the code

daveleek:
@sighphyre 
> If this field goes into the database, surely we can just read it from there rather than hoping it appears in our logs?

The existing origin stuff is in logs so we can put this in the same dashboard

> And why is only the logging behind a flag?

The thinking was to be able to remove this once we're satisfied but I guess that's a quick commit as well. Can remove flag if that's too redundant

@gastonfournier 
> but I wonder if we should are more information

Can absolutely improve the description of the field

> Can we instead try to guess based on the URL they're connecting to

Yeah could perhaps try to match domain and log that? Do we still want to log first and then populate known?

> Maybe it's also worthy to include UX on this one.

Good point!

daveleek:
After some discussions we're closing this one and doing it differently

PR #7921, opened 2024-08-19
Feat/metrics for outgoing integrations

daveleek:
Publishes metric-events results when addons have handled events, which then gets picked up and counted by metrics with result and destination for outgoing integrations

Skjermbilde 2024-08-19 kl  11 17 52

### DIFF
--- a/src/lib/addons/addon.test.ts
+++ b/src/lib/addons/addon.test.ts
@@ -16,6 +16,7 @@ const ARGS: IAddonConfig = {
     unleashUrl: url,
     integrationEventsService: {} as IntegrationEventsService,
     flagResolver: {} as IFlagResolver,
+    eventBus: {} as any,
 };
 
 test('Does not retry if request succeeds', async () => {
--- a/src/lib/addons/webhook.ts
+++ b/src/lib/addons/webhook.ts
@@ -102,6 +102,13 @@ export default class Webhook extends Addon {
             state = 'failed';
             const failedMessage = `Webhook request failed with status code: ${res.status}.`;
             stateDetails.push(failedMessage);
+            if (this.flagResolver.isEnabled('addonUsageMetrics')) {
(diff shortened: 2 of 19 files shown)


src/lib/addons/webhook.ts
```diff
@@ -94,6 +103,13 @@ export default class Webhook extends Addon {
             state = 'failed';
             const failedMessage = `Webhook request failed with status code: ${res.status}.`;
             stateDetails.push(failedMessage);
+            if (this.flagResolver.isEnabled('addonUsageMetrics')) {
+                this.eventBus.emit(ADDON_EVENTS_HANDLED, {
+                    result: state,
+                    destination: 'webhook',
```
sighphyre, line 109: Probably makes sense but intentional that this is `webhook` and not `serviceName`?

src/lib/addons/addon.test.ts
```diff
@@ -16,6 +16,7 @@ const ARGS: IAddonConfig = {
     unleashUrl: url,
     integrationEventsService: {} as IntegrationEventsService,
     flagResolver: {} as IFlagResolver,
+    eventBus: {} as any,
```
sighphyre, line 19: Maybe it's okay because it's a test, idk, but I hate `any`. I see this file `src/test/e2e/features-created-by-user-id-data-migration.e2e.test.ts` just seems to import the event bus without problems. Maybe we can do that?

sighphyre (approved):
Yeah I think it LGTM

src/lib/addons/webhook.ts
daveleek: Yeah we'll add known ones to a static list of known service names later and cross check against that so we don't fill metrics up with random stuff

src/lib/addons/addon.test.ts
daveleek: thanks, I'll have a look at that!

PR #7923, opened 2024-08-19
fix: last seen metrics exceeding table limits

gastonfournier:
## About the changes
When storing last seen metrics we no longer validate at insert time that the feature exists. Instead, there's a job cleaning up on a regular interval. 

Metrics for features with more than 255 characters, makes the whole batch to fail, resulting in metrics being lost.

This PR helps mitigate the issue while also logs long name feature names

### DIFF
--- a/src/lib/features/metrics/last-seen/last-seen-store.ts
+++ b/src/lib/features/metrics/last-seen/last-seen-store.ts
@@ -48,7 +48,6 @@ export default class LastSeenStore implements ILastSeenStore {
     async setLastSeen(data: LastSeenInput[]): Promise<void> {
         try {
             const inserts = prepareLastSeenInput(data);
-            console.log('inserts', inserts);
             const batchSize = 500;
 
             for (let i = 0; i < inserts.length; i += batchSize) {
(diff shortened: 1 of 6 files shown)


src/lib/features/metrics/last-seen/last-seen-store.ts
```diff
@@ -48,7 +48,7 @@ export default class LastSeenStore implements ILastSeenStore {
     async setLastSeen(data: LastSeenInput[]): Promise<void> {
         try {
             const inserts = prepareLastSeenInput(data);
-
+            console.log('inserts', inserts);
```
kwasniew, line 51: we can remove it

kwasniew (approved):

src/lib/features/metrics/last-seen/last-seen-store.ts
gastonfournier: [7f030b8](https://github.com/Unleash/unleash/pull/7923/commits/7f030b8f0f21fea72abdc026120f4dfe95824d7f)

PR #7929, opened 2024-08-20
feat: project actions count metric

kwasniew:
## About the changes

Added counter for project actions: created, archived, deleted and revived.

Design decision: 
* one counter with project and action labels instead of 4 different counters

Also updated one typo in a dialog.

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -19,6 +19,10 @@ import {
     CLIENT_METRICS,
     CLIENT_REGISTER,
     PROJECT_ENVIRONMENT_REMOVED,
+    PROJECT_CREATED,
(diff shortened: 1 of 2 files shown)


src/lib/metrics.ts
```diff
@@ -315,6 +319,12 @@ export default class MetricsMonitor {
             labelNames: ['stage'],
         });
 
+        const projectActionsCounter = createCounter({
+            name: 'project_actions_count',
+            help: 'Count project actions',
+            labelNames: ['project_id', 'action'],
```
Tymek, line 325: do we need `project_id`?

Tymek (approved):

PR #7930, opened 2024-08-20
fix: only show addon delete button when editing addons

daveleek:
Fixes so that delete button doesn't show up when creating integrations, only when editing them


gastonfournier (approved):
LGTM! It's excluding the divider as well :+1: 

This is the image before the change when trying to add a new webhook
image

PR #7931, opened 2024-08-20
fix: don't count flags multiple times (bonus: don't count non-project events)

thomasheartman:
This PR fixes an issue where the number of flags belonging to a project was wrong in the new getProjectsForAdminUi.

The cause was that we now join with the events table to get the most "lastUpdatedAt" data. This meant that we got multiple rows for each flag, so we counted the same flag multiple times. The fix was to use a "distinct".

Additionally, I used this as an excuse to write some more tests that I'd been thinking about. And in doing so also uncovered another bug that would only ever surface in verrry rare conditions: if a flag had been created in project A, but moved to project B AND the feature-project-change event hadn't fired correctly, project B's last updated could show data from that feature in project A.

I've also taken the liberty of doing a little bit of cleanup.

### DIFF
--- a/src/lib/features/project/project-owners-read-model.test.ts
+++ b/src/lib/features/project/project-owners-read-model.test.ts
@@ -20,7 +20,7 @@ const mockProjectData = (name: string): ProjectForUi => ({
     createdAt: new Date(),
     favorite: false,
     lastReportedFlagUsage: null,
-    lastFlagUpdate: null,
+    lastUpdatedAt: null,
 });
 
 describe('unit tests', () => {
--- a/src/lib/features/project/project-read-model.test.ts
+++ b/src/lib/features/project/project-read-model.test.ts
@@ -45,35 +45,11 @@ beforeEach(async () => {
     await flagStore.deleteAll();
 });
 
-test('it gets the right number of flags for a project', async () => {
-    await projectStore.create({ id: 'a', name: 'A' });
-    await projectStore.create({ id: 'b', name: 'B' });
-
-    const noFlags = await projectReadModel.getProjectsForAdminUi();
-    expect(noFlags).toEqual(
-        expect.arrayContaining([
-            expect.objectContaining({ id: 'a', featureCount: 0 }),
-            expect.objectContaining({ id: 'b', featureCount: 0 }),
-        ]),
-    );
-
-    await flagStore.create('a', { name: 'x', createdByUserId: 1 });
-    await flagStore.create('a', { name: 'y', createdByUserId: 1 });
-    await flagStore.create('b', { name: 'z', createdByUserId: 1 });
-
-    const withFlags = await projectReadModel.getProjectsForAdminUi();
-    expect(withFlags).toEqual(
-        expect.arrayContaining([
-            expect.objectContaining({ id: 'a', featureCount: 2 }),
-            expect.objectContaining({ id: 'b', featureCount: 1 }),
-        ]),
-    );
-});
-
-test('it gets the right number of flags for a project 2', async () => {
+test("it doesn't count flags multiple times when they have multiple events associated with them", async () => {
     await projectStore.create({ id: 'a', name: 'A' });
 
     await flagStore.create('a', { name: 'x', createdByUserId: 1 });
+
     await eventStore.store({
         type: 'feature-created',
         createdBy: 'admin',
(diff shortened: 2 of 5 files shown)


src/lib/features/project/project-owners-read-model.test.ts
```diff
@@ -20,7 +20,7 @@ const mockProjectData = (name: string): ProjectForUi => ({
     createdAt: new Date(),
     favorite: false,
     lastReportedFlagUsage: null,
-    lastFlagUpdate: null,
+    lastUpdatedAt: null,
```
thomasheartman, line 23: I've talked to stakeholder Jaanus about this, and we might want to include project updates here in the future, so we'll use a name that's less specifically about flags. Additionally: this name is still in development anyway, so it might change in the future.

src/lib/features/project/project-read-model.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        project: 'a',
+    });
+
+    const withFlags = await projectReadModel.getProjectsForAdminUi();
+    expect(withFlags).toEqual(
+        expect.arrayContaining([
+            expect.objectContaining({ id: 'a', featureCount: 1 }),
+        ]),
+    );
+});
```
thomasheartman: This was the cause of the bug.

src/lib/features/project/project-read-model.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        project: projectId,
+    });
+
+    const withFlags = await projectReadModel.getProjectsForAdminUi();
+    expect(withFlags).toEqual(
+        expect.arrayContaining([
+            expect.objectContaining({ id: projectId, featureCount: 1 }),
+        ]),
+    );
+});
```
thomasheartman, line 78: This is reproducing the bug that was encountered.

src/lib/features/project/project-read-model.test.ts
```diff
@@ -0,0 +1,162 @@
...
+            // has been moved from one project to the next (before the
+            // moving event has been counted)). In practice, you'll
+            // always get a "feature-project-change" event to count
+            expect.objectContaining({ id: projectId1, lastUpdatedAt: null }),
+            // this flag no longer exists in this project because it
+            // has been moved, so it should not be counted
+            expect.objectContaining({ id: projectId2, lastUpdatedAt: null }),
+        ]),
+    );
+});
```
thomasheartman, line 142: This uncovered the second, super edge-casey bug.

src/lib/features/project/project-read-model.ts
```diff
@@ -92,7 +98,7 @@ export class ProjectReadModel implements IProjectReadModel {
         let selectColumns = [
             this.db.raw(
                 'projects.id, projects.name, projects.description, projects.health, projects.created_at, ' +
-                    'count(features.name) FILTER (WHERE features.archived_at is null) AS number_of_features, ' +
+                    'count(DISTINCT features.name) FILTER (WHERE features.archived_at is null) AS number_of_features, ' +
```
thomasheartman, line 101: This fixes the main bug reported.

src/lib/features/project/project-read-model.ts
```diff
@@ -74,7 +74,13 @@ export class ProjectReadModel implements IProjectReadModel {
...
                 'projects.id',
             )
-            .leftJoin('events', 'events.feature_name', 'features.name')
+            .leftJoin('events', (join) => {
+                join.on('events.feature_name', '=', 'features.name').andOn(
+                    'events.project',
+                    '=',
+                    'projects.id',
+                );
+            })
```
thomasheartman, line 83: this fixes the secondary bug

src/lib/features/project/project-read-model.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        type: 'feature-updated',
+        createdBy: 'admin',
+        ip: '',
+        createdByUserId: 1,
+        featureName: flagName,
+        project: projectId,
+    });
+
+    const withFlags = await projectReadModel.getProjectsForAdminUi();
+    expect(withFlags).toEqual(
```
kwasniew, line 73: would `toMatchObject([{id: projectId, featureCount:1}])` work?

src/lib/features/project/project-read-model.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        ip: '',
+        createdByUserId: 1,
+        featureName: flagName,
+        project: projectId2,
+    });
+
+    const withEvents = await projectReadModel.getProjectsForAdminUi();
+
+    expect(withEvents).toEqual(
+        expect.arrayContaining([
```
kwasniew, line 131: again I'd use toMatchObject if possible so that we can visually see the important parts of the object

kwasniew (approved):

PR #7932, opened 2024-08-20
chore: change integration events api tag to addons

daveleek:
Changes the OpenAPI tag used for integration events to be 'Addons' like the rest of the addon APIs


gastonfournier (approved):
LGTM!

PR #7935, opened 2024-08-20
test: bump cypress timout temporarily to see if it helps identify hangup

daveleek:
Bump timeout temporarily with a second to see if that helps at all


daveleek:
Nope, didn't help

PR #7937, opened 2024-08-20
feat: Build frontend with SWC

Tymek:
Drop babel in favor of swc


alvinometric:
🥳 Let's add some of that Rust speed

PR #7943, opened 2024-08-20
Refactor: Remove `react-timeago`

Tymek:
## About the changes
Remove dependency 🫡
Replaced with ~~50~~ 58 LoC + 140 LoTC, using `date-fns`

### Related issues / Prior work
- https://github.com/nmn/react-timeago/issues/181
- https://github.com/Unleash/unleash/pull/7936
- Internal: [issue/1-2681](https://linear.app/unleash/issue/1-2681/api-tokens-table-can-show-incorrect-last-seen-date)

### Important files
- frontend/src/component/common/TimeAgo/TimeAgo.tsx
- frontend/src/component/common/TimeAgo/TimeAgo.test.tsx


frontend/src/component/common/TimeAgo/TimeAgo.tsx
```diff
@@ -0,0 +1,50 @@
+import { useEffect, useState, type FC } from 'react';
+import { formatDistanceToNow, secondsToMilliseconds } from 'date-fns';
+
+type TimeAgoProps = {
+    date: Date | number | string | null | undefined;
+    fallback?: string;
+    live?: boolean | number;
+};
```
thomasheartman, line 8: I was a little confused by the type here, because it serves two purposes, right? Either it's "auto-update for me", which is well described by live, or it's "update every n seconds", which is not very well explained by its name (`live`). 

I'm also surprised to find that `live: false` and `live: undefined` **don't** do the same thing. I think that in every other instance I've seen of boolean props in the app, `undefined` is the same as `false`.

Finally, it's not claer that refresh interval is seconds (and it's often miliseconds in JS, right?), so maybe we could be clearer about that.

I think that on a personal note, I'd prefer to see this split up as something like: 

```suggestion
type TimeAgoProps = {
    date: Date | number | string | null | undefined;
    fallback?: string;
} & ({  dontUpdate?: boolean }
  | { refreshIntervalSeconds: number });
```

So that you avoid weird conflicts such as `dontUpdate={true} refreshIntervalSeconds={5}`, but also so that the default is still to update, as in the original code.

Optionally, we could take a cue from our SDKs and do something like this:

```suggestion
type TimeAgoProps = {
    date: Date | number | string | null | undefined;
    fallback?: string;
    refreshIntervalSeconds?: number
};
```

Where the default refresh interval is 12 (if left out), but you can set it to `0` to disable refreshing?

I'm happy to hear other suggestions here, though, I was just a little surprised by how this works.

frontend/src/component/common/TimeAgo/TimeAgo.test.tsx
thomasheartman, line 1: Looks like a solid test suite 🙌🏼  But it doesn't appear to test how it acts when `live` is undefined, which we might wanna chuck in too 💁🏼

thomasheartman (approved):
This is a cool change 😄 I'm very in favor of this overall, but have a few questions:

1. Does this look exactly the same as before? There are no screenies attached, so I'm guessing it's a drop-in replacement visually?

2. In what cases do we **not** want the component to update? 

Question 2 is also pertinent to my inline comment regarding the props for that component. I'm not entirely convinced that the `live?: number | boolean` prop is the right way to format the API. But let's talk about it 💁🏼

frontend/src/component/common/TimeAgo/TimeAgo.tsx
Tymek: Good point. We should have it not configurable, using opinionated `12s` value. (60/5). It feels "alive" enough, we don't use seconds, it will be less then refetch interval on our SWR hooks.

frontend/src/component/common/TimeAgo/TimeAgo.tsx
```diff
@@ -4,47 +4,54 @@ import { formatDistanceToNow, secondsToMilliseconds } from 'date-fns';
...
-    }, [live]);
+    }, [refresh]);
+
+    if (!state.dateTime) {
+        return state.description;
+    }
 
-    return value;
+    return (
+        <time dateTime={state.dateTime.toISOString()}>{state.description}</time>
```
thomasheartman, line 53: Nice 💯

frontend/src/component/common/TimeAgo/TimeAgo.tsx
```diff
@@ -4,47 +4,54 @@ import { formatDistanceToNow, secondsToMilliseconds } from 'date-fns';
...
-    return secondsToMilliseconds(input);
-};
-
-const TimeAgo: FC<TimeAgoProps> = ({ date, fallback = '', live }) => {
-    const getValue = () => {
-        if (!date) return fallback;
+const TimeAgo: FC<TimeAgoProps> = ({
+    date,
+    fallback = '',
+    refresh = true,
```
thomasheartman, line 14: same point as before: it's weird to have an optional boolean prop default to `true`. We don't do that anywhere else as far as I'm aware. Would it make sense to use "negative" here instead, e.g. "dontRefresh", so that `undefined` and `false` have the same meaning?

Tymek: We have it in
- `frontend/src/component/feature/StrategyTypes/FlexibleStrategy/FlexibleStrategy.tsx`
- `frontend/src/component/project/ProjectList/ProjectGroup.tsx`
- `frontend/src/component/common/FeatureStatusChip/FeatureStatusChip.tsx`
- `frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx`

PR #7949, opened 2024-08-21
fix: project owner name overflow

kwasniew:
## About the changes

Truncate long project owner names. Had to add minWidth other than auto on the flex item that contains the avatar and the project name so that it doesn't grow outside of the parent. Previously the ellipsis had no effect.

### DIFF
--- a/frontend/src/component/project/NewProjectCard/ProjectOwners/ProjectOwners.tsx
+++ b/frontend/src/component/project/NewProjectCard/ProjectOwners/ProjectOwners.tsx
@@ -45,7 +45,6 @@ const StyledUserName = styled('p')(({ theme }) => ({
     textOverflow: 'ellipsis',
     textWrap: 'nowrap',
     alignSelf: 'end',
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/NewProjectCard/ProjectOwners/ProjectOwners.tsx
```diff
@@ -45,6 +45,7 @@ const StyledUserName = styled('p')(({ theme }) => ({
     textOverflow: 'ellipsis',
     textWrap: 'nowrap',
     alignSelf: 'end',
+    maxWidth: theme.spacing(21),
```
Tymek, line 48: maybe `100%` will work?

Tymek (approved):

thomasheartman (approved):
Nice one! Yeah, this might be another case where the same css-tricks article I used recently applies: https://css-tricks.com/preventing-a-grid-blowout/

PR #7950, opened 2024-08-21
fix: enable disabled strategies keeps settings

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/features/feature-toggle/tests/feature-toggle-service.e2e.test.ts
+++ b/src/lib/features/feature-toggle/tests/feature-toggle-service.e2e.test.ts
@@ -814,3 +814,51 @@ test('Should not allow to revive flags to archived projects', async () => {
         ),
     );
 });
+
+test('Should enable disabled strategies on feature environment enabled', async () => {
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):

src/lib/features/feature-toggle/tests/feature-toggle-service.e2e.test.ts
```diff
@@ -814,3 +814,51 @@ test('Should not allow to revive flags to archived projects', async () => {
         ),
     );
 });
+
+test('Should enable disabled strategies on feature environment enabled', async () => {
```
thomasheartman, line 818: Minor suggestion: can we be clearer about what we're checking here? Something like

```suggestion
test('Should not remove constraints or strategy variants on feature environment enabled', async () => {
```


```suggestion
test('Strategy configurations should not change on feature environment enabled', async () => {
```

thomasheartman: I guess I was late 🤷🏼

PR #7953, opened 2024-08-21
fix: serialize API token data correctly in instance stats

thomasheartman:
Turns out we've been trying to return API token data in instance stats for a while, but that the serialization has failed. Serializing a JS map just yields an empty object. 

This PR fixes that serialization and also adds API tokens to the instance stats schema (it wasn't before, but we did return it). Adding it to the schema is also part of making resource usage visible as part of the soft limits project.

### DIFF
--- a/src/lib/routes/admin-api/instance-admin.ts
+++ b/src/lib/routes/admin-api/instance-admin.ts
@@ -5,7 +5,6 @@ import type { IUnleashServices } from '../../types/services';
 import type { IUnleashConfig } from '../../types/option';
 import Controller from '../controller';
 import { NONE } from '../../types/permissions';
-import type { UiConfigSchema } from '../../openapi/spec/ui-config-schema';
 import type {
     InstanceStatsService,
     InstanceStatsSigned,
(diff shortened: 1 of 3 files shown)


src/lib/routes/admin-api/instance-admin.ts
```diff
@@ -129,11 +131,19 @@ class InstanceAdminController extends Controller {
...
 
     async getStatistics(
-        req: AuthedRequest,
-        res: Response<InstanceStatsSigned>,
+        _: AuthedRequest,
+        res: Response<InstanceAdminStatsSchema>,
     ): Promise<void> {
         const instanceStats = await this.instanceStatsService.getSignedStats();
-        res.json(instanceStats);
+        const apiTokensObj = Object.fromEntries(
```
kwasniew, line 138: why is it better to do it in the controller rather than the service itself?

kwasniew (approved):

src/lib/routes/admin-api/instance-admin.ts
thomasheartman: For visibility: slack thread with discussion: https://unleash-internal.slack.com/archives/C046LV6HH6W/p1724312294721579

PR #7957, opened 2024-08-22
feat: info-log webhooks domain part of url

daveleek:
Logging the domain part of the webhook url.
An effort to gain more knowledge around what type of integrations we would like to focus on

### DIFF
--- a/src/lib/services/addon-service.ts
+++ b/src/lib/services/addon-service.ts
@@ -225,6 +225,22 @@ export default class AddonService {
             `User ${auditUser.username} created addon ${addonConfig.provider}`,
         );
 
+        if (
+            addonConfig.provider === 'webhook' &&
+            addonConfig.parameters.url &&
+            addonConfig.parameters.url.indexOf('://') > 0
+        ) {
+            const start = addonConfig.parameters.url.indexOf('://') + 3;
+            let end = addonConfig.parameters.url.indexOf('/', start);
+            if (end === -1) {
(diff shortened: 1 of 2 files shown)


src/lib/services/addon-service.ts
```diff
@@ -259,6 +275,23 @@ export default class AddonService {
...
+            addonConfig.provider === 'webhook' &&
+            addonConfig.parameters.url &&
+            addonConfig.parameters.url.indexOf('://') > 0
+        ) {
+            const start = addonConfig.parameters.url.indexOf('://') + 3;
+            let end = addonConfig.parameters.url.indexOf('/', start);
+            if (end === -1) {
+                end = addonConfig.parameters.url.length;
+            }
+            const domain = addonConfig.parameters.url.substring(start, end);
```
gastonfournier, line 289: Duplicated code, can't this be extracted to a function?

gastonfournier (commented):
I think it'd be ideal to log the domain every time the webhook is called, not when created or modified.

src/lib/services/addon-service.ts
sighphyre: But also... why not just?
```
new URL(addonConfig.parameters.url).hostname
```

daveleek: Oh nice, did not even think about that. But obviously this exists 🙄

daveleek:
Closing this and redoing

PR #7959, opened 2024-08-22
chore: remove create feature component

sjaanus:
After we implemented new feature flag creation flow, this are not used anymore.

Creation is now handled by **CreateFeatureDialog**.

Also edit component can be minified, because it does not need so many fields anymore.


frontend/src/component/feature/FeatureForm/EditFeatureForm.tsx
```diff
@@ -106,73 +89,37 @@ const LimitContainer = styled(Box)(({ theme }) => ({
...
-            <ConditionallyRender
-                condition={displayFeatureNamingInfo}
-                show={
-                    <FeatureNamingPatternInfo featureNaming={featureNaming!} />
-                }
-            />
             <StyledInput
                 autoFocus
-                disabled={mode === 'Edit'}
+                disabled={true}
```
sjaanus, line 118: Leaved this as disabled input for now, maybe in future we can rework UX.

thomasheartman (approved):

PR #7962, opened 2024-08-22
feat: log domain every time a webhook is invoked

daveleek:
Log domain when webhook is invoked

### DIFF
--- a/src/lib/addons/webhook.ts
+++ b/src/lib/addons/webhook.ts
@@ -103,15 +103,20 @@ export default class Webhook extends Addon {
             state = 'failed';
             const failedMessage = `Webhook request failed with status code: ${res.status}.`;
             stateDetails.push(failedMessage);
-            if (this.flagResolver.isEnabled('addonUsageMetrics')) {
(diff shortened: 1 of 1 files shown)


sighphyre (approved):
Ye! LGTM

src/lib/addons/webhook.ts
```diff
@@ -103,16 +103,21 @@ export default class Webhook extends Addon {
...
         }
 
+        if (this.flagResolver.isEnabled('addonUsageMetrics')) {
+            this.eventBus.emit(ADDON_EVENTS_HANDLED, {
+                result: state,
+                destination: 'webhook',
+            });
+        }
+
+        const domain = new URL(url).hostname;
```
gastonfournier, line 116: @daveleek can this throw an exception? I'd recommend we try catch but you can quickly tell me if I'm wrong with a test

daveleek: This should occur after the webhook integration has already run, so if for example the protocol is missing, this integration will fail with an error message before we attempt to log the domain:

> Error querying <url-without-protocol> with method POST status code undefined TypeError: Only HTTP(S) protocols are supported

PR #7963, opened 2024-08-22
feat: command menu docs

sjaanus:
Also rebranded command bar to command menu, because that seems more suitable.
Command bar is more like a horizontal/vertical list/bar of icons, like sidebar. Command menu is more of a dropdown.

https://unleash-docs-git-command-docs-unleash-team.vercel.app/reference/command-menu


website/docs/reference/command-menu.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Command Menu
+---
+
+:::note Availability
```
melindafekete, line 5: Just for context: I'd like to begin to standardize how the plan/version availability info looks on all the pages. My dream is to have labels in the future but for now, this is what I've been going for

melindafekete (approved):
Really nice and concise, explains the use cases well

PR #7971, opened 2024-08-22
docs: mention what the minimum values are if you override the limits

thomasheartman:
Adds a note to the resource limit docs about what the minimum values are for some of the core resources if you override the limits.

These resources have a controlled minimum value. The resources that are not listed in the new table, do not. I don't know what happen if you set -42 for segmentValues, for instance. So ... I've left them out 🤷

Happy to take suggestions on how to fix it 😄

Also updates the availability note in preparation for 6.2 GA

### DIFF
--- a/website/docs/reference/resource-limits.mdx
+++ b/website/docs/reference/resource-limits.mdx
@@ -3,8 +3,8 @@ title: Resource Limits
 ---
 import Figure from '@site/src/components/Figure/Figure.tsx'
 
-:::info Availability
-Resource limits are currently in beta for hosted customers.
+:::note Availability
+**Plan**: All plans | **Version**: `6.2+`
 :::
 
 <Figure caption="When you're close to reaching the limit for a resource, you'll see a warning in the Unleash UI. The warning will tell you what the limit is, how close you are to reaching it, and how you can increase the limit." img="/img/project-limit-warning.png" />
--- a/website/docs/reference/resource-limits.mdx
+++ b/website/docs/reference/resource-limits.mdx
@@ -38,19 +38,14 @@ Existing hosted customers who are already above the limits will have their limit
 
 With a few exceptions, all the limits can be overridden through the use of their respective environment variables. However, while it is possible to override these limits, we advise against it in most cases. If you feel that you're unable to do what you need to do because of the limits, we'd love to hear from you.
 
-Some of the limits have a lower threshold. If you try to set them lower than that threshold, Unleash will instead use the minimum value listed below:
+The following resources have a lower limit of 1:
+- feature flags
+- strategies
+- constraint values
+- projects
+- environments
 
-| Resource          | Minimum value |
-|-------------------|--------------:|
-| Feature flags     |             1 |
-| Strategies        |             1 |
-| Constraints       |             0 |
-| Constraint values |             1 |
-| Segments          |             0 |
-| API tokens        |             0 |
-| Projects          |             1 |
-| Environments      |             1 |
-
+If you try to set their limits lower than that, Unleash will automatically adjust them to 1.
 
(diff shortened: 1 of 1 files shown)


website/docs/reference/resource-limits.mdx
```diff
@@ -38,6 +38,20 @@ Existing hosted customers who are already above the limits will have their limit
 
 With a few exceptions, all the limits can be overridden through the use of their respective environment variables. However, while it is possible to override these limits, we advise against it in most cases. If you feel that you're unable to do what you need to do because of the limits, we'd love to hear from you.
 
+Some of the limits have a lower threshold. If you try to set them lower than that threshold, Unleash will instead use the minimum value listed below:
+
+| Resource          | Minimum value |
+|-------------------|--------------:|
+| Feature flags     |             1 |
+| Strategies        |             1 |
+| Constraints       |             0 |
```
melindafekete, line 47: I don't get this one. What is the difference between having no minimum value or having a minimum value of 0? We can't have a negative amount of API keys, so not sure what the implications would be?

thomasheartman: Practically? Probably nothing, but I haven't checked what happens if you try to set segment values limit to -42. 

As for why some numbers are 0 and some are 1 in this table, that's because we decided when we set them up, that you might want to run Unleash without allowing people to use constraints, segments, or API tokens. But if you disallow flags or strategies entirely (i.e. set them to 0), then Unleash can't do anything. Unleash also needs at least 1 project and 1 environment to work. And finally, if you do activate constraints, then it doesn't make sense to not allow any values for them.

thomasheartman: That said, I have a little bit of time now. I'll see what happens

thomasheartman: Yeah, I'm guessing that this is what's gonna happen for all the other resources that don't have set limits:

image

I'd be happy to mention those with a 0 too?

melindafekete: So from a conceptual point of view, wouldn't it make sense to just mention the ones that *do* have a limit of 1? and leave the rest out (if the 0 limit is basically the same in practice as no limit)

So something like:

The following resources have a lower limit of 1:
- feature flags
- strategies
- constraints values
- projects
- environments

If you set any of these values below the minimum, Unleash automatically uses a value of 1 instead.

thomasheartman: Yeah, that's a good way to do it. I'll have a look.

website/docs/reference/resource-limits.mdx
```diff
@@ -3,8 +3,8 @@ title: Resource Limits
 ---
 import Figure from '@site/src/components/Figure/Figure.tsx'
 
-:::info Availability
-Resource limits are currently in beta for hosted customers.
+:::note Availability
+**Plan**: All plans | **Version**: `6.2+`
```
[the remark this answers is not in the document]
thomasheartman: Yeah, I'm down with that. I've only seen the enterprise-only updates you've done, so I didn't know what to do. I don't mind removing it, but does that make it less clear? Because it might be that you think we forgot to specify it? I like being explicit, so I might lean towards specifying it, but no strong feelings either way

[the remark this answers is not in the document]
thomasheartman: Hmm, but then again, that only applies to pages / sections where we talk about availability 🤷🏼  I dunno.

melindafekete (approved):

website/docs/reference/resource-limits.mdx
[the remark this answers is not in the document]
melindafekete: Yes, I think the way we don't mention on, for instance, the feature flags page that it's available for all plans, we don't need to specify it unless there is a restriction. I'll create a page on availability, version, beta, etc. that we can link to and explain in a bit more detail what these concepts mean.

PR #7973, opened 2024-08-23
fix: split features schema into archived and project features

kwasniew:
## About the changes

Removing features-schema that was used in archived features listing and get all features listing.
Created 2 dedicated schemas that match what we actually return in those.

The most important change is the underlying representation of a feature in a list. Neither of those lists (archive and get all project features) has strategies definitions and we want to be explicit about it.

I run both endpoints that were previously using features schema and analyzed fields one by one so you don't have to. The problem with the features-schema was that it was reusing single feature view that has much more information than we provide in the list endpoints.


src/lib/openapi/spec/archived-feature-schema.ts
```diff
@@ -0,0 +1,103 @@
+import type { FromSchema } from 'json-schema-to-ts';
+
+export const archivedFeatureSchema = {
+    $id: '#/components/schemas/archivedFeatureSchema',
+    type: 'object',
+    additionalProperties: true,
```
thomasheartman, line 6: Why have we set this to true? Do we expect to add additional properties? The default for data that we return is to set it to `false` so that we don't accidentally return more than what we say we will.

src/lib/openapi/spec/archived-feature-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+    required: ['name', 'project'],
```
thomasheartman, line 7: Can more of the properties be required? Should they? I'm not sure where we stand here in general, though. But I'm assuming that most (all?) properties on here will always be present, which is why description is `nullable`, for instance?

Marking properties as required also makes them come through as not potentially undefined in the front end API, which is useful for our sake.

src/lib/openapi/spec/archived-feature-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+        impressionData: {
+            type: 'boolean',
+            example: false,
+            description:
+                '`true` if the impression data collection is enabled for the feature, otherwise `false`.',
+        },
+        createdAt: {
+            type: 'string',
+            format: 'date-time',
+            nullable: true,
```
thomasheartman, line 48: Why is this nullable? All flags must have been created, right?

src/lib/openapi/spec/archived-feature-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+            example: '2023-01-28T15:21:39.975Z',
+            description: 'The date the feature was created',
+        },
+        archivedAt: {
+            type: 'string',
+            format: 'date-time',
+            nullable: true,
```
thomasheartman, line 55: If these are archived flags, don't they all have an `archivedAt` property?

src/lib/openapi/spec/archived-feature-schema.ts
```diff
@@ -0,0 +1,103 @@
...
+            deprecated: true,
+            example: '2023-01-28T16:21:39.975Z',
+            description:
+                'The date when metrics where last collected for the feature. This field was deprecated in v5, use the one in featureEnvironmentSchema',
+        },
+        environments: {
+            type: 'array',
+            description:
+                'The list of environments where the feature can be used',
```
thomasheartman, line 71: This is kinda confusing to me. Why is this here and what does it mean? A flag can be used in all environments, right? Is this the environments active in the flag's current project? Do we need this? Can we deprecate it?

src/lib/openapi/spec/project-feature-schema.ts
```diff
@@ -0,0 +1,99 @@
...
+        impressionData: {
+            type: 'boolean',
+            example: false,
+            description:
+                '`true` if the impression data collection is enabled for the feature, otherwise `false`.',
+        },
+        createdAt: {
+            type: 'string',
+            format: 'date-time',
+            nullable: true,
```
thomasheartman, line 61: Same question for this one: why is it nullable?

src/lib/openapi/spec/project-feature-schema.ts
```diff
@@ -0,0 +1,99 @@
...
+            description: 'The date the feature was created',
+        },
+        lastSeenAt: {
+            type: 'string',
+            format: 'date-time',
+            nullable: true,
+            deprecated: true,
+            example: '2023-01-28T16:21:39.975Z',
+            description:
+                'The date when metrics where last collected for the feature. This field was deprecated in v5, use the one in featureEnvironmentSchema',
```
thomasheartman, line 72: What is `featureEnvironmentSchema`? Is it projectFeatureEnvironmentSchema?

Isn't it better to point the user to where they'll find what they're looking for. How's this?

```suggestion
                'The date and time when metrics where last collected for this flag in any environment. This field was deprecated in v5. You should instead use the `lastSeenAt` property on the individual environments listed under the `environments` property.',
```

thomasheartman (approved):
Nice one! A few questions and comments, but the ground work is solid 😄

src/lib/openapi/spec/archived-feature-schema.ts
kwasniew: i was testing sth else and forgot to bring it back to false. Thanks for catching

src/lib/openapi/spec/archived-feature-schema.ts
kwasniew: I made all fields required but then TS checks failed for the types used in the services
```
export interface FeatureToggleDTO {
    name: string;
    description?: string;
    type?: string;
    stale?: boolean;
    archived?: boolean;
    archivedAt?: Date;
    createdAt?: Date;
    impressionData?: boolean;
    variants?: IVariant[];
    tags?: ITag[];
    createdByUserId?: number;
    createdBy?: {
        id: number;
        name: string;
        imageUrl: string;
    };
}
```
By making those optional I can merge my changes (splitting one schema into two schemas to remove fields we never return) and over time we can improve other types. Making it more strict over time is easy since it's not a breaking change.

src/lib/openapi/spec/archived-feature-schema.ts
kwasniew: I can possibly relax the nullable (giving it a try right now) but I can't make it required due to type conflicts with the model types.

src/lib/openapi/spec/archived-feature-schema.ts
kwasniew: Again, I can possibly relax the nullable (giving it a try right now) but I can't make it required due to type conflicts with the model types.

src/lib/openapi/spec/archived-feature-schema.ts
kwasniew: I originally removed it because in the project archived flags we return empty array but then I found that we also have a deprecated endpoint for listing all archived flags and we put 3 date fields there. I kept the name consistent with other environments fields in other schemas. But this field is only here to stay as long as we have a deprecated endpoint.

src/lib/openapi/spec/project-feature-schema.ts
kwasniew: I will try to relax this constraint to see if the tests pass

src/lib/openapi/spec/project-feature-schema.ts
kwasniew: Your suggestion is spot on. Applying it now

PR #7974, opened 2024-08-23
fix: add migration to normalize api token types

thomasheartman:
Adds a migration to normalize api token types already in the database, to remove any weird casing issues as currently demonstrated on sandbox

image

This is a companion piece to #7972


sjaanus (approved):

sjaanus (approved):
LG, if the down migration works .

PR #7984, opened 2024-08-27
feat: align insights charts

sjaanus:
Now the left widget will use the same that as the last data point on the right chart.
With this change, flags/users object never needs to be passed in.

**Next step is to remove the flags/users from the endpoint.**

Previous

image

Now

image

### DIFF
--- a/frontend/src/hooks/api/getters/useInsights/useInsights.ts
+++ b/frontend/src/hooks/api/getters/useInsights/useInsights.ts
@@ -30,7 +30,6 @@ export const useInsights = (
 
     return {
         insights: data || {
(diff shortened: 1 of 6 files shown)


frontend/src/hooks/api/getters/useInsights/useInsights.ts
```diff
@@ -31,7 +31,6 @@ export const useInsights = (
     return {
         insights: data || {
             users: { total: 0, inactive: 0, active: 0 },
-            flags: { total: 0 },
```
sjaanus, line 34: This becomes unused, and we can remove from API.

Tymek (approved):
Should "users" also be removed?

PR #7988, opened 2024-08-27
fix: project last seen at metrics

Tymek:
## About the changes
Read `last_seen_at` from correct table `last_seen_at_metrics`, instead of deprecated `feature.last_seen_at`


sjaanus (approved):
LG

PR #7989, opened 2024-08-27
refactor: make avatar group more customizable

Tymek:


frontend/src/component/common/AvatarGroup/AvatarGroup.tsx
```diff
@@ -34,13 +38,17 @@ type AvatarGroupProps = {
     avatarLimit?: number;
     AvatarComponent?: typeof UserAvatar;
     className?: string;
+    margin?: number;
```
kwasniew, line 41: I'm wondering if having those special case fields is a good way to go. Usually we compose new styling on top of the lower level elements. Could we make this component styleable with styled components/sx/styled(Avatar)?

kwasniew (approved):
Non blocking approve but please consider my comment

Tymek (commented):
Inverted how AvatarComponent is created

PR #8010, opened 2024-08-29
feat: start populating user first seen column

sjaanus:


src/lib/db/user-store.ts
```diff
@@ -69,9 +71,26 @@ class UserStore implements IUserStore {
...
+
+    private EXPERIMENTAL_USER_COLUMNS_PUBLIC = USER_COLUMNS_PUBLIC;
+    private EXPERIMENTAL_USER_COLUMNS = USER_COLUMNS;
+
+    constructor(db: Db, getLogger: LogProvider, flagResolver: IFlagResolver) {
         this.db = db;
         this.logger = getLogger('user-store.ts');
+        this.flagResolver = flagResolver;
+
+        if (this.flagResolver.isEnabled('onboardingMetrics')) {
```
sjaanus, line 84: The store is selecting columns in so many multiple places, so this made most sense.

kwasniew: this may not work at runtime for non transactional variant where we new up the store on startup only. However maybe it's ok here.

kwasniew (approved):

PR #8014, opened 2024-08-29
feat: projects onboarding metrics

kwasniew:
## About the changes

Project onboarding metrics calculated from the project created time

### DIFF
--- a/src/lib/features/onboarding/onboarding-read-model-type.ts
+++ b/src/lib/features/onboarding/onboarding-read-model-type.ts
@@ -9,6 +9,17 @@ export type InstanceOnboarding = {
     firstLive: number | null;
 };
 
+/**
+ * All the values are in minutes
+ */
+export type ProjectOnboarding = {
+    project: string;
+    firstFeatureFlag: number | null;
+    firstPreLive: number | null;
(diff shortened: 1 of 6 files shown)


src/lib/features/onboarding/onboarding-read-model-type.ts
```diff
@@ -9,6 +9,17 @@ export type InstanceOnboarding = {
...
+export type ProjectOnboarding = {
+    project: string;
+    firstFeatureFlag: number | null;
+    firstPreLive: number | null;
+    firstLive: number | null;
+};
+
 export interface IOnboardingReadModel {
     getInstanceOnboardingMetrics(): Promise<InstanceOnboarding>;
+    getProjectsOnboardingMetrics(): Promise<Array<ProjectOnboarding>>;
```
sjaanus, line 24: Why Array instead of [], what we usually have.

sjaanus (approved):

src/lib/features/onboarding/onboarding-read-model-type.ts
kwasniew: I use both. I can fix in the next PR

PR #8017, opened 2024-08-29
chore: strip +main from package version

gastonfournier:
This is not needed anymore as we are not publishing to npm from main.


daveleek (approved):
Great! :shipit:

PR #8020, opened 2024-08-30
feat: collect onboarding events in separate table

sjaanus:


src/migrations/20240830102144-onboarding-events.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+            CREATE TABLE onboarding_events (
+               event VARCHAR(50) NOT NULL,
+               diff INTEGER NOT NULL,
```
sjaanus, line 8: In seconds, rounded, can not be empty

src/migrations/20240830102144-onboarding-events.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+            CREATE TABLE onboarding_events (
+               event VARCHAR(50) NOT NULL,
```
sjaanus, line 7: Can not be empty, 50 seems to be good enough. Do not need long definition

src/migrations/20240830102144-onboarding-events.js
```diff
@@ -0,0 +1,23 @@
...
+               diff INTEGER NOT NULL,
+               project VARCHAR(255),
```
sjaanus, line 9: Project definition taken from project table. Same as project id. 
This can be null, because project can be deleted and we need the data to stay forever.

src/migrations/20240830102144-onboarding-events.js
```diff
@@ -0,0 +1,23 @@
...
+               PRIMARY KEY (event, project)
```
sjaanus, line 10: Composite primary key to avoid duplications.

src/migrations/20240830102144-onboarding-events.js
kwasniew: events have 255 and it feels like a standard default. 50 feels weird

src/migrations/20240830102144-onboarding-events.js
kwasniew: is it diff from project created or first user created?

src/migrations/20240830102144-onboarding-events.js
```diff
@@ -0,0 +1,30 @@
...
+            CREATE TABLE onboarding_events_instance (
+                event VARCHAR(255) NOT NULL,
+                time_to_event INTEGER NOT NULL,
+                PRIMARY KEY (event)
+            );
+
+            CREATE TABLE onboarding_events_project (
+               event VARCHAR(255) NOT NULL,
+               time_to_event INTEGER NOT NULL,
+               project VARCHAR(255) REFERENCES projects(id) ON DELETE CASCADE,
```
sjaanus, line 15: Now we can have foreign key and cascade delete, because we do not need to keep old data.

src/migrations/20240830102144-onboarding-events.js
```diff
@@ -0,0 +1,30 @@
...
+            CREATE TABLE onboarding_events_instance (
+                event VARCHAR(255) NOT NULL,
+                time_to_event INTEGER NOT NULL,
+                PRIMARY KEY (event)
+            );
+
+            CREATE TABLE onboarding_events_project (
+               event VARCHAR(255) NOT NULL,
+               time_to_event INTEGER NOT NULL,
+               project VARCHAR(255) NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
```
sjaanus, line 15: Now project is not nullable, has foreign key and deletes on cascade, because we not need to keep it around anymore.

src/migrations/20240830102144-onboarding-events.js
```diff
@@ -0,0 +1,30 @@
...
+        `
+            CREATE TABLE onboarding_events_instance (
+                event VARCHAR(255) NOT NULL,
+                time_to_event INTEGER NOT NULL,
+                PRIMARY KEY (event)
+            );
+
+            CREATE TABLE onboarding_events_project (
+               event VARCHAR(255) NOT NULL,
+               time_to_event INTEGER NOT NULL,
```
kwasniew, line 14: can we add comments about unit?

src/migrations/20240830102144-onboarding-events.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+            CREATE TABLE onboarding_events_instance (
```
kwasniew, line 6: CREATE TABLE IF NOT EXISTS?

kwasniew (approved):

PR #8022, opened 2024-08-30
feat: display the build number as extra data

gastonfournier:
## About the changes
Show the build number between parenthesis separated from the version.

When there's no build number:
image

When there is:
image


nunogois (approved):
I forgot we also had ` ({environment})` there so it doesn't look as nice as I hoped. Do we still need that?

Maybe UX can help figure out the ideal way of displaying this information.

gastonfournier:
> I forgot we also had ` ({environment})` there so it doesn't look as nice as I hoped. Do we still need that?
> 
> Maybe UX can help figure out the ideal way of displaying this information.

I also thought we might change that later, but I rather do this now and another improvement later. Despite that let me send this over to UX

gastonfournier:
Failing e2e cypress test don't seem related but they fail consistently. I'll merge without them

PR #8023, opened 2024-08-30
feat: project archive sort

Tymek:
## About the changes
Show newly archived projects first


kwasniew (commented):
The conventional commit says refactor. I see we're adding new production code. Is the conventional commit incorrect? If this is a new feature where can I test it? What flag do I need?

PR #8024, opened 2024-08-30
fix: project archive card

Tymek:
## About the changes
Fix archived project card after new projects list refactor.

Archive card was not showing button actions in footer.
image

After change:
image


kwasniew (changes requested):
Please add description  + screenshot what was broken, what is being fixed. It's useful when coming back a few months later and trying to understand a change.

PR #8025, opened 2024-08-30
ADR: Use of conditionals in JSX (`</ConditionallyRender>`)

Tymek:
Let's deprecate `<ConditionallyRender />`

### DIFF
--- a/website/docs/contributing/ADRs/front-end/jsx-conditionals.md
+++ b/website/docs/contributing/ADRs/front-end/jsx-conditionals.md
@@ -13,7 +13,7 @@ While most truthy and falsy values behave as expected with the `&&` operator, ce
 ```tsx
 {NaN && <p>❔</p>} // will render `NaN`
 {0 && <p>❔</p>} // will render `0`
-{arr?.length && <p>❔</p>} // can render `0` ❗
+{arr?.length && <p>❔</p>} // can render `0`
 ```
 
 These issues can cause bugs in components that conditionally render UI elements based on numeric values or other potentially falsy conditions. For this reason we use a wrapper.
@@ -88,16 +88,35 @@ This approach explicitly converts the condition to a boolean value using `Boolea
 The ternary operator is a more explicit and safer approach. This will cover some cases where we have to return `null` or `undefined`.
 
 ``` tsx
-{NaN ? <p>👍</p> : null}  // ✅ Won't render anything
+{NaN ? <p>👍</p> : null}  // Won't render anything
 ```
 
 It also plays nicely with TypeScript.
 
 ```tsx
 export const Test: FC<{ maybeString?: string }> = ({ maybeString }) =>
-    maybeString ? <SubComponent text={maybeString} /> : null; // ✅
+    maybeString ? <SubComponent text={maybeString} /> : null;
 ```
-
 
 This is what we will use from now onwards.
 
+## Consequences
+Positive: The codebase will become more type-safe and easier to understand.
+
+Negative: We import `<ConditionallyRender />` in almost 400 files. Significant refactoring effort is required.
+
+## Migration plan
+
+1. Mark <ConditionallyRender /> as deprecated in the codebase with a clear JSDoc comment.
+
+2. Automated refactoring with AST (Abstract Syntax Tree)
+There already is a script developed that can convert files between `ConditionallyRender` and ternary syntax. It is using jscodeshift, an  library. It will be put in `frontend/scripts/transform.js`.
+
+3. Each change will have to be reviewed. The order of refactoring should be:
+    1. New features that are behind feature flags.
+    2. Non-critical or not in demand pages, like new signals or feedback component.
+    3. Less complex pages, for example in `/src/component/admin`.
+    4. More complex and critical pages, like strategy editing.
+    5. Utilities and components used in many places (`/src/component/common`).
+
+3. Once all instances of <ConditionallyRender /> have been refactored, remove the component from the codebase.
--- a/website/docs/contributing/ADRs/front-end/jsx-conditionals.md
+++ b/website/docs/contributing/ADRs/front-end/jsx-conditionals.md
@@ -4,7 +4,7 @@ title: "ADR: Deprecation of `<ConditionallyRender />` Component"
 
 ## Background
 
-Using the `&&` operator in React can lead to unexpected rendering behavior when dealing with certain falsy values. In our codebase, the `<ConditionallyRender />` component has been used to render React elements based on a boolean condition. It has some certain drawbacks, and that's why we would like to replace it with ternary operator.
+Using the `&&` operator in React can lead to unexpected rendering behavior when dealing with certain falsy values. In our codebase, the `<ConditionallyRender />` component has been used to render React elements based on a boolean condition. It has certain drawbacks. This is why we would like to replace it with ternary operator.
 
 ### Pitfalls of `&&` operator
 
@@ -50,7 +50,7 @@ export const Test: FC<{ maybeString?: string }> = ({ maybeString }) => (
 ```
 
 #### Obfuscation of code smells and code cruft
-It's easier to spot nested ternaries, then nested `<ConditionallyRender />` elements.
+Nested ternaries are easier to spot then nested `<ConditionallyRender />` elements.
 
 ```tsx
 <div>
@@ -72,17 +72,17 @@ It's easier to spot nested ternaries, then nested `<ConditionallyRender />` elem
 To avoid these issues, safer alternatives to the `&&` operator can be used:
 
 ### **Convert to boolean**
-This approach explicitly converts the condition to a boolean value using `Boolean(expression)` or `!!expression`, ensuring that only `true` or `false` determine the rendering.
+We could try to explicitly converts the condition to a boolean, ensuring that only `true` or `false` determine the rendering.
 
 ```tsx
 {Boolean(NaN) && <p>❔</p>}  // Won't render anything
 {!!0 && <p>❔</p>}           // Also safe
 ```
 
-❌ **Unfortunately** Biome (linter we use) does not include rules to automatically enforce safer usage of the `&&` operator, as ESLint did.
+**Unfortunately** Biome (linter we use) does not include rules to automatically enforce safer usage of the `&&` operator, as ESLint did.
 
 ### Ternary Operator
-The ternary operator is a more explicit and safer approach. This will cover some cases where we have to return `null` or `undefined`.
+The ternary operator is a more explicit and safer approach. This covers cases where we need to return `null` or `undefined`.
 
 ``` tsx
 {NaN ? <p>👍</p> : null}  // Won't render anything
@@ -100,16 +100,16 @@ This is what we will use from now onwards.
 ## Consequences
 Positive: The codebase will become more type-safe and easier to understand.
 
-Negative: We import `<ConditionallyRender />` in almost 400 files. Significant refactoring effort is required.
+Negative: The `<ConditionallyRender />` component is imported in nearly 400 files. Significant refactoring effort is required.
 
 Performance: There was no measurable performance difference between code with and without this component. This was tested on production bundle, on the features search (table) and projects list pages.
 
 ## Migration plan
 
-1. Mark <ConditionallyRender /> as deprecated in the codebase with a clear JSDoc comment.
+1. Mark `<ConditionallyRender />` as deprecated in the codebase with a clear JSDoc comment
 
 2. Automated refactoring with AST (Abstract Syntax Tree)
-There already is a script developed that can convert files between `ConditionallyRender` and ternary syntax. It is using jscodeshift, an  library. It will be put in `frontend/scripts/transform.js`.
+There already is a script developed that can convert files between `ConditionallyRender` and ternary syntax. It uses jscodeshift, an  library. It will be put in `frontend/scripts/transform.js`.
 
 3. Each change will have to be reviewed. The order of refactoring should be:
     1. New features that are behind feature flags.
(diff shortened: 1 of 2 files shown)


website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+
+        show={<SubComponent text={maybeString} />}
+        // ❌ TS Error: Type 'string | undefined' is not assignable to type 'string'
+        // you have to use `maybeString!` here
+    />
+);
+```
+
+#### Obfuscation of code smells and code cruft
+Nested ternaries are easier to spot then nested `<ConditionallyRender />` elements.
```
melindafekete, line 53:
```suggestion
Nested ternaries are easier to spot than nested `<ConditionallyRender />` elements.
```

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+
+### **Convert to boolean**
+We could try to explicitly converts the condition to a boolean, ensuring that only `true` or `false` determine the rendering.
+
+```tsx
+{Boolean(NaN) && <p>❔</p>}  // Won't render anything
+{!!0 && <p>❔</p>}           // Also safe
+```
+
+**Unfortunately** Biome (linter we use) does not include rules to automatically enforce safer usage of the `&&` operator, as ESLint did.
```
melindafekete, line 82:
```suggestion
**Unfortunately** Biome (the linter we use) does not include rules to automatically enforce a safer usage of the `&&` operator, as ESLint did.
```

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
+---
+title: "ADR: Use of conditionals in JSX (deprecation of `<ConditionallyRender />`)"
+---
+
+## Background
+
+Using the `&&` operator in React can lead to unexpected rendering behavior when dealing with certain falsy values. In our codebase, the `<ConditionallyRender />` component has been used to render React elements based on a boolean condition. It has certain drawbacks. This is why we would like to replace it with ternary operator.
```
melindafekete, line 7:
```suggestion
Using the `&&` operator in React can lead to unexpected rendering behavior when dealing with certain falsy values. In our codebase, the `<ConditionallyRender />` component has been used to render React elements based on a boolean condition. However, it has certain drawbacks, which is why we would like to replace it with the ternary operator.
```

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+    />
+</div>
+```
+
+## Options considered
+
+To avoid these issues, safer alternatives to the `&&` operator can be used:
+
+### **Convert to boolean**
+We could try to explicitly converts the condition to a boolean, ensuring that only `true` or `false` determine the rendering.
```
melindafekete, line 75:
```suggestion
We could try to explicitly convert the condition to a boolean, ensuring that only `true` or `false` determine the rendering.
```

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+Negative: The `<ConditionallyRender />` component is imported in nearly 400 files. Significant refactoring effort is required.
+
+Performance: There was no measurable performance difference between code with and without this component. This was tested on production bundle, on the features search (table) and projects list pages.
+
+## Migration plan
+
+1. Mark `<ConditionallyRender />` as deprecated in the codebase with a clear JSDoc comment
+
+2. Automated refactoring with AST (Abstract Syntax Tree)
+There already is a script developed that can convert files between `ConditionallyRender` and ternary syntax. It uses jscodeshift, an  library. It will be put in `frontend/scripts/transform.js`.
```
melindafekete, line 112:
```suggestion
There already is a script developed that can convert files between `ConditionallyRender` and ternary syntax. It uses jscodeshift, a library. It will be put in `frontend/scripts/transform.js`.
```

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+
+While most truthy and falsy values behave as expected with the `&&` operator, certain falsy values can produce unintended outcomes:
+
+```tsx
+{NaN && <p>❔</p>} // will render `NaN`
+{0 && <p>❔</p>} // will render `0`
+{arr?.length && <p>❔</p>} // can render `0`
+```
+
+These issues can cause bugs in components that conditionally render UI elements based on numeric values or other potentially falsy conditions. For this reason we use a wrapper.
```
melindafekete, line 19:
```suggestion
These issues can cause bugs in components that conditionally render UI elements based on numeric values or other potentially falsy conditions. For this reason, we use a wrapper.
```

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+## Consequences
+Positive: The codebase will become more type-safe and easier to understand.
+
+Negative: The `<ConditionallyRender />` component is imported in nearly 400 files. Significant refactoring effort is required.
+
+Performance: There was no measurable performance difference between code with and without this component. This was tested on production bundle, on the features search (table) and projects list pages.
+
+## Migration plan
+
+1. Mark `<ConditionallyRender />` as deprecated in the codebase with a clear JSDoc comment
```
melindafekete, line 109:
```suggestion
1. Mark `<ConditionallyRender />` as deprecated in the codebase with a clear JSDoc comment.
```

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+
+2. Automated refactoring with AST (Abstract Syntax Tree)
```
melindafekete, line 111:
```suggestion
2. Automated refactoring with AST (Abstract Syntax Tree):
```

melindafekete (approved):

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+        show={(
+            <ConditionallyRender
+                condition={Boolean(b)}
+                show={<p>This is bad</p>}
+            />
+        )}
+        elseShow={'Should be refactored'}
+    />
+</div>
+```
```
nunogois, line 68: Would have been nice to see the suggested approach instead, so we could compare:

```tsx
<div>
    {Boolean(a) ? (
        Boolean(b) && <p>This is bad</p>
    ) : 'Should be refactored'}
</div>
```

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
```diff
@@ -0,0 +1,121 @@
...
+There already is a script developed that can convert files between `ConditionallyRender` and ternary syntax. It uses jscodeshift, a library. It will be put in `frontend/scripts/transform.js`.
+
+3. Each change will have to be reviewed. The order of refactoring should be:
+    1. New features that are behind feature flags.
+    2. Non-critical or not in demand pages, like new signals or feedback component.
+    3. Less complex pages, for example in `/src/component/admin`.
+    4. More complex and critical pages, like strategy editing.
+    5. Utilities and components used in many places (`/src/component/common`).
+
+3. Once all instances of `<ConditionallyRender />` have been refactored, remove the component from the codebase.
```
nunogois, line 121: Should be item `4` instead.

nunogois (commented):
Sure, I'm on board, but I can't help but wonder if we could improve the ConditionallyRender component instead.

sighphyre:
> Sure, I'm on board, but I can't help but wonder if we could improve the ConditionallyRender component instead.

Plus one. I don't think that's sufficiently answered as to why this isn't an option in this proposal. This is a big and dangerous change with a lot of scope to go wrong in subtle, difficult to notice ways. Why isn't just improving the ConditionallyRender component an option? What's stopping that?

sighphyre:
> > Sure, I'm on board, but I can't help but wonder if we could improve the ConditionallyRender component instead.
> 
> Plus one. I don't think that's sufficiently answered as to why this isn't an option in this proposal. This is a big and dangerous change with a lot of scope to go wrong in subtle, difficult to notice ways. Why isn't just improving the ConditionallyRender component an option? What's stopping that?

Okay, we spoke. I think I have a better sense of the problem. The issue is that the type information is split over both the `show` and the `condition` property. Because those properties are handled on the top layer of conditionally render that's already a problem in our codebase so changes to the internals of conditionally render don't help us with the type ergonomics

Tough problem, I see why we want to take this approach. Lotta risk here but might be worthwhile

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
Tymek: https://github.com/Unleash/unleash/pull/8063

website/docs/contributing/ADRs/front-end/jsx-conditionals.md
Tymek: https://github.com/Unleash/unleash/pull/8063

nunogois:
> > > Sure, I'm on board, but I can't help but wonder if we could improve the ConditionallyRender component instead.
> > 
> > 
> > Plus one. I don't think that's sufficiently answered as to why this isn't an option in this proposal. This is a big and dangerous change with a lot of scope to go wrong in subtle, difficult to notice ways. Why isn't just improving the ConditionallyRender component an option? What's stopping that?
> 
> Okay, we spoke. I think I have a better sense of the problem. The issue is that the type information is split over both the `show` and the `condition` property. Because those properties are handled on the top layer of conditionally render that's already a problem in our codebase so changes to the internals of conditionally render don't help us with the type ergonomics
> 
> Tough problem, I see why we want to take this approach. Lotta risk here but might be worthwhile

```tsx
interface IRenderIfProps<T> {
    condition: T | undefined;
    children: (value: T) => JSX.Element;
    elseShow?: ReactNode;
}

export const RenderIf = <T,>({
    condition,
    children,
    elseShow = null,
}: IRenderIfProps<T>): JSX.Element | null => {
    if (condition) {
        return children(condition);
    }
    return elseShow ? <>{elseShow}</> : null;
};

<RenderIf condition={maybeValue}>
    {(value) => <span>{value}</span>}
</RenderIf>
```

`value` will have the defined type of `maybeValue`. And we could always ignore it in cases where we don't care about it.

Tymek:
@nunogois You'll need to use `T | undefined | null | 0 | false`. Plus it puts `elseShow` before `show`, and this is even less readable then previous version. It's a step in a right direction, but I'd like to go directly to what is in https://react.dev/learn/conditional-rendering, without half-steps.

nunogois:
> @nunogois You'll need to use `T | undefined | null | 0 | false`.

Why is that?

Here's what (it feels like) a drop-in replacement, where existing cases seem to work fine:

```tsx
import type { ReactNode } from 'react';

type ShowFunction<T> = (condition: T) => ReactNode;
type ElseShowFunction = () => ReactNode;

interface IConditionallyRenderProps<T> {
    condition: T | undefined;
    show: ReactNode | ShowFunction<T>;
    elseShow?: ReactNode | ElseShowFunction;
}

const isFunction = (maybeFunction: ReactNode | Function) =>
    typeof maybeFunction === 'function';

export const ConditionallyRender = <T,>({
    condition,
    show,
    elseShow = null,
}: IConditionallyRenderProps<T>): ReactNode | null => {
    if (condition) {
        if (isFunction(show)) {
            return show(condition);
        }
        return show;
    }
    if (elseShow) {
        if (isFunction(elseShow)) {
            return elseShow();
        }
        return elseShow;
    }
    return null;
};
```

Tymek:
> Why is that?
image

It doesn't solve issues with TS. I tried more advanced generics, but that is just crashing TS language server.

Tymek:
My attempt:
```
type RenderFunc = () => JSX.Element;

type TargetElement =
    | JSX.Element
    | JSX.Element[]
    | RenderFunc
    | ReactNode
    | null;

type RunIfFunction<T extends TargetElement> = T extends RenderFunc
    ? ReturnType<T> | null
    : T;

type ConditionallyRenderType = <
    C extends boolean,
    A extends TargetElement,
    B extends TargetElement,
>({
    condition,
    show,
    elseShow,
}: {
    condition: C;
    show: A;
    elseShow?: B;
}) => RunIfFunction<C extends true ? A : B>;
```

nunogois:
> > Why is that?
> > image
> 
> It doesn't solve issues with TS. I tried more advanced generics, but that is just crashing TS language server.

Hi @Tymek,

As I shared on my example, the code on your screenshot should look like this instead:

```tsx
<ConditionallyRender
    condition={pwDialog.user}
    show={(user) => (
        <ChangePassword
            showDialog={pwDialog.open}
            closeDialog={closePwDialog}
            user={user}
        />
    )}
/>
```

Can you share an example of a more advanced generic that crash the TS language server?

nunogois:
> O, with your approach, with "show(input)" it is adding another way of doing things, and even more abstraction and options. Why write in 10 lines something that can be 7 and less indented?
> 
> ```tsx
> {pwDialog.user ? (
>     <ChangePassword
>         showDialog={pwDialog.open}
>         closeDialog={closePwDialog}
>         user={user}
>     />
> ) : null}
> ```

I think that's what we should discuss as team. I'm not sure everyone agrees that the ternaries are immediately more readable, especially since we're already familiar with the ConditionallyRender component. And there's also the argument that removing the ConditionallyRenders can have some risk.

I'm personally okay with either approach. This is just showing that there's a chance of improving ConditionallyRender, even if we end up not liking it.

Like I mentioned above:

> PS: I'm not saying this covers all the arguments from the ADR, it's just a quick example of a possible path forward where we slightly improve the current ConditionallyRender component by making it easier to use instead of dropping it altogether.

PR #8027, opened 2024-08-30
feat: onboarding store

kwasniew:
## About the changes

* adding onboarding service with SQL store implementation
* added project read model method and user store method (can extract to read model in a separate PR)
* fixed project create method to allow for date control from the code

Next PR:
* create tests that combine store and read model
* plug the service into the rest of the code
* create composition root for the service
* write service tests with in-memory store for the listen method testing

### Important files

Naming convention:
* when events progress through the system they change meaning. e.g. lifecycle `initial` event arrived from the lifecycle subdomain. It is translated to `flag-created` in the onboarding subdomain. One we confirm it's the first flag created we store it as `first-flag` in the DB.

### DIFF
--- a/src/lib/features/onboarding/onboarding-service.ts
+++ b/src/lib/features/onboarding/onboarding-service.ts
@@ -1,0 +1,63 @@
+import type { IFlagResolver, IUnleashConfig } from '../../types';
+import type EventEmitter from 'events';
+import type { Logger } from '../../logger';
+import { STAGE_ENTERED, USER_LOGIN } from '../../metric-events';
+import type { NewStage } from '../feature-lifecycle/feature-lifecycle-store-type';
+import type { IOnboardingStore } from './onboarding-store-type';
+
+export class OnboardingService {
+    private flagResolver: IFlagResolver;
+
+    private eventBus: EventEmitter;
+
+    private logger: Logger;
+
+    private onboardingStore: IOnboardingStore;
+
+    constructor(
+        { onboardingStore }: { onboardingStore: IOnboardingStore },
+        {
+            flagResolver,
+            eventBus,
+            getLogger,
+        }: Pick<IUnleashConfig, 'flagResolver' | 'eventBus' | 'getLogger'>,
+    ) {
+        this.onboardingStore = onboardingStore;
+        this.flagResolver = flagResolver;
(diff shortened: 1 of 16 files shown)


src/lib/features/onboarding/onboarding-service.ts
```diff
@@ -0,0 +1,63 @@
...
+                await this.onboardingStore.insert({ type: 'firstUserLogin' });
+            }
+            if (event.loginOrder === 1) {
+                await this.onboardingStore.insert({ type: 'secondUserLogin' });
+            }
+        });
+        this.eventBus.on(STAGE_ENTERED, async (stage: NewStage) => {
+            if (!this.flagResolver.isEnabled('onboardingMetrics')) return;
+
+            if (stage.stage === 'initial') {
```
kwasniew, line 45: we do this translation since there's not 100% domain vocab match between lifecycle and onboarding. lifecycle cares about initial flag, while onboarding cares about the first flag

src/lib/features/onboarding/onboarding-store.ts
```diff
@@ -0,0 +1,136 @@
+import type { Db } from '../../db/db';
+import type {
+    IOnboardingStore,
+    OnboardingEvent,
+    SharedEvent,
+} from './onboarding-store-type';
+import { millisecondsToSeconds } from 'date-fns';
+
+type DBInstanceType = {
```
kwasniew, line 9: sticking to dashed convention as in lifecycle events instead of camel casing

src/lib/features/onboarding/onboarding-store.ts
```diff
@@ -0,0 +1,136 @@
...
+    event: 'firstFlag' | 'firstPreLive' | 'firstLive';
+    time_to_event: number;
+    project: string;
+};
+
+const translateEvent = (
+    event: OnboardingEvent['type'],
+): DBInstanceType['event'] => {
+    if (event === 'flag-created') {
+        return 'first-flag';
```
kwasniew, line 29: can be also first-flag-created.
Overall progression: 'initial' in lifecycle -> 'flag-created' in service -> 'first-flag-created' when we confirm it's the first one

sjaanus (approved):
LG

PR #8030, opened 2024-09-02
chore: reorganize product data in footer

gastonfournier:
## About the changes
Instead of having the plan at the end, move it next to the product name and display the build number at last

### DIFF
--- a/frontend/src/component/menu/Footer/ApiDetails/ApiDetails.tsx
+++ b/frontend/src/component/menu/Footer/ApiDetails/ApiDetails.tsx
@@ -17,15 +17,11 @@ export const ApiDetails = (props: IApiDetailsProps): ReactElement => {
     const environment = props.uiConfig.environment;
     const updateNotification = formatUpdateNotification(props.uiConfig);
 
-    const buildInfo = buildNumber ? <small> ({buildNumber})</small> : '';
+    const buildInfo = buildNumber ? <small>({buildNumber})</small> : '';
     return (
         <section title='API details'>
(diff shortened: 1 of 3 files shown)


frontend/src/component/menu/Footer/ApiDetails/ApiDetails.tsx
```diff
@@ -17,15 +17,11 @@ export const ApiDetails = (props: IApiDetailsProps): ReactElement => {
...
+    const buildInfo = buildNumber ? <small>({buildNumber})</small> : '';
     return (
         <section title='API details'>
             <FooterTitle>
-                {name} {version} {buildInfo}
-                <ConditionallyRender
-                    condition={Boolean(environment)}
-                    show={<small> ({environment})</small>}
-                />
+                {name} {environment ? environment : ''} {version} {buildInfo}
```
nunogois, line 24: Maybe if environment is empty we want to show OSS as the default? 🤔

nunogois (approved):

frontend/src/component/menu/Footer/ApiDetails/ApiDetails.tsx
gastonfournier: The default environment is [Open Source](https://github.com/Unleash/unleash/blob/1bc0f97101b84dca3f0c56eab038dd627495a8f4/src/lib/create-config.ts#L175) I think we have a type issue but it's probably like this because of backward compatibility reasons.

PR #8032, opened 2024-09-02
task: remove upgrade edge banner.

chriswk:
All customers but one are now upgraded, and this banner no longer adds value, we also have the outdated SDK banner, which can fill in if needed.


sighphyre (approved):
LG!

PR #8034, opened 2024-09-02
feat: onboarding table to prometheus

sjaanus:
Instead of lifecycle table, we take the metrics directly from onboarding tables.

### DIFF
--- a/src/lib/features/onboarding/onboarding-read-model.ts
+++ b/src/lib/features/onboarding/onboarding-read-model.ts
@@ -4,14 +4,19 @@ import type {
     InstanceOnboarding,
     ProjectOnboarding,
 } from './onboarding-read-model-type';
-import { millisecondsToMinutes } from 'date-fns';
 
-const calculateTimeDifferenceInMinutes = (date1?: Date, date2?: Date) => {
-    if (date1 && date2) {
-        const diffInMilliseconds = date2.getTime() - date1.getTime();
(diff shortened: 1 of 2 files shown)


src/lib/features/onboarding/onboarding-read-model.ts
```diff
@@ -4,14 +4,19 @@ import type {
...
 } from './onboarding-read-model-type';
-import { millisecondsToMinutes } from 'date-fns';
 
-const calculateTimeDifferenceInMinutes = (date1?: Date, date2?: Date) => {
-    if (date1 && date2) {
-        const diffInMilliseconds = date2.getTime() - date1.getTime();
-        return millisecondsToMinutes(diffInMilliseconds);
-    }
-    return null;
+const instanceEventLookup = {
```
sjaanus, line 8: Need mapping for new column format

kwasniew: maybe we can use the - notation for the metrics so that we can skip mapping?

kwasniew: did we add camel case to be consistent with other metrics?

kwasniew (approved):

src/lib/features/onboarding/onboarding-read-model.ts
sjaanus: I just dont want to mess with json object with - in the key.

PR #8048, opened 2024-09-02
Upgrade millify

gastonfournier:

PR #8040, opened 2024-09-02
Add debug

gastonfournier:
## About the changes

Closes #

PR #8049, opened 2024-09-02
fix: force languages in cypress browser for tests

gastonfournier:
To fix this we had to create a free trial account on cypress and enable the recording of the test. That way we found out the issue was with a locale:
image (38)

Probably, this works well locally because our local machines do have a default locale, but probably we don't have one when running in CI, and millify library is causing the tests to fail specifically at this line: https://github.com/Unleash/unleash/blob/363911c4a1ce94dd467a950ec84ccdc9a3ec7d75/frontend/src/component/common/AvatarGroup/AvatarGroup.tsx#L89 (validated [here](https://github.com/Unleash/unleash/pull/8040/files#diff-afc857890da2221bd34feed0ff45dd7745ff32fb0b27055214cbe69896d5311dL89)).

Unfortunately, upgrading millify didn't help, but downgrading to v5 (which doesn't support locales), solve the issue at the cost of not having the up-to-date library: https://github.com/Unleash/unleash/pull/8048

I believe the issue is related to this locale `c` reported here: https://github.com/cypress-io/cypress/issues/7890#issuecomment-2105991364 because only after overriding the languages this worked


nunogois (approved):
LGTM

chriswk:
Ok, not trying to sound aggressive, but why are we even using millify here? Is it that important to get thousand separators?

gastonfournier:
> Ok, not trying to sound aggressive, but why are we even using millify here? Is it that important to get thousand separators?

We're only using it in a few places and we're consciously adding it for things like: https://github.com/Unleash/unleash/pull/7879

PR #8055, opened 2024-09-03
fix: reduce integration events requests

nunogois:
Cuts the total amount of integration event requests in half when browsing the integrations page, by only fetching the 20 latest events for each configured integration when it's actually needed (open modal).


daveleek (approved):
LGTM!

PR #8057, opened 2024-09-03
fix: onboarding events corner cases

kwasniew:
## About the changes

Fixing two corner cases for onboarding metrics:
* for the default project we don't take project creation date but instead we take first user creation date
* for the instance metrics we need to exclude system users who are created upfront


src/lib/db/user-store.ts
```diff
@@ -302,6 +302,7 @@ class UserStore implements IUserStore {
     async getFirstUserDate(): Promise<Date | null> {
         const firstInstanceUser = await this.db('users')
             .select('created_at')
+            .where('is_system', '=', false)
```
sjaanus, line 305: Nice

sjaanus (approved):

PR #8058, opened 2024-09-03
feat: start returning onboarding status with project overview

sjaanus:
To show/hide onboarding flow, we need to get extra info about onboarding status. This PR adds it to project overview.

### DIFF
--- a/src/lib/openapi/spec/project-overview-schema.ts
+++ b/src/lib/openapi/spec/project-overview-schema.ts
@@ -19,7 +19,7 @@ export const projectOverviewSchema = {
     $id: '#/components/schemas/projectOverviewSchema',
     type: 'object',
     additionalProperties: false,
-    required: ['version', 'name'],
+    required: ['version', 'name', 'onboardingStatus'],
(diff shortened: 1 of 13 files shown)


src/lib/openapi/spec/project-overview-schema.ts
```diff
@@ -135,6 +135,41 @@ export const projectOverviewSchema = {
...
         },
+        onboardingStatus: {
+            type: 'object',
+            oneOf: [
+                {
+                    type: 'object',
+                    properties: {
+                        status: {
+                            type: 'string',
+                            enum: ['onboarding-started', 'onboarded'],
```
kwasniew, line 146: wondering how orval will handle it :)

kwasniew (approved):

src/lib/openapi/spec/project-overview-schema.ts
sjaanus: We have similar for lifecycle schemas, should not be too big of an issue. Hopefully

PR #8060, opened 2024-09-03
fix: deps revert for chartjs plugin

kwasniew:
## About the changes

Revert https://github.com/Unleash/unleash/commit/a28d9651768c91fcb23970bc96c996dbbbaa3137


sjaanus (approved):

ivarconr (approved):

gastonfournier (approved):
Can we pin the dependency so it's not updated?

PR #8066, opened 2024-09-03
feat: can select client and frontend sdk

kwasniew:
## About the changes

Can select SDK and SDK type
Screenshot 2024-09-03 at 17 37 09

Details:
* can navigate between first two stages including going back to stage 1 with SDK selection
* each stage selects either SDK name and type of the environment
* once you select one env it stays persistent even when you go back and forth

Next PR:
* add icons for SDKs
* add tests when the whole feature is complete

### DIFF
--- a/frontend/src/component/onboarding/ConnectSdkDialog.tsx
+++ b/frontend/src/component/onboarding/ConnectSdkDialog.tsx
@@ -69,12 +69,12 @@ export const ConnectSdkDialog = ({
     const theme = useTheme();
     const isLargeScreen = useMediaQuery(theme.breakpoints.up('lg'));
     const [sdk, setSdk] = useState<Sdk | null>(null);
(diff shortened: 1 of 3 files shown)


frontend/src/component/onboarding/ConnectSdkDialog.tsx
```diff
@@ -68,25 +68,48 @@ export const ConnectSdkDialog = ({
 }: IConnectSDKDialogProps) => {
     const theme = useTheme();
     const isLargeScreen = useMediaQuery(theme.breakpoints.up('lg'));
+    const [sdk, setSdk] = useState<Sdk | null>(null);
```
kwasniew, line 71: parent component will keep track of sdk and env because we need to remember you previous setting when going back and forth

sjaanus (approved):

PR #8077, opened 2024-09-04
fix: editor can add dependencies in default project

kwasniew:
## About the changes

Editors should be able to add dependencies in a default project.

Details:
* adding new permission for editor
* editor is selection is dynamic by the role name instead of hardcoding 2 which is magical
* making sure the queries are idempotent and can be run multiple times

### DIFF
--- a/src/migrations/20240904084114-add-update-feature-dependency-editor.js
+++ b/src/migrations/20240904084114-add-update-feature-dependency-editor.js
@@ -1,0 +1,17 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        INSERT INTO role_permission (role_id, permission) SELECT id, 'UPDATE_FEATURE_DEPENDENCY' FROM roles WHERE name = 'Editor' LIMIT 1;
+        `,
+        cb
+    );
+};
+
+exports.down = function (db, cb) {
+    db.runSql(
+        `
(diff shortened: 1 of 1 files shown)


src/migrations/20240904084114-add-update-feature-dependency-editor.js
```diff
@@ -0,0 +1,29 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        INSERT INTO role_permission (role_id, permission)
+        SELECT id, 'UPDATE_FEATURE_DEPENDENCY'
+        FROM roles WHERE name = 'Editor'
+        AND EXISTS (SELECT 1 FROM roles WHERE name = 'Editor')
```
kwasniew, line 7: prevent a case where someone manually deleted editor

src/migrations/20240904084114-add-update-feature-dependency-editor.js
```diff
@@ -0,0 +1,29 @@
...
+        AND NOT EXISTS (
+            SELECT 1
```
kwasniew, line 9: prevent double insert

sjaanus (changes requested):
Verify that the role is rool role.

sjaanus (approved):
Query itself looks good.

PR #8080, opened 2024-09-04
feat: navigate between all stages

kwasniew:
## About the changes

* can navigate between all 3 stages
* last stage has only empty placeholder
* second stage sets the API key for the last stage to use
* simplified stage union type


frontend/src/component/onboarding/GenerateApiKey.tsx
```diff
@@ -159,11 +160,11 @@ const TokenExplanation = ({
                             </SecretExplanationDescription>
                         </ArcherElement>
                         <ArcherElement
-                            id='secreat-description'
+                            id='secret-description'
                             relations={[
                                 {
                                     targetId: 'secret',
-                                    targetAnchor: 'bottom',
+                                    targetAnchor: 'right',
```
kwasniew, line 167: bottom breaks react archer when the token is long and slips to the next line

PR #8084, opened 2024-09-04
Fix/project role permission grant

FredrikOseberg:
## Background

In #6380 we fixed a privilege escalation bug that allowed members of a project that had permission to add users to the project with roles that had a higher permission set than themselves. The PR linked essentially constricts you only be able to assign users to roles that you possess yourself if you are not an Admin or Project owner.

This fix broke expectations for another customer who needed to have a project owner without the DELETE_PROJECT permission. The fix above made it so that their custom project owner role only was able to assign users to the project with the role that they posessed.

## Fix

Instead of looking directly at which role the role granter has, this PR addresses the issue by making the assessment based on the permission sets of the user and the roles to be granted. If the granter has all the permissions of the role being granted, the granter is permitted to assign the role.

## Other considerations

The endpoint to get roles was changed in this PR. It previously only retrieved the roles that the user had in the project. This no-longer makes sense because the user should be able to see other project roles than the one they themselves hold when assigning users to the project. 

The drawback of returning all project roles is that there may be a project role in the list that the user does not have access to assign, because they do not hold all the permissions required of the role. This was discussed internally and we decided that it's an acceptable trade-off for now because the complexities of returning a role list based on comparing permissions set is not trivial. We would have to retrieve each project role with permissions from the database, and run the same in-memory check against the users permission to determine which roles to return from this endpoint. Instead we opted for returning all project roles and display an error if you try to assign a role that you do not have access to.

## Follow up
When this is merged, there's no longer need for the frontend logic that filters out roles in the role assignment form. I deliberately left this out of the scope for this PR because I couldn't wrap my head around everything that was going on there and I thought it was better to pair on this with @chriswk or @nunogois in order to make sure we get this right as the logic for this filtering seemed quite complex and was touching multiple different components.


src/lib/features/project/project-service.ts
```diff
@@ -940,15 +941,38 @@ export default class ProjectService {
...
         if (
             this.isAdmin(userAddingAccess.id, userRoles) ||
             this.isProjectOwner(userRoles, projectId)
         ) {
             return true;
         }
-        return rolesBeingAdded.every((roleId) =>
-            userRoles.some((userRole) => userRole.id === roleId),
-        );
+
```
FredrikOseberg, line 951: The following LoCs represent the real change in this PR

src/lib/features/project/project-service.e2e.test.ts
```diff
@@ -772,72 +772,92 @@ describe('Managing Project access', () => {
...
             email: 'otheruser_no_roles@example.com',
         });
-        const customRole = await stores.roleStore.create({
-            name: 'role_that_noone_has',
-            roleType: 'custom',
-            description:
-                'Used to prove that you can not assign a role you do not have via setRolesForUser',
-        });
+
+        const customRoleUserAccess = await accessService.createRole(
```
FredrikOseberg, line 796: Refactoring of the assignment e2e tests to make sure they also fail when trying to grant a role that has permissions that the granter does not have.

chriswk (approved):
This seems fair. Though at first the owner without delete project sounded like just a custom role, but of course the owner doesn't have the owner role without the delete project permission, so then it made sense. Let's go

Issue #7987, opened 2024-08-27
Update a Segment via API call


theomeuh:
### Discussed in https://github.com/orgs/Unleash/discussions/7980

<div type='discussions-op-text'>

<sup>Originally posted by **theomeuh** August 26, 2024</sup>
I'm trying to migrate some data we have in Unleash and I need to automatically update Segment via API. I'm referring to this endpoint: https://docs.getunleash.io/reference/api/unleash/update-segment

### Context
It says: "Any fields not specified will be left untouched", but if I try to PUT an empty body, I got a 400 Bad Request `{"id":"1dd37e9f-2b75-458b-b7ca-1424e0561479","name":"BadDataError","message":"Request validation failed: your request body or params contain invalid data. Refer to the `details` list for
more information.","details":[{"message":"\"name\" is required."}]}`

Then let's add a "dummy name".. I got the same message but this time, it is the field "constraints" which is apparently missing.

My question: is the doc up-to-date ? It really looks like a PUT (like the endpoint verb says) and not a PATCH (like the description says), do you confirm ?

PS: what I really want is to update "constraints" but unfortunately, I have the "name is required" error when I try to just push a payload with only "constraints"</div>

### Current behavior

A PUT requests on `api/admin/segments/:id` with in body a json containing only the key `constraints` returns a 400, saying the field `name` is required

### Expected behavior

A PUT requests on `api/admin/segments/:id` with in body a json containing only the key `constraints` must return a 200 (and update the `constraints` field)

gastonfournier:
Thanks @theomeuh we'll look into it, we have somewhat related work we're planning to take on soon, and I included this as part of that backlog, we'll keep you posted on this ticket.

nunogois:
Hi @theomeuh,

Thank you for bringing up this issue. We’ve reviewed your concerns and I’d like to provide some clarification and updates.

You’re correct in pointing out that the behavior of the endpoint can be confusing, especially regarding how it handles PUT requests. While the API documentation indicates that unspecified fields will remain untouched (which would align more with PATCH behavior), the endpoint currently requires `name` and `constraints` to be present. Other fields should be left untouched, however there's also an edge case for the `project` property, where it is reset to `null` if not present.

https://github.com/Unleash/unleash/pull/8085 doesn’t really fix the issue, but at least it includes the request body schema, which is the same as the POST endpoint.

[...]

PR #8085, opened 2024-09-04
fix: add request body schema in update segment endpoint

nunogois:
https://linear.app/unleash/issue/2-2592/updateimprove-a-segment-via-api-call

Related to https://github.com/Unleash/unleash/issues/7987

This does not make the endpoint necessarily better - It's still a PUT that acts as a PUT in some ways (expects specific required fields to be present, resets the project to `null` if it's not included in the body) and a PATCH in others (ignores most fields if they're not included in the body). We need to have a more in-depth discussion about developing long-term strategies for our API and respective OpenAPI spec.

However this at least includes the proper schema for the request body, which is slightly better than before.


gastonfournier (approved):
Great :sparkles:!

PR #8092, opened 2024-09-05
feat: add sdk example box

sjaanus:
image

### DIFF
--- a/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectOnboarding/SdkExample.tsx
+++ b/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectOnboarding/SdkExample.tsx
@@ -1,8 +1,8 @@
 import { type SelectChangeEvent, styled, Typography } from '@mui/material';
 import { Link } from 'react-router-dom';
 import Select from 'component/common/select';
-import { OFFICIAL_SDKS } from '../../../../integrations/IntegrationList/AvailableIntegrations/SDKs';
 import { useState } from 'react';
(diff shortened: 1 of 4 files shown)


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectOnboarding/SdkExample.tsx
```diff
@@ -0,0 +1,72 @@
+import { type SelectChangeEvent, styled, Typography } from '@mui/material';
+import { Link } from 'react-router-dom';
+import Select from '../../../../common/select';
```
kwasniew, line 3: component/common

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectOnboarding/SdkExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { OFFICIAL_SDKS } from '../../../../integrations/IntegrationList/AvailableIntegrations/SDKs';
```
kwasniew, line 4: I added the SDK list in the onboarding directory. The file is called sharedTypes.ts. We can use the integrations dir but I wanted to control which ones we display in the onboarding just in case the lists don't match between integrations and onboarding.

kwasniew: We have value level lists in the SelectSdk.tsx so you can export it from there. Then we don't depend on the other module.

kwasniew (approved):

PR #8095, opened 2024-09-05
fix: use the correct env vars for initial admin account

nunogois:
https://github.com/Unleash/unleash/pull/7795 mistakenly changed the initial admin account environment variables to be:

 - `INITIAL_ADMIN_USER_USERNAME`
 - `INITIAL_ADMIN_USER_PASSWORD`

Instead of the previous:

 - `UNLEASH_DEFAULT_ADMIN_USERNAME`
 - `UNLEASH_DEFAULT_ADMIN_PASSWORD`

See: https://github.com/Unleash/unleash/pull/7795/files?diff=unified&w=0#diff-a38cc7ae9f963264333b6535498970a33e046dba50a6af554f438343a6edd2ddR319-R334

This PR reverts them back to the correct environment variable names and adds a test to prevent something similar from happening in the future.

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -317,8 +317,8 @@ const defaultVersionOption: IVersionOption = {
 };
 
 const parseEnvVarInitialAdminUser = (): UsernameAdminUser | undefined => {
-    const username = process.env.INITIAL_ADMIN_USER_USERNAME;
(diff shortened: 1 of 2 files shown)


chriswk (approved):
Nice Nuno. Thanks for fixing this!

PR #8099, opened 2024-09-05
feat: connection status bubble

sjaanus:
Fully working cycle

image


frontend/src/component/onboarding/ConnectionInformation.tsx
```diff
@@ -0,0 +1,109 @@
...
+    fontSize: theme.fontSizes.smallBody,
+}));
+
+export const ConnectionInformation = ({
+    onConnection,
+    projectId,
+}: IConnectionInformationProps) => {
+    const theme = useTheme();
+    const { project } = useProjectOverview(projectId, {
+        refreshInterval: 1000,
```
kwasniew, line 57: now that we're pinging if so often and knowing the query complexity behind it I'm thinking about a dedicated endpoint that returns nothing but the onboarding status. The project overview can still return both but this page would get fit for purpose model without all the irrelevant information that we query over and over again. WDYT?

sjaanus: Perhaps when GA, this screen is open sooo rarely that it has almost no impact.

frontend/src/component/onboarding/ConnectionInformation.tsx
```diff
@@ -0,0 +1,109 @@
...
+                    <Typography fontWeight='bold' variant='body2'>
+                        Environment
+                    </Typography>
+                    <Typography variant='body2'>Development</Typography>
+                </Info>
+                <Info>
+                    <Typography fontWeight='bold' variant='body2'>
+                        SDK
+                    </Typography>
+                    <Typography variant='body2'>NodeJS</Typography>
```
kwasniew, line 82: Don't you want to parametrize the SDK and env?

frontend/src/component/onboarding/ConnectionInformation.tsx
```diff
@@ -0,0 +1,109 @@
...
+    const theme = useTheme();
+    const { project } = useProjectOverview(projectId, {
+        refreshInterval: 1000,
+    });
+
+    const onboarded = project.onboardingStatus.status === 'onboarded';
+
+    useEffect(() => {
+        if (onboarded) {
+            onConnection();
```
kwasniew, line 64: did you think about the scenario where people come back to the 3rd stage again? E.g. after closing the dialog and opening it again? Or when going back and forth between the stages?

kwasniew (approved):

PR #8111, opened 2024-09-06
feat: make node example more consistent

sjaanus:


frontend/src/component/onboarding/sdkSnippets.ts
```diff
@@ -29,12 +29,13 @@ dotnet add package Newtonsoft.Json`,
...
 
 export const initializeCodeSnippets: Record<SdkName, string> = {
-    Node: `import { initialize } from 'unleash-client';
+    Node: `const { initialize } = require('unleash-client');
 
 const unleash = initialize({
   url: '<YOUR_API_URL>',
   appName: 'unleash-onboarding-node',
   customHeaders: { Authorization: '<YOUR_API_TOKEN>' },
+  metricsInterval: 5000,
```
sjaanus, line 38: I was waiting for a while to get it connected, thought it is disabled by default, but it seems default is 60 seconds and it was not responsive enough, that I expected. Making it 5 seconds, so metrics are sent sooner.

frontend/src/component/onboarding/sdkSnippets.ts
```diff
@@ -29,12 +29,13 @@ dotnet add package Newtonsoft.Json`,
 };
 
 export const initializeCodeSnippets: Record<SdkName, string> = {
-    Node: `import { initialize } from 'unleash-client';
+    Node: `const { initialize } = require('unleash-client');
```
sjaanus, line 32: This is what we suggest in out sdk github page and this is what works out of the box.

kwasniew: I thought that most people use ESM modules nowadays so wanted to provide the most up to date example

frontend/src/component/onboarding/sdkSnippets.ts
kwasniew: We can even make it 1 second to see the results quicker

frontend/src/component/onboarding/sdkSnippets.ts
sjaanus: I am still not sure, because default package json is not ESM.

sjaanus: If you know that you have configured type module, then you know how to change the import.
But if you have no idea, about ESM vs commonjs, you will be behind a hurdle.

kwasniew (approved):

PR #8112, opened 2024-09-06
fix: give project_default_strategy_write the ability to update the default strategy

thomasheartman:
This appears to have been an oversight in the original implementation
of this endpoint. This seems to be the primary point of this
permission. Additionally, the docs mention that this permission should
allow you to do just that.

Note: I've not added any tests for this, because we don't typically add tests for it. If we have an example to follow, I'd be very happy to add it, though


kwasniew (approved):
It's declarative so fine without a test

PR #8114, opened 2024-09-06
fix: allow you to save default strategies with the right permissions

thomasheartman:
Allow you to edit default strategies in the UI if you have the update_project or project_default_strategy_write permissions. These are the same permissions that we use in the API.

Previously, we used the update_feature_strategy permission here, but that one is intended to be used for updating strategies belonging to actual flags.

One of the trickier bits here is that we use the `StrategyVariants` component, which previously had baked in the permission required (update_feature_environment_variants). Because the permissions are different for the default strategy, I updated the component to make it configurable, but for the default to be the old permission (so that other uses aren't affected).


kwasniew (approved):
I'm assuming you tested it manually

thomasheartman:
Sure did 🙋🏼

PR #8118, opened 2024-09-06
fix: user projects should exclude archived ones

kwasniew:
## About the changes

User projects should only list active, not archived projects
Screenshot 2024-09-06 at 12 11 12

### DIFF
--- a/src/lib/features/project/project-store.ts
+++ b/src/lib/features/project/project-store.ts
@@ -539,20 +539,28 @@ class ProjectStore implements IProjectStore {
     async getProjectsByUser(userId: number): Promise<string[]> {
         const projects = await this.db
             .from((db) => {
-                db.select('project')
+                db.select('role_user.project')
                     .from('role_user')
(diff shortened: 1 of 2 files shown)


src/lib/features/project/project-store.ts
```diff
@@ -539,20 +539,28 @@ class ProjectStore implements IProjectStore {
     async getProjectsByUser(userId: number): Promise<string[]> {
         const projects = await this.db
             .from((db) => {
-                db.select('project')
+                db.select('role_user.project')
                     .from('role_user')
                     .leftJoin('roles', 'role_user.role_id', 'roles.id')
+                    .leftJoin('projects', 'role_user.project', 'projects.id')
                     .where('user_id', userId)
+                    .andWhere('projects.archived_at', null)
```
kwasniew, line 547: other archived_at usages are behind a flag but since we're just about to remove the flag from the code I add it directly here

Tymek (approved):

PR #8151, opened 2024-09-16
Update Entra ID SAML SSO docs

melindafekete:
## About the changes
Updates the ~~Azure~~ Entra ID SAML SSO docs to reflect the latest product name changes from Microsoft, update (and remove) some screenshots, and ensure that claims + namespaces are correctly shown.

Docs deeplink-> **[How to Add SSO with SAML 2.0 and Microsoft Entra ID](https://unleash-docs-git-melinda-azure-sso-unleash-team.vercel.app/how-to/how-to/how-to-add-sso-azure-saml)**


website/static/img/microsoft-entra-admin-center.png
melindafekete, line 1: From @daveleek:
> Right now, it seems we do not support namespaces at all in our claims for firstName and lastName. We also show two different sets of claims names, one uses firstName & lastName which without namespaces would be correct. the other one under group sync shows givenname and surname with namespaces which doesn't appear to work with our without namespaces. 

I can't `user.firstName` or `user.lastName` attributes in the Entra admin UI anymore, I think they removed them. In the new screenshot, I'm only showing `user.mail` as an attribute. Not sure if we want to explicitly mention, that we don't support any others? 🤷‍♀️

daveleek: Oh wow, that's interesting! Going to have to check the other test setup

daveleek: Yeah gone there too, just a dropdown now with valid values.

daveleek: In our test setup we currently have these claims:
image

I'll have a look at them to see which ones are in use, I think we use firstName and lastName claims (transformed in Entra ID from givenname and surname) to set the full name of the user in Unleash

daveleek: > I think we use firstName and lastName claims (transformed in Entra ID from givenname and surname) to set the full name of the user in Unleash

Ok checked and we do. Looks like we need 3 claims to set up a user with an email + a name; Unique User Identifier, firstName, and lastName. And groups for group sync. I guess that makes the name parameter in my screenshot above redundant

melindafekete: Thanks!

So should we specifically state in the instructions to add `user.surname` and `user.givenname` claims without namespaces?

daveleek: Yeah I think we should at least explain that it's how user names get populated in Unleash, it'll work without them but that might look/feel weird as the users get listed with just the email address

PR #8159, opened 2024-09-17
chore: add eventTimeline feature flag

nunogois:
https://linear.app/unleash/issue/2-2658/create-eventtimeline-feature-flag

Adds a new `eventTimeline` feature flag for the new event timeline feature.

I think `eventTimeline` is an appropriate name given the feature description and the way it is evolving, but I'm open to suggestions. ~~This also assumes that this feature will target OSS.~~ Confirmed that this will be a premium feature.

### DIFF
--- a/frontend/src/interfaces/uiConfig.ts
+++ b/frontend/src/interfaces/uiConfig.ts
@@ -90,6 +90,7 @@ export type UiFlags = {
     archiveProjects?: boolean;
     projectListImprovements?: boolean;
     onboardingUI?: boolean;
+    eventTimeline?: boolean;
 };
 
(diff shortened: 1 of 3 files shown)


daveleek (approved):
You need the snapshot update don't you? Apart from that LGTM!

Edit: Did we disable the snapshots?

nunogois:
> You need the snapshot update don't you? Apart from that LGTM!
> 
> Edit: Did we disable the snapshots?

Had the same thought you did, so I went to check and apparently we omitted the flags from the snap in https://github.com/Unleash/unleash/pull/8041

PR #8170, opened 2024-09-18
fix: upgrade openapi to drop dompurify and fix path-to-regexp

gastonfournier:
This new version [does not depend on dompurify](https://github.com/Unleash/unleash/pull/8170/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL3853-L3858) which should fix https://github.com/Unleash/unleash/security/dependabot/195

Tested the UI locally
image


nunogois (approved):
Nice one!

PR #8176, opened 2024-09-18
chore: event timeline

nunogois:
https://linear.app/unleash/issue/2-2657/implement-a-first-iteration-of-an-horizontal-event-timeline

This implements the very first iteration of our event timeline.

This is behind a feature flag, which when enabled shows the new timeline at the top of our event log page.

It is missing some features, like:
 - Placement: It should show up as an option in the header, not in the event log;
 - Tooltip: It should show proper tooltips for all the events that we're displaying;
 - Grouping: It should group together events that occurred in a short span of time;
 - Signals: It should show signals along with the events;

Here's how it currently looks like, with some example events, in order from left to right:
 - A flag was disabled more than 30 min ago;
 - A flag was then enabled;
 - A segment was updated (didn't have an icon for segments, so I picked one);
 - A strategy was updated;
 - A flag was created;

image
image

(Time passed since I took the first screenshot, so you can see the events "moved" to the left slightly in the dark theme screenshot)

I have some concerns about the low contrast of `neutral` variant events, especially in dark mode. Maybe we should consider using `error` instead, for red? Or maybe add a border to our event circles?

I specifically changed my environment to be "development" for the screenshots. The default selection is the first enabled environment that is `type=production`, which in my case is "production".

Here are our filters:
 - Time Span
image
 - Environment
image

Here are a few more screenshots, with the different time spans (zooming out, since we're increasing the time span):

image
image
image
image
image
image

Again, when zooming out, some events should be grouped together, but that's a task for later.

### DIFF
--- a/frontend/src/component/events/EventTimeline/EventTimeline.tsx
+++ b/frontend/src/component/events/EventTimeline/EventTimeline.tsx
@@ -23,7 +23,7 @@ const StyledTimelineContainer = styled('div')(({ theme }) => ({
     width: '100%',
     display: 'flex',
     alignItems: 'center',
-    padding: theme.spacing(1, 0),
+    padding: theme.spacing(1.5, 0),
 }));
 
 const StyledTimeline = styled('div')(({ theme }) => ({
@@ -47,7 +47,7 @@ const StyledMarker = styled('div')(({ theme }) => ({
 }));
 
 const StyledMiddleMarker = styled(StyledMarker)(({ theme }) => ({
-    top: theme.spacing(-1.5),
+    top: theme.spacing(-2),
 }));
 
 const StyledMarkerLabel = styled('div')(({ theme }) => ({
--- a/frontend/src/component/events/EventTimeline/EventTimeline.tsx
+++ b/frontend/src/component/events/EventTimeline/EventTimeline.tsx
@@ -1,0 +1,151 @@
+import { styled } from '@mui/material';
+import type { EventSchemaType } from 'openapi';
+import { useState } from 'react';
+import { startOfDay, sub } from 'date-fns';
+import type { IEnvironment } from 'interfaces/environments';
+import { useEventSearch } from 'hooks/api/getters/useEventSearch/useEventSearch';
+import { EventTimelineEvent } from './EventTimelineEvent/EventTimelineEvent';
+import {
+    EventTimelineHeader,
+    type TimeSpanOption,
+    timeSpanOptions,
+} from './EventTimelineHeader/EventTimelineHeader';
+
+const StyledRow = styled('div')({
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+});
+
+const StyledTimelineContainer = styled('div')(({ theme }) => ({
+    position: 'relative',
+    height: theme.spacing(1),
+    width: '100%',
+    display: 'flex',
+    alignItems: 'center',
+    padding: theme.spacing(1, 0),
+}));
+
+const StyledTimeline = styled('div')(({ theme }) => ({
+    backgroundColor: theme.palette.divider,
+    height: theme.spacing(0.5),
+    width: '100%',
+}));
+
+const StyledMiddleMarkerContainer = styled('div')({
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    position: 'relative',
+});
+
+const StyledMarker = styled('div')(({ theme }) => ({
+    position: 'absolute',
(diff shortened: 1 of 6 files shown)


frontend/src/component/events/EventTimeline/EventTimeline.tsx
```diff
@@ -0,0 +1,151 @@
...
+    'feature-created',
+    'feature-updated',
+    'feature-variants-updated',
+    'feature-archived',
+    'feature-revived',
+    'feature-strategy-update',
+    'feature-strategy-add',
+    'feature-strategy-remove',
+    'feature-environment-enabled',
+    'feature-environment-disabled',
```
nunogois, line 80: These are the events we're currently considering for the timeline.

frontend/src/component/events/EventTimeline/EventTimeline.tsx
```diff
@@ -0,0 +1,151 @@
...
+    );
+
+    const filteredEvents = events.filter(
+        (event) =>
+            new Date(event.createdAt).getTime() >= startDate.getTime() &&
+            new Date(event.createdAt).getTime() <= endDate.getTime() &&
+            RELEVANT_EVENT_TYPES.includes(event.type) &&
+            (!event.environment ||
+                !environment ||
+                event.environment === environment.name),
```
nunogois, line 110: Here I decided that events without a specific environment should always show regardless of our environment selection.

frontend/src/component/events/EventTimeline/EventTimelineEvent/EventTimelineEvent.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+    if (type.startsWith('feature-')) {
+        return <FlagOutlinedIcon />;
+    }
+    if (type.startsWith('segment-')) {
+        return <GroupsIcon />;
+    }
+
+    return <QuestionMarkIcon />;
+};
```
nunogois, line 67: This is where we pick which icon will show for each event type.

frontend/src/component/events/EventTimeline/EventTimelineEvent/EventTimelineEvent.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const customEventVariants: Partial<
+    Record<EventSchemaType, CustomEventVariant>
+> = {
+    'feature-environment-enabled': 'success',
+    'feature-environment-disabled': 'neutral',
+    'feature-archived': 'neutral',
+};
```
nunogois, line 75: And this is where we specify special variants, like "green" or "grey" icons. Any other event types default to "secondary" which is our "light purple".

frontend/src/component/events/EventTimeline/EventTimelineEvent/EventTimelineEventTooltip/EventTimelineEventTooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return (
+        <div>
+            <div>{eventDateTime}</div>
+            <div>{event.createdBy}</div>
+            <div>{event.type}</div>
+            <div>{event.featureName}</div>
+            <div>{event.environment}</div>
+        </div>
+    );
+};
```
nunogois, line 50: This needs to be properly implemented. Will be addressed at a later point.

frontend/src/component/events/EventTimeline/EventTimelineHeader/EventTimelineHeader.tsx
```diff
@@ -0,0 +1,167 @@
...
+        () => environments.filter(({ enabled }) => enabled),
+        [environments],
+    );
+
+    useEffect(() => {
+        if (activeEnvironments.length > 0) {
+            const defaultEnvironment =
+                activeEnvironments.find(({ type }) => type === 'production') ||
+                activeEnvironments[0];
+            setEnvironment(defaultEnvironment);
```
nunogois, line 108: Here is the logic for the default environment selection, as mentioned in the PR description.

frontend/src/component/events/EventTimeline/EventTimelineEvent/EventTimelineEventTooltip/EventTimelineEventTooltip.tsx
gastonfournier: Cool, this looks like a rendering strategy. Agree to do it later!

gastonfournier (approved):
Very good and paying attention to details!

PR #8181, opened 2024-09-19
fix: archive project curl

kwasniew:
## About the changes

Remove extra ' from the curl command
Screenshot 2024-09-19 at 09 46 48


gastonfournier (approved):
Oh great! :)

PR #8182, opened 2024-09-19
fix: update playground SDK to increase the possible random numbers used for stickiness id

thomasheartman:
Same fix as done in some other sdks, such as the node one at
https://github.com/Unleash/unleash-client-node/pull/417o

Fixes an issue where 1% rollout would not yield any results with
random rollout for certain group ids

### DIFF
--- a/src/lib/features/playground/feature-evaluator/strategy/flexible-rollout-strategy.ts
+++ b/src/lib/features/playground/feature-evaluator/strategy/flexible-rollout-strategy.ts
@@ -10,7 +10,7 @@ const STICKINESS = {
 
 export default class FlexibleRolloutStrategy extends Strategy {
     private randomGenerator: Function = () =>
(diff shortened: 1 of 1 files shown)


src/lib/features/playground/feature-evaluator/strategy/flexible-rollout-strategy.ts
```diff
@@ -10,7 +10,7 @@ const STICKINESS = {
 
 export default class FlexibleRolloutStrategy extends Strategy {
     private randomGenerator: Function = () =>
-        `${Math.round(Math.random() * 100) + 1}`;
+        `${Math.round(Math.random() * 10_000) + 1}`;
```
kwasniew, line 13: I'd use the same 10000 as Node SDK for easy code searching between the two

PR #8205, opened 2024-09-20
chore: event timeline tooltips

nunogois:
https://linear.app/unleash/issue/2-2664/implement-event-tooltips

Implements event tooltips in the new event timeline.

This leverages our current `feature-event-formatter-md` to provide both a label and a summary of the event. Whenever our new `eventTimeline` flag is enabled, we enrich our events in our event search endpoint with this information. We've discussed different options here and reached the conclusion that this is the best path forward for now. This way we are being consistent, DRY, relatively performant and it also gives us a happy path forward if we decide to scope in the event log revamp, since this data will already be present there.

We also added a new `label` property to each of our event types currently in our event formatter. This way we can have a concise, human-readable name for each event type, instead of exposing the internal event type string.

~~We also fixed the way the event formatter handled bold text (as in, **bold**). Before, it was wrapping them in *single asterisks*, but now we're using **double asterisks**. We also abstracted this away into a helper method aptly named `bold`. Of course, this change meant that a bunch of snapshots and tests needed to be updated.~~

~~This new `bold` method also makes it super easy to revert this decision if we choose to, for any reason. However I believe we should stick with markdown formatting, since it is the most commonly supported formatting syntax, so I see this as an important fix. It's also in the name of the formatter (`md`). I also believe bold was the original intent. If we want italic formatting we should implement it separately at a later point.~~

Edit: It was _bold_ of me to assume this would work out of the box on Slack. It does when you manually try it on the app, but not when using the Slack client. See: https://github.com/Unleash/unleash/pull/8222

image

image


src/lib/openapi/spec/event-schema.ts
```diff
@@ -92,6 +92,16 @@ export const eventSchema = {
...
         },
+        label: {
+            type: 'string',
+            nullable: true,
+            description: 'The concise, human-readable name of the event.',
+        },
+        summary: {
+            type: 'string',
+            nullable: true,
+            description: 'A markdown-formatted summary of the event.',
```
daveleek, line 103: Could we mark this as Experimental somehow? `**experimental** A markdown-formatted summary of the event.` ?

src/lib/openapi/spec/event-schema.ts
```diff
@@ -92,6 +92,16 @@ export const eventSchema = {
             nullable: true,
             description: 'Any tags related to the event, if applicable.',
         },
+        label: {
+            type: 'string',
+            nullable: true,
+            description: 'The concise, human-readable name of the event.',
```
daveleek, line 98: Same comment as below - could we mark this as experimental somehow?

daveleek (approved):
I like it, looks good!
If we could mark the new properties on the returns from the API as experimental somehow that would be good. Why? Because we're not 100% sure yet if we want a separate endpoint for this further along, or if we want to do this a different way by exporting formatting as a separate project etc. So marking this as experimental will be a way to not have to commit to it.

@gastonfournier / @sighphyre i like the improvement around extracting the boldness into a separate helper, and as @nunogois explained - we're already styling this to be bold in slack but it's styled with single asterisks for some reason. As long as double asterisks still result in bold in Slack I don't see this as being a breaking change. WDYT?

src/lib/openapi/spec/event-schema.ts
nunogois: WDYT of this? https://github.com/Unleash/unleash/pull/8205/commits/8432e536a72e61db763f05b96f2ba7f1c3c4f4a5

daveleek: Yeah that'll work! Should be super clear :)

nunogois:
> > We also fixed the way the event formatter handled bold text (as in, bold). Before, it was wrapping them in single asterisks, but now we're using double asterisks. We also abstracted this away into a helper method aptly named bold. Of course, this change meant that a bunch of snapshots and tests needed to be updated.
> > This new bold method also makes it super easy to revert this decision if we choose to, for any reason. However I believe we should stick with markdown formatting, since it is the most commonly supported formatting syntax, so I see this as an important fix. It's also in the name of the formatter (md). I also believe bold was the original intent. If we want italic formatting we should implement it separately at a later point.
> 
> Should we make this a different PR? I see that for Slack messages, they use a different format than Markdown: https://slack.com/help/articles/202288908-Format-your-messages maybe this single asterisk was developed with Slack messages in mind. I agree we might want to change it to something more standard and then before sending Slack messages changing to single asterisks (at the presentation layer of Slack)...

Reverted this in https://github.com/Unleash/unleash/pull/8205/commits/701225af3cb093ae62cf664bac77405d0cdb63c6

I was mistakenly convinced that double asterisks would work out of the box on Slack because they do when you type them on the app. However that doesn't seem to be the case when using their client. 

I agree that it's probably simpler to leave this alone for now. I have another idea that will leave the integrations alone for now but still provide us with proper markdown if requested. I'll tackle that in another PR.

daveleek (approved):
LGTM!

PR #8216, opened 2024-09-23
fix: import svg viewbox misalignment

thomasheartman:
This change updates the import svg's viewbox. It was set to be 24 x
24, but the icon is only 16px wide, causing it to be shifted 4px to
the left, as shown in the following image:

image

By shrinking the viewbox (but keeping the width the same), the icon is
now more correctly aligned:

image


gastonfournier (approved):
LG!

PR #8217, opened 2024-09-23
display setup complete message when project is onboarded

thomasheartman:
This PR adds the new `ProjectSetupComplete` component (the name can be changed) that we display when a project has been set up with a flag and a connected SDK.

It uses the project overview to check the project's onboarding status.

image


kwasniew:
It's ok to use project overview for now. However I was also considering a fit for purpose read model that returns only what we need. Currently the list of projects is read from the user profile but it doesn't have the flags/members/health information. If we decide to create a new resource (e.g. /personal-projects or /personal-dashboard-projects) we could also enhance it with the onboarding information.

kwasniew:
Also once we settle for the API we will create a test that verifies conditional loading of different UI combinations depending on the state of the system

kwasniew (approved):

PR #8219, opened 2024-09-23
fix: trim messages longer than 3000 chars

gastonfournier:
## About the changes
When trying to send messages longer than 3000 chars we get this error:
```
[ERROR]  web-api:WebClient:0 failed to match all allowed schemas [json-pointer:/blocks/0/text]
[ERROR]  web-api:WebClient:0 must be less than 3001 characters [json-pointer:/blocks/0/text/text]
[2024-09-23T10:10:15.676] [WARN] addon/slack-app - All (1) Slack client calls failed with the following errors: A platform error occurred: {"ok":false,"error":"invalid_blocks","errors":["failed to match all allowed schemas [json-pointer:/blocks/0/text]","must be less than 3001 characters [json-pointer:/blocks/0/text/text]"],"response_metadata":{"messages":["[ERROR] failed to match all allowed schemas [json-pointer:/blocks/0/text]","[ERROR] must be less than 3001 characters [json-pointer:/blocks/0/text/text]"],"scopes":["incoming-webhook","users:read","channels:read","groups:read","mpim:read","im:read","users:read.email","chat:write"],"acceptedScopes":["chat:write"]}}
```

This PR trims the text length to 3000 chars.

We also upgrade slack API due to some security fixes: https://github.com/slackapi/node-slack-sdk/releases/tag/%40slack%2Fweb-api%407.3.4 

After checking the migration guide to v7 it seems that none of the breaking changes affect us: https://github.com/slackapi/node-slack-sdk/wiki/Migration-Guide-for-web%E2%80%90api-v7

## Testing
I did manual test this integration and the fix. The way to reproduce is adding a very long strategy name and sending that as an update on Slack:
image

Now the event succeeds and we notice on the integration event log that the message was trimmed:
image


src/lib/addons/slack-app.ts
```diff
@@ -138,7 +141,7 @@ export default class SlackAppAddon extends Addon {
             const requests = channels.map((name) => {
                 return this.slackClient!.chat.postMessage({
                     channel: name,
-                    text,
+                    text: text,
```
nunogois, line 144: Unneeded?

```suggestion
                    text,
```

src/lib/addons/slack.ts
```diff
@@ -73,8 +73,9 @@ export default class SlackAddon extends Addon {
             }
         }
 
-        const { text, url: featureLink } = this.msgFormatter.format(event);
-
+        const { text: formattedMessage, url: featureLink } =
+            this.msgFormatter.format(event);
+        const text = formattedMessage.substring(0, 3000);
```
nunogois, line 78: Just curious: Is this implementation simpler because this integration is deprecated?

nunogois (approved):
LGTM - Great job and fantastic use of the integration events feature 👍

src/lib/addons/slack.ts
gastonfournier: Simpler than what other implementation? I did it here because the "msgFormatter" is kind of generic, it doesn't say anything about Slack, so just for precaution I decided not to modify it. I was thinking "If this is a message formatter inside a Slack folder package by feature, then this might be only for Slack", but because it was not I just err on the safe side.

nunogois: Simpler than the one in Slack App. In my head it should be pretty much the same thing, so I was just curious why it wasn't.

gastonfournier: Oh, just because the slack-app is using the formatted text also as output for the event, this one is only using it for the push

nunogois: There's still a:

```ts
this.registerEvent({
    integrationId,
    state,
    stateDetails: stateDetails.join('\n'),
    event: serializeDates(event),
    details: {
        url,
        channels: slackChannels,
        username,
        message: text,
    },
});
```

here, which means that there's a small inconsistency between Slack and Slack App in the way they register integration events: Slack App will tell us the text was trimmed to 3000 characters, but Slack will just trim it.

It's a small detail, so I think it's fine, but again, was just curious 😄

gastonfournier: Oh! Thanks! I didn't understand. This is the PR for consistency:  https://github.com/Unleash/unleash/pull/8230

gastonfournier: I just didn't see the message at the bottom

PR #8221, opened 2024-09-23
feat: Personal flags UI component

kwasniew:
## About the changes

Screenshot 2024-09-23 at 16 56 01

Details:
* updated API to return flag type and project so that we can get correct icon for a flag type and a correct link to a flag
* UI wise it's the same scrollable list as for the projects


frontend/src/component/personalDashboard/PersonalDashboard.tsx
```diff
@@ -136,13 +138,56 @@ const useProjects = () => {
...
+};
+
 export const PersonalDashboard = () => {
     const { user } = useAuthUser();
 
     const name = user?.name;
 
     const { projects, activeProject, setActiveProject } = useProjects();
 
+    const { personalDashboard } = usePersonalDashboard();
```
kwasniew, line 180: I will extract this code into custom hook once we find shared items between flags and projects

frontend/src/component/personalDashboard/PersonalDashboard.tsx
```diff
@@ -136,13 +138,56 @@ const useProjects = () => {
...
+    const [activeFlag, setActiveFlag] = useState<string | null>(null);
+    useEffect(() => {
+        if (
+            personalDashboard?.flags.length &&
+            personalDashboard?.flags.length > 0
```
thomasheartman, line 185: Is this meant to be more like this (though both will work)?

```suggestion
            personalDashboard?.flags &&
            personalDashboard?.flags.length > 0
```

We should also be able to simplify it:
```suggestion
            personalDashboard?.flags.length > 0
```

though TS might require us to do 

```suggestion
            (personalDashboard?.flags.length ?? 0) > 0
```

at which point ... maybe?

thomasheartman (approved):

frontend/src/component/personalDashboard/PersonalDashboard.tsx
kwasniew: `if(personalDashboard?.flags.length)` will do

PR #8222, opened 2024-09-23
chore: feature event formatter md format style

nunogois:
https://linear.app/unleash/issue/2-2697/implement-proper-markdown-bold-format-in-feature-event-formatter-md

This is a follow up to https://github.com/Unleash/unleash/pull/8205, specifically [this comment](https://github.com/Unleash/unleash/pull/8205#issuecomment-2368207656) from @gastonfournier

Implements an easy way to switch between formatting styles in our event formatter. This enhancement will allow us to generate fully markdown-formatted event summaries when needed, while preserving the simplistic markdown formatting currently supported by platforms like Slack.

Also includes some slight scouting. See comments for details.


src/lib/addons/feature-event-formatter-md-events.ts
```diff
@@ -0,0 +1,359 @@
...
+        label: 'User deleted',
+        action: '{{b}}{{user}}{{b}} deleted user {{b}}{{event.preData.name}}{{b}}',
+        path: '/admin/users',
+    },
+    [USER_UPDATED]: {
+        label: 'User updated',
+        action: '{{b}}{{user}}{{b}} updated user {{b}}{{event.preData.name}}{{b}}',
+        path: '/admin/users',
+    },
+};
```
nunogois, line 359: I think it makes sense to split the previous code and put the events formatting map in its own file. It's still pretty big, but hopefully more manageable.

src/lib/addons/feature-event-formatter-md.ts
```diff
@@ -82,307 +21,39 @@ export enum LinkStyle {
...
     private readonly linkStyle: LinkStyle;
 
-    constructor(unleashUrl: string, linkStyle: LinkStyle = LinkStyle.MD) {
+    private readonly formatStyle: FormatStyle;
+
+    constructor({
+        unleashUrl,
+        linkStyle = LinkStyle.MD,
+        formatStyle = 'simple',
+    }: IFeatureEventFormatterMdArgs) {
```
nunogois, line 43: Named parameters improve code readability and make function calls more flexible by allowing arguments to be passed in any order.

src/lib/features/events/event-search-controller.ts
```diff
@@ -47,9 +47,10 @@ export default class EventSearchController extends Controller {
         this.eventService = eventService;
         this.flagResolver = config.flagResolver;
         this.openApiService = openApiService;
-        this.msgFormatter = new FeatureEventFormatterMd(
-            config.server.unleashUrl,
-        );
+        this.msgFormatter = new FeatureEventFormatterMd({
+            unleashUrl: config.server.unleashUrl,
+            formatStyle: 'markdown',
```
nunogois, line 52: We request properly formatted markdown only when necessary, minimizing the surface area of potential issues.

We can be confident that everything else remains unchanged and continues to meet previous expectations, as no updates to tests or snapshots were required.

src/lib/addons/feature-event-formatter-md.ts
```diff
@@ -82,307 +21,39 @@ export enum LinkStyle {
...
 
     private readonly linkStyle: LinkStyle;
 
-    constructor(unleashUrl: string, linkStyle: LinkStyle = LinkStyle.MD) {
+    private readonly formatStyle: FormatStyle;
+
+    constructor({
+        unleashUrl,
+        linkStyle = LinkStyle.MD,
+        formatStyle = 'simple',
```
daveleek, line 42: Could we rename and just use the linkStyle parameter? Seems to be options `SLACK | MD` just as what formatStyle is for: bold with one asterisk slack-style or bold with two markdown-style

nunogois: I'm not sure we can. Other integrations, aside from Slack, expect proper markdown links, but that doesn’t necessarily mean they fully support proper markdown. E.g. [Microsoft Teams](https://support.microsoft.com/en-us/office/use-markdown-formatting-in-microsoft-teams-4d10bd65-55e2-4b2d-a1f3-2bebdcd2c772)

For now, I’d prefer to leave the links as they are and simply add this additional option.

daveleek: Ok fair enough, they serve the same purpose more or less and work more or less the same, but since they looked so different I wondered if perhaps they could be merged.

daveleek (approved):
LGTM

src/lib/addons/feature-event-formatter-md.ts
nunogois: I think it's a very good point and something I definitely want to come back to sometime in the future. For now, I think I just want to ignore the different and inconsistent markdown implementations across integrations and leave them alone, as long as I'm still able to get properly formatted markdown out of our markdown formatter 😄

gastonfournier (commented):
A general comment: maybe if we try to package by feature, each component or feature could have a different formatter and makes it simpler to work with if we know that there are no side-effects when modifying a formatter (because each formatter would be used by just one component).

I think that makes sense because that allows us to customize them based on our needs (like trimming to 3000 chars in Slack)

gastonfournier (approved):
Agreed to tackle this later. Feature Markdown should be renamed but not now, and feature formatter is not only about feature changes, but also we should not address that here.

PR #8228, opened 2024-09-24
Chore: Add .vite to .gitignore

thomasheartman:
Noticed this file popped up suddenly. Let's ignore it.


frontend/.gitignore
```diff
@@ -62,3 +62,7 @@ cypress/screenshots/*
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
```
thomasheartman, line 65:
```suggestion
```

PR #8230, opened 2024-09-24
chore: send full message without trimming to the addon event

gastonfournier:
Address https://github.com/Unleash/unleash/pull/8219#discussion_r1772884465


nunogois (approved):
LGTM!

PR #8234, opened 2024-09-24
chore: event timeline header placement

nunogois:
https://linear.app/unleash/issue/2-2662/make-the-event-timeline-available-globally-through-a-new-header-button

https://github.com/user-attachments/assets/bde38ee8-cdd8-409d-a95e-0c06189e3d9b

(In the video, you’ll notice a slight delay before new events show up. This happens because the timeline automatically refreshes every 10 seconds)

Removes the event timeline from the event log and integrates it into a new header option.

I chose a middle-ground approach between options 1 and 2 from our Figma sketches. This solution provides the best of both worlds IMO: the timeline stands out as a global component, distinct from the current page context, while sliding in rather than overlapping the content. This way, users can view the timeline alongside the page content.


gastonfournier (approved):
LGTM!

PR #8237, opened 2024-09-24
feat: highlighting flags chart

kwasniew:
## About the changes

Adding highlighter shadow to each data point
Screenshot 2024-09-24 at 16 19 02

I had to explicitly disable annotations highlighting since ChartJS has a global registry of plugins and the annotation plugin from the data network tab was doing this registration. As a result if you open network data usage and then go to your dashboard you'd get the entire chart highlighted from the invisible annotation.

### DIFF
--- a/frontend/src/component/common/Chart/customHighlightPlugin.ts
+++ b/frontend/src/component/common/Chart/customHighlightPlugin.ts
@@ -1,9 +1,8 @@
 import type { Chart } from 'chart.js';
 
-export const customHighlightPlugin = {
+export const customHighlightPlugin = (width = 46, bottomOverflow = 34) => ({
     id: 'customLine',
     beforeDraw: (chart: Chart) => {
-        const width = 46;
         if (chart.tooltip?.opacity && chart.tooltip.x) {
             const x = chart.tooltip.caretX;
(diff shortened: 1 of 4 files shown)


frontend/src/component/personalDashboard/createChartOptions.ts
```diff
@@ -82,6 +82,11 @@ export const createBarChartOptions = (
     return {
         plugins: {
             legend: plugins?.legend,
+            // required to avoid the highlight plugin highlighting empty annotation
+            annotation: {
```
kwasniew, line 86: even though we don't use annotations on this chart it is used in the network data usage. ChartJs has global registrations so this page can be affected.

frontend/src/component/common/Chart/customHighlightPlugin.ts
```diff
@@ -1,9 +1,8 @@
 import type { Chart } from 'chart.js';
 
-export const customHighlightPlugin = {
+export const customHighlightPlugin = (width = 46, bottomOverflow = 34) => ({
```
kwasniew, line 3: defaults optimized for the network data usage where I took it from for reuse

thomasheartman (approved):

PR #8240, opened 2024-09-24
chore: event timeline persistent state

nunogois:
https://linear.app/unleash/issue/2-2700/persist-timeline-state-in-local-storage

Implements persistent state management for the event timeline using local storage. 

I believe this improves UX by persisting both the timeline toggle (visibility) state and applied filters across page refreshes.

Includes some scouting/refactoring and some workarounds to prevent the timeline from animating on page load (in most cases).


gastonfournier (approved):
LGTM!

PR #8244, opened 2024-09-25
fix: adjust welcome dialog size

kwasniew:
## About the changes

* make welcome dialog take 65% of the viewport width
* fix React warnings about div children of the p element

### DIFF
--- a/frontend/src/component/personalDashboard/ConnectSDK.tsx
+++ b/frontend/src/component/personalDashboard/ConnectSDK.tsx
@@ -1,4 +1,4 @@
-import { Button, styled, Typography } from '@mui/material';
+import { Button, styled } from '@mui/material';
 import type { FC } from 'react';
 
 const TitleContainer = styled('div')(({ theme }) => ({
@@ -35,10 +35,10 @@ export const CreateFlag: FC<{ project: string }> = ({ project }) => {
                 <NeutralCircleContainer>1</NeutralCircleContainer>
                 Create a feature flag
             </TitleContainer>
-            <Typography>
+            <div>
                 <div>The project currently holds no feature toggles.</div>
                 <div>Create a feature flag to get started.</div>
-            </Typography>
+            </div>
             <div>
                 <Button href={`projects/${project}`} variant='contained'>
                     Go to project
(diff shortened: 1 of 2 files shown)


frontend/src/component/personalDashboard/ConnectSDK.tsx
```diff
@@ -35,10 +35,10 @@ export const CreateFlag: FC<{ project: string }> = ({ project }) => {
                 <NeutralCircleContainer>1</NeutralCircleContainer>
                 Create a feature flag
             </TitleContainer>
-            <Typography>
+            <div>
                 <div>The project currently holds no feature toggles.</div>
                 <div>Create a feature flag to get started.</div>
-            </Typography>
+            </div>
```
thomasheartman, line 41: I think you might have this backward? Why not use a wrapping div and two `p` children (or typography, as it were)?

frontend/src/component/personalDashboard/WelcomeDialog.tsx
```diff
@@ -10,7 +10,7 @@ import { formatAssetPath } from 'utils/formatPath';
 const StyledDialog = styled(Dialog)(({ theme }) => ({
     '& .MuiDialog-paper': {
         borderRadius: theme.shape.borderRadiusLarge,
-        maxWidth: theme.spacing(140),
+        maxWidth: '65vw',
```
thomasheartman, line 13: Do we wanna set some sort of px max too? 🤷🏼  Maybe not?

frontend/src/component/personalDashboard/ConnectSDK.tsx
```diff
@@ -56,10 +56,13 @@ export const ConnectSDK: FC<{ project: string }> = ({ project }) => {
...
-                Your project is not yet connected to any SDK. In order to start
-                using your feature flag connect an SDK to the project.
-            </Typography>
+            <div>
+                <div>Your project is not yet connected to any SDK.</div>
+                <div>
+                    In order to start using your feature flag connect an SDK to
+                    the project.
+                </div>
+            </div>
```
thomasheartman, line 65: What's going on here? Again, if you want separate paragraphs, you can use multiple `p` tags.

thomasheartman (approved):
I'm ... not sure about this one.  I think the dialog size change is probably fine, but I'm not wild about wrapping text in divs instead of paragraph tags. Why are you doing it that way? Of course, if it's more appropriate semantically than p tags, that's all good, but it feels a little off to me.

PR #8245, opened 2024-09-25
feat: add group project roles to project roles

thomasheartman:
This PR adds project roles you get from groups to the list of roles listed for each project.


kwasniew (approved):
Do we have a similar query in the system where you get all the projects you have access to based on your role and group?

PR #8246, opened 2024-09-25
chore: remove the projects? Property from ICreateGroupModel

thomasheartman:
This property does not seem to be used anywhere, so we can remove it.

Can't find any references in code here or in enterprise. Let's try it and see if it breaks.


gastonfournier (approved):
LGTM!

PR #8247, opened 2024-09-25
feat: flag exposure in personal dashboard

kwasniew:
## About the changes

Show last seen widget and lifecycle widget when selecting a flag
Screenshot 2024-09-25 at 10 31 30

Same last seen as the project flags and single flag view
Screenshot 2024-09-25 at 10 31 34

Same lifecycle as the project flags and single flag view
Screenshot 2024-09-25 at 10 31 39

Details:
* create a new FlagExposure component to encapsulate exposure info
* had to improve lifecycle tooltip so that they don't reach to the URL for the project id. Instead we pass it explicitly from the parent. The reason for that is that personal dashboard is not tied to a project based URL 
* when archiving the flag from the lifecycle widget you end up on the same screen but we need to refresh the personal dashboard data. In other places we usually navigate to the project flags view after the archive but here it makes sense to stay in the dashboard
* if you archive a feature from the lifecycle tooltip with change requests enabled you get a change request created but you will only see it when you go to the project with a change request. Change request bar is tied to the project view. I think it's such a rare corner case that it's ok for now

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/FeatureLifecycleTooltip.test.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/FeatureLifecycleTooltip.test.tsx
@@ -8,7 +8,6 @@ import {
     DELETE_FEATURE,
     UPDATE_FEATURE,
 } from 'component/providers/AccessProvider/permissions';
-import { Route, Routes } from 'react-router-dom';
 
 const currentTime = '2024-04-25T08:05:00.000Z';
 const twoMinutesAgo = '2024-04-25T08:03:00.000Z';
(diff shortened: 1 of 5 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/FeatureLifecycleTooltip.test.tsx
```diff
@@ -23,24 +22,17 @@ const renderOpenTooltip = (
...
-        </Routes>,
+        <FeatureLifecycleTooltip
+            stage={stage}
+            onArchive={onArchive}
+            onComplete={onComplete}
+            onUncomplete={onUncomplete}
+            loading={loading}
+            project={'default'}
+        >
+            <span>child</span>
```
thomasheartman, line 33: curious: what is this child? is it just the text child in a span?

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/FlagExposure.tsx
```diff
@@ -0,0 +1,72 @@
...
+            )}
+
+            {feature.project ? (
+                <MarkCompletedDialogue
+                    isOpen={showMarkCompletedDialogue}
+                    setIsOpen={setShowMarkCompletedDialogue}
+                    projectId={feature.project}
+                    featureId={feature.name}
+                    onComplete={refetchFeature}
+                />
```
thomasheartman, line 68: Can a feature not have a project?

thomasheartman (approved):

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/FeatureLifecycleTooltip.test.tsx
kwasniew: this component accepts any ReactElement as children prop so wanted to specify sth that doesn't look like string. In tests we find this element by text so passing child without a span would also work

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureLifecycle/FlagExposure.tsx
kwasniew: Yes, a placeholder feature has an empty project

PR #8248, opened 2024-09-25
feat: add project owners to personal dashboard project payload

thomasheartman:
This PR adds project owner information to the personal dashboard's project payload.

To do so, it uses the existing project owners read model.

I've had to make a few changes to the project owners read model to accomodate this:
- make the input type to `addOwners` more lenient. We only need the project ids, so we can make that the only required property
- fall back to using email as the name if the user has no name or username (such as if you sign up with the demo auth)

## Discussion points

Should we pass the anonymize emails flag here? I've got a vague memory that we actually anonymize them in the db now, so we don't need to here, but I can't remember exactly. (Update: emails are encrypted on creation in demo; this shouldn't be necessary: https://unleash-internal.slack.com/archives/C046LV6HH6W/p1722514435046269)

I've also used a shortcut for populating the schema: just copied in the property from the project schema. Should we write it out in full?

### DIFF
--- a/src/lib/features/personal-dashboard/personal-dashboard-read-model.ts
+++ b/src/lib/features/personal-dashboard/personal-dashboard-read-model.ts
@@ -1,5 +1,4 @@
 import type { Db } from '../../db/db';
-import type { IProjectOwnersReadModel } from '../project/project-owners-read-model.type';
 import type {
     IPersonalDashboardReadModel,
     PersonalFeature,
@@ -15,11 +14,8 @@ type IntermediateProjectResult = Omit<PersonalProject, 'roles'> & {
 export class PersonalDashboardReadModel implements IPersonalDashboardReadModel {
     private db: Db;
 
-    private projectOwnersReadModel: IProjectOwnersReadModel;
-
-    constructor(db: Db, ProjectOwnersReadModel: IProjectOwnersReadModel) {
+    constructor(db: Db) {
         this.db = db;
-        this.projectOwnersReadModel = ProjectOwnersReadModel;
     }
 
     async getPersonalProjects(userId: number): Promise<PersonalProject[]> {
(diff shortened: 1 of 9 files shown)


src/lib/services/index.ts
```diff
@@ -406,7 +407,7 @@ export const createServices = (
 
     const personalDashboardService = new PersonalDashboardService(
         db
-            ? new PersonalDashboardReadModel(db)
+            ? new PersonalDashboardReadModel(db, new ProjectOwnersReadModel(db))
```
kwasniew, line 410: Since we already have a personal dashboard service maybe it should do the orchestration instead of one read model depending on the other read model?

src/lib/features/personal-dashboard/personal-dashboard-read-model.ts
```diff
@@ -14,8 +15,11 @@ type IntermediateProjectResult = Omit<PersonalProject, 'roles'> & {
 export class PersonalDashboardReadModel implements IPersonalDashboardReadModel {
     private db: Db;
 
-    constructor(db: Db) {
+    private projectOwnersReadModel: IProjectOwnersReadModel;
+
+    constructor(db: Db, ProjectOwnersReadModel: IProjectOwnersReadModel) {
         this.db = db;
+        this.projectOwnersReadModel = ProjectOwnersReadModel;
```
kwasniew, line 22: upper case ProjectOwnersReadModel, also I'd move this dependency to the service

src/lib/features/personal-dashboard/personal-dashboard-read-model.ts
```diff
@@ -80,6 +80,7 @@ export class PersonalDashboardReadModel implements IPersonalDashboardReadModel {
             (project: IntermediateProjectResult) => {
                 const roles = Object.values(project.roles);
                 roles.sort((a, b) => a.id - b.id);
+
```
kwasniew, line 83: not sure what happened on this line

kwasniew (approved):

src/lib/features/personal-dashboard/personal-dashboard-read-model.ts
thomasheartman: probably a stray console.log that i added and then removed
```suggestion
```

PR #8249, opened 2024-09-25
feat: group timeline events by their position on the timeline

daveleek:
Groups timeline events by their position on the timeline, defined by TIME_GROUPING_SIZE as number of percentages of width


nunogois:
Closed in favor of https://github.com/Unleash/unleash/pull/8254

PR #8251, opened 2024-09-25
test: reduce noise from test warnings

kwasniew:
## About the changes

Reducing noise in frontend tests

Most of the console warn logs we get in frontend tests are:
* wrap React in act warnings (since we use react testing library it shouldn't be needed)
* MSW HTTP mock not setup

The first warning will be hopefully gone with those issues fixed:
* https://github.com/testing-library/react-testing-library/issues/1338
* https://github.com/testing-library/react-testing-library/issues/1216

The second warning we can ignore since we don't need to setup every HTTP interaction to make our tests pass and returning error by default should be fine.

By reducing those warning we are left with the important warnings that we can act upon

### DIFF
--- a/frontend/src/setupTests.ts
+++ b/frontend/src/setupTests.ts
@@ -1,6 +1,7 @@
 import '@testing-library/jest-dom';
 import 'whatwg-fetch';
 import 'regenerator-runtime';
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
This is pretty clever :raised_hands:

PR #8254, opened 2024-09-25
chore: implement event grouping in the event timeline

nunogois:
https://linear.app/unleash/issue/2-2663/implement-event-grouping-when-multiple-events-happen-in-a-short-period

This PR introduces a grouping logic for timeline events, enhancing the way events are displayed when they occur close to each other.

We also updated and refactored components to support handling groups of events rather than individual events.

Also includes some minor code cleanups and optimizations as part of general refactoring efforts (scouting).

image


frontend/src/component/common/Markdown/Markdown.tsx
```diff
@@ -19,6 +19,9 @@ const LinkRenderer = ({
     );
 };
 
-export const Markdown = (props: ComponentProps<typeof ReactMarkdown>) => (
-    <ReactMarkdown components={{ a: LinkRenderer }} {...props} />
+export const Markdown = ({
+    components,
+    ...props
+}: ComponentProps<typeof ReactMarkdown>) => (
+    <ReactMarkdown components={{ a: LinkRenderer, ...components }} {...props} />
```
nunogois, line 26: For anyone coming across this: This was part of an experiment where we customized `em` components within our markdown renderer, in this case specifically for the timeline event tooltip. 

Even though we ended up choosing a different approach, we still see this as an improvement (scouting) that could be helpful for future cases where we need to override specific components in the same way.

daveleek (approved):
Awesome! :shipit:

PR #8267, opened 2024-09-26
fix: layout content width on large screens

nunogois:
I believe the changes in https://github.com/Unleash/unleash/pull/8234, more specifically [this change](https://github.com/Unleash/unleash/pull/8234/files#diff-3a5e0763fe1effa9cd485763fecf0633ce8f0877c9d1971e6bd5c5abe7426364L64), have caused the main layout content to no longer stretch to its maximum relative width on larger screens.

This adjustment seems to resolve this issue, ensuring proper behavior on larger screens without introducing any overflow issues on smaller screens.

### DIFF
--- a/frontend/src/component/layout/MainLayout/MainLayout.tsx
+++ b/frontend/src/component/layout/MainLayout/MainLayout.tsx
@@ -117,7 +117,7 @@ export const MainLayout = forwardRef<HTMLDivElement, IMainLayoutProps>(
         const { isChangeRequestConfiguredInAnyEnv } = useChangeRequestsEnabled(
             projectId || '',
         );
-        const eventTimeline = useUiFlag('eventTimeline'); // && !isOss();
+        const eventTimeline = useUiFlag('eventTimeline') && !isOss();
         const { open: showTimeline, setOpen: setShowTimeline } =
             useEventTimeline();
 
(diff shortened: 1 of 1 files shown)


thomasheartman:
Yeah, slick! That's this one, right? https://linear.app/unleash/issue/1-2895/project-list-should-always-be-full-width

frontend/src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -67,6 +67,9 @@ const NewMainLayoutContent = styled(Grid)(({ theme }) => ({
     margin: '0 auto',
     paddingLeft: theme.spacing(2),
     paddingRight: theme.spacing(2),
+    [theme.breakpoints.up(1856)]: {
+        width: '100%',
+    },
```
thomasheartman, line 72: Don't we have standardized breakpoint sizes? `L`, `XL`, etc?

thomasheartman (approved):
This is probably a good change, but it's hard to tell without screenshots showing exactly what the issue is. Is this linear ticket the same issue? https://linear.app/unleash/issue/1-2895/project-list-should-always-be-full-width

frontend/src/component/layout/MainLayout/MainLayout.tsx
nunogois: We do, and we use the defaults for most things: https://mui.com/material-ui/customization/breakpoints/ however there's one specifically for `1856` on this component's styles:

https://github.com/Unleash/unleash/pull/8267/files#diff-3a5e0763fe1effa9cd485763fecf0633ce8f0877c9d1971e6bd5c5abe7426364R73

PR #8269, opened 2024-09-26
feat: get projects by ids

kwasniew:
## About the changes

Adding ability to read projects by multiple ids.

Background:
In the personal dashboard we need to return a list of user projects with health/flags counts/members count (that a user is a member of or favorited). My plan is to have a separate query to read user project ids and then pass those ids to the existing query. This way we can orchestrate the desired behavior in a service without writing new complex queries. 

Screenshot 2024-09-26 at 10 38 54

## Discussion points

I considered changing query param id to query param ids (plural) in place, but since we use the single id in a few places I find it safer to add a new property and not touch the existing code.

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -83,6 +83,7 @@ import type {
     IProjectApplicationsSearchParams,
     IProjectEnterpriseSettingsUpdate,
     IProjectQuery,
+    IProjectsQuery,
 } from './project-store-type';
 import type { IProjectFlagCreatorsReadModel } from './project-flag-creators-read-model.type';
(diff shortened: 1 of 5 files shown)


src/lib/features/project/project-service.ts
```diff
@@ -230,13 +231,13 @@ export default class ProjectService {
     }
 
     async getProjects(
-        query?: IProjectQuery,
+        query?: IProjectQuery & IProjectsQuery,
         userId?: number,
     ): Promise<ProjectForUi[]> {
-        const getProjects = () =>
```
kwasniew, line 236: gitar leftover

thomasheartman (approved):

PR #8271, opened 2024-09-26
fix: add foreign key constraint to role_permissions table

FredrikOseberg:
# Background

#8084 surfaced an issue with permissions. 22 months ago we added a permission that replaced UPDATE_FEATURE_VARIANTS with UPDATE_FEATURE_ENVIRONMENT_VARIANTS. 17 months later we added this permission to logic in enterprise related to creating and deleting environments. In between those 17 months, if we created and deleted an environment the UPDATE_FEATURE_ENVIRONMENT_VARIANTS permission would not be deleted, and it would hang around associated to an environment that no longer existed. 

This also applies for environment permissions that are not automatically applied to built in roles, such as: SKIP_CHANGE_REQUEST, APPLY_CHANGE_REQUEST and APPROVE_CHANGE_REQUEST. Creating a custom role with these permissions in a specific environment and then deleting the environment would also cause these permissions to hang around forever in the database.

Further complicating the situation, many of permissions in the database have empty strings in the environment column instead of null.

# Solution

This PR attempts to solve the situation by doing the following: 

We create a migration that first (1) converts all empty string values in the role_permission table to null, then (2) delete all permissions in role_permission where environment is not null and environments value does not exist in the name column of the environments table and (3) finally add the foreign key constraint on role_permission environments column that references name column in environments table with an on delete cascade.

Secondly, we guard all incoming data when we create or update a role by cleaning the permissions and setting environment to null in the access service, to keep backwards compatability and because the UI is currently sending empty string values.

Thirdly, we add a transaction around updating and creating a role since these are not atomic actions. Especially update is dangerous because we first wipe all permissions, then re-add all the new permissions. Failing to re-add the permissions puts the role in a state where it has lost all permissions.

### DIFF
--- a/src/lib/features/access/createAccessService.ts
+++ b/src/lib/features/access/createAccessService.ts
@@ -1,4 +1,4 @@
-import type { Db, IUnleashConfig } from '../../server-impl';
+import type { Db, IUnleashConfig, Knex } from '../../server-impl';
 import GroupStore from '../../db/group-store';
 import { AccountStore } from '../../db/account-store';
 import RoleStore from '../../db/role-store';
@@ -17,6 +17,11 @@ import {
     createFakeEventsService,
 } from '../events/createEventsService';
 
+// We need this function to satisfy the type expectations of withTransactional
+export const curriedCreateAccessService =
+    (config: IUnleashConfig) => (db: Knex) =>
+        createAccessService(db, config);
+
 export const createAccessService = (
     db: Db,
     config: IUnleashConfig,
@@ -33,6 +38,7 @@ export const createAccessService = (
         { getLogger },
         eventService,
     );
+
     return new AccessService(
         { accessStore, accountStore, roleStore, environmentStore },
         { getLogger },
--- a/src/lib/services/access-service.ts
+++ b/src/lib/services/access-service.ts
@@ -108,6 +108,15 @@ export interface AccessWithRoles {
 
 const isProjectPermission = (permission) => PROJECT_ADMIN.includes(permission);
 
+export const cleanPermissions = (permissions: PermissionRef[] | undefined) => {
+    return permissions?.map((permission) => {
+        if (permission.environment === '') {
+            return { ...permission, environment: null };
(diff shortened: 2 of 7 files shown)


src/lib/features/access/createAccessService.ts
```diff
@@ -17,6 +17,11 @@ import {
     createFakeEventsService,
 } from '../events/createEventsService';
 
+// We need this function to satisfy the type expectations of withTransactional
+export const curriedCreateAccessService =
```
FredrikOseberg, line 21: withTransactional expects a function that takes config and returns an instantiated service

src/migrations/20240917071436-add-foreign-key-to-role-permissions.js
```diff
@@ -0,0 +1,20 @@
+exports.up = function (db, cb) {
+    db.runSql(
+        `
+        UPDATE role_permission SET environment = null where environment = '';
```
FredrikOseberg, line 4: Update all permissions in `role_permissions` where environment is set to empty string

src/migrations/20240917071436-add-foreign-key-to-role-permissions.js
```diff
@@ -0,0 +1,20 @@
...
+        DELETE FROM role_permission WHERE environment IS NOT NULL AND environment NOT IN (SELECT name FROM environments);
```
FredrikOseberg, line 5: Delete all permissions from role_permission where environment is not null and does not exist in the environments table

src/lib/services/access-service.ts
```diff
@@ -108,6 +108,15 @@ export interface AccessWithRoles {
 
 const isProjectPermission = (permission) => PROJECT_ADMIN.includes(permission);
 
+export const cleanPermissions = (permissions: PermissionRef[] | undefined) => {
```
FredrikOseberg, line 111: Ensure all incoming permissions use null instead of empty string in the service layer

PR #8274, opened 2024-09-26
feat: project details for personal dashboard

kwasniew:
## About the changes

Changing personal dashboard to return project details (health, members count, flags count).
Removing roles and owners for now (making them optional, keeping the definitions so that we can move them to single project endpoint)

### DIFF
--- a/src/lib/features/personal-dashboard/personal-dashboard-read-model-type.ts
+++ b/src/lib/features/personal-dashboard/personal-dashboard-read-model-type.ts
@@ -1,20 +1,24 @@
 import type { ProjectOwners } from '../project/project-owners-read-model.type';
 
 export type PersonalFeature = { name: string; type: string; project: string };
-export type PersonalProject = {
+export type BasePersonalProject = {
     name: string;
     id: string;
-    roles: {
+    roles?: {
         name: string;
         id: number;
         type: 'custom' | 'project' | 'root' | 'custom-root';
     }[];
 };
-export type PersonalProjectWithOwners = PersonalProject & {
-    owners: ProjectOwners;
+export type PersonalProject = BasePersonalProject & {
+    owners?: ProjectOwners;
+} & {
+    health: number;
+    memberCount: number;
+    featureCount: number;
 };
 
 export interface IPersonalDashboardReadModel {
     getPersonalFeatures(userId: number): Promise<PersonalFeature[]>;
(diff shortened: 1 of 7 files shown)


src/lib/features/personal-dashboard/personal-dashboard-read-model-type.ts
```diff
@@ -1,20 +1,24 @@
 import type { ProjectOwners } from '../project/project-owners-read-model.type';
 
 export type PersonalFeature = { name: string; type: string; project: string };
-export type PersonalProject = {
+export type BasePersonalProject = {
```
kwasniew, line 4: we will remove this type once we finish migration to the single project endpoint

thomasheartman: are we also removing the optional roles, then?

src/lib/features/personal-dashboard/personal-dashboard-read-model-type.ts
```diff
@@ -1,20 +1,24 @@
...
+    roles?: {
         name: string;
         id: number;
         type: 'custom' | 'project' | 'root' | 'custom-root';
     }[];
 };
-export type PersonalProjectWithOwners = PersonalProject & {
-    owners: ProjectOwners;
+export type PersonalProject = BasePersonalProject & {
+    owners?: ProjectOwners;
```
thomasheartman, line 14: shouldn't roles also be moved in here? And why are they optional? Because we're not returning them for now? Will we later, though? Isn't that a separate endpoint?

thomasheartman (approved):

src/lib/features/personal-dashboard/personal-dashboard-read-model-type.ts
kwasniew: I was thinking that we can read the roles lazily when you open single project together with owner, events and onboarding status fetched from the single project API in personal dashboard controller.
I kept owners and roles optional to allow easier migration (data structure copy paste to single project endpoint)

src/lib/features/personal-dashboard/personal-dashboard-read-model-type.ts
kwasniew: We're moving roles to a separate endpoint so will address the removal and adding to a new endpoint in a separate PR

PR #8279, opened 2024-09-26
feat: read projects from personal dashboard API

kwasniew:
## About the changes

Read project details from the personal dashboard API
Screenshot 2024-09-26 at 14 06 39

Also updated orval type

### DIFF
--- a/frontend/src/component/personalDashboard/PersonalDashboard.tsx
+++ b/frontend/src/component/personalDashboard/PersonalDashboard.tsx
@@ -13,7 +13,6 @@ import {
 import type { Theme } from '@mui/material/styles/createTheme';
 import { ProjectIcon } from 'component/common/ProjectIcon/ProjectIcon';
 import React, { type FC, useEffect, useState } from 'react';
-import { useProfile } from 'hooks/api/getters/useProfile/useProfile';
 import LinkIcon from '@mui/icons-material/Link';
 import { Badge } from '../common/Badge/Badge';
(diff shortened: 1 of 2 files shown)


frontend/src/component/personalDashboard/PersonalDashboard.tsx
```diff
@@ -81,21 +83,21 @@ export const StyledCardTitle = styled('div')<{ lines?: number }>(
...
 const ActiveProjectDetails: FC<{
-    project: { flags: number; members: number; health: number };
+    project: PersonalDashboardSchemaProjectsItem;
 }> = ({ project }) => {
     return (
         <Box sx={{ display: 'flex', gap: 2 }}>
             <Box sx={{ display: 'flex', flexDirection: 'column' }}>
                 <Typography variant='subtitle2' color='primary'>
-                    {project.flags}
+                    {project.featureCount}
```
thomasheartman, line 92: minor thing: since this is a new endpoint, should we use `flagCount` or stick with `featureCount`? Not sure we reached a conclusion on that part of the renaming anywhere, but we use `feature` all over the shop, so ... 🤷🏼

kwasniew: I decided to keep it consistent with the other endpoint I copied it from

kwasniew: We can still change our mind as long as the API is in the unstable land

PR #8285, opened 2024-09-27
feat: create page for when you have no projects

thomasheartman:
This adds a front end fallback screen for when you have no projects.

image

### DIFF
--- a/frontend/src/component/common/AvatarGroupFromOwners/AvatarGroupFromOwners.tsx
+++ b/frontend/src/component/common/AvatarGroupFromOwners/AvatarGroupFromOwners.tsx
@@ -9,7 +9,7 @@ type Props = {
     AvatarComponent?: typeof UserAvatar;
     className?: string;
 };
-export const OwnerAvatarGroup: React.FC<Props> = ({ users, ...props }) => {
+export const AvatarGroupFromOwners: React.FC<Props> = ({ users, ...props }) => {
     const { uiConfig } = useUiConfig();
 
     const mapOwners = (owner: ProjectSchemaOwners[number]) => {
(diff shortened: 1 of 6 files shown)


frontend/src/component/common/AvatarGroupFromOwners/AvatarGroupFromOwners.tsx
thomasheartman, line 1: We do this transformation in three different places now, so I thought it would be about time to break it out into a shared component.

frontend/src/component/personalDashboard/PersonalDashboard.tsx
```diff
@@ -213,88 +216,102 @@ export const PersonalDashboard = () => {
...
+            {noProjects ? (
+                <ContentGridNoProjects
+                    owners={[{ ownerType: 'system' }]}
+                    admins={[
+                        { name: 'admin' },
+                        {
+                            name: 'Christopher Tompkins',
+                            imageUrl:
+                                'https://avatars.githubusercontent.com/u/1010371?v=4',
+                        },
```
thomasheartman, line 228: This entire bit is largely just wrapping the old content in a ternary. we'll probably sort this out with the union type down the road, but this makes it work right now.

frontend/src/component/project/ProjectCard/ProjectCardFooter/ProjectOwners/ProjectOwners.tsx
```diff
@@ -106,7 +81,7 @@ export const ProjectOwners: FC<IProjectOwnersProps> = ({ owners = [] }) => {
                     <StyledOwnerName>
                         <StyledHeader data-loading>Owner</StyledHeader>
                         <StyledUserName data-loading>
-                            {users[0]?.name}
+                            {getOwnerName(owners[0])}
```
kwasniew, line 84: what if owners is an empty array?

kwasniew (approved):

frontend/src/component/project/ProjectCard/ProjectCardFooter/ProjectOwners/ProjectOwners.tsx
thomasheartman: Same thing happens as before; we show "system"

thomasheartman: Or rather: before, we wouldn't show anything (but theoretically it should never be empty). Now we show "system"

PR #8293, opened 2024-09-27
feat: add project owners to personal dashboard

thomasheartman:
This PR adds all user-type owners of projects that you have access to to the personal dashboard payload. It adds the new `projectOwners` property regardless of whether you have access to any projects or not because it required less code and fewer conditionals, but we can do the filtering if we want to.

To add the owners, it uses the private project checker to get accessible projects before passing those to the project owner read model, which has a new method to fetch user owners for projects.


src/lib/openapi/spec/personal-dashboard-schema.ts
```diff
@@ -8,6 +8,35 @@ export const personalDashboardSchema = {
     additionalProperties: false,
     required: ['projects', 'flags'],
     properties: {
+        projectOwners: {
+            type: 'array',
+            description:
+                'Users with the project owner role in Unleash. The list will not include owners of private projects that the user is not a member of, unless the owner is also an owner of a project that the is visible to the user.',
```
kwasniew, line 14: "unless the owner is also an owner of a project that the is visible to the user" isn't this repeating the same thing as "The list will not include owners of private projects that the user is not a member of"

src/lib/features/project/project-owners-read-model.test.ts
```diff
@@ -299,7 +299,7 @@ describe('integration tests', () => {
             projectId,
         );
 
-        const owners = await readModel.getAllProjectOwners();
+        const owners = await readModel.getProjectOwnersDictionary();
```
kwasniew, line 302: Why do you prefer Dictionary over All?

kwasniew (approved):

src/lib/openapi/spec/personal-dashboard-schema.ts
thomasheartman: Yes; yes it is. I just couldn't think of a better way to word it in the moment. Thanks for the suggestion?

PR #8295, opened 2024-09-27
fix: some integrations only counted errors not all results

daveleek:
Moves the addon/integration result metrics-counting to the finally block / outside the result successful if-else, to capture not just failed runs but successful runs as well

### DIFF
--- a/src/lib/addons/datadog.ts
+++ b/src/lib/addons/datadog.ts
@@ -113,14 +113,14 @@ export default class DatadogAddon extends Addon {
             state = 'failed';
             const failedMessage = `Datadog Events API request failed with status code: ${res.status}.`;
             stateDetails.push(failedMessage);
-            if (this.flagResolver.isEnabled('addonUsageMetrics')) {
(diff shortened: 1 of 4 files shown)


src/lib/addons/datadog.ts
```diff
@@ -113,16 +113,16 @@ export default class DatadogAddon extends Addon {
...
-
             this.logger.warn(failedMessage);
         }
 
+        if (this.flagResolver.isEnabled('addonUsageMetrics')) {
+            this.eventBus.emit(ADDON_EVENTS_HANDLED, {
+                result: state,
+                destination: 'datadog',
+            });
+        }
```
gastonfournier, line 124: This will work, but I'm thinking, maybe it's best to go with Ivar's approach... if in the next line we are registering an event, maybe we can rely on a listener that listens all events but only counts the ones that have an `integrationId` property set...

gastonfournier (approved):
Looks good as a fix, maybe we can follow up on my comment next week, but perhaps it's best to have the fix out to amend our metrics

PR #8299, opened 2024-09-30
feat: add Unleash admins to API payload

thomasheartman:
Adds Unleash admins to the personal dashboard payload.

Uses the access store (and a new method) to fetch admins and maps it to a new `MinimalUser` type. We already have a `User` class, but it contains a lot of information we don't care about here, such as `isAPI`, SCIM data etc.

In the UI, admins will be shown to users who are not part of any projects. This is the default state for new viewer users, and can also happen for editors if you archive the default project, for instance.

Tests in a follow-up PR

### DIFF
--- a/src/lib/openapi/spec/personal-dashboard-schema.ts
+++ b/src/lib/openapi/spec/personal-dashboard-schema.ts
@@ -7,6 +7,41 @@ export const personalDashboardSchema = {
     additionalProperties: false,
     required: ['projects', 'flags'],
     properties: {
+        admins: {
+            type: 'array',
+            description: 'Users with the admin role in Unleash.',
+            items: {
+                type: 'object',
+                required: ['id'],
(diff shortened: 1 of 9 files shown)


src/lib/openapi/spec/personal-dashboard-schema.ts
```diff
@@ -7,6 +7,41 @@ export const personalDashboardSchema = {
     additionalProperties: false,
     required: ['projects', 'flags'],
     properties: {
+        admins: {
```
kwasniew, line 10: do we want to make admins and project owners required?

thomasheartman: We *can*, but I didn't want to do it yet, in case we decide to only return those when the user has no projects.

kwasniew (approved):

PR #8303, opened 2024-09-30
test: test the dashboard admins property

thomasheartman:
This PR adds tests for the new admins property of the personal dashboard API payload.

It checks that only user admins are added and that their image URL is not an empty string. In doing this, also fixes an issue where the image URL wouldn't be generated correctly.

## Discussion points

Some of the test feels like it might be better testing on a deeper level (i.e. the account store). However, from an initial glance, I think that would require more setup and work, so I'm leaving it in the dashboard test for now as that's where it's ultimately useful. But we can discuss if we should move it.

### DIFF
--- a/src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
+++ b/src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
@@ -249,7 +249,16 @@ test('should return Unleash admins', async () => {
     const admins = body.admins;
     admins.sort((a, b) => a.id - b.id);
     expect(body.admins).toMatchObject([
-        { id: admin.id, username: admin.username },
-        { id: admin2.id, name: admin2.name, username: admin2.username },
+        {
+            id: admin.id,
+            username: admin.username,
+            imageUrl: expect.stringMatching(/^https:\/\/gravatar.com/),
+        },
+        {
+            id: admin2.id,
+            name: admin2.name,
+            username: admin2.username,
+            imageUrl: expect.stringMatching(/^https:\/\/gravatar.com/),
+        },
     ]);
 });
--- a/src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
+++ b/src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
@@ -250,8 +250,6 @@ test('should return Unleash admins', async () => {
 
     const { body } = await app.request.get(`/api/admin/personal-dashboard`);
 
-    const admins = body.admins;
-    admins.sort((a, b) => a.id - b.id);
     expect(body.admins).toMatchObject([
         {
             id: admin.id,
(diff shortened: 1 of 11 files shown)


src/lib/types/user.ts
```diff
@@ -33,7 +33,7 @@ export interface IUser {
     scimId?: string;
 }
 
-export type MinimalUser = Pick<
+export type MinimalUiUser = Pick<
```
kwasniew, line 36: What if this data is accessed through the API? Would it still make sense to call it "Ui"

src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
```diff
@@ -217,3 +217,52 @@ test('should return personal dashboard project details', async () => {
...
+        .returning('*');
+    await app.services.accessService.setUserRootRole(
+        serviceAdminId,
+        adminRoleId,
+    );
+
+    const { body } = await app.request.get(`/api/admin/personal-dashboard`);
+
+    const admins = body.admins;
+    admins.sort((a, b) => a.id - b.id);
```
kwasniew, line 254: I'm not a big fan of sorting in e2e/component tests. Is it changing between test runs?

kwasniew (approved):

src/lib/types/user.ts
thomasheartman: good point. The data is intended to be used to render a user (so it includes the image url, notably). I don't mind leaving or removing that part of the name. Or if you have a different suggestion, I'm very open to that too.

src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
thomasheartman: no idea. why don't you like sorting the data in tests like this? i'm thinking it'll keep it from breaking regardless of any changes that we do on the back end. We can leave it out, of course, but it'd be annoying if it suddenly broke later when we haven't specified the orde.r

src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
```diff
@@ -217,3 +217,52 @@ test('should return personal dashboard project details', async () => {
...
+    const admin2 = await userService.createUser({
+        username: 'John',
+        name: 'John Admin',
+        rootRole: adminRoleId,
+    });
+
+    // service account that shouldn't be listed in the output. Service
+    // accounts are enterprise functionality, so there's no service to
+    // call here
+    const [{ id: serviceAdminId }] = await db
```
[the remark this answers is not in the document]
thomasheartman: The controller, service, and store that create service accounts are all in the enterprise repo, if that's what you're asking?

src/lib/types/user.ts
kwasniew: Personally I'd stick to the previous name

thomasheartman: Done ✅

src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
thomasheartman: be151d1851dc0445d30a1b1193ac93fa647df85c

PR #8308, opened 2024-09-30
feat(onboarding): add links to examples

Tymek:
## About the changes
Links from Unleash UI to [Unleash/unleash-sdk-examples](https://github.com/Unleash/unleash-sdk-examples)

https://linear.app/unleash/issue/1-2869/add-codesandbox-links-to-unleashunleash

### DIFF
--- a/frontend/src/component/onboarding/flow/SdkExample.tsx
+++ b/frontend/src/component/onboarding/flow/SdkExample.tsx
@@ -20,39 +20,37 @@ const StyledLink = styled(Link)({
 
 const repositoryUrl =
     'https://github.com/Unleash/unleash-sdk-examples/tree/main';
-const exampleDirectories: Record<SdkName, string> = {
-    Android: 'Android',
-    '.NET': 'Csharp',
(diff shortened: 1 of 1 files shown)


frontend/src/component/onboarding/flow/SdkExample.tsx
```diff
@@ -18,17 +18,42 @@ const StyledLink = styled(Link)({
     textDecoration: 'none',
 });
 
+const repositoryUrl =
+    'https://github.com/Unleash/unleash-sdk-examples/tree/main';
+const exampleDirectories: Record<SdkName, string> = {
```
sjaanus, line 23: Could we get away without this mapping and name the directories the same as we have in app?

Tymek: Yes, if we deal with exceptions:
```
'.NET': 'Csharp',
'Node.js': 'NodeJS',
```

sjaanus (approved):
Nice

PR #8314, opened 2024-10-01
feat: return project roles

thomasheartman:
This PR updates the personal dashboard project endpoint to return owners and roles. It also adds the impl for getting roles (via the access store).

I'm filtering the roles for a project to only include project roles for now, but we might wanna change this later.

Tests and UI update will follow.

### DIFF
--- a/src/lib/features/personal-dashboard/personal-dashboard-service.ts
+++ b/src/lib/features/personal-dashboard/personal-dashboard-service.ts
@@ -10,6 +10,7 @@ import type {
 import type { IProjectReadModel } from '../project/project-read-model-type';
 import type { IPrivateProjectChecker } from '../private-project/privateProjectCheckerType';
 import type {
+    IAccessStore,
     IAccountStore,
     IEventStore,
     IOnboardingReadModel,
@@ -36,6 +37,8 @@ export class PersonalDashboardService {
 
     private onboardingReadModel: IOnboardingReadModel;
 
+    private accessStore: IAccessStore;
+
     constructor(
         personalDashboardReadModel: IPersonalDashboardReadModel,
         projectOwnersReadModel: IProjectOwnersReadModel,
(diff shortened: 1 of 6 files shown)


src/test/fixtures/fake-access-store.ts
thomasheartman, line 1: This file doesn't live where I expected it to. It also didn't expose the class as a named export (so the fact that it was mismatched didn't really matter much). Should we move it or leave it where it is?

src/lib/features/personal-dashboard/personal-dashboard-service.ts
```diff
@@ -117,11 +123,24 @@ export class PersonalDashboardService {
         const owners =
             await this.projectOwnersReadModel.getProjectOwners(projectId);
 
+        const allRoles = await this.accessStore.getAllProjectRolesForUser(
```
kwasniew, line 126: in the next PR we can Promise.all all those async operations

thomasheartman: Ah, yeah. Actually lemme do it right away, while I'm in that context.

src/test/fixtures/fake-access-store.ts
kwasniew: Access subdomain has not been fully moved to feature oriented architecture yet. I remember moving only the composition root because this was all I needed at that time to finish some other big migration. We can finish the migration (depending how much you're inclined to do it) to full feature oriented architecture but not in this PR :)

kwasniew (approved):

src/lib/features/personal-dashboard/personal-dashboard-service.ts
thomasheartman: Nah, I'll do it in a follow-up after all. Think it's better served that way

PR #8315, opened 2024-10-01
feat: handle owners and project roles in the UI

thomasheartman:
This commit uses the now-included project owner and role information
to populate the owner/role section. If you have no roles, we'll tell
you that you don't instead of displaying an empty set of badges.


kwasniew (approved):
Would be cool to add a screenshot to the PR for whoever gonna find this PR in future.

thomasheartman:
good point (though this is more about just using the right data; the original impls with screenies was in a previous PR). I'll add some after merging

PR #8319, opened 2024-10-01
chore: test that owners and roles are returned

thomasheartman:
Test that owners and roles are returned in the personal dashboard project details response as expected.

### DIFF
--- a/src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
+++ b/src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
@@ -215,9 +215,20 @@ test('should return projects where users are part of a group', async () => {
 });
 
 test('should return personal dashboard project details', async () => {
-    const { body: user } = await loginUser('new_user@test.com');
+    const { body: user1 } = await loginUser('new_user@test.com');
 
-    const project = await createProject(`x${randomId()}`, user);
+    const project = await createProject(`x${randomId()}`, user1);
+    const { body: user2 } = await loginUser('user2@test.com');
+    await app.services.projectService.addAccess(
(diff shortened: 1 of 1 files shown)


src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
```diff
@@ -253,9 +281,9 @@ test('should return personal dashboard project details', async () => {
                 ),
             },
             {
-                createdBy: 'audit user',
+                createdBy: 'unknown',
```
kwasniew, line 284: wondering what this change is about

kwasniew (approved):

src/lib/features/personal-dashboard/personal-dashboard-controller.e2e.test.ts
thomasheartman: Honestly, I don't know why `createdBy` is "unknown" here, but it feels like it's further down the stack and I didn't wanna start tangling with that now. We could remove that altogether and slice the array, though. to only get the flag events. 🤔  But since we're checking the whole object, maybe not.

kwasniew: I added previous one because it was there in the test result. But I'm surprised that it has changed. Let's keep it for now and observe this test. If it becomes unstable because of this we'll act accordingly.

thomasheartman: Oh, it changed because I added the user to the project after creating it, which triggers another event. That's why the event's changed. I don't think it'll be unstable. But we'll see.

PR #8320, opened 2024-10-01
fix: add missing awaits on events

FredrikOseberg:
This PR adds missing awaits on the call to eventStore which were revealed to be fire and forget when we added transactional support for these methods.


thomasheartman (approved):

gastonfournier (approved):
LGTM! Is this everywhere in the code? Maybe enterprise as well

PR #8321, opened 2024-10-01
refactor: implement an event timeline context and provider

nunogois:
https://linear.app/unleash/issue/2-2730/refactor-the-event-timeline-state-management-to-a-context-and-provider

This PR refactors the state management for the **Event Timeline** component by introducing a context and provider to improve accessibility of state across the component tree.

### DIFF
--- a/frontend/src/component/events/EventTimeline/EventTimelineContext.tsx
+++ b/frontend/src/component/events/EventTimeline/EventTimelineContext.tsx
@@ -1,0 +1,18 @@
+import { createContext, useContext } from 'react';
+import type { IEventTimelineContext } from './EventTimelineProvider';
+
+export const EventTimelineContext = createContext<
+    IEventTimelineContext | undefined
+>(undefined);
+
+export const useEventTimelineContext = (): IEventTimelineContext => {
+    const context = useContext(EventTimelineContext);
+
+    if (!context) {
+        throw new Error(
+            'useEventTimelineContext must be used within a EventTimelineProvider',
+        );
+    }
+
+    return context;
+};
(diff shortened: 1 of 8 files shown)


frontend/src/component/events/EventTimeline/EventTimelineContext.tsx
```diff
@@ -0,0 +1,18 @@
...
+    IEventTimelineContext | undefined
+>(undefined);
+
+export const useEventTimelineContext = (): IEventTimelineContext => {
+    const context = useContext(EventTimelineContext);
+
+    if (!context) {
+        throw new Error(
+            'useEventTimelineContext must be used within a EventTimelineProvider',
+        );
```
gastonfournier, line 14: Is this a common pattern? I'm also unsure if `useContext` can return undefined

gastonfournier (approved):
LGTM!

frontend/src/component/events/EventTimeline/EventTimelineContext.tsx
nunogois: Yes, and it can. See:
 - [Sticky](https://github.com/Unleash/unleash/blob/main/frontend/src/component/common/Sticky/Sticky.tsx#L40)
 - [useFeedback](https://github.com/Unleash/unleash/blob/main/frontend/src/component/feedbackNew/useFeedback.tsx#L23)

The idea is to wrap your component tree with a provider, allowing any nested components to access the context using `useContext`. If a component isn’t within the provider, `useContext` will return undefined. We control the component tree and typically place providers at a high level, so this should never happen. However, if we break it, the UI will loudly throw an error at us.

PR #8322, opened 2024-10-01
chore: timeline event group icon when events are all same type

nunogois:
https://linear.app/unleash/issue/2-2726/groups-should-show-the-timeline-event-type-icon-when-all-events-inside

Displays the event type icon for the group when all events within the group share the same type.

image

image


gastonfournier (approved):
LG!

PR #8329, opened 2024-10-02
a11y: add labels to env/period selectors

thomasheartman:
This PR adds labels to the environment and period selectors for the flag metrics chart on the personal dashboard page. Because the period selector is also used on the flag metrics page, I've also updated its use there by removing a previous text label (which was just text, not an associated `label` element).

Before:
Chart:
image

Flag metrics:
image

After:
Chart:
image

Flag metrics:
image


kwasniew (approved):
I like it

PR #8330, opened 2024-10-02
feat: remove bold/strong from personal dashboard events

kwasniew:
## About the changes

Closes #


nunogois (approved):
LGTM. 

Again, not happy with this being a permanent solution, but I think it's the best solution for now, and consistent with what we already did in the event timeline. Hopefully we'll revisit this sometime in the near future.

Cc @daveleek

PR #8335, opened 2024-10-02
feat: health trend insight

kwasniew:
## About the changes

Showing project health trend message for pro and enterprise customers. Currently it mimicks ios health app with a message how hour average health trend changes between last 4 weeks and previous 4 weeks. 
Screenshot 2024-10-02 at 12 26 28

We can add more info e.g. you're doing great job at keeping your project healthy etc. Looking forward to suggestions. the original version only states the facts without telling what's good or bad.


frontend/src/component/personalDashboard/MyProjects.tsx
```diff
@@ -137,8 +137,12 @@ export const MyProjects: FC<{
                 </List>
             </SpacedGridItem>
             <SpacedGridItem item lg={4} md={1}>
-                {activeProjectStage === 'onboarded' ? (
-                    <ProjectSetupComplete project={activeProject} />
+                {activeProjectStage === 'onboarded' &&
+                personalDashboardProjectDetails ? (
```
thomasheartman, line 141: question: can you have an active project stage without personaldashboardprojectdetails? is this redundant?

frontend/src/component/personalDashboard/ProjectSetupComplete.tsx
```diff
@@ -40,6 +40,91 @@ export const ProjectSetupComplete: FC<{ project: string }> = ({ project }) => {
...
+            insights.avgHealthCurrentWindow < insights.avgHealthPastWindow
+        ) {
+            trend = 'declined';
+        } else if (
+            insights.avgHealthPastWindow === insights.avgHealthCurrentWindow
+        ) {
+            trend = 'consistent';
+        }
+    }
+    return trend;
```
thomasheartman, line 69: we *could* do an early return here by checking if they're both null and just return the string directly. would decrease nesting by one level and remove the mutable variable. Your choice, though

frontend/src/component/personalDashboard/ProjectSetupComplete.tsx
```diff
@@ -40,6 +40,91 @@ export const ProjectSetupComplete: FC<{ project: string }> = ({ project }) => {
...
+                        {insights.avgHealthCurrentWindow}%
+                    </PercentageScore>{' '}
+                    during the last 4 weeks.
+                </Typography>
+            ) : null}
+            {projectHealthTrend !== 'unknown' ? (
+                <Link to={`/projects/${project}/insights`}>
+                    View more insights
+                </Link>
+            ) : null}
```
thomasheartman, line 127: I'd maybe consider extracting this into a variable above instead of a series of ternaries. Just to avoid any future issues 🤷🏼

thomasheartman (approved):

frontend/src/component/personalDashboard/MyProjects.tsx
kwasniew: no, but the orval type doesn't allow me to make this state more explicit so I ended up adding two expressions in here

frontend/src/component/personalDashboard/ProjectSetupComplete.tsx
kwasniew: 100%. Fixing

frontend/src/component/personalDashboard/ProjectSetupComplete.tsx
kwasniew: Refactored to a separate component with early returns

PR #8338, opened 2024-10-02
chore: timeline plausible tracking

daveleek:
Adds plausible events tracking of opening and closing events timeline


frontend/src/hooks/usePlausibleTracker.ts
```diff
@@ -66,7 +66,8 @@ export type CustomEvents =
     | 'new-in-unleash-click'
     | 'new-in-unleash-dismiss'
     | 'search-opened'
-    | 'events-exported';
+    | 'events-exported'
+    | 'timeline-usage';
```
nunogois, line 70: Just so we don't risk any overlaps with other eventual timelines in Unleash? Doesn't hurt to be a bit more specific.

```suggestion
    | 'event-timeline-usage';
```

nunogois (approved):
LGTM.

Note: I believe that in the past we struggled to evaluate Plausible events based on props, which is why you'll see some Plausible events that are very specific about the action. I wonder if we should go with one approach or the other. Maybe it's worth checking this in Plausible and see how it looks like for us to evaluate this.

frontend/src/hooks/usePlausibleTracker.ts
daveleek: Thank you, good catch!

daveleek:
Thanks @nunogois that's an interesting point. I could just add separate events, I'll have a look in the custom events spec and see if I see more of them. It would be like 3-5 different events depending on how detailed we want to be

frontend/src/component/events/EventTimeline/EventTimelineEventGroup/EventTimelineEventGroup.tsx
```diff
@@ -31,9 +32,13 @@ export const EventTimelineEventGroup = ({
     const eventTime = group[0].timestamp;
 
     const position = `${((eventTime - startTime) / timelineDuration) * 100}%`;
+    const { trackEvent } = usePlausibleTracker();
```
nunogois, line 35: Should we move this to the start of this component's code? That's what we usually do with hooks. So, in this case it would be before the `const timelineDuration = endTime - startTime;`

frontend/src/component/events/EventTimeline/EventTimelineEventGroup/EventTimelineEventGroup.tsx
```diff
@@ -31,9 +32,13 @@ export const EventTimelineEventGroup = ({
...
+    const trackHover = () => {
+        trackEvent('event-timeline-event-hover');
+    };
 
     return (
-        <StyledEvent position={position}>
+        <StyledEvent position={position} onMouseOver={trackHover}>
```
nunogois, line 41: Have we validated that this only happens **once** when you hover the element? I assumed the correct event would be onMouseEnter 🤔

daveleek: I can try debugging it and see what happens.

frontend/src/hooks/usePlausibleTracker.ts
```diff
@@ -66,7 +66,10 @@ export type CustomEvents =
     | 'new-in-unleash-click'
     | 'new-in-unleash-dismiss'
     | 'search-opened'
-    | 'events-exported';
+    | 'events-exported'
+    | 'event-timeline-click'
```
nunogois, line 70: Maybe this one is a bit more obvious than click?
```suggestion
    | 'event-timeline-open'
```

frontend/src/hooks/usePlausibleTracker.ts
```diff
@@ -66,7 +66,10 @@ export type CustomEvents =
...
+    | 'event-timeline-dismiss'
```
nunogois, line 71: Same as above.
```suggestion
    | 'event-timeline-close'
```

nunogois (approved):
Nice! 🚀

PR #8344, opened 2024-10-02
fix: add grid w/container query for projects

thomasheartman:
This PR is a work in progress, but the current state aims to suggest a direction. The main goals of this are:

1. Make it so that the layout grid doesn't break on small screens
2. Fix an issue where the border of the box didn't fit the outline
3. (Bonus): make the layout of the info box depend on the **box's** size, not the screen size.

To achieve those goals, this PR:
1. Switches to using a native CSS grid instead of MUI's grid component. This gives us more power over the layout in various different sizes.
2. Switches from putting borders on the boxes inside the grid, instead makes the grid container the color of the border and uses gaps to create borders.
3. If your browser supports it, it will use container queries to determine whether we should display the layout as a multi-column grid or in a single column.

Container query demo (both with the same screen sizes):

Sidebar closed: 
image

Sidebar open:
image

### DIFF
--- a/frontend/src/component/personalDashboard/Grid.tsx
+++ b/frontend/src/component/personalDashboard/Grid.tsx
@@ -12,7 +12,6 @@ const ContentGrid2 = styled('article')(({ theme }) => {
         overflow: 'hidden',
         border: `0.5px solid ${theme.palette.divider}`,
         gap: `2px`,
-
         display: 'grid',
 
         '&>*': {
--- a/frontend/src/component/personalDashboard/Grid.tsx
+++ b/frontend/src/component/personalDashboard/Grid.tsx
@@ -12,7 +12,8 @@ const ContentGrid2 = styled('article')(({ theme }) => {
         overflow: 'hidden',
         border: `0.5px solid ${theme.palette.divider}`,
         gap: `2px`,
-        display: 'grid',
+        display: 'flex',
+        flexFlow: 'column nowrap',
 
         '&>*': {
             backgroundColor: theme.palette.background.paper,
@@ -21,17 +22,9 @@ const ContentGrid2 = styled('article')(({ theme }) => {
 });
 
 export const ProjectGrid = styled(ContentGrid2)(({ theme }) => ({
-    gridTemplateAreas: `
-            "title"
-            "onboarding"
-            "projects"
-            "box1"
-            "box2"
-            "owners"
-            `,
-
     '@container (min-width: 1000px)': {
         gridTemplateColumns: '1fr 1fr 1fr',
+        display: 'grid',
         gridTemplateAreas: `
                 "title onboarding onboarding"
                 "projects box1 box2"
@@ -42,6 +35,7 @@ export const ProjectGrid = styled(ContentGrid2)(({ theme }) => ({
     '@supports not (container-type: inline-size)': {
         [theme.breakpoints.up('lg')]: {
             gridTemplateColumns: '1fr 1fr 1fr',
+            display: 'grid',
             gridTemplateAreas: `
                 "title onboarding onboarding"
                 "projects box1 box2"
--- a/frontend/src/component/personalDashboard/Grid.tsx
+++ b/frontend/src/component/personalDashboard/Grid.tsx
@@ -1,6 +1,75 @@
 import { Box, Grid, styled } from '@mui/material';
 import type { Theme } from '@mui/material/styles/createTheme';
 
+export const ContentGridContainer = styled('div')({
+    containerType: 'inline-size',
+});
+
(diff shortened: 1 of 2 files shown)


frontend/src/component/personalDashboard/Grid.tsx
```diff
@@ -1,6 +1,75 @@
 import { Box, Grid, styled } from '@mui/material';
 import type { Theme } from '@mui/material/styles/createTheme';
 
+export const ContentGridContainer = styled('div')({
+    containerType: 'inline-size',
+});
```
thomasheartman, line 6: You need external containers to be able to use container queries. This contains the grid so we can check the size of it.

frontend/src/component/personalDashboard/Grid.tsx
```diff
@@ -1,6 +1,74 @@
...
+        backgroundColor: theme.palette.divider,
+        borderRadius: `${theme.shape.borderRadiusLarge}px`,
+        overflow: 'hidden',
+        border: `0.5px solid ${theme.palette.divider}`,
+        gap: `2px`,
+        display: 'grid',
+
+        '&>*': {
+            backgroundColor: theme.palette.background.paper,
+        },
```
thomasheartman, line 19: make sure all direct children have a background color. This can be pushed to the gird items if we want.

frontend/src/component/personalDashboard/Grid.tsx
```diff
@@ -1,6 +1,74 @@
...
+            `,
+
+    '@container (min-width: 1000px)': {
+        gridTemplateColumns: '1fr 1fr 1fr',
+        gridTemplateAreas: `
+                "title onboarding onboarding"
+                "projects box1 box2"
+                ". owners owners"
+            `,
+    },
```
thomasheartman, line 40: because we use named areas in the items, that messes things up if we suddenly don't have any areas mentioned in the 1-column layout. As such, the workaround is to list them in all possible layouts.

frontend/src/component/personalDashboard/Grid.tsx
```diff
@@ -1,6 +1,74 @@
...
+
+    '@supports not (container-type: inline-size)': {
+        [theme.breakpoints.up('lg')]: {
+            gridTemplateColumns: '1fr 1fr 1fr',
+            gridTemplateAreas: `
+                "title onboarding onboarding"
+                "projects box1 box2"
+                ". owners owners"
+            `,
+        },
```
thomasheartman, line 50: If your browser doesn't support container queries AND we're on 'lg' screens, here's the fallback.

frontend/src/component/personalDashboard/Grid.tsx
```diff
@@ -1,6 +1,74 @@
...
+    '@container (min-width: 1000px)': {
+        display: 'block',
+    },
+
+    '@supports not (container-type: inline-size)': {
+        [theme.breakpoints.up('lg')]: {
+            display: 'block',
+        },
+    },
+}));
```
thomasheartman, line 70: This is an empty element to take up spaces in the grid when it's multi-column (the special `.` area). We hide it to avoid showing an empty div in the list when we collapse the grid.

frontend/src/component/personalDashboard/MyProjects.tsx
thomasheartman, line 1: This file is mostly just indentation caused by the container element that I added. The only relevant change here is removing `item`, `lg`, and `md` props from the grid items and switching to using the new components instead. Also: add grid areas.

frontend/src/component/personalDashboard/Grid.tsx
thomasheartman: Optionally (just thought of this): we could only use a grid layout when we're wider than X px, and use a flex layout otherwise. That's probably a cleaner option.

frontend/src/component/personalDashboard/Grid.tsx
kwasniew: why only inline and not block+inline?

frontend/src/component/personalDashboard/Grid.tsx
kwasniew: I think I prefer the grid/flex distinction depending on size

kwasniew (approved):
Very nice work!

frontend/src/component/personalDashboard/Grid.tsx
thomasheartman: That... I don't know. When playing with containers before, I've tried both, and I think this is usually the one that's worked out well for me. Must be something I don't quite get, but it works here, so 🤷🏼 

(side note: if you read up on it and grok it proper, please let me know 😄 )

PR #8347, opened 2024-10-03
refactor: use css grid for flags and no content grid

thomasheartman:
This PR uses the new CSS grid layout for the flag grid and the no content grid.

In doing so, it also improves how you use the grid item (giving them a `gridArea` prop) and extracts the breakpoint handling so that all sections that use breakpoints use the same breakpoints.

As with the previous PR, here's screenies of the same screen width, but with open and closed sidebar:
Open:
image

Closed:
image

### DIFF
--- a/frontend/src/component/personalDashboard/Grid.tsx
+++ b/frontend/src/component/personalDashboard/Grid.tsx
@@ -64,8 +64,11 @@ export const SpacedGridItem2 = styled('div', {
     gridArea,
 }));
 
-export const EmptyGridItem = styled('div')(({ theme }) => ({
+export const EmptyGridItem = styled('div', {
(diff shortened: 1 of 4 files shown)


frontend/src/component/personalDashboard/Grid.tsx
```diff
@@ -21,56 +21,60 @@ const ContentGrid2 = styled('article')(({ theme }) => {
...
+
+        return {
+            [`@container (min-width: ${containerBreakpoint})`]: css,
+
+            '@supports not (container-type: inline-size)': {
+                [screenBreakpoint]: css,
+            },
+        };
+    };
+
```
thomasheartman, line 38: Extract breakpoint handling / grid folding into a shareable function

kwasniew (approved):

PR #8348, opened 2024-10-03
feat: health score components in personal dashboard

kwasniew:
## About the changes

OSS users will always see this insight. Pro and Enterprise users will see this insights until they get 8 weeks of health trends data.
Screenshot 2024-10-03 at 09 57 45

Pro and Enterprise users will see this insight after they get 8 weeks of health trends data.
Screenshot 2024-10-03 at 09 57 56

Details:
* reusing health chart and health chart legend from the project insights. I decided to include the legend so that we know what each color means

### DIFF
--- a/frontend/src/component/project/Project/ProjectInsights/ProjectHealth/FlagCounts.tsx
+++ b/frontend/src/component/project/Project/ProjectInsights/ProjectHealth/FlagCounts.tsx
@@ -1,0 +1,80 @@
+import { Box, styled, useTheme } from '@mui/material';
+import { Link } from 'react-router-dom';
+import type { FC } from 'react';
+
+const Dot = styled('span', {
+    shouldForwardProp: (prop) => prop !== 'color',
+})<{ color?: string }>(({ theme, color }) => ({
+    height: '15px',
+    width: '15px',
+    borderRadius: '50%',
+    display: 'inline-block',
+    backgroundColor: color,
+}));
+
+const FlagCountsWrapper = styled(Box)(({ theme }) => ({
+    display: 'flex',
+    flexDirection: 'column',
+    gap: theme.spacing(2),
+}));
+
+const FlagsCount = styled('p')(({ theme }) => ({
+    color: theme.palette.text.secondary,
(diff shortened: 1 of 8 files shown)


frontend/src/component/project/Project/ProjectInsights/ProjectHealth/FlagCounts.tsx
```diff
@@ -0,0 +1,80 @@
...
+    marginLeft: theme.spacing(3),
+}));
+
+const StatusWithDot = styled(Box)(({ theme }) => ({
+    display: 'flex',
+    alignItems: 'center',
+    gap: theme.spacing(1),
+}));
+
+export const FlagCounts: FC<{
```
kwasniew, line 32: this code is extracted from the project insights health widget so that we can compose is inside personal dashboard

src/lib/features/personal-dashboard/personal-dashboard-service.ts
```diff
@@ -173,6 +183,11 @@ export class PersonalDashboardService {
             insights: {
                 avgHealthCurrentWindow,
                 avgHealthPastWindow,
+                totalFlags,
+                potentiallyStaleFlags,
+                staleFlags,
+                activeFlags: totalFlags - staleFlags - potentiallyStaleFlags,
```
thomasheartman, line 189: Don't we also get the total flags in the payload? Any reason we calculate again?

thomasheartman (approved):

src/lib/features/personal-dashboard/personal-dashboard-service.ts
kwasniew: We do get it in the personal dashboard global overview. This is the project details endpoint. Theoretically we could reuse the value from the global overview but this was just convenient. If we ever decide to re-use the project endpoint we get all the data in one go.

kwasniew: So in other words this change allows to move the project details around without thinking about the data source since it's self sufficient.

PR #8349, opened 2024-10-03
chore: plausible event for event timeline signals tip click

daveleek:
Triggers a plausible event when the signals tip is clicked in the events timeline

### DIFF
--- a/frontend/src/hooks/usePlausibleTracker.ts
+++ b/frontend/src/hooks/usePlausibleTracker.ts
@@ -67,10 +67,7 @@ export type CustomEvents =
     | 'new-in-unleash-dismiss'
     | 'search-opened'
     | 'events-exported'
-    | 'event-timeline-open'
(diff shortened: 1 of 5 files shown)


nunogois (approved):
LGTM

frontend/src/hooks/usePlausibleTracker.ts
```diff
@@ -69,7 +69,8 @@ export type CustomEvents =
     | 'events-exported'
     | 'event-timeline-open'
     | 'event-timeline-close'
-    | 'event-timeline-event-hover';
+    | 'event-timeline-event-hover'
```
sjaanus, line 72: Do we need separate event for all of these, we usually use properties?

daveleek: Thanks @sjaanus and @nunogois for the discussion on this, I have changed this back to being based on props

PR #8361, opened 2024-10-03
feat(onboarding): add comments to snippets

Tymek:
## About the changes
Add warning about plaintext API key and increased metrics interval


sjaanus (approved):

ivarconr (commented):
I am deeply worried about this. If most customers starts using 1 second interval for metrics and polling it will have severe impact on our hosting. 

1. I see no reason to change fetch interval as all SDK will fetch immediately anyway. And they should be aware of the delay when they update.

2. Updating the metrics interval is very dangerous for high loads. If we try to get metrics faster we could consider sending the first metrics a bit faster (we do that for the frontend javascript SDK).

PR #8362, opened 2024-10-04
fix: return 404 if the project doesn't exist

thomasheartman:
This change adds a check for whether the project exists in the
database before trying to fetch data for it. If it doesn't exist,
you'll get a 404.

### DIFF
--- a/src/lib/features/personal-dashboard/personal-dashboard-service.ts
+++ b/src/lib/features/personal-dashboard/personal-dashboard-service.ts
@@ -10,7 +10,6 @@ import type {
 import type { IProjectReadModel } from '../project/project-read-model-type';
 import type { IPrivateProjectChecker } from '../private-project/privateProjectCheckerType';
 import type {
-    IProjectStore,
     IAccessStore,
     IAccountStore,
     IEvent,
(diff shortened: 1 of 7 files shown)


src/lib/features/personal-dashboard/personal-dashboard-service.ts
```diff
@@ -105,6 +111,12 @@ export class PersonalDashboardService {
         userId: number,
         projectId: string,
     ): Promise<PersonalDashboardProjectDetailsSchema> {
+        if (!(await this.projectStore.hasProject(projectId))) {
```
kwasniew, line 114: I'm not sure about introducing projectStore since it brings coupling to the entire API of this store. What if we could call `getOnboardingStatusForProject(projectId: string)` before the whole Promise.all and let it return null when the status doesn't exist which essentially translates to 404. This will also make `getOnboardingStatusForProject` more resilient.

src/lib/features/project/project-service.ts
```diff
@@ -1546,7 +1546,9 @@ export default class ProjectService {
             updatedAt: project.updatedAt,
             archivedAt: project.archivedAt,
             createdAt: project.createdAt,
-            onboardingStatus,
+            onboardingStatus: onboardingStatus ?? {
+                status: 'onboarding-started',
+            },
```
thomasheartman, line 1551: This method already throws a not found error if the project id doesn't exist, so we can fall back to a default here. Is this the right default, though?

kwasniew (approved):

PR #8364, opened 2024-10-04
chore: make signals visible on event timeline for non admins

nunogois:
https://linear.app/unleash/issue/2-2743/open-the-signal-query-endpoint-to-everyone-not-only-admins

The new signal query endpoint is now open for every Unleash user, not only admins. 

This PR allows non-admins to view signals in the event timeline. It also updates the signals tooltip to be shown to all users, not just admins, under the following assumptions:

- `!signalsSuggestionSeen` - Current user has not dismissed the signals tip
- `isEnterprise()` - Enterprise instance
- `signalsEnabled` - The signals feature flag is enabled
- `!signalsLoading` - Signals have finished loading (avoids flickering)
- `signals.length === 0` - We can't find any signals in the selected timespan


gastonfournier (approved):
LGTM!

PR #8375, opened 2024-10-07
fix: handle project fetching error

thomasheartman:
This change handles cases where fetching project details results in errors. In our sandbox instance, that notably happens for the `..` project, because that ID breaks the API call.

In short, we now check the state of the request, and if we have an error, we show information about that to the user instead:

image


frontend/src/component/personalDashboard/MyProjects.tsx
```diff
@@ -63,127 +65,176 @@ const ActiveProjectDetails: FC<{
...
-                                                size='small'
-                                                sx={{ ml: 'auto' }}
-                                            >
-                                                <LinkIcon
-                                                    titleAccess={`projects/${project.id}`}
+                                        <ListItemButton
+                                            sx={listItemStyle}
+                                            selected={
+                                                project.id === activeProject
+                                            }
```
thomasheartman, line 183:
```suggestion
                                            selected={project.id === activeProject}
```

PR #8376, opened 2024-10-07
fix: fontend dev command to start in sandbox pro environment

Tymek:
Update development command

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -14,7 +14,7 @@
     "dev": "vite",
     "start": "vite",
     "start:prod": "vite build && vite preview",
-    "start:sandbox": "UNLEASH_API=https://sandbox.getunleash.io/ospro yarn run start",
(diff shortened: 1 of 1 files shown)


frontend/package.json
```diff
@@ -14,7 +14,7 @@
     "dev": "vite",
     "start": "vite",
     "start:prod": "vite build && vite preview",
-    "start:sandbox": "UNLEASH_API=https://sandbox.getunleash.io/ospro yarn run start",
+    "start:sandbox": "UNLEASH_API=https://sandbox.getunleash.io/ VITE_TEST_REDIRECT=true UNLEASH_BASE_PATH=/pro/ yarn run start",
```
chriswk, line 17: Shouldn't this include /pro in UNLEASH_API? or does the proxy automatically append UNLEASH_BASE_PATH to UNLEASH_API?

Tymek: I just copied what we have for enterprise sandbox

PR #8378, opened 2024-10-07
Frontend for additional environments

Tymek:
## About the changes
- [flag created](https://app.unleash-hosted.com/hosted/projects/dx/features/purchaseAdditionalEnvironments)
- added item to the menu
- styled banner and dialog
- added dialog
- styled dialog
- added dialog form state management
- added tests

### DIFF
--- a/frontend/src/component/menu/routes.ts
+++ b/frontend/src/component/menu/routes.ts
@@ -267,6 +267,15 @@ export const routes: IRoute[] = [
         component: EnvironmentTable,
         type: 'protected',
         flag: EEA,
+        notFlag: 'purchaseAdditionalEnvironments',
+        menu: { mobile: true, advanced: true },
+    },
+    {
+        path: '/environments',
+        title: 'Environments',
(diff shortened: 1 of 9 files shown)


frontend/src/component/menu/routes.ts
```diff
@@ -267,6 +267,15 @@ export const routes: IRoute[] = [
         component: EnvironmentTable,
         type: 'protected',
         flag: EEA,
+        notFlag: 'purchaseAdditionalEnvironments',
+        menu: { mobile: true, advanced: true },
+    },
+    {
+        path: '/environments',
```
sjaanus, line 274: Will these both pop up for enterprise?

sjaanus (approved):
One comment

frontend/src/component/menu/routes.ts
Tymek: No, because other "path" has `notFlag`. Only 1 item shows up on Enterprise - tested in sandbox.

PR #8379, opened 2024-10-07
chore: better show hide event timeline tooltips

nunogois:
https://linear.app/unleash/issue/2-2753/improve-current-showhide-tooltips-to-be-more-specific-about-the

Improves the "show" and "hide" tooltips of this feature to have a slightly more specific text. This is not just any timeline, this is the **event timeline**.

- "Hide timeline" -> "Hide event timeline"
- "Show timeline" -> "Show event timeline"


daveleek (approved):
Great!

PR #8392, opened 2024-10-08
Ab testing cleanup redirect

melindafekete:
## About the changes
Remove the old AB testing file, add a docusaurus redirect, and update the old link in readme.


website/docusaurus.config.js
```diff
@@ -856,6 +856,10 @@ module.exports = {
                         ],
                         to: '/topics/feature-flag-migration/feature-flag-migration-best-practices',
                     },
+                    {
```
melindafekete, line 859: @nnennandukwe for future reference, docs to docs redirects can be done in this file

melindafekete: Actually also added it in vercel, not sure if we need both

alvinometric: I think in the future we probably want to move every redirect to Vercel. But I'm not sure what the limit is 😅

alvinometric (approved):

PR #8394, opened 2024-10-08
Restructure features documentation

melindafekete:
## About the changes
Reorganize, update, and improve concepts documentation around feature flags.

This PR merges all feature flag-related concepts pages into a single page. This helps the discoverability of all closely related topics and helps tie together these concepts through explanations between the different sections.

Previously:
- [Archived flags](https://docs.getunleash.io/reference/archived-toggles)
- [Feature flag types](https://docs.getunleash.io/reference/feature-toggle-types)
- [Feature flag naming patterns](https://docs.getunleash.io/reference/feature-flag-naming-patterns)
- [Feature flag lifecycle](https://docs.getunleash.io/reference/feature-lifecycle)
- [Dependent features](https://docs.getunleash.io/reference/dependent-features)
- [Feature flags](https://docs.getunleash.io/reference/feature-toggles)

Are now all on:
- [Feature flags](https://unleash-docs-git-melinda-features-docs-reorg-unleash-team.vercel.app/reference/feature-toggles)

Additionally, removed the deprecated [Feature Flag Variants](https://docs.getunleash.io/reference/feature-toggle-variants) from the navigation, now only referenced at the [bottom of the strategy variants](https://unleash-docs-git-melinda-features-docs-reorg-unleash-team.vercel.app/reference/strategy-variants#environment-level-variants) page. Also, I removed the [how to create a naming pattern guide](https://docs.getunleash.io/how-to/how-to-add-feature-flag-naming-patterns), as it covers the same information as the concepts docs, and there is not enough specific content to fill a standalone guide.


website/docs/reference/feature-lifecycle.md
```diff
@@ -2,61 +2,3 @@
 title: Feature Lifecycle
 ---
```
alvinometric, line 4: Is this supposed to be blank?

gastonfournier (commented):
Good job with the redirects!

website/docs/reference/feature-lifecycle.md
melindafekete: Deleted, thanks!

website/docs/reference/events.mdx
```diff
@@ -291,7 +291,7 @@ Generated when you drop all existing tags as part of a configuration import. The
 
 #### `feature-potentially-stale-on`
 
-Generated when Unleash marks a feature flag as potentially stale due to exceeding the expected lifetime of its [feature flag type](/reference/feature-toggle-types). `preData` and `data` are null.
+Generated when Unleash marks a feature flag as potentially stale due to exceeding the expected lifetime of its [feature flag type](/reference/feature-toggles#feature-flag-types). `preData` and `data` are null.
```
alvinometric, line 294:
```suggestion
Generated when Unleash marks a feature flag as potentially stale due to exceeding the expected lifetime of its [feature flag type](/reference/feature-toggles#feature-flag-types). `preData` and `data` are `null`.
```

website/docs/reference/strategy-variants.md
```diff
@@ -131,3 +131,8 @@ To make use of strategy variants, you need to use a compatible client. Client SD
 
 
 If you would like to give feedback on this feature, experience issues or have questions, please feel free to open an issue on [GitHub](https://github.com/Unleash/unleash/).
+
+
+## Environment level variants
+
+[Environment level variants](./feature-toggle-variants) are deprecated. Use strategy variants instead.
```
thomasheartman, line 138: Do we have a deprecation notice on the old variants doc page?

website/docusaurus.config.js
```diff
@@ -356,7 +352,7 @@ module.exports = {
                     },
                     {
                         from: '/advanced/archived_toggles',
-                        to: '/reference/archived-toggles',
+                        to: '/reference/feature-toggles',
```
thomasheartman, line 355: We should probably update all the relevant redirects to target the new page.

It'll be (I think):

```json
from: ['/advanced/archived_toggles', '/reference/archived-toggles']
to: '/reference/feature-toggles'
```

Also, don't we do this server-side now? 🤔 

So this goes for this one and all the other deleted files (I noticed there's only two redirects here, but you've deleted at least 6 files)

thomasheartman: Ah, but I see you've also updated the vercel file, which should take care of it. In that case, are the changes here still necessary?

thomasheartman (approved):
Nice one 🔥 

Good work on getting the OpenAPI changes in too 🙌🏼 

A couple small questions, but no blockers. Solid effort 💯

website/docs/reference/strategy-variants.md
melindafekete: Yes

website/docs/reference/events.mdx
melindafekete: Why do you suggest code formatting for this? Given that it's an unrelated change, and there are many occurrences of it, let's tackle it separately

website/docusaurus.config.js
melindafekete: Yes, I added the new redirects in Vercel, these are some legacy ones. I am not sure if these `'/advanced'` pages are still getting any traffic. Eventually, all redirects should be in Vercel, but I wanted to tackle this separately, as we might not need to move all of them.

> We should probably update all the relevant redirects to target the new page.
I am pointing these legacy pages to the new page too, unless I am missing something 🤔

PR #8398, opened 2024-10-09
chore: remove all deprecated imports of act

thomasheartman:
don't use `act` from `react-dom`. Instead, use act from `react` directly, as advised by the deprecation notice.

This PR fixes all of the deprecated import warnings, updates some testing libraries we use (and tests), and fixes one or two other warnings.

### DIFF
--- a/frontend/src/component/admin/apiToken/CreateApiToken/CreateApiToken.test.tsx
+++ b/frontend/src/component/admin/apiToken/CreateApiToken/CreateApiToken.test.tsx
@@ -39,9 +39,6 @@ test('Enabled new token button when limits, version and permission allow for it'
         permissions,
     });
 
-    const button = await screen.findByText('Create token');
-    expect(button).toBeDisabled();
-
     await waitFor(async () => {
         const button = await screen.findByText('Create token');
         expect(button).not.toBeDisabled();
--- a/frontend/src/component/onboarding/flow/ProjectOnboarding.tsx
+++ b/frontend/src/component/onboarding/flow/ProjectOnboarding.tsx
@@ -120,7 +120,7 @@ export const ProjectOnboarding = ({
 }: IProjectOnboardingProps) => {
     const { project } = useProjectOverview(projectId);
     const isFirstFlagCreated =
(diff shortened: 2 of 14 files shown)


frontend/src/component/admin/apiToken/CreateApiToken/CreateApiToken.test.tsx
```diff
@@ -39,9 +39,6 @@ test('Enabled new token button when limits, version and permission allow for it'
         permissions,
     });
 
-    const button = await screen.findByText('Create token');
-    expect(button).toBeDisabled();
-
```
thomasheartman, line 44: This check was failing now. I suspect that it wasn't caught in the previous version of the dom for some reason. However, I think the important part is the next check, so I'm happy to remove this bit.

frontend/src/component/project/Project/CreateProject/NewCreateProjectForm/CreateProjectDialog.test.tsx
```diff
@@ -27,11 +27,6 @@ test('Enabled new project button when limits, version and permission allow for i
         permissions: [{ permission: CREATE_PROJECT }],
     });
 
-    const button = await screen.findByRole('button', {
-        name: 'Create project',
-    });
-    expect(button).toBeDisabled();
```
thomasheartman, line 33: Same thing as the other one: this isn't the important part of the test.

frontend/src/component/onboarding/flow/ProjectOnboarding.tsx
```diff
@@ -194,8 +194,8 @@ const CreateFlag = ({ projectId, refetchFeatures }: ICreateFlagProps) => {
                 Create a feature flag
             </TitleContainer>
             <Typography>
-                <div>The project currently holds no feature flags.</div>
-                <div>Create one to get started.</div>
+                The project currently holds no feature flags. Create one to get
+                started.
```
thomasheartman, line 198: You can't nest divs within p elements. If you want a separate paragraph, use an extra p element.

kwasniew (approved):

PR #8400, opened 2024-10-09
chore: add OpenAI API key configuration

nunogois:
https://linear.app/unleash/issue/2-2787/add-openai-api-key-to-our-configuration

Adds the OpenAI API key to our configuration and exposes a new `unleashAIAvailable` boolean in our UI config to let our frontend know that we have configured this. This can be used together with our flag to decide whether we should enable our experiment for our users.


sighphyre (approved):
LGTM

PR #8402, opened 2024-10-09
chore:origin middleware flag cleanup

daveleek:
Cleans up the originMiddleware flag and adds a originMiddlewareRequestLogging flag that only logs the request to info if enabled.

(Prevents cluttering OSS/Selfhosted logs)


src/lib/middleware/origin-middleware.ts
```diff
@@ -31,11 +26,15 @@ export const originMiddleware = ({
             const uaLabel = userAgent
                 ? determineIntegrationSource(userAgent)
                 : 'Other';
-            logger.info('API request', {
-                method: req.method,
-                userAgent: req.headers['user-agent'],
-                origin: getFilteredOrigin(req),
-            });
+
+            if (!flagResolver.isEnabled('originMiddlewareRequestLogging')) {
```
nunogois, line 30: Don't you mean this instead? Only log if it's enabled?

```suggestion
            if (flagResolver.isEnabled('originMiddlewareRequestLogging')) {
```

daveleek: Wow great catch, what on earth was I doing here? Confused by considering kill switch perhaps

nunogois (approved):
Other than what I mentioned [here](https://github.com/Unleash/unleash/pull/8413#pullrequestreview-2359517410), this LGTM 👍

PR #8408, opened 2024-10-09
chore: implement an Unleash AI service

nunogois:
https://linear.app/unleash/issue/2-2789/create-an-ai-service

Implements an Unleash AI service with 3 tools:
 - Search flag
 - Create flag
 - Toggle flag

Depends on #8400

### DIFF
--- a/src/lib/features/ai/ai-service.ts
+++ b/src/lib/features/ai/ai-service.ts
@@ -1,0 +1,85 @@
+import type { ChatCompletionMessageParam } from 'openai/resources/chat/completions';
+import type { ChatCompletionRunner } from 'openai/resources/beta/chat/completions';
+import OpenAI from 'openai';
+import type {
+    IAuthRequest,
+    IUnleashConfig,
+    IUnleashServices,
+    Logger,
+} from '../../server-impl';
+import type FeatureToggleService from '../feature-toggle/feature-toggle-service';
+import type { FeatureSearchService } from '../feature-search/feature-search-service';
+import { createFlag } from './tools/create-flag';
+import { toggleFlag } from './tools/toggle-flag';
+import { searchFlag } from './tools/search-flag';
+
+export class AIService {
+    private config: IUnleashConfig;
+
+    private logger: Logger;
+
+    private client: OpenAI | undefined;
+
+    private featureService: FeatureToggleService;
+
+    private featureSearchService: FeatureSearchService;
+
+    constructor(
+        config: IUnleashConfig,
+        {
+            featureToggleService,
+            featureSearchService,
+        }: Pick<
+            IUnleashServices,
+            'featureToggleService' | 'featureSearchService'
+        >,
+    ) {
+        this.config = config;
+        this.logger = config.getLogger('features/ai/ai-service.ts');
+        this.featureService = featureToggleService;
+        this.featureSearchService = featureSearchService;
+    }
+
+    getClient(): OpenAI {
+        if (this.client) {
+            return this.client;
+        }
+
+        const apiKey = this.config.openAIAPIKey;
+        if (!apiKey) {
+            throw new Error(
+                'Unleash AI is unavailable: Missing OpenAI API key',
+            );
+        }
+
+        this.client = new OpenAI({ apiKey });
+        return this.client;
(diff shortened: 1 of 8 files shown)


src/lib/features/ai/tools/create-flag.ts
```diff
@@ -0,0 +1,66 @@
...
+}: createFlagFunctionParams) => {
+    try {
+        const response = await featureService.createFeatureToggle(
+            project,
+            { name, description },
+            auditUser,
+        );
+
+        return response;
+    } catch (error) {
```
sighphyre, line 33: What's the try/catch here for? Debugging?

src/lib/features/ai/ai-service.ts
```diff
@@ -0,0 +1,85 @@
...
+                'Unleash AI is unavailable: Missing OpenAI API key',
+            );
+        }
+
+        this.client = new OpenAI({ apiKey });
+        return this.client;
+    }
+
+    chat(
+        messages: ChatCompletionMessageParam[],
```
sighphyre, line 60: What's the intent here? Does the chat history live on the UI at the moment? Can't quite see how this works behind a load balancer

src/lib/features/ai/tools/search-flag.ts
```diff
@@ -0,0 +1,161 @@
...
+    } catch (error) {
+        return error;
+    }
+};
+
+export const searchFlag = ({
+    featureSearchService,
+    user,
+}: searchFlagOperationalParams): RunnableToolFunctionWithParse<searchFlagParams> => ({
+    type: 'function',
```
sighphyre, line 84: This is pretty cool! How well does this work in practice? Does OpenAI use all of these terms effectively?

src/lib/features/ai/tools/toggle-flag.ts
```diff
@@ -0,0 +1,79 @@
...
+                    description: 'The project the flag belongs to.',
+                },
+                name: { type: 'string', description: 'The name of the flag.' },
+                environment: {
+                    type: 'string',
+                    description: 'The environment in which to toggle the flag',
+                },
+                enabled: {
+                    type: 'boolean',
+                    description: 'The desired enabled status of the flag',
```
sighphyre, line 73: Maybe your intent is to tackle this in the base prompt but I can see this causing confusion with constraints forcing a flag off

sighphyre (approved):
Yeah looks cool!

src/lib/features/ai/tools/create-flag.ts
nunogois: Anything can go wrong when calling this method, so this allows us to forward either the success response or the error message back to the assistant, allowing it to handle both scenarios gracefully.

src/lib/features/ai/ai-service.ts
nunogois: We're still on the early stage of this project where we're mostly just migrating code from the hackathon branch and polishing it a bit. So for now the chat history still lives in the UI. We have another task for later that moves this ownership to the server.

src/lib/features/ai/tools/search-flag.ts
nunogois: In general, based on the hackathon, this worked pretty well, so it should! We can always fine tune if needed.

If we're specifically talking about this `search-flag` method I'm not sure because I haven't tried it yet 😄 It's a new tool to replace the old `get-flag` one. Good news is that we'll find out soon enough whether this works like we expect it to!

src/lib/features/ai/tools/toggle-flag.ts
nunogois: I'm not sure I follow 🤔 If a flag is off because of constraints, that doesn't matter in the perspective of our assistant, because it's only concerned about the admin side of Unleash, right?

I'm curious to know what you have in mind.

src/lib/features/ai/ai-service.ts
nunogois: In the same vein, I'm pretty sure our `toggle-flag` tool is not respecting change requests currently, which is another task we have for later.

PR #8409, opened 2024-10-09
Fix linting error in GithubContributors file

melindafekete:
## About the changes
Fix linting error in GithubContributors file


nunogois (approved):
LGTM!

PR #8413, opened 2024-10-10
chore: flag for webhook domain logging

daveleek:
Introduces a flag for webhook info-logging of invocations against domain. This should be on for our cloud customers but default to off so it doesn't clutter logs of OSS/Selfhosted


nunogois (approved):
I assume these are supposed to be long-lived flags that are always enabled for cloud customers? If so, I wonder if we can find a better way to detect we are running a cloud instance, e.g., some property in UnleashConfig.

PR #8421, opened 2024-10-10
chore: use new signal meta properties in event timeline

nunogois:
https://linear.app/unleash/issue/2-2796/better-signals-integration

Adds support to the following signal payload meta properties:
 - `unleash_title`
 - `unleash_description`
 - `unleash_icon`
 - `unleash_variant`

Follows a logic similar to what we currently have for banners. E.g. [custom icon](https://docs.getunleash.io/reference/banners#custom-icon).

## Call signal endpoints
image
image

## View signals in event timeline
image
image


gastonfournier (approved):
Looks cool to me :P

nunogois:
https://github.com/Unleash/unleash/pull/8421/commits/118bd5abe04f003232ced4f40a2d712b74956125 prefixes these properties with `experimental_` so that they better represent our outlook on them.

PR #8428, opened 2024-10-11
chore: wrapTimer function types

gastonfournier:
This gives us better types for our wrapTimer function.

Maybe the type `(args: any) => any` could also be improved

### DIFF
--- a/src/lib/util/metrics-helper.ts
+++ b/src/lib/util/metrics-helper.ts
@@ -1,4 +1,4 @@
-/* eslint-disable no-param-reassign */
+import type EventEmitter from 'events';
 import timer from './timer';
 
 // wrapTimer keeps track of the timing of a async operation and emits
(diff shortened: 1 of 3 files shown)


nunogois (approved):
I think this could be a simpler approach, where we type our parameters and rely on the implicit type instead of a declared one:

```ts
const wrapTimer = (
    eventBus: EventEmitter,
    eventType: string,
    args: Record<string, unknown> = {},
) => {
    const t = timer.new();
    return (data: unknown) => {
        eventBus.emit(eventType, { ...args, time: t() });
        return data;
    };
};
```

If you prefer to have a type declaration, I would do something like this:

```ts
type WrapTimerFunction = (
    eventBus: EventEmitter,
    eventType: string,
    args?: Record<string, unknown>,
) => (data: unknown) => unknown;

const wrapTimer: WrapTimerFunction = (eventBus, eventType, args = {}) => {
    const t = timer.new();
    return (data) => {
        eventBus.emit(eventType, { ...args, time: t() });
        return data;
    };
};
```

Or maybe type it like this, if you think it's simpler to read:

```ts
type WrapTimerFunctionCallback = (data: unknown) => unknown;

type WrapTimerFunction = (
    eventBus: EventEmitter,
    eventType: string,
    args?: Record<string, unknown>,
) => WrapTimerFunctionCallback;
```

src/lib/util/metrics-helper.ts
```diff
@@ -8,11 +8,14 @@ import timer from './timer';
 // It transparently passes the data to the following .then(<func>)
 //
 // usage: promise.then(wrapTimer(bus, type, { payload: 'ok' }))
```
nunogois, line 10: Since we're touching this file, I also think these comments are outdated.

nunogois (approved):
Nice!

PR #8434, opened 2024-10-11
fix: same site api call with session cookie

### LINKED ISSUE #8029: Unleash-Session Cookie Sent In Headers
### Describe the bug

Hi:)
We are having the bug that I saw this [issue](https://github.com/Unleash/unleash/issues/5086) opened about.
Opening Unlish and the application in the same domain returns a 401 error for the toggle request. Because the  Cookie header has been sent.
I tried to use the customHeaders property to override the value in the cookie header but it doesn't seem to help and it still uses the cookie.
Is this bug known and have you fixed it?
And do you have a suggestion how to deal with it in the meantime
Thanks in advance!!

1. open the unleash ui and log in
2. open in another tab at the same domain' your web application
3. the request to unleash fails due to the unleash-session cookie header 

### Expected behavior

When using unleash token, i dont want the session cookie to be sent as well.

### Logs, error output, etc.

_No response_

### Screenshots

image
image

### Additional context

_No response_

### Unleash version

_No response_

### Subscription type

None

### Hosting type

Self-hosted

React v4

gastonfournier:
## About the changes
When a web app is hosted in the same domain as Unleash UI and the web app uses unleash SDK to make requests to Unleash, the browser automatically includes the cookie in the request headers if:

    - The request URL matches the cookie's Path attribute (which it does in this case).
    - The request is sent to the same domain (which it is, since both apps are under the same domain).

This PR aims to limit the scope of unleashSession for those URLs that are not APIs. This is an alternative to https://github.com/Unleash/unleash/pull/8435

Closes #8029

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -31,6 +31,7 @@ import NotFoundError from './error/notfound-error';
 import { bearerTokenMiddleware } from './middleware/bearer-token-middleware';
 import { auditAccessMiddleware } from './middleware';
 import { originMiddleware } from './middleware/origin-middleware';
+import { unlessHasHeader } from './middleware/unless-has-header-middleware';
 
 export default async function getApp(
     config: IUnleashConfig,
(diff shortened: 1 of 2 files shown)


gastonfournier:
Close in favor of https://github.com/Unleash/unleash/pull/8435

PR #8441, opened 2024-10-14
chore: Unleash AI API hook

nunogois:
https://linear.app/unleash/issue/2-2791/create-a-useaiapi-react-hook

Implements a basic Unleash AI API React hook that fits our initial needs for interacting with this API through our frontend.

Also adds a new nice-to-have script to run the frontend set to the `demo` base path, which matches our Cloud defaults. This way you can run the latest local cloud with the latest local frontend in an easy way.


gastonfournier (approved):
LGTM!

PR #8446, opened 2024-10-14
chore: add a class to handle aggreggation queries

gastonfournier:
## About the changes
We have many aggregation queries that run on a schedule: https://github.com/Unleash/unleash/blob/f63496d47f6808a246732f5ddee9b9136e23eca7/src/lib/metrics.ts#L714-L719

These staticCounters are usually doing db query aggregations that traverse tables and we run all of them in parallel: https://github.com/Unleash/unleash/blob/f63496d47f6808a246732f5ddee9b9136e23eca7/src/lib/metrics.ts#L410-L412

This can add strain to the db. This PR suggests a way of handling these queries in a more structured way, allowing us to run them sequentially (therefore spreading the load): https://github.com/Unleash/unleash/blob/f02fe87835c30d77ff8b6755268e123585820511/src/lib/metrics-gauge.ts#L38-L40

As an additional benefit, we get both the gauge definition and the queries in a single place: https://github.com/Unleash/unleash/blob/f02fe87835c30d77ff8b6755268e123585820511/src/lib/metrics.ts#L131-L141

This PR only tackles 1 metric, and it only focuses on gauges to gather initial feedback. The plan is to migrate these metrics and eventually incorporate more types (e.g. counters)


src/lib/metrics.ts
```diff
@@ -715,7 +720,6 @@ export default class MetricsMonitor {
             collectStaticCounters.bind(this),
             hoursToMilliseconds(2),
             'collectStaticCounters',
-            0, // no jitter
```
gastonfournier, line 718: Allowing these statistics to run with jitter as they are not needed at startup

src/lib/metrics.ts
```diff
@@ -125,11 +127,19 @@ export default class MetricsMonitor {
...
+        dbMetrics.registerGaugeDbMetric({
             name: 'max_feature_strategies',
             help: 'Maximum number of strategies in one feature',
             labelNames: ['feature'],
+            query: () =>
+                stores.featureStrategiesReadModel.getMaxFeatureStrategies(),
+            map: (result) => ({
+                count: result.count,
+                labels: { feature: result.feature },
+            }),
```
gastonfournier, line 140: Gauge and query are now declared together

src/lib/metrics.ts
```diff
@@ -408,7 +419,6 @@ export default class MetricsMonitor {
                     instanceOnboardingMetrics,
                     projectsOnboardingMetrics,
                 ] = await Promise.all([
```
sjaanus, line 421: We do not need to run these in parallel. I would be to get rid of promise all.

src/lib/metrics-gauge.ts
```diff
@@ -0,0 +1,50 @@
...
+    labelNames: string[];
+    query: Query<T>;
+    map: MapResult<T>;
+};
+
+export class DbMetricsMonitor {
+    private tasks: (() => Promise<void>)[] = [];
+    private gauges: Record<string, Gauge<string>> = {};
+
+    constructor() {}
```
nunogois, line 22: Are we planning on doing something with the constructor? It not, can we omit it?

src/lib/metrics-gauge.ts
```diff
@@ -0,0 +1,50 @@
+import { createGauge, type Gauge } from './util/metrics';
+
+type RestrictedRecord<T extends string[]> = Record<T[number], string>;
```
nunogois, line 3:
```suggestion
type RestrictedRecord<T extends readonly string[]> = Record<T[number], string>;
```

src/lib/metrics.ts
gastonfournier: Exactly, we are going to do that gradually. When we register a new metric into this new implementation, the new implementation will run them sequentially

src/lib/metrics-gauge.ts
gastonfournier: I went back and forth with adding some dependencies to it (like timer) and it just remained there... I think this is just details we can fix later

src/lib/metrics-gauge.ts
```diff
@@ -0,0 +1,50 @@
...
+    name: string;
+    help: string;
+    labelNames: string[];
+    query: Query<T>;
+    map: MapResult<T>;
+};
+
+export class DbMetricsMonitor {
+    private tasks: (() => Promise<void>)[] = [];
+    private gauges: Record<string, Gauge<string>> = {};
```
nunogois, line 20: Up to you but I would maybe consider making these a `Set` and a `Map` respectively.

nunogois (approved):
Super nice!

PR #8447, opened 2024-10-15
Add environment types environment order

Tymek:


frontend/src/component/common/GeneralSelect/GeneralSelect.tsx
```diff
@@ -51,6 +51,7 @@ function GeneralSelect<T extends string = string>({
     classes,
     fullWidth,
     visuallyHideLabel,
+    labelId,
```
Tymek, line 54: fixing accessibility and tests - `labelId` is a prop from MUI

sjaanus (approved):

PR #8450, opened 2024-10-15
chore: handle aggregation queries sequentially follow up

gastonfournier:
This is a follow up on https://github.com/Unleash/unleash/pull/8446
- migrating some metrics (check the commits)
- moving event listener registration outside metric calculation (every 2 hs we are adding new listeners)
- adding jitter to metric refresh
- manually executing some of the tasks immediately to avoid changing the test logic (with jitter now we'd have to force or wait until the first execution of the scheduled task that collect the metrics). We might decide to change this later after the migration
- try catch the execution of the task
- register metrics now "just registers" and returns functions for refreshing metrics. In the future, we'll return an object with the ability of independently updating specific metrics so we can tweak the frequency of updates

### DIFF
--- a/src/lib/features/instance-stats/instance-stats-service.ts
+++ b/src/lib/features/instance-stats/instance-stats-service.ts
@@ -29,7 +29,6 @@ import { CUSTOM_ROOT_ROLE_TYPE } from '../../util';
 import type { GetActiveUsers } from './getActiveUsers';
 import type { ProjectModeCount } from '../project/project-store';
 import type { GetProductionChanges } from './getProductionChanges';
-import { DbMetricsMonitor } from '../../metrics-gauge';
 
 export type TimeRange = 'allTime' | '30d' | '7d';
 
@@ -110,13 +109,11 @@ export class InstanceStatsService {
 
     private appCount?: Partial<{ [key in TimeRange]: number }>;
 
-    private getActiveUsers: GetActiveUsers;
+    getActiveUsers: GetActiveUsers;
 
-    private getProductionChanges: GetProductionChanges;
+    getProductionChanges: GetProductionChanges;
 
     private featureStrategiesReadModel: IFeatureStrategiesReadModel;
-
-    dbMetrics: DbMetricsMonitor;
 
     constructor(
         {
@@ -181,18 +178,16 @@ export class InstanceStatsService {
         this.clientMetricsStore = clientMetricsStoreV2;
         this.flagResolver = flagResolver;
         this.featureStrategiesReadModel = featureStrategiesReadModel;
-        this.dbMetrics = new DbMetricsMonitor({ getLogger });
-    }
-
-    async fromPrometheus(
-        name: string,
-        labels?: Record<string, string | number>,
-    ): Promise<number> {
-        return (await this.dbMetrics.findValue(name, labels)) ?? 0;
     }
 
     getProjectModeCount(): Promise<ProjectModeCount[]> {
         return this.projectStore.getProjectModeCounts();
+    }
+
+    getToggleCount(): Promise<number> {
+        return this.featureToggleStore.count({
+            archived: false,
+        });
     }
 
     getArchivedToggleCount(): Promise<number> {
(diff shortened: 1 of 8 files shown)


src/lib/metrics-gauge.ts
```diff
@@ -1,11 +1,14 @@
...
 import { createGauge, type Gauge } from './util/metrics';
 
 type RestrictedRecord<T extends readonly string[]> = Record<T[number], string>;
 type Query<R> = () => Promise<R | undefined | null>;
-type MapResult<R> = (result: R) => {
+type MetricValue<R> = {
     count: number;
     labels: RestrictedRecord<GaugeDefinition<R>['labelNames']>;
 };
+type MapResult<R> = (result: R) => MetricValue<R> | MetricValue<R>[];
```
gastonfournier, line 11: A single task can output different counts for different labels. This is useful for cases like `client_apps_total` at line 40

src/lib/metrics.ts
```diff
@@ -117,15 +117,29 @@ export default class MetricsMonitor {
             help: 'Number of times a feature flag has been used',
             labelNames: ['toggle', 'active', 'appName'],
         });
-        const featureFlagsTotal = createGauge({
+
+        // schedule and execute immediately
+        await dbMetrics.registerGaugeDbMetric({
```
gastonfournier, line 122: These tasks need to execute immediately because the tests expect that. I'd argue that some of these can be delayed, but maybe they are important because of other components depending on these. So I'm delaying this decision until we migrated all gauges to dbMetrics

src/lib/metrics.ts
```diff
@@ -39,376 +39,743 @@ import type { SchedulerService } from './services';
...
+    });
 
-        async function collectStaticCounters() {
+    return {
+        collectDbMetrics: dbMetrics.refreshDbMetrics,
+        collectStaticCounters: async () => {
             try {
-                dbMetrics.refreshDbMetrics();
-
-                const stats = await instanceStatsService.getStats();
```
gastonfournier, line 409: getStats was triggering many queries in parallel, some of which were not used in prometheus.

src/lib/metrics.ts
```diff
@@ -512,16 +858,6 @@ export default class MetricsMonitor {
                 legacyTokensActive.reset();
                 legacyTokensActive.set(deprecatedTokens.activeLegacyTokens);
 
-                if (maxEnvironmentStrategies) {
```
gastonfournier, line 515: This got migrated to: https://github.com/Unleash/unleash/pull/8450/files#diff-0610ce1595d1dcf150ca2efb2f0c5806d082f5bc6b8429a437970c7d60171cbdR181-R194 where both the gauge definition and refresh query are together

src/lib/metrics.ts
```diff
@@ -460,30 +827,6 @@ export default class MetricsMonitor {
...
-
-                eventBus.on(
-                    events.EXCEEDS_LIMIT,
-                    ({
-                        resource,
-                        limit,
-                    }: { resource: string; limit: number }) => {
-                        exceedsLimitErrorCounter.increment({ resource, limit });
-                    },
-                );
```
gastonfournier, line 485: Moved up to line 485 on the right side. Now lives inside `registerPrometheusMetrics` instead of `startMonitoring`

src/lib/metrics.ts
```diff
@@ -714,367 +1062,56 @@ export default class MetricsMonitor {
...
-            mapFeaturesForClientDuration.observe(duration);
-        });
+        const { collectStaticCounters, collectDbMetrics } =
+            registerPrometheusMetrics(
+                config,
+                stores,
+                version,
+                eventBus,
+                instanceStatsService,
+            );
```
gastonfournier, line 1086: Start monitoring now calls helper functions that register prometheus metrics and then adds schedules to frequently update these metrics

src/lib/features/instance-stats/instance-stats-service.ts
```diff
@@ -338,11 +369,12 @@ export class InstanceStatsService {
             this.clientInstanceStore.getDistinctApplicationsCount(30),
             this.clientInstanceStore.getDistinctApplicationsCount(),
         ]);
-        return {
+        this.appCount = {
```
gastonfournier, line 372: Store the result, so it can be used internally (line 261)

PR #8451, opened 2024-10-15
chore: fix invalid dom nesting

thomasheartman:
This PR fixes all `invalidDomNesting` errors we're getting in our tests. The culprit was the `Badge` icon we use, which wrapped its children in a div. When that's used as a child of a `p` tag, that'd cause this to trigger. 

What I've done is to change the wrapping element to a span instead. The Badge itself uses an `display: inline-flex`, so divs and spans should be treated the same, meaning there's no visual change for this.


FredrikOseberg (approved):
LGTM

PR #8453, opened 2024-10-15
fix: group and user has at least 1 project owner counter respects multirole project groups

daveleek:
Sometimes when a project had an owner and a group that has Owner role together with other roles you couldn't remove the owner access for the group.

The workaround were to remove the other group roles first, then you could remove the owner role when it was the only role left for the group in the project (given that any other group or user had owner role)

The reason was that we were checking an aggregated `roleId` property on the group roles returned object, which is set by a reduce function to be the first row of group ids returned from the db query. If any other role than 'Owner' was set as roleId, that group wasn't counted by the atleastOneOwner verification and if there was only 1 other 'Owner' on the project then the operation wasn't allowed.

Fixed this by removing the property altogether and check the roles property instead, which contains all roles the group has


gastonfournier (approved):
Well done! Tricky issue indeed!

PR #8454, opened 2024-10-15
feat: ability to configure when users are considered inactive

gastonfournier:
Give the ability to change when users are considered inactive via an environment variable `USER_INACTIVITY_THRESHOLD_IN_DAYS` or configuration option: `userInactivityThresholdInDays`. Default remains 180 days


nunogois (approved):
LGTM

PR #8456, opened 2024-10-15
chore: make the Unleash AI chat resizable

nunogois:
https://linear.app/unleash/issue/2-2840/make-the-unleash-ai-chat-window-resizable

This PR makes the Unleash AI chat resizable, providing users with a flexible way to adjust the chat window's size.

Implements a reusable `Resizable` wrapper component that allows configuration of:
 - Minimum, maximum, and default sizes.
 - Customizable resize handlers for each edge and corner of the container.
 - Optional resize event callbacks.

Double-clicking any resize handler maximizes the container along that axis (or both, if it's a corner). If the container is already maximized, double-clicking again will revert it to the default size.

### DIFF
--- a/frontend/src/component/ai/AIChat.tsx
+++ b/frontend/src/component/ai/AIChat.tsx
@@ -13,6 +13,7 @@ import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 import { AIChatInput } from './AIChatInput';
 import { AIChatMessage } from './AIChatMessage';
 import { AIChatHeader } from './AIChatHeader';
+import { Resizable } from 'component/common/Resizable/Resizable';
 
 const StyledAIIconContainer = styled('div')(({ theme }) => ({
     position: 'fixed',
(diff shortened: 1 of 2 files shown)


frontend/src/component/ai/AIChat.tsx
```diff
@@ -134,24 +142,32 @@ export const AIChat = () => {
...
-                    </AIChatMessage>
-                    {messages.map(({ role, content }, index) => (
-                        <AIChatMessage key={index} from={role}>
-                            {content}
+            <StyledResizable
+                handlers={['top-left', 'top', 'left']}
+                minSize={{ width: '270px', height: '200px' }}
+                maxSize={{ width: '90vw', height: '90vh' }}
+                defaultSize={{ width: '320px', height: '450px' }}
+                onResize={scrollToEnd}
```
daveleek, line 150: I can see cases where this would be annoying, ie you scroll up in history and then resize to read something without too many line breaks. But not sure the need would arise in here

daveleek (approved):
Nice, weird that there's no good component for this already in MUI

frontend/src/component/ai/AIChat.tsx
nunogois: If you try this locally I think you'll understand why I did this. If we don't do something like this, whenever you resize, the conversation content container will anchor to the top left.

nunogois: I'll keep it for now but we can remove this behavior if we get feedback about it.

nunogois:
> Nice, weird that there's no good component for this already in MUI

Honestly, I didn’t check MUI before starting this implementation, but from what I can tell, it doesn’t seem like there’s a built-in component for this. The resizing behavior we’re going for is a bit niche, so it’s not something you see commonly implemented in web UI libraries.

PR #8465, opened 2024-10-16
chore: silence remaining front end test warnings

thomasheartman:
This silences front end test warnings, errors, and logs that we don't care about. The
reason we don't care is that:
- we won't fix
- it's test-specific, doesn't appear to happen in real life

And it clogs the test logs.

### DIFF
--- a/frontend/src/setupTests.ts
+++ b/frontend/src/setupTests.ts
@@ -15,6 +15,14 @@ if (!window.ResizeObserver) {
 
 process.env.TZ = 'UTC';
 
+const messagesToIgnore = [
+    'Warning: An update to %s inside a test was not wrapped in act',
+    '[MSW] Found a redundant usage of query parameters in the request handler URL for',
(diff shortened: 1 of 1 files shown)


frontend/src/setupTests.ts
```diff
@@ -15,19 +15,40 @@ if (!window.ResizeObserver) {
...
+    'Warning: An update to %s inside a test was not wrapped in act',
+    "Failed to create chart: can't acquire context from the given item",
+];
+
+const warningsToIgnore = [
+    '[MSW] Found a redundant usage of query parameters in the request handler URL for',
+    'MUI: You have provided an out-of-range value',
+];
+
+const logsToIgnore = ['An exception was caught and handled.'];
```
thomasheartman, line 28: We *can* probably put them all in a single list, when I think about it. It's not like you're gonna get the same message as a log and as a warning, is it?

kwasniew (approved):

PR #8466, opened 2024-10-16
chore: adapt UI to server-side Unleash AI chat ownership

nunogois:
https://linear.app/unleash/issue/2-2847/adapt-unleash-ai-chat-logic-to-new-server-side-chat-ownership-logic

Adapts the Unleash AI chat logic on the UI to the new server-side chat ownership logic.

### DIFF
--- a/frontend/src/component/ai/AIChat.tsx
+++ b/frontend/src/component/ai/AIChat.tsx
@@ -15,9 +15,9 @@ import { AIChatMessage } from './AIChatMessage';
 import { AIChatHeader } from './AIChatHeader';
 import { Resizable } from 'component/common/Resizable/Resizable';
 
-const AI_LOADING_MESSAGE = {
+const AI_ERROR_MESSAGE = {
     role: 'assistant',
-    content: '_Unleash AI is typing..._',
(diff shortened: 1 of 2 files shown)


frontend/src/component/ai/AIChat.tsx
```diff
@@ -106,26 +104,34 @@ export const AIChat = () => {
...
+                { role: 'user', content },
+            ]);
+            const { messages: newMessages } = await chat(content);
             mutate(() => true);
             setMessages(newMessages);
-            setLoading(false);
         } catch (error: unknown) {
+            setMessages((currentMessages) => [
+                ...currentMessages,
+                AI_ERROR_MESSAGE,
```
daveleek, line 122: Should we log this into the chat messages table as well? It could be weird since it's coming from Unleash and not the AI integration nor the user, but if we're presenting it to the user we might want to have it present in the log as well

daveleek: If Unleash is down that will be hard to do, is this handled server side as well?

daveleek (approved):
LGTM!

chriswk (approved):
Yeah, looks good.

sighphyre (approved):
LGTM

PR #8469, opened 2024-10-17
feat: release plans nav menu item and icon suggestion

daveleek:
Adds a `Release management` item to the nav menu and an empty releasemgmt page component.
Temporarily suggests the MUI `Launch` icon as the release mgmt link icon.

image

### DIFF
--- a/frontend/src/component/menu/routes.ts
+++ b/frontend/src/component/menu/routes.ts
@@ -48,6 +48,7 @@ import { Application } from 'component/application/Application';
 import { Signals } from 'component/signals/Signals';
 import { LazyCreateProject } from '../project/Project/CreateProject/LazyCreateProject';
 import { PersonalDashboard } from '../personalDashboard/PersonalDashboard';
+import { ReleaseManagement } from 'component/releases/ReleaseManagement';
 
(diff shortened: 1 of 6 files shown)


gardleopard:
Is this behind a feature flag?

frontend/src/component/menu/routes.ts
```diff
@@ -246,6 +247,15 @@ export const routes: IRoute[] = [
         type: 'protected',
         menu: { mobile: true, advanced: true },
     },
+    {
+        path: '/releases-management',
+        title: 'Release management',
+        component: ReleaseManagement,
+        type: 'protected',
+        menu: { advanced: true, mode: ['enterprise'] },
+        flag: 'releasePlans',
```
daveleek, line 256: This @gardleopard makes sure it's shown only when the specified flag is 'true'
- Another interesting side effect, setting that flag to true negates any enterprise check for the link, so would show up in both OSS and Pro, but the route will still yield a 404 in that case.
Still, it's something we might want to look at (but I feel that's beyond the scope of this PR, I'm making a bug task for it though)

gardleopard (approved):
LGTM

PR #8470, opened 2024-10-17
chore: patch text when trying to add a root role group

sighphyre:
This changes the warning on adding group root roles to cover all roles rather than just Editor or Admin. This got opened to Viewers in a 2K line monster PR that was mostly refactor so I'm assuming this was an accident but it's not a dangerous accident

## Discussion

Okay so why change the warning rather than change the code?

Two reasons.
1) This has been like this for a year and a half. It was changed before the feature entered GA, so users are probably used to it by now. Seems rude to take things away and it's harmless to keep it
2) It's consistent with everything else to have all 3 roles displayed and removes an edge case in the code


chriswk (approved):
LGTM

sighphyre:
> I remember seeing something like this before. I can never fully understand why you can't assign groups with root roles to projects, but the text makes sense to me.

I'd have to check the backend code, it's been a while but iirc, the rather loose reasoning is that there's no point. Admins and Editors already have everything a group could grant so this adds confusion without benefit

PR #8478, opened 2024-10-18
chore: scroll-related UX adjustments in the Unleash AI chat

nunogois:
https://linear.app/unleash/issue/2-2857/make-some-scroll-related-ux-adjustments-to-the-unleash-ai-chat

Introduces scroll-related UX enhancements to the Unleash AI chat, providing a smoother and more refined user experience.

### DIFF
--- a/frontend/src/component/ai/AIChat.tsx
+++ b/frontend/src/component/ai/AIChat.tsx
@@ -20,6 +20,10 @@ const AI_ERROR_MESSAGE = {
     content: `I'm sorry, I'm having trouble understanding you right now. I've reported the issue to the team. Please try again later.`,
 } as const;
 
+type ScrollOptions = ScrollIntoViewOptions & {
+    onlyIfAtEnd?: boolean;
+};
+
 const StyledAIIconContainer = styled('div')(({ theme }) => ({
     position: 'fixed',
     bottom: 20,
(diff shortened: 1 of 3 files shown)


frontend/src/component/ai/AIChat.tsx
```diff
@@ -88,22 +92,45 @@ export const AIChat = () => {
...
+            { threshold: 1.0 },
+        );
+
+        const target = chatEndRef.current;
+        if (target) {
+            intersectionObserver.observe(target);
+        }
+
+        return () => {
+            if (target) {
```
gastonfournier, line 124: Should this be like this? target will get a reference to the target and the function will just keep it referenced
```suggestion
            if (chatEndRef.current) {
```

frontend/src/component/ai/AIChat.tsx
```diff
@@ -153,7 +180,7 @@ export const AIChat = () => {
                 minSize={{ width: '270px', height: '200px' }}
                 maxSize={{ width: '90vw', height: '90vh' }}
                 defaultSize={{ width: '320px', height: '450px' }}
-                onResize={scrollToEnd}
+                onResize={() => scrollToEnd({ onlyIfAtEnd: true })}
```
gastonfournier, line 183: Should we always scroll to end when resizing? I mean if you scroll up on a conversation and then resize...

nunogois: That's why we have `onlyIfAtEnd: true`. This is actually an improvement on: https://github.com/Unleash/unleash/pull/8456#discussion_r1802461279

frontend/src/component/ai/AIChat.tsx
nunogois: Good point! I simplified this to always use the ref: https://github.com/Unleash/unleash/pull/8478/commits/8d473e1ae504d8c8e1f7c1fb258e01a31897c464

gastonfournier (approved):
LGTM!

PR #8479, opened 2024-10-18
fix: don't allow . or .. in feature url

Tymek:
We do some validation on flag names, but there's some cases that slip through. These are some cases that we should handle better.

With `..` as a name, you can't go into the flag in Unleash and you can't activate any environments because the it is interpreted as "go up a level".

### DIFF
--- a/src/lib/routes/util.ts
+++ b/src/lib/routes/util.ts
@@ -13,7 +13,11 @@ export const customJoi = joi.extend((j) => ({
     },
     validate(value, helpers) {
         // Base validation regardless of the rules applied
-        if (encodeURIComponent(value) !== value) {
+        if (
+            encodeURIComponent(value) !== value ||
+            value === '..' ||
+            value === '.'
+        ) {
             // Generate an error, state and options need to be passed
             return { value, errors: helpers.error('isUrlFriendly.base') };
         }
--- a/src/lib/schema/feature-schema.test.ts
+++ b/src/lib/schema/feature-schema.test.ts
@@ -9,7 +9,29 @@ test('should require URL firendly name', () => {
     };
 
     const { error } = featureSchema.validate(toggle);
-    expect(error.details[0].message).toEqual('"name" must be URL friendly');
+    expect(error?.details[0].message).toEqual('"name" must be URL friendly');
+});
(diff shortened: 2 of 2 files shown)


src/lib/routes/util.ts
Tymek, line 1: this change applies to features, tags, projects, context fields and addons

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureTogglesHeader/hooks/useFeatureCreatedFeedback.ts
thomasheartman, line 1: Why is this part of this PR? Seems unrelated. I don't have the context to understand what this is about, I think.

src/lib/routes/util.ts
```diff
@@ -13,7 +13,11 @@ export const customJoi = joi.extend((j) => ({
     },
     validate(value, helpers) {
         // Base validation regardless of the rules applied
-        if (encodeURIComponent(value) !== value) {
+        if (
+            encodeURIComponent(value) !== value ||
+            value === '..' ||
+            value === '.'
```
thomasheartman, line 19: Yeh, this is good. Are there other cases we might have missed? 

I checked the docs for this function, and there's also `- _ ! ~ * ' ( )`:

image

Do we want to allow them? I guess there's no harm in it because they don't break urls the same way. Sure, let's go.

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureTogglesHeader/hooks/useFeatureCreatedFeedback.ts
Tymek: looks like I didn't start from main

Tymek: I'll fix it

thomasheartman (approved):

PR #8484, opened 2024-10-18
chore: Prometheus metrics refactor

gastonfournier:
Migrate some prometheus metrics to use the new and sequential metric updater


src/lib/metrics-gauge.ts
```diff
@@ -37,17 +37,39 @@ export class DbMetricsMonitor {
...
+                .filter((r) => !Number.isInteger(r.value))
+                .forEach((r) => {
+                    this.log.warn(
+                        `Invalid value for ${definition.name}: ${r.value}. Value must be an integer.`,
+                    );
+                });
+            return resultArray.filter((r) => Number.isInteger(r.value));
+        }
+        return [];
+    }
```
gastonfournier, line 60: Encapsulate transformation from the query result (which can either be a single result or an array), into an array of valid metrics for prometheus

src/lib/metrics-gauge.ts
```diff
@@ -14,7 +14,7 @@ type MapResult<R, L extends string> = (
 type GaugeDefinition<T, L extends string> = {
     name: string;
     help: string;
-    labelNames: L[];
+    labelNames?: L[];
```
gastonfournier, line 17: Support gauges without labels

chriswk (approved):
Nice

PR #8492, opened 2024-10-21
fix: handle loading states for project details for a single project

thomasheartman:
This PR updates the use of references on the project details page to handle the loading state for a single project.

Now, if a project is loading, it'll show skeleton loaders for the relevant boxes:

image

I've also updated the state type we use for this to be more accurate. However, it feels like it's getting kinda messy they way we handle it now. I wouldn't mind refactoring the output of the hook into something more like this data type to make handling more explicit. (Shamelessly stolen from Elm, but iykyk)

```ts
type RemoteData<T> = 
  | { state: 'error', error: Error } 
  | { state: 'loading' } 
  | { state: 'success', data: T } 
```

Also, I had some issues with applying the `data-loading` attribute to components, where it seemed to work in some cases, but not in others. For instance, for `DataError` and `ContactAdmins`, it didn't work just setting `data-loading` on the component invocation in `MyProjects`, so I wrapped them both in data-loading divs instead. But happy to hear if there's a trick I missed.

After refactoring: 
image

### DIFF
--- a/frontend/src/component/personalDashboard/MyProjects.tsx
+++ b/frontend/src/component/personalDashboard/MyProjects.tsx
@@ -122,7 +122,10 @@ const ProjectListItem: FC<{
     );
 };
 
-type MyProjectsState = 'no projects' | 'projects' | 'projects with error';
+type MyProjectsState =
+    | 'no projects'
+    | 'projects'
+    | 'projects with error or loading';
 
 export const MyProjects = forwardRef<
     HTMLDivElement,
@@ -149,7 +152,7 @@ export const MyProjects = forwardRef<
         const state: MyProjectsState = projects.length
             ? personalDashboardProjectDetails
                 ? 'projects'
-                : 'projects with error'
+                : 'projects with error or loading'
             : 'no projects';
 
         const activeProjectStage =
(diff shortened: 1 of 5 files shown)


frontend/src/component/personalDashboard/MyProjects.tsx
```diff
@@ -149,7 +152,7 @@ export const MyProjects = forwardRef<
         const state: MyProjectsState = projects.length
             ? personalDashboardProjectDetails
                 ? 'projects'
-                : 'projects with error'
+                : 'projects with error or loading'
```
kwasniew, line 155: so does it mean that on error we're stuck with the skeleton?

kwasniew (approved):
I wanted to combine state specific data myself when it starts to hurt me. If you feel like we're at this stage and you can timebox it to a resonable timeframe then why not. Otherwise I'd clean it up with the next major change in this area. I've shifted data-loading myself a few time it didn't work as expected :)

frontend/src/component/personalDashboard/MyProjects.tsx
thomasheartman: Nah. If we get an error the use loading ref is false, so we're good.

kwasniew (approved):

PR #8495, opened 2024-10-21
fix: break words first, break all second

kwasniew:
## About the changes

For the events markdown we want to prioritize breaking on words, but for very long strings we want to break on anything.
Screenshot 2024-10-21 at 14 47 21


thomasheartman (approved):
yesssss

PR #8506, opened 2024-10-22
chore: add 0BSD license

gastonfournier:
Seems to be very permissive and close to ISC: https://opensource.org/license/0bsd


chriswk (approved):
LGTM

PR #8518, opened 2024-10-23
feat: rework application overview db query

sjaanus:
Previously we were returning all instance names from database, but now we count them in database.

### DIFF
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -346,7 +346,6 @@ export default class ClientApplicationsStore
             .leftJoin('metrics as m', 'm.app_name', 'ca.app_name')
             .leftJoin('instances as i', 'i.environment', 'm.environment')
             .orderBy('m.environment', 'asc');
-        console.log(query.toQuery());
         const rows = await query;
         stopTimer();
         if (!rows.length) {
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -313,29 +313,40 @@ export default class ClientApplicationsStore
                     ),
                 ])
                     .from('client_metrics_env as cme')
+                    .where('cme.app_name', appName)
                     .leftJoin('features as f', 'f.name', 'cme.feature_name')
                     .groupBy('cme.app_name', 'cme.environment', 'f.project');
+            })
+            .with('instances', (qb) => {
+                qb.select([
+                    'ci.app_name',
+                    'ci.environment',
+                    this.db.raw(
+                        'COUNT(DISTINCT ci.instance_id) as unique_instance_count',
+                    ),
+                    this.db.raw(
+                        'ARRAY_AGG(DISTINCT ci.sdk_version) as sdk_versions',
+                    ),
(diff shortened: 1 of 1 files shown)


src/lib/db/client-applications-store.ts
```diff
@@ -313,29 +313,40 @@ export default class ClientApplicationsStore
                     ),
                 ])
                     .from('client_metrics_env as cme')
+                    .where('cme.app_name', appName)
```
sjaanus, line 316: Extra where clause to reduce query scope

src/lib/db/client-applications-store.ts
```diff
@@ -313,29 +313,40 @@ export default class ClientApplicationsStore
...
                     .leftJoin('features as f', 'f.name', 'cme.feature_name')
                     .groupBy('cme.app_name', 'cme.environment', 'f.project');
             })
+            .with('instances', (qb) => {
+                qb.select([
+                    'ci.app_name',
+                    'ci.environment',
+                    this.db.raw(
```
sjaanus, line 324: Now we count in memory, instead of joining and returning all.

src/lib/db/client-applications-store.ts
```diff
@@ -313,29 +313,40 @@ export default class ClientApplicationsStore
...
+                        'COUNT(DISTINCT ci.instance_id) as unique_instance_count',
+                    ),
+                    this.db.raw(
```
sjaanus, line 327: We also count these in database.

src/lib/db/client-applications-store.ts
```diff
@@ -313,29 +313,40 @@ export default class ClientApplicationsStore
...
+                        'ARRAY_AGG(DISTINCT ci.sdk_version) as sdk_versions',
+                    ),
+                    this.db.raw('MAX(ci.last_seen) as latest_last_seen'),
```
sjaanus, line 330: Also last seen, max will be taken, instead of doing it in memory.

src/lib/db/client-applications-store.ts
```diff
@@ -397,25 +405,14 @@ export default class ClientApplicationsStore
                 };
                 acc.push(env);
             } else {
-                if (instance_id) {
```
sjaanus, line 400: Got rid of all of this logic done in memory

src/lib/db/client-applications-store.ts
```diff
@@ -313,29 +313,40 @@ export default class ClientApplicationsStore
...
-                    '=',
-                    'm.environment',
-                );
-            })
-            .where('a.app_name', appName)
+            .from('client_applications as ca')
+            .leftJoin('metrics as m', 'm.app_name', 'ca.app_name')
+            .leftJoin('instances as i', 'i.environment', 'm.environment')
             .orderBy('m.environment', 'asc');
+        console.log(query.toQuery());
```
gastonfournier, line 349:
```suggestion

```

gastonfournier (approved):
This looks great, can we validate this in sandbox before rolling out? Maybe checking prometheus metrics before and after the change or just checking in Grafana

PR #8529, opened 2024-10-24
chore: add migration for release plans

daveleek:
Adds release plans migration.
- Creates release_plan_definitions
- Creates milestones
- Creates milestone_strategies
- Creates indexes on FK of milestones and milestone_strategies tables

Adds a down migration that removes indexes, columns, and tables in an order that addresses inter-dependencies

### DIFF
--- a/src/migrations/20241024062521-add-release-plans-milestones-strategies.js
+++ b/src/migrations/20241024062521-add-release-plans-milestones-strategies.js
@@ -1,0 +1,62 @@
+exports.up = function(db, cb) {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS release_plan_definitions
+        (
+            id TEXT PRIMARY KEY NOT NULL,
+            discriminator TEXT NOT NULL,
+            name TEXT NOT NULL,
+            description TEXT,
+            feature_name TEXT,
+			environment TEXT,
+            created_by_user_id INTEGER NOT NULL REFERENCES users(id),
+            created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+            CONSTRAINT release_plan_definitions_discriminator_values
+                CHECK (discriminator IN ('plan', 'template')),
+            CONSTRAINT feature_environments_fkey FOREIGN KEY (environment, feature_name)
+                REFERENCES feature_environments(environment, feature_name)
+                ON DELETE CASCADE
+        );
+
+        CREATE TABLE IF NOT EXISTS milestones
+        (
+            id SERIAL PRIMARY KEY NOT NULL,
+            name TEXT NOT NULL,
+            sort_order INTEGER NOT NULL,
+            release_plan_definition_id TEXT NOT NULL REFERENCES release_plan_definitions(id) ON DELETE CASCADE
+        );
+
+        CREATE INDEX IF NOT EXISTS idx_milestones_release_plan_definition_id ON milestones (release_plan_definition_id);
+
+        ALTER TABLE IF EXISTS release_plan_definitions
+            ADD COLUMN IF NOT EXISTS active_milestone_id SERIAL REFERENCES milestones(id);
+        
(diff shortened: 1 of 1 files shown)


src/migrations/20241024062521-add-release-plans-milestones-strategies.js
```diff
@@ -0,0 +1,62 @@
...
+            id SERIAL PRIMARY KEY NOT NULL,
+            name TEXT NOT NULL,
+            sort_order INTEGER NOT NULL,
+            release_plan_definition_id TEXT NOT NULL REFERENCES release_plan_definitions(id) ON DELETE CASCADE
+        );
+
+        CREATE INDEX IF NOT EXISTS idx_milestones_release_plan_definition_id ON milestones (release_plan_definition_id);
+
+        ALTER TABLE IF EXISTS release_plan_definitions
+            ADD COLUMN IF NOT EXISTS active_milestone_id SERIAL REFERENCES milestones(id);
```
chriswk, line 32: Certainly this is an INT (you don't want a separate sequence generated here)

src/migrations/20241024062521-add-release-plans-milestones-strategies.js
```diff
@@ -0,0 +1,62 @@
...
+        
+        CREATE TABLE IF NOT EXISTS milestone_strategies
+        (
+            id SERIAL PRIMARY KEY NOT NULL,
```
chriswk, line 36: Why not a textual id here as well?

src/migrations/20241024062521-add-release-plans-milestones-strategies.js
```diff
@@ -0,0 +1,62 @@
...
+            milestone_id SERIAL NOT NULL REFERENCES milestones(id) ON DELETE CASCADE,
```
chriswk, line 37: Same thing here, the type of the FK side of the relationship isn't another SERIAL, it's an INT.

chriswk: Though, I'd suggest maybe having all these new tables have textual PKs

src/migrations/20241024062521-add-release-plans-milestones-strategies.js
```diff
@@ -0,0 +1,62 @@
...
+        CREATE TABLE IF NOT EXISTS release_plan_definitions
+        (
+            id TEXT PRIMARY KEY NOT NULL,
+            discriminator TEXT NOT NULL,
+            name TEXT NOT NULL,
+            description TEXT,
+            feature_name TEXT,
+			environment TEXT,
+            created_by_user_id INTEGER NOT NULL REFERENCES users(id),
+            created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
```
chriswk, line 13:
```suggestion
            created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT (now() at time zone 'utc'),
```
to ensure we set created_at to UTC?

src/migrations/20241024062521-add-release-plans-milestones-strategies.js
daveleek: Yep 100% correct

src/migrations/20241024062521-add-release-plans-milestones-strategies.js
daveleek: Cool thanks! Good suggestion!

chriswk (commented):
So, we're making a cyclic dependency between release_plan_definitions and milestones?

chriswk (approved):
Nice. LGTM

daveleek:
> So, we're making a cyclic dependency between release_plan_definitions and milestones?

Addressed in chat - But for clarity: yes. It's intentional but shouldn't be cyclic. a release_plan_definition will be added, milestones will be added to this release plan, then we can set an active milestone via the release_plan_definition.active_milestone_id.

PR #8530, opened 2024-10-24
fix: error logger catchall

Tymek:
## About the changes
Some specially crafted URLs log 500 errors, instead of 400. This change should catch it in last middleware registered of an Express app.

Closes [1-3027/bug-logger-not-catching-unparsed-urls-correctly](https://linear.app/unleash/issue/1-3027/bug-logger-not-catching-unparsed-urls-correctly)

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -197,12 +197,6 @@ export default async function getApp(
         services.openApiService.useErrorHandler(app);
     }
 
-    if (process.env.NODE_ENV !== 'production') {
-        app.use(errorHandler());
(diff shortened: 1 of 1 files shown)


gastonfournier (commented):
I still would like to reproduce before making this change. I've been able to confirm that the status code errors we were seeing in logs were actually 4xx. But there were other requests at the same time which were POST and were returning 5xx responses. As you said that looked like a security scan looking for vulnerabilities, so it's abnormal traffic, but I still want to figure out how they caused the 5xx.

PR #8532, opened 2024-10-24
Email template for Productivity Report

Tymek:
## About the changes

image


package.json
```diff
@@ -34,7 +34,7 @@
   "main": "./dist/lib/server-impl.js",
   "scripts": {
     "start": "TZ=UTC node ./dist/server.js",
-    "copy-templates": "copyfiles -u 1 src/mailtemplates/**/*.mustache dist/",
+    "copy-templates": "copyfiles -u 1 src/mailtemplates/**/* dist/",
```
kwasniew, line 37: what else did we copy before?

src/lib/services/email-service.test.ts
```diff
@@ -144,7 +144,8 @@ test('Can send order environments email', async () => {
     expect(content.bcc).toBe('bcc@bcc.com');
 });
 
-test('Can send productivity report email', async () => {
+// TODO: test
+test.skip('Can send productivity report email', async () => {
```
kwasniew, line 148: why do we skip it?

src/lib/services/email-service.ts
```diff
@@ -28,6 +28,11 @@ export interface IEmailEnvelope {
     subject: string;
     html: string;
     text: string;
+    attachments?: {
```
kwasniew, line 31: what's the rationale for this? what did we do before?

src/lib/services/email-service.ts
```diff
@@ -522,32 +527,48 @@ export class EmailService {
     }
 
     async sendProductivityReportEmail(
+        userName: string,
```
kwasniew, line 530: it will be a breaking change for my enterprise PR so we need to cordinate

src/lib/services/email-service.ts
```diff
@@ -522,32 +527,48 @@ export class EmailService {
...
         userEmail: string,
-        customerId: string,
+        metrics: {
```
kwasniew, line 532: I have this structure in enterprise (we use underscores only in db types):    

   ```
       health: number;
        productionUpdates: number;
        flagsCreated: number;
```

File: src/lib/services/email-service.ts [hunk 6]
```diff
@@ -613,6 +634,29 @@ export class EmailService {
...
+        filename: string,
+        cid: string,
+    ): {
+        filename: string;
+        path: string;
+        cid: string;
+    } {
+        const topPath = path.resolve(__dirname, '../../mailtemplates');
+        const attachment = path.join(topPath, templateName, filename);
+        console.log({ attachment });
```
kwasniew, line 648: leftover

src/lib/services/email-service.ts
```diff
@@ -613,6 +634,29 @@ export class EmailService {
...
+    private resolveTemplateAttachment(
+        templateName: string,
+        filename: string,
+        cid: string,
+    ): {
+        filename: string;
+        path: string;
+        cid: string;
+    } {
+        const topPath = path.resolve(__dirname, '../../mailtemplates');
```
kwasniew, line 646: we need to check if everything works with base paths in sandbox

src/mailtemplates/productivity-report/productivity-report.html.mustache
```diff
@@ -1,320 +1,151 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
-<html xmlns="http://www.w3.org/1999/xhtml">
+<!DOCTYPE html>
```
kwasniew, line 1: how did you check if it works in older email clients?

src/lib/services/email-service.test.ts
Tymek: Not finished yet. It was testing previous template. I'll update it

src/lib/services/email-service.ts
Tymek: this is the same as templates, but yes, we can check

kwasniew (approved):

PR #8551, opened 2024-10-28
fix: invalid spreading of keys into table rows everywhere

thomasheartman:
This commit fixes invalid prop spreading warnings in all the table rows I
could find through a quick search in the code base.

The issue is that you can't spread the "key" prop into a component. It *must* be an explicit prop.

The process is the same everywhere:
1. Instead of spreading `row.getRowProps()` into the component, we extract and split it: `const {key, ...rowProps} = row.getRowProps()`.
2. Do the same thing for cellProps.

### DIFF
--- a/frontend/src/component/changeRequest/ProjectChangeRequests/ProjectChangeRequests.tsx
+++ b/frontend/src/component/changeRequest/ProjectChangeRequests/ProjectChangeRequests.tsx
@@ -16,8 +16,6 @@ export const ProjectChangeRequests = () => {
 
     const { changeRequests, loading } = useProjectChangeRequests(projectId);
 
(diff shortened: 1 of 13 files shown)


frontend/src/component/changeRequest/ProjectChangeRequests/ProjectChangeRequests.tsx
```diff
@@ -16,6 +16,8 @@ export const ProjectChangeRequests = () => {
 
     const { changeRequests, loading } = useProjectChangeRequests(projectId);
 
+    console.log('changeRequests', changeRequests);
+
```
thomasheartman, line 20:
```suggestion
```

kwasniew (approved):
Great!

PR #8553, opened 2024-10-28
chore: drop experimental from new signal meta properties

nunogois:
https://linear.app/unleash/issue/2-2855/change-our-experimental-title-to-camelcase-and-maybe-remove

Drops "experimental" from the new signal meta properties and makes them camelCase instead.


sighphyre (approved):
LGTM

PR #8570, opened 2024-10-29
chore: allow CC-BY-4.0

chriswk:
CC BY 4.0 This license lets others distribute, remix, tweak, and build upon your work, even commercially, as long as they credit you for the original creation. This is the most accommodating of licenses offered. Recommended for maximum dissemination and use of licensed material.

- Sounds compatible with us


melindafekete (approved):
Thank you! 🙌

PR #8571, opened 2024-10-29
feat(productiviy-report): email config

Tymek:
## About the changes
Add ability to customize email headers for non-transactional emails.


src/lib/services/email-service.ts
```diff
@@ -569,7 +583,9 @@ export class EmailService {
                         'unleashLogo',
                     ),
                 ],
-            };
+                headers,
+            } satisfies IEmailEnvelope;
```
kwasniew, line 587: not needed

kwasniew (approved):

PR #8572, opened 2024-10-29
fix: set path-to-regexp version in docker

gastonfournier:
Should fix: https://github.com/Unleash/unleash/security/dependabot/191


chriswk (approved):
LGTM

PR #8592, opened 2024-10-30
feat: search endpoint should return archived at date

sjaanus:
Include archived at to search response payload.


src/test/e2e/helpers/test-helper.ts
```diff
@@ -154,17 +152,6 @@ function httpApis(
                 .expect(expectedResponseCode);
         },
 
-        getFeatures(
```
sjaanus, line 157: We do not have this method in backend and it was not used.

thomasheartman (approved):

PR #8596, opened 2024-10-30
feat: traffic limits for enterprise-payg

Tymek:
## About the changes
image


nunogois (commented):
Looks like a clean approach! 
Shouldn't we add the feature flag in [experimental.ts](https://github.com/Unleash/unleash/blob/main/src/lib/types/experimental.ts) as well? 

Since it's still just being introduced, maybe we can rename it to camelCase so it's consistent with our other flags, i.e. `enterprisePAYG` instead of `enterprise-payg`. It's also a bit detached from enterprise in its implementation, even though that's only a detail, but perhaps it could be more generic like `paygBilling` or something.

nunogois (approved):
LGTM

PR #8609, opened 2024-10-31
Revert "fix: set path-to-regexp version in docker" (#8597)

gastonfournier:
Reverts Unleash/unleash#8572

Same fix applied in main


gastonfournier:
We know the packages are vulnerable but that'll be fixed later.

PR #8613, opened 2024-10-31
feat: use actionable change request data in UI

thomasheartman:
This PR hooks up the actionable change request data to the counter in the UI. It: 
- creates a getter for the data. It only exposes data. We don't really care about error or loading for this (it's not an important piece of data), so we don't expose that just yet.
- Adds orval-generated schema
- Uses the hook in the UI.

It also stwitches the previous "notification badge" for MUI's built-in badge. We already use that badge component for the event timeline, so I thought it would make sense to do it here too. Overall, the effect is pretty good, but there's a few kinks we might wanna work out. I'll make a follow-up for that:

image

image


frontend/src/component/project/Project/Project.tsx
```diff
@@ -160,6 +154,13 @@ export const Project = () => {
             path: `${basePath}/change-requests`,
             name: 'change-request',
             isEnterprise: true,
+            label: simplifyProjectOverview
+                ? () => (
+                      <ActionableChangeRequestsIndicator>
+                          Change requests
```
sjaanus, line 160: Why are we passing this in? Are we using any other texts?

sjaanus (approved):

frontend/src/component/project/Project/Project.tsx
thomasheartman: Good point. Yeah, We can remove that.

PR #8614, opened 2024-10-31
chore: add migration for release-plans-definitions that removes fk on users(id)

daveleek:
Removes a fk constraint release-plans-definitions->created_by_user_id on users table so we can delete users even if they have created release-plans/templates

Also adds an index on created_by_user_id


chriswk (approved):
LGTM

PR #8616, opened 2024-10-31
chore: upgrade to biome 1.9.4

chriswk:
The two lints being turned off are new for 1.9.x and caused a massive diff inside frontend if activated, to reduce impact, turned them off for the merge. We might want to look at turning them back on once we're ready to have a semantic / a11y refactor of our frontend

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,9 +1,12 @@
 {
-  "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
+  "$schema": "https://biomejs.dev/schemas/1.9.4/schema.json",
   "linter": {
     "enabled": true,
     "rules": {
       "recommended": true,
+      "a11y": {
+        "useSemanticElements": "off"
+      },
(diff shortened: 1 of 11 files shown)


biome.json
```diff
@@ -1,9 +1,12 @@
 {
-  "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
+  "$schema": "https://biomejs.dev/schemas/1.9.4/schema.json",
   "linter": {
     "enabled": true,
     "rules": {
       "recommended": true,
+      "a11y": {
+        "useSemanticElements": "off"
```
kwasniew, line 8: can we add explanation in the PR why those changes were needed? have some defaults changed?

kwasniew (approved):

biome.json
chriswk: Yes, good suggestion. Added as main comment to PR, will include in merge commit message

PR #8627, opened 2024-11-01
chore: use palette.background.alternative for background

thomasheartman:
The default badge styling uses theme.palette.primary.main for the
color, but this color doesn't give us the contrast we need in dark
mode. The solution is to use theme.palette.background.alternative instead.

Before:
image

After:
image


daveleek (approved):
LGTM! :)

PR #8628, opened 2024-11-01
feat: activity widget

sjaanus:
It is still raw, next PRs add styling and date filtering for only single year. 

image


frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
```diff
@@ -0,0 +1,90 @@
...
+import type { ProjectActivitySchema } from '../../../../openapi';
+import { styled, Tooltip } from '@mui/material';
+
+const StyledContainer = styled('div')(({ theme }) => ({
+    gap: theme.spacing(1),
+}));
+
+type Output = { date: string; count: number; level: number };
+
+export function transformData(inputData: ProjectActivitySchema): Output[] {
```
sjaanus, line 13: This can be cleaned up later,  but basically for each day we need a level(color), and this divides them evenly into level buckets.

frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const resultMap: Record<string, number> = {};
+
+    // Step 1: Count the occurrences of each date
+    inputData.forEach((item) => {
+        const formattedDate = new Date(item.date).toISOString().split('T')[0];
```
thomasheartman, line 18: Aren't they already formatted this way?

But anyway, this is just an accumulation, right? Assuming we have multiple entries for each date, it'll count each entry as one and sum them together? 

Maybe I've misunderstood the API, but I thought we returned each date only once, with already summed up activities? In that case, this'd be wrong, right?

thomasheartman (approved):

PR #8630, opened 2024-11-01
feat: project status backend structure

sjaanus:
Adding project status schema definition, controller, service, e2e test.

Next PR will add functionality for activity object.


src/lib/features/project-status/projects-status.e2e.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+beforeAll(async () => {
+    db = await dbInit('projects_status', getLogger);
+    app = await setupAppWithCustomConfig(
+        db.stores,
+        {
+            experimental: {
+                flags: {
+                    strictSchemaValidation: true,
+                },
```
thomasheartman, line 19: Do we need to set this flag manually still? I thought it was auto-on? 🤔

src/lib/openapi/spec/project-activity-schema.ts
```diff
@@ -0,0 +1,29 @@
...
+    description:
+        'An array of project activities, each containing a date and activity count',
+    items: {
+        type: 'object',
+        additionalProperties: false,
+        required: ['date', 'count'],
+        properties: {
+            date: {
+                type: 'string',
+                description: 'Activity date',
```
thomasheartman, line 15: Should this have a format? E.g. 
```suggestion
                type: 'string',
                format: 'date',
                description: 'Activity date',
```

src/lib/openapi/spec/project-activity-schema.ts
```diff
@@ -0,0 +1,29 @@
+import type { FromSchema } from 'json-schema-to-ts';
+
+export const projectActivitySchema = {
+    $id: '#/components/schemas/projectActivitySchema',
+    type: 'array',
+    description:
+        'An array of project activities, each containing a date and activity count',
```
thomasheartman, line 7:
```suggestion
        'An array of project activity information. Each item contains a date and the total number of activities for that date.,
```

src/lib/features/project-status/project-status-controller.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.route({
+            method: 'get',
+            path: '/:projectId/status',
+            handler: this.getProjectStatus,
+            permission: NONE,
+            middleware: [
+                this.openApiService.validPath({
+                    tags: ['Projects'],
+                    operationId: 'getProjectStatus',
+                    summary: 'Get an overview of a project status.',
```
thomasheartman, line 44: Is this the endpoint for the whole project status view? 
Maybe something like this is enough:

```suggestion
                    summary: 'Get project status',
```

src/lib/features/project-status/project-status-controller.ts
```diff
@@ -0,0 +1,71 @@
...
+                    description:
+                        'This endpoint returns status for the project, including activities.',
```
thomasheartman, line 46:
```suggestion
                        'This endpoint returns information on the status the project, including activities, health, resources, and aggregated flag lifecycle data.',
```

src/lib/openapi/spec/project-status-schema.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const projectStatusSchema = {
+    $id: '#/components/schemas/projectStatusSchema',
+    type: 'object',
+    additionalProperties: false,
+    required: ['activities'],
+    description:
+        'Schema representing the overall status of a project, including an array of activity records. Each record in the activity array contains a date and a count, providing a snapshot of the project’s activity level over time.',
+    properties: {
+        activities: {
```
thomasheartman, line 12: I'm not sure `activities` is a good name for this property. It took me a good while into reading this PR before I realized that activities weren't things such as "flag created", "strategy added", but instead an aggregated list of such activities (presumably?) divided into per-day data.

Not suuper important (and as long as the endpoint is unstable we can change it), but maybe we could use a different wording. I mean, this isn't really a list of activities, but a list of dates with activity information. Would `activityCountByDate` be too wordy?

thomasheartman (approved):
Nice one! Got a couple suggestions we could throw in, but I'm also not sure about the name of the schema / property. But there's a comment for that and we can change it alter.

src/lib/features/project-status/projects-status.e2e.test.ts
sjaanus: This is not auto on, we should enable it, but I think our schemas still suck enough that we cant :)

PR #8631, opened 2024-11-01
chore: stub out project resources for status model

thomasheartman:
This PR begins to stub out the project resources widget. I still need one more piece of data and then to work on the styling, but it's a placeholder for now. I've also moved the project status modal to its own folder so we can group the widgets etc. I'd like to get that merged quickly to avoid any future conflicts, which is why I'm making the PR ready now.


sjaanus (approved):
LG, I wonder if in future we should move everything under status endpoint, to not make so many ajax requests, but lets see.

PR #8634, opened 2024-11-04
feat: Make `disableMigration` option handled by environment variable

0417taehyun:
## About the changes

Closes #

## Discussion points

In some cases, people want to disable database migration. For example, some people or companies want to grant whole permissions to handle the schema by DBAs, not by application level hence I use `parseEnvVarBoolean` to handle `disableMigration` option by environment variable. I set the default value as `false` for the backward compatibility.

However, I wonder where to add test code of it because all the unit and e2e test is a kind of testing the whole configuration.

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -259,7 +259,7 @@ const defaultDbOptions: WithOptional<IDBOption, 'user' | 'password' | 'host'> =
             propagateCreateError: false,
         },
         schema: process.env.DATABASE_SCHEMA || 'public',
-        disableMigration: false,
+        disableMigration: parseEnvVarBoolean(process.env.DATABASE_DISABLE_MIGRATION, false),
         applicationName: process.env.DATABASE_APPLICATION_NAME || 'unleash',
     };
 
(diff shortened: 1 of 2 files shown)


gastonfournier:
Hi @0417taehyun I think this makes sense because this is already a configurable option. There's a linting issue you should just run `yarn lint:fix`. 

Regarding testing, we test that after the migrations applied Unleash works well, if migrations are not applied there's no guarantee that Unleash will work well and most likely it won't work, so I don't think there's need to test this. Maybe the only thing that comes to mind is testing that when this is disabled the migrations are not executed.

Maybe just checking if this log is written: https://github.com/Unleash/unleash/blob/a7d581aa7d6891c6ef6c2e3489b40c587c1501a0/src/lib/server-impl.ts#L146-L147

0417taehyun:
@gastonfournier 

Thanks for answering my question and checking linting issue! I fixed the issued by using `yarn lint:fix` and check the `DB migration: disabled` log. I also agree that it is okay not to test about this logic. Thanks!

gastonfournier (approved):
Looks good! Thanks for the contribution!

PR #8635, opened 2024-11-04
fix: text color in timeline event group for light mode

thomasheartman:
The recent fix to dark mode's background color for events also
caused an issue with the text color in light mode. This change makes
it so that the text color is correct in both instances.

Before (light):
image

After (light + dark):
image
image


thomasheartman:
Reviewed by Anniken; merging.

PR #8637, opened 2024-11-04
feat: unsubscribe link to user profile

kwasniew:
## About the changes

Show link to user profile where she can manager email subscription

### DIFF
--- a/src/lib/services/email-service.test.ts
+++ b/src/lib/services/email-service.test.ts
@@ -175,5 +175,5 @@ test('Can send productivity report email', async () => {
     expect(content.html.includes(`localhost/insights`)).toBe(true);
     expect(content.html.includes(`localhost/profile`)).toBe(true);
     expect(content.text.includes(`localhost/insights`)).toBe(true);
-    expect(content.text.includes(`localhost/insights`)).toBe(true);
+    expect(content.text.includes(`localhost/profile`)).toBe(true);
 });
--- a/src/lib/services/email-service.test.ts
+++ b/src/lib/services/email-service.test.ts
@@ -169,8 +169,11 @@ test('Can send productivity report email', async () => {
             health: 99,
         },
     );
-    console.log(content);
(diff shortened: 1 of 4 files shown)


src/lib/services/email-service.test.ts
```diff
@@ -169,8 +169,11 @@ test('Can send productivity report email', async () => {
...
         },
     );
-    console.log(content);
     expect(content.from).toBe('noreply@getunleash.ai');
     expect(content.subject).toBe('Unleash - productivity report');
     expect(content.html.includes(`Productivity Report`)).toBe(true);
+    expect(content.html.includes(`localhost/insights`)).toBe(true);
+    expect(content.html.includes(`localhost/profile`)).toBe(true);
+    expect(content.text.includes(`localhost/insights`)).toBe(true);
+    expect(content.text.includes(`localhost/insights`)).toBe(true);
```
Tymek, line 178:
```suggestion
    expect(content.text.includes(`localhost/insights`)).toBe(true);
    expect(content.text.includes(`localhost/profile`)).toBe(true);
```

src/lib/services/email-service.ts
```diff
@@ -537,14 +537,11 @@ export class EmailService {
         },
     ): Promise<IEmailEnvelope> {
         if (this.configured()) {
-            const unsubscribeUrl = '{{amazonSESUnsubscribeUrl}}'; // FIXME: Add unsubscribe URL
```
Tymek, line 540: We will probably need to also put unsubscripe url in headers. We need it on this level

Tymek (changes requested):

src/lib/services/email-service.ts
kwasniew: But it will be different from the profile link that I used and composed in the template. I will add a task for the header in issue tracking

Tymek (approved):

PR #8642, opened 2024-11-04
chore: add Unleash AI to New in Unleash

nunogois:
https://linear.app/unleash/issue/2-2910/add-unleash-ai-to-new-in-unleash

Adds a new "Unleash AI" item to the "New in Unleash" section.

We don’t have documentation to link to yet, as this is still an experimental feature. However, I’m considering adding a “Check it out” button that highlights the feature, which I can introduce in a separate PR.

image

PR #8643, opened 2024-11-04
refactor: introduce a highlight reusable component

nunogois:
Follow-up to: https://github.com/Unleash/unleash/pull/8642

Introduces a reusable `Highlight` component that leverages the Context API pattern, allowing highlight effects to be triggered from anywhere in the application.

This update refactors the existing highlight effect in the event timeline to use the new Highlight component and extends the highlight functionality to include the Unleash AI feature.


gastonfournier (approved):
Looks nice!

PR #8652, opened 2024-11-05
feat: add user preference change to event log

Tymek:


src/lib/types/events.ts
```diff
@@ -2024,3 +2027,18 @@ function mapUserToData(user: IUserEventData): any {
...
     };
 }
+
+export class UserPreferenceUpdatedEvent extends BaseEvent {
+    readonly userId;
+    readonly data: any;
+
+    constructor(eventData: {
+        userId: number;
+        data: any;
```
kwasniew, line 2037: do we know the structure?

PR #8663, opened 2024-11-05
chore: new feature flag overview metadata panel

nunogois:
https://linear.app/unleash/issue/2-2920/update-the-flag-overview-metadata-properties-to-match-the-new-design

Updates the feature flag overview metadata panel to match the new design.

This redesign is behind a feature flag, so we opted to go with a duplicate file approach. We should remember to clean this up if we decide to remove the flag.

image


chriswk (approved):
Nice. You got a :+1: from me at least.

PR #8665, opened 2024-11-06
feat: activity chart polish

sjaanus:
image

### DIFF
--- a/frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
+++ b/frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
@@ -5,22 +5,26 @@ import type { ProjectActivitySchema } from '../../../../openapi';
 import { styled, Tooltip } from '@mui/material';
 
 const StyledContainer = styled('div')(({ theme }) => ({
-    gap: theme.spacing(1),
+    display: 'flex',
+    flexDirection: 'column',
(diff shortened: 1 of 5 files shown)


frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
```diff
@@ -5,22 +5,26 @@ import type { ProjectActivitySchema } from '../../../../openapi';
...
+    justifyContent: 'center',
+    alignItems: 'center',
+    gap: theme.spacing(2),
 }));
 
+const TitleContainer = styled('div')({
+    width: '100%',
+    display: 'flex',
+    justifyContent: 'flex-start',
+});
```
thomasheartman, line 19: Couldn't we just make this an h4? If there's only one element in it, why make it a flex div at all? Wouldn't something like this be enough?

```suggestion
const TitleContainer = styled('h4')({
    margin: 0,
    width: '100%',
});
```

thomasheartman: Oh, right, because you put the whole thing inside a flex container. Well, we can just make the h4 take up 100% width. Or chuck an align self on the graph. (or remove flex and go old school with margin auto 😄 )

thomasheartman (approved):

PR #8667, opened 2024-11-06
fix: prevent rendering too many hooks error

thomasheartman:
We found an issue where we'd get a minified react error referencing the LazyProjectExport component.

We suspect that the issue might be the conditional rendering of this component, so the fix is to always render it, but to use the flag to check whether we should show the count or not.

### DIFF
--- a/frontend/src/component/project/Project/Project.tsx
+++ b/frontend/src/component/project/Project/Project.tsx
@@ -94,11 +94,14 @@ const TabText = styled('span')(({ theme }) => ({
 }));
 
 const ChangeRequestsLabel = () => {
+    const simplifyProjectOverview = useUiFlag('simplifyProjectOverview');
     const projectId = useRequiredPathParam('projectId');
     const { total } = useActionableChangeRequests(projectId);
 
+    const count = simplifyProjectOverview ? 0 : (total ?? 0);
+
     return (
-        <StyledCounterBadge badgeContent={total ?? 0} color='primary'>
+        <StyledCounterBadge badgeContent={count} color='primary'>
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/Project/Project.tsx
```diff
@@ -94,11 +94,14 @@ const TabText = styled('span')(({ theme }) => ({
 }));
 
 const ChangeRequestsLabel = () => {
+    const simplifyProjectOverview = useUiFlag('simplifyProjectOverview');
     const projectId = useRequiredPathParam('projectId');
     const { total } = useActionableChangeRequests(projectId);
 
+    const count = simplifyProjectOverview ? 0 : (total ?? 0);
```
FredrikOseberg, line 101: Isn't `if (!simplifyProjectOverview) return null;` a better check? If the flag is off we don't add empty elements to the DOM.

thomasheartman: No, because if we return `null`, we won't render a label at all. This way, at least the label is the same, though we could potentially just use a different label

FredrikOseberg: Ok, I thought this was just the count. I'll go ahead and approve it then.

FredrikOseberg (approved):
LGTM

frontend/src/component/project/Project/Project.tsx
thomasheartman: But I found a nicer way to do it in ad7aa879fa6b7ac1be54c97c8f14c8be8d7d3033. We'll just render the text.

PR #8672, opened 2024-11-06
Feat/release management overview

daveleek:
Release management overview page.
This is big enough already, so I'll add a separate PR with release plan template deletion, navigating to edit, and we need to resolve createdBy int to a username somehow

image


chriswk (approved):
SRTM (Seems reasonable to me) >)

PR #8673, opened 2024-11-06
fix: active flags double stale flags

sjaanus:
Do not count stale flags as potentially stale flags to remove duplicates.
Stale flags feel like more superior state and it should not show up under potentially stale.


frontend/vite.config.mts
```diff
@@ -26,10 +26,10 @@ const vitestConfig = vitestDefineConfig({
     css: {
         preprocessorOptions: {
             scss: {
-                api: 'modern-compiler'
-            }
-        }
-    }
+                api: 'modern-compiler',
```
sjaanus, line 29: Linter wants this now.

thomasheartman (approved):

PR #8678, opened 2024-11-06
feat: deleting release-plan templates

daveleek:
Deleting release plan templates (with confirmation dialog)
We should probably look at permission model for this, I'll make a note of that

image


chriswk (approved):
LGTM

daveleek:
And with your endpoints PR merged I have now validated that it actually works :)

PR #8679, opened 2024-11-06
1-3059: add initial visuals for lifecycle summary

thomasheartman:
Add rough implementation of the lifecycle summary components.

This PR adds components for all the different lifecycle stages. We don't have any data yet, so they're all hardcoded for now, just to get the visuals right. I'm expecting the lines of code to drop and to refactor/extract some structures as development continues.

For now, this is what they look like:

image

Things to note:
- The lifecycle stage icon colors don't match up with the sketches, but they match up with what we currently have in the app. If we change them, we should change them together.
- This implementation does not contain the "Flag lifecycle" header or the "view graphs" link.

### DIFF
--- a/frontend/src/component/project/Project/ProjectStatus/ProjectLifecycleSummary.tsx
+++ b/frontend/src/component/project/Project/ProjectStatus/ProjectLifecycleSummary.tsx
@@ -1,54 +1,65 @@
-import { Typography, styled } from '@mui/material';
+import { styled } from '@mui/material';
 import { FeatureLifecycleStageIcon } from 'component/feature/FeatureView/FeatureOverview/FeatureLifecycle/FeatureLifecycleStageIcon';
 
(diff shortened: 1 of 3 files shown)


frontend/src/component/project/Project/ProjectStatus/ProjectLifecycleSummary.tsx
```diff
@@ -0,0 +1,177 @@
...
+                    </Counter>
+                    <span>flags in archived</span>
+                </p>
+                <Stats>
+                    <dt>This month</dt>
+                    <dd>3 flags archived</dd>
+                </Stats>
+            </LifecycleBox>
+        </Wrapper>
+    );
```
thomasheartman, line 176: This is the complete, rolled-out version of what it'll look like. We'll very likely refactor this in a future iteration..

sjaanus (approved):

PR #8683, opened 2024-11-06
chore: flag overview page redesign - environments

nunogois:
https://linear.app/unleash/issue/2-2826/enabling-environment-via-feature-flag-environment-section-header
https://linear.app/unleash/issue/2-2825/feature-flag-list-bottom-left-to-be-a-nav-section

Follow-up to: https://github.com/Unleash/unleash/pull/8663

Implements most of the remaining work for our flag overview page redesign.

Most of the code you see is a straight copy/paste from our older existing components, with the slight improvement here and there.

Includes some improvements to our vertical tabs component to suit our use case. 
Also updates the Demo flow accordingly. I did some manual tests and it seems to work decently in both scenarios, whether `flagOverviewRedesign` is enabled or not. The demo needs some love but that's a story for a different PR and a different time.

Once again, due to the duplicate file pattern, we should remember to clean this up if we decide to remove the flag.

image

### DIFF
--- a/frontend/cypress/integration/demo/demo.spec.ts
+++ b/frontend/cypress/integration/demo/demo.spec.ts
@@ -40,6 +40,7 @@ describe('demo', () => {
                     res.body.flags = {
                         ...res.body.flags,
                         demo: true,
+                        flagOverviewRedesign: true,
                     };
                 }
             });
--- a/frontend/cypress/support/UI.ts
+++ b/frontend/cypress/support/UI.ts
@@ -227,7 +227,6 @@ export const deleteFeatureStrategy_UI = (
         },
     ).as('deleteUserStrategy');
     cy.visit(`/projects/${project}/features/${featureToggleName}`);
-    cy.get('[data-testid=FEATURE_ENVIRONMENT_ACCORDION_development]').click();
     cy.get('[data-testid=STRATEGY_REMOVE_MENU_BTN]').first().click();
     cy.get('[data-testid=STRATEGY_FORM_REMOVE_ID]').first().click();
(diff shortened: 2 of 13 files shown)


frontend/cypress/integration/demo/demo.spec.ts
```diff
@@ -40,6 +40,7 @@ describe('demo', () => {
                     res.body.flags = {
                         ...res.body.flags,
                         demo: true,
+                        flagOverviewRedesign: true,
```
daveleek, line 43: What's the implications of this?

daveleek (approved):
Nice one, big task! 😅

chriswk (approved):
I see you have a :+1: from David already, but I'll add mine as well. Kudos on doing this :1st_place_medal:

frontend/cypress/integration/demo/demo.spec.ts
nunogois: This makes the demo e2e tests, that we only run manually and rarely, use the new flag overview page redesign.

frontend/cypress/support/UI.ts
```diff
@@ -227,7 +227,6 @@ export const deleteFeatureStrategy_UI = (
         },
     ).as('deleteUserStrategy');
     cy.visit(`/projects/${project}/features/${featureToggleName}`);
-    cy.get('[data-testid=FEATURE_ENVIRONMENT_ACCORDION_development]').click();
```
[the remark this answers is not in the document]
nunogois: With this change we're only testing the new UI, not the old one. See: https://github.com/Unleash/unleash/pull/8683/files#diff-3c93375ad9f5acedec6fbc6c020d3927053759ee0b74bd129a1397e1e22ba81aR47

PR #8685, opened 2024-11-07
fix: show archive tooltip only on project screen

sjaanus:
Fixing archive tooltip appearing on every filter in Unleash

image


thomasheartman (approved):
😄 Would it be possible to add this on the project page instead? Would that be safer?

sjaanus:
> 😄 Would it be possible to add this on the project page instead? Would that be safer?

not easily, since the button is part of bigger component, so I can not wrap the whole big component. and then would need to pass it in with props.

PR #8687, opened 2024-11-07
feat: health score backend

sjaanus:
Implements the backend part for project health.

### DIFF
--- a/src/lib/features/project-status/project-status-service.ts
+++ b/src/lib/features/project-status/project-status-service.ts
@@ -6,26 +6,32 @@ import type {
     ISegmentStore,
     IUnleashStores,
 } from '../../types';
+import type { IPersonalDashboardReadModel } from '../personal-dashboard/personal-dashboard-read-model-type';
 
 export class ProjectStatusService {
     private eventStore: IEventStore;
     private projectStore: IProjectStore;
     private apiTokenStore: IApiTokenStore;
     private segmentStore: ISegmentStore;
+    private personalDashboardReadModel: IPersonalDashboardReadModel;
 
-    constructor({
-        eventStore,
-        projectStore,
-        apiTokenStore,
-        segmentStore,
-    }: Pick<
-        IUnleashStores,
-        'eventStore' | 'projectStore' | 'apiTokenStore' | 'segmentStore'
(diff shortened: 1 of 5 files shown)


src/lib/features/project-status/project-status-service.ts
```diff
@@ -35,14 +41,21 @@ export class ProjectStatusService {
...
         ] = await Promise.all([
             this.projectStore.getConnectedEnvironmentCountForProject(projectId),
             this.projectStore.getMembersCountByProject(projectId),
             this.apiTokenStore.countProjectTokens(projectId),
             this.segmentStore.getProjectSegmentCount(projectId),
             this.eventStore.getProjectRecentEventActivity(projectId),
+            this.personalDashboardReadModel.getLatestHealthScores(projectId, 4),
         ]);
 
+        const averageHealth = healthScores.length
```
sjaanus, line 54: This is a decision I made, in the personal dashboard, if you do not have 4 week data, it was showing 0. It felt silly so I made that in a "rare" case there is just 1-3 weeks, it will use that to calculate average. 

Althought the frontend will still show 4. But this is the first 3 weeks instance is set up. I think it makes sense to show average for 1-3 also, atleast I would like to see this chart already from week 1.

thomasheartman: Yea, that sounds sensible to me 😄 We could also send down either how many weeks of data we've used (so that it's correct) or calculate that in the front end based on the project's age 💁🏼

src/lib/openapi/spec/project-status-schema.ts
```diff
@@ -14,6 +14,12 @@ export const projectStatusSchema = {
             description:
                 'Array of activity records with date and count, representing the project’s daily activity statistics.',
         },
+        averageHealth: {
+            type: 'number',
```
thomasheartman, line 18: Actually, now that I think about it, this can / should probably be an `integer` because we round it, right?

```suggestion
            type: 'integer',
```

And if not for this one, it at least goes for the `resources` properties - but that's on me. I'll fix it 🙋🏼

thomasheartman (approved):

PR #8690, opened 2024-11-07
feat: activity chart now shows full year

sjaanus:
Previously when there was no data, the activity chart showed as really small. Now I am populating the start date and end date to draw out full chart.

image

### DIFF
--- a/frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
+++ b/frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
@@ -19,7 +19,31 @@ const TitleContainer = styled('h4')({
 
 type Output = { date: string; count: number; level: number };
 
-export function transformData(inputData: ProjectActivitySchema): Output[] {
+const ensureFullYearData = (data: Output[]): Output[] => {
+    const today = new Date();
+    const oneYearBack = new Date();
+    oneYearBack.setFullYear(today.getFullYear() - 1);
+
+    const formattedToday = today.toISOString().split('T')[0];
+    const formattedOneYearBack = oneYearBack.toISOString().split('T')[0];
+
+    const hasToday = data.some((item) => item.date === formattedToday);
+    const hasOneYearBack = data.some(
+        (item) => item.date === formattedOneYearBack,
+    );
+
+    if (!hasOneYearBack) {
+        data.unshift({ count: 0, date: formattedOneYearBack, level: 0 });
+    }
+
+    if (!hasToday) {
+        data.push({ count: 0, date: formattedToday, level: 0 });
+    }
+
+    return data;
+};
+
+const transformData = (inputData: ProjectActivitySchema): Output[] => {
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
```diff
@@ -46,14 +70,12 @@ export function transformData(inputData: ProjectActivitySchema): Output[] {
     };
 
     // Step 4: Convert the map back to an array and assign levels
-    return inputData
-        .map(({ date, count }) => ({
-            date,
-            count,
-            level: calculateLevel(count),
-        }))
-        .reverse();
```
sjaanus, line 55: Remove as data is already correctly ordered.

frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
```diff
@@ -19,7 +19,31 @@ const TitleContainer = styled('h4')({
...
+    const oneYearBack = new Date();
+    oneYearBack.setFullYear(today.getFullYear() - 1);
+
+    const formattedToday = today.toISOString().split('T')[0];
+    const formattedOneYearBack = oneYearBack.toISOString().split('T')[0];
+
+    const hasToday = data.some((item) => item.date === formattedToday);
+    const hasOneYearBack = data.some(
+        (item) => item.date === formattedOneYearBack,
+    );
```
thomasheartman, line 33: Can we check these without iteration? Not that the numbers are gonna be large enough for it to matter much, but if we have today, then that will always be `data[data.length-1]`, right? So we can just check the last element and see if that today or if it's less than? E.g. if `data[data.length-1].date < formattedToday`?

Likewise, will oneYearBack, ever be a different element than the first one? So we could check if `data[0].date > formattedOneYearBack`

Also, what are the cases where we don't have today or one year back? Is it edge cases where the date changes while the person is looking at the UI? Or is it dates when we just have no data, so we don't send anything?


And I'm not entirely sure how the chart works, so tell me if this is right: say you only have two data points a week apart, the chart will fill in the days between points A and B, but nothing else. So even if we only have a single data point from a few days ago, we can add only two points (today and a year ago) to draw a chart for a whole year. Yeah?

thomasheartman (approved):
Looks good to me! I think we can skip the iteration we're doing here, but it's likely that with the array sizes we use, it won't really make a difference 🤷🏼

frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
sjaanus: I will leave as is, because 1 year back might not compute the same way here and in database. Also like some timezone issues, where today is different frontend/backend.

Case when we do not today is very simple. no event today. It is not edge case. Or year back, also. if it is new project, it will not have any events.

Correct on the chart. it will take min and max and fill in the middle.

PR #8692, opened 2024-11-07
fix: console warnings and errors related to event timeline and strategy form

nunogois:
Fixes browser console warnings and errors related to the event timeline and strategy form.

- **Event Timeline**: Addressed a warning where the environment filter rendered with a default environment value (production) before environments were fully loaded.
- **Strategy Form**: Resolved an error caused by forwarding the enabled prop as a boolean.


daveleek (approved):
LGTM!

PR #8694, opened 2024-11-08
task: explicitly allows Artistic-2.0 license

chriswk:
Due to our use of docusaurus and the openapi plugin we need to allow Artistic-2.0, Having read the license, it allows for free use of the licensed code, provided the license is included when distributing, but does not require a relicensing of products using the licensed code.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -17,5 +17,5 @@ jobs:
         uses: actions/dependency-review-action@v4
         with:
           fail-on-severity: moderate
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Agree, I think this one's fine

daveleek (commented):
> (5) You may Distribute Compiled forms of the Standard Version without the Source, provided that you include complete instructions on how to get the Source of the Standard Version.

How does this affect us building and deploying docker images?

chriswk:
So far the docker images don't include the Artistic license code.

PR #8696, opened 2024-11-08
feat: archived toggles do not show switch

sjaanus:
Archived toggles will not show switch anymore, but a dash.

Screenshot from 2024-11-08 11-41-17

### DIFF
--- a/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/FeatureToggleCell/FeatureToggleCell.tsx
+++ b/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/FeatureToggleCell/FeatureToggleCell.tsx
@@ -96,4 +96,6 @@ export const PlaceholderFeatureToggleCell = () => (
         <div data-loading>toggle</div>
     </StyledSwitchContainer>
 );
-export const ArchivedFeatureToggleCell = () => <StyledDiv>-</StyledDiv>;
+export const ArchivedFeatureToggleCell = () => (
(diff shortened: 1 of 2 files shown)


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/FeatureToggleCell/FeatureToggleCell.tsx
```diff
@@ -90,3 +96,4 @@ export const PlaceholderFeatureToggleCell = () => (
         <div data-loading>toggle</div>
     </StyledSwitchContainer>
 );
+export const ArchivedFeatureToggleCell = () => <StyledDiv>-</StyledDiv>;
```
thomasheartman, line 99: Wonder if we should do `aria-hidden=true` here. It's not very helpful for screen readers if it reads out "development: hyphen, production: hyphen", etc. (which I *suspect* it will). Either that, or use ScreenReaderOnly to indicate that it has been disabled? But then, you'd get that for every row, which is also probably not super helpful. Hmm 🤔  What do you think?

thomasheartman (approved):

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/FeatureToggleCell/FeatureToggleCell.tsx
sjaanus: I think aria hidden makes sense.

PR #8698, opened 2024-11-08
tests: add more tests for the lifecycle avg calculation query

thomasheartman:
This PR adds more tests to check a few more cases for the lifecycle calculation query. Specifically, it tests that:
- If we don't have any data for a stage, we return `null`.
- We filter on projects
- It correctly takes `0` days into account when calculating averages.


website/docusaurus.config.ts
thomasheartman, line 1: Linter was failing because of this 🤷🏼

sjaanus (approved):

PR #8700, opened 2024-11-08
chore: put project lifecycle read model in own directory + add fake

thomasheartman:
This PR moves the project lifecycle summary to its own subdirectory and adds files for types (interface) and a fake implementation.

It also adds a query for archived flags within the last 30 days taken from `getStatusUpdates` in `src/lib/features/project/project-service.ts` and maps the gathered data onto the expected structure. The expected types have also been adjusted to account for no data.

Next step will be hooking it up to the project status service, adding schema, and exposing it in the controller.


src/lib/features/project-status/project-lifecycle-read-model/project-lifecycle-read-model-type.ts
```diff
@@ -0,0 +1,28 @@
+export interface IProjectLifecycleSummaryReadModel {
+    getProjectLifecycleSummary(
+        projectId: string,
+    ): Promise<ProjectLifecycleSummary>;
+}
+
+export type ProjectLifecycleSummary = {
+    initial: {
```
sjaanus, line 8: I feel like this should be type

{
        averageDays: number | null;
        currentFlags: number;
    };

sjaanus (approved):

src/lib/features/project-status/project-lifecycle-read-model/project-lifecycle-read-model-type.ts
thomasheartman: Hmm? Not sure I understand what you mean? As in extract out those two fields into its own thing? Sure, we can do that. Archived is different, but that might be alright.

thomasheartman: If I understood you correctly, it's done in 28236f4491b419b544c708994b61c397383c0b4f

PR #8703, opened 2024-11-08
chore: list release templates in strategy popover

nunogois:
https://linear.app/unleash/issue/2-2817/expand-the-strategy-dropdown-with-release-plan-templates

Does what it says on the tin, lists release plan templates in the strategy popover.

I think we should improve this popover soon, at least behind the `flagOverviewRedesign` flag, but not in this PR.

image


chriswk (approved):
Nice!

PR #8707, opened 2024-11-11
task: added migration for release plan template permissions

chriswk:
Due to the cascade on the permission in role_permission, the down migration does not need to remove the rows from role_permission.


daveleek (approved):
Looks great!

PR #8709, opened 2024-11-11
feat(1-3085): hook up lifecycle read model data to endpoint

thomasheartman:
This PR hooks up the project lifecycle summary read model to the service and exposes the lifecycle summary data in the controller.

### DIFF
--- a/src/lib/features/project-status/projects-status.e2e.test.ts
+++ b/src/lib/features/project-status/projects-status.e2e.test.ts
@@ -256,3 +256,33 @@ test('project health should be correct average', async () => {
 
     expect(body.averageHealth).toBe(40);
 });
+
+test('project status contains lifecycle data', async () => {
+    const { body } = await app.request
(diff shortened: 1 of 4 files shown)


src/lib/features/project-status/projects-status.e2e.test.ts
```diff
@@ -256,3 +256,33 @@ test('project health should be correct average', async () => {
 
     expect(body.averageHealth).toBe(40);
 });
+
+test('project status contains lifecycle data', async () => {
```
sjaanus, line 260: Could we have a small example also that data calculations work. Meaning, adding some data to database that will be returned.

thomasheartman: sure, I can add that in a follow-up, but that is essentially what all the tests in `src/lib/features/project-status/project-lifecycle-summary-read-model.test.ts` are there for.

PR #8715, opened 2024-11-12
chore: add release plan events

nunogois:
Adds new release plan events. 

These will be triggered by the new release plans API.


chriswk (approved):
Nice

PR #8718, opened 2024-11-12
1-3093: round the project health

thomasheartman:
This PR rounds the average health score we show for a project. With fractional numbers, it'd often overflow the graph. It also doesn't really give you much extra info, so we can round it. The rounding is then used both in the text, in the graph, and to calculate the graph fill percentage.

Before:
image

After:
image

### DIFF
--- a/frontend/src/component/project/Project/ProjectStatus/ProjectHealth.tsx
+++ b/frontend/src/component/project/Project/ProjectStatus/ProjectHealth.tsx
@@ -32,7 +32,6 @@ export const ProjectHealth = () => {
     const {
         data: { averageHealth },
     } = useProjectStatus(projectId);
-    const averageHealthRounded = Math.round(averageHealth);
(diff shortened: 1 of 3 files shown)


frontend/src/component/project/Project/ProjectStatus/ProjectHealth.tsx
```diff
@@ -32,6 +32,7 @@ export const ProjectHealth = () => {
     const {
         data: { averageHealth },
     } = useProjectStatus(projectId);
+    const averageHealthRounded = Math.round(averageHealth);
```
sjaanus, line 35: Sorry as this was my PR, but I think we should do it in backend.

thomasheartman: Yeah, I don't mind that at all. I'll update the PR.

sjaanus (approved):

PR #8723, opened 2024-11-12
feat: edit release plan template

daveleek:
This facilitates editing a release plan template
In order to split large UI projects up, this only allows you to change the name and description (but will show you the names of milestones in the UI), and will carry the existing milestones with it when making a request. We'll tackle editing milestones and their strategies in a different PR

Skjermbilde 2024-11-12 kl  15 15 53


chriswk (approved):
Nice

frontend/src/hooks/api/getters/useReleasePlanTemplates/useReleasePlanTemplateInstance.ts
```diff
@@ -0,0 +1,48 @@
...
+const DEFAULT_DATA: IReleasePlanTemplateInstance = {
+    id: '',
+    name: '',
+    description: '',
+    milestones: [],
+    createdAt: '',
+    createdByUserId: 0,
+};
+
+export const useReleasePlanTemplateInstance = (templateId: string) => {
```
nunogois, line 21: I assume this hook is needed to grab detailed data about the release plan template? Since we don't return all data in the GET all endpoint?

If so maybe this is more consistent with what we usually do?
```suggestion
export const useReleasePlanTemplate = (templateId: string) => {
```

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm.tsx
```diff
@@ -0,0 +1,57 @@
...
+                label='Template name'
+                value={name}
+                onChange={(e) => setName(e.target.value)}
+                error={Boolean(errors.name)}
+                errorText={errors.name}
+                onFocus={() => clearErrors()}
+                autoFocus
+            />
+            <StyledInputDescription>
+                What's the purpose of this template?
```
nunogois, line 45: I think the Figma sketch does not include labels. I think it's just a detail, but if you'd like you can confirm this.

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+    };
+
+    return (
+        <>
+            <FormTemplate
+                title={`Edit template ${template.name}`}
+                description='Edit a release plan template that makes it easier for you and your team to release features.'
+            >
```
nunogois, line 109: I assume we'll add other props like `formatApiCode` at a later stage?

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
```diff
@@ -0,0 +1,139 @@
...
+        clearErrors();
+        const isValid = validate();
+        if (isValid) {
+            const payload = getTemplatePayload();
+            try {
+                updateReleasePlanTemplate({
+                    ...payload,
+                    id: templateId,
+                    milestones: template.milestones,
+                }).then(() => {
```
nunogois, line 95: Prefer using async/await instead of then.

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+const StyledCancelButton = styled(Button)(({ theme }) => ({
+    marginLeft: theme.spacing(3),
+}));
+
+export const EditReleasePlanTemplate = () => {
+    const releasePlansEnabled = useUiFlag('releasePlans');
+    if (!releasePlansEnabled) {
+        return null;
+    }
```
nunogois, line 63: My gut feeling is that this will cause you issues. Usually you should do this right before the component's `return` at the end, so the same number of hooks always run for each render.

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
```diff
@@ -0,0 +1,139 @@
...
+    transition: 'background-color 0.2s ease-in-out',
+    backgroundColor: theme.palette.background.default,
+    '&:hover': {
+        backgroundColor: theme.palette.neutral.light,
+    },
+    borderRadius: theme.shape.borderRadiusMedium,
+}));
+
+const StyledMilestoneCardBody = styled(Box)(({ theme }) => ({
+    padding: theme.spacing(3, 2, 3, 2),
```
nunogois, line 41:
```suggestion
    padding: theme.spacing(3, 2),
```

frontend/src/component/releases/ReleaseManagement/ReleasePlanTemplateCardMenu.tsx
```diff
@@ -16,7 +16,8 @@ import { TemplateDeleteDialog } from './TemplateDeleteDialog';
 
 export const ReleasePlanTemplateCardMenu = ({
     template,
-}: { template: IReleasePlanTemplate }) => {
+    clickHandler,
```
nunogois, line 19: How about something like `onClick` instead? That's a bit more consistent with how we identify event handlers.

frontend/src/component/releases/ReleaseManagement/ReleasePlanTemplateCard.tsx
```diff
@@ -71,8 +77,16 @@ export const ReleasePlanTemplateCard = ({
...
+                    <StyledMenu
+                        onClick={(e) => {
+                            e.preventDefault();
+                            e.stopPropagation();
+                        }}
+                    >
+                        <ReleasePlanTemplateCardMenu
+                            template={template}
+                            clickHandler={clickHandler}
+                        />
```
nunogois, line 89: This one confuses me a bit. If we always want to navigate on click, why not wrap the card in a link instead?

nunogois (approved):
No hard blocks from me, but I think it's worth addressing some of the comments before merging.

frontend/src/hooks/api/getters/useReleasePlanTemplates/useReleasePlanTemplateInstance.ts
daveleek: Yeah fair point. The issue is we have a useReleasePlanTemplates (plural). I can maybe rename that in a different PR

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm.tsx
daveleek: Yeah, I'm setting up a task to go through UI-things that are off. The sketch doesn't represent how we do forms currently so need to understand if that's a change in direction scoped for this project

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
daveleek: Yeah, we have to look at extending this to fit other type of information as well

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
```diff
@@ -0,0 +1,139 @@
...
+                        name={name}
+                        setName={setName}
+                        description={description}
+                        setDescription={setDescription}
+                        errors={errors}
+                        clearErrors={clearErrors}
+                    />
+
+                    {template.milestones.map((milestone) => (
+                        <StyledMilestoneCard>
```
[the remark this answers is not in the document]
daveleek: Good point, yes there is! I'll fix

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
```diff
@@ -0,0 +1,139 @@
...
+        setName,
+        description,
+        setDescription,
+        errors,
+        clearErrors,
+        validate,
+        getTemplatePayload,
+    } = useTemplateForm(template.name, template.description);
+
+    const handleCancel = () => {};
```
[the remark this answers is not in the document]
daveleek: Yes! Will do

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
daveleek: Oh, interesting! thanks!

frontend/src/component/releases/ReleaseManagement/ReleasePlanTemplateCard.tsx
daveleek: The card is already wrapped in a click, but I have to override that for the menu, and thus also specify an onclick for the menu items

PR #8730, opened 2024-11-13
feat: metrics for password and scim enabled

kwasniew:
## About the changes

Added metrics that tell us if SCIM and password auth are enabled.
It's similar to SAML and OIDC checks we had before. Not we have full picture so that we can check if instance has SSO enabled but password disabled.

### DIFF
--- a/src/lib/features/instance-stats/instance-stats-service.ts
+++ b/src/lib/features/instance-stats/instance-stats-service.ts
@@ -220,11 +220,11 @@ export class InstanceStatsService {
     }
 
     async hasPasswordAuth(): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


src/lib/features/instance-stats/instance-stats-service.ts
```diff
@@ -219,6 +219,22 @@ export class InstanceStatsService {
         return settings?.enabled || false;
     }
 
+    async hasPasswordAuth(): Promise<boolean> {
+        const settings = await this.settingStore.get<{ disabled: boolean }>(
+            'unleash.auth.simple',
+        );
+
+        return settings?.disabled !== false;
```
kwasniew, line 227: password auth has different logic than others (disable instead of enabled check)

PR #8749, opened 2024-11-14
feat: show deleted user sessions

kwasniew:
## About the changes

After log in we show a toast telling user that they were logged out of stale sessions

Screenshot 2024-11-14 at 10 11 20

Next steps:
* add tests for toast


src/test/e2e/services/user-service.e2e.test.ts
```diff
@@ -95,8 +115,9 @@ afterAll(async () => {
     await db.destroy();
 });
 
-afterEach(async () => {
+beforeEach(async () => {
```
kwasniew, line 118: it's a better practice to clean data before not after since it allows debugging failed tests

Tymek (approved):

PR #8755, opened 2024-11-14
Add close button, fix layout, refactor

thomasheartman:
This PR consists of cleanup for the project status modal as well as adding the missing button.

I've added inline comments to explain the different steps.

It now looks like this:
image


frontend/src/assets/icons/projectStatus.svg
thomasheartman, line 1: The icon is still there, it just has no fill color by default. This is to make it easier to style.

frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
thomasheartman, line 1: I've moved section headers up to the project status modal level, so they're easier to align and make sure they share the same styles etc.

I left the container in the first instance because without it we lose the scrolling when it overflows.

frontend/src/component/project/Project/ProjectStatus/ProjectHealth.tsx
thomasheartman, line 1: We use the exact same container three places, so I extracted it.

frontend/src/component/project/Project/ProjectStatus/ProjectHealthGrid.tsx
thomasheartman, line 1: This used to be in `ProjectStatusModal`, but because of how involved it is, I'd rather treat it as "one widget" and extract it into its own file.

frontend/src/component/project/Project/ProjectStatus/ProjectLifecycleSummary.tsx
thomasheartman, line 1: This is mostly just whitespace-change: I've taken the header out and placed it on the status modal level.

sjaanus (approved):

PR #8766, opened 2024-11-15
chore: improve the performance of our instance stats

gastonfournier:
## About the changes
Our stats are used for many places and many times to publish prometheus metrics and some other things.

Some of these queries are heavy, traversing all tables to calculate aggregates.

This adds a feature flag to be able to memoize 1 minute (by default) how long to keep the calculated values in memory. 

We can use the key of the function to individually control which ones are memoized or not and for how long using a numeric variant.

Initially, this will be disabled and we'll test in our instances first


src/lib/features/instance-stats/instance-stats-service.ts
```diff
@@ -58,6 +59,8 @@ export interface InstanceStats {
     strategies: number;
     SAMLenabled: boolean;
     OIDCenabled: boolean;
+    passwordAuthEnabled: boolean;
```
sighphyre, line 62: Why this change? I can't see where these are consumed, only returned. Were these missing before this?

src/lib/features/instance-stats/instance-stats-service.ts
```diff
@@ -193,55 +196,95 @@ export class InstanceStatsService {
         this.trafficDataUsageStore = trafficDataUsageStore;
     }
 
+    memory = new Map<string, () => Promise<any>>();
+    memorize<T>(key: string, fn: () => Promise<T>): Promise<T> {
+        const enabled = this.flagResolver.isEnabled('memorizeStats', {
+            memoryKey: key,
+        });
+        if (enabled) {
+            const variant = this.flagResolver.getVariant('memorizeStats', {
```
sighphyre, line 205: If we resolve the flag then check the variant after I think it does weird things to the metrics. Might be an excellent place to use `variant.feature_enabled` :sweat_smile:

sighphyre (approved):
Cheers, nice change. How often is instance stats collected?

src/lib/features/instance-stats/instance-stats-service.ts
gastonfournier: Oh! Nice suggestion! I wonder if this is clear from our docs :D

src/lib/features/instance-stats/instance-stats-service.ts
gastonfournier: It's returned in line 401. It was previously added to getStats() as a calculation but never returned and never used. The thing is, getStats is only used to sign our stats that we require from our self-hosted customers, so I thought it made sense to have it there

gastonfournier:
> Cheers, nice change. How often is instance stats collected?

Never. I've started exposing methods so you can pick what you need, rather than using all stats and then discarding a lot of values. Well, actually, we used getStats to fetch signed stats that we request from customers.

But the individual values which are now memoized were used for version calculation (we send a bunch of data with the version but that's going to be handled with [this PR](https://github.com/Unleash/unleash/pull/8776)). And it's now being used for personal dashboard and also for some prometheus metrics every minute... so my suspicious is that at startup we query many of these aggregation methods more than once in less than a minute

src/lib/features/instance-stats/instance-stats-service.ts
gastonfournier: Thanks! [33cbd0f](https://github.com/Unleash/unleash/pull/8766/commits/33cbd0f04bccb16b6d091869d1313c672f8304cd) should address that

PR #8768, opened 2024-11-15
feat: create and edit release plan template milestones

daveleek:
Adding and editing milestones and their names

image

### DIFF
--- a/frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
+++ b/frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
@@ -35,7 +35,7 @@ const StyledCancelButton = styled(Button)(({ theme }) => ({
     marginLeft: theme.spacing(3),
 }));
 
-const StyledButton = styled(Button)(({ theme }) => ({
+const StyledAddMilestoneButton = styled(Button)(({ theme }) => ({
     marginTop: theme.spacing(1),
     maxWidth: theme.spacing(20),
 }));
@@ -88,21 +88,6 @@ export const CreateReleasePlanTemplate = () => {
         }
     };
 
-    const updateMilestone = (
-        index: number,
-        milestone: IReleasePlanMilestonePayload,
-    ) => {
-        /*
-        setMilestones(() => {
-            return milestones.map((m, i) => {
-                if (i === index) {
-                    return milestone;
-                }
-                return m;
-            });
-        });*/
-    };
-
     const onSidebarClose = () => {};
 
     if (!releasePlansEnabled) {
@@ -130,7 +115,7 @@ export const CreateReleasePlanTemplate = () => {
                     errors={errors}
                     clearErrors={clearErrors}
                 />
-                <StyledButton
+                <StyledAddMilestoneButton
                     variant='text'
                     color='primary'
                     onClick={(e) => {
@@ -145,7 +130,7 @@ export const CreateReleasePlanTemplate = () => {
                     }}
                 >
                     + Add milestone
-                </StyledButton>
+                </StyledAddMilestoneButton>
                 <StyledButtonContainer>
                     <CreateButton name='template' permission={ADMIN} />
                     <StyledCancelButton onClick={handleCancel}>
--- a/frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
+++ b/frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
@@ -49,7 +49,7 @@ export const CreateReleasePlanTemplate = () => {
     const [milestones, setMilestones] = useState<
         IReleasePlanMilestonePayload[]
     >([{ name: 'Milestone 1', sortOrder: 0 }]);
-    const [addStrategyOpen, setAddStrategyOpen] = useState<boolean>(false);
+    const [addStrategyOpen, setAddStrategyOpen] = useState(false);
     const {
         name,
         setName,
--- a/frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
+++ b/frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
@@ -12,6 +12,12 @@ import { scrollToTop } from 'component/common/util';
 import useToast from 'hooks/useToast';
 import { formatUnknownError } from 'utils/formatUnknownError';
 import { useUiFlag } from 'hooks/useUiFlag';
+import ReleaseTemplateIcon from '@mui/icons-material/DashboardOutlined';
+import type { IReleasePlanMilestonePayload } from 'interfaces/releasePlans';
+import { MilestoneList } from './MilestoneList';
+import { useState } from 'react';
+import { SidebarModal } from 'component/common/SidebarModal/SidebarModal';
+import { ReleasePlanTemplateAddStrategyForm } from './ReleasePlanTemplateAddStrategyForm';
 
 const StyledForm = styled('form')(() => ({
     display: 'flex',
@@ -29,12 +35,21 @@ const StyledCancelButton = styled(Button)(({ theme }) => ({
     marginLeft: theme.spacing(3),
 }));
 
+const StyledButton = styled(Button)(({ theme }) => ({
+    marginTop: theme.spacing(1),
+    maxWidth: theme.spacing(20),
+}));
+
 export const CreateReleasePlanTemplate = () => {
     const releasePlansEnabled = useUiFlag('releasePlans');
     usePageTitle('Create release plan template');
     const { setToastApiError, setToastData } = useToast();
     const navigate = useNavigate();
     const { createReleasePlanTemplate } = useReleasePlanTemplatesApi();
+    const [milestones, setMilestones] = useState<
+        IReleasePlanMilestonePayload[]
+    >([{ name: 'Milestone 1', sortOrder: 0 }]);
+    const [addStrategyOpen, setAddStrategyOpen] = useState<boolean>(false);
     const {
         name,
         setName,
@@ -57,7 +72,10 @@ export const CreateReleasePlanTemplate = () => {
         if (isValid) {
             const payload = getTemplatePayload();
             try {
-                const template = await createReleasePlanTemplate(payload);
+                const template = await createReleasePlanTemplate({
+                    ...payload,
+                    milestones,
+                });
                 scrollToTop();
                 setToastData({
                     type: 'success',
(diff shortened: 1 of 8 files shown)


frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
```diff
@@ -70,33 +88,69 @@ export const CreateReleasePlanTemplate = () => {
         }
     };
 
+    const onSidebarClose = () => {};
```
nunogois, line 91: Should we implement this one right away?

frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
```diff
@@ -70,33 +88,69 @@ export const CreateReleasePlanTemplate = () => {
...
+                    onClick={(e) => {
+                        e.preventDefault();
+                        setMilestones([
+                            ...milestones,
+                            {
+                                name: `Milestone ${milestones.length + 1}`,
+                                sortOrder: milestones.length,
+                            },
+                        ]);
+                    }}
```
nunogois, line 130: Should we extract this to a `addMilestone` method?

frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
```diff
@@ -70,33 +88,69 @@ export const CreateReleasePlanTemplate = () => {
...
             >
-                <StyledForm onSubmit={handleSubmit}>
-                    <TemplateForm
-                        name={name}
-                        setName={setName}
-                        description={description}
-                        setDescription={setDescription}
-                        errors={errors}
-                        clearErrors={clearErrors}
+                <>
```
nunogois, line 146: No need for this when there's only one child

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
```diff
@@ -0,0 +1,109 @@
...
+    errors,
+    clearErrors,
+}: IMilestoneCard) => {
+    const [milestoneName, setMilestoneName] = useState<string>(milestone.name);
+    const [editMode, setEditMode] = useState<boolean>(false);
+    const theme = useTheme();
+
+    useEffect(() => {
+        milestone.name = milestoneName;
+    }, [milestoneName]);
```
nunogois, line 53: I believe this will cause issues. You should set the state using `setMilestones` instead. Ideally this component would have a `onNameChange` or similar method in their props and you would handle this in the MilestoneList. MilestoneList can accept the original setMilestones as a prop and use it inside the new `onNameChange` method to update the correct milestone in the list.

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const editMilestoneNameClick = () => {
+        setEditMode(true);
+    };
```
nunogois, line 57: If it's going to be this simple then it's probably not needed and can be inlined.

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
```diff
@@ -0,0 +1,109 @@
...
+                        {!editMode && (
+                            <StyledMilestoneCardTitle
+                                onClick={editMilestoneNameClick}
+                            >
+                                {milestoneName}
+                            </StyledMilestoneCardTitle>
+                        )}
+                        {!editMode && (
+                            <StyledEditIcon onClick={editMilestoneNameClick} />
+                        )}
```
nunogois, line 94: Why isn't this a part of the above block, since the condition is the same?

frontend/src/component/releases/ReleasePlanTemplate/MilestoneList.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const [currentMilestone, setCurrentMilestone] = useState<number>(-1);
+
+    const showAddStrategyDialog = (index: number) => {
+        setCurrentMilestone(index);
+        setAddStrategyOpen(true);
+    };
+
+    return (
+        <>
+            {milestones.map((milestone, index) => (
```
nunogois, line 27: Try not to rely on the index here. It's going to be especially troublesome if we decide to allow the user to sort / change the order of the milestones.

Ideally you would have unique ids you could use for each milestone, however those are generated only in the backend. My suggestion is [creating temporary uuids](https://github.com/Unleash/unleash/pull/8768/files#diff-67383b99ce55a9de4c95252431abfc962e7410dc62fb3c43d5375ca289406d09R126) in the frontend for state management and omitting them from the final payload.

frontend/src/component/releases/ReleasePlanTemplate/ReleasePlanTemplateAddStrategyForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const StyledButtonContainer = styled('div')(() => ({
+    marginTop: 'auto',
+    display: 'flex',
+    justifyContent: 'flex-end',
+}));
+
+export const ReleasePlanTemplateAddStrategyForm = ({
+    onCancel,
+}: { onCancel: () => void }) => {
```
nunogois, line 21: Extract into `IReleasePlanTemplateAddStrategyFormProps` for consistency?

frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
```diff
@@ -29,12 +35,21 @@ const StyledCancelButton = styled(Button)(({ theme }) => ({
...
+    maxWidth: theme.spacing(20),
+}));
+
 export const CreateReleasePlanTemplate = () => {
     const releasePlansEnabled = useUiFlag('releasePlans');
     usePageTitle('Create release plan template');
     const { setToastApiError, setToastData } = useToast();
     const navigate = useNavigate();
     const { createReleasePlanTemplate } = useReleasePlanTemplatesApi();
+    const [milestones, setMilestones] = useState<
```
nunogois, line 49: Why are we using a new state instead of using the one from `useTemplateForm`?

frontend/src/interfaces/releasePlans.ts
```diff
@@ -24,5 +24,17 @@ export interface IReleasePlanTemplatePayload {
     id?: string;
     name: string;
     description: string;
-    milestones?: IReleasePlanMilestone[];
+    milestones?: IReleasePlanMilestonePayload[];
```
nunogois, line 27: Why is this a payload instead of just `IReleasePlanMilestone`?

nunogois (changes requested):
I believe we should take another look at some of these points before merging:
 - Duplicated code between `CreateReleasePlanTemplate` and `EditReleasePlanTemplate`
 - State management
 - Relying on index for milestone management

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
daveleek: I struggled finding a good abstraction for this as the list has a dynamic amount of elements and it all has to be saved as one big chunk. I'd love to get your thoughts/insights on that

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
daveleek: Fair enough!

frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
daveleek: it's a weird thing. I tried that but I kept getting rerenders/losing them again immediately after adding when using the `useTemplateForm`, this was a way to get across the finish line

frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
daveleek: It's a required prop on the sidebar, but not something I think we'll have use for? the use cases I can find all do some kind of navigation but there's not navigation to make here. could probably just set the prop to be `() => void` instead though

frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
daveleek: I suppose we could, especially if we're making this UI reusable by both pages.

frontend/src/component/releases/ReleasePlanTemplate/CreateReleasePlanTemplate.tsx
```diff
@@ -70,33 +88,69 @@ export const CreateReleasePlanTemplate = () => {
...
+                        setMilestones([
+                            ...milestones,
+                            {
+                                name: `Milestone ${milestones.length + 1}`,
+                                sortOrder: milestones.length,
+                            },
+                        ]);
+                    }}
+                >
+                    + Add milestone
```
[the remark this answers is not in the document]
daveleek: It was an Icon elsewhere I found now, swapping

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
daveleek: Argh, I moved things around and forgot to remove the duplicate condition 🙄

frontend/src/component/releases/ReleasePlanTemplate/EditReleasePlanTemplate.tsx
```diff
@@ -89,52 +78,78 @@ export const EditReleasePlanTemplate = () => {
...
-                        <StyledCancelButton onClick={handleCancel}>
-                            Cancel
-                        </StyledCancelButton>
-                    </StyledButtonContainer>
-                </StyledForm>
-            </FormTemplate>
-        </>
+                </>
+            </SidebarModal>
+        </FormTemplate>
```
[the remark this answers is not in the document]
daveleek: Wasn't sure how this would develop as I went along, and decided to keep them separate until I knew if I could merge them. I'll have another look now and see if I can merge them together

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
nunogois: I think you could have added the onClick handler on the parent component, catching any click in the children, instead of duplicating it.

frontend/src/component/releases/ReleasePlanTemplate/MilestoneList.tsx
```diff
@@ -0,0 +1,76 @@
...
+}));
+
+export const MilestoneList = ({
+    milestones,
+    setMilestones,
+    setAddStrategyOpen,
+    errors,
+    clearErrors,
+}: IMilestoneList) => {
+    const [currentMilestone, setCurrentMilestone] = useState(-1);
```
nunogois, line 30: Why not use `IReleasePlanMilestonePayload | undefined` here? Should be more explicit than `-1` and indexes.

frontend/src/component/releases/ReleasePlanTemplate/MilestoneList.tsx
nunogois: I still think relying on indexes here will bite you down the line, but if you're willing to accept that then feel free to proceed for now.

frontend/src/component/releases/ReleasePlanTemplate/ReleasePlanTemplateAddStrategyForm.tsx
nunogois: Should be an easy one to address.

frontend/src/component/releases/ReleasePlanTemplate/MilestoneList.tsx
```diff
@@ -0,0 +1,76 @@
+import type { IReleasePlanMilestonePayload } from 'interfaces/releasePlans';
+import { useState } from 'react';
+import { MilestoneCard } from './MilestoneCard';
+import { styled } from '@mui/material';
+import { Button } from '@mui/material';
+import Add from '@mui/icons-material/Add';
+
+interface IMilestoneList {
```
nunogois, line 8:
```suggestion
interface IMilestoneListProps {
```

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+const StyledInput = styled(Input)(({ theme }) => ({
+    width: '100%',
+}));
+
+const StyledMilestoneCardTitle = styled('span')(({ theme }) => ({
+    fontWeight: theme.fontWeight.bold,
+    fontSize: theme.fontSizes.bodySize,
+}));
+
+interface IMilestoneCard {
```
nunogois, line 52:
```suggestion
interface IMilestoneCardProps {
```

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm.tsx
```diff
@@ -10,48 +17,99 @@ const StyledInput = styled(Input)(({ theme }) => ({
     marginBottom: theme.spacing(2),
 }));
 
+const StyledForm = styled('form')(() => ({
+    display: 'flex',
+    flexDirection: 'column',
+    height: '100%',
+}));
+
 interface ITemplateForm {
```
nunogois, line 26:
```suggestion
interface ITemplateFormProps {
```

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm.tsx
```diff
@@ -10,48 +17,99 @@ const StyledInput = styled(Input)(({ theme }) => ({
...
+    mode: 'create' | 'edit';
```
nunogois, line 27: Where are we using this?

nunogois (approved):
This is definitely better now, especially the create / edit component part. 

We already discussed the points I'm worried about, so I'll leave you a non-blocking approval and leave the decision up to you. Some of the issues I'm concerned about you'll eventually be forced to tackle, even if it's later down the line.

frontend/src/component/releases/ReleasePlanTemplate/MilestoneList.tsx
daveleek: This one is for adding strategies to the milestone. Current milestone (should probably be named ...index) gets passed to the add strategy form. If we're adding something to that object (like we do with names) it should be bubbled up no?

frontend/src/component/releases/ReleasePlanTemplate/MilestoneCard.tsx
daveleek: Parent is the same for this and the other section which is only shown during edit mode, and would have a weird effect if you click inside. I'll have a think if I can introduce a parent just for these two while doing add strategies

frontend/src/interfaces/releasePlans.ts
daveleek: I'll have a look with add strategies :)

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm.tsx
daveleek: I'll remove as it's not needed for this PR

PR #8769, opened 2024-11-15
fix(a11y): Fix keyboard navigation issues with sidebar

thomasheartman:
This PR fixes a number of keyboard accessibility issues with the feedback sidebar. They are (in no particular order):
1. The radio inputs don't have a focus style for `focus-visible` (when keyboard focused).
2. There's two close buttons there for some reason? One is invisible, but you can tab to it?
3. The sidebar doesn't trap focus, so you can tab out of the modal and continue tabbing through the main page (with the modal still open)
4. The sidebar doesn't steal focus. When you open it, your focus remains on the button you used to open it. So if you want to navigate to it, you have to go through the entire page (behind the modal) to get to it.
5. The sidebar can't be closed by 'escape'.


sjaanus (approved):
Nice!

PR #8772, opened 2024-11-15
1-3121: fix wrong counting for unhealthy flags

thomasheartman:
This PR fixes the counting of unhealthy flags for the project status page. The issue was that we were looking for `archived = false`, but we don't set that flag in the db anymore. Instead, we set the `archived_at` date, which should be null if the flag is unarchived.


src/lib/features/project-status/project-stale-flags-read-model/project-stale-flags-read-model.ts
```diff
@@ -7,7 +7,8 @@ export class ProjectStaleFlagsReadModel implements IProjectStaleFlagsReadModel {
     async getStaleFlagCountForProject(projectId: string): Promise<number> {
         const result = await this.db('features')
             .count()
-            .where({ project: projectId, archived: false })
+            .whereNull('archived_at')
```
thomasheartman, line 10: This was the actual bug. Apparently, we don't set the `archived` flag anymore, so relying on `archived = false` led to lodas of wrong results.

src/lib/features/project-status/projects-status.e2e.test.ts
```diff
@@ -264,36 +264,46 @@ test('project status includes stale flags', async () => {
...
+    function cartesianProduct(...arrays: any[][]): any[][] {
+        return arrays.reduce(
+            (acc, array) => {
+                return acc.flatMap((accItem) =>
+                    array.map((item) => [...accItem, item]),
+                );
+            },
+            [[]] as any[][],
+        );
+    }
```
thomasheartman, line 277: We could also do 4 nested for loops if we prefer, but this solution feels descriptive enough to me, at least.

sjaanus (approved):

PR #8778, opened 2024-11-18
feat: implement call to add release-plan to feature environment

daveleek:
Implements the addReleasePlan api call and invoking it from FeatureReleasePlanCard.
Invokes the release-plan add event even if the call fails


chriswk (approved):
LGTM

PR #8791, opened 2024-11-19
feat: backfill licensed users

sjaanus:
**This migration introduces a query that calculates the licensed user counts and inserts them into the licensed_users table.** 

**The logic ensures that:**

1. All users created up to a specific date are included as active users until they are explicitly deleted.
2. Deleted users are excluded after their deletion date, except when their deletion date falls within the last 30 days or before their creation date.
3. The migration avoids duplicating data by ensuring records are only inserted if they don’t already exist in the licensed_users table.

**Logic Breakdown:**

**Identify User Events (user_events):** Extracts email addresses from user-related events (user-created and user-deleted) and tracks the type and timestamp of the event. This step ensures the ability to differentiate between user creation and deletion activities.
**Generate a Date Range (dates):** Creates a continuous range of dates spanning from the earliest recorded event up to the current date. This ensures we analyze every date, even those without events.
**Determine Active Users (active_emails):** Links dates with user events to calculate the status of each email address (active or deleted) on a given day. This step handles:

- The user's creation date.
- The user's deletion date (if applicable).

**Calculate Daily Active User Counts (result):** 
For each date, counts the distinct email addresses that are active based on the conditions:

- The user has no deletion date.
- The user's deletion date is within the last 30 days relative to the current date.
- The user's creation date is before the deletion date.

### DIFF
--- a/src/migrations/20241119105837-licensed-users-backfill.js
+++ b/src/migrations/20241119105837-licensed-users-backfill.js
@@ -1,6 +1,6 @@
 exports.up = (db, cb) => {
   db.runSql(`
-      WITH user_events AS (
+        WITH user_events AS (
           SELECT
               DISTINCT CASE WHEN type = 'user-deleted' THEN pre_data ->> 'email' ELSE data ->> 'email' END AS email,
                        type,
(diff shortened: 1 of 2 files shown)


src/migrations/20241119105837-licensed-users-backfill.js
```diff
@@ -0,0 +1,85 @@
+exports.up = (db, cb) => {
+  db.runSql(`
+        WITH user_events AS (
```
sjaanus, line 3: Just gets the correct events

src/migrations/20241119105837-licensed-users-backfill.js
```diff
@@ -0,0 +1,85 @@
...
+          SELECT
+              DISTINCT CASE WHEN type = 'user-deleted' THEN pre_data ->> 'email' ELSE data ->> 'email' END AS email,
+                       type,
+                       created_at AS event_date
+          FROM
+              events
+          WHERE
+              type IN ('user-created', 'user-deleted')
+        ),
+        dates AS (
```
sjaanus, line 13: Generates needed dates from first event to today.

src/migrations/20241119105837-licensed-users-backfill.js
```diff
@@ -0,0 +1,85 @@
...
+                   generated_dates
+               WHERE
+                   date + INTERVAL '1 day' <= CURRENT_DATE
+           )
+           SELECT
+               date :: date
+           FROM
+               generated_dates
+        ),
+        active_emails AS (
```
sjaanus, line 32: For each date, for each email, find the most recent created and deleted event.

src/migrations/20241119105837-licensed-users-backfill.js
```diff
@@ -0,0 +1,85 @@
...
+                   CASE WHEN ue.type = 'user-deleted' THEN ue.event_date ELSE NULL END
+               ) AS deleted_date
+           FROM
+               dates d
+                   LEFT JOIN user_events ue ON ue.event_date <= d.date
+           GROUP BY
+               d.date,
+               ue.email
+        ),
+        result AS (
```
sjaanus, line 49: Count the emails on date, 1 if the deleted date is empty, delete happened within last 30 days, delete happened before the creation.

src/test/e2e/licensed-users-backfill.test.ts
```diff
@@ -0,0 +1,204 @@
...
+        await dbm.up('20241119105837-licensed-users-backfill.js');
+
+        const { rows } = await client.query(
+            "SELECT TO_CHAR(date, 'YYYY-MM-DD') AS date, count FROM licensed_users ORDER BY date;",
+        );
+        expect(rows.find((row) => row.date === '2024-10-01').count).toBe(1);
+        expect(rows.find((row) => row.date === '2024-10-02').count).toBe(1);
+        expect(rows.find((row) => row.date === '2024-10-05').count).toBe(1);
+        expect(rows.find((row) => row.date === '2024-11-04').count).toBe(1);
+        expect(rows.find((row) => row.date === '2024-11-05').count).toBe(0); // 30 days has passed
```
thomasheartman, line 97: This might be the same thing as I pointed out in a separate PR, so feel free to ignore if you don't think it's relevant, but October is a 31-day month. So if you create the user on the 1st, and delete them on the 2nd, the 31st will be the 30th day after deletion. Extrapolating, that means that november 3rd is the 30 day mark. Should we switch this to check November 4th? Or should it be 3rd? Again, because this will in reality differentiate on milliseconds, it probably doesn't matter much. I'm sure youv'e got it covered.

If someone is deleted on the 5th, does the 5th count as the first day of 30 or is the 6th the first day?

thomasheartman (approved):

PR #8793, opened 2024-11-19
chore: stop using personal dashboard UI flag

kwasniew:
## About the changes

This PR stops using the flag. The next PR will remove the flag itself. In the transitional period we want to keep the flag returned from backend since all UIs are upgraded to no-flag version.


frontend/src/component/InitialRedirect.tsx
```diff
@@ -19,16 +17,8 @@ export const InitialRedirect = () => {
             return `/projects/${lastViewed}`;
         }
 
-        if (personalDashboardUiEnabled) {
-            return '/personal';
-        }
-
-        if (projects && !lastViewed && projects.length === 1) {
```
kwasniew, line 26: this code was unreachable with a flag on

sjaanus (approved):

PR #8799, opened 2024-11-19
chore: hide release plan template permissions behind feature flag

daveleek:
Hides config of release plan template permissions behind a feature flag.
It's not the prettiest, it checks the label of the section and checks the flag. Happy to take thoughts on improving this

Does not hide any of the template permissions in the permission explanation in the table from roles that have been configured with these permissions even if flag is false.
Enabled permissions do not get removed if re-saved after flag has been turned off


daveleek:
Could potentially strip these in the backend to keep the frontend clean. I'll have a look

nunogois (approved):

PR #8807, opened 2024-11-20
fix: don't break personal dashboard charts if the flag is called `.`

thomasheartman:
This PR fixes an issue where the personal dashboard would fail to render if the flag was called `.` (Curiously, it was not an issue with `..`; probably because they end up accessing different URLs).

I've taken the very pragmatic approach here of saying "right, we know that `.` and `..` cause issues, let's just not even try to fetch data for them".

The option, of course, is to bake in more error handling in the components, but due to how we've got hooks depending on each other, it's a bit of a rabbit hole to go down. I think this is a good compromise for now.

So now, you'll get this instead:
image

I've also gone and updated the text for when we get a metrics fetching error, because this probably isn't due to the flag name anymore. If it is, we want to know.

### DIFF
--- a/frontend/src/component/personalDashboard/FlagMetricsChart.tsx
+++ b/frontend/src/component/personalDashboard/FlagMetricsChart.tsx
@@ -235,7 +235,7 @@ const FlagMetricsChartInner: FC<{
         return (
             <ChartContainer>
                 <PlaceholderFlagMetricsChart
-                    label={`Couldn't fetch metrics for the current flag. This may be a transient error, or your flag name ("${flag.name}") may be causing issues.`}
+                    label={`Couldn't fetch metrics for the current flag right now. Please try again. Report this if it doesn't resolve itself.`}
                 />
             </ChartContainer>
         );
--- a/frontend/src/component/personalDashboard/FlagMetricsChart.tsx
+++ b/frontend/src/component/personalDashboard/FlagMetricsChart.tsx
@@ -215,7 +215,7 @@ export const PlaceholderFlagMetricsChartWithWrapper: React.FC<{
     );
 };
 
-export const FlagMetricsChart: FC<{
+const FlagMetricsChartInner: FC<{
     flag: { name: string; project: string };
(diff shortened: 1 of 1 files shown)


frontend/src/component/personalDashboard/FlagMetricsChart.tsx
```diff
@@ -97,7 +97,7 @@ const EmptyFlagMetricsChart = () => {
 const useMetricsEnvironments = (project: string, flagName: string) => {
     const [environment, setEnvironment] = useState<string | null>(null);
     const { feature } = useFeature(project, flagName);
-    const activeEnvironments = feature.environments.map((env) => ({
+    const activeEnvironments = (feature?.environments ?? []).map((env) => ({
```
kwasniew, line 100: Is the return type of useFeature lying to us? It would indicate the defensive technique is not needed looking at the useFeature return type

kwasniew (approved):

frontend/src/component/personalDashboard/FlagMetricsChart.tsx
thomasheartman: Yeah, good point. It's both lying and we're abusing it.

First, it's wrapping an http request. This means we might not have the data immediately, but we don't wait. Also, there may not be any data there (we might get an error), but we don't handle that (did someone say union types?).

But second, the useFeature hook returns the deprecated `IFeatureToggle` interface. If you check that interface, you'll find that `environments` is always present. But if you compare it to the `FeatureSchema` which supersedes it, you'll see that `environments` is optional (yikes!).

kwasniew: Ideally we should fix the source of truth (the useFeature hook type) so that we don't have to catch those mistakes at runtime

thomasheartman: Aye. I'll make a follow-up PR to gauge how much work that'd be.

PR #8809, opened 2024-11-20
feat: licensed users ui rework

sjaanus:
1. Moved link creation bottom next to licensed users view
2. Created licensed users component
3. Added flag

OSS:
image
All others
image


frontend/src/component/admin/users/InviteLinkBar/InviteLinkBar.tsx
```diff
@@ -15,24 +14,5 @@ export const InviteLinkBar: VFC = () => {
         });
     };
 
-    return (
```
sjaanus, line 18: This bordering and style will be used in 2 places, so moving it to shared component on bottom. Also all of these stylings were not needed.

frontend/src/component/admin/users/UsersAdmin.tsx
```diff
@@ -21,7 +20,6 @@ export const UsersAdmin = () => {
                         index
                         element={
                             <>
-                                <InviteLinkBar />
```
sjaanus, line 24: Moving inside UsersList

frontend/src/component/admin/users/UsersHeader/LicensedUsersBox.tsx
```diff
@@ -0,0 +1,49 @@
...
+const StyledLink = styled(Link)(({ theme }) => ({
+    fontSize: theme.spacing(1.75),
+}));
+
+export const LicensedUsersBox = () => {
+    return (
+        <StyledContainer>
+            <StyledRow>
+                <Typography variant='body1' fontWeight='bold'>
+                    11/25
```
sjaanus, line 24: Mock data.

frontend/src/component/admin/users/UsersHeader/LicensedUsersBox.tsx
```diff
@@ -0,0 +1,49 @@
...
+                        </Box>
+                    }
+                />
+            </StyledRow>
+            <StyledRow>
+                <Typography variant='body2'>
+                    {' '}
+                    Seats used last 30 days
+                </Typography>
+                <StyledLink onClick={() => {}}>
```
sjaanus, line 42: onClick be implemented in following PRs.

frontend/src/component/admin/users/UsersHeader/LicensedUsersBox.tsx
```diff
@@ -0,0 +1,43 @@
...
+                            A licensed seat is a unique user that had access to
+                            your instance within the last 30 days, and thereby
+                            occupied a seat.
+                        </Box>
+                    }
+                />
+            </StyledRow>
+            <StyledRow>
+                <Typography variant='body2'>Seats used last 30 days</Typography>
+                <StyledLink onClick={() => {}}>View graph over time</StyledLink>
```
thomasheartman, line 39: This opens the sidebar, right? If so, it should be a button, not a link. But then again, MUI might be smart enough to convert it for ya. Have you double checked that it's rendered as a button in the resulting HTML?

frontend/src/component/admin/users/UsersHeader/LicensedUsersBox.tsx
thomasheartman, line 1: If I'm reading this correctly, you've organized this box into rows first, right?
image

To me, it looks more like it's organized by columns: left with 11/25 seats used last 30 days, and right with ? View graph over time.

My immediate thought was that it might result in some strange wrapping on narrower displays, where the top and bottom rows don't align properly anymore.

But I think the more important point is that this divorces the number of seats from it's description: I think the most logical order of reading is:
"11/25" -> "seats used last 30 days" -> "(?)" -> "view graph over time". And in fact, "11/25" and "seats used last 30 days" should probably be the same `p` element (separate internal spans for styling and breaking).

Also, we might want to associate the help icon with the text through `aria-details` (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-details), but ... we don't do that anywhere else, and I wouldn't wanna mess that up, so maybe not if we don't have the time to read properly into it.

This can of course be done in a followup.

frontend/src/component/admin/users/UsersHeader/LicensedUsersBox.tsx
```diff
@@ -0,0 +1,43 @@
+import { Box, Link, styled, Typography } from '@mui/material';
+import { HelpIcon } from 'component/common/HelpIcon/HelpIcon';
+
+const StyledContainer = styled(Box)(({ theme }) => ({
```
thomasheartman, line 4: Just a suggestion: this can totally be an [`article`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) if you want. Doesn't matter much, though, but it's a standalone component that you could anywhere.

frontend/src/component/admin/users/UsersHeader/UsersHeader.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { LicensedUsersBox } from './LicensedUsersBox';
+
+interface StyledContainerProps {
+    licensedUsersEnabled: boolean;
+}
+
+const StyledContainer = styled(Box)<StyledContainerProps>(
+    ({ theme, licensedUsersEnabled }) => ({
+        display: 'grid',
+        gridTemplateColumns: licensedUsersEnabled ? '60% 40%' : '100%',
```
thomasheartman, line 14: We can also use fractions here, which is more idiomatic for grid (_gridiomatic_?) (but this comment is most important because of that awful pun, so do what you will)

```suggestion
        gridTemplateColumns: licensedUsersEnabled ? '3fr 2fr' : '100%',
```

thomasheartman (approved):

frontend/src/component/admin/users/UsersHeader/LicensedUsersBox.tsx
sjaanus: Done

PR #8821, opened 2024-11-21
1-3145: counteract the MUI button's negative margin

thomasheartman:
This change styles the project status svg for this by giving it a
negative margin in this case.

This isn't a native MUI icon, so handling it is a bit tricky. That
said, the ideal solution would be to make the icon conform better to
MUI standards, but this is a quick fix for now.

Before:
image

After:
image


sjaanus (approved):

thomasheartman:
Changes were included in https://github.com/Unleash/unleash/pull/8823.

PR #8830, opened 2024-11-21
chore: add traffic overage banner

nunogois:
https://linear.app/unleash/issue/2-2995/add-a-global-traffic-overage-banner-to-unleash

Adds a new global traffic overage banner to Unleash.

Introduces a new banner for Pro or PAYG instances, displayed only when one of the following conditions is met:
 - The total included requests for your plan have been exceeded
 - We estimate that the total included requests for your plan will be exceeded

The banner provides links to our [pricing & plans](https://www.getunleash.io/pricing) and [fair use policy](https://www.getunleash.io/fair-use-policy) pages. Additionally, it includes a "see data usage" link that redirects users to `/admin/network/data-usage` in their Unleash instance, where detailed data usage can be reviewed.

To avoid redundancy, the overage warning on the Network Data Usage page has been removed, as it is now covered by the new global traffic overage banner. In its place, a recommendation to set up Edge is now always displayed for plans with limited requests.

image

### DIFF
--- a/frontend/src/component/App.tsx
+++ b/frontend/src/component/App.tsx
@@ -23,6 +23,7 @@ import { LicenseBanner } from './banners/internalBanners/LicenseBanner';
 import { Demo } from './demo/Demo';
 import { LoginRedirect } from './common/LoginRedirect/LoginRedirect';
 import { SecurityBanner } from './banners/internalBanners/SecurityBanner';
+import { TrafficOverageBanner } from './banners/TrafficOverageBanner';
 
 const StyledContainer = styled('div')(() => ({
     '& ul': {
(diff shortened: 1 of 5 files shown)


ivarconr:
I understand why you are doing it, and the intention is good. 

But what if the customer wants to use a lot of traffic? We have customers who are fine paying for the additional traffic, as it is important for them to avoid any self-hosting complexities. Its not like AWS give me a lot of banners just because we have a lot of traffic towards our instance. 

It makes sense the first time you go over, as an early warning. But for every month you use Unleash. 

Another way would be to allow admins to subscribe to usage warning emails. (that's how Vercel and AWS and others solve it).

nunogois:
Closing this for now.

PR #8845, opened 2024-11-25
chore: switch to using milestoneId instead of index

daveleek:
Tracks changes using milestoneId and not index. creates a new temp id for new milestones


chriswk (approved):
Nice. Seems like the right way to go.

PR #8846, opened 2024-11-25
task: added unique index for release plan templates

chriswk:
We want to prevent our users from defining multiple templates with the same name. So this adds a unique index on the name column when discriminator is template.


daveleek (approved):
Cool!

PR #8848, opened 2024-11-25
feat: Add PROJECT_ARCHIVED event to send message to Slack

0417taehyun:
### Summary

- Add `PROJECT_ARCHIVED` event on `EVENT_MAP` to use
- Add a test case for `PROJECT_ARCHIVED` event formatting 
- Add `PROJECT_ARCHIVED` event when users choose which events they send to Slack
- Fix Slack integration document by adding `PROJECT_ARCHIVED`

### Example

The example message looks like the image below. I covered my email with a black rectangle to protect my privacy 😄 
The link refers `/projects-archive` to see archived projects.

Slack message example

## Discussion points

I considered the reason why Unleash didn't implement to send `PROJECT_ARCHIVED` message to Slack integration.

One thing I assumed that it is impossible to create a new project with open source codes, which means it is only enabled in the enterprise plan. However, [document](https://docs.getunleash.io/reference/integrations/slack-app#events) explains that users can send `PROJECT_CREATED` and `PROJECT_DELETED` events to Slack, which are also available only in the enterprise plan, hence it means we need to embrace all worthwhile events.

I think it is reasonable to add `PROJECT_ARCHIVED` event to Slack integration because users, especially operators, need to track them through Slack by separating steps, `_CREATED`, `_ARCHIVED`, and `_DELETED`.

### DIFF
--- a/src/lib/addons/__snapshots__/feature-event-formatter-md.test.ts.snap
+++ b/src/lib/addons/__snapshots__/feature-event-formatter-md.test.ts.snap
@@ -184,6 +184,14 @@ exports[`Should format specialised text for events when no specific text for str
 }
 `;
 
+exports[`Should format specialised text for events when project archived 1`] = `
+{
+  "label": "Project archived",
+  "text": "*user@company.com* archived project *my-other-project*",
+  "url": "unleashUrl/projects-archive",
+}
+`;
+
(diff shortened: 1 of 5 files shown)


nunogois (approved):
LGTM - Great catch, thank you for your contribution!

PR #8850, opened 2024-11-25
fix: show data for last 30 days where we say we do

thomasheartman:
This change updates the stat for archived flags "this month".

Turns out we were accessing the wrong property on the data object.

Additionally, changes the label to say "last 30 days" instead of "this
month"  because that's more accurate.


sjaanus (approved):

FredrikOseberg (approved):
LGTM

PR #8852, opened 2024-11-25
feat: view diff in change requests

kwasniew:
## About the changes

Closes #


sjaanus (approved):
LGTM

frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.tsx
```diff
@@ -69,6 +69,11 @@ const StyledContainer: FC<{ children?: React.ReactNode }> = styled('div')(
     }),
 );
 
+const ViewDiff = styled('span')(({ theme }) => ({
```
thomasheartman, line 72: Should this be a button? I'm not sure I understand exactly how this works. Do you click "view diff" to open it? If so, it should definitely render as a button. Do you just hover it? If so, we should make sure you can tab to it correctly.

PR #8853, opened 2024-11-25
fix: text overflow on project users access page

Tymek:
Remove scrollbar in project settings - user roles

before:
image

after:
image

### DIFF
--- a/frontend/src/component/common/Table/cells/RoleCell/RoleCell.tsx
+++ b/frontend/src/component/common/Table/cells/RoleCell/RoleCell.tsx
@@ -4,6 +4,7 @@ import { TooltipLink } from 'component/common/TooltipLink/TooltipLink';
 import { RoleDescription } from 'component/common/RoleDescription/RoleDescription';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 import { styled } from '@mui/material';
+import { Truncator } from 'component/common/Truncator/Truncator';
 
 const StyledRoleDescriptions = styled('div')(({ theme }) => ({
     display: 'flex',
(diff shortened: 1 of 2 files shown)


frontend/src/component/common/TooltipLink/TooltipLink.tsx
```diff
@@ -6,32 +6,49 @@ import {
...
     children: ReactNode;
+    clampText?: boolean;
 }
 
 export const TooltipLink = ({
     tooltip,
     highlighted,
     tooltipProps,
     children,
+    clampText,
```
kwasniew, line 42: does it need to be configurable? when do we want ellipsis vs scrollbar?

frontend/src/component/common/TooltipLink/TooltipLink.tsx
```diff
@@ -6,32 +6,49 @@ import {
...
+    shouldForwardProp: (prop) => prop !== 'highlighted' && prop !== 'clampText',
+})<{ highlighted?: boolean; clampText?: boolean }>(
+    ({ theme, highlighted, clampText }) => ({
+        backgroundColor: highlighted ? theme.palette.highlight : 'transparent',
+        color: theme.palette.text.primary,
+        textDecorationColor: theme.palette.text.disabled,
+        textDecorationStyle: 'dashed',
+        textUnderlineOffset: theme.spacing(0.5),
+        whiteSpace: 'nowrap',
+        ...(clampText
```
kwasniew, line 18: When I was clamp I though we gonna use css clamp so maybe we should pick a term that doesn't have other meaning in CSS?

frontend/src/component/common/TooltipLink/TooltipLink.tsx
thomasheartman: Same question 🙋🏼 But also: is it okay that we don't show the whole text. should we reconfigure the tables instead? 🤔

frontend/src/component/common/TooltipLink/TooltipLink.tsx
Tymek: `overflowEllipsis`?

Tymek:
Re-done with new `Truncator` element from Nuno.

frontend/src/component/common/TooltipLink/TooltipLink.tsx
Tymek: Column width updated a bit, to what can still fit.

kwasniew (approved):

PR #8856, opened 2024-11-26
task: enabled in OSS.

chriswk:
Add a column to projects and environments signifying if they're on even if instance is reverted back to OSS after having been in enterprise

This is still in draft, because I need to add the filters for project queries as well.


src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
```diff
@@ -60,6 +70,8 @@ export default class FeatureToggleClientStore
                 action,
             });
         this.flagResolver = flagResolver;
+        const isTest = process.env.NODE_ENV === 'test';
+        this.isOss = !isEnterprise && ui.environment !== 'pro' && !isTest;
```
gastonfournier, line 74: Instead of having this logic in the constructor, can we do it outside and get this value injected? That might simplify testing

src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
```diff
@@ -72,7 +84,7 @@ export default class FeatureToggleClientStore
         const isPlayground = requestType === 'playground';
         const environment = featureQuery?.environment || DEFAULT_ENV;
         const stopTimer = this.timer(`getAllBy${requestType}`);
-
+        this.logger.info(`Getting all features and we're OSS: ${this.isOss}`);
```
gastonfournier, line 87: Do we need this?

src/migrations/20241125115848-add-enabled-in-oss-to-project-and-environment.js
```diff
@@ -0,0 +1,16 @@
+exports.up = function(db, cb) {
+    db.runSql(`
+        ALTER TABLE environments ADD COLUMN enabled_in_oss BOOLEAN NOT NULL DEFAULT false;
+        ALTER TABLE projects ADD COLUMN enabled_in_oss BOOLEAN NOT NULL DEFAULT false;
```
gastonfournier, line 4: Hacking this would not be difficult... just run a query against the db changing all enabled_in_oss to true... it adds some complexity though

gastonfournier (commented):
Overall looks fine, I'm wondering (not requiring) if moving the calculation of isOss outside the constructor would be better... maybe a method in config object could be central enough

src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
chriswk: yeah. let's do that. That way I can reuse it all the stores. Good feedback :hand:

src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
chriswk: No. will remove

chriswk:
This failing e2e test is real, because it's testing additional projects, and we haven't enabled enterprise here. I'll look into how to solve that.

sighphyre (commented):
I think the approach is okay because I can't see another way but I would really like to see some tests around this, if for no other reason than the likelihood of future patches to adjacent code completely borking the queries

src/migrations/20241125115848-add-enabled-in-oss-to-project-and-environment.js
chriswk: And we hide default by default these days, so it's really only development and production.
Yeah ok, let's be explicit then. Updating

src/lib/features/client-feature-toggles/client-feature-toggle-service.ts
```diff
@@ -27,6 +27,7 @@ export class ClientFeatureToggleService {
         this.logger = getLogger('services/client-feature-toggle-service.ts');
         this.segmentReadModel = segmentReadModel;
         this.clientFeatureToggleStore = clientFeatureToggleStore;
+        const isTest = process.env.NODE_ENV === 'test';
```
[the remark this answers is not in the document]
chriswk: Because my initial thought had the oss check on services, and I forgot to remove the last trace. :)

src/lib/features/feature-toggle/feature-toggle-store.ts
```diff
@@ -439,6 +439,16 @@ export default class FeatureToggleStore implements IFeatureToggleStore {
             : queryBuilder.whereNull('archived_at');
     };
 
+    static filterByProjectsAccessibleByOss: Knex.QueryCallbackWithArgs = (
+        queryBuilder: Knex.QueryBuilder,
+        isOss: boolean,
+    ) => {
+        if (isOss) {
+            return queryBuilder.andWhere('project', '=', 'default');
+        }
```
gastonfournier, line 448: Looks easy to hack, but fair enough

gastonfournier (approved):
Ok, looks good to me, it's still easy to hack, but it's better than not doing anything

PR #8858, opened 2024-11-26
fix: user header layout on narrow screens

thomasheartman:
This PR throws in a number of fixes to the UsersHeader's invite link and licensed users components:

- Change the border colors from the primary purple to being a standard border color
- Fix text / button wrapping in the invite link component. It now wraps such that the button goes onto the row below the text if it cannot fit on the same row. The text within the button will not wrap until it absolutely has to (and is on its own line).
- Fix the wrapping behavior of the licensed users box: the bottom row ("seats used 30 days" and "view graph over time") will now wrap at the same time as the other button in the other box.
- Fix some text sizes within the licensed users box
- Fix the button to look more like a link in the licensed users box

Most of it is pretty standard fare, but I've taken a slightly different route when it comes to the licensed users box component.

I switched the whole component to be a "figure" instead of an article. I was trying to figure out how I could make it behave the way we wanted visually while still linking the "seats over 30 days" to the "11 /25" counter text. The examples on MDN use things such as code snippets, poems, quotes, etc, in addition to the more common image elements. And in a way, 11/25 is a figure representing the number of seats used, so I went with that for now. That said, I'd be very interested to hear some other takes on this.

Now, because the `figcaption` has to be the first or last element of a `figure` element, I had to include the "open sidebar" button inside the caption, which isn't ideal. But I can live with it, I think.

Before:
image
image
image

After:
image
image
image

### DIFF
--- a/frontend/src/component/admin/users/InviteLinkBar/InviteLinkBar.tsx
+++ b/frontend/src/component/admin/users/InviteLinkBar/InviteLinkBar.tsx
@@ -1,6 +1,15 @@
 import type { VFC } from 'react';
 import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
 import { InviteLinkBarContent } from './InviteLinkBarContent';
+import { styled } from '@mui/material';
+
+const Bar = styled('article')(({ theme }) => ({
+    display: 'flex',
+    flexFlow: 'row wrap',
+    gap: theme.spacing(2),
+    justifyContent: 'space-between',
+    alignItems: 'center',
+}));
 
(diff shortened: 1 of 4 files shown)


frontend/src/component/admin/users/InviteLinkBar/InviteLinkBar.tsx
thomasheartman, line 1: I moved the wrapper in here, because without it, it's not a *bar*, but two pieces of contents. It also made more sense for the usersheader component.

sjaanus (approved):

PR #8859, opened 2024-11-26
feat: Change request applied diff for update strategy

kwasniew:
## About the changes

Show correct title diff and payload diff (e.g. rollout %) when change request is applied
Screenshot 2024-11-26 at 10 01 58

Previously we were always comparing title and payload against the current strategy. But after the strategy is applied there's no diff anymore. What's more when the strategy is updated in future we're diffing against a future value. This PR always looks at the snapshot for change request when the change request is applied.
I added tests showing both cases in detail. 

There will be another PR in the backend to take a snapshot when the CR is being applied. 

Also we should add snapshots for deleted strategies since so that we can show diffs for them after applying CRs.
Update: added delete strategies improvements and extract method refactor in this PR

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
@@ -1,5 +1,5 @@
 import type React from 'react';
-import type { VFC, FC, ReactNode } from 'react';
+import type { FC, ReactNode } from 'react';
 import { Box, styled, Tooltip, Typography } from '@mui/material';
 import BlockIcon from '@mui/icons-material/Block';
 import TrackChangesIcon from '@mui/icons-material/TrackChanges';
@@ -20,6 +20,7 @@ import { ConditionallyRender } from 'component/common/ConditionallyRender/Condit
 import { flexRow } from 'themes/themeStyles';
 import { EnvironmentVariantsTable } from 'component/feature/FeatureView/FeatureVariants/FeatureEnvironmentVariants/EnvironmentVariantsCard/EnvironmentVariantsTable/EnvironmentVariantsTable';
 import { ChangeOverwriteWarning } from './ChangeOverwriteWarning/ChangeOverwriteWarning';
+import type { IFeatureStrategy } from 'interfaces/strategy';
 
 export const ChangeItemWrapper = styled(Box)({
     display: 'flex',
@@ -60,10 +61,7 @@ const StyledTypography: FC<{ children?: React.ReactNode }> = styled(Typography)(
     }),
 );
 
-const hasNameField = (payload: unknown): payload is { name: string } =>
-    typeof payload === 'object' && payload !== null && 'name' in payload;
-
-const DisabledEnabledState: VFC<{ show?: boolean; disabled: boolean }> = ({
+const DisabledEnabledState: FC<{ show?: boolean; disabled: boolean }> = ({
     show = true,
     disabled,
 }) => {
@@ -98,7 +96,7 @@ const DisabledEnabledState: VFC<{ show?: boolean; disabled: boolean }> = ({
     );
 };
 
-const EditHeader: VFC<{
+const EditHeader: FC<{
     wasDisabled?: boolean;
     willBeDisabled?: boolean;
 }> = ({ wasDisabled = false, willBeDisabled = false }) => {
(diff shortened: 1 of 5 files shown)


thomasheartman (approved):

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -1,5 +1,5 @@
 import type React from 'react';
-import type { VFC, FC, ReactNode } from 'react';
```
kwasniew, line 2: VFC is deprecated

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -119,7 +117,157 @@ const EditHeader: VFC<{
...
+const hasDiff = (object: unknown, objectToCompare: unknown) =>
+    JSON.stringify(object) !== JSON.stringify(objectToCompare);
+
+const DeleteStrategy: FC<{
+    change: IChangeRequestDeleteStrategy;
+    changeRequestState: ChangeRequestState;
+    currentStrategy: IFeatureStrategy | undefined;
+    actions?: ReactNode;
+}> = ({ change, changeRequestState, currentStrategy, actions }) => {
+    const name =
```
kwasniew, line 129: extracted DeleteStrategy component so that we can do the ifs checks outside of the inline calls

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -119,7 +117,157 @@ const EditHeader: VFC<{
...
+    );
+};
+
+const UpdateStrategy: FC<{
+    change: IChangeRequestUpdateStrategy;
+    changeRequestState: ChangeRequestState;
+    currentStrategy: IFeatureStrategy | undefined;
+    actions?: ReactNode;
+}> = ({ change, changeRequestState, currentStrategy, actions }) => {
+    const hasVariantDiff = hasDiff(
```
kwasniew, line 190: extracted component can have display data calculation outside of the inline calls

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -185,139 +326,35 @@ export const StrategyChange: VFC<{
                         <div>{actions}</div>
                     </ChangeItemCreateEditDeleteWrapper>
                     <StrategyExecution strategy={change.payload} />
-                    <ConditionallyRender
-                        condition={hasVariantDiff}
```
kwasniew, line 189: for add strategy diff checking doesn't make sense

frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.tsx
```diff
@@ -80,29 +78,32 @@ const Truncated = styled('div')(() => ({
 }));
 
 export const StrategyTooltipLink: FC<IStrategyTooltipLinkProps> = ({
-    change,
```
kwasniew, line 83: this was unnecessary coupling

frontend/src/component/changeRequest/ChangeRequest/StrategyTooltipLink/StrategyTooltipLink.tsx
```diff
@@ -80,29 +78,32 @@ const Truncated = styled('div')(() => ({
...
+    name,
```
kwasniew, line 81: strategy name and title passing give us more flexibility in caller code (can select current live strategy vs snapshot depending on CR state)

frontend/src/component/changeRequest/changeRequest.types.ts
```diff
@@ -241,14 +241,15 @@ export type ChangeRequestAddStrategy = Pick<
...
 export type ChangeRequestEditStrategy = ChangeRequestAddStrategy & {
     id: string;
-    snapshot?: Omit<IFeatureStrategy, 'title'> & { title?: string | null };
+    snapshot?: IFeatureStrategy;
 };
 
 type ChangeRequestDeleteStrategy = {
     id: string;
-    name: string;
+    name?: string;
```
kwasniew, line 249: for applied change requests this one is never provided

PR #8860, opened 2024-11-26
Remove admonitions from Proxy docs

melindafekete:
This PR contains two small fixes:
1. [SSO Overview](https://unleash-docs-git-docs-misc-quick-fixes-unleash-team.vercel.app/reference/sso): remove Google Authentication from the list of supported providers in the intro, as it has been deprecated.
2. [Unleash Proxy generated docs](https://unleash-docs-git-docs-misc-quick-fixes-unleash-team.vercel.app/reference/unleash-proxy): remove injected admonition containing tips. Proxy is now in maintenance mode, and we don't want other admonitions to distract from that message

### DIFF
--- a/website/remote-content/edge-proxy.js
+++ b/website/remote-content/edge-proxy.js
@@ -28,10 +28,7 @@ const DOCS = mapObject(enrich)({
 });
 
 const getAdmonitions = (data) => {
-    const admonitions = {
-        'unleash-proxy': ``,
-        'unleash-edge': ``,
-    };
+    const admonitions = {};
 
     return [admonitions[data.slugName]];
 };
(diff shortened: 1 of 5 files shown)


website/remote-content/edge-proxy.js
```diff
@@ -29,11 +29,7 @@ const DOCS = mapObject(enrich)({
 
 const getAdmonitions = (data) => {
     const admonitions = {
-        'unleash-proxy': `:::tip
-
-Looking for how to run the Unleash proxy? Check out the [_how to run the Unleash proxy_ guide](../how-to/how-to-run-the-unleash-proxy.mdx)!
-
-:::`,
+        'unleash-proxy': ``,
```
alvinometric, line 32: What happens if we remove the key? Rather than set an empty value

alvinometric (approved):
🚀

website/remote-content/edge-proxy.js
melindafekete: it's the same thing, I left the keys in there so it would be easier to edit later if we need to inject some admonitions, but I can remove them

PR #8861, opened 2024-11-26
feat: release plan template strategy types, constraints, segments

daveleek:
Adds different types of strategies and their properties to the strategy-add/edit dialog, along with constraints and segments

Skjermbilde 2024-11-26 kl  15 25 55
Skjermbilde 2024-11-26 kl  15 25 36

### DIFF
--- a/frontend/src/component/releases/ReleasePlanTemplate/MilestoneStrategyType.tsx
+++ b/frontend/src/component/releases/ReleasePlanTemplate/MilestoneStrategyType.tsx
@@ -1,0 +1,59 @@
+import type { IFormErrors } from 'hooks/useFormErrors';
+import type { IReleasePlanMilestoneStrategy } from 'interfaces/releasePlans';
+import type { IStrategy } from 'interfaces/strategy';
+import { MilestoneStrategyTypeFlexible } from './MilestoneStrategyTypeFlexible';
+import GeneralStrategy from 'component/feature/StrategyTypes/GeneralStrategy/GeneralStrategy';
+import UserWithIdStrategy from 'component/feature/StrategyTypes/UserWithIdStrategy/UserWithId';
+import DefaultStrategy from 'component/feature/StrategyTypes/DefaultStrategy/DefaultStrategy';
+
+interface IMilestoneStrategyTypeProps {
+    strategy: Omit<IReleasePlanMilestoneStrategy, 'milestoneId'>;
+    strategyDefinition?: IStrategy;
+    parameters: IReleasePlanMilestoneStrategy['parameters'];
+    updateParameter: (field: string, value: string) => void;
+    errors: IFormErrors;
+}
+export const MilestoneStrategyType = ({
+    strategy,
+    strategyDefinition,
+    parameters,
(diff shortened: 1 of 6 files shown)


daveleek:
There's some validation to add, but that will be a separate task

frontend/src/component/releases/ReleasePlanTemplate/MilestoneStrategyType.tsx
```diff
@@ -0,0 +1,59 @@
...
+    if (!strategyDefinition) {
+        return null;
+    }
+
+    switch (strategy.name) {
+        case 'default':
+            return <DefaultStrategy strategyDefinition={strategyDefinition} />;
+        case 'flexibleRollout':
+            return (
+                <MilestoneStrategyTypeFlexible
```
daveleek, line 32: This one in feature strategies uses projectid and environment info, so had to take the relevant parts out and make a separate one specifically for this.

frontend/src/component/releases/ReleasePlanTemplate/MilestoneStrategyType.tsx
```diff
@@ -0,0 +1,59 @@
...
+                    parameters={parameters}
+                    updateParameter={updateParameter}
+                    errors={errors}
+                    editable={true}
+                />
+            );
+        case 'userWithId':
+            return (
+                <UserWithIdStrategy
```
daveleek, line 41: This along with the other UIs (Default and General) are all imported from the feature strategies. Should we make our own here or is this good enough for now?

frontend/src/component/releases/ReleasePlanTemplate/ReleasePlanTemplateAddStrategyForm.tsx
```diff
@@ -127,9 +191,35 @@ export const ReleasePlanTemplateAddStrategyForm = ({
                     )}
                 </StyledTitle>
             </StyledHeaderBox>
+            {!BuiltInStrategies.includes(strategy.name || 'default') && (
+                <StyledAlertBox>
+                    <Alert severity='warning'>
+                        Custom strategies are deprecated. We recommend not
```
daveleek, line 197: Given that custom strategies are deprecated, should we support them here at all?

frontend/src/component/releases/ReleasePlanTemplate/MilestoneStrategyType.tsx
nunogois: This is probably good enough for now, let's come back to this later.

frontend/src/component/releases/ReleasePlanTemplate/ReleasePlanTemplateAddStrategyForm.tsx
nunogois: IMO we can do the same as we do in feature strategies for now and return to this later. But it's good to bring it up.

nunogois (approved):
👍 Following the philosophy of moving forward for now and coming back to clean this up later.

PR #8862, opened 2024-11-26
chore: adapt billing page to custom billing scenarios

nunogois:
https://linear.app/unleash/issue/2-3030/improve-the-billing-page-behavior-for-instances-with-custom-billing

Adapts the billing page to support custom billing scenarios.

Here’s how it will appear in such cases (notice a new "Your billing is managed by Unleash" text, with no visible button to update billing information):

image


ivarconr (approved):
this is nice.

PR #8864, opened 2024-11-27
chore: make milestone_strategies.title nullable

daveleek:
Adds a migration that makes title property of milestone strategies nullable


chriswk (approved):
LGTM

PR #8865, opened 2024-11-27
fix: Use a proper dark theme for the activity graph

thomasheartman:
This PR adds a proper dark theme for the activity graph. We previously used the exact same theme for both light and dark modes.

Before:
image

After (different chart):
image

I'm also passing in the theme explicitly as the `colorScheme` property. Without that prop, the graph uses your system color scheme (according to the docs), which may not be the same as your Unleash theme color scheme.

A side effect of this, however, is that we're now getting small borders on the light theme dots. I haven't found a prop in the docs to control this, but I kinda like them.

Before:
image

After (different chart):
image

### DIFF
--- a/frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
+++ b/frontend/src/component/project/Project/ProjectStatus/ProjectActivity.tsx
@@ -10,6 +10,9 @@ const StyledContainer = styled('div')(({ theme }) => ({
     display: 'flex',
     flexDirection: 'column',
     alignItems: 'center',
+    'svg rect': {
+        stroke: '#0000 !important',
+    },
 }));
 
 type Output = { date: string; count: number; level: number };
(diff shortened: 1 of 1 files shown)


sjaanus (commented):
I do not like the borders, because it draws attention to the empty cells, while previously they faded into background and the dark dots had more focus.

Can we try to fix it one more time?

thomasheartman:
@sjaanus I mean, sure. We can reach into the svg and remove all the rect strokes. It feels kinda hacky, but there doesn't seem to be a different way around it (at least not that I can find in [the docs](https://grubersjoe.github.io/react-activity-calendar/?path=/docs/react-activity-calendar--docs), but please tell me I'm wrong). 

That'd look like this for dark mode:
image

And this for light mode:
image


In reality, at current the stroke for light mode exists, but it is white and at .04 opacity. In other words, it's invisible in light mode. So if we just say the stroke's always invisible, then that should work fine.

sjaanus (approved):
Looks good now.

PR #8870, opened 2024-11-27
feat: add variants to release plan template strategies

daveleek:
Adds variant handling in release plan template milestone strategies


chriswk (approved):
Seems reasonable to me :)

nunogois (approved):

PR #8874, opened 2024-11-27
feat: view diff in edit segment CR

kwasniew:
## About the changes

Screenshot 2024-11-27 at 16 42 59

Changes:
* edit segment has the same View Diff as edit strategy
* removing snapshot from diff comparison. Segment snapshots broke diffs

Next PRs:
* take update segment snapshots on apply change request similarly to update strategy 
* add tests for this UI component
* consider removing delete segment CR visualization since we can only remove unused segments so CRs never take part in segments removal

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/SegmentTooltipLink.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/SegmentTooltipLink.tsx
@@ -35,7 +35,7 @@ export const SegmentDiff: FC<{
             <EventDiff
                 entry={{
                     preData: omit(currentSegment, ['createdAt', 'createdBy']),
-                    data: changeRequestSegment,
+                    data: omit(changeRequestSegment, ['snapshot']),
                 }}
             />
         </StyledCodeSection>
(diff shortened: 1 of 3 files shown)


frontend/src/component/changeRequest/ChangeRequest/SegmentTooltipLink.tsx
```diff
@@ -56,9 +56,15 @@ const StyledContainer: FC<{ children?: React.ReactNode }> = styled('div')(
...
 
+const ViewDiff = styled('span')(({ theme }) => ({
+    color: theme.palette.primary.main,
+    marginLeft: theme.spacing(1),
+}));
+
 const Truncated = styled('div')(() => ({
     ...textTruncated,
     maxWidth: 500,
+    display: 'flex',
```
thomasheartman, line 67: Out of curiosity, what's the flex doing here? Is it just aligning the items on a row (which used to not be necessary because there was only one item)?

thomasheartman (approved):

frontend/src/component/changeRequest/ChangeRequest/SegmentTooltipLink.tsx
kwasniew: yes, preventing elements from being put one below the other. We prefer a horizontal row similar to strategy changes

PR #8875, opened 2024-11-27
feat(cjux-278): maintenance root roles

Tymek:
Custom root roles for changing maintenance mode state and banners.

Internal ticket: [CJUX-278](https://linear.app/unleash/issue/CJUX-278/[roles]-custom-root-roles-missing-admin-responsibilities)

### DIFF
--- a/frontend/src/component/admin/maintenance/index.tsx
+++ b/frontend/src/component/admin/maintenance/index.tsx
@@ -6,10 +6,11 @@ import { Box, styled } from '@mui/material';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 import { MaintenanceTooltip } from './MaintenanceTooltip';
 import { MaintenanceToggle } from './MaintenanceToggle';
+import { UPDATE_MAINTENANCE_MODE } from '@server/types/permissions';
 
 export const MaintenanceAdmin = () => (
     <div>
-        <PermissionGuard permissions={ADMIN}>
+        <PermissionGuard permissions={[ADMIN, UPDATE_MAINTENANCE_MODE]}>
             <MaintenancePage />
         </PermissionGuard>
     </div>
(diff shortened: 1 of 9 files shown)


kwasniew (approved):
How did you test it? Manually?

frontend/src/component/admin/maintenance/index.tsx
```diff
@@ -6,10 +6,11 @@ import { Box, styled } from '@mui/material';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 import { MaintenanceTooltip } from './MaintenanceTooltip';
 import { MaintenanceToggle } from './MaintenanceToggle';
+import { UPDATE_MAINTENANCE_MODE } from '@server/types/permissions';
```
thomasheartman, line 9: I didn't know we could import stuff from the server? Should it go in `AccessProvider/permissions` to be consistent with how we're doing it today? I think only defining it in one place is cool, though, but why is it not something we're already doing?

src/lib/types/permissions.ts
```diff
@@ -141,4 +144,17 @@ export const ROOT_PERMISSION_CATEGORIES = [
...
+    },
+];
+
+// Used on Frontend, to allow admin panel use for users with custom root roles
+export const MAINTENANCE_MODE_PERMISSIONS = [
+    ADMIN,
+    READ_ROLE,
+    READ_CLIENT_API_TOKEN,
+    READ_FRONTEND_API_TOKEN,
+    UPDATE_MAINTENANCE_MODE,
```
thomasheartman, line 159: How does this work? Do you need one of them or all of them? Seems a little strange to gain access to the full admin panel if you have read client API token only? Maybe I'm misunderstanding how this works?

Tymek: We don't have a way to selectively show some tabs (yet). It's a way to at least show all options, and `PermissionGuard` wrapper will do block further UI

image

frontend/src/component/admin/maintenance/index.tsx
Tymek: Thanks for noticing this. It needs refactoring. I'll post another PR, because it turns out there are inconsistencies all over this.

PR #8877, opened 2024-11-28
refactor: extract add strategy diff

kwasniew:
## About the changes

Extract component for AddStrategy 

Screenshot 2024-11-28 at 09 54 56

Also:
* added tests
* making diffing less magical (details inline)

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
@@ -267,6 +267,56 @@ const UpdateStrategy: FC<{
     );
 };
 
+const AddStrategy: FC<{
+    change: IChangeRequestAddStrategy;
+    actions?: ReactNode;
+}> = ({ change, actions }) => {
+    return (
+        <>
+            <ChangeItemCreateEditDeleteWrapper>
+                <ChangeItemInfo>
+                    <Typography
+                        color={
+                            change.payload?.disabled
+                                ? 'action.disabled'
+                                : 'success.dark'
+                        }
+                    >
+                        + Adding strategy:
+                    </Typography>
+                    <StrategyTooltipLink
+                        name={change.payload.name}
+                        title={change.payload.title}
+                    >
+                        <StrategyDiff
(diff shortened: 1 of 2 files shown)


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -295,49 +345,7 @@ export const StrategyChange: FC<{
...
-                            >
-                                + Adding strategy:
-                            </Typography>
-                            <StrategyTooltipLink
-                                name={change.payload.name}
-                                title={change.payload.title}
-                            >
-                                <StrategyDiff
-                                    change={change}
-                                    currentStrategy={currentStrategy}
```
kwasniew, line 316: we were relying on the fact that current strategy is undefined because we look it up by change payload id that happens to be missing in the add strategy change payload. 
`'id' in change.payload && strategy.id === change.payload.id`
This is very difficult to reason about

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -267,6 +267,56 @@ const UpdateStrategy: FC<{
...
+                    >
+                        + Adding strategy:
+                    </Typography>
+                    <StrategyTooltipLink
+                        name={change.payload.name}
+                        title={change.payload.title}
+                    >
+                        <StrategyDiff
+                            change={change}
+                            currentStrategy={undefined}
```
kwasniew, line 293: it's much more obvious when we pass undefined as currentStrategy for the AddStrategy component. You don't have to analyze useCurrentStrategy why it returns undefined

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -267,6 +267,56 @@ const UpdateStrategy: FC<{
     );
 };
 
+const AddStrategy: FC<{
```
kwasniew, line 270: This is extract component refactoring with one minor change described below

PR #8884, opened 2024-11-28
chore: remove release-checks

gastonfournier:
## About the changes
If we're going to be releasing from tip of main, the commits are already tested and linted, there's no need to run all these checks


chriswk (approved):
LGTM. All for removing complexity

PR #8891, opened 2024-11-29
feat: add migration

FredrikOseberg:
Creates a data migration that allows us to remove all hanging permissions tied to deleted environments. In addition, this migration adds a foreign key relationship to environment column that references environments name. See rationale [here](https://github.com/Unleash/unleash/pull/8271).

### DIFF
--- a/src/lib/features/feature-search/feature.search.e2e.test.ts
+++ b/src/lib/features/feature-search/feature.search.e2e.test.ts
@@ -6,7 +6,12 @@ import {
 } from '../../../test/e2e/helpers/test-helper';
 import getLogger from '../../../test/fixtures/no-logger';
 import type { FeatureSearchQueryParameters } from '../../openapi/spec/feature-search-query-parameters';
-import { DEFAULT_PROJECT, type IUnleashStores } from '../../types';
+import {
+    CREATE_FEATURE_STRATEGY,
+    DEFAULT_PROJECT,
+    type IUnleashStores,
+    UPDATE_FEATURE_ENVIRONMENT,
+} from '../../types';
 import { DEFAULT_ENV } from '../../util';
 
 let app: IUnleashTest;
@@ -29,7 +34,7 @@ beforeAll(async () => {
     );
     stores = db.stores;
 
-    await app.request
+    const { body } = await app.request
         .post(`/auth/demo/login`)
         .send({
             email: 'user@getunleash.io',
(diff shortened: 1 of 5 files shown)


kwasniew:
Notable changes:
* the new migration transitively deletes role_permissions when then env is deleted. Then it re-creates the default environment from a file called database.json with hardcoded data
* one option would be to replicate all the 15 role permissions in the database.json when we re-create the default env
* in this PR what I did instead was to capture all role permissions and re-create them in-memory (read query + permissions DB insert)
* some of the tests were relying on non default env permissions being preset even when the env didn't exist (possible because the FK from permissions to envs was missing before the migration). I'm updating all those tests and adding comments inline

src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -43,12 +48,30 @@ beforeAll(async () => {
 
     await app.linkProjectToEnvironment('default', 'development');
 
+    await stores.accessStore.addPermissionsToRole(
```
kwasniew, line 51: we were enabling feature in development and it requires two permissions. Previously they were present from a default migration that adds permissions to default, development and production envs. However we delete all environments now and all role permissions are gone. We restore all 15 permissions for the default env but for all envs that we create explicitly in tests we need to add permissions explicitly

src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -43,12 +48,30 @@ beforeAll(async () => {
...
 
     await app.linkProjectToEnvironment('default', 'production');
+
+    await stores.accessStore.addPermissionsToRole(
+        body.rootRole,
+        [
+            { name: UPDATE_FEATURE_ENVIRONMENT },
+            { name: CREATE_FEATURE_STRATEGY },
+        ],
+        'production',
```
kwasniew, line 73: same as the previous comment for development

src/test/e2e/helpers/database-init.ts
```diff
@@ -110,15 +120,20 @@ export default async function init(
     const testDb = createDb(config);
     const stores = await createStores(config, testDb);
     stores.eventStore.setMaxListeners(0);
+    const defaultRolePermissions = await getDefaultEnvRolePermissions(testDb);
```
kwasniew, line 123: this is the main fix. Read all env permissions for the default environment

src/test/e2e/helpers/database-init.ts
```diff
@@ -110,15 +120,20 @@ export default async function init(
...
     await resetDatabase(testDb);
     await setupDatabase(stores);
+    await restoreRolePermissions(testDb, defaultRolePermissions);
```
kwasniew, line 126: re-apply the permissions

kwasniew: One alternative solution would be to store all 15 permissions in a file. What are the tradeoffs?
* with DB query and insert approach our tests are verifying actual permissions in migrations
* one drawback of DB query and insert is that it's different from all the other data that we seed from database.json

src/test/e2e/services/access-service.e2e.test.ts
```diff
@@ -94,8 +94,6 @@ const createRole = async (rolePermissions: PermissionRef[]) => {
 
 const hasCommonProjectAccess = async (user, projectName, condition) => {
     const defaultEnv = 'default';
-    const developmentEnv = 'development';
```
kwasniew, line 97: those 2 environments were tested directly against user permissions table that has changes and when we delete development and production in DB seed we also loose role permissions for those two environments

src/test/e2e/services/access-service.e2e.test.ts
```diff
@@ -637,15 +571,15 @@ test('should support permission with "ALL" environment requirement', async () =>
     await accessStore.addPermissionsToRole(
         customRole.id,
         [{ name: CREATE_FEATURE_STRATEGY }],
-        'production',
+        'default',
```
kwasniew, line 574: use the env that we actually have in tests

kwasniew (approved):

PR #8895, opened 2024-11-29
chore: improve release plan events and add them to event timeline

nunogois:
https://linear.app/unleash/issue/2-3043/improve-release-plan-events-and-add-them-to-the-event-timeline

Improves release plan events and adds them to the event timeline.

This will break the events in Enterprise but that's okay, we can follow up with the Enterprise PR to fix them.

image


daveleek (approved):
Great!

chriswk (approved):
Cool!

PR #8896, opened 2024-11-29
feat: only display oss included projects/environments when install is oss

chriswk:
Trying again, now with a tested function for resolvingIsOss.

Still want to test this on a pro instance in sandbox before we deploy this to our customers to avoid what happened Friday.


gastonfournier (commented):
Can we add a feature flag to release it in a more controlled way and validate in prod?

chriswk:
Will update this to not filter the client-features query. Let's not mess with SDK results

chriswk:
> Can we add a feature flag to release it in a more controlled way and validate in prod?

Yes. If you can turn the entire thing off with a feature flag though it defeats the purpose of the limitation, which is why I did not add a feature flag for this initially.

gastonfournier:
> > Can we add a feature flag to release it in a more controlled way and validate in prod?
> 
> Yes. If you can turn the entire thing off with a feature flag though it defeats the purpose of the limitation, which is why I did not add a feature flag for this initially.

makes sense

gastonfournier (approved):
LGTM! We'll test carefully before releasing

gastonfournier (approved):
LGTM!

PR #8913, opened 2024-12-04
fix: Change Open API validation middleware to specify and use path parameters

0417taehyun:
### Context

I removed `idNumberMiddleware` and changed to use Open API validation process to validate path parameters in #8734 .

However, Express can't parse path parameters if they don't exist in the path, which is the first argument of the method such as `use`, hence it is impossible to parse them by using `useErrorHandler` method like the one below.

```TypeScript
    useErrorHandler(app: Express): void {
        // We need to pass a specific parameter such as `/:segmentId` as the first argument of the `use` method.
        // ex. `app.use('/:segmentId/, (err, req, res, next) => { ... }`
        app.use((err, req, res, next) => {
            if (err?.status && err.validationErrors) {
                const apiError = fromOpenApiValidationErrors(
                    req,
                    err.validationErrors,
                );

                res.status(apiError.statusCode).json(apiError);
            } else {
                next(err);
            }
        });
    }

```

Nevertheless, we can't pass the specific path parameter to `useErrorHandler` because it is defined on the top level, `app.ts` file. Consequently, we need to handle it on the controller level, `/routs/controller.ts`,  like the one below.

```TypeScript
// Open API validation handler the same as inner logic of `useErrorHandler` method.
const checkOpenAPIValidationError = () => async (err, req, res, next) => {
    if (err?.status && err.validationErrors) {
        const apiError = fromOpenApiValidationErrors(req, err.validationErrors);

        res.status(apiError.statusCode).json(apiError);
    } else {
        next(err);
    }
};

export default class Controller {

    route(options: IRouteOptions): void {
        this.app[options.method](
            options.path,
            storeRequestedRoute,
            checkPermission(options.permission),
            checkPrivateProjectPermissions(),
            this.useContentTypeMiddleware(options),
            this.useRouteErrorHandler(options.handler.bind(this)),
        );

        // Add OpenAPI validation handler with specified path parameter, `options.path`, and now Express can handle them.
        this.app.use(options.path, checkOpenAPIValidationError());
    }
}
```

I moved Open API validation handler to the controller layer to reuse on all services such as project and segments, and also removed unnecessary middleware at the top level, `app.ts`, and method, `useErrorHandler` in `openapi-service.ts`.

#### Before

1  Before

Express cant' parse the path parameter because it doesn't be specified on the `use` method. Therefore, it returns `undefined` as an error message.

#### After

2  After

Express can parse the path parameter because I change to specify it on the controller layer. Accordingly, it returns `test`.

## Discussion points

- I think we don't need to add tests about this change because there are already some tests of the `fromOpenApiValidatoinError`.

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -193,10 +193,6 @@ export default async function getApp(
     // Setup API routes
     app.use(`${baseUriPath}/`, new IndexRouter(config, services, db).router);
 
-    if (services.openApiService) {
-        services.openApiService.useErrorHandler(app);
-    }
-
     if (process.env.NODE_ENV !== 'production') {
         app.use(errorHandler());
     } else {
--- a/src/lib/routes/controller.ts
+++ b/src/lib/routes/controller.ts
@@ -65,7 +65,7 @@ const checkPrivateProjectPermissions = () => async (req, res, next) => {
     return res.status(404).end();
 };
 
-const checkOpenAPIValidationError = async (err, req, res, next) => {
+const openAPIValidationMiddleware = async (err, req, res, next) => {
     if (err?.status && err.validationErrors) {
         const apiError = fromOpenApiValidationErrors(req, err.validationErrors);
 
@@ -126,7 +126,7 @@ export default class Controller {
             this.useRouteErrorHandler(options.handler.bind(this)),
         );
 
-        this.app.use(options.path, checkOpenAPIValidationError);
+        this.app.use(options.path, openAPIValidationMiddleware);
     }
 
     get(
--- a/src/lib/routes/controller.ts
+++ b/src/lib/routes/controller.ts
@@ -65,7 +65,7 @@ const checkPrivateProjectPermissions = () => async (req, res, next) => {
     return res.status(404).end();
 };
 
-const checkOpenAPIValidationError = () => async (err, req, res, next) => {
+const checkOpenAPIValidationError = async (err, req, res, next) => {
     if (err?.status && err.validationErrors) {
         const apiError = fromOpenApiValidationErrors(req, err.validationErrors);
 
@@ -126,7 +126,7 @@ export default class Controller {
             this.useRouteErrorHandler(options.handler.bind(this)),
         );
 
-        this.app.use(options.path, checkOpenAPIValidationError());
+        this.app.use(options.path, checkOpenAPIValidationError);
     }
 
     get(
--- a/src/lib/routes/controller.ts
+++ b/src/lib/routes/controller.ts
@@ -10,6 +10,7 @@ import { type IUnleashConfig, NONE } from '../types';
 import { handleErrors } from './util';
 import requireContentType from '../middleware/content_type_checker';
 import { PermissionError } from '../error';
+import { fromOpenApiValidationErrors } from '../error/bad-data-error';
 import { storeRequestedRoute } from '../middleware/response-time-metrics';
 
 type IRequestHandler<P = any, ResBody = any, ReqBody = any, ReqQuery = any> = (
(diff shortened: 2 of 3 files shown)


src/lib/routes/controller.ts
```diff
@@ -114,6 +125,8 @@ export default class Controller {
             this.useContentTypeMiddleware(options),
             this.useRouteErrorHandler(options.handler.bind(this)),
         );
+
+        this.app.use(options.path, checkOpenAPIValidationError());
```
0417taehyun, line 129: We can reuse this Open API validation handler for all services by using `options.path`.

0417taehyun, line 129: We can reuse this Open API validation handler for all services by using `options.path`.

gastonfournier:
Hi @0417taehyun thanks for the contribution! We'll try to add this to our next week planning and review it.

0417taehyun:
@gastonfournier 

Hi, is there any progress of this?

src/lib/app.ts
```diff
@@ -193,10 +193,6 @@ export default async function getApp(
     // Setup API routes
     app.use(`${baseUriPath}/`, new IndexRouter(config, services, db).router);
 
-    if (services.openApiService) {
-        services.openApiService.useErrorHandler(app);
-    }
```
gastonfournier, line 198: This can easily be removed because useErrorHandler only handles errors identified as coming from openapi

src/lib/routes/controller.ts
```diff
@@ -64,6 +65,16 @@ const checkPrivateProjectPermissions = () => async (req, res, next) => {
...
 
+const checkOpenAPIValidationError = () => async (err, req, res, next) => {
+    if (err?.status && err.validationErrors) {
+        const apiError = fromOpenApiValidationErrors(req, err.validationErrors);
+
+        res.status(apiError.statusCode).json(apiError);
+    } else {
+        next(err);
+    }
+};
```
gastonfournier, line 76: Do you see a problem on having this as a const? It's the same function all the time, we don't have to create new functions for each option path
```suggestion
const checkOpenAPIValidationError = async (err, req, res, next) => {
    if (err?.status && err.validationErrors) {
        const apiError = fromOpenApiValidationErrors(req, err.validationErrors);

        res.status(apiError.statusCode).json(apiError);
    } else {
        next(err);
    }
};
```

src/lib/routes/controller.ts
gastonfournier, line 129:
```suggestion
        this.app.use(options.path, checkOpenAPIValidationError);
```
Based on my previous comment, now just use the constant (we could rename the constant to something more meaningful like `openapiErrorValidationMiddleware` or something similar)

gastonfournier (commented):
I've tested locally and it works fine, just adding some minor tweaks.

src/lib/routes/controller.ts
0417taehyun: @gastonfournier 

I adjust your suggestion on [b4702e2](https://github.com/Unleash/unleash/pull/8913/commits/b4702e2eeff113e548ab27837a553034b768696a) commit.

I just followed the way of `checkPrivateProjectPermissions` function definition like the one below but I agreed with your suggestion.

```TypeScript
const checkPrivateProjectPermissions = () => async (req, res, next) => {
    if (
        !req.checkPrivateProjectPermissions ||
        (await req.checkPrivateProjectPermissions())
    ) {
        return next();
    }
    return res.status(404).end();
};
```

src/lib/routes/controller.ts
0417taehyun: @gastonfournier 

I changed it as the constant on b4702e2 commit and renamed the function on 0b71d03 commit.

gastonfournier (approved):
Looks good to me! (LGTM!) I'm just double checking some details with my colleagues and will merge then!

PR #8916, opened 2024-12-04
Add SOC2 docs

melindafekete:
This PR adds documentation around our [SOC2 compliance](https://unleash-docs-git-docs-add-soc2-unleash-team.vercel.app/using-unleash/compliance/soc2), specifically what Unleash features address each of the SOC2 controls


website/docs/using-unleash/compliance/soc2.mdx
```diff
@@ -0,0 +1,32 @@
...
+ 
+
+## How Unleash features map to SOC2 controls
+
+| SOC2 Control                                | SOC2 Control Description                                                                                             | Unleash Feature                                                                                                                                           |
+|---------------------------------------------|---------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| CC 2.1, CC 7.2 Log management utilized      | The company utilizes a log management tool to identify events that may have a potential impact on the company's ability to achieve its security objectives. | Unleash provides a log of all [events](/reference/events), such as configuration changes and access.                                                                         |
```
melindafekete, line 19: For the access logs, do we mean [login history](https://docs.getunleash.io/reference/login-history)?

website/docs/using-unleash/compliance/soc2.mdx
```diff
@@ -0,0 +1,32 @@
...
+| CC 3.2, CC 7.5, CC 9.1 Continuity and disaster recovery plans tested | The company has a documented business continuity/disaster recovery (BC/DR) plan and tests it at least annually.       | Unleash provides a business continuity disaster recovery (BCDR) policy available to customers in the Trust Center, and annual test results upon request.                                                 |
+| CC 3.4, CC 7.1 Configuration management system established | The company has a configuration management procedure in place to ensure that system configurations are deployed consistently throughout the environment. | Unleash provides Release Plan Templates to implement consistent release sequences aligned with the customer policy. Additionally, the [Change Request](/reference/change-requests) supports 4-eyes approval workflows for changes. |
```
melindafekete, line 23: I'm not sure about this one. On the one hand, I would not like to include features that are still in development. We should only put GA features here, I think. On the other hand, not sure how release templates (or change requests) would be a solution for managing 'system configuration'. If it's about the system configuration of the customer, then I think our general recommendation is that feature flags shouldn't manage configuration. If it's about the system configuration of Unleash, then not sure how change requests or release templates help with that.

If we're talking about the Unleash configuration/setup, then I think import/export and terraform are a better fit here.

website/sidebars.ts
```diff
@@ -555,6 +555,11 @@ const sidebars: SidebarsConfig = {
                             label: 'FedRAMP',
                             id: 'using-unleash/compliance/fedramp',
                         },
+                        {
+                            type: 'doc',
+                            label: 'SOC2',
```
melindafekete, line 560: Do we just say SOC2, or SOC2 Type II?

melindafekete (commented):
@dsusa72 @michaelferranti left a couple of questions for you. thanks!

website/docs/using-unleash/compliance/soc2.mdx
ferrantim: Yes, but more generally login history plus event log since access could be via the API.

website/docs/using-unleash/compliance/soc2.mdx
ferrantim: Configuration management here is a general term for the state of the system.  It is asking if you have a process in place to management system changes. In the case of feature flags that control production state, that is configuration.  So this is ok.  WE can remove the Release plans part for now. Can you remember to add back once we have rolled out?

website/sidebars.ts
ferrantim: Let's do SOC2 Type II. that has the specific and verifiable controls that we are talking about

ferrantim (approved):
LGTM with the couple of comments I added. Thanks!

dsusa72 (commented):
LGTM

website/docs/using-unleash/compliance/soc2.mdx
dsusa72: I agree to remove Release Plans for now, and add it when it reaches GA

PR #8920, opened 2024-12-05
fix: permission button unintended full width behavior

nunogois:
Follow-up to https://github.com/Unleash/unleash/pull/8882

The referenced PR caused an unintended behavior by making the button behave like a normal button would on certain parent containers. Previously, the span wrapper caused a side effect that restricted the button’s width, which we were relying on.

By setting some initial styling properties, this PR should hopefully satisfy both use cases.

image

image

image


FredrikOseberg (approved):
LGTM!

PR #8922, opened 2024-12-05
feat: add variant etag

gastonfournier:
## About the changes
This adds a variant that allows us to control client refreshes in case of need.


chriswk (approved):
Cool

PR #8923, opened 2024-12-05
fix: Concurrency issue when creating a user by using database error

0417taehyun:
### Context

I got `500 Internal Server Error` when trying to create a new user like the image below because the user service layer has a concurrency issue. 

1. Concurrency Issue

The previous validation, using `hasUser` method, can't guarantee to check existing users properly in a concurrent situation because `SELECT` transaction can't retrieve uncommitted data, which refers a read committed isolation. You can simply understand this concurrency issue with the image below.

2. Concurrency Issue Diagram

Therefore, I add `catch` statement to handle the an unique violation error from database because it is thrown by `INSERT` transaction, and add a test case to ensure resolving the concurrency issue. The test will be failed like the image below if it doesn't handle the concurrency situation.

3. Failed Test

## Discussion points

I removed `hasUser` method on `createUser` method because it is unnecessary. It is only used on `upsert` method and test codes. I think it is enough to check an existing user by the unique violation.

### DIFF
--- a/src/lib/services/user-service.ts
+++ b/src/lib/services/user-service.ts
@@ -45,6 +45,7 @@ import { PasswordPreviouslyUsedError } from '../error/password-previously-used';
 import { RateLimitError } from '../error/rate-limit-error';
 import type EventEmitter from 'events';
 import { USER_LOGIN } from '../metric-events';
+import { UNIQUE_CONSTRAINT_VIOLATION } from '../error';
 
 export interface ICreateUser {
     name?: string;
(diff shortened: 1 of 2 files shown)


src/lib/services/user-service.ts
```diff
@@ -239,50 +240,53 @@ class UserService {
...
-
-        await this.eventService.storeEvent(
-            new UserCreatedEvent({
-                auditUser,
-                userCreated,
-            }),
-        );
+            return userCreated;
+        } catch (error) {
+            if (error.code === UNIQUE_CONSTRAINT_VIOLATION) {
```
0417taehyun, line 284: This is how to handle a concurrent problem and replace the previous validation, `hasUser` method. There are several patterns like this on the code base hence I follow them.

src/test/e2e/services/user-service.e2e.test.ts
```diff
@@ -159,7 +159,23 @@ test('should not be allowed to create existing user', async () => {
             { username: 'test', rootRole: adminRole.id },
             TEST_AUDIT_USER,
         ),
-    ).rejects.toThrow(Error);
+    ).rejects.toThrow('User already exists');
+});
+
+test('should not be allowed to create existing user in a concurrency situation', async () => {
```
0417taehyun, line 165: This test case makes `createUser` method to handle a concurrent problem. It uses `Promise` to make the concurrency.

FredrikOseberg:
@0417taehyun 

Thank you for this PR! We're wrapping up some work in this sprint but we'll have a look at this early next week.

0417taehyun:
@FredrikOseberg 

Thanks! Feel free to let me know if I need to change or discuss about something with you.

sjaanus:
@0417taehyun Thank you for your pull request.

I attempted to reproduce the issue locally, but it seems nearly impossible to replicate.

Could you provide more context about how you encountered this issue in a real-life scenario? Specifically, having two user creation calls with the same email within the exact same milliseconds feels like an extremely rare edge case.

Given this, I would assume we have numerous other creation methods that could theoretically face the same problem. However, since this doesn't appear to occur in practice, we haven't implemented try-catch blocks for these cases.

0417taehyun:
@sjaanus 

Hi!

As you said that there are numerous cases that could face the same issue, and I also aforementioned that I want to discuss I reusing the database error handler on the middleware layer, I think it would be better to handle those numerous cases through middleware, which means we can handle them seamlessly, instead of using try-catch blocks.

Before we talk about the fundamental solution like the above, let me introduce my specific situation.

I've try to provider OAuth sing-in process for the users, and they often double-click before creating their account for some reasons. The response by the OAuth providers is sometimes delayed because of their traffic issues, and users can click the sign-in button doubly, which makes this concurrent issue.

I think it is weird that API response returned Internal Server Exception even though they can handle this as a 4xx error, and this pull request is a kind of start to handle them.

sjaanus:
> @sjaanus
> 
> Hi!
> 
> As you said that there are numerous cases that could face the same issue, and I also aforementioned that I want to discuss I reusing the database error handler on the middleware layer, I think it would be better to handle those numerous cases through middleware, which means we can handle them seamlessly, instead of using try-catch blocks.
> 
> Before we talk about the fundamental solution like the above, let me introduce my specific situation.
> 
> I've try to provider OAuth sing-in process for the users, and they often double-click before creating their account for some reasons. The response by the OAuth providers is sometimes delayed because of their traffic issues, and users can click the sign-in button doubly, which makes this concurrent issue.
> 
> I think it is weird that API response returned Internal Server Exception even though they can handle this as a 4xx error, and this pull request is a kind of start to handle them.

While this might be plausible, double clicks usually happen 0.5s-2s later. To have this concurrency, they would need to make 2 clicks within 5ms, and even then it should not happen, the first request should hit before the second. But I will review the PR one one more time and accept it.

src/lib/services/user-service.ts
```diff
@@ -239,50 +240,53 @@ class UserService {
...
+                username,
                 email,
-                Joi.string().email({ ignoreLength: true }),
-                'Email',
-            );
-        }
+                name,
+            });
 
-        const exists = await this.store.hasUser({ username, email });
```
sjaanus, line 254: I believe we should still include the `hasUser` check because READ operations are significantly less expensive than WRITE operations. Additionally, in 99.99% of cases, a READ is sufficient, as the concurrency issue described appears to be an extremely rare edge case.

sjaanus: I am wondering if you removed it because test failed?

sjaanus (changes requested):

src/lib/services/user-service.ts
sjaanus: Thanks! Yes, let's close it for now. I think we don't need to worry about concurrency issues at the moment.

PR #8935, opened 2024-12-09
chore(unl-204): make toasts smaller

thomasheartman:
This PR makes toasts smaller and less intrusive, and gives them a new color scheme.

Changes include:
- new color scheme
- no description, only title
- new padding

I'll remove the description prop in a separate PR and remove code references to it.

### DIFF
--- a/frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
+++ b/frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
@@ -8,7 +8,7 @@ export const useStyles = makeStyles()((theme) => ({
         zIndex: 500,
         margin: '0 0.8rem',
         borderRadius: theme.shape.borderRadiusMedium,
-        padding: theme.spacing(0.5),
+        padding: theme.spacing(1),
         paddingInlineStart: theme.spacing(2),
     },
     innerContainer: {
@@ -17,19 +17,11 @@ export const useStyles = makeStyles()((theme) => ({
     starting: {
         opacity: 0,
     },
-    headerContainer: {
-        display: 'flex',
-        alignItems: 'center',
-    },
     confettiContainer: {
         position: 'relative',
         maxWidth: '600px',
         margin: '0 auto',
         display: 'flex',
-    },
-    textContainer: {
-        marginLeft: '1rem',
-        wordBreak: 'break-word',
     },
     headerStyles: {
         fontSize: theme.typography.body1.fontSize,
--- a/frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
+++ b/frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
@@ -3,12 +3,13 @@ import { makeStyles } from 'tss-react/mui';
 export const useStyles = makeStyles()((theme) => ({
     container: {
         maxWidth: '450px',
-        background: theme.palette.background.paper,
+        background: theme.palette.background.paper, // todo: add new background colors here
         boxShadow: theme.boxShadows.popup,
         zIndex: 500,
         margin: '0 0.8rem',
-        borderRadius: '12.5px',
-        padding: '2rem',
+        borderRadius: theme.shape.borderRadiusMedium,
+        padding: theme.spacing(0.5),
+        paddingInlineStart: theme.spacing(2),
     },
     innerContainer: {
         position: 'relative',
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
```diff
@@ -2,56 +2,42 @@ import { makeStyles } from 'tss-react/mui';
...
         margin: '0 0.8rem',
-        borderRadius: '12.5px',
-        padding: '2rem',
-    },
-    innerContainer: {
-        position: 'relative',
+        maxWidth: '450px',
+        padding: theme.spacing(1),
+        paddingInlineStart: theme.spacing(2),
+        zIndex: 500,
```
kwasniew, line 17: As we touch this code shouldn't we switch zIndex to something from a design system token?

frontend/src/component/common/ToastRenderer/ToastRenderer.tsx
```diff
@@ -36,7 +35,7 @@ const ToastRenderer = () => {
                 right: 0,
                 left: 0,
                 margin: '0 auto',
-                maxWidth: '450px',
+                width: 'fit-content',
```
kwasniew, line 38: is it what you were asking for in slack?

thomasheartman: No, not quite, but it's part of it 💁🏼 This just makes the toast be as wide as it has to instead of always the same width.

frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
```diff
@@ -2,56 +2,42 @@ import { makeStyles } from 'tss-react/mui';
...
-        alignItems: 'center',
-        flexDirection: 'column',
     },
     anim: {
         animation: `$drop 10s 3s`,
     },
     checkMark: {
-        width: '65px',
-        height: '65px',
+        marginInlineStart: theme.spacing(1),
```
kwasniew, line 32: is it to support right-to-left writing mode in this one special case? If not, I'd stick to marginLeft as it's an established pattern in our code and is more familiar/idiomatic in this codebase

frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
```diff
@@ -2,56 +2,42 @@ import { makeStyles } from 'tss-react/mui';
...
+        gap: theme.spacing(2),
         margin: '0 0.8rem',
-        borderRadius: '12.5px',
-        padding: '2rem',
-    },
-    innerContainer: {
-        position: 'relative',
+        maxWidth: '450px',
+        padding: theme.spacing(1),
+        paddingInlineStart: theme.spacing(2),
```
kwasniew, line 16: same comment as for marginInlineStart.

kwasniew (approved):

frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
thomasheartman: It's to not have to think about writing modes, actually. If we ever switch (I doubt it), either to right-to-left or top-to-bottom, then this'll still work in the writing direction. So in my eyes, it's just a beefier version of left. I've used it other places, but don't mind switching it out if people hate it.

frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
```diff
@@ -2,56 +2,42 @@ import { makeStyles } from 'tss-react/mui';
 
 export const useStyles = makeStyles()((theme) => ({
     container: {
-        maxWidth: '450px',
-        background: theme.palette.background.paper,
+        alignItems: 'center',
+        background:
+            theme.mode === 'light' ? '#201E42' : theme.palette.background.paper,
```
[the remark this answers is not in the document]
thomasheartman: I can add a comment. The color doesn't exist as a value in the design system yet, hence the raw hex value and the inline switch.

frontend/src/component/common/ToastRenderer/Toast/Toast.styles.ts
kwasniew: 99% of the other code is not prepared for this change so I don't see much value in creating sth that stands out. In a consistent codebase every deviation from a norm is a signal that makes the code reader think. If we switch to margin inlines I'd do it consistently across the board. Otherwise I wouldn't bother

PR #8948, opened 2024-12-10
feat: add core feature alert gh action

FredrikOseberg:
Adds a GH action that notifies us that core features were changed. The action listens on paths in the source code where our client api and frontend api are located. If there is a change to these files, we add a comment to the PR to review carefully and add principals and CTO as reviewers.

### DIFF
--- a/src/lib/features/client-feature-toggles/client-feature-toggle-service.ts
+++ b/src/lib/features/client-feature-toggles/client-feature-toggle-service.ts
@@ -40,7 +40,6 @@ export class ClientFeatureToggleService {
             query || {},
         );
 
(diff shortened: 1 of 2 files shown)


src/lib/features/client-feature-toggles/client-feature-toggle-service.ts
```diff
@@ -40,6 +40,7 @@ export class ClientFeatureToggleService {
             query || {},
         );
 
+        console.log('ALTERING CLIENT FEATURES');
```
FredrikOseberg, line 43: Will be removed before merge. Used to trigger the action.

PR #8950, opened 2024-12-10
Chore/increase client api test coverage

FredrikOseberg:
Added more tests around specific plans. Also added snapshot as per our conversation @gastonfournier, but I'm unsure how much value it will give because it seems that the tests should already catch this using respondWithValidation and the OpenAPI schema. The problem here is that empty array is a valid state, so there were no reason for the schema to break the tests.


gastonfournier:
I think the snapshot helps us to identify if we are:
1. Adding a new field to the response
2. Removing a field from the response
3. Filtering out something

I think this is a good improvement, but I'd like to know if this would have helped us detect that there was something off with the PR #8856

chriswk (approved):
LGTM

PR #8952, opened 2024-12-10
fix(1-3173): clear "removed tags" when you bulk update tags

thomasheartman:
This PR fixes a bug wherein the list of tags to remove from a group of tags wouldn't be correctly updated.

## Repro steps
- Add a console log line to  `frontend/src/component/feature/FeatureView/FeatureOverview/ManageTagsDialog/ManageBulkTagsDialog.tsx`'s `ManagebulkTagsDialog`. Log the value of the`payload` variable.
- Pick a flag with no tags.
- Add tag A -> before submitting, you should have one added tag and zero removed flags. After submitting, both should be empty.
- Now remove tag A -> before submitting, you should have one removed tag and zero added tag. After submitting, both should be empty
- Notice that removed flags hasn't been emptied, but still contains tag A.
- Now add tab B -> before submitting, you should have tag B in added and nothing in removed. Notice that tag A is still in removed.

## Discussion points

This gives us both a `clear` and a `reset` event, which is unfortunate because they sound like they do the same thing. I'd suggest renaming the `clear` event (because it doesn't really clear the state completely), but I'm not sure to what. Happy to do that if you have a suggestion.

I have not tested that submission of the form actually resets the state. I spent about 45 minutes looking at it, but couldn't find a way that was sensible and worked (considered spying: couldn't make it work; considered refactoring and extracting components: think that's too much of a change). I think this is benign enough that it can go without a test for that thing actually being called.

I did, however, test the different reducer commands.

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/ManageTagsDialog/ManageBulkTagsDialog.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/ManageTagsDialog/ManageBulkTagsDialog.tsx
@@ -112,7 +112,7 @@ export const ManageBulkTagsDialog: FC<IManageBulkTagsDialogProps> = ({
         removedTags: [],
     });
 
-    const submitAndClear = () => {
+    const submitAndReset = () => {
         onSubmit(payload);
         dispatch({ type: 'reset' });
     };
@@ -241,7 +241,7 @@ export const ManageBulkTagsDialog: FC<IManageBulkTagsDialogProps> = ({
             secondaryButtonText='Cancel'
             primaryButtonText='Save tags'
             title='Update feature flag tags'
-            onClick={submitAndClear}
+            onClick={submitAndReset}
             disabledPrimaryButton={
                 payload.addedTags.length === 0 &&
                 payload.removedTags.length === 0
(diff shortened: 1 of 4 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/ManageTagsDialog/ManageBulkTagsDialog.tsx
```diff
@@ -106,6 +112,11 @@ export const ManageBulkTagsDialog: VFC<IManageBulkTagsDialogProps> = ({
         removedTags: [],
     });
 
+    const submitAndClear = () => {
+        onSubmit(payload);
+        dispatch({ type: 'reset' });
+    };
```
Tymek, line 118: maybe `submitAndReset`? Clear is related to removing also "initialValues"

Tymek (approved):

frontend/src/component/feature/FeatureView/FeatureOverview/ManageTagsDialog/ManageBulkTagsDialog.tsx
thomasheartman: Yes, good point. I agree with this. But do you have a different word for clear, because having both is really confusing me. Maybe "resetToInitialValues" or something?

PR #8953, opened 2024-12-10
fix: remove unused vulnerable transitive dependency 'prompt'

ivarconr:
My goal was to remove the "revalidator" dependency coming as a transitive dependency from db-migrate. Everything seems to still work on my machine...


chriswk (approved):
LGTM

PR #8959, opened 2024-12-11
task: Remove docker package setup.

chriswk:
Keeping the index.js inside the docker folder to avoid our tsconfig having to be changed. In addition, updated the build to copy less of the folders into our final image, since we really only needed build and node_modules.

Should reduce attack surface of the container.


gastonfournier (approved):
Looks great!

PR #8961, opened 2024-12-11
fix: make project ui query optimized

sjaanus:
From 13 seconds to 0.1 seconds.

1. Joining 1 million events to projects/features is slow. **Solved by using CTE.**
2. Running grouping on 1 million rows is slow. **Solved by adding index.**


chriswk (approved):
Nice, wasn't aware you could do order on column on index creation.

gastonfournier (approved):
Great improvement! Make sure we still return the same data, but :clap:

PR #8964, opened 2024-12-12
feat: make segments editing work with template milestone strategies

daveleek:
Makes segment editing work in template milestone strategies

### DIFF
--- a/frontend/src/component/releases/ReleasePlanTemplate/ReleasePlanTemplateAddStrategyForm.tsx
+++ b/frontend/src/component/releases/ReleasePlanTemplate/ReleasePlanTemplateAddStrategyForm.tsx
@@ -152,9 +152,30 @@ export const ReleasePlanTemplateAddStrategyForm = ({
             ? String(currentStrategy.parameters.stickiness)
             : 'default';
 
+    const segmentsMap = allSegments?.reduce(
+        (acc, segment) => {
+            acc[segment.id] = segment;
+            return acc;
+        },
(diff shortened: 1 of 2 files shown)


frontend/src/component/releases/ReleasePlanTemplate/ReleasePlanTemplateAddStrategyForm.tsx
```diff
@@ -152,10 +152,31 @@ export const ReleasePlanTemplateAddStrategyForm = ({
...
+            );
+        }
     }, []);
 
+    useEffect(() => {
+        setCurrentStrategy((prev) => ({
+            ...prev,
+            segments: segments.map((segment) => segment.id),
+        }));
+    }, [JSON.stringify(segments)]);
```
gastonfournier, line 178:
```suggestion
    }, segments);
```
why not just this? I think this is the state for React... maybe @nunogois will know better

gastonfournier (approved):
Looks fine, just a minor comment, I think it will work anyway

frontend/src/component/releases/ReleasePlanTemplate/ReleasePlanTemplateAddStrategyForm.tsx
daveleek: Ah thanks, I think you're right. This is not a deep object, so shouldn't have to use JSON.stringify here. I'll fix

PR #8967, opened 2024-12-12
feat: first revision of delta api

sjaanus:


src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -72,13 +73,17 @@ export default class ConfigurationRevisionService extends EventEmitter {
                 'Updating feature configuration with new revision Id',
                 revisionId,
             );
-            this.emit(UPDATE_REVISION, revisionId);
             this.revisionId = revisionId;
+            this.emit(UPDATE_REVISION, revisionId);
```
sjaanus, line 77: We should not send out event, before the local revisionId has been sent. This caused out of sync issues.

FredrikOseberg (approved):
LGTM

PR #8969, opened 2024-12-12
fix: do not initialize cache when flag is off

sjaanus:
This feature is still in early alpha. We do not want to populate any cache on startup. We might not want to do it ever.


src/lib/features/client-feature-toggles/cache/client-feature-toggle-cache.ts
```diff
@@ -158,7 +163,9 @@ export class ClientFeatureToggleCache {
     }
 
     private async onUpdateRevisionEvent() {
-        await this.pollEvents();
+        if (this.flagResolver.isEnabled('deltaApi')) {
+            await this.pollEvents();
```
sjaanus, line 167: Only poll when flag is enabled

FredrikOseberg (approved):

PR #8974, opened 2024-12-13
Fix/template edit UI snags

daveleek:
Milestone accordions no longer jump when opened and closed
The disturbing divider line is gone
Introduces an "edit mode" to the strategy, currently only affects button text.
You can now remove milestones from templates
Updates the text on the create/edit template save button


chriswk (approved):
LGTM

PR #8976, opened 2024-12-13
Feat/refactor polling

FredrikOseberg:
This PR refactors the method that listens on revision changes:
- Now supports all environments
- Removed unnecessary populate cache method

# Discussion point

In the listen method, should we implement logic to look into which environments the events touched? By doing this we would: 
- Reduce cache size
- Save some memory/CPU if the environment is not initialized in the cache, because we could skip the DB calls.

### DIFF
--- a/src/lib/features/client-feature-toggles/cache/client-feature-toggle-cache.ts
+++ b/src/lib/features/client-feature-toggles/cache/client-feature-toggle-cache.ts
@@ -182,48 +182,37 @@ export class ClientFeatureToggleCache {
     }
 
     public async pollEvents() {
+        const keys = Object.keys(this.cache);
+
+        if (keys.length === 0) return;
         const latestRevision =
             await this.configurationRevisionService.getMaxRevisionId();
 
-        if (this.currentRevisionId === 0) {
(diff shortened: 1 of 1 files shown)


src/lib/features/client-feature-toggles/cache/client-feature-toggle-cache.ts
```diff
@@ -177,53 +177,42 @@ export class ClientFeatureToggleCache {
...
+        const changedToggles = [
+            ...new Set(
+                changeEvents
+                    .filter((event) => event.featureName)
+                    .map((event) => event.featureName!),
+            ),
+        ];
+        const newToggles = await this.getChangedToggles(
+            changedToggles,
+            latestRevision, // TODO: this should come back from the same query to not be out of sync
```
sjaanus, line 205: This looks ok, but I was thinking about if we might not want to return the revisionId, from query .

Because the flow is getMaxRevisionId(), based on which we get the events, and if now with those events, go pull data and revision ID has changed, it will get out sync. I think best is

If we getMaxRevisionId, we use that across all the calls we make.

sjaanus (approved):

PR #8978, opened 2024-12-13
feat: deleted feature names should come from event

sjaanus:
This is still raw and experimental.
We started to pull deleted features from event payload.
Now we put full query towards read model.

Co-Author:  @FredrikOseberg


src/lib/features/client-feature-toggles/cache/client-feature-toggle-cache.test.ts
```diff
@@ -154,28 +134,3 @@ test('revision equal to the base case returns only later revisions ', () => {
         removed: [],
     });
 });
-
-test('project filter removes features not in project', () => {
```
sjaanus, line 158: We are not filtering by project anymore after we do the cache key change.

FredrikOseberg (approved):
LGTM

PR #8979, opened 2024-12-13
feat: enable docker version tag on main

gastonfournier:
Since we're releasing from main now...

### DIFF
--- a/.github/workflows/docker_publish.yaml
+++ b/.github/workflows/docker_publish.yaml
@@ -47,9 +47,9 @@ jobs:
           flavor: latest=${{ github.event.inputs.is-latest-version || 'auto' }}
           tags: |
             # only enabled for workflow dispatch except main (assume its a release):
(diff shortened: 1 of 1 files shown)


chriswk (commented):
Just make sure we're not pushing version, major.minor, major on every push to main. I see it says workflow_dispatch as event name, but maybe we should also check if we have a tag?

gastonfournier:
> Just make sure we're not pushing version, major.minor, major on every push to main. I see it says workflow_dispatch as event name, but maybe we should also check if we have a tag?

We are using input.version to figure out if checking out the tag or main https://github.com/Unleash/unleash/blob/374a4cf0422bb120e64ce57cb591a560dd9f28e6/.github/workflows/docker_publish.yaml#L29-L36 

So I used the same here [374a4cf](https://github.com/Unleash/unleash/pull/8979/commits/374a4cf0422bb120e64ce57cb591a560dd9f28e6)

chriswk (approved):
LGTM

PR #8980, opened 2024-12-13
feat: add prom metrics

FredrikOseberg:
This PR adds prometheus metrics that allows us to see whether or not tags and namePrefix is used at all in our cloud offering.


sjaanus (approved):

chriswk (approved):
LGTM

PR #8984, opened 2024-12-16
fix: now null sdks will also be handled nicely

sjaanus:
We had some error logs, that when SDK is null, then the application overview query was failing.

This solves it

### DIFF
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -325,7 +325,7 @@ export default class ClientApplicationsStore
                         'COUNT(DISTINCT ci.instance_id) as unique_instance_count',
                     ),
                     this.db.raw(
-                        'ARRAY_AGG(DISTINCT ci.sdk_version) as sdk_versions',
+                        'ARRAY_AGG(DISTINCT ci.sdk_version) FILTER (WHERE ci.sdk_version IS NOT NULL) as sdk_versions',
                     ),
                     this.db.raw('MAX(ci.last_seen) as latest_last_seen'),
                 ])
@@ -394,7 +394,7 @@ export default class ClientApplicationsStore
                 env = {
                     name: environment,
                     instanceCount: Number(unique_instance_count),
-                    sdks: sdk_versions,
+                    sdks: sdk_versions || [],
                     lastSeen: latest_last_seen,
                     issues: {
                         missingFeatures: featuresNotMappedToProject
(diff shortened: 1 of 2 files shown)


src/lib/db/client-applications-store.ts
```diff
@@ -325,7 +325,7 @@ export default class ClientApplicationsStore
                         'COUNT(DISTINCT ci.instance_id) as unique_instance_count',
                     ),
                     this.db.raw(
-                        'ARRAY_AGG(DISTINCT ci.sdk_version) as sdk_versions',
+                        'ARRAY_AGG(DISTINCT ci.sdk_version) FILTER (WHERE ci.sdk_version IS NOT NULL) as sdk_versions',
```
sjaanus, line 328: If the sdk_version is null, do not add it to list.

src/lib/db/client-applications-store.ts
```diff
@@ -394,7 +394,7 @@ export default class ClientApplicationsStore
                 env = {
                     name: environment,
                     instanceCount: Number(unique_instance_count),
-                    sdks: sdk_versions,
+                    sdks: sdk_versions || [],
```
sjaanus, line 397: If list is null, then use empty list.

src/lib/openapi/spec/application-environment-instances-schema.ts
```diff
@@ -24,6 +24,7 @@ export const applicationEnvironmentInstancesSchema = {
                     },
                     sdkVersion: {
                         type: 'string',
+                        nullable: true,
```
sjaanus, line 27: SDK version can be null

src/lib/db/client-applications-store.ts
```diff
@@ -412,7 +412,7 @@ export default class ClientApplicationsStore
             return acc;
         }, []);
         environments.forEach((env) => {
-            env.sdks?.sort();
+            env.sdks.sort();
```
gastonfournier, line 415: According to `Server failed executing request: {} Cannot read properties of null (reading 'sort')","name":"controller/MetricsController","plan":"Enterprise","region":"eu-central-1","stack":"TypeError: Cannot read properties of null (reading 'sort')\n    at /unleash/node_modules/unleash-server/dist/lib/db/client-applications-store.js:330:22\n    at Array.forEach (<anonymous>)\n    at ClientApplicationsStore.mapApplicationOverviewData (/unleash/node_modules/unleash-server/dist/lib/db/client-applications-store.js:329:22)\n    at ClientApplicationsStore.getApplicationOverview (/unleash/node_modules/unleash-server/dist/lib/db/client-applications-store.js:291:21)` there seems to be some edge cases where this is still undefined

PR #8985, opened 2024-12-16
fix: check for enterprise in the create/edit templates UI

daveleek:
Hiding with enterprise tag in routes wasn't enough anymore as we're showing those to Pro with upgrade messages. This temporarily disables the UI completely.

Before
Skjermbilde 2024-12-16 kl  11 26 36

After
image

Also disables the create screen if you know the direct url:
Skjermbilde 2024-12-16 kl  14 49 48

### DIFF
--- a/frontend/src/component/releases/ReleaseManagement/ReleaseManagement.tsx
+++ b/frontend/src/component/releases/ReleaseManagement/ReleaseManagement.tsx
@@ -10,11 +10,19 @@ import { useNavigate } from 'react-router-dom';
 import { useReleasePlanTemplates } from 'hooks/api/getters/useReleasePlanTemplates/useReleasePlanTemplates';
 import { EmptyTemplatesListMessage } from './EmptyTemplatesListMessage';
 import { ReleasePlanTemplateList } from './ReleasePlanTemplateList';
+import { useUiFlag } from 'hooks/useUiFlag';
+import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 
 export const ReleaseManagement = () => {
     usePageTitle('Release management');
     const navigate = useNavigate();
     const data = useReleasePlanTemplates();
+
+    const { isEnterprise } = useUiConfig();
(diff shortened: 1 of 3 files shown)


gastonfournier (commented):
Can you add screenshots before & after just to understand the PR better?

daveleek:
@gastonfournier thanks, yeah updated description now

gastonfournier (changes requested):
I don't think this improves the UI. If we don't want to show it, we should remove the item in the left side, making the UI blank doesn't seem like a good solution

frontend/src/component/releases/ReleaseManagement/ReleaseManagement.tsx
```diff
@@ -32,7 +40,7 @@ export const ReleaseManagement = () => {
                                 }}
                                 maxWidth='700px'
                                 permission={CREATE_RELEASE_TEMPLATE}
-                                disabled={false}
+                                disabled={!isEnterprise()}
```
gastonfournier, line 43: :+1:

frontend/src/component/releases/ReleaseManagement/ReleaseManagement.tsx
```diff
@@ -10,12 +10,20 @@ import { useNavigate } from 'react-router-dom';
...
 export const ReleaseManagement = () => {
     usePageTitle('Release management');
     const navigate = useNavigate();
     const data = useReleasePlanTemplates();
 
+    const { isEnterprise } = useUiConfig();
+    const releasePlansEnabled = useUiFlag('releasePlans');
+    if (!releasePlansEnabled) {
+        return null;
+    }
```
gastonfournier, line 25: I think this is fine if the feature is disabled you should not be able to see this... I don't think it's needed due to https://github.com/Unleash/unleash/blob/3696420c51dc9d673c6da1393c4d0a61b8c3af24/frontend/src/component/menu/routes.ts#L290-L292 but :shrug:

gastonfournier (approved):
LG! Thanks for addressing the comment!

PR #8987, opened 2024-12-16
feat: authentication configuration permission

Tymek:
## About the changes

New permission for SSO config.


kwasniew (approved):
I'm assuming this PR is only adding the data but we don't expose it yet

PR #8990, opened 2024-12-17
feat: segment delta

sjaanus:
Now the delta endpoint also always returns all the segments.


src/lib/features/client-feature-toggles/tests/client-feature-toggles.e2e.test.ts
```diff
@@ -338,4 +362,5 @@ test('should match with /api/client/delta', async () => {
         .expect(200);
 
     expect(body.features).toMatchObject(deltaBody.updated);
+    expect(body.segments).toMatchObject(deltaBody.segments);
```
sjaanus, line 365: Test that the segments structure looks same as the client/features.

PR #8996, opened 2024-12-18
feat: read logs and update cors maintenance root-role permissions

Tymek:
## About the changes
Additional granular permissions related to instance-level access.

- CORS settings
- Reading logs (both instance logs and login history)


src/lib/routes/admin-api/config.ts
```diff
@@ -198,6 +218,30 @@ class ConfigController extends Controller {
...
+                req.body.frontendApiOrigins,
+                req.audit,
+            );
+            res.sendStatus(204);
+            return;
+        }
+
+        throw new NotFoundError();
+    }
 }
```
Tymek, line 245: Previous "setUiConfig" is a generic name, but it only sets CORS. For easy transition into endpoint with proper permissions it's easier to create new one, and avoid potential errors or confusion about permissions in the future.

gastonfournier:
Hi @Tymek what's the reasoning behind this PR? It's hard for me to understand why this change (i.e. context)

Tymek:
@gastonfournier It's to enable creating custom root-roles, that have access to only some responsibilities of Admin user.

Ticket: https://linear.app/unleash/issue/CJUX-278/[roles]-custom-root-roles-missing-admin-responsibilities

src/lib/routes/admin-api/config.ts
```diff
@@ -99,6 +100,7 @@ class ConfigController extends Controller {
             ],
         });
 
+        // TODO: deprecate when removing `granularAdminPermissions` flag
```
gastonfournier, line 103: This would be a breaking change so we have to delay removing it. Meanwhile I believe you can flag the route as deprecated (for OpenAPI doc)

src/lib/features/frontend-api/frontend-api-service.ts
```diff
@@ -208,6 +208,23 @@ export class FrontendApiService {
...
 
+    async setFrontendCorsSettings(
+        value: FrontendSettings['frontendApiOrigins'],
+        auditUser: IAuditUser,
+    ): Promise<void> {
+        const error = validateOrigins(value);
+        if (error) {
+            throw new BadDataError(error);
+        }
+        const settings = (await this.getFrontendSettings()) || {};
```
gastonfournier, line 219: I noticed get frontend settings uses a cached value by default so either:
```suggestion
        const settings = (await this.getFrontendSettings(false)) || {};
```

or 


```suggestion
        const settings = (await this.fetchFrontendSettings()) || {};
```

src/lib/openapi/spec/set-cors-schema.ts
```diff
@@ -0,0 +1,20 @@
+import type { FromSchema } from 'json-schema-to-ts';
+
+export const setCorsSchema = {
+    $id: '#/components/schemas/setCorsSchema',
+    type: 'object',
+    additionalProperties: false,
+    description: 'Unleash configuration settings affect the admin UI.',
```
gastonfournier, line 7:
```suggestion
    description: 'Unleash CORS configuration.',
```

src/lib/routes/admin-api/config.ts
gastonfournier: Alternatively we could use a PATCh on the uiConfig, but we're not using PATCH that frequently... but arguably CORS is a configuration on itself, maybe it's fine not having it as part of uiConfig

gastonfournier (commented):
In general it looks good, but I think there might be a problem with cached values that might override changes made by others. Happy to approve after you check that comment

gastonfournier (approved):
LG!

PR #9000, opened 2024-12-19
chore: restrict on RBAC middleware

gastonfournier:


chriswk (approved):
Yeah. Better idea than identifying all stores/models accessing this data.

PR #9001, opened 2024-12-19
feat: store memory footprints to grafana

sjaanus:
When there is new revision, we will start storing memory footprint for old client-api and the new delta-api.
We will be sending it as prometheus metrics.

The memory size will only be recalculated if revision changes, which does not happen very often.

### DIFF
--- a/src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
+++ b/src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
@@ -173,12 +173,8 @@ export default class FeatureController extends Controller {
                 await this.clientFeatureToggleService.getActiveSegmentsForClient();
 
             try {
-                console.log(`storing features ${features.length}`);
                 const featuresSize = this.getCacheSizeInBytes(features);
                 const segmentsSize = this.getCacheSizeInBytes(segments);
-                console.log(
-                    `features storing ${featuresSize}, ${segmentsSize}`,
-                );
                 this.clientFeaturesCacheMap.set(
                     JSON.stringify(query),
                     featuresSize + segmentsSize,
--- a/src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
+++ b/src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
@@ -35,6 +35,7 @@ import {
 import type ConfigurationRevisionService from '../feature-toggle/configuration-revision-service';
 import type { ClientFeatureToggleService } from './client-feature-toggle-service';
 import {
+    CLIENT_FEATURES_MEMORY,
     CLIENT_METRICS_NAMEPREFIX,
     CLIENT_METRICS_TAGS,
 } from '../../internals';
@@ -68,6 +69,8 @@ export default class FeatureController extends Controller {
     private flagResolver: IFlagResolver;
 
     private eventBus: EventEmitter;
+
+    private clientFeaturesCacheMap = new Map<string, number>();
 
     private featuresAndSegments: (
         query: IFeatureToggleQuery,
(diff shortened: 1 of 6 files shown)


src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -69,6 +70,8 @@ export default class FeatureController extends Controller {
 
     private eventBus: EventEmitter;
 
+    private clientFeaturesCacheMap = new Map<string, number>();
```
sjaanus, line 73: We can not read the size of memoizee directly, so we will be storing the memory usage for each query key into new cache. This will be calculated on new revision id.

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -162,6 +165,32 @@ export default class FeatureController extends Controller {
     private async resolveFeaturesAndSegments(
         query?: IFeatureToggleQuery,
     ): Promise<[FeatureConfigurationClient[], IClientSegment[]]> {
+        if (this.flagResolver.isEnabled('deltaApi')) {
```
sjaanus, line 168: Only if flag on, do this logic, otherwise go default. We will start very gradual rollout.

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -162,6 +165,32 @@ export default class FeatureController extends Controller {
...
+
+            try {
+                const featuresSize = this.getCacheSizeInBytes(features);
+                const segmentsSize = this.getCacheSizeInBytes(segments);
+                this.clientFeaturesCacheMap.set(
+                    JSON.stringify(query),
+                    featuresSize + segmentsSize,
+                );
+
+                await this.clientFeatureToggleService.getClientDelta(
```
sjaanus, line 183: We are just making the same call in delta API to request for cache update.

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -162,6 +165,32 @@ export default class FeatureController extends Controller {
...
+            const features =
+                await this.clientFeatureToggleService.getClientFeatures(query);
+
+            const segments =
+                await this.clientFeatureToggleService.getActiveSegmentsForClient();
+
+            try {
+                const featuresSize = this.getCacheSizeInBytes(features);
+                const segmentsSize = this.getCacheSizeInBytes(segments);
+                this.clientFeaturesCacheMap.set(
```
sjaanus, line 178: Saving the current key into cache.

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -162,6 +165,32 @@ export default class FeatureController extends Controller {
...
+                    featuresSize + segmentsSize,
+                );
+
+                await this.clientFeatureToggleService.getClientDelta(
+                    undefined,
+                    query!,
+                );
+                this.storeFootprint();
+            } catch (e) {
+                this.logger.error('Delta diff failed', e);
```
sjaanus, line 189: Putting all in try catch if something fails for extra safety.

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -162,6 +165,32 @@ export default class FeatureController extends Controller {
...
+                const featuresSize = this.getCacheSizeInBytes(features);
+                const segmentsSize = this.getCacheSizeInBytes(segments);
+                this.clientFeaturesCacheMap.set(
+                    JSON.stringify(query),
+                    featuresSize + segmentsSize,
+                );
+
+                await this.clientFeatureToggleService.getClientDelta(
+                    undefined,
+                    query!,
```
thomasheartman, line 185: When will query not be present? This `!` is a little icky

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
thomasheartman: Again: can query be undefined here (because we force it with `!` later)?

So is this the size of the query *without* delta or with?

thomasheartman: Ah, I see now. This is specified in the store footprint method. This is for features.

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
```diff
@@ -294,4 +305,20 @@ export class ClientFeatureToggleDelta {
     private async updateSegments(): Promise<void> {
         this.segments = await this.segmentReadModel.getActiveForClient();
     }
+
+    storeFootprint() {
+        try {
+            const featuresMemory = this.getCacheSizeInBytes(this.delta);
```
thomasheartman, line 311: How many of these do we keep at the same time? Help me wrap my head around this: can the delta be different for each connected client (assuming they're polling and you're making a flag change every second)?

presumably this uses the revision id to store something, right, so it doesn't need that. I ... remember you saying something about that.

thomasheartman (approved):

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
sjaanus: We keep a delta for each environment. So all clients that connect from same environment, share it.

PR #9004, opened 2024-12-19
chore: Use 20.18.1 instead of 20.14 as our node base image

chriswk:
We already have 20.18.1 set as the default in our Dockerfile, but we override it with an arg from the workflow, so this brings the workflow up to date with what we use as our default.


gastonfournier (approved):
LGTM!

PR #9045, opened 2024-12-30
chore: new hosted auth handler

nunogois:
https://linear.app/unleash/issue/2-3074/add-sign-in-with-google-to-our-hosted-instances-when-only-password-is
https://linear.app/unleash/issue/2-3078/allow-viewer-access-from-auth-app-for-payg

Adds support for a new unified hosted auth handler.

This new auth handler provides a Google sign in option in case you don't have an SSO provider configured. It also provides a way for auth-app to sign in with Unleash's read-only user for Pro and Enterprise PAYG instances.

Check the PR comments for more details.

### DIFF
--- a/frontend/src/component/admin/auth/AuthSettings.tsx
+++ b/frontend/src/component/admin/auth/AuthSettings.tsx
@@ -1,4 +1,4 @@
-import { Alert, Tab, Tabs } from '@mui/material';
+import { Tab, Tabs } from '@mui/material';
 import { PageContent } from 'component/common/PageContent/PageContent';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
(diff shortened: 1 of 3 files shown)


frontend/src/component/admin/auth/AuthSettings.tsx
```diff
@@ -35,17 +34,14 @@ export const AuthSettings = () => {
...
     ].filter(
         (item) => uiConfig.flags?.googleAuthEnabled || item.label !== 'Google',
     );
 
-    if (isEnterprise()) {
-        tabs.push({
-            label: 'SCIM',
-            component: <ScimSettings />,
-        });
-    }
```
nunogois, line 47: I need some guidance here.

Isn't SSO an Enterprise-only feature anyways? 

AFAIK there's no way to get `authenticationType === 'enterprise'` in a hosted Pro instance. A Pro instance would simply display "Your Unleash instance is managed by the Unleash team."

This PR simplifies our logic to rely on the plan instead of the auth type and protect SSO behind the Enterprise plan.

If we're okay with this, then maybe we should protect the route from showing up at all.

daveleek: I think this is an improvement indeed, there used to be a flag there that got removed, and the isEnterprise check stayed behind. Can probably remove the `const { uiConfig, isEnterprise } = useUiConfig();` line as well

daveleek (approved):
Awesome, looks like an improvement!

ivarconr (approved):

PR #9047, opened 2025-01-02
feat: log warning if there is diff between client/features and delta api

sjaanus:
Log out the diff and warn if there is gap between old logic and new delta.

### DIFF
--- a/src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
+++ b/src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
@@ -195,17 +195,9 @@ export default class FeatureController extends Controller {
                     a.name.localeCompare(b.name),
                 );
 
-                const deepEqualIgnoreOrder = (obj1, obj2) => {
-                    const sortedObj1 = JSON.parse(
-                        JSON.stringify(obj1, Object.keys(obj1).sort()),
-                    );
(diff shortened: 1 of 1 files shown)


src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -180,10 +182,39 @@ export default class FeatureController extends Controller {
                     featuresSize + segmentsSize,
                 );
 
-                await this.clientFeatureToggleService.getClientDelta(
-                    undefined,
-                    query!,
+                const delta =
```
kwasniew, line 185: do you want to run this sequentially or in parallel with the query you compare it against?

kwasniew (approved):

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -407,4 +399,14 @@ export default class FeatureController extends Controller {
         const jsonString = JSON.stringify(value);
         return Buffer.byteLength(jsonString, 'utf8');
     }
+
+    deepEqualIgnoreOrder = (obj1, obj2) => {
```
kwasniew, line 403: types seem to be missing here and this function doesn't need to be attached to class

PR #9049, opened 2025-01-02
migration: read logs permission

Tymek:
Should be merged AFTER https://github.com/Unleash/unleash/pull/8996


gastonfournier (approved):
LG!

PR #9050, opened 2025-01-02
chore: new user avatar tooltip

nunogois:
https://linear.app/unleash/issue/2-3109/improve-avatar-tooltip

I noticed our current user avatar tooltip is a bit poor.

This PR tries to improve it a bit using only the data we already have available, without any drastic changes.

### Before

image

### After

image

### Other examples after the changes

image

image

image

image

### DIFF
--- a/frontend/src/component/common/UserAvatar/UserAvatar.tsx
+++ b/frontend/src/component/common/UserAvatar/UserAvatar.tsx
@@ -8,7 +8,6 @@ import {
 } from '@mui/material';
 import type { IUser } from 'interfaces/user';
 import { forwardRef } from 'react';
-import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 import { HtmlTooltip } from '../HtmlTooltip/HtmlTooltip';
 
 const StyledAvatar = styled(Avatar)(({ theme }) => ({
@@ -26,14 +25,10 @@ const StyledTooltip = styled(Box)(({ theme }) => ({
     gap: theme.spacing(2),
 }));
 
-const StyledTooltipAvatar = styled(Avatar)(({ theme }) => ({
+const StyledTooltipAvatar = styled(StyledAvatar)(({ theme }) => ({
     width: theme.spacing(5),
     height: theme.spacing(5),
-    margin: 'auto',
-    backgroundColor: theme.palette.secondary.light,
-    color: theme.palette.text.primary,
-    fontSize: theme.fontSizes.smallerBody,
-    fontWeight: theme.fontWeight.bold,
+    fontSize: theme.fontSizes.smallBody,
 }));
 
 const StyledTooltipContent = styled(Box)({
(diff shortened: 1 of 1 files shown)


frontend/src/component/common/UserAvatar/UserAvatar.tsx
```diff
@@ -19,6 +21,27 @@ const StyledAvatar = styled(Avatar)(({ theme }) => ({
...
 
+const StyledTooltip = styled(Box)(({ theme }) => ({
+    display: 'flex',
+    gap: theme.spacing(2),
+}));
+
+const StyledTooltipAvatar = styled(Avatar)(({ theme }) => ({
+    width: theme.spacing(5),
+    height: theme.spacing(5),
+    margin: 'auto',
```
thomasheartman, line 32: out of curiosity: what's the margin auto doing here? Does ...  does the avatar have some built-in margins?

frontend/src/component/common/UserAvatar/UserAvatar.tsx
```diff
@@ -19,6 +21,27 @@ const StyledAvatar = styled(Avatar)(({ theme }) => ({
...
+    backgroundColor: theme.palette.secondary.light,
+    color: theme.palette.text.primary,
+    fontSize: theme.fontSizes.smallerBody,
+    fontWeight: theme.fontWeight.bold,
```
thomasheartman, line 36: This, presumably, is to handle cases when we put letters instead of images?

frontend/src/component/common/UserAvatar/UserAvatar.tsx
```diff
@@ -94,19 +119,60 @@ export const UserAvatar = forwardRef<HTMLDivElement, IUserAvatarProps>(
...
-                            <TooltipSecondaryContent>
-                                {tooltip.secondary}
-                            </TooltipSecondaryContent>
-                            <TooltipMainContent>
-                                {tooltip.main}
-                            </TooltipMainContent>
-                        </>
+                        <StyledTooltip>
+                            <ConditionallyRender
+                                condition={Boolean(src || user?.imageUrl)}
```
thomasheartman, line 144: Isn't this condition redundant with the if check above?

frontend/src/component/common/UserAvatar/UserAvatar.tsx
```diff
@@ -56,8 +79,10 @@ const TooltipSecondaryContent = styled('div')(({ theme }) => ({
     color: theme.palette.text.secondary,
     fontSize: theme.typography.body2.fontSize,
 }));
+
 const TooltipMainContent = styled('div')(({ theme }) => ({
     fontSize: theme.typography.body1.fontSize,
+    wordBreak: 'break-word',
```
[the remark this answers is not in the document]
thomasheartman: Oh, and I also have some positive experiences with it. MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

frontend/src/component/common/UserAvatar/UserAvatar.tsx
thomasheartman: Which, ... I'm not sure we'll ever do? Because we only show this if we have src or imageurl?

frontend/src/component/common/UserAvatar/UserAvatar.tsx
thomasheartman: Also ... are we still using conditionally render? 🤔

frontend/src/component/common/UserAvatar/UserAvatar.tsx
nunogois: Not sure, was a copy paste from the existing styled avatar component.

frontend/src/component/common/UserAvatar/UserAvatar.tsx
nunogois: Yep, this made me rethink my approach. I'm [now](https://github.com/Unleash/unleash/pull/9050/commits/201c15a1c0c2d12f477fb5460f0b69cf10a7861d) always showing an avatar in the tooltip with whatever content is in the original avatar, which is probably more reasonable and consistent:

image

image

frontend/src/component/common/UserAvatar/UserAvatar.tsx
nunogois: >Isn't this condition redundant with the if check above?

Yep, it was. Good point.

>Also ... are we still using conditionally render? 🤔

I don't know, are we? 🤷 

I've dropped them in this file in the latest commit.

PR #9055, opened 2025-01-03
feat: rework frontend validation for release plan templates

daveleek:
Adds validation, error messages and UI indication for:
Minimum one strategy per milestone.
At least one milestone per template.
Milestone must have a name.

Moving milestone name editing component into separate component instead of duplicating.
Clicking to edit milestone name no longer triggers accordion movement

image


nunogois (approved):
LGTM. 

Are we planning on adding these validations server-side as well?

daveleek:
Thanks @nunogois! Yes, there will be server-side validation as well for this!

Issue #8934, opened 2024-12-09
PATCH method required for /api/admin/context/{contextField}


preetiakrohilla:
### Describe the feature request

Hi Team,

We have a use case where we are trying to update the Unleash UserId context field in parallel. Here are the steps we follow:

We are using REST APIs to update the context field.

1. **To get the current list of legal values for a given userId context field, use the following GET request:**

curl --header "Authorization: xxxxxxxxxxxxxxxxxx" --header "Content-Type: application/json" http://localhost:4242/api/admin/context/userId
Sample response:
{
  "name": "userId",
  "description": "Allows you to constrain on userId",
  "stickiness": false,
  "sortOrder": 1,
  "legalValues": [
    {
      "value": "aeb27c26-8952-434c-a724-e86fbd124e15",
      "description": "test1"
    }
  ],
  "createdAt": "2023-10-29T13:25:26.078Z"
}

2. **Add the new userId value to the list of legalValues received in the response from step #1**

For example, you may want to add a new user ID like this:
{
  "value": "6880e05c-df57-4a06-af65-418e57754d19",
  "description": "test2"
}

3. **Use a PUT request to update the userId context field with the modified legalValues list:**

curl --request PUT 'http://localhost:4242/api/admin/context/userId' \
--header 'Authorization: xxxxxxxxxxxxxx' \
--header 'Content-Type: application/json' \
--data-raw '{
  "name": "userId",
  "description": "Allows you to constrain on userId",
  "legalValues": [
    {
      "value": "aeb27c26-8952-434c-a724-e86fbd124e15",
      "description": "test1"
    },
    {
      "value": "6880e05c-df57-4a06-af65-418e57754d19",
      "description": "test2"
    }
  ],
  "stickiness": false
}'

**Potential Issue:**
Two concurrent requests might lead to data overwriting on the server, as both requests may fetch the same state in step #1 and update the context field simultaneously. This can result in the loss of changes made by one request when the other request is processed later.

### Solution suggestions

Introduce Patch Method for Unleash Rest API /api/admin/context/{contextField}

FredrikOseberg:
@preetiakrohilla 

Hi.

I can see the use-case for this functionality, however we have other priorities that are more pressing right now. If you would like to contribute this functionality we'd be happy to take a look at it, otherwise we'll have to revisit this in the future.

kwasniew:
[...]

PR #9056, opened 2025-01-03
feat: ability to upsert single legal values

kwasniew:
## About the changes

Ability to add/update a single legal value for the context. 

Important decisions:
* since it allows to add a new legal value or update description of a single legal value I decided to go for the PUT method since we know the identity of the legal value (context name + legal value name combination is unique).
* in the event log I'm using the same CONTEXT_FIELD_UPDATED event instead of creating a new event for single legal value updates 

Addresses: https://github.com/Unleash/unleash/issues/8934

Followup PRs:
* add delete single legal value capability
* wrap all context updates in transactions since they touch context and event log

### DIFF
--- a/src/lib/features/context/context-service.ts
+++ b/src/lib/features/context/context-service.ts
@@ -20,9 +20,10 @@ import {
 } from '../../types';
 import type { IPrivateProjectChecker } from '../private-project/privateProjectCheckerType';
 import type EventService from '../events/event-service';
-import { contextSchema } from '../../services/context-schema';
+import { contextSchema, legalValueSchema } from '../../services/context-schema';
 import { NameExistsError } from '../../error';
 import { nameSchema } from '../../schema/feature-schema';
+import type { LegalValueSchema } from '../../openapi';
 
 class ContextService {
     private eventService: EventService;
@@ -126,7 +127,6 @@ class ContextService {
         );
         const value = await contextSchema.validateAsync(updatedContextField);
 
-        // update
         await this.contextFieldStore.update(value);
 
         const { createdAt, sortOrder, ...previousContextField } = contextField;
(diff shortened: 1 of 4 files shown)


src/lib/features/context/context-service.ts
```diff
@@ -140,6 +140,45 @@ class ContextService {
...
+    async updateContextFieldLegalValue(
+        contextFieldLegalValue: { name: string; legalValue: LegalValueSchema },
+        auditUser: IAuditUser,
+    ): Promise<void> {
+        const contextField = await this.contextFieldStore.get(
+            contextFieldLegalValue.name,
+        );
+        const validatedLegalValue = await legalValueSchema.validateAsync(
+            contextFieldLegalValue.legalValue,
+        );
```
thomasheartman, line 152: Maybe I'm missing something, but why are we using the schema to validate here? Isn't that automatically taken care of at the endpoint level?

src/lib/features/context/context-service.ts
```diff
@@ -140,6 +140,45 @@ class ContextService {
...
+
+        const legalValues = contextField.legalValues
+            ? [...contextField.legalValues]
+            : [];
+
+        const existingIndex = legalValues.findIndex(
+            (legalvalue) => legalvalue.value === validatedLegalValue.value,
+        );
```
thomasheartman, line 160: Oh, my. we store this in a list, huh? I imagine using a map would be better, but it's probably also not something we want to touch now, I guess 😅

src/lib/features/context/context-service.ts
thomasheartman: Oh, also, `legalValueSchema` is not `LegalValueSchema` from `../../openapi`.  I *thought* it looked like Joi, but was a little confused by the import (and didn't look hard enough, apparently).

But everything that the joi schema does, we should be able to do with openapi directly, right? Set string min and max lengths and allow description to be `null`, `undefined`, or an arbitrary string?

thomasheartman (commented):
Regarding your decisions, I don't agree with this one:

> since it allows to add a new legal value or update description of a single legal value I decided to go for the PUT method since we know the identity of the legal value (context name + legal value name combination is unique).

Because we're putting to `:contextField/legalValues` (why isn't it `:contextField/legal-values`?), I would expect the PUT to replace all the legal values (in which case, it should've also been a list). So in this case, because you're only sending a single value and want that added to the collection, I think POST is the right method to use. 

(That way, it'd even make sense to allow sending either a single value or a list of values later, if that's something people want)

On the other hand, if you sent it to `:contextField/legal-values/:legalValueName`, then I think PUT would be the right verb. 

As for reusing the context field updated event: yeah, that sounds right to me.

src/lib/features/context/context-service.ts
thomasheartman: But then again, maybe order is important? 🤷🏼

src/lib/features/context/context-service.ts
kwasniew: It's consistent with our current approach in the context module where openapi validates simple string and joi actual 1-100 characters. I didn't want to introduce a new convention in this PR

src/lib/features/context/context-service.ts
kwasniew: Yes, this PR piggybacks on the current implementation

thomasheartman (approved):

src/lib/features/context/context-service.ts
thomasheartman: Yeah, that's fair. I suspect we'll just forget about it and never touch it, then. But that's fine too. It's a little added complexity, but not a big deal here.

PR #9058, opened 2025-01-03
feat: ability to delete single legal values

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/features/context/context.ts
+++ b/src/lib/features/context/context.ts
@@ -45,6 +45,10 @@ interface ContextParam {
     contextField: string;
 }
 
+interface DeleteLegalValueParam extends ContextParam {
+    legalValue: string;
+}
+
 export class ContextController extends Controller {
     private contextService: ContextService;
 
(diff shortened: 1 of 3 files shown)


src/lib/features/context/context.ts
```diff
@@ -188,6 +192,25 @@ export class ContextController extends Controller {
...
+            method: 'delete',
+            path: '/:contextField/legal-values/:legalValue',
+            handler: this.deleteLegalValue,
+            acceptAnyContentType: true,
+            permission: UPDATE_CONTEXT_FIELD,
+            middleware: [
+                openApiService.validPath({
+                    tags: ['Context'],
+                    summary: 'Delete legal value for the context field',
+                    description: `Endpoint that allows deleting a single custom context field legal value. Does not validate that legal value is not in use, but since context field configuration is stored in a json blob for the strategy, existing strategies are safe.`,
```
thomasheartman, line 205: This is an interesting point. I wonder if we can be clearer. The fact that we store stuff as JSON blobs is an implementation detail that we don't need to expose, I think. Does it prevent you from re-saving strategies that already use the legal value?

```suggestion
                    description: `Removes the specified custom context field legal value. Does not validate that the legal value is not in use and does not remove usage from constraints that use it.`,
```

thomasheartman (approved):

PR #9062, opened 2025-01-06
fix: delta do not return archived as changed

sjaanus:
Our delta API was returning archived feature as updated. Now making sure we do not put `archived-feature `event into `updated` event array.
Also stop returning removed as complex object.

### DIFF
--- a/src/lib/features/client-feature-toggles/client-feature-toggle-service.ts
+++ b/src/lib/features/client-feature-toggles/client-feature-toggle-service.ts
@@ -9,10 +9,8 @@ import type {
 import type { Logger } from '../../logger';
 
 import type { FeatureConfigurationClient } from '../feature-toggle/types/feature-toggle-strategies-store-type';
-import type {
-    RevisionDeltaEntry,
-    ClientFeatureToggleDelta,
-} from './delta/client-feature-toggle-delta';
+import type { ClientFeatureToggleDelta } from './delta/client-feature-toggle-delta';
+import type { ClientFeaturesDeltaSchema } from '../../openapi';
 
 export class ClientFeatureToggleService {
     private logger: Logger;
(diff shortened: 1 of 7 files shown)


src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
```diff
@@ -197,6 +200,9 @@ export class ClientFeatureToggleDelta {
         }
     }
 
+    /**
```
sjaanus, line 203: We needed a way to clear out delta between e2e test runs. We might find a better way in future.

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
```diff
@@ -217,6 +223,7 @@ export class ClientFeatureToggleDelta {
             ...new Set(
                 changeEvents
                     .filter((event) => event.featureName)
+                    .filter((event) => event.type !== 'feature-archived')
```
sjaanus, line 226: Change events should not have archived events. These belong to removed.

src/lib/features/client-feature-toggles/client-feature-toggle-service.ts
```diff
@@ -44,7 +42,7 @@ export class ClientFeatureToggleService {
     async getClientDelta(
         revisionId: number | undefined,
         query: IFeatureToggleQuery,
-    ): Promise<RevisionDeltaEntry | undefined> {
+    ): Promise<ClientFeaturesDeltaSchema | undefined> {
```
sjaanus, line 45: Fix the types. RevisionDeltaEntry has removed as complex objects, but we are returning string array.

PR #9063, opened 2025-01-06
feat: making context service transactional

kwasniew:
## About the changes

Using transactional context service so that context updates and event log are inside the same transaction.


sjaanus (approved):
LGTM

PR #9064, opened 2025-01-06
chore: make grid group cards have a consistent height

nunogois:
https://linear.app/unleash/issue/2-3114/group-cards-should-have-a-consistent-height-in-their-grid

Makes the group cards height consistent in the grid.

image


sighphyre (approved):
Nice!

PR #9067, opened 2025-01-08
feat: Unique connection tracking

kwasniew:
## About the changes

Enhancing request timing metrics with connection id (think instance id but guaranteed to be unique and handled internally by us). 

Details:
* I checked how appName was added in the first place: https://github.com/Unleash/unleash/pull/2117. Same tradeoffs apply here
* connection id is implemented here: https://github.com/Unleash/unleash-client-node/pull/690
* This is behind a flag to opt-into tracking and also it can be disabled with the killswitch for appName
* We have a new `x-unleash-appname` header that replaces `unleash-appname` but not all SDKs use it so we check both with preference given to the x-header
* I'm not using a new `x-unleash-sdk` header here to keep the metrics data as small as possible for now with the ability to track unique clients

Important:
* connection id has higher cardinality than the app name. We have as many unique connection ids as client pods AND frontend api instances (that's why I want this to be opt-in only for the time of debugging)

### DIFF
--- a/src/lib/middleware/response-time-metrics.ts
+++ b/src/lib/middleware/response-time-metrics.ts
@@ -55,12 +55,17 @@ export function responseTimeMetrics(
         // when pathname is undefined use a fallback
         pathname = pathname ?? collapse(req.path);
         let appName: string | undefined;
+        let connectionId: string | undefined;
         if (
             !flagResolver.isEnabled('responseTimeWithAppNameKillSwitch') &&
             (instanceStatsService.getAppCountSnapshot('7d') ??
                 appNameReportingThreshold) < appNameReportingThreshold
         ) {
+            // TODO: upgrade to x-unleash-appname
             appName = req.headers['unleash-appname'] ?? req.query.appName;
+            if (flagResolver.isEnabled('uniqueSdkTracking')) {
+                connectionId = req.headers['x-unleash-connection-id'];
+            }
         }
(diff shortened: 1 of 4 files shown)


src/lib/middleware/response-time-metrics.ts
```diff
@@ -55,12 +55,17 @@ export function responseTimeMetrics(
         // when pathname is undefined use a fallback
         pathname = pathname ?? collapse(req.path);
         let appName: string | undefined;
+        let connectionId: string | undefined;
         if (
             !flagResolver.isEnabled('responseTimeWithAppNameKillSwitch') &&
             (instanceStatsService.getAppCountSnapshot('7d') ??
                 appNameReportingThreshold) < appNameReportingThreshold
         ) {
+            // TODO: upgrade to x-unleash-appname
```
thomasheartman, line 64: We could also give preference to the new header if it's there?

This isn't the right line to give the suggestion on, but like this:
```suggestion
             appName = req.headers['x-unleash-appname'] ?? req.headers['unleash-appname'] ?? req.query.appName;
```

thomasheartman (approved):

src/lib/middleware/response-time-metrics.ts
kwasniew: Very good point. Incorporating the suggestion

PR #9068, opened 2025-01-08
chore: remove simplifyProjectOverview flag

sjaanus:


thomasheartman (approved):
Yessss, this is what we like to see 😄

PR #9071, opened 2025-01-08
chore: mark SSO as an Enterprise only feature

nunogois:
https://linear.app/unleash/issue/2-3124/mark-sso-as-an-enterprise-only-feature

For all intents and purposes, SSO seems to have always been an Enterprise only feature. It just wasn't very clear previously.

See: https://github.com/Unleash/unleash/pull/9045/files#r1899635618

image


gastonfournier (approved):
LGTM! I also checked documentation and it states it's an enterprise feature

PR #9074, opened 2025-01-09
feat: unique connection counting

kwasniew:
## About the changes

E2E POC for unique connection counting using windowed HyperLogLog. We want to know how many unique connections from SDKs we had in the previous hour and in the current hour. 

Screenshot 2025-01-09 at 16 47 45

Algorithm:
* on each request we emit the connection event when the connection id header is detected
* we have unique connection counting service listening on unique connections and updating in-memory HyperLogLog (fixed memory algorithm and data structure for approximate uniqueness tracking)
* we sync the in-memory HyperLogLog to postgres (BYTEA type) so that different pods can cross check uniqueness (merging in-memory with DB)
* in a DB we store current window unique count as byte array HyperLogLog and previous window unique count as another byte array. We swap the buckets every hour accommodating for different corner cases with multiple pods
* we start counting a new current window at every full hour (doing sliding window HyperLogLog is way more difficult than full hour window)

Followup:
* add grafana metric to read the previous bucket every hour


src/lib/middleware/response-time-metrics.ts
```diff
@@ -66,6 +66,11 @@ export function responseTimeMetrics(
                 req.query.appName;
         }
 
+        const connectionId = req.headers['x-unleash-connection-id'];
```
kwasniew, line 69: this is where it starts. on each detected connection id received we let the interested parties know

src/lib/features/unique-connection/unique-connection-service.ts
```diff
@@ -0,0 +1,77 @@
...
+        config: IUnleashConfig,
+    ) {
+        this.uniqueConnectionStore = uniqueConnectionStore;
+        this.logger = config.getLogger('services/unique-connection-service.ts');
+        this.flagResolver = config.flagResolver;
+        this.eventBus = config.eventBus;
+        this.activeHour = new Date().getHours();
+    }
+
+    listen() {
```
kwasniew, line 35: this service is interested in new connection ids received

src/lib/features/unique-connection/unique-connection-service.ts
```diff
@@ -0,0 +1,77 @@
...
+        this.eventBus.on(SDK_CONNECTION_ID_RECEIVED, this.count.bind(this));
+    }
+
+    async count(connectionId: string) {
+        if (!this.flagResolver.isEnabled('uniqueSdkTracking')) return;
+        this.hll.add(HyperLogLog.hash(connectionId));
```
kwasniew, line 41: this data structure tracks unique connections

src/lib/features/scheduler/schedule-services.ts
```diff
@@ -179,4 +180,10 @@ export const scheduleServices = async (
         minutesToMilliseconds(15),
         'cleanUpIntegrationEvents',
     );
+
+    schedulerService.schedule(
```
kwasniew, line 184: sync in-memory HyperLogLogs with DB

src/lib/features/unique-connection/unique-connection-service.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    async sync(): Promise<void> {
```
kwasniew, line 44: this logic will get detailed coverage. We track current and previous HyperLogLog in 2 DB rows. If we found the hour has changed we start a new current bucket and copy current to previous. We have to accommodate for multiple pods where any pod can be the first one to start a current bucket.

src/migrations/20250109150818-unique-connections-table.js
```diff
@@ -0,0 +1,15 @@
+exports.up = function(db, cb) {
+    db.runSql(`
+    CREATE TABLE IF NOT EXISTS unique_connections
+    (
+        id VARCHAR(255) NOT NULL,
+        updated_at TIMESTAMP DEFAULT now(),
+        hll BYTEA NOT NULL,
```
kwasniew, line 7: hyperlog is a buffer that we translate to BYTEA type in postgres

src/lib/features/unique-connection/unique-connection-service.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    uniqueConnectionService1.count('connection1');
+    uniqueConnectionService1.count('connection2');
+    await uniqueConnectionService1.sync();
+
+    uniqueConnectionService2.count('connection1');
+    uniqueConnectionService2.count('connection3');
+    await uniqueConnectionService2.sync();
+
+    const stats1 = await uniqueConnectionService1.getStats();
```
thomasheartman, line 107: is the get stats method implemented to make it easier to test? or is it also used for reporting? I suspect I'll find out later.

src/lib/features/unique-connection/unique-connection-service.ts
```diff
@@ -0,0 +1,98 @@
+import type { IUnleashConfig } from '../../types/option';
+import type { IFlagResolver, IUnleashStores } from '../../types';
+import type { Logger } from '../../logger';
+import type { IUniqueConnectionStore } from './unique-connection-store-type';
+import HyperLogLog from 'hyperloglog-lite';
+import type EventEmitter from 'events';
+import { SDK_CONNECTION_ID_RECEIVED } from '../../metric-events';
+
+const n = 12;
```
thomasheartman, line 9: I don't really mind, but I think someone's gonna get real mad if we just use `n` as a variable name. How about `numBitIndices` or `bitIndexCount`? (or do them in SCREAMING_SNAKE_CASE if you want)?

src/lib/features/unique-connection/unique-connection-service.ts
```diff
@@ -0,0 +1,98 @@
...
+    listen() {
+        this.eventBus.on(SDK_CONNECTION_ID_RECEIVED, this.count.bind(this));
+    }
+
+    count(connectionId: string) {
+        if (!this.flagResolver.isEnabled('uniqueSdkTracking')) return;
+        this.hll.add(HyperLogLog.hash(connectionId));
+    }
+
+    async getStats() {
```
thomasheartman, line 46: So, interestingly: this only reports what's in the db merged with what it's local value is. It does *not* tell you what's in the db, and it does not sync between the two, so they can diverge. 

Probably not important and wont make a big difference, but good to keep in mind, I think.

src/lib/features/unique-connection/unique-connection-service.ts
```diff
@@ -0,0 +1,98 @@
...
+            previousHll.merge({ n, buckets: previous.hll });
+        }
+        const currentHll = HyperLogLog(n);
+        if (current) {
+            currentHll.merge({ n, buckets: current.hll });
+        }
+        return { previous: previousHll.count(), current: currentHll.count() };
+    }
+
+    async sync(clock = () => new Date()): Promise<void> {
```
thomasheartman, line 62: Why use a function instead of a value here? Not that I mind; I'm just curious whether there's a specific reason?

src/lib/features/unique-connection/unique-connection-store-type.ts
```diff
@@ -0,0 +1,14 @@
+export type UniqueConnections = {
+    hll: Buffer;
```
thomasheartman, line 2: We should probably rename this `buckets` or `hllBuckets` for clarity.

src/lib/features/unique-connection/unique-connection-store.ts
```diff
@@ -0,0 +1,34 @@
...
+export class UniqueConnectionStore implements IUniqueConnectionStore {
+    private db: Db;
+
+    constructor(db: Db) {
+        this.db = db;
+    }
+
+    async insert(uniqueConnections: UniqueConnections): Promise<void> {
+        await this.db<UniqueConnections>('unique_connections')
+            .insert({ id: uniqueConnections.id, hll: uniqueConnections.hll })
```
thomasheartman, line 14: The db migration was already merged, right? So the column is `hll`?

src/lib/middleware/response-time-metrics.test.ts
```diff
@@ -66,6 +66,7 @@ describe('responseTimeMetrics new behavior', () => {
             },
             method: 'GET',
             path: 'should-not-be-used',
+            headers: {},
```
thomasheartman, line 69: this seems weird. why do you need to add headers to all of these?

... because without it, trying to access `headers[index]` yields an error, huh?

thomasheartman (approved):
Nice one 🔥  Thanks for finishing this up. Got a few small questions and comments, but nothing that should stop this from going in.

src/lib/features/unique-connection/unique-connection-service.test.ts
kwasniew: I was thinking about using it in metrics if we go for the explicit metrics polling for changes mechanism instead of event publishing

src/lib/middleware/response-time-metrics.test.ts
kwasniew: yes, it was to pass the tests

src/lib/features/unique-connection/unique-connection-service.ts
kwasniew: I wanted every instance to report the same values so that'w why we're reading from a DB

PR #9077, opened 2025-01-10
fix: etag support RFC 9110 standard

sjaanus:
ETag must be in double quotes based on standard. Fixing it.

```
If-None-Match: "<etag_value>"


FredrikOseberg (approved):
LGTM

PR #9081, opened 2025-01-10
chore: prioritize milestone strategies in sorting

nunogois:
https://linear.app/unleash/issue/2-3100/sorting-issues-with-milestone-strategies-and-regular-strategies

This PR ensures that milestone strategies are always prioritized and evaluated before regular strategies. It ensures that the order displayed in the UI matches the internal evaluation order.

### DIFF
--- a/src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
+++ b/src/lib/features/client-feature-toggles/client-feature-toggle-store.ts
@@ -22,6 +22,7 @@ import type EventEmitter from 'events';
 import FeatureToggleStore from '../feature-toggle/feature-toggle-store';
 import type { Db } from '../../db/db';
 import Raw = Knex.Raw;
+import { sortStrategies } from '../../util/sortStrategies';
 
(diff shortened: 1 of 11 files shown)


daveleek (approved):
Nice! The sorting looks ripe for a future extracting into a method tho haha 😅

nunogois:
> I think this looks good, but we also need to look at the event and how it looks like. Not a blocker for this PR though.

Which event are you talking about?

nunogois:
> Nice! The sorting looks ripe for a future extracting into a method tho haha 😅

Addressed in https://github.com/Unleash/unleash/pull/9081/commits/3f18024723f03902ae868d6f7a836821fc62cc7d — Let me know if this is better!

gastonfournier:
> > I think this looks good, but we also need to look at the event and how it looks like. Not a blocker for this PR though.
> 
> Which event are you talking about?

Event into the event log, because some of these changes will be reflected in the event log of modifying a feature environment.

PR #9090, opened 2025-01-13
feat: Remove create admin token from API

gastonfournier:
Admin tokens have been [deprecated in previous versions](https://docs.getunleash.io/reference/api-tokens-and-client-keys#admin-tokens) encouraging the usage of personal access tokens for better traceability of changes within Unleash.

This removes the ability of creating them from the API


nunogois (approved):
LGTM

PR #9104, opened 2025-01-16
chore: fix release template strategy padding and spacing

daveleek:
Fixes the weird padding issues and the low position of add/cancel buttons on the add/edit strategy dialog for release templates

Before:
image

After:
image


gastonfournier (approved):
LG!

PR #9109, opened 2025-01-17
feat: max parallel sessions config

kwasniew:
## About the changes

Removing deleteStaleUserSessions flag and introducing MAX_PARALLEL_SESSIONS config with env var.

Defaults:
* 100 for enterprise (considered undefined but a very high number leads to simpler code without special cases)
* 5 for pro and OSS

Self-hosted enterprise and OSS can override the env var.

What's the main use case for this env var? 1 parallel session enforcement for security purposes

### DIFF
--- a/src/lib/create-config.ts
+++ b/src/lib/create-config.ts
@@ -266,7 +266,7 @@ const defaultDbOptions: WithOptional<IDBOption, 'user' | 'password' | 'host'> =
         applicationName: process.env.DATABASE_APPLICATION_NAME || 'unleash',
     };
 
-const defaultSessionOption: ISessionOption = {
+const defaultSessionOption = (isEnterprise: boolean): ISessionOption => ({
     ttlHours: parseEnvVarNumber(process.env.SESSION_TTL_HOURS, 48),
     clearSiteDataOnLogout: parseEnvVarBoolean(
         process.env.SESSION_CLEAR_SITE_DATA_ON_LOGOUT,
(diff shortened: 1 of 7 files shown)


src/lib/create-config.ts
```diff
@@ -638,6 +642,12 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
         ui.environment,
         isTest,
     );
+
+    const session: ISessionOption = mergeAll([
```
kwasniew, line 646: moved so that isEnterprise is available

website/docs/using-unleash/deploy/configuring-unleash.mdx
```diff
@@ -201,6 +201,7 @@ unleash.start(unleashOptions);
       instructing the browser to clear all cookies on the same domain Unleash is running on. If disabled unleash will
       only destroy and clear the session cookie. Defaults to _true_. `SESSION_CLEAR_SITE_DATA_ON_LOGOUT`
     - _cookieName_ - Name of the cookies used to hold the session id. Defaults to 'unleash-session'.
+    - _maxParallelSessions_ - The maximum number of parallel user sessions with password based login. `MAX_PARALLEL_SESSIONS`
```
kwasniew, line 204: documenting the new option

melindafekete: thanks! looks good.

ivarconr (approved):

PR #9110, opened 2025-01-17
chore: fix traffic data timezone visualization issue with getUTCDate

daveleek:
Dates are hard. For a lot of timezones new Date(Date.parse('2025-01-01T00:00:00.000Z')).getDate() could be returned as 31 and not 1

By using getUTCDate instead we now adjust accordingly and return the right day number.

### DIFF
--- a/frontend/src/hooks/useTrafficData.ts
+++ b/frontend/src/hooks/useTrafficData.ts
@@ -132,7 +132,7 @@ const toChartData = (
 
             for (const dayKey in item.days) {
                 const day = item.days[dayKey];
-                const dayNum = new Date(Date.parse(day.day)).getDate();
+                const dayNum = new Date(Date.parse(day.day)).getUTCDate();
                 daysRec[`d${dayNum}`] = day.trafficTypes[0].count;
             }
             const epInfo = endpointsInfo[item.apiPath];


frontend/src/hooks/useTrafficData.ts
```diff
@@ -132,7 +132,7 @@ const toChartData = (
 
             for (const dayKey in item.days) {
                 const day = item.days[dayKey];
-                const dayNum = new Date(Date.parse(day.day)).getDate();
+                const dayNum = new Date(Date.parse(day.day)).getUTCDate();
```
kwasniew, line 135: I'd add a test for the function generating those dates that fails with previous code and succeeds with the new one

kwasniew (approved):

frontend/src/hooks/useTrafficData.ts
gastonfournier: He showed me the test but it needs to run vite with a different timezone

daveleek: Thanks, yep fully onboard with that @kwasniew - but I'm yet to find a good way to have a committable repro test for this :( Looking into spoofing timezones and such, but doesn't bite.

kwasniew: may not be worth the effort. ideally before the test we should change the timezone and reset to original value afterward.

kwasniew: another option would be to avoid overriding the timezone in the tested method explicitly

kwasniew:
```ts
beforeEach(() => {
  // Set the timezone to the desired value
  process.env.TZ = 'America/New_York';
});

afterEach(() => {
  // Reset the timezone to the original value
  process.env.TZ = originalTimeZone;
});
```

daveleek: Trying that but can't get it to "bite" unfortunately

PR #9113, opened 2025-01-17
chore(1-3267): use the user's locale settings for formatting the tooltip date

thomasheartman:
This PR updates the tooltip date display in the traffic usage chart to use the user's chosen locale settings, falling back to en-US if the settings are unavailable or otherwise unset.

So, for instance, if I have set my locale to "ja-JP", I'd get this instead of the en US format:
image


daveleek (approved):
Nice one! Thanks!

PR #9122, opened 2025-01-20
feat(1-3262): initial impl of new month/range picker

thomasheartman:
This PR implements a first version of the new month/range picker for the data usage graphs. It's minimally hooked up to the existing functionality to not take anything away. 

This primary purpose of this PR is to get the design and interaction out on sandbox so that UX can have a look and we can make adjustments.

As such, there are a few things in the code that we'll want to clean up before removing the flag later:
- for faster iteration, I've used a lot of CSS nesting and element selectors. this isn't usually how we do it here, so we'll probably want to extract into styled components later
- there is a temporary override of the value in the period selector so that you can select ranges. It won't affect the chart state, but it affects the selector state. Again, this lets you see how it acts and works.
- I've added a `NewHeader` component because the existing setup smushed the selector (it's a MUI grid setup, which isn't very flexible). I don't know what we want to do with this in the end, but the existing chart *does* have some problems when you resize your window, at least (although this is likely due to the chart, and can be solved in the same way that we did for the personal dashboards).

image


frontend/src/component/admin/network/NetworkTrafficUsage/PeriodSelector.tsx
```diff
@@ -0,0 +1,230 @@
...
+            currentDate.getFullYear(),
+            currentDate.getMonth() - subtractMonthCount,
+            1,
+        );
+        selectablePeriods.push(
+            toSelectablePeriod(date, undefined, date > new Date('2024-03-31')),
+        );
+    }
+    return selectablePeriods;
+};
```
thomasheartman, line 59: These functions are mostly copied from `frontend/src/hooks/useTrafficData.ts`, but slightly modified. We'll probably want to reconsider this as the API is built out. I don't want to touch them more than I need to right now.

frontend/src/hooks/useTrafficData.ts
```diff
@@ -48,8 +48,7 @@ const calculateTrafficDataCost = (
     return unitCount * trafficUnitCost;
 };
 
-const padMonth = (month: number): string =>
-    month < 10 ? `0${month}` : `${month}`;
+const padMonth = (month: number): string => month.toString().padStart(2, '0');
```
thomasheartman, line 51: Javascript can do this for us with a builtin instead of us checking the value

Tymek (approved):

PR #9127, opened 2025-01-21
feat(1-3260): store support for data traffic from a range

thomasheartman:
Add support for querying the traffic data usage store for the aggregated data for an arbitrary number of months back.

Adds a new `getTrafficDataForMonthRange(monthsBack: number)` method to the store that aggregates data on a monthly basis by status code and traffic group. Returns a new type with month data instead of day data.

### DIFF
--- a/src/lib/features/traffic-data-usage/traffic-data-usage-store-type.ts
+++ b/src/lib/features/traffic-data-usage/traffic-data-usage-store-type.ts
@@ -7,6 +7,13 @@ export type IStatTrafficUsage = {
     count: number;
 };
 
+export type IStatMonthlyTrafficUsage = {
+    month: Date;
+    trafficGroup: string;
+    statusCodeSeries: number;
+    count: number;
+};
+
 export interface IStatTrafficUsageKey {
     day: Date;
     trafficGroup: string;
@@ -17,4 +24,7 @@ export interface ITrafficDataUsageStore
     extends Store<IStatTrafficUsage, IStatTrafficUsageKey> {
     upsert(trafficDataUsage: IStatTrafficUsage): Promise<void>;
     getTrafficDataUsageForPeriod(period: string): Promise<IStatTrafficUsage[]>;
+    getTrafficDataForMonthRange(
+        monthsBack: number,
+    ): Promise<IStatMonthlyTrafficUsage[]>;
 }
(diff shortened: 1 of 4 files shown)


src/lib/features/traffic-data-usage/traffic-data-usage-store.test.ts
```diff
@@ -20,6 +21,10 @@ afterAll(async () => {
     await db.destroy();
 });
 
+beforeEach(async () => {
+    await trafficDataUsageStore.deleteAll();
+});
```
thomasheartman, line 26: We were calling this in all the tests, so it's probably better to extract it so that we don't need to remember.

src/lib/features/traffic-data-usage/traffic-data-usage-store.ts
```diff
@@ -97,4 +98,37 @@ export class TrafficDataUsageStore implements ITrafficDataUsageStore {
...
+                { column: 'traffic_group', order: 'asc' },
+            ]);
+        return rows.map(
+            ({ traffic_group, status_code_series, month, count }) => ({
+                trafficGroup: traffic_group,
+                statusCodeSeries: status_code_series,
+                month,
+                count: Number.parseInt(count),
+            }),
+        );
```
thomasheartman, line 132: Here's the actual implementation.

src/lib/features/traffic-data-usage/traffic-data-usage-store.test.ts
```diff
@@ -195,3 +196,57 @@ test('can query for data from specific periods', async () => {
...
+            };
+            await trafficDataUsageStore.upsert(dataB);
+        }
+        expectedValues.push({
+            groupA: monthAggregateA,
+            groupB: monthAggregateB,
+        });
+    }
+
+    for (const monthsBack of [3, 6, 12]) {
```
Tymek, line 236: this could be `test.each`

src/lib/features/traffic-data-usage/traffic-data-usage-store-type.ts
```diff
@@ -7,6 +7,13 @@ export type IStatTrafficUsage = {
     count: number;
 };
 
+export type IStatMonthlyTrafficUsage = {
+    month: string;
```
Tymek, line 11: Yup, month makes more sense here then `day`, for example for `differenceInCalendarMonths(now, entry.day)` 👍🏻

Tymek (approved):

PR #9133, opened 2025-01-22
feat: delta rework

sjaanus:
We are changing how the Delta API works, as discussed:

1. We have removed the `updated` and `removed` arrays and now keep everything in the `events` array.
2. We decided to keep the hydration cache separate from the events array internally. Since the hydration cache has a special structure and may contain not just one feature but potentially 1,000 features, it behaved differently, requiring a lot of special logic to handle it.
3. Implemented `nameprefix` filtering, which we were missing before.

Things still to implement:

1.  Segment hydration and updates to it.


src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -191,22 +192,34 @@ export default class FeatureController extends Controller {
...
+                                features.length
+                            }, new count ${hydrationEvent.features.length}, query ${JSON.stringify(query)},
                         diff ${JSON.stringify(
                             diff(sortedToggles, sortedNewToggles),
                         )}`,
+                        );
+                    }
+                } else {
+                    this.logger.warn(
+                        `Delta diff should have only hydration event, query ${JSON.stringify(query)}`,
```
FredrikOseberg, line 219: Is this the correct implementation of the diff checker? If we only compare on hydration event, how do we know that the deltas are producing the correct result?

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta-types.ts
```diff
@@ -0,0 +1,63 @@
+import type { ClientFeatureSchema } from '../../../openapi';
```
FredrikOseberg, line 1: Love this file. Very clean.

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
```diff
@@ -18,92 +18,75 @@ import { CLIENT_DELTA_MEMORY } from '../../../metric-events';
...
-        revisionId: last.revisionId,
-        updated: Array.from(updatedMap.values()),
-        removed: Array.from(removedMap.values()),
+    const allProjects = projects.includes('*');
+    const startsWithPrefix = (revision: DeltaEvent) => {
+        return (
+            (isDeltaFeatureUpdatedEvent(revision) &&
+                revision.feature.name.startsWith(namePrefix)) ||
+            (isDeltaFeatureRemovedEvent(revision) &&
+                revision.featureName.startsWith(namePrefix))
```
FredrikOseberg, line 46: Why are these different?

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
```diff
@@ -18,92 +18,75 @@ import { CLIENT_DELTA_MEMORY } from '../../../metric-events';
...
+            (isDeltaFeatureRemovedEvent(revision) &&
+                projects.includes(revision.project))
+        );
+    };
 
-    return { ...revision, updated, removed };
+    return targetedEvents.filter((revision) => {
+        return (
+            startsWithPrefix(revision) && (allProjects || isInProject(revision))
+        );
```
FredrikOseberg, line 62: We always filter with namePrefix? What if nameprefix is empty?

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
```diff
@@ -171,26 +155,48 @@ export class ClientFeatureToggleDelta {
         if (!hasSegments) {
             await this.updateSegments();
         }
-
         if (requiredRevisionId >= this.currentRevisionId) {
             return undefined;
         }
+        if (requiredRevisionId === 0) {
```
FredrikOseberg, line 161: Might be missing some context, but if this is the revisionId coming from the SDK, couldn't it possibly be undefined? Why a strict check on 0?

FredrikOseberg (changes requested):
Some feedback we'll need to address before merging. I might just be missing context but let's discuss it before we go ahead.

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
sjaanus: Well, every time you add new events to cache, they will instantly applied to hydration event. 
We only take features from database at first, but after that we will start manipulation the hydration event every second.

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
sjaanus: Feature updated has complex object of feature, with all the description, strategies, constraints. Removed has just string value.

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
sjaanus: https://github.com/Unleash/unleash/pull/9133/files#diff-117921481602afc7d7b53fef3f43e7c22c8e08d938f28512444b1021bdcb86c5R145

Normalizes it.

src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
sjaanus: https://github.com/Unleash/unleash/pull/9133/files#diff-117921481602afc7d7b53fef3f43e7c22c8e08d938f28512444b1021bdcb86c5R145 normalizes it.

FredrikOseberg (approved):
LGTM

PR #9136, opened 2025-01-22
Update lifcyecle open API schema with new enums

melindafekete:
Updates the enum names for the lifecycle stages in OpenAPI spec


melindafekete:
Enum values are not changing in the API yet

PR #9139, opened 2025-01-23
feat: show a dialog when adding a release plan to a change request enabled feature environment

daveleek:
Displays a dialogue for user to confirm adding changes to draft when user adds release-plan to change request enabled feature environment.

Currently does nothing more than display a success toast when confirmed, pending remaining change request implementation

image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
@@ -8,9 +8,7 @@ import useToast from 'hooks/useToast';
 import { formatUnknownError } from 'utils/formatUnknownError';
 import { useReleasePlans } from 'hooks/api/getters/useReleasePlans/useReleasePlans';
 import { useChangeRequestsEnabled } from 'hooks/useChangeRequestsEnabled';
-import { useNavigate } from 'react-router-dom';
 import { useUiFlag } from 'hooks/useUiFlag';
-import { formatReleasePlanChangeRequestPath } from 'component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlanAddChangeRequestDialog';
 
 const StyledIcon = styled('div')(({ theme }) => ({
     width: theme.spacing(4),
@@ -55,6 +53,7 @@ interface IFeatureReleasePlanCardProps {
     featureId: string;
     environmentId: string;
     releasePlanTemplate: IReleasePlanTemplate;
+    setAddingTemplateId: (templateId: string) => void;
 }
 
 export const FeatureReleasePlanCard = ({
@@ -62,9 +61,9 @@ export const FeatureReleasePlanCard = ({
     featureId,
     environmentId,
     releasePlanTemplate,
+    setAddingTemplateId,
 }: IFeatureReleasePlanCardProps) => {
     const Icon = getFeatureStrategyIcon('releasePlanTemplate');
-    const navigate = useNavigate();
     const { trackEvent } = usePlausibleTracker();
     const { refetch } = useReleasePlans(projectId, featureId, environmentId);
     const { addReleasePlanToFeature } = useReleasePlansApi();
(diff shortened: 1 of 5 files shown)


nunogois (approved):
I'm confused by this. I think this is the first dialog based on a route on our whole codebase. But if it helps you keep moving forward, it should be okay for now.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -51,33 +53,46 @@ interface IFeatureReleasePlanCardProps {
...
     const { refetch } = useReleasePlans(projectId, featureId, environmentId);
     const { addReleasePlanToFeature } = useReleasePlansApi();
     const { setToastApiError, setToastData } = useToast();
+    const { isChangeRequestConfigured } = useChangeRequestsEnabled(projectId);
+    const releasePlanChangeRequestsEnabled = useUiFlag(
+        'releasePlanChangeRequests',
+    );
 
-    const addReleasePlan = async () => {
+    const addReleasePlan = async (e: React.MouseEvent) => {
```
nunogois, line 76:
```suggestion
    const addReleasePlan = async () => {
```

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlanAddChangeRequestDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+            type: 'success',
+            text: 'Added to draft',
+        });
+        onClosing();
+    };
+
+    return (
+        <Dialogue
+            title='Request changes'
+            open={true}
```
nunogois, line 38: Maybe it would be more consistent to manage the open state from the caller, like we do for other dialogs?

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlanAddChangeRequestDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const ReleasePlanAddChangeRequestDialog = ({
+    featureId,
+    environmentId,
+    releaseTemplateId,
+    onClosing,
+}: {
+    featureId: string;
+    environmentId: string;
+    releaseTemplateId: string | undefined;
```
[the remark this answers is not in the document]
nunogois: Would probably be more consistent to create an `IReleasePlanAddChangeRequestDialogProps` interface.

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlanAddChangeRequestDialog.tsx
nunogois: Even if it's something like open={Boolean(changeRequestReleaseTemplateId)}, since that seems to be your logic here.

daveleek: Yeah thanks, I agree and tried that, but there's a call to get template that needs a defined string. So I ended up getting a runtime crash due to number of hooks being rendered. I might be able to get out of that by changing the type of what goes into the templateId state to hold all I need instead

nunogois: [This](https://github.com/Unleash/unleash/pull/9139/files#diff-5a7b0924651d248db4f47c573a7e8c907f71d6c7f4de84db7193c9fe569619f1R24) one? You already have a guard above so you shouldn't have that issue anymore.

What I meant is that `open` would be a prop in this component (`ReleasePlanAddChangeRequestDialog`, so here you would simply have `open={open}`. But maybe the caller decides that `open={Boolean(changeRequestReleaseTemplateId)}` if you don't want to manage a separate open state and instead just rely on that ID being set.

nunogois: >I might be able to get out of that by changing the type of what goes into the templateId state to hold all I need instead

I like this one as well, you can simply pass the template itself if you have it instead.

PR #9149, opened 2025-01-24
feat(1-3267): use new API for chart creation

thomasheartman:
Adds support for the new /traffic-search API behind a flag. When active, you'll be able to select month ranges as well as specific single months.

Largely copies the existing network traffic component, and adds some minor tweaks to make it work with the new data.

This is quite rough, but it gives us a base to build on for later. There's still things that we need to solve for in following PRs.


daveleek (approved):
I see the description is a wip, but we discussed the PR and went through together and it looks good!

PR #9150, opened 2025-01-24
chore(1-3293): add new query for daily data that uses  date ranges

thomasheartman:
Add new methods to the store behind data usage metrics that accept date ranges instead of a single month. The old data collection methods re-route to the new ones instead, so the new methods are tested implicitly.

Also deprecates the new endpoint that's not in use anywhere except in an unused service method in Enterprise yet.

## Discussion point:

Should the new methods accept `from` and `to` as dates? Or as strings? If they're strings, there's a chance the parsing can go wrong, so dates feel safer to me.

There's *one* instance of a place where call the old daily data method (in `src/lib/features/instance-stats/instance-stats-service.ts`). Should we update that to use the new method? Should we keep the old one and have both?

### DIFF
--- a/src/lib/features/traffic-data-usage/fake-traffic-data-usage-store.ts
+++ b/src/lib/features/traffic-data-usage/fake-traffic-data-usage-store.ts
@@ -6,9 +6,11 @@ import type {
 import type { ITrafficDataUsageStore } from '../../types';
 import {
     differenceInCalendarMonths,
+    endOfDay,
     format,
     isSameMonth,
     parse,
+    startOfDay,
 } from 'date-fns';
 
 export class FakeTrafficDataUsageStore implements ITrafficDataUsageStore {
@@ -90,22 +92,26 @@ export class FakeTrafficDataUsageStore implements ITrafficDataUsageStore {
         return Object.values(data);
     }
 
-    async getDailyTrafficDataForPeriod(
+    async getDailyTrafficUsageDataForPeriod(
         from: Date,
         to: Date,
     ): Promise<IStatTrafficUsage[]> {
         return this.trafficData.filter(
-            (data) => data.day >= from && data.day <= to,
+            (data) => data.day >= startOfDay(from) && data.day <= endOfDay(to),
         );
     }
 
-    async getMonthlyTrafficDataForPeriod(
+    async getMonthlyTrafficUsageDataForPeriod(
         from: Date,
         to: Date,
     ): Promise<IStatMonthlyTrafficUsage[]> {
         const data: { [key: string]: IStatMonthlyTrafficUsage } =
             this.trafficData
-                .filter((entry) => entry.day >= from && entry.day <= to)
+                .filter(
+                    (data) =>
+                        data.day >= startOfDay(from) &&
+                        data.day <= endOfDay(to),
+                )
                 .reduce((acc, entry) => {
                     const month = format(entry.day, 'yyyy-MM');
                     const key = `${month}-${entry.trafficGroup}-${entry.statusCodeSeries}`;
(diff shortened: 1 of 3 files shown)


src/lib/features/traffic-data-usage/fake-traffic-data-usage-store.ts
```diff
@@ -89,4 +89,41 @@ export class FakeTrafficDataUsageStore implements ITrafficDataUsageStore {
...
+    async getDailyTrafficDataForPeriod(
+        from: Date,
+        to: Date,
+    ): Promise<IStatTrafficUsage[]> {
+        return this.trafficData.filter(
+            (data) => data.day >= from && data.day <= to,
+        );
+    }
+
+    async getMonthlyTrafficDataForPeriod(
```
thomasheartman, line 102: "traffic data"? "traffic usage data"? got a preference?

src/lib/features/traffic-data-usage/fake-traffic-data-usage-store.ts
```diff
@@ -89,4 +89,41 @@ export class FakeTrafficDataUsageStore implements ITrafficDataUsageStore {
 
         return Object.values(data);
     }
+
+    async getDailyTrafficDataForPeriod(
+        from: Date,
+        to: Date,
+    ): Promise<IStatTrafficUsage[]> {
+        return this.trafficData.filter(
+            (data) => data.day >= from && data.day <= to,
```
daveleek, line 98: from/to, will they contain time? day field in DB is day, so might potentially catch one less day than intentional.

thomasheartman: This is just the fake store, so I'm not too worried about this one, but yeah, we should add a "startOfday/endOfDay here too. I'll fix.

daveleek: ah good point, dunno why I missed that 😅

src/lib/features/traffic-data-usage/traffic-data-usage-store.ts
```diff
@@ -89,18 +96,20 @@ export class TrafficDataUsageStore implements ITrafficDataUsageStore {
...
+    async getDailyTrafficDataForPeriod(
+        from: Date,
+        to: Date,
     ): Promise<IStatTrafficUsage[]> {
-        const rows = await this.db<IStatTrafficUsage>(TABLE).whereRaw(
-            `to_char(day, 'YYYY-MM') = ?`,
-            [period],
-        );
+        const rows = await this.db<IStatTrafficUsage>(TABLE)
+            .where('day', '>=', startOfDay(from))
```
daveleek, line 104: 👍

src/lib/features/traffic-data-usage/fake-traffic-data-usage-store.ts
daveleek: traffic usage * is the most accurate term, but I might have opted for keeping method names shorter some places

thomasheartman: Cool, I'll put the extra usage in there , then

daveleek (approved):
Looks good!

> There's one instance of a place where call the old daily data method (in src/lib/features/instance-stats/instance-stats-service.ts). Should we update that to use the new method? Should we keep the old one and have both?

Wasn't aware of this - the method looks to be aggregating for the current month however, so could probably change it to just fetch using the new method and skip the aggregating of data. Should also be fine to just leave it

src/lib/features/traffic-data-usage/fake-traffic-data-usage-store.ts
thomasheartman: fixed in 9a5863b7b04015c6ebbb25083665e72f97a4c455

src/lib/features/traffic-data-usage/fake-traffic-data-usage-store.ts
thomasheartman: addressed in 9a5863b7b04015c6ebbb25083665e72f97a4c455

thomasheartman:
I ... think I wanna leave the instance stats for now. Keep the changes as isolated as possible.

PR #9154, opened 2025-01-27
chore: remove feature toggle kill switch for admin token UI

gastonfournier:
## About the changes
This was released with 6.6 and now this flag is no longer needed. The default value is already set to true, so it's the same as removing it


daveleek (approved):
LGTM!

PR #9155, opened 2025-01-27
feat: configurable global font size

FredrikOseberg:
This PR sets up the application to accept a value from a variant we control to set the font size of the application on a global level. If it fails, the value falls back to the previously set CSS value.


thomasheartman (approved):
As discussed in slack, I suggest following Tymek's advice and trying it out with changing `theme.typography.fontSize` instead, but otherwis it looks good.

Giving you an approval so you can merge once you've figured it out.

PR #9156, opened 2025-01-27
fix:  correct name order -> traffic usage data -> traffic data usage

thomasheartman:
Correcting a slight oversight from when we created the new methods. This brings the name in line with the other store methods.


daveleek (approved):
LGTM!

PR #9157, opened 2025-01-27
docs: fix background img dark mode opacity

nunogois:
https://linear.app/unleash/issue/2-3200/update-docs-bottom-left-background-image-in-dark-mode

The background mountain texture was a bit too striking visually. We should set an opacity of `0.06` in dark mode, the same as in Unleash dark mode.

### Before

image

### After

image


melindafekete (approved):
Thank you!

PR #9160, opened 2025-01-28
task: make setting service handle conflict on insert

chriswk:
Currently, in enterprise we're struggling with setting service and transactionality; all our verfications says that the setting key is not present, so setting-store happily tries to insert a new row with a new PK. However, somehow at the same time, the key already exists. This commit adds conflict handling to the insertNewRow.


gastonfournier (approved):
LGTM!

PR #9161, opened 2025-01-28
fix: tabs not growing to fill all available space

FredrikOseberg:
Fix an issue where tabs would fill out all available space, now it's left aligned: 

Skjermbilde 2025-01-28 kl  13 40 56

Previously:

Skjermbilde 2025-01-28 kl  13 41 32


thomasheartman (approved):
Huh. Didn't realize that they **weren't** supposed to take the whole space. Yeah, this looks fine.

PR #9166, opened 2025-01-29
fix: dynamic highlight width

thomasheartman:
This makes the width of the highlight bars in the network dynamic and based on the number of labels included in the chart.

Since the number of labels should always correspond to the number of data points, this seems like a sensible approach.

With this, the label width will now be calculated on the fly, so even if you resize the window or change the number of labels, the highlighting will still work as expected.

Daily view:
image

Monthly aggregate:
image

The labels are now a little narrower on the daily graphs, but it avoids them being super wide on the monthly graphs

### DIFF
--- a/frontend/src/component/common/Chart/customHighlightPlugin.ts
+++ b/frontend/src/component/common/Chart/customHighlightPlugin.ts
@@ -1,37 +1,35 @@
 import type { Chart } from 'chart.js';
 
-export const customHighlightPlugin = (bottomOverflow = 34) => {
-    return {
-        id: 'customLine',
-        beforeDraw: (chart: Chart) => {
-            if (chart.tooltip?.opacity && chart.tooltip.x) {
-                const x = chart.tooltip.caretX;
-                const yAxis = chart.scales.y;
-                const ctx = chart.ctx;
-                ctx.save();
-                const gradient = ctx.createLinearGradient(
-                    x,
-                    yAxis.top,
-                    x,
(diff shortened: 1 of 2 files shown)


frontend/src/component/common/Chart/customHighlightPlugin.ts
```diff
@@ -17,10 +17,14 @@ export const customHighlightPlugin = (width = 46, bottomOverflow = 34) => ({
             gradient.addColorStop(0, 'rgba(129, 122, 254, 0)');
             gradient.addColorStop(1, 'rgba(129, 122, 254, 0.12)');
             ctx.fillStyle = gradient;
+
+            const barWidth =
+                (chart.width / (chart.data.labels?.length ?? 1)) *
+                (chart.options.datasets?.bar?.categoryPercentage ?? 1);
```
daveleek, line 23: Cool, is categoryPercentage set anywhere or is that a separate PR?

daveleek (approved):
Nice!

frontend/src/component/common/Chart/customHighlightPlugin.ts
thomasheartman: No, categoryPercentage is set by default to `0.8`. This is a little wider than the actual bar, which appears to be categoryPercentage * barPercentage (defaults to `0.9`). So this should make it so that the bar is roughly 90% of the highlight width. Ish ...

PR #9169, opened 2025-01-29
feat(1-3281): wraps the new datepicker in a dropdown

thomasheartman:
Wraps the datepicker in a popover, making it function largely the same as a dropdown list.

The dropdown displays one of:
- "current month" if you've selected the current month
- "<month> <year>" (e.g. "December 2024") if you've selected a month that isn't the current month
- "Last n months" (e.g. "Last 3 months") if you have selected a range

Additionally, the range selections have been updated to span the whole row, aligning with the look of generic dropdown lists.

image

Like with the rest of this file (`PeriodSelector`), the code is rough and not according to Unleash standards. However, I'm prioritizing fast changes so UX can have a look before I clean up the code to switch to using styled components etc later. It's still behind a flag, so I'm not very worried about it.


daveleek (approved):
Nice!

PR #9176, opened 2025-01-30
chore: drag to reorder release plan template milestones

nunogois:
https://linear.app/unleash/issue/2-2821/drag-to-reorder-template-milestones

This PR introduces reordering release plan template milestones by dragging and dropping them.

Was a bit undecided on the approach, but it seems like using an old `useDragItem` hook we have is pretty elegant and behaves as expected.

I suggest reviewers try it out themselves.

image


daveleek (approved):
Nice, looks good and seems to work well when trying it out as well!

PR #9180, opened 2025-01-31
chore(1-3316): update request info boxes to new design

thomasheartman:
Updates the existing number of requests and overage info boxes to the new design.

The existing versions of the boxes had some issues on narrower screens, so it was easier to just leave them as is and start from scratch.

The previous boxes on narrow screens:
image

The current ones (from wide to narrower):
Wide
image

Mid (the text should probably ideally wrap at the same time here, but I'm not sure how at the moment)
image

Narrow
image

Extra narrow:
image

There's still some work we **could** do, but we should have UX have a look first. In particular, it's about how the text wraps in certain places etc, but I think it's good enough for now.

I'll come back with tests for the calculations and some refactoring and cleanup in a followup.

### DIFF
--- a/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
+++ b/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
@@ -290,7 +290,6 @@ const NewNetworkTrafficUsage: FC = () => {
                             overageCost > 0
                         }
                         show={
-                            // todo: we the text here should be modified based on the selection.
                             <Alert severity='warning' sx={{ mb: 4 }}>
                                 <BoldText>Heads up!</BoldText> You are currently
                                 consuming more requests than your plan includes
@@ -310,13 +309,6 @@ const NewNetworkTrafficUsage: FC = () => {
                     />
                     <StyledBox>
                         <NewHeader>
-                            {/* <NetworkTrafficUsagePlanSummary
-                                usageTotal={usageTotal}
-                                includedTraffic={includedTraffic}
-                                overageCost={overageCost}
-                                estimatedMonthlyCost={estimatedMonthlyCost}
-                            /> */}
-
                             <TrafficInfoBoxes>
                                 <RequestSummary
                                     period={newPeriod}
--- a/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
+++ b/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
@@ -282,7 +282,13 @@ const NewNetworkTrafficUsage: FC = () => {
             elseShow={
                 <>
                     <ConditionallyRender
-                        condition={includedTraffic > 0 && overageCost > 0}
+                        condition={
+                            (newPeriod.grouping === 'monthly' ||
+                                newPeriod.month ===
+                                    format(new Date(), 'yyyy-MM')) &&
+                            includedTraffic > 0 &&
+                            overageCost > 0
+                        }
                         show={
                             // todo: we the text here should be modified based on the selection.
                             <Alert severity='warning' sx={{ mb: 4 }}>
@@ -290,20 +296,14 @@ const NewNetworkTrafficUsage: FC = () => {
                                 consuming more requests than your plan includes
                                 and will be billed according to our terms.
                                 Please see{' '}
-                                <Link
-                                    component={RouterLink}
-                                    to='https://www.getunleash.io/pricing'
-                                >
+                                <RouterLink to='https://www.getunleash.io/pricing'>
                                     this page
-                                </Link>{' '}
+                                </RouterLink>{' '}
                                 for more information. In order to reduce your
                                 traffic consumption, you may configure an{' '}
-                                <Link
-                                    component={RouterLink}
-                                    to='https://docs.getunleash.io/reference/unleash-edge'
-                                >
+                                <RouterLink to='https://docs.getunleash.io/reference/unleash-edge'>
                                     Unleash Edge instance
-                                </Link>{' '}
+                                </RouterLink>{' '}
                                 in your own datacenter.
                             </Alert>
                         }
--- a/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
+++ b/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
@@ -274,6 +274,15 @@ const NewNetworkTrafficUsage: FC = () => {
         .map((info) => info.label.toLowerCase())
         .toReversed();
     const requestTypes = `${otherLabels.toReversed().join(', ')}, and ${lastLabel}`;
+    const chartLabel =
+        newPeriod.grouping === 'daily'
+            ? `A bar chart showing daily traffic usage for ${new Date(
+                  newPeriod.month,
+              ).toLocaleDateString('en-US', {
+                  month: 'long',
+                  year: 'numeric',
+              })}. Each date shows ${requestTypes} requests.`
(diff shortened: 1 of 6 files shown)


frontend/src/component/common/Badge/Badge.tsx
thomasheartman, line 1: For some reason, the badges are tabbable by default which is confusing, because they aren't interactive. We can't set them to be unreachable by keyboard by default because it might break existing uses, but we can allow the user to override it.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -248,15 +261,22 @@ const NewNetworkTrafficUsage: FC = () => {
         }
     }, [data]);
 
+    // todo: extract this (and also endpoints info)
+    const [lastLabel, ...otherLabels] = Object.values(endpointsInfo)
+        .map((info) => info.label.toLowerCase())
+        .toReversed();
+    const requestTypes = `${otherLabels.toReversed().join(', ')}, and ${lastLabel}`;
```
thomasheartman, line 276: endpoints info is a constant, so we can extract it into its own file. I'll do that in a follow-up.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -248,49 +269,75 @@ const NewNetworkTrafficUsage: FC = () => {
...
             elseShow={
                 <>
                     <ConditionallyRender
-                        condition={includedTraffic > 0 && overageCost > 0}
+                        condition={
+                            (newPeriod.grouping === 'monthly' ||
+                                newPeriod.month ===
+                                    format(new Date(), 'yyyy-MM')) &&
+                            includedTraffic > 0 &&
+                            overageCost > 0
```
thomasheartman, line 290: Telling people they're "currently" using more data when the data we're measuring is from a past month seems misleading. So let's only show this if your chart includes data for the current month.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -248,49 +269,74 @@ const NewNetworkTrafficUsage: FC = () => {
...
+                                                  Object.keys(endpointsInfo),
+                                                  traffic.usage,
+                                              )
+                                    }
+                                    includedTraffic={includedTraffic}
+                                />
+                                {newPeriod.grouping === 'daily' &&
+                                    includedTraffic > 0 &&
+                                    usageTotal - includedTraffic > 0 &&
+                                    estimateTrafficDataCost && (
```
thomasheartman, line 328: 🤮 Yeah, this looks horrible. I'll extract it into a separate variable in the follow-up.

frontend/src/component/admin/network/NetworkTrafficUsage/average-traffic-previous-months.ts
thomasheartman, line 1: tests in a followup

frontend/src/hooks/useTrafficData.ts
```diff
@@ -172,7 +172,9 @@ const toMonthlyChartData = (
     );
 
     const labels = Array.from({ length: numMonths }).map((_, index) =>
-        formatMonth(addMonths(from, index)),
+        index === numMonths - 1
+            ? 'Current month'
+            : formatMonth(addMonths(from, index)),
```
thomasheartman, line 177: we want to mark the current month (that hasn't been completed) as such, so that there's no confusion.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -192,7 +209,11 @@ const NewNetworkTrafficUsage: FC = () => {
                         },
                     );
                 } else {
-                    return new Date(tooltipItems[0].label).toLocaleDateString(
+                    const timestamp = Date.parse(tooltipItems[0].label);
+                    if (Number.isNaN(timestamp)) {
```
daveleek, line 213: Interesting! so essentially .. Date.parse('current') -> NaN ?

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -248,49 +269,74 @@ const NewNetworkTrafficUsage: FC = () => {
...
             condition={isOss()}
             show={<Alert severity='warning'>Not enabled.</Alert>}
             elseShow={
                 <>
                     <ConditionallyRender
-                        condition={includedTraffic > 0 && overageCost > 0}
+                        condition={
+                            (newPeriod.grouping === 'monthly' ||
+                                newPeriod.month ===
+                                    format(new Date(), 'yyyy-MM')) &&
```
daveleek, line 288: There's a `currentPeriod` in `useTrafficDataEstimation`, but maybe that's not available here?

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -300,7 +346,16 @@ const NewNetworkTrafficUsage: FC = () => {
...
-                            aria-label='An instance metrics line chart with two lines: requests per second for admin API and requests per second for client API' // todo: this isn't correct at all!
+                            aria-label={
+                                newPeriod.grouping === 'daily'
+                                    ? `A bar chart showing daily traffic usage for ${new Date(
+                                          newPeriod.month,
+                                      ).toLocaleDateString('en-US', {
+                                          month: 'long',
+                                          year: 'numeric',
+                                      })}. Each date shows ${requestTypes} requests.`
+                                    : `A bar chart showing monthly total traffic usage for the current month and the preceding ${newPeriod.monthsBack} months. Each month shows ${requestTypes} requests.`
```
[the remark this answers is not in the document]
daveleek: variable - sorry

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
thomasheartman: Yup. [According to mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse), it returns:
> A number representing the [timestamp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date) of the given date. If dateString fails to be parsed as a valid date, [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) is returned.

thomasheartman: Of course, we could also check whether the label is equal to "Current", but that'd break if the we changed the label, so I went with this strange check.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
thomasheartman: Ah, good call. Should be actually!

thomasheartman: Mmm ... So, there **is**, but that variable has: the current date and a label 'current month'. Neither of those really help out much here. We'd have to take the value from that variable and format that. It's probably slightly better than instantiating a new date object every time, but only marginally.

I'd rather extract this into a separate function or variable (that might be tied to the selection object). But I'd like to look at that separately too, if you're alright with it.

daveleek (approved):
Nice! Couple of questions/pointing something out but nothing serious

PR #9190, opened 2025-02-03
feat: allow SCIM user deletion

sighphyre:
# Summary 

Adds a way to remove SCIM entities if you're an admin

## Frontend Changes

Groups and users can now be deleted if they originate from SCIM, the usual deletion buttons have just been re-enabled.

The UI for SCIM setup itself has been given two new buttons, one each for nuking all SCIM groups and users. This requires a [companion PR](https://github.com/bricks-software/unleash-enterprise/pull/35) on enterprise to work correctly. I've opted specifically to add these buttons to the SCIM setup interface because this is more a SCIM problem that a group/user problem from a user perspective and because this really shouldn't be part of the standard manage groups/users flow,

## Backend Changes

Group/user deletion code no longer rejects calls if the group/user is a SCIM entity.

A method for deleting all users has been added to the user-service. This is specifically not added to the SCIM code because that code is for handling callers from the outside world wanting to work against the SCIM protocol. 

## Discussion

This has added two new events for bulk deletion. Those events do not do a pre-data/post data check, which means they show up in the event log as just "ScimGroupsDeleted". I'm not convinced this is the correct choice because audit logs should show a full record of what happened. That being said, this is very much an escape hatch for SCIM if something goes wrong and I'm worried about dropping a 70K entity change into the event log, both from a UI sanity perspective and from a melt-the-database perspective

### DIFF
--- a/frontend/src/component/admin/auth/ScimSettings/ScimDeleteUsersDialog.tsx
+++ b/frontend/src/component/admin/auth/ScimSettings/ScimDeleteUsersDialog.tsx
@@ -24,7 +24,7 @@ export const ScimDeleteEntityDialog = ({
         open={open}
         primaryButtonText={`Delete SCIM ${entityType}`}
         secondaryButtonText='Cancel'
-        title={`Really Delete All SCIM ${entityType}?`}
+        title={`Do you really want to delete ALL SCIM ${entityType}?`}
         onClose={closeDialog}
         onClick={removeUser}
     >
--- a/frontend/src/component/admin/auth/ScimSettings/ScimDeleteUsersDialog.tsx
+++ b/frontend/src/component/admin/auth/ScimSettings/ScimDeleteUsersDialog.tsx
@@ -1,0 +1,36 @@
+import { Alert, styled, Typography } from '@mui/material';
+import { Dialogue } from 'component/common/Dialogue/Dialogue';
+
+const StyledAlert = styled(Alert)(({ theme }) => ({
+    marginBottom: theme.spacing(3),
+}));
+
+export type EntityType = 'Users' | 'Groups';
+
+interface IScimDeleteUsersProps {
+    open: boolean;
+    entityType: EntityType;
+    closeDialog: () => void;
+    deleteEntities: () => void;
+}
+
+export const ScimDeleteEntityDialog = ({
+    open,
+    closeDialog,
+    deleteEntities: removeUser,
+    entityType,
+}: IScimDeleteUsersProps) => (
+    <Dialogue
+        open={open}
+        primaryButtonText={`Delete SCIM ${entityType}`}
+        secondaryButtonText='Cancel'
+        title={`Really Delete All SCIM ${entityType}?`}
+        onClose={closeDialog}
+        onClick={removeUser}
+    >
+        <Typography variant='body1'>
+            This will delete all {entityType.toLocaleLowerCase()} created or
+            managed by SCIM.
+        </Typography>
+    </Dialogue>
+);
--- a/frontend/src/component/admin/auth/ScimSettings/ScimSettings.tsx
+++ b/frontend/src/component/admin/auth/ScimSettings/ScimSettings.tsx
@@ -200,7 +200,7 @@ export const ScimSettings = () => {
                     </Grid>
                     <Grid item md={10.5} mb={2}>
                         <StyledTitleDiv>
-                            <strong>Clear SCIM Groups</strong>
+                            <strong>Delete SCIM Groups</strong>
                         </StyledTitleDiv>
                         <p>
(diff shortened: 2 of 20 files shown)


frontend/src/component/admin/auth/ScimSettings/ScimDeleteUsersDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+    open,
+    closeDialog,
+    deleteEntities: removeUser,
+    entityType,
+}: IScimDeleteUsersProps) => (
+    <Dialogue
+        open={open}
+        primaryButtonText={`Delete SCIM ${entityType}`}
+        secondaryButtonText='Cancel'
+        title={`Really Delete All SCIM ${entityType}?`}
```
gastonfournier, line 27:
```suggestion
        title={`Do you really want to delete ALL SCIM ${entityType}?`}
```
Who am I to judge your English :see_no_evil: :rofl:

frontend/src/component/admin/auth/ScimSettings/ScimSettings.tsx
```diff
@@ -138,6 +173,57 @@ export const ScimSettings = () => {
...
+                            onClick={() => {
+                                setDeleteUsersDialog(true);
+                            }}
+                        >
+                            Delete Users
+                        </Button>
+                    </Grid>
+                    <Grid item md={10.5} mb={2}>
+                        <StyledTitleDiv>
+                            <strong>Clear SCIM Groups</strong>
```
gastonfournier, line 203:
```suggestion
                            <strong>Delete SCIM Groups</strong>
```
Maybe for consistency with the previous title about SCIM users?

frontend/src/component/admin/groups/Group/Group.tsx
```diff
@@ -270,11 +270,8 @@ export const Group: VFC = () => {
                                     onClick={() => setRemoveOpen(true)}
                                     permission={ADMIN}
                                     tooltipProps={{
-                                        title: isScimGroup
-                                            ? scimGroupTooltip
-                                            : 'Delete group',
```
gastonfournier, line 275: Do we have any indication that the group is managed by SCIM? Maybe something for later, but as a way of identifying that it's managed we could add an icon or a text next to the name.

gastonfournier (approved):
> That being said, this is very much an escape hatch for SCIM if something goes wrong and I'm worried about dropping a 70K entity change into the event log, both from a UI sanity perspective and from a melt-the-database perspective

Agree, we can still recover added groups from the existing events when the group was added.

Just a few suggestions but looks good

frontend/src/component/admin/groups/Group/Group.tsx
sighphyre: That's not a bad idea I think. Right now you kinda have to guess from not being able to edit it/add users to it. But that needs quite a few clicks

src/lib/db/user-store.ts
```diff
@@ -272,6 +272,10 @@ class UserStore implements IUserStore {
         await this.activeUsers().del();
     }
 
+    async deleteScimUsers(): Promise<void> {
+        await this.db(TABLE).whereNotNull('scim_id').del();
```
ivarconr, line 276: normally we delete users by clearing the email and setting the deleted_at attribute. Is there  a reason to do it differently here?

sighphyre: I think so, yes. If you're pushing the button that triggers this code it's because something has gone terribly wrong and you're looking for an undo button. So I've chosen to wipe them completely

Usually we'd want soft delete so we can restore or track those users later. In this flow, if you need to do that, you just your SCIM provider run again and they'll get restored, that's really the source of truth for those users anyway

I'm open to changing this though

chriswk: Hmm, if they sync again, we risk that they can get an error due to the scimExternalId already being in use.

I was thinking we might want to then remove scimExternalId and scimId as well as the email and setting the deleted_at attribute. If I remember correctly the sync for users is done by looking for user by email when configured correctly, and if the deleted_at flag is set we claim it isn't there, which will cause scim to try to write the user and then it will fail the uniqueness check, since the email has already been used?

@sighphyre ?

sighphyre: I thought about this some more. I actually do want to hard delete these. @ivarconr Objections?

gastonfournier: We're also under the assumption that this functionality will be needed only in the case of an error configuring SCIM, so this works more or less like a DB reset, just focused on the users table. I think this should be ok... we don't want to store emails (in event logs) that were not meant to be added to the instance.

PR #9191, opened 2025-02-03
refactor(1-3336): useTrafficData / NetworkTrafficUsage.tsx cleanup

thomasheartman:
This PR refactors the `NetworkTrafficUsage.tsx` and `useTrafficData` files a bit.

The primary objective was to make the network traffic usage component easier to work with, so I suggest to the reviewer that they start there.

Part of that refactoring, was taking things out of the useTraffic hook that didn't need to be there. In the end, I'd removed so much that I didn't even need the hook itself in the new component, so I switched that to a regular useState.

It made more sense to me to put some of the functions inside the hook into a separate file and import them directly (because they don't rely on any hook state), so I have done that and removed those functions from the trafficData hook. In this case, I also moved the tests.

I have not added any new tests in this PR, but will do so in a follow-up. The functions I intend to test have been marked as such.


frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -149,8 +156,7 @@ const createBarChartOptions = (
     },
 });
 
-// this is primarily for dev purposes. The existing grid is very inflexible, so we might want to change it, but for demoing the design, this is enough.
-const NewHeader = styled('div')(({ theme }) => ({
+const TopRow = styled('div')(({ theme }) => ({
```
thomasheartman, line 162: This isn't a dev component anymore, but part of the new component.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -178,24 +187,21 @@ const NewNetworkTrafficUsage: FC = () => {
...
-        calculateOverageCost,
-        calculateEstimatedMonthlyCost,
-        endpointsInfo,
-    } = useTrafficDataEstimation();
+
+    const [chartDataSelection, setChartDataSelection] =
+        useState<ChartDataSelection>({
+            grouping: 'daily',
+            month: selectablePeriods[0].key,
+        });
```
thomasheartman, line 195: Turns out all I needed was the selection, so there's no point in using the old big hook.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -225,65 +231,42 @@ const NewNetworkTrafficUsage: FC = () => {
...
+        toDateRange(chartDataSelection),
+    );
 
     const data = newToChartData(traffic.usage);
+    const usageTotal = calculateTotalUsage(traffic.usage);
+    const overageCost = calculateOverageCost(
+        usageTotal,
+        includedTraffic,
+        BILLING_TRAFFIC_BUNDLE_PRICE,
+    );
```
thomasheartman, line 247: We don't need all the `useState` stuff to calculate this, so I've changed it to just regular functions.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -390,8 +362,6 @@ const OldNetworkTrafficUsage: FC = () => {
         toChartData,
         toTrafficUsageSum,
         endpointsInfo,
-        calculateOverageCost,
-        calculateEstimatedMonthlyCost,
```
thomasheartman, line 394: These two functions are now imported directly from the neighboring traffic calculations file. The functions have been copied over directly along with their tests, so I'm not worried about this change.

frontend/src/component/admin/network/NetworkTrafficUsage/PeriodSelector.tsx
```diff
@@ -1,70 +1,14 @@
...
-            currentDate.getMonth() - subtractMonthCount,
-            1,
-        );
-        selectablePeriods.push(
-            toSelectablePeriod(date, undefined, date > new Date('2024-03-31')),
-        );
-    }
-    return selectablePeriods;
-};
-
```
thomasheartman, line 67: This has been moved to `selectable-periods`. It's pretty close to what's in useTraffic data, but because we also want to show unselectable months in the grid, we make some small changes. This is also what's used in the new component.

frontend/src/component/admin/network/NetworkTrafficUsage/chart-data-selection.ts
thomasheartman, line 1: Extracting the type and transformation from the hook. It deserves its separate file

frontend/src/component/admin/network/NetworkTrafficUsage/chart-functions.ts
thomasheartman, line 1: This is primarily just moving the functions that the new traffic usage component uses from useTrafficData.

frontend/src/component/admin/network/NetworkTrafficUsage/traffic-calculations.ts
```diff
@@ -0,0 +1,154 @@
...
+export const filterData = (
+    data?: TrafficUsageDataSegmentedCombinedSchema,
+): TrafficUsageDataSegmentedCombinedSchema | undefined => {
+    if (!data) {
+        return;
+    }
+    // filter out endpoints not mentioned in endpointsInfo
+    // filter out any data from before May 2024
+    return data;
+};
```
thomasheartman, line 21: This function will be implemented and tested in the next iteration.

frontend/src/component/admin/network/NetworkTrafficUsage/traffic-calculations.ts
thomasheartman, line 1: This file now contains all the traffic data calculations we're doing.

frontend/src/component/admin/billing/BillingDashboard/BillingPlan/BillingDetailsPAYG.tsx
thomasheartman, line 1: Didn't realize these two components would need an update too. From a cursory glance, it looks like they're both importing a lot of stuff from useTrafficEstimation, including functions to transform the data into chart format, but without ever using that chart data. The only thing they seem to do with it is to use it to calculate the overage cost because the function to do so uses chart data and not API data (I don't quite understand why).

I might suggest also switching these to use the new endpoint and a new function behind a flag in a followup.

daveleek: Hard to remember details right now, but it's linked to filtering and grouping/pricing of requests done in `toChartData`. We're only interested in showing the numbers we're interested in charging for.

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
daveleek: The reason for useStates littered across here is some uncertainty in whether or not a variable would be updated when dependencies updated. This might not be a problem, but potentially something to be aware of

frontend/src/component/admin/network/NetworkTrafficUsage/chart-functions.ts
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+        return toDailyChartData(traffic);
+    }
+};
+
+type SegmentedSchemaApiData =
+    TrafficUsageDataSegmentedCombinedSchema['apiData'][0];
+
+// todo: integrate filtering `filterData` frontend/src/component/admin/network/NetworkTrafficUsage/util.ts
+const prepareApiData = (
```
daveleek, line 33: Since you've implemented the filtering here, the overage calculation could just be fed this output instead of the output from toChartData

daveleek (approved):
Huge and much needed refactoring, thank you! Bit too many changes to grasp the full extent of, but looks good! A round of click-testing might not be a bad idea :)

frontend/src/component/admin/billing/BillingDashboard/BillingPlan/BillingDetailsPAYG.tsx
thomasheartman: Yeah, that's what I figured too. I'll set that up in the follow-up, I think. Thanks for clarifying!

frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
thomasheartman: yeah. I've not seen any issues with that. I'll double check and then again when this goes out to sandbox, but I think it's fine.

PR #9193, opened 2025-02-03
fix: make traffic usage test more robust by using this month's actual numbers

thomasheartman:
The test was breaking because it assumed a month would have at least
30 days.

Because the test relies on the current month, this isn't necessarily true.

Further, there's parts of the code that relies on "impure" state via
the "current date" (which will change based on when you run it), so
setting a specific month in the test won't work.

As such, this test makes the calculation explicit and uses the number
of days in the current month.


nunogois (approved):
LGTM. Thanks for having a look!

sighphyre (approved):
LGTM

gastonfournier (approved):
Nice!

PR #9195, opened 2025-02-03
chore: drop release plan template view permissions

nunogois:
https://linear.app/unleash/issue/2-3225/drop-the-release-template-view-permissions

Drops the release plan template view permissions in favor of an "open by default" approach.

Should merge the Enterprise PR first.


gastonfournier (approved):
LGTM!

PR #9197, opened 2025-02-03
fix: heroku reset for tests

kwasniew:
## About the changes

Heroku CLI needs to be installed explicitly
https://github.com/AkhileshNS/heroku-deploy/issues/188#issuecomment-2556617195


gastonfournier (approved):
Thanks! LGTM!

PR #9203, opened 2025-02-04
fix: now hydration event is being returned if revision does not exist in cache

sjaanus:
Now when customer is coming with revision ID that does not exist in cache, we return hydration.

### DIFF
--- a/src/lib/features/client-feature-toggles/delta/client-feature-delta-api.e2e.test.ts
+++ b/src/lib/features/client-feature-toggles/delta/client-feature-delta-api.e2e.test.ts
@@ -270,3 +270,54 @@ test('should get segment updated and removed events', async () => {
         ],
     });
 });
+
+test('should return hydration if revision not in cache', async () => {
+    await app.createFeature('base_feature');
+    await syncRevisions();
+    const { body, headers } = await app.request
+        .get('/api/client/delta')
+        .expect(200);
+    const etag = headers.etag;
+
+    expect(body).toMatchObject({
+        events: [
+            {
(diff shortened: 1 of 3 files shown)


src/lib/features/client-feature-toggles/delta/client-feature-delta-api.e2e.test.ts
```diff
@@ -270,3 +270,54 @@ test('should get segment updated and removed events', async () => {
...
+
+    expect(deltaBody).toMatchObject({
+        events: [
+            {
+                type: DELTA_EVENT_TYPES.FEATURE_UPDATED,
+            },
+        ],
+    });
+
+    const { body: rehydrationBody } = await app.request
```
sjaanus, line 311: We are sending 1, which does not exist in cache.

src/lib/features/client-feature-toggles/delta/delta-cache.ts
```diff
@@ -12,6 +12,22 @@ export class DeltaCache {
     constructor(hydrationEvent: DeltaHydrationEvent, maxLength: number = 20) {
         this.hydrationEvent = hydrationEvent;
         this.maxLength = maxLength;
+
+        this.addBaseEventFromHydration(hydrationEvent);
```
sjaanus, line 16: This is important change. Previously when we were initializing environment, we just filled hydration event. But hydration event changes in time, so we also need to add an event with same revision id as base event to event queue.

This way if customer got hydration with rev 100, and hydration now moves to 101, customer is coming back with 100, we need to have reference to 100.

FredrikOseberg (approved):
LGTM

PR #9205, opened 2025-02-04
chore(1-3335): filters data coming from the API to remove data points we're not  interested in

thomasheartman:
Implements a function that cleans and filters incoming data from the
new traffic API.

Specifically, it:
- Removes `/edge` data points
- Removes any data from before may 2024, which is the first full month we have on record

Because all uses of the existing hook do this filtering themselves, I
have added the filtering at the hook level. This is to avoid
forgetting this filtering later. If we find out we need this data, we
can move the filtering.


daveleek (approved):
Great! Love the naming and the constants, very clear!

PR #9207, opened 2025-02-04
feat: render markdown in toats

gastonfournier:
## About the changes
Toasts sometimes render messages that are in markdown format. This changes how we display them from 
image
to this:
Screenshot from 2025-02-04 15-23-36


nunogois (approved):
Bit unsure about the usage of `h3` as a wrapper of our markdown. Maybe @thomasheartman will have some thoughts.

nunogois:
> > Bit unsure about the usage of `h3` as a wrapper of our markdown. Maybe @thomasheartman will have some thoughts.
> 
> It was already there and doesn't look bad to the eye :D
> 
> Happy to change this later if needed, I'll delay merging this until later today to give Thomas a chance

I also think it's fine to merge as is and follow up on a separate PR if needed 👍

PR #9209, opened 2025-02-04
fix: license banner using message type

gastonfournier:
## About the changes
Actually uses messageType and maps from warn to warning.


nunogois (approved):
Nice! However I wonder if the `Banner` component should be smart enough to understand `warn` instead (move this logic to the banner component).

gastonfournier:
> Nice! However I wonder if the `Banner` component should be smart enough to understand `warn` instead (move this logic to the banner component).

I think the banner component defines an interface and the client needs to comply with it. Unfortunately I used the short version of warning in the license and it's only intended to be used in the banner, but eventually this can be used elsewhere... so I'm fine mapping from one domain to the other

PR #9217, opened 2025-02-05
chore: clear scim fields when deleting user + migration for existing cases

daveleek:
users.scim_id is unique and will cause collision errors when reinserted.

PR adds a migration that clears scim_id and scim_external_id fields on deleted users, and clears scim_id and scim_external_id when deleting users through user store


chriswk:
scim_id is our generated id though, so new insert === new scim_id. scim_external_id on the other hand, that comes from whatever system is trying to sync with us.

chriswk (approved):
LGTM

PR #9219, opened 2025-02-05
refactor: add functions to estimate monthly usage from data directly

thomasheartman:
Adds new monthly estimation functions that operate on raw usage data instead of chart data. This brings those methods in line with the rest of the traffic calculation functions that we have in that file and means we can remove other external dependencies. 

 This is somewhat inspired by #9218, but not directly linked.

### DIFF
--- a/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
+++ b/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
@@ -47,7 +47,6 @@ import {
     calculateEstimatedMonthlyCost,
 } from 'utils/traffic-calculations';
 import { currentDate, currentMonth } from './dates';
-import { endpointsInfo } from './endpoint-info';
 import { type ChartDataSelection, toDateRange } from './chart-data-selection';
 import {
     type ChartDatasetType,
@@ -301,7 +300,6 @@ const NewNetworkTrafficUsage: FC = () => {
                                         chartDataSelection.grouping === 'daily'
                                             ? usageTotal
                                             : averageTrafficPreviousMonths(
-                                                  Object.keys(endpointsInfo),
                                                   traffic.usage,
                                               )
                                     }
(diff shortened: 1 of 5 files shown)


frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -245,10 +247,7 @@ const NewNetworkTrafficUsage: FC = () => {
     );
 
     const estimatedMonthlyCost = calculateEstimatedMonthlyCost(
-        chartDataSelection.grouping === 'daily'
-            ? chartDataSelection.month
-            : currentMonth,
-        data.datasets,
+        traffic.usage?.apiData,
```
thomasheartman, line 250: This ? is because `usage` might be undefined. This is because the hook doesn't account for whether you have data or not. I'll address this in (yet another) followup.

frontend/src/component/admin/network/NetworkTrafficUsage/average-traffic-previous-months.ts
thomasheartman, line 1: This was an incidental refactoring. Because we now filter the data before it goes to the compnoent, we can remove this filtering right here.

thomasheartman, line 1: This was an incidental refactoring. We now filter the data before it goes into the component, so we can drop it here. Additionally, we have the current month defined elsewhere.

frontend/src/hooks/useTrafficData.ts
thomasheartman, line 1: This is moving the old functions back to where they were originally defined. Because we no longer want them in traffic-calculations, we might as well put them back so we remember to delete them.

frontend/src/utils/traffic-calculations.test.ts
thomasheartman, line 1: The changes here are all about testing the new implementations against the old one. Given the same data in, they should get the same result out.

daveleek (approved):

PR #9221, opened 2025-02-05
chore: use union type for traffic search data

thomasheartman:
Makes the data returned from the traffic search a union type to avoid
nasty object-is-undefined errors at runtime.

It requires more explicit handling, sure, but it means we don't need
to accept undefined.

### DIFF
--- a/frontend/src/component/admin/billing/BillingDashboard/BillingPlan/useOverageCost.ts
+++ b/frontend/src/component/admin/billing/BillingDashboard/BillingPlan/useOverageCost.ts
@@ -30,16 +30,19 @@ const useNewOverageCostCalculation = (includedTraffic: number) => {
     const from = formatDate(startOfMonth(now));
     const to = formatDate(endOfMonth(now));
 
-    const { usage } = useTrafficSearch('daily', { from, to });
-
+    const { result } = useTrafficSearch('daily', { from, to });
     const overageCost = useMemo(() => {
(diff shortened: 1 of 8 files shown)


frontend/src/hooks/api/getters/useInstanceTrafficMetrics/useInstanceTrafficMetrics.ts
thomasheartman, line 1: This is the main part of this PR: change the type to only have a data property if we actually have something. This means we don't need to keep checking "what if data isn't there" means we need to handle it more explicitly.

frontend/src/component/admin/billing/BillingDashboard/BillingPlan/useOverageCost.ts
```diff
@@ -30,16 +30,19 @@ const useNewOverageCostCalculation = (includedTraffic: number) => {
...
-
+    const { result } = useTrafficSearch('daily', { from, to });
     const overageCost = useMemo(() => {
-        const totalUsage = calculateTotalUsage(usage);
+        if (result.state !== 'success') {
+            return 0;
+        }
+
+        const totalUsage = calculateTotalUsage(result.data);
         return calculateOverageCost(
```
daveleek, line 40: Neat!

daveleek (approved):
Cool! 😎

PR #9223, opened 2025-02-05
chore(1-3333): add new pnps reset

thomasheartman:
This migration resets the pNPS survey results older than 6 months, so that the survey appears again for users who have already completed it.


FredrikOseberg (approved):
LGTM

PR #9224, opened 2025-02-05
chore: remove lifecycle v2 flag

Tymek:
## About the changes
Cleanup


frontend/src/component/common/FeatureLifecycle/FeatureLifecycleStageIcon.tsx
```diff
@@ -1,47 +1,25 @@
...
-        ) : (
-            <CompletedStageIcon {...props} />
-        );
-    } else {
-        return newIcons ? (
-            <Stage1 {...props} />
-        ) : (
-            <InitialStageIcon {...props} />
-        );
+        return <Stage5 {...props} />;
```
kwasniew, line 13: do we want to sync Stage names instead of using index numbers?

kwasniew (approved):

frontend/src/component/common/FeatureLifecycle/FeatureLifecycleStageIcon.tsx
Tymek: Good question. I think numbers here make sense, because SVGs are created this way. This component maps it so outside we can use stage names.

PR #9228, opened 2025-02-05
feat: decorate emails in banner view

gastonfournier:
Some responses in the backend include emails, in order to show them as links we're returning them from the backend as markdown. I believe this creates issues because we're coupling backend and frontend (it's similar to returning html: `<a href="mailto:...">email</a>`). 

I believe it should be the frontend who decides how to render these links.

Having these emails returned as markdown also creates issues as now all interfaces should support markdown. One example we forgot about is the toast, but also CLI or using the REST API now receive the markdown.

This moves the responsibility to the UI, the solution could be better (HATEOAS comes to mind, but I feel it's an overkill)


gastonfournier:
For now we're going to stick to the markdown format, it seems this requires a slightly wider conversation and I don't want to introduce this parsing of messages to convert to markdown in the UI as it would degrade the performance should this become a common practice

PR #9240, opened 2025-02-06
feat: implement dialogs for changerequest milestone handling and removing release plans

daveleek:
image
Skjermbilde 2025-02-06 kl  10 20 10


frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlan.tsx
```diff
@@ -187,6 +225,27 @@ export const ReleasePlan = ({
...
+                environmentId={environment}
+                featureId={featureName}
+                onClosing={() => {
+                    setPlanForChangeRequestStartMilestoneDialog(undefined);
+                    setMilestoneForChangeRequestDialog(undefined);
+                }}
+                projectId={projectId}
+                releasePlan={planForChangeRequestStartMilestoneDialog}
+                milestone={milestoneForChangeRequestDialog}
+            />
```
daveleek, line 248: I set two up here so I wouldn't have to let the action be nullable & have another react state for it. I could change the logic in the dialog to looking at milestone being set, only release plan being set, or only template being set though. WDYT @nunogois?

nunogois: Uh sure. Honestly I would have expected separate dialog components for each action, but if this works and you're happy I don't think it matters much 🤷

nunogois (approved):

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ChangeRequest/ReleasePlanAddChangeRequestDialog.tsx
```diff
@@ -1,30 +1,45 @@
...
+} from 'interfaces/releasePlans';
 import { useChangeRequestApi } from 'hooks/api/actions/useChangeRequestApi/useChangeRequestApi';
+import { usePendingChangeRequests } from 'hooks/api/getters/usePendingChangeRequests/usePendingChangeRequests';
 
 const StyledBoldSpan = styled('span')(({ theme }) => ({
     fontWeight: theme.typography.fontWeightBold,
 }));
 
 interface IReleasePlanAddChangeRequestDialogProps {
+    action: 'addReleasePlan' | 'deleteReleasePlan' | 'startMilestone';
```
nunogois, line 17: Is this needed now that we split them up?

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ChangeRequest/ReleasePlanAddChangeRequestDialog.tsx
```diff
@@ -1,30 +1,45 @@
...
     projectId: string;
     featureId: string;
     environmentId: string;
-    releaseTemplate: IReleasePlanTemplate | undefined;
+    releaseTemplate?: IReleasePlanTemplate | undefined;
+    releasePlan?: IReleasePlan | undefined;
+    milestone?: IReleasePlanMilestone | undefined;
+    environmentActive?: boolean;
```
nunogois, line 24: Seems like these 2 are unused

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ChangeRequest/ReleasePlanAddChangeRequestDialog.tsx
```diff
@@ -45,7 +62,7 @@ export const ReleasePlanAddChangeRequestDialog = ({
     return (
         <Dialogue
             title='Request changes'
-            open={Boolean(releaseTemplate)}
+            open={Boolean(releaseTemplate) || Boolean(releasePlan)}
```
nunogois, line 65: Not a huge fan of this ambiguity. What is this dialog doing that it shows up if one of these is truthy? I also prefer to handle the open state explicitly instead of relying on other state variables, but that's mostly a personal preference.

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ChangeRequest/RemoveReleasePlanChangeRequestDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+        });
+
+        refetchChangeRequests();
+
+        setToastData({
+            type: 'success',
+            text: 'Added to draft',
+        });
+        onClosing();
+    };
```
nunogois, line 50: IMO the caller should own this. I think dialogs should only care about displaying the required information and emitting the "confirmation" callback to the parent, which then handles it.

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlan.tsx
```diff
@@ -96,6 +100,33 @@ export const ReleasePlan = ({
...
+        setPlanForChangeRequestRemoveDialog,
+    ] = useState<IReleasePlan | undefined>();
+    const [
+        planForChangeRequestStartMilestoneDialog,
+        setPlanForChangeRequestStartMilestoneDialog,
+    ] = useState<IReleasePlan | undefined>();
+    const [
+        milestoneForChangeRequestDialog,
+        setMilestoneForChangeRequestDialog,
+    ] = useState<IReleasePlanMilestone | undefined>();
```
nunogois, line 114: You can use something like `selectedReleasePlan` and share that state across multiple dialogs, since only one of them will be open at a time. I'm also pretty sure you don't need to specify `| undefined` here, it's implicit.

nunogois: What's preventing us from making these dialogs work the same way as the existing `ReleasePlanRemoveDialog`?

daveleek: > since only one of them will be open at a time

We moved the "am I open" part inside the dialog, so this follows that pattern. And to check if it's open - it checks to see if the provided property is defined (or boolean). So if we share, then both delete and start milestone will open in some sequence.

We could move to having an isOpen for them, but then that would be 3 isOpens instead.

nunogois: > but then that would be 3 isOpens instead.

That feels more intuitive for me.

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ChangeRequest/ReleasePlanAddChangeRequestDialog.tsx
daveleek: Strange, I removed these. why doesn't that show up?

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ChangeRequest/ReleasePlanAddChangeRequestDialog.tsx
daveleek: This is also gone, very strange

PR #9242, opened 2025-02-06
Fix image zoom bug

melindafekete:
This PR addresses the following changes:

- Fixes a bug (see video below) where images wouldn't render correctly when zoomed in
- Adds error handling to the script that fetches repo stars (thank you @nunogois)
- Removes the custom Figure component to make sure all images render correctly and consistently across the docs
- Updates an outdated architecture image for Unleash introductory overview (to remove 'Proxy)

https://github.com/user-attachments/assets/1d0599e5-216c-4b20-aaa3-f308466b7bb9


nnennandukwe (approved):
Nice clean up. So we won't need the `Figure` component moving forward?

melindafekete:
> Nice clean up. So we won't need the `Figure` component moving forward?

No, let's just use the regular images. At the moment, the component is broken, and it was only used in some of the places, so I wanted to remove it anyway for consistency reasons.

PR #9250, opened 2025-02-07
feat: emit delta updated event

sjaanus:
Start emitting delta updated event and make delta cache service singleton.


FredrikOseberg (approved):
LGTM

PR #9252, opened 2025-02-07
chore(1-3349): no border for event timeline / dropdowns to the right

thomasheartman:
Removes the border between the accordion summary and its contents for the event timeline, and moves dropdown selectors to the right to avoid overcrowding on the left.

image


nunogois (approved):
LGTM

PR #9253, opened 2025-02-07
fix: open/close animation on personal dashboard is choppy

thomasheartman:
Extracts each panel into its own component for the personal dashboard. This lets us use separate states for each panel, which in turn lets each panel change its open / close state without causing the other panels to re-render.

When you have a lot of flags and/or projects, the list to render becomes very long, which causes performance problems, especially when you need to rerender both flags and projects and the timeline whenever one of them changes.

The problems were especially noticeable in Firefox for me. Even with this, the event timeline is a little choppy. I suspect that's because of it might take a long time to paint? But we can look into that later.

Also updates the dashboard state hook to let you only pass in the flags/projects you want. We could extract this into three different hooks that all use the same localhost key, but I'm not sure whether that's better or worse 🤷🏼

### DIFF
--- a/frontend/src/component/personalDashboard/useDashboardState.ts
+++ b/frontend/src/component/personalDashboard/useDashboardState.ts
@@ -5,10 +5,11 @@ import type {
 } from 'openapi';
 import { useEffect } from 'react';
 
-export const useDashboardState = (
-    projects: PersonalDashboardSchemaProjectsItem[],
-    flags: PersonalDashboardSchemaFlagsItem[],
-) => {
+type StateProps = {
+    projects?: PersonalDashboardSchemaProjectsItem[];
+    flags?: PersonalDashboardSchemaFlagsItem[];
+};
+export const useDashboardState = (props?: StateProps) => {
     type State = {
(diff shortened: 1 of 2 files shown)


frontend/src/component/personalDashboard/useDashboardState.ts
```diff
@@ -35,34 +36,31 @@ export const useDashboardState = (
...
-                !flags.some((flag) => flag.name === state.activeFlag?.name));
 
-        if (setDefaultFlag) {
-            updates.activeFlag = flags[0];
+        if (
+            props?.flags?.length &&
+            (!state.activeFlag ||
+                !props.flags.some(
+                    (flag) => flag.name === state.activeFlag?.name,
+                ))
```
thomasheartman, line 45: I took these out of variables because the linter didn't recognize `props.flags` as definitely being there if the variable was true. So to avoid using `!`, I put it in there.

kwasniew (approved):
I'd keep the hook as you did so that the storage key doesn't get out of sync between the 3 hooks

PR #9255, opened 2025-02-07
chore: make the milestone status a button

nunogois:
https://linear.app/unleash/issue/2-3251/make-the-milestone-status-action-link-and-icon-a-proper-button-that

Makes the milestone status a button while trying to keep most of the original design intact.

image

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlanMilestone/ReleasePlanMilestoneStatus.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlanMilestone/ReleasePlanMilestoneStatus.tsx
@@ -1,22 +1,35 @@
-import { Button, styled } from '@mui/material';
+import { styled } from '@mui/material';
 import PlayCircleIcon from '@mui/icons-material/PlayCircle';
 import PauseCircleIcon from '@mui/icons-material/PauseCircle';
 import TripOriginIcon from '@mui/icons-material/TripOrigin';
 
 export type MilestoneStatus = 'not-started' | 'active' | 'paused' | 'completed';
 
-const StyledStatusButton = styled(Button, {
+const StyledStatusButton = styled('button', {
     shouldForwardProp: (prop) => prop !== 'status',
-})<{ status: MilestoneStatus }>(({ theme, status }) => ({
-    backgroundColor:
-        status === 'active' ? theme.palette.success.light : 'transparent',
-    '&&&': {
+})<{ status: MilestoneStatus; disabled?: boolean }>(
+    ({ theme, status, disabled }) => ({
+        display: 'flex',
+        alignItems: 'center',
+        border: 'none',
+        gap: theme.spacing(1),
+        padding: 0,
+        paddingRight: theme.spacing(1),
+        cursor: 'pointer',
+        backgroundColor:
+            status === 'active' ? theme.palette.success.light : 'transparent',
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlanMilestone/ReleasePlanMilestoneStatus.tsx
```diff
@@ -1,36 +1,48 @@
...
     backgroundColor:
         status === 'active' ? theme.palette.success.light : 'transparent',
-    color:
-        status === 'active'
-            ? theme.palette.success.contrastText
-            : status === 'completed'
-              ? theme.palette.text.secondary
-              : theme.palette.text.primary,
-    '& svg': {
+    '&&&': {
```
thomasheartman, line 13: I've seen this before: is this just a hack to override the specificity? In this case, why use a MUI button at all? Why not just use a regular `button`?

thomasheartman (approved):

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlanMilestone/ReleasePlanMilestoneStatus.tsx
nunogois: Yep, it is. How do you feel about this instead? https://github.com/Unleash/unleash/pull/9255/commits/e60385c30533a87470bd9fcc08ce874a0f5a4f62

thomasheartman: Nice! I think that's better, but why use `pointer-events: none`? Wouldn't it be better to use `cursor: default`? You're already using the disabled attribute, which makes it unreachable by tab anyway.

nunogois: This brings back our previous behavior of the button being "click-through" when disabled, which means that if the milestone is running or paused, and you click the disabled button, you still toggle the accordion. The cursor is still pointer because, again, we "click-through" the button and toggle the accordion.

Would you expect a different behavior?

thomasheartman: Ah, got it. Yeah, I can see the logic there. I'm not sure it's what I'd expect, but if it's UX approved, then I don't really have any arguments against it. maybe that it's unintuitive? Do disabled buttons still get hover styles? Because it might be strange if it looks like it can be interacted with but it can't? It can't get tab focus, so that's not an issue though.

nunogois: > Do disabled buttons still get hover styles?

No, they don't. They act the same as before this PR.

I'm proceeding with this for now but we can revisit this later if needed.

thomasheartman: Great! My point was I think it'd be confusing if they did. So yeah, happy for you to merge it 😄

PR #9257, opened 2025-02-07
chore: add release management to new in unleash

nunogois:
https://linear.app/unleash/issue/2-3244/add-release-management-to-new-in-unleash

Adds release management to "New in Unleash".

image


daveleek (approved):
Brilliant!

PR #9258, opened 2025-02-07
docs: remove notifications docs

kwasniew:
## About the changes

Closes #


website/sidebars.ts
```diff
@@ -444,10 +444,9 @@ const sidebars: SidebarsConfig = {
                             type: 'category',
                             link: {
                                 type: 'doc',
-                                id: 'reference/notifications',
```
kwasniew, line 447: it looks like this one has to point to the first selected item. Since notifications are gone I picked the next one from the list

melindafekete (approved):

PR #9259, opened 2025-02-07
fix(1-3296): don't navigate through the list on tab

thomasheartman:
Prevents tab from navigating you through the list of results. Instead
makes it so that the tab key always takes you to the next item in the
same hierarchy.

As a bonus: also automatically closes the menu when you navigate
away (the previous implementation has a bug where it wouldn't if you
shift-tab).

The behavior of not letting you navigate the list with tab is
consistent with native HTML select elements as well as MUI select
elements. You typically navigate them with the arrow keys.

### DIFF
--- a/frontend/src/component/commandBar/CommandBar.tsx
+++ b/frontend/src/component/commandBar/CommandBar.tsx
@@ -288,6 +288,10 @@ export const CommandBar = () => {
         },
     );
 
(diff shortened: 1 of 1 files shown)


daveleek (approved):
I suppose we don't have any data on how many people this would potentially affect (ie people used to tabbing through the list)
But maybe it's not a big deal as arrows still work

PR #9265, opened 2025-02-09
chore: Create test db from template

gastonfournier:
## About the changes
Based on the first hypothesis from https://github.com/Unleash/unleash/pull/9264, I decided to find an alternative way of initializing the DB, mainly trying to run migrations only once and removing that from the actual test run.

I found in [Postgres template databases](https://www.postgresql.org/docs/current/manage-ag-templatedbs.html) an interesting option in combination with jest global initializer.

### Changes on how we use DBs for testing

Previously, we were relying on a single DB with multiple schemas to isolate tests, but each schema was empty and required migrations or custom DB initialization scripts.

With this method, we don't need to use different schema names (apparently there's no templating for schemas), and we can use new databases. We can also eliminate custom initialization code. 

### Legacy tests

This method also highlighted some wrong assumptions in existing tests. One example is the existence of `default` environment, that because of being deprecated is no longer available, but because tests are creating the expected db state manually, they were not updated to match the existing db state. 

To keep tests running green, I've added a configuration to use the `legacy` test setup (24 tests). By migrating these, we'll speed up tests, but the code of these tests has to be modified, so I leave this for another PR.

## Downsides
1. The template db initialization happens at the beginning of any test, so local development may suffer from slower unit tests. As a workaround we could define an environment variable to disable the db migration
2. Proliferation of test dbs. In ephemeral environments, this is not a problem, but for local development we should clean up from time to time. There's the possibility of cleaning up test dbs using the db name as a pattern: https://github.com/Unleash/unleash/blob/2ed2e1c27418b92e815d06c351504005cf083fd0/scripts/jest-setup.ts#L13-L18 but I didn't want to add this code yet. Opinions?

## Benefits
1. It allows us migrate only once and still get the benefits of having a well known state for tests.
3. It removes some of the custom setup for tests (which in some cases ends up testing something not realistic)
4. It removes the need of testing migrations: https://github.com/Unleash/unleash/blob/main/src/test/e2e/migrator.e2e.test.ts as migrations are run at the start
5. Forces us to keep old tests up to date when we modify our database


src/lib/features/feature-toggle/tests/feature-toggle-last-seen-at.e2e.test.ts
```diff
@@ -117,10 +91,7 @@ test('response should include last seen at per environment for multiple environm
     const { body } = await app.request.get(`/api/admin/archive/features`);
 
     const featureEnvironments = body.features[0].environments;
-    const [def, development, production] = featureEnvironments;
```
gastonfournier, line 120: Default env no longer returned

src/lib/features/instance-stats/getProductionChanges.e2e.test.ts
```diff
@@ -44,12 +44,6 @@ const noEnvironmentEvent = (days: number) => {
 
 beforeAll(async () => {
     db = await dbInit('product_changes_serial', getLogger);
-    await db.rawDatabase('environments').insert({
-        name: 'production',
-        type: 'production',
-        enabled: true,
-        protected: false,
-    });
```
gastonfournier, line 52: No need to initialize

src/lib/types/core.test.ts
```diff
@@ -24,36 +10,21 @@ describe('System user definitions in code and db', () => {
...
-
-        await initSchema(config.db);
-
-        const e2e = {
-            ...config.db,
-            connectionTimeoutMillis: 2000,
-        };
+        db = await dbInit('system_user_alignment_test', getLogger);
 
-        await migrateDb(config);
```
gastonfournier, line 45: Much simpler using a db that's already migrated

src/test/e2e/helpers/database-init.ts
```diff
@@ -94,46 +96,79 @@ export interface ITestDb {
...
+    dbInitMethod?: 'legacy' | 'template';
+};
+
 export default async function init(
     databaseSchema = 'test',
     getLogger: LogProvider = noLoggerProvider,
-    configOverride: Partial<IUnleashOptions> = {},
+    configOverride: Partial<IUnleashOptions & DBTestOptions> = {},
 ): Promise<ITestDb> {
+    const testDbName = `unleashtestdb_${uuidv4().replace(/-/g, '')}`;
```
gastonfournier, line 108: Generates a random db name to be used during testing

src/test/e2e/helpers/database-init.ts
```diff
@@ -94,46 +96,79 @@ export interface ITestDb {
...
+        (configOverride.dbInitMethod ?? 'template') === 'template';
+    const testDBTemplateName = process.env.TEST_DB_TEMPLATE_NAME;
     const config = createTestConfig({
         db: {
             ...getDbConfig(),
             pool: { min: 1, max: 4 },
-            schema: databaseSchema,
+            ...(useDbTemplate
+                ? { database: testDbName }
+                : { schema: databaseSchema }),
```
gastonfournier, line 118: Schemas are no longer needed, we'll use a new db instead

src/test/e2e/helpers/database-init.ts
```diff
@@ -94,46 +96,79 @@ export interface ITestDb {
...
+
 export default async function init(
     databaseSchema = 'test',
     getLogger: LogProvider = noLoggerProvider,
-    configOverride: Partial<IUnleashOptions> = {},
+    configOverride: Partial<IUnleashOptions & DBTestOptions> = {},
 ): Promise<ITestDb> {
+    const testDbName = `unleashtestdb_${uuidv4().replace(/-/g, '')}`;
+    const useDbTemplate =
+        (configOverride.dbInitMethod ?? 'template') === 'template';
```
gastonfournier, line 110: Some tests require more work to migrate to this strategy, so with this option I can still run tests that need refactoring. It also makes it easy to identify them

src/test/e2e/helpers/database-init.ts
```diff
@@ -94,46 +96,79 @@ export interface ITestDb {
...
+            throw new Error(
+                'TEST_DB_TEMPLATE_NAME environment variable is not set',
+            );
+        }
+        const client = new Client(getDbConfig());
+        await client.connect();
+
+        await client.query(
+            `CREATE DATABASE ${testDbName} TEMPLATE ${testDBTemplateName}`,
+        );
```
gastonfournier, line 138: Assumes TEST_DB_TEMPLATE_NAME should exist as it's created by jest-setup.ts script. It also assumes the db contains an already initialized and migrated db for Unleash

src/test/e2e/helpers/database-init.ts
```diff
@@ -94,46 +96,79 @@ export interface ITestDb {
...
-    await resetDatabase(testDb);
-    await setupDatabase(stores);
-    await restoreRolePermissions(testDb, defaultRolePermissions);
+
+    if (!useDbTemplate) {
+        const defaultRolePermissions =
+            await getDefaultEnvRolePermissions(testDb);
+        await resetDatabase(testDb);
+        await setupDatabase(stores);
+        await restoreRolePermissions(testDb, defaultRolePermissions);
```
gastonfournier, line 158: All these functions simulate the migration. Once all tests are migrated we can remove them

src/test/e2e/helpers/database-init.ts
```diff
@@ -94,46 +96,79 @@ export interface ITestDb {
...
-            await resetDatabase(testDb);
-            await setupDatabase(stores);
-            await restoreRolePermissions(testDb, defaultRolePermissions);
+            if (!useDbTemplate) {
+                const defaultRolePermissions =
+                    await getDefaultEnvRolePermissions(testDb);
+                await resetDatabase(testDb);
+                await setupDatabase(stores);
+                await restoreRolePermissions(testDb, defaultRolePermissions);
+            }
```
gastonfournier, line 171: No need to reset a db when you can get a new one at a cheaper cost than reseting

scripts/jest-setup.ts
```diff
@@ -0,0 +1,39 @@
+import { Client, type ClientConfig } from 'pg';
+import { migrateDb } from '../src/migrator';
+import { getDbConfig } from '../src/test/e2e/helpers/database-config';
+
+let initializationPromise: Promise<void> | null = null;
+const initializeTemplateDb = (db: ClientConfig): Promise<void> => {
+    if (!initializationPromise) {
+        initializationPromise = (async () => {
+            const testDBTemplateName = process.env.TEST_DB_TEMPLATE_NAME;
+            const client = new Client(db);
```
kwasniew, line 10: I'd extract into resetDB fn so that we have similar level of abstraction inside insitalizeTempalteDb between migrateDb function (encapsulated migrations) and this new code (encapsulated reset db).

kwasniew (approved):

PR #9266, opened 2025-02-09
fix: Killwitch to block usage-metrics from non-exiting flag-names

ivarconr:
Adds a killswitch called "filterExistingFlagNames". When enabled it will filter out reported SDK metrics and remove all reported metrics for names that does not match an exiting feature flag in Unleash. 

This have proven critical in the rare case of an SDK that start sending random flag-names back to unleash, and thus filling up the database. At some point the database will start slowing down due to the noisy data. 

In order to not resolve the flagNames all the time we have added a small cache (10s) for feature flag names. This gives a small delay (10s) from flag is created until we start allow metrics for the flag when kill-switch is enabled. We should probably listen to the event-stream and use that invalidate the cache when a flag is created.

### DIFF
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -67,7 +67,6 @@ exports[`should create default config 1`] = `
     "_events": {},
     "_eventsCount": 0,
     "_maxListeners": undefined,
-    Symbol(shapeMode): false,
     Symbol(kCapture): false,
   },
   "feedbackUriPath": undefined,
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -67,6 +67,7 @@ exports[`should create default config 1`] = `
     "_events": {},
     "_eventsCount": 0,
     "_maxListeners": undefined,
+    Symbol(shapeMode): false,
     Symbol(kCapture): false,
   },
   "feedbackUriPath": undefined,
--- a/src/lib/features/metrics/client-metrics/client-metrics-store-v2.ts
+++ b/src/lib/features/metrics/client-metrics/client-metrics-store-v2.ts
@@ -155,6 +155,7 @@ export class ClientMetricsStoreV2 implements IClientMetricsStoreV2 {
         throw new NotFoundError(`Could not find metric`);
     }
 
+    //TODO: Consider moving this to a specific feature store
     async getFeatureFlagNames(): Promise<string[]> {
         return this.db(FEATURES_TABLE).distinct('name').pluck('name');
     }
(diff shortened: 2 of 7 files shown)


src/lib/features/metrics/client-metrics/client-metrics-store-v2.ts
```diff
@@ -33,6 +33,8 @@ const DAILY_TABLE = 'client_metrics_env_daily';
 const HOURLY_TABLE_VARIANTS = 'client_metrics_env_variants';
 const DAILY_TABLE_VARIANTS = 'client_metrics_env_variants_daily';
 
+const FEATUREs_TABLE = 'features';
```
nunogois, line 36:
```suggestion
const FEATURES_TABLE = 'features';
```

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -103,6 +113,29 @@ export default class ClientMetricsServiceV2 {
...
+                        `Filtered out ${toggleNames.length - existingNames.length} toggles with non-existing names`,
+                    );
+                }
+                return this.filterValidToggleNames(existingNames);
+            } catch (e) {
+                this.logger.error(e);
+                return this.filterValidToggleNames(toggleNames);
+            }
+        } else {
+            return this.filterValidToggleNames(toggleNames);
```
nunogois, line 135: Can be simplified to always return this.filterValidToggleNames(toggleNames); without the else and the return in the catch.

src/lib/__snapshots__/create-config.test.ts.snap
```diff
@@ -67,7 +67,6 @@ exports[`should create default config 1`] = `
     "_events": {},
     "_eventsCount": 0,
     "_maxListeners": undefined,
-    Symbol(shapeMode): false,
```
nunogois, line 70: Not sure where this change is coming from 🤔

nunogois (approved):
LGTM!

Not entirely sure if we should add this new query in the metrics store compared to adding it in the features store and providing the features store in the metrics service. Fetching flag names feels like something very much in the feature flag domain.

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
ivarconr: yes, makes sense

ivarconr:
> LGTM!
> 
> Not entirely sure if we should add this new query in the metrics store compared to adding it in the features store and providing the features store in the metrics service. Fetching flag names feels like something very much in the feature flag domain.

In general I agree, this was a short-cut to make it work. I was not entirely sure where to put it right now due to our package-by-feature, and did not want make the feature stored shared right now.

nunogois:
> > LGTM!
> > 
> > Not entirely sure if we should add this new query in the metrics store compared to adding it in the features store and providing the features store in the metrics service. Fetching flag names feels like something very much in the feature flag domain.
> 
> In general I agree, this was a short-cut to make it work. I was not entirely sure where to put it right now due to our package-by-feature, and did not want make the feature stored shared right now. 

Yep, sounds reasonable! 👍

PR #9270, opened 2025-02-10
chore: removes extra border on collapse for the event timeline

thomasheartman:
Fixes a small visual glitch where the event timeline panel (which
usually doesn't have a bottom border on the summary) would get a
bottom border during the collapsing animation.

This happens because to make the border act as we want, we switch
between using the summary's bottom border and the content's top
border, and I'd only updated one of the borders to respect the new design.


nunogois (approved):
LGTM

PR #9271, opened 2025-02-10
chore: make playground code editor height dynamic

thomasheartman:
The playground code editor had a fixed height of `150px`. This works
well with the current font size, but if we're changing it, we'll end
up with too much height compared to the font size.

So instead, let's calculate the font size based on the current font
size.


frontend/src/component/playground/Playground/PlaygroundForm/PlaygroundCodeFieldset/PlaygroundEditor/PlaygroundEditor.tsx
```diff
@@ -97,6 +97,9 @@ export const PlaygroundEditor: VFC<IPlaygroundEditorProps> = ({
         [setContext],
     );
 
+    // (placeholder line count * font size * line height) + padding
+    const codeInputHeight = `calc(6 * ${theme.typography.body1.fontSize} * ${theme.typography.body1.lineHeight}) + 8px`;
```
Tymek, line 101: \+ `theme.spacing(2)`?

Tymek (approved):

PR #9274, opened 2025-02-10
Update license policy doc

melindafekete:
Updates license policy document with the changes:
- Simplify content, remove screenshots of self-explanatory UI
- Explain new license enforcement behaviours

### DIFF
--- a/website/docs/using-unleash/deploy/license-keys.mdx
+++ b/website/docs/using-unleash/deploy/license-keys.mdx
@@ -1,119 +1,83 @@
 ---
-title: License Keys for Unleash Enterprise
+title: License Keys
 ---
 
-import VideoContent from "@site/src/components/VideoContent.jsx";
+:::note Availability
 
-This document describes all aspects and workflows of license key management in Unleash, available from 5.8 and required for self-hosted enterprise plans from v6.0 onwards.
-
-_For questions, please reach out to Unleash Customer Success using the [linked address](mailto:customersuccess@getunleash.io)_.
-
-:::info Warning
-
-License key management applies only to **self-hosted, Enterprise** versions of Unleash.
-
--   For **Unleash hosted** (Pro/Enterprise), license management is not necessary and is handled automatically.
--   For **Unleash Open Source**, no license is necessary and these steps do not apply.
+**Plan**: [Enterprise](https://www.getunleash.io/pricing) | **Version**: `5.8+`
 
 :::
 
-Areas of license key management that will be covered include:
-
--   [License policy](#license-policy)
-    -   [Enforcement](#enforcement)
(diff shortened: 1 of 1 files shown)


website/docs/using-unleash/deploy/license-keys.mdx
```diff
@@ -1,119 +1,83 @@
...
--   **Expired license key**  
-    ![Banner showing an expired license](/img/license-banner-expired.png)
--   **Enroled user count higher than the number of seats supported by the license**  
-    ![Banner showing a high user count](/img/license-banner-highuser.png)
-
----
+:::note Availability
 
-## Obtaining a new or updated license
+**Plan**: [Enterprise](https://www.getunleash.io/pricing) self-hosted | **Version**: `6.7+`
```
melindafekete, line 31: What do we mean when we say that hosted Enterprise 'license management is handled automatically'. Does it mean that the following enforcement never applies to them?

gastonfournier: Handled automatically is probably not accurate as there's no license for hosted Enterprise. We just use different methods to validate that users stay within their contracted limits

website/docs/using-unleash/deploy/license-keys.mdx
```diff
@@ -1,119 +1,83 @@
...
+1. Go to **Admin** > **License**.
+2. Paste the new license key and click **Update license key**.
+
+After updating a license key, you may need to refresh your browser for the license warning banners to be removed.
 
-### Environment Variable
+### Using an environment variable
 
-License key installation or update through an environment variable on container deployment is also possible, using the format `UNLEASH_LICENSE`.
+You can also install or update your license key on container deployment, using the environment variable `UNLEASH_LICENSE`.
```
gastonfournier, line 77:
```suggestion
You can also install or update your license key on container deployment, using the environment variable `UNLEASH_LICENSE`.

If you try to update your license with an invalid one, or one that was issued before your existing license, the existing license will prevail.
```
Not sure if this is well written, but I can foresee questions about "what happens if" and this could clarify some of them.

gastonfournier (approved):
LGTM!

website/docs/using-unleash/deploy/license-keys.mdx
melindafekete: Ok, thanks! But do we have the same restrictions in place (read-only mode etc) in case they do not stay within the limits?

gastonfournier: We don't go into read-only mode, no. We handle things differently for hosted customers, mainly reaching out by email or customer success team.

ivarconr (approved):

PR #9275, opened 2025-02-10
chore(1-3356): make mui buttons respect font size changes

thomasheartman:
Makes the MUI buttons use `body1.fontSize` for their own font size.

Turns out we had a hardcoded `16px` in the `app.css` file, which made it hard to override. This change removes mui button styling in `app.css` and puts it into the `theme.ts` file instead.

Also updates the border radius to match the one we use in the theme (from 3px to 4px). It's a subtle change:

before:
image

after:
image

### DIFF
--- a/frontend/src/themes/theme.ts
+++ b/frontend/src/themes/theme.ts
@@ -323,6 +323,17 @@ export default createTheme({
             },
         },
 
+        // Buttons
+        MuiButton: {
+            styleOverrides: {
+                root: ({ theme }) => ({
+                    fontSize: theme.typography.body1.fontSize,
+                    borderRadius: '3px',
+                    textTransform: 'none',
+                }),
+            },
+        },
+
         // Links
         MuiLink: {
             styleOverrides: {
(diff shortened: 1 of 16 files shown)


frontend/src/themes/theme.ts
```diff
@@ -323,6 +323,17 @@ export default createTheme({
             },
         },
 
+        // Buttons
+        MuiButton: {
+            styleOverrides: {
+                root: ({ theme }) => ({
+                    fontSize: theme.typography.body1.fontSize,
+                    borderRadius: '3px',
```
Tymek, line 331: We don't have this size in `theme.shape`. Should this maybe be 4px to match `theme.shape.borderRadius`?

Tymek (approved):

frontend/src/themes/theme.ts
thomasheartman: Maybe. It's been this way since forever, it seems. But yeah, I don't mind using the theme variable instead.

thomasheartman: Done ✅

PR #9277, opened 2025-02-10
fix: don't check license in OSS

thomasheartman:
The license check API call was giving me 404s in the console of the
OSS version of Unleash.

This changes the `useLicense` hook to use `useEnterpriseSWR` instead
of `useSWR` to avoid making the API call in the OSS version. This is
consistent with the `useLicenseCheck` hook in the same file.


gastonfournier (approved):
Thanks!!

PR #9289, opened 2025-02-11
chore: use gitcliff changelog for release

gastonfournier:
This is to reuse the gitcliff output for the release


chriswk (approved):
Nice!

PR #9293, opened 2025-02-11
fix(1-3363): frontend flag type should not return truthy for variant

Tymek:
## About the changes
Prevent issues with disabled variants resolving as truthy in frontend.

### DIFF
--- a/frontend/src/hooks/useUiFlag.ts
+++ b/frontend/src/hooks/useUiFlag.ts
@@ -4,9 +4,12 @@ type flags = ReturnType<typeof useUiConfig>['uiConfig']['flags'];
 
 export const useUiFlag = <K extends keyof flags>(flag: K): boolean => {
     const { uiConfig } = useUiConfig();
-    if (typeof uiConfig?.flags?.[flag] !== 'boolean') {
-        console.error(`Flag ${flag} is undefined or a variant.`);
+    const value = uiConfig?.flags?.[flag];
+    if (typeof value === 'boolean') {
+        return value;
+    } else if (typeof value !== 'undefined') {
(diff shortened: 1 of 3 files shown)


src/lib/types/experimental.ts
```diff
@@ -310,17 +310,10 @@ const flags: IFlags = {
...
-        ),
-        payload: {
-            type: PayloadType.JSON,
-            value: '14',
-        },
-    },
+    uiGlobalFontSize: parseEnvVarBoolean(
+        process.env.EXPERIMENTAL_UI_GLOBAL_FONT_SIZE_NAME,
+        false,
+    ),
```
thomasheartman, line 316: Didn't you do this already in https://github.com/Unleash/unleash/pull/9290. Is your branch up to date?

frontend/src/hooks/useUiFlag.ts
```diff
@@ -2,8 +2,14 @@ import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
...
     const { uiConfig } = useUiConfig();
+    const value = uiConfig?.flags?.[flag];
+    if (typeof value === 'boolean') {
+        return value;
+    } else if (typeof value !== 'undefined') {
+        console.error(`Flag ${flag} does not return a boolean.`);
+    }
 
-    return uiConfig?.flags?.[flag] || false;
+    return false;
```
thomasheartman, line 14: While I think this is fine in theory, I'm not confident about changing it now because we have places where we depend on the contents. Yeh, ts found the one in feedback component, but I don't really trust it to monitor everywhere. I don't see what we gain by changing this function right now. I'd rather look at other options too before changing this one.

PR #9294, opened 2025-02-11
feat: connection count usage

kwasniew:
## About the changes

Network data usage may show connections instead of raw request count behind a flag. 
Screenshot 2025-02-11 at 15 37 06

This PR adjusts chart data to visualize connections count (1 connection = 7200 requests per day) in a day or in a month

### DIFF
--- a/frontend/src/component/admin/network/NetworkTrafficUsage/chart-functions.ts
+++ b/frontend/src/component/admin/network/NetworkTrafficUsage/chart-functions.ts
@@ -6,6 +6,8 @@ import {
     addMonths,
     differenceInCalendarDays,
     differenceInCalendarMonths,
+    getDaysInMonth,
+    parseISO,
 } from 'date-fns';
 import { formatDay, formatMonth } from './dates';
 import type { ChartDataSelection } from './chart-data-selection';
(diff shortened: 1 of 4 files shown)


frontend/src/component/admin/network/NetworkTrafficUsage/chart-functions.ts
```diff
@@ -40,6 +42,46 @@ export const toChartData = (
     return { datasets, labels };
 };
 
+export const toConnectionChartData = (
```
kwasniew, line 45: this is a copy of existing method with minor modifications

frontend/src/component/admin/network/NetworkTrafficUsage/chart-functions.ts
```diff
@@ -40,6 +42,46 @@ export const toChartData = (
...
+                endpointsInfo[item1.apiPath].order -
+                endpointsInfo[item2.apiPath].order,
+        )
+        .map((item) => {
+            const record = newRecord();
+            for (const dataPoint of Object.values(item.dataPoints)) {
+                const date = parseISO(dataPoint.period);
+                const requestCount = dataPoint.trafficTypes[0].count;
+
+                if (dataPoint.period.length === 7) {
```
kwasniew, line 61: if YYYY-MM then divide requests by 1 day connection (7200 req) and number of days in a month

thomasheartman:
```suggestion
                if (traffic.grouping === 'monthly') {
```

thomasheartman (approved):
I don't have the context, but by and large this looks sane to me. Added a suggestion and a comment to have a look at, but otherwise fine.

thomasheartman:
Also, so this only affects visualisations, right? It's not supposed to touch the aggregation metrics in the boxes etc? Because if so, you probably want to do it earlier on.

PR #9295, opened 2025-02-11
chore: add migration that backfills scim user email hashes

daveleek:
A migration that sets already added SCIM users' email_hash


sjaanus (approved):

src/migrations/20250211114205-scim-email-hash-backfill.js
```diff
@@ -0,0 +1,14 @@
+exports.up = (db, cb) => {
+    db.runSql(`
+        UPDATE users
+        SET email_hash = md5(email::text)
+        WHERE scim_external_id IS NOT NULL;
```
sighphyre, line 5: Why `scim_external_id` and not `scim_id`? Providers don't need to set `scim_external_id` so there's cases where it's null. `scim_id` should always be set for a scim user or a user where scim has taken over

daveleek: Nice one, changed it!

sighphyre (approved):
LGTM

PR #9302, opened 2025-02-12
feat: new flag header

thomasheartman:
Initial spike to add the new design for the flag header. This PR uses the flag and adds a new header if it's active.

As far as I can tell, all elements of the new header are present in this design.

image

### Narrow windows

Handle narrow window widths. There is no handling of narrow windows here and the entire thing is set as nowrap, so it ends up with the actions hiding the tabs:
image

However, we also don't handle this very well today:
image

That said, I think now woud be a good time to handle this.

### Icon buttons with tooltips

Todays buttons for copying the flag, archiving it, etc, use `PermissionIconButton`. We want the same behavior, but without behavior checks. As far as I can tell, that's not a component we have today? (although I didn't look very hard). In this first instance, I've made a very simple version that's based on the permission icon button. Before we close off the header work, I'd like to make these two different button classes based of the same design. But I'd rather address it in a followup.

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureView.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureView.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { type PropsWithChildren, useState, type FC } from 'react';
 import {
     IconButton,
     styled,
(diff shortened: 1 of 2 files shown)


frontend/src/component/feature/FeatureView/FeatureView.tsx
```diff
@@ -231,144 +278,228 @@ export const FeatureView = () => {
...
-                                        isFeatureNameCopied
-                                            ? 'Copied!'
-                                            : 'Copy name'
-                                    }
-                                    arrow
-                                >
-                                    <IconButton
-                                        onClick={handleCopyToClipboard}
-                                        style={{ marginLeft: 8 }}
+            {flagOverviewRedesign ? (
```
Tymek, line 281: Is the header coupled with a lot of other things on the page? Maybe it makes sense to refactor it into a separate file?

thomasheartman: It's definitely couple to a few fair few things (which is why I did it inline initially instead of as a standalone function). Happy to take a look at it in a follow-up, though.

Tymek (approved):

PR #9303, opened 2025-02-12
fix: mutating delta events

FredrikOseberg:
Fix a bug where delta events would be mutated due to Object.assign.

### DIFF
--- a/src/lib/features/client-feature-toggles/delta/delta-cache.test.ts
+++ b/src/lib/features/client-feature-toggles/delta/delta-cache.test.ts
@@ -184,7 +184,6 @@ describe('RevisionCache', () => {
     });
 
     it('should not mutate previous feature-updated events when new events with the same feature name are added', () => {
-        // Arrange: Create a minimal base hydration event.
         const baseEvent: DeltaHydrationEvent = {
             eventId: 1,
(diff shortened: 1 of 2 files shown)


kwasniew (approved):

src/lib/features/client-feature-toggles/delta/delta-cache.test.ts
```diff
@@ -182,4 +182,63 @@ describe('RevisionCache', () => {
...
+                name: 'streaming-test',
+                description: null,
+                project: 'streaming-deltas',
+                stale: false,
+                type: 'release',
+                variants: [],
+                strategies: [],
+            },
+        };
+        deltaCache.addEvents([JSON.parse(JSON.stringify(initialFeatureEvent))]);
```
kwasniew, line 223: is it important for this test? if so I'd explain why

FredrikOseberg: Yeah.. in order for the test to fail when the mutation was happening the input needs to be cloned otherwise the reference is kept to the original object in this test and the check doesn't fail because the in place mutation also applies to the referenced object. I'll add a comment.

PR #9308, opened 2025-02-13
feat: new flag info box

Tymek:
- updated spacing of elements
- modified header and "flag type" 
- added "collaborators"
- refactored tags

New design:
image

What's not done yet:
- dropdown for environments visibility

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/DependencyRow.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/DependencyRow.tsx
@@ -26,6 +26,7 @@ import {
 
 const StyledPermissionButton = styled(PermissionButton)(({ theme }) => ({
     fontSize: theme.fontSizes.smallBody,
+    lineHeight: theme.typography.body1.lineHeight,
 }));
 
 const useDeleteDependency = (project: string, featureId: string) => {
@@ -112,17 +113,19 @@ export const DependencyRow = ({ feature }: IDependencyRowProps) => {
                     <StyledMetaDataItemLabel>
                         Dependency:
                     </StyledMetaDataItemLabel>
-                    <StyledPermissionButton
-                        size='small'
-                        permission={UPDATE_FEATURE_DEPENDENCY}
-                        projectId={feature.project}
-                        variant='text'
-                        onClick={() => {
-                            setShowDependencyDialogue(true);
-                        }}
-                    >
-                        Add parent flag
-                    </StyledPermissionButton>
+                    <div>
+                        <StyledPermissionButton
+                            size='small'
+                            permission={UPDATE_FEATURE_DEPENDENCY}
+                            projectId={feature.project}
+                            variant='text'
+                            onClick={() => {
+                                setShowDependencyDialogue(true);
+                            }}
+                        >
+                            Add parent flag
+                        </StyledPermissionButton>
+                    </div>
                 </StyledMetaDataItem>
             ) : null}
             {hasParentDependency ? (
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/DependencyRow.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/DependencyRow.tsx
@@ -1,4 +1,3 @@
-import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 import { AddDependencyDialogue } from 'component/feature/Dependencies/AddDependencyDialogue';
 import type { IFeatureToggle } from 'interfaces/featureToggle';
 import { useState } from 'react';
@@ -26,11 +25,8 @@ import {
 } from './FeatureOverviewMetaData';
 
 const StyledPermissionButton = styled(PermissionButton)(({ theme }) => ({
-    '&&&': {
-        fontSize: theme.fontSizes.smallBody,
-        lineHeight: 1,
-        margin: 0,
-    },
+    fontSize: theme.fontSizes.smallBody,
+    alignSelf: 'auto',
 }));
 
 const useDeleteDependency = (project: string, featureId: string) => {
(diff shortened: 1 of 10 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/DependencyRow.tsx
```diff
@@ -112,115 +108,82 @@ export const DependencyRow = ({ feature }: IDependencyRowProps) => {
...
+                </StyledMetaDataItem>
+            ) : null}
+            {hasParentDependency && !feature.dependencies[0]?.enabled ? (
+                <StyledMetaDataItem>
+                    <StyledMetaDataItemLabel>
+                        Dependency value:
+                    </StyledMetaDataItemLabel>
+                    <span>disabled</span>
+                </StyledMetaDataItem>
+            ) : null}
```
thomasheartman, line 157: Not very important, but can we use `&&` instead of ternaries in this file to get rid of all the `: null`? It's kinda distracting when there's this many of them,  I think.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/FeatureOverviewMetaData.tsx
```diff
@@ -97,55 +79,46 @@ const FeatureOverviewMetaData = () => {
...
+                </div>
                 <StyledBody>
                     <StyledMetaDataItem>
                         <StyledMetaDataItemLabel>
-                            Project:
+                            Flag type:
                         </StyledMetaDataItemLabel>
                         <StyledMetaDataItemText data-loading>
-                            {project}
+                            {capitalizeFirst(type || '')} flag
```
thomasheartman, line 101: This can also be done by CSS `::first-letter: { text-transform: uppercase }` which will also capitalize "flag" if there is no type 💁🏼 Not important, but this feels more like a CSS thing to me 🤷🏼

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/FeatureOverviewMetaData.tsx
```diff
@@ -97,55 +79,46 @@ const FeatureOverviewMetaData = () => {
...
+                    <StyledTitle>Flag details</StyledTitle>
+                    {description ? (
                         <StyledMetaDataItem data-loading>
                             <StyledMetaDataItemText>
                                 {description}
                             </StyledMetaDataItemText>
                         </StyledMetaDataItem>
-                    }
-                />
+                    ) : null}
```
thomasheartman, line 93: Same question regarding ternaries vs double ampersands. Wouldn't `&&` be easier? But did we say not to use double ampersands?

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/TagRow.tsx
```diff
@@ -99,80 +96,77 @@ export const TagRow = ({ feature }: IFeatureOverviewSidePanelTagsProps) => {
...
+        <StyledAddTagButton
+            size='small'
+            permission={UPDATE_FEATURE}
+            projectId={feature.project}
+            variant='text'
+            onClick={() => {
+                setManageTagsOpen(true);
+            }}
+        >
+            <StyledAddIcon /> Add tag
```
thomasheartman, line 109: Isn't there typically a prop to add icons to buttons? 
Should this be something like:

```suggestion
            }}
            startIcon={<StyledAddIcon />}
        >
            Add tag
```

I know the permission button is a custom component, but I think it inherits most of the button props. And if it doesn't inherit this one, maybe it should? Not really important, though, as MUI icons typically get all the needed stuff to be hidden for screen readers etc anyway.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/TagRow.tsx
```diff
@@ -99,80 +96,77 @@ export const TagRow = ({ feature }: IFeatureOverviewSidePanelTagsProps) => {
         }
     };
 
+    const addTagButton = (
```
thomasheartman, line 99: Should this be a function component and rendered as `<AddTagButton/>` instead?

I thought of the ["Never call function component functions directly" react docs](https://react.dev/reference/rules/react-calls-components-and-hooks#never-call-component-functions-directly), but this is a variable, so I'm not sure whether it applies here or not 🤔

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/TagRow.tsx
```diff
@@ -99,80 +96,77 @@ export const TagRow = ({ feature }: IFeatureOverviewSidePanelTagsProps) => {
...
-                                                    <ClearIcon />
-                                                </Tooltip>
+            {!tags.length ? (
+                <StyledMetaDataItem>
+                    <StyledLabel>Tags:</StyledLabel>
+                    <StyledTagContainer>{addTagButton}</StyledTagContainer>
+                </StyledMetaDataItem>
+            ) : (
+                <StyledTagRow>
+                    <StyledLabel>Tags:</StyledLabel>
```
thomasheartman, line 122: Small point, but is there a reason why you swap out the metadata item container here? 

In fact, I'd expect these two branches to be exactly the same: Container -> Label + Tag Container -> Tags + Button. So if the list of tags is empty, then `tags.map` gives you no components back and it'll still work.

Does that check out?

thomasheartman (approved):
Nice one! Got a couple questions and comments throughout but nothing blocking, so I'm gonna give this a ✅  and you can address them either now or later (or never if you disagree).

Regarding the screenies in the PR description: are the tags supposed to wrap like they do in the final, light screeshot? It looks a little weird to me that they're suddenly left-aligned. I don't expect we have sketches for that, so we should probably check in with UX.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/DependencyRow.tsx
Tymek: We don't have reliable way of catching issues related to `&&` operator. https://docs.getunleash.io/contributing/ADRs/front-end/jsx-conditionals#pitfalls-of--operator

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/TagRow.tsx
Tymek: This doesn't apply for things rendered into a variable. But thanks for pointing it out. I refactored it to a separate component

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/TagRow.tsx
Tymek: It didn't align tags properly, because we have 2 cases here: if tags fit in one line, keep it this way (aligned to the right), but if it doesn't fit - align to the left in starting from next line.
I'll spend 5min checking if I can make it work.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/FeatureOverviewMetaData.tsx
Tymek: I'm using what we already have, plus I did not have good time debugging and finding what's causing transform with `::` pseudo-classes in the past.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/DependencyRow.tsx
```diff
@@ -26,11 +25,8 @@ import {
 } from './FeatureOverviewMetaData';
 
 const StyledPermissionButton = styled(PermissionButton)(({ theme }) => ({
-    '&&&': {
-        fontSize: theme.fontSizes.smallBody,
-        lineHeight: 1,
-        margin: 0,
-    },
+    fontSize: theme.fontSizes.smallBody,
+    alignSelf: 'auto',
```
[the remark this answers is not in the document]
Tymek: Turns out this was not needed - leftover from previous tests

PR #9309, opened 2025-02-13
chore: add change request events to teams integration

nunogois:
https://linear.app/unleash/issue/2-3265/investigate-teams-integration-for-change-request-notifications

Adds change request events to the Teams integration.


daveleek (approved):
Cool, yup that should do it

src/lib/addons/teams.ts
```diff
@@ -23,6 +41,26 @@ export default class TeamsAddon extends Addon {
     flagResolver: IFlagResolver;
 
     constructor(args: IAddonConfig) {
+        if (args.flagResolver.isEnabled('teamsIntegrationChangeRequests')) {
+            teamsDefinition.events = [
```
daveleek, line 45: This is the only place the teamsDefinition is used yeah?

daveleek (approved):
Quick question, but apart from that 👍

src/lib/addons/teams.ts
nunogois: Just double checked: Yes.

PR #9314, opened 2025-02-14
fix: add a metric to track client registrations

ivarconr:
Adding a counter to track every time a client registers with Unleash.

### DIFF
--- a/src/lib/features/metrics/instance/instance-service.ts
+++ b/src/lib/features/metrics/instance/instance-service.ts
@@ -1,3 +1,4 @@
+import type EventEmitter from 'events';
 import { APPLICATION_CREATED, CLIENT_REGISTER } from '../../../types/events';
 import type { IApplication, IApplicationOverview } from './models';
 import type { IUnleashStores } from '../../../types/stores';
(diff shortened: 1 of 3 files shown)


src/lib/metrics.ts
```diff
@@ -807,6 +812,9 @@ export function registerPrometheusMetrics(
         clientDeltaMemory.reset();
         clientDeltaMemory.set(event.memory);
     });
+    eventBus.on(events.CLIENT_REGISTERED, ({ appName, environment }) => {
```
kwasniew, line 815: I'm assuming we don't need to batch those before calling inc

src/lib/features/metrics/instance/instance-service.ts
```diff
@@ -105,6 +111,7 @@ export default class ClientInstanceService {
         value.clientIp = clientIp;
         value.createdBy = SYSTEM_USER.username!;
         this.seenClients[this.clientKey(value)] = value;
+        this.eventBus.emit(CLIENT_REGISTERED, value);
```
kwasniew, line 114: it's slightly confusing the we have CLIENT_REGISTER and CLIENT_REGISTERED events

kwasniew (approved):

src/lib/metrics.ts
ivarconr: no need, its just in-memory counters.

PR #9315, opened 2025-02-17
fix: delete project segment with change request

kwasniew:
## About the changes

Fixing a bug where non admin users couldn't delete empty project segments in projects with change requests. The request was going directly to delete segment API instead of adding a change to the "shopping cart"

When deleting an empty project segment with change requests enabled you get add to draft option instead of instant delete:
Screenshot 2025-02-17 at 08 35 18

Then the toast message is changed according to change request naming convention:
Screenshot 2025-02-17 at 08 35 23

We already had delete segment UI in the "shopping cart" but I fixed the crossed out name when the change request is applied to read the data from the snapshot and not current strategy:
Screenshot 2025-02-17 at 08 36 16

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/SegmentChangeDetails.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/SegmentChangeDetails.tsx
@@ -59,6 +59,10 @@ export const SegmentChangeDetails: FC<{
 }> = ({ actions, change, changeRequestState }) => {
     const { segment: currentSegment } = useSegment(change.payload.id);
     const snapshotSegment = change.payload.snapshot;
+    const previousName =
+        changeRequestState === 'Applied'
+            ? change.payload?.snapshot?.name
+            : currentSegment?.name;
     const referenceSegment =
         changeRequestState === 'Applied' ? snapshotSegment : currentSegment;
 
@@ -74,10 +78,13 @@ export const SegmentChangeDetails: FC<{
                         >
                             - Deleting segment:
                         </Typography>
-                        <SegmentTooltipLink change={change}>
+                        <SegmentTooltipLink
+                            name={change.payload.name}
+                            previousName={previousName}
+                        >
                             <SegmentDiff
                                 change={change}
-                                currentSegment={currentSegment}
+                                currentSegment={referenceSegment}
                             />
                         </SegmentTooltipLink>
                     </ChangeItemInfo>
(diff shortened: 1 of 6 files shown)


frontend/src/component/segments/SegmentDelete/SegmentDeleteConfirm/SegmentDeleteConfirm.tsx
```diff
@@ -37,7 +39,7 @@ export const SegmentDeleteConfirm = ({
         <Dialogue
             title='Are you sure you want to delete this segment?'
             open={open}
-            primaryButtonText='Delete segment'
+            primaryButtonText={title}
```
kwasniew, line 42: had to be dynamic to accomodate CR and regular mode

frontend/src/component/changeRequest/ChangeRequest/SegmentTooltipLink.tsx
```diff
@@ -42,8 +42,9 @@ export const SegmentDiff: FC<{
     );
 };
 interface IStrategyTooltipLinkProps {
-    change: IChangeRequestUpdateSegment | IChangeRequestDeleteSegment;
```
kwasniew, line 45: we were only reading current name from the change so removing [stamp coupling](https://wiki.c2.com/?StampCoupling) to the entire change

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/SegmentChangeDetails.tsx
```diff
@@ -97,7 +104,7 @@ export const SegmentChangeDetails: FC<{
                     <ChangeItemCreateEditWrapper>
                         <ChangeItemInfo>
                             <Typography>Editing segment:</Typography>
-                            <SegmentTooltipLink change={change}>
+                            <SegmentTooltipLink name={change.payload.name}>
```
kwasniew, line 107: removing stamp coupling

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/SegmentChangeDetails.tsx
```diff
@@ -74,10 +78,13 @@ export const SegmentChangeDetails: FC<{
                         >
                             - Deleting segment:
                         </Typography>
-                        <SegmentTooltipLink change={change}>
+                        <SegmentTooltipLink
+                            name={change.payload.name}
```
kwasniew, line 82: removing stamp coupling

frontend/src/component/changeRequest/ChangeRequest/SegmentTooltipLink.tsx
thomasheartman: Not sure I got much from the article you linked (it felt like it was arguing with itself too much), but sure, I don't mind this change.

thomasheartman (approved):

PR #9317, opened 2025-02-17
fix(1-3377):  handle singular counts in project status lifecycle boxes

thomasheartman:
If the average number of days in a stage is 1, use `1 day` instead of
`1 days`.

Likewise, if your total number of archived flags is 1, use `1 flag archived` instead of `1 flags archived`.

I grepped through the file, but couldn't find any other hardcoded instances of "flags" or "days", so I think this is everything.

### DIFF
--- a/frontend/src/component/project/Project/ProjectStatus/ProjectLifecycleSummary.tsx
+++ b/frontend/src/component/project/Project/ProjectStatus/ProjectLifecycleSummary.tsx
@@ -167,10 +167,10 @@ export const ProjectLifecycleSummary = () => {
         );
     };
 
-    const totalArchivedStat = () => {
-        const count = data?.lifecycleSummary.archived.last30Days ?? 0;
-        return `${count} ${count === 1 ? 'flag' : 'flags'} archived`;
-    };
+    const archivedLast30DaysCount =
+        data?.lifecycleSummary.archived.last30Days ?? 0;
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/Project/ProjectStatus/ProjectLifecycleSummary.tsx
```diff
@@ -165,6 +167,10 @@ export const ProjectLifecycleSummary = () => {
         );
     };
 
+    const totalArchivedStat = () => {
+        const count = data?.lifecycleSummary.archived.last30Days ?? 0;
+        return `${count} ${count === 1 ? 'flag' : 'flags'} archived`;
+    };
```
Tymek, line 173: this can be a variable, not a function

Tymek (approved):

frontend/src/component/project/Project/ProjectStatus/ProjectLifecycleSummary.tsx
thomasheartman: It *can* be, but I'm not sure what you get from it. doing it as a function gives you scope to name variables and separate it from other stuff. I don't mind, though 🤷🏼

Tymek: Scoping is a for of an abstraction - to me it makes this code less declarative, less "linear" to read if you need to understand everything that's going on, instead of trusting the function name

PR #9320, opened 2025-02-18
feat(1-3379): add stale badge to the new header

thomasheartman:
Without this, there's no way to tell if a flag is stale or not on the flag page, so we're adding it back in. However, we'll only show the badge if it's stale, not if it's an active flag.

To make the badge line up properly when it wraps, I also added a way to opt out of the hard-coded left margin in the badge.

### DIFF
--- a/frontend/src/component/common/FeatureStatusChip/FeatureStatusChip.tsx
+++ b/frontend/src/component/common/FeatureStatusChip/FeatureStatusChip.tsx
@@ -3,11 +3,13 @@ import { Badge } from 'component/common/Badge/Badge';
 interface IStatusChip {
     stale: boolean;
     showActive?: boolean;
+    withLeftMargin?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


frontend/src/component/common/FeatureStatusChip/FeatureStatusChip.tsx
```diff
@@ -19,7 +21,7 @@ export const FeatureStatusChip = ({
     const value = stale ? 'Stale' : 'Active';
 
     return (
-        <div data-loading style={{ marginLeft: '8px' }}>
+        <div data-loading style={{ marginLeft: withLeftMargin ? '8px' : 0 }}>
```
Tymek, line 24: You can change 8px -> spacing(1)

`<Box sx={theme => ({ marginLeft: theme.spacing(1) })}>`

Tymek (approved):

PR #9321, opened 2025-02-18
chore(1-3380): handle narrow windows for the flag header.

thomasheartman:
Makes it so that the actions/tabs wrap on narrow width screens.

Constraints:
- When wrapping, the actions should go **before** the tabs, when not wrapping, they should be placed **after**
- Need to maintain a logical tab order for wrapping, so just using `flex-flow: row wrap-reverse` doesn't work because the tab order will be wrong
- When the elements switch, you shouldn't lose your tab place in the document

This solution uses container queries to determine the container size and uses that to set the wrapping. Falls back to media queries if container queries aren't supported (it's supported on >93% of browsers according to caniuse).

The wrapping points don't use predefined media queries because:
- containers don't care about the size of the screen. It's the intrinsic size of the container that matters.
- wrapping at 900px seemed too far out if container queries are unsupported. But it's a fallback, so we can switch to that if we want.

If your keyboard focus is on one of the actions on a wide screen, and the screen goes narrow, your focus will still be after the tabs (staying consistent), so tabbing to the next element will take you into the flag details, while backtab takes you back to the tabs.

Before wrapping:
image

After wrapping:
image

## A note on accessibility:

The spec for flexbox (taken from [MDN's doc](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items)) states:

>  "Authors must not use order or the *-reverse values of [flex-flow](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-flow)/flex-direction as a substitute for correct source ordering, as that can ruin the accessibility of the document."

So even if wrap-reverse works visually, it's not a good solution for this.

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureViewHeader.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureViewHeader.tsx
@@ -266,8 +266,6 @@ type Props = {
     feature: IFeatureToggle;
 };
 
-// todo: make the breakpoint 700px
-
 export const FeatureViewHeader: FC<Props> = ({ feature }) => {
     const projectId = useRequiredPathParam('projectId');
     const featureId = useRequiredPathParam('featureId');
--- a/frontend/src/component/feature/FeatureView/FeatureViewHeader.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureViewHeader.tsx
@@ -178,6 +178,8 @@ type Props = {
     feature: IFeatureToggle;
 };
 
+// todo: make the breakpoint 700px
+
 export const FeatureViewHeader: FC<Props> = ({ feature }) => {
     const projectId = useRequiredPathParam('projectId');
     const featureId = useRequiredPathParam('featureId');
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureViewHeader.tsx
```diff
@@ -174,10 +189,85 @@ const useLegacyVariants = (environments: IFeatureToggle['environments']) => {
...
+    feature: IFeatureToggle;
+    showOnNarrowScreens?: boolean;
+    onFavorite: () => void;
+    handleCopyToClipboard: () => void;
+    isFeatureNameCopied: boolean;
+    openStaleDialog: () => void;
+    openDeleteDialog: () => void;
+};
+
+const HeaderActionsComponent = ({
```
thomasheartman, line 202: Moves the action into a separate component so that they can be rendered multiple times

frontend/src/component/feature/FeatureView/FeatureViewHeader.tsx
```diff
@@ -49,8 +49,16 @@ const NewStyledHeader = styled('div')(({ theme }) => ({
...
     borderBottom: `1px solid ${theme.palette.divider}`,
+    containerType: 'inline-size',
 }));
 
+const onNarrowHeader = (css: object) => ({
+    '@container (max-width: 650px)': css,
+    '@supports not (container-type: inline-size)': {
+        '@media (max-width: 700px)': css,
+    },
+});
```
thomasheartman, line 60: Use container queries to decide when to stack actions over tabs.

Doesn't use the predefined breakpoints because:
- for container queries: It doesn't make sense for containers to use predefined screen widths
- for the media query: 650px ('sm') is too small (causes wrapping or overlap) and 900px ('md') feel too wide a place to start wrapping

Tymek (approved):

PR #9322, opened 2025-02-18
feat: project field to feature removed event

sjaanus:
Start sending out project for feature deleted, as we need it for edge to keep exact copy.

### DIFF
--- a/src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
+++ b/src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
@@ -205,9 +205,9 @@ export class ClientFeatureToggleDelta extends EventEmitter {
                 return undefined;
             }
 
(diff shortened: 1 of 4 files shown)


src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts
```diff
@@ -205,17 +205,9 @@ export class ClientFeatureToggleDelta extends EventEmitter {
...
-                    if (event.type === 'feature-removed') {
-                        const { project, ...rest } = event;
-                        return rest;
-                    }
-                    return event;
-                }),
-            };
-
-            return Promise.resolve(response);
+            return Promise.resolve({
```
kwasniew, line 208: you probably don't need the promise wrapper

kwasniew (approved):

PR #9323, opened 2025-02-18
feat: show and hide environments

Tymek:
## About the changes
- Button to show and hide environments
- Refactored hook storing state of hidden environments
- Changed the way flag is triggered for feature overview
- Visual updates for new page look

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
@@ -1,0 +1,70 @@
+import { Button, Checkbox, Menu, MenuItem, styled } from '@mui/material';
+import { useState, type FC } from 'react';
+
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
+import ExpandLessIcon from '@mui/icons-material/ExpandLess';
+import type { FeatureEnvironmentSchema } from 'openapi';
+
+type EnvironmentVisibilityMenuProps = {
+    environments: Pick<FeatureEnvironmentSchema, 'name'>[];
+    hiddenEnvironments: FeatureEnvironmentSchema['name'][];
+    onChange: (name: FeatureEnvironmentSchema['name']) => void;
+};
+
+const buttonId = 'environment-visibility-button';
+const menuId = 'environment-visibility-menu';
+
+const StyledContainer = styled('div')(({ theme }) => ({
+    display: 'flex',
+    justifyContent: 'center',
+    paddingTop: theme.spacing(4),
+}));
+
+export const EnvironmentVisibilityMenu: FC<EnvironmentVisibilityMenuProps> = ({
+    environments,
+    hiddenEnvironments,
+    onChange,
+}) => {
+    const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+    const isOpen = Boolean(anchorEl);
+    const handleOpen = (event: React.MouseEvent<HTMLButtonElement>) => {
+        setAnchorEl(event.currentTarget);
+    };
+    const handleClose = () => {
+        setAnchorEl(null);
+    };
+
+    return (
+        <StyledContainer>
+            <Button
+                onClick={handleOpen}
+                endIcon={isOpen ? <ExpandLessIcon /> : <ExpandMoreIcon />}
+                variant='outlined'
+                id={buttonId}
+                aria-controls={isOpen ? menuId : undefined}
+                aria-haspopup='true'
+                aria-expanded={isOpen ? 'true' : undefined}
+                data-loading
+            >
+                Hide/show environments
+            </Button>
+            <Menu
+                id={menuId}
+                anchorEl={anchorEl}
+                open={isOpen}
+                onClose={handleClose}
+                MenuListProps={{ 'aria-labelledby': buttonId }}
+            >
+                {environments.map(({ name }) => (
+                    <MenuItem key={name} onClick={() => onChange(name)}>
+                        <Checkbox
+                            onChange={() => onChange(name)}
+                            checked={!hiddenEnvironments?.includes(name)}
+                        />
+                        {name}
+                    </MenuItem>
+                ))}
+            </Menu>
+        </StyledContainer>
+    );
+};
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
@@ -3,12 +3,11 @@ import { useState, type FC } from 'react';
 
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import ExpandLessIcon from '@mui/icons-material/ExpandLess';
-import type { FeatureEnvironmentSchema } from 'openapi';
 
 type EnvironmentVisibilityMenuProps = {
-    environments: Pick<FeatureEnvironmentSchema, 'name'>[];
-    hiddenEnvironments: FeatureEnvironmentSchema['name'][];
-    onChange: (name: FeatureEnvironmentSchema['name']) => void;
+    environments: Array<{ name: string }>;
+    hiddenEnvironments: string[];
+    onChange: (name: string) => void;
 };
 
 const buttonId = 'environment-visibility-button';
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/hooks/useEnvironmentVisibility.ts
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/hooks/useEnvironmentVisibility.ts
@@ -1,0 +1,39 @@
+import { useLocalStorageState } from 'hooks/useLocalStorageState';
+import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
+import { createLocalStorage } from 'utils/createLocalStorage';
+
+// Reading legacy value be safely refactored out in a next version - related to `flagOverviewRedesign` flag
+const { value: legacyStoreValue } = createLocalStorage<{
+    hiddenEnvironments?: Array<string>;
+}>('global:v1', {});
+
+export const useEnvironmentVisibility = () => {
+    const [value, setValue] = useLocalStorageState<Array<string>>(
+        'environment-visibiilty',
+        legacyStoreValue?.hiddenEnvironments || [],
+    );
+    const { trackEvent } = usePlausibleTracker();
+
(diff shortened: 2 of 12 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/LegacyFeatureOverview.tsx
Tymek, line 1: Copy for keeping a new version behind a flag

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
```diff
@@ -0,0 +1,70 @@
...
+                variant='outlined'
+                id={buttonId}
+                aria-controls={isOpen ? menuId : undefined}
+                aria-haspopup='true'
+                aria-expanded={isOpen ? 'true' : undefined}
+                data-loading
+            >
+                Hide/show environments
+            </Button>
+            <Menu
```
thomasheartman, line 51: I'm not very familiar with the Menu component, so this may be a stupid question, but I just looked it up on the docs: do they autoclose on every selection? I would assume that you might want to show/hide multiple environments in one go with this, but maybe the auto-close on selection is an opt-in thing?

And out of my curiosity: is there a reason why you picked Menu over, say, Select?

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
```diff
@@ -0,0 +1,70 @@
...
+                open={isOpen}
+                onClose={handleClose}
+                MenuListProps={{ 'aria-labelledby': buttonId }}
+            >
+                {environments.map(({ name }) => (
+                    <MenuItem key={name} onClick={() => onChange(name)}>
+                        <Checkbox
+                            onChange={() => onChange(name)}
+                            checked={!hiddenEnvironments?.includes(name)}
+                        />
```
thomasheartman, line 63: Again: I'm not sure how Menu works, but there's a couple things here that we might wanna change:

All form controls (that includes checkboxes), should have labels (according to [MUI's accessibility notes for this component](https://mui.com/material-ui/react-checkbox/#accessibility)). In this case, you'd probably want to label it with the name of the env (so probably wrap the name of the env in a label and mark it as being for this checkbox). 

But I'm not even sure we should have the checkbox here. How does it work with keyboard focus? Does it give you multiple tab stops within it? It looks like it's designed to be navigated by arrow keys and that tab takes you to the next item in the document flow, but I'm not sure whether adding an input field here messes with that or not. Regardless, I can imagine it being confusing for screen readers. 

But based on how MUI uses the menu in their docs, maybe it's better to just use an icon instead of a form control? 💁🏼

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/hooks/useEnvironmentVisibility.ts
```diff
@@ -0,0 +1,39 @@
+import { useLocalStorageState } from 'hooks/useLocalStorageState';
+import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
+import { createLocalStorage } from 'utils/createLocalStorage';
+
+// Reading legacy value be safely refactored out in a next version - related to `flagOverviewRedesign` flag
```
thomasheartman, line 5: Is this missing a word? I'm not sure I understand the comment (but maybe that's not important)

```suggestion
// Reading legacy value will be safely refactored out in a next version - related to `flagOverviewRedesign` flag
```

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/FeatureOverviewMetaData.tsx
```diff
@@ -156,6 +166,13 @@ const FeatureOverviewMetaData = () => {
                         <DependencyRow feature={feature} />
                     ) : null}
                     <TagRow feature={feature} />
+                    {onEnvironmentVisibilityChange ? (
+                        <EnvironmentVisibilityMenu
+                            environments={feature.environments || []}
+                            hiddenEnvironments={hiddenEnvironments || []}
+                            onChange={onEnvironmentVisibilityChange}
+                        />
+                    ) : null}
```
thomasheartman, line 175: Why is the onEnvironmentVisibilityChange function optional? Are there cases where we don't want you to do that?

frontend/src/component/feature/FeatureView/FeatureOverview/NewFeatureOverviewEnvironment/FeatureOverviewEnvironmentBody.tsx
thomasheartman, line 1: Are the changes in this file just conditionally render -> ternary or should I be looking for something else?

frontend/src/component/feature/FeatureView/FeatureOverview/NewFeatureOverviewEnvironment/NewFeatureOverviewEnvironment.tsx
```diff
@@ -52,62 +52,75 @@ const StyledHeaderTitle = styled('span')(({ theme }) => ({
...
-        <StyledFeatureOverviewEnvironment>
-            <StyledHeader data-loading>
-                <StyledHeaderToggleContainer>
-                    <FeatureOverviewEnvironmentToggle
-                        environment={featureEnvironment}
+    return environments.map(({ name: environmentId }) => {
+        const featureMetrics = getFeatureMetrics(
+            feature?.environments,
+            metrics,
+        );
```
thomasheartman, line 83: Probably not super important, but can we move this function call to outside the loop? It doesn't appear to be dependent on the environment id, so the response should be the same for each environment. I'm guessing it doesn't do any very computation, but in case it fires off some requests or does something silly 💁🏼

frontend/src/component/feature/FeatureView/FeatureOverview/NewFeatureOverviewEnvironment/NewFeatureOverviewEnvironment.tsx
```diff
@@ -52,62 +52,75 @@ const StyledHeaderTitle = styled('span')(({ theme }) => ({
 }));
 
 interface INewFeatureOverviewEnvironmentProps {
-    environmentId: string;
+    hiddenEnvironments: string[];
```
thomasheartman, line 55: This seems like a cool change. how are we going from one env to multiple? I couldn't work it out from the description. Presumably this gives us some new capabilities? what component is this (do you have a screenshot)?

frontend/src/hooks/useGlobalLocalStorage.ts
```diff
@@ -5,6 +5,9 @@ interface IGlobalStore {
     hiddenEnvironments?: Array<string>;
 }
 
+/**
+ * @deprecated use tested `useLocalStorageState` hook instead
+ */
```
thomasheartman, line 10: How many places do we use this today? I migrating off it going to be hard? Also, *why* are we deprecating it? I don't remember any conversation about that.

thomasheartman (approved):
Got a fair few comments and questions, but nothing blocking. Some things will have to be checked/addressed before we start rolling it out (a11y stuff, babyyyyy), but I'm happy to put it in behind the flag.

Nice work!

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
Tymek: Select displays it's value in the button. If menu should close on every select is up to our implementation - in this case it will not.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/FeatureOverviewMetaData.tsx
Tymek: For tests that. We have some "parent flag" tests that don't pass environments.

frontend/src/component/feature/FeatureView/FeatureOverview/NewFeatureOverviewEnvironment/FeatureOverviewEnvironmentBody.tsx
Tymek: I think that's it in this file. It's what Nuno initially sketched

frontend/src/component/feature/FeatureView/FeatureOverview/NewFeatureOverviewEnvironment/NewFeatureOverviewEnvironment.tsx
Tymek: I'll refactor it later

frontend/src/component/feature/FeatureView/FeatureOverview/NewFeatureOverviewEnvironment/NewFeatureOverviewEnvironment.tsx
Tymek: Entire right side (list of environments)

frontend/src/hooks/useGlobalLocalStorage.ts
Tymek: Just this place. I didn't find it anywhere else. We already got rid of this for favorites

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
thomasheartman: Sure; sounds sensible.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/FeatureOverviewMetaData.tsx
thomasheartman: Hmm. I'm not sure I want the tests to determine whether something should be required or not. And in this case, seems like we fall back to an empty list of environments anyway? Optionally, you could keep the function optional but give it a default impl that does nothing?

frontend/src/component/feature/FeatureView/FeatureOverview/NewFeatureOverviewEnvironment/NewFeatureOverviewEnvironment.tsx
thomasheartman: I ... don't quite understand what you're referring to 😅 But that's alright

frontend/src/hooks/useGlobalLocalStorage.ts
thomasheartman: Safe. Shouldn't be a problem, then. But I still don't understand why we're deprecating it? Does it do something we've moved away from? Does it not work?

PR #9325, opened 2025-02-18
chore: visualize connected edges

nunogois:
https://linear.app/unleash/issue/2-3233/visualize-connected-edge-instances

Adds a new tab in the Network page to visualize connected Edges.

This is behind a `edgeObservability` flag.

Also opens up the Network page even if you don't have a Prometheus API configured. When accessing the tabs that require it to set, and it isn't, we show some extra information about this and redirect you to the respective section in our docs.

image

image

image

### DIFF
--- a/frontend/src/component/admin/network/NetworkPrometheusAPIWarning.tsx
+++ b/frontend/src/component/admin/network/NetworkPrometheusAPIWarning.tsx
@@ -1,0 +1,24 @@
+import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
+
+export const NetworkPrometheusAPIWarning = () => {
+    const {
+        uiConfig: { networkViewEnabled },
+    } = useUiConfig();
+
+    if (networkViewEnabled) return null;
+
+    return (
+        <p>
+            This view requires the <strong>PROMETHEUS_API</strong> environment
+            variable to be set. Refer to our{' '}
+            <a
+                href='https://docs.getunleash.io/reference/network-view#data-source'
+                target='_blank'
+                rel='noreferrer'
+            >
+                documentation
+            </a>{' '}
+            for more information.
+        </p>
+    );
+};
--- a/frontend/src/hooks/api/getters/useConnectedEdges/useConnectedEdges.ts
+++ b/frontend/src/hooks/api/getters/useConnectedEdges/useConnectedEdges.ts
@@ -1,0 +1,35 @@
+import { useMemo } from 'react';
+import useUiConfig from '../useUiConfig/useUiConfig';
+import { formatApiPath } from 'utils/formatPath';
+import handleErrorResponses from '../httpErrorResponseHandler';
+import { useConditionalSWR } from '../useConditionalSWR/useConditionalSWR';
+import type { ConnectedEdge } from 'interfaces/connectedEdge';
+
+const DEFAULT_DATA: ConnectedEdge[] = [];
+
+export const useConnectedEdges = () => {
+    const { isEnterprise } = useUiConfig();
+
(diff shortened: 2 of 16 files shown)


frontend/src/component/admin/network/NetworkPrometheusAPIWarning.tsx
```diff
@@ -0,0 +1,24 @@
+import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
+
+export const NetworkPrometheusAPIWarning = () => {
+    const {
+        uiConfig: { networkViewEnabled },
```
nunogois, line 5: We should probably rename `networkViewEnabled` to `prometheusAPIAvailable` or something.

I can do it in a follow-up PR if you agree, however it may be a breaking change according to our OpenAPI schema since it renames an existing property: https://github.com/Unleash/unleash/blob/14b6b38238fe5d9c8997600d33347bab539fa046/src/lib/openapi/spec/ui-config-schema.ts#L102

gastonfournier: Yes, but at the same time we discussed couple of times that our ui specific API should not be as strict as other APIs, so I don't mind too much

frontend/src/component/admin/network/NetworkPrometheusAPIWarning.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+    return (
+        <p>
+            This view requires the <strong>PROMETHEUS_API</strong> environment
+            variable to be set. Refer to our{' '}
+            <a
+                href='https://docs.getunleash.io/reference/network-view#data-source'
+                target='_blank'
+                rel='noreferrer'
+            >
```
gastonfournier, line 18: Neat!

frontend/src/hooks/api/getters/useConnectedEdges/useConnectedEdges.ts
```diff
@@ -0,0 +1,39 @@
...
+import type { SWRConfiguration } from 'swr';
+import { useUiFlag } from 'hooks/useUiFlag';
+
+const DEFAULT_DATA: ConnectedEdge[] = [];
+
+export const useConnectedEdges = (options?: SWRConfiguration) => {
+    const { isEnterprise } = useUiConfig();
+    const edgeObservabilityEnabled = useUiFlag('edgeObservability');
+
+    const { data, error, mutate } = useConditionalSWR<ConnectedEdge[]>(
```
gastonfournier, line 16: Should we use `useEnterpriseSWR` instead? It already validates is enterprise, but maybe because of the flag is not easy to use

gastonfournier (approved):
LGTM! As always I didn't go into much detail of the UI as I prioritize iterating fast on it

frontend/src/hooks/api/getters/useConnectedEdges/useConnectedEdges.ts
nunogois: Unfortunately `useEnterpriseSWR` does not support a condition, so it wouldn't be practical with the flag, like you guessed. `useConditionalSWR` allows to specify the exact condition we want, which feels like the way to go in most cases. I would go as far as suggest we drop `useEnterpriseSWR` completely at some point.

frontend/src/component/admin/network/NetworkPrometheusAPIWarning.tsx
nunogois: Here you go: https://github.com/Unleash/unleash/pull/9330

PR #9327, opened 2025-02-19
fix: use bucket stop as timestamp for metrics bucket

ivarconr:
This fixes a problem where the yggdrasil-engine does not send the correct value for bucket.start. In practice clients sends metrics every 60s and it does not matter if we use start or the stop timestamp to resolve the nearest full hour.


sighphyre (approved):
LGTM!

PR #9328, opened 2025-02-19
chore: await so change requests banner shows up after adding release plans changes

daveleek:
Await adding changes and refetching change requests so that the banner can show up immediately after adding release plan changes


gastonfournier (approved):
LG!

PR #9330, opened 2025-02-19
refactor: rename networkViewEnabled to prometheusAPIAvailable

nunogois:
https://linear.app/unleash/issue/2-3290/rename-networkviewenabled-to-prometheusapiavailable

Follow-up to: https://github.com/Unleash/unleash/pull/9325

See: https://github.com/Unleash/unleash/pull/9325#discussion_r1960085303

Renames `networkViewEnabled` to `prometheusAPIAvailable`. This new name better reflects what it stands for.


gastonfournier (approved):
LG! To reiterate this API is for our UI only so it's ok being more flexible than with other APIs in terms of backward compatibility

PR #9335, opened 2025-02-20
chore: make Edge instance node width consistent

nunogois:
https://linear.app/unleash/issue/2-3295/make-edge-instance-nodes-width-consistent-in-the-connected-edges-view

When manually testing this feature I noticed that sometimes Edge instance nodes had different widths depending on their content.

This makes it so they stretch to the available space, making all instance nodes in a group (app-name) the same width.


gastonfournier (approved):
LG!

PR #9337, opened 2025-02-20
chore: upgrade vitest to 3.0.6

nunogois:
https://linear.app/unleash/issue/2-3296/upgrade-vitest

Follow-up to https://github.com/Unleash/unleash/pull/9212

The ☝️ PR was failing some tests because `chartjs-adapter-date-fns` is imported as a [side effect](https://stackoverflow.com/questions/41127479/es6-import-for-side-effects-meaning) and seems to be CommonJS-only. When Vitest loads it, it fails to find named exports (like `_adapters`). By inlining `chartjs-adapter-date-fns` in our Vitest config, esbuild transforms it into a proper ES module for testing.

Additionally, we’re upgrading from v1.4.0 to v3.0.6 directly since our code seems to be fully compatible.
>Vitest requires Vite >=v5.0.0 and Node >=v18.0.0

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -124,7 +124,7 @@
     "vite-plugin-env-compatible": "2.0.1",
     "vite-plugin-svgr": "3.3.0",
     "vite-tsconfig-paths": "4.3.2",
-    "vitest": "1.4.0",
+    "vitest": "3.0.6",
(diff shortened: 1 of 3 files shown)


gastonfournier (approved):
Nice work! Thanks!

PR #9340, opened 2025-02-20
feat: make env selector filterable

thomasheartman:
Makes the env selector on the flag page act the same way as the env selector on the new project page or any of the filterable buttons in the new project/flag dialogs. 

Also slightly changes the styles of the existing dropdown lists to bring them in line with the new env selector (more padding, full-width highlights).

Selector:

image

Project/flag creation:
Before:
image

After:

image

## Technical notes

I was a little unsure how best to share the padding/spacing styles between the search field and popover at first (as was requested by UX). The easiest way (and most compliant with how we do it today) was to define the spacing in a variable and move the relevant components into the same file.

However, I actually think that using a CSS variable (e.g. `--popover-spacing`) would be "better" here, but we don't really use them much, so I've left that out for now. That said, if you agree, I'd be more than happy to use that instead 🙋🏼

### DIFF
--- a/frontend/src/component/common/DialogFormTemplate/ConfigButtons/ConfigButton.styles.tsx
+++ b/frontend/src/component/common/DialogFormTemplate/ConfigButtons/ConfigButton.styles.tsx
@@ -1,13 +1,5 @@
 import { Popover, styled } from '@mui/material';
 
-export const StyledDropdown = styled('div')(({ theme }) => ({
-    paddingInline: 0,
-    display: 'flex',
-    flexDirection: 'column',
-    gap: theme.spacing(1),
-    maxHeight: '70vh',
-}));
-
 export const StyledPopover = styled(Popover)(({ theme }) => ({
     '& .MuiPaper-root': {
         borderRadius: `${theme.shape.borderRadiusMedium}px`,
--- a/frontend/src/component/common/DialogFormTemplate/ConfigButtons/ConfigButton.styles.tsx
+++ b/frontend/src/component/common/DialogFormTemplate/ConfigButtons/ConfigButton.styles.tsx
@@ -1,16 +1,4 @@
-import { Popover, styled } from '@mui/material';
-
-export const StyledPopover = styled(Popover)(({ theme }) => ({
-    '& .MuiPaper-root': {
-        borderRadius: `${theme.shape.borderRadiusMedium}px`,
-        paddingInline: 0,
-        paddingTop: theme.spacing(1.5),
-        display: 'flex',
-        flexDirection: 'column',
-        gap: theme.spacing(1),
-        maxHeight: '70vh',
-    },
-}));
+import { styled } from '@mui/material';
 
 export const ButtonLabel = styled('span', {
     shouldForwardProp: (prop) => prop !== 'labelWidth',
--- a/frontend/src/component/common/DialogFormTemplate/ConfigButtons/ConfigButton.styles.tsx
+++ b/frontend/src/component/common/DialogFormTemplate/ConfigButtons/ConfigButton.styles.tsx
@@ -1,7 +1,8 @@
 import { Popover, styled } from '@mui/material';
 
 export const StyledDropdown = styled('div')(({ theme }) => ({
-    padding: theme.spacing(2),
+    paddingInline: 0,
+    paddingTop: theme.spacing(1.5),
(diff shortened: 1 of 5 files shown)


frontend/src/component/common/DialogFormTemplate/ConfigButtons/DropdownList.styles.tsx
```diff
@@ -1,8 +1,10 @@
 import { Checkbox, ListItem, styled } from '@mui/material';
 
 export const StyledListItem = styled(ListItem)(({ theme }) => ({
-    paddingLeft: theme.spacing(1),
+    paddingLeft: theme.spacing(2),
+    paddingBlock: theme.spacing(1),
```
thomasheartman, line 5: These shouldn't be made here; this is just a placeholder while I work out the PR.

frontend/src/component/common/DialogFormTemplate/ConfigButtons/ConfigButton.styles.tsx
thomasheartman, line 1: Elements have either been moved or removed because they were unnecessary

frontend/src/component/common/DialogFormTemplate/ConfigButtons/shared.styles.tsx
```diff
@@ -12,11 +12,27 @@ const visuallyHiddenStyles = {
     whiteSpace: 'nowrap',
 };
 
+const padding = (theme: Theme) => theme.spacing(1.5);
+
+export const StyledPopover = styled(Popover)(({ theme }) => ({
+    '& .MuiPaper-root': {
+        borderRadius: `${theme.shape.borderRadiusMedium}px`,
+        paddingInline: 0,
+        paddingTop: padding(theme),
```
thomasheartman, line 21: This is actually a great use case for css variables, but I don't think we use them many places, so I've added it as a js const for now. I'd be very happy to use variables instead, though.

frontend/src/component/common/DialogFormTemplate/ConfigButtons/shared.styles.tsx
```diff
@@ -12,11 +12,27 @@ const visuallyHiddenStyles = {
...
+
+        gap: theme.spacing(1),
+        maxHeight: '70vh',
+    },
+}));
+
 export const StyledDropdownSearch = styled(TextField, {
     shouldForwardProp: (prop) => prop !== 'hideLabel',
 })<{ hideLabel?: boolean }>(({ theme, hideLabel }) => ({
+    paddingInline: padding(theme),
```
Tymek, line 33:
```suggestion
    paddingInline: theme.spacing(dropdownPadding),
```
Wrapping it in a function is an unnecessary misdirection. Every closure is another thing to keep on the "stack" in your head when reading this. Does this need to always match the '& .MuiInputBase-root'? Then maybe setting `const dropdownPadding = 1.5` is a solution? It's an alternative to a CSS variable.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewMetaData/EnvironmentVisibilityMenu/EnvironmentVisibilityMenu.tsx
```diff
@@ -33,37 +35,62 @@ export const EnvironmentVisibilityMenu: FC<EnvironmentVisibilityMenuProps> = ({
...
                 onClose={handleClose}
-                MenuListProps={{ 'aria-labelledby': buttonId }}
+                anchorOrigin={{
+                    vertical: 'bottom',
+                    horizontal: 'left',
+                }}
+                transformOrigin={{
+                    vertical: 'top',
+                    horizontal: 'left',
+                }}
```
Tymek, line 77: 👌🏻

Tymek (approved):

frontend/src/component/common/DialogFormTemplate/ConfigButtons/shared.styles.tsx
thomasheartman: good call! Addressed in d8e97fa08aed7280ee0bc450630d110c8c4b9d31. Of course, me being me, I still like CSS variables, but hey .. I can do that in my own time 😄

PR #9342, opened 2025-02-20
feat: new environment box

Tymek:
## About the changes
image

## Discussion points
- I decided not to show "metrics" box before metrics are available, for 2 reasons: 1. screen is less cluttered in the beginning, when user doesn't have any SDKs connected. 2. If for some reason user doesn't send metrics (it's optional), it's not needed here.


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+    expandable,
+    ...props
+}) => {
+    return (
+        <StyledAccordionSummary
+            {...props}
+            expandIcon={
+                expandable ? <ExpandMore titleAccess='Toggle' /> : <></>
+            }
+            sx={{ cursor: expandable ? 'pointer' : 'default !important' }}
```
Tymek, line 61: I'll maybe try to avoid `important` here

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
```diff
@@ -135,115 +75,70 @@ const FeatureOverviewEnvironment = ({
...
-                    </StyledAccordion>
-                </StyledFeatureOverviewEnvironment>
-            }
-        />
+        <StyledFeatureOverviewEnvironment>
+            <StyledAccordion
+                TransitionProps={{ mountOnEnter: true }}
+                data-testid={`${FEATURE_ENVIRONMENT_ACCORDION}_${env.name}`}
+                expanded={isOpen && hasActivations}
+                onChange={() => setIsOopen(isOpen ? !isOpen : hasActivations)}
```
thomasheartman, line 96: We should also double check the rendered component here. Again, according to the [MUI docs](https://mui.com/material-ui/react-accordion/#changing-heading-level):

> By default, the Accordion uses an h3 element for the heading. You can change the heading element using the slotProps.heading.component prop to ensure the correct heading hierarchy in your document.

I think h2 is right here (flag -> environment):

```tsx
<Accordion slotProps={{ heading: { component: 'h2' } }}>
```

But we should also make sure that there's no invalid components inside the h2, so checking the rendered hierarchy is probably smart.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const StyledHeader = styled('div')(({ theme }) => ({
+    display: 'flex',
+    width: '100%',
+    color: theme.palette.text.primary,
+    alignItems: 'center',
+    minHeight: theme.spacing(8),
+}));
+
+const StyledHeaderTitle = styled('div')(({ theme }) => ({
```
thomasheartman, line 26: If this is rendered inside a `h2`, then `div` isn't a valid element (but `span` is).

Anyway, this **might** (emphasis on might) be a good use case for [hgroup](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup).

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+                expandable ? <ExpandMore titleAccess='Toggle' /> : <></>
+            }
+            sx={{ cursor: expandable ? 'pointer' : 'default !important' }}
+        >
+            <StyledHeader data-loading>
+                <StyledHeaderTitle>
+                    <StyledHeaderTitleLabel>Environment</StyledHeaderTitleLabel>
+                    <StyledStringTruncator
+                        text={environmentId}
+                        maxWidth='100'
```
thomasheartman, line 68: Is this a percentage?

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/FeatureOverviewEnvironmentMetrics/FeatureOverviewEnvironmentMetrics.tsx
```diff
@@ -0,0 +1,90 @@
...
+    return (
+        <StyledContainer>
+            <StyledInfo>
+                <StyledPercentage>{percentage}%</StyledPercentage>
+                <StyledInfoParagraph>
+                    The flag has been requested{' '}
+                    <b>
+                        <PrettifyLargeNumber value={total} /> times
+                    </b>
+                    <br />
```
thomasheartman, line 70: This is a personal thing, but I don't `br` elements. I know we use them lots of places, but I'd rather wrap it in a span and set display to block or something. But this isn't important and you're free to disagree.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/FeatureOverviewEnvironmentMetrics/FeatureOverviewEnvironmentMetrics.tsx
```diff
@@ -0,0 +1,90 @@
...
+interface IFeatureOverviewEnvironmentMetrics {
+    environmentMetric?: IFeatureEnvironmentMetrics;
+    disabled?: boolean;
+}
+
+const StyledContainer = styled('div')({
+    marginLeft: 'auto',
+    display: 'flex',
+    alignItems: 'center',
+});
```
thomasheartman, line 16: I think this would be a perfect use case for the [`figure`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure) component, using the info as the `figcaption`.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/FeatureOverviewEnvironmentToggle/FeatureOverviewEnvironmentToggle.tsx
```diff
@@ -1,17 +1,25 @@
...
 import { useChangeRequestsEnabled } from 'hooks/useChangeRequestsEnabled';
 import { useFeature } from 'hooks/api/getters/useFeature/useFeature';
 import { useRequiredPathParam } from 'hooks/useRequiredPathParam';
 import type { IFeatureEnvironment } from 'interfaces/featureToggle';
 
-interface IFeatureOverviewEnvironmentToggleProps {
-    environment: IFeatureEnvironment;
-}
+const StyledContainer = styled('div')(() => ({
+    order: -1,
```
thomasheartman, line 10: Ooooh, I wanna be real careful about this. After my deep dive on ordering the other day, I'm not sure this is what we wanna do. How does this affect tab order?

From [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/order):

> Using the order property will create a disconnect between the visual presentation of content and DOM order. This will adversely affect low vision users navigating with the aid of assistive technology such as a screen reader. If the visual order differs from the DOM order, your users will have different experiences depending on how they access your content.

Can we do this without order?

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
thomasheartman, line 1: Is this the old component (not behind a flag)? I thought the new one was `...Environments`? Ah, no, it's probably `Legacy...`, right?

frontend/src/component/common/PercentageCircle/PercentageCircle.tsx
```diff
@@ -36,13 +40,21 @@ const PercentageCircle = ({
...
         <svg viewBox={`0 0 ${d} ${d}`} style={style} aria-hidden>
             <title>A circle progress bar with {percentage}% completion.</title>
+            <circle
+                r={r}
+                cx={r}
+                cy={r}
+                fill='none'
+                stroke={theme.palette.background.elevation2}
+                strokeWidth={d * strokeWidth}
+            />
```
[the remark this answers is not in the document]
Tymek: background gray circle

frontend/src/themes/theme.ts
```diff
@@ -509,13 +509,13 @@ export const lightTheme = createTheme({
...
         },
 
-        // Environment accordion
         MuiAccordion: {
             styleOverrides: {
                 root: ({ theme }) => ({
                     '&:first-of-type, &:last-of-type': {
                         borderRadius: theme.shape.borderRadiusLarge,
                     },
+                    // Environment accordion -- clean up with `flagOverviewRedesign` flag
```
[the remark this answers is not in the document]
Tymek: Used in "Legacy" file. 🤔 But maybe we can inline it there?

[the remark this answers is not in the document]
thomasheartman: So by "clean up", you mean "remove", right? That was my main point. It wasn't clear to me what you meant.

As long as that's the only place (the class selector seems to suggest so), then that sounds sensible to me, but ... I'd grep through just to make sure that it isn't used anywhere else first.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -56,7 +54,12 @@ export const EnvironmentHeader: FC<EnvironmentHeaderProps> = ({
         <StyledAccordionSummary
             {...props}
             expandIcon={
-                expandable ? <ExpandMore titleAccess='Toggle' /> : <></>
+                <ExpandMore
+                    titleAccess='Toggle'
+                    sx={{
+                        opacity: expandable ? 1 : 0,
```
thomasheartman, line 60: In case that's what you're doing here: opacity doesn't affect whether something is reachable by tab, but that might be handled in a different manner.

frontend/src/component/common/PercentageCircle/PercentageCircle.tsx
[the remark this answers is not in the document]
Tymek: I moved this to "PercentageDonut"

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -56,7 +54,12 @@ export const EnvironmentHeader: FC<EnvironmentHeaderProps> = ({
         <StyledAccordionSummary
             {...props}
             expandIcon={
-                expandable ? <ExpandMore titleAccess='Toggle' /> : <></>
+                <ExpandMore
+                    titleAccess='Toggle'
```
thomasheartman, line 58: I'd put the env name in here, too; e.g. `Toggle ${environment} information` or something.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
Tymek: That's a link to MUI v6. Right now we have `div`s all the way down

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/FeatureOverviewEnvironmentToggle/FeatureOverviewEnvironmentToggle.tsx
Tymek: Yup. We do not have navigation elements in between, and I think it is an issue for screen readers, to read environment name first and toggle later.

frontend/src/themes/theme.ts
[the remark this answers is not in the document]
Tymek: It will be tricky to copy all the `-expanded` behavior without bugs. I'll leave it for now

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
thomasheartman: You're right! In that case, I'm assuming you've assessed the situation and checked the header levels, then. Resolving.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/FeatureOverviewEnvironmentToggle/FeatureOverviewEnvironmentToggle.tsx
thomasheartman: Hmm, might be. But if this is on the feature page, why would they even need the flag name? 🤔 But yeah, I think I support your motion here.

frontend/src/themes/theme.ts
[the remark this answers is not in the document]
thomasheartman: Sounds good. I'd still maybe advise to change the comment to "remove" instead of clean up so that it's clearer if it's not you removing the flag.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
```diff
@@ -135,115 +75,70 @@ const FeatureOverviewEnvironment = ({
...
+                            variant='outlined'
+                            size='small'
+                        />
+                    )}
+                </EnvironmentHeader>
+                <StyledAccordionDetails>
+                    <StyledEnvironmentAccordionContainer>
+                        <EnvironmentAccordionBody
+                            featureEnvironment={featureEnvironment}
+                            isDisabled={!env.enabled}
```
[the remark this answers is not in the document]
thomasheartman: Gotcha!

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
Tymek: I wrapped accordion 'summary' in a `header` tag, and added `h2` on environment name

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/FeatureOverviewEnvironmentMetrics/FeatureOverviewEnvironmentMetrics.tsx
Tymek: Replaced with by doing it with CSS

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
Tymek: Handled with `visibility` + `aria-hidden`

frontend/src/component/common/PercentageCircle/PercentageCircle.tsx
```diff
@@ -1,17 +1,22 @@
 import { useTheme } from '@mui/material';
 import type { CSSProperties } from 'react';
 
-interface IPercentageCircleProps {
+type PercentageCircleProps = {
```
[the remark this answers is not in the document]
Tymek: If something depends on this being an `interface`, TS will throw an error, so it's a safe change. I do it kind of automatically. If a component is old, it shows me that it has been modified and updated to current syntax standard without looking at blame/history.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
Tymek: (replaced with new Truncator, instead of deprecated one)

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
thomasheartman: Yeah, I'll take that!

frontend/src/component/common/PercentageCircle/PercentageCircle.tsx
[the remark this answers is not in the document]
thomasheartman: Cool. Again, I like the change, but I'm curious as to what your reasoning for doing it is. You say you do it automatically, but why? As mentioned, I've done it myself in lots of places because I find it easier to extend the types `type A = B & C`, for instance, but I wanna hear *your* take on it.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -0,0 +1,87 @@
...
+                    sx={{ visibility: expandable ? 'visible' : 'hidden' }}
+                    aria-hidden={!expandable}
+                />
+            }
+            expandable={expandable}
+        >
+            <StyledHeader data-loading>
+                <StyledHeaderTitle>
+                    <StyledHeaderTitleLabel>Environment</StyledHeaderTitleLabel>
+                    <StyledTruncator component='h2'>
```
thomasheartman, line 79: Truncator only does line-based truncation, right? Maybe that's fine for this use case, though.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -0,0 +1,87 @@
...
+    expandable = true,
+    ...props
+}) => {
+    return (
+        <StyledAccordionSummary
+            {...props}
+            expandIcon={
+                <ExpandMore
+                    sx={{ visibility: expandable ? 'visible' : 'hidden' }}
+                    aria-hidden={!expandable}
```
thomasheartman, line 71: From [mdn's docs](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility):

> Using a visibility value of hidden on an element will remove it from the [accessibility tree](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/What_is_accessibility#accessibility_apis). This will cause the element and all its descendant elements to no longer be announced by screen reading technology.

So aria-hidden shouldn't be required here.

```suggestion
```

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -0,0 +1,87 @@
...
+                />
+            }
+            expandable={expandable}
+        >
+            <StyledHeader data-loading>
+                <StyledHeaderTitle>
+                    <StyledHeaderTitleLabel>Environment</StyledHeaderTitleLabel>
+                    <StyledTruncator component='h2'>
+                        {environmentId}
+                    </StyledTruncator>
```
thomasheartman, line 81: I'm not entirely sure about this. It puts the label before the header, whereas headers introduce the next section.

It'd be like this markdown:
```markdown
Environment

## Development
```

There is no logical connection between "environment" and the name of the environment. Again, going from the [`hgroup` docs on mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup), you're allowed to put `p`s before the header (but they must be `p`s):

> The `<hgroup>` element allows the grouping of a heading with any secondary content, such as subheadings, an alternative title, or tagline. Each of these types of content represented as a `<p>` element within the `<hgroup>`.

So I'd suggest changing `StyledHeaderTitle` to an `hgroup` and `StyledHeaderTitleLabel` to a `p`, so the rendered HTML is something like:

```html
<hgroup>
  <p>Environment</p>
  <h2>Development</h2>
</hgroup>
```

Seems more semantically correct to me.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/FeatureOverviewEnvironmentMetrics/FeatureOverviewEnvironmentMetrics.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { PrettifyLargeNumber } from 'component/common/PrettifyLargeNumber/PrettifyLargeNumber';
+import {
+    styled,
+    Tooltip,
+    tooltipClasses,
+    type TooltipProps,
+} from '@mui/material';
+import { PercentageDonut } from 'component/common/PercentageCircle/PercentageDonut';
+
+const StyledContainer = styled('figure')(({ theme }) => ({
```
thomasheartman, line 12: You love to see it

thomasheartman (approved):
Nice one! I still think we should use an `hgroup` as per my comment, but otherwise I'm pretty happy for this to go in.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
Tymek: It does both, I tested it here with `line={1}` (default)

PR #9351, opened 2025-02-24
chore: add file and component names to `styled` output class names in dev

thomasheartman:
Updates the `vite.config.mts` file to include config for emotion,
telling it to add `<file name>--<component name>` to the class names
of generated components.

The class name modification only happens in development mode.


nunogois (approved):
LGTM!

PR #9352, opened 2025-02-24
chore: months old version banner

nunogois:
https://linear.app/unleash/issue/2-3309/warn-users-of-lagging-versions-3-months-old

Adds a banner warning users in case they are running an outdated Unleash version (>=3 months old) in case the build date data is present.

When clicking "Changelog" the users are redirected to https://github.com/Unleash/unleash/releases.

image


chriswk (approved):
Nice

PR #9357, opened 2025-02-25
fix: get all strategies and check if env to be disabled

daveleek:
Context for this PR:

We noticed that when we had release plans enabled for the environment, the environment would still be switched off if we removed the last release-plans-unrelated strategy.
The reason is we do a `getFeature` in `optionallyDisableFeature`, then check the matching environments strategies if all of them are `disabled=true`. `getFeature` results in a db call to get from `features_view` which filters out release plan related strategies.

This PR introduces a flag `optionallyDisableWithStrategiesCheckOnly ` which when enabled sidesteps all that logic to simply call `feature-strategies-store.getStrategiesForFeatureEnv`, then does the same as before, checks all strategies for `disabled=true`, and disables the environment if so.
Side bonus here should be that it's a little more performant as it doesn't have to do all those requests to the db and all the mapping, when none of it beyond feature environment strategies is being used.

Working on writing some tests for this


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -793,6 +793,33 @@ class FeatureToggleService {
...
+                await this.unprotectedUpdateEnabled(
+                    projectId,
+                    featureName,
+                    environment,
+                    false,
+                    auditUser,
+                    user,
+                );
+            }
+            return Promise.resolve();
```
nunogois, line 820: I'm pretty sure this can be simplified to `return;`

nunogois (approved):
Looks reasonable to me, and it's behind a flag so it should be fine 🙂

src/lib/features/feature-toggle/tests/feature-toggle-service.limit.test.ts
```diff
@@ -97,7 +97,17 @@ describe('Strategy limits', () => {
         const { featureToggleService, featureStrategiesStore } =
             createFakeFeatureToggleService({
                 getLogger,
-                flagResolver: alwaysOnFlagResolver,
+                flagResolver: {
```
daveleek, line 100: The reason these tests broke needs to be investigated, I suspect the tests might rely on lacking implementations in fake methods

daveleek: Ok, so they're failing because the environment property is undefined. It's not marked as optional in `optionallyDisableFeature` though, so I'm a bit unsure what the right direction here is. Fixing the tests and always assuming the environment parameter is set, or escaping early if it is not set

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -793,6 +793,33 @@ class FeatureToggleService {
         auditUser: IAuditUser,
         user?: IUser,
     ): Promise<void> {
+        if (
```
kwasniew, line 796: something like this would be easier for me to parse (haven't checked correctness of the code):
```ts
async optionallyDisableFeature(
    featureName: string,
    environment: string,
    projectId: string,
    auditUser: IAuditUser,
    user?: IUser,
): Promise<void> {
    const disableFeature = async (strategies: IStrategy[]) => {
        const hasOnlyDisabledStrategies = strategies.every(strategy => strategy.disabled);
        if (hasOnlyDisabledStrategies) {
            await this.unprotectedUpdateEnabled(projectId, featureName, environment, false, auditUser, user);
        }
    };

    if (this.flagResolver.isEnabled('optionallyDisableWithStrategiesCheckOnly')) {
        const strategies = await this.featureStrategiesStore.getStrategiesForFeatureEnv(projectId, featureName, environment);
        await disableFeature(strategies);
        return;
    }

    const feature = await this.getFeature({ featureName });
    const env = feature.environments.find(e => e.name === environment);
    if (env) {
        await disableFeature(env.strategies);
    }
}
```

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -793,6 +793,33 @@ class FeatureToggleService {
...
+            this.flagResolver.isEnabled(
+                'optionallyDisableWithStrategiesCheckOnly',
```
kwasniew, line 798: This flag name is difficult to remember and use in conversations. Maybe simplifyOptionalDisableFeature or simplifyDisableFeature?

kwasniew (approved):

src/lib/features/feature-toggle/feature-toggle-service.ts
daveleek: Thank you, yeah might make sense to refactor it a little

src/lib/features/feature-toggle/feature-toggle-service.ts
daveleek: simplifyDisableFeature sounds good, thanks! I wasn't sure what to use, so I picked a hard one to provoke feedback 😁

PR #9358, opened 2025-02-25
feat: now project moving is split into two events

sjaanus:
When project is moved, then Unleash creates only one event, which is for target project.
We also need one for source project, to know that project was moved out of it.

Test will be in enterprise repo.


FredrikOseberg (approved):
LGTM

kwasniew (approved):

PR #9364, opened 2025-02-25
feat: ignore events in log when nothing has changed

gastonfournier:
## About the changes
Some automation may keep some data up-to-date (e.g. segments). These updates sometimes don't generate changes but we're still storing these events in the event log and triggering reactions to those events.

Arguably, this could be done in each service domain logic, but it seems to be a pretty straightforward solution: if preData and data are provided, it means some change happened. Other events that don't have preData or don't have data are treated as before.

Tests were added to validate we don't break other events.

### DIFF
--- a/src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
+++ b/src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
@@ -3609,11 +3609,16 @@ test('Updating feature strategy sort-order should trigger a an event', async () 
     );
 
     const strategies: FeatureStrategySchema[] = body;
-    let order = 1;
     const sortOrders: SetStrategySortOrderSchema = [];
 
-    strategies.forEach((strategy) => {
-        sortOrders.push({ id: strategy.id!, sortOrder: order++ });
+    // swap two strategies with different sort orders (note: first and second have the same sort order)
+    sortOrders.push({
+        id: strategies[0].id!,
+        sortOrder: strategies[2].sortOrder ?? 0,
+    });
+    sortOrders.push({
+        id: strategies[2].id!,
+        sortOrder: strategies[0].sortOrder ?? 0,
     });
 
     await app.request
--- a/src/lib/services/api-token-service.test.ts
+++ b/src/lib/services/api-token-service.test.ts
@@ -111,7 +111,7 @@ test('Api token operations should all have events attached', async () => {
         (e) => e.type === API_TOKEN_UPDATED,
     );
     expect(updatedApiTokenEvents).toHaveLength(1);
-    expect(updatedApiTokenEvents[0].preData.expiresAt).toBeDefined();
+    expect(updatedApiTokenEvents[0].preData.expiresAt).toBeUndefined();
     expect(updatedApiTokenEvents[0].preData.secret).toBeUndefined();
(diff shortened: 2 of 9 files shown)


nunogois (approved):
If I understood correctly, we're filtering out events where `preData` and `data` are present and equal.

If either one of them is not present, or if they are different from one another, we store the event as usual.

Sounds like an elegant solution to me 👍

gastonfournier:
> If I understood correctly, we're filtering out events where `preData` and `data` are present and equal.
> 
> If either one of them is not present, or if they are different from one another, we store the event as usual.

Correct, adding this to the description

src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
```diff
@@ -3609,11 +3609,16 @@ test('Updating feature strategy sort-order should trigger a an event', async ()
     );
 
     const strategies: FeatureStrategySchema[] = body;
-    let order = 1;
     const sortOrders: SetStrategySortOrderSchema = [];
 
-    strategies.forEach((strategy) => {
-        sortOrders.push({ id: strategy.id!, sortOrder: order++ });
```
gastonfournier, line 3616: Increasing all sortOrders by one was actually not changing the order of the strategies, so the event was not emitted after this change

src/test/fixtures/fake-api-token-store.ts
```diff
@@ -37,9 +37,10 @@ export default class FakeApiTokenStore
         return this.tokens.some((token) => token.secret === key);
     }
 
-    async get(key: string): Promise<IApiToken> {
-        // get can return undefined. See api-token-store.e2e.test.ts
-        return this.tokens.find((t) => t.secret === key);
+    async get(key: string): Promise<IApiToken | undefined> {
+        const found = this.tokens.find((t) => t.secret === key);
+        // clone the object to get a copy
+        return found ? { ...found } : undefined;
```
gastonfournier, line 43: Tests were failing because updating a property was actually modifying the object returned making the event to have the same payload in preData and data

src/lib/services/api-token-service.test.ts
```diff
@@ -111,7 +111,7 @@ test('Api token operations should all have events attached', async () => {
         (e) => e.type === API_TOKEN_UPDATED,
     );
     expect(updatedApiTokenEvents).toHaveLength(1);
-    expect(updatedApiTokenEvents[0].preData.expiresAt).toBeDefined();
+    expect(updatedApiTokenEvents[0].preData.expiresAt).toBeUndefined();
```
gastonfournier, line 114: When we first create the token the expiration is not provided. Probably this was working because of the fix in fake api token store

src/lib/types/stores/store.ts
```diff
@@ -1,5 +1,5 @@
 export interface Store<T, K> {
-    get(key: K): Promise<T>;
+    get(key: K): Promise<T | undefined>;
```
gastonfournier, line 2: I've been willing to make this change for a long time and it seems nothing is breaking, you might expect a few compiler warnings here and there (probably everywhere), but the truth is when you get an element by key you may not get the element if the key does not exist

nunogois (approved):
LG!

PR #9369, opened 2025-02-26
feat: SQL performance optimization to count instances

sjaanus:
The original query used `COUNT(DISTINCT app_name)` on a dataset with **12 million rows**, resulting in **16-second execution times** for some customers. This was due to how PostgreSQL processes `COUNT(DISTINCT)`:

- **Single-threaded execution**: PostgreSQL does not efficiently parallelize `COUNT(DISTINCT)`.
- **Sorting overhead**: Aggregation requires tracking all distinct values **before counting**.

#### Optimization  
We replaced `COUNT(DISTINCT app_name)` with a `GROUP BY` subquery, significantly improving performance:

- **Parallelized execution (`O(N / P)`, where `P` is the number of parallel workers)**: `GROUP BY` distributes work across multiple CPU cores.  
- **Faster deduplication**: Instead of sorting or hashing all values at once, PostgreSQL removes duplicates first.

#### Results  
- **Before:** 16 seconds ⏳ (`O(N log N)`)  
- **After:** **0.5 seconds 🚀** (`O(N / P)`, parallelized with `GROUP BY`) (30x faster!)

### DIFF
--- a/src/lib/db/client-instance-store.ts
+++ b/src/lib/db/client-instance-store.ts
@@ -253,17 +253,26 @@ export default class ClientInstanceStore implements IClientInstanceStore {
     }
 
     async getDistinctApplicationsCount(daysBefore?: number): Promise<number> {
-        let query = this.db.from(TABLE);
-        if (daysBefore) {
-            query = query.where(
-                'last_seen',
-                '>',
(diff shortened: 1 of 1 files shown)


FredrikOseberg (approved):
Nice!

PR #9370, opened 2025-02-26
feat: allow command bar navigation to wrap around

thomasheartman:
Lets you navigate to the top of the list when you're at the bottom,
and vice versa.

Arrow down at the end of the list takes you to the search field and
arrow up from the search field takes you to the end of the list.


sjaanus (approved):
LG

PR #9382, opened 2025-02-27
chore: adjust env dropdown

thomasheartman:
Adjusts styling of the env dropdown now that we have both release plans and strategies.

Key points:
- simplifies strategy separator, removes inherent height. Also: extracts it from the draggable component (it has no business knowing whether to add that or not)
- Puts release plans and strategies in the same list so that it becomes:
```markdown
- Release plan
  - strategy 1
  - strategy 2
- (OR) Strategy A
- (OR) Strategy B
```
- Adjusts some padding around to make it line up properly
- Swaps a couple conditional renders for ternaries

Rendered:
image

## Still todo:

Handle cases where you have >50 strats and we show the warning etc. It's a little trickier because of how it interacts with release plans, so I wanna leave that for later.

I'm also unsure about how we handle spacing today. All the little items have their own different spacing and I'm not sure it won't get out of sync, but I'm also not sure how else to handle it. We should look at it later.

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
@@ -20,7 +20,7 @@ import type { IFeatureStrategy } from 'interfaces/strategy';
 import { usePlausibleTracker } from 'hooks/usePlausibleTracker';
 import { useUiFlag } from 'hooks/useUiFlag';
 import { useReleasePlans } from 'hooks/api/getters/useReleasePlans/useReleasePlans';
-import { StrategyDraggableItem as NewStrategyDraggableItem } from './StrategyDraggableItem/StrategyDraggableItem';
+import { StrategyDraggableItem } from './StrategyDraggableItem/StrategyDraggableItem';
 import { ReleasePlan } from '../../../ReleasePlan/ReleasePlan';
 import { StrategySeparator } from 'component/common/StrategySeparator/StrategySeparator';
 
@@ -252,7 +252,7 @@ export const EnvironmentAccordionBody = ({
                                             <StrategySeparator text='OR' />
                                         ) : null}
 
-                                        <NewStrategyDraggableItem
+                                        <StrategyDraggableItem
                                             strategy={strategy}
                                             index={index}
                                             environmentName={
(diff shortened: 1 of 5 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -232,122 +231,102 @@ export const EnvironmentAccordionBody = ({
...
+                                                    featureEnvironment.name
+                                                }
+                                                otherEnvironments={
+                                                    otherEnvironments
+                                                }
+                                                isDragging={false}
+                                                onDragStartRef={
+                                                    (() => {}) as any
+                                                }
+                                                onDragOver={(() => {}) as any}
```
Tymek, line 305: I can see why drag&drop didn't work for paginated items. Maybe parameters like that should be optional, instead of forcing `any`?

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -22,52 +20,53 @@ import type { IFeatureStrategy } from 'interfaces/strategy';
...
+    borderBottom: `1px solid ${theme.palette.divider}`,
+    background:
+        type === 'release plan'
+            ? theme.palette.background.elevation2
+            : theme.palette.background.elevation1,
+    position: 'relative',
+    paddingBlock: theme.spacing(2.5),
+    '&:first-of-type': {
+        paddingTop: theme.spacing(1),
+    },
```
Tymek, line 57: maybe `flex` + `gap` do the trick here

Tymek (approved):

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
thomasheartman: Huh, doesn't it work for paginated items? Maybe it doesn't. I've not touched this bit at all, so I didn't even realize that it was like this 😅 I'll ... inspect it later.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
thomasheartman: In this case, I don't think it will. We need the list items to take up their full height for the borders (and different background colors) to work as expected. I'd like to find a way around this, but I'm not sure how yet. I anticipate there'll be more iterations, though, so I'm gonna leave it in for now.

PR #9385, opened 2025-02-27
feat: frontend traffic tab

kwasniew:
## About the changes

When consumption model is enabled we have frontend traffic tab in addition to backend connections
Screenshot 2025-02-27 at 15 47 35

This PR does not take Edge traffic into account. Edge traffic will need to be incorporated server side


frontend/src/component/admin/network/NetworkTrafficUsage/chart-functions.ts
```diff
@@ -15,9 +15,11 @@ export type ChartDatasetType = ChartDataset<'bar'>;
 
 export const toTrafficUsageChartData = (
     traffic: TrafficUsageDataSegmentedCombinedSchema,
+    filter?: string,
```
kwasniew, line 18: adjusted a function used for data usage with an option to filter out only one traffic type e.g. frontend

PR #9389, opened 2025-02-27
chore: new add release plan dialog

nunogois:
https://linear.app/unleash/issue/2-3249/adding-a-release-plan-to-a-non-cr-environment-feels-too-immediate

This adds a new confirmation / preview dialog when adding a release plan.

What's cool about it is that it will describe what will happen before you confirm. It also acts as the "add to CR" dialog, so we now only have 1 dialog instead of 2 separate ones.

This also refactors quite a bit of our code here, hopefully simplifying it.

### Simple (env disabled)
image

### CR protected (env enabled)
image


daveleek (approved):
Nice!

PR #9394, opened 2025-02-28
refactor(1-3429): makes the drag args optional, defaulting to noops

thomasheartman:
Removes the `() => {} as any` args from the StrategyDraggableItem invocation when you have paginated strats. Instead makes all the drag params optional. It defaults to a no op if not provided.

Also, the reason it had to be typed as `any` before is probably because it was missing a function. The correct empty param is `() => () => {}` 💁


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyDraggableItem.tsx
```diff
@@ -23,26 +23,28 @@ interface IStrategyDraggableItemProps {
...
-    onDragOver: (
+    onDragOver?: (
         ref: RefObject<HTMLDivElement>,
         index: number,
     ) => DragEventHandler<HTMLDivElement>;
-    onDragEnd: () => void;
+    onDragEnd?: () => void;
 }
 
+const noop = () => () => {};
```
Tymek, line 37: maybe `onDragNoop`? Because `noop` us usually `() => {}`

Tymek (approved):

PR #9395, opened 2025-02-28
chore: flip UI and backend validation for project env disabling

sighphyre:
Moves the validation for whether or not a project can have 0 environments from the backend to the frontend. 

Currently the frontend doesn't care, which means that the user can trigger a disable for the last environment for a project, which creates a vague error message. This PR disables the slider for the last env on a project on the frontend

Conversely, preventing the API from ever creating a project with no environments means automation tools like terraform have no way to manage the state correctly and effectively. This relaxes that validation, allowing 0 environment projects to be created if and only if they're explicitly requested to have no environments. There's no code path that uses this in the UI so that should be safef


src/lib/features/project/project-service.ts
```diff
@@ -383,7 +377,7 @@ export default class ProjectService {
 
         await this.projectStore.create(data);
 
-        const envsToEnable = newProject.environments?.length
+        const envsToEnable = newProject.environments
```
sighphyre, line 380: This is actually the salient change. The original code relies on length of 0 being falsy.

src/lib/features/project/project-service.e2e.test.ts
```diff
@@ -2853,13 +2853,6 @@ describe('create project with environments', () => {
         expect(created).toMatchObject(allEnabledEnvs);
     });
 
-    test('an empty list throws an error', async () => {
```
sighphyre, line 2856: Intentionally taking this out. We want this to now happen

gastonfournier (approved):
LGTM!

PR #9400, opened 2025-02-28
chore: release plan replacement UI

nunogois:
https://linear.app/unleash/issue/2-3099/when-attempting-to-add-a-release-plan-to-a-feature-environment-already

Adapts our UI to better provide hints about our release plan replacement behavior, both on the add dialog and change requests.

### Add dialog

image

### CR

image

image

image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ReleasePlanChange.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ReleasePlanChange.tsx
@@ -1,5 +1,5 @@
 import type React from 'react';
-import type { FC, ReactNode } from 'react';
+import { useRef, useState, type FC, type ReactNode } from 'react';
 import { Box, styled, Typography } from '@mui/material';
 import type {
     ChangeRequestState,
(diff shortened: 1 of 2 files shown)


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ReleasePlanChange.tsx
```diff
@@ -177,24 +146,107 @@ const StartMilestone: FC<{
...
+        if (currentTooltipCloseTimeoutRef.current) {
+            clearTimeout(currentTooltipCloseTimeoutRef.current);
+        }
+        setCurrentTooltipOpen(true);
+    };
+    const closeCurrentTooltip = () => {
+        currentTooltipCloseTimeoutRef.current = setTimeout(() => {
+            setCurrentTooltipOpen(false);
+        }, 100);
+    };
```
nunogois, line 172: Unfortunately the default tooltip behavior did not work correctly with the accordions. If you toggled one of them after the other, it would close the tooltip.

This required some manual plumbing (manually handle the tooltip open state), but the final result seems worth it.

PR #9414, opened 2025-03-04
chore: archive release plan template event

daveleek:
creates an event for archiving release plan templates

### DIFF
--- a/src/lib/types/events.ts
+++ b/src/lib/types/events.ts
@@ -210,6 +210,8 @@ export const RELEASE_PLAN_TEMPLATE_UPDATED =
     'release-plan-template-updated' as const;
 export const RELEASE_PLAN_TEMPLATE_DELETED =
     'release-plan-template-deleted' as const;
+export const RELEASE_PLAN_TEMPLATE_ARCHIVED =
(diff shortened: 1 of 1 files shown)


src/lib/types/events.ts
```diff
@@ -2094,6 +2097,20 @@ export class ReleasePlanTemplateDeletedEvent extends BaseEvent {
...
+    readonly preData: any;
+    readonly data: any;
+    constructor(eventData: {
+        data: any;
+        preData: any;
+        auditUser: IAuditUser;
+    }) {
+        super(RELEASE_PLAN_TEMPLATE_ARCHIVED, eventData.auditUser);
+        this.data = eventData.data;
+        this.preData = eventData.preData;
```
nunogois, line 2110: Should it have both data and preData? What would you expect to see in these fields?

See removed event below, for example.

nunogois (approved):
LGTM

src/lib/types/events.ts
daveleek: There will be an archived_at column in the data, so figured that warranted a data field

PR #9415, opened 2025-03-04
feat: UI for archiving release templates

daveleek:
This reworks the delete release template UI to revolve around soft deleting (archiving) release templates. Still requires the delete permission


gastonfournier (approved):
LG!

PR #9416, opened 2025-03-04
Downgrade openapi-docs-plugin

melindafekete:


gastonfournier (approved):
It looks like this has fixed the issue with docs search! :clap:

PR #9427, opened 2025-03-05
feat: project environments include visible property

sighphyre:
Adds a new property to the project environments API to indicate whether or not a given environment is visible on the project requested. This helps us not have to depend on UI specific endpoints for integrations.

Also removes a check which prevents you from removing the last environment on a project. That should have been caught in a previous PR but wasn't


src/lib/features/project-environments/environment-service.ts
```diff
@@ -77,8 +76,24 @@ export default class EnvironmentService {
 
     async getProjectEnvironments(
         projectId: string,
-    ): Promise<IProjectEnvironment[]> {
-        return this.environmentStore.getProjectEnvironments(projectId);
+    ): Promise<IProjectsAvailableOnEnvironment[]> {
+        // This function produces an object for every environment, in that object is a boolean
```
sighphyre, line 80: I'm usually not a fan of explaining what's going on with comments but this behaviour is so surprising to me that I had to reverse engineer it from API behaviour and it'd be cool to save the next person the trouble

src/lib/features/project-environments/environment-service.ts
```diff
@@ -254,22 +269,13 @@ export default class EnvironmentService {
...
-            await this.eventService.storeEvent(
-                new ProjectEnvironmentRemoved({
-                    project: projectId,
-                    environment,
-                    auditUser,
-                }),
-            );
-            return;
-        }
-        throw new MinimumOneEnvironmentError(
```
sighphyre, line 271: Removed this validation, we should no longer need this

gastonfournier: Would be nice to remove the validation class, can be done as a follow up if you prefer, but this is probably the last place we used it

src/lib/openapi/spec/environment-project-schema.ts
```diff
@@ -57,6 +57,12 @@ export const environmentProjectSchema = {
                 'The strategy configuration to add when enabling a feature environment by default',
             $ref: '#/components/schemas/createFeatureStrategySchema',
         },
+        visible: {
+            type: 'boolean',
+            example: true,
+            description:
+                'Indicates if the environment is available to be enabled for feature flags in the project',
```
gastonfournier, line 64: This phrase looks a bit weird, but my suggestion might be too simplistic
```suggestion
                'Indicates if the environment is visible in the project',
```

src/lib/features/project-environments/environment-service.ts
gastonfournier: Maybe something to add to the list of major API changes for next release, but first we'd have to deprecate it in the API in the next major

gastonfournier (approved):
LG with some minor improvement suggestions

src/lib/features/project-environments/environment-service.ts
sighphyre: Cheers, good call, removed

src/lib/openapi/spec/environment-project-schema.ts
sighphyre, line 64: Yeah I think that doesn't really capture what this property actually represents and this is actually really, really important behaviour. I've asked ChatGPT to clean it up, we can always change it later

```suggestion
                'Indicates whether the environment can be enabled for feature flags in the project',
```

PR #9432, opened 2025-03-05
docs: SDK refresh intervals

kwasniew:
## About the changes

Document backend and client/frontend SDK intervals.

Details:
* most backend SDKs have synced intervals (except from Rust). Some SDKs are still waiting for a final publish.
* adding client SDKs interval comparison table. Full client/frontend SDK feature comparison table is outside of this PR which focuses only on intervals.

Screenshot 2025-03-05 at 09 54 46


sjaanus (approved):
LG

melindafekete (approved):
Thanks!

PR #9439, opened 2025-03-06
Fix: drag and drop reverts to previous position on drop

thomasheartman:
Fixes an issue where, when dragging large expanded milestone cards, the position would revert from the current visual state to the previous one when you drop the item.


nunogois (approved):
LGTM

PR #9443, opened 2025-03-07
chore(1-3450): Place strategy names and titles on the same line (and fix list nesting issues)

thomasheartman:
Moves strategy titles and names onto the same line, as per the new designs.

In doing so, I've also updated the component to use a more semantic hgroup with the header being the strategy title if it exists or the strategy name if not.

The downside of being more semantically correct here is that we need to know what header level we want the strategy to use. In most cases, that's 3 (e.g. flag name > environment > strategy, release plan > milestone > strategy), but for plans on flag envs, it's 4 (flag name > env > milestone name > strategy).

I've also taken the opportunity to fix a little mistake I made earlier. `ol`s can only have `li` children, and I'd forgotten to wrap a nested `ol` inside an `li`. The changes in  `EnvironmentAccordionBody` all relate to that change. Because we now have several layers of lists nested within each other, dealing with styling and padding gets a little tricky, but CSS has the power do help us out here. 

Rendered:
image


frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -44,15 +44,17 @@ const StyledCustomTitle = styled('div')(({ theme }) => ({
...
-    verticalAlign: 'middle',
-});
+const StyledHeaderContainer = styled('hgroup')(({ theme }) => ({
+    display: 'flex',
+    flexFlow: 'row',
+    columnGap: '1ch',
+    fontSize: theme.typography.body1.fontSize,
+    '*:first-child': {
+        fontWeight: 'bold',
+    },
```
thomasheartman, line 54: We don't know whether the first item is going to be a header or a p element, but we always want the first item to be bold. We could also solve it by applying a class, but this feels easier.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -44,15 +44,17 @@ const StyledCustomTitle = styled('div')(({ theme }) => ({
...
 }));
-const StyledHeaderContainer = styled('div')({
-    flexDirection: 'column',
-    justifyContent: 'center',
-    verticalAlign: 'middle',
-});
+const StyledHeaderContainer = styled('hgroup')(({ theme }) => ({
+    display: 'flex',
+    flexFlow: 'row',
+    columnGap: '1ch',
```
thomasheartman, line 50: Approximates one space.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -112,33 +114,39 @@ export const StrategyItemContainer: FC<StrategyItemContainerProps> = ({
...
-                                            String(strategy.title),
+                                            String(strategy.name),
                                         )}
-                                    </StyledCustomTitle>
+                                        :
+                                    </p>
+                                    <Typography
+                                        component={`h${strategyHeaderLevel}`}
+                                    >
+                                        {strategy.title}
```
thomasheartman, line 130: Not sure why we tried to format the strategy title as a strategy name before. pretty sure that's an oversight. If you call your strategy "flexibleRollout", you probably want to see that and not "Gradual rollout" 💁🏼

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -36,11 +36,25 @@ const StyledAccordionBodyInnerContainer = styled('div')(({ theme }) => ({
...
+    },
+    '&:not(li > &) > li:first-of-type': {
+        // select first list elements in lists that are not directly nested
+        // within other lists.
+        paddingTop: theme.spacing(1),
+    },
+    '& > li:has(> ol)': {
+        // nested lists add their own padding to their list items, so we don't want to double it up.
+        paddingTop: 0,
+    },
```
thomasheartman, line 56: Yay, CSS selectors 🥳  I moved all the li spacing and positioning up here so that it's colocated.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -243,75 +252,83 @@ export const EnvironmentAccordionBody = ({
...
-                                <StyledAlert severity='warning'>
-                                    We noticed you're using a high number of
-                                    activation strategies. To ensure a more
-                                    targeted approach, consider leveraging
-                                    constraints or segments.
-                                </StyledAlert>
+                        <li>
+                            {releasePlans.length > 0 ? (
+                                <StrategySeparator />
+                            ) : null}
```
thomasheartman, line 258: Nest nested lists within list items and add the separator between release plans and strategies to this element instead of the first strategy.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
thomasheartman: Update: use strategy name instead of first child.

Emotion complains that:
> The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".

So let's use a class instead.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -36,11 +36,25 @@ const StyledAccordionBodyInnerContainer = styled('div')(({ theme }) => ({
...
+    },
+    '&:not(li > &) > li:first-of-type': {
+        // select first list elements in lists that are not directly nested
+        // within other lists.
+        paddingTop: theme.spacing(1),
+    },
+    '& > li:has(> ol)': {
+        // nested lists add their own padding to their list items, so we don't want to double it up.
+        paddingBlock: 0,
+    },
```
thomasheartman, line 56: Yay, CSS selectors 🥳 I moved all the li spacing and positioning up here so that it's colocated.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
nunogois: Aren't they essentially the same thing, with "Gradual rollout" being the friendly, more readable, name?

nunogois: Oh I see what you mean now. Yes, looks like an oversight, users should be able too call their strategies whatever they want 😄

nunogois (approved):
LG!

PR #9446, opened 2025-03-10
feat: always show features table

kwasniew:
## About the changes

We have gotten two feedbacks related to the onboarding on project level that would be nice to fix

If you create a flag, and archive it, you get back to the onboarding on project overview. The problem is that you can't find your archived flag (since archive option is in the table)

We don't show the import icon before you have created a flag (since it belongs to the list), which hides that option which is a bit inconvenient

Solution: show list all the time even though there are no flags, as it will solve both cases
Screenshot 2025-03-10 at 10 12 55

### DIFF
--- a/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureToggles.tsx
+++ b/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureToggles.tsx
@@ -148,14 +148,9 @@ export const ProjectFeatureToggles = ({
         'hide-setup' | 'show-setup'
     >(`onboarding-state:v1-${projectId}`, 'hide-setup');
 
-    const notOnboarding =
-        project.onboardingStatus.status === 'onboarded' ||
-        onboardingFlow === 'closed';
     const isOnboarding =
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -495,128 +490,113 @@ export const ProjectFeatureToggles = ({
                     />
                 }
             />
-            <ConditionallyRender
```
kwasniew, line 498: this code removes conditional render and only keeps the show part

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -148,14 +148,9 @@ export const ProjectFeatureToggles = ({
         'hide-setup' | 'show-setup'
     >(`onboarding-state:v1-${projectId}`, 'hide-setup');
 
-    const notOnboarding =
```
kwasniew, line 151: those become unused after the conditional render is gone

PR #9447, opened 2025-03-10
chore: use new designs for project default strategies

thomasheartman:
Implements the new strategy list design for default strategies. Moves the old impl into a legacy file. Also: removes the description from the strategy item. From my digging, we only showed this for default strategy items and it didn't really provide any useful information. The only other place you can add a description is for custom strategies (at least that I could find), but these are deprecated and we never show the description when you apply the strategy anyway.

Rendered:
image

Without the flag (nothing changes):
image


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -71,6 +69,8 @@ export const EnvironmentHeader: FC<EnvironmentHeaderProps> = ({
                     sx={{ visibility: expandable ? 'visible' : 'hidden' }}
                 />
             }
+            id={id}
+            aria-controls={`environment-accordion-${id}-content`}
```
thomasheartman, line 73: https://mui.com/material-ui/react-accordion/#accessibility

frontend/src/component/project/Project/ProjectSettings/ProjectDefaultStrategySettings/ProjectEnvironment/ProjectEnvironmentDefaultStrategy/ProjectEnvironmentDefaultStrategy.tsx
```diff
@@ -51,44 +48,42 @@ const ProjectEnvironmentDefaultStrategy = ({
...
-                <StrategyExecution strategy={strategy} />
-
-                {strategy.variants && strategy.variants.length > 0 ? (
-                    <SplitPreviewSlider variants={strategy.variants} />
-                ) : null}
-            </StrategyItemContainer>
-        </>
+        <StrategyItem
+            strategy={strategy}
+            headerItemsRight={
```
Tymek, line 67: Maybe "actions"/"headerActions" prop is better, because it's less coupled to the visual design?

Tymek (approved):

frontend/src/component/project/Project/ProjectSettings/ProjectDefaultStrategySettings/ProjectEnvironment/ProjectEnvironmentDefaultStrategy/ProjectEnvironmentDefaultStrategy.tsx
thomasheartman: Yeah, maybe. The thing is that when we get to the playground, there's also left-aligned items. We can either solve that by the caller passing in a container that wraps it or by adding right and left props. But I wanna delay this decision until we've found all the places we use it and then take it from there.

PR #9497, opened 2025-03-10
chore: remove deprecated getAdmin from client features store

gastonfournier:
This method has been deprecated 2 years ago and is not used


daveleek (approved):
Looks good, could not find it in use anywhere either!

PR #9498, opened 2025-03-10
chore: remove deprecated method from project-service

gastonfournier:
Depends on https://github.com/bricks-software/unleash-enterprise/pull/137

This method has been deprecated and has an alternative. Migrating to that alternative

### DIFF
--- a/src/lib/features/project/project-service.e2e.test.ts
+++ b/src/lib/features/project/project-service.e2e.test.ts
@@ -934,7 +934,7 @@ test('should add admin users to the project', async () => {
     await isProjectUser(adminUsers[2].id, project.id, true);
 });
 
-test('add user should fail if user already have access', async () => {
+test('add user do nothing if user already has access', async () => {
     const project = {
         id: 'add-users-twice',
         name: 'New project',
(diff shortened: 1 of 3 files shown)


src/lib/features/project/project-service.e2e.test.ts
```diff
@@ -2328,20 +2321,17 @@ test('should get correct amount of project members for current and past window',
...
+        [memberRole.id],
+        [], // no groups
+        createdUsers.map((u) => u.id),
+        auditUser,
     );
 
     const result = await projectService.getStatusUpdates(project.id);
     expect(result.updates.projectMembersAddedCurrentWindow).toBe(6); // 5 members + 1 owner
-    expect(result.updates.projectActivityCurrentWindow).toBe(6);
+    expect(result.updates.projectActivityCurrentWindow).toBe(2);
```
gastonfournier, line 2334: @FredrikOseberg @kwasniew this one caught my attention. Instead of adding 5 users in 5 independent operations, we could add all of them in one operation. I'm just curious what we use this projectActivityCurrentWindow for, because if it's intended to count 5 in a bulk update instead of 1, maybe it's not counting as it's supposed to.

src/lib/features/project/project-service.e2e.test.ts
```diff
@@ -944,7 +934,7 @@ test('should add admin users to the project', async () => {
     await isProjectUser(adminUsers[2].id, project.id, true);
 });
 
-test('add user should fail if user already have access', async () => {
+test('add user do nothing if user already has access', async () => {
```
gastonfournier, line 937: Highlighting this change in behaviour

src/lib/features/project/project-service.e2e.test.ts
gastonfournier: Just checked with them on Slack and this used to serve an endpoint that's now deprecated:
```
/** @deprecated use project insights instead */
this.route({
    method: 'get',
    path: '/:projectId/dora',
```

So this change shouldn't have a bad impact in our UX

daveleek (approved):
Awesome, LG!

PR #9499, opened 2025-03-10
chore: remove legacy feature type

gastonfournier:
This has been deprecated 4 years ago and is not used


daveleek (approved):
Couldn't find any usage of it either! LG

PR #9503, opened 2025-03-11
feat: archived indicators and disabled button when editing archived template

daveleek:
Adds an indication that the template has been archived and disables the save changes button
image


gastonfournier:
@daveleek rebase :D

chriswk (approved):
LGTM

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm/TemplateForm.tsx
```diff
@@ -81,6 +83,11 @@ export const TemplateForm: React.FC<ITemplateFormProps> = ({
             description={<TemplateFormDescription />}
             formatApiCode={formatApiCode}
         >
+            {archived && (
+                <Alert severity='warning'>
+                    This template has been archived and can no longer be edited.
+                </Alert>
+            )}
```
gastonfournier, line 90: Should we also have this condition on the save button?

gastonfournier (approved):
I think if we're not allowing the customer to save, besides the banner we should have a condition in the button, maybe it's already there

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm/TemplateForm.tsx
daveleek: Thank you, yes it's in EditTemplate where the button is passed in

PR #9504, opened 2025-03-11
chore(1-3422): playground strategies list

thomasheartman:
Initial rough work on adapting the playground strategies to the new designs. This PR primarily splits components into Legacy files and adds new replacements. There are *some* updates (including spacing and text color), but nothing juicy yet. However, I wanted to get this in now, before this PR grows even bigger.


frontend/src/component/playground/Playground/PlaygroundResultsTable/FeatureResultInfoPopoverCell/FeatureDetails/FeatureDetails.test.tsx
```diff
@@ -10,7 +11,7 @@ const testCases = [
             hasUnsatisfiedDependency: true,
             isEnabledInCurrentEnvironment: false,
         } as PlaygroundFeatureSchema,
-        expectedText1: 'This feature flag is False in development because',
+        expectedText1: /This feature flag is False in development because/,
```
thomasheartman, line 14: The exact HTML structure has changed a little bit in the new component (from `<p><span><span><span></p>` to `<p>text<span>text</p>`), so the old matcher didn't work correctly. But making it a regex seems harmless enough and shows that it can still find the text.

frontend/src/component/playground/Playground/PlaygroundResultsTable/FeatureResultInfoPopoverCell/FeatureDetails/FeatureDetails.tsx
```diff
@@ -124,61 +111,43 @@ export const FeatureDetails = ({
...
-                    <ConditionallyRender
-                        condition={feature?.strategies?.result !== 'unknown'}
-                        show={() => (
+            <HeaderRow>
+                <HeaderGroup>
+                    <StyledTypographyName>{feature.name}</StyledTypographyName>
+                    <p>
+                        {feature?.strategies?.result !== 'unknown' ? (
                             <PlaygroundResultChip
+                                tabindex={-1}
```
thomasheartman, line 120: These shouldn't be tabbable. They don't do anything.

frontend/src/component/playground/Playground/PlaygroundResultsTable/PlaygroundResultChip/PlaygroundResultChip.tsx
```diff
@@ -11,12 +11,14 @@ interface IResultChipProps {
     label: string;
     // Result icon - defaults to true
     showIcon?: boolean;
+    tabindex?: number;
```
thomasheartman, line 14: Allows you to make these non-tabbable. I don't think there is a reason they should be tabbable at all, so I might default it to -1 later.

Tymek (approved):

PR #9509, opened 2025-03-11
chore: use a local image of unleash enterprise for testing FE

gastonfournier:
This removes the need of running a heroku instance with enterprise to run frontend tests. We can also get rid of vercel deployments of the frontend for PRs (ofc we'd be loosing the preview if we do this)

It relies on the free enterprise docker image to spin up a clean Unleash instance on each frontend spec test.

It also seems to speed up tests, I guess because we don't have to wait for vercel deployment queue:
image

Compared with other recent PRs:
image

Unrelated to this PR but something is making the features spec tests very slow since Cypress upgrade: https://github.com/Unleash/unleash/actions/runs/13784760762/job/38550079932#step:4:138

### DIFF
--- a/.github/docker-compose.test.yml
+++ b/.github/docker-compose.test.yml
@@ -1,0 +1,56 @@
+# This docker compose setup configures:
+# - the Unleash server instance + the necessary backing Postgres database
+# - the Unleash proxy
+#
+# To learn more about all the parts of Unleash, visit
+# https://docs.getunleash.io
+#
+# NOTE: please do not use this configuration for production setups.
+# Unleash does not take responsibility for any data leaks or other
+# problems that may arise as a result.
+#
+# This is intended to be used for demo, development, and learning
+# purposes only.
+
+services:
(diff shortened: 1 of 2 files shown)


.github/docker-compose.test.yml
```diff
@@ -0,0 +1,56 @@
+# This docker compose setup configures:
+# - the Unleash server instance + the necessary backing Postgres database
+# - the Unleash proxy
```
gastonfournier, line 3:
```suggestion
```

.github/docker-compose.test.yml
```diff
@@ -0,0 +1,56 @@
+# This docker compose setup configures:
+# - the Unleash server instance + the necessary backing Postgres database
```
gastonfournier, line 2:
```suggestion
# - the Unleash enterprise server instance + the necessary backing Postgres database
```

ivarconr (approved):

PR #9510, opened 2025-03-11
chore: Playground Strategy Lists

thomasheartman:
Continue the implementation of Playground strategy lists. This PR also adjusts some existing strategy container and list items to accomodate more use cases (such as this).

The playground strategy execution component is still the old design.

After (playground results):
image

After (env strategy list):
image

Before (env strategy list):
image


frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -16,16 +15,20 @@ type StrategyItemContainerProps = {
     onDragStart?: DragEventHandler<HTMLButtonElement>;
     onDragEnd?: DragEventHandler<HTMLButtonElement>;
     headerItemsRight?: ReactNode;
+    headerItemsLeft?: ReactNode;
```
thomasheartman, line 18: This is where this becomes relevant. Because they're styled differently from the right header items, I think it makes sense for them to handled as a separate prop. Also, as evidenced here, they're not all actions, so "headerActions" etc isn't that accurate a name.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -16,16 +15,20 @@ type StrategyItemContainerProps = {
...
     className?: string;
     style?: React.CSSProperties;
     children?: React.ReactNode;
 };
 
-const DragIcon = styled(IconButton)({
+const inlinePadding = 3;
```
thomasheartman, line 24: Offset the drag icon so that the strategy design matches the sketches (strategy title is aligned with strategy content such as constraints.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -38,9 +41,14 @@ const StyledHeaderContainer = styled('hgroup')(({ theme }) => ({
...
-const StyledContainer = styled('article')({
+const StyledContainer = styled('article')(({ theme }) => ({
     background: 'inherit',
-});
+    padding: theme.spacing(inlinePadding),
+    paddingTop: theme.spacing(0.5),
+    display: 'flex',
+    flexDirection: 'column',
+    rowGap: theme.spacing(0.5),
+}));
```
thomasheartman, line 51: Move spacing that was previously distributed among two elements (header, content) and let the container handle how it lays out its items.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -49,102 +57,101 @@ const StyledTruncator = styled(Truncator)(({ theme }) => ({
...
+    display: 'flex',
+    alignItems: 'center',
+    color: disabled ? theme.palette.text.secondary : theme.palette.text.primary,
+}));
+
+const StyledHeaderInner = styled('div')(({ theme }) => ({
+    display: 'flex',
+    alignItems: 'center',
+    gap: theme.spacing(1),
+}));
```
thomasheartman, line 71: Allow for the draggable item with an offset and use a nested div for header and headerItemsLeft (badges etc)

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -49,102 +57,101 @@ const StyledTruncator = styled(Truncator)(({ theme }) => ({
...
+                                )}
+                            </StyledHeaderContainer>
+                        </StrategyHeaderLink>
 
-                    {strategy.disabled ? (
-                        <Badge color='disabled'>Disabled</Badge>
-                    ) : null}
+                        {strategy.disabled ? (
+                            <Badge color='disabled'>Disabled</Badge>
+                        ) : null}
```
thomasheartman, line 141: In the future, we should probably not apply this in the playgroud.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -30,9 +30,7 @@ interface IEnvironmentAccordionBodyProps {
 }
 
 const StyledAccordionBodyInnerContainer = styled('div')(({ theme }) => ({
-    [theme.breakpoints.down(400)]: {
-        padding: theme.spacing(1),
-    },
+    borderBottom: `1px solid ${theme.palette.divider}`,
```
thomasheartman, line 33: Add a border at the end of the container, before the accordion footer.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -58,7 +56,9 @@ export const StyledContentList = styled('ol')(({ theme }) => ({
 export const StyledListItem = styled('li', {
     shouldForwardProp: (prop) => prop !== 'type',
 })<{ type?: 'release plan' | 'strategy' }>(({ theme, type }) => ({
-    borderBottom: `1px solid ${theme.palette.divider}`,
+    '& + &, li > ol > &:first-of-type': {
```
thomasheartman, line 59: We don't always want to apply the border: it should only be applied to subsequent list items OR to list items that are at the beginning of a nested list

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -290,34 +290,29 @@ export const EnvironmentAccordionBody = ({
...
+                                <StrategySeparator />
+                                <Strategies />
+                            </li>
+                        ) : null}
+                    </>
+                ) : strategies.length > 0 ? (
+                    <Strategies />
+                ) : null}
+            </StyledContentList>
+        </StyledAccordionBodyInnerContainer>
```
thomasheartman, line 316: Just removes the completely redundant wrapping div.

frontend/src/component/playground/Playground/PlaygroundResultsTable/FeatureResultInfoPopoverCell/FeatureResultInfoPopoverCell.tsx
```diff
@@ -65,32 +60,96 @@ export const FeatureResultInfoPopoverCell = ({
...
+                />
             </Popover>
         </FeatureResultPopoverWrapper>
     );
 };
+
+export const FeatureResultInfoPopoverCell = (
+    props: FeatureResultInfoPopoverCellProps,
+) => {
+    const useNewStrategyDesign = useUiFlag('flagOverviewRedesign');
```
thomasheartman, line 144: Split the two versions into entirely separate components. We can consider creating a new file too, but this feels like the root of the playground changes, so I don't mind doing the split in this file alone.

frontend/src/component/playground/Playground/PlaygroundResultsTable/FeatureResultInfoPopoverCell/FeatureStrategyList/StrategyList/StrategyItem/FeatureStrategyItem.tsx
thomasheartman, line 1: This one is interesting. It's essentially the same as `frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/StrategyItem.tsx`, but with a different execution section (to add all the extra playground stuff). We might be able to align them later, but I wanna leave that for now. Regardless, we use the same StrategyItemContainer, so this alredy feels pretty sensible to me.

frontend/src/component/playground/Playground/PlaygroundResultsTable/FeatureResultInfoPopoverCell/FeatureResultInfoPopoverCell.tsx
```diff
@@ -65,32 +60,96 @@ export const FeatureResultInfoPopoverCell = ({
...
+            <IconButton onClick={togglePopover}>
+                <InfoOutlined ref={ref} />
+            </IconButton>
+            <Popover
+                open={open}
+                onClose={() => setOpen(false)}
+                anchorEl={ref.current}
+                PaperProps={{
+                    sx: (theme) => ({
+                        '--popover-inline-padding': theme.spacing(4),
```
thomasheartman, line 104: We can't set the padding here because the strategy separators should expand all the way to the edges. Instead, we can set a CSS variable and access it in the downstream components that need it.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
Tymek: I'm not sure about this structure. We have syntax that is very different from how it visually looks like. 
```
- ol
  - li // release plan
  - li
    - ol
      - li // strategy
      - li
      - li
```
It's because strategies are "draggable", right? 
And `<Strategies />` in-place component makes this harder to read.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
Tymek: This looks like magic I will be afraid to touch later. If CSS selectors that need comments are too complex for what it does. I much prefer what legacy component does with JS (`releasePlans.length > 0 && strategies.length > 0`).

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -49,102 +57,101 @@ const StyledTruncator = styled(Truncator)(({ theme }) => ({
...
+                            onDragStart={onDragStart}
+                            onDragEnd={onDragEnd}
+                            sx={{ cursor: 'move' }}
+                        >
+                            <DragIndicator
+                                titleAccess='Drag to reorder'
+                                cursor='grab'
+                                sx={{ color: 'action.active' }}
+                            />
+                        </DragIcon>
```
Tymek, line 107: I think attempting to extract "draggable" feature from this into another container makes sense. It's a separate functionality and has a lot of complexity. Right now behind a flag it is broken for me on `main` and on this branch

Tymek (changes requested):

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
thomasheartman: Hmm. I see your point about CSS selectors, but I also don't think that we should abstain from using selectors when they are useful to solve a problem. 

There's a couple points here:
1. I don't think the markup/html should know anything about when it should apply borders. That's a styling thing and should be separate.
2. If we change that, then you still need to do this logic check in the code. It's just in more places and intermingled with other rendering logic.

So in this case, we'd have to check if the `index >= 1 || releasePlans.length > 0` in the list and then apply some attribute to the HTML that we then parse in the style code. But then we'd have to do the same thing in the release plan code and the milestone template code and the playground code. They all use the same list (on purpose; I'll extract it to shared component later), and using CSS selectors this way allows us to handle all of this once instead of over and over again.

So with that context, I think they're the right tool.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
thomasheartman: Yeah, this is a good point. It might. I did notice dragging was acting weird, but wasn't sure what was up. Considering this is already wonky on main, I'd rather merge this first and try to solve the dragging problem in a followup, though. Makes it easier to isolate the solution and get that merged.

Tymek (approved):

PR #9512, opened 2025-03-11
chore: show events IP

nunogois:
https://linear.app/unleash/issue/2-3392/visualize-event-log-ip-address-on-unleash-enterprise-ui

Shows events IP for Unleash Enterprise.

image


gastonfournier (approved):
It makes sense to me

PR #9518, opened 2025-03-12
Fix features table migrations in 20220603081324-add-archive-at-to-fea…

vickylance:
Changes to migration file - 20220603081324-add-archive-at-to-feature-toggle.js

1. Fixes the migration script where features table is updated with archived_at column with the latest date instead of taking the date from the events table. 
2. Also it fails for the latest toggle which was archived and then revived but after migration it updates the toggle as archived toggle.
3. Also because of the buggy reference to the outer join's events table `e` its taking a huge time to run the migration.

This PR fixes all the above issues.

## About the changes

We have faced an issue during migration of unleash-server from 4.8.2 to 6.6.0 where one of our toggle which was archived and unarchived once before migration and that was the latest toggle in terms of the archived date, but after the migration was ran it was in archived status. 

Upon further debugging and running the SQL command in the features table migration file we noticed that we should not be referencing the outer join's events table `e` for the  feature_name check and additionally we should add the features table's archived toggle check instead.

### Important files

The change in only in this file.

[20220603081324-add-archive-at-to-feature-toggle.js](https://github.com/Unleash/unleash/pull/9518/files#diff-b91c299b96edc46ca3a1963bf54966aa777c9fa107f3bd8b45f5fb54dc57460e)

## Discussion points

Let me know if any further details is required.

### DIFF
--- a/src/migrations/20220603081324-add-archive-at-to-feature-toggle.js
+++ b/src/migrations/20220603081324-add-archive-at-to-feature-toggle.js
@@ -13,7 +13,7 @@ exports.up = function (db, callback) {
                           AND e.created_at =
                               (SELECT Max(created_at) date
                                FROM   events
-                               WHERE  type = 'feature-archived'
+                               WHERE  e.type = 'feature-archived'
                                       AND e.feature_name = f.NAME)) res
         WHERE  res.NAME = f.NAME;
         UPDATE features
--- a/src/migrations/20220603081324-add-archive-at-to-feature-toggle.js
+++ b/src/migrations/20220603081324-add-archive-at-to-feature-toggle.js
@@ -13,8 +13,9 @@ exports.up = function (db, callback) {
                           AND e.created_at =
                               (SELECT Max(created_at) date
                                FROM   events
-                               WHERE  e.type = 'feature-archived'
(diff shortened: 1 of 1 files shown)


vickylance:
@gastonfournier   kind review the above PR

gastonfournier:
Hi @vickylance! Thanks for the contribution! I see this migration applies because you're migrating from 4.8 and this was introduced in 4.14. I was curious why you were changing such an old migration. I think this makes sense as this won't affect users on newer versions.

If you have a lot of flags that are archived, I'm not sure this will help you though... I'm not sure I can **easily** a patch of 4.8 or 4.14 so you can do a minor upgrade just with this migration before you upgrade to 6.6. 

Also, make sure you pay attention to these migration upgrade notes: https://docs.getunleash.io/using-unleash/deploy/upgrading-unleash#upgrading-to-57-and-later-from-versions--5611 since you're 2 major versions behind.

gastonfournier (approved):

gastonfournier:
> If you have a lot of flags that are archived, I'm not sure this will help you though... 

Ok, I checked again the code and it will help

vickylance:
Thank you

PR #9523, opened 2025-03-12
feat: frontend consumption table

sjaanus:
Similar to backend traffic, we have table for frontend traffic. We might want to rename the backend on to align with this one.

### DIFF
--- a/src/migrations/20250312162155-connection-count-consumption-table.js
+++ b/src/migrations/20250312162155-connection-count-consumption-table.js
@@ -2,7 +2,7 @@
 exports.up = (db, callback) => {
     db.runSql(
         `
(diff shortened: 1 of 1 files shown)


src/migrations/20250312162155-connection-count-consumption-table.js
```diff
@@ -0,0 +1,23 @@
+
+exports.up = (db, callback) => {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS frontend_consumption_count(
```
kwasniew, line 5: I called the other one connection_count_consumption so that it's not tied to backend/frontend but to connection count. This way we can change our mind when we use connections.

I'd call this one request_count_consumption so that it's related to request counting and not to frontend or backend

kwasniew (approved):

PR #9525, opened 2025-03-12
chore: users actions menu

nunogois:
https://linear.app/unleash/issue/2-3342/new-entrance-point-create-dot-dot-dot-menu-instead-of-icons

Adds a new users actions menu.

Should this change be behind a flag? I'm leaning towards no, but if you think otherwise let me know.

### Previous
image

### After
image

### If user is SCIM-managed
image

### DIFF
--- a/frontend/src/component/admin/users/UsersList/UsersActionsCell/UsersActionsCell.tsx
+++ b/frontend/src/component/admin/users/UsersList/UsersActionsCell/UsersActionsCell.tsx
@@ -9,8 +9,6 @@ import {
     Tooltip,
     Typography,
 } from '@mui/material';
-import { PermissionHOC } from 'component/common/PermissionHOC/PermissionHOC';
-import { ADMIN } from 'component/providers/AccessProvider/permissions';
 import { useState } from 'react';
 
 const StyledActions = styled('div')(({ theme }) => ({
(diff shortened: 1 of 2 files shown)


frontend/src/component/admin/users/UsersList/UsersActionsCell/UsersActionsCell.tsx
```diff
@@ -1,101 +1,173 @@
...
+    onClick,
+    permission,
+    isScimUser,
+    children,
+}: IUserActionProps) => {
+    const scimTooltip =
+        'This user is managed by your SCIM provider and cannot be changed manually';
+
+    return (
+        <PermissionHOC permission={permission}>
```
nunogois, line 150: I noticed the users page is only accessible to ADMINs. That means we can simplify this further.

nunogois: Addressed in https://github.com/Unleash/unleash/pull/9525/commits/2cee099b70b42afdd2f9a02ae58b99d075783d23

gastonfournier (approved):
I like it, I think it changes the visuals but it's not a big change that'd require a feature flag. :+1: from my side

PR #9527, opened 2025-03-12
refactor: remove unused access overview component

nunogois:
https://linear.app/unleash/issue/2-3396/remove-unused-access-overview-component

Removes an unused access overview component.

A nice bonus is that we're adding something called [Access Overview](https://github.com/Unleash/unleash/pull/9525) so we avoid some confusion.


daveleek (approved):
Yup, looks unused!

PR #9530, opened 2025-03-13
fix: Data usage graphs don't work in UTC-n time zones

thomasheartman:
Fixes a number of issues that would surface in UTC-n (where n > 1) timezones. I've not found a way to check this with tests (and it looks like [we weren't able to last time either](https://github.com/Unleash/unleash/pull/9110/files#r1919746328)), so all the testing's been done manually by adjusting my system time and zone. (From what I understand, you can't generate a Date with a specific TZ offset in JS: it's only utc or local time)

Resolved:
- [x] Selecting "Jan" in the dropdown results in the selection being "December" (off by one in the selector)

- [x] Selecting a month view only gives you one data point (and it's probably empty). Wrong date parsing on the way out resulted in sending `{ from: "2025-02-28", to: "2025-02-28"}` instead of `{ from: "2025-03-01", to: "2025-03-31"}`

- [x] The dates we create when making "daysRec" need to be adjusted. They showed the wrong month, so the dates were off.

- [x] Make sure the labels are correct when hovering over. Again: we used the wrong month for generating these.

- [x] The available months are wrong. Incorrect month parsing again.

- [x] The request summary month is wrong. You guessed it: incorrect month parsing


daveleek (approved):
LG

PR #9531, opened 2025-03-13
chore: rename access matrix to access overview

nunogois:
https://linear.app/unleash/issue/2-3344/new-name-access-overview-instead-of-access-matrix

Renames Access Matrix to Access Overview, both internally (code) and externally (UI).

### DIFF
--- a/src/lib/routes/admin-api/user-admin.ts
+++ b/src/lib/routes/admin-api/user-admin.ts
@@ -741,7 +741,7 @@ export default class UserAdminController extends Controller {
                 ),
             );
         }
-        const matrix = await this.accessService.permissionsMatrixForUser(
+        const overview = await this.accessService.getAccessOverviewForUser(
             user,
             project,
             environment,
@@ -749,7 +749,7 @@ export default class UserAdminController extends Controller {
 
         // TODO add response validation based on the schema
         res.status(200).json({
-            matrix,
+            overview,
             user,
             rootRole,
             projectRoles,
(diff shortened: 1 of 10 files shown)


src/lib/services/access-service.ts
```diff
@@ -244,14 +244,14 @@ export class AccessService {
      * Provided a project, project permissions will be checked against that project.
      * Provided an environment, environment permissions will be checked against that environment (and project).
      */
-    async permissionsMatrixForUser(
+    async getAccessOverviewForUser(
```
nunogois, line 247: This change breaks Enterprise temporarily. Should be followed up with the Enterprise PR fixing it.

src/lib/routes/admin-api/user-admin.ts
```diff
@@ -741,15 +741,15 @@ export default class UserAdminController extends Controller {
...
+        const overview = await this.accessService.getAccessOverviewForUser(
             user,
             project,
             environment,
         );
 
         // TODO add response validation based on the schema
         res.status(200).json({
-            matrix,
+            overview,
```
gastonfournier, line 752: Are we changing an API schema? Is it public? Just watching out for potential breaking changes

gastonfournier: Ok, maybe the matrix field is not "exposed" in our public schema. Probably done intentionally... actually the response body is not declared:

image

gastonfournier (approved):
Looks good, but maybe we should also consider a task to add the response body to the schema (and maybe label it as experimental or something is possible)

src/lib/routes/admin-api/user-admin.ts
nunogois: Good point, I'll add a task for this!

PR #9532, opened 2025-03-13
Chore(1-3390)/playground strategy execution: constraints

thomasheartman:
Implements the new design for playground constraints. They're not in use in segments yet, and strategy parameters have not been touched. This PR establishes a pattern that we can follow for strategies and parameters later.

image

The PR also includes a change in how the constraint item organizes its children: it now takes care adding padding and spacing itself, instead of the children doing that. It looks right most places, but segments aren't quite right anymore. However, as this is behind a flag, I'd rather fix that in a separate PR.


frontend/src/component/playground/Playground/PlaygroundResultsTable/FeatureResultInfoPopoverCell/FeatureStrategyList/StrategyList/StrategyItem/StrategyExecution/ConstraintExecution/ConstraintExecution.tsx
```diff
@@ -1,59 +1,78 @@
...
+    gap: theme.spacing(1),
+    color:
+        variant === 'ok'
+            ? theme.palette.success.dark
+            : theme.palette.error.dark,
+
+    fontSize: theme.typography.body2.fontSize,
+    svg: {
+        // todo: how best to size this?
+        fontSize: `calc(${theme.typography.body1.fontSize} * 1.1)`,
```
Tymek, line 30: `fontSize: theme.spacing(...)` ?

PR #9533, opened 2025-03-13
chore: clean up frontend routes

daveleek:
This PR aims to clean up a little in our frontend routing (mostly unused stuff)

- The old mobile menu is gone, the new one doesn't use the mobile: true attributes in the IRoute.menu
- admin links are not sourced from routes.ts anymore - so formatting that and returning a collection for it is no longer neccessary
- DrawerMenu doesn't use the routes provided to it - a sub component resolves its own routes from useRoutes. Removing that parameter and the providing of it in Header
- There's no 'advanced' menu anymore - the main menu is populated with items marked as menu.advanced: true however - so changing that to be menu.main: true.
- routes.ts getRoutes renamed to getNavRoutes and changed to just returning the routes with menu.main: true

### DIFF
--- a/frontend/src/component/menu/Header/DrawerMenu/DrawerMenu.tsx
+++ b/frontend/src/component/menu/Header/DrawerMenu/DrawerMenu.tsx
@@ -3,7 +3,6 @@ import { Link } from 'react-router-dom';
 import { Divider, Drawer, styled } from '@mui/material';
 import { ReactComponent as UnleashLogo } from 'assets/img/logoDarkWithText.svg';
 import { ReactComponent as UnleashLogoWhite } from 'assets/img/logoWithWhiteText.svg';
-import type { INavigationMenuItem } from 'interfaces/route';
 import styles from './DrawerMenu.module.scss'; // FIXME: useStyle - theme
 import theme from 'themes/theme';
 import { ThemeMode } from 'component/common/ThemeMode/ThemeMode';
@@ -32,18 +31,12 @@ interface IDrawerMenuProps {
         href: string;
         title: string;
     }>;
-    routes: {
-        mainNavRoutes: INavigationMenuItem[];
-        mobileRoutes: INavigationMenuItem[];
-        adminRoutes: INavigationMenuItem[];
-    };
 }
 
 export const DrawerMenu: VFC<IDrawerMenuProps> = ({
     links = [],
     open = false,
     toggleDrawer,
-    routes,
 }) => {
     return (
         <Drawer
--- a/frontend/src/component/menu/Header/Header.tsx
+++ b/frontend/src/component/menu/Header/Header.tsx
@@ -21,13 +21,12 @@ import { ReactComponent as CelebatoryUnleashLogo } from 'assets/img/unleashHolid
 import { DrawerMenu } from './DrawerMenu/DrawerMenu';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 import { focusable } from 'themes/themeStyles';
-import { getCondensedRoutes, getRoutes } from 'component/menu/routes';
+import {} from 'component/menu/routes';
 import DarkModeOutlined from '@mui/icons-material/DarkModeOutlined';
 import LightModeOutlined from '@mui/icons-material/LightModeOutlined';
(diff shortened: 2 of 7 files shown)


frontend/src/interfaces/route.ts
```diff
@@ -29,8 +29,7 @@ export interface INavigationMenuItem {
 }
 
 interface IRouteMenu {
-    mobile?: boolean;
-    advanced?: boolean;
+    main?: boolean;
```
daveleek, line 32: There's not 'advanced' menu anymore. Renaming to 'main' - as that's where it ends up in downstream properties

frontend/src/component/menu/routes.ts
```diff
@@ -549,4 +532,7 @@ export const getCondensedRoutes = (routes: IRoute[]): INavigationMenuItem[] => {
     });
 };
 
-export const getRoutes = computeRoutes();
+export const baseRoutes = routes.filter((route) => !route.hidden);
+export const getNavRoutes = () => {
+    return getCondensedRoutes(baseRoutes.filter((route) => route.menu.main));
```
daveleek, line 537: Might as well export them condensed out of the box - there's only one reference to this and that also imported the getCondensed and applied it to the returned collection

frontend/src/component/menu/routes.ts
```diff
@@ -518,24 +518,7 @@ export const routes: IRoute[] = [
 export const getRoute = (path: string) =>
     routes.find((route) => route.path === path);
 
-export const baseRoutes = routes.filter((route) => !route.hidden);
-
-const computeRoutes = () => {
-    const mainNavRoutes = baseRoutes.filter((route) => route.menu.advanced);
-    const adminRoutes = routes.filter((route) => route.menu.adminSettings);
-    const mobileRoutes = routes.filter((route) => route.menu.mobile);
```
daveleek, line 526: Out of all these - only the mainNavRoutes were being used (mobile no longer in use, and admin has it's own adminRoutes.ts). No point returning this object

frontend/src/component/menu/Header/Header.tsx
```diff
@@ -21,13 +21,12 @@ import { ReactComponent as CelebatoryUnleashLogo } from 'assets/img/unleashHolid
 import { DrawerMenu } from './DrawerMenu/DrawerMenu';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 import { focusable } from 'themes/themeStyles';
-import { getCondensedRoutes, getRoutes } from 'component/menu/routes';
+import {} from 'component/menu/routes';
```
daveleek, line 24: oops - I'll clean these two up

frontend/src/component/menu/Header/Header.tsx
```diff
@@ -105,19 +104,6 @@ const Header = () => {
...
-    const filteredMainRoutes = {
-        mainNavRoutes: getCondensedRoutes(routes.mainNavRoutes)
-            .filter(filterByConfig(uiConfig))
-            .map(mapRouteLink),
-        mobileRoutes: getCondensedRoutes(routes.mobileRoutes)
-            .filter(filterByConfig(uiConfig))
-            .map(mapRouteLink),
-        adminRoutes,
-    };
-
```
daveleek, line 120: None of this was being used in the component anyways, a sub component was fetching the same thing from useRoutes

frontend/src/component/menu/Header/DrawerMenu/DrawerMenu.tsx
```diff
@@ -32,18 +31,12 @@ interface IDrawerMenuProps {
...
-        mobileRoutes: INavigationMenuItem[];
-        adminRoutes: INavigationMenuItem[];
-    };
 }
 
 export const DrawerMenu: VFC<IDrawerMenuProps> = ({
     links = [],
     open = false,
     toggleDrawer,
-    routes,
```
daveleek, line 46: unused parameter

Tymek (approved):
I bet this can be simplified further. `INavigationMenuItem['menu']` can be made optional. And further, if we have a split between admin menu and main nav menu, it makes sense to get rid of `adminSettings` prop and collapse `route.menu<{}>` into `route.showInNavigation<boolean>`. But that's for later. Great job in cleaning this up!

PR #9535, opened 2025-03-13
chore: change access overview to lists in accordions

nunogois:
https://linear.app/unleash/issue/2-3343/accordions-not-a-must-have
https://linear.app/unleash/issue/2-3345/indicator-of-how-many-permissions

Changes our Access Overview from tables to lists in accordions. Also includes the total permissions in the accordion summary.

Looking at the designs it seems like lists would make the most sense, both visually and in terms of semantics. This will also allow us to group the permissions both visually and semantically in a future task.

image

### Update

Also improved our project permissions label.

image

image


frontend/src/component/admin/users/AccessOverview/AccessOverview.tsx
```diff
@@ -104,19 +106,24 @@ export const AccessOverview = () => {
...
-            <AccessOverviewTable permissions={overview?.environment ?? []} />
+            <StyledAccessOverviewContainer>
+                <AccessOverviewAccordion permissions={overview?.root ?? []}>
+                    Root permissions for role {rootRole?.name}
+                </AccessOverviewAccordion>
+                <AccessOverviewAccordion permissions={overview?.project ?? []}>
+                    Project permissions
+                    {project
+                        ? ` for project ${project}${projectRoles?.length ? ` with project role${projectRoles.length !== 1 ? 's' : ''} ${projectRoles?.map((role: any) => role.name).join(', ')}` : ''}`
+                        : ''}
```
gastonfournier, line 117: I really like this, but maybe lowercase? It looks weird to have some camel case some lower case, but on the flip side the name is the name... so it might be ok.
```suggestion
                    {project
                        ? ` for project ${project}${projectRoles?.length ? ` with project role${projectRoles.length !== 1 ? 's' : ''} ${projectRoles?.map((role: any) => role.name.lowercase()).join(', ')}` : ''}`
                        : ''}
```

frontend/src/component/admin/users/AccessOverview/AccessOverview.tsx
```diff
@@ -104,19 +106,24 @@ export const AccessOverview = () => {
...
+                </AccessOverviewAccordion>
+                {environment && (
+                    <AccessOverviewAccordion
+                        permissions={overview?.environment ?? []}
+                    >
+                        Environment permissions for environment {environment}
```
gastonfournier, line 123: Maybe less redundant
```suggestion
                        Environment permissions for {environment}
```

gastonfournier (approved):
UI looks great!

frontend/src/component/admin/users/AccessOverview/AccessOverview.tsx
nunogois: I'm not sure I agree. We're not doing the same with the project and environment. If you have roles like "Role A", "Role B", "Role C", or some other initials, it might look weird.

If you'd like a better visually hierarchal separation we can try adding a colon (:). Something like:
"with project roles: Role A, Role B, Role C".

PR #9544, opened 2025-03-14
chore: implement design for disabled environments in playground

thomasheartman:
Adds the new design for strategy lists in disabled environments.

image


Tymek (approved):
Please refactor out nested a ternary before merging.

PR #9545, opened 2025-03-14
fix: use the correct permission for the create template button

daveleek:
We were using the wrong permission in the create template permission button


chriswk (approved):
Do we use the @server package other places for permissions?

nunogois (approved):

daveleek:
@chriswk yes we do :)

PR #9546, opened 2025-03-14
chore: categorize access overview

nunogois:
https://linear.app/unleash/issue/2-3346/only-show-description-and-not-permission-first-column-and-group

Categorizes the permissions in our Access Overview.

image


gastonfournier (approved):
Looks good!

PR #9548, opened 2025-03-16
Feat(UI): add ability to collapse/expand strategies

### LINKED ISSUE #9529: Allow to collapse strategy

I would like the ability to collapse strategy in the UI. Additionally the ability to "Collapse all" and "Expand all".

Happy to also contribute if the feature request gets approved

### Background

We tend to have several strategies for a feature flag. The highest I have seen is 29 for an experiment flag. But there are several other flags where we have somewhere between 10-15 strategies. This might be largely related to how my company operates and how we run AB tests (where we need to keep old strategies to allow the users to age).

### Solution suggestions

1. For each strategy there is a way to collapse and expand it.
2. Somewhere in the top bar in each environment, there is an option to collapse all and expand all.

Not something to be done for v1, but for v2 we can also consider storing the collapse state for the user in localStorage. So that the next time they open the flag, the collapse state for each strategy can be restored.

wuguishifu:
## About the changes

- Added a button to collapse/expand the targeting section of the strategies.
- When collapsed, the rollout percentage and variants are shown.

### Expanded (default):
Screenshot 2025-03-15 at 10 51 58 PM

### Collapsed:
Screenshot 2025-03-15 at 10 52 06 PM

Closes #9529 

### Important files

- The main functionality is located in the directory `/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/CollapseStrategyIcon`

## Discussion points

- Right now it seems like for environments the expanded/collapsed state isn't cached at all so refreshing the page resets the state. I followed this behavior for this PR as well, but maybe instead the state should be stored in cookies so it persists across sessions.

### DIFF
--- a/frontend/src/component/common/ConstraintsList/StrategyEvaluationItem/StrategyEvaluationItem.tsx
+++ b/frontend/src/component/common/ConstraintsList/StrategyEvaluationItem/StrategyEvaluationItem.tsx
@@ -5,14 +5,17 @@ type StrategyItemProps = {
     type?: ReactNode;
     children?: ReactNode;
     values?: string[];
+    reduceMargin?: boolean;
 };
 
-const StyledContainer = styled('div')(({ theme }) => ({
+const StyledContainer = styled('div', {
+    shouldForwardProp: prop => prop !== 'reduceMargin',
+})<{ reduceMargin: boolean }>(({ theme, reduceMargin }) => ({
     display: 'flex',
     gap: theme.spacing(1),
     alignItems: 'center',
     fontSize: theme.typography.body2.fontSize,
-    margin: theme.spacing(2, 3),
+    margin: reduceMargin ? 0 : theme.spacing(2, 3),
 }));
 
 const StyledContent = styled('div')(({ theme }) => ({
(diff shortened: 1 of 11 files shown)


thomasheartman:
Hi, @wuguishifu 👋🏼  Thanks for the contribution 😄  As was mentioned in the [issue you linked](https://github.com/Unleash/unleash/issues/9529), we've registered this as a feature request and are talking about it internally at the moment. 

As you might have noticed, we're working on a pretty big strategy UI update right now, and this falls squarely in the middle of it (and also aligns with our stated goals of the updated interface). Because we're right in the middle of it, we'd like to take this chance to have our UX team think about this fully (what to show / hide, consequences of showing/hiding, tradeoffs, etc) before we implement it. That also means that we're not going to merge this right now.

But to help us understand what the need for this is, it would be interesting to hear *your* take on why you want this functionality. Could you tell me what problems this solves for you?

Anyway, because we are right in the middle of an update of the strategy rendering code, any files you touch there are likely to cause conflicts, so to keep us all from going bananas with merge conflicts, I'd recommend waiting a little bit before updating this PR anyway 😅

PR #9550, opened 2025-03-17
feat: requests consumption UI for frontend

sjaanus:
Switching frontend traffic tab to use the requests consumption API:

image


website/sidebars.ts
```diff
@@ -135,7 +135,7 @@ const sidebars: SidebarsConfig = {
                                     type: 'doc',
                                     label: 'Security and Compliance',
                                     id: 'feature-flag-tutorials/use-cases/security-compliance',
-                                }
+                                },
```
sjaanus, line 138: This seems to be somekind of linter.

PR #9552, opened 2025-03-17
chore: add user access overview schema

nunogois:
https://linear.app/unleash/issue/2-3403/add-response-schema-for-access-overview

Adds a response schema for the user access overview.


gastonfournier (approved):
LG! Should we move this to experimental? Rather than Auth? I think it's not a big problem cause we've been using it for a long time... it's just that if under experimental we'd be able to modify it without breaking compatibility

PR #9556, opened 2025-03-18
feat: rework navigation sidebar admin section

daveleek:
This PR moves the admin section of the navigation sidebar into a separate component.
In that component we check if the flag is enabled, if so we change what we show in the menu to be a single link to admin settings. 

We also change the MainLayout where we remove the whole menu if the flag is enabled **and** we're visiting a /admin route. That's because we'll add a separate menu in a slightly different location later for just the menu sub section

image
image

### DIFF
--- a/frontend/src/component/layout/MainLayout/MainLayout.tsx
+++ b/frontend/src/component/layout/MainLayout/MainLayout.tsx
@@ -102,9 +102,10 @@ export const MainLayout = forwardRef<HTMLDivElement, IMainLayoutProps>(
 
         const theme = useTheme();
         const isSmallScreen = useMediaQuery(theme.breakpoints.down('lg'));
-        const hideAdmin =
+        const showOnlyAdminMenu =
             newAdminUIEnabled && location.pathname.indexOf('/admin') === 0;
-        const showRegularNavigationSideBar = !isSmallScreen && !hideAdmin;
+        const showRegularNavigationSideBar =
+            !isSmallScreen && !showOnlyAdminMenu;
 
         return (
             <EventTimelineProvider>
--- a/frontend/src/component/layout/MainLayout/MainLayout.tsx
+++ b/frontend/src/component/layout/MainLayout/MainLayout.tsx
@@ -1,5 +1,6 @@
 import { forwardRef, type ReactNode } from 'react';
 import { Box, Grid, styled, useMediaQuery, useTheme } from '@mui/material';
+import { useLocation } from 'react-router-dom';
 import Header from 'component/menu/Header/Header';
 import Footer from 'component/menu/Footer/Footer';
 import Proclamation from 'component/common/Proclamation/Proclamation';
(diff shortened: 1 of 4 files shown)


frontend/src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -90,14 +92,19 @@ const MainLayoutContentContainer = styled('main')(({ theme }) => ({
...
+        const location = useLocation();
         const projectId = useOptionalPathParam('projectId');
         const { isChangeRequestConfiguredInAnyEnv } = useChangeRequestsEnabled(
             projectId || '',
         );
 
         const theme = useTheme();
         const isSmallScreen = useMediaQuery(theme.breakpoints.down('lg'));
+        const hideAdmin =
+            newAdminUIEnabled && location.pathname.indexOf('/admin') === 0;
```
daveleek, line 106: Needs some testing on sandbox, but we use a similar pattern elsewhere and since we're setting the basename in the root this should behave as expected here

frontend/src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -90,14 +92,20 @@ const MainLayoutContentContainer = styled('main')(({ theme }) => ({
...
+        const location = useLocation();
         const projectId = useOptionalPathParam('projectId');
         const { isChangeRequestConfiguredInAnyEnv } = useChangeRequestsEnabled(
             projectId || '',
         );
 
         const theme = useTheme();
         const isSmallScreen = useMediaQuery(theme.breakpoints.down('lg'));
+        const showOnlyAdminMenu =
+            newAdminUIEnabled && location.pathname.indexOf('/admin') === 0;
```
kwasniew, line 106: will it break with a route to feature flag named admin?

daveleek: it shouldn't, but I decided to test:
image

src/server-dev.ts
```diff
@@ -58,7 +58,7 @@ process.nextTick(async () => {
                         filterExistingFlagNames: true,
                         teamsIntegrationChangeRequests: true,
                         simplifyDisableFeature: true,
-                        adminNavUI: true,
+                        adminNavUI: false,
```
nunogois, line 61: I wonder why we're turning this off?

nunogois (approved):
LGTM

frontend/src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -90,14 +92,20 @@ const MainLayoutContentContainer = styled('main')(({ theme }) => ({
...
         const { uiConfig } = useUiConfig();
+        const location = useLocation();
         const projectId = useOptionalPathParam('projectId');
         const { isChangeRequestConfiguredInAnyEnv } = useChangeRequestsEnabled(
             projectId || '',
         );
 
         const theme = useTheme();
         const isSmallScreen = useMediaQuery(theme.breakpoints.down('lg'));
+        const showOnlyAdminMenu =
```
thomasheartman, line 105: Wait, so how do you navigate out of the admin menu if you've landed there? If the other nav is gone, ... how do you get out?

src/server-dev.ts
daveleek: As @thomasheartman comments further down - currently there's no menu-way of getting back out. So I didn't want it to break life for us during dev, but I also want to break this up into reviewable PRs

frontend/src/component/layout/MainLayout/MainLayout.tsx
daveleek: Right now - you'd have to edit URL or through command bar. It's turned off for devs for the time being, eventually there will be an admin menu that looks a little different (and also has a link back to dashboard)

thomasheartman: Got it. Seems weird to trap users there, but the link back should be good.

PR #9557, opened 2025-03-18
chore: fix missing primary routes for commandbar

daveleek:
Cleaning up mobile routes also removed the primary routes (what we call them in navigation sidebar) to be unresolvable, resulting in:
image

This PR fixes this by adding a menu property "primary?: boolean", and some code mapping matching items into a route collection for command bar. The collection should be reusable for menu-rendering in the future if we choose to
image


gastonfournier (approved):
LRTM and it fixes a bug so :+1:

PR #9567, opened 2025-03-18
feat: new constraint view for flag edit page

Tymek:
## About the changes
Refactor components in Targeting (Edit strategy)
Before:
image

After:
image


frontend/src/component/common/NewConstraintAccordion/NewConstraintAccordionList/NewConstraintAccordionList.tsx
```diff
@@ -43,30 +42,6 @@ const StyledContainer = styled('div')({
...
-    color: theme.palette.text.secondary,
-}));
-
-const StyledAddCustomLabel = styled('div')(({ theme }) => ({
-    marginTop: theme.spacing(1),
-    marginBottom: theme.spacing(1),
-    color: theme.palette.text.primary,
-    display: 'flex',
-}));
-
```
Tymek, line 69: dead code

frontend/src/component/common/ConstraintsList/ConstraintsList.tsx
```diff
@@ -10,17 +10,21 @@ const StyledList = styled('ul')(({ theme }) => ({
...
     border: `1px solid ${theme.palette.divider}`,
     borderRadius: theme.shape.borderRadiusMedium,
     background: theme.palette.background.default,
-    padding: theme.spacing(2, 3),
+    padding: theme.spacing(1.5, 3),
     display: 'flex',
     flexFlow: 'column',
-    gap: theme.spacing(2),
+    gap: theme.spacing(1),
 }));
```
Tymek, line 22: Decorating list items with container block (border, padding, background) automatically was a bad idea. It's easier to be more explicit where and how it is used.

frontend/src/component/common/ConstraintsList/StrategyEvaluationItem/StrategyEvaluationItem.tsx
```diff
@@ -35,45 +39,33 @@ const StyledType = styled('span')(({ theme }) => ({
...
+                >
+                    {values[0]}
+                </Truncator>
+            ) : null}
+            {values && values?.length > 1 ? (
+                <Truncator title='' lines={2} onSetTruncated={onSetTruncated}>
+                    {values.join(', ')}
+                </Truncator>
             ) : null}
         </StyledContent>
```
Tymek, line 70: I didn't find any reliable way of truncating many long values when each is wrapped in it's own chip. It's far easier to do it for strings.

thomasheartman: Ah, mm. Right. I see what you mean. At the same time, this makes the structure less clear to screen readers etc. It'd be nice to retain it as a list if possible, but I'm not gonna halt the PR for it. When it's not a list, there's no indication in the rendered html that it's not just one long value with commas in it. If it's a list, you'll typically have screen readers announce how many items it contains and then go through them in order (as far as I understand).

So what is the actual issue here? What are you truncating? "all the items"? I'd think the list itself has an intrinsic width that you could check? Probably?  I'd think the truncator handles that, but maybe it don't 🤷🏼

frontend/src/component/common/SegmentItem/SegmentItem.tsx
```diff
@@ -32,16 +39,14 @@ const StyledAccordion = styled(Accordion)(({ theme }) => ({
...
     fontSize: theme.typography.body2.fontSize,
     minHeight: 'unset',
-    '.MuiAccordionSummary-content, .MuiAccordionSummary-content.Mui-expanded': {
-        margin: 0,
-    },
 }));
 
 const StyledAccordionDetails = styled(AccordionDetails)(({ theme }) => ({
-    padding: theme.spacing(2, 0, 1),
+    borderTop: `1px dashed ${theme.palette.divider}`,
```
thomasheartman, line 48: Ooh, a dashed border? That's ... dashing? I don't think I've seen that in sketches before? Is it a new style we're trying out?

frontend/src/component/common/Truncator/Truncator.tsx
```diff
@@ -59,6 +60,10 @@ export const Truncator = ({
         return () => resizeObserver.disconnect();
     }, [title, children]);
 
+    useEffect(() => {
+        onSetTruncated?.(isTruncated);
```
thomasheartman, line 64: Is that optional chaining gonna work? I thought you had to do `onSetTruncated && onSetTruncated(args)`? Also, what exactly do you need this one for? There's a lot of things being passed everywhere; I don't remember exactly what it was now

thomasheartman (approved):

frontend/src/component/common/SegmentItem/SegmentItem.tsx
Tymek: We have it in context. I think it looks neat

frontend/src/component/common/Truncator/Truncator.tsx
Tymek: That's for "should this list be expandable"

frontend/src/component/common/ConstraintsList/StrategyEvaluationItem/StrategyEvaluationItem.tsx
Tymek: Valid point. I'll experiment I there is an approach that will be the best of both words.

PR #9579, opened 2025-03-19
chore: finish up disabled strategies

thomasheartman:
Aligns the design of disabled strategies with the sketches. Most notable changes:
- makes the disabled badge warning yellow
- greys out the preceding "or" separator
- makes the segment "preview" button *not* grey (because it's still interactable)

As a bonus: uses a list for the constraint value lists instead of a div and updates the design to match the sketches (no chips). 

image

With strat variants:
image

Bonus fix: 
Lets the constraint value list wrap so that we avoid this kind of blowout:
image

Instead: 
image

### DIFF
--- a/frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
+++ b/frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
@@ -8,6 +8,7 @@ import type { PlaygroundStrategySchema } from 'openapi';
 import { Badge } from '../Badge/Badge';
 import { Link } from 'react-router-dom';
 import { Truncator } from '../Truncator/Truncator';
+import { disabledStrategyClassName } from './disabled-strategy-utils';
 
 type StrategyItemContainerProps = {
     strategyHeaderLevel?: 1 | 2 | 3 | 4 | 5 | 6;
@@ -87,7 +88,10 @@ export const StrategyItemContainer: FC<StrategyItemContainerProps> = ({
             : ({ children }) => <> {children} </>;
 
     return (
-        <Box sx={{ position: 'relative' }}>
+        <Box
+            className={strategy.disabled ? disabledStrategyClassName : ''}
+            sx={{ position: 'relative' }}
+        >
             <StyledContainer style={style} className={className}>
                 <StyledHeader disabled={Boolean(strategy?.disabled)}>
                     {onDragStart ? (
--- a/frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
+++ b/frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
@@ -136,9 +136,6 @@ export const StrategyItemContainer: FC<StrategyItemContainerProps> = ({
                             </StyledHeaderContainer>
                         </StrategyHeaderLink>
 
-                        {strategy.disabled ? (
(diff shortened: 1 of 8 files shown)


frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -87,7 +88,10 @@ export const StrategyItemContainer: FC<StrategyItemContainerProps> = ({
             : ({ children }) => <> {children} </>;
 
     return (
-        <Box sx={{ position: 'relative' }}>
+        <Box
+            className={strategy.disabled ? disabledStrategyClassName : ''}
```
thomasheartman, line 92: This lets us easily do all kinds of style tweaks to disabled strategies without having to pass around "disabled" props everywhere.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -77,6 +78,13 @@ const StyledAlert = styled(Alert)(({ theme }) => ({
     marginInline: theme.spacing(2), // should consider finding a variable here
 }));
 
+const StyledStrategySeparator = styled(StrategySeparator)(({ theme }) => ({
+    [`&:has(+ *:not(ol) .${disabledStrategyClassName}), &:has(+ ol > li:first-of-type .${disabledStrategyClassName})`]:
```
thomasheartman, line 82: Yeah, this one's kinda nasty. I've been coming around to the suggestion that we don't need to nest the strategy list if you have milestones, which would make this into the more manageable `&:has(+ * .${disabledStrategyClassName})`. But I'll do that in a separate PR.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -77,6 +78,13 @@ const StyledAlert = styled(Alert)(({ theme }) => ({
...
+        {
+            filter: 'grayscale(1)',
```
Tymek, line 84: previously it was `grayscale` + `opacity` tweak

Tymek (approved):

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
thomasheartman: Yes, but that was for the general contsraints (not the separator). I've removed the opacity tweak for now because it looked fine without it to my eyes and there's less of a risk of introducing contrast issues. If UX has any issues with it, we can always change it later

PR #9580, opened 2025-03-19
feat(docs): Added ADR for logging levels

chriswk:
This PR starts an ADR for logging levels, and how to make your coworkers' days who are on-call less noisy.

This is my initial suggestion. Please feel free to help me with fleshing out both the table for what log levels means to you or if you disagree with what I've put as a background.

My impression is that we are pretty good at this, but we could be better, and having an ADR to lean on when pushing back on PR's log levels could help us. 

Thanks to @kwasniew for the push to start the ADR.

Examples: 
* https://github.com/Unleash/unleash/blob/038c10f6125c4cce200a0bf49f38c7bddada7093/src/lib/db/account-store.ts#L164-L173

And probably the storeFootprint method in our client-feature-toggle-delta
* https://github.com/Unleash/unleash/blob/038c10f6125c4cce200a0bf49f38c7bddada7093/src/lib/features/client-feature-toggles/delta/client-feature-toggle-delta.ts#L363-L370

* https://github.com/bricks-software/unleash-enterprise/blob/293304a5d67231d584c3fa4c28589af23fb395e3/src/traffic-data/traffic-data-usage-service.ts#L69

### DIFF
--- a/website/docs/contributing/ADRs/overarching/logging-levels.md
+++ b/website/docs/contributing/ADRs/overarching/logging-levels.md
@@ -72,4 +72,4 @@ async markSeenAt(secrets: string[]): Promise<void> {
         }
     }
 ```
-Not being able to update lastSeen is not something we can anything about as on-calls, so this is also a good candidate for downgrading to warn
+Not being able to update lastSeen is not something we can do anything about as on-calls, so this is also a good candidate for downgrading to WARN.
--- a/website/docs/contributing/ADRs/overarching/logging-levels.md
+++ b/website/docs/contributing/ADRs/overarching/logging-levels.md
@@ -16,7 +16,7 @@ We should agree on the semantic information carried in each level, and which lev
 
 Current suggestion
 
-| Log level | Frequency in healthy application | Standard Availability                     | Configurable                |   
+| Log level | Frequency in healthy application | Standard Availability                     | Configurable                |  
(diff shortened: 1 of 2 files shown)


website/docs/contributing/ADRs/overarching/logging-levels.md
```diff
@@ -0,0 +1,35 @@
...
+| INFO      | 10-100                           | Default deploy config sets LOG_LEVEL=info | YES                         |
+| DEBUG     | 100 - 1000                       | Local development                         | YES (specific deployments)  |
+| TRACE     | 1000 - 10000                     | NO                                        | YES (specific deployments)  |
+
+
+
+
+### Change
+
+Previously we might've logged an ERROR for a self-healable issue, this should change to WARN, and not be an ERROR.
```
kwasniew, line 32: would be nice to have links to those examples in the PR description so that we can learn from the actual examples

chriswk: Good point. I'll try to find them

chriswk: Added a couple of examples both to the description and the Change part of the ADR.

kwasniew (approved):

gastonfournier (approved):
I like the (bad) examples and explanation why these logs should not be error

PR #9581, opened 2025-03-20
Flatten the list

thomasheartman:
wip

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
@@ -246,80 +246,74 @@ export const EnvironmentAccordionBody = ({
         );
     };
 
-    const strategyList =
-        strategies.length < 50 || !manyStrategiesPagination ? (
-            <StyledContentList>
-                {strategies.map((strategy, index) => (
-                    <StyledListItem key={strategy.id}>
-                        {index > 0 ? <StyledStrategySeparator /> : null}
+    const paginateStrategies =
+        strategies.length >= 50 && manyStrategiesPagination;
 
-                        <ProjectEnvironmentStrategyDraggableItem
-                            strategy={strategy}
-                            index={index}
-                            environmentName={featureEnvironment.name}
-                            otherEnvironments={otherEnvironments}
-                            isDragging={dragItem?.id === strategy.id}
-                            onDragStartRef={onDragStartRef}
-                            onDragOver={onDragOver(strategy.id)}
-                            onDragEnd={onDragEnd}
-                        />
-                    </StyledListItem>
-                ))}
-            </StyledContentList>
-        ) : (
-            <PaginatedStrategyContainer>
+    const strategyList = paginateStrategies ? (
+        <>
+            {page.map((strategy, index) => (
+                <StyledListItem key={strategy.id}>
+                    {index > 0 ? <StyledStrategySeparator /> : null}
+
+                    <ProjectEnvironmentStrategyDraggableItem
+                        strategy={strategy}
+                        index={index + pageIndex * pageSize}
+                        environmentName={featureEnvironment.name}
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -46,44 +46,41 @@ export const StyledContentList = styled('ol')(({ theme }) => ({
...
+    theme.palette.background.elevation1;
 
-const PaginatedStrategyContainer = styled('div')(({ theme }) => ({
-    paddingTop: theme.spacing(2.5),
-    position: 'relative',
-    display: 'flex',
-    flexFlow: 'column nowrap',
-    gap: theme.spacing(2),
+export const StyledListItem = styled('li')<{
+    'data-type'?: 'release-plan' | 'strategy';
```
thomasheartman, line 60: With the data-type attribute, we can pass it down to html and target it with css.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -46,44 +46,41 @@ export const StyledContentList = styled('ol')(({ theme }) => ({
...
 }));
 
-const StyledAlert = styled(Alert)(({ theme }) => ({
-    marginInline: theme.spacing(2), // should consider finding a variable here
+const AlertContainer = styled('div')(({ theme }) => ({
+    padding: theme.spacing(2),
+    paddingBottom: theme.spacing(0),
+    backgroundColor: strategyBackground(theme),
+    ':has(+ ol>li[data-type="release-plan"])': {
+        backgroundColor: releasePlanBackground(theme),
```
thomasheartman, line 73: The reason we need to set the background color here is that the list items use their own background color to cover the list background. So if you have release plans, that background color is slightly different from the one you'd have for regular strategies. This just makes sure we stay in sync here.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
```diff
@@ -46,44 +46,41 @@ export const StyledContentList = styled('ol')(({ theme }) => ({
...
+    },
 }));
 
-const StyledStrategySeparator = styled(StrategySeparator)(({ theme }) => ({
-    [`&:has(+ *:not(ol) .${disabledStrategyClassName}), &:has(+ ol > li:first-of-type .${disabledStrategyClassName})`]:
+// todo: consider exporting this into a shared thing or move it into the separator itself (either as a disabled prop or using the css here)
+export const StyledStrategySeparator = styled(StrategySeparator)({
+    [`&:has(+ * .${disabledStrategyClassName}, + .${disabledStrategyClassName})`]:
```
thomasheartman, line 79: The extra addition here makes the disabled work in playground too. I'm thinking it might be worth extracting this component along with the styled list and list item above. Table it for a different PR though.

Tymek (approved):

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/EnvironmentAccordionBody.tsx
thomasheartman: Ah, yes! Now you're speaking my language 🙌🏼  But it's a little tricky, because the color doesn't get set until further down. Moving it up the chain *could* work, but it also makes it a little trickier 🤔 I like the idea, but I don't have the solution immediately in my head, so I'll leave it for now (there's more important things to look at). But if you got a solution for it, I'm listening!

PR #9582, opened 2025-03-20
fix: access overview fallback to email

nunogois:
https://linear.app/unleash/issue/2-3430/fix-undefined-in-access-overview-when-user-name-is-unavailable

Adds a fallback to email in case the name is not available.

Also switches the priority of the fallbacks to be consistent with other places in our codebase (email > username) and uses `||` instead of `??` because falsy values don't provide much informational value anyways.

image

### DIFF
--- a/frontend/src/component/user/Profile/ProfileTab/ProfileTab.tsx
+++ b/frontend/src/component/user/Profile/ProfileTab/ProfileTab.tsx
@@ -131,7 +131,7 @@ export const ProfileTab = ({ user }: IProfileTabProps) => {
                 <StyledAvatar user={user} />
                 <StyledInfo>
                     <StyledInfoName>
-                        {user.name || user.username}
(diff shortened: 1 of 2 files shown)


frontend/src/component/user/Profile/ProfileTab/ProfileTab.tsx
```diff
@@ -131,7 +131,7 @@ export const ProfileTab = ({ user }: IProfileTabProps) => {
                 <StyledAvatar user={user} />
                 <StyledInfo>
                     <StyledInfoName>
-                        {user.name || user.username}
+                        {user.name || user.email || user.username}
```
nunogois, line 134: Scouting. Noticed this had the same issue: A user without name or username (email only) would show undefined here.

gastonfournier (approved):
LGTM!

PR #9584, opened 2025-03-20
fix: all users have a root role and warning if not

gastonfournier:
## About the changes
SCIM provisioned users ended up without a root role. Unleash was assigning them the Viewer role by code but some queries using the db to resolve the role did not have the same logic leading to weird behaviors. 

This amends the situation by assigning the Viewer role to those users following the least privilege principle. 

Also adds a warning when assuming the Viewer role. That should never happen but we want to be confident before removing it.

### DIFF
--- a/src/lib/services/access-service.ts
+++ b/src/lib/services/access-service.ts
@@ -457,8 +457,9 @@ export class AccessService {
     async getRootRoleForUser(userId: number): Promise<IRole> {
         const rootRole = await this.store.getRootRoleForUser(userId);
         if (!rootRole) {
-            const defaultRole = await this.getPredefinedRole(RoleName.VIEWER);
-            return defaultRole;
+            // this should never happen, but before breaking we want to know if it does.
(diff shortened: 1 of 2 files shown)


src/lib/services/access-service.ts
```diff
@@ -457,8 +457,9 @@ export class AccessService {
     async getRootRoleForUser(userId: number): Promise<IRole> {
         const rootRole = await this.store.getRootRoleForUser(userId);
         if (!rootRole) {
-            const defaultRole = await this.getPredefinedRole(RoleName.VIEWER);
-            return defaultRole;
+            // this should never happen, but before breaking we want to know if it does.
+            this.logger.warn(`Could not find root role for user=${userId}.`);
+            return this.getPredefinedRole(RoleName.VIEWER);
```
gastonfournier, line 462: In the future this should be an exception

sighphyre (approved):
Awesome!

PR #9588, opened 2025-03-20
chore: instance status prices

nunogois:
https://linear.app/unleash/issue/2-3429/use-the-correct-prices-for-each-instance-in-unleashs-ui

Uses the instance prices exposed through instance status to display the correct price amounts in Unleash's UI.

### DIFF
--- a/frontend/src/component/admin/billing/BillingDashboard/BillingPlan/BillingPlan.tsx
+++ b/frontend/src/component/admin/billing/BillingDashboard/BillingPlan/BillingPlan.tsx
@@ -130,7 +130,7 @@ export const BillingPlan = () => {
                         </GridCol>
                         <GridCol>
                             <ConditionallyRender
-                                condition={!isPAYG && baseProPrice > 0}
+                                condition={!isPAYG}
                                 show={
                                     <StyledPriceSpan>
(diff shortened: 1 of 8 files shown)


frontend/src/component/admin/billing/BillingDashboard/BillingPlan/BillingPlan.tsx
```diff
@@ -9,16 +9,14 @@ import { Badge } from 'component/common/Badge/Badge';
 import { BillingDetails } from './BillingDetails';
 import { useInstanceStatus } from 'hooks/api/getters/useInstanceStatus/useInstanceStatus';
 
-export const BILLING_PLAN_PRICES: Record<string, number> = {
-    [InstancePlan.PRO]: 80,
-};
+export const BILLING_PRO_BASE_PRICE = 80;
+export const BILLING_PRO_SEAT_PRICE = 15;
+export const BILLING_PAYG_SEAT_PRICE = 75;
+export const BILLING_TRAFFIC_PRICE = 5;
```
nunogois, line 15: I'd love to get rid of these static constants but I kept them for now as defaults, in case we have any mistakes in our logic.

We can remove them once we feel more confident.

sjaanus (approved):

PR #9592, opened 2025-03-21
Constraint values list

Tymek:
## About the changes

Closes #


frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
Tymek, line 1: Extracted from `frontend/src/component/common/ConstraintsList/StrategyEvaluationItem/StrategyEvaluationItem.tsx` and extended

frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
```diff
@@ -0,0 +1,66 @@
...
+                    onSetTruncated={() => onSetTruncated?.(false)}
+                >
+                    <Tooltip title={tooltips?.[values[0]] || ''}>
+                        <span>{values[0]}</span>
+                    </Tooltip>
+                </Truncator>
+            </StyledSingleValue>
+        ) : null}
+        {values && values?.length > 1 ? (
+            <Truncator title='' lines={2} onSetTruncated={onSetTruncated}>
```
thomasheartman, line 55: Why are these two truncator invocations different? In particular, why do they have different values for "arrow" and for the onSetTruncated function? But also, do we need separate handling of single- and multi-element lists? They're both lists, right?

frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
```diff
@@ -0,0 +1,66 @@
...
+export type ValuesListProps = {
+    values?: string[];
+    tooltips?: Record<string, string | undefined>;
+} & Pick<TruncatorProps, 'onSetTruncated'>;
+
+const StyledValuesContainer = styled('div')({
+    flex: '1 1 0',
+});
+
+const StyledValueItem = styled('span')(({ theme }) => ({
```
thomasheartman, line 17: So I take it you weren't able to make this work with actual list items?

frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
```diff
@@ -0,0 +1,66 @@
...
+    padding: theme.spacing(0.25),
+    display: 'inline-block',
+    span: {
```
thomasheartman, line 20: What's the nested span do? Does the top-level one use padding for spacing other elements? Can't that be handled by the container and `gap`?

frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
```diff
@@ -0,0 +1,66 @@
+import type { FC } from 'react';
+import { styled, Tooltip } from '@mui/material';
```
thomasheartman, line 2: Most other places (that I know of) either use the tooltip resolver (`frontend/src/component/common/TooltipResolver/TooltipResolver.tsx`) or the HTML tooltip (`frontend/src/component/common/HtmlTooltip/HtmlTooltip.tsx`) directly. This should also make it look like it does in the sketches and match other tooltips we have.

thomasheartman (commented):
Based on previous conversations, I'm guessing you weren't able to make truncation work right with `li`s? (I'm curious as to why not; you should be able to use them just as any other element).

Also:  it looks like values are displayed in chips again, but the sketches use regular text separated by commas. Was it reverted by accident?

But I'm happy for you to merge this now and we can get back to the other issues later.

frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
Tymek: the entire thing has to be `inline` for truncator to work

frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
Tymek: The second one is not for a single value - it's for the entire list, so it doesn't have a `title` - it will never display a tooltip. Instead this case should trigger the ability to "open" accordion and see values there

frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
Tymek: Will do in next PR

frontend/src/component/common/ConstraintsList/ValuesList/ValuesList.tsx
thomasheartman: Huh. So `inline-flex` doesn't work?

PR #9595, opened 2025-03-24
chore: rename release-management to release-templates, remove "plan" from descriptions

daveleek:
Changes all the routes in frontend from /release-management to /release-templates.
Also removes 'plan' from descriptions that read 'release plan templates'


gastonfournier (approved):
I was thinking maybe we should have a redirect in the frontend from /release-management to /release-templates, but first I couldn't find a place where we do that (although using <Navigate should work) and on the other side, it's a beta feature so it's ok to have this breaking change in the UI. It will auto-heal

PR #9596, opened 2025-03-24
fix: wrap env header grid on narrow containers

thomasheartman:
When the container is <500px, make it so that the strategy count wraps onto its own line to prevent it from obscuring the environment name.

I decided to go for 500px because that allows our default names to not get cut off before wrapping. It seems like a sensible default.

Before:
image

After:

image

It's still not perfect when you have envs without strategies, but it's better. We'll get back to that edge case later.

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
@@ -17,9 +17,6 @@ const StyledAccordionSummary = styled(AccordionSummary, {
     padding: theme.spacing(0.5, 3, 0.5, 2),
     display: 'flex',
     alignItems: 'center',
-    [theme.breakpoints.down('sm')]: {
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -17,9 +17,6 @@ const StyledAccordionSummary = styled(AccordionSummary, {
     padding: theme.spacing(0.5, 3, 0.5, 2),
     display: 'flex',
     alignItems: 'center',
-    [theme.breakpoints.down('sm')]: {
-        fontWeight: 'bold',
-    },
```
thomasheartman, line 22: This line was marked as new in the recent changes according to my editor. @Tymek Is this intentional? I don't see why we would want to bold it on smaller screens?

Tymek: Probably was intentional at the time. Requirements change :P

Tymek (approved):

PR #9597, opened 2025-03-24
fix: now hydration event will have segments inside

sjaanus:
This was broken by https://github.com/Unleash/unleash/pull/9254.

Previously, if no IDs were provided, it returned all segments. After the previous PR, it started returning zero.

This PR fixes that: if no IDs are defined, it will return all segments, as originally intended.

Tests coming in enterprise repo.

### DIFF
--- a/src/lib/features/segment/segment-read-model.ts
+++ b/src/lib/features/segment/segment-read-model.ts
@@ -111,7 +111,7 @@ export class SegmentReadModel implements ISegmentReadModel {
     }
 
     async getAllForClientIds(ids?: number[]): Promise<IClientSegment[]> {
-        if (!ids || ids.length === 0) {
+        if (ids?.length === 0) {
             return [];
         }
 
--- a/src/lib/features/segment/fake-segment-read-model.ts
+++ b/src/lib/features/segment/fake-segment-read-model.ts
@@ -23,7 +23,7 @@ export class FakeSegmentReadModel implements ISegmentReadModel {
         return [];
     }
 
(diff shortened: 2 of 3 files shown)


kwasniew (approved):

nunogois (approved):
So, essentially the old version early returned `[]` if ids were falsy or empty. The new version only early returns `[]` if ids are truthy but empty. Did I get that right?

LGTM.

src/lib/features/segment/segment-read-model.ts
```diff
@@ -111,7 +111,7 @@ export class SegmentReadModel implements ISegmentReadModel {
     }
 
     async getAllForClientIds(ids?: number[]): Promise<IClientSegment[]> {
-        if (!ids || ids.length === 0) {
+        if (ids?.length === 0) {
```
gastonfournier, line 114: Is this actually doing the thing?

In the left side it was returning [] if id was undefined or null or if length was zero. In the right side, it looks like the same...

gastonfournier: Ok, in this case if ids is undefined it will provide all the segments, while it only returns empty if the ids provided is the empty array.

nunogois: Yep, messed with my brain a bit as well. Essentially the happy path is `ids` falsy but not an empty array, the highlighted part on my screenshot (no "early return" is logged).

image

sjaanus: Yes, correct

PR #9599, opened 2025-03-24
Fix(1-3462)/janky drag n drop

thomasheartman:
Fixes janky drag and drop behavior and updates the styling of the drag handle focus.

The solution uses the same method to prevent oscillation as we do for strategies. To get access to the same context, I've added some extra parameters to the OnMoveItem function and passed along the extra data from the `useDragItem` hook. No new information, just making more of it available, and turning it into an object so that you can declare the properties you need (and get rid of potential wrong ordering of drag/drop indices).

For the drag and drop behavior: If the dragged element is the same size or smaller than the element you're dragging over, they will swap places as soon as you enter that space. If the target element is larger, however, they won't swap until you reach the drag/drop handle, even if they could theoretically switch somewhere in the middle. This appears to be a limitation of how the drag/drop event system works. New drag events are only fired when you "dragenter" a new element, so it never fires anywhere in the middle. Technically, we could insert more empty spans inside the drag handle to trigger more events, but I wanna hold off on that because it doesn't sound great. 

When dragging, only the handle is visible; the rest of the card stays in place. For strategies, we show a "ghost" version of the config you're dragging. However, if you apply the drag handle to the card itself, all of it becomes draggable, but you can no longer select the text inside it, which is unfortunate. Strategies do solev this, though, but I haven't been able to figure out why. If you know, please share!

Before:
image

After:
image

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/ProjectEnvironmentStrategyDraggableItem.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/ProjectEnvironmentStrategyDraggableItem.tsx
@@ -17,7 +17,6 @@ import { UPDATE_FEATURE_STRATEGY } from '@server/types/permissions';
 import { StrategyDraggableItem } from './StrategyDraggableItem';
 
 type ProjectEnvironmentStrategyDraggableItemProps = {
-    className?: string;
     strategy: IFeatureStrategy;
     environmentName: string;
     index: number;
@@ -35,7 +34,6 @@ type ProjectEnvironmentStrategyDraggableItemProps = {
 };
 
 export const ProjectEnvironmentStrategyDraggableItem = ({
-    className,
     strategy,
     index,
     environmentName,
--- a/frontend/src/component/releases/ReleasePlanTemplate/TemplateForm/MilestoneList/MilestoneCard/MilestoneCard.tsx
+++ b/frontend/src/component/releases/ReleasePlanTemplate/TemplateForm/MilestoneList/MilestoneCard/MilestoneCard.tsx
@@ -10,7 +10,7 @@ import {
     FormHelperText,
 } from '@mui/material';
 import type { IReleasePlanMilestoneStrategy } from 'interfaces/releasePlans';
-import { type DragEventHandler, type RefObject, useRef, useState } from 'react';
+import { type DragEventHandler, type RefObject, useState } from 'react';
 import ExpandMore from '@mui/icons-material/ExpandMore';
 import { MilestoneCardName } from './MilestoneCardName';
 import { MilestoneStrategyMenuCards } from './MilestoneStrategyMenu/MilestoneStrategyMenuCards';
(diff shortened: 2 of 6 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/ProjectEnvironmentStrategyDraggableItem.tsx
thomasheartman, line 1: the prop was never used.

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm/MilestoneList/MilestoneCard/MilestoneCard.tsx
thomasheartman, line 1: extract drag handle to a separate file

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm/MilestoneList/MilestoneCard/MilestoneCardDragHandle.tsx
```diff
@@ -0,0 +1,54 @@
...
+};
+
+export const MilestoneCardDragHandle: FC<Props> = ({ dragItemRef }) => {
+    return (
+        <DragButton tabIndex={-1} type='button'>
+            <DraggableContent ref={dragItemRef}>
+                <DraggableHoverIndicator className='draggable-hover-indicator'>
+                    <DragIndicator aria-hidden />
+                </DraggableHoverIndicator>
+                <ScreenReaderOnly>Drag to reorder</ScreenReaderOnly>
```
thomasheartman, line 50: The double span here serves a few purposes:
1. It allows you to start the drag anywhere the side of the card (buttons don't work great with drag n drop, so we'll use a span to make up for it)
2. Because the outer span is the height of the card, it also works to calculate whether we can or can not flip the card.
3. The inner span is just there for hover / focus indication.

frontend/src/component/releases/ReleasePlanTemplate/TemplateForm/MilestoneList/MilestoneCard/MilestoneCardDragHandle.tsx
```diff
@@ -0,0 +1,54 @@
...
+    },
+}));
+
+type Props = {
+    dragItemRef: React.RefObject<HTMLElement>;
+};
+
+export const MilestoneCardDragHandle: FC<Props> = ({ dragItemRef }) => {
+    return (
+        <DragButton tabIndex={-1} type='button'>
```
thomasheartman, line 45: These don't support keyboard drag-n-drop (yet?), so I'm taking them out of the flow for now.

frontend/src/hooks/useDragItem.ts
```diff
@@ -3,7 +3,9 @@ import { useRef, useEffect, type RefObject } from 'react';
 export type OnMoveItem = (
     dragIndex: number,
     dropIndex: number,
-    save?: boolean,
+    save: boolean,
+    event: DragEvent,
+    draggedElement: HTMLElement,
```
thomasheartman, line 8: The extra parameters here is what allows us to do the same rapid oscillation prevention as we do for strategies.

Tymek (approved):

PR #9605, opened 2025-03-25
feat: add use template button to strategies selector on flag config pane

sighphyre:
Adds a new button to the strategy options on the toggle config pane to be able to add a release plan directly

image

When clicked this just renders the usual popup but with only the release plans populated.

image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
@@ -20,6 +20,7 @@ import { useReleasePlans } from 'hooks/api/getters/useReleasePlans/useReleasePla
 import { useChangeRequestsEnabled } from 'hooks/useChangeRequestsEnabled';
 import { formatUnknownError } from 'utils/formatUnknownError';
 import { useUiFlag } from 'hooks/useUiFlag';
+import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 
 interface IFeatureStrategyMenuProps {
     label: string;
@@ -57,6 +58,7 @@ export const FeatureStrategyMenu = ({
     disableReason,
 }: IFeatureStrategyMenuProps) => {
     const [anchor, setAnchor] = useState<Element>();
+    const [onlyReleasePlans, setOnlyReleasePlans] = useState<boolean>(false);
     const navigate = useNavigate();
     const { trackEvent } = usePlausibleTracker();
     const [selectedTemplate, setSelectedTemplate] =
@@ -71,8 +73,9 @@ export const FeatureStrategyMenu = ({
         usePendingChangeRequests(projectId);
     const { refetch } = useReleasePlans(projectId, featureId, environmentId);
     const { addReleasePlanToFeature } = useReleasePlansApi();
+    const { isOss } = useUiConfig();
     const releasePlansEnabled = useUiFlag('releasePlans');
-
+    const displayReleasePlanButton = !isOss() && releasePlansEnabled;
     const crProtected =
         releasePlansEnabled && isChangeRequestConfigured(environmentId);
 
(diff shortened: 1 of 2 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
```diff
@@ -154,6 +168,27 @@ export const FeatureStrategyMenu = ({
 
     return (
         <StyledStrategyMenu onClick={(event) => event.stopPropagation()}>
+            {displayReleasePlanButton() ? (
```
sighphyre, line 171: I see these components are using the <ConditionallyRender> but there's a whole [ADR](https://docs.getunleash.io/contributing/ADRs/front-end/jsx-conditionals) to stop doing that so... I've chosen to use ternaries instead

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
```diff
@@ -10,6 +10,7 @@ interface IFeatureStrategyMenuCardsProps {
     projectId: string;
     featureId: string;
     environmentId: string;
+    onlyReleasePlans: boolean;
```
sighphyre, line 13: This component is used in literally one place so rather than duplicate it, I've just given it the option to hide the other strategy stuff

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
```diff
@@ -22,10 +23,12 @@ export const FeatureStrategyMenuCards = ({
     projectId,
     featureId,
     environmentId,
+    onlyReleasePlans,
```
gastonfournier, line 26: I think this is fine, but I was wondering, what if we just want to display strategies?
I'd think of something like:
```typescript
displayComponents = ['strategies', 'releasePlans']
```
as a default value but you can choose... Alternatively
```typescript
displayComponents = { strategies: true, releasePlans: true }
```

gastonfournier (approved):
It looks nice, just two comments to think about :D

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
sighphyre: I'm open to this, but only when that becomes a requirement. Let's leave this one for when we have a better sense of what we're doing with the UI here

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
sighphyre: I don't think that'll matter, actually. Both of those come from hooks so it should cause a re-render anyway. I don't have strong feelings either way so I'm happy to change it to a const

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
nunogois: Alternative: This component could accept `children` so that the caller decides what will be displayed here.

nunogois: But remember that this will be rebuilt soon, so it probably doesn't matter much anyways.

sighphyre: > But remember that this will be rebuilt soon, so it probably doesn't matter much anyways.

Honestly this is what's going through my mind. I'd rather not make this open to change when the next planned change is to throw this away. Right now it needs to get us over the initial feedback phase so we know what we should do next

PR #9606, opened 2025-03-25
chore: consolidate release plan flags

sighphyre:
Removes the `releasePlanChangeRequests` flag and swaps existing references to it to the `releasePlans` flag instead


gastonfournier (approved):
LG! First merge the other one so we don't break in CI

PR #9609, opened 2025-03-25
feat: improve constraints item on small screens

Tymek:
## About the changes
Fixing constraint operator item, items alignment and padding for better presentation on mobile devices.
image

### DIFF
--- a/frontend/src/component/common/ConstraintsList/ConstraintItemHeader/ConstraintItemHeader.tsx
+++ b/frontend/src/component/common/ConstraintsList/ConstraintItemHeader/ConstraintItemHeader.tsx
@@ -19,6 +19,7 @@ const Operator: FC<{
     <Tooltip title={inverted ? `Not ${label}` : label} arrow>
         <StrategyEvaluationChip
             label={formatOperatorDescription(label, inverted)}
+            multiline
         />
     </Tooltip>
 );
@@ -49,10 +50,15 @@ const CaseSensitive: FC = () => {
 };
 
 const StyledConstraintContainer = styled('div')(({ theme }) => ({
-    display: 'grid',
-    gridTemplateColumns: 'repeat(3, auto)',
-    gap: theme.spacing(2),
-    placeItems: 'center',
+    display: 'flex',
+    flexDirection: 'column',
+    gap: theme.spacing(1),
+    [theme.breakpoints.up('sm')]: {
+        gap: theme.spacing(2),
+        display: 'grid',
+        gridTemplateColumns: 'repeat(3, auto)',
+        placeItems: 'center',
+    },
 }));
 
 const StyledOperatorGroup = styled('div')(({ theme }) => ({
@@ -65,6 +71,9 @@ const StyledConstraintName = styled('div')(({ theme }) => ({
     maxWidth: '150px',
     paddingRight: theme.spacing(0.5),
     overflow: 'hidden',
+    [theme.breakpoints.down('sm')]: {
+        maxWidth: 'unset',
+    },
 }));
 
 type ConstraintItemHeaderProps = ConstraintSchema & {
--- a/frontend/src/component/common/ConstraintsList/StrategyEvaluationItem/StrategyEvaluationItem.tsx
+++ b/frontend/src/component/common/ConstraintsList/StrategyEvaluationItem/StrategyEvaluationItem.tsx
@@ -13,6 +13,9 @@ const StyledContainer = styled('div')(({ theme }) => ({
     alignItems: 'center',
     fontSize: theme.typography.body2.fontSize,
     minHeight: theme.spacing(4),
+    [theme.breakpoints.down('sm')]: {
+        flexDirection: 'column',
+    },
 }));
 
 const StyledContent = styled('div')(({ theme }) => ({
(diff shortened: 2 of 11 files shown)


frontend/src/component/common/ConstraintsList/ConstraintItemHeader/ConstraintItemHeader.tsx
```diff
@@ -49,10 +50,15 @@ const CaseSensitive: FC = () => {
...
 
 const StyledConstraintContainer = styled('div')(({ theme }) => ({
-    display: 'grid',
-    gridTemplateColumns: 'repeat(3, auto)',
-    gap: theme.spacing(2),
-    placeItems: 'center',
+    display: 'flex',
+    flexDirection: 'column',
+    gap: theme.spacing(1),
+    [theme.breakpoints.up('sm')]: {
```
thomasheartman, line 56: Can we use container queries instead of media queries for these style changes? Lets us focus more on the actual size of the constraint container than on the width of the screen. Especially because the sidebar can expand contract, this might play better.

frontend/src/component/common/ConstraintsList/StrategyEvaluationItem/StrategyEvaluationItem.tsx
```diff
@@ -13,6 +13,9 @@ const StyledContainer = styled('div')(({ theme }) => ({
     alignItems: 'center',
     fontSize: theme.typography.body2.fontSize,
     minHeight: theme.spacing(4),
+    [theme.breakpoints.down('sm')]: {
```
thomasheartman, line 16: Same thing for these breakpoints. Would it be better to use container queries?

frontend/src/component/common/SegmentItem/SegmentItem.tsx
```diff
@@ -48,12 +48,13 @@ const StyledAccordionDetails = styled(AccordionDetails)(({ theme }) => ({
     padding: theme.spacing(0.5, 3, 2.5),
 }));
 
-const StyledLink = styled(Link)({
+const StyledLink = styled(Link)(({ theme }) => ({
     textDecoration: 'none',
+    paddingRight: theme.spacing(0.5),
```
thomasheartman, line 53: This is probably totally fine, but why are you adding padding to the link? Doesn't look like it's got a different background color, so it'll just be invisible? If this is about spacing, isn't it better to add it to the relevant parent element?

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/StrategyExecution/RolloutParameter/RolloutParameter.tsx
```diff
@@ -28,17 +28,19 @@ export const RolloutParameter: FC<{
...
-                    {hasConstraints ? 'who match constraints ' : ' '}
-                    is included.
-                </span>
-                {displayGroupId && parameters?.groupId ? (
-                    <StrategyEvaluationChip
-                        label={`groupId: ${parameters?.groupId}`}
-                    />
-                ) : null}
+                <StrategyEvaluationChip label={`${percentage}%`} />
+                <div>
```
thomasheartman, line 32: can this be a `p` instead? It's text content, so it seems reasonable to me. Or in fact, if the chip is a span, then the entire child content (chip + rollout info) can (arguably should?) be in a single `p` tag.

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentAccordionBody/StrategyDraggableItem/StrategyItem/StrategyExecution/RolloutParameter/RolloutParameter.tsx
```diff
@@ -28,17 +28,19 @@ export const RolloutParameter: FC<{
...
-                <span>
-                    {hasConstraints ? 'who match constraints ' : ' '}
-                    is included.
-                </span>
-                {displayGroupId && parameters?.groupId ? (
-                    <StrategyEvaluationChip
-                        label={`groupId: ${parameters?.groupId}`}
-                    />
-                ) : null}
+                <StrategyEvaluationChip label={`${percentage}%`} />
```
thomasheartman, line 31: Not sure what this is rendered as, but it should probably be a span, I think. 
```suggestion
                <StrategyEvaluationChip component='span' label={`${percentage}%`} />
```

thomasheartman (approved):
This looks fine to me; good job on getting it to render nicely on narrow screens 🙌🏼 

Overall, I'd suggest using container queries instead of media queries for those breakpoints, but I'm not particularly bothered if you wanna keep them as media queries. Your call.

Additionally, there's a point about HTML semantics, and what elements to use. Have a look and see what you think.

Tymek:
I improved on `p`/`span` suggestions. For container queries, it has it's issues because right now a lot of spacing depends on child-width. Container query requires a more work to get it right. I'd like to attempt it, but not now.

PR #9610, opened 2025-03-25
feat: backend for retrieving tag colors

FredrikOseberg:
# Add tag color support to feature tags

## Description
This PR adds support for including tag color information when retrieving feature tags. The color information is propagated from the tag type to the feature tags, allowing the UI to display tags with their associated colors.

## Changes
1. **Schema Updates**
   - Updated `tag-schema.ts` to include color information in the tag schema
   - Added tagSchema to schemas already using tagSchema but lacking the component import

2. **Store Updates**
   - Modified `FeatureTagStore` to join with the `tag_types` table when retrieving feature tags and retrieve color
   - Updated `getAllTagsForFeature` method to include color information from tag types

3. **Test Updates**
   - Added new test case in `tags.e2e.test.ts` to verify color information propagation
   - Added new test case in  `feature.search.e2e.test.ts` to verify color information propagation

## Related
This change enables the UI to display tags with their associated colors, improving the visual organization and categorization of features.

### DIFF
--- a/src/lib/db/feature-tag-store.ts
+++ b/src/lib/db/feature-tag-store.ts
@@ -102,11 +102,18 @@ class FeatureTagStore implements IFeatureTagStore {
         const stopTimer = this.timer('getAllForFeature');
         if (await this.featureExists(featureName)) {
             const rows = await this.db
-                .select(COLUMNS)
+                .select([...COLUMNS, 'tag_types.color as color'])
                 .from<FeatureTagTable>(TABLE)
+                .leftJoin('tag_types', 'tag_types.name', 'feature_tag.tag_type')
                 .where({ feature_name: featureName });
+
             stopTimer();
-            return rows.map(this.featureTagRowToTag);
+
+            return rows.map((row) => ({
+                type: row.tag_type,
+                value: row.tag_value,
+                color: row.color,
+            }));
         } else {
             throw new NotFoundError(
                 `Could not find feature with name ${featureName}`,
--- a/src/lib/features/feature-search/feature.search.e2e.test.ts
+++ b/src/lib/features/feature-search/feature.search.e2e.test.ts
@@ -1252,3 +1252,39 @@ test('should return archived when query param set', async () => {
         ],
     });
 });
+
+test('should return tags with color information from tag type', async () => {
+    await app.createFeature('my_feature_a');
+
+    await app.request
+        .put('/api/admin/tag-types/simple')
+        .send({
+            name: 'simple',
+            description: 'Simple tag type',
+            icon: 'tag',
+            color: '#FF0000',
+        })
+        .expect(200);
+
+    await app.addTag('my_feature_a', {
+        type: 'simple',
+        value: 'my_tag',
+    });
+
+    const { body } = await searchFeatures({});
+
+    expect(body).toMatchObject({
+        features: [
+            {
+                name: 'my_feature_a',
+                tags: [
+                    {
+                        type: 'simple',
+                        value: 'my_tag',
+                        color: '#FF0000',
+                    },
+                ],
+            },
+        ],
+    });
+});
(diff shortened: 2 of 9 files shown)


src/lib/db/feature-tag-store.ts
```diff
@@ -102,11 +102,18 @@ class FeatureTagStore implements IFeatureTagStore {
         const stopTimer = this.timer('getAllForFeature');
         if (await this.featureExists(featureName)) {
             const rows = await this.db
-                .select(COLUMNS)
+                .select([...COLUMNS, 'tag_types.color as color'])
```
kwasniew, line 105: what is the reason that color is not part of the COLUMNS?

src/lib/db/feature-tag-store.ts
```diff
@@ -102,11 +102,18 @@ class FeatureTagStore implements IFeatureTagStore {
...
                 .from<FeatureTagTable>(TABLE)
+                .leftJoin('tag_types', 'tag_types.name', 'feature_tag.tag_type')
                 .where({ feature_name: featureName });
+
             stopTimer();
-            return rows.map(this.featureTagRowToTag);
```
kwasniew, line 109: is the function we stopped using still useful in other places and we couldn't modify it?

src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -1252,3 +1252,39 @@ test('should return archived when query param set', async () => {
...
 });
+
+test('should return tags with color information from tag type', async () => {
+    await app.createFeature('my_feature_a');
+
+    await app.request
+        .put('/api/admin/tag-types/simple')
+        .send({
+            name: 'simple',
+            description: 'Simple tag type',
```
kwasniew, line 1263: what are we communicating by including this description in this test? I'd omit this if this is not relevant or call it 'irrelevant'

src/lib/db/feature-tag-store.ts
FredrikOseberg: I'm not refactoring all endpoints to retrieve the tag color. Only two views currently need the tag color information. Therefore I'm limiting this iteration to only provide the joins with the tag type table and retrieve the color for these two specific endpoints.

src/lib/db/feature-tag-store.ts
FredrikOseberg: It's still useful, see answer above.

src/test/e2e/stores/feature-tag-store.e2e.test.ts
```diff
@@ -45,7 +45,7 @@ test('should tag feature', async () => {
         createdByUserId: TESTUSERID,
     });
     expect(featureTags).toHaveLength(1);
-    expect(featureTags[0]).toStrictEqual(tag);
+    expect(featureTags[0]).toStrictEqual({ ...tag, color: '#FFFFFF' });
```
kwasniew, line 48: maybe env var DEFAULT_COLOR so that we remember what FFFFFF is in future

src/test/e2e/api/admin/tags.e2e.test.ts
```diff
@@ -219,3 +223,48 @@ test('backward compatibility: the API should return invalid tag names if they ex
...
+
+    await app.request.post('/api/admin/projects/default/features').send({
+        name: featureName,
+        type: 'kill-switch',
+        enabled: true,
+        strategies: [{ name: 'default' }],
+    });
+
+    await app.request
+        .put('/api/admin/tag-types/simple')
```
kwasniew, line 242: it's weird that we parametrize ${featureName} in the next call but hardcode simple as a string. I'd either duplicate string literals or parametrize everything

kwasniew (approved):

PR #9613, opened 2025-03-25
refactor: deprecate old ConstraintAccordion and point to new version

Tymek:
## About the changes
Code for constraint accordion was copy-pasted before previous improvement. Old version is still in use for Segments. When we get to improving constraint editing we should rebuild segments editing, without use of this code.

Internal ticket: [issue/1-3536](https://linear.app/unleash/issue/1-3536/deprecate-old-constraintaccordion)


thomasheartman (commented):
Wouldn't the right thing here be to move it into a "LegacyConstraintAccordion" folder? And I'm not sure I understand the linear ticket: 
> There is a new one, since before this project. Old one is used only on segments. 

Are you saying that even before the project, there was a different constraint accordion in use somewhere? What happened to it; have we replaced it in this project? So now we have three? (this one, the other old one, and the new one?)

thomasheartman:
Great! Do we have a path forward for removing this component completely? Or is it too tangled up for now?

Tymek:
> Great! Do we have a path forward for removing this component completely? Or is it too tangled up for now?

Best time to handle it will be if(/when?) we refactor Segments-Edit page.

PR #9616, opened 2025-03-25
feat: persist env required approvals

kwasniew:
## About the changes

Environment store create/update/get/getAll should work with required_approvals. 
We have extra flag protection for explicit selects to make sure we control when the new column is being returned.
Wherever we do select * we are safe. And on the write side of things it's on the caller side to decide if we insert/update a new field or not (will be done in the enterprise repo)

### Important files

We want to fully roll out this migration first: https://github.com/Unleash/unleash/pull/9612

### DIFF
--- a/src/lib/features/project-environments/environment-service.test.ts
+++ b/src/lib/features/project-environments/environment-service.test.ts
@@ -3,6 +3,7 @@ import { createTestConfig } from '../../../test/config/test-config';
 import dbInit, { type ITestDb } from '../../../test/e2e/helpers/database-init';
 import NotFoundError from '../../error/notfound-error';
 import {
+    type IExperimentalOptions,
     type IUnleashStores,
     SYSTEM_USER,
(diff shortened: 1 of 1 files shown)


src/lib/features/project-environments/environment-store.ts
```diff
@@ -112,24 +117,35 @@ export default class EnvironmentStore implements IEnvironmentStore {
...
         this.isOss = isOss;
+        this.flagResolver = flagResolver;
         this.timer = (action) =>
             metricsHelper.wrapTimer(eventBus, DB_TIME, {
                 store: 'environment',
                 action,
             });
     }
 
+    private allColumns() {
```
kwasniew, line 137: we have previous PR with a migration so this code should never run before the migration is applied. I decided to be extra careful anyway so we have extra protection with a flag before we start asking for required_approvals in the explicit query

PR #9617, opened 2025-03-25
feat: admin menu

daveleek:
Adds an admin menu to the page when visiting a /admin sub page

- introduces a new adminRoutes.ts and moves the existing usage to oldAdminRoutes.ts (we'll remove that when we remove the flag)
- introduces a new AuthSettings so that we can control page with a route - the old behavior is kept in OldAuthSettings.tsx which will be removed when we remove the flag
- also introduces a WrapIfAdminSubpage component that just renders the children prop if flag disabled or page not a sub page of /admin. This is the scariest part of this PR so if you want to be spending some extra time reviewing thoroughly - this is a good candidate for that time

image

### DIFF
--- a/frontend/src/component/admin/Admin.tsx
+++ b/frontend/src/component/admin/Admin.tsx
@@ -2,6 +2,7 @@ import { Routes, Route } from 'react-router-dom';
 import { ApiTokenPage } from './apiToken/ApiTokenPage/ApiTokenPage';
 import { CreateApiToken } from './apiToken/CreateApiToken/CreateApiToken';
 import { AuthSettings } from './auth/AuthSettings';
+import { OldAuthSettings } from './auth/OldAuthSettings';
 import { Billing } from './billing/Billing';
 import FlaggedBillingRedirect from './billing/FlaggedBillingRedirect/FlaggedBillingRedirect';
 import { CorsAdmin } from './cors';
@@ -47,7 +48,11 @@ export const Admin = () => {
                 <Route path='banners' element={<Banners />} />
                 <Route path='license' element={<License />} />
                 <Route path='cors' element={<CorsAdmin />} />
-                <Route path='auth' element={<AuthSettings />} />
+                {newAdminUIEnabled ? (
+                    <Route path='auth/*' element={<AuthSettings />} />
+                ) : (
+                    <Route path='auth' element={<OldAuthSettings />} />
+                )}
                 <Route
                     path='admin-invoices'
                     element={<FlaggedBillingRedirect />}
--- a/frontend/src/component/admin/adminRoutes.ts
+++ b/frontend/src/component/admin/adminRoutes.ts
@@ -3,25 +3,24 @@ import type { INavigationMenuItem } from 'interfaces/route';
 export const adminGroups: Record<string, string> = {
     users: 'User administration',
     access: 'Access control',
+    sso: 'Single sign-on',
+    network: 'Network',
     instance: 'Instance configuration',
-    log: 'Logs',
-    other: 'Other',
 };
 
 export const adminRoutes: INavigationMenuItem[] = [
+    // Admin home
+    {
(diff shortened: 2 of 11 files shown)


frontend/src/component/admin/Admin.tsx
```diff
@@ -47,7 +48,11 @@ export const Admin = () => {
                 <Route path='banners' element={<Banners />} />
                 <Route path='license' element={<License />} />
                 <Route path='cors' element={<CorsAdmin />} />
-                <Route path='auth' element={<AuthSettings />} />
+                {newAdminUIEnabled ? (
+                    <Route path='auth/*' element={<AuthSettings />} />
+                ) : (
+                    <Route path='auth' element={<OldAuthSettings />} />
```
daveleek, line 54: We'll remove this when removing the flag

frontend/src/component/admin/adminRoutes.ts
```diff
@@ -3,45 +3,74 @@ import type { INavigationMenuItem } from 'interfaces/route';
...
     access: 'Access control',
+    sso: 'Single sign-on',
+    network: 'Network',
     instance: 'Instance configuration',
-    log: 'Logs',
-    other: 'Other',
 };
 
 export const adminRoutes: INavigationMenuItem[] = [
+    // Admin home
```
daveleek, line 12: Moved all the original routes to oldAdminRoutes.ts - which will be deleted when we remove the flag

frontend/src/component/admin/auth/AuthSettings.tsx
```diff
@@ -51,44 +53,16 @@ export const AuthSettings = () => {
...
+                <PageContent header={activeTab}>
+                    <Routes>
+                        <Route path='/' element={<OidcAuth />} />
+                        <Route path='/saml' element={<SamlAuth />} />
+                        <Route path='/password' element={<PasswordAuth />} />
+                        {googleAuthEnabled && (
+                            <Route path='/google' element={<GoogleAuth />} />
+                        )}
+                        <Route path='/scim' element={<ScimSettings />} />
+                    </Routes>
```
daveleek, line 65: We moved the old way of doing this to OldAuthSettings.tsx. This allows us to link directly to a SSO type from the menu. The old interface will be removed when we remove the flag

frontend/src/component/layout/MainLayout/AdminMenu/AdminMenuIcons.tsx
```diff
@@ -0,0 +1,43 @@
...
+    instance: BuildOutlinedIcon,
+    '/admin/billing': BillingIcon,
+    '/history': EventNoteIcon,
+};
+
+const findIcon = (key: string) => {
+    return icons[key] || EmptyIcon;
+};
+
+export const IconRenderer: FC<{ path: string; active: boolean }> = ({
```
daveleek, line 30: Stolen with pride from NavigationSidebar

Tymek: Maybe it should be an `AdminIconRenderer`? Spare some time for devs who use auto-import by component name :P

daveleek: Yeah not a bad point, I'll have a look at that in a later PR

frontend/src/component/admin/useAdminRoutes.ts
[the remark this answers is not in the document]
daveleek: Yeah, when cleaning up the flag those items will be changed properly and there will only be one set of menu links.

daveleek:
Thank you @Tymek!

> Another thing is, on some resolutions layout looks a bit off, asymmetric.

Yeah looking into that, it's in MainLayout. Didn't want to make changes to those breakpoint rules there in the scope of this PR as well.

PR #9618, opened 2025-03-26
feat: show seats used only for full enterprise customers

sjaanus:
This component will be hidden for pro and payg, as they are not billed by 30 days window.
image

### DIFF
--- a/frontend/src/component/admin/users/UsersHeader/UsersHeader.tsx
+++ b/frontend/src/component/admin/users/UsersHeader/UsersHeader.tsx
@@ -23,8 +23,12 @@ const StyledElement = styled(Box)(({ theme }) => ({
 }));
 
 export const UsersHeader = () => {
-    const { isOss } = useUiConfig();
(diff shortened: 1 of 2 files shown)


frontend/src/component/admin/users/UsersHeader/UsersHeader.tsx
```diff
@@ -23,8 +23,12 @@ const StyledElement = styled(Box)(({ theme }) => ({
...
 
 export const UsersHeader = () => {
-    const { isOss } = useUiConfig();
-    const licensedUsersEnabled = !isOss();
+    const {
+        isEnterprise,
+        uiConfig: { billing },
+    } = useUiConfig();
+
+    const licensedUsersEnabled = isEnterprise() && billing !== 'pay-as-you-go';
```
kwasniew, line 31: shouldn't this be modelled in the backend code so that frontend doesn't have to do the multi conditional checking?

kwasniew (approved):

PR #9621, opened 2025-03-26
feat: create and edit environment required approvals

kwasniew:
## About the changes

Can create and edit environment with required change request approvals
Screenshot 2025-03-26 at 14 34 19

Details:
* change behind a flag
* current update needs to distinguish between ignoring field in update (undefined) and reseting field to null in update

### DIFF
--- a/frontend/src/component/environments/EnvironmentForm/ChangeRequestSelector.tsx
+++ b/frontend/src/component/environments/EnvironmentForm/ChangeRequestSelector.tsx
@@ -1,0 +1,97 @@
+import {
+    FormControl,
+    FormControlLabel,
+    Radio,
+    RadioGroup,
+    styled,
+} from '@mui/material';
+import KeyboardArrowDownOutlined from '@mui/icons-material/KeyboardArrowDownOutlined';
+import GeneralSelect from '../../common/GeneralSelect/GeneralSelect';
+import { useTheme } from '@mui/material/styles';
+
+interface IEnvironmentChangeRequestProps {
+    onChange: (approvals: number | null) => void;
+    value: number | null;
+}
+
+const StyledRadioGroup = styled(RadioGroup)({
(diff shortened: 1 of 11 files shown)


src/lib/types/model.ts
```diff
@@ -198,7 +198,7 @@ export interface IEnvironment {
     projectCount?: number;
     apiTokenCount?: number;
     enabledToggleCount?: number;
-    requiredApprovals?: number;
+    requiredApprovals?: number | null;
```
kwasniew, line 201: we have to distinguish null and undefined. undefined means skip this field in update and keep the previous value, null means reset this field to null in update

frontend/src/component/environments/EnvironmentForm/ChangeRequestSelector.tsx
```diff
@@ -0,0 +1,97 @@
...
+                        control={<Radio />}
+                    />
+                </StyledRadioButtonGroup>
+            </StyledRadioGroup>
+            {value ? (
+                <>
+                    <StyledRequiredApprovals>
+                        Required approvals
+                    </StyledRequiredApprovals>
+                    <GeneralSelect
```
kwasniew, line 83: copied from the table with change request settings

frontend/src/component/environments/EnvironmentForm/ChangeRequestSelector.tsx
```diff
@@ -0,0 +1,97 @@
...
+    display: 'flex',
+    flexDirection: 'column',
+});
+
+const StyledRequiredApprovals = styled('p')(({ theme }) => ({
+    marginTop: theme.spacing(1),
+    marginBottom: theme.spacing(0.5),
+}));
+
+const useApprovalOptions = () => {
```
kwasniew, line 31: copied from change requests table

PR #9624, opened 2025-03-27
chore: add a message stating no templates when there aren't any

sighphyre:
When there's no release templates and the user clicks 'Use Template'. They now get a little message telling them that and a link to where they can add one.


gastonfournier (approved):
LG!

PR #9627, opened 2025-03-27
chore: add some getting started info and images when you have no release plans

sighphyre:
Adds a nice description about how to get started with release plans if and only if you have no current release plans. I've added included both a light and dark mode svg theme and a conditional check, it's entirely possible to mark the SVG with classes and style it directly but this approach means UX can iterate on the images easily if they want.


gastonfournier (approved):
LG!

PR #9634, opened 2025-03-27
chore: ux feedback on release templates

sighphyre:
Makes two small changes to the release template UI based on walkthrough feedback with UX

1) The how-to descriptions for creating release plans won't get hidden when the user has created release plans. We think too much is better than too little. At a later point we'll push users to documentation more aggressively
2) The warning for when the user taps the "Use template" button now has a line break to give it some breathing room and will render anchored to the bottom left of the originating button rather than covering it


gastonfournier (approved):
LG!

PR #9635, opened 2025-03-27
fix: focus styles for env headers

thomasheartman:
Adds focus styles to the env accordion header only when the focus is on the header itself (not on the env toggle inside the header). The focus style is consistent with what we do for other accordions (dashboard, milestones).

Middle one is focused:
image

Focus is on the toggle inside the top one (yeh, we should have better focus styles for toggles; but that's not for now):
image

Open and focused: 
image

Getting the consistent background for the header when it's open is a little tricky because the accordion container and summary are split into different files. ~~This first iteration used a class name for the specific header (because envs can have multiple accordion headers inside them, e.g. release plans) and setting a CSS variable in the summary, so that the background matches.~~ I found out that I only need to set it in the parent anyway 😄 

Without it, you get this (notice that there is a little white outside the lower corners): 
image

### DIFF
--- a/frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionView/ConstraintAccordionView.tsx
+++ b/frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionView/ConstraintAccordionView.tsx
@@ -95,6 +95,7 @@ export const ConstraintAccordionView = ({
                         cursor: expandable ? 'pointer' : 'default!important',
                     },
                 }}
+                tabIndex={expandable ? 0 : -1}
             >
                 <StyledWrapper>
                     <ConstraintAccordionViewHeader
--- a/frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
+++ b/frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
@@ -96,6 +96,7 @@ export const StrategyItemContainer: FC<StrategyItemContainerProps> = ({
                 <StyledHeader disabled={Boolean(strategy?.disabled)}>
                     {onDragStart ? (
                         <DragIcon
+                            tabIndex={-1}
                             className='strategy-drag-handle'
                             draggable
                             disableRipple
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
@@ -20,15 +20,13 @@ const StyledAccordionSummary = styled(AccordionSummary, {
     borderRadius: theme.shape.borderRadiusLarge,
     pointerEvents: 'auto',
     opacity: 1,
+    border: `1px solid #0000`,
     '&&&': {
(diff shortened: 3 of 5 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -17,9 +17,19 @@ const StyledAccordionSummary = styled(AccordionSummary, {
     padding: theme.spacing(0.5, 3, 0.5, 2),
     display: 'flex',
     alignItems: 'center',
+    borderRadius: theme.shape.borderRadiusLarge,
+    pointerEvents: 'auto',
+    opacity: 1,
```
thomasheartman, line 22: I moved these from the parent. Not sure why they were there in the first place? Besides, the way they were added (using the `MuiAccordionSummary--root` class), they'd also target any sub-accordion summaries (such as for release plans). This seems more sensible to me because I'd guess that this is what they're targeting specifically.

frontend/src/component/feature/FeatureView/FeatureOverview/ReleasePlan/ReleasePlan.tsx
thomasheartman, line 1: Unused variable.

Tymek (commented):
image

For some reason this will disable other environments on the list

thomasheartman:
Ah, I know what's happening, I think. Hang on.

thomasheartman:
Yeah, looks like it's happening because we use an accordion with the `disabled` prop when there's no strategies. I see now that this micro-targeting was to override the styles there and force it enabled. That feels kinda hacky to me. I think we might be better off parroting what they did for release templates and just use a div / article instead of an accordion when it's not enabled. looking into it.

frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionView/ConstraintAccordionView.tsx
```diff
@@ -95,6 +95,7 @@ export const ConstraintAccordionView = ({
                         cursor: expandable ? 'pointer' : 'default!important',
                     },
                 }}
+                tabIndex={expandable ? 0 : -1}
```
thomasheartman, line 98: This created an invisible tab stop on constraints that you can't expand or do anything with. Now, there is still no way to tell that your focus is on an expandable constraint accordion, but I'll look at that in a separate PR.

frontend/src/component/common/StrategyItemContainer/StrategyItemContainer.tsx
```diff
@@ -96,6 +96,7 @@ export const StrategyItemContainer: FC<StrategyItemContainerProps> = ({
                 <StyledHeader disabled={Boolean(strategy?.disabled)}>
                     {onDragStart ? (
                         <DragIcon
+                            tabIndex={-1}
```
thomasheartman, line 99: We don't have the capability to let you reorder by keyboard today, so let's remove the drag icon from the tab list. It also has no focus styles, so it's confusing to leave it in.

thomasheartman:
@Tymek Updated the PR to handle the case you mentioned. I discovered a few other issues in the meantime, but I'll tackle those separately.

Tymek (approved):

PR #9638, opened 2025-03-27
fix: only display feedback component on opened strategy panes

sighphyre:
Fixes an issue where the new feedback button on the config strategy pane would affect environments without strategies configured

I've made the following changes here:
- The feedback button is now disabled by default, the parent needs to request that it be turned on and that's only done in one place
- CSS has been reworked so that it behaves when that component doesn't exist on the DOM (just flex align to the right and a spacer component that expands to eat whatever space it finds)
- Swapped the CSS on the parent that was originally causing the parent component not to fill the container (looks to be flex used to align on a different axis)

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
@@ -33,17 +33,16 @@ interface IFeatureStrategyMenuProps {
     matchWidth?: boolean;
     size?: IPermissionButtonProps['size'];
     disableReason?: string;
+    allowReleasePlanFeedback?: boolean;
 }
 
 const StyledStrategyMenu = styled('div')(({ theme }) => ({
     flexShrink: 0,
     display: 'flex',
-    width: '100%',
     flexFlow: 'row',
+    flex: 1,
+    justifyContent: 'flex-end',
     gap: theme.spacing(1),
-    '& > :nth-child(2)': {
-        marginLeft: 'auto',
-    },
 }));
 
 const StyledAdditionalMenuButton = styled(PermissionButton)(({ theme }) => ({
@@ -62,6 +61,10 @@ const StyledLink = styled(Link<typeof RouterLink | 'a'>)(({ theme }) => ({
     textDecoration: 'none',
 }));
 
+const Spacer = styled('div')(({ theme }) => ({
+    flex: 1,
+}));
+
 export const FeatureStrategyMenu = ({
     label,
     projectId,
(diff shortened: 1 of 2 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
```diff
@@ -33,17 +33,16 @@ interface IFeatureStrategyMenuProps {
     matchWidth?: boolean;
     size?: IPermissionButtonProps['size'];
     disableReason?: string;
+    allowReleasePlanFeedback?: boolean;
 }
 
 const StyledStrategyMenu = styled('div')(({ theme }) => ({
     flexShrink: 0,
     display: 'flex',
-    width: '100%',
```
sighphyre, line 41: Can't use this here because something else also occupies this container in other cases so the parent is flex now along the horizontal axis

frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/FeatureOverviewEnvironment.tsx
```diff
@@ -43,8 +43,8 @@ const NewStyledAccordionDetails = styled(AccordionDetails)(({ theme }) => ({
 const StyledAccordionFooter = styled('footer')(({ theme }) => ({
     padding: theme.spacing(2, 3, 3),
     display: 'flex',
-    flexDirection: 'column',
-    alignItems: 'flex-end',
+    flexDirection: 'row',
+    justifyContent: 'end',
```
sighphyre, line 47: These two properties were causing the child to not actually be flexed along the horizontal axis. They seem to be used here for centering and I believe this should do the same thing

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
```diff
@@ -33,17 +33,16 @@ interface IFeatureStrategyMenuProps {
     matchWidth?: boolean;
     size?: IPermissionButtonProps['size'];
     disableReason?: string;
+    allowReleasePlanFeedback?: boolean;
```
Tymek, line 36: If it's passed in one place, maybe it should be slot-like instead of boolean? That's very optional.

Tymek (approved):

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
sighphyre: I missed this 'cos I set this to auto merge (wanted this to go out asap if it was good) but that's a really good idea! Totally do that tomorrow!

PR #9640, opened 2025-03-28
fix: useLocation instead of browser location for resolving pathname

daveleek:
We accidentally used `location` (the browser variable) instead of the `const location = useLocation()` from react in 2 of 3 places. this results in main menu being hidden while admin menu not being shown


sighphyre (approved):
LGTM

PR #9641, opened 2025-03-28
fix: Back to Unleash onClick+useNavigate instead of href

daveleek:
makes the button use useNavigate() instead of href - lets react sort the routing (works better with hosted instances path variables)


sighphyre (approved):
LGTM

PR #9642, opened 2025-03-28
chore: swap release template instruction svgs for pngs

sighphyre:
Swaps out the release template user guidance images for pngs rather than svgs. They're about a fifth of the size, but still larger than I'd like.

May or may not be a problem, the plan is to remove these over time and push users to our docs insteaad


daveleek (approved):
LGTM!

PR #9644, opened 2025-03-28
fix: feedback button

Tymek:
## About the changes
Fix alignment of "add strategy" & "add template" buttons


sighphyre (approved):
I think this looks good. Putting an approve so I don't block but shout if you want a re-review after cleanups!

PR #9646, opened 2025-03-28
feat: enforce env change request on project create

kwasniew:
## About the changes

* change request settings from the environment always take precedence over user selected values for those environments
* changing project env tests from interaction based to state based 

### Important files

Fake composition roots are supposed to return all collaborators that are needed for testing in addition to the service itself.
We will need to update enterprise repo manually since the signature of the fake composition root has changed.

### DIFF
--- a/src/lib/features/project-environments/environment-service.test.ts
+++ b/src/lib/features/project-environments/environment-service.test.ts
@@ -78,10 +78,19 @@ test('Can manage required approvals', async () => {
     const groupRetrieved = (await service.getAll()).find(
         (env) => env.name === 'approval_env',
     );
+    const changeRequestEnvs =
+        await db.stores.environmentStore.getChangeRequestEnvironments([
+            'approval_env',
+            'default',
+            'other',
+        ]);
 
     expect(retrieved).toEqual(created);
     expect(updated).toEqual({ ...created, requiredApprovals: 2 });
     expect(groupRetrieved).toMatchObject({ ...created, requiredApprovals: 2 });
+    expect(changeRequestEnvs).toEqual([
+        { name: 'approval_env', requiredApprovals: 2 },
+    ]);
 });
 
 test('Can connect environment to project', async () => {
--- a/src/lib/features/project-environments/environment-store-type.ts
+++ b/src/lib/features/project-environments/environment-store-type.ts
@@ -30,4 +30,7 @@ export interface IEnvironmentStore extends Store<IEnvironment, string> {
         projectId: string,
         query?: Object,
     ): Promise<IProjectEnvironment[]>;
+    getChangeRequestEnvironments(
+        environments: string[],
+    ): Promise<{ name: string; requiredApprovals: number }[]>;
 }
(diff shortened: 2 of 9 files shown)


src/lib/features/project-environments/environment-store-type.ts
```diff
@@ -30,4 +30,7 @@ export interface IEnvironmentStore extends Store<IEnvironment, string> {
         projectId: string,
         query?: Object,
     ): Promise<IProjectEnvironment[]>;
+    getChangeRequestEnvironments(
```
kwasniew, line 33: tells us which of the environments that we asked for have change requests enabled and how many approvals are needed

src/lib/features/project/project-service.test.ts
```diff
@@ -134,31 +132,89 @@ describe('enterprise extension: enable change requests', () => {
...
+                isAPI: false,
+            },
+            TEST_AUDIT_USER,
+            async () => crEnvs,
+        );
 
-        // @ts-expect-error: for testing purposes
-        service.eventService.storeEvent = storeEvent;
+        const { events } = await eventService.getEvents();
+        expect(events).toMatchObject([
```
kwasniew, line 171: this is what replaced mock verifications. We check what event was actually stored. I think it's more explicit

src/lib/features/project-environments/environment-service.test.ts
```diff
@@ -78,10 +78,19 @@ test('Can manage required approvals', async () => {
     const groupRetrieved = (await service.getAll()).find(
         (env) => env.name === 'approval_env',
     );
+    const changeRequestEnvs =
+        await db.stores.environmentStore.getChangeRequestEnvironments([
```
kwasniew, line 82: testing actual SQL

src/lib/features/project/project-service.test.ts
```diff
@@ -24,12 +35,17 @@ describe('enterprise extension: enable change requests', () => {
             TEST_AUDIT_USER,
         );
 
-        return service;
+        return {
```
kwasniew, line 38: this makes code testable without breaking service encapsulation and without ts-expect-error

src/lib/features/project-environments/environment-store-type.ts
thomasheartman: so this only returns envs that **have** CRs enabled. Non-CR envs are filtered out, right? Is there ever a use case where we'd want both? Maybe not, but I was a little surprised that it would only return envs with CRs. Maybe the name's a bit weird because it's "**get**", but you pass in a list that it filters? 🤷🏼 This is not important, by the way. Feel free to disregard it 😄 

And finally: is it more useful to return a list of items or an object that has `{ [name]: requiredApprovals }` instead? Again, not important. If you need it in that format, you can always reduce it later, I guess 💁🏼

thomasheartman (approved):
Looks good to me 😄

src/lib/features/project-environments/environment-store-type.ts
kwasniew: TBH This method was use-case driven. So I literally designed the method API from a service. I was thinking of the ideal method I'd like to have. And it lead to the method where you tell the list of envs and get only those that have change requests. We could also call it filter

thomasheartman: Yeh, all good. It made sense the further into the PR I got. There's no need to make a "theoretically perfect" method if it doesn't serve your use case. Totes fine to leave it.

PR #9648, opened 2025-03-31
docs: environment change requests

kwasniew:
## About the changes

Explaining environment level change request approval settings that can be either defaults or enforcements depending on the permissions of the project members.


website/docs/reference/change-requests.mdx
```diff
@@ -104,8 +104,27 @@ While change requests are environment-specific, project segments are not. For th
 
 Only Admin users can bypass the change request process for project segments through API calls.
 
+## Environment-level change requests
```
melindafekete, line 107: @kwasniew what do you think about calling it 'Environment-level change requests'? 'Global environment change requests' makes me think there is such a thing as a 'global environment'. but happy to change it back if you think it's better

kwasniew: I like it :)

PR #9650, opened 2025-03-31
refactor: environments in feature directory

kwasniew:
## About the changes

Environments controller and tests in feature oriented directory. There is still an environment service that is shared between global environments and project environments modules but I will not split it for now.


FredrikOseberg (approved):
LGTM

PR #9654, opened 2025-03-31
feat: remove access overview flag making it GA ready

gastonfournier:
This is exposing information we already have about permissions in a UI that should help users have an overview of the permissions of a user with regards to projects and environments


chriswk (approved):
LGTM

PR #9655, opened 2025-03-31
Fix(1-3553): fix focus styles and tab navigation for segments + constraints

thomasheartman:
Fixes a few small focus styling and tab navigation issues for the new segments + constraints.

Before:
The segment acc header was tabbable even though you can't do anything with it (by mouse or by keyboard)
image

The focus color would stay even when you're focusing stuff inside the accordion
image

Constraint items have focus styles that exceed the accordion borders, causing a weird squaring effect:
image

After:
The segment acc header isn't reachable by tab anymore.

The segment acc header doesn't change its background color on elements inside getting focus:
image

The constraint item has rounded corners for focus styles too:
image
image


frontend/src/component/common/ConstraintsList/ConstraintItemHeader/ConstraintItemHeader.tsx
thomasheartman, line 1: The extra dive here shouldn't have any effect as far as I can tell. There doesn't seem to be any styling that targets it, and I don't see any visual difference. Seems to have been an oversight.

frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionView/ConstraintAccordionView.tsx
```diff
@@ -35,6 +35,7 @@ const StyledAccordion = styled(Accordion)(({ theme }) => ({
     '&:before': {
         opacity: '0',
     },
+    overflow: 'hidden',
```
thomasheartman, line 38: Fix the focus style exceeding the constraint item issue.

frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionView/ConstraintAccordionViewHeader/ConstraintAccordionViewHeaderInfo.tsx
```diff
@@ -20,7 +20,7 @@ const StyledHeaderMetaInfo = styled('div')(({ theme }) => ({
     },
 }));
 
-const StyledExpandItem = styled('div')(({ theme }) => ({
+const StyledExpandItem = styled('p')(({ theme }) => ({
```
thomasheartman, line 23: Text is usually better with paragraphs. This one is kinda optional. I was unsure about whether we want to hide this for screen readers or not, but I think it's fine as is.

frontend/src/component/common/SegmentItem/SegmentItem.tsx
```diff
@@ -41,6 +41,9 @@ const StyledAccordion = styled(Accordion)(() => ({
 const StyledAccordionSummary = styled(AccordionSummary)(({ theme }) => ({
     fontSize: theme.typography.body2.fontSize,
     minHeight: 'unset',
+    ':focus-within': {
+        backgroundColor: 'inherit',
+    },
```
thomasheartman, line 46: Prevent background color from being set to a darker shade when focused. we don't set an explicit background color for it, so we'll set it to "inherit" on focus-within.

frontend/src/component/common/SegmentItem/SegmentItem.tsx
```diff
@@ -110,7 +114,11 @@ export const SegmentItem: FC<SegmentItemProps> = ({
                 disableGutters
                 TransitionProps={{ mountOnEnter: true, unmountOnExit: true }}
             >
-                <StyledAccordionSummary id={`segment-accordion-${segment.id}`}>
+                <StyledAccordionSummary
+                    id={`segment-accordion-${segment.id}`}
+                    tabIndex={-1}
+                    aria-controls={segmentDetailsId}
```
thomasheartman, line 120: The [a11y recommendations for accordion](https://v5.mui.com/material-ui/react-accordion/#accessibility) say to use id and aria-controls. If you set aria-controls, MUI will set that as the id of the details panel. In our case, the accordion summary doesn't do anything, but because the id is now set on the details, we can use that to indicate that our preview button now controls the section.

I thiiink this is a sensible thing to do.

Tymek (approved):

PR #9656, opened 2025-03-31
chore: admin menu enterprise badge

daveleek:
Adds enterprise badge to menu items that are enterprise only when instance is pro

Also fixes Instance stats button in admin home
Skjermbilde 2025-03-31 kl  14 34 07

### DIFF
--- a/frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
+++ b/frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
@@ -1,5 +1,5 @@
 import type React from 'react';
-import { type FC, useCallback } from 'react';
+import type { FC } from 'react';
 import type { INavigationMenuItem } from 'interfaces/route';
 import type { NavigationMode } from './NavigationMode';
 import { ShowAdmin } from './ShowHide';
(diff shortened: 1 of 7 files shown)


frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationList.tsx
```diff
@@ -33,15 +33,15 @@ const StyledBadgeContainer = styled('div')(({ theme }) => ({
     display: 'flex',
 }));
 
-const EnterprisePlanBadge = () => (
+export const EnterprisePlanBadge = () => (
```
gastonfournier, line 36: Not sure if this should be exported from here for it's usage in AdminHome as it kind of breaks the idea of having components that are self contained in a folder. But maybe double check this with someone with more context than me. This is the ADR I'm referring to https://docs.getunleash.io/contributing/ADRs/front-end/preferred-folder-structure

daveleek: Was planning on adding some context to this. My other option was to copy it. But I chose to export it and add a reference to it because it's being looked at for changing by UX, so referencing it means both places get updated more easily, and I don't add another place to change it

@kwasniew thoughts on this?

kwasniew: Can we move it to a separate component that both NavigationList and AdminNavigationItems reference?

gastonfournier (approved):
LG!

PR #9657, opened 2025-03-31
chore: expose custom strategy metrics in prometheus

gastonfournier:
## About the changes
These metrics are sent to version info but not exposed in prometheus and they can provide valuable data about their usage


sastromskis (approved):
Nice!

PR #9660, opened 2025-03-31
fix: linter rule for useHookAtTopLevel

Tymek:
React can crash an app if rules of hooks are broken. This occurs if there is a return statement before a hook, causing a different number of hook calls between component renders. To prevent this I'm reintroducing a linting rule. We probably had it before we switched from ESLint to Biome. This PR is also fixing all of the issues introduced in the code that where flagged by Biome.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -11,7 +11,8 @@
         "noUnsafeOptionalChaining": "off",
         "useExhaustiveDependencies": "off",
         "noUnusedImports": "warn",
-        "useJsxKeyInIterable": "off"
+        "useJsxKeyInIterable": "off",
+        "useHookAtTopLevel": "error"
       },
       "complexity": {
         "noBannedTypes": "off",
--- a/biome.json
+++ b/biome.json
@@ -104,7 +104,7 @@
   },
   "overrides": [
     {
-      "include": ["!frontend/**"],
+      "include": ["src/**"],
       "linter": {
         "rules": {
           "correctness": {
--- a/biome.json
+++ b/biome.json
@@ -101,5 +101,17 @@
     "formatter": {
       "indentWidth": 2
     }
-  }
+  },
+  "overrides": [
+    {
+      "include": ["!frontend/**"],
+      "linter": {
+        "rules": {
+          "correctness": {
+            "useHookAtTopLevel": "off"
+          }
+        }
+      }
+    }
+  ]
 }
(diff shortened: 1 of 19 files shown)


frontend/src/component/admin/billing/BillingDashboard/BillingPlan/useOverageCost.ts
```diff
@@ -9,10 +9,6 @@ import { BILLING_TRAFFIC_PRICE } from './BillingPlan';
 import { useInstanceStatus } from 'hooks/api/getters/useInstanceStatus/useInstanceStatus';
 
 export const useOverageCost = (includedTraffic: number) => {
-    if (!includedTraffic) {
-        return 0;
-    }
-
```
Tymek, line 15: Redundant - `includedTraffic` value is always `BILLING_INCLUDED_REQUESTS`

biome.json
```diff
@@ -100,5 +101,17 @@
...
-  }
+  },
+  "overrides": [
+    {
+      "include": ["src/**"],
+      "linter": {
+        "rules": {
+          "correctness": {
+            "useHookAtTopLevel": "off"
+          }
```
thomasheartman, line 112: I think it might be better to switch this one, so that the general rule is off, but it's error in the front end directory. Primarily because this will also impact "/website" (the docs). But maybe that's good? 🤷🏼

frontend/src/component/common/Table/cells/FeatureOverviewCell/FeatureOverviewCell.test.tsx
thomasheartman, line 1: Is this related to hooks? I think the change here (and in the file itself) makes sense, but it doesn't seem hook-related?

frontend/src/component/project/Project/ProjectSettings/ChangeRequestConfiguration/ChangeRequestTable.tsx
```diff
@@ -139,43 +140,33 @@ export const ChangeRequestTable: VFC = () => {
             },
             {
                 Header: 'Required approvals',
-                Cell: ({ row: { original } }: any) => {
-                    const { hasAccess } = useContext(AccessContext);
```
thomasheartman, line 143: Is this actually wrong? It's at the top level of a component, right? It's just an anonymous component nested within another one. dosen't seem conditional to me. Still, don't mind changing it if it makes the linter happy 💁🏼

thomasheartman (approved):
very nice work here 🙌🏼

biome.json
Tymek: This way it works when you run in `frontend` as root. I didn't get any errors in `website`, and it's easy to fix when we do

PR #9667, opened 2025-04-01
chore: add flags to control network route links from menu

daveleek:
makes sure network section routes/links are covered with the correct flags


gastonfournier (approved):
LG!

PR #9670, opened 2025-04-01
Remove caseInsensitiveInOperators flag

thomasheartman:
The flag has never made it out of dev, and has not been turned on consistently since 2023. There was a project for it, but we have since abandoned it.


FredrikOseberg (approved):
LGTM

FredrikOseberg (changes requested):
Hang on. I need to verify something.

FredrikOseberg (approved):
Merge when comfortable

PR #9671, opened 2025-04-01
chore: admin menu main layout tweaks

daveleek:
- Fixes so Admin subpage doesn't stick to the left (with a huge open space to the right) in certain screen resolutions
- Fixes the breakpoints in admin home so we switch to having two columns of stats much sooner (avoids having orphaned fourth stat on second row)
image


gastonfournier (approved):
LG!

PR #9683, opened 2025-04-02
feat: new styling for strategy selector

sjaanus:
- Updated all card titles and descriptions to use `theme.typography.caption.fontSize` (12px)
- Added proper spacing between card sections with `marginBottom: theme.spacing(3)`
- Made the section scrollable
- Added info icons with helpful tooltips to all strategy sections:
  - **Pre-defined strategy types**: "Select a starting setup, and customize the strategy to your need with targeting and variants"
  - **Apply a release template**: "Use one of the pre-defined templates defined in your company for rolling out features to users"
  - **Custom strategies**: "Custom strategies you have defined in Unleash"
- Fixed section organization by combining the default strategy with other predefined strategies under a single "Pre-defined strategy types" section

image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
@@ -24,6 +24,7 @@ const StyledContentContainer = styled('div')(() => ({
 
 const StyledName = styled(StringTruncator)(({ theme }) => ({
     fontWeight: theme.typography.fontWeightBold,
+    fontSize: theme.typography.caption.fontSize,
     display: 'block',
     marginBottom: theme.spacing(0.5),
 }));
@@ -71,7 +72,7 @@ export const FeatureReleasePlanCard = ({
                     title={description}
                     arrow
                     sx={{
-                        fontSize: (theme) => theme.typography.body2.fontSize,
+                        fontSize: (theme) => theme.typography.caption.fontSize,
                         fontWeight: (theme) =>
                             theme.typography.fontWeightRegular,
                         width: '100%',
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
@@ -228,7 +228,6 @@ export const FeatureStrategyMenu = ({
                 }}
                 PaperProps={{
                     sx: (theme) => ({
-                        paddingBottom: theme.spacing(0),
                         width: 'auto',
                         maxWidth: '95vw',
                         maxHeight: '80vh',
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
@@ -226,15 +226,6 @@ export const FeatureStrategyMenu = ({
                     vertical: 'top',
                     horizontal: 'left',
                 }}
-                PaperProps={{
-                    sx: (theme) => ({
-                        width: 'auto',
-                        maxWidth: '95vw',
-                        maxHeight: '80vh',
-                        display: 'flex',
-                        flexDirection: 'column',
-                    }),
-                }}
                 disableScrollLock={true}
             >
                 {newStrategyDropdownEnabled ? (
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
@@ -222,14 +222,21 @@ export const FeatureStrategyMenu = ({
                     vertical: 'bottom',
                     horizontal: 'left',
                 }}
+                transformOrigin={{
+                    vertical: 'top',
+                    horizontal: 'left',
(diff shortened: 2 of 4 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
```diff
@@ -222,14 +222,11 @@ export const FeatureStrategyMenu = ({
                     vertical: 'bottom',
                     horizontal: 'left',
                 }}
-                PaperProps={{
```
sjaanus, line 225: These were not needed actually. The child container will do the pushing.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
```diff
@@ -102,113 +133,140 @@ export const FeatureStrategyMenuCards = ({
...
-                                defaultStrategy={true}
-                            />
-                        </CardWrapper>
-                    </GridSection>
-                </>
-            ) : null}
-            <ConditionallyRender
-                condition={templates.length > 0}
-                show={
+            <ScrollableContent>
```
sjaanus, line 136: These big chunks is just moving some components around.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -24,6 +24,7 @@ const StyledContentContainer = styled('div')(() => ({
 
 const StyledName = styled(StringTruncator)(({ theme }) => ({
     fontWeight: theme.typography.fontWeightBold,
+    fontSize: theme.typography.caption.fontSize,
```
thomasheartman, line 27: Ah, I see you found a different size that is 12px 🕵🏼‍♂️

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
```diff
@@ -30,13 +38,23 @@ const StyledLink = styled(Link)(({ theme }) => ({
...
+    overflowY: 'auto',
+    padding: theme.spacing(0, 0, 1, 0),
 }));
 
 const GridSection = styled(Box)(({ theme }) => ({
     display: 'grid',
     gridTemplateColumns: 'repeat(2, 1fr)',
     gap: theme.spacing(1.5),
     padding: theme.spacing(0, 2),
+    marginBottom: theme.spacing(3),
```
thomasheartman, line 57: The GridSection component is housed within the GridContainer, which is a flex container. Wouldn't it be better to use `gap: theme.spacing(3)` to GridSection instead? That way, the children (grid containers) don't need to worry about how they're positioned next to their siblings. The container already covers layout direction, so I think it'd be useful to move that up there too (colocates the responsibility)

```suggestion
```

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
```diff
@@ -30,13 +38,23 @@ const StyledLink = styled(Link)(({ theme }) => ({
 
 const GridContainer = styled(Box)(() => ({
     width: '100%',
+    display: 'flex',
+    flexDirection: 'column',
```
thomasheartman, line 42: As per the comment (but you'll also need to add `{theme}` to the style function)

```suggestion
    flexDirection: 'column',
    gap: theme.spacing(3)
```

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
thomasheartman: Oh, interesting. So you're creating a separate scroll zone for strategies? Shouldn't release templates also get that treatment? Or .. what happens if you have 100 of each? (more realistically 20, but you know)

thomasheartman (approved):

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
sjaanus: Actually the `FeatureReleasePlanCard` live inside `FeatureStrategyMenuCards`. There is no `FeatureReleasePlanCards`.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
thomasheartman: Or is that not right because you also have a nested box? In that case, you can add the same gap to that one. Or: how about just applying spacing when there's more siblings?

```tsx
'& + &': { marginTop: theme.spacing(3) }
```
 will only add the margin between the items if there's more than one in a row 💁🏼

sjaanus: Not exactly, then all the cards will have this spacing.

PR #9684, opened 2025-04-02
fix: link to "view more insights" from dashboard goes to status modal

thomasheartman:
Updates the link from the project dashboard page to take you to the project status modal instead of the old insights page.

We didn't have a way to auto-open the modal before, so I added a query param to control it.

### DIFF
--- a/frontend/src/component/project/Project/Project.tsx
+++ b/frontend/src/component/project/Project/Project.tsx
@@ -32,7 +32,13 @@ import ProjectFlags from './ProjectFlags';
 import ProjectHealth from './ProjectHealth/ProjectHealth';
 import { useRequiredPathParam } from 'hooks/useRequiredPathParam';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
-import { Navigate, Route, Routes, useLocation } from 'react-router-dom';
+import {
+    Navigate,
+    Route,
+    Routes,
+    useLocation,
+    useSearchParams,
+} from 'react-router-dom';
 import { DeleteProjectDialogue } from './DeleteProject/DeleteProjectDialogue';
 import { ProjectLog } from './ProjectLog/ProjectLog';
 import { ChangeRequestOverview } from 'component/changeRequest/ChangeRequestOverview/ChangeRequestOverview';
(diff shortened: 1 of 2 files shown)


frontend/src/component/project/Project/Project.tsx
```diff
@@ -116,19 +122,32 @@ const ProjectStatusSvgWithMargin = styled(ProjectStatusSvg)(({ theme }) => ({
...
+    );
+    const toggleStatusModal = (open: boolean) => () => {
+        if (open) {
+            searchParams.set('project-status', '');
+        } else {
+            searchParams.delete('project-status');
+        }
+        setSearchParams(searchParams);
+        setProjectStatusOpen(open);
+    };
```
Tymek, line 137: In theory, this should be wrapped in a useCallback, but it's probably in 1 instance on the entire page, so it's not worth the optimization

Tymek (approved):

frontend/src/component/project/Project/Project.tsx
thomasheartman: I've seen that a couple times in the code base, but I don't quite understand what it does or why it's there? What's the point of it?

thomasheartman: Looks like it caches it. Yeah, I can wrap it .

thomasheartman: Actually, nah. According to the docs:

> Caching a function with useCallback  is only valuable in a few cases:
>
> - You pass it as a prop to a component wrapped in [memo.](https://react.dev/reference/react/memo) You want to skip re-rendering if the value hasn’t changed. Memoization lets your component re-render only if dependencies changed.
>  - The function you’re passing is later used as a dependency of some Hook. For example, another function wrapped in useCallback depends on it, or you depend on this function from [useEffect.](https://react.dev/reference/react/useEffect)
>
> There is no benefit to wrapping a function in useCallback in other cases. There is no significant harm to doing that either, so some teams choose to not think about individual cases, and memoize as much as possible. The downside is that code becomes less readable. Also, not all memoization is effective: a single value that’s “always new” is enough to break memoization for an entire component.

I don't think either of those is true in this case, so I'll revert the wrap. If it turns out to be an issue, we can look at it later.

PR #9693, opened 2025-04-03
fix: project status modal links don't work

thomasheartman:
Fixes a bug where project status modal links wouldn't work.

The reason they didn't work is because we modified the query params on modal close, and because we manually close the modal when you click a link (because otherwise it'd stay open when you navigated to other project pages), we inadverdently reset the URL.

I'm not entirely sure why setting the search params would modify the URL itself, but I'm guessing that's related to the implementation.

One way to solve this is to indicate whether we're closing the modal because a link was clicked or not, and only modify the query params if that is not the case.


kwasniew (approved):
Approving to have a fix out but let's think about a simple API without boolean params

PR #9700, opened 2025-04-04
we don't have historical data available for purchased traffic

chriswk:
So this PR removes the purchased dt point, as well as the total available. Using the same check our overageCalculation does (it also only works on current month)

### DIFF
--- a/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
+++ b/frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
@@ -65,11 +65,11 @@ const NetworkTrafficUsage: FC = () => {
         usageTotal - totalTraffic > 0 &&
         estimateTrafficDataCost;
 
+    const isCurrentMonth =
+        chartDataSelection.grouping === 'monthly' ||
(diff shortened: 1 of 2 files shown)


frontend/src/component/admin/network/NetworkTrafficUsage/NetworkTrafficUsage.tsx
```diff
@@ -65,11 +65,11 @@ const NetworkTrafficUsage: FC = () => {
         usageTotal - totalTraffic > 0 &&
         estimateTrafficDataCost;
 
+    const isCurrentMonth =
+        chartDataSelection.grouping === 'monthly' ||
```
daveleek, line 69: I'd double check this with @thomasheartman if you haven't already

gastonfournier (approved):
LG! I think there might be some request at the 1st day of the month to know the previous month... but I believe this is good enough now

daveleek (approved):
I recommend you verify the grouping === 'monthly' => isCurrentMonth true thing with @thomasheartman, but apart from that LG!

chriswk:
> I recommend you verify the grouping === 'monthly' => isCurrentMonth true thing with @thomasheartman, but apart from that LG!

good catch, of course it's 'daily'

PR #9702, opened 2025-04-04
feat: flag overview change requests

kwasniew:
## About the changes

Modifying a search query to include corresponding change requests for each of the feature environments. We will use this info to enhance flags search where release managers can see which features have actionable change requests.

### DIFF
--- a/src/lib/features/feature-search/feature-search-store.ts
+++ b/src/lib/features/feature-search/feature-search-store.ts
@@ -71,6 +71,7 @@ class FeatureSearchStore implements IFeatureSearchStore {
             hasEnabledStrategies: r.has_enabled_strategies,
             yes: Number(r.yes) || 0,
             no: Number(r.no) || 0,
+            changeRequestIds: r.change_request_ids ?? [],
         };
     }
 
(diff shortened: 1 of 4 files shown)


src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -71,6 +71,7 @@ class FeatureSearchStore implements IFeatureSearchStore {
             hasEnabledStrategies: r.has_enabled_strategies,
             yes: Number(r.yes) || 0,
             no: Number(r.no) || 0,
+            changeRequestIds: r.change_request_ids ?? [],
```
kwasniew, line 74: no need for a flag since we can fall back to empty array

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -326,6 +327,43 @@ class FeatureSearchStore implements IFeatureSearchStore {
                 'ranked_features.feature_name',
                 'lifecycle.stage_feature',
             );
+        if (this.flagResolver.isEnabled('flagsOverviewSearch')) {
+            finalQuery
```
kwasniew, line 331: we add it as late as possible so that we operate on pre-filtered features. We find all change request events for a give feature and then find which of those change request events have a corresponding change requests in a given environment.

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -326,6 +327,43 @@ class FeatureSearchStore implements IFeatureSearchStore {
...
+                        )
+                        .select('cre.feature')
+                        .select(
+                            this.db.raw(
+                                'array_agg(distinct cre.change_request_id) AS change_request_ids',
+                            ),
+                        )
+                        .select('cr.environment')
+                        .groupBy('cre.feature', 'cr.environment')
+                        .whereNotIn('cr.state', [
```
kwasniew, line 347: the tradeoff here is that the query needs to be aware of the states to exclude. Since we don't need to add new states very often I think it's ok.

src/lib/features/feature-search/feature-search-store.ts
kwasniew: I was considering moving the code into a new method but other subqueries are inlined so it's consistent and also I'm not sure if jumping to the extracted method will help a lot

sjaanus (approved):
LG

PR #9703, opened 2025-04-07
feat: release plan hover buttons

sjaanus:
image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
@@ -52,7 +52,6 @@ const StyledCard = styled(Button)(({ theme }) => ({
     borderRadius: theme.spacing(1),
     textAlign: 'left',
     overflow: 'hidden',
-    // position: 'relative',
     [`&:hover ${CardContent}`]: {
         opacity: 0.5,
     },
@@ -69,7 +68,7 @@ const StyledTopRow = styled('div')(({ theme }) => ({
     width: '100%',
 }));
 
-const PrimaryButton = styled(Button)(({ theme }) => ({
+const UseButton = styled(Button)(({ theme }) => ({
     backgroundColor: theme.palette.primary.main,
     color: theme.palette.primary.contrastText,
     '&:hover': {
@@ -77,7 +76,7 @@ const PrimaryButton = styled(Button)(({ theme }) => ({
     },
 }));
 
-const SecondaryButton = styled(Button)(({ theme }) => ({
+const PreviewButton = styled(Button)(({ theme }) => ({
     backgroundColor: 'transparent',
     border: `1px solid ${theme.palette.primary.main}`,
     color: theme.palette.primary.main,
@@ -136,20 +135,20 @@ export const FeatureReleasePlanCard = ({
             </CardContent>
 
             <HoverButtonsContainer>
-                <PrimaryButton
+                <UseButton
                     variant='contained'
                     size='small'
                     onClick={handleUseClick}
                 >
                     Use
-                </PrimaryButton>
-                <SecondaryButton
+                </UseButton>
+                <PreviewButton
                     variant='outlined'
                     size='small'
                     onClick={handlePreviewClick}
                 >
                     Preview
-                </SecondaryButton>
+                </PreviewButton>
             </HoverButtonsContainer>
         </StyledCard>
     );
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
@@ -30,10 +30,10 @@ const HoverButtonsContainer = styled('div')(({ theme }) => ({
     right: theme.spacing(2),
     display: 'flex',
     gap: theme.spacing(1),
-    visibility: 'hidden',
+    opacity: 0,
 }));
 
-const StyledCard = styled(Button)(({ theme }) => ({
+const StyledCard = styled('div')(({ theme }) => ({
     display: 'flex',
     flexDirection: 'column',
     width: '100%',
@@ -48,13 +48,13 @@ const StyledCard = styled(Button)(({ theme }) => ({
     borderRadius: theme.spacing(1),
     textAlign: 'left',
     overflow: 'hidden',
-    [`&:hover ${CardContent}, &:focus-within ${CardContent}`]: {
+    position: 'relative',
+    [`&:hover ${CardContent}`]: {
         opacity: 0.5,
     },
-    [`&:hover ${HoverButtonsContainer}, &:focus-within ${HoverButtonsContainer}`]:
-        {
-            visibility: 'visible',
-        },
+    [`&:hover ${HoverButtonsContainer}`]: {
+        opacity: 1,
+    },
 }));
 
 const StyledTopRow = styled('div')(({ theme }) => ({
@@ -118,6 +118,7 @@ export const FeatureReleasePlanCard = ({
                     variant='contained'
                     size='small'
                     onClick={handleUseClick}
+                    tabIndex={0}
                 >
                     Use
                 </Button>
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
@@ -27,14 +27,10 @@ const CardContent = styled('div')(({ theme }) => ({
 
 const HoverButtonsContainer = styled('div')(({ theme }) => ({
     position: 'absolute',
-    top: '50%',
     right: theme.spacing(2),
-    transform: 'translateY(-50%)',
     display: 'flex',
     gap: theme.spacing(1),
-    opacity: 0,
     visibility: 'hidden',
-    transition: 'opacity 0.2s ease-in-out, visibility 0.2s ease-in-out',
 }));
 
 const StyledCard = styled(Button)(({ theme }) => ({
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -20,6 +20,23 @@ const StyledName = styled(StringTruncator)(({ theme }) => ({
     marginBottom: theme.spacing(0.5),
 }));
 
+const CardContent = styled('div')(({ theme }) => ({
+    width: '100%',
+    transition: 'opacity 0.2s ease-in-out',
+}));
+
+const HoverButtonsContainer = styled('div')(({ theme }) => ({
```
sjaanus, line 28: Container going on top of the card. Positioning it in the middle of card, towards right side.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -35,8 +52,13 @@ const StyledCard = styled(Button)(({ theme }) => ({
     borderRadius: theme.spacing(1),
     textAlign: 'left',
     overflow: 'hidden',
-    '&:hover, &:focus': {
-        borderColor: theme.palette.primary.main,
+    // position: 'relative',
+    [`&:hover ${CardContent}`]: {
```
sjaanus, line 55: Card should gray out.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -35,8 +52,13 @@ const StyledCard = styled(Button)(({ theme }) => ({
...
+        opacity: 0.5,
+    },
+    [`&:hover ${HoverButtonsContainer}`]: {
```
sjaanus, line 58: Bring visibility back on hover

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -47,37 +69,88 @@ const StyledTopRow = styled('div')(({ theme }) => ({
...
+    color: theme.palette.primary.main,
+    '&:hover': {
+        backgroundColor: theme.palette.action.hover,
+    },
+}));
+
 interface IFeatureReleasePlanCardProps {
     template: IReleasePlanTemplate;
     onClick: () => void;
+    onPreviewClick?: (e: React.MouseEvent) => void;
```
sjaanus, line 91: Next PRs will implement this.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -20,6 +20,23 @@ const StyledName = styled(StringTruncator)(({ theme }) => ({
...
+const CardContent = styled('div')(({ theme }) => ({
+    width: '100%',
+    transition: 'opacity 0.2s ease-in-out',
+}));
+
+const HoverButtonsContainer = styled('div')(({ theme }) => ({
+    position: 'absolute',
+    top: '50%',
+    right: theme.spacing(2),
+    transform: 'translateY(-50%)',
```
thomasheartman, line 32: Why do you need to 50% and then translate -50%? Wouldn't `top: 0` do the same thing here?

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
thomasheartman: woah. I did not know you could use component names in css selectors like that 👀

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -47,37 +68,88 @@ const StyledTopRow = styled('div')(({ theme }) => ({
     width: '100%',
 }));
 
+const UseButton = styled(Button)(({ theme }) => ({
```
thomasheartman, line 71: Wouldn't buttons be easier to just use as variants of the MUI button? E.g. `<Button variant="outlined">` and `variant="contained"`? (Or is it "text"?). That way it deals with the colors and hover states etc for you? Maybe you're getting something extra here, but it doesn't look like that to me 🤔

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
thomasheartman: Also, reducing opacity doesn't really grey it out, it just makes it kinda see-through. It's probably fine here, but if you actually wanna grey it out, you can use [`filter: grayscale(n)`](https://developer.mozilla.org/en-US/docs/Web/CSS/filter) 💁🏼

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -20,6 +20,23 @@ const StyledName = styled(StringTruncator)(({ theme }) => ({
...
+
+const HoverButtonsContainer = styled('div')(({ theme }) => ({
+    position: 'absolute',
+    top: '50%',
+    right: theme.spacing(2),
+    transform: 'translateY(-50%)',
+    display: 'flex',
+    gap: theme.spacing(1),
+    opacity: 0,
+    visibility: 'hidden',
```
thomasheartman, line 36: According to [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility), when an element is `hidden`:
> The element cannot receive focus (such as when navigating through [tab indexes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)).

Also:
> Using a visibility value of hidden on an element will remove it from the [accessibility tree](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/What_is_accessibility#accessibility_apis). This will cause the element and all its descendant elements to no longer be announced by screen reading technology.

So I don't think using visibility hidden is the right move here. But that being said, setting opacity to 0 (which you already do) should be enough to hide it visually without taking it out of the accessibility tree. Does that not work on its own?

```suggestion
```

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
thomasheartman: I see you already add those variants in the implementation. Are the colors not right? There's also a `color` prop on mui buttons that you can use.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -47,37 +68,88 @@ const StyledTopRow = styled('div')(({ theme }) => ({
     width: '100%',
 }));
 
+const UseButton = styled(Button)(({ theme }) => ({
+    backgroundColor: theme.palette.primary.main,
+    color: theme.palette.primary.contrastText,
+    '&:hover': {
```
thomasheartman, line 74: If you're styling these yourself, make sure that `focus-visible` is also getting the appropriate styling (for tab focus)

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -47,37 +68,88 @@ const StyledTopRow = styled('div')(({ theme }) => ({
...
+        backgroundColor: theme.palette.primary.dark,
+    },
+}));
+
+const PreviewButton = styled(Button)(({ theme }) => ({
+    backgroundColor: 'transparent',
```
thomasheartman, line 80: Should this actually be transparent? What if you have so much text in the description that it runs to the end of the container and underneath the button. That might make the button text hard to read. Would making it the normal background color be better? Or it might not be a problem 🤷🏼

thomasheartman (approved):

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
sjaanus: Yeap I think opacity is what we want in this case, as black does not gray out in grayscale. It just stays black, but we need things to go lighter.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
sjaanus: Good catch, was playing around with this a bit, so this was leftover.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
sjaanus: Should it be in accessibility tree, when element is not shown in UI? I expected that it should be added to accessibility tree, when user moves on the element. So I would remove opacity and keep visibility?

PR #9704, opened 2025-04-07
docs: .NET supports impression events

daveleek:
Lists .NET as SDK that supports impression events
https://github.com/Unleash/unleash-client-dotnet/tree/main?tab=readme-ov-file#handling-events


sighphyre (approved):
LGTM

PR #9705, opened 2025-04-07
feat: search by lifecycle stage

kwasniew:
## About the changes

Allow to search by lifecycle stage
Screenshot 2025-04-07 at 12 20 48

### Important files

This PR is smal but I had to be very careful where I add the query for the lifecycle so that SQL is syntactically correct and all the information for the query is available


src/lib/openapi/spec/feature-search-query-parameters.ts
```diff
@@ -34,6 +34,18 @@ export const featureSearchQueryParameters = [
             'The state of the feature active/stale. The state can be specified with an operator. The supported operators are IS, IS_NOT, IS_ANY_OF, IS_NONE_OF.',
         in: 'query',
     },
+    {
```
kwasniew, line 37: in the UI we'll mostly gonna use IS or IS_ANY_OF but there's not reason for the API not to support all of those

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -327,7 +332,15 @@ class FeatureSearchStore implements IFeatureSearchStore {
                 'ranked_features.feature_name',
                 'lifecycle.stage_feature',
             );
+
         if (this.flagResolver.isEnabled('flagsOverviewSearch')) {
+            const parsedLifecycle = lifecycle
+                ? parseSearchOperatorValue('lifecycle.latest_stage', lifecycle)
+                : null;
+            if (parsedLifecycle) {
+                applyGenericQueryParams(finalQuery, [parsedLifecycle]);
```
kwasniew, line 341: the important part is that we can't add generic query param with other query params because we don't have lifecycle information in the ranked_features query. So I decided to add the extra query for lifecycle at the very end where the latest stage is already selected. Conveniently it can be covered by the same feature flag as the very last query for the change requests that the feature is used in.

sjaanus (approved):

PR #9708, opened 2025-04-07
fix: return bad data error when failing to patch env variants

nunogois:
https://linear.app/unleash/issue/2-3483/variants-endpoint-should-return-400-or-other-more-appropriate-status

Throws BadDataError (returns 400) instead of 500 by wrapping the patch logic with a try catch. 

Added a test that validates the new behavior.


chriswk (approved):
LGTM :)

PR #9709, opened 2025-04-07
feat: adjust columns for a search page refactor

Tymek:
## About the changes
- Change "Project ID" to Project name column in search
- Remove certain columns that are not present in new design

### DIFF
--- a/frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
+++ b/frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useState, type VFC } from 'react';
+import { useCallback, useEffect, useMemo, useState, type FC } from 'react';
 import { Box, Link, useMediaQuery, useTheme } from '@mui/material';
 import { Link as RouterLink } from 'react-router-dom';
 import { createColumnHelper, useReactTable } from '@tanstack/react-table';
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
```diff
@@ -144,30 +145,40 @@ export const FeatureToggleListTable: VFC = () => {
...
+                              header: 'Tags',
+                              cell: FeatureTagCell,
+                              enableSorting: false,
+                              meta: {
+                                  width: '1%',
+                              },
+                          },
+                      ),
+                  ]
+                : ([] as never[])),
```
Tymek, line 181: A trick to preserve TS type inference

kwasniew, line 181: what are we communicating here with the never array?

frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
```diff
@@ -144,30 +145,40 @@ export const FeatureToggleListTable: VFC = () => {
...
+                                  />
+                              ),
+                              enableSorting: false,
+                              meta: {
+                                  width: '1%',
+                              },
+                          },
+                      ),
+                      columnHelper.accessor(
+                          (row) =>
```
kwasniew, line 167: can we write explicit type so that the next person doesn't have to check the data structure at runtime?

kwasniew (approved):

PR #9713, opened 2025-04-08
feat: lifecycle filters UI

kwasniew:
## About the changes

Lifecycle filters are managed outside of the other filters. Currently I put them in a separate row but another PR will add a new filter icon too. Also numbers of items in each stage is not provided since it requires a big change in the backend code.
Screenshot 2025-04-08 at 09 17 30


frontend/src/component/feature/FeatureToggleList/FeatureToggleFilters/LifecycleFilters.tsx
```diff
@@ -0,0 +1,82 @@
+import { Box, Chip, styled } from '@mui/material';
+import type { FC } from 'react';
+import type { FilterItemParamHolder } from '../../../filter/Filters/Filters';
+
+const StyledChip = styled(Chip, {
```
kwasniew, line 5: inspired by the chips in other filters but with some minor modifications

PR #9714, opened 2025-04-08
feat: remove + from the segment dropdown

sjaanus:
Removing it from strategy and also from milestone creation.

image

### DIFF
--- a/frontend/src/component/common/AutocompleteBox/AutocompleteBox.tsx
+++ b/frontend/src/component/common/AutocompleteBox/AutocompleteBox.tsx
@@ -8,6 +8,7 @@ import {
 } from '@mui/material';
 import type { AutocompleteRenderInputParams } from '@mui/material/Autocomplete';
 import { useState } from 'react';
+import type { ReactNode } from 'react';
 
 interface IAutocompleteBoxProps {
     label: string;
@@ -15,6 +16,7 @@ interface IAutocompleteBoxProps {
     value?: IAutocompleteBoxOption[];
     onChange: (value: IAutocompleteBoxOption[]) => void;
     disabled?: boolean;
+    icon?: ReactNode | null;
 }
 
 export interface IAutocompleteBoxOption {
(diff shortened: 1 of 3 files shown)


frontend/src/component/common/AutocompleteBox/AutocompleteBox.tsx
```diff
@@ -32,23 +34,6 @@ const StyledContainer = styled('div')(({ theme }) => ({
     },
 }));
 
-const StyledIcon = styled('div', {
```
sjaanus, line 35: These small cleanups were just unused pieces of code

frontend/src/component/common/AutocompleteBox/AutocompleteBox.tsx
```diff
@@ -9,13 +8,16 @@ import {
...
 import { useState } from 'react';
+import type { ReactNode } from 'react';
 
 interface IAutocompleteBoxProps {
     label: string;
     options: IAutocompleteBoxOption[];
     value?: IAutocompleteBoxOption[];
     onChange: (value: IAutocompleteBoxOption[]) => void;
     disabled?: boolean;
+    icon?: ReactNode | null;
```
sjaanus, line 19: Making icon and width configurable. Previously was hard coded.

frontend/src/component/common/AutocompleteBox/AutocompleteBox.tsx
```diff
@@ -59,37 +44,42 @@ export const AutocompleteBox = ({
     value = [],
     onChange,
     disabled,
+    icon,
+    width = '215px',
 }: IAutocompleteBoxProps) => {
-    const [placeHolder, setPlaceholder] = useState('Add Segments');
-    const { classes: styles } = useStyles();
+    const [_, setPlaceholder] = useState('Add Segments');
```
thomasheartman, line 50: This is a little confusing. Why is the default placeholder "add segments"? And if we never use the actual value, why is this line here at all?

frontend/src/component/common/AutocompleteBox/AutocompleteBox.tsx
thomasheartman, line 19: Can we make it just normal or undefined? I don't think the `null` state adds anything here (except for possibly confusion about what the difference between null and undefined is).

```suggestion
    icon?: ReactNode;
```

thomasheartman: On closer inspection, I see you treat them differently: `undefined` is the default icon, `null` is no icon. Maybe that is sensible? 🤔

thomasheartman (approved):

Issue #9710, opened 2025-04-07
Security Vulnerability: CVE-2025-26519 in Unleash Docker Image (musl libc)


iamrayofhope:
### Describe the bug

We're currently using Unleash version 6.9.0 and have identified a potential security vulnerability due to https://github.com/advisories/GHSA-xpv5-92cc-8f65, which affects musl libc (used in Alpine-based Docker images).

Details:

CVE ID: [CVE-2025-26519](https://nvd.nist.gov/vuln/detail/CVE-2025-26519)

Vulnerability Type: Out-of-bounds write in musl libc

Affected Component: Docker image based on Alpine Linux with vulnerable musl libc

Relevant Alpine Versions:

Alpine 3.16 with musl < 1.2.3-r4

Alpine 3.17 with musl < 1.2.3-r6

Alpine 3.18 with musl < 1.2.4-r3

Impact:

If Unleash is deployed via a Docker image using Alpine Linux, it may include a vulnerable version of musl, exposing systems to potential memory corruption issues when processing specific character encodings.

### Steps to reproduce the bug

Steps to reproduce the bug
Steps to Reproduce the Vulnerability (Scan Method)
✅ Prerequisites:
Docker installed

A vulnerability scanner like:

[Trivy](https://github.com/aquasecurity/trivy)

[Grype](https://github.com/anchore/grype)

Docker Scout (built-in for Docker Desktop)

🔍 Using Trivy (Recommended & Fast)
Install Trivy (if not already installed):

bash
Copy
Edit
brew install aquasecurity/trivy/trivy # macOS
sudo apt install trivy # Ubuntu
Pull the Unleash Docker image (vulnerable version):

bash
Copy
Edit
docker pull unleashorg/unleash-server:6.9.0
Scan the image:

bash
Copy
Edit
trivy image unleashorg/unleash-server:6.9.0
Look for musl-related CVEs: You'll see output like this:

yaml
Copy
Edit
musl
├── https://github.com/advisories/GHSA-xpv5-92cc-8f65
│ Severity: HIGH
│ Installed Version: 1.2.3-r2
│ Fixed Version: 1.2.3-r4

### Expected behavior

Please update the base image used in the Dockerfile to one with a patched version of musl libc.

### Logs, error output, etc.

```bash

```

### Additional context

![Image](https://github.com/user-attachments/assets/aba9f47d-c05d-41bf-b541-03b75b2a69b0)

### Unleash version

6.9.0

### Subscription type

Open source

[...]

PR #9716, opened 2025-04-08
Task/bump docker base image

chriswk:
As reported in #9710 . This PR updates to 20.19.0-alpine3.21 which according to docker's own scan now longer is affected by the CVE


sastromskis (approved):
LG

PR #9721, opened 2025-04-08
fix: fix hover on strategy card

sjaanus:
For some reason, using template literals do not work after its in enterprise repo. I will try something that matches more over codebase style.


frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureReleasePlanCard/FeatureReleasePlanCard.tsx
```diff
@@ -49,13 +50,12 @@ const StyledCard = styled('div')(({ theme }) => ({
     textAlign: 'left',
     overflow: 'hidden',
     position: 'relative',
-    [`&:hover ${CardContent}, &:focus-within ${CardContent}`]: {
+    '&:hover > div:first-of-type, &:focus-within > div:first-of-type': {
```
thomasheartman, line 53: Right, if I'm reading these right, you want the main card to lose opacity and the buttons to gain opacity, yeah? The selectors here might work, but they're also pretty brittle. Why not use html classes for this? Alternatively, you can apply the styling within the affected components directly.

thomasheartman (approved):

PR #9724, opened 2025-04-08
feat: total count respect lifecycle filter

kwasniew:
## About the changes

Originally I added lifecycle as a separate CTE. Later I found that total counting will not take it into account. In this PR I moved the lifecycle join into the first CTE that is later used for total counting. I also added a test showing the previous behavior where total was not counted correctly.

### DIFF
--- a/src/lib/features/feature-search/feature-search-store.ts
+++ b/src/lib/features/feature-search/feature-search-store.ts
@@ -1,4 +1,4 @@
-import type { Knex } from 'knex';
+import { Knex } from 'knex';
 import type EventEmitter from 'events';
 import metricsHelper from '../../util/metrics-helper';
 import { DB_TIME } from '../../metric-events';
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -79,24 +79,6 @@ class FeatureSearchStore implements IFeatureSearchStore {
         };
     }
 
-    private getLatestLifecycleStageQuery() {
```
kwasniew, line 82: we're inlining this

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -245,19 +230,48 @@ class FeatureSearchStore implements IFeatureSearchStore {
                         'users',
                         'users.id',
                         'features.created_by_user_id',
-                    );
-
-                query.leftJoin('last_seen_at_metrics', function () {
```
kwasniew, line 250: not sure why this one didn't follow chaining pattern and started on a new line

sjaanus (approved):
Ha, good catch

PR #9726, opened 2025-04-09
feat: styling of AND operator in strategies

sjaanus:
I am not even putting behind flag, as its minor change.

image


thomasheartman (approved):
Nice! I'm ... kinda torn between putting it behind a flag or not. No strong feelings, so yeh, go with your gut.

I was surprised to find that this wasn't the same component that we use for general constraint separation, though (`frontend/src/component/common/ConstraintsList/ConstraintSeparator/ConstraintSeparator.tsx`). Maybe it's worth seeing if we can use that? It feels like they're the same thing to me. 

And if we don't: can we add `role="separator"` to the separator elements where they're used? We did that for the recent strategy and constraint separators, and I think it's the right thing to do.

PR #9727, opened 2025-04-09
Chore(1-3598): new constraint edit design iteration 1

thomasheartman:
Implements the first step towards implementing the new design for constraint editing. All the edit functionalities work as and when you do them now, but there is no validation of the values you put in that's happening.

The inverted / not inverted button and the case sensitivity button are placeholders. They should use icons and have proper descriptions of what they do. I'll do that in a follow-up.

The way to enter values is currently always in the section below the main controls. Again, more work on this is coming.

Current look:

With case sensitive options:
image

With legal values:
image


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/ConstraintOperatorSelect.tsx
thomasheartman, line 1: This is essentially adapting `frontend/src/component/common/NewConstraintAccordion/ConstraintOperatorSelect.tsx`, which is a copy of `frontend/src/component/common/ConstraintAccordion/ConstraintOperatorSelect.tsx`. 

I think that both of those old components can be deleted (maybe?) after this project, but I'm not wild about having three files with the exact same (and pretty specific) name rolling around in here. Suggestions are welcome!

PR #9732, opened 2025-04-09
feat: strategy menu interaction between two dialogues.

sjaanus:
- Significantly improved dialog navigation flow — the "Go back" button now properly returns users to the strategy menu instead of closing the preview dialog with no way to return
- Now release cards expand to 100% height if neighbour card has bigger height.


frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCard/FeatureStrategyMenuCard.tsx
```diff
@@ -86,6 +88,9 @@ export const FeatureStrategyMenuCard = ({
                 buttonTitle: strategy.displayName || strategyName,
             },
         });
+        if (onClose) {
+            onClose();
+        }
```
thomasheartman, line 93: Not important, but we can simplify this to:
```suggestion
        onClose?.()
```

thomasheartman (approved):

PR #9733, opened 2025-04-09
fix: wrap env data before truncating env name.

thomasheartman:
Solves an issue where the new buttons would potentially obscure the entire name of the env:

image

 
Now, instead of using grid and container queries to find the right point to break, we're using flex to always wrap before needing to truncate the environment name. The name will still truncate if necessary. 

Why didn't we do this originally? I ... couldn't think of a way to make flex work in a 2D layout, but the `width: 100%` property seems to do the trick 😄 

image

PS: the buttons only stack when they have to. They stay single-line for as long as possible. (just in case you were wondering).
image

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
@@ -41,20 +41,16 @@ const StyledHeader = styled('header')(({ theme }) => ({
 }));
 
 const StyledHeaderTitle = styled('hgroup')(({ theme }) => ({
-    display: 'grid',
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvironments/FeatureOverviewEnvironment/EnvironmentHeader/EnvironmentHeader.tsx
```diff
@@ -37,24 +37,19 @@ const StyledHeader = styled('header')(({ theme }) => ({
...
+    flexFlow: 'row wrap',
     flex: 1,
     columnGap: theme.spacing(1),
-    '@container (max-width: 500px)': {
-        gridTemplateColumns: '1fr',
-    },
 }));
 
 const StyledHeaderTitleLabel = styled('p')(({ theme }) => ({
+    width: '100%',
```
thomasheartman, line 50: effectively puts this element on its own row

sjaanus (approved):

PR #9739, opened 2025-04-10
feat: add milestones to search results

kwasniew:
## About the changes

Feature search should include milestone name, milestone order and total number of milestones
Screenshot 2025-04-10 at 08 57 10

## Discussion points

@Tymek Jaanus has a good point that writing a query here allows for a faster iteration
@sjaanus please take a look at the query enhancement. I will be running explain plan for this part and change requests soon but I don't expect perf degradation due to early filtering happening before we join extra data.
@daveleek are you ok with me moving stores for release plans into OSS so that we can use them for the test setup? Also please check if the query makes sense to you since you know it well. I basically join two tables that you created with feature name and environments from the search result features.

### DIFF
--- a/src/lib/features/feature-search/feature-search-store.ts
+++ b/src/lib/features/feature-search/feature-search-store.ts
@@ -76,6 +76,13 @@ class FeatureSearchStore implements IFeatureSearchStore {
             yes: Number(r.yes) || 0,
             no: Number(r.no) || 0,
             changeRequestIds: r.change_request_ids ?? [],
+            ...(r.milestone_name
+                ? {
+                      milestoneName: r.milestone_name,
+                      milestoneOrder: r.milestone_order,
+                      totalMilestones: Number(r.total_milestones || 0),
+                  }
+                : {}),
         };
     }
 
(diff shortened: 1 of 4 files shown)


src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -371,6 +378,53 @@ class FeatureSearchStore implements IFeatureSearchStore {
                     },
                 )
                 .select('feature_cr.change_request_ids');
+
+            finalQuery
```
kwasniew, line 382: This could be a view or extracted method but the query is here to stay

src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -1425,3 +1425,93 @@ test('should return change request ids per environment', async () => {
         ],
     });
 });
+
+const createReleasePlan = async ({
```
kwasniew, line 1429: we don't have release plan stores in OSS so will move them in another PR instead of manual queries

src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -1425,3 +1425,93 @@ test('should return change request ids per environment', async () => {
...
+
+    expect(body).toMatchObject({
+        features: [
+            {
+                name: 'my_feature_a',
+                environments: [
+                    { name: 'default' },
+                    {
+                        name: 'development',
+                        totalMilestones: 3,
```
kwasniew, line 1506: this is the new info we need per env

src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -1425,3 +1425,93 @@ test('should return change request ids per environment', async () => {
...
+                        milestoneName: 'Milestone 2',
+                        milestoneOrder: 1,
+                    },
+                    { name: 'production' },
+                ],
+            },
+        ],
+    });
+    expect(body.features[0].environments[0].milestoneName).toBeUndefined();
```
kwasniew, line 1515: envs without release plans will return nothing

src/lib/openapi/spec/feature-environment-schema.ts
```diff
@@ -71,6 +71,24 @@ export const featureEnvironmentSchema = {
             description:
                 'Experimental. A list of change request identifiers for actionable change requests that are not Cancelled, Rejected or Approved.',
         },
+        milestoneName: {
```
kwasniew, line 74: decided to follow the pattern and not nest anything under release plans

src/lib/features/feature-search/feature-search-store.ts
sjaanus: I think extracting the method is a good direction. We do not have many of that currently, but if we can have these parts as separate methods, the method naming can give good indication what is being done in each joins.

daveleek (approved):
Nice! Yeah we can move the stores to OSS, shouldn't be a problem!

src/lib/features/feature-search/feature-search-store.ts
```diff
@@ -371,6 +378,53 @@ class FeatureSearchStore implements IFeatureSearchStore {
...
+                .leftJoin(
+                    this.db
```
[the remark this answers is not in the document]
daveleek: Agree with @sjaanus here!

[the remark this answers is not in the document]
kwasniew: I will merge this to test in sandbox and will run explain plans before finishing this piece of work

PR #9747, opened 2025-04-11
refactor: move release plan stores to OSS

kwasniew:
## About the changes

Moving release plan stores to OSS. No new code so no need to review line by line.

I'm already using those stores in search tests. Will do the enterprise cleanup in the next PR.


daveleek (approved):
Nice one!

PR #9755, opened 2025-04-14
fix: improve status job performance

kwasniew:
## About the changes

statusJob was running project status calculation for multiple projects in parallel. Now we spread it into single project chunks run every 30s. 
Also adding some instrumentation for the function execution time and query execution time.

### DIFF
--- a/src/lib/db/project-stats-store.ts
+++ b/src/lib/db/project-stats-store.ts
@@ -117,7 +117,7 @@ class ProjectStatsStore implements IProjectStatsStore {
     async getTimeToProdDates(
         projectId: string,
     ): Promise<ICreateEnabledDates[]> {
-        const stopTimer = this.timer('get_time_to_prod_dates');
+        const stopTimer = this.timer('getTimeToProdDates');
         const result = await this.db
             .select('events.feature_name')
             // select only first enabled event, distinct works with orderBy
--- a/src/lib/db/project-stats-store.ts
+++ b/src/lib/db/project-stats-store.ts
@@ -117,6 +117,7 @@ class ProjectStatsStore implements IProjectStatsStore {
     async getTimeToProdDates(
         projectId: string,
     ): Promise<ICreateEnabledDates[]> {
+        const stopTimer = this.timer('get_time_to_prod_dates');
         const result = await this.db
             .select('events.feature_name')
             // select only first enabled event, distinct works with orderBy
@@ -143,6 +144,7 @@ class ProjectStatsStore implements IProjectStatsStore {
             .orderBy('events.feature_name')
             // first enabled event
             .orderBy('events.created_at', 'asc');
+        stopTimer();
         return result;
     }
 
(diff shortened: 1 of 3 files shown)


src/lib/features/project/project-service.ts
```diff
@@ -1282,21 +1292,18 @@ export default class ProjectService {
...
-        await Promise.all(
-            statusUpdates.map((statusUpdate) => {
-                return this.projectStatsStore.updateProjectStats(
-                    statusUpdate.projectId,
-                    statusUpdate.updates,
-                );
-            }),
-        );
+        // run one project status update at a time every
+        void batchExecute(projects, 1, 30_000, async (project) => {
```
kwasniew, line 1296: This will relax the load on DB when you have many projects. Previously each project was running ~10 queries. Now we spread it over time

kwasniew: Previously status job was waiting for the completion of all status updates but now we early exit without waiting. batchExecute runs with setTimeout over many minutes but we don't wait on the final result upon job time reporting

src/lib/features/project/project-service.ts
```diff
@@ -1282,21 +1292,18 @@ export default class ProjectService {
...
+            const statusUpdate = await this.getStatusUpdates(project.id);
+            await this.projectStatsStore.updateProjectStats(
+                statusUpdate.projectId,
+                statusUpdate.updates,
+            );
+        });
     }
 
     async getStatusUpdates(projectId: string): Promise<ICalculateStatus> {
```
kwasniew, line 1305: I have a hunch that this query may be slow so want to confirm this

src/lib/features/project/project-service.ts
```diff
@@ -1282,21 +1292,18 @@ export default class ProjectService {
...
+        const stopTimer = this.timer('getStatusUpdates');
```
kwasniew, line 1306: because we don't have a total scheduled job timer working due to async batching we need another timer on the most expensive function itself

src/lib/db/project-stats-store.ts
```diff
@@ -117,6 +117,7 @@ class ProjectStatsStore implements IProjectStatsStore {
     async getTimeToProdDates(
         projectId: string,
     ): Promise<ICreateEnabledDates[]> {
+        const stopTimer = this.timer('getTimeToProdDates');
```
kwasniew, line 120: I have a hunch that this query may be slow so want to get numbers

src/lib/util/index.ts
```diff
@@ -30,3 +30,4 @@ export * from './arraysHaveSameItems';
 export * from './constantTimeCompare';
 export * from '../features/metrics/client-metrics/collapseHourlyMetrics';
 export * from '../features/playground/offline-unleash-client';
+export * from './batchExecute';
```
kwasniew, line 33: moved this fn from another repo and no I need to export it

sjaanus (approved):
LG, I'm wondering if we can measure how long the job takes now. Then again, we're intentionally slowing it down — 72 × 30 seconds, which comes to 36 minutes — so the measurement might not tell us much.

Is there an issue where the pod closes midway?

kwasniew:
@sjaanus With the async processing every 30s there's no good reason to measure the total time since as you said it's known upfront. That's why I added function level measurements around status calculation for each project. This is our closest approximation with the new async batch processing model.

PR #9756, opened 2025-04-14
fix: demo QR code

nunogois:
https://linear.app/unleash/issue/CJUX-340/bug-qr-code-doesnt-work

Generated a new QR code using [qrcode-monkey.com](https://www.qrcode-monkey.com/), following a similar design to our previous one.

Unlike the previous QR generator, this one does **not** proxy the URL—so the QR code will never break in the future. It’s also in SVG format, allowing for infinite scaling without quality loss.

image


chriswk (approved):
Nice

PR #9758, opened 2025-04-15
fix(frontend): Now only shows average daily requests when viewing daily

chriswk:
As reported by Ivar, there's no point in displaying the average daily traffic when grouping monthly.


frontend/src/component/admin/network/NetworkTrafficUsage/hooks/useChartDataSelection.ts
```diff
@@ -17,6 +17,7 @@ export const useChartDataSelection = (includedTraffic?: number) => {
         });
 
     const options = useMemo(() => {
+        console.log(includedTraffic);
```
daveleek: Console.log :)

PR #9761, opened 2025-04-15
fix: improve health rating job performance

kwasniew:
## About the changes

setHealthRating job was running project health calculation for multiple projects in parallel. Now we spread it into single project chunks run every 5s.
Also adding some instrumentation for the function execution time and query execution time.

This is similar improvement to statusJob: https://github.com/Unleash/unleash/pull/9755

### DIFF
--- a/src/lib/services/project-health-service.ts
+++ b/src/lib/services/project-health-service.ts
@@ -10,6 +10,9 @@ import {
     calculateProjectHealth,
     calculateProjectHealthRating,
 } from '../domain/project-health/project-health';
+import { batchExecute } from '../util';
+import metricsHelper from '../util/metrics-helper';
+import { FUNCTION_TIME } from '../metric-events';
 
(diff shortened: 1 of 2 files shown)


src/lib/services/project-health-service.ts
```diff
@@ -22,7 +25,9 @@ export default class ProjectHealthService {
 
     private projectService: ProjectService;
 
-    calculateHealthRating: (project: IProject) => Promise<number>;
+    calculateHealthRating: (project: Pick<IProject, 'id'>) => Promise<number>;
```
Tymek, line 28: this one could also be `private`, if it isn't public for testing

Tymek (approved):

src/lib/services/project-health-service.ts
kwasniew: it is called by the scheduler

PR #9765, opened 2025-04-15
fix: demo topics should work with new flag design

nunogois:

Fixes our demo topics by making them work with the new flag page design.

We achieve this by adding 2 new interactive steps in `demoApp.step2` and `demoApp.step4` to expand the respective environment accordions. We mark them as optional so they are not strictly required and will be skipped if not found. This means the demo will be resilient to rolling back the `flagOverviewRedesign` flag, for example.

We also mark 2 steps as optional: saving constraints in `demoApp.step2` and `demoApp.step4`. It seems like we no longer have an extra button to save the constraints after adding them, so by marking these steps as optional the demo flow is able to proceed without breaking.

### DIFF
--- a/frontend/src/component/demo/demo-topics.tsx
+++ b/frontend/src/component/demo/demo-topics.tsx
@@ -131,6 +131,17 @@ export const TOPICS: ITutorialTopic[] = [
             },
             {
                 href: `/projects/${PROJECT}/features/demoApp.step2`,
+                target: `div[data-testid="FEATURE_ENVIRONMENT_ACCORDION_${ENVIRONMENT}"] .MuiAccordionSummary-expandIconWrapper`,
+                content: (
+                    <Description>
+                        Expand the environment card to see all the defined
+                        strategies by using the arrow button.
+                    </Description>
+                ),
+                optional: true,
+            },
+            {
+                href: `/projects/${PROJECT}/features/demoApp.step2`,
                 target: 'button[data-testid="ADD_STRATEGY_BUTTON"]',
                 content: (
                     <Description>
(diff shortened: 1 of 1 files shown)


frontend/src/component/demo/demo-topics.tsx
```diff
@@ -467,6 +479,17 @@ export const TOPICS: ITutorialTopic[] = [
                 ),
                 preventDefault: true,
             },
+            {
+                href: `/projects/${PROJECT}/features/demoApp.step4`,
```
gastonfournier, line 483: Is it ok having step4 twice?

gastonfournier (approved):
I'm not sure I understand the change but feel safe, feel free to merge, test and iterate

sighphyre (approved):
That's pretty clever. LGTM

frontend/src/component/demo/demo-topics.tsx
nunogois: Should be safe, as this only triggers a navigation if the current URL does not match, but I can double check just in case.

frontend/src/component/demo/demo-topics.tsx
```diff
@@ -129,6 +129,17 @@ export const TOPICS: ITutorialTopic[] = [
                 ),
                 preventDefault: true,
             },
+            {
+                href: `/projects/${PROJECT}/features/demoApp.step2`,
+                target: `div[data-testid="FEATURE_ENVIRONMENT_ACCORDION_${ENVIRONMENT}"] .MuiAccordionSummary-expandIconWrapper`,
```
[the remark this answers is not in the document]
nunogois: Agreed. We're leveraging our `data-testid` but still selecting the MUI expand icon wrapper, which there's no guarantee it will always be present in the future. The reason I did this was because this seems to have been working well for some time now. See in `demoApp.step3`: https://github.com/Unleash/unleash/blob/1d94c261e6435ade29e9654c7a4f3d980c6ee7ba/frontend/src/component/demo/demo-topics.tsx#L368

I'll look into maybe highlighting the full accordion (summary) instead of specifically the icon wrapper.

PR #9770, opened 2025-04-16
fix: demo flow consistency with new flag page

nunogois:

Follow-up to #9765 

This should make the demo flow a bit more consistent with the new page design.

The page seems a bit slower in general, so we're being a bit more aggressive on the cleaning up of demo data to make it faster as well as adding some delays in key steps of the flow.

### DIFF
--- a/frontend/src/component/demo/demo-setup.ts
+++ b/frontend/src/component/demo/demo-setup.ts
@@ -5,12 +5,47 @@ import { formatApiPath } from 'utils/formatPath';
 const PROJECT = 'demo-app';
 const ENVIRONMENT = 'dev';
 
-const ensureUserIdContextExists = async () => {
+const CONTEXT_FIELDS_TO_KEEP = [
+    'appName',
+    'country',
+    'currentTime',
+    'sessionId',
+    'userId',
+];
+
+const getContextFields = async () => {
     const contextFields: IUnleashContextDefinition[] =
         (await fetch(formatApiPath('api/admin/context')).then((res) =>
             res.json(),
         )) || [];
 
+    return contextFields;
+};
+
+const deleteOldContextFields = async (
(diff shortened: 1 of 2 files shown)


frontend/src/component/demo/demo-setup.ts
```diff
@@ -27,46 +62,10 @@ const ensureUserIdContextExists = async () => {
...
-                        stickiness: 'default',
-                        groupId: featureId,
-                    },
-                }),
-            },
-        );
-    }
+const cleanUpContext = async () => {
+    const contextFields = await getContextFields();
+    await deleteOldContextFields(contextFields);
```
gastonfournier, line 67: Mmmm this looks a bit odd... it looks like we're doing something via API that should be done via a migration...

frontend/src/component/demo/demo-topics.tsx
```diff
@@ -481,18 +484,19 @@ export const TOPICS: ITutorialTopic[] = [
             },
             {
                 href: `/projects/${PROJECT}/features/demoApp.step4`,
-                target: `div[data-testid="FEATURE_ENVIRONMENT_ACCORDION_${ENVIRONMENT}"] .MuiAccordionSummary-expandIconWrapper`,
+                target: `div[data-testid="FEATURE_ENVIRONMENT_ACCORDION_${ENVIRONMENT}"] > div[aria-expanded="false"]`,
```
gastonfournier, line 487: I would try to do what @kwasniew suggested of adding a class to the div, maybe `demoApp.step4` so we can target it more precisely. Maybe a css class is not "the best" but something that specifically indicates this is the demoApp step4 target would be nice

gastonfournier (approved):
It feels really odd, but we'll let this one go

PR #9771, opened 2025-04-16
fix: demo flow step 4 is too eager on add strategy

nunogois:

Follow-up to #9770 

This "add strategy" step of `demoApp.step4` was a bit too eager and did not properly wait for the accordion to be fully expanded. This change makes it consistent with the same step in `demoApp.step2`, also improving its "back" behavior.


gastonfournier (approved):
LG!

PR #9793, opened 2025-04-17
feat: run cypress against current branch

gastonfournier:
## About the changes
Currently, we're running against the older version of our UI. When making changes to it we want to make sure we're testing the current code

**Details in comments**


.github/docker-compose.test.yml
```diff
@@ -14,11 +14,27 @@
...
+      npx wait-on http://unleash:4242/health && \
+      echo \"Unleash is up, starting frontend\" && \
+      yarn --cwd ./frontend run dev"
+    ports:
+      - "3000:3000"
+    environment:
+      UNLEASH_API: "http://unleash:4242/"
   unleash:
-    image: 726824350591.dkr.ecr.eu-central-1.amazonaws.com/unleash-enterprise:latest
+    image: unleashorg/unleash-enterprise:latest
```
gastonfournier, line 34: Use the new publicly available image

.github/docker-compose.test.yml
```diff
@@ -14,11 +14,27 @@
...
     pull_policy: "always"
-    ports:
-      - "4242:4242"
+    expose:
+      - "4242"
```
gastonfournier, line 37: Expose unleash server only internally

.github/docker-compose.test.yml
```diff
@@ -14,11 +14,27 @@
...
+    working_dir: /app
+    volumes:
+      - ..:/app
+    command: |
+      sh -c "echo \"Building frontend\" && \
+      yarn --cwd ./frontend install && \
+      echo \"Waiting for Unleash to start\" && \
+      npx wait-on http://unleash:4242/health && \
+      echo \"Unleash is up, starting frontend\" && \
+      yarn --cwd ./frontend run dev"
```
gastonfournier, line 28: This is what's needed to run the frontend. 

This also made me realize that our frontend depends on code from the server:
`Failed to resolve import "@server/util/escape-regex" from "src/component/common/Highlighter/Highlighter.tsx". Does the file exist?`
so we need to load the whole server into this docker image

.github/workflows/e2e.frontend.yaml
```diff
@@ -23,7 +23,6 @@ jobs:
         uses: actions/checkout@v4
       - name: Start Unleash test instance
         run: |
-          curl https://app.unleash-hosted.com/docker-login/token/${{ secrets.ECR_ENTERPRISE_TOKEN }} | docker login --username AWS --password-stdin 726824350591.dkr.ecr.eu-central-1.amazonaws.com
```
gastonfournier, line 26: Not needed anymore with the new image

.github/workflows/e2e.frontend.yaml
```diff
@@ -32,6 +31,6 @@ jobs:
         with:
           working-directory: frontend
           env: AUTH_USER=admin,AUTH_PASSWORD=unleash4all
-          config: baseUrl=http://localhost:4242
+          config: baseUrl=http://localhost:3000
```
gastonfournier, line 34: Now use the UI port

frontend/package.json
```diff
@@ -17,7 +17,6 @@
     "start:sandbox": "UNLEASH_API=https://sandbox.getunleash.io/ VITE_TEST_REDIRECT=true UNLEASH_BASE_PATH=/pro/ yarn run start",
     "start:sandbox:enterprise": "UNLEASH_API=https://sandbox.getunleash.io/ VITE_TEST_REDIRECT=true UNLEASH_BASE_PATH=/enterprise/ yarn run start",
     "start:demo2": "UNLEASH_API=https://sandbox.getunleash.io/ UNLEASH_BASE_PATH=/demo2/ yarn run start",
-    "start:enterprise": "UNLEASH_API=https://unleash.herokuapp.com VITE_TEST_REDIRECT=true yarn run start",
```
gastonfournier, line 20: Heroku is no longer available

frontend/package.json
```diff
@@ -30,7 +29,6 @@
     "ts:check": "tsc",
     "e2e": "NODE_OPTIONS=\"${NODE_OPTIONS:-} --no-experimental-fetch\" yarn run cypress open --config baseUrl='http://localhost:3000' --env AUTH_USER=admin,AUTH_PASSWORD=unleash4all",
     "e2e:oss": "yarn --cwd frontend run cypress run --spec \"cypress/oss/**/*.spec.ts\" --config baseUrl=\"http://localhost:${EXPOSED_PORT:-4242}\" --env AUTH_USER=admin,AUTH_PASSWORD=unleash4all",
-    "e2e:heroku": "NODE_OPTIONS=\"${NODE_OPTIONS:-} --no-experimental-fetch\" yarn run cypress open --config baseUrl='https://unleash.herokuapp.com' --env AUTH_USER=admin,AUTH_PASSWORD=unleash4all",
```
gastonfournier, line 33: Heroku is no longer available

.github/docker-compose.test.yml
```diff
@@ -14,11 +14,35 @@
 services:
   # The Unleash server contains the Unleash configuration and
   # communicates with server-side SDKs and the Unleash Proxy
+  frontend:
+    image: node:20-alpine
+    working_dir: /app
+    volumes:
+      - ..:/src
+    command: |
+      sh -c " cp -r /src/* /app && \
```
gastonfournier, line 23: Because this is mounted as a volume, installing dependencies in this node-20-alpine image results in different binaries being installed than those for ubuntu-latest. This copy makes sure we don't touch the source files when building

.github/docker-compose.test.yml
```diff
@@ -14,15 +14,27 @@
...
   # The Unleash server contains the Unleash configuration and
   # communicates with server-side SDKs and the Unleash Proxy
+  frontend:
+    image: node:alpine
+    working_dir: /app
+    command: npx http-server ./build -p 80 --cors --proxy http://unleash:4242
+    ports:
+      - "3000:80"
+    volumes:
+      - ../frontend/build:/app:ro
```
gastonfournier, line 24: Frontend is built before running docker compose up (line 25 of .github/workflows/e2e.frontend.yaml)

Then it's served as static content and everything else is proxy-ed to unleash server

.github/workflows/e2e.frontend.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
           - groups/groups.spec.ts
           - projects/access.spec.ts
           - segments/segments.spec.ts
-          - login/login.spec.ts
+          #- login/login.spec.ts
```
gastonfournier, line 16: @Tymek this spec seems to be new and it's not working in CI. I'm commenting it out for now.

src/lib/util/rewriteHTML.ts
```diff
@@ -9,7 +9,7 @@ export const rewriteHTML = (
     result = result.replace(/::baseUriPath::/gi, rewriteValue);
     result = result.replace(/::cdnPrefix::/gi, cdnPrefix || '');
 
-    const faviconPrefix = cdnPrefix ? 'https://cdn.getunleash.io' : '';
+    const faviconPrefix = cdnPrefix ? cdnPrefix : '';
```
gastonfournier, line 12: This could be an oversight... but if we have a cdn prefix I think we should use it. Maybe it was always the same originally but this seems correct

Tymek: Maybe not?
`https://cdn.getunleash.io/v4.1.0/favicon.ico` is not available, but `https://cdn.getunleash.io/favicon.ico` is.

.github/docker-compose.test.yml
Tymek: If we have all node packages available, maybe we can run `start:prod --port 80` or `./node_modules/.bin/vite preview --port 80`?

Tymek (approved):

frontend/package.json
nunogois: Wonder if we should clean up our scripts more aggressively in a separate PR, maybe as part of v7.

src/lib/util/rewriteHTML.test.ts
```diff
@@ -49,7 +49,7 @@ test('rewriteHTML swaps out faviconPath if cdnPrefix is set', () => {
     const result = rewriteHTML(input, '', 'https://cdn.getunleash.io/v4.1.0');
     expect(
         result.includes(
-            '<link rel="icon" href="https://cdn.getunleash.io/favicon.ico" />',
+            '<link rel="icon" href="https://cdn.getunleash.io/v4.1.0/favicon.ico" />',
```
nunogois, line 52: Based on https://github.com/Unleash/unleash/pull/9793/files#r2049346256 — Maybe this needs to be reverted?

nunogois (approved):
LGTM

frontend/package.json
gastonfournier: Maybe? But this is currently useless so I see no point in delaying its removal

src/lib/util/rewriteHTML.ts
gastonfournier: Good catch! But maybe it's just a coincidence that that version does not have it. I've checked in our current version and I can find the icon under https://cdn.getunleash.io/unleash/commits/d60ea1acd45e30ddb85e0996dcc24c7bb27e7ff7/favicon.ico on our continuous build. I'll check if it's also present in our recent releases

.github/docker-compose.test.yml
gastonfournier: I considered it, but I think it is more realistic to use static files cause that's how we usually serve our frontend in production.

PR #9800, opened 2025-04-17
chore!: removing userId strategy for new installations of Unleash

gastonfournier:
This removes a strategy that was already deprecated, but only for new installations.

I tested starting with an installation with this strategy being used and then updating, and I was still able to edit the strategy, so this should not impact current users.

On a fresh install the strategy is no longer available.


src/migrations/20210421133845-add-sort-order-to-strategies.js
```diff
@@ -6,7 +6,6 @@ exports.up = function (db, cb) {
     ALTER TABLE strategies ADD COLUMN sort_order integer DEFAULT 9999;
     UPDATE strategies SET sort_order = 0 WHERE name = 'default';
     UPDATE strategies SET sort_order = 1 WHERE name = 'flexibleRollout';
-    UPDATE strategies SET sort_order = 2 WHERE name = 'userWithId';
```
nunogois, line 9: I'm pretty sure we shouldn't update existing migrations and that this would silently fail if it can't find any row with `name = 'userWithId'`.

src/migrations/20210421135405-add-display-name-and-update-description-for-strategies.js
```diff
@@ -6,7 +6,6 @@ exports.up = function (db, cb) {
     ALTER TABLE strategies ADD COLUMN display_name text;
     UPDATE strategies SET display_name = 'Standard', description = 'The standard strategy is strictly on / off for your entire userbase.' WHERE name = 'default';
     UPDATE strategies SET display_name = 'Gradual rollout', description = 'Roll out to a percentage of your userbase, and ensure that the experience is the same for the user on each visit.' WHERE name = 'flexibleRollout';
-    UPDATE strategies SET display_name = 'UserIDs', description = 'Enable the feature for a specific set of userIds.' WHERE name = 'userWithId';
```
nunogois, line 9: Same as https://github.com/Unleash/unleash/pull/9800/files#r2052191420

src/migrations/20230420125500-v5-strategy-changes.js
```diff
@@ -9,13 +9,9 @@ exports.up = function (db, callback) {
             and deprecated
             and not exists (select * from feature_strategies where strategy_name = name limit 1);
 
-        -- deprecate strategies on v5
-        update strategies set deprecated = true where name in ('userWithId');
-
         -- update strategy descriptions and sort order
         update strategies set sort_order = 1, description = 'This strategy turns on / off for your entire userbase. Prefer using "Gradual rollout" strategy (100%=on, 0%=off).' WHERE name = 'default';
         update strategies set sort_order = 0 WHERE name = 'flexibleRollout';
-        update strategies set description = 'Enable the feature for a specific set of userIds. Prefer using "Gradual rollout" strategy with user id constraints.' WHERE name = 'userWithId';
```
nunogois, line 18: Same as https://github.com/Unleash/unleash/pull/9800/files#r2052191420

src/migrations/20230420125500-v5-strategy-changes.js
```diff
@@ -9,13 +9,9 @@ exports.up = function (db, callback) {
             and deprecated
             and not exists (select * from feature_strategies where strategy_name = name limit 1);
 
-        -- deprecate strategies on v5
-        update strategies set deprecated = true where name in ('userWithId');
```
nunogois, line 13: Same as https://github.com/Unleash/unleash/pull/9800/files#r2052191420

src/migrations/default-strategies.json
```diff
@@ -4,18 +4,6 @@
...
-    "description": "Active for users with a userId defined in the userIds-list",
-    "parameters": [
-      {
-        "name": "userIds",
-        "type": "list",
-        "description": "",
-        "required": false
-      }
-    ]
-  },
```
nunogois, line 18: This affects an existing migration, `20170211090541-add-default-strategies.js`, so same as /Users/nunogois/dev/unleash/unleash/src/migrations/20170211090541-add-default-strategies.js

src/migrations/20210421133845-add-sort-order-to-strategies.js
nunogois: Instead, we should add a new migration that removes the strategy.

gastonfournier: Yes, it's intentional, that way we only affect new installations and don't change existing installations. We don't want to remove the strategy on existing installations, only if it's new.

It's true that the other migrations would not apply if you don't find the strategy due to the where filter, but it feels cleaner to remove those.

chriswk (approved):
Looks good to me. We still support this via constraints, so no need for a specific user id strategy. Have a :+1:

gastonfournier:
> Looks good to me. We still support this via constraints, so no need for a specific user id strategy. Have a 👍

yes! added that to docs :)

PR #9806, opened 2025-04-21
Remove user ids strategy 2

nunogois:
Expands on the work done in https://github.com/Unleash/unleash/pull/9800


src/migrations/20250422070810-drop-userwithid-strategy.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `
+            DELETE FROM strategies
+            WHERE name = 'userWithId' AND built_in = 1
```
nunogois, line 7: Unlike e.g. https://github.com/Unleash/unleash/blob/bc7856a23a422b6232766168edce3113dab25a6f/src/migrations/20230420125500-v5-strategy-changes.js#L10
Here we're deleting the strategy even if it's in use by a feature. Based on #9800 it seems like this is what we want to do. However, if it's not, this is something we probably need to change.

gastonfournier: I think we don't want to delete it. The idea is:
1. New installations will not receive it
2. Old installations requires manual action from the user

PR #9807, opened 2025-04-21
feat: enter will select the first filtered value

sjaanus:
image


frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
```diff
@@ -196,7 +203,9 @@ export const RestrictiveLegalValues = ({
                     </>
                 }
             />
-            <ConstraintValueSearch filter={filter} setFilter={setFilter} />
+            <div onKeyDown={handleSearchKeyDown}>
```
sjaanus, line 206: Best option is to wrap it with div, so no need to modify shared component.

PR #9816, opened 2025-04-22
chore: remove demo plans dialog

nunogois:
https://linear.app/unleash/issue/2-3531/remove-pricing-pop-up

We decided to remove this dialog in favor of linking directly to https://www.getunleash.io/pricing


gastonfournier (approved):
LG!

PR #9817, opened 2025-04-23
Always show the value list + hide "add values" on non-free text entries

thomasheartman:
Removes the condition to hide the value list if we use legal values. 

In doing so, I also realized that focus handling when you delete the last item in the constraint values list doesn't work if the add values button isn't there (which it shouldn't be for legal values and more). So I've hidden the add values button when it doesn't do anythnig helpful (or for cases where we don't have designs yet). In cases where you don't have the add values button and you delete the last constraint value, we'll move the focus to the "delete constraint" button (that was easier than making sure we pass refs all the way down into the operator select, but we can change that later).

To facilitate this (refs coming from the parent component), I refactored the value list component to accept the add values widget as a child (and extracted it to its own file).


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/AddValuesWidget.tsx
thomasheartman, line 1: This is just extracted from ValueList. No other changes.

sjaanus (approved):

PR #9818, opened 2025-04-23
feat: legal values context fields should only have correct operators

sjaanus:
image

instead of showing following

image


frontend/src/component/common/NewConstraintAccordion/NewConstraintAccordionList/NewConstraintAccordionList.tsx
```diff
@@ -44,37 +41,6 @@ const StyledContainer = styled('div')({
     display: 'flex',
     flexDirection: 'column',
 });
-
-export const useConstraintAccordionList = (
```
sjaanus, line 48: Unused code path.

PR #9824, opened 2025-04-23
feat: lifecycle count query

kwasniew:
## About the changes

API to return counts for each lifecycle stage:
Screenshot 2025-04-23 at 16 01 18

### DIFF
--- a/src/lib/features/feature-lifecycle/feature-lifecycle-count-controller.ts
+++ b/src/lib/features/feature-lifecycle/feature-lifecycle-count-controller.ts
@@ -1,0 +1,80 @@
+import {
+    type IFeatureLifecycleReadModel,
+    type IUnleashConfig,
+    type IUnleashServices,
+    NONE,
+} from '../../types';
+import type { OpenApiService } from '../../services';
+import { createResponseSchema, getStandardResponses } from '../../openapi';
+import Controller from '../../routes/controller';
+import type { Request, Response } from 'express';
+import {
+    type FeatureLifecycleCountSchema,
+    featureLifecycleCountSchema,
+} from '../../openapi/spec/feature-lifecycle-count-schema';
+
+export default class FeatureLifecycleCountController extends Controller {
+    private featureLifecycleReadModel: IFeatureLifecycleReadModel;
+
+    private openApiService: OpenApiService;
+
+    constructor(
+        config: IUnleashConfig,
+        {
+            openApiService,
+            featureLifecycleReadModel,
+        }: Pick<
+            IUnleashServices,
+            'openApiService' | 'featureLifecycleReadModel'
+        >,
+    ) {
+        super(config);
+        this.featureLifecycleReadModel = featureLifecycleReadModel;
+        this.openApiService = openApiService;
+
+        this.route({
+            method: 'get',
+            path: '/count',
+            handler: this.getStageCount,
+            permission: NONE,
+            middleware: [
+                openApiService.validPath({
+                    tags: ['Unstable'],
+                    summary: 'Get all features lifecycle stage count',
+                    description:
+                        'Information about the number of features in each lifecycle stage.',
+                    operationId: 'getFeatureLifecycleStageCount',
+                    responses: {
+                        200: createResponseSchema(
+                            'featureLifecycleCountSchema',
+                        ),
+                        ...getStandardResponses(401, 403, 404),
+                    },
+                }),
+            ],
+        });
+    }
+
+    async getStageCount(
+        _: Request<any, any, any, any>,
+        res: Response<FeatureLifecycleCountSchema>,
+    ): Promise<void> {
+        const stageCounts =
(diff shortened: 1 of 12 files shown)


src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
```diff
@@ -115,6 +112,10 @@ const expectFeatureStage = async (featureName: string, stage: StageName) => {
     });
 };
 
+const getFeaturesLifecycleCount = async () => {
+    return app.request.get(`/api/admin/lifecycle/count`).expect(200);
```
kwasniew, line 116: I'm wondering if this is the correct name for the new API

src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
```diff
@@ -173,6 +174,15 @@ test('should return lifecycle stages', async () => {
 
     eventStore.emit(FEATURE_REVIVED, { featureName: 'my_feature_a' });
     await reachedStage('my_feature_a', 'initial');
+
+    const { body: lifecycleCount } = await getFeaturesLifecycleCount();
+    expect(lifecycleCount).toEqual({
```
kwasniew, line 179: example of the reponse

src/lib/services/index.ts
```diff
@@ -490,6 +490,7 @@ export const createServices = (
         projectStatusService,
         transactionalUserSubscriptionsService,
         uniqueConnectionService,
+        featureLifecycleReadModel,
```
kwasniew, line 493: we don't have separate grouping of read models and services so adding it to services. controller uses the read model directly since it's read only path

src/lib/features/feature-lifecycle/feature-lifecycle-count-controller.ts
```diff
@@ -0,0 +1,80 @@
...
+    async getStageCount(
+        _: Request<any, any, any, any>,
+        res: Response<FeatureLifecycleCountSchema>,
+    ): Promise<void> {
+        const stageCounts =
+            await this.featureLifecycleReadModel.getStageCount();
+
+        const result: Record<string, number> = stageCounts.reduce(
+            (acc, { stage, count }) => {
+                acc[stage === 'pre-live' ? 'preLive' : stage] = count;
```
Tymek, line 67: this mapping is present elsewhere, right?

src/lib/features/feature-lifecycle/feature-lifecycle.e2e.test.ts
Tymek: I can't think of anything better

src/lib/features/feature-lifecycle/feature-lifecycle-count-controller.ts
```diff
@@ -0,0 +1,80 @@
...
+                            'featureLifecycleCountSchema',
+                        ),
+                        ...getStandardResponses(401, 403, 404),
+                    },
+                }),
+            ],
+        });
+    }
+
+    async getStageCount(
```
Tymek, line 58: I'm not sure it has to be memoized, but I didn't check how expensive this query is

Tymek (approved):

src/lib/features/feature-lifecycle/feature-lifecycle-count-controller.ts
kwasniew: yes but in the feature toggle directory for stats. in the feature lifecycle directory itself it's the first time

src/lib/features/feature-lifecycle/feature-lifecycle-count-controller.ts
kwasniew: i was checking the explain plan and it was very cheap. I can add query tracking in the next PR if we find in the UI that those are slow. Also we can set infrequent interval in the SWR hook since this doesn't need to update very often

src/lib/services/index.ts
gastonfournier: Hi, just FYI, I'm modifying this, as this read model is a store, not a service. So with ESM it messes things a bit up, but I'll just explicitly add the read model dependency, not through the services interface.

kwasniew: I think there was a problem with testing since tests access either stores and services and we don't have another grouping for read models

gastonfournier: The biggest problem was that when we're creating controllers, we only have access to services interface and the db. If we start using read models for simple cases I think it makes sense to also have access to stores (as read models are currently exposed there), then it's a matter of propagating stores to the controller constructor. Does it make sense? 

Alternatively, we could have a list of read models (neither stores nor services)

kwasniew: Yeah, we can piggyback on the stores we have too.

PR #9825, opened 2025-04-23
feat: change requests, use new accordion list not legacy

sjaanus:
This should be final piece where old accordion is used.

image


frontend/src/component/common/NewConstraintAccordion/NewConstraintAccordionList/NewConstraintAccordionList.tsx
```diff
@@ -150,16 +151,25 @@ export const NewConstraintAccordionList = forwardRef<
...
-                                constraint={constraint}
-                                onCancel={onCancel.bind(null, index)}
-                                onDelete={onRemove?.bind(null, index)}
-                                onSave={onSave!.bind(null, index)}
-                                onAutoSave={onAutoSave?.(
-                                    constraint[constraintId],
-                                )}
-                            />
+                            state.get(constraint)?.editing ? (
+                                <EditableConstraintWrapper
```
sjaanus, line 155: So this is interesting. Previously NewConstraintAccordion handled both, edit and view. Now EditableConstraintWrapper  only handles edit, so I need to manually use ConstraintAccordionView for view.

PR #9827, opened 2025-04-23
chore: add terraform to integrations

nunogois:
https://linear.app/unleash/issue/2-2565/add-terraform-to-list-of-integrations

Adds Terraform to our integrations list.

It links to: https://docs.getunleash.io/reference/terraform

image


gastonfournier (approved):
Looks like a good initial step. Maybe in the future we can do a "Learn more" like with Jira but let's not over complicate it now.

nunogois:
> Looks like a good initial step. Maybe in the future we can do a "Learn more" like with Jira but let's not over complicate it now.

Yes, I thought about it, but it's not easy to come up with content that doesn't end up being redundant / potentially outdated compared to just linking to the docs.

Like you said, maybe something for later 👍

PR #9828, opened 2025-04-23
chore: fix admin routes should respect plan data

nunogois:
https://linear.app/unleash/issue/2-2852/sidebar-bug-with-enterprisepro-only-route-constraints

Fixes an issue where admin routes didn't respect plan data if their flag was enabled.

First noticed here: https://github.com/Unleash/unleash/pull/8469#discussion_r1804361222

Issue was that only `adminRoutes` respected plan data. `mainNavRoutes` and `primaryRoutes` did not follow the same filtering logic.

We can probably clean this up even further in the future, but didn't want to extend the PR too much.

Also adds tests to validate the intended behavior.

### DIFF
--- a/frontend/src/component/admin/filterAdminRoutes.test.ts
+++ b/frontend/src/component/admin/filterAdminRoutes.test.ts
@@ -1,9 +1,9 @@
-import { filterAdminRoutes } from './filterAdminRoutes';
+import { filterRoutesByPlanData } from './filterRoutesByPlanData';
 
-describe('filterAdminRoutes - open souce routes', () => {
+describe('filterRoutesByPlanData - open souce routes', () => {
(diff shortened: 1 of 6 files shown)


daveleek (approved):
Nice one, are we sure exposing these wasn't intentional? (we do have the enterprise badge and all that)

nunogois:
> Nice one, are we sure exposing these wasn't intentional? (we do have the enterprise badge and all that)

The tests should cover for that, specifically this one:

"filters routes based on pro access (still shows enterprise routes with badge" — If you're only Pro we still show Enterprise routes, so we can show them with the premium badge.

However if you notice any scenario in the tests that is not intended, or would like me to add one hypothesis, let me know.

PR #9829, opened 2025-04-24
feat: 1-3652/legal value selector visual update

thomasheartman:
Handles the visual changes for the legal value selector widget.

Before:
image

After:
image

I'm still working on improving the functionality of selecting from the search input and not losing focus when you select/deselect an item (both of these work (mostly) as expected on hosted, so we've introduced a regression somewhere).

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/ConstraintValueSearch.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/ConstraintValueSearch.tsx
@@ -6,39 +6,42 @@ import { ScreenReaderOnly } from 'component/common/ScreenReaderOnly/ScreenReader
 interface IConstraintValueSearchProps {
     filter: string;
     setFilter: React.Dispatch<React.SetStateAction<string>>;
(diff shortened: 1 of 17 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/ConstraintValueSearch.tsx
thomasheartman, line 1: Mostly a copy of the old constraint value search, but using a hidden label instead and supporting an "onKeyDown" prop. The point of that prop is to allow this to perform the "check" function in the legal values selector. The functionality isn't perfectly implemented yet, but I wanna do that in a follow-up.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/LegalValuesSelector.tsx
thomasheartman, line 1: Based on RestricitiveLegalValues, but with a couple notable changes:
- Item order in the header
- Borderless checkboxes
- Additionally, there'll be more in terms of functionality coming up later.

sjaanus (approved):

PR #9830, opened 2025-04-24
feat: remove warning boxes for empty values

sjaanus:
We had issues, where when you selected the operator, these boxes were jumping up. The problem was that the illegal value checker was marking empty values as illegal also.

Now empty value is not included in illegal values.

image


frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
```diff
@@ -47,7 +47,9 @@ export const getIllegalValues = (
 ) => {
     const deletedValuesSet = getLegalValueSet(deletedLegalValues);
 
-    return constraintValues.filter((value) => deletedValuesSet.has(value));
+    return constraintValues.filter(
+        (value) => value !== '' && deletedValuesSet.has(value),
```
thomasheartman, line 51: This can probably be simplified just to this, right? 

```suggestion
        (value) => value && deletedValuesSet.has(value),
```

But also ... why would the deleted values set have `''` as a value? 🤷🏼 Eh.

thomasheartman (approved):

PR #9833, opened 2025-04-24
feat: remove old instances based on last seen not created at

kwasniew:
## About the changes

When we clean unused instances I'd suggest we do it based on last seen date not created date.

Screenshot 2025-04-24 at 13 29 42


src/lib/db/client-instance-store.ts
```diff
@@ -62,7 +62,7 @@ export default class ClientInstanceStore implements IClientInstanceStore {
 
     async removeInstancesOlderThanTwoDays(): Promise<void> {
         const rows = await this.db(TABLE)
-            .whereRaw("created_at < now() - interval '2 days'")
+            .whereRaw("last_seen < now() - interval '2 days'")
```
kwasniew, line 65: at db level it default to now

PR #9835, opened 2025-04-24
feat: remove inactive applications

kwasniew:
## About the changes

Remove inactive applications (last_seen older than 30 days) from the application list

Screenshot 2025-04-24 at 13 54 42

There is a delete cascade to client_applications_usage so this will be removed too.

The new capability is behind a flag.

## Discussion points

There's seet_at and updated_at column and they seem to have the same value

### DIFF
--- a/src/lib/db/client-applications-store.ts
+++ b/src/lib/db/client-applications-store.ts
@@ -449,4 +449,16 @@ export default class ClientApplicationsStore
             }));
         }
     };
+
+    async removeInactiveApplications(): Promise<number> {
+        const rows = await this.db(TABLE)
+            .whereRaw("seen_at < now() - interval '30 days'")
+            .del();
+
+        if (rows > 0) {
+            this.logger.debug(`Deleted ${rows} applications`);
+        }
+
+        return rows;
+    }
 }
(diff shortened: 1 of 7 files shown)


src/lib/db/client-applications-store.ts
```diff
@@ -449,4 +449,16 @@ export default class ClientApplicationsStore
             }));
         }
     };
+
+    async removeInactiveApplications(): Promise<number> {
+        const rows = await this.db(TABLE)
+            .whereRaw("seen_at < now() - interval '30 days'")
```
kwasniew, line 455: we could also use updated_at

sjaanus (approved):

src/lib/db/client-applications-store.ts
sjaanus: It makes sense to use both.

kwasniew: yeah, good point. We can look at either seen_at which is optional and updated_at which is required

kwasniew: ok I checked the code and we always have seen_at set in the app:      ```updated_at: input.updatedAt || new Date(),
        seen_at: input.lastSeen || new Date(),```

PR #9836, opened 2025-04-24
feat: separate constraint edit/view screens

sjaanus:
Next PR is to start using these.

### DIFF
--- a/frontend/src/component/common/NewConstraintAccordion/ConstraintsList/ViewableConstraintsList.tsx
+++ b/frontend/src/component/common/NewConstraintAccordion/ConstraintsList/ViewableConstraintsList.tsx
@@ -1,0 +1,72 @@
+import { forwardRef } from 'react';
+import { styled } from '@mui/material';
+import type { IConstraint } from 'interfaces/strategy';
+import produce from 'immer';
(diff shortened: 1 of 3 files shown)


frontend/src/component/common/NewConstraintAccordion/ConstraintsList/ViewableConstraintsList.tsx
```diff
@@ -0,0 +1,72 @@
...
+});
+
+export const ViewableConstraintsList = forwardRef<
+    IViewableConstraintsListRef | undefined,
+    IViewableConstraintsListProps
+>(({ constraints, setConstraints }, ref) => {
+    const { context } = useUnleashContext();
+    const state = useWeakMap<IConstraint, IViewableConstraintsListItemState>();
+
+    const onEdit =
```
sjaanus, line 36: I could not get rid of these yet, as the problem is that these are used very deep in some components, but its a start.

thomasheartman: Yup, I noticed when I was looking. But I'm hopuful for later 😄

thomasheartman (approved):

PR #9837, opened 2025-04-24
chore: unknown flags

nunogois:
https://linear.app/unleash/issue/2-3406/hold-unknown-flags-in-memory-and-show-them-in-the-ui-somehow

This PR introduces a suggestion for a “unknown flags” feature.

When clients report metrics for flags that don’t exist in Unleash (e.g. due to typos), we now track a limited set of these unknown flag names along with the appnames that reported them. The goal is to help users identify and clean up incorrect flag usage across their apps.

We store up to 10 unknown flag + appName combinations, keeping only the most recent reports. Data is collected in-memory and flushed periodically to the DB, with deduplication and merging to ensure we don’t exceed the cap even across pods.

We were especially careful to make this implementation defensive, as unknown flags could be reported in very high volumes. Writes are batched, deduplicated, and hard-capped to avoid DB pressure.

No UI has been added yet — this is backend-only for now and intended as a step toward better visibility into client misconfigurations.

I would suggest starting with a simple banner that opens a dialog showing the list of unknown flags and which apps reported them.

image

### DIFF
--- a/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
+++ b/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
@@ -125,37 +125,49 @@ export default class ClientMetricsServiceV2 {
         validatedToggleNames: string[];
         unknownToggleNames: string[];
     }> {
+        let toggleNamesToValidate: string[] = toggleNames;
         let unknownToggleNames: string[] = [];
 
-        if (this.flagResolver.isEnabled('filterExistingFlagNames')) {
+        const shouldFilter = this.flagResolver.isEnabled(
+            'filterExistingFlagNames',
+        );
+        const shouldReport = this.flagResolver.isEnabled('reportUnknownFlags');
+
+        if (shouldFilter || shouldReport) {
             try {
-                const validNames = await this.cachedFeatureNames();
+                const existingFlags = await this.cachedFeatureNames();
 
                 const existingNames = toggleNames.filter((name) =>
-                    validNames.includes(name),
+                    existingFlags.includes(name),
                 );
-                if (this.flagResolver.isEnabled('reportUnknownFlags')) {
-                    unknownToggleNames = toggleNames
-                        .filter((name) => !existingNames.includes(name))
-                        .slice(0, MAX_UNKNOWN_FLAGS);
+                const nonExistingNames = toggleNames.filter(
+                    (name) => !existingFlags.includes(name),
+                );
+
+                if (shouldFilter) {
+                    toggleNamesToValidate = existingNames;
+
+                    if (existingNames.length !== toggleNames.length) {
+                        this.logger.info(
+                            `Filtered out ${toggleNames.length - existingNames.length} toggles with non-existing names`,
+                        );
+                    }
                 }
-                if (existingNames.length !== toggleNames.length) {
-                    this.logger.info(
-                        `Filtered out ${toggleNames.length - existingNames.length} toggles with non-existing names`,
+
+                if (shouldReport) {
+                    unknownToggleNames = nonExistingNames.slice(
+                        0,
+                        MAX_UNKNOWN_FLAGS,
                     );
                 }
-
-                const validatedToggleNames =
-                    await this.filterValidToggleNames(existingNames);
-
-                return { validatedToggleNames, unknownToggleNames };
             } catch (e) {
                 this.logger.error(e);
             }
         }
 
-        const validatedToggleNames =
-            await this.filterValidToggleNames(toggleNames);
+        const validatedToggleNames = await this.filterValidToggleNames(
+            toggleNamesToValidate,
+        );
 
         return { validatedToggleNames, unknownToggleNames };
     }
--- a/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
+++ b/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
@@ -26,6 +26,10 @@ import {
 import type { ClientMetricsSchema } from '../../../../lib/openapi';
 import { nameSchema } from '../../../schema/feature-schema';
 import memoizee from 'memoizee';
+import {
+    MAX_UNKNOWN_FLAGS,
+    type UnknownFlagsService,
+} from '../unknown-flags/unknown-flags-service';
 
 export default class ClientMetricsServiceV2 {
     private config: IUnleashConfig;
@@ -35,6 +39,8 @@ export default class ClientMetricsServiceV2 {
     private clientMetricsStoreV2: IClientMetricsStoreV2;
 
     private lastSeenService: LastSeenService;
+
+    private unknownFlagsService: UnknownFlagsService;
 
     private flagResolver: Pick<IFlagResolver, 'isEnabled' | 'getVariant'>;
 
@@ -46,9 +52,11 @@ export default class ClientMetricsServiceV2 {
         { clientMetricsStoreV2 }: Pick<IUnleashStores, 'clientMetricsStoreV2'>,
         config: IUnleashConfig,
         lastSeenService: LastSeenService,
+        unknownFlagsService: UnknownFlagsService,
     ) {
         this.clientMetricsStoreV2 = clientMetricsStoreV2;
         this.lastSeenService = lastSeenService;
+        this.unknownFlagsService = unknownFlagsService;
         this.config = config;
         this.logger = config.getLogger(
             '/services/client-metrics/client-metrics-service-v2.ts',
@@ -113,7 +121,12 @@ export default class ClientMetricsServiceV2 {
         }
     }
 
-    async filterExistingToggleNames(toggleNames: string[]): Promise<string[]> {
+    async filterExistingToggleNames(toggleNames: string[]): Promise<{
+        validatedToggleNames: string[];
+        unknownToggleNames: string[];
+    }> {
+        let unknownToggleNames: string[] = [];
+
         if (this.flagResolver.isEnabled('filterExistingFlagNames')) {
             try {
                 const validNames = await this.cachedFeatureNames();
@@ -121,17 +134,30 @@ export default class ClientMetricsServiceV2 {
                 const existingNames = toggleNames.filter((name) =>
                     validNames.includes(name),
                 );
+                if (this.flagResolver.isEnabled('reportUnknownFlags')) {
+                    unknownToggleNames = toggleNames
+                        .filter((name) => !existingNames.includes(name))
+                        .slice(0, MAX_UNKNOWN_FLAGS);
+                }
                 if (existingNames.length !== toggleNames.length) {
                     this.logger.info(
                         `Filtered out ${toggleNames.length - existingNames.length} toggles with non-existing names`,
                     );
                 }
-                return this.filterValidToggleNames(existingNames);
+
+                const validatedToggleNames =
+                    await this.filterValidToggleNames(existingNames);
+
+                return { validatedToggleNames, unknownToggleNames };
             } catch (e) {
                 this.logger.error(e);
             }
         }
-        return this.filterValidToggleNames(toggleNames);
+
+        const validatedToggleNames =
+            await this.filterValidToggleNames(toggleNames);
+
+        return { validatedToggleNames, unknownToggleNames };
     }
 
     async filterValidToggleNames(toggleNames: string[]): Promise<string[]> {
@@ -181,7 +207,7 @@ export default class ClientMetricsServiceV2 {
                 ),
         );
 
-        const validatedToggleNames =
+        const { validatedToggleNames, unknownToggleNames } =
             await this.filterExistingToggleNames(toggleNames);
 
         this.logger.debug(
@@ -202,6 +228,15 @@ export default class ClientMetricsServiceV2 {
             };
 
             this.config.eventBus.emit(CLIENT_REGISTER, heartbeatEvent);
+        }
+
+        if (unknownToggleNames.length > 0) {
+            const unknownFlags = unknownToggleNames.map((name) => ({
+                name,
+                appName: value.appName,
+                seenAt: value.bucket.stop,
+            }));
+            this.unknownFlagsService.register(unknownFlags);
         }
 
         if (validatedToggleNames.length > 0) {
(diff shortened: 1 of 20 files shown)


src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -204,6 +230,15 @@ export default class ClientMetricsServiceV2 {
             this.config.eventBus.emit(CLIENT_REGISTER, heartbeatEvent);
         }
 
+        if (unknownToggleNames.length > 0) {
+            const unknownFlags = unknownToggleNames.map((name) => ({
+                name,
+                appName: value.appName,
+                seenAt: value.bucket.stop,
```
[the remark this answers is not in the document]
nunogois: `value` is not undefined here, so I think Copilot is wrong on this one.

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -113,25 +121,43 @@ export default class ClientMetricsServiceV2 {
...
                     this.logger.info(
                         `Filtered out ${toggleNames.length - existingNames.length} toggles with non-existing names`,
                     );
                 }
-                return this.filterValidToggleNames(existingNames);
+
+                const validatedToggleNames =
+                    await this.filterValidToggleNames(existingNames);
+
+                return { validatedToggleNames, unknownToggleNames };
```
gastonfournier, line 151: Do we need to return unknownToggleNames? I may find the answer later.

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
```diff
@@ -0,0 +1,61 @@
...
+            await tx(TABLE).delete();
+            if (flags.length > 0) {
+                const rows = flags.map((flag) => ({
+                    name: flag.name,
+                    app_name: flag.appName,
+                    seen_at: flag.seenAt,
+                }));
+                await tx(TABLE).insert(rows);
+            }
+        });
```
gastonfournier, line 37: Maybe this is to keep only 10 when we store... the latest 10... but I'm not convinced with the delete all... cause it needs to lock the whole table...

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export interface IUnknownFlagsStore {
+    replaceAll(flags: UnknownFlag[]): Promise<void>;
+    getAll(): Promise<UnknownFlag[]>;
+    clear(hoursAgo: number): Promise<void>;
+    deleteAll(): Promise<void>;
+}
+
+export class UnknownFlagsStore implements IUnknownFlagsStore {
```
gastonfournier, line 19: Should we use CRUD?

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
```diff
@@ -0,0 +1,61 @@
...
+        return rows.map((row) => ({
+            name: row.name,
+            appName: row.app_name,
+            seenAt: new Date(row.seen_at),
+        }));
+    }
+
+    async clear(hoursAgo: number): Promise<void> {
+        return this.db(TABLE)
+            .whereRaw(`seen_at <= NOW() - INTERVAL '${hoursAgo} hours'`)
```
gastonfournier, line 54: I think seen at is not the best to clear flags, because it might just keep being updated if it's a frequently accessed flag... maybe use created at

src/migrations/20250424185110-unknown-flags.js
```diff
@@ -0,0 +1,23 @@
+
+exports.up = (db, callback) => {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS unknown_flags (
+            name TEXT NOT NULL,
```
gastonfournier, line 6: I would add created_at and allow seen_at to be updated... maybe we should stop writing more once we hit the maximum defined rather than replacing the state...

gastonfournier (commented):
Looks promising

nunogois:
> The code looks good. Have an approval! Small musing: Have we stopped using dbTimer? I notice the last couple of stores there is no timer being used

Can't find `dbTimer` in our source code, but I think you mean the timer some stores have for some of their methods. I can add it here, if you think it adds value.

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
nunogois: In case you haven't found the answer, this is what ends up being registered. See: https://github.com/Unleash/unleash/pull/9837/files#diff-8989b03379ec73aa0cebd572e8d7e819ee39b2fccf0cf4a93769007c73fc5054R239

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
nunogois: It’s true that multiple pods might flush around the same time, which means concurrent reads and writes are possible. That said, the current approach is designed to handle that safely. Each pod fetches the full current state from the database before flushing, merges it with its local in-memory cache, deduplicates by `name` + `appName`, and keeps only the 10 most recent entries. So even if two pods flush close together, the end result is still consistent — just depending on which write lands last.

The DELETE + INSERT is intentional. Since the table is capped at 10 rows and flushes are infrequent, this approach keeps the logic simple and avoids the need for upserts or row-level diffs. 

Also worth noting: the only reason we’re storing this in the DB at all is to ensure persistence across pod restarts and availability across pods. If this were purely local state, the in-memory cache would be enough. 

If the feature grows in scope, we can revisit the update strategy, but for now this tradeoff feels right.

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
nunogois: I think `seen_at` is the right field here, since what we care about is whether the flag has been reported recently. If an unknown flag keeps getting reported, it’s still relevant — so we want to keep it in the list. Using `created_at` would remove flags that are still active just because they were first seen a while ago, which feels like the wrong signal for this use case.

src/migrations/20250424185110-unknown-flags.js
nunogois: I think this goes against the intended design. We’re treating this table as a capped, rolling snapshot of the most recently reported unknown flags, not a historical log. seenAt is updated as needed, and we replace the full state on flush to ensure convergence across pods and avoid partial or stale data.

Allowing unbounded inserts or stopping writes after a cap would make the behavior less predictable and could leave outdated data in place. If the feature evolves into something more persistent or audit-like, we could revisit this, but for now I think the current approach is a better fit for the problem we’re solving.

gastonfournier: > the most recently reported unknown flags, not a historical log

The problem I see with this is that it might not be useful if you constantly see a changing list of flags. You may find an unknown flag and tell your team mate, hey, there's this report about this feature flag that doesn't exist, and you may go back to the list and not find it. 

I rather have a list that you can work on and clean it up. The idea of created_at is to have a combination of recently seen, but keeping the list more stable. As long as the flag was seen recently, we'll keep the oldest one by created_at, if the flag has not been seen recently, we can remove it from the table (despite its created_at), that means that actively accessed flags will usually stay visible.

Contrary, we may expect few unknown flags, and therefore 10 would be more than the total of unknown flags, meaning that optimizing this would be pointless.

Let  me know what you think

nunogois:
Discussed with @gastonfournier and we're interested in gathering some data before we invest more time on this feature.

https://github.com/Unleash/unleash/pull/9837/commits/67aa13087305399b76f6eba4a91b0965ef20c94c adds a new gauge for the total of unknown flags seen in the last 24 hours, if any (max of 10).

image

`reportUnknownFlags` is also no longer dependent on `filterExistingFlagNames`. We've also confirmed that unknown flags includes invalid flags:

image

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
gastonfournier: Makes sense!

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
gastonfournier: I think the idea of created_at might still be valuable. It will tell you the "age" or how long ago this flag was first seen (if we don't remove them). Which makes me think... maybe we should not delete all and recreate... instead, what if we make a count on the table and we only insert up to 10-count? 

We should still delete flags that stopped appearing in the "unknown" group from the table, so it adds some complexity, so let's leave it as an idea here

src/migrations/20250424185110-unknown-flags.js
gastonfournier: Resolved on 1-1 conversation

gastonfournier (approved):
LGTM! I'm still thinking on what the behavior should be... Also, what'd happen when multiple instances try to write to the db, what happens if an instance has seen a flag that doesn't exist in Unleash but another instance has not seen it (running the flush in a single instance would miss this unknown flag)? 

Overall, this will be a nice first step, we don't have to have all the answers atm.

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
nunogois: I'm still not convinced of what `created_at` really gives us here. What's the benefit you're seeing? How valuable is it to know when an unknown flag was first seen? And are the first 10 really more valuable than the last 10?

I'll leave it for now, but I'm curious about your point of view here.

nunogois:
> Also, what'd happen when multiple instances try to write to the db, what happens if an instance has seen a flag that doesn't exist in Unleash but another instance has not seen it (running the flush in a single instance would miss this unknown flag)?

That should be covered by the fact that we're merging the local unknown flags cache with the ones already registered in the DB when we flush: https://github.com/Unleash/unleash/pull/9837/files#diff-544e3c7c0bac8e29a5c0954c0b046e580ad6b11eca94dcb097fac0eedf7a8b7eR61-R79

src/lib/features/scheduler/schedule-services.ts
```diff
@@ -194,4 +195,16 @@ export const scheduleServices = async (
         minutesToMilliseconds(10),
         'uniqueConnectionService',
     );
+
+    schedulerService.schedule(
+        unknownFlagsService.flush.bind(unknownFlagsService),
+        minutesToMilliseconds(2),
+        'flushUnknownFlags',
+    );
```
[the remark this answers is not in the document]
nunogois: But we don't want this to be a single instance job, right? 

Each Unleash instance (pod) will receive unknown flags as part of metrics and store them in their own cache, so it makes sense that each Unleash instance (pod) has their own job to flush that cache into the DB.

If anything we could do it for the "clear", but I don't think doing that gives us much either.

src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
gastonfournier: Feel free to leave it for now. I'm thinking of differentiating the 2 scenarios we discussed:
1. You made a typo or accidentally referenced a flag that doesn't exist. This will stay in code for a long time and therefore the difference between created_at (or when it was first registered as unknown) and seen_at should be large.
2. Your normal workflow implies releasing the flag to prod first and then creating it in Unleash. In this case, the difference between created_at and last_seen should be at max a few hours to 1 or 2 days.

Keeping the created_at as "first time the flag was registered as unknown" comes with challenges, which is why I'm ok dropping this, but I hope my point is clearer now.

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -204,6 +242,15 @@ export default class ClientMetricsServiceV2 {
             this.config.eventBus.emit(CLIENT_REGISTER, heartbeatEvent);
         }
 
+        if (unknownToggleNames.length > 0) {
+            const unknownFlags = unknownToggleNames.map((name) => ({
+                name,
+                appName: value.appName,
+                seenAt: value.bucket.stop,
+            }));
+            this.unknownFlagsService.register(unknownFlags);
```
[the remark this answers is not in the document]
gastonfournier: no, that's fine! I thought it was storing to db.

src/lib/features/scheduler/schedule-services.ts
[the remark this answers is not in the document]
gastonfournier: Yeah, I wanted to comment on the clear one :D but yeah, maybe it's not going to give us much, it was just because I didn't see us using it and it came up in some conversations last week with either @kwasniew @sjaanus or @sastromskis. That we should have the ability of electing a leader for one of these scheduled jobs and I remember we had that already

gastonfournier:
> > Also, what'd happen when multiple instances try to write to the db, what happens if an instance has seen a flag that doesn't exist in Unleash but another instance has not seen it (running the flush in a single instance would miss this unknown flag)?
> 
> That should be covered by the fact that we're merging the local unknown flags cache with the ones already registered in the DB when we flush: https://github.com/Unleash/unleash/pull/9837/files#diff-544e3c7c0bac8e29a5c0954c0b046e580ad6b11eca94dcb097fac0eedf7a8b7eR61-R79

Does this mean we never clean? Imagine a flag `'I_do_not_exist'`, you add it to the db. When is it deleted? If the answer is the clear job every 24hs, that works for me :)

src/lib/features/scheduler/schedule-services.ts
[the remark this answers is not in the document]
nunogois: Honestly I didn't remember / wasn't aware of it being a thing in our codebase until now, so thank you for the reminder.

nunogois:
> Does this mean we never clean? Imagine a flag `'I_do_not_exist'`, you add it to the db. When is it deleted? If the answer is the clear job every 24hs, that works for me :)

Yes, it is. We automatically clean old unknown flags due to our 10 max logic and the 24h clean up job that cleans unknown flags seen >24h ago.

PR #9838, opened 2025-04-25
1-3658/fix legal values selector issues

thomasheartman:
Fixes an issue with the new legal values selector where selecting an item from filtering or changing the checkbox state would move your focus to the top of the page. I think it's because we'd re-render the whole tree because of it, and this would clear your focus selection. To get around it, I've used the existing ResolveInput component. We might want to change this later as we get around to more input components (single values, etc), but for now, I think this is good enough.

As a bonus, I get to delete the most annoying part of the EditableConstraints file 😄 

The constraint still opens in edit mode for now, but I expect that to get resolved once we properly implement the split between editable and non-editable constraints that was started yesterday.

### DIFF
--- a/frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
+++ b/frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
@@ -158,10 +158,12 @@ export const RestrictiveLegalValues = ({
     };
 
     const handleSearchKeyDown = (event: React.KeyboardEvent) => {
-        if (event.key === 'Enter' && filteredValues.length > 0) {
+        if (event.key === 'Enter') {
             event.preventDefault();
-            const firstValue = filteredValues[0].value;
-            onChange(firstValue);
+            if (filteredValues.length > 0) {
+                const firstValue = filteredValues[0].value;
+                onChange(firstValue);
+            }
         }
     };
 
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/RestrictiveLegalValues/RestrictiveLegalValues.tsx
thomasheartman, line 1: This fixes a bug in the logic. the issue was: if you filter, but had no results, then enter would submit the form and save the strategy. Splitting it up like this means that "enter" never submits the form.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
thomasheartman, line 1: Use the existing resolveInput component. I think the core issue was because when you changed the constraints, this whole component would re-render, taking your focus away from the search field (because it had been removed and re-rendered). So we need the component to be declared outside of the editable constraint component.

We could duplicate ResolveInput, but I don't see much value in that for now.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraintWrapper.tsx
thomasheartman, line 1: The existing impl differentiates between the local constraint and the constraint, and I'm guessing there's a good reason for it. So I want to bring the existing constraint back into the mix here to not accidentally mess up some subtle differences. Presumably it's got to do with cleaned values.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/LegalValuesSelector.tsx
thomasheartman, line 1: Same fix as in RestrictiveLegalValues

sjaanus (approved):
LG

PR #9839, opened 2025-04-25
feat: split and clean up constraint lists

sjaanus:
Instead of NewConstraintAccordionList, we are now using EditableConstraintsList and ViewableConstraintsList.

Was able to fully get rid of state, which was hard to understand, but only used to know if constraint is being edited or not.

Most of the changes are in the new files EditableConstraintsList/ViewableConstraintsList, so we can always fall back to old if needed.


frontend/src/component/common/NewConstraintAccordion/ConstraintsList/EditableConstraintsList.tsx
```diff
@@ -1,24 +1,23 @@
...
+}
 
-export const editableConstraintsListId = 'editableConstraintsListId';
+export interface IEditableConstraintsListProps {
+    constraints: IConstraint[];
+    setConstraints?: React.Dispatch<React.SetStateAction<IConstraint[]>>;
+}
 
-export const useEditableConstraintsList = useConstraintsList;
+export const editableConstraintsListId = 'editableConstraintsListId';
```
thomasheartman, line 20: Can we use react's `useId` hook instead? So we never need to worry about having crashing IDs on a page
```suggestion
export const editableConstraintsListId = useId();
```

thomasheartman (approved):
Love it!

PR #9840, opened 2025-04-25
chore: remove flag for global change request config

kwasniew:
## About the changes

Closes #


src/test/e2e/system-user-migration.test.ts
kwasniew, line 1: this test prevents adding new columns and it verifies a transitional migration from 1.5y ago

PR #9842, opened 2025-04-25
Don't show extra input component if we have "add values" button

thomasheartman:
Makes it so that the InputContainer is only rendered if we don't have the add values button up top. We might need to adjust this later (as we get more sketches for other input types, such as single numbers, single semvers etc), but it works for now.

With legal values  (no add values button)
image

Without legal values (but add values button)
image

Because we don't handle single value cases yet, some of those inputs are still stuck in an in-between state:
image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
@@ -9,10 +9,7 @@ import {
     type Operator,
 } from 'constants/operators';
 import useUnleashContext from 'hooks/api/getters/useUnleashContext/useUnleashContext';
-import type {
-    ILegalValue,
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
```diff
@@ -43,35 +40,8 @@ const TopRow = styled('div')(({ theme }) => ({
     flexFlow: 'row nowrap',
     alignItems: 'flex-start',
     justifyItems: 'space-between',
-    borderBottom: `1px dashed ${theme.palette.divider}`,
 }));
 
-const resolveLegalValues = (
```
thomasheartman, line 49: This was unused after the previous PR. I just didn't notice it before now.

sjaanus (approved):

PR #9843, opened 2025-04-25
chore: second design pass for editable constraints

thomasheartman:
Fix a number of visual issues with the main editable constraint component.

I've introduced a few more layers of container nesting to make the layout break the right ways:
- Put everything on the same line when on wide.
- At 700px place selected values on the row below
- From 700px down, when necessary, also wrap operator options

Support super long context names without breaking layout
image

Wrap values at 700px width container:
image

Wrap operator options when necessary
image

Absolutely position delete button to allow to not push it out of the container on narrow screens: 
image

Remove extra focus styling from MUI (darken select background):
Before:
image

image

After:
image
image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
@@ -107,6 +107,9 @@ const InputContainer = styled('div')(({ theme }) => ({
 const StyledSelect = styled(GeneralSelect)(({ theme }) => ({
     fieldset: { border: 'none', borderRadius: 0 },
     maxWidth: '25ch',
+    ':focus-within .MuiSelect-select': {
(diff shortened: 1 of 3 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
```diff
@@ -246,52 +287,55 @@ export const EditableConstraint: FC<Props> = ({
...
+                                        Make match
+                                        {localConstraint.caseInsensitive
+                                            ? ' '
+                                            : ' not '}
+                                        case sensitive
+                                    </ScreenReaderOnly>
+                                </CaseButton>
+                            ) : null}
+                        </OperatorOptions>
+                    </ConstraintOptions>
```
thomasheartman, line 338: This is just indentation changes due to extra wrapping logic

sjaanus (approved):

PR #9845, opened 2025-04-25
feat: count per lifecycle stage

Tymek:
## About the changes

image

Show count per stage, and include count if flags are filtered.


frontend/src/hooks/api/getters/useLifecycleCount/useLifecycleCount.ts
```diff
@@ -0,0 +1,23 @@
...
+    return {
+        lifecycleCount: data,
+        error,
+        loading: !error && !data,
+    };
+};
+
+const fetcher = (path: string) => {
+    return fetch(path)
+        .then(handleErrorResponses('Instance Stats'))
```
kwasniew, line 21: wrong error message

frontend/src/component/feature/FeatureToggleList/FeatureToggleFilters/LifecycleFilters.tsx
```diff
@@ -47,22 +49,45 @@ const lifecycleOptions: {
...
+    if (lifecycle === null) {
+        return (
+            (lifecycleCount.initial || 0) +
+            (lifecycleCount.preLive || 0) +
+            (lifecycleCount.live || 0) +
+            (lifecycleCount.completed || 0)
+        );
+    }
+
+    const key = lifecycle === 'pre-live' ? 'preLive' : lifecycle;
```
kwasniew, line 69: wondering if we should avoid pre-live in frontend altogether and stick to preLive. Does our backend ever return pre-live?

Tymek: I think it does in some places

PR #9846, opened 2025-04-25
feat: Spike frontend applications registration

kwasniew:
## About the changes

When we register frontend applications based on the metrics that they report. 

In the next PR I will add tests and move the registration from the controller to the service itself.


src/lib/features/metrics/instance/instance-service.ts
```diff
@@ -104,13 +108,22 @@ export default class ClientInstanceService {
...
+    }
+
     public async registerClient(
         data: PartialSome<IClientApp, 'instanceId'>,
         clientIp: string,
     ): Promise<void> {
         const value = await clientRegisterSchema.validateAsync(data);
         value.clientIp = clientIp;
         value.createdBy = SYSTEM_USER.username!;
+        value.sdkType = 'backend';
```
kwasniew, line 126: existing code will add data to the new DB column

src/lib/features/frontend-api/frontend-api-service.ts
```diff
@@ -127,6 +127,8 @@ export class FrontendApiService {
             },
             ip,
         );
+
+        return environment;
```
kwasniew, line 131: return env so that other code doesn't have to resolve it again

src/lib/features/frontend-api/frontend-api-controller.ts
```diff
@@ -229,11 +248,31 @@ export default class FrontendAPIController extends Controller {
...
-            req.ip,
-        );
+        const environment =
+            await this.services.frontendApiService.registerFrontendApiMetrics(
+                req.user,
+                req.body,
+                req.ip,
+            );
+
+        if (
```
kwasniew, line 258: I'd like to move this code to frontendApiService in the next PR. Added it here since it required minimal code changes to verify this behind a flag in production

src/lib/features/frontend-api/frontend-api-controller.ts
```diff
@@ -216,6 +228,13 @@ export default class FrontendAPIController extends Controller {
         );
     }
 
+    private resolveProject(user: IUser | IApiUser) {
```
kwasniew, line 231: copied from the backend SDK API controller

src/lib/db/client-instance-store.ts
```diff
@@ -35,6 +35,7 @@ const mapToDb = (client) => ({
     app_name: client.appName,
     instance_id: client.instanceId,
     sdk_version: client.sdkVersion || '',
+    sdk_type: client.sdkType,
```
kwasniew, line 38: we have already applied this migration in all envs

sjaanus (approved):

PR #9853, opened 2025-04-28
Chore/constraint value styling

thomasheartman:
Fixes a few small styling issues with the constraint value chips:
- Background color was wrong
- They shouldn't have a border when they're not focused
 
Different styles: 
1. Keyboard focus
2. Mouse hover
3. No focus
4. No focus
5. Add values button for reference.
image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/ValueList.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/ValueList.tsx
@@ -21,9 +21,8 @@ const ValueChipBase = styled(
 )(({ theme }) => ({
     transition: 'all 0.3s ease',
     outline: `1px solid #0000`,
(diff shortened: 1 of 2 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/ValueList.tsx
```diff
@@ -14,16 +14,26 @@ const StyledList = styled('ul')({
     display: 'contents',
 });
 
+export const baseChipStyles = (theme: Theme) => ({
```
thomasheartman, line 17: I'm not entirely convinced by this extraction. Happy to revert it if you think it's useful. It's only in two places so it's ... actually a step **before** I think I'd normally like to extract it. But at the same time, it's also something that should change together 🤷🏼

sjaanus (approved):

PR #9856, opened 2025-04-29
chore: replace inherit color on menu item hover to common.white

daveleek:
Hover text was black, now it's white


gastonfournier (approved):
LG!

PR #9857, opened 2025-04-29
feat: show backend and frontend sdks separately

kwasniew:
## About the changes

We want to split SDKs into Frontend SDKs and Backend SDKs both int he UI and in the API
Screenshot 2025-04-29 at 09 28 52


sjaanus (approved):
Generally looks good, but lets make instances also more clear, that these are backend instances perhaps or something. 

Just an idea, but something like that

Backend instances connected: 3
Frontend connected: ✓

PR #9860, opened 2025-04-29
feat: view for recently used constraints

sjaanus:
Not connected to storage and do not have use button.

image


thomasheartman (approved):
Very nice! One minor thing (that I'm sure we'll sort out): I **think** the idea was to only show these when you have no constraints added, but let's try this out for now and change it later.

PR #9861, opened 2025-04-29
feat: now updating/editing strategy will store constraints in recents

sjaanus:
Now updating/editing strategy will store constraints in recents


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/RecentlyUsedConstraints/useRecentlyUsedConstraints.test.tsx
```diff
@@ -91,26 +91,22 @@ describe('useRecentlyUsedConstraints', () => {
     });
 
     it('should initialize with empty array when no items in localStorage', () => {
-        const { result } = renderHook(() =>
-            useRecentlyUsedConstraints('test-key'),
-        );
+        const { result } = renderHook(() => useRecentlyUsedConstraints());
```
sjaanus, line 94: useRecentlyUsedConstraints had key removed, as at first I wanted to make generic one, but now its specific to constraints, so it does not need key.

frontend/src/hooks/api/actions/useFeatureStrategyApi/useFeatureStrategyApi.ts
```diff
@@ -2,14 +2,27 @@ import type {
...
 
+    const { addItem: addToRecentlyUsedConstraints } =
+        useRecentlyUsedConstraints();
+
+    const saveConstraintsToRecentlyUsed = (constraints: IConstraint[]) => {
+        if (Array.isArray(constraints)) {
+            constraints.forEach((constraint) => {
+                addToRecentlyUsedConstraints(constraint);
+            });
+        }
```
thomasheartman, line 23: Small suggestion: would it be a better experience to have the addItem function instead be addItems, and you just add whatever. Then that function will take care of looping and all of that? Also encapsulates that logic a bit, I guess.

thomasheartman (approved):

PR #9863, opened 2025-04-29
feat: separate frontend and backend applications from edge

kwasniew:
## About the changes

Changes:
* separates backend and frontend application registration when using batch reporting from edge
* relaxes strategies reporting constraint since it's not longer used

### DIFF
--- a/src/lib/features/metrics/instance/instance-service.ts
+++ b/src/lib/features/metrics/instance/instance-service.ts
@@ -114,7 +114,7 @@ export default class ClientInstanceService {
         this.seenClients[this.clientKey(data)] = data;
     }
 
-    public async registerClient(
+    public async registerBackendClient(
         data: PartialSome<IClientApp, 'instanceId'>,
         clientIp: string,
     ): Promise<void> {
--- a/src/lib/features/metrics/instance/metrics.ts
+++ b/src/lib/features/metrics/instance/metrics.ts
@@ -145,16 +145,21 @@ export default class ClientMetricsController extends Controller {
                 for (const app of applications) {
                     if (
                         app.sdkType === 'frontend' &&
-                        typeof app.sdkVersion === 'string' &&
-                        this.flagResolver.isEnabled('registerFrontendClient')
+                        typeof app.sdkVersion === 'string'
                     ) {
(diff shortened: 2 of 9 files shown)


src/lib/types/model.ts
```diff
@@ -466,7 +466,7 @@ export interface IFrontendClientApp {
     sdkVersion: string;
     sdkType: 'frontend';
     environment: string;
-    projects: string[];
+    projects?: string[];
```
kwasniew, line 469: edge doesn't provide projects and it always reports * as project

src/lib/openapi/spec/bulk-registration-schema.ts
```diff
@@ -70,6 +70,12 @@ export const bulkRegistrationSchema = {
             example: 'unleash-client-java:8.0.0',
             type: 'string',
         },
+        sdkType: {
```
kwasniew, line 73: we have a new optional sdkType that edge will soon support

src/lib/features/metrics/instance/metrics.ts
```diff
@@ -143,12 +143,26 @@ export default class ClientMetricsController extends Controller {
...
-                            app,
-                            clientIp,
-                        ),
-                    );
+                    if (
+                        app.sdkType === 'frontend' &&
+                        typeof app.sdkVersion === 'string' &&
+                        this.flagResolver.isEnabled('registerFrontendClient')
+                    ) {
+                        this.clientInstanceService.registerFrontendClient({
```
kwasniew, line 151: frontend registration is more lightweight since it doesn't send any events to eventBus/eventStore

src/lib/features/metrics/instance/instance-service.ts
```diff
@@ -114,7 +114,7 @@ export default class ClientInstanceService {
         this.seenClients[this.clientKey(data)] = data;
     }
 
-    public async registerClient(
+    public async registerBackendClient(
```
kwasniew, line 117: since we have separate paths for backend and frontend let's be explicit about it

sjaanus (approved):
Nice

PR #9877, opened 2025-04-30
feat: Unleash v7 ESM migration

gastonfournier:
We're migrating to ESM, which will allow us to import the latest versions of our dependencies.


biome.json
```diff
@@ -52,6 +67,7 @@
       "website/build",
       "website/docs/generated",
       "website/global.js",
+      "website",
```
gastonfournier, line 70: @chriswk ignored website temporarily as linter would fail otherwise

.husky/update-openapi-spec-list.js
```diff
@@ -1,28 +1,30 @@
-const fs = require('fs');
-const path = require('path');
-
-const directoryPath = path.join(`${__dirname}/..`, 'src/lib/openapi/spec');
-const indexPath = path.join(directoryPath, 'index.ts');
+import { readdir, writeFileSync } from 'fs';
```
chriswk, line 1:
```suggestion
import { readdir, writeFileSync } from 'node:fs';
```

.husky/update-openapi-spec-list.js
```diff
@@ -1,28 +1,30 @@
...
+import { join, basename, dirname } from 'path';
```
chriswk, line 2:
```suggestion
import { join, basename, dirname } from 'node:path';
```

biome.json
```diff
@@ -12,7 +12,22 @@
...
+        "useImportExtensions": {
+          "level": "error",
+          "options": {
+            "suggestedExtensions": {
+              "ts": {
+                "module": "js",
+                "component": "jsx"
+              },
+              "tsx": {
+                "module": "js",
```
chriswk, line 25:
```suggestion
                "module": "ts",
```
Since we use vite to build here, turns out the correct extension is actually the extension of the file, not what tsc compiles it down to.

frontend/index.js
```diff
@@ -1,9 +1,10 @@
 require('pkginfo')(module, 'version');
-const path = require('path');
-
+import path from 'path';
```
chriswk, line 2:
```suggestion
import path from 'node:path';
```

chriswk (approved):
Let's do this!

PR #9880, opened 2025-05-01
refactor: stabilize frontend apps reporting

kwasniew:
## About the changes

This PR stabilizes reporting of frontend applications/instances.

Details:
* moving application/instance registration from the frontend api controller to frontend api service
* removing unused dependencies
* adding a test showing that apps/instances from the metrics are persisted

### DIFF
--- a/src/lib/features/frontend-api/frontend-api-controller.ts
+++ b/src/lib/features/frontend-api/frontend-api-controller.ts
@@ -4,11 +4,10 @@ import {
     type IFlagResolver,
     type IUnleashConfig,
     type IUnleashServices,
-    type IUser,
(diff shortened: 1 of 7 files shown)


src/lib/features/frontend-api/frontend-api-controller.ts
```diff
@@ -248,28 +240,12 @@ export default class FrontendAPIController extends Controller {
             return;
         }
 
-        const environment =
```
kwasniew, line 251: in the spike controller was doing orchestration of services but in the prod ready version we want controller to use only one service (frontend API service) doing the orchestration

PR #9881, opened 2025-05-02
feat: recently used segments

sjaanus:
Currently it is using the old badge, I need to create new or edit existing to support different actions.

image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegments.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegments.tsx
@@ -3,6 +3,7 @@ import { useRecentlyUsedSegments } from './useRecentlyUsedSegments';
 import type { ISegment } from 'interfaces/segment';
 import { FeatureStrategySegmentChip } from '../FeatureStrategySegmentChip';
 import { useSegments } from 'hooks/api/getters/useSegments/useSegments';
+import { useUiFlag } from 'hooks/useUiFlag';
 
 type IRecentlyUsedSegmentsProps = {
     setSegments?: React.Dispatch<React.SetStateAction<ISegment[]>>;
@@ -29,12 +30,17 @@ export const RecentlyUsedSegments = ({
 }: IRecentlyUsedSegmentsProps) => {
     const { items: recentlyUsedSegmentIds } = useRecentlyUsedSegments();
     const { segments: allSegments } = useSegments();
+    const addEditStrategyEnabled = useUiFlag('addEditStrategy');
 
-    if (recentlyUsedSegmentIds.length === 0 || !setSegments || !allSegments) {
+    if (
+        !addEditStrategyEnabled ||
+        recentlyUsedSegmentIds.length === 0 ||
+        !setSegments ||
+        !allSegments
+    ) {
         return null;
     }
 
-    // Get the full segment objects from the IDs
     const segmentObjects = recentlyUsedSegmentIds
         .map((id) => allSegments.find((segment) => segment.id === id))
         .filter((segment) => segment !== undefined) as ISegment[];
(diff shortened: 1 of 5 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegments.tsx
```diff
@@ -0,0 +1,68 @@
+import { styled, Typography } from '@mui/material';
+import { useRecentlyUsedSegments } from './useRecentlyUsedSegments';
+import type { ISegment } from 'interfaces/segment';
+import { FeatureStrategySegmentChip } from '../FeatureStrategySegmentChip';
+import { useSegments } from 'hooks/api/getters/useSegments/useSegments';
+import { useUiFlag } from 'hooks/useUiFlag';
+
+type IRecentlyUsedSegmentsProps = {
```
thomasheartman, line 8: Not important, but isn't the `I` prefix only for interfaces? I know we've used it for interfaces, but types are typically without the leading I, I think. I'd suggest either removing the `I` or changing it to an interface. My preference would be removing the prefix, but it's up to you.

```suggestion
type RecentlyUsedSegmentsProps = {
```

frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegments.tsx
```diff
@@ -0,0 +1,68 @@
...
+    setSegments?: React.Dispatch<React.SetStateAction<ISegment[]>>;
+};
+
+const StyledContainer = styled('div')(({ theme }) => ({
+    marginTop: theme.spacing(3),
```
thomasheartman, line 13: Wouldn't it be better to handle spacing around the container in the parent component? Feels like placing the component on the page is the responsibility of the caller and not the component itself. Also makes it easier if you want to evenly space items.

But yeah, not super important 💁🏼

frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/useRecentlyUsedSegments.ts
```diff
@@ -0,0 +1,28 @@
...
+    const addItem = (newItem: number | number[]) => {
+        setItems((prevItems) => {
+            const itemsToAdd = Array.isArray(newItem) ? newItem : [newItem];
+
+            let updatedItems = [...prevItems];
+            itemsToAdd.forEach((id) => {
+                updatedItems = updatedItems.filter(
+                    (existingId) => existingId !== id,
+                );
+                updatedItems = [id, ...updatedItems];
```
thomasheartman, line 18: Out of curiosity, I wanted to see if there was a way to insert at the start of an array, and there is! 
You could use [`unshift`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift):

```suggestion
                updatedItems.unshift(id);
```

Not sure it has any other benefits here, though. (and if there was a perf difference, this list is always gonna be so small that it don't matter anyway 🤷🏼 )

thomasheartman (approved):

frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/useRecentlyUsedSegments.ts
sjaanus: It is generally good practice to not mutate array. Specially in react world. Due to how states etc work.

thomasheartman: Fair point. But in this case, you're creating a new array, anyway, so it doesn't really matter for states, right? It's not gonna be an issue here, but I have run into cases where spreading an object absolutely murdered the responsiveness because it had to chug through too many items. In that case, directly mutating the object instead was wildly more effective. That may or may not apply to arrays, but again: it was more out of curiosity 💁🏼

PR #9882, opened 2025-05-02
1-3687/input mode separation

thomasheartman:
Slight organizational update to get rid of the four different boolean states we had going on.

Primarily:
- creates `TopRowInput` component that gives you the right kind of input for the top row (all cases except for legal values, for which it returns null).
- Checks whether the input type is `legal values` before rendering the legal values input container.

Secondarily:
- Fixes the jank when adding new constraints
- extracts resolveLegalValues and improves performance (refer to the inline note)

### DIFF
--- a/frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
+++ b/frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
@@ -69,7 +69,7 @@ const resolveLegalValues = (
  *
  * For the case of `ProjectActionsFilterItem.tsx`: it already excludes legal
  * values and date operators. This leaves only free text and single value
- * text/numeric operators.
+ * text/numeric operators. Alternatively, consider rewriting this component to only handle those cases.
  */
 export const ResolveInput = ({
     input,
--- a/frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
+++ b/frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
@@ -64,6 +64,13 @@ const resolveLegalValues = (
     };
 };
 
+/**
+ * @deprecated; remove with `addEditStrategy` flag. Need an input? Prefer using specific input components.
+ *
+ * For the case of `ProjectActionsFilterItem.tsx`: it already excludes legal
+ * values and date operators. This leaves only free text and single value
+ * text/numeric operators.
+ */
 export const ResolveInput = ({
     input,
     contextDefinition,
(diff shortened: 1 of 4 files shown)


frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
thomasheartman, line 1: We should look at changing the inputs for actions too, but it's not urgent. In any case, we can at the very least update this function when we remove the flag, removing date and legal value constraint selectors.

frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/useConstraintInput/useConstraintInput.tsx
thomasheartman, line 1: This fixes the janky constraint addition. The reason we'd sometimes get an initial rendering of the legal value input is that this would run with the default set to legal values. It would change pretty immediately because of the `useEffect` hook later in the file, however. 

This method is used in three places:
- the editable constraint
- the deprecated constraint edit accordion
- the project actions filter item

Of those, the actions filter item doesn't support legal values, and for the editable constraints, it fixes the janky rendering. For the existing item, it just changes what it renders, but the size is about the same or smaller. 

Regardless, if the default here isn't the same as what the operator resolves to, it'll fix itself immediately. 

This isn't behind a flag.

I also considered passing in the default input type, but that didn't seem to have the desired effect.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
```diff
@@ -386,22 +400,19 @@ export const EditableConstraint: FC<Props> = ({
                     </StyledIconButton>
                 </HtmlTooltip>
             </TopRow>
-            {showInputField ? (
-                <InputContainer>
-                    <ResolveInput
-                        setValues={setValues}
+            {inputType === 'legal values' ? (
+                <LegalValuesContainer>
+                    <LegalValuesSelector
```
thomasheartman, line 405: We don't need resolveinput anymore, because the legal value input is the only input type we want.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/resolve-legal-values.ts
thomasheartman, line 1: This is extracted from `frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx`.

Some minor changes:
- uses empty arrays as default if the values are undefined
- has a complexity of O(n + m) instead of O(n * m) (where n = number of selected values and m = total number of available values). 

The perf thing  shouldn't matter for small lists, but I thought it might for cases where you have very many values (e.g. sandbox's clientId list is 636 entries and super laggy). It didn't seem to have much of an effect, but I'm happy to leave it in place regardless. 

Memoizing this also didn't seem to fix it, so I made a [linear task to investigate](https://linear.app/unleash/issue/1-3688/nice-to-have-performance-improvements-for-large-constraint-lists). I suspect it might be HTML rendering that's the bottle neck and that we might want to look into virtualizing this.

thomasheartman: I also thought about extracting this to a "common" file, but after the refactoring, this will only be relevant for the editable constraints anyway, so I'm leaving it next to the other files.

sjaanus (approved):

PR #9888, opened 2025-05-05
refactor: remove flagOverviewRedesign flag

Tymek:
## About the changes
When removing the flag I encountered following issues, also fixed in this PR:
- Change requests tests where out of date - method of switching flag on/off was relying on `data-testid=feature-flag-status`
- Change request summary had behavior not properly covered. Table with details on variants will now appear on adding variants to an existing strategy or editing variants - not on deleting a strategy or adding a fresh one, because it's what visual flags split is for

Without changes tests where not passing, that's why I have it all in one PR. See individual commits for a diff easier to parse.

Other things I'll fix in a next PR.


frontend/src/component/changeRequest/ChangeRequest.test.tsx
```diff
@@ -298,7 +291,8 @@ test('add flag change to pending change request', async () => {
 
     await verifyBannerForPendingChangeRequest();
 
-    await changeFlag('production');
+    const flag = screen.getByLabelText('production');
+    fireEvent.click(flag);
```
Tymek, line 295: we don't have buttons on the left to toggle if flag is enabled/disabled in an environment

thomasheartman: Sure is! We could even use `getByRole` instead (`getByRole('button', {name: 'production'}`) to follow [React Testing Library's priority guidelines](https://testing-library.com/docs/queries/about/#priority). I think that makes it clearer what we're doing too. (and maybe rename the variable to `environmentToggle` or something).

But overall, nah. That's fine. This is good.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.tsx
```diff
@@ -346,6 +327,8 @@ export const StrategyChange: FC<{
         environmentName,
     );
 
+    console.log('currentStrategy', currentStrategy);
+
```
thomasheartman, line 331: whoopsie 😄 
```suggestion
```

frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionView/ConstraintAccordionViewHeader/ConstraintAccordionViewHeader.tsx
```diff
@@ -13,9 +10,15 @@ interface IConstraintAccordionViewHeaderProps {
     onDelete?: () => void;
     onEdit?: () => void;
     onUse?: () => void;
+    /**
+     * @deprecated
+     */
     singleValue: boolean;
```
thomasheartman, line 16: Why don't we just delete these entirely? They don't do anything anymore, right? Both this and the next

frontend/src/component/common/StrategySeparator/LegacyStrategySeparator.tsx
```diff
@@ -29,6 +29,9 @@ const StyledCenteredContent = styled(StyledContent)(({ theme }) => ({
     padding: theme.spacing(0.75, 1.5),
 }));
 
+/**
+ * @deprecated remove with 'flagOverviewRedesign' flag. This pollutes a lot of places in the codebase 😞
+ */
```
thomasheartman, line 34: But you're removing the flag now? Why did you add the deprecation notice here? What does it mean?

frontend/src/component/playground/Playground/PlaygroundResultsTable/PlaygroundResultChip/LegacyPlaygroundResultChip.tsx
```diff
@@ -16,74 +8,17 @@ interface IResultChipProps {
     showIcon?: boolean;
 }
 
+/**
+ * @deprecated remove with 'flagOverviewRedesign' flag. This pollutes a lot of places in the codebase 😞
+ */
```
thomasheartman, line 13: Same thing as the previous comment: I don't understand what you mean here?

thomasheartman (approved):
Finally! 🔥  Very nice job with this 😄  My main question here is what the new deprecation notices telling people to remove it with the flagOverviewRedesign flag is about. This is the PR that removes that flag, so what does that mean? 

But also, I think we can just remove some of the other properties that aren't used anymore instead of deprecating them. You'll see in the inline comment.

frontend/src/component/common/StrategySeparator/LegacyStrategySeparator.tsx
Tymek: This is not breaking tests, and it's not a case of "just remove this file". I'll split this into a separate PR.

frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionView/ConstraintAccordionViewHeader/ConstraintAccordionViewHeader.tsx
Tymek: Yup, but this can be done in a separate PR. It tried to make this one as simple as possible, only changing things that didn't pass tests/tsc.

PR #9892, opened 2025-05-05
chore: flag cleanup simplifyDisableFeature

daveleek:
Cleans up the simplifyDisableFeature flag, and in doing so we're choosing the behavior that was put behind the flag.
This has been out for our hosted customers for a long time and we've not received any bug reports around it


daveleek:
I did some local testing before pushing the PR. The issue this feature flag was fixing was the usage of a specific index in the DB that excluded release plans related strategies, a little harder to repro in a test I think. It's been out a long while now, but I intend to do some testing after I merge it

PR #9894, opened 2025-05-05
WIP

gastonfournier:
Changed locally:
- **Migrations treated as CommonJS**: there's a problem with running migrations from within tests (the legacy tests that are not using the template db)
- The change above required undoing the migration of migrations to ESM. Alternatively, we could consider knex-migrate
- Added `import { jest } from '@jest/globals';` where needed

With these changes I got down to 28 test suites failures out of 345:
```shell
Test Suites: 28 failed, 317 passed, 345 total
Tests:       72 failed, 3 skipped, 2954 passed, 3029 total
Snapshots:   82 passed, 82 total
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,11 +113,16 @@
       "/src/migrations",
       "/src/test"
     ],
-    "extensionsToTreatAsEsm": [".ts", ".tsx"],
+    "extensionsToTreatAsEsm": [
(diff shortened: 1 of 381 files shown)


chriswk (approved):
Looks fine to me

PR #9895, opened 2025-05-06
feat: recently used segment chip

sjaanus:
I took the old segment chip as baseline/inspiration and then altered to my own needs.

image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegmentChip.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegmentChip.tsx
@@ -11,16 +11,11 @@ interface IRecentlyUsedSegmentChipProps {
 const StyledChip = styled(Chip)(({ theme }) => ({
     background: 'transparent',
     height: 'auto',
-    '.MuiChip-label': {
-        fontSize: theme.typography.body2.fontSize,
-        padding: theme.spacing(0.5, 1),
-    },
-    '&.MuiChip-outlined': {
-        border: `1px dashed ${theme.palette.divider}`,
-    },
+    fontSize: theme.typography.body2.fontSize,
+    padding: theme.spacing(0.5),
+    border: `1px dashed ${theme.palette.divider}`,
     '& .MuiChip-icon': {
         color: theme.palette.neutral.main,
-        marginLeft: theme.spacing(1),
     },
 }));
 
@@ -43,8 +38,8 @@ export const RecentlyUsedSegmentChip = ({
                 label={segment.name}
                 icon={<AddIcon />}
                 onClick={onAdd}
+                size='small'
                 variant='outlined'
-                clickable
             />
         </Tooltip>
     );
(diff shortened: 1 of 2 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegmentChip.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Chip, styled, Tooltip } from '@mui/material';
+
+interface IRecentlyUsedSegmentChipProps {
+    segment: ISegment;
+    setSegments: React.Dispatch<React.SetStateAction<ISegment[]>>;
+}
+
+const StyledChip = styled(Chip)(({ theme }) => ({
+    background: 'transparent',
+    height: 'auto',
```
thomasheartman, line 13: Do you need this for some reason? Isn't the height kinda right? If not, you can set `size='small` on the chip. That might also do what you need if there's too much padding.

frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegmentChip.tsx
```diff
@@ -0,0 +1,51 @@
...
+                return prev;
+            }
+            return [...prev, segment];
+        });
+    };
+
+    return (
+        <Tooltip title='Add segment' arrow>
+            <StyledChip
+                label={segment.name}
```
thomasheartman, line 43: For the above comment: if you want it smaller, you can add a size here

```suggestion
                label={segment.name}
                size='small'
```

frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegmentChip.tsx
```diff
@@ -0,0 +1,51 @@
...
+                icon={<AddIcon />}
+                onClick={onAdd}
+                variant='outlined'
+                clickable
```
thomasheartman, line 47: I don't think this is necessary (haven't seen it before). From the MUI docs:

> If true, the chip will appear clickable, and will raise when pressed, even if the onClick prop is not defined. If false, the chip will not appear clickable, even if onClick prop is defined. This can be used, for example, along with the component prop to indicate an anchor Chip is clickable. Note: this controls the UI and does not affect the onClick event.

I think that if you don't define it, it'll appear clickable as long as it has an onClick

frontend/src/component/feature/FeatureStrategy/FeatureStrategySegment/RecentlyUsedSegments/RecentlyUsedSegments.tsx
```diff
@@ -54,12 +54,10 @@ export const RecentlyUsedSegments = ({
             <StyledHeader>Recently used segments</StyledHeader>
             <StyledSegmentsContainer>
                 {segmentObjects.map((segment) => (
```
thomasheartman, line 56: Should we do some kind of filtering here to hide segments that already exist on that strategy? but ... If we only show recent segments when you have no segments, then that doesn't matter.

thomasheartman (approved):

PR #9900, opened 2025-05-06
feat: feature link migration

kwasniew:
## About the changes

Create table for feature links. ids will be ulids since it's a new convention introduced by release plans. The ids will be generated in the application. Also feature links will be deleted when features are deleted.

This migration is tested with e2e test suite that is coming up in the next PR

### DIFF
--- a/src/migrations/20250506074807-feature-link-table.js
+++ b/src/migrations/20250506074807-feature-link-table.js
@@ -1,0 +1,24 @@
+
+exports.up = (db, callback) => {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS feature_link(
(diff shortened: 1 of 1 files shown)


src/migrations/20250506074807-feature-link-table.js
```diff
@@ -0,0 +1,24 @@
+
+exports.up = (db, callback) => {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS feature_link(
+            id TEXT NOT NULL,
+            feature_name VARCHAR(255) NOT NULL,
```
kwasniew, line 7: has to mach features table name

Tymek (approved):

PR #9907, opened 2025-05-06
chore: clean up adminNavUI flag

daveleek:
Removes the usage of adminNavUI flag, thus turning on the new admin area/menu for everyone


gastonfournier (approved):
I think we should remove additional files which are no longer used

daveleek:
Thank you @gastonfournier - yes we should. Wasn't sure if it should be part of this PR or a followup

daveleek:
removed them @gastonfournier

PR #9909, opened 2025-05-06
chore(1-3639): constraint validation

thomasheartman:


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint.tsx
```diff
@@ -205,17 +206,19 @@ export const EditableConstraint: FC<Props> = ({
...
     setValues,
     setValuesWithRecord,
-    setError,
     removeValue,
-    error,
 }) => {
+    const { input } = useConstraintInput({
+        contextDefinition,
+        localConstraint,
+    });
```
thomasheartman, line 220: moving this here from the wrapper for now. I don't actually think we need this hook at all. I wanna spend the next PR looking at if we can get rid of that (and .. a lot of other stuff in those components).

frontend/src/component/layout/MainLayout/NavigationSidebar/NavigationSidebar.tsx
thomasheartman, line 1: these are unrelated to the overall change, but they were throwing errors in the console because they forwarded their non-html props and attributes.

sjaanus (approved):

PR #9913, opened 2025-05-07
fix: make setConstraints work in editable constraints list

sjaanus:
Was missing implementation for addConstraint through ref.

Added it together with test.

### DIFF
--- a/frontend/src/component/common/NewConstraintAccordion/ConstraintsList/EditableConstraintsList.tsx
+++ b/frontend/src/component/common/NewConstraintAccordion/ConstraintsList/EditableConstraintsList.tsx
@@ -1,10 +1,14 @@
 import type React from 'react';
+import { useImperativeHandle } from 'react';
 import { forwardRef } from 'react';
 import { styled } from '@mui/material';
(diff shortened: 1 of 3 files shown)


frontend/src/component/common/NewConstraintAccordion/ConstraintsList/EditableConstraintsList.tsx
```diff
@@ -29,6 +33,15 @@ export const EditableConstraintsList = forwardRef<
 >(({ constraints, setConstraints }, ref) => {
     const { context } = useUnleashContext();
 
+    useImperativeHandle(ref, () => ({
```
sjaanus, line 36: Implement addConstraint method for ref.

thomasheartman: Huh. Interesting. Don't think I've seen us do this kinda thing before. In fact, I didn't know you could 😅 But hey, if it works, it works

thomasheartman (approved):
Huh.

PR #9921, opened 2025-05-07
Refactor: get rid of editable constraint wrapper

thomasheartman:
This (admittedly pretty big) PR removes a component layer, moves all logic for updating constraint values into a single module, and dumbs down other components.

The main changes are:
- EditableConstraintWrapper is gone. All the logic in there has been moved into the new `useEditableConstraint` hook. Previously it was split between the wrapper, editableConstraint itself, the legalValues component.
- the `useEditableConstraint` hook accepts a constraint and a save function and returns an editable version of that constraint, the validator for input values, a function that accepts update commands, and, when relevant, existing and deleted legal values.
- All the logic for updating a constraint now exists in the `constraint-reducer` file. As a pure function, it'll be easy to unit test pretty thoroughly to make sure all commands work as they should (tests will come later)
- The legal values selector has been dumbed down consiberably as it no longer needs to create its own internal weak map. The internal representation of selected values is now a set, so any kind of lookup is now constant time, which should remove the need for the extra layer of abstraction. 

## Discussion points

I know the reducer pattern isn't one we use a *lot* in Unleash, but I found a couple examples of it in the front end and it's also quite similar to how we handle state updates to change request states. I'd be happy to find a different way to represent it if we can keep it in a single, testable interface.

Semi-relatedly: I've exposed the actions to submit for the updates at the moment, but we could map these to functions instead. It'd make invocations a little easier (you wouldn't need to specify the action yourself; only use the payload as a function arg if there is one), but we'd end up doing more mapping to create them. I'm not sure it's worth it, but I also don't mind if we do 💁🏼


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/LegalValuesSelector.tsx
thomasheartman, line 1: We can get rid of the extra values map we're creating now that the input is a set. Lookup is constant time that way. 

I've also removed the logic to figure out which values have been deleted in favor of moving that logic outside.

frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEditBody/ResolveInput/ResolveInput.tsx
```diff
@@ -84,23 +82,11 @@ export const ResolveInput = ({
...
-                        data={resolveLegalValues(
-                            constraintValues,
-                            contextDefinition.legalValues,
-                        )}
-                        constraintValues={constraintValues}
-                        values={localConstraint.values || []}
-                        setValuesWithRecord={setValuesWithRecord}
-                        setValues={setValues}
-                    />
-                ) : (
```
thomasheartman, line 103: This is not needed. The new component should only be used by the editable constraint for now, which uses it directly. Moreover, the project actions filter component doesn't have legal values, so it's not relevant there.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/useEditableConstraint/constraint-reducer.ts
thomasheartman, line 1: This is the main new attraction. In short, it moves all the logic for updating the constraint in here. Should make it very easy to test and verify that the different constraint updates work as correctly.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/useEditableConstraint/editable-constraint-type.ts
thomasheartman, line 1: Just a bunch of types to help differentiate between the different constraint types

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/useEditableConstraint/useEditableConstraint.tsx
thomasheartman, line 1: This hook is the spiritual successor to the editable constraint wrapper. It takes care of handling the logic that was originally handled outside the constraint.

In this version, it exposes the constraint, an update function, the validator, and legal values where necessary.

In particular, it wraps the update function to take care of saving state and updating context field information for you.

sjaanus (approved):

PR #9924, opened 2025-05-07
refactor: remove deprecated GET archive features endpoint

nunogois:
https://linear.app/unleash/issue/2-3366/remove-get-apiadminarchivefeatures-deprecated-in-4100

Removes GET `/api/admin/archive/features` which was deprecated in v4.10.
Also cleans up related code.

May include some slight scouting.

**P.S.** Should we merge this into main, or is there a `v7` branch we should be targeting instead?


gastonfournier (commented):
LGTM! I think we need to solve this conversation: https://unleash-internal.slack.com/archives/C048ELND3QD/p1746693473141949 before we merge

website/docs/reference/api/legacy/unleash/admin/archive.md
```diff
@@ -4,31 +4,6 @@ title: /api/admin/archive
 
 > In order to access the admin API endpoints you need to identify yourself. Unless you're using the `none` authentication method, you'll need to [create an ADMIN token](/how-to/how-to-create-api-tokens) and add an Authorization header using the token.
 
-### Fetch archived flags {#fetch-archived-toggles}
-
-`GET http://unleash.host.com/api/admin/archive/features`
```
gastonfournier, line 9: Do we need to replace this with another API or there's no other API?

gastonfournier (approved):
LGTM!

website/docs/reference/api/legacy/unleash/admin/archive.md
nunogois: An alternative would be: GET `/api/admin/search/features?archived=IS%3Atrue`. That said, my thinking is that since these are legacy docs, this particular detail may not be relevant here.

We've removed a deprecated endpoint, so this particular documentation change should reflect that IMO.

Looping in @melindafekete in case she has thoughts. Happy to follow up in a separate PR if needed.

PR #9930, opened 2025-05-08
feat: add domain to links

kwasniew:
## About the changes

After some research I found it's easiest to store domain on the write side of links. It will be reported to a DB using a dedicated library. We will store metrics in victoria metrics. I think it's better for this than plausible

### DIFF
--- a/src/migrations/20250508102832-feature-link-domain.js
+++ b/src/migrations/20250508102832-feature-link-domain.js
@@ -1,0 +1,15 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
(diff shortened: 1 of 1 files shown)


src/migrations/20250508102832-feature-link-domain.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+exports.up = function (db, callback) {
+    db.runSql(
+        `ALTER TABLE feature_link ADD COLUMN domain varchar(255);`,
```
kwasniew, line 5: domains should not be longer than 253 chars

PR #9932, opened 2025-05-08
chore: make operator checking more ergonomic and type-ful

thomasheartman:
This is a helper PR for a refactor I'm working on for the new constraint inputs. In the refactoring, it's useful to have individual subtypes for the various subgroups of operators and to be able to easily assert whether something is X operator or not.

The only change required in the code base is a single check for operators, which is now handled by using the new `isXOperator` functions instead.

Yes, the operator file in constants now includes functions, but it seemed useful to put the identification functions there instead of somewhere unrelated. The tests are primarily to ensure that the identifier function works, and I'd be happy to remove them if we think it's necessary. That said, they're pretty simple unit tests, so I think it's fine to leave them.

The main bulk of the change is: removing the explicit `: Operator[]` typing to the various sub-sets of operators and instead adding explicit types. Additionally, there's the new identifier functions.

### DIFF
--- a/frontend/src/component/common/ConstraintsList/ConstraintItemHeader/isCaseSensitive.test.ts
+++ b/frontend/src/component/common/ConstraintsList/ConstraintItemHeader/isCaseSensitive.test.ts
@@ -1,6 +1,7 @@
 import {
     allOperators,
     inOperators,
+    isInOperator,
     stringOperators,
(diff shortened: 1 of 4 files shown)


kwasniew (approved):
Looks fine. I'd manually test the usage of this change in the UI

PR #9938, opened 2025-05-08
refactor: remove deprecated get archive featured by project endpoint

nunogois:
https://linear.app/unleash/issue/2-3367/remove-get-apiadminarchivefeaturesprojectid-deprecated-in-4110

Removes GET `/api/admin/archive/features/{projectId}` which was deprecated in v4.11.
Also cleans up related code.

I leveraged our search features endpoint where needed, since that's what we're using now in our UI (to see archived features you filter by archived=true).

Builds on top of https://github.com/Unleash/unleash/pull/9924 — Since they're both related.


gastonfournier (approved):
LG! Clever usage of search! We should wait until we make a decision on versioning

melindafekete (approved):

PR #9939, opened 2025-05-09
chore: undeprecate custom strategies

daveleek:
Removes the deprecation notice on custom strategies, also removes these notices from docs


gastonfournier (approved):
LGTM! I've also added @sighphyre as FYI ;)

PR #9940, opened 2025-05-09
chore: prevent duplicate key errors in unknown flags

nunogois:
https://linear.app/unleash/issue/2-3561/fix-duplicate-key-errors-in-unknown-flags

This should prevent `duplicate_key` errors in unknown flags.

Follow-up to: https://github.com/Unleash/unleash/pull/9837


gastonfournier (approved):
Nice

PR #9944, opened 2025-05-09
feat: remove instances older than 1 day not 2 days

kwasniew:
## About the changes

This was showing instances from 2 days
Screenshot 2025-05-09 at 13 00 52

This is showing instances from 1 day
Screenshot 2025-05-09 at 12 58 24

This PR removes all instances older than 1 day. The first example above will start showing only last day instances as a result

### DIFF
--- a/src/lib/db/client-instance-store.ts
+++ b/src/lib/db/client-instance-store.ts
@@ -61,9 +61,9 @@ export default class ClientInstanceStore implements IClientInstanceStore {
             });
     }
 
-    async removeInstancesOlderThanTwoDays(): Promise<void> {
+    async removeOldInstances(): Promise<void> {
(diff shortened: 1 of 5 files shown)


src/lib/db/client-instance-store.ts
```diff
@@ -61,9 +61,9 @@ export default class ClientInstanceStore implements IClientInstanceStore {
             });
     }
 
-    async removeInstancesOlderThanTwoDays(): Promise<void> {
+    async removeOldInstances(): Promise<void> {
         const rows = await this.db(TABLE)
-            .whereRaw("last_seen < now() - interval '2 days'")
+            .whereRaw("last_seen < now() - interval '1 days'")
```
kwasniew, line 66: I think days and day work the same

PR #9948, opened 2025-05-09
refactor: remove deprecated post project role access endpoint

nunogois:
https://linear.app/unleash/issue/2-3363/remove-post-apiadminprojectsprojectidroleroleidaccess-deprecated-in

Removes POST `/api/admin/projects/{projectId}/role/{roleId}/access` which was deprecated in v5.5.
Also cleans up related code.

Needs follow up PRs.

### DIFF
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -843,6 +843,9 @@ export default class ProjectService {
         );
     }
 
+    /**
+     * @deprecated use `addAccess` instead
+     */
     async addRoleAccess(
         projectId: string,
         roleId: number,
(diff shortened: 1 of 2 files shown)


src/lib/features/project/project-service.ts
```diff
@@ -843,6 +843,9 @@ export default class ProjectService {
         );
     }
 
+    /**
+     * @deprecated use `addAccess` instead
+     */
     async addRoleAccess(
```
nunogois, line 849: Reason why this isn't deleted right away is because this would break Enterprise. Let's merge the Enterprise PR before we remove this.

gastonfournier (approved):
LG! Great work identifying these nuances! :medal_sports:

PR #9949, opened 2025-05-09
refactor: clean up dead code after removing deprecated project role access endpoint

nunogois:
https://linear.app/unleash/issue/2-3363/remove-post-apiadminprojectsprojectidroleroleidaccess-deprecated-in

Removes POST `/api/admin/projects/{projectId}/role/{roleId}/access` which was deprecated in v5.5.
Also cleans up related code.

Should only be merged after the Enterprise PR.


gastonfournier (approved):
LG!

PR #9953, opened 2025-05-12
chore: update styling for variant box

sjaanus:
Did both variations, release templates and normal.

image

### DIFF
--- a/frontend/src/component/feature/StrategyTypes/NewStrategyVariants.tsx
+++ b/frontend/src/component/feature/StrategyTypes/NewStrategyVariants.tsx
@@ -118,7 +118,7 @@ export const NewStrategyVariants: FC<{
 
     return (
         <>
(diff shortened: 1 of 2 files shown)


thomasheartman (approved):

frontend/src/component/feature/StrategyTypes/NewStrategyVariants.tsx
```diff
@@ -123,10 +118,10 @@ export const NewStrategyVariants: FC<{
 
     return (
         <>
-            <StyledVariantsHeader>
+            <Alert severity='info' sx={{ mb: 2 }} icon={false}>
```
thomasheartman, line 121: Technically, shouldn't it be up to the containing parent to decide on spacing between this and the next item? But it's not important. Up to you 🤷🏼

PR #9954, opened 2025-05-12
feat: export feature links

kwasniew:
## About the changes

Adding feature links to the export capability. Next PR will add import support.

The main change is to make read model support multiple features query instead of a single feature query (WHERE IN vs WHERE).

### DIFF
--- a/src/lib/features/export-import-toggles/export-import.e2e.test.ts
+++ b/src/lib/features/export-import-toggles/export-import.e2e.test.ts
@@ -464,7 +464,6 @@ test('should export custom context fields from strategies and variants', async (
         },
         strategy,
     );
-
     const variantStickinessContext = {
         name: 'variant-stickiness-context',
         legalValues: [
--- a/src/lib/features/export-import-toggles/export-import.e2e.test.ts
+++ b/src/lib/features/export-import-toggles/export-import.e2e.test.ts
@@ -112,6 +112,17 @@ const createVariants = async (feature: string, variants: IVariant[]) => {
     );
 };
 
+const addLink = async (
+    feature: string,
+    link: { url: string; title: string },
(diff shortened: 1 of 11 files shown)


src/lib/features/feature-links/feature-links-read-model-type.ts
```diff
@@ -2,9 +2,10 @@ export interface IFeatureLink {
     id: string;
     url: string;
     title: string | null;
+    feature: string;
 }
 
 export interface IFeatureLinksReadModel {
-    getLinks(feature: string): Promise<IFeatureLink[]>;
+    getLinks(...features: string[]): Promise<IFeatureLink[]>;
```
kwasniew, line 9: making this method generic so that it can serve a single feature view and feature export

src/lib/types/model.ts
```diff
@@ -124,7 +124,7 @@ export interface FeatureToggleView extends FeatureToggleWithEnvironment {
     children: string[];
     lifecycle: IFeatureLifecycleStage | undefined;
     collaborators?: { users: Collaborator[] };
-    links: IFeatureLink[];
+    links: Omit<IFeatureLink, 'feature'>[];
```
kwasniew, line 127: single feature view doesn't need to repeat feature name in the output structure

src/lib/features/export-import-toggles/export-import.e2e.test.ts
```diff
@@ -331,6 +353,15 @@ test('exports features', async () => {
                 ],
             },
         ],
+        links: [
```
kwasniew, line 356: this is a format modelled after the dependencies handling. Link of objects with feature name and corresponding links. This is easy to handle on the import side

PR #9957, opened 2025-05-12
chore: remove filterExistingFlagNames feature flag

nunogois:
https://linear.app/unleash/issue/2-3564/remove-filterexistingflagnames-feature-flag

We're removing the `filterExistingFlagNames` feature flag since we've decided we want this to be the default behavior.

We don't need to rush to merge it, just in case we need to disable this for any reason. However it should also be pretty easy to just revert if needed.

Changes in tests are a bit tricky since they assumed the previous behavior where we always registered metrics, even for non existing flag names. `cachedFeatureNames` is also memoized with a TTL of 10s, so the easiest way to overcome this was to override `cachedFeatureNames` to return what we expected. As long as they return the same flag names that we expect, we're able to register their metrics.

Let me know if you can think of a better approach.

### DIFF
--- a/src/lib/features/frontend-api/frontend-api.e2e.test.ts
+++ b/src/lib/features/frontend-api/frontend-api.e2e.test.ts
@@ -363,7 +363,7 @@ test('should store frontend api client metrics', async () => {
 
     // @ts-expect-error - cachedFeatureNames is a private property in ClientMetricsServiceV2
     app.services.clientMetricsServiceV2.cachedFeatureNames = jest
-        .fn()
+        .fn<() => Promise<string[]>>()
         .mockResolvedValue([featureName]);
 
     await app.request
--- a/src/lib/features/frontend-api/frontend-api.e2e.test.ts
+++ b/src/lib/features/frontend-api/frontend-api.e2e.test.ts
@@ -360,6 +360,12 @@ test('should store frontend api client metrics', async () => {
         projects: ['*'],
         environment: '*',
     });
+
+    // @ts-expect-error
+    app.services.clientMetricsServiceV2.cachedFeatureNames = jest
+        .fn()
+        .mockResolvedValue([featureName]);
+
     await app.request
         .get(`/api/admin/client-metrics/features/${featureName}`)
         .set('Authorization', adminToken.secret)
--- a/src/lib/features/frontend-api/frontend-api.e2e.test.ts
+++ b/src/lib/features/frontend-api/frontend-api.e2e.test.ts
@@ -361,7 +361,7 @@ test('should store frontend api client metrics', async () => {
         environment: '*',
     });
 
-    // @ts-expect-error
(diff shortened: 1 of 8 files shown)


src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -125,49 +125,36 @@ export default class ClientMetricsServiceV2 {
...
-                        MAX_UNKNOWN_FLAGS,
-                    );
-                }
+                unknownToggleNames = nonExistingNames.slice(
+                    0,
+                    MAX_UNKNOWN_FLAGS,
+                );
             } catch (e) {
                 this.logger.error(e);
             }
```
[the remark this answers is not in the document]
nunogois: Oops, I mentioned the wrong flag. The one that we removed was: `filterExistingFlagNames`. We're keeping `reportUnknownFlags`.

src/lib/features/frontend-api/frontend-api.e2e.test.ts
```diff
@@ -356,6 +356,12 @@ test('should store frontend api client metrics', async () => {
         projects: ['*'],
         environment: '*',
     });
+
+    // @ts-expect-error
```
gastonfournier, line 360: What's the expected error? It's usually a good idea to explain why. Also, ts-expect-error usually hides other potential errors. 
If it's a type issue with jest.fn, you can add type params to it: `jest.fn<string[]>` IIRC

src/test/e2e/api/client/metrics.access.e2e.test.ts
```diff
@@ -33,6 +33,12 @@ test('should enrich metrics with environment from api-token', async () => {
         project: '*',
     });
 
+    const featureName = Object.keys(metricsExample.bucket.toggles)[0];
+    // @ts-expect-error
```
gastonfournier, line 37: Same comment about ts-expect-error

gastonfournier (approved):
I think it looks good! Just nitpicking on the ts-expect-error that I don't like much

src/lib/features/frontend-api/frontend-api.e2e.test.ts
nunogois: It's because `cachedFeatureNames` is not a publicly accessible property of `clientMetricsServiceV2`. I don't like adapting our code to make our tests happy (making this public instead of private when it doesn't make sense otherwise) but maybe that would be the easiest way of preventing this `// @ts-expect-error`. Thoughts?

src/test/e2e/api/client/metrics.access.e2e.test.ts
nunogois: Addressed in https://github.com/Unleash/unleash/pull/9957/commits/88d71ec1c4b4490acd23d647dfe6bd1d751bb51b

gastonfournier (approved):
LG!

PR #9958, opened 2025-05-12
feat: Import feature links

kwasniew:
## About the changes

Can import feature links by deleting previous ones if they existed and creating new ones.


src/lib/features/export-import-toggles/export-import-service.ts
```diff
@@ -355,6 +364,27 @@ export default class ExportImportService
         await this.importDependencies(dto, user, auditUser);
     }
 
+    private async importLinks(dto: ImportTogglesSchema, auditUser: IAuditUser) {
```
kwasniew, line 367: similar algorithm to feature tags. First delete previous links then create new ones using existing method for creation to get all event logs recorded

PR #9960, opened 2025-05-12
feat: add input box for gradual rollout slider

sjaanus:
Add input box for gradual rollout slider.
Was not able to get it aligned nicely in the middle of horizontal slider, so its just inline with the row for now. Yes, just adding random padding did work, but that felt too artificial.

This is also behind feature flag. We can improve small CSS stylings in future iterations.

Added % also for the input, as it felt more correct.

image

### DIFF
--- a/frontend/src/component/feature/StrategyTypes/RolloutSlider/RolloutSlider.tsx
+++ b/frontend/src/component/feature/StrategyTypes/RolloutSlider/RolloutSlider.tsx
@@ -1,7 +1,15 @@
-import { makeStyles, withStyles } from 'tss-react/mui';
-import { Slider, Typography, Box, styled } from '@mui/material';
+import { withStyles } from 'tss-react/mui';
+import {
+    Slider,
+    Typography,
+    Box,
+    styled,
+    TextField,
+    InputAdornment,
+} from '@mui/material';
 import { ROLLOUT_SLIDER_ID } from 'utils/testIds';
 import { HelpIcon } from 'component/common/HelpIcon/HelpIcon';
+import { useState, useEffect } from 'react';
 
 const StyledSlider = withStyles(Slider, (theme) => ({
     root: {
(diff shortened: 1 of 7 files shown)


frontend/src/component/feature/StrategyTypes/RolloutSlider/RolloutSlider.tsx
```diff
@@ -104,18 +157,18 @@ const RolloutSlider = ({
...
                             <Typography variant='body2'>
-                                The rollout percentage determines the proportion
-                                of users exposed to a feature. It's based on the
-                                MurmurHash of a user's unique identifier,
-                                normalized to a number between 1 and 100. If the
-                                normalized hash is less than or equal to the
-                                rollout percentage, the user sees the feature.
-                                This ensures a consistent, random distribution
-                                of the feature among users.
+                                T2he rollout percentage determines the
```
thomasheartman, line 160: This looks like a typo! Probably  has knock-on effects for the rest of the formatting too 😄 

```suggestion
                                The rollout percentage determines the
```

frontend/src/component/feature/StrategyTypes/RolloutSlider/RolloutSlider.tsx
```diff
@@ -104,18 +157,18 @@ const RolloutSlider = ({
...
+                                identifier, normalized to a number between 1 and
+                                100. If the normalized hash is less than or
+                                equal to the rollout percentage, the user sees
+                                the feature. This ensures a consistent, random
+                                distribution of the feature among users.
                             </Typography>
 
                             <StyledSubheader variant='h3'>
-                                Stickiness
+                                Stickiness1
```
thomasheartman, line 171: Uh, what's with the numbers this time around? 😅 

```suggestion
                                Stickiness
```

frontend/src/component/feature/StrategyTypes/RolloutSlider/RolloutSlider.tsx
```diff
@@ -142,20 +195,47 @@ const RolloutSlider = ({
...
+                    />
+                </SliderContent>
+                <StyledInputBox>
+                    <StyledTextField
+                        size='small'
+                        type='number'
+                        value={inputValue}
+                        onChange={handleInputChange}
+                        onBlur={handleInputBlur}
+                        onKeyDown={handleKeyDown}
```
thomasheartman, line 221: Why does this one need onBlur and onKeyDown? Shouldn't onChange be enough?

frontend/src/component/feature/StrategyTypes/RolloutSlider/RolloutSlider.tsx
```diff
@@ -142,20 +195,47 @@ const RolloutSlider = ({
...
+                        data-testid={ROLLOUT_SLIDER_ID}
+                        marks={marks}
+                        onChange={onChange}
+                        valueLabelDisplay='on'
+                        disabled={disabled}
+                    />
+                </SliderContent>
+                <StyledInputBox>
+                    <StyledTextField
+                        size='small'
```
thomasheartman, line 216: This input definitely needs a label. We can reuse the same one we use for the slider, though.

But also, unless that ID is used by something else, I'd suggest replacing it with `react.useId` to guarantee that it's unique even if there's multiple rollout sliders on a page. (don't know if that can happen atm? change requests? but best to be sure anyway.)

```suggestion
                        size='small'
                        aria-labelledby='discrete-slider-always'
```

thomasheartman (approved):

frontend/src/component/feature/StrategyTypes/RolloutSlider/RolloutSlider.tsx
sjaanus: If we used only onChange on the input field, the slider would update immediately with each keystroke. This would create a poor user experience:
As a user types "75", the slider would first jump to "7" and then to "75"

thomasheartman: Right. I see the logic there. I don't understand why the onKeyDown handler is there, though. Shouldn't enter submit the form (as is standard with input fields)? Not a big deal, but I don't think we need it 🤷🏼

thomasheartman (approved):

PR #9962, opened 2025-05-12
chore: remove proxy from integrations

gastonfournier:
Removing proxy from integrations as it's been deprecated while ago and currently unmaintained


nunogois (approved):
LGTM

PR #9964, opened 2025-05-12
refactor: simplify feature toggle service deps

kwasniew:
## About the changes

Grouping feature toggle service deps into stores, config and services/read models. This simulated named parameters instead of having a long list of positional params.


Tymek (approved):
👍🏻

PR #9967, opened 2025-05-12
refactor: lax newUserInviteLink user arg

nunogois:
Part of: https://linear.app/unleash/issue/2-3560/payg-flow-sends-2-email-one-from-bluebird-and-one-from-unleash-we-want

This laxes our `user` argument in `newUserInviteLink` to just pick the id, since that's the only property we're interested in. This allows us to send `{ id }` when we don't have the full `IUserWithRootRole` available while keeping compatibility intact.

### DIFF
--- a/src/lib/services/user-service.ts
+++ b/src/lib/services/user-service.ts
@@ -295,7 +295,7 @@ class UserService {
     }
 
     async newUserInviteLink(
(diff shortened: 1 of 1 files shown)


sastromskis (approved):
LGTM

gastonfournier (approved):
LGTM!

PR #9973, opened 2025-05-13
feat: make all feature toggle service write methods transactional

kwasniew:
## About the changes

* making all feature toggle service write methods transactional
* switching from previous tx mechanism to the recent one
* removing unused method on feature toggle controller
* fixing tests that didn't get db.rawDatabase passed in the app constructor

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-controller.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-controller.ts
@@ -48,10 +48,7 @@ import type {
 } from '../../services';
 import { querySchema } from '../../schema/feature-schema';
 import type { BatchStaleSchema } from '../../openapi/spec/batch-stale-schema';
-import type {
-    TransactionCreator,
-    UnleashTransaction,
-} from '../../db/transaction';
+import type { WithTransactional } from '../../db/transaction';
 import { BadDataError } from '../../error';
 import { anonymise } from '../../util';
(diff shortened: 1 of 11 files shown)


src/lib/features/feature-toggle/feature-toggle-controller.ts
```diff
@@ -680,14 +675,18 @@ export default class ProjectFeaturesController extends Controller {
...
-            {
-                ...req.body,
-                description: req.body.description || undefined,
-            },
-            req.audit,
-        );
+        const created =
+            await this.transactionalFeatureToggleService.transactional(
+                (service) =>
+                    service.createFeatureToggle(
```
kwasniew, line 681: this is the most important change. feature creation should be transactional

src/lib/features/feature-toggle/feature-toggle-controller.ts
```diff
@@ -949,8 +953,8 @@ export default class ProjectFeaturesController extends Controller {
             return;
         }
 
-        await this.startTransaction(async (tx) =>
-            this.transactionalFeatureToggleService(tx).bulkUpdateEnabled(
+        await this.transactionalFeatureToggleService.transactional((service) =>
```
kwasniew, line 956: this is an example of a transition from out previous transaction handling mechanism to the current one

Tymek (approved):

PR #9974, opened 2025-05-13
chore: accept 429 responses

gastonfournier:
When being throttled assume the link is ok, eventually someday it will be checked

Already tested it works on this branch: https://github.com/Unleash/unleash/actions/runs/14993127371


nunogois (approved):
LG!

PR #9977, opened 2025-05-13
chore: enable strict schema validation

sjaanus:


src/lib/types/experimental.ts
```diff
@@ -96,7 +96,7 @@ const flags: IFlags = {
     },
     strictSchemaValidation: parseEnvVarBoolean(
         process.env.UNLEASH_STRICT_SCHEMA_VALIDTION,
-        false,
+        true,
```
gastonfournier, line 99: UNLEASH_STRICT_SCHEMA_VALIDTION should be UNLEASH_STRICT_SCHEMA_VALIDATION,
 and I would keep it false but enabled for tests and dev (which we might already have)

PR #9981, opened 2025-05-13
feat: track last seen clients using bulk update 

Tymek:
Let's not update `lastSeen` in the db on each client call


src/lib/features/metrics/instance/metrics.test.ts
```diff
@@ -108,29 +108,6 @@ test('should accept client metrics with yes/no', () => {
         .expect(202);
 });
 
-test('should accept client metrics with yes/no with metricsV2', async () => {
```
Tymek, line 111: same as test above it

kwasniew (approved):

PR #10000, opened 2025-05-15
fix: now feature is not pulled from server if not defined

sjaanus:
All our servers have had this for a while on every feature open.
image

This PR fixes it by making request only conditionally.

The featureFetcher function also specifically handles 404 responses by returning `{ status: 404 }`, which makes the behavior consistent whether the API returns a 404 or the condition prevents the API call from happening at all.

This approach avoids unnecessary API calls when parameters are missing and provides a consistent interface for handling "not found" scenarios in the component that uses this hook.


kwasniew (approved):

Tymek:
10000!

PR #10003, opened 2025-05-15
Vitest

chriswk:
Points of discussion:
* Do we want to migrate away from Jest? 
* Fast check has a vitest module, which we should probably rewrite our fast-check tests to use.

Vitest Pros:
* Automated failing test comments on github PRs
* A nice local UI with incremental testing when changing files (`yarn test:ui`)
* Also nicely supported in all major IDEs, click to run test works (so we won't miss what we had with jest).
* Works well with ESM

Vitest Cons:
* The ESBuild transformer vitest uses takes a little longer to transform than our current SWC/jest setup, however, it is possible to setup SWC as the transformer for vitest as well (though it only does one transform, so we're paying ~7-10 seconds instead of ~ 2-3 seconds in transform phase).
* Exposes how slow our tests are (tongue in cheek here)


package.json
```diff
@@ -53,82 +53,22 @@
...
     "test:docker": "./scripts/docker-postgres.sh",
-    "test:report": "NODE_OPTIONS=--experimental-vm-modules NODE_ENV=test PORT=4243 node node_modules/.bin/jest --reporters=\"default\" --reporters=\"jest-junit\"",
+    "test:report": "NODE_ENV=test PORT=4243 vitest --reporter=junit",
     "test:docker:cleanup": "docker rm -f unleash-postgres",
-    "test:watch": "yarn test --watch",
-    "test:coverage": "NODE_ENV=test PORT=4243 node --experimental-vm-modules node_modules/.bin/jest --coverage --testLocationInResults --outputFile=\"coverage/report.json\" --forceExit",
-    "test:coverage:jest": "NODE_ENV=test PORT=4243 node --experimental-vm-modules node_modules/.bin/jest --silent --ci --json --coverage --testLocationInResults --outputFile=\"report.json\" --forceExit",
-    "test:updateSnapshot": "NODE_ENV=test PORT=4243 node --experimental-vm-modules node_modules/.bin/jest --updateSnapshot",
+    "test:watch": "vitest",
+    "test:coverage": "NODE_ENV=test PORT=4243 vitest run --coverage --outputFile=\"coverage/report.json\"",
```
chriswk, line 62: NODE_ENV is set by default to test, so we probably don't need to set that.

kwasniew (approved):
I see 3 categories of changes:
* jest -> vi
* Rejection syntax (toThrow -> errorWithMessage)
* Forced await usage in async tests
anything else?

I would expect test speedup only if we switched to a tool without import mocking (e.g. native test runner or mocha). Tools that do import mocking usually add significant overhead since they bypass native import caching and read file system over and over again.

gastonfournier (approved):
I vote for yes, for consistency and also jest mocking of ESM modules is not straightforward and actually experimental. Also, the migration cost is not huge and we can afford choosing a different tool later if we have to

nunogois (approved):
✅

PR #10004, opened 2025-05-15
fix: hide project archive in OSS

nunogois:
https://linear.app/unleash/issue/2-3569/fix-hide-project-archive-in-oss

Hides "project archive" in OSS.

I believe this is a bug. OSS only has one project and the project archive was acting unexpectedly anyways since it was showing the same default project as being archived. This is because in OSS we use the OSS project-controller, not the Enterprise version (override) of it.

### DIFF
--- a/frontend/src/component/menu/routes.ts
+++ b/frontend/src/component/menu/routes.ts
@@ -125,6 +125,7 @@ export const routes: IRoute[] = [
         component: ArchiveProjectList,
         type: 'protected',
         menu: {},
+        enterprise: true,
     },
 
     // Flags overview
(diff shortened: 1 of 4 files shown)


frontend/src/component/menu/routes.ts
```diff
@@ -125,6 +125,7 @@ export const routes: IRoute[] = [
         component: ArchiveProjectList,
         type: 'protected',
         menu: {},
+        enterprise: true,
```
nunogois, line 128: Even though this is called `enterprise` I believe (and validated locally that) it does what we want: Only mount this route on Pro and Enterprise.

kwasniew (approved):

PR #10009, opened 2025-05-16
docs: explain health API properly

### LINKED ISSUE #9965: /health api always return status code 200
### Describe the bug

Hello everyone,

When invoking the health API at "/health", a response with status code 200 OK is returned: { "health": "GOOD" } even though the PostgreSQL instance is offline.

According to the documentation, in this exact scenario the API should return: 500 { "health": "BAD" }.

The issue can be reproduced by starting with both the web application and the database instance online, then forcibly shutting down the database server. The API response remains unchanged.

This has been replicated both with containerized applications and with a local Windows 11 installation.
Unleash version: 6.10.0
PostgreSQL version: 17.4-1

1. start both unleash and PostgresSQL;
2. Invoke /health API; it returns 200 correctly
3. turn off PostgreSQL instance;
4. Invoke /health API, it still returns 200 

### Expected behavior

1. start both unleash and PostgresSQL;
2. Invoke /health API; it returns 200 correctly
3. turn off PostgreSQL instance;
4. Invoke /health API, it returns 500 

### Logs, error output, etc.

```bash

```

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

6.10.0

### Subscription type

Open source

### Hosting type

Self-hosted

_No response_

gastonfournier:
## About the changes
Our health implementation always returns GOOD if the server is up: https://github.com/Unleash/unleash/blob/beb29f5b5bbff620097fa822e2dff6b68a11adab/src/lib/routes/health-check.ts#L46-L51

Currently our documentation of the endpoint is misleading saying that it will return BAD if unable to connect to PostgreSQL

Closes #9965

### DIFF
--- a/src/lib/openapi/spec/health-check-schema.ts
+++ b/src/lib/openapi/spec/health-check-schema.ts
@@ -10,7 +10,7 @@ export const healthCheckSchema = {
     properties: {
         health: {
             description:
-                'The state this Unleash instance is in. GOOD if everything is ok, BAD if the instance should be restarted',
+                'The state this Unleash instance is in. GOOD if the server is up and running. It never returns BAD, if the server is unhealthy you will get an unsuccessful http response.',
             type: 'string',
             enum: ['GOOD', 'BAD'],
             example: 'GOOD',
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Nice

PR #10012, opened 2025-05-16
fix: now feature component is not loaded before we have feature infor…

sjaanus:
We were seeing strange errors when the feature component was rendered before the feature data was returned from the backend. Now, we ensure the component is not rendered until the feature is available.

### DIFF
--- a/frontend/src/hooks/api/getters/useFeature/useFeature.ts
+++ b/frontend/src/hooks/api/getters/useFeature/useFeature.ts
@@ -1,10 +1,9 @@
-import type { SWRConfiguration } from 'swr';
+import useSWR, { type SWRConfiguration } from 'swr';
 import { useCallback } from 'react';
-import { emptyFeature } from './emptyFeature.ts';
-import handleErrorResponses from '../httpErrorResponseHandler.ts';
+import { emptyFeature } from './emptyFeature.js';
+import handleErrorResponses from '../httpErrorResponseHandler.js';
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/api/getters/useFeature/useFeature.ts
```diff
@@ -26,9 +25,7 @@ export const useFeature = (
 ): IUseFeatureOutput => {
     const path = formatFeatureApiPath(projectId, featureId);
 
-    const { data, error, mutate } = useConditionalSWR<IFeatureResponse>(
-        Boolean(featureId && projectId),
-        { status: 404 },
+    const { data, error, mutate } = useSWR<IFeatureResponse>(
```
sjaanus, line 28: Rollback of conditional swr.

frontend/src/hooks/api/getters/useFeature/useFeature.ts
```diff
@@ -1,10 +1,9 @@
-import type { SWRConfiguration } from 'swr';
+import useSWR, { type SWRConfiguration } from 'swr';
 import { useCallback } from 'react';
-import { emptyFeature } from './emptyFeature.ts';
-import handleErrorResponses from '../httpErrorResponseHandler.ts';
+import { emptyFeature } from './emptyFeature.js';
```
kwasniew, line 3: js extension?

PR #10025, opened 2025-05-23
chore: remove console.log and reuse package scripts

gastonfournier:
Make `start:dev` re-use `dev:backend`, the only difference is start:dev runs a clean before running

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "dev:frontend:cloud": "UNLEASH_BASE_PATH=/demo/ yarn run dev:frontend",
     "dev": "concurrently \"yarn:dev:backend\" \"yarn:dev:frontend\"",
     "prepare:backend": "concurrently \"yarn:copy-templates\" \"yarn:build:backend\"",
-    "start:dev": "yarn run clean && TZ=UTC NODE_ENV=development tsc-watch --onSuccess \"node dist/server-dev.js\"",
+    "start:dev": "yarn run clean && yarn dev:backend",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
     "dev:frontend:cloud": "UNLEASH_BASE_PATH=/demo/ yarn run dev:frontend",
     "dev": "concurrently \"yarn:dev:backend\" \"yarn:dev:frontend\"",
     "prepare:backend": "concurrently \"yarn:copy-templates\" \"yarn:build:backend\"",
-    "start:dev": "yarn run clean && TZ=UTC NODE_ENV=development tsc-watch --onSuccess \"node dist/server-dev.js\"",
+    "start:dev": "yarn run clean && yarn dev:backend",
```
nunogois, line 49: `dev:backend` is probably the most consistent with the previous behavior, but I wonder if:
 - We should `dev` instead. Is there any value in running just the backend?
 - We should remove this script. It seems like we already mention `yarn dev` in our docs, so this can probably be cleaned up.

nunogois: That's interesting, because I always run `yarn dev` with both if I'm running OSS. Running the frontend should be quick anyways, and I think I've run into weird unexpected behaviors in the past with my backend serving an outdated UI.

gastonfournier: I'd be ok removing it, so in that way of thinking it doesn't really make sense if we run UI or not. For now I'll keep it because others might be using it but it shouldn't be a big problem if we remove it

PR #10031, opened 2025-05-27
chore: prefer searchEvents over deprecated methods

nunogois:
https://linear.app/unleash/issue/2-3577/prefer-the-new-searchevents-method-over-deprecated-methods

We should favor our new `searchEvents` method over the other deprecated methods. This PR removes these deprecated methods, replaces them with the new `searchEvents` and `searchEventsCount` methods, and marks the old endpoints as deprecated.

Follow-up to: https://github.com/Unleash/unleash/pull/10030


gastonfournier (approved):
Looks good!

PR #10032, opened 2025-05-27
fix: make revision id not be so reactive

sjaanus:


src/lib/features/events/event-store.ts
```diff
@@ -196,7 +198,14 @@ class EventStore implements IEventStore {
             .max('id')
             .where((builder) =>
                 builder
-                    .whereNotNull('feature_name')
+                    .andWhere((inner) =>
```
sjaanus, line 201: This the the main change, the core.

src/lib/features/events/event-store.ts
```diff
@@ -196,7 +198,14 @@ class EventStore implements IEventStore {
...
+                        inner
+                            .whereNotNull('feature_name')
+                            .whereNotIn('type', [
+                                FEATURE_CREATED,
+                                FEATURE_TAGGED,
+                            ]),
+                    )
```
gastonfournier, line 208: Doesn't this work the same way? 
```
        const row = await this.db(TABLE)
            .max('id')
            .where((builder) =>
                builder
                    .whereNotNull('feature_name')
                    .orWhereIn('type', [
                        SEGMENT_UPDATED,
                        FEATURE_IMPORT,
                        FEATURES_IMPORTED,
                    ]),
            )
            .andHavingNotIn('type', [FEATURE_CREATED, FEATURE_TAGGED])
            .andWhere('id', '>=', largerThan)
            .first();
```

chriswk (approved):
LGTM

src/lib/features/events/event-store.ts
gastonfournier: Also, note we use a similar query for delta calculation (method `getRevisionRange` below), maybe both would have to share some code?

gastonfournier (commented):
LGTM just minor details I trust you on this one

src/lib/features/events/event-store.ts
sjaanus: Yes, in theory that would do the same, just wanted to keep all filtering in same location. Hmm, but the code is a more readable.

sjaanus:
> LGTM just minor details I trust you on this one

Thanks! I appreciate the trust 😊 Just to make sure we are all aligned and nothing slips through. Always good to have a second pair of eyes!

src/lib/features/events/event-store.ts
sjaanus: Updated getRevisionRange also.

sjaanus: Actually I reverted this, as where not does not work with lists and this broke the code.

PR #10035, opened 2025-05-27
Chore(1 3755)/split insights in three

thomasheartman:
Creates sections for the insights dashboard and moves charts around into the same order as the sketches and into the right sections. There's no charts for the top section (lifecycle currently) yet, and the sections also don't have their own filters. 

To make this re-ordering easier, I've also moved the previous insights chart into a legacy file and set up a proxy component that handles switching based on the flag.

image

Next step is separating the filters.


frontend/src/component/insights/LegacyInsightsCharts.tsx
thomasheartman, line 1: This is the insights chart as it was before https://github.com/Unleash/unleash/pull/10034, i.e. with the median time to prod chart.

PR #10037, opened 2025-05-27
feat: update sidebar navigation

Tymek:
## About the changes

Closes #


Tymek:
> What will happen to recent projects and flags?

this is now available easily from dashboard and command bar (quick suggestions) that we didn't have before

PR #10038, opened 2025-05-28
feat: clean up old OAS and ability to add badges to descriptions

gastonfournier:
## About the changes
This removes some old and unused files from an initial test of setting up openapi that's currently not working: 
image

Also, adds the ability of marking endpoints as (note, the endpoints are made up just for illustration purposes):
- Enterprise only 
image

- Beta: 
image

- Both: 
image

### DIFF
--- a/src/lib/services/openapi-service.ts
+++ b/src/lib/services/openapi-service.ts
@@ -46,7 +46,7 @@ export class OpenApiService {
         const { beta, enterpriseOnly, ...rest } = op;
 
         const betaBadge = beta
-            ? `![Beta](${this.docsStaticsPath()}/Beta.svg)
+            ? `![Beta](${this.docsStaticsPath()}/Beta.svg) This is a beta endpoint and it may change or be removed in the future. 
             
             `
             : '';


src/lib/openapi/util/api-operation.ts
```diff
@@ -14,4 +14,6 @@ export interface ApiOperation<Tag = OpenApiTag | DeprecatedOpenAPITag>
     extends Omit<OpenAPIV3.OperationObject, 'tags'> {
     operationId: string;
     tags: [Tag];
+    beta?: boolean;
+    enterpriseOnly?: boolean;
```
gastonfournier, line 18: This way we can also specify an operation to be beta or enterpriseOnly

src/lib/services/openapi-service.ts
```diff
@@ -38,12 +43,40 @@ export class OpenApiService {
...
-        return this.api.validPath(op);
+        const { beta, enterpriseOnly, ...rest } = op;
+
+        const betaBadge = beta
+            ? `![Beta](${this.docsStaticsPath()}/Beta.svg)
+            
+            `
+            : '';
+        const enterpriseBadge = enterpriseOnly
+            ? `![Enterprise](${this.docsStaticsPath()}/Enterprise.svg) **Enterprise feature**
```
thomasheartman, line 54: Small nit, but I think this is a clearer alt text

```suggestion
            ? `![Unleash Enterprise](${this.docsStaticsPath()}/Enterprise.svg) **Enterprise feature**
```

thomasheartman (commented):
Sounds nice, but a couple questions:

On beta:
What's the process for marking something as beta? When is it not beta? If it's in development, but not being rolled out yet, is it in beta? How does the "beta" tag interact with "unstable"? What does it mean for something to be unstable AND beta? What about unstable but not beta? Should this "beta" thing replace the "unstable" tag? That actually sounds like a good idea to me, but in that case, does "beta" accurately capture the intention? (maybe it does, not all beta features make it to stable, after all).

On enterprise-only:
Am I correct in assuming that this will apply to all endpoints in enterprise and none of the endpoints in oss? If so, is there a way that we can apply this automatically instead of having to do it manually? (Maybe we can override the implementation in enterprise?)

gastonfournier:
> Sounds nice, but a couple questions:

Thanks!

> On beta: What's the process for marking something as beta? When is it not beta? If it's in development, but not being rolled out yet, is it in beta? How does the "beta" tag interact with "unstable"? What does it mean for something to be unstable AND beta? What about unstable but not beta? Should this "beta" thing replace the "unstable" tag? That actually sounds like a good idea to me, but in that case, does "beta" accurately capture the intention? (maybe it does, not all beta features make it to stable, after all).

Not really, relevant for the PR, this is adding the capability, not the process. We currently do use the "unstable" tag, but I noticed we rush to move it into a more stable state. I was thinking it'd be awesome to add a version property to the operation stating when it was added, and then if it was added in the current or previous minor version, automate the `Beta` badge... but that's probably as a follow up. Now it's just up to each one of us.

I like the idea of replacing Unstable with Beta, but following expand and contract, this would be the expand phase and removing Unstable would be the contract phase. We still have time to learn and decide

> On enterprise-only: Am I correct in assuming that this will apply to all endpoints in enterprise and none of the endpoints in oss? If so, is there a way that we can apply this automatically instead of having to do it manually? (Maybe we can override the implementation in enterprise?)

Yes, I'd like to have this out first and then figure out if we can automate it somehow or override enterprise impl.

PR #10039, opened 2025-05-28
Feat(1-3760)/filters per analytics section

thomasheartman:
Splits the insights page into separate sections with their own separate filters. Each filter is sticky to the top of the page, similar to how the previous filters worked.

In doing this, I have also moved around a lot of code. Refer to the inline comments for more specific examples, but on a high level, this PR:
- Moves the flag check from InsightsCharts into Insights itself. Because the old Insights had filters and state, and the new one does not, it made sense to fork off higher up in the tree. Because the new version doesn't have state, I have also simplified it and removed an intermediate component (InsightsCharts) that doesn't feel necessary anymore.
- Because InsightsCharts isn't used anymore, I've moved the LegacyInsightsCharts file back into InsightsCharts. However, I'm happy to move it back if we think that's useful.
- Instead of all charts living in InsightsCharts, I've split each section into its own file in the new sections directory. Because the sections have separate filters, they don't share any state anymore, so there's nothing they share.
- I'm reusing the existing hook and endpoint. As it stands, the performance insights use **most** of the data from that payload (but not all of it). The user insights use some of it. Flag lifecycle insights doesn't use anything, but I've wired it up to make the filters work. I expect that we'll iterate on the API endpoints etc later.

image


frontend/src/component/filter/Filters/Filters.tsx
```diff
@@ -42,7 +42,7 @@ type ITextFilterItem = IBaseFilterItem & {
     pluralOperators: [string, ...string[]];
 };
 
-type IDateFilterItem = IBaseFilterItem & {
+export type IDateFilterItem = IBaseFilterItem & {
```
thomasheartman, line 45: I've exported this for a type assertion that TS suddenly started complaining about in the InsightsFilter component.

frontend/src/component/insights/InsightsCharts.tsx
thomasheartman, line 1: I have reverted this to the legacy version of the file (and deleted the legacy file). Additionally, I've deleted some scroll handling code that wasn't doing anything (only setting variables that weren't used anywhere) but I believe was causing console.errors.

frontend/src/component/insights/InsightsCharts.styles.ts
thomasheartman, line 1: This is extracting styled components from the InsightsCharts file because they're now in use both in Insights.tsx and in the section files.

frontend/src/component/insights/Insights.tsx
thomasheartman, line 1: I've extracted the old version into LegacyInsights. This file contains a shared entry point that checks the flag state and delegates further. The new version doesn't have filters and can thus be simplified quite a bit.

I have also removed the in-between component of InsightCharts: All the sections need to have their own filter state, so there's nothing we can share between them anymore. As such, I've extracted each section to its own file and simplified this wrapper.

frontend/src/component/insights/InsightsFilters.tsx
thomasheartman, line 1: This file now supports setting a prefix for the local storage to use. This allows us to use the same component for multiple filters (e.g. performance, users). 

Note: we might revert this later if we decide to use different endpoints / filters.

frontend/src/component/insights/sections/PerformanceInsights.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { InsightsFilters } from 'component/insights/InsightsFilters';
+import {
+    StyledChartContainer,
+    StyledWidget,
+    StyledWidgetContent,
+    StyledWidgetStats,
+} from '../InsightsCharts.styles';
+
+export const PerformanceInsights: FC = () => {
+    const statePrefix = 'performance-';
```
thomasheartman, line 28: Each section has its own state now. It's not the prettiest, but I suspect we'll change this during the lifetime of the project as we decide on API endpoints etc., so I don't want to spend too much time on it right now.

PR #10040, opened 2025-05-28
feat: add required cdn events

gastonfournier:
## About the changes
Exposes configuration revision service UPDATE_EVENT constant so it can be used, and we also add a new event for cdn token created


sighphyre (approved):
LGTM!

PR #10046, opened 2025-05-28
chore: simplify serving of static openapi assets

gastonfournier:
Use existing public folder from frontend to serve opeanapi static assets. We also benefit from frontend image optimizations


sastromskis (approved):
LG!

nunogois (approved):

PR #10060, opened 2025-05-29
feat: remove isAPI from userSchema response

gastonfournier:
isAPI is not needed for responses, removing it from the system is much more complex, so I had to cut it with a scalpel...


nunogois (approved):
LG. I assume we're still using `isAPI` internally? Have we considered only including it in our queries when needed, instead of filtering it out for responses?

PR #10061, opened 2025-05-29
chore!: remove deprecated get project health report

nunogois:
https://linear.app/unleash/issue/2-3365/remove-get-apiadminprojectsprojectidhealth-report-deprecated-in-660

Removes GET `/api/admin/projects/{projectId}/health-report` which was deprecated in v6.6.
Also cleans up related code.

Seems like the only use we have for `project-health-service` now is updating project health, which is still used in e.g. insights. I wonder if this logic should move somewhere else.


gastonfournier (approved):
LG! I'm not sure if someone is using this API, but if you checked, go ahead

nunogois:
v7 is already out, so I'm closing this one for now.

PR #10062, opened 2025-05-30
chore: print out 10 unknown flag names

gastonfournier:
- Move logs, and replace a debug log with just the info log while adding appName to it
- Print out at most 10 invalid flag names


gastonfournier:
> LGTM! I'm afraid the unknown flags might be a bit too noisy though.

We're not adding new print statements, but we're printing out a few names. The idea is to learn with minimum effort before we build a feature around this.

PR #10066, opened 2025-06-02
feat: lifecycle trends migration

sjaanus:
Adding a single table to capture all lifecycle trends data.

One field presents a challenge: `median_time_in_stage_days`. This value is calculated per `stage`, not per `flag_type`. As a result, we would need to duplicate the total median time across each flag type. This isn’t a major issue.

An alternative would be to create a separate table solely for the median values, but that seems like overkill.

### DIFF
--- a/src/migrations/20250530103216-lifecycle-trends.js
+++ b/src/migrations/20250530103216-lifecycle-trends.js
@@ -6,7 +6,7 @@ exports.up = function(db, cb) {
             CREATE TABLE IF NOT EXISTS lifecycle_trends (
                 id TEXT NOT NULL,
                 stage TEXT NOT NULL CHECK (stage IN ('develop', 'production', 'cleanup')),
-                flag_type TEXT NOT NULL CHECK (category IN ('experimental', 'release', 'permanent')),
+                flag_type TEXT NOT NULL CHECK (flag_type IN ('experimental', 'release', 'permanent')),
                 median_time_in_stage_days DECIMAL(10,2) DEFAULT 0,
                 flags_older_than_week INTEGER DEFAULT 0,
                 new_flags_this_week INTEGER DEFAULT 0,
--- a/src/migrations/20250530103216-lifecycle-trends.js
+++ b/src/migrations/20250530103216-lifecycle-trends.js
@@ -11,7 +11,7 @@ exports.up = function(db, cb) {
                 flags_older_than_week INTEGER DEFAULT 0,
                 new_flags_this_week INTEGER DEFAULT 0,
                 created_at TIMESTAMP DEFAULT now(),
-                PRIMARY KEY (id, stage, flag_type)
+                PRIMARY KEY (id, stage, flag_type, project)
                 );
         `,
         cb,
--- a/src/migrations/20250530103216-lifecycle-trends.js
+++ b/src/migrations/20250530103216-lifecycle-trends.js
@@ -5,14 +5,15 @@ exports.up = function(db, cb) {
         `
             CREATE TABLE IF NOT EXISTS lifecycle_trends (
                 id TEXT NOT NULL,
-                stage TEXT NOT NULL CHECK (stage IN ('develop', 'production', 'cleanup')),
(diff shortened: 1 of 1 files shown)


src/migrations/20250530103216-lifecycle-trends.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+exports.up = function(db, cb) {
+    db.runSql(
+        `
+            CREATE TABLE IF NOT EXISTS lifecycle_trends (
+                id TEXT NOT NULL,
```
thomasheartman, line 7: I assume we'll follow the pattern established by flag_trends and format this is yyyy-ww, e.g. 2024-43, right?

src/migrations/20250530103216-lifecycle-trends.js
```diff
@@ -0,0 +1,22 @@
...
+                stage TEXT NOT NULL CHECK (stage IN ('develop', 'production', 'cleanup')),
+                flag_type TEXT NOT NULL CHECK (flag_type IN ('experimental', 'release', 'permanent')),
+                median_time_in_stage_days DECIMAL(10,2) DEFAULT 0,
```
thomasheartman: If it's a median, does it really need to be a decimal? Won't an integer be sufficient? Or do we wanna record the number of hours spent in a stage? E.g. if the median flag for production was there for 36 hours, do we log 1 day or 1.5 days?

thomasheartman (commented):
I'm not sure this is everything we need. As [mentioned in the slack thread](https://unleash-internal.slack.com/archives/C046LV6HH6W/p1748858732295079?thread_ts=1748845692.751559&cid=C046LV6HH6W), here's some thoughts:

On your comment about duplicating the time in stage for each flag type: yeah, this might be fine. If one of them ever gets out of sync (e.g. experimental says 5, but killswitch says 7), what do we do? For guaranteed data consistency, it might be worth breaking it out, anyway.

Additionally for time in stage: I think we’ll need to record the number of data points used to calculate this. ~~If we want to be able to show accurate averages across any combination of projects,~~ Hang on … if we’re showing median instead of averages, how do we … average that across projects? Or do we median it across projects too? That can get into some really screwy skewing. Do you have a solution for this? If not, I think we’ll need to think about it a bit. 

And speaking of projects, we’ll need to store this on a per-project basis (the UI includes a project selector), so that’s also missing.

We will also need to store how many flags were created and archived for each week, but that should probably be its own table to avoid even more data duplication. In fact, maybe we should extract time in stage too in that case?

src/migrations/20250530103216-lifecycle-trends.js
sjaanus: Yes

src/migrations/20250530103216-lifecycle-trends.js
```diff
@@ -0,0 +1,23 @@
...
+        `
+            CREATE TABLE IF NOT EXISTS lifecycle_trends (
+                id TEXT NOT NULL,
+                stage TEXT NOT NULL CHECK (stage IN ('initial','develop', 'production', 'cleanup', 'archived')),
+                flag_type TEXT NOT NULL CHECK (flag_type IN ('experimental', 'release', 'permanent')),
+                project VARCHAR(255) NOT NULL REFERENCES projects (id) ON DELETE CASCADE,
+                flags_older_than_week INTEGER DEFAULT 0,
+                new_flags_this_week INTEGER DEFAULT 0,
+                created_at TIMESTAMP DEFAULT now(),
+                PRIMARY KEY (id, stage, flag_type)
```
thomasheartman: Oh, wait, we might need to add "project" here too, right? Otherwise 2025-04 initial experimental would cause conflicts for projects A and B.
```suggestion
                PRIMARY KEY (id, project, stage, flag_type)
```

thomasheartman (approved):
Nice one! This should address all the concerns I had previously 🙌🏼  The only thing left is that I think you'll need to expand the primary key here to include project too.

PR #10069, opened 2025-06-02
chore(adminapi)!: Remove deprecated project overview

chriswk:
BREAKING CHANGE: As part of the preparation for a new major (7.0) this removes /api/admin/projects/{projectId} endpoint. It has been deprecated since 5.8, and we don't use it anymore in our frontend.


chriswk:
Did not drop the feature-variants-schema here, because we reused the same schema for environment specific variants as well

src/test/e2e/api/admin/favorites.e2e.test.ts
```diff
@@ -122,28 +115,6 @@ beforeEach(async () => {
...
 });
 
-test('should be favorited in project endpoint', async () => {
-    const featureName = 'test-feature';
-    await createFeature(featureName);
-    await favoriteFeature(featureName);
-    await favoriteProject();
-
-    const { body } = await app.request
-        .get('/api/admin/projects/default')
```
gastonfournier, line 132: Should we still keep these tests and replace the url with `get('/api/admin/projects/default/overview')` ?

gastonfournier (approved):
LG!

PR #10070, opened 2025-06-02
fix: add event listeners for release plan updates in metrics

Tymek:
## About the changes
Milestone changes should be visible in metrics, in the same way as strategy changes are.


daveleek (approved):
Great! Thanks!

PR #10072, opened 2025-06-03
chore(apitoken)!: remove ILegacyCreateApiToken

daveleek:


src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -109,7 +110,9 @@ export class ProjectApiTokenController extends Controller {
                 openApiService.validPath({
                     tags: ['Projects'],
                     operationId: 'createProjectApiToken',
-                    requestBody: createRequestSchema('createApiTokenSchema'),
+                    requestBody: createRequestSchema(
+                        'createProjectApiTokenSchema',
+                    ),
```
daveleek, line 115: This in effect removes the project and projects fields (see src/lib/openapi/spec/create-project-api-token-schema.ts) without proper deprecation before removal. That said, they shouldn't throw any warnings, and will be ignored. We also didn't support them properly (see removed code further down) and we keep the usage of projectId from the req.params as before, so maybe it's a little less extreme change?

src/lib/routes/admin-api/api-token.ts
```diff
@@ -299,25 +300,24 @@ export class ApiTokenController extends Controller {
...
-                `Admin tokens are disabled in this instance. Use a Service account or a PAT to access admin operations instead`,
-            );
-        }
+
         const hasPermission = await this.accessService.hasPermission(
             req.user,
             permissionRequired,
         );
         if (hasPermission) {
             const token = await this.apiTokenService.createApiToken(
```
daveleek: this.apiTokenService.createApiToken is deprecated completely, will have a look at replacing and removing

src/lib/routes/admin-api/api-token.ts
```diff
@@ -299,25 +300,24 @@ export class ApiTokenController extends Controller {
...
-                createToken,
+                {
+                    ...createToken,
+                    projects: createToken.projects ?? [createToken.project],
```
daveleek: going to double check this, could possibly yeet project here

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -178,30 +181,17 @@ export class ProjectApiTokenController extends Controller {
...
-                201,
-                res,
-                apiTokenSchema.$id,
-                serializeDates(token),
-                { location: `api-tokens` },
-            );
-        } else {
-            res.statusMessage =
-                'Project level tokens can only be created for one project';
-            res.status(400);
```
daveleek, line 203: No point supporting the property in theory if it fails in practice. Removing this, and changing the endpoints schema to a new one that doesn't have project/projects

src/lib/services/api-token-service.ts
```diff
@@ -277,11 +275,16 @@ export class ApiTokenService {
      * @deprecated This may be removed in a future release, prefer createApiTokenWithProjects
      */
     public async createApiToken(
-        newToken: Omit<ILegacyApiTokenCreate, 'secret'>,
```
daveleek, line 280: I'll have a look at removing this method

src/lib/db/api-token-store.ts
```diff
@@ -65,8 +64,8 @@ const tokenRowReducer = (acc, tokenRow) => {
 };
 
 const toRow = (newToken: IApiTokenCreate) => ({
-    username: newToken.tokenName ?? newToken.username,
-    token_name: newToken.tokenName ?? newToken.username,
+    username: newToken.tokenName,
```
gastonfournier, line 67: Should we get rid of username in the db? I don't mean do it now, I mean, let's create a task in the current project so we don't forget. This would not be a breaking change but it's nice if we can do it later

src/lib/routes/admin-api/api-token.ts
```diff
@@ -299,25 +300,24 @@ export class ApiTokenController extends Controller {
...
             permissionRequired,
         );
         if (hasPermission) {
-            const token = await this.apiTokenService.createApiToken(
-                createToken,
+            const token = await this.apiTokenService.createApiTokenWithProjects(
+                {
+                    ...createToken,
+                    projects: createToken.projects,
+                },
```
gastonfournier: This should be equivalent:
```suggestion
            const token = await this.apiTokenService.createApiTokenWithProjects(createToken,
```

src/lib/routes/admin-api/project/api-token.ts
gastonfournier: I think this is fine, you're not using `additionalProperties: false` which is the right choice for input, and it also means it will still accept old payloads, but we're now honest to say we don't expect to do anything with them if you send them :+1:

src/lib/routes/admin-api/project/api-token.ts
```diff
@@ -160,10 +163,10 @@ export class ProjectApiTokenController extends Controller {
     }
 
     async createProjectApiToken(
-        req: IAuthRequest,
+        req: IAuthRequest<{ projectId: string }, CreateProjectApiTokenSchema>,
```
gastonfournier, line 166: Nice!

src/lib/schema/api-token-schema.ts
```diff
@@ -6,29 +6,18 @@ import { DEFAULT_ENV } from '../util/constants.js';
 export const createApiToken = joi
     .object()
     .keys({
-        username: joi.string().optional(),
         tokenName: joi.string().optional(),
```
gastonfournier: Shouldn't tokenName be mandatory now? I think it was optional due to the "either username or tokenName" logic

src/lib/schema/create-project-api-token-schema.ts
```diff
@@ -0,0 +1,20 @@
+import joi from 'joi';
+import { ApiTokenType } from '../types/model.js';
+import { DEFAULT_ENV } from '../util/constants.js';
+
+export const createProjectApiToken = joi
+    .object()
+    .keys({
+        tokenName: joi.string().optional(),
```
gastonfournier: Should this be mandatory?

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -230,36 +188,6 @@ test('Token-admin should not be allowed to create token', async () => {
     await destroy();
 });
 
-test('An admin should be forbidden to create an admin token', async () => {
```
gastonfournier, line 233: Isn't this test still valid?

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -188,40 +179,7 @@ test('Only token-admins should be allowed to create token', async () => {
...
-            username: 'default-admin',
-            type: 'admin',
-        })
-        .set('Content-Type', 'application/json')
-        .expect(403);
-
-    await destroy();
-});
-
-test('Token-admin should not be allowed to create token', async () => {
```
gastonfournier, line 200: I think this test is now named `Only token-admins should be allowed to create token` and it's actually doing something else.

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
```diff
@@ -431,70 +359,13 @@ describe('Fine grained API token permissions', () => {
...
                 .send({
-                    username: 'default-frontend',
+                    tokenName: 'default-frontend',
                     type: 'frontend',
                 })
                 .set('Content-Type', 'application/json')
                 .expect(403);
             await destroy();
         });
-        test('should NOT be allowed to create ADMIN tokens', async () => {
```
gastonfournier, line 441: Isn't this still a valid test?

gastonfournier (approved):
Great job @daveleek! Leaving a few comments but I feel this is good :clap:

src/lib/db/api-token-store.ts
daveleek: It's outlived its usefulness, so I guess all that is left is to remove it (we'd have to wait for some time though since it's a subtractive db change?)

src/lib/routes/admin-api/api-token.ts
daveleek: Argh yes, good point!

src/lib/schema/create-project-api-token-schema.ts
daveleek: Good point, making it required

src/lib/db/api-token-store.ts
gastonfournier: absolutely, but it'd make this change unnecessary bigger. That's why I'd suggest we let it be for now

src/test/e2e/api/admin/api-token.e2e.test.ts
```diff
@@ -257,22 +254,21 @@ test('should not create token for invalid environment', async () => {
...
-            type: 'admin',
+            type: 'client',
             environment: '*',
         })
         .set('Content-Type', 'application/json')
         .expect(400);
 });
 
-test('only one of tokenName and username can be set', async () => {
+test('can not create token with admin type', async () => {
```
daveleek, line 268: We have this test @gastonfournier

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
daveleek: It kinda is, but we already have a 400-check on creating admin tokens. How many tests should we have?

src/test/e2e/api/admin/api-token.e2e.test.ts
gastonfournier: Nice

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
gastonfournier: One is enough

src/test/e2e/api/admin/api-token.auth.e2e.test.ts
daveleek: I think this is git playing tricks, I checked the file locally and this is how this test looks:

```
test('Only token-admins should be allowed to create token', async () => {
    expect.assertions(0);

    const preHook = (app, config, { userService, accessService }) => {
        app.use('/api/admin/', async (req, res, next) => {
            const role = await accessService.getPredefinedRole(RoleName.EDITOR);
            req.user = await userService.createUser({
                email: 'editor2@example.com',
                rootRole: role.id,
            });
            next();
        });
    };

    const { request, destroy } = await setupAppWithCustomAuth(
        stores,
        preHook,
        undefined,
        db.rawDatabase,
    );

    await request
        .post('/api/admin/api-tokens')
        .send({
            tokenName: 'default-admin',
            type: 'admin',
        })
        .expect(403);

    await destroy();
});
```

PR #10080, opened 2025-06-04
chore!: remove deprecated default env from new installs

gastonfournier:
**BREAKING CHANGE**: DEFAULT_ENV changed from `default` (should not be used anymore) to `development`

## About the changes
- Only delete default env if the install is fresh new.
- Consider development the new default. The main consequence of this change is that the default is no longer considered `type=production` environment but also for frontend tokens due to this assumption: https://github.com/Unleash/unleash/blob/724c4b78a27ef6d974468a7e11065a6460b8b8c7/src/lib/schema/api-token-schema.test.ts#L54-L59 (I believe this is mostly due to the [support for admin tokens](https://github.com/Unleash/unleash/pull/10080#discussion_r2126871567))
- `feature_toggle_update_total` metric reports `n/a` in environment and environment type as it's not environment specific

### DIFF
--- a/src/lib/features/feature-search/feature.search.e2e.test.ts
+++ b/src/lib/features/feature-search/feature.search.e2e.test.ts
@@ -9,11 +9,9 @@ import {
 import getLogger from '../../../test/fixtures/no-logger.js';
 import type { FeatureSearchQueryParameters } from '../../openapi/spec/feature-search-query-parameters.js';
 import {
-    CREATE_FEATURE_STRATEGY,
     DEFAULT_PROJECT,
     type IUnleashStores,
     TEST_AUDIT_USER,
-    UPDATE_FEATURE_ENVIRONMENT,
 } from '../../types/index.js';
 import { DEFAULT_ENV } from '../../util/index.js';
 
@@ -22,9 +20,7 @@ let db: ITestDb;
 let stores: IUnleashStores;
 
 beforeAll(async () => {
-    db = await dbInit('feature_search', getLogger, {
-        dbInitMethod: 'legacy' as const,
-    });
+    db = await dbInit('feature_search', getLogger);
     app = await setupAppWithAuth(
         db.stores,
         {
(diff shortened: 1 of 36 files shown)


src/lib/features/feature-toggle/tests/archive-feature-toggles.e2e.test.ts
```diff
@@ -125,26 +123,6 @@ test('Should disable all environments when reviving a toggle', async () => {
...
-        enabled: true,
-        type: 'production',
-        sortOrder: 2,
-    });
-
-    await db.stores.featureEnvironmentStore.addEnvironmentToFeature(
-        'feat-proj-1',
-        'default',
-        true,
-    );
```
gastonfournier, line 147: Not needed due to migrating to template db which is already migrated

src/lib/features/frontend-api/global-frontend-api-cache.ts
```diff
@@ -141,7 +141,7 @@ export class GlobalFrontendApiCache extends EventEmitter {
 
     private environmentNameForToken(token: IApiUser): string {
         if (token.environment === ALL_ENVS) {
-            return 'default';
+            return DEFAULT_ENV;
```
gastonfournier: This is kind of odd, not sure how this should work in reality because we should return a list of environments maybe... cc @sjaanus @kwasniew maybe you can advise what's best in this case, at least now the tests will pass but I'm worried of some unpredictable behavior

src/lib/features/feature-search/feature.search.e2e.test.ts
```diff
@@ -46,38 +42,6 @@ beforeAll(async () => {
...
-    await app.linkProjectToEnvironment('default', 'production');
-
-    await stores.accessStore.addPermissionsToRole(
-        body.rootRole,
-        [
-            { name: UPDATE_FEATURE_ENVIRONMENT },
-            { name: CREATE_FEATURE_STRATEGY },
-        ],
-        'production',
-    );
```
gastonfournier, line 79: No need to setup db when using a migrated database

src/lib/metrics.ts
```diff
@@ -902,8 +902,8 @@ export function registerPrometheusMetrics(
         featureFlagUpdateTotal.increment({
             toggle: featureName,
             project,
-            environment: 'default',
-            environmentType: 'production',
+            environment: 'n/a',
+            environmentType: 'n/a',
```
gastonfournier, line 906: Feature update is not environment-specific. Following the same pattern as in other metrics

src/migrations/20250604110546-remove-default-env-from-new-installs.js
```diff
@@ -0,0 +1,23 @@
...
+                // If there are exactly 4 events, it means this is a new install
+                db.runSql(
+                        `DELETE FROM environments WHERE name = 'default';`,
+
+                    cb);
+            } else {
+                // If there are not exactly 4 events, do nothing
+                cb();
+            }
+        },
```
gastonfournier, line 17: Conditional deletion of the environment

src/test/e2e/api/admin/metrics.e2e.test.ts
```diff
@@ -179,9 +177,9 @@ test('should save multiple projects from token', async () => {
         .expect('Content-Type', /json/)
         .expect(200);
 
-    expect(body).toMatchObject({
-        applications: [
-            {
+    expect(body.applications).toEqual(
+        expect.arrayContaining([
+            expect.objectContaining({
```
gastonfournier, line 182: Because the DB persist across tests in this test case there are more applications registered, and it's not predictable which ones will exist, so it's better to assert this

src/test/e2e/helpers/database-init.ts
```diff
@@ -102,7 +31,6 @@ export interface ITestDb {
 }
 
 type DBTestOptions = {
-    dbInitMethod?: 'legacy' | 'template';
```
gastonfournier, line 105: Clean up and no longer use legacy db init

PR #10081, opened 2025-06-04
fix(docs): enterprise image zoom

gastonfournier:


website/docs/how-to/how-to-create-feature-toggles.md
```diff
@@ -131,12 +131,6 @@ This can be done at any point after you've created your flag.
 
 :::
 
-:::tip API: add variants
-
-Use the [create variants endpoint](/reference/api/unleash/overwrite-feature-variants) to create or overwrite existing variants in all environments. The payload should be your desired variant configuration.
-
-:::
```
gastonfournier, line 138: @melindafekete the problem with this is that the endpoint was removed, so there's no page for overwrite-feature-variants anymore. I believe this will require some rewriting of this specific page

melindafekete: But the page still exists: https://docs.getunleash.io/reference/api/unleash/overwrite-feature-variants 🤔

gastonfournier: That's because publish is failing to publish new docs and therefore it's still not deleting it. But it raises an interesting question, does removing an endpoint mean we have to get rid of this documentation page? Maybe we have to move it to an archive or something...

melindafekete: Hmm good question, I don't think an archive in the docs is necessary. The release notes will have a full list of the things we deprecated with guidance on what to use instead. I assume we'll also have a migration guide for v7?

website/docs/how-to/how-to-create-feature-toggles.md
```diff
@@ -131,14 +131,8 @@ This can be done at any point after you've created your flag.
 
 :::
 
-:::tip API: add variants
-
-Use the [create variants endpoint](/reference/api/unleash/overwrite-feature-variants) to create or overwrite existing variants in all environments. The payload should be your desired variant configuration.
```
melindafekete, line 136: Should we point to https://docs.getunleash.io/reference/api/unleash/overwrite-environment-feature-variants instead?

gastonfournier: That's not available "by default". The default is strategy variants

melindafekete (approved):
Thank you, @gastonfournier!

website/docs/how-to/how-to-create-feature-toggles.md
gastonfournier: Yes, I just pushed a change with the link to strategy-variants and removing the image which was old

PR #10086, opened 2025-06-05
fix: orval for ESM

Tymek:
## About the changes
I updated Orval and configured it to be compatible with v7.


frontend/orval.config.ts
```diff
@@ -14,6 +16,7 @@ module.exports = {
             schemas: 'models',
             client: 'swr',
             clean: true,
+            propertySortOrder: 'Alphabetical',
```
Tymek, line 19: generates smaller diff vs previous generation. It would be nice to swap that to default `Specification` value, but not in this PR

frontend/src/hooks/api/getters/useInsights/useInsights.ts
```diff
@@ -27,6 +27,7 @@ export const useInsights = (
     return {
         insights:
             data ||
+            /* @ts-expect-error FIXME: lifecycle trends */
```
Tymek: CC @thomasheartman @sjaanus

Tymek: Return type when loading needs an update

thomasheartman: Aye, aye! 🫡 Not using data from the hook yet, but we'll get there. For now, let's also add the flag name, so that at least we'll catch it by then.
```suggestion
            /* @ts-expect-error FIXME (lifecycleMetrics): lifecycle trends */
```

PR #10088, opened 2025-06-05
feat: update backend to include `technicalDebt` field

Tymek:
## About the changes
Deprecate `health` for a more descriptive `technicalDebt` across project-related services and schemas.

Changes:
- added `technicalDebt` field across project models, calculated as `100 - health`
- marked original `health` field as deprecated with JSDoc
- Updated OpenAPI schemas
- Modified services to calculate tech debt alongside health

### Important files
- `src/lib/types/model.ts
- `src/lib/features/project/project-read-model.ts`
- `src/lib/openapi/spec/project-insights-schema.ts`


src/lib/features/project-insights/project-insights-service.ts
```diff
@@ -113,6 +124,7 @@ export class ProjectInsightsService {
 
         return {
             health: project?.health || 0,
+            technicalDebt: 100 - (project?.health || 0),
```
Tymek, line 127: I'm hesitant to modify this deeper then on a service level. If we introduce duplication on a store level we will be fetching all of this twice from DB for as long as `health` is deprecated and not removed.

src/lib/features/project-insights/project-insights-service.ts
```diff
@@ -151,9 +163,23 @@ export class ProjectInsightsService {
...
             members,
+            /**
+             * @deprecated
+             */
+            health: {
+                rating: health.rating,
+                activeCount: health.activeCount,
+                potentiallyStaleCount: health.potentiallyStaleCount,
+                staleCount: health.staleCount,
+            },
```
Tymek, line 182: cannot be directly mapped as `health: health,` because this object now has `health.technicalDebt` property, but there are snapshot tests for this part.

kwasniew (approved):

PR #10089, opened 2025-06-05
Fix(1-3804)/store flag creation form state

thomasheartman:
Add flag creation form data to localstorage.


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureTogglesHeader/CreateFeatureDialog.tsx
```diff
@@ -103,6 +106,15 @@ const CreateFeatureDialogContent = ({
     const navigate = useNavigate();
     const openFeatureCreatedFeedback = useFeatureCreatedFeedback();
 
+    const [storedFlagConfig, storeFlagConfig] =
```
kwasniew: this hook is modelled after useState that normally returns a tuple [val, setVal] with an explicit value and a setter. The naming convention here confused me since it's not very idiomatic and I thought it's the value part listed twice.

kwasniew (approved):

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureTogglesHeader/CreateFeatureDialog.tsx
thomasheartman: Sure, I'll change it 🙋🏼

PR #10101, opened 2025-06-10
chore!: remove project health report frontend

daveleek:
Frontend only-task of https://github.com/Unleash/unleash/pull/10061

Removes the frontend for deprecated project health report.


frontend/src/component/project/Project/Project.tsx
```diff
@@ -373,8 +372,6 @@ export const Project = () => {
                 }}
             />
             <Routes>
-                {/* FIXME: remove /health with `healthToTechDebt` flag - redirect to project status */}
```
gastonfournier, line 376: Should we look into removing this flag? It seems to be in quite some places
https://app.unleash-hosted.com/hosted/projects/dx/features/healthToTechDebt is from DX and not rolled out

PR #10110, opened 2025-06-11
delete legacy constraint accordion

thomasheartman:
This PR continues the cleanup after removing the addEditStrategy flag (part 2 of ???). The primary purpose of this PR is to delete and remove all references to the LegacyConstraintAccordion.

I've gone and updated all references to the legacy files in external components and verified manually that they still work. 

Most of the files in this PR are changing references. I've extracted two bits into more general constants/utils:
1. Constraint IDs are a symbol. it was exported as a const from the previous createEmptyConstraint file. I've moved it into constants.
2. formatOperatorDescription was similarly used all over the place, so I've placed it in the shared utils directory.

In reviewing this, you can ignore any changes in the legacy constraint accordion folder, because that's all been deleted. Instead, focus on the changes in the other files. It's primarily just import updates, but would be good to get a second set of eyes, anyway.


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/FeatureStrategyConstraintAccordionList/FeatureStrategyConstraintAccordionList.tsx
```diff
@@ -47,11 +46,52 @@ const useConstraintLimit = (constraintsCount: number) => {
...
+    }));
+
+    const onAdd =
+        addConstraint &&
+        (() => {
+            addConstraint(context[0].name);
+        });
+
+    return { onAdd, state, context };
+};
```
thomasheartman, line 88: These functions were previously exported from the legacy ConstraintAccordionList. They're not used anywhere else, so I've moved them in here.

PR #10111, opened 2025-06-11
fix: use technicalDebt property from backend

Tymek:
## About the changes
Frontend should load `technicaDebt` from backend instead of re-calculating it.

### DIFF
--- a/src/lib/features/productivity-report/productivity-report-view-model.ts
+++ b/src/lib/features/productivity-report/productivity-report-view-model.ts
@@ -27,6 +27,7 @@ export const productivityReportViewModel = ({
     userName,
     userEmail,
     ...metrics,
+    technicalDebt: Math.max(0, 100 - metrics.health).toString(),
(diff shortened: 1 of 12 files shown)


src/lib/features/productivity-report/productivity-report-view-model.ts
```diff
@@ -27,6 +27,7 @@ export const productivityReportViewModel = ({
     userName,
     userEmail,
     ...metrics,
+    technicalDebt: Math.max(0, 100 - metrics.health).toString(),
     unleashUrl,
     healthColor() {
```
kwasniew, line 32: what about healthColor? is it gonna change to technicalDebt color?

Tymek: I could rename it, but for 'color', it will work the same. I think minimizing changes is better then re-writing it in this case.

Issue #10121, opened 2025-06-11
Creating Unleash user on /admin/create-user fails when FIPS is enabled


pandrija:
### Describe the bug

After enabling FIPS on our Unleash (enterprise) instance, we started getting errors when creating new users:

Insert into "users" ("created_at", "email", "email_hash", "image_url", "name", "username")
values ($1, $2, md5($3), DEFAULT, $4, DEFAULT)
returning "id", "name", "username", "email", "image_url", "seen_at", "is_service", "scim_id",
"login_attempts", "created_at" - could not compute MD5 hash: disabled for FIPS

Are there any plans to replace MD5 with something like SHA-256?
Also, are there any other functionalities that might be failing for us (due to FIPS) that we are not aware yet?

### Steps to reproduce the bug

1. Log-in into Unleash UI
2. Navigate to /admin/create-user
3. Fill all the data and click"Create User"

### Expected behavior

User is created

### Logs, error output, etc.

```bash
Toast displayed:
Insert into "users" ("created_at", "email", "email_hash", "image_url", "name", "username")
values ($1, $2, md5($3), DEFAULT, $4, DEFAULT)
returning "id", "name", "username", "email", "image_url", "seen_at", "is_service", "scim_id",
"login_attempts", "created_at" - could not compute MD5 hash: disabled for FIPS
```

### Unleash version

Unleash Enterprise version	6.10.1

### Subscription type

Enterprise

### Hosting type

Self-hosted

### SDK information (language and version)

N/A

gastonfournier:
This should be fixed from version [v7.0.1](https://github.com/Unleash/unleash/releases/tag/v7.0.1) onwards.

With regards to your question: 

> Also, are there any other functionalities that might be failing for us (due to FIPS) that we are not aware yet?

I wouldn't know for sure but I'll ask the team. To my knowledge, we're not FIPS certified, so we may find other things, but as long as they're reported, we can analyse and prioritize them.

gastonfournier:
Just to clarify on the certification side — Unleash isn’t FIPS-certified as a cryptographic module, but we are [FedRAMP-aligned](https://docs.getunleash.io/using-unleash/compliance/fedramp?utm_source=chatgpt.com).
FedRAMP is a broader framework that includes the use of FIPS-validated crypto when deployed in those environments, so the two are related but not equivalent.
If any other FIPS-related issues surface, please do keep reporting them — we’ll continue to review and prioritize as needed.

PR #10125, opened 2025-06-12
fix: md5 is deprecated and fails FEDRAMP. Replace with sha256

chriswk:
#10121 points out that we're using md5 functions still. This PR updates our migrations to no longer use md5 at all (so if you haven't run the migrations, you won't get email hashes until you get to the included migration with this PR). If you've already run the migrations, we'll drop the existing `email_hash varchar(32)` column and replace it with a `email_hash TEXT` column.

We're also replacing the md5 function with `encode(sha256(email), 'hex')`. encode has been supported since PG10, sha256 came with PG11.

Do we want an index on the email_hash? I wasn't sure, but if we want to do lookup we probably should have an index on it (though not a unique one)

### DIFF
--- a/src/migrations/20250612114001-backfill-email-hashes-from-event-log.js
+++ b/src/migrations/20250612114001-backfill-email-hashes-from-event-log.js
@@ -1,16 +1,16 @@
 exports.up = function(db, cb) {
   db.runSql(`
     WITH deleted_users AS (
-        SELECT (pre_data ->> 'id' AS user_id,
-                pre_data ->> 'email' AS email)
+        SELECT (pre_data ->> 'id')::int AS user_id,
+                pre_data ->> 'email' AS email
         FROM events
         WHERE type = 'user-deleted'
         AND created_at >= now() - interval '30 days'
-    );
+    )
     UPDATE users u
-    SET email_hash = encode(sha256(du.email), 'hex')
+    SET email_hash = encode(sha256(du.email::bytea), 'hex')
     FROM deleted_users du
-    WHERE u.id = du.id AND email_hash IS NULL
+    WHERE u.id = du.user_id AND u.email IS NULL;
   `, cb);
 };
 


src/migrations/20250612070900-backfill-email-hashes-using-sha256.js
```diff
@@ -0,0 +1,12 @@
+exports.up = function(db, cb) {
+  db.runSql(`
+    ALTER TABLE users DROP COLUMN email_hash;
```
sjaanus, line 3: We do run into the issue of losing all current email hashes for users who have been deleted.

This might not be a major problem, as email hashes are only used for the last 30 days. If we collect data over the next 30 days, we should be fine.

However, we may want to check with Ivar to confirm whether this data is currently being used for any major expansions. At the moment, it only impacts our hosted offering.

Just as a side note, we don't know exactly when this will happen, since customers upgrade at random times, but it will eventually impact self-hosted as well. Again, the 30-day window applies.

```
export const createGetLicensedUsers =
    (db: Db): GetLicensedUsers =>
    async () => {
        const result = await db('users')
            .countDistinct('email_hash as activeCount')
            .whereNotNull('email_hash')
            .andWhere(function () {
                this.whereNull('deleted_at').orWhere(
                    'deleted_at',
                    '>=',
                    db.raw("NOW() - INTERVAL '30 days'"),
                );
            })
            .first();

        return Number(result?.activeCount ?? 0);
    };

chriswk: Ah, true, we keep the email_hash after deletion, but scrub the email.

chriswk: Could add a new column, leave the old one in place?

chriswk:
This PR has been updated with another migration that backfills from our user_deleted events for the last month prior to the migration run.

PR #10134, opened 2025-06-13
refactor: migrate from make-fetch-happen to ky and use ky natively

gastonfournier:
Migrate from make-fetch-happen to ky

## Summary:

- Replaced all usages of make-fetch-happen with [ky](https://github.com/sindresorhus/ky) for HTTP requests.
- Upgraded nock to v14 so it's capable of mocking native fetch implementation
- Removed the make-fetch-happen dependency
- Ensured all fetch logic is compatible with ky API.

## Why:

- ky provides a modern, lightweight, and promise-based HTTP client with a simpler API.
- Reduces dependencies and simplifies codebase.

## Testing:

We'll do testing in sandbox environment of the modified functionality (which is not much) and we also rely on automated testing.


src/lib/addons/addon.ts
```diff
@@ -75,15 +71,13 @@ export default abstract class Addon {
             this.logger.warn(
                 `Error querying ${url} with method ${
                     method || 'GET'
-                } status code ${e.code}`,
+                } status code ${(e as any).code}`,
```
sighphyre: Pretty sure this cast isn't necessary

```suggestion
                } status code ${e.code}`,
```

src/lib/addons/addon.ts
```diff
@@ -75,15 +71,13 @@ export default abstract class Addon {
...
                 e,
             );
-            res = { status: e.code, ok: false };
+            return { status: e.code, ok: false } as Response;
```
sighphyre, line 77: Okay this cast is janky. But I see why we need it and it's not making things worse, we already had an implicit any here

sighphyre (approved):
Wow, okay, LGTM. That's cool

PR #10135, opened 2025-06-13
Chore(1-3831)/add edit strategy cleanup pt iv

thomasheartman:
This PR addresses and removes the last comment related to the addEditStrategy flag. In doing so, I have also removed the remaining dangling files from the new constraint accordion directory. I believe that everything that's left in there now is currently in use.


frontend/src/component/common/NewConstraintAccordion/ConstraintAccordionView/ConstraintAccordionViewHeader/ConstraintAccordionViewHeader.tsx
```diff
@@ -60,16 +37,7 @@ export const ConstraintAccordionViewHeader = ({
...
-                <ConstraintAccordionViewActions onUse={onUse} />
-            ) : (
-                // @deprecated : remove onEdit and onDelete from current file together with NewConstraintAccordionList and addEditStrategy flag
-                <ConstraintAccordionEditActions
-                    onEdit={onEdit}
-                    onDelete={onDelete}
-                    disableEdit={disableEdit}
-                />
-            )}
+            {onUse ? <ConstraintAccordionViewActions onUse={onUse} /> : null}
```
sjaanus, line 40: I think you changed behaviour here. Previously there always was a component being rendered.

thomasheartman: I don't think it matters, but this is why I wanted you to have a look at this.

Here's a constraint that doesn't have onEdit or onDelete from hosted: 
image


And here it is from my local, where nothing is rendered
image

There's no actions rendered. This is in line with what the ConstraintAccordionEditActions component did: it would only render actions that were passed to it. And because both onEdit and onDelete were always empty, it would never render anything here.

So I think it's safe.

sjaanus: Ok, just wanted to double check. If there was nothing visible before, its fine.

PR #10138, opened 2025-06-13
fix: correct upgrade link

sidoruk-sv:
Description:

| State | URL |
| ------------- | ------------- |
| Not working (404) | https://www.getunleash.io/upgrade_unleash |
| Working (200) | https://www.getunleash.io/upgrade-unleash |


chriswk (approved):
Thank you for catching this.

PR #10146, opened 2025-06-16
chore(1-3835): improve json diff view

thomasheartman:
Replaces the existing JSON diff implementation we use with `json-diff-react`, a react-fork of the popular `json-diff` library. The change is behind a new flag.

The new library has several advantages:
- nicer formatting (including nested objects)
- we don't need to calculate the diff manually anymore
- option to hide/reveal unchanged properties

There's still a few more things to put in place (such as handling of no changes) and overflow handling when you have very long properties.

Here's a few comparison screenies:

Old (below) vs new (above):
image

Fold and unfold:
image
image

In change requests:
image

Strategy re-ordering:
Folded:
image

Unfolded:
image

Old:
image


chriswk (approved):
Nice. I was able to follow what was happening, so have a :+1:  :)

PR #10151, opened 2025-06-17
feat: expose impact metrics

kwasniew:
## About the changes

Screenshot 2025-06-17 at 13 21 55

### DIFF
--- a/src/lib/routes/backstage.ts
+++ b/src/lib/routes/backstage.ts
@@ -44,7 +44,7 @@ class BackstageController extends Controller {
             });
 
             if (this.flagResolver.isEnabled('impactMetrics')) {
-                this.get('/prometheus/impact', async (req, res) => {
+                this.get('/impact/metrics', async (req, res) => {
                     res.set('Content-Type', impactRegister.contentType);
 
                     const metricsOutput = await impactRegister.metrics();


src/lib/features/metrics/impact/impact-register.ts
```diff
@@ -0,0 +1,5 @@
+import { Registry } from 'prom-client';
+
+// Deliberately keep it separate from the global instance metrics register
+// It is a singleton that everyone participating in the impact metrics can import
+export const impactRegister = new Registry();
```
kwasniew, line 5: the most idiomatic way for doing singleton in JS: "just create once" and export

kwasniew: I think it's important to separate instance own metrics from user submitted impact/custom metrics

src/lib/routes/backstage.ts
```diff
@@ -41,6 +42,16 @@ class BackstageController extends Controller {
 
                 res.end(metricsOutput);
             });
+
+            if (this.flagResolver.isEnabled('impactMetrics')) {
```
kwasniew, line 46: I know this will be only resolved at the startup time but for now it's fine

src/lib/features/metrics/impact/impact-register.ts
kwasniew: Code writing the metrics will write to the same registry

PR #10159, opened 2025-06-18
task: added table for requested approvers for CRs

chriswk:
As part of the task to make it possible to send notifications to approvers for a CR, this PR adds a table that can link users to CRs they've been requested to make an approval for.


chriswk:
My thoughts exactly, we just need a way to flag that you've been assigned. CR already has logic for tracking who approved it.

PR #10160, opened 2025-06-18
chore(1-3842): don't reorder constraint properties / make id's non-optional

thomasheartman:
This PR takes two steps towards better constraint handling:

## New type: `IConstraintWithId`

Introduces a new type, `IConstraintWithId`. This is the same as an `IConstraint`, except the constraint id property is required. The idea is that the list of editable constraints should move towards using this instead of just `IConstraint`. That should prevent us (on a type-level) from seeing more of the same kind of errors we saw with the segment constraints yesterday.

I don't want to go ahead and update all the upstream uses of this to IConstraintWithId in this PR, so I'll look at that separately.

## API payload constraint replacer

Introduces an api payload constraint "replacer", which we can use for [JSON.stringify's `replacer` parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter). The current implementation works both for strategies and for segments and has been added to edit + create forms for both of these resources.

This has a couple benefits:
1. We can clearly state exactly how we want them to be rendered, including property order. I've decided to go with context -> operator -> value(s) as the main one (check the screenie), as I believe this is the most logical reading order.
2. We can exclude value/values (whichever one doesn't work with the operator)
3. It doesn't matter how we treat constraints internally, we can still present the payload how we want
4. Importantly: this only affects the stringification for the user-facing API payload, so it's very low risk. It does not affect anything that we actually send to the api.

Here's what it can look like with ordered properties:

image

### DIFF
--- a/frontend/src/utils/createEmptyConstraint.ts
+++ b/frontend/src/utils/createEmptyConstraint.ts
@@ -1,5 +1,5 @@
 import { constraintId } from 'constants/constraintId';
-import { isDateOperator, isSingleValueOperator } from 'constants/operators';
+import { isDateOperator } from 'constants/operators';
 import type { IConstraintWithId } from 'interfaces/strategy';
 import { operatorsForContext } from 'utils/operatorsForContext';
 import { v4 as uuidv4 } from 'uuid';
@@ -9,24 +9,15 @@ export const createEmptyConstraint = (
 ): IConstraintWithId => {
     const operator = operatorsForContext(contextName)[0];
 
-    const makeConstraint = (
-        props: { value: string } | { values: string[] },
-    ): IConstraintWithId => {
-        return {
-            contextName,
-            operator,
-            ...props,
-            caseInsensitive: false,
-            inverted: false,
-            [constraintId]: uuidv4(),
-        };
+    const value = isDateOperator(operator) ? new Date().toISOString() : '';
+
+    return {
+        contextName,
+        operator,
+        value,
+        values: [],
+        caseInsensitive: false,
+        inverted: false,
+        [constraintId]: uuidv4(),
     };
-
-    if (isSingleValueOperator(operator)) {
-        const value = isDateOperator(operator) ? new Date().toISOString() : '';
-
-        return makeConstraint({ value });
-    }
-
-    return makeConstraint({ values: [] });
 };


frontend/src/component/common/NewConstraintAccordion/ConstraintsList/EditableConstraintsList.tsx
```diff
@@ -39,6 +40,17 @@ export const EditableConstraintsList = forwardRef<
...
+    useEffect(() => {
+        if (!constraints.every((constraint) => constraintId in constraint)) {
+            setConstraints(
+                constraints.map((constraint) => ({
+                    [constraintId]: uuidv4(),
+                    ...constraint,
+                })),
+            );
+        }
+    }, [constraints, setConstraints]);
```
thomasheartman, line 52: I'm not ready to make this component only accept constraints with ids yet, so this helps us at least set ids on constraints if they don't have them yet.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.test.ts
```diff
@@ -23,22 +24,9 @@ test('mapping to an editable constraint adds a constraint id if there is none',
...
-    const transformed = toIConstraint(fromIConstraint(constraint));
-
-    expect(Object.keys(transformed)).toEqual([
-        'values',
-        'inverted',
-        'operator',
-        'contextName',
-        'caseInsensitive',
-    ]);
-});
```
thomasheartman, line 44: We won't need this anymore with the new serializers.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.ts
```diff
@@ -90,29 +90,13 @@ export const fromIConstraint = (
...
-        operator,
-        contextName,
-        caseInsensitive,
-        [constraintId]: id,
-    };
     if ('value' in constraint) {
-        return {
-            value: constraint.value,
-            ...baseValues,
-        };
```
thomasheartman, line 111: Same: this isn't necessary anymore.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyForm/FeatureStrategyForm.tsx
thomasheartman, line 1: Unrelated cleanup, but I stumbled across this file, so just trying to keep things clean

kwasniew (approved):

PR #10161, opened 2025-06-18
task: added requested cr approval email template and emailService method

chriswk:
Initial template and emailService method in place here. We should check with stakeholders that this email says what we want to say.

### DIFF
--- a/src/mailtemplates/requested-cr-approval/requested-cr-approval.html.mustache
+++ b/src/mailtemplates/requested-cr-approval/requested-cr-approval.html.mustache
@@ -1,0 +1,380 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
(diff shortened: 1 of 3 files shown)


src/mailtemplates/requested-cr-approval/requested-cr-approval.html.mustache
```diff
@@ -0,0 +1,380 @@
...
+                                    <td valign="top" class="footerContent" mc:edit="footer_content00">
+                                        <a href="https://www.getunleash.io/blog">Unleash blog</a>
+                                        <a href="https://github.com/Unleash/unleash">Github</a>
+                                        <a href="https://slack.unleash.run">Slack community</a>
+                                        <a href="https://www.getunleash.io/support" style="border-right:0;">Help center</a>
+                                    </td>
+                                </tr>
+                                <tr>
+                                    <td valign="top" class="footerContent" mc:edit="footer_content01">
+                                        <em>Copyright &copy; {{ year }} | Bricks Software | All rights reserved.</em>
```
daveleek, line 365: Is the year provided by the framework?

chriswk: no. forgot that key

PR #10163, opened 2025-06-18
fix: Order properties correctly when mapping from Editable Constraint to IConstraint

thomasheartman:
Prevents the property order from changing when constraints are set from the editable constraint component. When we render out the API command, we don't specify the order of properties in the objects, which means that it can change dramatically, which can be a little jarring. 

As it is right now, it first renders in one order when you open the strategy form:

image

And when you navigate to the targeting section of the strategy form, it changes to:
image

This also applies to constraints in segments.

With this change, the order will remain the same before and after.

Additionally, there's some extra tests around constraint ids being kept the same and being set if it doesn't exist.

## Further work

This came about as part of the issue we had with constraint editing in segments being broken as of now. As part of that, I would like to make some further improvements (such as making the ID required when you use a list of editable constraints), but that will be in a follow-up. There are some complications that might not make it a viable option, sadly.

We could also try to stabilize the property order in the API rendering methods (which I think might be a good idea), but because there's multiple different ones, this seems to be a faster solution.

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.test.ts
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.test.ts
@@ -12,7 +12,6 @@ test('mapping to and from retains the constraint id', () => {
 
 test('mapping to an editable constraint adds a constraint id if there is none', () => {
     const constraint = createEmptyConstraint('context');
-    // @ts-expect-error: this violates the type constraint
     delete constraint[constraintId];
 
     const editableConstraint = fromIConstraint(constraint);
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.test.ts
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.test.ts
@@ -24,10 +24,10 @@ test('mapping to an editable constraint adds a constraint id if there is none', 
 
 test('mapping from an empty constraint removes redundant value / values', () => {
     const constraint = createEmptyConstraint('context');
-    expect('value' in constraint).toBe(true);
+    expect(constraint).toHaveProperty('value');
 
     const transformed = toIConstraint(fromIConstraint(constraint));
-    expect('value' in transformed).toBe(false);
+    expect(transformed).not.toHaveProperty('value');
 });
 
 test('mapping to constraint returns properties in expected order', () => {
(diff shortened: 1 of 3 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    expect(editableConstraint[constraintId]).toBeDefined();
+
+    const iConstraint = toIConstraint(editableConstraint);
+    expect(iConstraint[constraintId]).toEqual(editableConstraint[constraintId]);
+});
+
+test('mapping from an empty constraint removes redundant value / values', () => {
+    const constraint = createEmptyConstraint('context');
+    expect('value' in constraint).toBe(true);
```
kwasniew: is there toHaveProperty assertion?

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.ts
```diff
@@ -86,13 +90,29 @@ export const fromIConstraint = (
...
+        operator,
+        contextName,
+        caseInsensitive,
+        [constraintId]: id,
+    };
     if ('value' in constraint) {
-        return constraint;
+        return {
+            value: constraint.value,
+            ...baseValues,
```
kwasniew, line 110: is the order of baseValues spread after the value deliberate. If baseValues happens to have value it will take precedence

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.ts
```diff
@@ -86,13 +90,29 @@ export const fromIConstraint = (
...
+        };
     } else {
-        const { values, ...rest } = constraint;
         return {
-            ...rest,
-            values: Array.from(values),
+            values: Array.from(constraint.values),
+            ...baseValues,
```
kwasniew, line 115: same comment as before

kwasniew (approved):

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.test.ts
thomasheartman: huh. didn't think of that, but yes, there is. Good shout.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/editable-constraint-type.ts
thomasheartman: Yep, this is deliberate. `baseValues` is declared as a specific set of values on line 100. Unless you add it, it will never contain  `value` or `values`. And if it did, you'd probably get red squiggles telling you about it.

PR #10169, opened 2025-06-18
feat: ingest new impact metrics

sighphyre:
Accepts the new impact metrics into the singleton registry and then does nothing with them. If the relevant flag is off, the metrics are stripped from the existing metrics data format and dropped on the floor

### DIFF
--- a/src/lib/features/metrics/instance/metrics.ts
+++ b/src/lib/features/metrics/instance/metrics.ts
@@ -27,6 +27,7 @@ import { CLIENT_METRICS } from '../../../events/index.js';
 import type { CustomMetricsSchema } from '../../../openapi/spec/custom-metrics-schema.js';
 import type { StoredCustomMetric } from '../custom/custom-metrics-store.js';
 import type { CustomMetricsService } from '../custom/custom-metrics-service.js';
+import type { MetricsTranslator } from '../impact/metrics-translator.js';
 
 export default class ClientMetricsController extends Controller {
     logger: Logger;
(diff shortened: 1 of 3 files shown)


src/lib/routes/client-api/index.ts
```diff
@@ -4,13 +4,20 @@ import MetricsController from '../../features/metrics/instance/metrics.js';
...
 import type { IUnleashConfig } from '../../types/index.js';
 import type { IUnleashServices } from '../../services/index.js';
+import { impactRegister } from '../../features/metrics/impact/impact-register.js';
+import { MetricsTranslator } from '../../features/metrics/impact/metrics-translator.js';
 
 export default class ClientApi extends Controller {
     constructor(config: IUnleashConfig, services: IUnleashServices) {
         super(config);
 
+        const metricsTranslator = new MetricsTranslator(impactRegister);
```
kwasniew, line 14: I'm leaning towards inlining this inside the MetricsController or event better MetricsService if possible. The translator is like a calculator rather than I/O that needs to be swappable.

src/lib/features/metrics/instance/metrics.ts
```diff
@@ -150,16 +155,25 @@ export default class ClientMetricsController extends Controller {
         } else {
             try {
                 const { body: data, ip: clientIp, user } = req;
-                data.environment = this.metricsV2.resolveMetricsEnvironment(
-                    user,
-                    data,
-                );
+                const { impactMetrics, ...metricsData } = data;
```
kwasniew, line 158: would it be possible to do the application logic in the service and keep the controller as dumb as possible?

kwasniew (approved):
I'd consider moving everything to a service and not injecting a translator. I like to think of dependencies as swappable options. If you're never gonna exercise the option to swap a dependency then the option is not worth it.

PR #10170, opened 2025-06-18
feat: do not lock until migrations are needed

gastonfournier:
Test to see if we can skip locking the db until we actually realize we need to run migrations


sastromskis (approved):
LGTM!

PR #10199, opened 2025-06-23
feat: expand resolver interface with impact metrics

kwasniew:
## About the changes

Preparing resolver interface to accept new experiment node sdk instance with instance metrics (separate PR in enterprise/cloud repo)

### DIFF
--- a/src/lib/types/experimental.ts
+++ b/src/lib/types/experimental.ts
@@ -314,10 +314,19 @@ export interface IFlagResolver {
     isEnabled: (expName: IFlagKey, context?: IFlagContext) => boolean;
     getVariant: (expName: IFlagKey, context?: IFlagContext) => Variant;
     getStaticContext: () => IFlagContext;
+    impactMetrics?: IImpactMetricsResolver;
 }
 
 export interface IExternalFlagResolver {
     isEnabled: (flagName: IFlagKey, context?: IFlagContext) => boolean;
     getVariant: (flagName: IFlagKey, context?: IFlagContext) => Variant;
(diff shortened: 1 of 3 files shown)


src/lib/types/experimental.ts
```diff
@@ -314,10 +314,19 @@ export interface IFlagResolver {
...
 }
 
 export interface IExternalFlagResolver {
     isEnabled: (flagName: IFlagKey, context?: IFlagContext) => boolean;
     getVariant: (flagName: IFlagKey, context?: IFlagContext) => Variant;
     getStaticContext: () => IFlagContext;
+    impactMetrics?: IImpactMetricsResolver;
+}
+
+export interface IImpactMetricsResolver {
```
kwasniew, line 327: following a pattern where Unleash controls the interface even though it mimicks node interface 1 to 1

src/lib/types/experimental.ts
```diff
@@ -314,10 +314,19 @@ export interface IFlagResolver {
     isEnabled: (expName: IFlagKey, context?: IFlagContext) => boolean;
     getVariant: (expName: IFlagKey, context?: IFlagContext) => Variant;
     getStaticContext: () => IFlagContext;
+    impactMetrics?: IImpactMetricsResolver;
```
sighphyre, line 317: Can we make this non nullable? For our case we should always be able to know it's present, even if it means telling the TS compiler to get over itself with a `!`

kwasniew: We can. I'd prefer to do this gradually by first making it optional, catching up in the other repo and only then making this required here.

PR #10201, opened 2025-06-24
feat: count created feature links with impact metrics

kwasniew:
## About the changes

Changes:
* adding first impact metrics: counter for feature links created
* by strategically placing it in a service we count feature links created directly and via template
* suggested a way to centralize metrics definitions
* added a fake impact metrics resolver for testing purposes

### DIFF
--- a/src/lib/features/metrics/impact/define-impact-metrics.ts
+++ b/src/lib/features/metrics/impact/define-impact-metrics.ts
@@ -1,0 +1,10 @@
+import type { IFlagResolver } from '../../../types/index.js';
+
+export const FEAUTRE_LINK_COUNT = 'feature_link_count';
+
+export const defineImpactMetrics = (flagResolver: IFlagResolver) => {
+    flagResolver.impactMetrics?.defineCounter(
+        FEAUTRE_LINK_COUNT,
+        'Count of feature links',
+    );
+};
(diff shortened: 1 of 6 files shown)


src/lib/features/metrics/impact/define-impact-metrics.ts
```diff
@@ -0,0 +1,10 @@
+import type { IFlagResolver } from '../../../types/index.js';
+
+export const FEAUTRE_LINK_COUNT = 'feature_link_count';
+
+export const defineImpactMetrics = (flagResolver: IFlagResolver) => {
+    flagResolver.impactMetrics?.defineCounter(
```
kwasniew, line 6: I think it's a good idea to define counters/gauges when then app starts

src/lib/features/metrics/impact/define-impact-metrics.ts
```diff
@@ -0,0 +1,10 @@
+import type { IFlagResolver } from '../../../types/index.js';
+
+export const FEAUTRE_LINK_COUNT = 'feature_link_count';
```
kwasniew, line 3: adding const to avoid typos in counting

src/lib/features/metrics/impact/define-impact-metrics.ts
Tymek: Ok, no point in discussing usage ergonomics now

Tymek (approved):

PR #10211, opened 2025-06-25
feat: add transaction context store

sjaanus:
Create transaction context, that generates a random ID for each transaction, but also allows you to define your own id, for example change-request.


src/lib/db/transaction.ts
```diff
@@ -75,13 +76,14 @@ export function withTransactional<S>(
...
 
-    service.transactional = async <R>(fn: (service: S) => R) =>
-        // Maybe: inTransaction(db, async (trx: Knex.Transaction) => fn(serviceFactory(trx)));
-        // this assumes that the caller didn't start a transaction already and opens a new one.
-        db.transaction(async (trx: Knex.Transaction) => {
-            const transactionalService = serviceFactory(trx);
-            return fn(transactionalService);
+    service.transactional = async <R>(fn: (service: S) => R): Promise<R> => {
+        return db.transaction(async (trx) => {
+            return transactionContext.run(async () => {
```
sjaanus: It keeps same logic, but wrapped inside transactioncontext.

src/lib/util/transactionContext.ts
```diff
@@ -0,0 +1,48 @@
...
+    const timestamp = Date.now();
+    const random = Math.floor(Math.random() * 1000);
+    return timestamp * 1000 + random;
+}
+
+const storage = new AsyncLocalStorage<OperationContext>();
+
+export const transactionContext = {
+    run<T>(callback: () => Promise<T>): Promise<T> {
+        const data: OperationContext = {
```
sjaanus: By default a transaction gets a unique transaction id.

src/lib/util/transactionContext.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    getOperationType(): OperationContext['type'] | undefined {
+        return storage.getStore()?.type;
+    },
+
+    getOperationId(): number | undefined {
+        return storage.getStore()?.id;
+    },
+
+    setOperation(operation: OperationContext): void {
```
sjaanus: It is possible to overwrite it with custom operation, like change request.

src/lib/db/transaction.ts
```diff
@@ -86,6 +93,24 @@ export function withTransactional<S>(
...
+    db: Knex,
+): withContextualTransactional<S> {
+    const service = serviceFactory(db) as withContextualTransactional<S>;
+
+    service.trackedTransactional = async <R>(
+        fn: (transactionContext: TransactionContext) => R,
+    ): Promise<R> => {
+        return db.transaction(async (trx) => {
+            const transactionContext = new TransactionContext(trx);
+            return fn(transactionContext);
```
gastonfournier: I'd not be following this... `fn` would be the original function from the service, but here you're changing it's method signature to accept a `TransactionContext`, is that correct?

src/lib/db/transaction.ts
```diff
@@ -41,6 +42,12 @@ export type WithTransactional<S> = S & {
     transactional: <R>(fn: (service: S) => R) => Promise<R>;
 };
 
+export type withContextualTransactional<S> = S & {
+    trackedTransactional: <R>(
+        fn: (transactionContext: TransactionContext) => R,
```
gastonfournier: Shouldn't this be something like this? So you can use the service inside the transaction?
```suggestion
        fn: (service: S, transactionContext: TransactionContext) => R,
```

src/lib/db/transaction.ts
```diff
@@ -75,10 +89,17 @@ export function withTransactional<S>(
...
+        fn: (service: S) => R,
+        transactionContext?: TransactionUserParams,
+    ) =>
         db.transaction(async (trx: Knex.Transaction) => {
+            const defaultContext: TransactionUserParams = {
+                type: 'transaction',
+                value: generateNumericTransactionId(),
+            };
+
+            trx.userParams = transactionContext || defaultContext;
```
sjaanus, line 102: This is the main line, we are populating usersParams of Transaction.

gastonfournier (approved):
I like it!

src/lib/db/transaction.ts
```diff
@@ -1,6 +1,17 @@
 import type { Knex } from 'knex';
 import type { IUnleashConfig } from '../types/index.ts';
 
+export interface TransactionUserParams {
+    type: 'change-request' | 'transaction';
+    value: number;
+}
+
+function generateNumericTransactionId(): number {
```
kwasniew, line 9: we already have ulid library which is a standardized lexographically sortable and unique

PR #10217, opened 2025-06-26
feat: ability to search events by type with pagination

gastonfournier:
This helps us to use this feature to implement partial user sync by observing the changes in the audit log


.github/workflows/build_prs_jest_report.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull_requests: write
+      pull-requests: write
```
gastonfournier, line 17: Unrelated but just spotted this typo: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#example-setting-the-github_token-permissions-for-one-job-in-a-workflow

src/lib/features/events/event-store.ts
```diff
@@ -101,7 +101,7 @@ export interface IEventTable {
 
 const TABLE = 'events';
 
-class EventStore implements IEventStore {
+export class EventStore implements IEventStore {
```
gastonfournier, line 104: This is so we can start importing it by name following https://docs.getunleash.io/contributing/ADRs/back-end/preferred-export

src/lib/features/events/event-store.ts
```diff
@@ -375,7 +375,7 @@ class EventStore implements IEventStore {
     ): Promise<IEvent[]> {
         const query = this.buildSearchQuery(queryParams, params.query)
             .select(options?.withIp ? [...EVENT_COLUMNS, 'ip'] : EVENT_COLUMNS)
-            .orderBy('created_at', 'desc')
+            .orderBy('created_at', params.order || 'desc')
```
gastonfournier, line 378: This is the main change

nunogois (approved):
LGTM!

PR #10230, opened 2025-06-27
feat: transaction grouping, use ulids and strings

sjaanus:
Replace generating random number id based on timestamp and now use ulids.


gastonfournier (approved):
LGTM!

PR #10235, opened 2025-06-30
task: Added name and email of requester to track down spammers

chriswk:
As the title says. This adds name and email of requester to CR approval mails. This will hopefully have users complain to their coworkers rather than Unleash if they get too many mails.


daveleek (changes requested):
Needs requesterName and requesterEmail passed into the email context

daveleek (approved):
LGTM!

PR #10236, opened 2025-06-30
feat: start storing every transaction id in events table

sjaanus:
Every time an event gets inserted, we check if there is transaction data, that we can include.

### DIFF
--- a/src/test/e2e/stores/event-store.e2e.test.ts
+++ b/src/test/e2e/stores/event-store.e2e.test.ts
@@ -484,7 +484,6 @@ test('Should store and retrieve transaction context fields', async () => {
         id: '01HQVX5K8P9EXAMPLE123456',
     };
 
-    // Create a service with transactional capability
     const eventStoreService = withTransactional(
         (db) => new EventStore(db, getLogger),
         db.rawDatabase,


src/lib/features/events/event-store.ts
```diff
@@ -157,7 +161,9 @@ export class EventStore implements IEventStore {
 
     async batchStore(events: IBaseEvent[]): Promise<void> {
         try {
-            await this.db(TABLE).insert(events.map(this.eventToDbRow));
+            await this.db(TABLE).insert(
+                events.map((event) => this.eventToDbRow(event)),
```
kwasniew, line 165: did you see problem with this binding and you changed the syntax?

src/lib/features/events/event-store.ts
```diff
@@ -472,10 +478,14 @@ export class EventStore implements IEventStore {
             featureName: row.feature_name,
             project: row.project,
             environment: row.environment,
+            groupType: row.group_type,
+            groupId: row.group_id,
         };
     }
 
     eventToDbRow(e: IBaseEvent): Omit<IEventTable, 'id' | 'created_at'> {
+        const transactionContext = (this.db as any).userParams;
```
kwasniew: maybe user defined type guard would be more elegant: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/test/e2e/stores/event-store.e2e.test.ts
```diff
@@ -472,3 +473,108 @@ test('Should return empty result when filtering by non-existent ID', async () =>
 
     expect(filteredEvents).toHaveLength(0);
 });
+
+test('Should store and retrieve transaction context fields', async () => {
+    const mockTransactionContext = {
+        type: 'change-request' as const,
+        id: '01HQVX5K8P9EXAMPLE123456',
+    };
+    (eventStore as any).db.userParams = mockTransactionContext;
```
kwasniew: did we mess sth up upstream and we need to do this?

src/lib/events/index.ts
```diff
@@ -397,6 +397,8 @@ export interface IEvent extends Omit<IBaseEvent, 'ip'> {
     id: number;
     createdAt: Date;
     ip?: string;
+    groupType?: string | null;
```
kwasniew: why both null and undefined? could we just do undefined?

src/lib/features/events/event-store.ts
sjaanus: Yes, since I am using **this** now in eventToDbRow method, I needed to not do it inline like before. It broke the code.

PR #10251, opened 2025-06-30
feat: bulk impact metrics

kwasniew:
## About the changes

Support for impact metrics in a bulk endpoint. After this PR is merged we can start implementing bulk impact metrics inside edge (I have WiP code for edge).

Details:
* bulk metrics has strict schema definition in openapi so had to provide detailed description
* saving impactMetrics in bulk behind a flag
* when storing impact metrics fails it should not affect other metrics
* since bulk metrics has strict schema providing incorrect values will reject the data but for single metrics endpoint it will not be the case. single metrics have joi validation but bulk metrics have both joi and openapi validation

### DIFF
--- a/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
+++ b/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
@@ -196,8 +196,14 @@ export default class ClientMetricsServiceV2 {
     }
 
     async registerImpactMetrics(impactMetrics: Metric[]) {
-        const value = await impactMetricsSchema.validateAsync(impactMetrics);
-        this.impactMetricsTranslator.translateMetrics(value);
+        try {
+            const value =
+                await impactMetricsSchema.validateAsync(impactMetrics);
+            this.impactMetricsTranslator.translateMetrics(value);
+        } catch (e) {
+            // impact metrics should not affect other metrics on failure
+            this.logger.warn(e);
+        }
     }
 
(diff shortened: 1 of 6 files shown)


src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -196,8 +196,14 @@ export default class ClientMetricsServiceV2 {
     }
 
     async registerImpactMetrics(impactMetrics: Metric[]) {
-        const value = await impactMetricsSchema.validateAsync(impactMetrics);
-        this.impactMetricsTranslator.translateMetrics(value);
+        try {
```
kwasniew, line 199: this will affect direct SDK metrics uploads and relax the checking

src/lib/features/metrics/instance/metrics.ts
```diff
@@ -275,6 +275,17 @@ export default class ClientMetricsController extends Controller {
...
                     this.config.eventBus.emit(CLIENT_METRICS, data);
                 }
+
+                if (
+                    this.flagResolver.isEnabled('impactMetrics') &&
+                    impactMetrics &&
+                    impactMetrics.length > 0
+                ) {
+                    promises.push(
+                        this.metricsV2.registerImpactMetrics(impactMetrics),
```
kwasniew, line 285: same method as for SDK metric uploads

PR #10262, opened 2025-07-01
chore: remove view diff links in new components + remove colons

thomasheartman:
Removes the view diff hover links (and strategy icons) in the new views and removes trailing colons.

In removing the hover links, I have split up the content of their files (`StrategyTooltipLink` and `SegmentTooltipLink`) into `Change{Segment|Strategy}Name` and `{Segment|Strategy}Diff`. I have reverted the existing tooltip files to their state before I began changing this and added deprecation notices. These old tooltips are only used by the old components, so we don't need to work on them after all.

In doing this work, I've also updated the strategy change diff to handle the new functionality (tabs instead of hover)

The removal of the trailing colons (so that it's `adding strategy gradual rollout` instead of `adding strategy: gradual rollout` is in preparation for the remaining changes to the header that we're introducing with this change. The removal of these is not behind a flag, so I've also done it in the legacy components. This feels like a very low risk change, so it felt like more work to have to check a flag for each of the different instances where we use it.

image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.test.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/StrategyChange.test.tsx
@@ -27,6 +27,12 @@ const strategy = {
 };
 
 const setupApi = () => {
+    testServerRoute(server, '/api/admin/ui-config', {
+        flags: {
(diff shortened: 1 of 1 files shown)


sjaanus (approved):
LG

PR #10274, opened 2025-07-01
chore: AI flag cleanup action

nunogois:
https://linear.app/unleash/issue/2-3662/explore-ai-flag-cleanup-through-a-gh-action

This builds upon:
 - https://github.com/Unleash/unleash/pull/10242
 - https://github.com/Unleash/unleash/pull/10249
 - https://github.com/Unleash/unleash/pull/10250
 - https://github.com/Unleash/unleash/pull/10254

...and **Henry**, our hackathon prototype.

This GH action automatically creates PRs to clean up flags marked as completed.

See example: https://github.com/Unleash/unleash/pull/10273

### DIFF
--- a/.github/workflows/ai-flag-cleanup-pr.yml
+++ b/.github/workflows/ai-flag-cleanup-pr.yml
@@ -254,7 +254,7 @@ jobs:
             const { data: pr } = await github.rest.pulls.create({
               owner,
               repo,
-              title: `[AI] ${flag} flag cleanup`,
+              title: `chore(AI): ${flag} flag cleanup`,
               head: branch,
               base: '${{ inputs.base-branch }}',
               body


.github/workflows/ai-flag-cleanup-pr.yml
```diff
@@ -33,138 +33,116 @@ on:
...
+      - name: Check out new branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.create_branch.outputs.result }}
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
gastonfournier, line 129: Why this email?

.github/workflows/ai-flag-cleanup-pr.yml
```diff
@@ -173,86 +151,122 @@ jobs:
...
-
-      - name: Extract summary from Aider output
-        id: extract_summary
+          python-version: '3.12'
+
+      - name: Install build tools & Aider
         run: |
-          SUMMARY=$(awk '/AI Flag Cleanup Summary/ {found=1} found' .aider.chat.history.md)
+          python -m pip install --upgrade pip
+          pip install aider-chat
```
gastonfournier, line 199: Not sure how long it takes but we could cut some time by having a custom image with these installed instead of `ubuntu-latest` (just an idea for the future)

gastonfournier (approved):
Looks nice! I was thinking if we should open up room for custom instructions (such as: keep the entry in experimental or move the experimental entry to stable-config [ref](https://github.com/Unleash/unleash/pull/10273/files#diff-1c570bc657c8a97a334a43d8e35d8936c00edab077ff725abfda30e15e9fb838L263)), which will be very specific to each code base, but that also opens the door for misuses... so maybe not a good idea...

.github/workflows/ai-flag-cleanup-pr.yml
```diff
@@ -173,86 +151,122 @@ jobs:
...
-              head: branchRef,
-              base: 'refs/heads/${{ inputs.base-branch }}',
-              body: [
-                `This PR cleans up the ${flagName} flag. These changes were automatically generated by AI and should be reviewed carefully.`,
-                '',
-                `Fixes #${{ inputs.issue-number }}`,
-                '',
-                `${{ steps.extract_summary.outputs.summary }}`
-              ].join('\n')
+              title: `[AI] ${flag} flag cleanup`,
```
gastonfournier:
```suggestion
              title: `chore(AI): ${flag} flag cleanup`,
```
Just to stay close to conventional commits

.github/workflows/ai-flag-cleanup-pr.yml
nunogois: Copied from the action I took inspiration from: https://github.com/mirrajabi/aider-github-action/blob/1137d61a782d2d2e589a3b5624ee2d4d7e6e75fb/entrypoint.sh#L29

But this seems to be the email we should use for the GH actions bot:
- https://github.com/orgs/community/discussions/119597#discussion-6506958
- https://github.com/orgs/community/discussions/26560#discussioncomment-3252340

I'm also okay with trying something else if you'd like.

gastonfournier: I was just curious :) thanks for the refs!

nunogois:
> Looks nice! I was thinking if we should open up room for custom instructions (such as: keep the entry in experimental or move the experimental entry to stable-config [ref](https://github.com/Unleash/unleash/pull/10273/files#diff-1c570bc657c8a97a334a43d8e35d8936c00edab077ff725abfda30e15e9fb838L263)), which will be very specific to each code base, but that also opens the door for misuses... so maybe not a good idea...

Technically you can customize / edit the issue description to try and guide it along. But I think sooner or later we'll need to support reacting to comments in that PR.

Regarding misuses, it's up to the "caller" action. Here are some options:
- Tighten the condition for it to run
```
if: github.event.label.name == 'unleash-flag-completed' &&
    contains(fromJson(${{ toJson(needs.approve.outputs.owners) }}), github.actor)
```
- Use a protected environment, requiring human approval
```
environment: ai-flag-cleanup
```

Also, since you need to supply your own AI API key, the caller takes the AI costs.

PR #10276, opened 2025-07-02
chore: prepend changes with "change:" 

thomasheartman:
Adds  "change:" to the beginning of all changes and does some work to align the use of compononents and structure across them (supersedes https://github.com/Unleash/unleash/pull/10260).

In doing so, I have also added new and legacy variants for all different change components, because this has required some hierarchy restructuring every now and then. A reason for doing that was adding the correct wrapping behavior for components, such that on smaller screens, we wouldn't entirely blow out and make the kebab menu invisible and inaccessible.

It also makes it so that we switch to full-width change view earlier (at breakpoint md instead of sm), because at sm, a lot of stuff got hidden before we switched to full-width.

Most changes are trivial updates; I've called out bits of the code that are not in comments.

Rendered, it looks like this:
image 

One interesting and potentially quite useful side-effect, is that all change types now use the exact same set of components in the same fashion, as evidenced by this screenie where I've added outlines to the hierarchy:

image

The one difference is that components without a diff place the "more" kebab menu one layer further inside to facilitate prettier wrapping (the kebab menu can stay on the same line as the other text when wrapping):
image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/Change.styles.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequest/Changes/Change/Change.styles.tsx
@@ -20,7 +20,7 @@ export const Added = styled(Action)(({ theme }) => ({
 export const AddedStrategy = styled(Added, {
     shouldForwardProp: (prop) => prop !== 'disabled',
 })<{ disabled?: boolean }>(({ theme, disabled }) => ({
-    color: disabled ? theme.palette.text.disabled : undefined,
+    color: disabled ? theme.palette.text.secondary : undefined,
 }));
 
 const Change = styled('span')({


frontend/src/component/changeRequest/ChangeRequest/Changes/Change/ChangeSegmentName.tsx
thomasheartman, line 1: This ended up just being a straightforward prop forwarding, so I removed it.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/Change.styles.tsx
```diff
@@ -1,9 +1,53 @@
...
+    </Typography>
+);
+
+export const Deleted = styled(Action)(({ theme }) => ({
+    color: theme.palette.error.main,
+    '::before': { content: '"- "' },
+}));
+
+export const Added = styled(Action)(({ theme }) => ({
+    '::before': { content: '"+ "' },
```
thomasheartman, line 16: The "+" and "-" icons aren't really useful as parts of text that are read out by screen readers (as far as I understand), but they're a useful visual tool, so I'm putting them in via CSS instead.

frontend/src/component/changeRequest/ChangeRequest/Changes/Change/NameWithChangeInfo/NameWithChangeInfo.tsx
thomasheartman, line 1: There were changes made to this component a while back that removed the `ins` element to not have it underlined. However, the `ins` element has semantic meaning, so I added it back in for the meaning to be clearer. The right way to remove the underline (without changing the meaning of the element) is to use CSS, as done here.

sjaanus (approved):
LG. A little hard to follow, a lot of small changes in each file, but nothing alarming by the looks of it.

PR #10277, opened 2025-07-02
feat: group id clickable in event search

sjaanus:
Now when pressing the group id, the query params get updated.
Also the FilterItem appears and it is possible to discard the group id selection through it.

image

### DIFF
--- a/frontend/src/component/events/EventCard/EventCard.tsx
+++ b/frontend/src/component/events/EventCard/EventCard.tsx
@@ -6,6 +6,7 @@ import { Link } from 'react-router-dom';
 import { styled } from '@mui/material';
 import type { EventSchema } from 'openapi';
 import { useUiFlag } from 'hooks/useUiFlag';
+import { useLocation } from 'react-router-dom';
 
 interface IEventCardProps {
     entry: EventSchema;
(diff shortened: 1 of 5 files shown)


frontend/src/component/events/EventCard/EventCard.tsx
```diff
@@ -74,17 +75,41 @@ export const StyledCodeSection = styled('div')(({ theme }) => ({
...
     const { locationSettings } = useLocationSettings();
     const eventGroupingEnabled = useUiFlag('eventGrouping');
+    const location = useLocation();
 
     const createdAtFormatted = formatDateYMDHMS(
         entry.createdAt,
         locationSettings.locale,
     );
 
+    const getGroupIdLink = () => {
```
sjaanus, line 85: After clicking on it, update query params manually.

frontend/src/component/events/EventLog/EventLogFilters.tsx
```diff
@@ -50,6 +57,22 @@ export const useEventLogFilters = (
                 value: env.name,
             })) ?? [];
 
+        const groupIdOptions =
```
sjaanus, line 60: Build the group id options based on what is in query params, so basically it should only show the one that is selected/queryparams.

kwasniew (approved):

PR #10291, opened 2025-07-03
feat: reset pnps feedback

sjaanus:
This PR resets pnps prompts by removing entries in the user_feedback table


kwasniew (approved):
are you sure we don't care about it?

sjaanus:
Closing as there is already automatic reset every 6 months.

PR #10295, opened 2025-07-03
fix: tear down also event handlers in addon service

sjaanus:
This is attempt to fix tests, as I caught this service failing in the stacktrace.

Basically the event handlers were not cleaned up during stop.

1. Test ends → stopUnleash() is called
2. services.addonService.destroy() called → but this only destroyed addon providers, not event handlers
3. await db.destroy() called → destroys database connection pool
4. Event comes in → triggers handleEvent → calls fetchAddonConfigs() → tries to query destroyed database
5. Database query fails with "Error: aborted"
6. Node.js catches this → db-migrate's fatal handler → process.exit(1)

### DIFF
--- a/src/lib/features/events/event-service.ts
+++ b/src/lib/features/events/event-service.ts
@@ -110,6 +110,13 @@ export default class EventService {
         return this.eventStore.on(eventName, listener);
     }
 
+    off(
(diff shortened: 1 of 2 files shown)


chriswk (approved):
LGTM

PR #10301, opened 2025-07-03
fix: frontend API CORS

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -103,10 +103,7 @@ export default async function getApp(
     if (config.enableOAS && services.openApiService) {
         services.openApiService.useDocs(app);
     }
-    // Support CORS preflight requests for the frontend endpoints.
-    // Preflight requests should not have Authorization headers,
-    // so this must be handled before the API token middleware.
-    app.options(
+    app.use(
         `${baseUriPath}/api/frontend*`,
         corsOriginMiddleware(services, config),
     );
--- a/src/lib/features/frontend-api/frontend-api-controller.ts
+++ b/src/lib/features/frontend-api/frontend-api-controller.ts
@@ -19,7 +19,6 @@ import {
 } from '../../openapi/index.js';
 import type { Context } from 'unleash-client';
 import { enrichContextWithIp } from './index.js';
-import { corsOriginMiddleware } from '../../middleware/index.js';
 import NotImplementedError from '../../error/not-implemented-error.js';
 import rateLimit from 'express-rate-limit';
 import { minutesToMilliseconds } from 'date-fns';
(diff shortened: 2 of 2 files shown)


src/lib/app.ts
```diff
@@ -103,10 +103,7 @@ export default async function getApp(
     if (config.enableOAS && services.openApiService) {
         services.openApiService.useDocs(app);
     }
-    // Support CORS preflight requests for the frontend endpoints.
-    // Preflight requests should not have Authorization headers,
-    // so this must be handled before the API token middleware.
-    app.options(
+    app.use(
```
kwasniew, line 106: will handle all HTTP methods

src/lib/features/frontend-api/frontend-api-controller.ts
```diff
@@ -65,10 +64,6 @@ export default class FrontendAPIController extends Controller {
                 functionName,
             });
 
-        // Support CORS requests for the frontend endpoints.
-        // Preflight requests are handled in `app.ts`.
-        this.app.use(corsOriginMiddleware(services, config));
```
kwasniew, line 70: when API token was removed and JS SDK was making GET flags it was not reaching this middleware because of the earlier auth middleware returning 401 without CORS headers

src/lib/app.ts
```diff
@@ -103,10 +103,7 @@ export default async function getApp(
     if (config.enableOAS && services.openApiService) {
         services.openApiService.useDocs(app);
     }
-    // Support CORS preflight requests for the frontend endpoints.
-    // Preflight requests should not have Authorization headers,
-    // so this must be handled before the API token middleware.
-    app.options(
```
kwasniew, line 109: not sure why we split OPTIONS from other HTTP methods in the original version. It's not a typical CORS handling pattern I am used to

nunogois (approved):
LGTM. 

Two points:
 - Should we do the same for `/api/streaming*`?
 - Can we feature flag this somehow?

kwasniew:
@nunogois express routes are startup time so you can't really change them at runtime 
I want to setup streaming in another PR

nunogois:
> @nunogois express routes are startup time so you can't really change them at runtime 

Yeah, it's tricky. Maybe the middleware could evaluate this flag internally? And if we wanted the old behavior we would early return with next() except if req method "OPTIONS".

> I want to setup streaming in another PR

Fair 👍

Tymek (approved):

kwasniew:
@nunogois I'm worried about the complexity of the runtime control for this introducing bugs. I'll carefully start rolling this out in small batches to see if everything works

PR #10302, opened 2025-07-03
Add "feature status will change"

thomasheartman:
Updates handling of default changes by letting the change components themselves handle adding necessary text instead of passing it down as "actions" (when text isn't really an action ...).

There's two cases where we add default changes to a CR:
1. If you enable an environment that has no strategies, then we add the default strategy.
2. If you disable or remove the last strategy on an active environment (such that the environment would now have 0 active strategies), then we disable the environment too and add that to the change request.

Here's what they look like in the new design:
image

image

And because it seemed the easiest way to do it, I've also updated the old design to do the same thing. It looks the same as it used to:
image

image


sjaanus (approved):
LG

PR #10310, opened 2025-07-04
feat(1-3873)/warn you when adding existing values

thomasheartman:
wip


frontend/src/component/feature/FeatureStrategy/FeatureStrategyConstraints/EditableConstraint/useEditableConstraint/useEditableConstraint.tsx
```diff
@@ -76,7 +77,30 @@ export const useEditableConstraint = (
         [JSON.stringify(context), localConstraint.contextName],
     );
 
-    const validator = constraintValidator(localConstraint.operator);
+    const baseValidator = constraintValidator(localConstraint.operator);
+
+    const validator = useCallback(
```
kwasniew: do we need this useCallback? It makes code more difficult to read with the extra array

thomasheartman: Probably fine. I'm a bit paranoid about this section because I've done some fairly hefty (and pretty effective (if I must say so myself)) caching to not improve the perf on large lists of legal values. However, the validator isn't used for legal values, so we're probably fine. I'll double check and remove if I don't see any negative side effects.

PR #10322, opened 2025-07-07
fix: audit scim user deleted events

gastonfournier:
SCIM users deleted in bulk are not captured in the event log. We just add an event like this:
image

This prevents partial user sync because we don't get an event when the user was deleted.

### DIFF
--- a/src/lib/services/user-service.ts
+++ b/src/lib/services/user-service.ts
@@ -403,14 +403,26 @@ class UserService {
     }
 
     async deleteScimUsers(auditUser: IAuditUser): Promise<void> {
-        await this.store.deleteScimUsers();
-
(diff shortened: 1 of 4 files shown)


src/lib/services/user-service.ts
```diff
@@ -403,14 +403,26 @@ class UserService {
...
+                    deletedUser: { ...user, rootRole: viewerRole.id },
+                    auditUser,
+                });
+            });
+            await this.eventService.storeEvents([
+                ...deletions,
+                new ScimUsersDeleted({
+                    data: null,
+                    auditUser,
+                }),
```
gastonfournier, line 423: If we have a single event per user, do we need this?

PR #10324, opened 2025-07-07
feat: move apply button to a separate section

daveleek:
This moves the apply button to a separate section (when request approvers is enabled), to prevent accidentally applied CRs stemming from requesting more than necessary approvers and the extra approver arrives at an approved and ready CR and then accidentally clicks apply.

We move the Changes overview StyledPaper into a StyledDiv so we can stack StyledPapers as rows inside the column

image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
@@ -546,11 +546,11 @@ export const ChangeRequestOverview: FC = () => {
                                         </StyledButtonContainer>
                                         <StyledBox>
                                             <StyledTypography>
-                                                Changes approved
+                                                Apply changes
                                             </StyledTypography>
                                             <Typography>
-                                                One approving review from
-                                                requested approvers
+                                                The change request has been
+                                                reviewed and approved
                                             </Typography>
                                         </StyledBox>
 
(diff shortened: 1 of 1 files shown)


frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
```diff
@@ -306,175 +338,244 @@ export const ChangeRequestOverview: FC = () => {
...
+                                            bgColor={
+                                                theme.palette.primary.main!
+                                            }
+                                            svgColor={
+                                                theme.palette.background.paper
+                                            }
+                                        >
+                                            <ApprovedIcon
+                                                style={{
+                                                    transform: `scale(1.5)`,
```
gastonfournier, line 562: :thinking: this doesn't look like a common practice in the UI.

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
```diff
@@ -306,175 +338,244 @@ export const ChangeRequestOverview: FC = () => {
...
+                        </StyledInnerContainer>
+                    </StyledPaper>
+                    <ConditionallyRender
+                        condition={
+                            changeRequest.state === 'Approved' &&
+                            approversEnabled
+                        }
+                        show={
+                            <StyledPaper sx={{ mt: 2 }} elevation={0}>
+                                <StyledInnerContainer sx={{ pt: 0, pb: 1.5 }}>
```
gastonfournier: Shouldn't these properties be added to the Styled component? Or maybe have a different component style?

gastonfournier (approved):
I mean, sure, makes sense to me, due to the refactor it's not easy to review, but looks correct to me

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
daveleek: It's taken directly from the other place the icon was in use, to look the same

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestOverview.tsx
daveleek: This is a local override of a reuse. But I'm thinking maybe I should just copy it and make changes directly on it instead. I'll update with that

PR #10325, opened 2025-07-07
chore: unknown flags with environment

nunogois:
https://linear.app/unleash/issue/2-3681/add-environment-to-unknown-flag-reports

This does a few things:
 - Adds environment to our unknown flag reports
 - Increases our max unknown flag reports from 10 to 100
 - Flattens our response (easier to get started with and display in a table, we can later decide if we want to group this data)

image

### DIFF
--- a/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
+++ b/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
@@ -244,14 +244,17 @@ export default class ClientMetricsServiceV2 {
             this.config.eventBus.emit(CLIENT_REGISTER, heartbeatEvent);
         }
 
+        const environment = value.environment ?? 'default';
+
         if (unknownToggleNames.length > 0) {
             const unknownFlags = unknownToggleNames.map((name) => ({
                 name,
                 appName: value.appName,
                 seenAt: value.bucket.stop,
+                environment,
             }));
             this.logger.info(
-                `Registering ${unknownFlags.length} unknown flags from ${value.appName}, i.e.: ${unknownFlags
+                `Registering ${unknownFlags.length} unknown flags from ${value.appName} in the ${environment} environment. Some of the unknown flag names include: ${unknownFlags
                     .slice(0, 10)
                     .map((f) => f.name)
                     .join(', ')}`,
@@ -264,7 +267,7 @@ export default class ClientMetricsServiceV2 {
                 (name) => ({
                     featureName: name,
                     appName: value.appName,
-                    environment: value.environment ?? 'default',
+                    environment,
                     timestamp: value.bucket.stop, //we might need to approximate between start/stop...
                     yes: value.bucket.toggles[name].yes ?? 0,
                     no: value.bucket.toggles[name].no ?? 0,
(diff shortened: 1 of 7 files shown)


src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -244,14 +244,17 @@ export default class ClientMetricsServiceV2 {
             this.config.eventBus.emit(CLIENT_REGISTER, heartbeatEvent);
         }
 
+        const environment = value.environment ?? 'default';
```
gastonfournier, line 238: I think we should use development after https://github.com/Unleash/unleash/pull/10080
```suggestion
        const environment = value.environment ?? 'development';
```
But maybe we should filter out toggles without an environment, as it shouldn't be the case that you evaluate a toggle without an environment...

src/lib/features/metrics/unknown-flags/unknown-flags-service.ts
```diff
@@ -80,24 +81,9 @@ export class UnknownFlagsService {
...
-            grouped.get(name)!.push({ appName, seenAt });
-        }
-
-        return Array.from(grouped.entries()).map(([name, reportedBy]) => ({
-            name,
-            reportedBy,
-        }));
+    async getAll(): Promise<UnknownFlag[]> {
+        if (!this.flagResolver.isEnabled('reportUnknownFlags')) return [];
+        return this.unknownFlagsStore.getAll();
```
gastonfournier: I think Ivar's idea here was to do something like `this.unknownFlagsStore.getAllAfter("7 days ago");` and keep all the flags in the database. Doing an insert and on conflict merge the date when it was seen.

That way we can keep the clean up process running once a day/week and deleting flags with more than 7 days

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
nunogois: Are we comfortable making this change here? I think it's been this way for a while now, and it feels out of scope to change it now.

https://github.com/Unleash/unleash/blob/068ef585be3d6914f049e9615c2c0e2b8a3e7dc3/src/lib/features/metrics/client-metrics/metrics-service-v2.ts#L267

src/lib/features/metrics/unknown-flags/unknown-flags-service.ts
nunogois: I think that would simplify our logic, but it feels worse in terms of being defensive:
 - Unknown flags always hit the DB right away instead of being buffered in-memory per pod and only flushed/merged in a scheduled manner
 - In those 7 days (or whatever cadence we decide) we can be quite destructive by spamming Unleash with a huge volume of unknown flags

But maybe this is how our metrics work currently and we're okay with it 🤷
Let's chat about it.

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
gastonfournier: Ok, I didn't realize it was something moved from another place.

src/lib/features/metrics/unknown-flags/unknown-flags-service.ts
gastonfournier: Yes, as we discussed, let's try to learn first, be less defensive and disable the flag if we see some issue, and then fix it

gastonfournier (approved):
Looks nice!

PR #10327, opened 2025-07-08
fix: make user creation transactional

gastonfournier:
## About the changes
When inserting a user with an invalid role id, the user creation will succeed but there will be no record in the audit log.

The API call returns a 400 misleading you to believe the user was not created, but it actually was.

This makes the whole user creation transactional, so if something fails, data will be in the right state.

### DIFF
--- a/src/lib/features/users/factory.ts
+++ b/src/lib/features/users/factory.ts
@@ -1,0 +1,55 @@
+import { ResetTokenStore } from '../../db/reset-token-store.js';
+import SettingStore from '../../db/setting-store.js';
+import {
+    createAccessService,
+    createEventsService,
+    type Db,
+    EmailService,
+    type IUnleashConfig,
+    ResetTokenService,
+    SessionService,
+    SessionStore,
+    SettingService,
+    UserService,
+} from '../../server-impl.js';
+import { UserStore } from './user-store.js';
+
+export const createUserService = (
+    db: Db,
+    config: IUnleashConfig,
+): UserService => {
+    const userStore = new UserStore(db, config.getLogger);
+    const resetTokenStore = new ResetTokenStore(
+        db,
+        config.eventBus,
+        config.getLogger,
+    );
+    const resetTokenService = new ResetTokenService(
+        { resetTokenStore },
+        config,
+    );
+    const eventService = createEventsService(db, config);
+    const sessionStore = new SessionStore(
+        db,
+        config.eventBus,
+        config.getLogger,
+    );
+    const sessionService = new SessionService({ sessionStore }, config);
+    const settingStore = new SettingStore(db, config.getLogger);
+    const settingService = new SettingService(
+        { settingStore },
+        config,
+        eventService,
+    );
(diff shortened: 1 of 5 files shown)


src/test/e2e/api/admin/user-admin.e2e.test.ts
```diff
@@ -32,463 +31,536 @@ let sessionStore: ISessionStore;
...
+            {
+                experimental: {
+                    flags: {
+                        strictSchemaValidation: true,
+                        showUserDeviceCount: true,
+                    },
+                },
+                rateLimiting: {
+                    createUserMaxPerMinute: 100,
+                },
```
gastonfournier, line 107: Some tests were failing with a rate limit error

src/test/e2e/api/admin/user-admin.e2e.test.ts
```diff
@@ -32,463 +31,536 @@ let sessionStore: ISessionStore;
...
+            },
+            db.rawDatabase,
```
gastonfournier, line 109: Sending the db, makes the test to actually use the template db.

src/test/e2e/api/admin/user-admin.e2e.test.ts
```diff
@@ -32,463 +31,536 @@ let sessionStore: ISessionStore;
...
+                },
+                experimental: {
+                    flags: {
+                        strictSchemaValidation: true,
+                        showUserDeviceCount: true,
+                    },
+                },
             },
-        },
+            db.rawDatabase,
```
gastonfournier, line 52: This ensures we use the  transactional service instead of the fake transactional one

src/lib/features/users/factory.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    return new UserService({ userStore }, config, {
+        accessService,
+        resetTokenService,
+        emailService,
+        eventService,
+        sessionService,
+        settingService,
+    });
+};
```
nunogois: Should we call this file `createUserService.ts` instead for consistency?

src/lib/services/index.ts
```diff
@@ -252,14 +248,18 @@ export const createServices = (
...
+        : withFakeTransactional(
+              new UserService(stores, config, {
+                  accessService,
+                  resetTokenService,
+                  emailService,
+                  eventService,
+                  sessionService,
+                  settingService,
+              }),
+          );
```
nunogois, line 262: I wonder if our createUserService/factory file should expose a `createFakeUserService` method instead of us doing it here.

nunogois (approved):
LGTM!

src/lib/features/users/factory.ts
gastonfournier: Yeah, that makes sense

src/lib/services/index.ts
gastonfournier: Yes, I'm not a big fan of fake services, so this way it remains hidden... and it depends on some fake services created here, it might be good to refactor it later.

PR #10335, opened 2025-07-09
chore: remove unused method

thomasheartman:
I found this method when running through the environment store that has 0 references. I also can't find any references to it in enterprise and it's not in the interface. I think it's safe to remove.


src/lib/features/project-environments/environment-store.ts
```diff
@@ -301,20 +301,6 @@ export default class EnvironmentStore implements IEnvironmentStore {
         return present;
     }
 
-    async getByName(name: string): Promise<IEnvironment> {
```
thomasheartman, line 304: This has no references anywhere as far as I could find 🤷

PR #10337, opened 2025-07-09
Impact metrics errors with rate per second option

Tymek:
## About the changes
- checkbox to select 'rate' vs 'increase' - always available for now, but does nothing for gauge. I can improve it later on
- better preview - it will show resolved query underneath
- cleaner error handling that doesn't overflow widgets


frontend/src/component/impact-metrics/ImpactMetricsChart.tsx
```diff
@@ -16,33 +16,38 @@ type ImpactMetricsChartProps = {
     selectedRange: 'hour' | 'day' | 'week' | 'month';
     selectedLabels: Record<string, string[]>;
     beginAtZero: boolean;
+    showRate?: boolean;
```
kwasniew, line 19: boolean is almost never a good modelling option. I'd go for a union type because the moment you get a third option it's trivial to extend.

PR #10348, opened 2025-07-11
chore: prevent unknown flag deadlocks by sorting and batching inserts

nunogois:
https://linear.app/unleash/issue/2-3692/prevent-deadlocks-by-sorting-and-batching-inserts

Tries to prevent deadlocks by sorting and batching inserts when flushing.

### DIFF
--- a/src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
+++ b/src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
@@ -1,6 +1,8 @@
 import type { Db } from '../../../db/db.js';
+import type { Logger, LogProvider } from '../../../logger.js';
 
 const TABLE = 'unknown_flags';
+const MAX_INSERT_BATCH_SIZE = 100;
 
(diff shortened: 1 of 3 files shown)


src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
```diff
@@ -28,22 +30,36 @@ export interface IUnknownFlagsStore {
...
+            try {
+                await this.db(TABLE)
+                    .insert(chunk)
+                    .onConflict(['name', 'app_name', 'environment'])
+                    .merge(['seen_at']);
+            } catch (error) {
+                this.logger.debug(
+                    `unknown_flags: batch ${i / MAX_INSERT_BATCH_SIZE + 1} failed and was skipped.`,
+                    error,
+                );
```
nunogois, line 61: We don't really care much here, we're okay with losing a few unknown flag reports since we're flushing every 2 minutes.

gastonfournier (approved):

PR #10350, opened 2025-07-14
feat: prefix unleash impact metrics labels

kwasniew:
## About the changes

So far we've been using only 2 labels: appName and environment. However with dynamic flag names we don't have a way to include only relevant labels and filter out infra level labels. That's why this PR prefixes all impact metrics domain labels with unleash_ so that we can filter those out. So we will have e.g. unleash_counter_server_errors{unleash_env=prod, unleash_flag1=enabled, unleash_flag2=variantA, unleash_flag3=disabled}


sighphyre (approved):
LGTM

PR #10355, opened 2025-07-14
feat: Using impact metrics with flags

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -1145,6 +1145,7 @@ export function registerPrometheusMetrics(
                 config.flagResolver.impactMetrics?.updateGauge(
                     HEAP_MEMORY_TOTAL,
                     process.memoryUsage().heapUsed,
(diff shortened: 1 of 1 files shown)


src/lib/metrics.ts
```diff
@@ -792,7 +792,11 @@ export function registerPrometheusMetrics(
                     appName,
                 })
                 .observe(time);
-            config.flagResolver.impactMetrics?.incrementCounter(REQUEST_COUNT);
+            config.flagResolver.impactMetrics?.incrementCounter(
+                REQUEST_COUNT,
+                1,
+                { flagNames: ['consumptionModel'], context: {} },
```
kwasniew, line 798: This API is matching isEnabled signature in flag resolver. However it's not very ergonomic since we usually provide static context when SDK is instantiated and never override it. We only need flag names here. I will adjust it in our flag resolver wrapper later once I learn about the metrics+flags correlation.

sighphyre (approved):
LG

PR #10357, opened 2025-07-15
fix: don't overwrite `seenClients` in instance-service; merge if same client appears again

thomasheartman:
Fixes a bug where `registerInstance` and `register{Frontend|Backend}Client` would overwrite each other's data in the instance service, leading to the bulk update being made with partial data, often missing SDK version. There's a different issue in the actual store that causes sdk version and type to be overwritten when it's updated (because we don't use `setLastSeen` anymore), but I'll handle that in a different PR.

This PR adds tests for the changes I've made. Additionally, I've made these semi-related bonus changes:
- In registerInstance, don't expect a partial `IClientApp`. We used to validate that it was actual a metrics object instead. Instead, update the signature to expect the actual properties we need from the cilent metrics schema and set a default for instanceId the way Joi did.
- In `metrics.ts`, use the `ClientMetricsSchema` type in the function signature, so that the request body is correctly typed in the function (instead of being `any`).
- Delete two unused properties from the`createApplicationSchema`. They would get ignored and were never used as far as I can tell. (`appName` is taken from the URL, and applications don't store `sdkVersion` information).
- Add `sdkVersion` to `IClientApp` because it's used in instance service.

I've been very confused about all the weird type shenanigans we do in the instance service (expecting `IClientApp`, then validating with a different Joi schema etc). I think this makes it a little bit better and updates the bits I'm touching, but I'm happy to take input if you disagree.

### DIFF
--- a/src/lib/features/metrics/instance/instance-service.test.ts
+++ b/src/lib/features/metrics/instance/instance-service.test.ts
@@ -263,3 +263,129 @@ test('filter out private projects from overview', async () => {
         featureCount: 0,
     });
 });
+
+test('`registerInstance` sets `instanceId` to `default` if it is not provided', async () => {
+    const instanceService = new ClientInstanceService(
+        {} as any,
+        config,
+        {} as any,
+    );
+
+    await instanceService.registerInstance(
+        {
+            appName: 'appName',
+            environment: '',
+        },
+        '::1',
+    );
+
+    expect(instanceService.seenClients.appName_default).toMatchObject({
+        appName: 'appName',
+        instanceId: 'default',
+    });
+});
+
+describe('upserting into `seenClients`', () => {
+    test('registerInstance merges its data', async () => {
+        const instanceService = new ClientInstanceService(
+            {} as any,
+            config,
+            {} as any,
+        );
+
+        const client = {
+            appName: 'appName',
+            instanceId: 'instanceId',
+        };
+
+        const key = instanceService.clientKey(client);
+
+        instanceService.seenClients = {
+            [key]: { ...client, sdkVersion: 'my-sdk' },
+        };
+
+        await instanceService.registerInstance(
+            {
+                ...client,
+                environment: 'blue',
+            },
+            '::1',
+        );
+
+        expect(instanceService.seenClients[key]).toMatchObject({
+            appName: 'appName',
+            instanceId: 'instanceId',
+            environment: 'blue',
+            sdkVersion: 'my-sdk',
+        });
+    });
+    test('registerBackendClient merges its data', async () => {
+        const instanceService = new ClientInstanceService(
+            {} as any,
+            config,
+            {} as any,
+        );
+
+        const client = {
+            appName: 'appName',
+            instanceId: 'instanceId',
+        };
+
+        const key = instanceService.clientKey(client);
+
+        instanceService.seenClients = {
+            [key]: { ...client, environment: 'blue' },
+        };
+
+        await instanceService.registerBackendClient(
+            {
+                ...client,
+                sdkVersion: 'my-sdk',
+                started: new Date(),
+                interval: 5,
+            },
+            '::1',
+        );
+
+        expect(instanceService.seenClients[key]).toMatchObject({
+            appName: 'appName',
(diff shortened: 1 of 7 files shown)


src/lib/features/metrics/instance/instance-service.test.ts
```diff
@@ -263,3 +263,129 @@ test('filter out private projects from overview', async () => {
...
+
+        const client = {
+            appName: 'appName',
+            instanceId: 'instanceId',
+        };
+
+        const key = instanceService.clientKey(client);
+
+        instanceService.seenClients = {
+            [key]: { ...client, sdkVersion: 'my-sdk' },
```
kwasniew, line 304: why do you prefer to set it using a backdoor mechanism instead of doing a preliminary registration with a public API?

thomasheartman: Probably just because this struck me as easier / quicker. I just needed to prepopulate it with something it wouldn't overwrite. I'll switch it up if you want.

src/lib/features/metrics/instance/instance-service.ts
```diff
@@ -99,24 +98,33 @@ export default class ClientInstanceService {
...
+
     public async registerInstance(
-        data: PartialSome<IClientApp, 'instanceId'>,
+        data: Pick<
+            ClientMetricsSchema,
+            'appName' | 'instanceId' | 'environment'
+        >,
         clientIp: string,
     ): Promise<void> {
-        const value = await clientMetricsSchema.validateAsync(data);
```
kwasniew, line 106: is this now done by openapi?

src/lib/features/metrics/instance/metrics.ts
```diff
@@ -169,7 +176,9 @@ export default class ClientMetricsController extends Controller {
                     this.flagResolver.isEnabled('impactMetrics') &&
                     impactMetrics
                 ) {
-                    await this.metricsV2.registerImpactMetrics(impactMetrics);
+                    await this.metricsV2.registerImpactMetrics(
+                        impactMetrics as Metric[],
```
kwasniew, line 180: is it possible to do it without type assertion or do we need to help the compiler

src/lib/features/metrics/instance/instance-service.ts
thomasheartman: Partially. 

- OpenAPI guarantees that `instanceId` and `appName` are there. These are the bits we care the most about. 
- There's also `environment` which is optional in OpenAPI, but which clientMetricsSchema.validate required. This might be useful if it's not already present, but it'll show up when you register the client anyway. Besides, I don't know when it wouldn't be there. Finally, even if it's not there, we can manually default it to `default`, which is what the validation method did. (Although I'm leaning towards removing it in a follow-up)
- The real reason I removed it is that that schema validation also requires your input to have `buckets: {...}` defined, which a) we don't use, and b) is a pain to set up during testing.

So yes. OpenAPI takes care of what we need it to.

src/lib/features/metrics/instance/metrics.ts
thomasheartman: Not as it is right now. Or at least, my compiler was nagging. I think it's because `impactMetrics` isn't defined on that type yet? But that'll be up to what you're working on right now, so I didn't wanna go in and start messing around there 🤷🏼

src/lib/features/metrics/instance/instance-service.test.ts
thomasheartman: Oh, and because Joi will throw an error here if you don't provide it with all the properties it expects, such as "started", and "interval" etc.

thomasheartman: And because we need to directly inspect the internal property anyway for this test to make sense, I'd like to leave it as is.

src/lib/features/metrics/instance/metrics.ts
thomasheartman: Yeah, this is because the schema doesn't have `impactMetrics`, I suppose. If you think it should, then you should update the schema, but I'm guessing that'll happen at the end of the impact metrics project?

PR #10360, opened 2025-07-16
fix(1-3928): prevent overwriting existing values in instance store

thomasheartman:
Fixes a bug in the instance store where insert and bulkUpsert would overwrite existing properties if there was a row there already. Now it'll ignore any properties that are undefined.

The implementation is lifted directly from `src/lib/db/client-applications-store.ts` (line 107 atm).

Additionally, I've renamed the `insert` method to `upsert` to make it clearer what it does (and because we already have `bulkUpsert`). The method seems to only be used in tests, anyway. I do not anticipate any changes to be required in enterprise (I've checked).

## Discussion points:

This implementation uses `delete` to remove properties from the object. Why didn't I do it some other way? Two main reasons:
1. We've had this implementation for 4 years in the client applications store. If there were serious issues with it, we'd probably know by know. (Probably.)
2. The only way I can think of without deleting, would be to use `Object.fromEntries` and `Object.toEntries` and either map or reduce. That'll double the amount of property iterations we'll need to do.

So naively, this strikes me as being more efficient. If you know better solutions, I will of course be happy to take them. If not, I'd like to leave this as is and then change it if we see that it's causing issues.

### DIFF
--- a/src/lib/db/client-instance-store.ts
+++ b/src/lib/db/client-instance-store.ts
@@ -33,7 +33,7 @@ const mapRow = (row): IClientInstance => ({
 });
 
 const mapToDb = (client: INewClientInstance) => {
-    const mapped = {
+    const temp = {
         app_name: client.appName,
         instance_id: client.instanceId,
         sdk_version: client.sdkVersion,
@@ -43,14 +43,14 @@ const mapToDb = (client: INewClientInstance) => {
         environment: client.environment,
     };
 
-    Object.keys(mapped).forEach((k) => {
-        if (mapped[k] === undefined) {
-            // not using !temp[k] to allow false and null values to get through
-            delete mapped[k];
+    const result = {};
+    for (const [key, value] of Object.entries(temp)) {
+        if (value !== undefined) {
+            result[key] = value;
         }
-    });
+    }
 
-    return mapped;
+    return result;
 };
 
 export default class ClientInstanceStore implements IClientInstanceStore {


src/lib/db/client-instance-store.ts
```diff
@@ -21,25 +21,37 @@ const COLUMNS = [
 ];
 const TABLE = 'client_instances';
 
-const mapRow = (row) => ({
+const mapRow = (row): IClientInstance => ({
     appName: row.app_name,
     instanceId: row.instance_id,
     sdkVersion: row.sdk_version,
+    sdkType: row.sdk_type,
```
thomasheartman, line 28: I'm pretty sure this is a safe change. As far as I can tell, this type is only used directly in the instance-service, which always remaps properties before propagating it. I do not think we'll suddenly start exposing the sdkType property through an API. (and even if we did, this ain't exactly the client/features api)

src/lib/db/client-instance-store.ts
```diff
@@ -21,25 +21,37 @@ const COLUMNS = [
...
+        last_seen: client.lastSeen || 'now()',
+        environment: client.environment,
+    };
+
+    Object.keys(mapped).forEach((k) => {
+        if (mapped[k] === undefined) {
+            // not using !temp[k] to allow false and null values to get through
+            delete mapped[k];
+        }
+    });
```
thomasheartman: This is the bit of lifted code that makes everything just work

src/test/e2e/stores/client-instance-store.e2e.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import getLogger from '../../fixtures/no-logger.js';
+import type {
+    IClientInstanceStore,
+    IUnleashStores,
+} from '../../../lib/types/index.js';
+import type { INewClientInstance } from '../../../lib/types/stores/client-instance-store.js';
+
+let db: ITestDb;
+let stores: IUnleashStores;
+let clientInstanceStore: IClientInstanceStore;
```
thomasheartman, line 12: The tests are based on (and placed based on) `src/test/e2e/stores/client-application-store.e2e.test.ts`. I don't think this is the right time to start moving this store into a feature-based architecture.

src/lib/db/client-instance-store.ts
kwasniew: JS engines don't like delete. I'd rather do `Object.fromEntries(
    Object.entries(mapped).filter(([_, v]) => v !== undefined)
);`

thomasheartman: Handled in 840849a70bcfc92c2f0d136fefa8e4bb8df46000

PR #10369, opened 2025-07-17
docs: prioritize search results

melindafekete:
**What**
This PR introduces a system to prioritize search results within the docs, improving the search experience.

**Why?**
Currently, all pages hold equal weight, which can cause specific, low-level API endpoints to rank higher than our best pages for conceptual overviews or important reference guides. This change implements a priority system to ensure that key pages are more visible in search results, while still leveraging Algolia's powerful relevance algorithm. In addition, deprecated pages should not show up in search.

**Details**

- **Algolia crawler config**: The hosted crawler's configuration has been modified to extract a numerical priority attribute from a <meta name="search_priority" ...> tag present in the page's HTML. This attribute has been added to our index's customRanking settings, causing pages with a higher priority value to rank higher.

- **SearchPriority component**: `src/components/SearchPriority.jsx` provides an API  to assign a search priority or exclude a page from search by adding a single line to an .mdx file. This component is responsible for rendering the correct <meta> tag that the crawler reads.

This new system is documented in the docs README.md.

This PR also removes the deprecated Slack integration documentation.


nunogois (approved):
Great job! 🎖️

PR #10375, opened 2025-07-21
chore: sift metrics on both endpoints

nunogois:
https://linear.app/unleash/issue/2-3696/report-unknown-flags-when-sent-to-the-bulk-metrics-endpoint

Unifies metrics sifting logic across both metrics endpoints:
- `/metrics`
- `/metrics/bulk`

This PR improves consistency between the `/metrics` and `/metrics/bulk` endpoints by introducing a shared `siftMetrics` method, now used within `registerBulkMetrics`. Both endpoints already call this method at the end of their respective logic flows, ensuring that metrics are sifted in the same way regardless of the path taken.

While the primary goal was to enable reporting of unknown flags via the `/metrics/bulk` endpoint, this change also improves bulk processing by consistently dropping invalid or unknown flags before insertion, just like in the regular `/metrics` endpoint.

### DIFF
--- a/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
+++ b/src/lib/features/metrics/client-metrics/metrics-service-v2.ts
@@ -187,9 +187,12 @@ export default class ClientMetricsServiceV2 {
         const metricsByToggle = new Map<string, IClientMetricsEnv[]>();
         for (const m of metrics) {
             if (m.yes === 0 && m.no === 0) continue;
-            const arr = metricsByToggle.get(m.featureName) ?? [];
+            let arr = metricsByToggle.get(m.featureName);
+            if (!arr) {
+                arr = [];
+                metricsByToggle.set(m.featureName, arr);
+            }
             arr.push(m);
-            metricsByToggle.set(m.featureName, arr);
         }
         if (metricsByToggle.size === 0) return [];
 


src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -178,12 +179,75 @@ export default class ClientMetricsServiceV2 {
...
+
+        const { validatedToggleNames, unknownToggleNames } =
+            await this.filterExistingToggleNames(toggleNames);
+
+        const validatedSet = new Set(validatedToggleNames);
+        const unknownSet = new Set(unknownToggleNames);
+
+        const invalidCount = toggleNames.length - validatedSet.size;
+        this.logger.debug(
+            `Got ${toggleNames.length} metrics (${invalidCount > 0 ? `${invalidCount} invalid` : 'all valid'}).`,
```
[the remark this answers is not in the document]
nunogois: Nah.

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -178,12 +179,75 @@ export default class ClientMetricsServiceV2 {
...
+        }
+        if (metricsByToggle.size === 0) return [];
+
+        const toggleNames = Array.from(metricsByToggle.keys());
+
+        const { validatedToggleNames, unknownToggleNames } =
+            await this.filterExistingToggleNames(toggleNames);
+
+        const validatedSet = new Set(validatedToggleNames);
+        const unknownSet = new Set(unknownToggleNames);
```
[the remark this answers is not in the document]
nunogois: Out of scope, maybe in a future PR.

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -178,12 +179,75 @@ export default class ClientMetricsServiceV2 {
...
+
+        const invalidCount = toggleNames.length - validatedSet.size;
```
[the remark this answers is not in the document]
nunogois: Both should already be deduped at this point, so we're comparing deduped with deduped, which is fine.

src/lib/features/metrics/client-metrics/metrics-service-v2.test.ts
```diff
@@ -113,7 +113,7 @@ test('process metrics properly even when some names are not url friendly, filter
     );
 
     // only toggle with a bad name gets filtered out
-    expect(eventBus.emit).not.toHaveBeenCalled();
+    expect(eventBus.emit).toHaveBeenCalledTimes(1);
```
gastonfournier, line 116: Is this with some event for unknown flags?

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -178,12 +179,75 @@ export default class ClientMetricsServiceV2 {
...
+    ): Promise<IClientMetricsEnv[]> {
+        if (!metrics.length) return [];
+
+        const metricsByToggle = new Map<string, IClientMetricsEnv[]>();
+        for (const m of metrics) {
+            if (m.yes === 0 && m.no === 0) continue;
+            const arr = metricsByToggle.get(m.featureName) ?? [];
+            arr.push(m);
+            metricsByToggle.set(m.featureName, arr);
+        }
```
gastonfournier:
```ts
for (const m of metrics) {
    if (m.yes === 0 && m.no === 0) continue;
    let arr = metricsByToggle.get(m.featureName);
    if (!arr) {
        arr = [];
        metricsByToggle.set(m.featureName, arr);
    }
    arr.push(m);
}
```
Micro-optimization, I think the cost might be negligible. Just sharing what I considered but I also think this is fine

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -178,12 +179,75 @@ export default class ClientMetricsServiceV2 {
...
+            this.unknownFlagsService.register(unknownFlags);
+        }
+
+        const siftedMetrics: IClientMetricsEnv[] = [];
+        for (const [featureName, group] of metricsByToggle) {
+            if (validatedSet.has(featureName)) {
+                siftedMetrics.push(...group);
+            }
+        }
+        return siftedMetrics;
```
gastonfournier, line 242: I stopped a second to consider the memory footprint of this, because we ingest metrics and generate 2 new arrays out of those metrics (the arrays are mutually exclusive). But I couldn't find an easy workaround... so this seems to be optimal already

src/lib/features/metrics/client-metrics/metrics-service-v2.test.ts
nunogois, line 116: We used to only emit `CLIENT_METRICS` if we had at least one valid flag name in our `/metrics`. However this is not consistent with the `/metrics/bulk` endpoint, which emits it anyways. 

This means that we need to make a choice here if we want to be consistent:
 - Always emit it
 - Only emit it after sifting the metrics, so we can check that we have at least one valid metric

Option 1 seemed easier.

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
```diff
@@ -178,12 +179,75 @@ export default class ClientMetricsServiceV2 {
         return toggleNames;
     }
 
+    private async siftMetrics(
```
nunogois, line 182: This method essentially extracts what we were already doing for filtering/validating in our `registerClientMetrics` method, makes it work for `IClientMetricsEnv[]` instead and optimizes it.

src/lib/features/metrics/client-metrics/metrics-service-v2.test.ts
nunogois: No, this is the `CLIENT_METRICS` event. Since we no longer do our sifting _before_ emitting the event, only after, this now always gets emitted. See: https://github.com/Unleash/unleash/pull/10375#discussion_r2218816656

gastonfournier: But I'd argue we don't want to process additional events if there's nothing to do with it... the event handler would later discard the event, so I think the sooner we discard it the better (unless there's some reason behind emitting the event)

src/lib/features/metrics/client-metrics/metrics-service-v2.ts
nunogois: Thanks, addressed in https://github.com/Unleash/unleash/pull/10375/commits/28060a37c646675be5c9159a1997b4123af2cb05

src/lib/features/metrics/client-metrics/metrics-service-v2.test.ts
nunogois: Are you suggesting option 2 would be preferable? Mind if I leave it for a follow up PR?

nunogois: Just to be clear, these are the 2 places where we emit it:

### /metrics

https://github.com/Unleash/unleash/blob/c1df04548dd8742804d50efe3f8038a2423128cf/src/lib/features/metrics/client-metrics/metrics-service-v2.ts#L270

### /metrics/bulk

https://github.com/Unleash/unleash/blob/c1df04548dd8742804d50efe3f8038a2423128cf/src/lib/features/metrics/instance/metrics.ts#L279

If we want to go with option 2, my suggestion would be to remove these 2 calls and instead emit only after sifting, in `registerBulkMetrics`.

gastonfournier: It doesn't hurt to do it later.

> If we want to go with option 2, my suggestion would be to remove these 2 calls and instead emit only after sifting, in registerBulkMetrics.

I think it's a good suggestion. But we should try to understand why we emit this event and whether it's safe to be removed when there's no data to process. From the top of my head I can't find a good reason for emitting it...

nunogois: This is the only listener I found:

https://github.com/Unleash/unleash/blob/c1df04548dd8742804d50efe3f8038a2423128cf/src/lib/metrics.ts#L1083

It's used to increment our `featureFlagUsageTotal` (`feature_toggle_usage_total`) prom counter based on the metrics we're sending along with the event. 

I assume it's fine (and an optimization, even) to only increment these counters for actual, valid, real flags, instead of all flags that we're receiving. So I'm still leaning towards my suggestion. 

WDYT?

gastonfournier: > I assume it's fine (and an optimization, even) to only increment these counters for actual, valid, real flags, instead of all flags that we're receiving. 

Yes, I think so.

> So I'm still leaning towards my suggestion.

Which was instead emit only after sifting, in registerBulkMetrics? I think to emit after sifting only if there's something left, makes sense

nunogois: Yes, exactly! I'll link the PR here once it's ready, for reference.

nunogois: Here it is: https://github.com/Unleash/unleash/pull/10376

PR #10376, opened 2025-07-21
chore: emit CLIENT_METRICS event after sifting

nunogois:
https://linear.app/unleash/issue/2-3705/only-emit-client-metrics-after-sifting-metrics

Only emits the `CLIENT_METRICS` event after metric sifting.

This ensures the event is emitted only for valid metrics tied to known flags, instead of all flags included in the metrics payload.

See: https://github.com/Unleash/unleash/pull/10375#discussion_r2218974109


gastonfournier (approved):
LGTM!

PR #10388, opened 2025-07-22
Show change request stage timestamps in UI

thomasheartman:
Adds a timestamp for each state we have time (and that isn't a state downstream from the current state) in the CR timeline.

image

The timestamp respects the user's preferred locale and uses the `time` element.

I've used the current name of the API payload's timestamps as it stands on the enterprise main branch for now. This name is not set in any schemas yet, so it is likely to change. Because it's not currently exposed to any users, that will not cause any issues. Name suggestions are welcome if you have them.

We only show timestamps for states up to and including the current state. Timestamps for downstream states (such as "approved" if you're in the "in review" state), will not be shown, even if they exist in the payload. (There are some decisions to make on whether to include these in the payload at all or not.)

There's no flags in this PR. They're not necessary If the API payload doesn't contain the timestamp, we just don't render the timestamp, and the timeline looks the way it always did:
image

## Bonus work

In the `ChangeRequestTimeline.tsx` file, I've made a few extra changes:
- `createTimelineItem` and `createScheduledTimelineItem` have become normal React components (`TimelineItem` and `ScheduledTimelineItem`) and are being called as such (in accordance with [React recommendations](https://react.dev/reference/rules/react-calls-components-and-hooks#never-call-component-functions-directly)).
- I've updated the subtitles for schedules to also use the time element, to improve HTML structure.

## Outstanding work

There's a few things that still need to be sorted out (primarily with UX). Mainly about how we handle scheduled items, which already have time information. For now, it looks like this:

image

image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestTimeline/ChangeRequestTimeline.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestTimeline/ChangeRequestTimeline.tsx
@@ -13,8 +13,7 @@ import type {
 import { HtmlTooltip } from '../../../common/HtmlTooltip/HtmlTooltip.tsx';
 import ErrorIcon from '@mui/icons-material/Error';
 import { useLocationSettings } from 'hooks/useLocationSettings';
-import { formatDateYMDHM } from 'utils/formatDate';
-import type { ReactNode } from 'react-markdown/lib/react-markdown';
+import { formatDateYMDHM } from 'utils/formatDate.ts';
 
 export type ISuggestChangeTimelineProps = {
     timestamps?: Record<ChangeRequestState, string>;
@@ -111,9 +110,10 @@ export const ChangeRequestTimeline: FC<ISuggestChangeTimelineProps> = ({
             <StyledBox>
                 <StyledTimeline>
                     {data.map((title, index) => {
-                        const timestampComponent = timestamps?.[title] ? (
-                            <Timestamp timestamp={timestamps?.[title]} />
-                        ) : undefined;
+                        const timestampComponent =
+                            index <= activeIndex && timestamps?.[title] ? (
+                                <Timestamp timestamp={timestamps?.[title]} />
+                            ) : undefined;
 
                         if (schedule && title === 'Scheduled') {
                             return (
--- a/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestTimeline/ChangeRequestTimeline.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestTimeline/ChangeRequestTimeline.tsx
@@ -9,6 +9,7 @@ import TimelineContent from '@mui/lab/TimelineContent';
 import type {
     ChangeRequestSchedule,
     ChangeRequestState,
+    ChangeRequestType,
 } from '../../changeRequest.types';
 import { HtmlTooltip } from '../../../common/HtmlTooltip/HtmlTooltip.tsx';
 import ErrorIcon from '@mui/icons-material/Error';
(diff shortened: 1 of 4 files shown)


frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestTimeline/ChangeRequestTimeline.tsx
```diff
@@ -102,17 +109,24 @@ export const ChangeRequestTimeline: FC<ISuggestChangeTimelineProps> = ({
...
 
     return (
         <StyledPaper elevation={0}>
             <StyledBox>
                 <StyledTimeline>
                     {data.map((title, index) => {
+                        const timestampComponent =
+                            index <= activeIndex && timestamps?.[title] ? (
+                                <Timestamp timestamp={timestamps?.[title]} />
+                            ) : undefined;
```
thomasheartman, line 121: This is a stopgap solution: only show timestamps up to and including the current state. In other words, don't show timestamps for states after the current. 

The prime example as to why we want this is where you make changes to a CR after getting it approved. That will reset the CR to in review, but we'll still have a timestamp for when you entered the approved state. 

This is probably something we want to filter out on the back end, but for now, this takes care of it.

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestTimeline/ChangeRequestTimeline.tsx
```diff
@@ -132,11 +146,17 @@ export const ChangeRequestTimeline: FC<ISuggestChangeTimelineProps> = ({
                             timelineDotProps = { variant: 'outlined' };
                         }
 
-                        return createTimelineItem(
-                            color,
-                            title,
-                            index < data.length - 1,
-                            timelineDotProps,
+                        return (
+                            <TimelineItem
```
thomasheartman, line 150: I've changed TimelineScheduleItem and TimelineItem to be react components instead of functions. This, as far as I understand, is in line with React best practices ([never call component functions directly](https://react.dev/reference/rules/react-calls-components-and-hooks#never-call-component-functions-directly))

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestTimeline/ChangeRequestTimeline.tsx
```diff
@@ -102,17 +109,24 @@ export const ChangeRequestTimeline: FC<ISuggestChangeTimelineProps> = ({
...
 
     return (
         <StyledPaper elevation={0}>
             <StyledBox>
                 <StyledTimeline>
                     {data.map((title, index) => {
+                        const timestampComponent =
+                            index <= activeIndex && timestamps?.[title] ? (
+                                <Time dateTime={timestamps?.[title]} />
+                            ) : undefined;
```
thomasheartman, line 121: This is a stopgap solution: only show timestamps up to and including the current state. In other words, don't show timestamps for states after the current.

The prime example as to why we want this is where you make changes to a CR after getting it approved. That will reset the CR to in review, but we'll still have a timestamp for when you entered the approved state.

This is probably something we want to filter out on the back end, but for now, this takes care of it.

kwasniew: so maybe it's worth extracting into intention revealing name for the condition. Or to leave a comment. Your inline github comment will not be visible in code after merging this PR

kwasniew (approved):
quality change

frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestTimeline/ChangeRequestTimeline.tsx
thomasheartman: Good call. Added a comment for now.

PR #10389, opened 2025-07-22
chore: remove unkonwn flags

gastonfournier:
We're iterating over the keys of experimental and checking if they're enabled: https://github.com/Unleash/unleash/blob/e140ab63e1a730775f589ad1d07e70a6a9b6a3bc/src/lib/util/flag-resolver.ts#L26 which leads to this big list of unknown flags

This comes from our report:
image

### DIFF
--- a/src/lib/types/experimental.ts
+++ b/src/lib/types/experimental.ts
@@ -9,7 +9,6 @@ export type IFlagKey =
     | 'anonymiseEventLog'
     | 'encryptEmails'
     | 'enableLicense'
-    | 'enableLicenseChecker'
     | 'responseTimeWithAppNameKillSwitch'
     | 'maintenanceMode'
     | 'messageBanner'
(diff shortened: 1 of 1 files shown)


gastonfournier:
> LGTM.
> 
> Do you know why we're doing this?

Maintenance and because we've noticed these flags existing in the unknown flags and they're not used in our codebase, so I think this is a valid use case for our unknown flags feature, which helps us optimize the payload size of our ui-config response

nunogois:
> > LGTM.
> > Do you know why we're doing this?
> 
> Maintenance and because we've noticed these flags existing in the unknown flags and they're not used in our codebase, so I think this is a valid use case for our unknown flags feature, which helps us optimize the payload size of our ui-config response

Oh I agree completely, what I meant is: Why are we evaluating every flag / calling getAll? Just curious. I can also look into it on my own sometime later, no worries.

gastonfournier:
> > > LGTM.
> > > Do you know why we're doing this?
> > 
> > 
> > Maintenance and because we've noticed these flags existing in the unknown flags and they're not used in our codebase, so I think this is a valid use case for our unknown flags feature, which helps us optimize the payload size of our ui-config response
> 
> Oh I agree completely, what I meant is: Why are we evaluating every flag / calling getAll? Just curious. I can also look into it on my own sometime later, no worries.

It's for the ui-config. Reason why we should migrate to frontend API instead. This was our poorsman solution to not having frontend API 3y ago...

PR #10393, opened 2025-07-22
Fix each child in a list should have a key error.

thomasheartman:
The console was complaining. I suspect it was because of the wrapping fragment. So instead of doing everything within react, I switched to using a standard case statement.

Also: because name is optional and not guaranteed to be unique, let's use id for the key instead.


frontend/src/component/changeRequest/ChangeRequestOverview/ChangeRequestRequestedApprovers/ChangeRequestRequestedApprovers.tsx
```diff
@@ -350,31 +350,21 @@ export const ChangeRequestRequestedApprovers: FC<{
...
-                        <ChangeRequestPending
-                            key={reviewer.name}
-                            name={reviewer.name || 'Unknown user'}
-                            imageUrl={reviewer.imageUrl}
-                        />
-                    )}
-                </>
-            ))}
+            {reviewers.map((reviewer) => {
+                const key = reviewer.name || reviewer.id;
```
kwasniew: is name unique?

PR #10399, opened 2025-07-23
Add timestampsInChangeRequestTimeline flag

thomasheartman:
Adds flag to OSS for the new timestamps in CR timeline capability.

We might not need them in the UI, but might as well add it in case.


kwasniew (approved):
I usually try to keep flag names shorter :)

PR #10413, opened 2025-07-24
feat: series query warning

Tymek:
## About the changes
Warnings and limit when trying to show high cardinality data

image


frontend/src/hooks/api/getters/useImpactMetricsData/useImpactMetricsData.ts
```diff
@@ -10,6 +10,7 @@ export type ImpactMetricsSeries = {
     data: TimeSeriesData;
 };
 
+// TODO: use OpenAPI types
```
thomasheartman: Can we tag this with a flag or something so that we remember to fix this at some point? 

Something like this, just with the right flag name?
```suggestion
// TODO(impactMetrics): use OpenAPI types
```

thomasheartman (approved):

PR #10421, opened 2025-07-29
feat: Timeline for cancelled CRs

thomasheartman:
Shows the cancelled state in the timeline for cancelled CRs. Extracts steps into a separate file.

Also, if timestamps are present, it will dynamically show the steps we have times for for rejected and cancelled CRs.

If we do not have timestamps, it'll use the old behavior for displaying rejected steps, but does add a new one for cancelled steps, which includes 'Draft', 'In review', 'Approved', 'Cancelled'. 

CRs can be cancelled after:
- In review
- Approved

CRs can be rejected after:
- In review
- Schedule

For other states, use the existing steps. Cancelled and Rejected are both terminal states, so there's no future steps to show.

Regardless of what we have timestamps for, always show 'Draft', 'In review' and the final state (Rejected | Cancelled). 

## Gallery

Cancellations
image

image

Rejections 
image

image

New cancelled steps without timestamps:
image


nunogois (approved):
Nice

PR #10436, opened 2025-07-29
Fix: Horizontal hyperscroll on project flag list

Tymek:
## About the changes
Sometimes, the horizontal scroll bar for flags was showing scroll way past where it should.

Internal ticket details: https://linear.app/unleash/issue/1-3779/horizontal-hyperscroll-on-project-flag-list


frontend/src/component/common/Table/SortableTableHeader/CellSortable/CellSortable.styles.ts
```diff
@@ -54,6 +54,7 @@ export const StyledButton = styled('button', {
     ...(isSorted && {
         fontWeight: theme.typography.fontWeightBold,
     }),
+    contain: 'paint',
```
Tymek, line 57: That line.
https://developer.mozilla.org/en-US/docs/Web/CSS/contain

Everything else is a refactor I did when familiarizing myself with this part of the code.

PR #10439, opened 2025-07-30
fix: stickiness should be preserved on strategy updates

gastonfournier:
## About the changes
When stickiness was set to empty or undefined while updating a strategy via API the stickiness would be lost.

This adds a validation step after creating a strategy, that updating with the same data used to create the strategy yields the same result.

The main change was lifting the default logic from the store layer to the service layer and adapting tests accordingly

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-service.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-service.ts
@@ -676,7 +676,7 @@ export class FeatureToggleService {
     ) {
         if (strategyName === 'flexibleRollout') {
             const stickiness =
-                params?.stickiness === undefined || params?.stickiness === ''
+                !params?.stickiness || params?.stickiness === ''
                     ? await this.featureStrategiesStore.getDefaultStickiness(
                           projectId,
                       )


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -680,6 +670,59 @@ export class FeatureToggleService {
...
+
+        return {
+            strategyName: name,
+            title,
+            disabled,
+            sortOrder,
+            constraints,
+            variants,
+            parameters,
+        };
```
gastonfournier: This groups logic from both create and update, so we don't have to maintain them separately. This probably led to the logic drift on how we handle stickiness (present at creation time but missing when updating)

src/lib/features/feature-toggle/tests/feature-toggle-service.e2e.test.ts
```diff
@@ -758,34 +762,79 @@ test('Should return last seen at per environment', async () => {
...
-    const feature = {
-        name: 'test-feature-stickiness-1',
-        description: 'the #1 feature',
-    };
-    const projectId = 'default';
+test.each([
+    ['empty stickiness', { rollout: '100', stickiness: '' }],
+    ['undefined stickiness', { rollout: '100' }],
+    ['undefined parameters', undefined],
+])(
```
gastonfournier: In all these scenarios, we should always end up with stickiness set to the default stickiness of the project

src/lib/features/feature-toggle/tests/feature-toggle-service.e2e.test.ts
```diff
@@ -758,34 +762,79 @@ test('Should return last seen at per environment', async () => {
...
+        };
+        const feature = {
+            name: `test-feature-create-${description.replaceAll(' ', '-')}`,
+        };
+        const projectId = 'default';
+        const defaultStickiness = `not-default-${description.replaceAll(' ', '-')}`;
+        const project = await stores.projectStore.update({
+            id: projectId,
+            name: 'stickiness-project-test',
+            defaultStickiness,
```
gastonfournier, line 800: Set default stickiness to something else than `'default'` for testing purposes

src/lib/features/project/project-store-type.ts
```diff
@@ -13,16 +13,23 @@ import type {
...
+    mode: ProjectMode;
+    defaultStickiness: string;
+    featureLimit?: number;
+    featureNamingPattern?: string;
+    featureNamingExample?: string;
+    featureNamingDescription?: string;
+    linkTemplates?: IProjectLinkTemplate[];
+}
+
+export interface IProjectInsert extends Partial<IProjectSettings> {
```
gastonfournier, line 26: IProjectInsert type was not accepting defaultStickiness. I guess this was an oversight as when sending defaultStickiness in the create object it was respected, so it was just the type not matching. Some properties like mode and featureLimit were duplicated between Insert and ProejctSettings but `defaultStickiness` was not.

There's also a bit of a type mismatch between `featureNamingPattern`, `featureNamingExample`, and `featureNamingDescription` from `ProjectSettings` and `featureNaming?: IFeatureNaming;` from `IProjectInsert`

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -680,6 +673,94 @@ export class FeatureToggleService {
...
+    private async defaultParameters(
+        projectId: string,
+        strategyName: string,
+        featureName: string,
+        params: IFeatureStrategy['parameters'] | undefined,
+    ) {
+        if (strategyName === 'flexibleRollout') {
+            if (params?.stickiness === '') {
+                // If stickiness is an empty string, we remove it to use the default stickiness.
+                delete params?.stickiness;
```
sjaanus: Can we avoid mutation, this seems fragile. Something like this, we clean beforehand and then use the cleaned version. This code is not tested.


```
private cleanParams(params: IFeatureStrategy['parameters'] | undefined) {
    if (!params) return {};
    const { stickiness, ...rest } = params;
    return stickiness === '' ? rest : params;
}
private async defaultParameters(
    projectId: string,
    strategyName: string,
    featureName: string,
    params: IFeatureStrategy['parameters'] | undefined,
) {
    const cleanedParams = this.cleanParams(params);

    if (strategyName === 'flexibleRollout') {
        return {
            rollout: '100',
            stickiness:
                cleanedParams.stickiness ??
                (await this.featureStrategiesStore.getDefaultStickiness(
                    projectId,
                )),
            groupId: featureName,
        };
    } else {
        return {};
    }
}

private async parametersWithDefaults(
    projectId: string,
    strategyName: string,
    featureName: string,
    params: IFeatureStrategy['parameters'] | undefined,
) {
    const cleanedParams = this.cleanParams(params);
    return mergeAll([
        await this.defaultParameters(
            projectId,
            strategyName,
            featureName,
            cleanedParams,
        ),
        cleanedParams,
    ]);
}

```

gastonfournier: I did it like this:
```typescript
const stickiness =
                params?.stickiness === undefined || params?.stickiness === ''
                    ? await this.featureStrategiesStore.getDefaultStickiness(
                          projectId,
                      )
                    : params?.stickiness;
```

src/lib/features/metrics/sizes/largest-resources-read-model.test.ts
```diff
@@ -79,7 +79,11 @@ test('can calculate resource size', async () => {
 
     await createFeature({
         featureName: 'featureB',
-        parameters: {},
+        parameters: {
+            groupId: 'featureB',
+            rollout: '100',
+            stickiness: 'default',
```
gastonfournier, line 85: These are now injected at the service layer but the test was expecting them to be inserted at the store level

src/lib/features/feature-toggle/tests/feature-toggle-service.e2e.test.ts
```diff
@@ -758,34 +760,82 @@ test('Should return last seen at per environment', async () => {
...
+    ['empty stickiness', { rollout: '100', stickiness: '' }],
+    ['undefined stickiness', { rollout: '100' }],
+    ['undefined parameters', undefined],
+    [
+        'different group id and stickiness',
+        { rollout: '100', groupId: 'test-group', stickiness: 'userId' },
+    ],
+    ['different rollout', { rollout: '25' }],
+    ['empty parameters', {}],
+    ['extra parameters are preserved', { extra: 'value', rollout: '100' }],
```
gastonfournier, line 773: These tests now account for the ones removed at the store level

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -680,6 +668,78 @@ export class FeatureToggleService {
...
 
+    private async parametersWithDefaults(
+        projectId: string,
+        featureName: string,
+        strategyName: string,
+        params: IFeatureStrategy['parameters'] | undefined,
+    ) {
+        if (strategyName === 'flexibleRollout') {
+            const stickiness =
+                params?.stickiness === undefined || params?.stickiness === ''
```
sjaanus: Can it be null?

gastonfournier: Even if the type does not allow it, it might... so this would work in both cases of null and undefined:
```suggestion
                !params?.stickiness || params?.stickiness === ''
```

PR #10441, opened 2025-07-30
fix: should be allowed to change strategy

gastonfournier:
## About the changes
Previous PR: https://github.com/Unleash/unleash/pull/10439 introduced a bug not allowing to update the strategy name

Because there was no test covering this, it was only caught later on integration.

This PR modifies a test to also validate the change of strategy and fixes the problem

### DIFF
--- a/src/lib/features/feature-toggle/feature-toggle-service.ts
+++ b/src/lib/features/feature-toggle/feature-toggle-service.ts
@@ -899,17 +899,11 @@ export class FeatureToggleService {
         const existingSegments = await this.segmentService.getByStrategy(id);
 
         if (existingStrategy.id === id) {
-            console.log(
-                `Updating strategy ${id} for feature ${featureName} in environment ${environment}: ${JSON.stringify(updates)}`,
-            );
             const standardizedUpdates = await this.standardizeStrategyConfig(
                 projectId,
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -898,12 +899,18 @@ export class FeatureToggleService {
         const existingSegments = await this.segmentService.getByStrategy(id);
 
         if (existingStrategy.id === id) {
+            console.log(
```
sjaanus: I think we can remove console logs

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -898,12 +899,18 @@ export class FeatureToggleService {
...
+                `Updating strategy ${id} for feature ${featureName} in environment ${environment}: ${JSON.stringify(updates)}`,
+            );
             const standardizedUpdates = await this.standardizeStrategyConfig(
                 projectId,
                 featureName,
                 { ...updates, name: updates.name! },
                 existingStrategy,
             );
+            console.log(
```
sjaanus: Also this

PR #10444, opened 2025-07-31
feat: Display example date for date formatting reasons

thomasheartman:
wip


frontend/src/component/user/Profile/ProfileTab/ProfileTab.tsx
```diff
@@ -215,30 +225,36 @@ export const ProfileTab = ({ user }: IProfileTabProps) => {
...
-                        {possibleLocales.map((locale) => {
-                            return (
-                                <option key={locale} value={locale}>
-                                    {locale}
-                                </option>
-                            );
-                        })}
-                    </Select>
-                </StyledFormControl>
+                <LocaleSelector>
```
thomasheartman, line 229: This whole bit is largely just indentation because I've wrapped the form control in a new element to facilitate the display we want. The only new parts are:

I've added an `aria-details` prop to the select, which points to the example. The example isn't necessary as a label or a description, but it does fit MDN's definition of details as:
> the element (or elements) that provide additional information related to the object.

And then, of course, I've added the element itself.

frontend/src/utils/formatDate.ts
```diff
@@ -14,7 +14,7 @@ export const formatDateYMDHMS = (
 
 export const formatDateYMDHM = (
     date: number | string | Date,
-    locale: string,
+    locale?: string,
```
thomasheartman, line 17: Locale isn't required in the `toLocaleString` function, and we may not have it in the spot where we call it, so I've marked it as optional here.

nunogois (approved):
LGTM!

>The example date is the date and time of the very first commit in the Unleash repo. By some stroke of luck, it happens to have everything we're looking for:

That's not luck, that's destiny ✨

PR #10446, opened 2025-07-31
chore: clear unknown flags every 24h instead of every 7d

nunogois:
https://linear.app/unleash/issue/2-3738/clear-unknown-flags-every-24h-instead-of-every-7d

Clears unknown flags every 24h instead of every 7d.

This ensures the list stays more relevant by removing stale entries sooner, allowing users to focus on actively reported unknown flags.

Also includes small improvements, including a new paragraph on the unknown flags page that better explains the concept of unknown flag reports.


gastonfournier (approved):
LG!

PR #8814, opened 2024-11-20
feat: info about unlimited projects option

Tymek:
- refactored projects list header
- added info about unlimited projects to open-source version

image
image

PR #10447, opened 2025-07-31
chore: overhaul project list pages

nunogois:
https://linear.app/unleash/issue/2-3743/overhaul-project-list-pages

This is essentially scouting work on our project list pages so we have a better baseline before proceeding with the cards/list view toggle.

Includes refactoring, fixes and visual improvements ensuring better consistency and alignment with our designs.

Took some liberties, so feel free to tell me I'm wrong.

### Searching

image

### Search results only in one of the groups

image

### No results found

image

### Helper text

image

### Title truncation with tooltip

image

### Archived projects

image

### DIFF
--- a/frontend/src/component/project/ProjectList/ProjectGroup.tsx
+++ b/frontend/src/component/project/ProjectList/ProjectGroup.tsx
@@ -4,9 +4,7 @@ import { ConditionallyRender } from 'component/common/ConditionallyRender/Condit
 import { ProjectCard as DefaultProjectCard } from '../ProjectCard/ProjectCard.tsx';
 import type { ProjectSchema } from 'openapi';
 import loadingData from './loadingData.ts';
-import { TablePlaceholder } from 'component/common/Table';
 import { styled } from '@mui/material';
-import { useSearchHighlightContext } from 'component/common/Table/SearchHighlightContext/SearchHighlightContext';
 import { UpgradeProjectCard } from '../ProjectCard/UpgradeProjectCard.tsx';
 import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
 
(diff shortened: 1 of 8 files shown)


frontend/src/component/project/ProjectList/ProjectGroup.tsx
```diff
@@ -40,86 +38,57 @@ type ProjectGroupProps = {
...
-                                No projects found matching &ldquo;
-                                {searchQuery}
-                                &rdquo;
-                            </TablePlaceholder>
-                        }
-                        elseShow={
-                            <TablePlaceholder>{placeholder}</TablePlaceholder>
-                        }
-                    />
-                }
```
nunogois, line 69: Extracted one level up. Easier to understand if you check the screenshots.

frontend/src/component/project/ProjectList/ProjectList.tsx
```diff
@@ -66,7 +74,7 @@ export const ProjectList = () => {
                     actions={
                         <>
                             <ConditionallyRender
-                                condition={!isOss && !isSmallScreen}
+                                condition={!isOss() && !isSmallScreen}
```
nunogois, line 77: Fun fact: We don't have project search in "larger than small screens" for [some time now](https://github.com/Unleash/unleash/pull/8814/files#diff-6fe5132bf326860735eea0148fe1c40c9cfb06e5836a9e02bc02e35165638bc0R69) and I guess no one noticed 🙈

chriswk (approved):
LGTM

frontend/src/component/project/ProjectList/ProjectList.tsx
gastonfournier: I hate this... function type shouldn't be cast to boolean without a compilation error... I wonder if that's TS config we're missing

nunogois: Likewise... but this is JS, and this is still a common way to check for truthy.

But I agree, would be nice if our linter/editor helped us catch this. Maybe it should require us to be explicit for truthy checks for non-boolean types.

I think this is what we're looking for: https://typescript-eslint.io/rules/strict-boolean-expressions/
But it seems like it's not supported in Biome yet.

PR #10450, opened 2025-08-01
chore: re-generate openapi

Tymek:
Update documentation to replace 'server-side' with 'backend' in API schemas


gastonfournier (approved):
:+1:

PR #10451, opened 2025-08-01
chore: fix interactive demo steps around constraint values

nunogois:
https://linear.app/unleash/issue/2-3744/fix-interactive-demo-steps-around-constraint-values

I noticed our interactive demo was broken around the constraint value steps.

This isn't a great fix, but it at least allows us to complete the interactive demo for now.


gastonfournier (approved):
LG!

PR #10462, opened 2025-08-04
chore: AI flag cleanup should focus on relevant flag

nunogois:
https://linear.app/unleash/issue/2-3749/prevent-ai-flag-cleanup-from-removing-the-wrong-flag

This slightly refines our prompt so we're strictly focusing on the relevant flag, instead of mistakenly targeting other flags.

Also includes:
- Adding our prompt to our output so we can more easily debug it 
- Only grab the last cleanup summary in case there are multiple (more than one step from the agent)
- Add a warning to the summary in case the agent couldn't find conditional logic related to the flag, only definitions / configurations

Did a few manual tests and it seemed to work correctly. Example: https://github.com/Unleash/unleash/pull/10461

### DIFF
--- a/.github/workflows/ai-flag-cleanup-pr.yml
+++ b/.github/workflows/ai-flag-cleanup-pr.yml
@@ -256,11 +256,14 @@ jobs:
               exit 1
           }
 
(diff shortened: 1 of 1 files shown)


gastonfournier (approved):
LGTM!

PR #10466, opened 2025-08-06
chore: project list table view

nunogois:
https://linear.app/unleash/issue/2-3740/implement-the-project-list-view

Implements the list (table) view of the projects page.

image

image

### DIFF
--- a/frontend/src/component/common/Table/cells/TextCell/TextCell.tsx
+++ b/frontend/src/component/common/Table/cells/TextCell/TextCell.tsx
@@ -26,6 +26,7 @@ const StyledWrapper = styled(Box, {
 
 const StyledSpan = styled('span')(() => ({
     display: 'inline-flex',
+    flexDirection: 'column',
     maxWidth: '100%',
 }));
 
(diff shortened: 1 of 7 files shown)


frontend/src/component/common/Table/cells/TextCell/TextCell.tsx
```diff
@@ -25,7 +25,7 @@ const StyledWrapper = styled(Box, {
 }));
 
 const StyledSpan = styled('span')(() => ({
-    display: 'inline-block',
+    display: 'inline-flex',
```
gastonfournier, line 28: Can this break other screens using the TextCell? Might be worthy of a deep check in sandbox before rolling out...

nunogois: Good catch. I went through all the tables, and the change only caused a visual difference in the segments table. So I've updated the span to use [flex-direction: column](https://github.com/Unleash/unleash/pull/10466/commits/95cd6fc212a9da32fbb07694a521bc4eacda4411) to better match the original behavior.

The main reason for this change (inline-flex) is that it fixes a subtle visual misalignment compared to other cells. Using a flex container also gives us more flexibility (pun intended) for future improvements.

I did one more pass through all the tables at the end, and they look okay to me.

PR #10467, opened 2025-08-06
chore: if flag is disabled automatically revert to showing cards

nunogois:
https://linear.app/unleash/issue/2-3753/fix-if-the-flag-is-disabled-we-should-automatically-revert-back-to

If the flag is disabled we should automatically revert back to showing cards.


gastonfournier (approved):
LG!

PR #10468, opened 2025-08-06
chore: add project-list-view-toggle Plausible event

nunogois:
https://linear.app/unleash/issue/2-3754/add-plausible-event-project-list-view-toggle-with-view

Adds a `project-list-view-toggle` Plausible event with a `view` prop so we can track if and how the new toggle is used.


gastonfournier (approved):
LG!

PR #10473, opened 2025-08-07
feat: promote segments script

leszekbulawa:
## About the changes

This PR adds a new script `promote-segments.sh` that enables promoting segments from one Unleash instance to another. This is particularly important because **importing feature toggles requires that the referenced segments already exist** in the target environment.

The script provides:
- Export segments from a source Unleash instance
- Create or update segments in a target Unleash instance  
- Optional filtering to promote only a specific segment by name (`--segment-name` flag)
- Comprehensive error handling and status reporting

### Key Features
- **Segment synchronization**: Ensures segments exist before importing dependent feature toggles
- **Selective promotion**: Use `--segment-name` to promote only specific segments
- **Upsert behavior**: Creates new segments or updates existing ones as needed
- **Data sanitization**: Removes readonly fields (`id`, `createdAt`, `createdBy`) before API calls
- **JSON validation**: Validates API responses to ensure data integrity

### Usage
```bash
# Configure the script variables at the top of the file
SOURCE_URL="https://source.unleash.com/api/admin/segments"
SOURCE_API_TOKEN="your-source-token"
TARGET_URL="https://target.unleash.com/api/admin/segments"  
TARGET_API_TOKEN="your-target-token"

# Promote all segments
./scripts/promote-segments.sh

# Promote only a specific segment
./scripts/promote-segments.sh --segment-name "mobile-users"
```


gastonfournier:
Hi @leszekbulawa, thanks for the proposal, but we don't believe this script belongs to Unleash codebase. It's true that export explicitly ignores the segment contents. When we made this decision, we considered that including the content of the segments could lead to some edge cases and decided to leave them out:

1. If you don't have segments in the target (happy path), you could have all the segment's contents in the export file.
2. When you already have segments in the target, there's a myriad of problems that can happen: mismatching contents with same, segment name, partially some segments could be missing, etc. if we want, we could decide to fail in this case until the user amends the situation in the target instance, but we considered it'd be simpler to amend by just creating the segments.
3. Size of the segments could become an issue.

Your script is a good way of accomplishing this, but we also don't consider import/export a regular operation our users perform daily. If anything, this feature should be built inside Unleash, but we have not considered it yet.

leszekbulawa:
> Hi @leszekbulawa, thanks for the proposal, but we don't believe this script belongs to Unleash codebase. It's true that export explicitly ignores the segment contents. When we made this decision, we considered that including the content of the segments could lead to some edge cases and decided to leave them out:
> 
> 1. If you don't have segments in the target (happy path), you could have all the segment's contents in the export file.
> 2. When you already have segments in the target, there's a myriad of problems that can happen: mismatching contents with same, segment name, partially some segments could be missing, etc. if we want, we could decide to fail in this case until the user amends the situation in the target instance, but we considered it'd be simpler to amend by just creating the segments.
> 3. Size of the segments could become an issue.
> 
> Your script is a good way of accomplishing this, but we also don't consider import/export a regular operation our users perform daily. If anything, this feature should be built inside Unleash, but we have not considered it yet.

Hey @gastonfournier. Thanks for reviewing the PR and for sharing your concerns — they make sense. Given that, how would you propose we approach migrating from one Unleash instance to another while ensuring data parity and integrity, without having to manually verify each configuration entry, which can be both time-consuming and error-prone? Would a database dump and restore be a good way of achieving this?

gastonfournier:
@leszekbulawa, migrating from one instance to another is not a common use case. Why are you trying to do that? We expect this to be a one-of operation, which is why you could migrate the segments manually one by one, and then do the import.

Using a script like you suggested should work fine.

leszekbulawa:
> @leszekbulawa, migrating from one instance to another is not a common use case. Why are you trying to do that? We expect this to be a one-of operation, which is why you could migrate the segments manually one by one, and then do the import.
> 
> Using a script like you suggested should work fine.

While it's not a common use case it is not non-existent as well eg. infrastructure changes, moving from one orchestrator to another, changing database from self-hosted to managed. In my case it's changing cloud provider. Maybe it's worth mentioning it in docs: https://github.com/Unleash/unleash/compare/main...leszekbulawa:unleash:patch-1

leszekbulawa:
@gastonfournier I've performed basic `pg_dump` and `pg_restore` and can confirm that it's a better way to migrate the data than using API and scripts. Restoring database using API and scripts does not pin Segments to archived feature flags. It also preserves "Created by" field.
In short - I would not recommend to use `scripts/promote.sh` for anything other that the actual sync (most likely for toggles promotion as the name indicates).
Leaving this here in case someone has a similar idea.

gastonfournier:
Thanks @leszekbulawa! That's a wonderful insight! Indeed, when I was thinking yesterday about your use case, I was considering that for a full migration, I would not use APIs. It's great that you confirmed what was going through my mind.

Also, it might be something we can include in our self-hosting documentation. I'm tagging @melindafekete our docs lead, in case this resonates with something she's planned.

PR #10475, opened 2025-08-07
chore: timeagocell default fallback

nunogois:
https://linear.app/unleash/issue/2-3759/set-timeagocell-default-fallback-to-never

Sets the default `emptyText` fallback to 'Never'. This should help with consistency and DRYness.

Also noticed some of our TimeAgoCell `title` properties were redundant, so I removed them for simplicity.

Examples:

image

image

### DIFF
--- a/frontend/src/component/admin/groups/Group/Group.tsx
+++ b/frontend/src/component/admin/groups/Group/Group.tsx
@@ -118,13 +118,7 @@ export const Group: VFC = () => {
             {
                 Header: 'Last login',
                 accessor: (row: IGroupUser) => row.seenAt || '',
-                Cell: ({ row: { original: user } }: any) => (
-                    <TimeAgoCell
-                        value={user.seenAt}
-                        emptyText='Never'
-                        title={(date) => `Last login: ${date}`}
-                    />
-                ),
+                Cell: TimeAgoCell,
                 maxWidth: 150,
             },
             {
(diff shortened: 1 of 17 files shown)


frontend/src/component/admin/groups/Group/Group.tsx
```diff
@@ -118,13 +118,7 @@ export const Group: VFC = () => {
             {
                 Header: 'Last login',
                 accessor: (row: IGroupUser) => row.seenAt || '',
-                Cell: ({ row: { original: user } }: any) => (
-                    <TimeAgoCell
-                        value={user.seenAt}
-                        emptyText='Never'
-                        title={(date) => `Last login: ${date}`}
```
gastonfournier, line 125: You mentioned it being redundant. Is it in the sense that the cell content says the same? This might be an accessibility thing. Not sure if removing it is the best, unless TimeAgoCell also has the title property

nunogois: TimeAgoCell will have a title with the date by default anyways, and you know it's the "last login" because it says so in the table header.


### Before

image

### After

image

gastonfournier: I understand, but if you use a screen reader, it may work better if you have the full "Last login: 07/01/2025".

gastonfournier (approved):
LGTM with one comment, but it's not a big deal I think

frontend/src/component/admin/groups/Group/Group.tsx
nunogois: If that's the case, I vote we go for consistency the other way around: https://github.com/Unleash/unleash/pull/10475/commits/d032c2e07f5f3c353fcd4afe28e947e14b8290a1

image

gastonfournier (approved):
LG!

PR #10477, opened 2025-08-07
chore: set min version for new SDK names

gastonfournier:
## About the changes
This sets the minimum version for the new SDK names to the current latest version. The versions with the new SDK name will be newer, but this will set the baseline for the future.

### DIFF
--- a/src/lib/features/metrics/instance/findOutdatedSdks.ts
+++ b/src/lib/features/metrics/instance/findOutdatedSdks.ts
@@ -12,6 +12,14 @@ const config: SDKConfig = {
     'unleash-client-ruby': '5.0.0',
     'unleash-client-dotnet': '4.1.3',
     'unleash-client-php': '2.3.0',
+    // new values after sdk registration rename:
+    'unleash-node-sdk': '6.6.0',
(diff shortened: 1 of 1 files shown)


chriswk (approved):
LGTM

src/lib/features/metrics/instance/findOutdatedSdks.ts
```diff
@@ -12,6 +12,14 @@ const config: SDKConfig = {
     'unleash-client-ruby': '5.0.0',
     'unleash-client-dotnet': '4.1.3',
     'unleash-client-php': '2.3.0',
+    // new values after sdk registration rename:
```
kwasniew, line 15: I trust you checked which versions introduce new naming scheme correctly :)

gastonfournier: It's still not included, but will be newer than the ones listed here, so they will not report as outdated :) at least that's my expectation :D

gastonfournier: BTW if we see something off we can change it

PR #10482, opened 2025-08-08
fix: now lifecycle trends will be added for flags that have official flag type

sjaanus:
We have instances that have fake flag types in features table. This ensures, we only count in flags have official types.


nunogois (approved):
LGTM!

PR #10488, opened 2025-08-11
Add prerequisites how-to-synchronize-unleash-instances.md

leszekbulawa:
Add info that Context fields and Segments require manual migration

## About the changes

Closes #


melindafekete:
We've since then specified things not imported and other validation rules: https://docs.getunleash.io/concepts/import-export

Thanks for flagging @leszekbulawa!

PR #10489, opened 2025-08-11
Fix/website performance

FredrikOseberg:
Work in progress.


website/src/components/VideoContent.jsx
```diff
@@ -1,29 +1,98 @@
...
+                        <path d='M8 5v14l11-7z' />
+                    </svg>
+                </div>
+
+                <div className={styles.loadingHint}>Click to load video</div>
+            </div>
+        );
+    }
+
+    return (
```
FredrikOseberg, line 61: Only load iframe when user has interacted with video image and play button

website/src/theme/Root.tsx
```diff
@@ -0,0 +1,161 @@
+import React, { useEffect } from 'react';
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+
+export default function Root({ children }: { children: React.ReactNode }) {
+    useEffect(() => {
```
FredrikOseberg, line 5: Lazy load kapa, google analytics and gtm

melindafekete (commented):
thanks @FredrikOseberg it's looking really good 🤩 

1. Removing mountain image -> I'm checking this with UX in #customer-journey-and-ux

2. Lower res videos -> is there a way to keep the lower res videos but have a slightly higher quality thumbnail? On my screen, they look quite pixelated
Screenshot 2025-08-25 at 11 02 51

3. Kapa: It throws an error when I try to ask it a question, but I think this should just be because it's a build preview

4. Contributors: Would it give us additional performance benefits if we completely remove the contributors feature? I think the original intent was to encourage community contributions to docs, but this hasn't happened, so we could also consider getting rid of it completely

website/src/types/global.d.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
```
kwasniew: is it related to web perf?

website/src/theme/DocItem/Footer/GitHubContributors.jsx
```diff
@@ -79,7 +79,7 @@ const GitHubContributors = ({ owner, repo, filePath }) => {
                                 title={`@${name}`}
                             >
                                 <img
-                                    src={contributor.avatar_url}
+                                    src={`${contributor.avatar_url}&s=66`}
```
kwasniew, line 82: maybe a comment in code for the magical 66

website/docusaurus.config.ts
```diff
@@ -421,16 +438,12 @@ class="header-github-link"
...
                     theme: {
                         customCss: './src/css/custom.css',
                     },
-                    googleAnalytics: {
-                        trackingID: 'UA-134882379-1',
-                    },
-                    googleTagManager: {
-                        containerId: 'GTM-KV5PRR2',
-                    },
+                    // Analytics/GTM moved to lazy loading in Root.tsx
```
kwasniew: this comment may not make sense for someone no familiar with the previous version

website/docusaurus.config.ts
```diff
@@ -903,28 +916,7 @@ class="header-github-link"
...
-                'data-button-height': '100px',
-                'data-button-image-height': '55px',
-                'data-button-image-width': '55px',
-                'data-button-text-font-weight': '400',
-                'data-button-text-font-size': '16px',
-                'data-button-text-font-family': '"Sen", sans-serif',
-                'data-button-text': 'ASK AI',
-                defer: true,
-            },
+            // Kapa AI widget is now lazy-loaded via src/theme/Root.tsx
```
kwasniew: same here, this comment may not make sense for someone not familiar with a previous version

FredrikOseberg:
> thanks @FredrikOseberg it's looking really good 🤩
> 
> 1. Removing mountain image -> I'm checking this with UX in #customer-journey-and-ux
> 2. Lower res videos -> is there a way to keep the lower res videos but have a slightly higher quality thumbnail? On my screen, they look quite pixelated
> 
> Screenshot 2025-08-25 at 11 02 51
> 3. Kapa: It throws an error when I try to ask it a question, but I think this should just be because it's a build preview
> 4. Contributors: Would it give us additional performance benefits if we completely remove the contributors feature? I think the original intent was to encourage community contributions to docs, but this hasn't happened, so we could also consider getting rid of it completely

1. Fantastic.

2. We're currently using the hd thumbnail from youtube. We can revert back to the maxres default thumbnail, but it will come with a slight performance cost. 

3. I assume it's not configured for Vercel, but it would be more comfortable to merge if we'd known for sure whether this was the case. Otherwise we can monitor when we deploy and quickly revert if there are production issues with Kapa.

4. I don't think there would be a significant difference, the current size of the profile images are around 1.7kb per profile picture (down from 32kb). So they are already optimized quite a lot. I'll leave the decision to keep / remove to you.

website/src/theme/Root.tsx
```diff
@@ -0,0 +1,161 @@
...
+        if (!ExecutionEnvironment.canUseDOM) {
+            return;
+        }
+
+        const loadGoogleAnalytics = () => {
```
kwasniew, line 10: if we use GTM I think it can load/manage GA without loading both

website/src/theme/Root.tsx
```diff
@@ -0,0 +1,161 @@
...
+            passive: true,
+        });
+
+        fallbackTimer = setTimeout(() => {
+            if (!hasInteracted) {
+                hasInteracted = true;
+                loadGoogleAnalytics();
+                loadGoogleTagManager();
+            }
+        }, 3000);
```
kwasniew, line 147: I'd explain this number and why we picked it

website/src/theme/Root.tsx
```diff
@@ -0,0 +1,161 @@
+import React, { useEffect } from 'react';
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+
+export default function Root({ children }: { children: React.ReactNode }) {
```
kwasniew, line 4: Maybe AnalyticsRoot to explain what's inside?

website/src/components/VideoContent.jsx
```diff
@@ -1,29 +1,98 @@
 // biome-ignore lint/correctness/noUnusedImports: Needs this for React to work
-import React from 'react';
+import React, { useState, useCallback } from 'react';
 import Admonition from '@theme/Admonition';
+import styles from './VideoContent.module.css';
+
+// Extract YouTube video ID from various URL formats
+const extractVideoId = (url) => {
+    const regex =
+        /(?:youtube\.com\/(?:[^\/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)([^"&?\/\s]{11})/;
```
kwasniew, line 9: wondering if using URLSearchParams to get v query param would be more readable

kwasniew (commented):
I don't see anything related to reo.js in this PR

FredrikOseberg:
> I don't see anything related to reo.js in this PR

Removed global.js, it's loaded by gtm regardless.

website/src/components/VideoContent.jsx
kwasniew: sometimes web perf folks recommend doing this on hover on desktop too

website/src/theme/Root.tsx
FredrikOseberg: Root is a special docusaurus concept: https://docusaurus.io/docs/swizzling#wrapper-your-site-with-root

kwasniew: can it use other components with responsibility driven names?

melindafekete:
@FredrikOseberg

Thanks!

2. Let's go with the hd thumbnail, then we can always revert to maxres later if needed
3. I'm happy to YOLO this, I checked another preview with unrelated changes, and it also threw an error there, so pretty sure it'll be okay. I'll check after deployment. 
4. Okay then, let's keep the contributors for now :)

FredrikOseberg:
> * I'm happy to YOLO this, I checked another preview with unrelated changes, and it also threw an error there, so pretty sure it'll be okay. I'll check after deployment.

Fantastic. I'll just finish up another task then I'll clean this PR and merge it. Let you know when I do.

melindafekete:
@FredrikOseberg cool, thanks. just waiting for Elise to confirm she's happy with the background removal

PR #10490, opened 2025-08-12
fix: move update approvers after cr change state

daveleek:
This moves the updateRequestedApprovers call until after changing state to in review.
We do this to make sure the approvers updated event gets processed after the change request makes it into in review.

That paves the way for the other change request that removes the notification listener for state change on change request, which should take care of the double notifications issue


daveleek:
Thanks @gastonfournier! Indeed, but they are in a diff repo :) PR is up now

PR #10494, opened 2025-08-13
feat: add origin column for consumption based tables

sjaanus:
We will start separating traffic, when it comes from different origins.
Also will add it as primary key so merging will work correctly.

There is a problem with using normal down migration, is that if we split data by origin, when down migrating and removing column, and re adding the old migration, we will have conflicts and it will fail.

This means before we down migrate, we need to aggregate/sum the data and only then we can reapply old schema.


src/migrations/202508131142433-consumption-origin.js
```diff
@@ -0,0 +1,68 @@
...
+                ADD PRIMARY KEY(day, metered_group, origin);
+        `,
+        cb
+    );
+};
+
+exports.down = function (db, cb) {
+    db.runSql(
+        `
+          CREATE TEMP TABLE tmp_request_count AS
```
sjaanus: We sum all origins up.

src/migrations/202508131142433-consumption-origin.js
```diff
@@ -0,0 +1,68 @@
...
+            SELECT
+                day,
+                metered_group,
+                SUM(requests) AS requests
+            FROM request_count_consumption
+            GROUP BY day, metered_group;
+
+            DELETE FROM request_count_consumption;
+
+            INSERT INTO request_count_consumption (day, metered_group, requests)
```
sjaanus: We insert the summed data.

src/migrations/202508131142433-consumption-origin.js
```diff
@@ -0,0 +1,68 @@
...
+        `
+          CREATE TEMP TABLE tmp_request_count AS
+            SELECT
+                day,
+                metered_group,
+                SUM(requests) AS requests
+            FROM request_count_consumption
+            GROUP BY day, metered_group;
+
+            DELETE FROM request_count_consumption;
```
sjaanus: We clear up the table that has origin.

src/migrations/202508131142433-consumption-origin.js
```diff
@@ -0,0 +1,68 @@
...
+
+            INSERT INTO request_count_consumption (day, metered_group, requests)
+            SELECT day, metered_group, requests FROM tmp_request_count;
+
+            DROP TABLE tmp_request_count;
```
sjaanus: We can get rid of temp table.

src/migrations/202508131142433-consumption-origin.js
```diff
@@ -0,0 +1,68 @@
...
+
+            ALTER TABLE request_count_consumption
```
sjaanus: We can drop old constraint and add back new primary key.

src/migrations/20250813142433-consumption-origin.js
```diff
@@ -0,0 +1,68 @@
...
+                ADD PRIMARY KEY(day, metered_group, origin);
+        `,
+        cb
+    );
+};
+
+exports.down = function (db, cb) {
+    db.runSql(
+        `
+          CREATE TEMP TABLE tmp_request_count AS
```
kwasniew, line 25: wondering if sth like this would be easier: 

```
        WITH aggregated_requests AS (
            SELECT
                day,
                metered_group,
                SUM(requests) AS requests
            FROM request_count_consumption
            GROUP BY day, metered_group
        )
        TRUNCATE request_count_consumption;
        INSERT INTO request_count_consumption (day, metered_group, requests)
        SELECT day, metered_group, requests FROM aggregated_requests;

        ALTER TABLE request_count_consumption
            DROP CONSTRAINT request_count_consumption_pkey,
            DROP COLUMN origin,
            ADD PRIMARY KEY(day, metered_group);

        -- Collapse connection_count_consumption
        WITH aggregated_connections AS (
            SELECT
                day,
                metered_group,
                SUM(requests) AS requests,
                SUM(connections) AS connections
            FROM connection_count_consumption
            GROUP BY day, metered_group
        )
        TRUNCATE connection_count_consumption;
        INSERT INTO connection_count_consumption (day, metered_group, requests, connections)
        SELECT day, metered_group, requests, connections FROM aggregated_connections;

        ALTER TABLE connection_count_consumption
            DROP CONSTRAINT connection_count_consumption_pkey,
            DROP COLUMN origin,
            ADD PRIMARY KEY(day, metered_group);
```

kwasniew: no temp tables, should be faster and safer

PR #10500, opened 2025-08-18
feat: incorporate backend as a valid api token type replacing client

gastonfournier:
This PR deprecates `CLIENT` api token type in favor of `BACKEND` but both will continue working.

Also replaces:
- `INIT_CLIENT_API_TOKENS` with `INIT_BACKEND_API_TOKENS`. The former is kept for backward compatibility.


src/lib/openapi/spec/create-api-token-schema.ts
```diff
@@ -20,8 +20,8 @@ const clientFrontendSchema = {
         type: {
             type: 'string',
             pattern:
-                '^([Cc][Ll][Ii][Ee][Nn][Tt]|[Ff][Rr][Oo][Nn][Tt][Ee][Nn][Dd])$',
-            description: `A client or frontend token. Must be one of the strings "client" or "frontend" (not case sensitive).`,
+                '^([Cc][Ll][Ii][Ee][Nn][Tt]|[Bb][Aa][Cc][Kk][Ee][Nn][Dd]|[Ff][Rr][Oo][Nn][Tt][Ee][Nn][Dd])$',
+            description: `A client or frontend token. Must be one of the strings "client" (deprecated), "backend" (preferred over "client") or "frontend" (not case sensitive).`,
```
nunogois, line 24: Should non-deprecated types show up first in our description?

src/lib/openapi/spec/create-project-api-token-schema.ts
```diff
@@ -10,8 +10,8 @@ export const createProjectApiTokenSchema = {
         type: {
             type: 'string',
             pattern:
-                '^([Cc][Ll][Ii][Ee][Nn][Tt]|[Ff][Rr][Oo][Nn][Tt][Ee][Nn][Dd])$',
-            description: `A client or frontend token. Must be one of the strings "client" or "frontend" (not case sensitive).`,
+                '^([Cc][Ll][Ii][Ee][Nn][Tt]|[Bb][Aa][Cc][Kk][Ee][Nn][Dd]|[Ff][Rr][Oo][Nn][Tt][Ee][Nn][Dd])$',
+            description: `A client or frontend token. Must be one of the strings "client" (deprecated), "backend" (preferred over "client") or "frontend" (not case sensitive).`,
```
nunogois, line 14: Same as https://github.com/Unleash/unleash/pull/10500/files#r2285338771

src/lib/routes/admin-api/api-token.ts
```diff
@@ -3,7 +3,7 @@ import type { Response } from 'express';
 import Controller from '../controller.js';
 import {
     ADMIN,
-    CREATE_CLIENT_API_TOKEN,
+    CREATE_CLIENT_API_TOKEN as CREATE_BACKEND_API_TOKEN,
```
nunogois: Why are we doing this for the CREATE but not DELETE or READ?

nunogois (approved):
I think this is fine overall, but a few things stand out.

- Reusing the existing permissions feels a bit odd, but I understand it's the easiest path forward.
- If we're planning to support both for now, it's a bit inconsistent that some tests directly replace one for the other. Not a huge deal though.
- Have we confirmed that our UI shows BACKEND correctly for both client and backend type tokens?

What if we simplify this? We can keep the changes that make our user-facing interfacing elements (UI and ENV vars) consistent (BACKEND instead of CLIENT), but continue calling it "client" internally for now.

If we'd rather reduce technical debt right away, I think we should instead go all in and consider a DB migration (including permissions).

src/lib/routes/admin-api/api-token.ts
gastonfournier: I think this is due to some code-refactoring. I'll leave it as it was

gastonfournier:
> * Reusing the existing permissions feels a bit odd, but I understand it's the easiest path forward.

Yes, I was thinking that a migration could be done, but in order to do it properly, we should follow the expand and contract pattern. So this would be the expand part. I was thinking that the contract could be done with v8. We could potentially use [joi.replace](https://github.com/Unleash/unleash/pull/10500/commits/515f6ad8224648f06e40b6de6c4eddbcc1d812e7#diff-6be1f2387d6d9e57d26c7e05a09c6c04f5373f16bfd66286dec403a0a2b4bf52R13) to replace CLIENT with BACKEND but you end up in a situation where you send `type: client` to a create API and you get a `type: backend` in response, which feels odd.

> * If we're planning to support both for now, it's a bit inconsistent that some tests directly replace one for the other. Not a huge deal though.

I tried to identify the most relevant tests and add both backend and client, but other tests I decided to leave them as client. The only places where this had to be changed is in e2e tests that expect a token, because the init script will now create backend tokens: https://github.com/Unleash/unleash/pull/10500/files#diff-a38cc7ae9f963264333b6535498970a33e046dba50a6af554f438343a6edd2ddR439-R442

> * Have we confirmed that our UI shows BACKEND correctly for both client and backend type tokens?

Yes, there are a few leftovers like `CLIENT API URL` in some UI components.

> What if we simplify this? We can keep the changes that make our user-facing interfacing elements (UI and ENV vars) consistent (BACKEND instead of CLIENT), but continue calling it "client" internally for now.

That's the intent of this PR and why we're not migrating permissions.

> If we'd rather reduce technical debt right away, I think we should instead go all in and consider a DB migration (including permissions). If this requires breaking changes, let's add it to a backlog for the next major.

We don't have a backlog yet, we're 10~11 months away from v8, so I'd rather review this again when we're planning for v8 (same as we did for v7)

src/lib/openapi/spec/create-api-token-schema.ts
gastonfournier: I think it doesn't change much. The comment should cover the documentation part.

alexcasalboni:
@gastonfournier FYI this PR breaks Unleash Edge (v19.15.0):

```shell
WARN unleash_edge::http::unleash_client: Failed to parse validation response with error: reqwest::Error { kind: Decode, source: Error("unknown variant `backend`, expected one of `FRONTEND`, `frontend`, `CLIENT`, `client`, `ADMIN`, `admin`, `invalid`", line: 1, column: 85) }
```

gastonfournier:
> @gastonfournier FYI this PR breaks Unleash Edge (v19.15.0):
> 
> ```shell
> WARN unleash_edge::http::unleash_client: Failed to parse validation response with error: reqwest::Error { kind: Decode, source: Error("unknown variant `backend`, expected one of `FRONTEND`, `frontend`, `CLIENT`, `client`, `ADMIN`, `admin`, `invalid`", line: 1, column: 85) }
> ```

Hi @alexcasalboni, thanks for the report! I think it'd only break if you're using an initial token, right? Because now initial tokens are created as backend... If that's the case, I should probably change it.

Can you confirm that's the case?

alexcasalboni:
@gastonfournier Yes, correct. Unfortunately, that is required in strict mode (which is the recommended option).

In other words, right now it's impossible to launch Unleash Edge in strict mode using `Unleash>=v7.1.1`.

Not 100% sure if this should be fixed in this repo (to provide backward compatibility) or directly in the unleash-edge repo (to support the new API format). Either way, happy to help and open an issue in the right place :)

gastonfournier:
> @gastonfournier Yes, correct. Unfortunately, that is required in strict mode (which is the recommended option).
> 
> In other words, right now it's impossible to launch Unleash Edge in strict mode using `Unleash>=v7.1.1`.
> 
> Not 100% sure if this should be fixed in this repo (to provide backward compatibility) or directly in the unleash-edge repo (to support the new API format). Either way, happy to help and open an issue in the right place :)

Thanks! We've fixed it and released a patch 7.1.2. The issue was only present in 7.1.1, not in 7.1.0. If you can validate it'd be great! 

Moving forward, we'll make edge support backend tokens, but the challenge is that not everyone upgrades, so we're considering returning different responses depending on who the client is. We did this in the past, but for now, the backend token types will be returned as client token types for backward compatibility. This should not be a problem because both are the same. It's mostly a rename, but we'll keep it under the hood until we can guarantee this doesn't break clients.

alexcasalboni:
@gastonfournier thanks 🙏 

and yes, I can confirm it is only happening in 7.1.1 (not 7.1.0) 👌

PR #10503, opened 2025-08-19
chore: ignore CR events with feature name

gastonfournier:
## About the changes
This ignores Change Request event types when calculating the etag because Change Request events don't change data.

They were being included when the change request event contained a featureName. After this change, those should be excluded.


src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -139,6 +140,14 @@ describe.each([
             TEST_AUDIT_USER,
         );
 
+        await app.services.eventService.storeEvent({
```
gastonfournier, line 143: This is added as the last event to prevent the IDs of the existing events above this one from shifting.

sastromskis (approved):
LGTM!

PR #10512, opened 2025-08-19
feat: support different etags per environment

gastonfournier:
## About the changes
This PR introduces environment-specific etags. This way clients will not react by updating features when there are changes in environments the SDK doesn't care about.

## Details
There's a bit of scouting work (please don't make me split this :pray:) and other details are in comments, but the most relevant for the lazy ones:
- Important **decision** on how we detect changes, unifying polling and delta: https://github.com/Unleash/unleash/pull/10512#discussion_r2285677129
- **Decision** on how we update revision id per environment: https://github.com/Unleash/unleash/pull/10512#discussion_r2291888401 
- and how we do initial fetch on the read path: https://github.com/Unleash/unleash/pull/10512#discussion_r2291884777
- The singleton pattern that gave me **nightmares**: https://github.com/Unleash/unleash/pull/10512#discussion_r2291848934
- **Do we still have ALL_ENVS tokens?** https://github.com/Unleash/unleash/pull/10512#discussion_r2291913249

## Feature flag
To control the rollout introduced `etagByEnv` feature: [0da567d](https://github.com/Unleash/unleash/pull/10512/commits/0da567dd9b6577fb316144b81babaa5fc1b27727)


src/lib/features/events/event-store.ts
```diff
@@ -191,29 +191,38 @@ export class EventStore implements IEventStore {
...
+                    }
+                    return inner;
+                })
+                .orWhereIn('type', [
+                    SEGMENT_UPDATED,
+                    FEATURE_IMPORT,
+                    FEATURES_IMPORTED,
+                    SEGMENT_CREATED,
+                    SEGMENT_DELETED,
+                ]);
```
gastonfournier: Extracted this method because I think both `getMaxRevisionId` and `getRevisionRange` should be in sync about what type of events they care about. I noticed they were off.

For example, SEGMENT_DELETED was not included in the max revision by id.

src/lib/app.ts
```diff
@@ -9,7 +9,7 @@ import {
     corsOriginMiddleware,
 } from './middleware/index.js';
 import rbacMiddleware from './middleware/rbac-middleware.js';
-import apiTokenMiddleware from './middleware/api-token-middleware.js';
+import { apiAccessMiddleware } from './middleware/api-token-middleware.js';
```
gastonfournier, line 12: Just switching to import by name rather than the default import.

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -357,11 +357,11 @@ export default class FeatureController extends Controller {
     }
 
     async calculateMeta(query: IFeatureToggleQuery): Promise<IMeta> {
-        // TODO: We will need to standardize this to be able to implement this a cross languages (Edge in Rust?).
```
gastonfournier, line 360: This probably came from a doubt when we were first implementing this. But I think both Unleash and edge should be resolving this in the same way

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
```diff
@@ -357,11 +357,11 @@ export default class FeatureController extends Controller {
...
         const revisionId =
-            await this.configurationRevisionService.getMaxRevisionId();
+            await this.configurationRevisionService.getMaxRevisionId(
+                query.environment,
```
gastonfournier: We should now consider the environment if calculating the etag for an environment specific token.

src/lib/features/events/event-store.ts
```diff
@@ -225,27 +234,11 @@ export class EventStore implements IEventStore {
...
-                        FEATURES_IMPORTED,
-                        SEGMENT_CREATED,
-                        SEGMENT_DELETED,
-                    ]),
-            )
+            .andWhere(this.typeIsInteresting())
             .orderBy('id', 'asc');
 
         const rows = await query;
+        stopTimer();
```
gastonfournier, line 247: Missing stop timer means we were not receiving the metric and maybe leaking memory

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -83,5 +111,6 @@ export default class ConfigurationRevisionService extends EventEmitter {
 
     destroy(): void {
         ConfigurationRevisionService.instance?.removeAllListeners();
+        ConfigurationRevisionService.instance = undefined;
```
gastonfournier, line 114: ## Debugging war story (a.k.a. “Why did this eat my entire day?”)
This service is a classic singleton: the first test suite that calls `ConfigurationRevisionService.getInstance(...)` wins, and every later call just gets the same in-memory instance back. 

**That means**:
- Same revisionId
- Same maxRevisionId map
- Same eventStore reference (from the FIRST app’s stores)

In normal runs (one app per process) this is fine. But in tests that spin up multiple apps (e.g. calling `setupApp` twice), only the FIRST `eventStore` is actually wired in. Later apps _happily_ think they have
their own "fresh" services, but **this singleton** keeps pointing at the original one. 

**The symptoms looked spooky:**
- “Stolen” DB state (events not showing up where expected)
- Mismatched revision counters
- Irreproducible unless two apps were created in the same test process
- Running a single test file in isolation = works; running the suite = chaos

**The fix**: 
In test teardown, we usually call `app.destroy()`, which calls the destroy method on each service. The missing piece was nuking the singleton so the next app build really gets a clean `ConfigurationRevisionService`. 

**Moral**: singletons + test with more than one initApp = sneaky shared state gremlins.

src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -2538,7 +2538,10 @@ export class FeatureToggleService {
                         environment,
                     ));
                 if (!canAddStrategies) {
-                    throw new PermissionError(CREATE_FEATURE_STRATEGY);
+                    throw new PermissionError(
+                        CREATE_FEATURE_STRATEGY,
+                        environment,
```
gastonfournier, line 2543: This is a little detail. `CREATE_FEATURE_STRATEGY` is an environment permission. Including the environment here provides a better error message.

src/lib/middleware/api-token-middleware.ts
```diff
@@ -31,7 +31,8 @@ export const TOKEN_TYPE_ERROR_MESSAGE =
 
 export const NO_TOKEN_WHERE_TOKEN_WAS_REQUIRED =
     'This endpoint requires an API token. Please add an authorization header to your request with a valid token';
-const apiAccessMiddleware = (
+
+export const apiAccessMiddleware = (
```
gastonfournier, line 35: Another scouting task: provide a named export besides the default export. Later, we should remove the default after confirming it's no longer in use (expand and contract)

src/lib/middleware/rbac-middleware.ts
```diff
@@ -87,6 +87,7 @@ const rbacMiddleware = (
                 )
             ) {
                 const { featureName } = params;
+                // TODO track if this deprecated path is still in use
```
gastonfournier, line 90: Just a reminder from an old hack we said had to be removed in v5

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -51,14 +53,36 @@ export default class ConfigurationRevisionService extends EventEmitter {
         return ConfigurationRevisionService.instance;
     }
 
-    async getMaxRevisionId(): Promise<number> {
+    async getMaxRevisionId(environment?: string): Promise<number> {
+        if (environment && !this.maxRevisionId[environment]) {
+            await this.updateMaxEnvironmentRevisionId(environment);
+        }
```
gastonfournier, line 59: If querying by environment and the max revision for that environment is not set, try to set it.

This is on the read path, so we only want to update if it's not set. For the asynchronous updates, the scheduled job will handle.

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -69,8 +93,12 @@ export default class ConfigurationRevisionService extends EventEmitter {
...
         if (this.revisionId !== revisionId) {
             this.logger.debug(
-                'Updating feature configuration with new revision Id',
-                revisionId,
+                `Updating feature configuration with new revision Id ${revisionId} and all envs: ${Object.keys(this.maxRevisionId).join(', ')}`,
+            );
+            await Promise.allSettled(
+                Object.keys(this.maxRevisionId).map((environment) =>
+                    this.updateMaxEnvironmentRevisionId(environment),
+                ),
```
gastonfournier, line 101: This is the update path, only when the global max revision id changed, there's a chance that some of the environments also need to change, so we need to try to update them all. 

An optimization idea could be to do it in a single query, but this happens in a background thread, so I do not consider this optimization that valuable

src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -1,220 +1,314 @@
...
+                },
+            },
+        },
+        db.rawDatabase,
+    );
+
+    return { app, db };
+}
+
+async function initialize({ app, db }: { app: IUnleashTest; db: ITestDb }) {
```
gastonfournier, line 76: Function that sets up an initial, predictable state for the database

src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -1,220 +1,314 @@
...
+
+async function validateInitialState({
+    app,
+    db,
+}: { app: IUnleashTest; db: ITestDb }) {
+    /**
+     * This helps reason about the etag, which is formed by <query-hash>:<event-id>
+     * To see the output you need to run this test with --silent=false
+     * You can see the expected output in the expect statement below
+     */
```
gastonfournier, line 104: This function is just a visual helper to understand what's in the database at the moment of testing (helps reasoning about the event ids), and also protects us from accidental modifications of the ids.

src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -1,220 +1,314 @@
...
+async function initialize({ app, db }: { app: IUnleashTest; db: ITestDb }) {
+    const allEnvs = await app.services.environmentService.getAll();
+    const nonDefaultEnv = allEnvs.find((env) => env.name !== DEFAULT_ENV)!.name;
+
+    await app.createFeature('X');
+    await app.createFeature('Y');
+    await app.archiveFeature('Y');
+    await app.createFeature('Z');
+    await app.enableFeature('Z', DEFAULT_ENV);
+    await app.enableFeature('Z', nonDefaultEnv);
```
gastonfournier, line 85: Using some test helpers makes the code more readable

src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -1,220 +1,314 @@
...
+
+    await app.services.eventService.storeEvent({
+        type: CHANGE_REQUEST_CREATED,
+        createdBy: 'some@example.com',
+        createdByUserId: 123,
+        ip: '127.0.0.1',
+        featureName: `X`,
+    });
```
gastonfournier, line 93: This is a random change request event to prove it's not invalidating the cache (from previous optimization)

src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -1,220 +1,314 @@
...
+            `"ae443048:15${enabled ? `:${name}` : ''}"`,
+        );
     });
-});
 
-// running after all inside describe block, causes some of the queries to fail to acquire a connection
-// this workaround is to run the afterAll outside the describe block
-afterAll(async () => {
-    await Promise.all(shutdownHooks.map((hook) => hook()));
+    test('production environment gets a different etag than development', async () => {
```
gastonfournier: Test that different environments get different etags.

src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -1,220 +1,314 @@
...
-    test('returns 200 when content updates and hash does not match anymore', async () => {
-        await app.services.featureToggleService.createFeatureToggle(
-            'default',
-            {
-                name: `featureNew304${apendix}`,
-                description: 'the #1 feature',
-            },
-            TEST_AUDIT_USER,
-        );
+    test('creating a new feature does not modify etag', async () => {
```
gastonfournier: This is because a feature creation doesn't change the configuration (there's no strategy in it). The previous test was wrong because it did not include the double quotes in the if-none-match header, and therefore it was returning a 200 response.

src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -1,220 +1,314 @@
...
-        const res = await app.request
+        await app.request
             .get('/api/client/features')
-            .set('if-none-match', 'ae443048:16')
+            .set('Authorization', devTokenSecret)
+            .set('if-none-match', `"76d8bb0e:13${enabled ? `:${name}` : ''}"`)
+            .expect(304);
+    });
+
+    test('a token with all envs should get the max id regardless of the environment', async () => {
```
gastonfournier: Test for all environments token (`'*'`). I didn't know this was supported. At least invalid from the UI, but it's still part of our code, but it behaves as a fallback to the default environment: https://github.com/Unleash/unleash/blob/18bea2538318cb7aaa26e6583a65547a18eeae16/src/lib/features/frontend-api/global-frontend-api-cache.ts#L142-L146

I think this approach of defaulting to the max revision id is also valid and covers either case.

src/test/e2e/api/client/feature.optimal304.e2e.test.ts
```diff
@@ -1,220 +1,314 @@
...
+            .get('/api/client/features')
+            .set('Authorization', devTokenSecret)
+            .expect(200);
+
+        expect(devHeaders.etag).toEqual(
+            `"76d8bb0e:13${enabled ? `:${name}` : ''}"`,
+        );
+    });
+
+    test('modifying dev environment should only invalidate dev tokens', async () => {
```
gastonfournier: This test validates that modifying a flag in a single environment only invalidates that environment.

src/lib/features/events/event-store.ts
gastonfournier: @sjaanus let me know if this difference was intentional or not. I guess it's not, and that's why the change. Not sure if that should be the case moving forward also

gastonfournier: Ok, we chatted about it and it was intentionally different:

**In polling:**
1. When you create or delete a segment, it's never active until you change a strategy.
2. When you change a strategy you send all data (including the segment).

**In Delta/Streaming:**
1. When you change the strategy you just send the strategy
2. This means that if you had not send the segment before-hand you won't get it

gastonfournier: Did this: [84cee36](https://github.com/Unleash/unleash/pull/10512/commits/84cee36018163d3f72611d3dfd9e31930d7b19f7) cause I still want to keep the code somewhat reusable.

src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
chriswk: Not really. Edge inherits the meta information, but does its own etag calculation. But I agree with removing the comment.

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -51,14 +53,36 @@ export default class ConfigurationRevisionService extends EventEmitter {
...
+        if (environment && !this.maxRevisionId[environment]) {
+            await this.updateMaxEnvironmentRevisionId(environment);
+        }
+        if (
+            environment &&
+            this.maxRevisionId[environment] &&
+            this.maxRevisionId[environment] > 0
+        ) {
+            return this.maxRevisionId[environment];
+        }
```
nunogois, line 66: Could this be simplified to:

```ts
if (environment) {
    const maxRevId =
        this.maxRevisionId.get(environment) ||
        (await this.updateMaxEnvironmentRevisionId(environment));

    if (maxRevId > 0) {
        return maxRevId;
    }
}
```

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -51,14 +53,36 @@ export default class ConfigurationRevisionService extends EventEmitter {
...
+            this.maxRevisionId[environment],
+            environment,
+        );
+        if (this.maxRevisionId[environment] ?? 0 < envRevisionId) {
+            this.maxRevisionId[environment] = envRevisionId;
+        }
+
+        return this.maxRevisionId[environment];
+    }
+
```
nunogois, line 85: Same here. I also don't think you should directly access the map. You should use the `get` and `set` methods instead.

```ts
async updateMaxEnvironmentRevisionId(environment: string): Promise<number> {
    let maxRevId = this.maxRevisionId.get(environment) ?? 0;
    
    const envRevisionId = await this.eventStore.getMaxRevisionId(
        maxRevId,
        environment,
    );
    
    if (maxRevId < envRevisionId) {
        this.maxRevisionId.set(environment, envRevisionId);
        maxRevId = envRevisionId;
    }

    if (this.flagResolver.isEnabled('debugEtag')) {
        this.logger.info(
            `[etag] Computed ETag for environment ${environment}: ${envRevisionId} stored as "${maxRevId}"`,
        );
    }

    return maxRevId;
}
```

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -69,8 +93,12 @@ export default class ConfigurationRevisionService extends EventEmitter {
         );
         if (this.revisionId !== revisionId) {
             this.logger.debug(
-                'Updating feature configuration with new revision Id',
-                revisionId,
+                `Updating feature configuration with new revision Id ${revisionId} and all envs: ${Object.keys(this.maxRevisionId).join(', ')}`,
+            );
+            await Promise.allSettled(
+                Object.keys(this.maxRevisionId).map((environment) =>
+                    this.updateMaxEnvironmentRevisionId(environment),
```
nunogois, line 100: Again, this is a Map and should be handled as such. I don't think this is doing what we're expecting 🤔 

image



```ts
await Promise.allSettled(
    [...this.maxRevisionId.keys()].map((environment) =>
        this.updateMaxEnvironmentRevisionId(environment),
    ),
);
```

src/lib/middleware/api-token-middleware.ts
nunogois: Not sure I'm a fan of calling this `apiAccessMiddleware` if the file name is `api-token-middleware`. Should we make it consistent?

PR #10519, opened 2025-08-21
chore: implement unknown flags UX feedback

nunogois:
https://linear.app/unleash/issue/2-3809/implement-the-latest-feedback-from-ux

Implements the latest feedback from UX regarding **unknown flags**.

Bit unsure about our column headers. E.g. instead of "Last seen" and "Seen in Unleash" we could call them "Reported" and "Last event".

image


src/lib/features/metrics/unknown-flags/unknown-flags-store.ts
```diff
@@ -64,11 +66,17 @@ export class UnknownFlagsStore implements IUnknownFlagsStore {
...
+            'uf.name',
+            'uf.app_name',
+            'uf.seen_at',
+            'uf.environment',
+            this.db.raw(
+                `(SELECT MAX(e.created_at)
+                FROM ${TABLE_EVENTS} AS e
+                WHERE e.feature_name = uf.name)
+                AS last_event_at`,
+            ),
```
gastonfournier, line 79: I'd be worried about performance if this was a widely used functionality, but it is not, so :rocket:

nunogois: What's your concern? Can you think of a better way of doing this? I believe we're correctly leveraging indexes here and should be fine.

gastonfournier: I mean, it's a nested query for each feature name, but as mentioned, no concern because this is not a widely used feature.

PR #10523, opened 2025-08-22
docs: some details spotted to align with new token types

gastonfournier:
## About the changes
Few places where we can use Backend token instead of CLIENT.


chriswk (approved):
Nice

PR #10529, opened 2025-08-25
Remove ratio graph & column to the right

thomasheartman:
This PR removes the flags archived / flags created trend line as well as the right-hand column.

Before:
image

After:
image

I'm not deleting the tooltip yet, because I think we'll want to re-use that for the group tooltip.


sjaanus (approved):
LG

PR #10533, opened 2025-08-25
fix: pagination limit and placeholder

kwasniew:
## About the changes

* make sure that only page limit available in the UI can be selected and choose 25 otherwise. This will solve a problem when someone has bookmarked URL with limit=1 or has this value in local storage
* make sure that when paginating to next pages we see loading placeholder instead of "No feature flags available"

This is what we have now when changin a page on slow connection
Screenshot 2025-08-25 at 16 13 18

## Discussion points

Another issue when paginating is that total count drops to 0 for a short period. I know we had a cache component for this but it doesn't seem to be working anymore.
Screenshot 2025-08-25 at 16 13 23

### DIFF
--- a/frontend/src/component/common/Table/PaginationBar/PaginationBar.tsx
+++ b/frontend/src/component/common/Table/PaginationBar/PaginationBar.tsx
@@ -1,4 +1,5 @@
 import type React from 'react';
+import { useEffect } from 'react';
 import { Box, Typography, Button, styled } from '@mui/material';
 import { ConditionallyRender } from '../../ConditionallyRender/ConditionallyRender.tsx';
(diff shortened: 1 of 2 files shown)


frontend/src/component/common/Table/PaginationBar/PaginationBar.tsx
```diff
@@ -60,6 +61,12 @@ export const PaginationBar: React.FC<PaginationBarProps> = ({
     fetchNextPage,
     setPageLimit,
 }) => {
+    useEffect(() => {
```
kwasniew, line 64: I considered different places to have this code and the least effort was to add it to PaginationBar that allows 4 selection options

PR #10542, opened 2025-08-26
chore: Etag log and code improvements

gastonfournier:
This is a follow-up on the comments in https://github.com/Unleash/unleash/pull/10512

Mostly using get and set when accessing the Map

### DIFF
--- a/src/lib/features/feature-toggle/configuration-revision-service.ts
+++ b/src/lib/features/feature-toggle/configuration-revision-service.ts
@@ -99,15 +99,14 @@ export default class ConfigurationRevisionService extends EventEmitter {
             this.revisionId,
         );
         if (this.revisionId !== revisionId) {
+            const knownEnvironments = [...this.maxRevisionId.keys()];
             this.logger.debug(
-                `Updating feature configuration with new revision Id ${revisionId} and all envs: ${Object.keys(this.maxRevisionId).join(', ')}`,
+                `Updating feature configuration with new revision Id ${revisionId} and all envs: ${knownEnvironments.join(', ')}`,
             );
             await Promise.allSettled(
-                this.maxRevisionId
-                    .keys()
-                    .map((environment) =>
-                        this.updateMaxEnvironmentRevisionId(environment),
-                    ),
+                knownEnvironments.map((environment) =>
+                    this.updateMaxEnvironmentRevisionId(environment),
+                ),
             );
             this.revisionId = revisionId;
             if (emit) {
--- a/src/lib/features/feature-toggle/configuration-revision-service.ts
+++ b/src/lib/features/feature-toggle/configuration-revision-service.ts
@@ -54,15 +54,15 @@ export default class ConfigurationRevisionService extends EventEmitter {
     }
 
     async getMaxRevisionId(environment?: string): Promise<number> {
-        if (environment && !this.maxRevisionId[environment]) {
-            await this.updateMaxEnvironmentRevisionId(environment);
-        }
-        if (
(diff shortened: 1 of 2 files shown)


src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -98,19 +99,15 @@ export default class ConfigurationRevisionService extends EventEmitter {
...
-                this.logger.info(
-                    `[etag] Updating feature configuration with new revision Id ${revisionId} and all envs: ${Object.keys(this.maxRevisionId).join(', ')}`,
-                );
-            } else {
-                this.logger.debug(
-                    `Updating feature configuration with new revision Id ${revisionId} and all envs: ${Object.keys(this.maxRevisionId).join(', ')}`,
-                );
-            }
+            this.logger.debug(
+                `Updating feature configuration with new revision Id ${revisionId} and all envs: ${Object.keys(this.maxRevisionId).join(', ')}`,
```
nunogois: Like I mentioned [here](https://github.com/Unleash/unleash/pull/10512/files#r2298391704) I don't think this will work as expected since it is a Map instead of a plain object:

image

However this will work as expected:

image

```suggestion
                `Updating feature configuration with new revision Id ${revisionId} and all envs: ${[...this.maxRevisionId.keys()].join(', ')}`,
```

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -98,19 +99,15 @@ export default class ConfigurationRevisionService extends EventEmitter {
...
+            );
             await Promise.allSettled(
-                Object.keys(this.maxRevisionId).map((environment) =>
-                    this.updateMaxEnvironmentRevisionId(environment),
-                ),
+                this.maxRevisionId
+                    .keys()
```
nunogois: Same as: https://github.com/Unleash/unleash/pull/10542/files#r2301421007

Try:

```ts
[...this.maxRevisionId.keys()].map((environment) => ....
```

instead.

src/lib/features/feature-toggle/configuration-revision-service.ts
gastonfournier: Why 
```
${[...this.maxRevisionId.keys()]}
```
and not just:
```
${this.maxRevisionId.keys()}
```

src/lib/features/feature-toggle/configuration-revision-service.ts
```diff
@@ -98,17 +99,12 @@ export default class ConfigurationRevisionService extends EventEmitter {
...
-            if (this.flagResolver.isEnabled('debugEtag')) {
-                this.logger.info(
-                    `[etag] Updating feature configuration with new revision Id ${revisionId} and all envs: ${Object.keys(this.maxRevisionId).join(', ')}`,
-                );
-            } else {
-                this.logger.debug(
-                    `Updating feature configuration with new revision Id ${revisionId} and all envs: ${Object.keys(this.maxRevisionId).join(', ')}`,
-                );
-            }
+            const knownEnvironments = [...this.maxRevisionId.keys()];
```
nunogois, line 102: Yep, nice!

nunogois (approved):
LGTM!

src/lib/features/feature-toggle/configuration-revision-service.ts
gastonfournier: Ok, maxRevisionId.keys returns a MapIterator and it doesn't have the map method

PR #10546, opened 2025-08-27
Show tooltip on group instead of single graph bar

thomasheartman:
The tooltip should show you the ratio of the group and the numbers for both archived and created.

The tooltip's position is averaged between all the bars in the group.

image

This still throws the same errors as before. Not sure exactly what's going on with the custom tooltip, but I'll investigate it in a follow-up.

Closes 1-4017.


frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -144,6 +141,9 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
                 data={data}
                 options={{
                     responsive: true,
+                    interaction: {
+                        mode: 'index',
```
thomasheartman, line 145: Shows tooltips for groups instead of individual data points.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -156,7 +157,7 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
                         },
                         tooltip: {
                             enabled: false,
```
thomasheartman, line 159: Places the tooltip on the average position of the group bars instead of using the nearest bar. Using `nearest` here would give make the tooltip move from one bar to the next inside the same group if you moved your cursor from one to the next. Instead, it's now on the average position regardless of which one you hover over.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -144,6 +141,10 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
                 data={data}
                 options={{
                     responsive: true,
+                    interaction: {
+                        mode: 'index',
+                        intersect: false,
```
thomasheartman, line 146: Makes it so that the tooltip shows regardless of whether you're directly over the bar or not. This prevents the tooltip from disappearing when the cursor is in the space between the bars (grouped or otherwise) and allows you to get tooltip info even if the bar is hard to pinpoint.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveRatioTooltip.tsx
```diff
@@ -1,24 +1,38 @@
...
     display: 'flex',
     justifyContent: 'space-between',
     alignItems: 'center',
-    marginBottom: theme.spacing(0.5),
+    whiteSpace: 'nowrap',
+    fontSize: theme.typography.body2.fontSize,
+    '::before': {
+        content: '" "',
+        display: 'inline-block',
+        height: '.65rem',
```
thomasheartman, line 27: This is the right height for the circles. I don't know why. I wish there was a better size that fit better, but I couldn't find it.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveRatioTooltip.tsx
```diff
@@ -1,24 +1,38 @@
...
+const DataRow = styled('div', {
+    shouldForwardProp: (prop) => prop !== 'dataType',
+})<{ dataType: 'archived' | 'created' }>(({ theme, dataType }) => ({
     display: 'flex',
     justifyContent: 'space-between',
     alignItems: 'center',
-    marginBottom: theme.spacing(0.5),
+    whiteSpace: 'nowrap',
+    fontSize: theme.typography.body2.fontSize,
+    '::before': {
```
thomasheartman, line 24: Place the colored circle before the data point. Doing it as pure CSS means it doesn't show up for screen readers etc, which is what we want in this case. It also isn't text.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveRatioTooltip.tsx
```diff
@@ -28,30 +42,14 @@ interface CreationArchiveRatioTooltipProps {
...
-        return null;
-    }
-
-    const rawData = ratioDataPoint.raw as WeekData;
-
-    if (!rawData) {
-        return null;
-    }
-
+    const rawData = tooltip.dataPoints[0].raw as WeekData;
```
thomasheartman, line 49: Isn't this the same in all the raw data points? Looked like it to me, and the numbers seem right?

sjaanus: Yep possibly, if it works, then pretty sweet.

sjaanus (approved):
LG

PR #10551, opened 2025-08-27
feat: impact metrics migration

kwasniew:
## About the changes

We want to store impact metrics configuration per flag. Previously we were storing impact metrics in global settings. This new table will accommodate both metrics for a flag and global ones (null feature field).
I decoded to model config as JSONB since we're not going to do any joins and we need to store data of this shape:
```
{
        id: string;
        selectedSeries: string;
        selectedRange: 'hour' | 'day' | 'week' | 'month';
        beginAtZero: boolean;
        aggregationMode: string;
        selectedLabels: Record<string, string[]>;
        title?: string;
    }
```

This is basically CRUD for this screen
Screenshot 2025-08-27 at 16 03 03

### DIFF
--- a/src/migrations/20250827135502-add-impact-metrics.js
+++ b/src/migrations/20250827135502-add-impact-metrics.js
@@ -1,0 +1,23 @@
+
+exports.up = (db, callback) => {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS impact_metrics(
+            id TEXT NOT NULL PRIMARY KEY,
+            feature VARCHAR(255),
+            config JSONB NOT NULL,
+            FOREIGN KEY (feature) REFERENCES features(name) ON DELETE CASCADE
+        );
+        `,
+        callback,
+    );
+};
+
+exports.down = (db, callback) => {
+    db.runSql(
+        `
+        DROP TABLE IF EXISTS impact_metrics;
+        `,
+        callback,
+    );
+};


src/migrations/20250827135502-add-impact-metrics.js
```diff
@@ -0,0 +1,23 @@
+
+exports.up = (db, callback) => {
+    db.runSql(
+        `
+        CREATE TABLE IF NOT EXISTS impact_metrics(
+            id TEXT NOT NULL PRIMARY KEY,
```
kwasniew, line 6: it will be ulid generated by code

src/migrations/20250827135502-add-impact-metrics.js
```diff
@@ -0,0 +1,23 @@
...
+            feature VARCHAR(255),
```
kwasniew, line 7: can be null for global impact metrics

src/migrations/20250827135502-add-impact-metrics.js
```diff
@@ -0,0 +1,23 @@
...
+            config JSONB NOT NULL,
```
kwasniew, line 8: implicit schema for CRUD config

src/migrations/20250827135502-add-impact-metrics.js
```diff
@@ -0,0 +1,23 @@
...
+            FOREIGN KEY (feature) REFERENCES features(name) ON DELETE CASCADE
```
kwasniew, line 9: if related to feature it has to be a valid feature

PR #10552, opened 2025-08-27
fix: consider both client and backend token types from the DB

gastonfournier:
## About the changes
In the previous fix: https://github.com/Unleash/unleash/pull/10543, we made sure client token types were displayed in the UI. Here, we're also making sure that the Backend token types are displayed as well.

1. A test validates that if a backend token exists in the db it will be returned in the API response. 
2. The UI has been adapted to also consider backend token types

### DIFF
--- a/src/lib/routes/admin-api/api-token.ts
+++ b/src/lib/routes/admin-api/api-token.ts
@@ -44,6 +44,7 @@ import {
 import type { FrontendApiService } from '../../features/frontend-api/frontend-api-service.js';
 import { OperationDeniedError } from '../../error/index.js';
 import type { CreateApiTokenSchema } from '../../internals.js';
+import type { IUserPermission } from '../../server-impl.js';
 
 interface TokenParam {
     token: string;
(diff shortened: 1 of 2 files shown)


src/lib/routes/admin-api/api-token.ts
```diff
@@ -64,33 +64,6 @@ export const tokenTypeToCreatePermission: (tokenType: ApiTokenType) => string =
...
-            ].includes(permission)
-        ) {
-            return ApiTokenType.CLIENT;
-        } else if (ADMIN === permission) {
-            return ApiTokenType.ADMIN;
-        } else {
-            return undefined;
-        }
-    };
-
```
nunogois, line 93: I think we're losing something here. 

Previously we said that anyone with any of the CRUD permissions for that token could read them, but now we specifically require the READ permission.

src/lib/routes/admin-api/api-token.ts
```diff
@@ -419,23 +392,33 @@ export class ApiTokenController extends Controller {
...
+                        (type === ApiTokenType.FRONTEND &&
+                            permission === READ_FRONTEND_API_TOKEN) ||
+                        ([ApiTokenType.CLIENT, ApiTokenType.BACKEND].includes(
+                            type,
+                        ) &&
+                            permission === READ_CLIENT_API_TOKEN),
+            );
+
+        const accessibleTokens = allTokens.filter((token) =>
+            readTokenFns.some((canRead) => canRead(token.type)),
```
nunogois: Personally I don't find this much more readable than before 😕 Something tells me we may be able to do better.

src/lib/routes/admin-api/api-token.ts
gastonfournier: Good catch! I completely overlooked this! Added them back in the last commit

nunogois (approved):
Awesome!

PR #10561, opened 2025-08-28
fix: backend tokens mapped to client tokens

gastonfournier:
Edge does not support backend token type, so until then, we need to keep returning backend token type as if they're client token types.


chriswk (approved):
LGTM

PR #10567, opened 2025-08-28
fix: incorrect current calculation of ratio

thomasheartman:
Looks like the sinner was looking at the last entry of each data list, when they are sorted by most recent first.

### DIFF
--- a/frontend/src/component/insights/componentsStat/CreationArchiveStats/CreationArchiveStats.tsx
+++ b/frontend/src/component/insights/componentsStat/CreationArchiveStats/CreationArchiveStats.tsx
@@ -13,10 +13,7 @@ function getCurrentArchiveRatio(
         InstanceInsightsSchema['creationArchiveTrends']
     >,
 ) {
-    if (
-        !groupedCreationArchiveData ||
-        Object.keys(groupedCreationArchiveData).length === 0
-    ) {
+    if (!groupedCreationArchiveData) {
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


frontend/src/component/insights/componentsStat/CreationArchiveStats/CreationArchiveStats.tsx
```diff
@@ -13,18 +13,15 @@ function getCurrentArchiveRatio(
...
+    if (!groupedCreationArchiveData) {
         return 0;
     }
 
     let totalArchived = 0;
     let totalCreated = 0;
 
     Object.values(groupedCreationArchiveData).forEach((projectData) => {
-        const latestData = projectData[projectData.length - 1];
+        const latestData = projectData[0];
```
thomasheartman, line 24: The data are sorted by date descending as far as I can tell, so the most recent stats will always be the first entry in each project.

frontend/src/component/insights/componentsStat/CreationArchiveStats/CreationArchiveStats.tsx
```diff
@@ -13,18 +13,15 @@ function getCurrentArchiveRatio(
         InstanceInsightsSchema['creationArchiveTrends']
     >,
 ) {
-    if (
-        !groupedCreationArchiveData ||
-        Object.keys(groupedCreationArchiveData).length === 0
-    ) {
+    if (!groupedCreationArchiveData) {
```
thomasheartman, line 16: If there are no keys, the rest of this function is essentially a no-op (a couple variable assignments), and this extra check tripped me up a bit. I suggest removing it because it's redundant, but happy to take this out too.

sjaanus (approved):

PR #10569, opened 2025-08-28
chore: allows you to add a flag from the unknown flags list

nunogois:
https://linear.app/unleash/issue/2-3826/allow-you-to-create-the-flag-from-the-unknown-flags-list

Allows you to add a flag from the unknown flags list.

image

image

### DIFF
--- a/frontend/src/component/unknownFlags/UnknownFlagsActionsCell.tsx
+++ b/frontend/src/component/unknownFlags/UnknownFlagsActionsCell.tsx
@@ -6,6 +6,8 @@ import type { UnknownFlag } from './hooks/useUnknownFlags.js';
 import { Link } from 'react-router-dom';
 import useProjects from 'hooks/api/getters/useProjects/useProjects.js';
 import { DEFAULT_PROJECT_ID } from 'hooks/api/getters/useDefaultProject/useDefaultProjectId.js';
+import AccessContext from 'contexts/AccessContext.js';
+import { useContext } from 'react';
 
 const StyledBox = styled(Box)(() => ({
     display: 'flex',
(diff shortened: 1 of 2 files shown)


frontend/src/component/unknownFlags/UnknownFlagsActionsCell.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const UnknownFlagsActionsCell = ({
+    unknownFlag,
+}: IUnknownFlagsActionsCellProps) => {
+    const { projects } = useProjects();
+
+    const project = projects.find(({ id }) => id === DEFAULT_PROJECT_ID)
+        ? DEFAULT_PROJECT_ID
+        : projects[0]?.id || DEFAULT_PROJECT_ID;
```
gastonfournier: I would get the default or first project, and keep the project object (to use it below). In the future it would be nice to have a selector for the project, but not needed now because you can move the feature later. The thing would be if the current user has or not permissions to create a flag in that project...
```suggestion
    const project = projects.find(({ id }) => id === DEFAULT_PROJECT_ID);
```

frontend/src/component/unknownFlags/UnknownFlagsActionsCell.tsx
```diff
@@ -0,0 +1,43 @@
...
+                to={`/projects/${project}?create=true&name=${unknownFlag.name}`}
+                permission={CREATE_FEATURE}
+                tooltipProps={{
+                    title: 'Create feature flag',
+                }}
+            >
+                <Add />
+            </PermissionIconButton>
+        </StyledBox>
+    );
```
gastonfournier:
```suggestion
    return !project? <></> : (
        <StyledBox>
            <PermissionIconButton
                component={Link}
                data-loading
                to={`/projects/${project.id}?create=true&name=${unknownFlag.name}`}
                permission={CREATE_FEATURE}
                tooltipProps={{
                    title: "Create feature flag in ${project.name}",
                }}
            >
                <Add />
            </PermissionIconButton>
        </StyledBox>
    );
```

ivarconr (approved):

frontend/src/component/unknownFlags/UnknownFlagsActionsCell.tsx
nunogois: There's actually a selector on the flag creation modal, but only if you're on a paid plan.

image

How would you feel about this instead? https://github.com/Unleash/unleash/pull/10569/commits/fe2ef2ac8f30b37dcba29c1d1cee27cf87a482cf

Basically we default to the "default" project. If not found, we grab the first.

However if you don't have the `CREATE_FEATURE` permission in it, we look for the first project where you do, so we can suggest it.

frontend/src/component/unknownFlags/UnknownFlagsActionsCell.tsx
nunogois: I think I disagree with this one. 

You're either in a free or paid plan.
 - Free: You can only create the flag in the default/first project, so why specify it?
 - Paid: You can change the project where you will create the flag, so why specify our suggestion like it's a limiting factor?

My suggestion: https://github.com/Unleash/unleash/pull/10569#discussion_r2309718570

gastonfournier: > You can change the project where you will create the flag

Awesome! I'm all in for this idea

PR #10575, opened 2025-08-29
fix: Show 100% instead of Infinity% if you've not created any flags

thomasheartman:
JS gives you positive infinity if you divide a positive number by 0, which isn't very helpful here. Instead, let's show 100%. If you divide 0 by 0, then you get NaN, which we also need to handle explicitly because it doesn't work with math.min.

Fixes 1-4033.


thomasheartman:
Eh, might be 🤷🏼 I'll extract it and pop a test for it. Good shout.

PR #10586, opened 2025-09-01
fix: lifecycle metrics will now be posted across all environments

sjaanus:


src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -88,18 +90,25 @@ export class FeatureLifecycleService {
...
-                    for (const [environment, metrics] of Object.entries(
-                        groupedByEnvironment,
-                    )) {
-                        const features = metrics.map(
-                            (metric) => metric.featureName,
-                        );
-                        await this.featuresReceivedMetrics(
-                            features,
-                            environment,
+                    if (this.flagResolver.isEnabled('optimizeLifecycle')) {
```
sjaanus, line 93: We will keep the new logic behind flag until its stable.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -88,18 +90,25 @@ export class FeatureLifecycleService {
...
+                        await this.handleBulkMetrics(events);
+                    } else {
+                        const groupedByEnvironment = groupBy(
```
sjaanus, line 96: We do no changes to existing logic.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,134 @@ export class FeatureLifecycleService {
...
+        }
+    }
+
+    private extractUniqueEnvironmentsAndFeatures(events: IClientMetricsEnv[]) {
+        const environments = [...new Set(events.map((e) => e.environment))];
+        const allFeatures = [...new Set(events.map((e) => e.featureName))];
+        return {environments, allFeatures};
+    }
+
+    private async buildEnvironmentMap(): Promise<Map<string, IEnvironment>> {
```
sjaanus, line 217: We do one database call to gather all `environments`.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,134 @@ export class FeatureLifecycleService {
...
+            }
+        } catch (e) {
+            this.logger.warn(
+                `Error handling bulk metrics for ${events.length} events`,
+                e,
+            );
+        }
+    }
+
+    private extractUniqueEnvironmentsAndFeatures(events: IClientMetricsEnv[]) {
```
sjaanus, line 211: First from the `metrics`, we gather unique `environments` and `features`.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,134 @@ export class FeatureLifecycleService {
...
+        const allFeatures = [...new Set(events.map((e) => e.featureName))];
+        return {environments, allFeatures};
+    }
+
+    private async buildEnvironmentMap(): Promise<Map<string, IEnvironment>> {
+        const allEnvs = await this.environmentStore.getAll();
+        return new Map(allEnvs.map((env) => [env.name, env]));
+    }
+
+    private async buildFeatureEnvironmentMap(allFeatures: string[]) {
```
sjaanus, line 222: We create a Map to get every environment configuration for every flag.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,134 @@ export class FeatureLifecycleService {
...
+        envMap: Map<string, IEnvironment>,
+        featureEnvMap: Map<string, Map<string, IFeatureEnvironment>>,
+    ): Array<{ feature: string; stage: 'pre-live' | 'live' }> {
+        const allStagesToInsert: Array<{
+            feature: string;
+            stage: 'pre-live' | 'live';
+        }> = [];
+
+        for (const environment of environments) {
+            const env = envMap.get(environment);
```
sjaanus, line 255: First we check, if in database environment does not exist, we skip the metric.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,134 @@ export class FeatureLifecycleService {
...
+            if (!env) continue;
+
+            const envFeatures = this.getFeaturesForEnvironment(
```
sjaanus, line 258: We find all features that were sent with metrics under this environment.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,134 @@ export class FeatureLifecycleService {
...
+                events,
+                environment,
+            );
+            allStagesToInsert.push(...this.createPreLiveStages(envFeatures));
```
sjaanus, line 262: We add these stages automatically to the list.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,134 @@ export class FeatureLifecycleService {
...
+
+            if (env.type === 'production') {
+                const enabledFeatures = this.getEnabledFeaturesForEnvironment(
```
sjaanus, line 265: For production, we filter in only features that are enabled in that environment.

src/lib/features/feature-lifecycle/feature-lifecycle-service.test.ts
```diff
@@ -95,10 +104,251 @@ test('can insert and read lifecycle stages', async () => {
...
+            appName: 'test-app',
+            timestamp: new Date(),
+            yes: 150,
+            no: 30,
+        },
+    ];
+
+    eventBus.emit(CLIENT_METRICS_ADDED, bulkMetricsEvent);
+
+    await new Promise((resolve) => setTimeout(resolve, 50));
```
sjaanus: This is semi bad, but I could not find a way to get it working else.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,137 @@ export class FeatureLifecycleService {
...
+    private getFeaturesForEnvironment(
+        events: IClientMetricsEnv[],
+        environment: string,
+    ): string[] {
+        return events
+            .filter((e) => e.environment === environment)
+            .map((e) => e.featureName);
+    }
+
+    private createPreLiveStages(
```
sjaanus, line 288: Mapper function for inserting pre live data.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,137 @@ export class FeatureLifecycleService {
...
+        features: string[],
+    ): Array<{ feature: string; stage: 'pre-live' }> {
+        return features.map((feature) => ({
+            feature,
+            stage: 'pre-live' as const,
+        }));
+    }
+
+    private createLiveStages(
```
sjaanus, line 297: Mapper function for inserting live data.

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
gastonfournier: Can we memoize it?

src/lib/features/feature-lifecycle/feature-lifecycle-service.ts
```diff
@@ -169,6 +178,137 @@ export class FeatureLifecycleService {
...
     }
 
+    private async handleBulkMetrics(events: IClientMetricsEnv[]) {
+        try {
+            const { environments, allFeatures } =
+                this.extractUniqueEnvironmentsAndFeatures(events);
+            const envMap = await this.buildEnvironmentMap();
+            const featureEnvMap =
+                await this.buildFeatureEnvironmentMap(allFeatures);
+            const allStagesToInsert = this.determineLifecycleStages(
```
gastonfournier, line 191: Do we need to do it on handleBulkMetrics? Can we do it using a job that executes on a schedule? I'm thinking that if you have 1000 clients sending metrics this would be expensive and I think we should keep this calculation off the critical path.

sjaanus: We are not doing it on every client send. We do it once every 5 seconds. So it does not matter how many clients really.

gastonfournier: oh, right, because this is done on bulkMetrics endpoint, which holds data in memory, and only after 5 seconds from a schedule it stores them into the DB. Maybe a simple workaround would be to store every 10 seconds instead of 5.

PR #10589, opened 2025-09-01
chore: Use custom highlighter for archive:creation chart

thomasheartman:
Uses the same custom highlighter that we've used for the other analytics charts, but customizes it for the archive : creation ratio chart. Also removes the color change on hover from the bars.

Because this plugin (or a version of it) is also used in the network traffic graphs, I've consolidated all the uses into a more configurable one (because I was mistakenly working in the wrong file).

Before:
image

After:

image

### DIFF
--- a/frontend/src/component/admin/network/NetworkTrafficUsage/BackendConnections.tsx
+++ b/frontend/src/component/admin/network/NetworkTrafficUsage/BackendConnections.tsx
@@ -11,7 +11,6 @@ import {
 } from '@mui/material';
 import { PeriodSelector } from './PeriodSelector.tsx';
 import { Bar } from 'react-chartjs-2';
-import { customHighlightPlugin } from 'component/common/Chart/customHighlightPlugin';
 import { getChartLabel } from './chart-functions.ts';
 import { useConsumptionStats } from './hooks/useStats.ts';
 import { StyledBox, TopRow } from './SharedComponents.tsx';
@@ -27,6 +26,7 @@ import {
 import annotationPlugin from 'chartjs-plugin-annotation';
 import { useChartDataSelection } from './hooks/useChartDataSelection.ts';
 import HelpOutline from '@mui/icons-material/HelpOutline';
+import { networkTrafficUsageHighlightPlugin } from './networkTrafficUsageHighlightPlugin.ts';
 
 const ConnectionExplanationBox = styled(Box)(({ theme }) => ({
     display: 'flex',
@@ -73,7 +73,7 @@ export const BackendConnections: FC = () => {
                         </TopRow>
                         <Bar
                             data={chartData}
-                            plugins={[customHighlightPlugin()]}
+                            plugins={[networkTrafficUsageHighlightPlugin]}
                             options={options}
                             aria-label={getChartLabel(chartDataSelection)}
                         />
--- a/frontend/src/component/common/Chart/customHighlightPlugin.ts
+++ b/frontend/src/component/common/Chart/customHighlightPlugin.ts
@@ -1,8 +1,34 @@
 import type { Chart } from 'chart.js';
 
-export const customHighlightPlugin = (bottomOverflow = 34) => ({
+type CustomHighlightPluginOptions = {
+    width?: number | ((chart: Chart) => number);
+    maxHighlightOpacity?: number;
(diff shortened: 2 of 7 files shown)


frontend/src/component/admin/network/NetworkTrafficUsage/BackendConnections.tsx
```diff
@@ -73,7 +73,7 @@ export const BackendConnections: FC = () => {
                         </TopRow>
                         <Bar
                             data={chartData}
-                            plugins={[customHighlightPlugin()]}
+                            plugins={[networkTrafficUsageHighlightPlugin]}
```
thomasheartman, line 76: The three charts in network traffic all share the same config here, so it seemed sensible to extract that configuration to a separate spot that they can all reference.

frontend/src/component/common/Chart/customHighlightPlugin.ts
```diff
@@ -1,8 +1,34 @@
 import type { Chart } from 'chart.js';
 
-export const customHighlightPlugin = (bottomOverflow = 34) => ({
+type CustomHighlightPluginOptions = {
+    width?: number | ((chart: Chart) => number);
+    maxHighlightOpacity?: number;
+    bottomOverflow?: number;
```
thomasheartman, line 6: These are the current configurations we're doing. Can be expanded later.

frontend/src/component/common/Chart/customHighlightPlugin.ts
```diff
@@ -1,8 +1,34 @@
...
+// Vertical line on the hovered chart, filled with gradient. Highlights a section of a chart when you hover over datapoints
+export const customHighlightPlugin = (
+    options?: CustomHighlightPluginOptions,
+) => ({
     id: 'customLine',
     beforeDraw: (chart: Chart) => {
+        const {
+            width = categoryWidth,
+            maxHighlightOpacity = 0.12,
+            bottomOverflow = 0,
```
thomasheartman, line 27: These seem like sensible defaults to me:
- the highlight should typically be the same width as the category (either a single bar or a group of bars depending on your data source)
- the highlight opacity is the color we'd used everywhere up until now
- no bottom overflow (stop where the chart is).

frontend/src/component/common/Chart/customHighlightPlugin.ts
```diff
@@ -12,23 +38,27 @@ export const customHighlightPlugin = (bottomOverflow = 34) => ({
...
                 yAxis.bottom - yAxis.top + bottomOverflow,
-                5,
-            );
-            ctx.fill();
+            ];
+            if (bottomOverflow) {
+                ctx.roundRect(...args, 5);
+                ctx.fill();
+            } else {
+                ctx.fillRect(...args);
```
thomasheartman, line 60: These cases are slightly different. If there is overflow, draw a rounded rectangle. If not, draw a perfect rectangle. Also, the drawing functions have slightly different behavior, which causes a weird full overlay at times with roundRect on the analytics page.

frontend/src/component/insights/components/LineChart/LineChartComponent.tsx
```diff
@@ -126,7 +97,7 @@ const LineChartComponent: FC<{
                 key={cover ? 'cover' : 'chart'}
                 options={options}
                 data={data}
-                plugins={[customHighlightPlugin]}
+                plugins={[customHighlightPlugin({ width: 26 })]}
```
thomasheartman, line 100: The analytics line charts have a set width in their custom function, so we'll retain that here.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -203,6 +208,11 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
                 options={options}
                 height={100}
                 width={250}
+                plugins={[
+                    customHighlightPlugin({
+                        maxHighlightOpacity: 0.24,
```
thomasheartman, line 213: The default highlight opacity was much too subtle on this graph, so we had to up the max opacity. Here's what it looks like with 0.12:

Image

PR #10590, opened 2025-09-01
chore: one unknown flag per row

nunogois:
https://linear.app/unleash/issue/2-3833/show-one-flag-name-per-row

Groups unknown flags by flag name, showing a single flag name per row. This greatly simplifies the way we show unknown flags.

Just to be safe, we're limiting the app names we're showing to 20, and environments per app name to 10.

Required some plumbing.

### Basic example
image

### App name search example, with highlight

image

### DIFF
--- a/src/lib/features/metrics/unknown-flags/fake-unknown-flags-store.ts
+++ b/src/lib/features/metrics/unknown-flags/fake-unknown-flags-store.ts
@@ -1,25 +1,67 @@
 import type {
     IUnknownFlagsStore,
     UnknownFlag,
+    UnknownFlagReport,
     QueryParams,
 } from './unknown-flags-store.js';
 
 export class FakeUnknownFlagsStore implements IUnknownFlagsStore {
-    private unknownFlagMap = new Map<string, UnknownFlag>();
+    private unknownFlagMap = new Map<string, UnknownFlagReport>();
 
-    private getKey(flag: UnknownFlag): string {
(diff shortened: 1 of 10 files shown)


src/lib/features/metrics/unknown-flags/fake-unknown-flags-store.ts
```diff
@@ -1,25 +1,67 @@
...
 
-    async insert(flags: UnknownFlag[]): Promise<void> {
+    async insert(flags: UnknownFlagReport[]): Promise<void> {
         this.unknownFlagMap.clear();
         for (const flag of flags) {
             this.unknownFlagMap.set(this.getKey(flag), flag);
         }
     }
 
+    private groupFlags(flags: UnknownFlagReport[]): UnknownFlag[] {
```
daveleek, line 22: This is just for testing yeah?

daveleek (approved):
This seems good, we might end up getting asked about a way to retrieve a full report though. Do we have that anywhere?

src/lib/features/metrics/unknown-flags/fake-unknown-flags-store.ts
nunogois: Yeah, it's the fake store used for tests.

nunogois:
> This seems good, we might end up getting asked about a way to retrieve a full report though. Do we have that anywhere?

What do you mean by full report? You can always fetch the endpoint directly.

PR #10591, opened 2025-09-01
chore: add unknown flags link in flags overview

nunogois:
https://linear.app/unleash/issue/2-3834/add-link-to-unknown-flags-in-flags-overview

Adds an "Unknown flags" link on the top right corner of the flags overview page.

image


frontend/src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
```diff
@@ -272,7 +280,7 @@ export const FeatureToggleListTable: FC = () => {
                                     });
                                 }}
                             >
-                                View archive
+                                Archived flags
```
nunogois, line 287: Noticed the designs had "Archived flags" instead of "View archive". This feels intentional so it's more consistent with the new link.

PR #10592, opened 2025-09-01
chore: make create feature flag button in unknown flags a text button

nunogois:
https://linear.app/unleash/issue/2-3835/make-create-feature-flag-a-variant=text-button-instead-of-an-icon

Makes "create feature flag" button in unknown flags a text button.

image

### DIFF
--- a/frontend/src/component/unknownFlags/UnknownFlagsTable.tsx
+++ b/frontend/src/component/unknownFlags/UnknownFlagsTable.tsx
@@ -122,7 +122,7 @@ export const UnknownFlagsTable = () => {
                         suggestedProject={suggestedProject}
                     />
                 ),
-                width: 160,
+                width: 120,
                 disableSortBy: true,
             },
             // Always hidden -- for search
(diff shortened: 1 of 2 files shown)


frontend/src/component/unknownFlags/UnknownFlagsTable.tsx
```diff
@@ -43,6 +47,23 @@ export const UnknownFlagsTable = () => {
...
+        if (!hasAccess(CREATE_FEATURE, project?.id)) {
+            for (const proj of projects) {
+                if (hasAccess(CREATE_FEATURE, proj.id)) {
+                    project = proj;
+                    break;
+                }
+            }
+        }
+        return project;
+    }, [projects, hasAccess]);
```
nunogois, line 65: Feels smarter to calculate the suggested project once at the parent level, memoize it, and pass it as a prop.

daveleek: I'm wondering a bit about this, would this potentially suggest projects the user can't create in and not suggest projects the user CAN create in?

daveleek: Disregard, I read the rest of it again

daveleek (approved):
LGTM!

PR #10593, opened 2025-09-02
chore: startup investigation

gastonfournier:
This was a short investigation into the startup performance. These are the findings:
1. Server binds in ~2.9 s, yet dev workflow shows ~20 s wall (includes build + watcher). 
2. Startup profiler highlighted large async churn and heavy synchronous module initialization before the first listen.

Some suggestions (from ChatGPT that I think are relevant but I don't feel we'll gain much):
1. Migration gating: requires migration scans ~350 migration files and runs db-migrate logic every boot (FS + promise storm) before listen; no cached manifest or version fast path.
2. OpenAPI boot cost: OpenApiService constructs full schema and pulls in @wesleytodd/openapi + swagger-parser stack even if docs are never hit during early life; validators compiled at runtime each boot.
3. Validation coupling: API route validation tightly coupled to heavy OpenApiService instance; cannot partially initialize (validators vs docs UI). Lazy-load swagger UI & parser only on /docs
4. Heavy module loads: date-fns root import (~275 ms chain) and openapi/swagger, unleash-client, knex stack each ~80–130 ms; some appear twice (path + bare specifier) causing duplicate parse/compile.
5. Promise churn: ~38k PROMISE async ops (massive summed duration) dominated by trivial then/new Promise chains from library initialization (Bluebird transitive usage, swagger, date-fns aggregation).
6. Repetitive runtime computation: Simple helpers (e.g. secondsToMilliseconds) imported from full date-fns bundle; no tree-shaking in CommonJS/ESM mixed environment.
7. Precompile validators: generate AJV compiled code at build


startup-report.txt
```diff
@@ -0,0 +1,402 @@
+Node v22.17.1 | PID 2141941
```
gastonfournier, line 1: Output of the report

startup-trace.json
gastonfournier, line 1: This file can be loaded in https://ui.perfetto.dev/ for a timeline but it's not really useful IMO

gastonfournier:
Close in favor of https://github.com/Unleash/unleash/pull/11599

PR #10601, opened 2025-09-02
chore: add unknown flag metric for unique flag names

nunogois:
https://linear.app/unleash/issue/2-3845/add-metric-for-unique-unknown-flag-names

Adds a new unknown flag metric for unique unknown flag names.

image

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -761,7 +761,12 @@ export function registerPrometheusMetrics(
 
     const unknownFlagsGauge = createGauge({
         name: 'unknown_flags',
-        help: 'Number of unknown flags reported in the last 24 hours, if any. Maximum of 1000.',
+        help: 'Number of unknown flag reports (name + app + env) in the last 24 hours, if any.',
+    });
+
+    const unknownFlagsUniqueNamesGauge = createGauge({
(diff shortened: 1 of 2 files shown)


src/lib/metrics.ts
```diff
@@ -1230,6 +1235,11 @@ export function registerPrometheusMetrics(
                 const unknownFlags = await stores.unknownFlagsStore.count();
                 unknownFlagsGauge.reset();
                 unknownFlagsGauge.set(unknownFlags);
+
+                const unknownFlagsUniqueNames =
+                    await stores.unknownFlagsStore.count({ unique: true });
```
gastonfournier, line 1240: Is it going to be one per flag or one per flag + env + app?

nunogois: We'll have both (check the screenshot in the PR description). This one in your comment only cares about unique unknown flag names.

nunogois: > LG! Nice to have only 1 gauge

Should we drop one of them?

gastonfournier: No, I think it's fine. I was giving a compliment that you did not add one metric per flag name

PR #10607, opened 2025-09-03
fix: flag resolver isEnabled for variant configs

kwasniew:
## About the changes

Internal flag resolver was not allowing to use external flag resolver when experimental flag was configured with an object config.

### DIFF
--- a/src/lib/util/flag-resolver.ts
+++ b/src/lib/util/flag-resolver.ts
@@ -49,7 +49,7 @@ export default class FlagResolver implements IFlagResolver {
         const exp = this.experiments[expName];
         if (exp) {
             if (typeof exp === 'boolean') return exp;
-            else return exp.enabled;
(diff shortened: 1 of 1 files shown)


src/lib/util/flag-resolver.ts
```diff
@@ -49,7 +49,7 @@ export default class FlagResolver implements IFlagResolver {
         const exp = this.experiments[expName];
         if (exp) {
             if (typeof exp === 'boolean') return exp;
-            else return exp.enabled;
+            else if (exp.enabled) return exp.enabled;
         }
         return this.externalResolver.isEnabled(expName, context);
```
kwasniew, line 54: this line was unreachable for object configs no matter what we put there. The fix was made for getVariant but this one was still buggy

nunogois (approved):

sjaanus (approved):
LG, now matches with getVariant logic.

PR #10609, opened 2025-09-03
chore: add IAM db auth support

nunogois:
https://linear.app/unleash/issue/2-3829/investigate-aws-iam-connection-support-for-unleash-docker

Adds AWS IAM DB connection support.

### DIFF
--- a/src/lib/db/db-pool.ts
+++ b/src/lib/db/db-pool.ts
@@ -3,19 +3,61 @@ import knexpkg from 'knex';
 const { knex } = knexpkg;
 import type { IUnleashConfig } from '../types/option.js';
 
+import { Signer } from '@aws-sdk/rds-signer';
+
 export function createDb({
     db,
     getLogger,
 }: Pick<IUnleashConfig, 'db' | 'getLogger'>): Knex {
     const logger = getLogger('db-pool.js');
+
+    const {
+        host,
+        port,
+        user,
+        database,
+        ssl,
+        applicationName,
+        password,
+        awsIamAuth,
+        awsRegion,
+        pool,
+    } = db;
+
+    let resolvedPassword: string | (() => Promise<string>) | undefined =
+        password;
+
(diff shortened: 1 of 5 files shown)


src/lib/db/db-pool.ts
```diff
@@ -3,19 +3,61 @@ import knexpkg from 'knex';
...
+        pool,
+    } = db;
+
+    let resolvedPassword: string | (() => Promise<string>) | undefined =
+        password;
+
+    if (awsIamAuth) {
+        if (!awsRegion) {
+            throw new Error(
+                'AWS_REGION is required when DATABASE_AWS_IAM=true',
```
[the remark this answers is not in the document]
nunogois: Uh...

src/lib/types/option.ts
```diff
@@ -20,7 +20,7 @@ export interface ISSLOption {
 
 export interface IDBOption {
     user: string;
```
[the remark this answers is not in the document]
nunogois: It's not like this was previously enforced or anything, we're just updating a type. I think this is fine as is.

src/lib/db/db-pool.ts
sastromskis, line 33: There's a chance that the Signer will be able to infer the region, so this might not be strictly needed. Not entirely sure about that, but other aws sdk componets do do that

nunogois: Should we try to remove it once we validate that this works?

src/lib/db/db-pool.ts
```diff
@@ -3,19 +3,61 @@ import knexpkg from 'knex';
...
+    }
+
+    const connection = {
+        host,
+        port,
+        user,
+        database,
+        ssl,
+        application_name: applicationName,
+        password: resolvedPassword,
```
sjaanus, line 53: Does knex accept async function as password?

PR #10420, opened 2025-07-28
chore: add feature configuration based on addConfiguration flag

nunogois:
https://linear.app/unleash/issue/2-3729/single-add-configuration-button-based-on-flag

Shows a single "Add configuration" button based on whether the new `addConfiguration` flag is enabled.

This button then shows our "Add configuration" modal which allows you to choose how to proceed in terms of your feature flag configuration.

Includes scouting.

### Single "Add configuration" button
image

### Modal
image

PR #10629, opened 2025-09-08
chore: add newStrategyModal flag

nunogois:
https://linear.app/unleash/issue/2-3865/add-newstrategymodal-feature-flag

Adds `newStrategyModal` feature flag.

The approach here is to duplicate the existing `FeatureStrategyMenuCards` into a `LegacyFeatureStrategyMenuCards`. We'll continue working on the `FeatureStrategyMenuCards` component while leaving `LegacyFeatureStrategyMenuCards` untouched. Once we're done with the implementation and remove the flag we can drop the legacy file.

I think it's easier to reduce our add strategy buttons to a single one right away (like we did with the `addConfiguration` flag in https://github.com/Unleash/unleash/pull/10420). This allows us to focus on the end result instead of having to implement things like "clicking the 'add release template' button should show the modal filtered to only release templates".

image

image


gastonfournier (approved):
Agree, having 2 buttons, one to add release templates and another to add strategies, complicates things unnecessarily, given we want to eliminate the multiple buttons.

PR #10633, opened 2025-09-08
chore: new add strategy modal design

nunogois:
https://linear.app/unleash/issue/2-3870/new-modal-design

New "add strategy" modal base design.

We're still missing a few details, like the new card design and the filters at the top, but those will come in follow-up PRs.

### New layout

image

### Default strategy tooltip, with a link to the default strategy settings

image

### Default strategy with a custom title

image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
@@ -1,68 +1,51 @@
-import { Link, styled, Typography, Box, IconButton } from '@mui/material';
+import { styled, Typography, Box, IconButton, Button } from '@mui/material';
 import { useStrategies } from 'hooks/api/getters/useStrategies/useStrategies';
 import { FeatureStrategyMenuCard } from '../FeatureStrategyMenuCard/FeatureStrategyMenuCard.tsx';
 import { useReleasePlanTemplates } from 'hooks/api/getters/useReleasePlanTemplates/useReleasePlanTemplates';
 import { FeatureReleasePlanCard } from '../FeatureReleasePlanCard/FeatureReleasePlanCard.tsx';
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCards.tsx
```diff
@@ -1,72 +1,55 @@
...
-interface IFeatureStrategyMenuCardsProps {
-    projectId: string;
-    featureId: string;
-    environmentId: string;
-    onlyReleasePlans: boolean;
-    onAddReleasePlan: (template: IReleasePlanTemplate) => void;
-    onReviewReleasePlan: (template: IReleasePlanTemplate) => void;
-    onClose: () => void;
-}
+const RELEASE_TEMPLATE_LIMIT = 5;
```
gastonfournier: I think we have it from resourcesLimits, we should use it from there. I think we already have a use for this type of limit

nunogois: I don't think we should use resource limits here. The idea is that we only show you a limited number of release templates if you have more than that number, as to not be overwhelmed visually. So in reality this is only relevant if you already have a higher resource limit for release templates, as that's the only way you would have more than 5 and have the option to "see all release templates".

gastonfournier: Ok, so it's just a coincidence release template (visual) limit is the same as the release template limit. If anything, to avoid confusion, I'd suggest renaming this constant to `RELEASE_TEMPLATES_DISPLAY_LIMIT` or something similar

nunogois: Agreed. Addressed in https://github.com/Unleash/unleash/pull/10633/commits/726d1ab86e6cbc0cc5c236ab558b9f89cd308243

PR #10634, opened 2025-09-09
fix: limit reset when no pagination bar

kwasniew:
## About the changes

Allow only predefined pagination limits even when no pagination bar.
Previously we were enforcing page limit only when pagination bar was enabled.

### DIFF
--- a/frontend/src/component/common/Table/PaginationBar/PaginationBar.tsx
+++ b/frontend/src/component/common/Table/PaginationBar/PaginationBar.tsx
@@ -1,5 +1,4 @@
 import type React from 'react';
-import { useEffect } from 'react';
 import { Box, Typography, Button, styled } from '@mui/material';
 import { ConditionallyRender } from '../../ConditionallyRender/ConditionallyRender.tsx';
 import { ReactComponent as ArrowRight } from 'assets/icons/arrowRight.svg';
@@ -61,12 +60,6 @@ export const PaginationBar: React.FC<PaginationBarProps> = ({
     fetchNextPage,
     setPageLimit,
 }) => {
-    useEffect(() => {
-        if (![25, 50, 75, 100].includes(pageSize)) {
-            setPageLimit(25);
-        }
-    }, [pageSize]);
-
     const itemRange =
         totalItems !== undefined && pageSize && totalItems > 1
             ? `${pageIndex * pageSize + 1}-${Math.min(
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/usePersistentTableState.ts
```diff
@@ -51,6 +51,19 @@ export const usePersistentTableState = <T extends QueryParamConfigMap>(
         return reorderObject(tableState, [...searchParams.keys()]);
     }, [searchParams, tableState, reorderObject]);
 
+    useEffect(() => {
```
kwasniew, line 54: centralized limit setting to predefined values. even when no pagination bar is included this will work

frontend/src/utils/safeNumberParam.ts
```diff
@@ -0,0 +1,9 @@
+import { encodeNumber, decodeNumber } from 'serialize-query-params';
+
+export const SafeNumberParam = {
```
kwasniew, line 6: default NumberParam parser from our library didn't handle NaN as we expected

frontend/src/component/project/Project/PaginatedProjectFeatureToggles/useProjectFeatureSearch.ts
```diff
@@ -28,8 +28,8 @@ export const useProjectFeatureSearch = (
     refreshInterval = 15 * 1000,
 ) => {
     const stateConfig = {
-        offset: withDefault(NumberParam, 0),
-        limit: withDefault(NumberParam, DEFAULT_PAGE_LIMIT),
+        offset: withDefault(SafeNumberParam, 0),
```
kwasniew, line 31: withDefault returns default when null/undefined is returned from SafeNumberParam. Previously on invalid string we were not getting 0 because param parser was returning NaN

frontend/src/component/common/Table/PaginationBar/PaginationBar.tsx
```diff
@@ -61,12 +60,6 @@ export const PaginationBar: React.FC<PaginationBarProps> = ({
     fetchNextPage,
     setPageLimit,
 }) => {
-    useEffect(() => {
```
kwasniew, line 64: this was wrong level to add the fixing logic

frontend/src/hooks/usePersistentTableState.ts
```diff
@@ -51,6 +51,19 @@ export const usePersistentTableState = <T extends QueryParamConfigMap>(
...
+        if (
+            tableState.limit &&
+            ![25, 50, 75, 100].includes(tableState.limit as number)
```
Tymek, line 57: Page sizes should not be magic numbers

frontend/src/utils/safeNumberParam.ts
Tymek: Let's link to where this is from in some way

```suggestion

/**
 * @see: https://github.com/pbeshai/use-query-params/issues/175#issuecomment-982791559
 */
export const SafeNumberParam = {
```

PR #10635, opened 2025-09-09
chore: add verbose logs to AWS IAM auth logic, add DB access checker

nunogois:
https://linear.app/unleash/issue/2-3874/add-verbose-logs-to-aws-iam-auth-logic-including-a-db-access-checker

Adds verbose logs to our new AWS IAM DB auth logic. Also adds a DB access checker that runs before our migrator and fails fast in case something is wrong with DB access.

This should not affect the regular auth path.

### DIFF
--- a/src/lib/db/aws-iam.ts
+++ b/src/lib/db/aws-iam.ts
@@ -16,7 +16,12 @@ export const getDBPasswordResolver = (db: IDBOption): PasswordResolver => {
             port: db.port,
             username: db.user,
         });
-        return async () => signer.getAuthToken();
+        return async () => {
+            console.log('[AWS RDS SIGNER] Getting token...');
+            const token = await signer.getAuthToken();
+            console.log(`[AWS RDS SIGNER] Got token: ${token}`);
+            return token;
+        };
     }
 
     return async () => db.password;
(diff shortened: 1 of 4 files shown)


src/lib/db/aws-iam.ts
```diff
@@ -16,7 +16,12 @@ export const getDBPasswordResolver = (db: IDBOption): PasswordResolver => {
             port: db.port,
             username: db.user,
         });
-        return async () => signer.getAuthToken();
+        return async () => {
+            console.log('[AWS RDS SIGNER] Getting token...');
+            const token = await signer.getAuthToken();
+            console.log(`[AWS RDS SIGNER] Got token: ${token}`);
```
sastromskis: Maybe we shouldn't expose the full token here. I think it's enough to know if we got it or not.

sastromskis (approved):
LG!

src/lib/db/aws-iam.ts
nunogois: Agreed, addressed here: https://github.com/Unleash/unleash/pull/10635/commits/6ff9ded4142323864fb5e851f0e12f527b5933ac

PR #10643, opened 2025-09-10
chore: add proxy-aware regional STS authentication for AWS IAM DB auth

nunogois:
https://linear.app/unleash/issue/2-3875/add-proxy-aware-regional-sts-authentication-for-aws-iam-db-auth

Adds proxy-aware regional STS authentication for AWS IAM DB auth.

We also added optional support for explicit IAM role assumption through the `DATABASE_AWS_ROLE_ARN` environment variable.

### DIFF
--- a/src/lib/__snapshots__/create-config.test.ts.snap
+++ b/src/lib/__snapshots__/create-config.test.ts.snap
@@ -34,6 +34,7 @@ exports[`should create default config 1`] = `
     "applicationName": "unleash",
     "awsIamAuth": false,
     "awsRegion": undefined,
+    "awsRoleArn": undefined,
     "database": "unleash_db",
     "disableMigration": false,
     "driver": "postgres",


src/lib/db/aws-iam.ts
```diff
@@ -1,25 +1,63 @@
...
-            region: db.awsRegion,
-            hostname: db.host,
-            port: db.port,
-            username: db.user,
-        });
-        return async () => signer.getAuthToken();
-    }
-
-    return async () => db.password;
+    if (!db.awsIamAuth) return async () => db.password;
```
gastonfournier, line 13: I'd be more specific:
```typescript
if (db.password) {
    return async () => db.password
}
if (db.awsIamAuth) {
...
}
```

nunogois: It's subjective, but I personally think early returning is more readable.

PR #10659, opened 2025-09-11
chore: update strategy cards in new add strategy modal

nunogois:
https://linear.app/unleash/issue/2-3867/update-strategy-cards

Updates the new "add strategy" modal strategy cards.

Followed the same logic as before where I took a snapshot of existing components and prefixed them with `Legacy`.

Does not include the new icons.

image

image

### DIFF
--- a/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCard/FeatureStrategyMenuCard.tsx
+++ b/frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCard/FeatureStrategyMenuCard.tsx
@@ -46,6 +46,7 @@ const StyledCard = styled('div', {
         backgroundColor: theme.palette.secondary.light,
         borderColor: theme.palette.secondary.border,
     }),
+    userSelect: 'none',
 }));
 
 const StyledCardContent = styled('div')(({ theme }) => ({


frontend/src/component/common/Truncator/Truncator.tsx
```diff
@@ -73,12 +75,15 @@ export const Truncator = ({
     const { title: tooltipTitle, ...otherTooltipProps } =
         overridableTooltipProps;
 
+    const defaultWordBreak = lines === 1 ? 'break-all' : 'break-word';
```
nunogois, line 78: Seems like I forgot to implement the `wordBreak` prop when I implemented the `Truncator` 🤦‍♂️

This also gives it what I believe is a sane default: If it's truncated to 1 line, break anywhere in our text. Otherwise try to preserve words.

See: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

Happy to split this into its own PR if you prefer.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCard/FeatureStrategyMenuCard.tsx
```diff
@@ -1,119 +1,109 @@
...
+                        {description}
+                    </Truncator>
+                )}
+            </StyledCardDescription>
+        </StyledCardContent>
+        <StyledCardActions className='cardActions'>
+            {children}
+        </StyledCardActions>
+    </StyledCard>
+);
```
nunogois, line 110: Unlike the previous `FeatureStrategyMenuCard` component, this implementation is a lot more generic, and we can re-use the same component for both strategies and release templates.

frontend/src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCards/FeatureStrategyMenuCardsDefaultStrategy.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+            setToastData({
+                text: 'Strategy applied',
+                type: 'success',
+            });
+        }
+        refetchFeature();
+
+        onClose();
+    };
```
nunogois, line 90: Most of this is taken out of: https://github.com/Unleash/unleash/blob/e79fa336471e44ed077a9356a36a2ff65dee536f/frontend/src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx#L145

PR #10660, opened 2025-09-12
chore: use paginated table for change request list

thomasheartman:
Adds a paginated table to the change request overview page and integrates it with the search API hook.

The current implementation still has some rough edges to work out, but it's getting closer. 

There's no sort buttons in this implementation. I've got it working on the side, but TS is complaining about types not matching up, so I'm spinning that out to a separate PR.

image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
@@ -34,7 +34,6 @@ const columnHelper = createColumnHelper<ChangeRequestSearchItemSchema>();
 const ChangeRequestsInner = () => {
     const { user } = useAuthUser();
 
-    // Check URL parameters directly to avoid double fetching
     const shouldApplyDefaults = useMemo(() => {
         const urlParams = new URLSearchParams(window.location.search);
         return (
@@ -53,7 +52,6 @@ const ChangeRequestsInner = () => {
         requestedApproverId: FilterItemParam,
     };
 
-    // Apply initial defaults if needed
     const initialState = shouldApplyDefaults
         ? {
               createdBy: {
@@ -67,7 +65,6 @@ const ChangeRequestsInner = () => {
         updateType: 'replaceIn',
     });
 
-    // Merge with initial state on first load only
     const effectiveTableState = useMemo(
         () => ({
             ...initialState,
@@ -111,7 +108,6 @@ const ChangeRequestsInner = () => {
                     },
                 }) => {
                     const features = getValue();
-                    // Convert string array to object array for FeaturesCell compatibility
                     const featureObjects = features.map((name: string) => ({
                         name,
                     }));


frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -1,285 +1,178 @@
...
-        createdAt: '2024-01-07T13:10:00Z',
-        environment: 'Production',
-        state: 'Scheduled',
-        schedule: {
-            scheduledAt: '2024-01-12T09:46:51+05:30',
-            status: 'failed',
-            reason: 'Mr Freeze',
-        },
-    },
-];
```
thomasheartman, line 150: Remove mock data (we hook up to the API now).

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -1,285 +1,178 @@
...
+                header: 'Environment',
+                meta: { width: '100px' },
+                cell: ({ getValue }) => <HighlightCell value={getValue()} />,
+            }),
+            columnHelper.accessor('state', {
+                id: 'Status',
+                header: 'Status',
+                meta: { width: '170px' },
+                cell: ({ getValue, row }) => (
+                    // @ts-expect-error (`globalChangeRequestList`) The schema (and query) needs to be updated
```
thomasheartman, line 147: We don't send data with the `schedule` property yet, so TS doesn't like this impl. I've got a workaround locally, but it will crash if run without the right props if you have any scheduled crs.

Luckily, this is all behind a flag 💁🏼

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -1,285 +1,178 @@
...
+            data,
+        }),
+    );
+
+    const bodyLoadingRef = useLoading(loading);
 
     return (
         <PageContent
-            isLoading={loading}
+            bodyClass='no-padding'
```
thomasheartman, line 166: Make table full-width as in the sketches.

frontend/src/component/changeRequest/ChangeRequests/GlobalChangeRequestTitleCell.tsx
```diff
@@ -41,12 +42,12 @@ export const GlobalChangeRequestTitleCell = ({
         id,
         title,
         project,
-        projectName,
         features: featureChanges,
         segments: segmentChanges,
     } = original;
+    const projectName = useProjectOverviewNameOrId(project);
```
thomasheartman, line 48: The real API payload doesn't send project names, so we need to fetch them from a hook.

frontend/src/component/changeRequest/ChangeRequestStatusBadge/ChangeRequestStatusBadge.tsx
```diff
@@ -9,20 +8,27 @@ import ErrorIcon from '@mui/icons-material/Error';
...
+    UnscheduledChangeRequest,
+} from '../changeRequest.types';
 
-interface IChangeRequestStatusBadgeProps {
-    changeRequest: ChangeRequestType | undefined;
+export interface IChangeRequestStatusBadgeProps {
+    changeRequest:
+        | Pick<UnscheduledChangeRequest, 'state'>
+        | Pick<ScheduledChangeRequest, 'state' | 'schedule'>
+        | undefined;
```
thomasheartman, line 20: This retyping only picks the props that it needs, allowing us to use the badges for more CR-like data, as long as it has state and schedule (if required).

frontend/src/component/changeRequest/ProjectChangeRequests/ChangeRequestsTabs/ChangeRequestStatusCell.tsx
```diff
@@ -1,14 +1,18 @@
...
+    ChangeRequestStatusBadge,
+    type IChangeRequestStatusBadgeProps,
+} from 'component/changeRequest/ChangeRequestStatusBadge/ChangeRequestStatusBadge';
+import type { FC } from 'react';
 
 interface IChangeRequestStatusCellProps {
     value?: string | null; // FIXME: proper type
-    row: { original: ChangeRequestType };
+    row: {
+        original: IChangeRequestStatusBadgeProps['changeRequest'];
```
thomasheartman, line 11: Propagates the type loosening from before.

frontend/src/hooks/api/getters/useChangeRequestSearch/useChangeRequestSearch.ts
```diff
@@ -31,6 +31,12 @@ const fallbackData: ChangeRequestSearchResponseSchema = {
 const SWR_CACHE_SIZE = 10;
 const PATH = 'api/admin/search/change-requests?';
 
+export type SearchChangeRequestsInput = {
+    [K in keyof SearchChangeRequestsParams]?:
+        | SearchChangeRequestsParams[K]
+        | null;
+};
```
thomasheartman, line 38: Loosen the implementation so that you can pass in both "null" and "undefined" to the api hook, and it'll smartly strip both of those from the query string.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -1,285 +1,178 @@
...
-        },
-    },
-];
+const columnHelper = createColumnHelper<ChangeRequestSearchItemSchema>();
 
 const ChangeRequestsInner = () => {
-    const loading = false;
+    const { user } = useAuthUser();
+
+    // Check URL parameters directly to avoid double fetching
```
sjaanus: We usually dont have these comments in code.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -1,285 +1,178 @@
...
+                cell: ({
+                    getValue,
                     row: {
                         original: { title, project },
                     },
-                }: any) => (
-                    <FeaturesCell project={project} value={value} key={title} />
+                }) => {
+                    const features = getValue();
+                    // Convert string array to object array for FeaturesCell compatibility
```
sjaanus: Can we get rid of inline comments

PR #10663, opened 2025-09-12
chore: persist strategy filter when going back from release template preview

nunogois:
https://linear.app/unleash/issue/2-3878/persist-strategy-filter-when-going-back-from-the-release-template

Persists the strategy filter in the new "add strategy" modal when going back from the release template preview modal.

This is done by moving the filter to the parent component, so the filter state persists across these navigations.

Also updates the button text in the release templates preview to say "Apply template" for consistency, but only if the `newStrategyModal` flag is enabled.


daveleek (approved):
Nice!

PR #10664, opened 2025-09-12
Add use change request search hook for UI.

thomasheartman:
Adds a use change request search hook. The hook (and tests) are based closely on the `useFeatureSearch` files.

I will wire them up to the table in an upcoming PR.

Also: fixes the orval schema to use numbers for offset and limit instead of strings (enterprise pr incoming). Plus: updates a variable usage in the use feature search hook.

### DIFF
--- a/frontend/src/hooks/api/getters/useChangeRequestSearch/useChangeRequestSearch.test.tsx
+++ b/frontend/src/hooks/api/getters/useChangeRequestSearch/useChangeRequestSearch.test.tsx
@@ -1,0 +1,76 @@
+import type { FC } from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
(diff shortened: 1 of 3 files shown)


frontend/src/hooks/api/getters/useFeatureSearch/useFeatureSearch.ts
```diff
@@ -63,7 +63,7 @@ const createFeatureSearch = () => {
         useClearSWRCache(swrKey, PATH, SWR_CACHE_SIZE);
 
         useEffect(() => {
-            initCache(params.project || '');
+            initCache(cacheId);
```
thomasheartman, line 66: This variable is assigned on line 62 with the exact same definition. I think reusing the variable here makes more sense. (Unless I'm missing something that the useeffect does)

PR #10672, opened 2025-09-22
feat: batch histogram metric type

kwasniew:
## About the changes

Add BatchHistogram for pre-Aggregated histogram metrics

Problem: regular Prometheus histograms are designed for individual observations (histogram.observe(value)), but in high-throughput distributed systems, clients need to send pre-aggregated histogram data to reduce network overhead and server load. E.g. 100 apps with local Unleash SDKs can collect histogram impact metrics locally.

  Before (individual observations):
```
  // 1000 network calls per second
  for (let i = 0; i < 1000; i++) {
      histogram.observe(responseTime);
  }
```

  After (batch aggregation):
```
  // Single network call for 10,000 observations
  batchHistogram.recordBatch(
      { service: 'api', region: 'us-east' },
      {
          count: 10000,        // Total observations
          sum: 3420.5,         // Sum of all values
          buckets: [           // Cumulative counts
              { le: 0.1, count: 2000 },
              { le: 0.5, count: 7000 },
              { le: 1, count: 9500 },
              { le: '+Inf', count: 10000 }
          ]
      }
  );
```

  Implementation Details

  - Stores cumulative bucket counts directly with the last bucket having all data points
  - Uses "+Inf" strings consistently throughout for infinity buckets
  - Integrates with existing prom-client Registry

### Important files

It will handle data from SDK histogram impact metrics: https://github.com/Unleash/unleash-node-sdk/pull/767
however it can be also used as a general purpose prometheus metric enhancing it capabilities

### DIFF
--- a/src/lib/features/metrics/impact/batch-histogram.ts
+++ b/src/lib/features/metrics/impact/batch-histogram.ts
@@ -1,7 +1,7 @@
 import type { Registry } from 'prom-client';
 
 interface BucketData {
-    le: number;
+    le: number | '+Inf';
     count: number;
 }
 
@@ -22,7 +22,7 @@ export class BatchHistogram {
         {
             count: number;
             sum: number;
-            buckets: Map<number, number>;
+            buckets: Map<number | '+Inf', number>;
         }
     > = new Map();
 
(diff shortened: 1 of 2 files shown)


src/lib/features/metrics/impact/batch-histogram.ts
```diff
@@ -0,0 +1,126 @@
...
+    constructor(config: {
+        name: string;
+        help: string;
+        registry: Registry;
+    }) {
+        this.name = config.name;
+        this.help = config.help;
+        this.registry = config.registry;
+
+        this.registry.registerMetric(this as any);
```
kwasniew, line 38: tried adding explicit type but not worth the effort

chriswk (approved):
Nice.

PR #10679, opened 2025-09-23
feat: add change request table filter buttons

thomasheartman:
Adds filter buttons for filtering between "CRs created by me" and "CRs where I've been requested as an approver".

The current implementation is fairly simplistic and the buttons are not connected to the actual table state directly (instead being set up with their own simple state and onChange hooks), but it covers the simple scenario. I want to defer a more complex solution until we know we need it and until we know exactly what we need. The implementation is based on the lifecycle filters that we have on the project flags page.

The current logic is such that: when you land on the page, there's no query params in the URL, but the data fetch applies `createdBy:IS<your user>`. If you switch to "approval requested" (and back again), the URL will reflect this.

For reference, the github workflow works like this, where each URL has a set of default filters, e.g.:
- `/pulls`: `is:open is:pr assignee:thomasheartman archived:false`
- `/pulls/review-requested`: `is:open is:pr review-requested:thomasheartman archived:false`

But if you change the default filters or add new ones, the URL will update to `pulls?<query-string>` (e.g. `/pulls?q=is%3Aopen+is%3Apr+review-requested%3Athomasheartman+archived%3Atrue`)

So this takes a similar approach, but better suited to the way we do tables in general.

Rendered:

image

image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useId, useMemo } from 'react';
 import { PageContent } from 'component/common/PageContent/PageContent';
 import { PageHeader } from 'component/common/PageHeader/PageHeader';
 import { PaginatedTable } from 'component/common/Table';
@@ -11,7 +11,6 @@ import { GlobalChangeRequestTitleCell } from './GlobalChangeRequestTitleCell.js'
 import { FeaturesCell } from '../ProjectChangeRequests/ChangeRequestsTabs/FeaturesCell.js';
 import { useUiFlag } from 'hooks/useUiFlag.js';
 import { withTableState } from 'utils/withTableState';
-import { useAuthUser } from 'hooks/api/getters/useAuth/useAuthUser';
 import {
     useChangeRequestSearch,
     DEFAULT_PAGE_LIMIT,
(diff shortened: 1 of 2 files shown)


frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -158,6 +161,30 @@ const ChangeRequestsInner = () => {
             data,
         }),
     );
+    const tableId = useId();
+    const handleQuickFilterChange = (filter: ChangeRequestQuickFilter) => {
+        if (!user) {
+            // todo (globalChangeRequestList): handle this somehow? Or just ignore.
```
FredrikOseberg, line 167: These are meant for a later cleanup?

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -158,6 +161,30 @@ const ChangeRequestsInner = () => {
...
+            return;
+        }
+        const [targetProperty, otherProperty] =
+            filter === 'Created'
+                ? ['createdBy', 'requestedApproverId']
+                : ['requestedApproverId', 'createdBy'];
+
+        // todo (globalChangeRequestList): extract and test the logic for wiping out createdby/requestedapproverid
```
FredrikOseberg, line 175: Same question as above

FredrikOseberg (approved):

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
thomasheartman: Yes, they are!

PR #10685, opened 2025-09-24
chore: update view more templates button design

nunogois:
https://linear.app/unleash/issue/2-3898/update-the-design-of-the-view-more-button

Updates the design of the "view more templates" button in the new "add strategy" modal.

image


gastonfournier (approved):
LG!

PR #10688, opened 2025-09-24
chore: limit total custom strategies displayed

nunogois:
https://linear.app/unleash/issue/2-3897/limit-custom-strategies-like-were-doing-with-release-templates

Limits total custom strategies displayed, like we're doing for release templates, in the new "add strategy" modal.

Added a more generic logic to `FeatureStrategyMenuCardsSection.tsx` so we can reuse it for both.


chriswk (approved):
LGTM

PR #10689, opened 2025-09-24
fix: cache eviction bug and the eternal loading screen

thomasheartman:
Fixes an issue where the project feature list (and potentially other places in the app that use the `useClearSWRCache` hook) would end up in an infinite loading screen because the latest entry that we want to show was overwritten.

The primary reason this happened is that we used `keysToDelete = array.slice(SWR_CACHE_SIZE - 1)`. Because the map keys are returned in insertion order, this would make us never delete the oldest keys, but always anything after the cache reached it's maximum size. The fix was to instead do `slice(0, -(SWR_CACHE_SIZE - 1))`, unless that is `0, 0`. If so, then just delete the entire filtered keys set. 

As a bonus: this PR also deduplicates cache entries that have the same query params but in different order for the feature search. This further reduces the cache space needed.

### DIFF
--- a/frontend/src/hooks/useClearSWRCache.test.ts
+++ b/frontend/src/hooks/useClearSWRCache.test.ts
@@ -14,7 +14,7 @@ describe('manageCacheEntries', () => {
         expect(cacheMock.has('prefix-3')).toBe(true);
     });
 
-    it('should keep the SWR_CACHE_SIZE entries and delete the rest', () => {
+    it('should keep the SWR_CACHE_SIZE entries and delete the rest, keeping the most recent entries', () => {
         const cacheMock = new Map();
         cacheMock.set('prefix-1', {});
         cacheMock.set('prefix-2', {});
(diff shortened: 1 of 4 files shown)


frontend/src/hooks/useClearSWRCache.ts
```diff
@@ -13,7 +13,12 @@ export const clearCacheEntries = (
     const filteredKeys = keys.filter(
         (key) => key.startsWith(clearPrefix) && key !== currentKey,
     );
-    const keysToDelete = filteredKeys.slice(SWR_CACHE_SIZE - 1);
+
+    const entriesToLeave = SWR_CACHE_SIZE - 1;
+    const keysToDelete =
+        entriesToLeave <= 0
+            ? filteredKeys
+            : filteredKeys.slice(0, -entriesToLeave);
```
thomasheartman, line 21: This is it 🐛

frontend/src/hooks/useClearSWRCache.test.ts
```diff
@@ -23,8 +23,19 @@ describe('manageCacheEntries', () => {
...
+    });
+
+    it('should not delete the current key, even if it would be cleared as part of the cache size limit', () => {
+        const cacheMock = new Map();
+        cacheMock.set('prefix-1', {});
+        cacheMock.set('prefix-2', {});
+        cacheMock.set('prefix-3', {});
+        cacheMock.set('prefix-4', {});
+
+        clearCacheEntries(cacheMock, 'prefix-2', 'prefix-', 2);
```
kwasniew, line 36: wondering if we should also clean the signature of clearCacheEntries. With 4 params it's totally unclear what each of those means.

thomasheartman: Ah, good point, yeah. Happy to do that in a follow-up?

frontend/src/hooks/useClearSWRCache.test.ts
```diff
@@ -50,4 +61,16 @@ describe('manageCacheEntries', () => {
         expect(cacheMock.has('other-2')).toBe(true);
         expect(cacheMock.has('prefix-3')).toBe(true);
     });
+
+    it('treats a negative cache size as zero', () => {
+        const cacheMock = new Map();
+        cacheMock.set('prefix-1', {});
```
kwasniew, line 67: is the order of prefixes supposed to be alphabetically sorted? the way those tests are written indicates so even though it shouldn't be the case

thomasheartman: Shouldn't matter. Insertion order does, though. We can use random keys instead if you prefer 🤷🏼

frontend/src/hooks/useClearSWRCache.test.ts
kwasniew: sure. next PR is fine

PR #10692, opened 2025-09-24
chore: use params object for swr cache clearing functions

thomasheartman:
With three and four different parameters (of which two are strings that are easily interchanged), it makes sense to rewrite these two functions to take named parameters instead. This is a follow-up to https://github.com/Unleash/unleash/pull/10689 based on one of the review comments.

### DIFF
--- a/frontend/src/hooks/useClearSWRCache.ts
+++ b/frontend/src/hooks/useClearSWRCache.ts
@@ -30,7 +30,7 @@ export const clearCacheEntries = ({
     keysToDelete.forEach((key) => cache.delete(key));
 };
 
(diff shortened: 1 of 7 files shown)


frontend/src/hooks/useClearSWRCache.ts
```diff
@@ -23,16 +30,27 @@ export const clearCacheEntries = (
     keysToDelete.forEach((key) => cache.delete(key));
 };
 
+type useClearSWRCacheProps = {
```
kwasniew: types usually start with the upper case letter

kwasniew (approved):

PR #10693, opened 2025-09-24
feat: add users updated read model

gastonfournier:
## About the changes
This adds a user read model using the updated_at new field in the users table.

### DIFF
--- a/src/lib/features/users/user-updates-read-model.ts
+++ b/src/lib/features/users/user-updates-read-model.ts
@@ -1,0 +1,45 @@
+import type { Logger, LogProvider } from '../../logger.js';
+
+import type { Db } from '../../db/db.js';
+import { rowToUser, USER_COLUMNS_PUBLIC, USERS_TABLE } from './user-store.js';
+import type { User } from '../../internals.js';
+
+export class UserUpdatesReadModel {
+    private db: Db;
+
+    private logger: Logger;
+
+    constructor(db: Db, getLogger: LogProvider) {
+        this.db = db;
+        this.logger = getLogger('user-updates-read-model.ts');
+    }
+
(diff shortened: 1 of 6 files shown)


src/test/e2e/users/user-updates-read-model.e2e.test.ts
```diff
@@ -0,0 +1,113 @@
+import dbInit, { type ITestDb } from '../helpers/database-init.js';
+import getLogger from '../../fixtures/no-logger.js';
+import type { IUnleashStores } from '../../../lib/types/index.js';
+import { beforeAll, test, expect } from 'vitest';
+let stores: IUnleashStores;
+let db: ITestDb;
+
+beforeAll(async () => {
+    db = await dbInit('user_store_serial', getLogger, {
+        experimental: { flags: {} },
```
nunogois: Might as well remove this?

gastonfournier: Done and keeping it simple: `db = await dbInit();`

src/lib/features/users/user-updates-read-model.ts
```diff
@@ -0,0 +1,74 @@
+import type { Logger, LogProvider } from '../../logger.js';
+
+import type { Db } from '../../db/db.js';
+import { USER_COLUMNS_PUBLIC, USERS_TABLE } from './user-store.js';
+import type { Row } from '../../db/crud/row-type.js';
+
+type Response = {
```
nunogois: Should we call this something else? I know it's not exported, but if I noticed a method returning `Response` my first instinct would be that this is an `express` `Response` instead of something user related.

gastonfournier: Yeah, good call, renaming and then merging

PR #10694, opened 2025-09-24
chore: rename variable for clarity

thomasheartman:
entriesToLeave is more ambiguous than entriesToKeep, so let's go with the latter.


thomasheartman:
Oh, I stopped caring about branch names for changes not related to tickets, so `jj` just handles it for me. It's pretty neat 😄

PR #10704, opened 2025-09-29
chore: extract UI config logic into its own service

nunogois:
https://linear.app/unleash/issue/2-3921/extract-ui-config-logic-into-its-own-service

Extracts UI config logic into its own service.

This is the first step to accomplish resource limits and license key resources alignment.


gastonfournier (approved):
This makes sense!

PR #10709, opened 2025-10-01
chore: resource limits service

nunogois:
https://linear.app/unleash/issue/2-3927/implement-resource-limits-service

Implements a resource limits service.

The implementation looks trivial (or even redundant) in OSS, but by implementing a resource limits service we can make this potentially dynamic and overridable.


gastonfournier (approved):
LG!

PR #10710, opened 2025-10-01
feat: change "archived" flags filter placement on project flags list

Tymek:
Archived flags link will now it will show up on the right side, next to import/export, which makes it more in line with flags overview.

image

### DIFF
--- a/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureToggles.tsx
+++ b/frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureToggles.tsx
@@ -119,7 +119,7 @@ const LinkToggle = styled('button')(({ theme }) => ({
 }));
 
 export const ProjectFeatureToggles = ({
-    environments: passedEnvironments,
+    environments: availableEnvironments,
 }: ProjectFeatureTogglesProps) => {
     const { trackEvent } = usePlausibleTracker();
     const projectId = useRequiredPathParam('projectId');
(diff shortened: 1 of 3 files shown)


frontend/src/component/project/Project/PaginatedProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -527,6 +553,24 @@ export const ProjectFeatureToggles = ({
                             isLoading={initialLoad}
                             totalItems={total}
                             environmentsToExport={environments}
+                            actions={
+                                flagsUiFilterRefactorEnabled ? (
+                                    <LinkToggle
+                                        type='button'
+                                        onClick={toggleArchived}
+                                        aria-pressed={showArchived}
```
thomasheartman: I'm gonna suggest you take this aria attribute out. According to [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-pressed):

> Do not change the contents of the label on a toggle when the state changes. If a button label says "pause", do not change it to "play" when pressed.

and

> If you want the label to toggle between "Paused" and "Play", don't use aria-pressed.

In other words, I don't think the attribute adds anything valuable here, so we should remove it:

> The first rule of ARIA use is "if you can use a native feature with the semantics and behavior you require already built in, instead of repurposing an element and adding an ARIA role, state or property to make it accessible, then do so."

```suggestion
```

Optionally, we could also just not change the text.

thomasheartman (approved):
Good! Just double check the aria stuff, please.

PR #10723, opened 2025-10-02
chore: new method that considers users updated at the same time

gastonfournier:
We could have users updated at the exact same time, so we need to include that timestamp in the next fetch to avoid missing users, but also include the latest user id so we can exclude the ones already fetched.

The following test shows how to process pages with this new method:
https://github.com/Unleash/unleash/blob/c03df86ee006f531024137a1cfc846eb282ab672/src/lib/features/users/user-updates-read-model.test.ts#L39-L65

### DIFF
--- a/src/lib/features/users/user-updates-read-model.test.ts
+++ b/src/lib/features/users/user-updates-read-model.test.ts
@@ -1,0 +1,65 @@
+import dbInit, {
+    type ITestDb,
+} from '../../../test/e2e/helpers/database-init.js';
+
+let db: ITestDb;
+const TABLE = 'users';
+const INSERT_INSTANT = new Date();
+beforeAll(async () => {
+    db = await dbInit();
+
+    for (let i = 0; i < 10; i += 1) {
+        await db.rawDatabase(TABLE).insert({
+            name: `test-user-${i}`,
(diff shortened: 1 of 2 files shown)


nunogois (approved):
What if we updated more than 100 users at the exact time? 😄

gastonfournier:
> What if we updated more than 100 users at the exact time? 😄

It's what I've been thinking about. The only workaround I can think of (now), is to also send the latest id (after ordering by id), so when equal updated_at we can also remove the ones with user id lower than the provided id. That way you'll always be getting a new page with newly unseen users.

nunogois (approved):
So, if I understood correctly, we're returning the first 100 (limit) that match one of:

- updated_at > lastUpdate
- updated_at === lastUpdate && id > lastId

LGTM.

gastonfournier:
> So, if I understood correctly, we're returning the first 100 (limit) that match one of:
> 
> * updated_at > lastUpdate
> * updated_at === lastUpdate && id > lastId
> 
> LGTM.

Yes, it's not as simple as I'd like it to be, but I couldn't come up with something simpler... maybe I should've used more precision in updated_at, but it'd just reduce the chances of having same value.

PR #10732, opened 2025-10-03
chore: return user id alongside max updated_at

gastonfournier:
This is for internal use

### DIFF
--- a/src/lib/features/users/user-updates-read-model.ts
+++ b/src/lib/features/users/user-updates-read-model.ts
@@ -50,20 +50,18 @@ export class UserUpdatesReadModel {
         return result ? result.last_updated_at : null;
     }
 
-    /**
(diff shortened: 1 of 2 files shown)


src/lib/features/users/user-updates-read-model.ts
```diff
@@ -38,16 +38,26 @@ export class UserUpdatesReadModel {
...
+            .orderBy('updated_at', 'desc')
+            .orderBy('id', 'desc')
             .first();
-        return result ? result.last_updated_at : null;
+        return result
+            ? {
+                  lastUpdatedAt: result.updated_at,
+                  userId: result.id,
+              }
+            : null;
```
gastonfournier, line 60: The change of shape will be a breaking change but I already have a PR almost ready

PR #10737, opened 2025-10-06
fix: Information exposure through a stack trace

gastonfournier:
Fix for [https://github.com/Unleash/unleash/security/code-scanning/81](https://github.com/Unleash/unleash/security/code-scanning/81)

To prevent information exposure through stack traces, ensure that the HTTP response sent to clients contains only sanitized, generic error information, such as a status code and a simple message. Internal details (including stack traces, error types, or internal error codes) should not be sent to the client. These can be safely logged on the server for debugging.

**The fix:**  
- Do not return the entire `finalError` object as JSON to the client, as it may include fields like `stack` or `internalMessage`.
- Instead, return only a subset of fields that are safe to expose to the user, in this case just `message` .
- Log the full error and any debugging details using the server-side logger **as currently done**.

---
_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


sighphyre (approved):
LGTM

src/lib/routes/util.ts
```diff
@@ -61,5 +61,23 @@ export const handleErrors: (
...
+    // hangs the execution of this method. Returning it as finalError.details doesn't
+    // work returning undefined. Printing out the finalError object using JSON.stringify
+    // shows that the details property is there and is an array.
+    // Running JSON.stringify(finalError.details) also hangs.
+    // As a workaround, we do a roundabout way of getting to the details property
+    // by doing JSON.parse(JSON.stringify(finalError))['details']
+    const details =
+        // @ts-expect-error - details might not be present on all UnleashErrors
+        // biome-ignore lint/complexity/useLiteralKeys: see above
+        finalError.details ?? JSON.parse(JSON.stringify(finalError))['details'];
```
gastonfournier, line 74: Note: if `finalError.details` is available, we use it without going through the stringification of the final error

gastonfournier: After thinking about this, the finalError was being serialized as JSON in the previous solution to send it over http protocol, so this shouldn't cause any issue.

PR #10745, opened 2025-10-07
feat: batch week data if the shown time span is greater than 12 weeks.

thomasheartman:
Implements batching of data points in the archived:created chart: when there's 12 or more weeks of data, batch data into batches of 4 weeks at a time. When we batch data, we also switch the labeling to be month-based and auto-generated (cf the inline comment with more details).

image

The current implementation batches into groups of 4 weeks, but this can easily be parameterized to support arbitrary batch sizes.

Because of the batching, we also now need to adjust the tooltip title in those cases. This is handled by a callback.

### DIFF
--- a/frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
+++ b/frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
@@ -15,10 +15,11 @@ import {
     TimeScale,
     Chart as ChartJS,
     Filler,
+    type TooltipItem,
 } from 'chart.js';
 import { useLocationSettings } from 'hooks/useLocationSettings';
 import type { TooltipState } from 'component/insights/components/LineChart/ChartTooltip/ChartTooltip';
-import type { WeekData, RawWeekData } from './types.ts';
+import type { WeekData, RawWeekData, BatchedWeekData } from './types.ts';
 import { createTooltip } from 'component/insights/components/LineChart/createTooltip.ts';
 import { CreationArchiveRatioTooltip } from './CreationArchiveRatioTooltip.tsx';
 import { getDateFnsLocale } from '../../getDateFnsLocale.ts';
@@ -27,6 +28,8 @@ import { NotEnoughData } from 'component/insights/components/LineChart/LineChart
 import { placeholderData } from './placeholderData.ts';
 import { Bar } from 'react-chartjs-2';
 import { GraphCover } from 'component/insights/GraphCover.tsx';
+import { format, startOfWeek } from 'date-fns';
+import { batchWeekData } from './batchWeekData.ts';
 
 ChartJS.register(
     CategoryScale,
@@ -47,6 +50,8 @@ interface ICreationArchiveChartProps {
     isLoading?: boolean;
 }
 
+type DataResult = 'Not Enough Data' | 'Batched' | 'Weekly';
+
 export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
     creationArchiveTrends,
     isLoading,
@@ -56,7 +61,7 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
     const { locationSettings } = useLocationSettings();
     const [tooltip, setTooltip] = useState<null | TooltipState>(null);
 
-    const { notEnoughData, aggregateOrProjectData } = useMemo(() => {
+    const { dataResult, aggregateOrProjectData } = useMemo(() => {
         const labels: string[] = Array.from(
             new Set(
                 creationVsArchivedChart.datasets.flatMap((d) =>
@@ -65,19 +70,21 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
             ),
         );
 
-        const aggregateWeekData = (acc: WeekData, item: RawWeekData) => {
-            if (item) {
-                acc.archivedFlags += item.archivedFlags || 0;
+        const aggregateWeekData = (acc: WeekData, item?: RawWeekData) => {
+            if (!item) return acc;
 
-                if (item.createdFlags) {
-                    Object.entries(item.createdFlags).forEach(([_, count]) => {
-                        acc.totalCreatedFlags += count;
-                    });
-                }
+            acc.archivedFlags += item.archivedFlags || 0;
+
+            if (item.createdFlags) {
+                Object.entries(item.createdFlags).forEach(([_, count]) => {
+                    acc.totalCreatedFlags += count;
+                });
             }
+
             if (!acc.date) {
-                acc.date = item?.date;
+                acc.date = item.date;
             }
+
             return acc;
         };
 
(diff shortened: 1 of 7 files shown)


frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -65,19 +70,21 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
             ),
         );
 
-        const aggregateWeekData = (acc: WeekData, item: RawWeekData) => {
-            if (item) {
-                acc.archivedFlags += item.archivedFlags || 0;
+        const aggregateWeekData = (acc: WeekData, item?: RawWeekData) => {
```
thomasheartman, line 73: This function has just been flipped to reduce nesting. The behavior should be exactly the same.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -86,6 +93,7 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
             totalCreatedFlags: 0,
             archivePercentage: 0,
             week: label,
+            date: '',
```
thomasheartman, line 96: Add an empty date here to allow the `date` property to be non-optional. I'm not really sure why we allow for empty week data points, but it may have to do with padding out the chart? Anyway, `''` is also falsy in JS and I haven't seen any breakage, so I think this is a worthwhile change.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -138,10 +156,26 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
...
+            startOfWeek(new Date(rawData.date), {
+                locale,
+                weekStartsOn: 1,
+            }),
+            `PP`,
+            { locale },
+        );
+        const endDate = format(new Date(rawData.endDate), `PP`, { locale });
+        return `${startDate} – ${endDate}`;
+    };
```
thomasheartman, line 177: This is the bit that gives us the new tooltips. Only useful with batched week data, it gives us a from date and a to date separated by an en dash:

Image

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -138,10 +156,26 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
...
     const data = useGraphCover ? placeholderData : aggregateOrProjectData;
 
+    const locale = getDateFnsLocale(locationSettings.locale);
+    const batchedTooltipTitle = (datapoints: TooltipItem<any>[]) => {
+        const rawData = datapoints[0].raw as GroupedWeekData;
+        const startDate = format(
+            startOfWeek(new Date(rawData.date), {
+                locale,
+                weekStartsOn: 1,
+            }),
```
thomasheartman, line 171: Metrics are collected on Sundays. This makes it so that the first "from" date is the monday of the week the data is collected for. Without it, it looks ilke there's only data for three weeks.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -172,27 +206,39 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
                     enabled: false,
                     position: 'average' as const,
                     external: createTooltip(setTooltip),
+                    callbacks: {
+                        title:
+                            dataResult === 'Batched'
+                                ? batchedTooltipTitle
+                                : undefined,
```
thomasheartman, line 213: If this is `undefined` the default title is used instead.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -172,27 +206,39 @@ export const CreationArchiveChart: FC<ICreationArchiveChartProps> = ({
...
                     },
                     grid: {
                         display: false,
                     },
                     ticks: {
-                        source: 'data' as const,
+                        source:
+                            dataResult === 'Batched'
+                                ? ('auto' as const)
+                                : ('data' as const),
```
thomasheartman, line 241: If using batched data, show auto-adjusted, monthly labels instead. That evenly distributes the labels independently of the data points' start times:

Image

If we instead keep using `data` (but switch to month), we get this (notice the weirdly wide March marking (3月 2025)):

Image

And if we use `week` with `auto` (not good because we'll get way too many ticks):

Image

Finally, `week` with `data` (this is okay-ish, but I think seeing dates like this is less useful because of the batched nature of the data):

Image

The one downside to using `auto`-labeling, however, is that it always seems to miss the first label I'd expect. However, I consider this to be a worthwile tradeoff.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveRatioTooltip.tsx
```diff
@@ -44,6 +44,7 @@ interface CreationArchiveRatioTooltipProps {
 }
 
 const Timestamp = styled('span')(({ theme }) => ({
+    whiteSpace: 'nowrap',
```
thomasheartman, line 47: Prevent the "from date - to date" text from wrapping, instead making the box wider if it needs to.

frontend/src/component/insights/hooks/useInsightsData.ts
thomasheartman, line 1: The changes in this file is just colocating the variable assignments to make the flow easier to follow, e.g.

instead of:
```
const A = ...
const B = ...
const C = ...

const A' = ... // use A
const B' = ... // use B
const C' = ... // use C
```

then this:
```
const A = ...
const A' = ... // use A

const B = ...
const B' = ... // use B

const C = ...
const C' = ... // use C
```

This is obviously preferential and doesn't really make a difference to the execution, but I think it reads better. Happy to revert if you prefer.

frontend/src/component/insights/sections/PerformanceInsights.tsx
```diff
@@ -71,13 +71,6 @@ export const PerformanceInsights: FC = () => {
     const lastFlagTrend = flagTrends[flagTrends.length - 1];
     const flagsTotal = lastFlagTrend?.total ?? 0;
 
-    function getFlagsPerUser(flagsTotal: number, usersTotal: number) {
```
thomasheartman, line 74: This function was unused, so this is a bit of cleanup.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -47,6 +50,8 @@ interface ICreationArchiveChartProps {
     isLoading?: boolean;
 }
 
+type DataResult = 'Not Enough Data' | 'Batched' | 'Weekly';
```
thomasheartman, line 53: I'd love a better name for this type if you can think of one 😅

sjaanus: Local type, good enough I guess.

PR #10746, opened 2025-10-07
feat: move release plans to feature environments

kwasniew:
## About the changes

Release plans now are queries as part of the GET feature API call in each environment they belong to. It moves them to the same place as strategies.
Screenshot 2025-10-07 at 11 46 30

The result of this change is moving data join to a DB from the browser making multiple API calls per env.

Implementation details:
* introduced read model encapsulating query for all release plans for all feature environments. The query was taken from the release plan store and generalized for multiple envs.
* earlier PR moved release plan schemas to OSS and now we're updating all transitive schemas to use it
* actual environment enhancement with release plans is flagged so that we can merge it safely and add tests in another repo


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1182,6 +1205,27 @@ export class FeatureToggleService {
         }
     }
 
+    private async addReleasePlansToEnvironments(
```
kwasniew, line 1208: this is the least invasive way to add release plans. For non flag enabled clients it does nothing

src/lib/features/release-plans/release-plan-read-model.ts
```diff
@@ -0,0 +1,184 @@
+import type { Db } from '../../db/db.js';
```
kwasniew, line 1: 95% of this file was copied from release-plan-store getByFeatureFlagAndEnvironment method

src/lib/features/feature-toggle/feature-toggle-service.ts
gastonfournier: This looks good, but `add` could be misleading you into thinking that this method is actually adding a release plan to an environment. If I may, I'd suggest naming this method `mapReleasePlansToEnvironments`

FredrikOseberg (approved):
LGTM

PR #10755, opened 2025-10-07
fix: clone environments

nunogois:
https://linear.app/unleash/issue/2-3932/cloned-environments-enable-disabled-strategies-unexpectedly

Cloning environments didn't work as expected. This fixes a few of issues:
 - Disabled strategies remain disabled after cloning
 - All strategy properties are cloned (including e.g. title)
 - Strategy cloning respects the selected projects
 - Release plans and their milestones are now correctly cloned

### DIFF
--- a/src/lib/db/feature-environment-store.ts
+++ b/src/lib/db/feature-environment-store.ts
@@ -475,71 +475,81 @@ export class FeatureEnvironmentStore implements IFeatureEnvironmentStore {
     async cloneStrategies(
         sourceEnvironment: string,
         destinationEnvironment: string,
+        projects: string[],
     ): Promise<void> {
         const stopTimer = this.timer('cloneStrategies');
-        const sourceFeatureStrategies = await this.db(
-            'feature_strategies',
-        ).where({
-            environment: sourceEnvironment,
-        });
 
-        const clonedStrategyRows = sourceFeatureStrategies.map(
-            (featureStrategy) => {
-                return {
+        await this.db.transaction(async (trx) => {
+            const sourceFeatureStrategies = await trx(
+                'feature_strategies as fs',
+            )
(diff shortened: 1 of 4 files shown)


src/lib/db/feature-environment-store.ts
```diff
@@ -475,71 +475,81 @@ export class FeatureEnvironmentStore implements IFeatureEnvironmentStore {
...
-            (featureStrategy) => {
-                return {
+        await this.db.transaction(async (trx) => {
+            const sourceFeatureStrategies = await trx(
+                'feature_strategies as fs',
+            )
+                .join('features as f', 'f.name', 'fs.feature_name')
+                .select('fs.*')
+                .where('fs.environment', sourceEnvironment)
+                .whereIn('f.project', projects);
```
gastonfournier, line 489: I think the projects are used as target, not as source:
image

gastonfournier (changes requested):
I think this is good but projects should not be used to filter source strategies

src/lib/db/feature-environment-store.ts
nunogois: I don't think this is correct, I believe they should be used as both. Copying strategies for projects you haven't selected feels like a bug to me.

There's also no way of doing the same thing for release plans. It is much easier to assume only the selected projects. And release plans and strategies are closely related, so both should follow a similar logic.

gastonfournier: Ok, I understand now. When you select the project, you're only interested in cloning the environment configuration just for these projects. Otherwise, after enabling the environment in a project where you did not choose to clone, you'll end up with the same configuration you had in the strategies at cloning time. This is unexpected because you're clearly targeting those specific projects when cloning; the rest should be without configuration.

gastonfournier (approved):
LG!

PR #10756, opened 2025-10-08
feat: Add batching functionality to new flags in production

thomasheartman:
Adds the same batching functionality that was added to the archived:created chart to the new flags in production chart.

In doing so, I've extracted the batching algorithm and the batched tooltip title creation, as well as the ChartDataResult type (though naming suggestions are still welcome on that front).

Locale 'ja':
image

Locale 'no':
image

### DIFF
--- a/frontend/src/component/insights/componentsChart/CreationArchiveChart/batchCreationArchiveData.ts
+++ b/frontend/src/component/insights/componentsChart/CreationArchiveChart/batchCreationArchiveData.ts
@@ -1,8 +1,8 @@
 import { batchData } from '../batchData.ts';
 import type { BatchedWeekData, WeekData } from './types.ts';
 
-export const batchCreationArchiveData = batchData(
-    (accumulated: BatchedWeekData, next: WeekData) => {
+export const batchCreationArchiveData = batchData({
+    merge: (accumulated: BatchedWeekData, next: WeekData) => {
         accumulated.totalCreatedFlags += next.totalCreatedFlags;
         accumulated.archivedFlags += next.archivedFlags;
 
@@ -15,7 +15,7 @@ export const batchCreationArchiveData = batchData(
         accumulated.endDate = next.date;
         return accumulated;
     },
-    (item: WeekData) => {
+    map: (item: WeekData) => {
         const { week: _, ...shared } = item;
 
         return {
@@ -23,4 +23,4 @@ export const batchCreationArchiveData = batchData(
             endDate: item.date,
         };
     },
-);
+});
--- a/frontend/src/component/insights/componentsChart/batchData.test.ts
+++ b/frontend/src/component/insights/componentsChart/batchData.test.ts
@@ -1,21 +1,15 @@
 import { batchData } from './batchData.ts';
 
 it('handles empty input', () => {
-    expect(
-        batchData(
-            (x, _) => x,
-            (x) => x,
-        )([]),
-    ).toEqual([]);
+    expect(batchData({ merge: (x, _) => x, map: (x) => x })([])).toEqual([]);
 });
 
 it('batches by 4, starting from the first entry', () => {
     const input = [7, 11, 13, 19, 23];
 
     expect(
-        batchData<number, number>(
-            (x, y) => x + y,
-            (x) => x,
-        )(input),
+        batchData<number, number>({ merge: (x, y) => x + y, map: (x) => x })(
+            input,
+        ),
     ).toStrictEqual([50, 23]);
 });
--- a/frontend/src/component/insights/componentsChart/batchData.ts
+++ b/frontend/src/component/insights/componentsChart/batchData.ts
@@ -1,10 +1,17 @@
-const batchSize = 4;
+const defaultBatchSize = 4;
+
+export type BatchDataOptions<T, TBatched> = {
+    merge: (accumulated: TBatched, next: T) => TBatched;
+    map: (item: T) => TBatched;
+    batchSize?: number;
+};
 
 export const batchData =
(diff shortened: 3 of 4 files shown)


frontend/src/component/insights/components/LineChart/ChartTooltip/ChartTooltip.tsx
```diff
@@ -91,12 +91,12 @@ export const ChartTooltipContainer: FC<IChartTooltipProps> = ({
...
 );
 
-export const ChartTooltip: VFC<IChartTooltipProps> = ({ tooltip }) => (
+export const ChartTooltip: FC<IChartTooltipProps> = ({ tooltip }) => (
     <ChartTooltipContainer tooltip={tooltip}>
         <Paper
             elevation={3}
             sx={(theme) => ({
-                width: 220,
+                width: 'max-content',
```
thomasheartman, line 99: Allows the tooltip to be grow to whatever size is needed for the content. Seems safe to me. The only other chart using this tooltip is flags, and I don't think that's in any danger of overflowing.

Image

frontend/src/component/insights/componentsChart/CreationArchiveChart/batchCreationArchiveData.test.ts
thomasheartman, line 1: This is a modified and simplified test suite that was previously `frontend/src/component/insights/componentsChart/CreationArchiveChart/batchWeekData.test.ts`. I've removed tests that test the batching algorithm, and instead focused on the merging behavior, which leads to simpler tests.

frontend/src/component/insights/componentsChart/CreationArchiveChart/batchCreationArchiveData.ts
thomasheartman, line 1: This was `frontend/src/component/insights/componentsChart/CreationArchiveChart/batchWeekData.ts` before. But because the batching algorithm has been extracted, all we need to supply are the merging functions and the new item function.

frontend/src/component/insights/componentsChart/NewProductionFlagsChart/batchProductionFlagsData.test.ts
thomasheartman, line 1: Similar tests as for archived:created flags. Just ensure that the merging / creation works as expected.

frontend/src/component/insights/componentsChart/NewProductionFlagsChart/NewProductionFlagsChart.tsx
```diff
@@ -19,10 +22,45 @@ interface IProjectHealthChartProps {
...
+                },
+                plugins: {
+                    tooltip: {
+                        callbacks: {
+                            title: batchedTooltipTitle,
+                        },
+                    },
+                },
+            };
+        }
```
thomasheartman, line 56: When the data is batched, we want to switch time unit and update the tooltip.

frontend/src/component/insights/componentsChart/batchData.test.ts
thomasheartman, line 1: These are the algorithm tests that were extracted. Because it's generic over any two types now, we can test it in a simpler fashion.

frontend/src/component/insights/componentsChart/batchData.ts
thomasheartman, line 1: This is the generic batching algorithm. It just needs to know how to add batched and unbatched together, and how to create a batched item from a single unbatched one.

frontend/src/component/insights/componentsChart/chartDataResult.ts
thomasheartman, line 1: Not really sure this needs to be extracted (it could stay as two local types), but ... eh? 🤷🏼

frontend/src/component/insights/componentsChart/useBatchedTooltipDate.ts
```diff
@@ -0,0 +1,33 @@
...
+                `PP`,
+                { locale },
+            );
+            const endDate = format(new Date(dataPoint.endDate), `PP`, {
+                locale,
+            });
+            return `${startDate} – ${endDate}`;
+        }
+
+        return undefined;
```
thomasheartman: According to [the chart.js docs](https://www.chartjs.org/docs/latest/configuration/tooltip.html#tooltip-callbacks):
> If the callback returns `undefined`, then the default callback will be used

So if we can't generate an appropriate label, we'll fall back to the default generation.

frontend/src/component/insights/componentsChart/batchData.ts
```diff
@@ -0,0 +1,19 @@
+const batchSize = 4;
+
+export const batchData =
```
sjaanus: This is hardcoded 4 batch size, can we rename this function to say it and/or add JSdoc commenting what it exactly does.

sjaanus: or put it as param.

sjaanus (approved):
Approved with comment

frontend/src/component/insights/componentsChart/batchData.ts
thomasheartman: Sure thing. I've got an idea, actually! 💡

thomasheartman: Updated in e4bfe568c20cb3eb400ee00f9427d6eb308d5694 by changing using an options object to start.

thomasheartman: Or rather: 1122e25a9975da6a1fe1cbd48abbe0dba0ad9a1e

PR #10759, opened 2025-10-08
Improve release plan write model

kwasniew:
## About the changes

Matching release plan types with enterprise repo expectations

### DIFF
--- a/src/lib/features/release-plans/release-plan-milestone-store.ts
+++ b/src/lib/features/release-plans/release-plan-milestone-store.ts
@@ -16,7 +16,10 @@ const fromRow = (row: any): ReleasePlanMilestone => {
     };
 };
 
-export type ReleasePlanMilestoneWriteModel = Omit<ReleasePlanMilestone, 'id'>;
+export type ReleasePlanMilestoneWriteModel = Omit<
+    ReleasePlanMilestone,
+    'id' | 'startedAt' | 'transitionCondition'
+>;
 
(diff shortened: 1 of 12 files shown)


src/lib/features/release-plans/release-plan-milestone-store.ts
```diff
@@ -16,7 +16,10 @@ const fromRow = (row: any): ReleasePlanMilestone => {
     };
 };
 
-export type ReleasePlanMilestoneWriteModel = Omit<ReleasePlanMilestone, 'id'>;
+export type ReleasePlanMilestoneWriteModel = Omit<
+    ReleasePlanMilestone,
+    'id' | 'startedAt' | 'transitionCondition'
```
kwasniew, line 21: startedAt and transitionCondition are fetched at read time only

src/lib/features/release-plans/release-plan-milestone.ts
```diff
@@ -5,7 +5,7 @@ export interface ReleasePlanMilestone {
     name: string;
     sortOrder: number;
     releasePlanDefinitionId: string;
-    startedAt?: Date;
+    startedAt?: string;
```
kwasniew, line 8: matching enterprise repo expectations form openapi schema

kwasniew, line 8: this is what enterprise repo expects and since it depends on openapi schema which is string. since we don't need date manipulations in the code we can return string and pass it up to the API layer

PR #10763, opened 2025-10-08
feat: add median per week calculation to the new flags in production widget

thomasheartman:
Based on the creation archive widget.

### DIFF
--- a/frontend/src/component/insights/componentsStat/NewProductionFlagsStats/NewProductionFlagsStats.tsx
+++ b/frontend/src/component/insights/componentsStat/NewProductionFlagsStats/NewProductionFlagsStats.tsx
@@ -1,0 +1,29 @@
+import type { FC } from 'react';
+import type { GroupedDataByProject } from 'component/insights/hooks/useGroupedProjectTrends';
+import type { InstanceInsightsSchema } from 'openapi';
+import { calculateMedian } from './calculateMedian.ts';
+import { StatCard } from '../shared/StatCard.tsx';
+
+interface NewProductionFlagsStatsProps {
+    groupedLifecycleData: GroupedDataByProject<
+        InstanceInsightsSchema['lifecycleTrends']
+    >;
+    isLoading?: boolean;
+}
+
+export const NewProductionFlagsStats: FC<NewProductionFlagsStatsProps> = ({
+    groupedLifecycleData,
+    isLoading,
+}) => {
+    const median = calculateMedian(groupedLifecycleData);
+
+    return (
+        <StatCard
+            value={median}
+            label='Median per week'
+            tooltip='The median number of flags that have entered production for the selected time range.'
+            explanation='How often do flags go live in production?'
+            isLoading={isLoading}
+        />
+    );
+};
--- a/frontend/src/component/insights/sections/PerformanceInsights.tsx
+++ b/frontend/src/component/insights/sections/PerformanceInsights.tsx
@@ -22,13 +22,12 @@ import {
     StyledWidget,
     StyledWidgetContent,
     StyledWidgetStats,
-    StatsExplanation,
 } from '../InsightsCharts.styles';
 import { useUiFlag } from 'hooks/useUiFlag';
 import { NewProductionFlagsChart } from '../componentsChart/NewProductionFlagsChart/NewProductionFlagsChart.tsx';
-import Lightbulb from '@mui/icons-material/LightbulbOutlined';
 import { CreationArchiveChart } from '../componentsChart/CreationArchiveChart/CreationArchiveChart.tsx';
 import { CreationArchiveStats } from '../componentsStat/CreationArchiveStats/CreationArchiveStats.tsx';
+import { NewProductionFlagsStats } from '../componentsStat/NewProductionFlagsStats/NewProductionFlagsStats.tsx';
 
 export const PerformanceInsights: FC = () => {
     const statePrefix = 'performance-';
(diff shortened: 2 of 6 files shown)


frontend/src/component/insights/componentsChart/NewProductionFlagsChart/types.ts
```diff
@@ -4,6 +4,6 @@ export type WeekData = {
     date: string;
 };
 
-export type BatchedWeekData = Omit<WeekData, 'week'> & {
+export type BatchedWeekData = WeekData & {
```
thomasheartman, line 7: Keep the `week` prop as a key to be used in median calculation. We could use a different key if we wanted, but this keeps it simpler 🤷🏼

frontend/src/component/insights/componentsChart/CreationArchiveChart/batchCreationArchiveData.ts
thomasheartman, line 1: Slightly restructure the exported function to add a parameterized batch size. This accounts for the new chart data type, which includes batch size.

frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveChart.tsx
```diff
@@ -51,6 +53,42 @@ interface ICreationArchiveChartProps {
...
+            order: 2,
+            parsing: {
+                yAxisKey: 'totalCreatedFlags',
+                xAxisKey: 'date',
+            },
+            ...sharedDatasetOptions(theme.palette.charts.A1),
+        },
+    ],
+});
+
```
thomasheartman, line 91: These have been extracted from the `useMemo` invocation to make it easier to read and reason about.

frontend/src/component/insights/componentsChart/NewProductionFlagsChart/batchProductionFlagsData.ts
thomasheartman, line 1: Same as for the other partial application of the batchData function.

frontend/src/component/insights/componentsChart/NewProductionFlagsChart/NewProductionFlagsChart.tsx
thomasheartman, line 1: The content that has disappeared from this file has been extracted into the new `useNewProductionFlagsData` hook.

frontend/src/component/insights/componentsChart/NewProductionFlagsChart/useNewProductionFlagsData.ts
thomasheartman, line 1: This is the extracted content from the NewProductionFlagsChart component. It also adds the median calculation after the potential week batching.

frontend/src/component/insights/componentsChart/chartData.ts
thomasheartman, line 1: I've introduced this new type to keep things easy to handle and colocated. It allows us to put the current state of the data and the data in the same variable, and also adds some extra type hints to facilitate type narrowing.

It supersedes the previous `ChartDataResult` type, which was only a set of strings.

frontend/src/component/insights/componentsStat/NewProductionFlagsStats/NewProductionFlagsStats.tsx
thomasheartman, line 1: Uses the new stat card introduced in a previous PR to display the median for the new production flags stats.

frontend/src/component/insights/sections/PerformanceInsights.tsx
```diff
@@ -22,13 +22,42 @@ import {
...
+                    chartData={chartData}
+                    median={median}
+                />
+            </StyledWidgetStats>
+            <StyledChartContainer>
+                <NewProductionFlagsChart chartData={chartData} />
+            </StyledChartContainer>
+        </StyledWidget>
+    );
+};
```
thomasheartman, line 60: This is the reason why there's so much refactoring in this PR. We previously aggregated the data within the newproductionflags component. But because we now need to share it between that component and the stat card, we need to extract it and bring it up a level in the tree.

frontend/src/component/insights/componentsChart/batchData.ts
thomasheartman, line 1: The data is never null after all, so I'm removing the extra complexity that I added here previously. Unfortunately, because this changes the return type, this can't be taken out of this PR.

frontend/src/component/insights/hooks/usePlaceholderData.ts
```diff
@@ -9,7 +10,7 @@ type PlaceholderDataOptions = {
 
 export const usePlaceholderData = (
     placeholderDataOptions?: PlaceholderDataOptions,
-) => {
+): ChartData<'line', number[]> => {
```
sjaanus, line 13: Are you sure this type is correct? I see that this data is also piped into `CreationArchiveChart` eventually, which is bar chart, so it feels weird to type this underlying method to line chart.

frontend/src/component/insights/componentsChart/NewProductionFlagsChart/NewProductionFlagsChart.tsx
```diff
@@ -58,112 +47,29 @@ const useOverrideOptions = (chartDataResult: ChartDataResult) => {
...
-        : {
-              datasets: padProjectData(),
-          };
-
-    const notEnoughData = chartDataResult === 'Not Enough Data';
-    const data =
-        notEnoughData || isLoading ? placeholderData : aggregateOrProjectData;
-
-    const overrideOptions = useOverrideOptions(chartDataResult);
+    const key = useId();
```
sjaanus: Maybe we do not need to generate new key every render?

frontend/src/component/insights/hooks/usePlaceholderData.ts
thomasheartman: Good question, but yeah, I'm pretty sure it's correct. If you try to change it to `ChartData<'bar', number[]>`, you get a long error message that ends with 'type "bar" is not assignable to type "line"':

image

But also, the `CreationArchiveChart` doesn't use this data for that reason precisely. Instead, it uses static placeholder data defined in `frontend/src/component/insights/componentsChart/CreationArchiveChart/placeholderData.ts`.

frontend/src/component/insights/componentsChart/NewProductionFlagsChart/NewProductionFlagsChart.tsx
thomasheartman: Yeh, good catch. I don't think we even need the key, actually. 

...
...

Nope, we're good. I'll remove it.

PR #10792, opened 2025-10-14
chore: resolve validator dependency to 13.15.15

gastonfournier:
## About the changes
This pins the validator dependency to one with a vulnerability fix and relaxes pinned dependencies inside resolutions.


gastonfournier:
There's still no fix for the vulnerability report

PR #10794, opened 2025-10-14
fix: hide enable strategy dialog when there are release plans

nunogois:
https://linear.app/unleash/issue/2-3953/fix-enable-environment-modal-when-there-are-release-plans

Prevents the "environment needs at least one enabled strategy" dialog from showing up if you have added a release plan.

Unfortunately we need to propagate `hasReleasePlans` all the way from the DB, which means changes across a bunch of layers.

PS: Maybe we should revisit this dialog in the future. We don't usually see it, and maybe it should trigger the new "add strategy" modal?

image


nunogois:
Closed in favor of https://github.com/Unleash/unleash/pull/10796

PR #10800, opened 2025-10-14
fix: populate all weeks in charts

thomasheartman:
When selecting a date range, the chart now shows you the entire range, even if there's only data for parts of it.

This ensures that both the new flags in production chart and the created vs archived chart show the same period and not two different ones (other charts can be updated later). 

Before:
image

After: 
image

It also ensures that the 4 week batching works the same for both charts.

For line charts, only the dates with data will be rendered as a line. For the bar chart, dates without data can still be hovered and show empty (zero) data.

image

image

Also prevents us from showing the wrong period if there is no data for the selected project

Before:
image

After:
image

Finally, it happily graphs multiple projects at their individual data points in time:

image

### DIFF
--- a/frontend/src/component/insights/componentsChart/CreationArchiveChart/aggregateCreationArchiveData.test.ts
+++ b/frontend/src/component/insights/componentsChart/CreationArchiveChart/aggregateCreationArchiveData.test.ts
@@ -74,7 +74,8 @@ it('combines data from all data sets for a specified week', () => {
         },
     ]);
 });
-it('combines data from all data sets for a specified week', () => {
+
+it('handles scenarios with empty, combined, and single-project data', () => {
     const firstWeek = { week: '2025-11', date: '2025-03-05' };
     const secondWeek = { week: '2025-12', date: '2025-03-12' };
     const thirdWeek = { week: '2025-13', date: '2025-03-19' };
--- a/frontend/src/component/insights/componentsChart/batchData.ts
+++ b/frontend/src/component/insights/componentsChart/batchData.ts
@@ -30,9 +30,8 @@ export const batchData =
                     acc[currentAggregatedIndex] = merge(data, curr);
                 } else if (!hasData && hasCurr) {
                     acc[currentAggregatedIndex] = map(curr);
-                } else if (hasData && !hasCurr) {
-                    // do nothing
                 }
+
                 return acc;
             },
             [] as (TBatched | null)[],
--- a/frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveRatioTooltip.tsx
+++ b/frontend/src/component/insights/componentsChart/CreationArchiveChart/CreationArchiveRatioTooltip.tsx
@@ -2,7 +2,7 @@ import type { FC } from 'react';
 import { Paper, Typography, styled } from '@mui/material';
 import type { TooltipState } from 'component/insights/components/LineChart/ChartTooltip/ChartTooltip';
 import { ChartTooltipContainer } from 'component/insights/components/LineChart/ChartTooltip/ChartTooltip';
-import type { WeekData } from './types.ts';
+import type { FinalizedWeekData } from './types.ts';
 import { calculateRatio } from 'component/insights/calculate-ratio/calculate-ratio.ts';
 
 const StyledTooltipItemContainer = styled(Paper)(({ theme }) => ({
@@ -56,11 +56,13 @@ export const CreationArchiveRatioTooltip: FC<
         return null;
     }
 
-    const rawData = tooltip.dataPoints[0].raw as WeekData;
-    const archivedCount = rawData.archivedFlags || 0;
-    const createdCount = rawData.totalCreatedFlags || 0;
+    const rawData = tooltip.dataPoints[0].raw as FinalizedWeekData;
+    const { archivedFlags, totalCreatedFlags } =
+        rawData.state === 'withRatio'
+            ? rawData
+            : { archivedFlags: 0, totalCreatedFlags: 0 };
 
-    const ratio = calculateRatio(archivedCount, createdCount);
+    const ratio = calculateRatio(archivedFlags, totalCreatedFlags);
 
     return (
         <ChartTooltipContainer tooltip={tooltip}>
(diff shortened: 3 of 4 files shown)


frontend/src/component/insights/componentsChart/CreationArchiveChart/aggregateCreationArchiveData.ts
thomasheartman, line 1: I wanted to extract this function because I was struggling to wrap my head around what was going on. That allowed me to add the tests to make sure that I wasn't breaking anything. Nothing's really changed from its prior implementation, save for typing.

frontend/src/component/insights/componentsChart/CreationArchiveChart/batchCreationArchiveData.ts
thomasheartman, line 1: This batching now handles potentially empty data sets without falling back to adding together zeroes everywhere.

frontend/src/component/insights/componentsChart/CreationArchiveChart/types.ts
thomasheartman, line 1: Most of these types are added to make it easier to keep track of the different states of the data as it goes through the processing. Only Finalized and BatchedWeekData are really used outside of the aggregation, but the discrimanators (`state`) make it much easier to know if you're dealing with a date/batch without entries or with.

frontend/src/component/insights/componentsChart/NewProductionFlagsChart/NewProductionFlagsChart.tsx
```diff
@@ -132,9 +118,40 @@ export const NewProductionFlagsChart: FC<IProjectHealthChartProps> = ({
...
+                    },
+            );
+            return shouldBatchData ? batchProductionFlagsData(padded) : padded;
+        };
+
+        return lifecycleData.datasets.map(({ data, ...dataset }) => ({
+            ...dataset,
+            data: padData(data),
+        }));
+    };
```
thomasheartman, line 147: We also need to pad and potentially batch data if you select specific projects. The process is a little different, but mostly similar to the grouped aggregation.

frontend/src/component/insights/componentsChart/batchData.test.ts
thomasheartman, line 1: Chart.js supports "sparse datasets", where some entries are `null`. This is what allows certain lines to only have entries on specific dates. We need to make sure the aggregation handles that properly.

frontend/src/component/insights/componentsChart/batchData.test.ts
```diff
@@ -13,3 +13,38 @@ it('batches by 4, starting from the first entry', () => {
         ),
     ).toStrictEqual([50, 23]);
 });
+
+describe('null entry handling', () => {
+    it('creates a new entry from the next item, if the accumulated is null', () => {
+        const input = [null, 9];
+        const result = batchData<number, number>({
+            merge: (x, y) => x * y,
```
thomasheartman, line 21: Use multiplication because `*` is treated as `0` in arithmetics (wat).

frontend/src/component/insights/componentsChart/batchData.test.ts
```diff
@@ -13,3 +13,38 @@ it('batches by 4, starting from the first entry', () => {
...
+        })(input);
+        expect(result).toStrictEqual([13]);
+    });
+    it('it returns null if both the accumulated and the next item are null', () => {
+        const input = [null, null];
+        const result = batchData<number, number>({
+            merge: (x, y) => x + y,
+            map: (x) => x,
+        })(input);
+        expect(result).toStrictEqual([null]);
```
thomasheartman, line 40: This would be `0` if we allowed JS to do `null + null`. (Another wat)

frontend/src/hooks/api/getters/useInsights/useInsights.ts
```diff
@@ -1,12 +1,39 @@
...
+        { start: parseISO(start), end: parseISO(end) },
+        {
+            weekStartsOn: 1,
+        },
+    ).map((date) => ({
+        week: format(date, 'RRRR-II'),
+        date: startOfDay(
+            lastDayOfWeek(date, { weekStartsOn: 1 }),
+        ).toISOString(),
+    }));
```
thomasheartman, line 35: This is an easy way to generate the date/week labels for the entire period. I'd like to keep it here in the first instance, because removing this is much easier than changing the API. I'd like to move it to the API later, however.

frontend/src/component/insights/componentsChart/batchData.ts
```diff
@@ -6,21 +6,34 @@ export type BatchDataOptions<T, TBatched> = {
...
-            }
-            return acc;
-        }, [] as TBatched[]);
+                if (!hasData && !hasCurr) {
+                    acc[currentAggregatedIndex] = null;
+                } else if (hasData && hasCurr) {
+                    acc[currentAggregatedIndex] = merge(data, curr);
+                } else if (!hasData && hasCurr) {
+                    acc[currentAggregatedIndex] = map(curr);
+                } else if (hasData && !hasCurr) {
```
sjaanus: I feel like this if clause is not needed.

frontend/src/component/insights/componentsChart/CreationArchiveChart/aggregateCreationArchiveData.test.ts
```diff
@@ -0,0 +1,163 @@
...
+            week: week.week,
+            date: week.date,
+            state: 'withRatio',
+            archivedFlags: 7,
+            totalCreatedFlags: 28,
+            archivePercentage: 25,
+        },
+    ]);
+});
+it('combines data from all data sets for a specified week', () => {
```
sjaanus: There is already a test with this name.

thomasheartman: Thanks, good catch!

frontend/src/component/insights/componentsChart/batchData.ts
thomasheartman: Yeah, fair. I thought it might be nice to leave in to show that we handle it explicitly, but I've removed it now. There's tests checking it anyway.

PR #10805, opened 2025-10-15
chore: add warning about release plans in import-export

nunogois:
https://linear.app/unleash/issue/2-3965/add-a-note-if-were-exporting-that-we-dont-understand-release-plans-in

Adds a warning about release plans in import/export.

It's not trivial to know every flag that will be exported in every scenario, and whether they have release plans, so our logic here is "have you configured release templates?"

image

image


gastonfournier (approved):
Looks good, thanks!

PR #10806, opened 2025-10-15
chore: remove UUID package refactoring code to use node:crypto instead

gastonfournier:
## About the changes
This PR drops the uuid package from node modules and replaces it with standard randomUUID usage that is available from 14.17 onwards, and we have a minimum requirement of node 20 at Unleash.

[Node.js crypto](https://nodejs.org/api/crypto.html#cryptorandomuuidoptions)
[Web crypto](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)

Co-Authored-By: @AnastasiyaHladina

### DIFF
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -3395,13 +3395,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@types/uuid@npm:^9.0.0":
-  version: 9.0.8
-  resolution: "@types/uuid@npm:9.0.8"
-  checksum: 10c0/b411b93054cb1d4361919579ef3508a1f12bf15b5fdd97337d3d351bece6c921b52b6daeef89b62340fd73fd60da407878432a1af777f40648cbe53a01723489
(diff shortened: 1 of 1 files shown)


src/lib/util/random-id.ts
```diff
@@ -1,5 +1,5 @@
-import { v4 as uuidv4 } from 'uuid';
+import { randomUUID } from 'node:crypto';
 
 export const randomId = (): string => {
-    return uuidv4();
+    return randomUUID();
```
sighphyre, line 4: Does this abstraction add anything for us? Less refactor? Seems like an abstraction that adds very little

gastonfournier: It does very little... maybe the idea was to use randomId instead of the specific underlying function, but we're not doing that...

gastonfournier:
Closing in favor of https://github.com/Unleash/unleash/pull/10807 and https://github.com/Unleash/unleash/pull/10808

PR #10807, opened 2025-10-15
chore: remove uuid from the backend

gastonfournier:
Backend only of: https://github.com/Unleash/unleash/pull/10806

## About the changes
This PR drops the uuid package from node modules and replaces it with standard randomUUID usage that is available from 14.17 onwards, and we have a minimum requirement of node 20 at Unleash.

[Node.js crypto](https://nodejs.org/api/crypto.html#cryptorandomuuidoptions)
[Web crypto](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)

Co-Authored-By: @AnastasiyaHladina

### DIFF
--- a/src/lib/util/random-id.ts
+++ b/src/lib/util/random-id.ts
@@ -1,5 +1,5 @@
-import { v4 as uuidv4 } from 'uuid';
+import { randomUUID } from 'node:crypto';
 
 export const randomId = (): string => {
-    return uuidv4();
+    return randomUUID();
 };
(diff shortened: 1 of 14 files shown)


src/lib/util/random-id.ts
```diff
@@ -1,5 +1,5 @@
-import { v4 as uuidv4 } from 'uuid';
+import { randomUUID } from 'node:crypto';
 
 export const randomId = (): string => {
```
sighphyre, line 3: Be cool if we could decide on whether to use this everywhere or nowhere. But maybe a different PR

gastonfournier: We use this function in 201 places but we just use randomUUID directly in some helper methods, so I think we can assume we're using randomId currently

PR #10810, opened 2025-10-15
chore: show edge instances in license info

nunogois:
https://linear.app/unleash/issue/2-3981/show-edge-instances-in-license-information-in-unleash

Show edge instances in license info.

Adapts to new logic, so resources are only shown if they are present.

### DIFF
--- a/frontend/src/component/admin/license/License.tsx
+++ b/frontend/src/component/admin/license/License.tsx
@@ -69,9 +69,9 @@ export const License = () => {
     };
 
     const resources = {
-        Seats: license.resources?.seats,
-        'Release templates': license.resources?.releaseTemplates,
-        'Edge instances': license.resources?.edgeInstances,
+        Seats: license.resources.seats,
+        'Release templates': license.resources.releaseTemplates,
+        'Edge instances': license.resources.edgeInstances,
     };
 
     return (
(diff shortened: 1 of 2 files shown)


frontend/src/component/admin/license/License.tsx
```diff
@@ -111,6 +113,16 @@ export const License = () => {
                                     {license.releaseTemplates}
                                 </StyledPropertyDetails>
                             </StyledDataCollectionPropertyRow>
+                            {enterpriseEdgeEnabled && (
+                                <StyledDataCollectionPropertyRow>
+                                    <StyledPropertyName>
+                                        Edge instances
+                                    </StyledPropertyName>
+                                    <StyledPropertyDetails>
+                                        {license.edgeInstances}
```
gastonfournier: I'm always considering that this can be undefined, if undefined we should not show this section. Not sure if this is taken into account or if we're always going to return zero when undefined in the license.

sighphyre: Definitely agree from a visualization perspective. Not having bought any enterprise edge nodes feels very different from not even knowing what that is

sighphyre (approved):
I'm not the right person to approve this but I'm putting an approve anyway because it looks safe

PR #10815, opened 2025-10-16
refactor: split the stat card widget into its own component

thomasheartman:
Prepares it for being shared between the new flags in production and the archived to created ration sections.

The card in question is this one: 
image


sjaanus (approved):
LG

PR #10838, opened 2025-10-21
chore: show avatars in cross-project change requests table.

thomasheartman:
Updates the CR table cell to show the avatar (again). This applies both to the project-based table and the global table.

Also adds a truncator to the created by cell, to avoid weird name stacking.

Because the truncator already has a popup functionality on truncation, we disable it for the avatars.

To make it compile, I've manually updated the types based on the schemas. I'll regenerate them at a later date. The fields that were marked as nullable are typically not nullable (although often they may not be required) in our schemas, so I'll take it out for now. If we find that we *are* returning null there, it still shouldn't cause any problems for the UI, so it's a low risk fix that can be done later if necessary.

Before:
image

image

After:
image
image

### DIFF
--- a/frontend/src/component/changeRequest/ProjectChangeRequests/ChangeRequestsTabs/AvatarCell.tsx
+++ b/frontend/src/component/changeRequest/ProjectChangeRequests/ChangeRequestsTabs/AvatarCell.tsx
@@ -2,11 +2,9 @@ import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
 import { styled, Typography } from '@mui/material';
 import { useSearchHighlightContext } from 'component/common/Table/SearchHighlightContext/SearchHighlightContext';
 import { Highlighter } from 'component/common/Highlighter/Highlighter';
-import {
-    type AvatarUser,
-    UserAvatar,
-} from 'component/common/UserAvatar/UserAvatar';
+import { UserAvatar } from 'component/common/UserAvatar/UserAvatar';
 import { Truncator } from 'component/common/Truncator/Truncator';
+import type { ComponentProps } from 'react';
 
 const StyledContainer = styled('div')(({ theme }) => ({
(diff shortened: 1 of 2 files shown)


frontend/src/component/changeRequest/ProjectChangeRequests/ChangeRequestsTabs/AvatarCell.tsx
```diff
@@ -2,24 +2,36 @@ import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
...
     display: 'flex',
     flexDirection: 'row',
     alignItems: 'center',
     margin: 0,
+    columnGap: theme.spacing(0.5),
 }));
 
-export const AvatarCell = ({ value }: any) => {
+type AvatarCellProps = {
+    value: AvatarUser;
```
Tymek: There's a way of doing it without export/import - `ComponentProps<typeof ...>`. I don't think we are consistent with that across the codebase.

Tymek (approved):
I assume this isn't an issue on other tables.

frontend/src/component/changeRequest/ProjectChangeRequests/ChangeRequestsTabs/AvatarCell.tsx
thomasheartman: Mmm, yeah, good shout. Seeing as all it needs are the props to the actual avatar, we should be able to just "forward" them, so to speak. I'll have a look.

thomasheartman:
Not sure what you're referencing in terms of it being an "issue" on other tables, but if it's about the name wrapping weirdly, then: 
- this cell is only used in CR tables
- the truncation happens in the cell now, so other tables (project CRs) will also get truncated

This is intentional, however. We'll see how it fares and can change it later if need be.

PR #10839, opened 2025-10-21
chore: add edge instances to instance stats

nunogois:
https://linear.app/unleash/issue/2-3979/add-edge-instances-to-self-reported-instance-stats

Adds edge instances to instance stats.

### DIFF
--- a/frontend/src/openapi/models/instanceAdminStatsSchema.ts
+++ b/frontend/src/openapi/models/instanceAdminStatsSchema.ts
@@ -25,7 +25,7 @@ export interface InstanceAdminStatsSchema {
      * @minimum 0
      */
     contextFields?: number;
-    /** The billable number of edge instances in the last 30, 60 and 90 days */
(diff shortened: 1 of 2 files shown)


src/lib/features/instance-stats/getEdgeInstances.ts
```diff
@@ -0,0 +1,59 @@
+import type { Db } from '../../types/index.js';
+
+const TABLE = 'edge_node_presence';
+const GRACE_PERCENTAGE = 0.05;
```
sighphyre: Gut feel but I don't think we should do this in the db query. The grace factor can be  calculated in billing elsewhere but if we put this here, it ties the grace value to a version of Unleash which feels weird. It's not challenging to tack this on where the bill is actually calculated

nunogois: Yeeted: https://github.com/Unleash/unleash/pull/10839/commits/2f584b9064c983ed2899cfa32986d52a27b3b273

PR #10841, opened 2025-10-22
chore: never wrap other text fields in global change requests table

thomasheartman:
Prevents text wraps in the wrapping columns of the table:
- Title
- environment
- status

Also prevents the change request badge cells (change request badges in tables, such as the project CR table) from wrapping. 

Before:
image

<details>
<summary>In extreme cases</summary>
image
</details>

After:
image

and when it's wide enough to not horizontally scroll:
image

## Note: bug introduced (but fix is coming)

This does introduce a rendering bug in Chrome, however, where the columns will take up all the width that they can without being truncated:

image

This does not happen in Firefox. However, it can be fixed by specifying a max width for the columns. I'll fix this in the next PR.

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
@@ -102,14 +102,12 @@ const ChangeRequestsInner = () => {
 
     const columns = useMemo(
         () => [
-            {
+            columnHelper.accessor('id', {
                 id: 'Title',
                 header: 'Title',
                 meta: { width: '300px' },
-                cell: ({ row }) => (
-                    <GlobalChangeRequestTitleCell {...row.original} />
-                ),
-            },
+                cell: GlobalChangeRequestTitleCell,
+            }),
             columnHelper.accessor('features', {
                 id: 'Updated feature flags',
                 header: 'Updated feature flags',
--- a/frontend/src/component/changeRequest/ChangeRequests/GlobalChangeRequestTitleCell.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequests/GlobalChangeRequestTitleCell.tsx
@@ -4,6 +4,7 @@ import { Link as RouterLink, type LinkProps } from 'react-router-dom';
 import { useProjectOverviewNameOrId } from 'hooks/api/getters/useProjectOverview/useProjectOverview';
 import { Truncator } from 'component/common/Truncator/Truncator';
 import type { ChangeRequestSearchItemSchema } from 'openapi';
+import type { Row } from '@tanstack/react-table';
 
 const LinkContainer = styled('div')(({ theme }) => ({
     color: theme.palette.text.secondary,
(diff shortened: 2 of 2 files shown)


frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -98,16 +98,11 @@ const ChangeRequestsInner = () => {
 
     const columns = useMemo(
         () => [
-            columnHelper.accessor('title', {
```
thomasheartman, line 101: We don't need the title accessor, because we access the whole row anyway.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
```diff
@@ -151,7 +146,9 @@ const ChangeRequestsInner = () => {
                 id: 'Environment',
                 header: 'Environment',
                 meta: { width: '100px' },
-                cell: ({ getValue }) => <HighlightCell value={getValue()} />,
+                cell: ({ getValue }) => (
+                    <HighlightCell maxTitleLines={1} value={getValue()} />
```
thomasheartman, line 152: Don't wrap the environment name. By default, if there is no subtitle, it can span two lines.

frontend/src/component/changeRequest/ProjectChangeRequests/ChangeRequestsTabs/ChangeRequestStatusCell.tsx
```diff
@@ -16,14 +16,13 @@ export const ChangeRequestStatusCell: FC<IChangeRequestStatusCellProps> = ({
     value,
     row: { original },
 }) => {
-    const renderState = () => {
-        if (!value) return null;
-        return <ChangeRequestStatusBadge changeRequest={original} />;
-    };
-
```
thomasheartman, line 23: Not sure why there was a double check for `!value` here previously. Anyway, I simplified it.

frontend/src/component/changeRequest/ProjectChangeRequests/ChangeRequestsTabs/ChangeRequestStatusCell.tsx
```diff
@@ -16,14 +16,13 @@ export const ChangeRequestStatusCell: FC<IChangeRequestStatusCellProps> = ({
...
     if (!value) {
         return <TextCell />;
     }
 
-    return <TextCell>{renderState()}</TextCell>;
+    return (
+        <TextCell sx={{ whiteSpace: 'nowrap' }}>
```
thomasheartman: Don't wrap change request status badges in tables. This applies both to the global change request table and to project change request tables. 

It does not currently apply to status badges outside tables. Before merging, we should figure out whether it should.

frontend/src/component/common/Table/cells/HighlightCell/HighlightCell.tsx
```diff
@@ -68,7 +70,7 @@ export const HighlightCell: FC<IHighlightCellProps> = ({
     return (
         <StyledContainer>
             <Truncator
-                lines={subtitle ? 1 : 2}
+                lines={maxTitleLines ?? (subtitle ? 1 : 2)}
```
thomasheartman, line 73: Being able to pass in a `maxTitleLines` prop made more sense to me than passing in `subtitle=' '`, so I made this override. Happy to discuss it, though.

frontend/src/component/changeRequest/ChangeRequestStatusBadge/ChangeRequestStatusBadge.tsx
```diff
@@ -12,6 +12,11 @@ import type {
     ScheduledChangeRequest,
     UnscheduledChangeRequest,
 } from '../changeRequest.types';
+import { styled } from '@mui/material';
+
+const Badge = styled(MuiBadge)({
+    whiteSpace: 'nowrap',
+});
```
thomasheartman, line 19: Prevent change request badges from wrapping text.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
Tymek: If helper isn't decorating with other stuff, sure. `id` is there so react-table will deal with it

thomasheartman: Yeah, I don't know if it does anything else (the docs weren't very clear), so sure, I'll add it back in.

PR #10850, opened 2025-10-23
feat: include readiness check option

### LINKED ISSUE #10742: Expose a readiness check endpoint to allow validating that Postgres is still up
The readiness check should be an aggregate of your ability to serve traffic and the health of the infra you depend on. When postgres is unavailable on startup, Unleash exits correctly, but if it becomes unavailable after startup, the health check continues to return healthy. 

Have the ability to opt in on a more aggressive readiness check that also validates that the underlying infrastructure on which Unleash depends, is up.

_Originally posted by @msanssouci in https://github.com/Unleash/unleash/discussions/10670#discussioncomment-14605547_

gastonfournier:
## About the changes
This introduces a new endpoint to allow users to check for readiness of Unleash to serve traffic, in particular validating that the DB is up. It's an opt-in feature that has to be enabled with the environment variable `CHECK_DB_ON_READY=true` or via the configuration option `checkDbOnReady`.

Closes #10742

### DIFF
--- a/src/lib/routes/ready-check.ts
+++ b/src/lib/routes/ready-check.ts
@@ -81,9 +81,22 @@ export class ReadyCheckController extends Controller {
         }
 
         try {
-            connection.query({
-                text: 'SELECT 1',
+            await connection.query('BEGIN');
+            await connection.query({
+                text: `SET LOCAL statement_timeout = ${timeoutMs}`,
             });
+            await connection.query('SELECT 1');
+            await connection.query('COMMIT');
+        } catch (error) {
+            try {
+                await connection.query('ROLLBACK');
+            } catch (rollbackError) {
+                this.logger.debug(
+                    'Failed to rollback readiness timeout transaction',
+                    rollbackError,
+                );
+            }
+            throw error;
         } finally {
             if (connection) {
                 await client.releaseConnection(connection);
--- a/src/lib/types/option.ts
+++ b/src/lib/types/option.ts
@@ -302,6 +302,5 @@ export interface IUnleashConfig {
     userInactivityThresholdInDays: number;
     buildDate?: string;
     unleashFrontendToken?: string;
-    /** If true, the readiness endpoint will attempt a simple Postgres query to verify DB availability */
     checkDbOnReady?: boolean;
 }
--- a/src/test/e2e/ready.e2e.test.ts
+++ b/src/test/e2e/ready.e2e.test.ts
@@ -29,6 +29,46 @@ describe('DB is up', () => {
         );
         await request.get('/ready').expect(200).expect('{"health":"GOOD"}');
     });
+
+    test('fails fast when readiness query hangs', async () => {
+        const { request } = await setupAppWithCustomConfig(
+            db.stores,
+            { checkDbOnReady: true },
+            db.rawDatabase,
+        );
+
+        const pool = db.rawDatabase.client.pool;
+        const originalAcquire = pool.acquire.bind(pool);
+
+        pool.acquire = () => {
+            const pending = originalAcquire();
+            pending.promise = pending.promise.then((conn: any) => {
+                const originalQuery = conn.query;
+                conn.query = (queryConfig: any, ...args: any[]) => {
+                    const isSelectOne =
+                        queryConfig?.toUpperCase() === 'SELECT 1';
+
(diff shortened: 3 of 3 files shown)


src/lib/create-config.ts
```diff
@@ -786,6 +786,11 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
         options.prometheusImpactMetricsApi ||
         process.env.PROMETHEUS_IMPACT_METRICS_API;
 
+    const checkDbOnReady =
```
sighphyre, line 789: We can't just nullish coalesce here?

```
const checkDbOnReady =
    options.checkDbOnReady ??
    parseEnvVarBoolean(process.env.CHECK_DB_ON_READY, false);

```

src/lib/routes/health-check.test.ts
```diff
@@ -28,11 +28,11 @@ afterEach(() => {
     getLogger.setMuteError(false);
 });
 
-test('should give 200 when ready', async () => {
+test('should give 200 when healthy', async () => {
```
sighphyre, line 31: Nice change!

src/lib/routes/ready-check.test.ts
```diff
@@ -0,0 +1,44 @@
+import supertest, { type Test } from 'supertest';
```
sighphyre: Be cool to include a test asserting the value when we're not ready too

src/lib/routes/ready-check.ts
```diff
@@ -0,0 +1,62 @@
...
+            middleware: [
+                openApiService.validPath({
+                    tags: ['Operational'],
+                    operationId: 'getReady',
+                    summary: 'Get instance readiness status',
+                    description:
+                        'This operation returns information about whether this Unleash instance is ready to serve requests or not. Typically used by your deployment orchestrator (e.g. Kubernetes, Docker Swarm, Mesos, et al.).',
+                    responses: {
+                        200: createResponseSchema('readyCheckSchema'),
+                        500: createResponseSchema('readyCheckSchema'),
```
sighphyre: 503 maybe? Feels more semantically correct but won't split hairs if you like 500

src/lib/openapi/spec/ready-check-schema.ts
```diff
@@ -0,0 +1,22 @@
+import type { FromSchema } from 'json-schema-to-ts';
+
+export const readyCheckSchema = {
```
sighphyre, line 3: Something feels off to me here. Glancing at our API docs, this will only manifest for 200 responses. We can't ever send "BAD" on a 200 response though, the 500 tag will get it's own schema, which I don't think will include "BAD"

Maybe not an issue, since this is unlikely to be used a lot by open api generators, just feels funny

sighphyre (approved):
Yeah this is cool! Thank you for doing this!

src/lib/routes/ready-check.ts
gastonfournier: 503 seems more accurate :+1:

src/lib/openapi/spec/ready-check-schema.ts
gastonfournier: Yes, good call, I think it's same with the health endpoint. Let me address that

src/lib/routes/ready-check.test.ts
gastonfournier: I'll use e2e cause that'd be simpler and makes this one test pointless

src/lib/create-config.ts
```diff
@@ -786,6 +786,10 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
...
```
sastromskis, line 789: I might be wrong here, but i think that there's no need to have it optional. The way i see it that the application can not be ready if it can't establish the DB connection. 

I understand why you'd want it optional in the healthcheck. This would keep the pods running from their internal state, and they could survive the loss of DB connectivity. But in the ready check i don't see a reason to not do the DB check.

gastonfournier: >  But in the ready check i don't see a reason to not do the DB check.

We know, but this comes at the cost of eating up one of the available connections. 

This change also respects the current behavior: checking readiness against our health check.

PR #10857, opened 2025-10-24
Update the design of change request filters to be one contiguous element

thomasheartman:
Removes spacing between the filters, making them stick together.

This is in prep for adding a new set of filters, which will make each group visually separated.

Also handles the bolding of text that happens when a filter is selected by a little css `::before` trick.
 
Before:
image

After:
image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters.tsx
@@ -1,22 +1,48 @@
-import { Box, Chip, styled } from '@mui/material';
+import { Box, Chip, styled, type ChipProps } from '@mui/material';
 import { useState, type FC } from 'react';
 
-const StyledChip = styled(Chip)(({ theme }) => ({
-    borderRadius: `${theme.shape.borderRadius}px`,
-    padding: theme.spacing(0.5),
-    fontSize: theme.typography.body2.fontSize,
-    height: 'auto',
-    '&[aria-current="true"]': {
-        backgroundColor: theme.palette.secondary.light,
-        fontWeight: 'bold',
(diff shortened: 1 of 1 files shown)


frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters.tsx
```diff
@@ -1,22 +1,48 @@
...
+            '&::before': {
+                content: `'${label}'`,
+                fontWeight: 'bold',
+                visibility: 'hidden',
+                height: 0,
+                display: 'block',
+                overflow: 'hidden',
+                userSelect: 'none',
+            },
+        },
```
thomasheartman, line 44: Add an invisible, unselectable before element (which is also hidden from the a11y tree) with the label of the chip in bold, reserving space for the bold text so it doesn't increase the size of the chip. Also, center the text inside the label to avoid it drifting to the right.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters.tsx
```diff
@@ -1,22 +1,48 @@
...
-        color: theme.palette.primary.main,
-    },
-    ':focus-visible': {
-        outline: `1px solid ${theme.palette.primary.main}`,
-        borderColor: theme.palette.primary.main,
-    },
-}));
+const makeStyledChip = (ariaControlTarget: string) =>
+    styled(({ ...props }: ChipProps) => (
+        <Chip variant='outlined' aria-controls={ariaControlTarget} {...props} />
```
thomasheartman, line 6: extract shared props instead of repeating yourself over and over. Will be more useful when we add the open/closed filter too.

Tymek (approved):

PR #10858, opened 2025-10-24
chore: Add state filter to UI query; default to open

thomasheartman:
Add an open/closed filter to the global change requests table.

To accomodate this, the PR:
- refactors the previous `ChangeRequestFilters` into its own directory with additional files for each of the filter groups. 
- updates the change request filters to work based on the table state instead of its own external state (this was fine with only one param, but would've gotten too complicated with two).

image
image

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/ChangeRequestFilters.styles.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/ChangeRequestFilters.styles.tsx
@@ -28,6 +28,15 @@ export const makeStyledChip = (ariaControlTarget: string) =>
             borderBottomRightRadius: theme.shape.borderRadius,
         },
 
+        '&:not(&[aria-current="true"])': {
+            '&:not(&:first-of-type)': {
+                borderLeftWidth: 0,
+            },
+            '&:not(&:last-of-type)': {
+                borderRightWidth: 0,
+            },
+        },
+
         '& .MuiChip-label': {
             position: 'relative',
             textAlign: 'center',


frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/ChangeRequestFilters.styles.tsx
thomasheartman, line 1: This is extracted directly from the previous changeRequestFilters.tsx component. No changes.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/ChangeRequestFilters.tsx
thomasheartman, line 1: This is the old Filters file once all the individual filter group logic has been moved out. 

Also, instead of keeping its own internal state, it now uses the table's state.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/StateFilterChips.tsx
thomasheartman, line 1: This is the new state filter.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/UserFilterChips.tsx
thomasheartman, line 1: These filter chips have been reworked to work with the table state for determining focus states and setting table states. Otherwise, they're pretty similar to what they were before. Parts of this is from the main `ChangeRequests.tsx` file.

frontend/src/component/changeRequest/ChangeRequests/ChangeRequests.tsx
thomasheartman, line 1: The deleted bits are either no longer needed or have been moved out into their own files or integrated in the filters.

PR #10868, opened 2025-10-27
feat: warning when empty segment

sjaanus:
Decisions made:

1. If there are other constraints and segments, **the warning still appears**, because I still think it's a misconfiguration. Adding an empty segment should not be done unless it’s really necessary, as it can be dangerous.

Screenshot from 2025-10-27 13-57-56


frontend/src/component/releases/ReleasePlanTemplate/TemplateForm/MilestoneStrategy/MilestoneStrategySegmentList.tsx
```diff
@@ -44,6 +50,14 @@ export const MilestoneStrategySegmentList = ({
         return null;
     }
 
+    const hasEmptySegments = segments.some(
```
kwasniew: can be derived from emptySegments variable on line 57

PR #10869, opened 2025-10-27
refactor: milestone progression natural key

kwasniew:
## About the changes

original milestone_progressions table had a synthetic id column with ulids generated by the app. Then we realized we have a natural key: source_milestone. Now that the code is not using the id column anymore we can drop it.


src/migrations/20251027094509-milestone-progression-natural-pk.js
```diff
@@ -0,0 +1,29 @@
+exports.up = function(db, cb) {
+    db.runSql(
+        `
+        DROP INDEX IF EXISTS milestone_progressions_source_milestone_idx;
```
kwasniew, line 4: source milestone had a unique index

src/migrations/20251027094509-milestone-progression-natural-pk.js
```diff
@@ -0,0 +1,27 @@
...
+        ALTER TABLE milestone_progressions DROP COLUMN id;
+
+        ALTER TABLE milestone_progressions ADD PRIMARY KEY (source_milestone);
+        `,
+        cb,
+    );
+};
+
+exports.down = function(db, cb) {
+    db.runSql(
```
kwasniew, line 15: not sure if down migration is worth the effort or if its' better to keep it empty

kwasniew: this is a feature we haven't rolled out

src/migrations/20251027094509-milestone-progression-natural-pk.js
```diff
@@ -0,0 +1,27 @@
+exports.up = function(db, cb) {
+    db.runSql(
+        `
+        DROP INDEX IF EXISTS milestone_progressions_source_milestone_idx;
+
+        ALTER TABLE milestone_progressions DROP COLUMN id;
```
sjaanus, line 6: Is it referenced somewhere as foreign key? If yes, this might cause problems.

src/migrations/20251027094509-milestone-progression-natural-pk.js
```diff
@@ -0,0 +1,27 @@
...
+
+        ALTER TABLE milestone_progressions ADD PRIMARY KEY (source_milestone);
```
sjaanus, line 8: Its fine if all these values are unique right now. Also there should be no nulls.

kwasniew: it has unique not null constraint right now and we use it as PK since the very beginning

src/migrations/20251027094509-milestone-progression-natural-pk.js
kwasniew: we used source_milestone in relations

PR #10870, opened 2025-10-27
chore: increase performance of application upsert

sjaanus:
This is our most longest running database query. Takes most time to execute overall.

Screenshot from 2025-10-27 15-29-33

### DIFF
--- a/src/lib/features/scheduler/schedule-services.ts
+++ b/src/lib/features/scheduler/schedule-services.ts
@@ -89,7 +89,7 @@ export const scheduleServices = (
 
     schedulerService.schedule(
         clientInstanceService.bulkAdd.bind(clientInstanceService),
-        secondsToMilliseconds(10),
+        secondsToMilliseconds(5),
         'bulkAddInstances',
     );
 


src/lib/db/client-applications-store.ts
```diff
@@ -174,11 +174,18 @@ export default class ClientApplicationsStore
...
+            .merge({
+                updated_at: this.db.raw('EXCLUDED.updated_at'),
+                seen_at: this.db.raw('EXCLUDED.seen_at'),
+            });
+
         await this.db(TABLE_USAGE)
             .insert(uniqueUsageRows)
             .onConflict(['app_name', 'project', 'environment'])
-            .merge();
+            .ignore();
```
sjaanus, line 188: Usage is just being added if missing, no value is being merged, so we can skip the merge and just do ignore.

src/lib/db/client-applications-store.ts
```diff
@@ -174,11 +174,18 @@ export default class ClientApplicationsStore
...
         );
 
-        await this.db(TABLE).insert(uniqueRows).onConflict('app_name').merge();
+        await this.db(TABLE)
+            .insert(uniqueRows)
+            .onConflict('app_name')
+            .merge({
+                updated_at: this.db.raw('EXCLUDED.updated_at'),
+                seen_at: this.db.raw('EXCLUDED.seen_at'),
+            });
```
sjaanus, line 183: We do not need to merge the whole object, we can just update the needed columns.

PR #10878, opened 2025-10-28
fix: missing await in user splash update

kwasniew:
## About the changes

Missing await for error handling

### DIFF
--- a/src/lib/services/user-splash-service.ts
+++ b/src/lib/services/user-splash-service.ts
@@ -46,7 +46,7 @@ export default class UserSplashService {
 
     async updateSplash(splash: IUserSplash): Promise<IUserSplash> {
         try {
-            return this.userSplashStore.updateSplash(splash);
+            return await this.userSplashStore.updateSplash(splash);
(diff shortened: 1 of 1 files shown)


src/lib/services/user-splash-service.ts
```diff
@@ -46,7 +46,7 @@ export default class UserSplashService {
 
     async updateSplash(splash: IUserSplash): Promise<IUserSplash> {
         try {
-            return this.userSplashStore.updateSplash(splash);
+            return await this.userSplashStore.updateSplash(splash);
```
gastonfournier, line 49: You don't have to if you're just returning, right?

kwasniew: that'a what I thought but it didn't catch the error without it

kwasniew: i think without try/catch it would be true

kwasniew: i was wondering why my fix from yesterday didn't work and replicated it locally

Issue #9593, opened 2025-03-21
Custom strategies - Disable Create and/or Edit


juliomariano-hotmart:
### Describe the feature request

Hello! 

Considering that custom strategies are deprecated, I would like to have the possibility (as Admin) of disabling the creation of new ones and also the possibility of blocking the editing of existing custom strategies. 

This would allow applications to start using the predefined strategies.

### Background

This feature would help prevent new custom strategies from being created or edited. It would prevent the propagation and expansion of a deprecated feature.

It would help in the future migration process to a Major where custom strategies do not exist.

### Solution suggestions

It could look like the `Maintenance Mode` functionality but for a deprecated functionality, like custom strategies.

- Enable/Disable the creation of new Custom Strategies
- Enable/Disable the editing of new Custom Strategies
- Log as WARN in the applications about the use of custom strategies

gastonfournier:
Hi @juliomariano-hotmart we have a set of permissions targeting directly to custom strategies:
- [CREATE_STRATEGY](https://github.com/Unleash/unleash/blob/d11f39e401e5d2ebe3d12bf3a642803d82d9412e/src/lib/routes/admin-api/strategy.ts#L122)
- [UPDATE_STRATEGY](https://github.com/Unleash/unleash/blob/d11f39e401e5d2ebe3d12bf3a642803d82d9412e/src/lib/routes/admin-api/strategy.ts#L144): this allows you also to deactivate and reactivate strategies
- [DELETE_STRATEGY](https://github.com/Unleash/unleash/blob/d11f39e401e5d2ebe3d12bf3a642803d82d9412e/src/lib/routes/admin-api/strategy.ts#L102)

In the UI they show as follows:

![Image](https://github.com/user-attachments/assets/e5d37505-6312-45c3-8c61-05e761cc5f50)

Let me know if this works for you

gastonfournier:
Hi @juliomariano-hotmart yes, this is part of an enterprise feature: https://docs.getunleash.io/how-to/how-to-create-and-assign-custom-root-roles

We're discussing if it makes sense to allow users to edit the existing pre-defined roles as for this use case it might be useful.

juliomariano-hotmart:
@gastonfournier Is it possible to open this feature to the free version? Since when the feature stops existing it will affect all versions (Free or Enterprise)

Today we use the self-hosted version

gastonfournier:
[...]

PR #10885, opened 2025-10-29
feat: ability to disable custom strategies

gastonfournier:
## About the changes
This adds the ability to disable custom strategy creation and editing by using two env variables: `UNLEASH_DISABLE_CUSTOM_STRATEGY_CREATION` and `UNLEASH_DISABLE_CUSTOM_STRATEGY_EDITING`. 

Fixes: #9593

This could be useful if you want to remove the ability to create new custom strategies and rely on the built-in ones

### DIFF
--- a/src/lib/types/option.ts
+++ b/src/lib/types/option.ts
@@ -180,7 +180,7 @@ export interface IUnleashOptions {
     resourceLimits?: Partial<ResourceLimits>;
     userInactivityThresholdInDays?: number;
     unleashFrontendToken?: string;
-    customStrategySettings?: Partial<ICustomStrategySettings>;
+    customStrategySettings?: ICustomStrategySettings;
     checkDbOnReady?: boolean;
 }
 


gastonfournier:
@copilot can you resolve the conflicts as a senior engineer would do?

src/lib/create-config.ts
```diff
@@ -585,6 +586,29 @@ export function createConfig(options: IUnleashOptions): IUnleashConfig {
...
     ];
 
+    const customStrategyEnv: Partial<ICustomStrategySettings> = {};
+    if (process.env.UNLEASH_DISABLE_CUSTOM_STRATEGY_CREATION !== undefined) {
+        customStrategyEnv.disableCreation = parseEnvVarBoolean(
+            process.env.UNLEASH_DISABLE_CUSTOM_STRATEGY_CREATION,
+            false,
+        );
+    }
+    if (process.env.UNLEASH_DISABLE_CUSTOM_STRATEGY_EDITING !== undefined) {
```
nunogois: I think the whole point of `parseEnvVarBoolean` is that you don't need to check that the env var is `!== undefined`, so this should be simplified.

It also means this would simplify your type and checks, since we would always have these properties set, even if they are false.

nunogois (approved):
LG, but I think it could be simplified.

src/lib/create-config.ts
chriswk: What Nuno said. parseEnvVarBoolean is non-nullable. It will always have a value

gastonfournier: [9f64ed7](https://github.com/Unleash/unleash/pull/10885/commits/9f64ed7fbaff51973a99301220ff2ac36fc88f3b) refactored here

PR #10891, opened 2025-10-29
feat(metrics): prioritize production environments first in FeatureExposureMetrics

mohammedlaniyan-unleash:
## About the changes
This PR updates the `useFeatureMetricsEnvironments` hook inside `FeatureExposureMetrics.tsx` to ensure that production environments are always the default selection.

Previously, environments were displayed in the order they appeared in the feature configuration. This could cause non-production environments to appear first, which wasn’t ideal for metrics visibility.  
Now, the hook identifies the first environment with `type === 'production'` and moves it to the start of the list.

### Why
This improves UX by making sure production metrics are the default selection as most user are mainly interested in the production.

### Important files
- `FeatureExposureMetrics.tsx`
- `FeatureExposureMetrics.test.ts` (added tests for environment ordering)

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.test.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.test.tsx
@@ -1,96 +1,86 @@
-import { vi, beforeEach, describe, it, expect } from 'vitest';
-import type React from 'react';
 import { render } from 'utils/testRenderer';
 import { screen } from '@testing-library/react';
+import { testServerRoute, testServerSetup } from 'utils/testServer';
+import { FeatureExposureMetrics } from './FeatureExposureMetrics.tsx';
+import { Route, Routes } from 'react-router-dom';
 
-type Env = {
-    name: string;
-    type: 'production' | 'development' | 'staging';
-};
-let ENVS: Env[] = [];
-let PROJECT_ID = 'p';
-let FEATURE_ID = 'f';
+describe('FeatureExposureMetrics – default production selection', () => {
+    const server = testServerSetup();
+    const PROJECT_ID = 'default';
+    const FEATURE_ID = 'feature_name';
+    const ROUTE = `/projects/${PROJECT_ID}/features/${FEATURE_ID}/metrics`;
+    const PATH = `/projects/:projectId/features/:featureId/metrics`;
 
-vi.mock('hooks/useRequiredPathParam', () => ({
-    useRequiredPathParam: (k: string) =>
-        k === 'projectId' ? PROJECT_ID : k === 'featureId' ? FEATURE_ID : '',
-}));
-
-vi.mock('hooks/api/getters/useFeature/useFeature', () => ({
-    useFeature: () => ({
-        feature: { name: FEATURE_ID, project: PROJECT_ID, environments: ENVS },
-        loading: false,
-        error: null,
-        refetchFeature: vi.fn(),
-    }),
-}));
-
-vi.mock('hooks/api/getters/useFeatureMetrics/useFeatureMetrics', () => ({
-    useFeatureMetricsRaw: () => ({
-        featureMetrics: [],
-        loading: false,
-        error: null,
-    }),
-}));
-
-import { useFeatureMetricsEnvironments } from './FeatureExposureMetrics.tsx';
-
-const IndexProbe: React.FC = () => {
-    const { environments, defaultProductionIndex: dpi } =
-        useFeatureMetricsEnvironments(PROJECT_ID, FEATURE_ID);
-    return Array.from(environments).map((env, idx) => (
-        <div key={`${env}-${idx}`} data-testid={dpi === idx ? 'dpi' : idx}>
-            {env}
-        </div>
-    ));
-};
-
-const mount = () =>
-    render(<IndexProbe />, {
-        route: `/projects/${PROJECT_ID}/features/${FEATURE_ID}/metrics`,
-    });
-
-beforeEach(() => {
-    PROJECT_ID = 'p';
-    FEATURE_ID = 'f';
-    ENVS = [];
-});
-
-describe('useFeatureMetricsEnvironments', () => {
     it('returns the name of the first production env (prod in middle)', async () => {
-        ENVS = [
-            { name: 'alpha', type: 'development' },
-            { name: 'prod', type: 'production' },
-            { name: 'beta', type: 'development' },
-        ];
-        mount();
-        expect(await screen.findByTestId('dpi')).toHaveTextContent('prod');
+        testServerRoute(
+            server,
+            `/api/admin/projects/${PROJECT_ID}/features/${FEATURE_ID}?variantEnvironments=true`,
+            {
+                environments: [
+                    {
+                        name: 'development',
+                        type: 'development',
+                        sortOrder: 2,
+                    },
+                    {
+                        name: 'production-1',
+                        type: 'production',
+                        sortOrder: 3,
+                    },
+                    {
+                        name: 'production-2',
+                        type: 'production',
+                        sortOrder: 4,
+                    },
+                ],
+            },
+        );
+
+        render(
+            <Routes>
+                <Route path={PATH} element={<FeatureExposureMetrics />} />
+            </Routes>,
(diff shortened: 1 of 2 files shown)


frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.tsx
```diff
@@ -30,7 +30,33 @@ export const FeatureExposureMetrics = () => {
 
     usePageTitle('Metrics');
 
-    const defaultEnvironment = Array.from(environments)[0];
+    const [defaultEnvironment, setDefaultEnvironment] = useState<string>(environments[0])
```
sjaanus: Environments is not array, why `Array.from()` removed?

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.tsx
```diff
@@ -30,7 +30,33 @@ export const FeatureExposureMetrics = () => {
...
+    const { featureMetrics } = useFeatureMetricsRaw(featureId, 48);
```
sjaanus: Why is 48 hardcoded, it was hoursback before.

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.tsx
```diff
@@ -30,7 +30,33 @@ export const FeatureExposureMetrics = () => {
...
+        return Object.entries(totals).reduce((maxEnv, [env, val]) =>
+            val > (totals[maxEnv] ?? -Infinity) ? env : maxEnv,
+            environments?.[0] ?? ''
+        )
+    }, [cachedMetrics, environments]);
+    
+    useEffect(() => {
+        const environment = highestRequest || environments[0];
+        setDefaultEnvironment(environment)
+    }, [highestRequest, environments])
```
sjaanus: On second thought, lets just select first production environment, lets make it simpler. So we do not need to go through metrics payload.

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.tsx
sjaanus, line 1: Can we add a test to validate the new behaviour?

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.tsx
```diff
@@ -171,16 +171,23 @@ export const FeatureExposureMetrics = () => {
...
+    const environments = feature.environments.map((environment, index) => {
+        if (!productionIndex && environment.type === 'production') {
+            productionIndex = index;
+        }
         return environment.name;
     });
 
+    const productionValue = environments.splice(productionIndex, 1)[0];
+    environments.unshift(productionValue);
+
```
sjaanus: This changes the order of environments, which is defined by backend, we do not want to do that. What we want to do instead is that the selected environment is production type. So I think we need to move the logic to the component where the selection happens.

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.test.ts
sjaanus, line 1: For tests I suggest to check how we do UI component based testing. We render the component and check the visuals.

For example check tests `UsersHeader.test.tsx `or `ProductivityEmailSubscription.test.tsx`

sjaanus (changes requested):
Added few comments

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.tsx
```diff
@@ -198,3 +199,16 @@ const useFeatureMetricsApplications = (
 
     return new Set(applications);
 };
+
+export const useDefaultEnvironments = (
+    projectId: string,
+    featureId: string,
+): string => {
+    const { feature } = useFeature(projectId, featureId);
```
kwasniew: we already have useFeatureMetricsEnvironments that calls useFeature. You could change it to return feature name and feature type objects. Then you can create another function that selects the first production type env from the previous hook return type.

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.tsx
```diff
@@ -30,13 +30,14 @@ export const FeatureExposureMetrics = () => {
 
     usePageTitle('Metrics');
 
-    const defaultEnvironment = Array.from(environments)[0];
+    const defaultEnvironment = useDefaultEnvironments(projectId, featureId);
```
sjaanus: When we already have environments here, why are we pulling environments again in the hook?

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.test.ts
```diff
@@ -0,0 +1,119 @@
+import { renderHook } from '@testing-library/react';
+import { useDefaultEnvironments } from './FeatureExposureMetrics.tsx';
+import { vi } from 'vitest';
+import * as useFeatureHook from 'hooks/api/getters/useFeature/useFeature';
+import type { IFeatureEnvironment } from 'interfaces/featureToggle.ts';
+
+describe('useDefaultEnvironments', () => {
```
kwasniew: I'd check other frontend tests to keep testing strategy consistent. we don't have spies in the frontend code tests.  I'd search for import { render } from 'utils/testRenderer'; and use it to test the entire component using the hook you're changing so that it's tested in a full component context. Also I'd check import { testServerRoute, testServerSetup } from 'utils/testServer'; if you need to do any HTTP mocking

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+    }),
+}));
+
+vi.mock('hooks/api/getters/useFeatureMetrics/useFeatureMetrics', () => ({
+    useFeatureMetricsRaw: () => ({
+        featureMetrics: [],
+        loading: false,
+        error: null,
+    }),
+}));
```
sjaanus: As previously stated, please use testServerRoute so we do not need to use `vi.mock.` Lets try to completely get rid of `vi.mock` in this test. https://github.com/Unleash/unleash/pull/10891#discussion_r2480964710

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const IndexProbe: React.FC = () => {
+    const { environments, defaultProductionIndex: dpi } =
+        useFeatureMetricsEnvironments(PROJECT_ID, FEATURE_ID);
+    return Array.from(environments).map((env, idx) => (
+        <div key={`${env}-${idx}`} data-testid={dpi === idx ? 'dpi' : idx}>
+            {env}
+        </div>
+    ));
+};
```
sjaanus: You are currently creating an arbitrary `div` element that you are testing. Please try to render actual `FeatureExposureMetrics` element. This makes the test more real and can catch unexpected corner cases.

frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureExposureMetrics.tsx
```diff
@@ -171,17 +173,21 @@ export const FeatureExposureMetrics = () => {
 
 // Get all the environment names for a feature,
 // not just the one's we have metrics for.
-const useFeatureMetricsEnvironments = (
+export const useFeatureMetricsEnvironments = (
     projectId: string,
     featureId: string,
-): Set<string> => {
+): { environments: Set<string>; defaultProductionIndex: number } => {
```
sjaanus, line 179: Can we avoid inline types?

kwasniew (approved):

sjaanus (approved):

PR #10896, opened 2025-10-30
docs: explain that IP allow list is configured by us

melindafekete:
Docs: clarify that you need to contact support to configure an IP allow list (can't be configured in the Admin UI)

[Support ticket](https://unleash-internal.slack.com/archives/C05QDP90V9D/p1761639039275959)


sighphyre (approved):
LG

PR #10897, opened 2025-10-31
feat: start now button with background

kwasniew:
## About the changes

without progression automations
Screenshot 2025-10-31 at 12 20 01

with progression automations
Screenshot 2025-10-31 at 12 15 52

start now for milestone after the currently running automation will require state re-modelling


sjaanus (approved):
LG

PR #10898, opened 2025-10-31
fix: enterprise edge stats should take into account full month

nunogois:
https://linear.app/unleash/issue/2-3993/fix-enterprise-edge-stats

Fixes Enterprise Edge stats to correctly reflect the average across the whole month.

Now returns a rounded average with 3 decimal places.

Also includes the average of the last 12 months.

### DIFF
--- a/src/lib/features/instance-stats/getEdgeInstances.ts
+++ b/src/lib/features/instance-stats/getEdgeInstances.ts
@@ -28,7 +28,7 @@ export const createGetEdgeInstances =
                         `),
                 ),
             )
-            .with('rng', (qb) =>
+            .with('range', (qb) =>
                 qb
                     .from('months')
                     .select(
@@ -39,9 +39,9 @@ export const createGetEdgeInstances =
             .with('buckets', (qb) =>
                 qb
                     .from(TABLE)
-                    .joinRaw('CROSS JOIN rng')
+                    .joinRaw('CROSS JOIN range')
                     .whereRaw(
-                        'bucket_ts >= rng.min_start AND bucket_ts < rng.max_end',
+                        'bucket_ts >= range.min_start AND bucket_ts < range.max_end',
                     )
                     .groupBy('bucket_ts')
                     .select(


src/lib/features/instance-stats/getEdgeInstances.e2e.test.ts
```diff
@@ -56,77 +117,122 @@ afterAll(async () => {
...
+        monthBeforeLast: expectedPrev,
     });
 });
 
-test('ignores current month data', async () => {
-    const { thisMonthStart, lastMonthStart } = monthWindows();
-    const lMid = atMidMonth(lastMonthStart);
-    const lLate = atLateMonth(lastMonthStart);
-    const tMid = atMidMonth(thisMonthStart);
+test('ignores current month data even when current month is heavy', async () => {
```
sighphyre: Why would current month being heavy affect this?

src/lib/features/instance-stats/getEdgeInstances.e2e.test.ts
```diff
@@ -56,77 +117,122 @@ afterAll(async () => {
...
+});
 
-    await db
-        .rawDatabase(TABLE)
-        .insert([
-            ...rowsForBucket(10, tMid),
-            ...rowsForBucket(2, lMid),
-            ...rowsForBucket(4, lLate),
-        ]);
+test('exact integral average is preserved after rounding', async () => {
```
sighphyre: What does this test guard us against?

src/lib/features/instance-stats/getEdgeInstances.e2e.test.ts
nunogois: Just to guard against window-leak bugs I guess. I can drop this if you prefer.

src/lib/features/instance-stats/getEdgeInstances.e2e.test.ts
nunogois: Just making sure we're actually doing what we say we're doing. Happy to drop this one as well.

frontend/src/openapi/models/instanceAdminStatsSchema.ts
```diff
@@ -25,8 +25,8 @@ export interface InstanceAdminStatsSchema {
      * @minimum 0
      */
     contextFields?: number;
-    /** The rounded up average number of edge instances in the last month and month before last */
-    edgeInstances?: InstanceAdminStatsSchemaEdgeInstances;
+    /** The average number of edge instances, per month, in the last 12 months, rounded to 3 decimal places */
+    edgeInstanceAveragesPerMonth?: InstanceAdminStatsSchemaEdgeInstanceAveragesPerMonth;
```
sighphyre: Damn that's a mouthful of a name. What about EdgeInstanceUsage or something a bit less paragraph-as-a-name?

nunogois: whatsWrongWithMySuperDescriptiveVariableNameThatIsJustATinyBitLong? Yeah, I agree, let's go with your suggestion for now

src/lib/features/instance-stats/getEdgeInstances.ts
```diff
@@ -2,68 +2,83 @@ import type { Db } from '../../types/index.js';
...
+                    ),
+                    db.raw(`
+                            FLOOR(EXTRACT(EPOCH FROM (
+                                (date_trunc('month', NOW()) - ((gs - 1) * INTERVAL '1 month'))
+                                - (date_trunc('month', NOW()) - (gs * INTERVAL '1 month'))
+                            )) / 300)::int AS expected
+                        `),
+                ),
+            )
+            .with('rng', (qb) =>
```
chriswk: nit: small thing here. please use range, not rng (random number generator) :)

I understood it from the context, but my mind went (where's the random? ) :)

chriswk (approved):
This looks good to me, as a selfish BigQuery reporter, it would be good to have current month so far as well. That would be valuable data on the BigQuery side.

nunogois:
> This looks good to me, as a selfish BigQuery reporter, it would be good to have current month so far as well. That would be valuable data on the BigQuery side.

Thank you, I'll add that on a separate PR 👍

PR #10923, opened 2025-11-05
fix: don't spread props + nest li correctly

thomasheartman:
Fixes a few errors appearing on the "assign user/group" on the project settings pages. Namely:
- spreading "key" into props in two places (option list and selected chips)
- incorrect HTML nesting ([`li`'s only permitted parents are `ul`, `ol`,  and `menu`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/li#technical_summary))
- missing key in the list (caused by nesting the `li`)

The renderOption type update is based on a [fix that was added in a more recent version of mui](https://github.com/mui/material-ui/pull/42689/files).

For some reason, the default tag rendering spreads the key in somehow, so I've had to add a manual `renderTags` prop, and as such extracted the `getOptionLabel` function. If you know a better way of sorting out the fact that the `key` prop is spread into the default MUI chips, I'd be very happy to implement that instead.

### DIFF
--- a/frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
+++ b/frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
@@ -4,6 +4,7 @@ import {
     Button,
     capitalize,
     Checkbox,
+    Chip,
     styled,
     TextField,
     Tooltip,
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
```diff
@@ -272,7 +285,7 @@ export const ProjectAccessAssign = ({
     };
 
     const renderOption = (
-        props: React.HTMLAttributes<HTMLLIElement>,
+        { key, ...props }: React.HTMLAttributes<HTMLLIElement> & { key?: any },
```
kwasniew, line 288: is key any or string?

thomasheartman: When we set it, it's always string, I think, but the MUI impl (https://github.com/mui/material-ui/pull/42689/files) uses `any`, so I wanted to be as close to that as possible. Happy to change it though.

PR #10926, opened 2025-11-05
refactor: clean up dead code after removing newStrategyModal flag

nunogois:
https://linear.app/unleash/issue/2-3885/clean-up-legacy-code

Cleans up dead code after removing the `newStrategyModal` feature flag.


daveleek (approved):
Awesome! 🧹

PR #10927, opened 2025-11-05
chore: cache projects response

gastonfournier:
## About the changes
This could backfire if our terraform integration might be expecting to see new data after creating a project... we'd need cache invalidation and that starts to feel like a lot of complexity...

### DIFF
--- a/src/lib/features/project/project-read-model.ts
+++ b/src/lib/features/project/project-read-model.ts
@@ -58,7 +58,7 @@ export class ProjectReadModel implements IProjectReadModel {
         this.db = db;
         this.timer = (action) =>
             metricsHelper.wrapTimer(eventBus, DB_TIME, {
-                store: 'project',
+                store: 'project-read-model',
                 action,
             });
         this.flagResolver = flagResolver;
--- a/src/lib/features/project/project-service.ts
+++ b/src/lib/features/project/project-service.ts
@@ -1,4 +1,4 @@
-import { subDays } from 'date-fns';
+import { subDays, secondsToMilliseconds } from 'date-fns';
 import joi from 'joi';
 const { ValidationError } = joi;
 import createSlug from 'slug';
@@ -86,6 +86,7 @@ import { canGrantProjectRole } from './can-grant-project-role.js';
 import { batchExecute } from '../../util/index.js';
 import metricsHelper from '../../util/metrics-helper.js';
 import { FUNCTION_TIME } from '../../metric-events.js';
+import memoizee from 'memoizee';
 import type { ResourceLimitsService } from '../resource-limits/resource-limits-service.js';
 
 type Days = number;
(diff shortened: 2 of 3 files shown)


src/lib/features/project/project-read-model.ts
```diff
@@ -58,7 +58,7 @@ export class ProjectReadModel implements IProjectReadModel {
         this.db = db;
         this.timer = (action) =>
             metricsHelper.wrapTimer(eventBus, DB_TIME, {
-                store: 'project',
+                store: 'project-read-model',
```
gastonfournier, line 61: Just to differentiate this from the project store that also has a similar query

src/lib/features/project/project-service.ts
```diff
@@ -221,16 +233,35 @@ export default class ProjectService {
...
 
     async getProjects(
         query?: IProjectQuery & IProjectsQuery,
         userId?: number,
     ): Promise<ProjectForUi[]> {
-        const projects = await this.projectReadModel.getProjectsForAdminUi(
-            query,
-            userId,
-        );
+        const projects = await this.getProjectsForAdminUiCached(query, userId);
```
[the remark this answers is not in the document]
gastonfournier: :+1: should be fixed in a0c5b41

kwasniew (commented):
shouldn't we have CachedProjectReadModel that is a drop-in replacement for ProjectReadModel? IMO it would be much more clear than another method in the project service that is already overloaded with capabilities

src/lib/features/project/project-service.ts
```diff
@@ -249,6 +330,27 @@ export default class ProjectService {
         return projects;
     }
 
+    private buildProjectsCacheKey(
```
sighphyre, line 333: Pretty sure this will work but it feels a bit dirty to be using big ass JSON blobs as hash keys, we can't just do this?

```
private buildProjectsCacheKey(
    query?: IProjectQuery & IProjectsQuery,
    userId?: number,
): string {
    const hash = createHash('sha256');
    if (query?.ids) {
        const sorted = [...query.ids].sort();
        hash.update(sorted.join(','));
    }
    hash.update(String(userId ?? ''));
    hash.update(String(query?.id ?? ''));
    hash.update(String(query?.archived ?? ''));
    return hash.digest('base64url');
}
```

gastonfournier:
> shouldn't we have CachedProjectReadModel that is a drop-in replacement for ProjectReadModel? IMO it would be much more clear than another method in the project service that is already overloaded with capabilities

I think the idea is to always return the cached version. It should be smart enough to know when to query the cache or not, but it's still a workaround, to me the problem is having the API overloaded with features that can be different resources. So I prefer to keep it yanky for now.

I'm also tempted to just enable this to the customer having issues

src/lib/features/project/project-service.ts
gastonfournier: Yeah, we could just concatenate strings... hsould be better, will do!

gastonfournier: But I like your approach

PR #10932, opened 2025-11-06
chore: add Change Requests to sidebar and fix command menu icon

thomasheartman:
Adds the new change request overview to the sidebar and fixes the entry in the command bar to use the right title and icon.

image
image

Additionally, standardize the change requests icon. Instead of using a custom one from a svg, let's use the material icon (which is the same, just -90deg rotated and flipped around an axis)

This standardization, makes the icon a tiny bit smaller in the new project modal, but it's the same size (20x20) as the other icons, so I don't think anyone's gonna notice or care, and I think it's better to use the basic material icon.

Of course, we could scale it up (and add a `scale` prop to the change request icon), but I'd rather leave it out for consistency and simplicity for now.

old:
image

new: 
image

As for the scale, I've set it to 1.1, because that makes it damn near identical on the change request page (where I've also updated the raw svg to the new icon) (old one on left, new one on right):
image


kwasniew (approved):
Thomas level attention to details :)

PR #10933, opened 2025-11-06
chore: slightly simpler quick start setup

sighphyre:
Throws together a quick and dirty compose that will spin up postgres and setup the dbs, which dodges the issue that our setup scripts aren't portable

### DIFF
--- a/website/docs/contributing/backend/overview.md
+++ b/website/docs/contributing/backend/overview.md
@@ -30,7 +30,7 @@ To run and develop Unleash, you need to have PostgreSQL 14.0+ locally.
 
 Unleash currently also works with PostgreSQL v14.0+, but this might change in a future feature release, and we have stopped running automatic integration tests below PostgreSQL 14. The current recommendation is to use a role with Owner privileges since Unleash uses Postgres functions to simplify our database usage.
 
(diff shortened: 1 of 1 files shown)


website/docs/contributing/backend/overview.md
```diff
@@ -32,6 +32,18 @@ Unleash currently also works with PostgreSQL v14.0+, but this might change in a
 
 ### Create a local unleash databases in postgres {#create-a-local-unleash-databases-in-postgres}
```
melindafekete:
```suggestion
### Create a local Unleash database in Postgres
```

gastonfournier (approved):
LG! I couldn't test it, but it seems similar to my setup. If something's wrong, we can fix it later.

In general, I'd also add unleash or unleash-enterprise to the docker-compose file so you have all-in-one, but this is a good starting point.

sighphyre:
> LG! I couldn't test it, but it seems similar to my setup. If something's wrong, we can fix it later.
> 
> In general, I'd also add unleash or unleash-enterprise to the docker-compose file so you have all-in-one, but this is a good starting point.

We actually already have a compose for that. This is just a way of getting postgres correctly configured without running a bunch of fragile scripts. The point here is to get a better compile/run loop for folks who don't live their life in this code

PR #10937, opened 2025-11-06
fix: don't clear "implicit" filters when updating table state outside quick filters

thomasheartman:
Fixes bug that would only occur if the first thing you do (when there's no url query params) on the page is to try to change the sort order or change the number of results per page.

In those cases, the table state would be replaced with only the new state from the sorting/page limit (and probably page).

In more specific terms: if you're on the page with no query params, then that means that you're seeing your open change requests.

But if you tried to change the sorting, say, then the "state" and "createdBy" filters would be cleared, and you would end up showing "all change requests ever".

The fix is to spread the implicit table state into the new state before updating the actual state, such that implicit filters become explicit when that happens.

## Implicit filters?

So why do we have implicit filters? Partly aesthetic, partly because that's how it works on github (github.com/pulls), and partly because that makes it easier to share with coworkers. You just need to go to the change requests page and copy the url. With no query params, they'll see their own results instead of yours.

### DIFF
--- a/frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/ChangeRequestFilters.tsx
+++ b/frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/ChangeRequestFilters.tsx
@@ -12,7 +12,7 @@ export const ChangeRequestFilters: FC<ChangeRequestFiltersProps> = ({
     ariaControlTarget,
 }) => {
     const updateTableState = (update: Partial<TableState>) => {
-        setTableState({ ...tableState, ...update, offset: 0 });
+        setTableState({ ...update, offset: 0 });
     };
 
     const StyledChip = makeStyledChip(ariaControlTarget);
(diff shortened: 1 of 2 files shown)


frontend/src/component/changeRequest/ChangeRequests/ChangeRequestFilters/ChangeRequestFilters.tsx
```diff
@@ -12,7 +12,7 @@ export const ChangeRequestFilters: FC<ChangeRequestFiltersProps> = ({
     ariaControlTarget,
 }) => {
     const updateTableState = (update: Partial<TableState>) => {
-        setTableState({ ...tableState, ...update, offset: 0 });
+        setTableState({ ...update, offset: 0 });
```
thomasheartman, line 15: This is handled by the `set` function now. It shouldn't be the filters' responsibility anyway.

PR #10940, opened 2025-11-07
feat: add safeguards to release plans on feature level

sjaanus:
I just found out that this main endpoint does not have OpenAPI validation. So currently not adding schemas, because after I added validation it all broke. Next PR I will check if it is reasonable amount of work to add openapi validation and also move schemas over.

### DIFF
--- a/src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
+++ b/src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
@@ -28,7 +28,6 @@ import type {
 } from '../../../openapi/index.js';
 import { ForbiddenError } from '../../../error/index.js';
 import { beforeAll, afterEach, afterAll, test, describe, expect } from 'vitest';
-import { ulid } from 'ulidx';
 let app: IUnleashTest;
 let db: ITestDb;
 let defaultToken: IApiToken;
@@ -525,7 +524,8 @@ test('Can create feature flag without strategies', async () => {
     const name = 'new.flag.without.strategy.2';
     await app.request
         .post('/api/admin/projects/default/features')
-        .send({ name });
+        .send({ name })
+        .expect(201);
     const { body: flag } = await app.request.get(
         `/api/admin/projects/default/features/${name}`,
     );
@@ -848,7 +848,8 @@ test('Trying to patch variants on a feature flag should trigger an OperationDeni
     const name = 'flag.variants.on.patch';
     await app.request
         .post(url)
-        .send({ name, description: 'some', type: 'release', stale: false });
+        .send({ name, description: 'some', type: 'release', stale: false })
+        .expect(201);
     await app.request
         .patch(`${url}/${name}`)
         .send([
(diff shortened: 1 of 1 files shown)


src/lib/features/feature-toggle/feature-toggle-service.ts
```diff
@@ -1220,10 +1220,24 @@ export class FeatureToggleService {
...
+        return environments.map((env) => {
+            const releasePlans = (
+                releasePlansByEnvironment[env.name] || []
+            ).map((plan) => {
+                if (!safeguardsEnabled && plan.safeguards) {
+                    const { safeguards, ...planWithoutSafeguards } = plan;
+                    return planWithoutSafeguards;
+                }
+                return plan;
+            });
```
sjaanus, line 1234: If safeguards are not enabled, remove safeguards from payload. This will look simpler after flag removal.

src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
```diff
@@ -3544,3 +3547,71 @@ test("Should not be able to create flag with a type that doesn't exist", async (
         })
         .expect(400);
 });
+
+test('GET feature returns release plans with safeguards', async () => {
```
sjaanus: We were missing e2e for release plans also. So adding release plans together with safeguards.

src/lib/features/release-plans/release-plan-read-model.ts
```diff
@@ -153,6 +191,24 @@ export class ReleasePlanReadModel implements IReleasePlanReadModel {
             });
     }
 
+    private getSelectColumns() {
```
sjaanus, line 194: If we are working with json, we need to keep them as raw queeries and cant be just strings.

src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
```diff
@@ -3323,19 +3328,6 @@ test('Enabling a feature environment should add the default strategy when only d
...
-            experimental: {
-                flags: {
-                    strictSchemaValidation: true,
-                    strategyVariant: true,
-                },
-            },
-        },
-        db.rawDatabase,
-    );
-
```
sjaanus: Andreas' legacy. This breaks all my flags that I initially set up.

src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
```diff
@@ -122,6 +125,8 @@ afterEach(async () => {
                 ),
             ),
     );
+    await db.stores.dependentFeaturesStore.deleteAll();
+    await db.stores.featureToggleStore.deleteAll();
```
sjaanus, line 128: Making tests not depend on each other.

src/lib/features/release-plans/release-plan-read-model.ts
```diff
@@ -78,11 +90,37 @@ const processReleasePlanRows = (templateRows): ReleasePlan[] =>
...
+                    safeguards: [],
                     milestones: [],
                 };
                 acc.push(plan);
             }
 
+            if (safeguardId) {
+                const hasSafeguard = (plan.safeguards || []).some(
+                    (sg) => sg.id === safeguardId,
+                );
```
sjaanus, line 102: Follows same pattern as milestones. I think rows are duplicated, so we reduce it back.

src/lib/features/release-plans/release-plan.ts
```diff
@@ -1,5 +1,32 @@
...
+    aggregationMode: 'rps' | 'count' | 'avg' | 'sum' | 'p95' | 'p99' | 'p50';
+    labelSelectors: Record<string, string[]>;
+}
+
+export interface ThresholdCondition {
+    operator: '>' | '<' | '>=' | '<=' | '==' | '!=';
+    threshold: number;
+}
+
+export interface Safeguard {
```
kwasniew, line 23: I wouldn't keep safeguard type in release plans domain but create safeguards domain. Safeguards will potentially span across release plans and strategies

src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
kwasniew: I am pretty sure I added this type of test in the last month. Have to find it

src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
```diff
@@ -3544,3 +3523,71 @@ test("Should not be able to create flag with a type that doesn't exist", async (
         })
         .expect(400);
 });
+
+test('GET feature returns release plans with safeguards', async () => {
+    const featureName = 'feature-with-release-plans';
+    await app.createFeature(featureName);
+
+    const planId = ulid();
+    await db.rawDatabase('release_plan_definitions').insert({
```
kwasniew: are we missing store methods in this repo?

src/lib/features/release-plans/release-plan-read-model.test.ts
```diff
@@ -167,6 +167,28 @@ test('should return release plans with complete milestone data', async () => {
         executedAt: executedAt,
     });
 
+    const impactMetricId = ulid();
+    await db.rawDatabase('impact_metrics').insert({
```
kwasniew, line 171: I'd treat raw inserts as last resort because on migration changes I want to ideally change the corresponding store and/or read models and not tests

src/lib/features/feature-toggle/tests/feature-toggles.e2e.test.ts
kwasniew: ok found it. I tested release plans in milestone-progression.e2e.test.ts in enterprise repo. I prefer to test enterprise features in enterprise repo as it's super easy to setup since all the infra is already there and we don't need to write raw inserts

PR #10951, opened 2025-11-10
fix: allow flag resolver to handle get variant flag enabled better

thomasheartman:
Fixes an edge case in the flag resolver, where if you call `isEnabled` for a variant flag, but the flag returns a disabled (or default) variant (but its feature is enabled), then the flag should still be "enabled" if `feature_enabled` is true.

### DIFF
--- a/src/lib/util/flag-resolver.ts
+++ b/src/lib/util/flag-resolver.ts
@@ -49,7 +49,7 @@ export default class FlagResolver implements IFlagResolver {
         const exp = this.experiments[expName];
         if (exp) {
             if (typeof exp === 'boolean') return exp;
-            else if (exp.enabled) return exp.enabled;
+            else if (exp.enabled || exp.feature_enabled) return true;
         }
         return this.externalResolver.isEnabled(expName, context);
     }


src/lib/util/flag-resolver.ts
```diff
@@ -49,7 +49,7 @@ export default class FlagResolver implements IFlagResolver {
         const exp = this.experiments[expName];
         if (exp) {
             if (typeof exp === 'boolean') return exp;
-            else if (exp.enabled) return exp.enabled;
+            else if (exp.enabled || exp.feature_enabled) return true;
```
kwasniew, line 52: int he experimental.file we have enabled not feature_enabled. I was adding this line here: https://github.com/Unleash/unleash/pull/10607

kwasniew: e.g.  this is one example where we have enabled field only and my change was addressing what we actually do in the experimenta.ts
```
   streaming: {
        name: 'disabled',
        enabled: parseEnvVarBoolean(
            process.env.UNLEASH_EXPERIMENTAL_STREAMING,
            false,
        ),
    },
```

thomasheartman: That's what we do, but the type is:

```ts
export interface Variant {
    name: string;
    enabled: boolean;
    payload?: Payload;
    feature_enabled?: boolean;
}
```

So we still allow you to have a feature_enabled field. We just don't treat it the same way that the regular unleash client does.

PR #10961, opened 2025-11-11
chore: allow custom messages in maintenance mode banner (UI)

thomasheartman:
Updates the maintenance mode banner to accept string variants, allowing for custom maintenance mode messages.

Because the banner is almost the same as the existing banner component we have, we can simplify the impl and just reuse the existing banner instead. The one difference is that the maintenance mode banner used to be taller. However, after talking to UX, we agreed that the banner should be the same size, anyway.

image

### DIFF
--- a/frontend/src/component/maintenance/MaintenanceBanner.tsx
+++ b/frontend/src/component/maintenance/MaintenanceBanner.tsx
@@ -1,22 +1,5 @@
 import { Banner } from 'component/banners/Banner/Banner';
-import useUiConfig from 'hooks/api/getters/useUiConfig/useUiConfig';
-import { PayloadType } from 'utils/variants';
-
-const defaultMessage =
-    '**Maintenance mode!** During maintenance mode, any changes made will not be saved and you may receive an error. We apologize for any inconvenience this may cause.';
-
-const useMaintenanceBannerMessage = (): string => {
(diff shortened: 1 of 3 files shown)


frontend/src/component/maintenance/MaintenanceBanner.tsx
```diff
@@ -1,39 +1,32 @@
...
+    if (typeof flag === 'boolean') {
+        return defaultMessage;
+    }
 
-const MaintenanceBanner = () => {
+    if (flag?.payload?.type === PayloadType.STRING) {
+        return flag.payload.value;
+    }
+    return defaultMessage;
+};
```
thomasheartman: This dance here feels really awkward. Is there a better way to read this? We have `useVariant`, but I'm not sure that's really what we're looking for here 🤔

frontend/src/component/maintenance/useMaintenanceBannerMessage.test.ts
thomasheartman, line 1: It's hard to properly test that the default value is returned when the variant is not of type string, has no payload, or when the flag is just a boolean. This is because you can't `waitFor` the value to change into something other than the default. Claude suggested putting multiple hooks in there to render the useUiConfig hook as well, but that didn't really work and felt like it was getting more complex than it should have been.

PR #10966, opened 2025-11-12
fix: allow external flag resolver to override false experiments with variants in getAll

thomasheartman:
Fixes a bug / uncovered edge case in the flag resolver in Unleash:
If a local experiment was defined as false (the typical default value), then that flag could only ever be returned as a boolean from the `ui-config` endpoint. In other words, even if the external resolver has a variant for that flag, the UI would never get the variant.

The fix is to not just check `isEnabled` for false flags, but instead:
- use `getVariant`
- then check `variant.enabled` (in which case we have a variant and can return it)
- else check `variant.feature_enabled`, falling back to `isEnabled` only if `feature_enabled` is null/undefined.

### DIFF
--- a/src/lib/util/flag-resolver.test.ts
+++ b/src/lib/util/flag-resolver.test.ts
@@ -29,12 +29,11 @@ test('should produce UI flags with extra dynamic flags', () => {
 
 test('should use external resolver for dynamic flags', () => {
     const externalResolver = {
-        isEnabled: (name: string) => {
-            if (name === 'extraFlag') {
-                return true;
-            }
-        },
-        getVariant: () => defaultVariant,
+        isEnabled: (name: string) => name === 'extraFlag',
+        getVariant: (name: string) => ({
+            ...defaultVariant,
+            feature_enabled: name === 'extraFlag',
+        }),
         getStaticContext: () => ({}),
     };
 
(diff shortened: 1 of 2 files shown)


src/lib/util/flag-resolver.test.ts
```diff
@@ -29,12 +29,11 @@ test('should produce UI flags with extra dynamic flags', () => {
...
-            if (name === 'extraFlag') {
-                return true;
-            }
-        },
-        getVariant: () => defaultVariant,
+        isEnabled: (name: string) => name === 'extraFlag',
+        getVariant: (name: string) => ({
+            ...defaultVariant,
+            feature_enabled: name === 'extraFlag',
+        }),
```
thomasheartman, line 36: This change in the existing test seems meaningful because:
- `isEnabled` returns `true`
- but `feature_enabled` in the default variant is `false`

These two should always be the same, which was causing breakages in this test (due to invalid input data). I think this change makes sense and better reflects reality.

thomasheartman:
@FredrikOseberg Close, but not quite! Adding the env parsing that reads it into a variant or a boolean is actually unrelated. That works regardless of this. I didn't realize this was an issue until I tried to hook it up in sandbox, where I realized that the ui config returned the flag as `true`, but didn't return the variant.

So the env parsing isn't actually related to this.

PR #10968, opened 2025-11-12
test: add extra test to ensure we don't return a disabled variant

thomasheartman:
Adds a test to ensure that the `getAll` method of the flag resolver doesn't return the disabled variant if a flag is defined as a boolean in the settings. 

We have some places in the UI where we check `if (uiConfig.flags.<flagname>) {...}`. If one of these flags were suddenly returned as the disabled variant instead of `false`, then it'd be impossible to turn it off. 

As such, to maintain backwards compatibility and adhere to the principle of least surprise, I'd like to add this test to ensure this doesn't change going forward.


FredrikOseberg (approved):
LG

PR #10970, opened 2025-11-12
chore: unique project names validation on creation

nunogois:
https://linear.app/unleash/issue/2-4024/we-should-validate-that-new-project-names-are-unique-ui-only

Validates that new project names must be unique.

Covers both:
 - Creating a new project
 - Editing an existing project

### DIFF
--- a/frontend/src/component/project/Project/hooks/useProjectForm.test.ts
+++ b/frontend/src/component/project/Project/hooks/useProjectForm.test.ts
@@ -122,18 +122,10 @@ describe('name validation', () => {
         ['An empty string', ''],
         ['Just whitespace', '     '],
     ])(`%s is not valid`, (_, value) => {
-        const { result } = renderHook(() => useProjectForm());
+        const { result } = renderHook(() => useProjectForm(undefined, 'valid'));
 
-        act(() => {
-            result.current.setProjectName(value);
-        });
-
-        let isValid = true;
-        act(() => {
-            isValid = result.current.validateName();
-        });
-
-        expect(isValid).toBeFalsy();
+        act(() => result.current.setProjectName(value));
+        expect(result.current.validateName()).toBeFalsy();
     });
 
     test('accepts a unique project name on creation', () => {
@@ -143,17 +135,9 @@ describe('name validation', () => {
 
         const { result } = renderHook(() => useProjectForm());
 
-        act(() => {
-            result.current.setProjectName('Brand New Project');
-        });
+        act(() => result.current.setProjectName('Brand New Project'));
 
-        let isValid = false;
-        act(() => {
-            isValid = result.current.validateName();
-        });
-
-        expect(isValid).toBeTruthy();
-        expect(result.current.errors).not.toHaveProperty('name');
+        expect(result.current.validateName()).toBeTruthy();
     });
 
     test('rejects duplicate project name on creation', () => {
@@ -163,9 +147,7 @@ describe('name validation', () => {
 
         const { result } = renderHook(() => useProjectForm());
 
-        act(() => {
-            result.current.setProjectName('Project One');
-        });
+        act(() => result.current.setProjectName('Project One'));
 
         let isValid = true;
         act(() => {
@@ -179,22 +161,16 @@ describe('name validation', () => {
         );
     });
 
-    test('allows keeping the original name when editing', () => {
+    test('allows keeping the original name when editing, even if that name is also used by a different project', () => {
         mockUseProjects.mockReturnValue(
             createProjectsResponse(existingProjects),
         );
 
         const { result } = renderHook(() =>
-            useProjectForm('project1', 'Project One'),
+            useProjectForm('project3', 'Project One'),
         );
 
-        let isValid = false;
-        act(() => {
-            isValid = result.current.validateName();
-        });
-
-        expect(isValid).toBeTruthy();
-        expect(result.current.errors).not.toHaveProperty('name');
+        expect(result.current.validateName()).toBeTruthy();
     });
 
     test('rejects renaming to another existing project name', () => {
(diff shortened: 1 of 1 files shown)


frontend/src/component/project/Project/hooks/useProjectForm.ts
```diff
@@ -172,6 +175,11 @@ const useProjectForm = (
             return false;
         }
 
+        if (projects.some(({ name }) => name === projectName)) {
+            setErrors((prev) => ({ ...prev, name: 'Name must be unique.' }));
```
thomasheartman: Maybe we could be even clearer here?
```suggestion
            setErrors((prev) => ({ ...prev, name: 'This name is already taken by a different project.' }));
```

frontend/src/component/project/Project/hooks/useProjectForm.ts
```diff
@@ -172,6 +175,11 @@ const useProjectForm = (
             return false;
         }
 
+        if (projects.some(({ name }) => name === projectName)) {
```
thomasheartman: Will this also trigger on re-using the existing name on edit? E.g. if you begin editing project "A", and decide against it, but instead of cancelling, you input "A" and try to save: will it tell you that the name is already taken?

nunogois: I believe this should help: https://github.com/Unleash/unleash/pull/10970/commits/f9c02f28587f721784858c4c6233ac991d82d079

We also refetch projects after submitting so we take into consideration the most up-to-date current project name.

frontend/src/component/project/Project/hooks/useProjectForm.test.ts
```diff
@@ -93,13 +113,134 @@ describe('payload generation', () => {
...
+        act(() => {
+            result.current.setProjectName(value);
+        });
+
+        let isValid = true;
+        act(() => {
+            isValid = result.current.validateName();
+        });
+
+        expect(isValid).toBeFalsy();
```
thomasheartman: It's caught an actual issue here: this test wasn't doing anything without the `act` call. However, it can be 4 lines instead of 11. Adding a valid project name to the initial useProjectForm hook also means that if the name isn't set correctly (as it wasn't before), the test will fail.

```suggestion
        const { result } = renderHook(() => useProjectForm(undefined, 'valid'));

        act(() => result.current.setProjectName(value));
        expect(result.current.validateName()).toBeFalsy();
```

frontend/src/component/project/Project/hooks/useProjectForm.test.ts
```diff
@@ -93,13 +113,134 @@ describe('payload generation', () => {
...
+            result.current.setProjectName('Brand New Project');
+        });
+
+        let isValid = false;
+        act(() => {
+            isValid = result.current.validateName();
+        });
+
+        expect(isValid).toBeTruthy();
+        expect(result.current.errors).not.toHaveProperty('name');
```
thomasheartman: Similarly, we should be able to make this more concise. I'm also not sure that asserting on the missing `name` property has any point to it?

```suggestion
        act(() => result.current.setProjectName('Brand New Project'));

        expect(result.current.validateName()).toBeTruthy();
```

frontend/src/component/project/Project/hooks/useProjectForm.test.ts
```diff
@@ -93,13 +113,134 @@ describe('payload generation', () => {
...
+        let isValid = true;
+        act(() => {
+            isValid = result.current.validateName();
+        });
+
+        expect(isValid).toBeFalsy();
+        expect(result.current.errors).toHaveProperty(
+            'name',
+            'This name is already taken by a different project.',
+        );
```
thomasheartman: And here:

```suggestion
        act(() => result.current.setProjectName('Project One'));

        expect(result.current.validateName()).toBeFalsy();
        expect(result.current.errors).toHaveProperty(
            'name',
            'This name is already taken by a different project.',
        );
```

I also don't like asserting on exact strings, but eh.

frontend/src/component/project/Project/hooks/useProjectForm.test.ts
```diff
@@ -93,13 +113,134 @@ describe('payload generation', () => {
...
+
+        let isValid = false;
+        act(() => {
+            isValid = result.current.validateName();
+        });
+
+        expect(isValid).toBeTruthy();
+        expect(result.current.errors).not.toHaveProperty('name');
+    });
```
thomasheartman: I'm not sure what this test is doing, tbh. I think it's more important to check that you can keep the current name, even if it's already taken by a different project. E.g. something like:

```suggestion
    test('allows keeping the original name when editing, even if that name is also used by a different project', () => {
        mockUseProjects.mockReturnValue(
            createProjectsResponse(existingProjects),
        );

        const { result } = renderHook(() =>
            useProjectForm('project3', 'Project One'),
        );

        expect(result.current.validateName()).toBeTruthy();
    });
```

frontend/src/component/project/Project/hooks/useProjectForm.test.ts
```diff
@@ -93,13 +113,134 @@ describe('payload generation', () => {
...
+        act(() => {
+            result.current.setProjectName('Project Two');
+        });
+
+        let isValid = true;
+        act(() => {
+            isValid = result.current.validateName();
+        });
+
+        expect(isValid).toBeFalsy();
```
thomasheartman:
```suggestion
        act(() => result.current.setProjectName('Project Two'));

        expect(result.current.validateName()).toBeFalsy();
```

frontend/src/component/project/Project/hooks/useProjectForm.test.ts
```diff
@@ -93,13 +113,134 @@ describe('payload generation', () => {
...
+            result.current.setProjectName('Project Three');
+        });
+
+        let isValid = false;
+        act(() => {
+            isValid = result.current.validateName();
+        });
+
+        expect(isValid).toBeTruthy();
+        expect(result.current.errors).not.toHaveProperty('name');
```
thomasheartman:
```suggestion
        act(() => result.current.setProjectName('Project Three'));

        expect(result.current.validateName()).toBeTruthy();
```

thomasheartman (approved):

PR #10971, opened 2025-11-12
chore: use proper express types

gastonfournier:
This PR adheres to express param types and query types whose values are always strings

### DIFF
--- a/src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
+++ b/src/lib/features/client-feature-toggles/client-feature-toggle.controller.ts
@@ -165,7 +165,7 @@ export default class FeatureController extends Controller {
     private async resolveFeaturesAndSegments(
         query?: IFeatureToggleQuery,
     ): Promise<[FeatureConfigurationClient[], IClientSegment[]]> {
-        if (this.flagResolver.isEnabled('deltaApi')) {
+        if (this.flagResolver.isEnabled('deltaDiff')) {
             const features =
                 await this.clientFeatureToggleService.getClientFeatures(query);
 


src/lib/routes/unleash-types.ts
```diff
@@ -15,10 +16,10 @@ export interface IAuthRequest<
...
 
 export interface IApiRequest<
-    PARAM = any,
-    ResBody = any,
-    ReqBody = any,
-    ReqQuery = any,
+    PARAM extends ParamsDictionary = ParamsDictionary,
+    ResBody = any, // TODO: consider `unknown` later to enforce explicit typing of res bodies
+    ReqBody = any, // TODO: consider `unknown` later to force narrowing/validation of req bodies
+    ReqQuery extends Query = Query,
```
gastonfournier, line 22:
```suggestion
    PARAM extends ParamsDictionary | unknown = unknown,
    ResBody = unknown,
    ReqBody = unknown,
    ReqQuery extends Query | unknown = unknown,
```
This would force us to declare the type before using params, body, or query params. The challenge is that any param is optional, despite not being optional in the signature, which makes this a bit tricky.

As an example:  `req: IAuthRequest<{ id: string }>` means that we expect an id, but it can also be undefined if the id key is not provided in the params dictionary. So it ends up being as good as  `req: IAuthRequest<Record<string, string>>,`

gastonfournier:
Currently, if we annotate our endpoints with parameters, we can get the types validated with the schema. So this means the PR is less valuable than adding the parameters everywhere.

gastonfournier:
> If you annotate the endpoint with the parameters property in openApiService.validPath, then types are forced. 

@thomasheartman quote :) 

That is more valuable than having all strings, so I'm closing this

PR #10972, opened 2025-11-12
feat: add safeguards schema to oss

sjaanus:
TypeScript throws `TS7056` because the schema object becomes too large for the compiler to fully serialize when using deep literal inference. 
Splitting the components object and explicitly reconstructing the type prevents the error while preserving correct type inference.

### DIFF
--- a/src/lib/openapi/spec/import-toggles-schema.ts
+++ b/src/lib/openapi/spec/import-toggles-schema.ts
@@ -28,6 +28,8 @@ import { safeguardSchema } from './safeguard-schema.js';
 import { metricQuerySchema } from './metric-query-schema.js';
 import { safeguardTriggerConditionSchema } from './safeguard-trigger-condition-schema.js';
 
+// error TS7056: The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.
+// We needed to do `as any` and then later recreate the type to bypass the maximum length serialization error.
 const importTogglesSchemaComponents = {
     schemas: {
         exportResultSchema,


src/lib/openapi/spec/import-toggles-schema.ts
```diff
@@ -24,6 +24,42 @@ import { featureDependenciesSchema } from './feature-dependencies-schema.js';
 import { dependentFeatureSchema } from './dependent-feature-schema.js';
 import { featureLinksSchema } from './feature-links-schema.js';
 import { featureLinkSchema } from './feature-link-schema.js';
+import { safeguardSchema } from './safeguard-schema.js';
+import { metricQuerySchema } from './metric-query-schema.js';
+import { safeguardTriggerConditionSchema } from './safeguard-trigger-condition-schema.js';
+
+const importTogglesSchemaComponents = {
```
sjaanus, line 33: I was getting `TS7056: The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.` So needed to bring this out as const

src/lib/openapi/spec/metric-query-schema.ts
```diff
@@ -0,0 +1,48 @@
+import type { FromSchema } from 'json-schema-to-ts';
+
+export const metricQuerySchema = {
```
sjaanus, line 3: This is direct copy paste from enterprise repo. I will be deleting from enterprise later.

src/lib/openapi/spec/safeguard-trigger-condition-schema.ts
```diff
@@ -0,0 +1,27 @@
+import type { FromSchema } from 'json-schema-to-ts';
+
+export const safeguardTriggerConditionSchema = {
```
sjaanus, line 3: This is also copy paste of enterprise. Will use in enterprise.

src/lib/openapi/spec/safeguard-schema.ts
```diff
@@ -0,0 +1,49 @@
+import type { FromSchema } from 'json-schema-to-ts';
+import { metricQuerySchema } from './metric-query-schema.js';
+import { safeguardTriggerConditionSchema } from './safeguard-trigger-condition-schema.js';
+
+export const safeguardSchema = {
```
sjaanus, line 5: This is moved from enterprise.

src/lib/openapi/spec/import-toggles-schema.ts
```diff
@@ -49,35 +85,44 @@ export const importTogglesSchema = {
...
-            featureLinkSchema,
-        },
-    },
+    components: importTogglesSchemaComponents,
 } as const;
 
-export type ImportTogglesSchema = FromSchema<typeof importTogglesSchema>;
+export const importTogglesSchema = _importTogglesSchema;
+
+export type ImportTogglesSchema = FromSchema<
```
sjaanus, line 93: Reconstruct the schema for `FromSchema`

src/lib/openapi/spec/import-toggles-schema.ts
```diff
@@ -24,6 +24,42 @@ import { featureDependenciesSchema } from './feature-dependencies-schema.js';
...
+        tagTypeSchema,
+        featureDependenciesSchema,
+        dependentFeatureSchema,
+        featureLinksSchema,
+        featureLinkSchema,
+        safeguardSchema,
+        metricQuerySchema,
+        safeguardTriggerConditionSchema,
+    },
+} as const as any;
```
kwasniew, line 64: what is the purpose of as const if you do as any afterwards?

src/lib/openapi/spec/import-toggles-schema.ts
kwasniew: this may be difficult to maintain in future (remembering to update this file)

src/lib/openapi/spec/import-toggles-schema.ts
sjaanus: Yes that can be skipped, but we need any :/

kwasniew (approved):

PR #10987, opened 2025-11-17
fix(frontend): use extendedUsageMetrics flag in metrics UI

daveleek:
Reuse the extendedUsageMetrics flag in the UI to show the day options only when enterprise customer has the flag set

### DIFF
--- a/frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureMetricsHours/FeatureMetricsHours.test.tsx
+++ b/frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureMetricsHours/FeatureMetricsHours.test.tsx
@@ -11,6 +11,9 @@ test('Display extended daily metrics', async () => {
         versionInfo: {
             current: { oss: 'irrelevant', enterprise: 'some value' },
         },
+        flags: {
+            extendedUsageMetrics: true,
+        },
     });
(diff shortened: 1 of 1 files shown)


frontend/src/component/feature/FeatureView/FeatureMetrics/FeatureMetricsHours/FeatureMetricsHours.tsx
```diff
@@ -30,9 +31,11 @@ export const FeatureMetricsHours = ({
         });
     };
     const { isEnterprise } = useUiConfig();
-    const options = isEnterprise()
-        ? [...hourOptions, ...daysOptions]
-        : hourOptions;
+    const extendedUsageMetricsEnabled = useUiFlag('extendedUsageMetrics');
+    const options =
+        isEnterprise() && extendedUsageMetricsEnabled
```
kwasniew, line 36: enterprise customers should still see 30 days even without extended usage metrics. extended is for 90 days

daveleek: Recap: After some testing and validation we realized that anything > 48h is ignored backend when flag isn't set, so we'll go ahead with this PR as is

PR #10995, opened 2025-11-18
chore: show hosting in connected edges

nunogois:
https://linear.app/unleash/issue/2-4037/show-hosting-in-connected-edges-edge-observability

Show "hosting" in Connected Edges.

This can be one of:
 - Cloud
 - Self-hosted
 - Unknown

image

### DIFF
--- a/frontend/src/component/admin/network/NetworkConnectedEdges/NetworkConnectedEdgeInstance.tsx
+++ b/frontend/src/component/admin/network/NetworkConnectedEdges/NetworkConnectedEdgeInstance.tsx
@@ -82,14 +82,16 @@ const StyledBadge = styled(Badge)(({ theme }) => ({
     padding: theme.spacing(0, 1),
 }));
 
-const getHosting = ({ hosting }: ConnectedEdge) => {
+const getHosting = ({
+    hosting,
+}: ConnectedEdge): 'Cloud' | 'Self-hosted' | 'Unknown' => {
     switch (hosting) {
         case 'hosted':
             return 'Cloud';
         case 'enterprise-self-hosted':
             return 'Self-hosted';
         default:
-            return 'Unknown';
+            return hosting ? `Unknown: ${hosting}` : 'Unknown';
     }
 };
 


frontend/src/component/admin/network/NetworkConnectedEdges/NetworkConnectedEdgeInstance.tsx
```diff
@@ -82,6 +82,17 @@ const StyledBadge = styled(Badge)(({ theme }) => ({
...
 }));
 
+const getHosting = ({ hosting }: ConnectedEdge) => {
+    switch (hosting) {
+        case 'hosted':
+            return 'Cloud';
+        case 'enterprise-self-hosted':
+            return 'Self-hosted';
+        default:
+            return 'Unknown';
```
gastonfournier: I hope this is not a valid use case :D but maybe this will lead customers to contact us which could be a good thing

nunogois: It shouldn't be, but I added it as a fallback just in case. Maybe I was influenced by my recent experience in Rust where we must cover all cases.

frontend/src/component/admin/network/NetworkConnectedEdges/NetworkConnectedEdgeInstance.tsx
```diff
@@ -82,6 +82,17 @@ const StyledBadge = styled(Badge)(({ theme }) => ({
     padding: theme.spacing(0, 1),
 }));
 
+const getHosting = ({ hosting }: ConnectedEdge) => {
```
sighphyre: Probably fine but if we do this

``` js
const getHosting = ({ hosting }: ConnectedEdge): string => {
    switch (hosting) {
        case 'hosted':
            return 'Cloud';
        case 'enterprise-self-hosted':
            return 'Self-hosted';
        case undefined:
            return 'Unknown';
    }

    const _exhaustive: never = hosting;
    return _exhaustive;
};

```

Later when someone patches this type, they get warned by the compiler that they need to make a decision here rather than suffering what you and I suffered - having a bunch of places quietly fail in the wrong ways without warning

Yes, I'm a ho for exhaustive matching, why do you ask?

frontend/src/component/admin/network/NetworkConnectedEdges/NetworkConnectedEdgeInstance.tsx
sighphyre: It's one of those "this should never happen" cases. In theory there's a third variant here "self-hosted" but we filter that out on the API level. There's also older Edges which report nothing so undefined is a possibility. We do also filter that out though.

So... should never happen unless we gone done goofed somewhere haha

gastonfournier: Agreed! I was just thinking the text could be "Unknown value, please contact us" :P half-joking half-serious

frontend/src/component/admin/network/NetworkConnectedEdges/NetworkConnectedEdgeInstance.tsx
nunogois: Huh interesting. I think we don't want the `string` return type though, since that loosens our return type from `'Cloud' | 'Self-hosted' | 'Unknown'` to any `string`.

Speaking of that, would this be a more natural way of accomplishing the same thing?

```ts
const getHosting = ({
    hosting,
}: ConnectedEdge): 'Cloud' | 'Self-hosted' | 'Unknown' => {
    switch (hosting) {
        case 'hosted':
            return 'Cloud';
        case 'enterprise-self-hosted':
            return 'Self-hosted';
        default:
            return hosting ? `Unknown: ${hosting}` : 'Unknown';
    }
};
```

